PR #3303, opened 2022-03-13
DBZ-4832 Don't set truststore/keystore parameters to system variables

harveyyue:
https://issues.redhat.com/browse/DBZ-4832

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnection.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnection.java
@@ -6,7 +6,6 @@
 
 package io.debezium.connector.mysql;
 
-import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.time.Duration;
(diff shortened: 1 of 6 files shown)


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnection.java
```diff
@@ -575,6 +580,22 @@ public boolean sslModeEnabled() {
             return sslMode() != SecureConnectionMode.DISABLED;
         }
 
+        public String sslKeyStore() {
+            return config.getString(MySqlConnectorConfig.SSL_KEYSTORE);
+        }
+
+        public String sslKeyStorePassword() {
```
gunnarmorling, line 587: Let's return it as `char[]` right here (same for trust store PW).

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnection.java
```diff
@@ -502,7 +492,22 @@ public MySqlConnectionConfiguration(Configuration config) {
             final Builder jdbcConfigBuilder = dbConfig
                     .edit()
                     .with("connectTimeout", Long.toString(getConnectionTimeout().toMillis()))
-                    .with("useSSL", Boolean.toString(useSSL));
+                    .with("sslMode", sslMode().name());
```
gunnarmorling, line 495:
```suggestion
                    .with("sslMode", sslMode().getValue());
```

I.e. let's not rely on the enum constant name, but the assigned value.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/legacy/MySqlJdbcContext.java
```diff
@@ -73,7 +73,22 @@ public MySqlJdbcContext(MySqlConnectorConfig config) {
         Builder jdbcConfigBuilder = jdbcConfig
                 .edit()
                 .with("connectTimeout", Long.toString(config.getConnectionTimeout().toMillis()))
-                .with("useSSL", Boolean.toString(useSSL));
+                .with("sslMode", sslMode().name());
```
gunnarmorling, line 76: Same as above.

gunnarmorling (commented):
LGTM overall, some minor comments inline. And I'll log a follow-up issue to get rid of the duplications between `MySqlJdbcContext` and `MySqlConnection`.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/legacy/MySqlJdbcContext.java
harveyyue: fix these issue.

gunnarmorling:
LGTM, applied. Thanks a lot, @harveyyue!

PR #3304, opened 2022-03-14
DBZ-4850 Add several retriable SQL error codes

EthanZ328:
(ORA-00310,ORA-01343,ORA-01371)

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -105,6 +105,7 @@ Eric Slep
 Eric Weaver
 Eric S. Kreiseir
 Erik Malm
+Ethan Zou
 Ewen Cheslack-Postava
 Fabian Aussems
 Fabian Martinez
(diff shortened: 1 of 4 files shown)


gunnarmorling:
Hey @Naros, could you review and merge this one? Thx!

Naros:
Hi @EthanZ328 so I'm not really comfortable adding these as "retriable" errors because really we should never face these if we are managing logs properly.  So before, I took a look at your log snippets and I saw something quite interesting in the snippet related to the data dictioanry:

```
Missing log file(s) for thread 2, sequence(s) 64055 to 64056
```

So I noticed we did add `64055` and `64056` to the LogMiner session, but these were only for Thread 1 and not Thread 2 is likely why these errors happened in the first place.  The logs that were added for Thread 1 were specifically ONLINE redo logs while the others were ARCHIVE logs.  This leads me to think there may be an issue with the redo log gathering query when multiple threads are in-play.  I'll take a look.

EthanZ328:
@Naros - I found that when we check if log switch occurs, we may misunderstand the sequence when there're more than one threads. It may cause many unnecessary logminer behaviors. I updated code accordingly. But I'm not sure if it is root cause of the Exception yet. Please check.

Naros:
Hi @EthanZ328 so yes I agree that the comparison check can be faulty in an Oracle RAC situation where you have multiple redo threads unlike in a standalone environment.  I think a better solution however would be to adjust the actual underlying SQL inside `SqlUtils#currentRedoLogSequenceQuery`.  There instead you would apply something like `ORDER BY SEQUENCE#` and we can avoid the `Collections.sort()` call.  What do you think?

EthanZ328:
@Naros - Make sense.

Naros (approved):
LGTM, thanks @EthanZ328

PR #3305, opened 2022-03-14
Update documentation to reflect mysql 8 changes

josetesan:
According to https://dev.mysql.com/doc/refman/5.7/en/information-schema-variables-table.html, information_schema.global_variables has been deprecated ( and removed on 8.x ) in favour of performance_schema, so this documentation reflects the new way to achieve this


josetesan:
This doc change affects dbz 1.6, 1.7 and 1.8 branches.

gunnarmorling (approved):
Good spot, thx!

gunnarmorling:
I've merged this to main (=1.9) and 1.8; we don't maintain earlier versions of Debezium any longer. Thanks again.

PR #3306, opened 2022-03-15
DBZ-4436 Restructure upstream steps for obtaining JDBC driver

roldanbob:
[DBZ-4436](https://issues.redhat.com/browse/DBZ-4436)

This change updates the upstream Oracle connector deployment documentation to provide separate instructions for obtaining the OJDBC driver for use with a standard Oracle database and Oracle XStream, and points to Maven Central vs. the Oracle website as the download source for the standard JAR file.
Information related to the XStream deployment is moved to the XStream section of the documentation.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/oracle.adoc
+++ b/documentation/modules/ROOT/pages/connectors/oracle.adoc
@@ -1792,13 +1792,13 @@ To deploy a {prodname} Oracle connector, you install the {prodname} Oracle conne
 .Prerequisites
 * link:https://zookeeper.apache.org/[Apache ZooKeeper], link:http://kafka.apache.org/[Apache Kafka], and link:{link-kafka-docs}.html#connect[Kafka Connect] are installed.
 * Oracle Database is installed, and is xref:{link-oracle-connector}#setting-up-oracle[configured to work with the {prodname} connector].
-* You have a copies of the Oracle JDBC driver and the XStream API JAR.
+* You have a copy of the Oracle JDBC driver.
+  For more information, see xref:obtaining-the-oracle-jdbc-driver[Obtaining the Oracle JDBC driver].
 +
 [IMPORTANT]
 ====
-Due to licensing requirements, the {prodname} Oracle connector does not ship with the Oracle JDBC driver or XStream API files.
-You must download these files directly from Oracle and add them to your environment.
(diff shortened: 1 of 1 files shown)


documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -3566,6 +3525,50 @@ The following configuration example adds the properties `database.connection.ada
...
+...
+
+├── ojdbc8.jar
+├── ucp.jar
+├── uidrvci
+└── xstreams.jar
+
+----
+
+. Copy the `ojdbc8.jar` and `xstreams.jar` files, and add them to the `_<kafka_home>_/libs` directory, for example, `kafka/libs`.
```
gunnarmorling, line 3564: @Naros, is there a reason for copying that into kafka/libs rather the connector directory?

roldanbob: @gunnarmorling @Naros Let me know if that should be changed. I brought that path forward from the current version of the docs.

gunnarmorling: Right, I saw that. And we could go with that, but I'm curious about whether that's truly necessary. Waiting to hear back from @Naros.

Naros: @gunnarmorling @roldanbob there is no specific reason; both are equally fine.  

The only thing that comes to mind is we may have chosen `kafka/libs` since its a static directory reference and would be universal across any type of Kafka Connect deployment whereas user-specific installations may actually specify the connector's plugin directory differently from our images versus others such as those from Confluent Hub.

FWIW, I did check the entire documentation and we do make the same call out in the Db2 documentation as well.

roldanbob: So, bottom line, we can leave that as-is? 

Note that there will be further changes to the content change in this PR based on [DBZ-4392](https://issues.redhat.com/browse/DBZ-4392).

gunnarmorling: Ok, logged https://issues.redhat.com/browse/DBZ-4883 as a follow-up and assigned it to you, @roldanbob. It describes some more changes which arguable could have been done as part of this PR already, but given it represents an improvement by itself, I went ahead and merged it, hoping we can get in those suggested follow-up changes soon, too. Thx!

PR #2946, opened 2021-11-17
DBZ-4478: Partition-scoped metrics for the SQL Server connector

morozov:
### Change summary

Introducing an absolutely separate API for multi-partition metrics would complicate a lot of code, so I decided to find a common ground. Below are the key steps.

#### Introduce interfaces and default implementations for change event source metrics 

See the details in [#3047](https://github.com/debezium/debezium/pull/3047).

### Add `P extends Partition` parameter to `io.debezium.pipeline.source.spi.*Listener` interfaces

All existing connectors are already aware of the `Partition` interface. Passing partition to the listeners won't require any fundamental rework of the existing connectors but at the same time will enable building single- and multi-partition metrics using the same API.

Technically, the `Partition` interface contains all the properties needed for exposing partition-scoped metrics but the interface stores them in a map without defining semantics for any of the properties. Since from the metrics perspective each property (e.g. server, database) has its own semantics, it's better to accept a parameterized type.

### Implement multi-partition metrics for SQL Server connector

In the single-partition mode, the SQL Server connector will use the default implementations of the metrics. This will provide a backward compatibility layer without any additional code.

In the multip-partition mode, it will use a separate metrics factory which will initialize multi-partition metrics. The metrics will have the following hierarchy:

1. The `Metrics` classes will act as multi-partition listeners and implement task-level MBean interfaces.
2. Internally, each metric will maintain a collection of task-scoped objects (currently called meters but we may think of a better name). Each object is in fact a group of atomic references which implements a single-partition listener API and a partition-level MBean.
3. When registering or unregistering, each `Metrics` class registers or unregisters itself and all internal objects.
4. Now, the default metrics can (and have been) reimplemented to delegate their implementation to the meter objects. This way, the same implementation is shared between the single- and multi-partition metrics.
5. As an intended side-effect, the integration tests that use a multi-partition configuration should use `TestHelper.waitForDatabaseSnapshotToBeCompleted(TestHelper.TEST_DATABASE)` to synchronize with the connector.

### Future scope

1. The existing single-partition and multi-partition `Snapshot` and `Streaming` MXBean interfaces can be replaced with combinations of smaller "meter" interfaces introduced above. For instance:
    * Single-partition streaming metrics = connection metrics (e.g. is connected) + streaming metrics (e.g. time behind source).
    * Multi-partition streaming task metrics = connection metrics; multi-partition partition metrics = streaming metrics.
[...]

PR #3309, opened 2022-03-16
DBZ-3535: Introduce schema.name.adjustment.mode

morozov:
https://issues.redhat.com/browse/DBZ-3535

**TODO**:
- [x] Update `debezium/debezium-connector-db2` (https://github.com/debezium/debezium-connector-db2/pull/47)
- [x] Update `debezium/debezium-connector-vitess` (https://github.com/debezium/debezium-connector-vitess/pull/77)
- [x] Update `debezium/debezium-connector-cassandra` (https://github.com/debezium/debezium-connector-cassandra/pull/49)
- [x] Add tests for `schema.name.adjustment.mode=none`
- [x] Update documentation
- [x] Decide what to do with `ByLogicalTableRouter` and `CloudEventsConverter` which are instantiated by Kafka Connect and are configured separately.
- [x] Rework `ConnectTableChangeSerializer` and `TransactionMonitor` which currently use a static instance of the default `SchemaNameAdjuster`.
- [x] Should the default value depend on `isUsingAvroConverter()`? Not at this time.

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/LegacyV1MongoDbSourceInfoStructMaker.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/LegacyV1MongoDbSourceInfoStructMaker.java
@@ -21,7 +21,7 @@ public class LegacyV1MongoDbSourceInfoStructMaker extends LegacyV1AbstractSource
     public LegacyV1MongoDbSourceInfoStructMaker(String connector, String version, CommonConnectorConfig connectorConfig) {
         super(connector, version, connectorConfig);
         schema = commonSchemaBuilder()
-                .name(SchemaNameAdjuster.defaultAdjuster().adjust("io.debezium.connector.mongo.Source"))
+                .name(SchemaNameAdjuster.avroAdjuster().adjust("io.debezium.connector.mongo.Source"))
                 .version(SourceInfo.SCHEMA_VERSION)
                 .field(SourceInfo.SERVER_NAME_KEY, Schema.STRING_SCHEMA)
                 .field(SourceInfo.REPLICA_SET_NAME, Schema.STRING_SCHEMA)
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
@@ -614,7 +614,8 @@ public class MongoDbConnectorConfig extends CommonConnectorConfig {
             .connector(
                     MAX_COPY_THREADS,
                     SNAPSHOT_MODE,
-                    CAPTURE_MODE)
+                    CAPTURE_MODE,
+                    SCHEMA_NAME_ADJUSTMENT_MODE)
             .create();
 
     /**
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorTask.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorTask.java
@@ -71,13 +71,13 @@ public final class MongoDbConnectorTask extends BaseSourceTask<MongoDbPartition,
     @Override
     public ChangeEventSourceCoordinator<MongoDbPartition, MongoDbOffsetContext> start(Configuration config) {
         final MongoDbConnectorConfig connectorConfig = new MongoDbConnectorConfig(config);
-        final SchemaNameAdjuster schemaNameAdjuster = SchemaNameAdjuster.create();
+        final SchemaNameAdjuster schemaNameAdjuster = connectorConfig.schemaNameAdjustmentMode().createAdjuster();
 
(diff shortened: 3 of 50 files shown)


gunnarmorling:
> Decide what to do with ByLogicalTableRouter and CloudEventsConverter which are instantiated by Kafka Connect and are configured separately.

Could you clarify the concern here some more?

> Should the default value depend on isUsingAvroConverter()? Probably, not at this time. In theory, the key and the value converters may be different while the adjuster is the same for all schema names.

Agreed. Let's use Avro-style adjustment by default, no matter what the converter is.

morozov:
> > Decide what to do with ByLogicalTableRouter and CloudEventsConverter which are instantiated by Kafka Connect and are configured separately.
> 
> Could you clarify the concern here some more?

In my understanding, unlike other components where the correct adjuster instance can be injected via a constructor (which is done in this patch), Debezium doesn't instantiate `ByLogicalTableRouter` and `CloudEventsConverter`. It is done by the Kafka Connect framework based on the `transforms.*` and the `key|value.converter` configuration respectively.

Currently, the `CommonConnectorConfig` class is responsible for instantiating the adjuster based on the configuration, and the connector task classes are responsible for passing it down the line.

The concern is that neither the adjuster nor the connector config instantiated by the task are available in the transform's or converter's constructor. So it's not clear how and whether these components should take the connector configuration into account. The possible options are:

1. Extract the logic of adjuster instantiation into some separate component and reuse it in the tasks and the transform and converter.
2. Duplicate the logic if instantiation of the adjuster in the transform and converter.
3. Leave it as is for now.

Furthermore, it's not clear to me whether it's a good idea for the transform and the converter to be even aware of the connector configuration. In theory, it should be possible to use them with any source connector. In this case, it's unclear how they should behave.

It's not clear if these components should be in the business of adjusting the schema names in the first place. Assuming that they process the messages produced by the connector that is configured to adjust them to a certain format, isnt' it fair to expect that the names are already adjusted?

morozov:
@jpechane regarding your suggestion in https://github.com/debezium/debezium/pull/2946#issuecomment-1041217481:

> A tip for future - if you will use the same branch names accross all repos then the Db2 and Vitess builds will not be run against `main` but against the corresponging branch

I am using `DBZ-3535` as the branch name in the pull requests against all four repos but the Cassandra build is currently failing because Cassandra is [built](https://github.com/debezium/debezium/runs/5608761301?check_suite_focus=true#step:3:42) from the `main` branch:

> Retrieving the default branch name
> ...
> Default branch 'main'

The DB2 and Vitess builds are passing but they are also built with the `main` branch from those repos.

Was the pipeline supposed to checkout `DBZ-3535` of all other repos or your recommendation was about something else?

morozov:
>  Yeah, it would make sense for this logic to apply in both directions.
[…]
> @jpechane IIRC, we had decided of keeping this cross repo build one-directional i.e. cross repo is built against main repo PR and not mess with our `main` repo builds.

After giving it another thought, I believe it makes some sense to use the `main` versions of the incubating connectors. This way, we sort of identify the breaking internal API changes. So if this is the intention, it should be fine to keep the existing behavior. Otherwise, each incubating connector's build will be duplicated: once here, and once in the PR against its own repo.

gunnarmorling:
> After giving it another thought, I believe it makes some sense to use the main versions of the incubating connectors. This way, we sort of identify the breaking internal API changes. So if this is the intention, it should be fine to keep the existing behavior. Otherwise, each incubating connector's build will be duplicated: once here, and once in the PR against its own repo.

Yes, exactly. I think the CI set-up works exactly as intended in this regard.

gunnarmorling:
@morozov, so regarding the SMT/converter question, that's a tricky one indeed. Piggy-backing on the actual connector option is not ideal, for the reasons you describe. IMO, we should add similar options to those you added to the connectors to the SMT and the converter too.

> Assuming that they process the messages produced by the connector that is configured to adjust them to a certain format, isnt' it fair to expect that the names are already adjusted?

No, I don't think that's the case. For the router, the user creates new topic names via regexp replacements, which may require adjustment for Avro. For the CE converter, this impacts the name of the CE schema into which messages are wrapped, those names are not coming in from the connectors themselves

In terms of a timeline, if we could get this into this PR today or tomorrow, we could add it to 1.9 CR1 (due on Friday). Otherwise, we can also keep it for 1.9 Final (next week, this kind of isolated small-risk is fine IMO), or 2.0 in the worst case. Just let us know. Thanks!

rk3rn3r:
Thx a lot for the explanation @morozov!

> But given the backward compatibility concern, "none" isn't yet the default. It will be the default in 2.0. @rk3rn3r could you clarify where exactly in Jira did you see the intention to have the "real-AVRO and legacy-AVRO adjusters"?

I think that was only my assumption because in some comment you said that the current behavior is too restrictive and you want a less-restrictive behavior, but still be AVRO-compatible. That was maybe my mistake of that argument that we gonna have a restictive-AVRO and a less-restricitve AVRO adjuster with 2.0.

morozov:
> IMO, we should add similar options to those you added to the connectors to the SMT and the converter too.

I added two more commits that update the SMT and the converter.

gunnarmorling:
> I added two more commits that update the SMT and the converter.

Perfect, thanks a lot. LGTM now, merging.

PR #3310, opened 2022-03-16
DBZ-4860 Use JdbcConfiguration class

jpechane:
https://issues.redhat.com/browse/DBZ-4860

### DIFF
--- a/debezium-microbenchmark-oracle/src/main/java/io/debezium/performance/connector/oracle/EndToEndPerf.java
+++ b/debezium-microbenchmark-oracle/src/main/java/io/debezium/performance/connector/oracle/EndToEndPerf.java
@@ -251,10 +251,7 @@ public class EndToEndPerf {
         }
 
         private OracleConnection getTestConnection() {
-            Configuration config = testConfig().build();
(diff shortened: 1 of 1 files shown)


debezium-microbenchmark-oracle/src/main/java/io/debezium/performance/connector/oracle/EndToEndPerf.java
```diff
@@ -252,7 +252,7 @@ private JdbcConfiguration testJdbcConfig() {
 
         private OracleConnection getTestConnection() {
             Configuration config = testConfig().build();
-            Configuration jdbcConfig = config.subset("database.", true);
+            JdbcConfiguration jdbcConfig = JdbcConfiguration.adapt(config.subset("database.", true));
```
gunnarmorling, line 255: Could we call `testJdbcConfig()` instead?

jpechane: @gunnarmorling Yes, you are right, code updated.

PR #3311, opened 2022-03-16
DBZ-4842 Relax supplemental log check when table does not exist

Naros:
https://issues.redhat.com/browse/DBZ-4842


gunnarmorling (approved):
LGTM, merging.

PR #3313, opened 2022-03-17
DBZ-4878 Re-read incremental snapshot chunk on DDL event

kgalieva:
A new schema can be applied to the chunk with an old schema if DDL event is within the chunk select window.
photo_2022-03-16_23-35-10

The chunk's JDBC schema in this case is the same as before. Schema change won't be detected. The binlog schema will get updated with the DDL event. The chunk is inserted after window close event. The events mapping uses the latest schema for the chunk.

The fix is simple: re-read chunk on DDL event. In this case chunk will have schema that is up to date with the binlog schema.

### DIFF
--- a/debezium-connector-mysql/README.md
+++ b/debezium-connector-mysql/README.md
@@ -77,17 +77,17 @@ This instructs Maven to run the normal Maven lifecycle through `integration-test
 
 ### Using MySQL with GTIDs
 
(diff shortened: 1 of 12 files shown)


jpechane:
@kgalieva It seems that `Db2` and Oracle connectors needs to be updated as well.

jpechane:
@kgalieva The code LGTM, unfortunately testing it would be quite difficult - but not impossible.
I think we can merge it when the project becomes buildable. Thanks a lot!

jpechane:
@kgalieva Applied, thanks a lot!

PR #3319, opened 2022-03-18
DBZ-3949 Ignore DML records in binlog in MySQL connector

vjuranek:
In some cases (e.g. in some cases when MEMORY engine is used), DML
records can appear in binlog in spite of the fact it's set to ROW
format. Always ignore these records.

https://issues.redhat.com/browse/DBZ-3949

### DIFF
--- a/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlConnectorIT.java
+++ b/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlConnectorIT.java
@@ -2427,7 +2427,7 @@ public class MySqlConnectorIT extends AbstractConnectorTest {
 
     @Test
     @FixFor("DBZ-3949")
-    public void testDMLInChangeEvents() throws Exception {
+    public void testDmlInChangeEvents() throws Exception {
         config = DATABASE.defaultConfig()
                 .with(MySqlConnectorConfig.TABLE_INCLUDE_LIST, DATABASE.qualifiedTableName("products"))
                 .with(MySqlConnectorConfig.INCLUDE_SCHEMA_CHANGES, false)
--- a/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlConnectorIT.java
+++ b/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlConnectorIT.java
@@ -2424,4 +2424,61 @@ public class MySqlConnectorIT extends AbstractConnectorTest {
 
         stopConnector(value -> assertThat(logInterceptor.containsWarnMessage(DatabaseSchema.NO_CAPTURED_DATA_COLLECTIONS_WARNING)).isFalse());
     }
+
+    @Test
+    @FixFor("DBZ-3949")
+    public void testDMLInChangeEvents() throws Exception {
+        config = DATABASE.defaultConfig()
+                .with(MySqlConnectorConfig.TABLE_INCLUDE_LIST, DATABASE.qualifiedTableName("products"))
+                .with(MySqlConnectorConfig.INCLUDE_SCHEMA_CHANGES, false)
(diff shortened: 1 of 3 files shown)


gunnarmorling:
So with the MEMORY engine, we do receive the legitimate ROW-style events (which we should consume), and _in addition_ ilegitimate STATEMENT-style events?

vjuranek:
I wasn't able to reproduce with MEMORY engine, so it seems that to get DML records, more specific environment if needed (possibly some older MySql version, maybe something else - I asked the requester in the Jira for details). That being said, I cannot exclude that there are also illegitimate  events, but what is in Jira is legitimate event. The idea is that we check binlog uses ROW format when connector starts and if we get DML messages, they come from non-standard sources like tables with MEMORY engine, which the user is very likely not interested in anyway (no matter if they are legitimate or not).

To make it more resilient, we can introduce dedicate option for this situation - by default fail when we get DML and in error message suggest use to use this option if the user is sure these events come from the source (s)he is not interested in, as proposed by @jpechane in the Jira.

gunnarmorling:
Ok, I agree that if we validate that the binlog mode is ROW upon start-up and fail if that's not the case, then it's ok to ignore DML records. In fact, it seems exactly to be what I've suggested [here](https://issues.redhat.com/browse/DBZ-3949?focusedCommentId=18969468&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18969468)?

> we can introduce dedicate option for this situation

Don't we already have this in form of `eventDeserializationFailureHandlingMode`? Do we really need it though? In the light of the above, could we even deprecate and later remove that option? Or is it used elsewhere still?

vjuranek:
>  In fact, it seems exactly to be what I've suggested [here](https://issues.redhat.com/browse/DBZ-3949?focusedCommentId=18969468&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18969468)?

yes, it is

 
> Don't we already have this in form of `eventDeserializationFailureHandlingMode`? 

Kind of yes. This is more general option - ignore any failure. So it can be used as a workaround for original issue, but if there are any other issues with binlog, the use won't see it. Another option could make sense as MySQL sends events for all databases, not only for those the user is interested in, so it could make sense to introduce another option which would ignore only these events, which are very likely from unrelated databases. But always ignoring it (without any option), is IMHO also fine.

> Do we really need it though? In the light of the above, could we even deprecate and later remove that option? Or is it used elsewhere still?

IMHO still makes sense when the user wants to ignore any kind of failures when reading binlog, so I'd keep it.

jpechane:
@gunnarmorling This PR is correct in the light of ignoring DMLs in the binlog.
If this is the final decision then it can be safely merged.

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlConnectorIT.java
```diff
@@ -2424,4 +2424,61 @@ public void testNoEmptySchemaLogWarningWithSnapshotNever() throws Exception {
 
         stopConnector(value -> assertThat(logInterceptor.containsWarnMessage(DatabaseSchema.NO_CAPTURED_DATA_COLLECTIONS_WARNING)).isFalse());
     }
+
+    @Test
+    @FixFor("DBZ-3949")
+    public void testDMLInChangeEvents() throws Exception {
```
gunnarmorling, line 2430:
```suggestion
    public void testDmlInChangeEvents() throws Exception {
```

(It's the convention we adhere to for abbeviations in identifiers; I know it's not applied consistently yet in existing code, but let's try and do it right in new/changed code).

gunnarmorling (approved):
LGTM. Will squash and merge. Thx, @vjuranek!

gunnarmorling:
Thx for your explanation, too, @vjuranek! In that light, I'd keep the option as is. It provides some value, at the same time I'd not introduce yet another knob for some more fine-grained control. The configuration surface of Debezium is already huge.

PR #3320, opened 2022-03-18
DBZ-4243

jpechane:


jpechane:
@dude0001 @gunnarmorling ok, thanks a lot guys but unfortunately this PR is not going to make to 1.9.0.Final.
I spend a lot of time with it and tried to align it with schema change support but I am afraid this is not technically feasible. At the same time I don't think we should drop schema evolution support in incremental snapshots as this feature was already delivered by community (@kgalieva) and is used in the field.

The main problem is that there is no way how to keep in memory list of table structure in syn with database. We cannot rely on infromation from binlog as it can arrive with delay and we cannot even rely on JDBC as even if we read the current table metadata then if we don't lock the table the schema can change between reading the metadata and data.

The current PR can be quite easily extended to support added columns but not those changed or removed. I can see few ways forward

1. The projection will contain the list of columns if and only if the explicit column list is defined for the table using oulumn include/exclude list, In this case we can support schema evolution unless it touches the captured columns,
2. We can extend the PR with dropping/changing the columns but then we would need to capture SQL exception while reading the chunk. In case of en error we will re-read the schema hopefully getting in sync. In this case we would either need to capture specific exception which I'd avoid as this is nothing standardized or introduce advanced mechanism so retry counter, backoff etc.
3. We will keep the code as is and there will be a new config option that will enable/disable explicit coulmn projection. When enabled the schema evolution will not be supported and vice versa. This will be probably easiest one to implement,

Please let me know your thoughts and preferences.

flobuc:
Hello everyone,
I have followed this discussion since the Bug was filed because my project also stucks on it.

I understand that the schema evolution support will not be kicked by the developers.
But for me development of debezium connectors goes in the wrong direction maybe.
It must be weighed which use case is more important in real projects:
1. Schema changes while anincremental snapshot or
2. only partially CDC events

Also as i can see in the documentation the feature "schema changes while an incremental snapshot" is only supported by MySQL and is not supported by PostgreSQL, Oracle, SQL Server and Db2. So for most of the projects (70% [https://www.statista.com/statistics/1131568/worldwide-popularity-ranking-relational-database-management-systems/]) this feature doesn't work. 

My personal experience is also, that changing the Schema during the runtime of an application never happened. Changing schema is such an important dangerous action, that this never happens during the runtime of connected applications.
But that not all data is published over CDC, this is for me a very common use case. 
(again, this is just my view).

Maybe it is a solution to manage an incremental snapshot with partial CDC columns without "schema evolution support" by configuration.

So for me, i will try to fork and apply the pull requests to workaround this problem. 

Just my 2 cents
Florian

jpechane:
@flobuc Nic ooints for disucssion. Speaking of PR - would you be willing to give a try to the above proposed option 3?

Naros:

Hi @jpechane, please sign off all commits with:

```
git commit -s
```

PR #3321, opened 2022-03-18
DBZ-4869 Log warning rather than exception on incorrect table-level logging level

Naros:
https://issues.redhat.com/browse/DBZ-4869


gunnarmorling:
So does this reflect the outcome of the discussion in the Jira issue?

PR #3323, opened 2022-03-21
DBZ-3762 Filter out MySQL inline comments by default

vjuranek:
https://issues.redhat.com/browse/DBZ-3762


jpechane:
@vjuranek Applied, thanks.

PR #3324, opened 2022-03-21
DBZ-4887 Allow ANALYZE TABLES statement in MySQL grammar

vjuranek:
https://issues.redhat.com/browse/DBZ-4887


jpechane:
@vjuranek Thanks a lot, it is fine. Now, please also add a test to https://github.com/debezium/debezium/tree/main/debezium-ddl-parser/src/test/resources/mysql/examples

When completed please open a PR against upstream https://github.com/antlr/grammars-v4/tree/master/sql/mysql/Positive-Technologies - it should contain the grammar update and the example.

For the next time if only grammar is affected and not Debezium itself it is enough to add a test just to SQL examples and it is not necessary to update unit test too.

vjuranek:
> @vjuranek Thanks a lot, it is fine. Now, please also add a test to https://github.com/debezium/debezium/tree/main/debezium-ddl-parser/src/test/resources/mysql/examples

done

> When completed please open a PR against upstream https://github.com/antlr/grammars-v4/tree/master/sql/mysql/Positive-Technologies - it should contain the grammar update and the example.

https://github.com/antlr/grammars-v4/pull/2553

> For the next time if only grammar is affected and not Debezium itself it is enough to add a test just to SQL examples and it is not necessary to update unit test too.

thanks, I removed it not to clutter the code with test which are not needed and are duplicit

jpechane:
@vjuranek Applied, thanks.

PR #3331, opened 2022-03-22
DBZ-4900 Use exponential back-off for log availability check

Naros:
https://issues.redhat.com/browse/DBZ-4900

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
@@ -48,15 +48,15 @@ public class LogMinerHelper {
      * @param archiveLogOnlyMode true to mine only archive lgos, false to mine all available logs
      * @param archiveDestinationName configured archive log destination name to use, may be {@code null}
      * @param maxRetries the number of retry attempts before giving up and throwing an exception about log state
-     * @param initialDelayMs the initial delay in milliseconds
-     * @param maxDelayMs the maximum delay in milliseconds
+     * @param initialDelayMs the initial delay
+     * @param maxDelayMs the maximum delay
      * @throws SQLException if anything unexpected happens
      * @return current redo log sequences
      */
     // todo: check RAC resiliency
     public static List<BigInteger> setLogFilesForMining(OracleConnection connection, Scn lastProcessedScn, Duration archiveLogRetention,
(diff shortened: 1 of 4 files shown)


gunnarmorling (commented):
@Naros, LGTM, the only thing that should be changed consistently is to use `Duration` rather than `long` ms. As discussed before, the semantic Java 8 time/date types should be used internally as much as possible, only at the system boundary (i.e. when consuming config options or when calling external APIs not aware of `Duration`) the value should be converted.

jpechane:
@gunnarmorling I've added a commit using Duration so please take a look

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
```diff
@@ -49,34 +48,37 @@
...
+                                                        boolean archiveLogOnlyMode, String archiveDestinationName, int maxRetries,
+                                                        Duration initialDelay, Duration maxDelay)
+            throws SQLException {
         List<BigInteger> ret = new LinkedList<>();
         removeLogFilesFromMining(connection);
 
         // Restrict max attempts to 0 or greater values (sanity-check)
         // the code will do at least 1 attempt and up to maxAttempts extra polls based on configuration
         final int maxAttempts = Math.max(maxRetries, 0);
+        final DelayStrategy retryStrategy = DelayStrategy.exponential(initialDelay.toMillis(), maxDelay.toMillis());
```
gunnarmorling, line 67: We should add an overload for `exponential()` to accept `Duration` too (or better yet, just change it). Can happen separately though.

gunnarmorling:
LGTM now, merging. Thx for the quick update, @jpechane, and the original work, @Naros!

gunnarmorling:
Logged https://issues.redhat.com/browse/DBZ-4902 as a follow-up. Probably a matter of minutes to do this change.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
jpechane: Well, it is a separate task. *All* factory methods need to be updated.

gunnarmorling: Yep, it's tracked via a follow-up issue now.

PR #3334, opened 2022-03-23
DBZ-4896 Allow `ON` as an expression value

vjuranek:
`ON` can be used as a value for e.g. some server system variables
like unique_checks [1].

[1] https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_unique_checks

### DIFF
--- a/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlParser.g4
+++ b/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlParser.g4
@@ -1717,8 +1717,8 @@ uninstallPlugin
 //    Set and show statements
 
 setStatement
-    : SET variableClause ('=' | ':=') expression
-      (',' variableClause ('=' | ':=') expression)*                             #setVariable
+    : SET variableClause ('=' | ':=') (expression | ON)
+      (',' variableClause ('=' | ':=') (expression | ON))*                      #setVariable
     | SET (CHARACTER SET | CHARSET | CHAR SET) (charsetName | DEFAULT)          #setCharset
     | SET NAMES
         (charsetName (COLLATE collationName)? | DEFAULT)                        #setNames
--- a/debezium-ddl-parser/src/test/resources/mysql/examples/ddl_create.sql
+++ b/debezium-ddl-parser/src/test/resources/mysql/examples/ddl_create.sql
@@ -413,6 +413,14 @@ COMMIT;
 END
 #end
 #begin
+-- delimiter //
+CREATE PROCEDURE set_unique_check()
+BEGIN
+    SET unique_checks=off;
+    SET unique_checks=on;
+END; -- //-- delimiter ;
+#end
+#begin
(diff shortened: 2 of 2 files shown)


vjuranek:
Antlr grammar PR: https://github.com/antlr/grammars-v4/pull/2554

jpechane:
DOes it support `set unique_checks=off`?

vjuranek:
> DOes it support `set unique_checks=off`?

Yes, as `OFF` is not in lexer, so it's `off` works without any additional fixes.

debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlParser.g4
```diff
@@ -1717,8 +1717,8 @@ uninstallPlugin
 //    Set and show statements
 
 setStatement
-    : SET variableClause ('=' | ':=') expression
-      (',' variableClause ('=' | ':=') expression)*                             #setVariable
+    : SET variableClause ('=' | ':=') (expression | ON)
```
gunnarmorling, line 1720: It works, but wouldn't it make more sense to make `ON` part of `expression`? Or is it really _only_ allowed exactly here?

gunnarmorling (commented):
One question inline, @vjuranek.

debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlParser.g4
vjuranek: I'm not MySQL expert to be able to say it for sure that this is the only place where it's allowed. And at the same time I'm not able to say for sure that adding it into `expression`, which is widely used, won't cause any issues as `ON` can have different meaning in some other circumstances/expressions. So this seems to me to be more safe change.

vjuranek:
Also updated example with `off` case - for clarity and also to fail fast in the future if `off` is added into the lexer.

debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlParser.g4
gunnarmorling: Ok, makes sense, let's do it the proposed way then.

For some context, in general, we don't care about the semantics of these grammar parts at all from a Debezium perspective (just need to be able to parse it _somehow_, so we can properly parse what we're actually after, which is table DDL). So we can sometimes take shortcuts, though the aspiration still is to have a grammar that makes sense (not the least given we feed back those changes upstream to Antlr).

vjuranek: Thanks for explanation - this is what I thought, that for Debezium it doesn't matter, at as I submitted this change also to Antlr, I don't want to break anything. Someone more experience can improve it in upstream if needed.

PR #3335, opened 2022-03-23
DBZ-4891 Oracle Logminer : escaped single quote handling in DML

AleksejsSibanovs:
Fix Oracle Logminer's escaped single quotes parsing in DMLs, related unit test and fix for the old DBZ-3413 test


Naros:
Hi @AleksejsSibanovs thanks for the contribution, a few things.

First, I noticed that you included a merge commit in the commit history.  We would prefer to avoid this so could you rework the commits so that such a commit doesn't exist?

Secondly, I noticed you added a `Stack<>` to the `LogMinerDmlParser` to deal with the nested single quotes.  Did you happen to check the performance impact of this versus other alternatives?  The parser is a very contentious hotspot in the code and anything we can do to minimize GC and overhead is important.  If you didn't that's okay, I can do so but just wanted to check before I spent time profiling otherwise.

AleksejsSibanovs:
Hi @Naros , thanks for your feedback!

I was thinking about different possible implementation, the one most efficient, probably, would be to collect characters one by one into StringBuffer as we are iterating over the value. But it would make me rewrite perfectly optimized logic, which I wanted to avoid. From the other hand I feel pretty challenged to create faster implementation. But it will take some time.

I'm not quite sure how to get rid of the commit smoothly, in the scope of the existing pull request. Anyway, as I'm going to completely change the implementation, looks like I should delete my repo, create a new one and make changes there, which leads us to the new pull request, is it OK?
Should I close this PR or it should be done by you?

PR #3336, opened 2022-03-23
DBZ-4892: Migrate test-suite fixtures to JUnit extension

jcechace:
https://issues.redhat.com/browse/DBZ-4892

- Tests are now using https://github.com/jcechace/fixture5 extension
- All fixtures rewritten to classes
- Simplified inheritance hierarchy of test classes


jcechace:
@jpechane Is it ok to temporarily add sonatype snapshot repository before ve make sure that the extension is good enough to be released?

jpechane:
@jcechace Applied, really nice job!

PR #3290, opened 2022-03-08
DBZ-4052 Updates properties table in SQL Server connector doc

roldanbob:
[DBZ-4052](https://issues.redhat.com/browse/DBZ-4052)

This change updates entries in the table of configuration properties for the SQL Server  connector based on the recent review. 

Need to clarify whether the discrepancies in the default value for `max.queue.size` are correct: 
* The Db2, MongoDB, MySQL, Oracle, and SQL Server docs say `8192`. 
* The PostgreSQL docs says `20240`. 

Tested in local Antora and downstream builds.

Please backport to 1.9 after merging, if necessary. Thanks.

PR #3338, opened 2022-03-23
DBZ-4906 Fixes typo in link text

roldanbob:
[DBZ-4906](https://issues.redhat.com/browse/DBZ-4906)

Fixes link error introduced in DBZ-4052 [79fd0a6](https://github.com/debezium/debezium/pull/3290/commits/79fd0a6f9dadac6d5aa847e4c0ba8c7038bbb793)

Upstream, this fix does not result in any apparent change, as  the links worked correctly despite the typo.
But the fix is significant downstream, because previously, the error did not pass the validator, and prevented the build from completing. 

This change is required in the 1.9 branch.


roldanbob:
Thanks for the quick review, @gunnarmorling!

PR #3340, opened 2022-03-24
DBZ-2514 Update release & documentation for Cassandra

Naros:
https://issues.redhat.com/browse/DBZ-2514

- [x] Release pipeline
- [x] Documentation
~~- [ ] Debezium BOM version references~~

@jpechane I need a bit of guidance on whether we intend to explicitly list Cassandra 4 in the BOM or do we intend to just continue to reference only Cassandra 3 there for now?


gunnarmorling:
@Naros, all the versions should be managed in the BOM. I.e. let's do the same for 3.x what we do for 4.x. An exception may be needed if there's conflicts between transitive dependencies of the two.

Naros:
Added sibling PR https://github.com/debezium/debezium-connector-cassandra/pull/50

Naros:
@gunnarmorling Since the BOM would reference the same dependency with only version differences, this creates a coordinate conflict in the BOM.  For now, I'm leaving those as the oldest which would be based on `version.cassandra3` for now unless either you or @jpechane have a suggestion to handle this differently.

PR #3341, opened 2022-03-24
DBZ-4891 Oracle Logminer : escaped single quote handling in DML

AleksejsSibanovs:
Fix Oracle Logminer's escaped single quotes parsing in DMLs, related unit test and fix for the old DBZ-3413 test

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/parser/LogMinerDmlParser.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/parser/LogMinerDmlParser.java
@@ -276,8 +276,24 @@ public class LogMinerDmlParser implements DmlParser {
         index += VALUES_LENGTH;
 
         int columnIndex = 0;
+        StringBuilder collectedValue = null;
         for (; index < sql.length(); ++index) {
+            char lookAhead = (index + 1 < sql.length()) ? sql.charAt(index + 1) : 0;
             char c = sql.charAt(index);
+
+            if (inQuote) {
+                if (c != '\'') {
+                    collectedValue.append(c);
+                }
+                else {
+                    if (sql.charAt(index + 1) == '\'') {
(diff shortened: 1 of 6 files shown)


AleksejsSibanovs:
Looks like locally I just run the style reformatting, which doesn't case about StringBuffer. Investigated, understood, changed to StringBuilder

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/parser/LogMinerDmlParser.java
```diff
@@ -276,8 +276,24 @@ private int parseColumnValuesClause(String sql, int start, String[] columnNames,
         index += VALUES_LENGTH;
 
         int columnIndex = 0;
+        StringBuilder collectedValue = null;
         for (; index < sql.length(); ++index) {
+            char lookAhead = (index + 1 < sql.length()) ? sql.charAt(index + 1) : 0;
```
Naros, line 281: This is used, is this intended?

Naros:
Hi @AleksejsSibanovs I pushed one commit here to address a similar concern with `SEL_LOB_LOCATOR` events.  I'm going to remove the unused `lookAhead` in that one method so we can wrap up this PR in preparation for this week's release.

Naros:
Reworked commit messages, squashed my commits & applied to `main`.
Thanks @AleksejsSibanovs !!!

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/parser/LogMinerDmlParser.java
AleksejsSibanovs: My fault, looks like this one introduced in some intermediary version and later somehow survived the refactoring.

quydx:
Has this bug been fixed on version 1.9.0 CR1, I can not find version 1.9 final in download page ?

PR #3342, opened 2022-03-24
Add test for SQL server schema filters

vjuranek:
Add test for SQL server schema filters. This requires smaller change in test helper which currently allows to enable CDC only for tables under default schema.

https://issues.redhat.com/browse/DBZ-2793

### DIFF
--- a/debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SqlServerConnectorIT.java
+++ b/debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SqlServerConnectorIT.java
@@ -8,6 +8,8 @@ package io.debezium.connector.sqlserver;
 import static io.debezium.connector.sqlserver.util.TestHelper.TYPE_LENGTH_PARAMETER_KEY;
 import static io.debezium.connector.sqlserver.util.TestHelper.TYPE_NAME_PARAMETER_KEY;
 import static io.debezium.connector.sqlserver.util.TestHelper.TYPE_SCALE_PARAMETER_KEY;
+import static io.debezium.relational.RelationalDatabaseConnectorConfig.SCHEMA_EXCLUDE_LIST;
+import static io.debezium.relational.RelationalDatabaseConnectorConfig.SCHEMA_INCLUDE_LIST;
 import static org.fest.assertions.Assertions.assertThat;
 import static org.fest.assertions.MapAssert.entry;
 import static org.junit.Assert.assertNull;
@@ -55,6 +57,7 @@ import io.debezium.pipeline.spi.Offsets;
 import io.debezium.relational.RelationalDatabaseConnectorConfig;
 import io.debezium.relational.RelationalDatabaseSchema;
 import io.debezium.relational.RelationalSnapshotChangeEventSource;
+import io.debezium.relational.TableId;
 import io.debezium.relational.Tables;
 import io.debezium.relational.ddl.DdlParser;
 import io.debezium.relational.history.DatabaseHistory;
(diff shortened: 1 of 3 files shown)


debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/util/TestHelper.java
```diff
@@ -314,7 +315,27 @@ protected static void disableDbCdc(SqlServerConnection connection, String name)
...
+     *
+     * @param connection
+     *            sql connection
+     * @param tableName
+     *            the name of the schema, may not be {@code null}
+     * @param tableName
+     *            the name of the table, may not be {@code null}
+     * @throws SQLException if anything unexpected fails
+     */
+    public static void enableSchemaTableCdc(SqlServerConnection connection, String schemaName, String tableName) throws SQLException {
```
vjuranek, line 329: Would be nicer to overload `enableTableCdc()`, but unfortunately `enableTableCdc(SqlServerConnection, String, String)` already exists for enabling CDC for a table with a custom capture name.

gunnarmorling:
@vjuranek, could you check whether those options are in SQL Server connector docs yet, and if not, add them there? Thx!

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/util/TestHelper.java
gunnarmorling: We may use `TableId` alternatively? It's the safer way for identifying tables to begin with.

vjuranek: Good, idea, thanks for the pointer!

vjuranek (commented):
Documentation added

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SqlServerConnectorIT.java
```diff
@@ -2569,6 +2572,98 @@ public void shouldIncludeDatabaseNameIntoTopicAndSchemaNamesInMultiPartitionMode
         assertThat(record.valueSchema().name()).isEqualTo("server1.testDB.dbo.tablea.Envelope");
     }
 
+    @Test
+    @FixFor("DBZ-2793")
+    public void shouldApplySchemaFilters() throws Exception {
+        connection.execute("CREATE SCHEMA s1");
```
jpechane, line 2578: @vjuranek LGTM, thanks! One question though, if I run this test twice, will not it fail as the schemas will already exist?

vjuranek: Right, if your run it again same DB, it would fail. Fixed now.

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SqlServerConnectorIT.java
```diff
@@ -2569,6 +2572,100 @@ public void shouldIncludeDatabaseNameIntoTopicAndSchemaNamesInMultiPartitionMode
         assertThat(record.valueSchema().name()).isEqualTo("server1.testDB.dbo.tablea.Envelope");
     }
 
+    @Test
+    @FixFor("DBZ-2793")
+    public void shouldApplySchemaFilters() throws Exception {
+        connection.execute("DROP SCHEMA IF EXISTS s1");
```
jpechane, line 2578: @vjuranek I think this will fail if the schem is non empty.
How about reordering the commands? delete all tables, schemas and then create all schemas and tables?

vjuranek: Hm, it worked for me, but maybe the during each test run new container was started. I re-ordered the commands and double check that it works when run twice (and no other container is started during the test), so hopefully should be fine now.

jpechane: @vjuranek Thanks for fixing. The issue would be visible if a previous test execution would not clean up the schema - either due to test failure or bug in test which would spread here.

jpechane:
@vjuranek Applied, thanks!

PR #3344, opened 2022-03-26
DBZ-3541: Round decimal default values of integer columns

morozov:
See [DBZ-3541](https://issues.redhat.com/browse/DBZ-3541).


jpechane:
@morozov LGTM. I'd probably use double instead of float. Also, do you now if MySQL would accept numbers in exponential formats as the default value? If yes then such test case should be added too.

jpechane:
@morozov Applied, thanks. I've just added one more test.

PR #3354, opened 2022-03-30
DBZ-4924 Use new action to setup openjdk in all build steps

ani-sha:
https://issues.redhat.com/browse/DBZ-4924


jpechane:
@ani-sha Applied, thanks!

PR #3357, opened 2022-03-30
DBZ-4814 exclude spatial_ref_sys

zalmane:
https://issues.redhat.com/browse/DBZ-4814


gunnarmorling:
@ani-sha, @Naros, any idea why no build seems to have run for this one?

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TestHelper.java
```diff
@@ -245,7 +247,7 @@ public static PostgresSchema getSchema(PostgresConnectorConfig config, TypeRegis
 
     protected static Set<String> schemaNames() throws SQLException {
         try (PostgresConnection connection = create()) {
-            return connection.readAllSchemaNames(Filters.IS_SYSTEM_SCHEMA.negate());
+            return connection.readAllSchemaNames(((Predicate<String>) Arrays.asList("pg_catalog", "information_schema")::contains).negate());
```
gunnarmorling, line 250: Not super happy about this duplication, but also lacking a good idea for how to avoid it. Will merge as-is. Thx, @zalmane!

gunnarmorling (approved):

Naros:
> any idea why no build seems to have run for this one? 

That's an excellent question and one for which I have no answer.  The original contributor isn't _new_ so therefore the actions should have been queued up.  I did see that Github Actions did have a small incident report a few hours ago so perhaps that might be why?

gunnarmorling:
> I did see that Github Actions did have a small incident report a few hours ago so perhaps that might be why?

Ah yeah, that could be it. Seems I should not just have merged, because now there's this failure: https://github.com/debezium/debezium/runs/5759229201?check_suite_focus=true. Did that PR build for you locally, @zalmane?

zalmane:
it did. let me try rebasing and build again locally.

zalmane:
@gunnarmorling I rebased and it actually now fails on the same test. It seems to stem from the change to line 60 in PostgresSchema.
I change it to `config.getTableFilters().dataCollectionFilter()` as you had suggested in the ticket. Seems it is throwing a NullPointerException in that test. 
Thoughts?

jpechane:
@zalmane Got the error, checking.

PR #3359, opened 2022-03-30
DBZ-4911 Use Jedis' clientSetname when establishing Redis connections

spicy-sauce:


jpechane:
@spicy-sauce This is good feature! Could you please just convert the hard-coded strings into constants and ideally centralize them in a single class? Thanks.

PR #3360, opened 2022-03-31
DBZ-4927 Allow `OR REPLACE` when creating trigger

vjuranek:
`CREATE OR REPLACE TRIGGER` is not possible in MySQL [1], however this
is possible in MariaDB [2].

[1] https://dev.mysql.com/doc/refman/8.0/en/create-trigger.html
[2] https://mariadb.com/kb/en/create-trigger/

https://issues.redhat.com/browse/DBZ-4927


vjuranek:
Upstream PR: https://github.com/antlr/grammars-v4/pull/2562

jpechane:
@vjuranek Thanks for ointing this out! Could you please add a comment to the grammer then? We've it already at different places so you can take a lok.

vjuranek:
@jpechane done

jpechane:
@vjuranek Applied, thanks!

PR #3362, opened 2022-03-31
DBZ-4933 Fix NPE on null column data in MySQL

nenad:
When doing a snapshot on a MySQL database that has char type columns (VARCHAR, TEXT, CHAR) and they are null, and a custom converter is present, the MySQL reader throws an NPE.

This fix addresses the problem by returning an object from the `null` value, as well as it checks if the converter is used for the column which is being read from.

https://issues.redhat.com/browse/DBZ-4933

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/AbstractMysqlFieldReader.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/AbstractMysqlFieldReader.java
@@ -40,38 +40,46 @@ public abstract class AbstractMysqlFieldReader implements MysqlFieldReader {
         if (column.jdbcType() == Types.TIME) {
             return readTimeField(rs, columnIndex);
         }
-        else if (column.jdbcType() == Types.DATE) {
+
+        if (column.jdbcType() == Types.DATE) {
             return readDateField(rs, columnIndex, column, table);
         }
+
         // This is for DATETIME columns (a logical date + time without time zone)
         // by reading them with a calendar based on the default time zone, we make sure that the value
         // is constructed correctly using the database's (or connection's) time zone
-        else if (column.jdbcType() == Types.TIMESTAMP) {
+        if (column.jdbcType() == Types.TIMESTAMP) {
             return readTimestampField(rs, columnIndex, column, table);
         }
+
         // JDBC's rs.GetObject() will return a Boolean for all TINYINT(1) columns.
         // TINYINT columns are reported as SMALLINT by JDBC driver
-        else if (column.jdbcType() == Types.TINYINT || column.jdbcType() == Types.SMALLINT) {
+        if (column.jdbcType() == Types.TINYINT || column.jdbcType() == Types.SMALLINT) {
             // It seems that rs.wasNull() returns false when default value is set and NULL is inserted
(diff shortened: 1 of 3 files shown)


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/AbstractMysqlFieldReader.java
```diff
@@ -40,38 +40,46 @@ public Object readField(ResultSet rs, int columnIndex, Column column, Table tabl
         if (column.jdbcType() == Types.TIME) {
             return readTimeField(rs, columnIndex);
         }
-        else if (column.jdbcType() == Types.DATE) {
+
+        if (column.jdbcType() == Types.DATE) {
```
jpechane, line 44: Why do you replace `else if` with `if`, is there any issue with it?

Naros: I was about to mention this.  It looks like the PR uses a slightly different approach to resolving the value converter near the end of the if blocks.  If we don't want to inline the `.getValueConverter(table.id(), column).isPresent()` call in the if check itself, then my suggestion would be to pull the original last `else if` and `else` into a single `else` block, calculate the value of `hasConverterForColumn` and then simply use a nested if-else block to keep the changes as localized as possible, i.e.:

```
else {
  boolean hasConverterForColumn = connectorConfig.customConverterRegistry()
                .getValueConverter(table.id(), column)
                .isPresent();
        if (hasConverterForColumn && TEXT_DATATYPES.contains(column.typeName())) {
            try {
                String columnData = rs.getString(columnIndex);
                if (columnData != null) {
                    return columnData.getBytes(column.charsetName());
                }
            }
            catch (UnsupportedEncodingException e) {
                logger.warn("Unsupported encoding '{}' for column '{}', sending value as String", e.getMessage(), column.name());
            }
        }
        return rs.getObject(columnIndex);
}
```

nenad: @jpechane The `else if` is reduced to just an `if` statement as all `else if` and `else` branches return value. There is no value in having a chained `else if` conditional statement if all branches return value.

To contrast what my change is about:
```java
if (a == 1) {
	return one;
} 
else if (a == 2) {
	return two;
} 
else {
	return three;
}
```
vs
```java
if (a == 1) {
	return one;
}
if (a == 2) {
	return two;
}
return three;
```

Both approaches do the same, and the second approach doesn't have an `else` condition. I'd consider using `else if` only in cases where the branches do not return value but do state modification.

I've removed the `else` conditional because it will be the last thing evaluated and returns, so there is no value in nesting.

This code change does not change the flow of execution of statements and, in my opinion, makes for a more readable code as it reduces the nesting and chained conditionals.

I'd love your feedback and I'm crossing my fingers we can get this merged.

jpechane: @nenad ok, having a second thought for it it is not exactly necessary but you are right that with `return` in place this is not exactly `switch` simulation.
It now looks more the method style I prefer - do as much you can do at top and return early. So I am merging it.
Thanks a lot.

PR #3365, opened 2022-03-31
DBZ-4727 Implement incremental snaphost pause/resume 

vjuranek:
https://issues.redhat.com/browse/DBZ-4727

### DIFF
--- a/debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/IncrementalSnapshotIT.java
+++ b/debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/IncrementalSnapshotIT.java
@@ -7,6 +7,7 @@ package io.debezium.connector.mongodb;
 
 import java.nio.file.Path;
 import java.sql.SQLException;
+import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
@@ -233,6 +234,16 @@ public class IncrementalSnapshotIT extends AbstractMongoConnectorIT {
 
     protected void sendAdHocSnapshotSignal() throws SQLException {
         sendAdHocSnapshotSignal(fullDataCollectionName());
+    }
+
+    protected void sendPauseSignal() throws SQLException {
+        insertDocuments("dbA", "signals",
+                new Document[]{ Document.parse("{\"type\": \"pause-snapshot\", \"payload\": \"{}\"}") });
+    }
+
+    protected void sendResumeSignal() throws SQLException {
+        insertDocuments("dbA", "signals",
+                new Document[]{ Document.parse("{\"type\": \"resume-snapshot\", \"payload\": \"{}\"}") });
     }
 
     protected Map<Integer, Integer> consumeMixedWithIncrementalSnapshot(int recordCount) throws InterruptedException {
@@ -632,6 +643,40 @@ public class IncrementalSnapshotIT extends AbstractMongoConnectorIT {
         final int expectedRecordCount = ROW_COUNT * 2;
         final Map<Integer, Integer> dbChanges = consumeMixedWithIncrementalSnapshot(expectedRecordCount, topicNames.get(1));
         for (int i = 0; i < expectedRecordCount; i++) {
+            Assertions.assertThat(dbChanges).includes(MapAssert.entry(i + 1, i));
+        }
+    }
+
+    @Test
+    public void pauseDuringSnapshot() throws Exception {
+        populateDataCollection();
+        startConnector(x -> x.with(CommonConnectorConfig.INCREMENTAL_SNAPSHOT_CHUNK_SIZE, 1));
+        waitForConnectorToStart();
+
+        sendAdHocSnapshotSignal();
+
+        List<SourceRecord> records = new ArrayList<>();
+        String topicName = topicName();
+        consumeRecords(100, record -> {
+            if (topicName.equalsIgnoreCase(record.topic())) {
+                records.add(record);
+            }
+        });
+
+        sendPauseSignal();
+
+        consumeAvailableRecords(record -> {
+            if (topicName.equalsIgnoreCase(record.topic())) {
+                records.add(record);
+            }
+        });
+        int beforeResume = records.size();
+
+        sendResumeSignal();
+
+        final int expectedRecordCount = ROW_COUNT;
+        Map<Integer, Integer> dbChanges = consumeMixedWithIncrementalSnapshot(expectedRecordCount - beforeResume);
+        for (int i = beforeResume + 1; i < expectedRecordCount; i++) {
             Assertions.assertThat(dbChanges).includes(MapAssert.entry(i + 1, i));
         }
     }
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresMetricsIT.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresMetricsIT.java
@@ -154,6 +154,8 @@ public class PostgresMetricsIT extends AbstractRecordsProducerTest {
         Assertions.assertThat(mBeanServer.getAttribute(getSnapshotMetricsObjectName(), "SnapshotRunning")).isEqualTo(false);
         Assertions.assertThat(mBeanServer.getAttribute(getSnapshotMetricsObjectName(), "SnapshotAborted")).isEqualTo(false);
         Assertions.assertThat(mBeanServer.getAttribute(getSnapshotMetricsObjectName(), "SnapshotCompleted")).isEqualTo(true);
+        Assertions.assertThat(mBeanServer.getAttribute(getSnapshotMetricsObjectName(), "SnapshotPaused")).isEqualTo(false);
+        Assertions.assertThat(mBeanServer.getAttribute(getSnapshotMetricsObjectName(), "SnapshotPausedDurationInSeconds")).isEqualTo(0L);
     }
 
     private void assertSnapshotNotExecutedMetrics() throws Exception {
--- a/debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
@@ -103,6 +103,23 @@ public abstract class AbstractIncrementalSnapshotChangeEventSource<P extends Par
     }
 
     @Override
+    public void pauseSnapshot(P partition, OffsetContext offsetContext) throws InterruptedException {
+        context = (IncrementalSnapshotContext<T>) offsetContext.getIncrementalSnapshotContext();
+        if (context.snapshotRunning() && !context.isSnapshotPaused()) {
+            context.pauseSnapshot();
+        }
+    }
+
+    @Override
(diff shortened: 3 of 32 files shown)


debezium-embedded/src/test/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotTest.java
```diff
@@ -470,6 +493,40 @@ public void snapshotPreceededBySchemaChange() throws Exception {
...
+        startAndConsumeTillEnd(connectorClass(), config);
+        waitForConnectorToStart();
+
+        waitForAvailableRecords(1, TimeUnit.SECONDS);
+        // there shouldn't be any snapshot records
+        assertNoRecordsToConsume();
+
+        sendAdHocSnapshotSignal();
+        // Sleep a bit to stream some events.
+        Thread.sleep(200);
```
jpechane, line 509: I am not sure if this is reliable, would not it be better to either listen on JMX metrics or catching some incoming messages to make sure the few snapshot events have arrived?

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/IncrementalSnapshotIT.java
```diff
@@ -430,6 +441,34 @@ public void updatesLargeChunk() throws Exception {
...
     }
 
+    @Test
+    public void pauseDuringSnapshot() throws Exception {
+        populateDataCollection();
+        startConnector();
+
+        sendAdHocSnapshotSignal();
+        // Sleep a bit to stream some events.
+        Thread.sleep(200);
```
jpechane, line 451: See the comment in AbstractIncrementalSnapshotIT.

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -100,6 +100,25 @@ public void closeWindow(P partition, String id, OffsetContext offsetContext) thr
...
+            context.pauseSnapshot();
+        }
+    }
+
+    @Override
+    public void resumeSnapshot(P partition, OffsetContext offsetContext) throws InterruptedException {
+        context = (IncrementalSnapshotContext<T>) offsetContext.getIncrementalSnapshotContext();
+        if (context.snapshotRunning() && context.isSnapshotPaused()) {
+            context.resumeSnapshot();
+            window.clear();
```
jpechane, line 116: This code is quite similar to `reReadChunk()`. Cannot it be re-used? At the same time, should all those operations be called? IIRC the pause is done after the current chunk is processed, is not it? If yes then this cleanup should be unnecessary.

jpechane (commented):
@vjuranek Thanks a lot, this is really great start.
I've left couple of comments. Could you also please check the Db2 failure?
My key question is - what is the planned general design? Should snapshot be paused in the middle of chunk processing - when part of of the chunk is already processed and loaded in memory or should hit happen after the chunk in progress is completed?

vjuranek:
>  Could you also please check the Db2 failure? 

It fails as it use default chunk size and probably send all the events before pause, with smaller chunk size tests are passing. I'll send a PR for it shortly.

> My key question is - what is the planned general design? Should snapshot be paused in the middle of chunk processing - when part of of the chunk is already processed and loaded in memory or should hit happen after the chunk in progress is completed?

See commit message. When pause is called, currently running chunk is processed, but no other chunk is read. Snapshot is kept in running state while paused.

debezium-embedded/src/test/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotTest.java
vjuranek: Yes, this is not a very good way how to do it. Thanks for suggestions, will take a look on it.

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
vjuranek: I had some issue when using `reReadChunk()`. I'll investigate if the cleanup can be removed - this would be the best way, reuse `reReadChunk()` second one.

gunnarmorling:
> When pause is called, currently running chunk is processed, but no other chunk is read.

+1

> Snapshot is kept in running state while paused.

WDYM by that exactly? Is it visible in metrics that a snapshot is paused? There should be some indication about that fact to the user.

jpechane:
@gunnarmorling Good point, we should either extend `io.debezium.pipeline.metrics.traits.SnapshotMetricsMXBean`. Is probably best way. I wa thinking about having combination of flags like not running/not completed but the separate one will be the most understandable.

jpechane:
@vjuranek If the pause is executed after the window is processed is it then necessary to do anyting around the state change? As after the chunk is processed the window should be empty no read started etc.

debezium-embedded/src/test/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotTest.java
vjuranek: done

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
vjuranek: yes, the cleanup is not needed

vjuranek:
> @vjuranek If the pause is executed after the window is processed is it then necessary to do anyting around the state change? As after the chunk is processed the window should be empty no read started etc.

yes, the cleanup is not needed, removed in the current version

> I wa thinking about having combination of flags like not running/not completed but the separate one will be the most understandable.

so I should add `boolean getSnaphostPaused()` into `SnapshotMetricsMXBean`, right? I guess also `getSnapshotDurationInSeconds()` should be adjusted not to count time during which snapshot was paused. Or better to provide equivalent for pausing, something line `getSnapshotPauseInSeconds()`?

jpechane:
@vjuranek Exactly. Related to the runtime I prefer your latter proposal - `getSnapshotPauseInSeconds`. This will give the user bot htotal time time spent from start till end of the snapshot and also the opportunity to calculate the "hot" time only.

jpechane:
@vjuranek I saw a recent activity. Is this sitll a draft?

vjuranek:
@jpechane yes, recent pushes were related to rebasing and fixing stuff and failing tests after rebase. There are still TBD items:
* JMX
* kafka based signals for MySQL

Working on it now and will convert to PR once finished.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresMetricsIT.java
```diff
@@ -154,6 +154,8 @@ private void assertSnapshotMetrics() throws Exception {
         Assertions.assertThat(mBeanServer.getAttribute(getSnapshotMetricsObjectName(), "SnapshotRunning")).isEqualTo(false);
         Assertions.assertThat(mBeanServer.getAttribute(getSnapshotMetricsObjectName(), "SnapshotAborted")).isEqualTo(false);
         Assertions.assertThat(mBeanServer.getAttribute(getSnapshotMetricsObjectName(), "SnapshotCompleted")).isEqualTo(true);
+        Assertions.assertThat(mBeanServer.getAttribute(getSnapshotMetricsObjectName(), "SnapshotPaused")).isEqualTo(false);
+        Assertions.assertThat(mBeanServer.getAttribute(getSnapshotMetricsObjectName(), "SnapshotPausedDurationInSeconds")).isEqualTo(0L);
```
Naros, line 158: I noticed we use the "SnapshotPaused" prefix here and in other places and I'm wondering if there is any concern with these metrics being confused with traditional snapshot-based metrics around running, aborted, and completed rather than using "IncrementalSnapshotPaused" or "AdhocSnapshotPaused".  

Also, we should also document these metrics in the documentation.

Naros (commented):
Overall this LGTM to me @vjuranek but I have left one comment below.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresMetricsIT.java
vjuranek: @Naros good point. I was also thinking about it, but as we already have metrics for in incremental snapshot without this prefix (`SnapshotRunning`, `SnapshotDurationInSeconds` etc), I decided for consistency and use it without "Incremental" prefix. It would be probably even more confusing for the users if part of the metrics would start with "Incremental" and part not, although it's related to incremental metrics. 

As we work on 2.0 where we can introduce backward incomapatible changes, we may rename all the metrics related to incremental snapshot.

As for the documentation, I completely forgot on it. Will added. Thanks!

vjuranek: @Naros documentation added. Looking more into JXM (and also signalling documentation), it looks like we mix traditional and incremental freely (use one JXM metric for both of them). Similar signalling documentation distinguish between `ad-hoc` snapshot and `incremental` snapshot, but e.g. in case of `stop-snapshot`, it speaks about `ad-hoc` snapshot and later on uses `incremental` snapshot. So I'm little but confused what is the right approach here. I included pause and resume signals under `Incremental snapshots` sub-chapter.

Naros: @vjuranek ah, I didn't realize those others were related to incremental snapshot and didn't have a prefix, so I can see your point. Perhaps what we can take is a small action item to discuss with @jpechane when he returns if he feels we need to improve the naming of these metrics to provide a clear distinction and if so we can create a follow-up task for that.

As for the signal documentation itself, I would agree this sounds a bit messy.  I'll take a look at the documentation and I'll add some further comments on that commit if I feel they're necessary.

Naros: Looking at what you did, I think I would agree that stop should be done similar to what you did for pause/resume.  I'll open a small follow-up for this and I'll take care of that tomorrow.

Naros (approved):

Naros:
I'll take care of merging this tomorrow after I get a small change from @roldanbob for the `signaling.adoc`.  Could you do me one favor and adjust the Oracle references you added to not be community only, I mean to include that in my review.

vjuranek:
> Could you do me one favor and adjust the Oracle references you added to not be community only, I mean to include that in my review.

Not sure I follow here. What Oracle references you mean @Naros ? Or this was meant for @roldanbob (the change for `signaling.adoc`)?

vjuranek:
Not sure if I got it right, but I removed `ifdef::community[]` and `endif::community[]` for Orcale link as it's done in PR #3756

Naros:
That's perfect, thanks @vjuranek!

Naros:
Applied to main (=2.0), I'll check with Jiri on whether we want to back port this to 1.9 when he's back.

PR #3369, opened 2022-04-05
DBZ-4955 Bump Jackson to 2.23.2

jpechane:
https://issues.redhat.com/browse/DBZ-4955

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -97,8 +97,9 @@
         <!-- Kafka and it's dependencies MUST reflect what the Kafka version uses -->
         <version.kafka>3.1.0</version.kafka>
         <version.zookeeper>3.6.3</version.zookeeper>
-        <version.jackson>2.12.3</version.jackson>
-        <version.jackson.databind>2.12.3</version.jackson.databind>
+        <!-- Kafka uses 2.12.3 but we need to bump due to CVE-2020-36518 before Kafka fixes it -->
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -97,8 +97,9 @@
         <!-- Kafka and it's dependencies MUST reflect what the Kafka version uses -->
         <version.kafka>3.1.0</version.kafka>
         <version.zookeeper>3.6.3</version.zookeeper>
-        <version.jackson>2.12.3</version.jackson>
-        <version.jackson.databind>2.12.3</version.jackson.databind>
+        <!-- Kafka uses 2.12.3 but we need to bump due to CVE-2020-36518 before Kafka fixes it -->
+        <version.jackson>2.13.2</version.jackson>
```
gunnarmorling, line 101: LGTM, any reason though for having two variables? Couldn't `version.jackson` be used for all the artifacts?

jpechane: @gunnarmorling It is for productization purposes. There were some versions we needed to decouple jackson and databind as the redhat suffix was different.

gunnarmorling: Ok, I see. Worth a comment there probably then. Other than that, LGTM. Leaving it to you to merge, as I'm not sure what status main is in. Thanks!

gunnarmorling (approved):

PR #3373, opened 2022-04-06
DBZ-4950 Remove MySQL legacy implementation

ani-sha:
https://issues.redhat.com/browse/DBZ-4950
https://issues.redhat.com/browse/DBZ-4965


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnection.java
```diff
@@ -468,9 +467,7 @@ public String connectionString() {
 
     public static class MySqlConnectionConfiguration {
 
-        protected static final String JDBC_PROPERTY_LEGACY_DATETIME = "useLegacyDatetimeCode";
```
jpechane, line 471: @ani-sha I've commented on this one yesterday but the comment probably got lost :-(.
This and the following property are related to JDBC driver not our legacy implementation. Could you please create a separate Jira for this change and extract into a separate commit just in case we would need to revert it for some reason? Thanks

ani-sha: @jpechane I should add the props back & its related change with a separate commit & Jira task?

jpechane:
Failure tracked via https://issues.redhat.com/browse/DBZ-4966

jpechane:
@ani-sha Is this still a draft?

ani-sha:
@jpechane Ready for the review!

jpechane:
@ani-sha COuld you please change it a bit. Right now you've introduced the legacy handling back. What I'd like to see is to take the original [DBZ-4950](https://issues.redhat.com/browse/DBZ-4950) commit, and split it intow two. One that will contain DBZ-4950 related removals and the other DBZ-4965 removals.

jpechane:
@ani-sha That's exactly what I was looking for. I'll merge it when the CI is completed.

Naros:
Hi @ani-sha, that might be a timing problem with some of the recent changes with the Cassandra 3 & 4 split.  Let's log a Jira issue for that, and could you attach the raw "8_Build Debezium Conenctor Cassandra.txt" file to the issue in case the logs get wiped from GH actions before we can investigate?

jpechane:
@ani-sha Applied, thanks. Can you just please create a Jira as proposed by @Naros ?

ani-sha:
@jpechane @Naros Logged https://issues.redhat.com/browse/DBZ-4972 for the cassandra failure.

PR #3380, opened 2022-04-07
DBZ-4879 Correctly resolve archive-log-only SCN mining range

Naros:
https://issues.redhat.com/browse/DBZ-4879

At a high level, the issue here is that a separate query for calculating the maximum SCN from the `V$ARCHIVED_LOG` table was being used and this could return a SCN that was higher than the logs that had previously been added to the mining session due to a redo log aging out to an archive log in between the collection of logs and calculation of the maximum archive log SCN.

The goal here was to instead return a `List<LogFile>` that way we resolve the logs once and we then use this list instead as the basis for the SCN boundary calculations when archive log mode is enabled, completely avoiding this race condition.


jpechane (approved):
@Naros LGTM, will merge when Oracle CI issue is resolved.

jpechane:
@Naros Applied, thanks

PR #3381, opened 2022-04-07
DBZ-4936 Don't re-emit transaction on restart with commit SCN equal to offset commit SCN

Naros:
https://issues.redhat.com/browse/DBZ-4936


jpechane:
@Naros Applied, thanks

PR #2871, opened 2021-10-29
DBZ-2975 schema recovery

gunnarmorling:

PR #3382, opened 2022-04-07
DBZ-4968 Remove server name from source partition

morozov:
**Change summary**:

1. Remove `SERVER_PARTITION_KEY` from all `Partition` implementations.
2. Remove `connectorConfig` from the `Partition.Provider` constructors that used it only to source the server name.
3. Update `Partition` unit tests. With the server name removed, for most of the `Partition` implementations, there are no other properties that would make different `Partition` instances non-equal.
4. Remove [DBZ-3347](https://issues.redhat.com/browse/DBZ-3347) tests for Oracle and SQL Server (single-partition mode) since the record source partition in these cases no longer contains any properties. The SQL Server connector source partition in multi-partition mode still contains the database name, so in this case, the test is still valid.

#### Backward compatibility considerations

This is a breaking API change.

1. The connector won't be able to read the offsets produced by its older version. The Kafka Connect connector requests the committed offsets for each source partition, the format of which is being changed.
   
   The users of the connectors will have to produce the offset in the new format during the upgrade. There don't seem to be any standard tools for that (see [KAFKA-4107](https://issues.apache.org/jira/browse/KAFKA-4107), [KIP-199](https://cwiki.apache.org/confluence/display/KAFKA/KIP-199%3A+Add+Kafka+Connect+offset+tool)) but there is an [article](https://rmoff.net/2019/08/15/reset-kafka-connect-source-connector-offsets/) online that describes the principles.
2. The connector won't be able to consume database history produced by its older version. As of https://github.com/debezium/debezium/pull/2871 ([DBZ-2975](https://issues.redhat.com/browse/DBZ-2975)), during the database history recovery, the messages are filtered by the source partition(s) being processed by the task.
   
   The users of the connectors will have to rewrite the database history during the upgrade and remove the `source.server` field from each message:
   ```bash
   kcat -b <brokers> -C -t <history-topic> -e > history.json
   cat history.json | jq 'del(.source.server)' > history-2.0.json
   # delete and create <history-topic>
   kcat -b <brokers> -P -t <history-topic> < history-2.0.json
   ```

**TODO**:
- [ ] Provide an example of producing a new connector offset.
- [x] Provide an example of rewriting database history.


jpechane:
@morozov Could you please rebase your PR? https://github.com/debezium/debezium/pull/3373 removed all legacy MySQL related code.

gunnarmorling:
Thanks, @morozov! Given the highly disruptive nature of this change, I'm wondering though whether it actually is worth it?

While we are ready to make breaking changes for 2.0, defaulting to making the upgrade not a drop-in one in each and every case, requiring complex operations like rewriting the history, I'm quite doubtul. After all, that's a non user-facing nuisance, which fixing IMHO isn't worth such disruption of users.

Curious what other folks here think?

morozov:
> Given the highly disruptive nature of this change, I'm wondering though whether it actually is worth it?

That's a fair concern. I didn't realize that this change would affect the database history until I implemented it.

Here are a few considerations:
1. A similar migration will be required for the SQL Server connectors anyways if the support for single-partition mode is removed (the committed offsets and the database history will have to contain the database name for the connector to recognize them).
2. As discussed offline `database.server.name` is part of the API exposed by the connector (topic names, schema names). The fact that it's also imprinted into each database history message an undesired implementation detail. Being an internal topic, the database history topic shouldn't be coupled to the public API.
   1. The current implementation may make certain operational scenarios unnecessarily complex. For instance, there is an SQL Server instance `srv1` with databases `a` and `b` and a corresponding multi-database connector `kctr1`. There's also a similar setup `srv2` / `kctr2`. In order to reduce the load on `srv1`, I want to move `a` from `srv1` to `srv2`. Besides just copying the database history of `a` from `srv1.history` to `srv2.history`, I also need to alter every message by replacing `srv1` with `srv2` because it's part of the source metadata. At the same time, it's not used for anything and is redundant.
3. If there are other, more user-facing changes planned for 2.0 that would require a similar migration anyways, it would be justified to piggyback and make this change as well.

As for intrusiveness, I'd say that the migration process itself is fairly simple but it indeed may be complicated by different operational factors, e.g. the number of connectors to migrate, the size of the history to rewrite, etc.

PR #3385, opened 2022-04-08
DBZ-4971 Add triggers for 2.x paths in Github CI

ani-sha:
https://issues.redhat.com/browse/DBZ-4971


jpechane:
@ani-sha Applied, thanks

PR #3386, opened 2022-04-08
DBZ-4949 Use Java 11 as a baseline

vjuranek:
Debezium now uses Java 11 as baseline and Java 17 for tests, e.g.
```
$ javap -verbose debezium-core/target/classes/io/debezium/annotation/GuardedBy.class |grep "major version"
  major version: 55
$ javap -verbose debezium-core/target/test-classes/io/debezium/config/ConfigDefinitionMetadataTest.class |grep "major version"                                                                                        
  major version: 61
```
TBD:
* test changes in Jenkins job
* update Cassandra connector to be able to run tests with Java 17, see [DBZ-4910](https://issues.redhat.com/browse/DBZ-4910)

https://issues.redhat.com/browse/DBZ-4949

### DIFF
--- a/.github/workflows/debezium-workflow.yml
+++ b/.github/workflows/debezium-workflow.yml
@@ -510,11 +510,11 @@ jobs:
       - name: Checkout Action
         uses: actions/checkout@v2
 
-      - name: Set up Java 11
+      - name: Set up Java 17
         uses: actions/setup-java@v2
         with:
           distribution: 'temurin'
-          java-version: 11
+          java-version: 17
 
       - name: Cache Maven Repository
         uses: actions/cache@v2
@@ -542,11 +542,11 @@ jobs:
       - name: Checkout Action
         uses: actions/checkout@v2
 
-      - name: Set up Java 11
+      - name: Set up Java 17
         uses: actions/setup-java@v2
         with:
           distribution: 'temurin'
-          java-version: 11
+          java-version: 17
 
       - name: Cache Maven Repository
         uses: actions/cache@v2
--- a/debezium-connect-rest-extension/pom.xml
+++ b/debezium-connect-rest-extension/pom.xml
@@ -15,7 +15,6 @@
     <maven.build.timestamp.format>yyyy-MM-dd HH:mm:ss</maven.build.timestamp.format>
     <timestamp>${maven.build.timestamp}</timestamp>
     <maven.compiler.parameters>true</maven.compiler.parameters>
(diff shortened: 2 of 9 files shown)


gunnarmorling:
> update Cassandra connector to be able to compile it with Java 11, see [DBZ-4910](https://issues.redhat.com/browse/DBZ-4910)

That needs to happen first though, right?

> Debezium now uses Java 11 as baseline and Java 17 for tests

Loving the move to 17 for tests; this could even go in right away.

vjuranek:
> > update Cassandra connector to be able to compile it with Java 11, see [DBZ-4910](https://issues.redhat.com/browse/DBZ-4910)
> 
> That needs to happen first though, right?

yes


> Loving the move to 17 for tests; this could even go in right away.

No, we have to fix Cassandra test first, they require Java 8 as they run embedded Cassandra 3.

vjuranek:
> No, we have to fix Cassandra test first, they require Java 8 as they run embedded Cassandra 3.

the summary was misleading (that Cassandra has issue with Java 11 during compilation), fixed now

gunnarmorling:
Yes, true; the Cassandra tests will need updating first. Gnarf.

jpechane:
@vjuranek ok, it seems that Cassandra has now issues with Java 17 :-(. Please take a look whether this is something that can be solved via JVM runtime config and if not then let's use Java 11 for Cassandra tests.

vjuranek:
@jpechane yes, looking on it. But FIY, if configuring JVM args won't help, we unfortunately cannot simply use JDK 11 only for Cassandra. Tests depends on Debezium core and embedded and fail with
```
java.lang.UnsupportedClassVersionError: io/debezium/kafka/KafkaCluster has been compiled by a more recent version of the Java Runtime (class file version 61.0), this version of the Java Runtime only recognizes class file versions up to 58.0
        at java.base/java.lang.ClassLoader.defineClass1(Native Method)
        at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1017)
```

jpechane:
@vjuranek Cassnd build now passes. So the only thing left is Jenkins?

vjuranek:
> @vjuranek Cassnd build now passes. So the only thing left is Jenkins?

@jpechane yes, I need to do at least some testing of changes in Jenkins jobs, that it works. Once verified, PR should be ready.

jpechane:
@vjuranek Applied, thanks a lot

PR #3388, opened 2022-04-08
DBZ-4953 Correctly handle TRUNCATE when LOB is enabled

Naros:
https://issues.redhat.com/browse/DBZ-4953


jpechane:
@Naros Applied, thanks

PR #3389, opened 2022-04-08
DBZ-4932 Use LCR v1 format with Oracle 12.0 and 12.1

Naros:
https://issues.redhat.com/browse/DBZ-4932


jpechane:
@Naros Applied, thanks

PR #2660, opened 2021-08-31
DBZ-3942 Load correct last commit LSN from offsets

jpechane:
https://issues.redhat.com/browse/DBZ-3942

PR #3393, opened 2022-04-11
DBZ-4865 PostgreSQL connector stuck in loop after auto restart on error

scott2112:
Postgres: Added a startup/restarting flag to prevent LSNs from being flushed during a restart.
In rare circumstances, if an LSN is flushed during restart, it can cause an infinite loop if it flushes to an LSN in the future past a WAL resume position.


jpechane:
@scott2112 Thanks for the PR. Unfortunately the tests are failing :-(. Can you please take a look?

Naros:
Hi @scott2112, not sure if you saw Jiri's prior comment but it seems the PostgreSQL tests are failing with your changes.  Could you take a look into why and make any adjustments that are necessary?

Naros:
Hi @scott2112 a gentle ping, could you take a look at the test failures?

scott2112:
Hi, yes, it's on my to-do list.  I've been busy with my main job lately because we're right at the end of a very large project, but I'll take a look first chance I get.

jpechane:
Closing in favour of https://github.com/debezium/debezium/pull/2660 and https://github.com/debezium/debezium/pull/3620

PR #3398, opened 2022-04-12
DBZ-4983 Use correct Java encoding Charset in AbstractMysqlFieldReader

LarsWerkman:
https://issues.redhat.com/browse/DBZ-4983

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/AbstractMysqlFieldReader.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/AbstractMysqlFieldReader.java
@@ -28,6 +28,7 @@ public abstract class AbstractMysqlFieldReader implements MysqlFieldReader {
     protected final Logger logger = LoggerFactory.getLogger(getClass());
 
     private static final Set<String> TEXT_DATATYPES = Collect.unmodifiableSet("CHAR", "VARCHAR", "TEXT");
+    private static final MySqlValueConverters.CharsetMappingWrapper CHARSET_MAPPING_WRAPPER = new MySqlValueConverters.CharsetMappingWrapper();
 
     private final MySqlConnectorConfig connectorConfig;
 
@@ -71,7 +72,7 @@ public abstract class AbstractMysqlFieldReader implements MysqlFieldReader {
             try {
                 String columnData = rs.getString(columnIndex);
                 if (columnData != null) {
-                    return columnData.getBytes(column.charsetName());
+                    return columnData.getBytes(CHARSET_MAPPING_WRAPPER.getJavaEncodingForMysqlCharSet(column.charsetName()));
                 }
             }
             catch (UnsupportedEncodingException e) {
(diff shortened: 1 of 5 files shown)


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/AbstractMysqlFieldReader.java
```diff
@@ -28,6 +28,7 @@ public abstract class AbstractMysqlFieldReader implements MysqlFieldReader {
     protected final Logger logger = LoggerFactory.getLogger(getClass());
 
     private static final Set<String> TEXT_DATATYPES = Collect.unmodifiableSet("CHAR", "VARCHAR", "TEXT");
+    private static final MySqlValueConverters.CharsetMappingWrapper CHARSET_MAPPING_WRAPPER = new MySqlValueConverters.CharsetMappingWrapper();
```
jpechane, line 31: @LarsWerkman Do we need a separate instance? Can we expose the one on MySQLValueConverters as it is static?

jpechane:
@LarsWerkman LGTM, thanks! I left one comment related to havi separate/re-used singleton.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/AbstractMysqlFieldReader.java
LarsWerkman: @jpechane That would be possible! I didn't do that as I was unsure what the standards are of exposing static final members in the project. What also would be possible is to extract the `CharsetMappingWrapper` class and make the function static like so,

```suggestion
    public final static class CharsetMappingWrapper extends CharsetMapping {
        static String getJavaEncodingForMysqlCharSet(String mySqlCharsetName) {
            return CharsetMapping.getStaticJavaEncodingForMysqlCharset(mySqlCharsetName);
        }
    }
```

I was unsure why an instance of this class was created in the first place. ( `CHARSET_MAPPING_WRAPPER` )

Let me know what you prefer and I can update the pull request accordingly

jpechane: @LarsWerkman Let me explain the reason. We introduced it as it happened to us already that the mapping was done differently (different classes) for two versions of JDBC drivers and we needed to support both so this was effectivelly isolation code.

I think right now you can create a sttaic method on `io.debezium.connector.mysql.MySqlConnection` that would do the call to the JDBC driver. It would probably make most sense to be placed there.

LarsWerkman: @jpechane I see! thank you for the explanation. I can definitely move the method to `io.debezium.connector.mysql.MySqlConnection` 
however I don't think we can remove the wrapping class, because the method `CharsetMapping.getStaticJavaEncodingForMysqlCharset` is a protected method so we cannot call it directly.
But I can move that wrapping class to the MySqlConnection class and expose it with a nice static method if that is preferred

jpechane: @LarsWerkman Yes, that's how I envisioned it, thanks. I started the CI and when passed the PR can be merged.

LarsWerkman: @jpechane it seems I didn't sort the imports correctly, If you could start the CI once more for me that would be much appreciated

jpechane:
@LarsWerkman Applied, thanks a lot!

PR #3399, opened 2022-04-13
DBZ-4976 Handle NPEs for column definitions of Oracle DDL statements

ani-sha:
https://issues.redhat.com/browse/DBZ-4976


jpechane:
@Naros Applied, thanks

PR #3400, opened 2022-04-13
DBZ-4980 Support CTE with create view statement in MySQL parser

ani-sha:
https://issues.redhat.com/browse/DBZ-4980

### DIFF
--- a/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlLexer.g4
+++ b/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlLexer.g4
@@ -613,6 +613,7 @@ QUERY:                               'QUERY';
 QUICK:                               'QUICK';
 REBUILD:                             'REBUILD';
 RECOVER:                             'RECOVER';
+RECURSIVE:                           'RECURSIVE';
 REDO_BUFFER_SIZE:                    'REDO_BUFFER_SIZE';
 REDUNDANT:                           'REDUNDANT';
 RELAY:                               'RELAY';
--- a/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlParser.g4
+++ b/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlParser.g4
@@ -241,6 +241,23 @@ createTrigger
       routineBody
     ;
 
+withClause
+    : WITH RECURSIVE? commonTableExpressions
+    ;
+
+commonTableExpressions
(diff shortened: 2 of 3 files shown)


debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlLexer.g4
```diff
@@ -613,6 +613,7 @@ QUERY:                               'QUERY';
 QUICK:                               'QUICK';
 REBUILD:                             'REBUILD';
 RECOVER:                             'RECOVER';
+RECURSIVE:                           'RECURSIVE';
```
jpechane, line 616: @ani-sha As you are adding a new keyword, can t be used as table/column name? If yes then this should be reflected in the grammart too - `keywordCanBeId`

debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlParser.g4
```diff
@@ -241,14 +241,30 @@ createTrigger
       routineBody
     ;
 
+withClause
+    : WITH RECURSIVE? commonTableExpressions
+    ;
+commonTableExpressions
```
jpechane, line 247: Probably an empty line should be inserted

debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlParser.g4
```diff
@@ -241,14 +241,30 @@ createTrigger
...
+    : cteName (('(' cteColumnName (',' cteColumnName)* ')')?  AS '(' dmlStatement ')')
+      (',' commonTableExpressions)?
+    ;
+
+cteName
```
jpechane, line 252: Can you please prepare a PR for upstream? I wonder if they would like to have the rules done this way or they'd prefer to use `uid` directly in the rules above. Ideally we should do it in the same way.

ani-sha:
Upstream PR - https://github.com/antlr/grammars-v4/pull/2577

debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlParser.g4
ani-sha: @jpechane created the upstream PR. I will wait for any suggestions from them and apply them here also.

ani-sha:
@jpechane Applied changes as per upstream & upstream PR is merged.

jpechane:
@ani-sha Applied, thanks!

PR #3402, opened 2022-04-13
DBZ-4978 Avoid NPE during Oracle connector shutdown

Naros:
https://issues.redhat.com/browse/DBZ-4978


jpechane:
@Naros Applied, thanks

PR #3403, opened 2022-04-14
DBZ-4987 Propagate up all exceptions from LCR handler

Naros:
https://issues.redhat.com/browse/DBZ-4987


jpechane:
@Naros Applied, thanks

PR #3405, opened 2022-04-14
DBZ-4994 Fix LOB locator event parsing with escaped single quotes

Naros:
https://issues.redhat.com/browse/DBZ-4994


jpechane:
@Naros Applied, thanks

PR #3406, opened 2022-04-14
DBZ-4996 Parse references_clause without referenced column list

Naros:
https://issues.redhat.com/browse/DBZ-4996

### DIFF
--- a/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/oracle/generated/PlSqlParser.g4
+++ b/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/oracle/generated/PlSqlParser.g4
@@ -3351,7 +3351,7 @@ foreign_key_clause
     ;
 
 references_clause
-    : REFERENCES tableview_name paren_column_list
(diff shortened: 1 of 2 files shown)


debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/oracle/generated/PlSqlParser.g4
```diff
@@ -3351,7 +3351,7 @@ foreign_key_clause
     ;
 
 references_clause
-    : REFERENCES tableview_name paren_column_list
+    : REFERENCES tableview_name paren_column_list? (ON DELETE (CASCADE| SET NULL_))?
```
jpechane, line 3354: @Naros Is the underscore in the `NULL` intentional?
Could you please add an example that would verffy the added grammar fragment?
Could you pleas eprovide an upstream PR?
Thanks

Naros: Hi @jpechane, yes the underscore is intentional, the Oracle grammar specifically uses `NULL_` as the lexical token for `NULL`.

jpechane:
@Naros Applied, thanks

PR #3407, opened 2022-04-17
DBZ-4998 Should store event header timestamp in HistoryRecord

harveyyue:
https://issues.redhat.com/browse/DBZ-4998
To identify the event time and process time from schema change event, I want to add the "ts_ms" processing time field to SchemaChangeEvent and HistoryRecord.


jpechane:
@harveyyue Nice improvement, thanks!

PR #3409, opened 2022-04-19
DBZ-4966 Redis Sink - Ping before creating pipeline

spicy-sauce:


jpechane:
@spicy-sauce Applied, thanks.

PR #3410, opened 2022-04-19
DBZ-5003 Can't use 'local' database through mongos

jchipmunk:
https://issues.redhat.com/browse/DBZ-5003

MongoDb connector should use "config" database if the database exists. Otherwise, "local" database should be used to get MongoDB version during connector validation.

This is similar to the approach implemented in ReplicaSetDiscovery class.


jpechane:
@jchipmunk Applied, thanks.

PR #3411, opened 2022-04-19
DBZ-5005 adjust LogMiner batch size based on comparison with currently used batch size

domsj:
https://issues.redhat.com/browse/DBZ-5005

seems more logical to use the current batch size in order to determine if said batch size needs to be increased or decreased.
(additionally it feels weird that the default batch size currently has an extra - hidden - meaning; this makes it harder to pick a good value for it.)
not sure if there's any formal theory to point to? (makes intuitive sense for me ...)

opening PR towards 1.9 (so I'll be able to enjoy the change faster, and because it feels like a 'fix').
is there any kind of formal statement you need from me or the company I work for regarding IP?

cc @Naros

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -139,6 +139,7 @@ Guy Korland
 Guy Pascarella
 Grzegorz Kołakowski
 Jacob Gminder
+Jan Doms
 Jan Hendrik Dolling
 Jason Schweier
 Jiabao Sun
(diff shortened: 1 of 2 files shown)


Naros:
Hi @domsj could you raise a Jira enhancement issue and then reword the commits to be prefixed with the Jira issue?

domsj:
@Naros https://issues.redhat.com/browse/DBZ-5005 created and PR updated

domsj:
> Hi @domsj could you please make sure this targets `main` rather than `1.9`; thanks!

updated

Naros:
There are some Oracle LogMiner test failures I am looking into.  I'll provide an update after I've investigated the errors.

Naros (approved):

PR #3413, opened 2022-04-20
DBZ-5007 Run doc changes workflow only on main repo

vjuranek:
* Don't run doc changes workflow on fork where it fails anyway.
* Fix cron comment to reflect the reality.

https://issues.redhat.com/browse/DBZ-5007


jpechane:
@vjuranek Applied, thanks

PR #3414, opened 2022-04-20
DBZ-4947 Build Debezium UI upon changes in debezium-core

ani-sha:
https://issues.redhat.com/browse/DBZ-4947

### DIFF
--- a/.github/workflows/debezium-workflow.yml
+++ b/.github/workflows/debezium-workflow.yml
@@ -34,6 +34,7 @@ jobs:
       rest-extension-changed: ${{ steps.changed-files-rest-extension.outputs.any_changed }}
       schema-generator-changed: ${{ steps.changed-files-schema-generator.outputs.any_changed }}
       debezium-testing-changed: ${{ steps.changed-files-debezium-testing.outputs.any_changed }}
+      debezium-ui-changed: ${{ steps.changed-files-debezium-ui.outputs.any_changed }}
       ddl-parser-changed: ${{ steps.changed-files-ddl-parser.outputs.any_changed }}
       documentation-only-changed: ${{ steps.changed-files-documentation.outputs.only_changed}}
     steps:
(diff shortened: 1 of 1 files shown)


jpechane:
@ani-sha Can we have the UI build also part of the PR build as we have for Db2 and other non-core connectors?

ani-sha:
@jpechane Yes. But do we want this kind of build also or just the PR build would be fine? This current workflow is triggered only upon `push` in the UI repository.

jpechane:
@ani-sha Ideally I'd like to have a PR failure if Debezium UI build is not working with the new PR.

jpechane:
@ani-sha Could you please take a look? The UI build is failing.

.github/workflows/debezium-workflow.yml
```diff
@@ -768,3 +781,36 @@ jobs:
...
+          key: maven-debezium-test-build-${{ hashFiles('core/**/pom.xml') }}
+          restore-keys: |
+            maven-debezium-test-build-${{ hashFiles('core/**/pom.xml') }}
+
+      - name: Build Debezium UI
+        run: >
+          ./mvnw clean install -f debezium-ui/backend/pom.xml -am -pl backend -B
+          -Ddebezium.test.records.waittime=20
+          -Dformat.formatter.goal=validate
+          -Dformat.imports.goal=check -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
```
ani-sha, line 816: @jpechane I tried running with this command, but it fails. What do you think would be the right command?

Naros: Hi @ani-sha so the issue is that the UI uses the Maven wrapper which is only available from the `./debezium-ui` path so perhaps for this build step we should use:
```
cd debezium-ui && ./mvnw clean install -f backend/pom.xml ...
```
Give that a try and see if that works.

Naros: Another option is to add `working-directory` to the build step and specify `debezium-ui` and omit the `cd` portion of the runnable command too.

jpechane: Already tested in https://github.com/debezium/debezium/pull/3481
I am going to get the commit here
But we need to fix the UI build first
It fails for me locally too

jpechane:
The merge is blocked by https://issues.redhat.com/browse/DBZ-5125

jpechane:
@ani-sha Applied, thanks a lot. The UI tests are now passing.

PR #3416, opened 2022-04-20
DBZ-4600 - Add a new ByteArrayConverter with delegate support

naznaz00:
**Overview**
Debezium currently provides `ByteBufferConverter`. This converter works great when converting from PostgreSQL DBs, but doesn’t work for MongoDB (where the data shape is different - a Binary BSON object internally). See: https://issues.redhat.com/browse/DBZ-4600

Instead we ended up using `ByteArrayConverter` from core Kafka-Connect - which works fine for our basic use-case.

However - we want to take advantage of the delegate config and pattern Debezium provides. 
I.e. the user should be able to provide a value for `delegate.converter.type` in the configuration to specify a “backup” converter which can be used in the event the shape of the data doesn’t match that expected. 

We've recently added heartbeats to our Debezium config - following the docs - and immediately discovered that the Kafka-Connect `ByteArrayConverter` fails to cope with the Json sent by the heartbeat. Given this "real" use-case it seemed sensible to me to add a Debezium-packaged `ByteArrayConverter`. Also I want my application to work :)

**Note**
It's worth noting that we've tested this code locally at AT; running the build Debezium snapshot jars and testing this converter with heartbeats enabled; and things worked correctly and as expected.

### DIFF
--- a/debezium-core/src/main/java/io/debezium/converters/BinaryDataConverter.java
+++ b/debezium-core/src/main/java/io/debezium/converters/BinaryDataConverter.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package io.debezium.converters;
+
+import java.io.IOException;
+import java.nio.ByteBuffer;
+import java.util.Map;
+
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.SchemaAndValue;
+import org.apache.kafka.connect.errors.DataException;
+import org.apache.kafka.connect.storage.Converter;
+import org.apache.kafka.connect.storage.ConverterConfig;
+import org.apache.kafka.connect.storage.HeaderConverter;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.debezium.config.Configuration;
+import io.debezium.config.Instantiator;
+
+/**
+ * A custom value converter that allows Avro messages to be delivered as raw binary data to kafka. <p/>
+ *
+ * Designed to be used in the outbox pattern where payloads are pre-serialized by KafkaAvroSerializer
+ * within the application, then get written as either a ByteBuffer or Byte[] to the database. <p/>
+ *
+ * These raw payloads become the body of the resulting Kafka message; allowing the consumer to deserialize
+ * using KafkaAvroDeserializer. <p/>
+ *
+ * To enable the converter in a connector, the following value need to be specified:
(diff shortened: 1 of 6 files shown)


Naros:
Hi @nathan-bradshaw-at this looks like a great contribution.  This originally stemmed from working with the Outbox pattern and a MongoDB environment based on the Jira description.  We do discuss the `ByteBufferConverter` in the outbox documentation.  Could you also update the outbox docs with information about the `ByteArrayConverter` as well for completeness?

naznaz00:
Hi @Naros - Cheers, I've updated the Mongo SMT docs, swapping out `ByteBufferConverter` for `ByteArrayConverter` in the Avro section. I'm fairly confident this is the better converter to use (from our battles with it :D) - and added an explanatory note to try and avoid confusion around the similar converters.

Naros (approved):
This LGTM, do you have any objections @jpechane ?

jpechane:
@naznaz00 This LGTM but I have one question. Do we need a separate converter? Can't we just extend the ByteBufferConverter (and rename it to `BinaryDataConverter`) to support both `byte[]` and ByteBuffer?

naznaz00:
> @naznaz00 This LGTM but I have one question. Do we need a separate converter? Can't we just extend the ByteBufferConverter (and rename it to `BinaryDataConverter`) to support both `byte[]` and ByteBuffer?

Hi @jpechane - this is difficult for me to answer as an external contributor - but I lean towards keeping them separate to avoid a breaking-change to any existing config people have - which renaming `ByteBufferConverter` would be.

I'm not totally against changing the _new_ converter to `BinaryDataConverter` and deprecating the existing one? That would give people time to adjust to the change, but still eventually give a single converter for all database types.

Really though; I'm broadly happy to do what I'm told :D

jpechane:
@naznaz00 Thanks for raising that point. In fact normally we'd follow the path you descibes. But given we are moving to Debezium 2.0 we can and will afford some breaking changes.
Under normal conditions you'd just rename the class and then create a new subcalls with the same name as the old one marked as deprecated but as I mentiond I feel we can afford some shortcuts here.

Naros:
I'm curious since 1.9 will be an effective "LTS" version until Debezium 2.0 debuts later this year, is there an advantage to take this PR as is and apply it to both 1.9 and 2.0 and then follow up specifically on 2.0 to consolidate the behavior?  I'm not sure how important this converter would be for users on 1.9 that use MongoDB, perhaps doing this in the context of 2.0 is enough?

jpechane:
@Naros You are right, this issue borders on bug/enhancement boundary. So I still would prefer to have it in a single class I mentioned above but with a separate commit that would provide `ByteBufferConverter` as deprecated class. This commit will be applied to 1.9 only but not to main.
The documentation will be updated with note like this was originally `ByteBufferConverter` which is deprecated.

jpechane:
@naznaz00 Applied, thanks

PR #3417, opened 2022-04-20
DBZ-5015 Get the json value of a Document when parsing signaling message

chadthman:
Incremental snapshots on a MongoDB connector are broken. In the signaling table you are expected to provide json in the `data` (ex: `"data": { "data-collections": ["database.collection"] }`) that specifies which tables to make the snapshot. However when parsing that `data` using `Document.parse` and looping over the values of the fields, if it finds nested json will be treated as another `Document`. Just calling the `.toString()` on that will not give the value in json. We need the resulting string for that field to be in json still as there is another json parser later in the method `Signal.process`

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -60,6 +60,7 @@ Brandon Maguire
 Brennan Vincent
 Camile Sing
 Cao Manh Dat
+Chad Marmon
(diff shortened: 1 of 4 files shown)


jpechane:
@chadthman Applied, thanks!

PR #3418, opened 2022-04-21
DBZ-5019 Redis-Sink - Check if client is not null before closing it

spicy-sauce:


jpechane:
@spicy-sauce Applied, thanks

PR #3424, opened 2022-04-22
[release] Changes for 2.0 release process

jpechane:


debezium-parent/pom.xml
```diff
@@ -29,7 +29,7 @@
         <version.docker.maven.plugin>0.31.0</version.docker.maven.plugin>
         <version.staging.plugin>1.6.12</version.staging.plugin>
         <version.protoc.maven.plugin>3.8.0</version.protoc.maven.plugin>
-        <version.javadoc.plugin>3.3.2</version.javadoc.plugin>
+        <version.javadoc.plugin>3.4.0</version.javadoc.plugin>
```
vjuranek, line 32: Is there any reason to have version defined also here? It seems to me quite confusing, I'd remove it from here and keep it only in main `pom.xml`

jpechane: There is https://issues.redhat.com/browse/DBZ-5026 dedicated for the task

vjuranek: I see, thanks

PR #3427, opened 2022-04-25
DBZ-4910 Don't build Cassandra with JDK8

vjuranek:
Cassandra now allows building with JDK 11 and therefore modified
enforcer rules were removed and default one (requiring at least JDK11)
is used.


jpechane:
@vjuranek Applied

PR #3428, opened 2022-04-25
DBZ-5006 Allow "/" characters in Oracle table names

Naros:
https://issues.redhat.com/browse/DBZ-5006


Naros:
I've also added 2 small commits to this PR to address scoping two tests to their respective Jira issues (as part of `main` or `2.0`) which are necessary to avoid test failures under the Oracle XStream adapter.

jpechane:
@Naros LGTM! Could you please extend the tests with the following cases if supported

* slash at the beginning of the table name
* slash at the end
* name with two slashes
* name with two slashes just one behind the other `//`

Naros:
@jpechane I added the requested test coverage.

jpechane:
@Naros LGTM! Please merge yourself this time to make sure I will not mixup 1.9. Thanks

PR #3429, opened 2022-04-26
DBZ-5026 Remove duplicate definition of maven plugin

ani-sha:
https://issues.redhat.com/browse/DBZ-5026


jpechane:
@ani-sha Applied, thanks!

PR #3430, opened 2022-04-26
DBZ-5007 Use single quotes in repo name

vjuranek:
Github actions allows only signle quote in expressions [1].

[1] https://docs.github.com/en/actions/learn-github-actions/expressions#literals


jpechane:
@vjuranek Thanks but @ani-sha was faster :-) - see https://github.com/debezium/debezium/pull/3431

PR #3438, opened 2022-04-27
DBZ-5037 update HTTP connector pom

zalmane:


jpechane:
@zalmane Applied, thanks.

PR #3439, opened 2022-04-27
DBZ-5062 Upgrade to Quarkus 2.8.2.Final

Naros:
https://issues.redhat.com/browse/DBZ-5062


jpechane:
@Naros Applied, thanks a lot!

PR #3440, opened 2022-04-28
DBZ-5064 Use Maven 3.8.4 for GitHub CI

ani-sha:
https://issues.redhat.com/browse/DBZ-5064

### DIFF
--- a/.github/workflows/debezium-workflow.yml
+++ b/.github/workflows/debezium-workflow.yml
@@ -168,6 +168,11 @@ jobs:
           distribution: 'temurin'
           java-version: 17
 
+      - name: Set up Maven
+        uses: stCarolas/setup-maven@v4.3
+        with:
+          maven-version: 3.8.4
+
       - name: Cache Maven Repository
         id: maven-cache-check
         uses: actions/cache@v2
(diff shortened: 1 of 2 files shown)


pom.xml
```diff
@@ -316,7 +316,7 @@
                         <configuration>
                             <rules>
                                 <requireMavenVersion>
-                                    <version>3.6.3</version>
+                                    <version>3.8.4</version>
```
vjuranek, line 319: Looks like this enforce is not used. I'd either use (preferred) it or remove it.

vjuranek (commented):
I think right now we use default (i.e. random) maven version in github workflows. IMHO we should add some [maven action](https://github.com/marketplace?type=actions&query=maven) (e.g. [sdkman](https://github.com/sdkman/sdkman-action)) to use maven with specified version.

pom.xml
jpechane: @vjuranek Please create a Jira for enforcing

jpechane:
@ani-sha Thanks

pom.xml
vjuranek: https://issues.redhat.com/browse/DBZ-5069

PR #3445, opened 2022-04-29
DBZ-5072 Config the cache size property for ByLogicalTableRouter caches

harveyyue:
https://issues.redhat.com/browse/DBZ-5072


jpechane:
@harveyyue LGTM, thanks! Could you please also extend the documentation of the SMT?

harveyyue:
> ! Could you please also extend the documentation of the SMT?

Yes, add `schema.name.adjustment.mode` and `logical.table.cache.size` to topic-routing doc.

jpechane:
@harveyyue Applied, thanks a lot

PR #3447, opened 2022-04-30
DBZ-5075: Clean up MySQL connector codebase

morozov:
1. `DdlChanges#groupStatementStringsByDatabase()` and `DatabaseStatementStringConsumer` interface were introduced in #49 but weren't used anywhere since then.
2. `DdlChanges#groupStatementsByDatabase`, `DatabaseStatementConsumer` and `HaltingPredicate` are unused since #3373.
3. `MySqlSchema#applyDdl()` is also no longer available.
4. The only terminator used in `DdlChanges` is a semicolon.


debezium-connector-mysql/src/test/java/io/debezium/relational/ddl/DdlChangesTest.java
```diff
@@ -1,97 +0,0 @@
-/*
```
jpechane, line 1: Why is this test completely removed?

jpechane:
@morozov Hi, nice initiative! Overall LGTM, I've just dropped a single question.

debezium-connector-mysql/src/test/java/io/debezium/relational/ddl/DdlChangesTest.java
morozov: All its methods test `DdlChanges#groupEventsByDatabase()` which isn't used anywhere else than in this test.

jpechane:
@morozov Applied, go for more chopping :-)

PR #3451, opened 2022-05-03
DBZ-4939 Fall back to JDBC driver if we fail to read timestamp

vjuranek:
MySQL protocol can change during connection and as a result our custom
field reader may fail. In such case fall back to JDBC to read the value.
This is workaround until proper solution, which will require bigger
refactoring, is implemented, see DBZ-5084 for more details.

https://issues.redhat.com/browse/DBZ-4939

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/AbstractMySqlFieldReader.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/AbstractMySqlFieldReader.java
@@ -42,14 +42,30 @@ public abstract class AbstractMySqlFieldReader implements MySqlFieldReader {
         }
 
         if (column.jdbcType() == Types.DATE) {
-            return readDateField(rs, columnIndex, column, table);
+            try {
+                return readDateField(rs, columnIndex, column, table);
+            }
+            catch (RuntimeException e) {
+                logger.warn("Failed to read date value: '{}'. Trying default ResultSet implementation.", e.getMessage());
+                // If our field reader failed, let's try JDBC as the last resort.
+                // Workaround for DBZ-5084.
+                return rs.getObject(columnIndex);
+            }
         }
 
         // This is for DATETIME columns (a logical date + time without time zone)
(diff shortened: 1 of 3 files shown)


vjuranek:
There were some random failures in `IncrementalSnapshotIT#schemaChanges`, but probably not related to this PR (i.e. already there), as this test doesn't use datetime AFAICT.

jpechane:
@vjuranek Before the final merge of the workaround - could you please check what {{ResultSet}} class is returened in both cases? If it is different then the refactoring should be pretty easy.

vjuranek:
Always returns `com.mysql.cj.jdbc.result.ResultSetImpl`

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/AbstractMySqlFieldReader.java
```diff
@@ -49,7 +49,15 @@ public Object readField(ResultSet rs, int columnIndex, Column column, Table tabl
         // by reading them with a calendar based on the default time zone, we make sure that the value
         // is constructed correctly using the database's (or connection's) time zone
         if (column.jdbcType() == Types.TIMESTAMP) {
-            return readTimestampField(rs, columnIndex, column, table);
+            try {
+                return readTimestampField(rs, columnIndex, column, table);
+            }
+            catch (RuntimeException e) {
+                logger.warn(String.format("Failed to read timestamp value: '%s'. Trying default ResultSet implementation.", e.getMessage()));
```
jpechane, line 56: @vjuranek Could you please use standard `{}` formatting placeholder in log message instead of `String.format`?

vjuranek: Sure, fixed, @jpechane

vjuranek:
@jpechane should be ready now. Tests now cover datetime/timestamp, date and time. Time for some reason works fine without fixing anything.

jpechane:
@vjuranek Applied, thanks

PR #3452, opened 2022-05-04
DBZ-4180 Pluggable topic selector

harveyyue:
https://issues.redhat.com/browse/DBZ-4180


harveyyue:
Related vitess pr: https://github.com/debezium/debezium-connector-vitess/pull/81
db2 pr: https://github.com/debezium/debezium-connector-db2/pull/51

debezium-api/src/main/java/io/debezium/spi/topic/TopicNamingStrategy.java
```diff
@@ -0,0 +1,44 @@
...
+import io.debezium.common.annotation.Incubating;
+
+/**
+ * An interface that defines the topic naming strategy, including DataChange, SchemaChange, Transaction, Heartbeat events etc.
+ *
+ * @param <I>
+ * @author Harvey Yue
+ */
+@Incubating
+public interface TopicNamingStrategy<I> {
```
jpechane, line 19: `I` should extend `DataCollectionId`?

jpechane:
@harveyyue Thank a lot, this is really nice piece of work. I've few comments though

- [ ] There should not be per-connector naming strategies. IIRC there are only two - based on database name and database schema name
- [ ] I really don't think the `topic.regex` and `topic.replacement` should be part of the default naming strategy. There should be a separate strategy for their use
- [ ] `String dataChangeTopic(I id, TopicNameProvider<I> provider)` should not be part of the interface
- [ ] Could you please think about moving `DataCollectionId` into API module and introducing a method like `List<String>` name parts? That would be used in topic naming strategy. I feel it is time to expose a collection/table name in a generic way in the API. There should be one more method returning array/list of stings describing the components - `schema`, `table`, etc.

harveyyue:
@jpechane I have fixed these issue.

debezium-api/src/main/java/io/debezium/spi/topic/TopicNamingStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+
+    String schemaChangeTopic();
+
+    String heartbeatTopic();
+
+    String transactionTopic();
+
+    void setLogicalName(String logicName);
+
+    String sanitizedTopicName(String topicName);
```
morozov, line 34: Does the naming strategy have to be responsible for sanitization? As discussed in [DBZ-4180](https://issues.redhat.com/browse/DBZ-4180), how the topic name is sanitized pretty much doesn't depend on how it's built.

The sanitization logic is primarily defined by the topic naming constraints of Kafka, not the user preferences.

debezium-api/src/main/java/io/debezium/spi/topic/TopicNamingStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+
+    String dataChangeTopic(I id);
+
+    String schemaChangeTopic();
+
+    String heartbeatTopic();
+
+    String transactionTopic();
+
+    void setLogicalName(String logicName);
```
morozov, line 32: It looks like each strategy instance will have one logical name during its entire lifetime. Instead of accepting it via a separate method, would it be possible to make it part of the constructor?

E.g. the `OffsetCommitPolicy` classes accept their `Properties` via the constructor. This approach might eliminate the need for `configure` as well.

morozov (commented):
@harveyyue thank you for the patch. On top of what Jiri suggested, I have a couple more comments.

debezium-core/src/main/java/io/debezium/relational/TableId.java
```diff
@@ -145,6 +148,21 @@ public String identifier() {
         return id;
     }
 
+    @Override
+    public List<String> parts() {
+        return Collect.arrayListOf(catalogName, schemaName, tableName);
```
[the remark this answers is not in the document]
morozov: If we add the `CONNECTOR` part to the mix, it may eliminate the need for the introduction of the `topic.prefix` property and the `LOGIC_NAME_PLACEHOLDER`. It would be just another variable that could be optionally put anywhere in the `topic.parts` property.

morozov:
> I really don't think the `topic.regex` and `topic.replacement` should be part of the default naming strategy. There should be a separate strategy for their use

There is already the [org.apache.kafka.connect.transforms.RegexRouter](https://docs.confluent.io/platform/current/connect/transforms/regexrouter.html) SMT. Why would Debezium need a similar feature?

harveyyue:
> 

We met following performance issue when using RegexRouter to capture 400 shards to 4 topics.
1. The cache size hard code had been fixed.
2. Build the SourceRecord for twice time is wasting cpu time, we want to merge these actions to `DefaultRegexTopicNamingStrategy`.

jpechane:
@morozov IMHO this is valid reason, I also really hate the need to rebuild the Connect record from scratch if a small change is needed :-(. That's IMHO the Achilless heel of SMTs. I am aware of advantages of immutable objects but sometimes on need to do an optimization.

@harveyyue Slightly related but also unrelated question. `ByLogicalTableRouter` also renames the schemas. I really woder if it would be useful to provide not only pluggable topic selector but also pluggable schema name generator. Not as a part of this PR - just food for thought.

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -531,15 +534,14 @@ public static SchemaNameAdjustmentMode parse(String value) {
             .withImportance(Importance.MEDIUM)
             .withDescription("The name of the data collection that is used to send signals/commands to Debezium. Signaling is disabled when not set.");
 
-    public static final Field TRANSACTION_TOPIC = Field.create("transaction.topic")
-            .withDisplayName("Transaction topic name")
+    public static final Field CUSTOM_TOPIC_NAMING_STRATEGY = Field.create("custom.topic.naming.strategy")
+            .withDisplayName("Custom topic naming strategy")
             .withGroup(Field.createGroupEntry(Field.Group.ADVANCED, 21))
             .withType(Type.STRING)
```
jpechane, line 540: Can't we use type `CLASS` and `withDefault` like in `io.debezium.connector.mysql.MySqlConnectorConfig.DATABASE_HISTORY`?

debezium-core/src/main/java/io/debezium/relational/TableId.java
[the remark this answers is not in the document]
jpechane: Definitely interesting idea for the default naming strategy. It will also remove the need for most regexrouter usescases -like use only table name. But this means that every connector will have its own unique default configuartion of `topic.parts` as some use catalog, some schema, right?

morozov:
Thanks for the clarification regarding the regex-based naming strategy. We use the regex router SMT on quite heavily loaded MySQL connectors (up to ~200,000 tables per connector) and have never identified the SMT to be a bottleneck. But it definitely makes sense to optimize it if it's known to be an issue.

harveyyue:
> 

Clarification one thing, we figured out the performance issue connector bottlenect is reading binlog from the aws mysql (version5.6.mysql_aurora.1.22.4) finally.
But use regex-based naming strategy instead of SMT, we can improve 10-20% cpu time in method WorkSourceTask.sendRecords() according the VisualVM Sampler.

jpechane:
@harveyyue Excellent work, I am going to merge this PR. I've just rebased and force push on the latest main and as soons as CI is completed it is done.
Could you please create another PR with documentation so we don't delay this one any longer?
Thanks a lot for your patience!

harveyyue:
> @harveyyue Excellent work, I am going to merge this PR. I've just rebased and force push on the latest main and as soons as CI is completed it is done. Could you please create another PR with documentation so we don't delay this one any longer? Thanks a lot for your patience!

@jpechane Ok, I will do it.

PR #3453, opened 2022-05-04
DBZ-5087 Added a decorator for result sets that can cancel the parent statements

alwaysbemark:
**Overview**
The changes and rationale are detailed in the DBZ ticket but the gist is that some JDBC drivers (for example, SQL Server for JDBC > 2.0) cannot close() a `ResultSet` without first looping through all the remaining rows. This means that when we try to interrupt a snapshot process it might hang indefinitely for all intents and purposes.

The fix is to cancel() the parent `Statement` which usually involves a round trip to the server, which means that no further results will be returned within the result set.

I tested that it definitely solves our problem with the event change coordinator thread taking too long to stop and doing it not-so-gracefully when it does.

https://issues.redhat.com/browse/DBZ-5087

Open to feedback, thank you!

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -235,6 +235,7 @@ Marc Zbyszynski
 Mario Mueller
 Mariusz Strzelecki
 Mark Drilling
+Mark Bereznitsky
 Martin Medek
 Martin Sillence
 Matt Beary
--- a/debezium-core/src/main/java/io/debezium/jdbc/CancellableResultSet.java
+++ b/debezium-core/src/main/java/io/debezium/jdbc/CancellableResultSet.java
@@ -1,0 +1,1032 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.jdbc;
+
+import java.io.InputStream;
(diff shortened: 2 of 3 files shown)


alwaysbemark:
Hey all, is there anything that needs to be done to run the remaining builds and to target 1.9 in addition to main? Thank you!

jpechane:
@alwaysbemark Thanks a lot! Do you think that other Debezium code would benefit form this as well?
If yes then I can imagine `io.debezium.jdbc.JdbcConnection.connection(boolean)` to return decorated connection that would return decorated resultset.

alwaysbemark:
> @alwaysbemark Thanks a lot! Do you think that other Debezium code would benefit form this as well? If yes then I can imagine `io.debezium.jdbc.JdbcConnection.connection(boolean)` to return decorated connection that would return decorated resultset.

Hey @jpechane , thanks for having a look! I went through the main flow again and could not identify any other potentially massive queries (like the initial table snapshot) so considering the fact that this adds another potential round trip to the server (and even more, depending on the implementations of JDBC driver) I can't recommend adding it anywhere else for now, but I tried to keep it generic so that it can be! 

Is this something that could target 1.9 as well by any chance? Thanks again!

jpechane:
@alwaysbemark Applied, thanks.  As this is pretty isolated change I am going to add it to 1.9 too.

PR #3456, opened 2022-05-04
DBZ-5085 Enforce read-consistency determining snapshot offsets

Naros:
https://issues.redhat.com/browse/DBZ-5085

This fix is to guarantee the read consistency between resolving the current SCN and the in-progress transaction state by using a single database query rather than relying on multiple queries as we did previously.

- [x] Code change
- [x] Introduce integration test (if possible)
- [x] Execute tests for Xstream and LogMiner

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
@@ -5,13 +5,10 @@
  */
 package io.debezium.connector.oracle;
 
-import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.sql.Savepoint;
 import java.sql.Statement;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -20,17 +17,14 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.debezium.pipeline.EventDispatcher;
-import io.debezium.pipeline.source.snapshot.incremental.SignalBasedIncrementalSnapshotContext;
 import io.debezium.pipeline.source.spi.SnapshotProgressListener;
 import io.debezium.pipeline.source.spi.StreamingChangeEventSource;
-import io.debezium.pipeline.txmetadata.TransactionContext;
 import io.debezium.relational.RelationalSnapshotChangeEventSource;
 import io.debezium.relational.Table;
 import io.debezium.relational.TableId;
 import io.debezium.schema.SchemaChangeEvent;
 import io.debezium.schema.SchemaChangeEvent.SchemaChangeEventType;
 import io.debezium.util.Clock;
-import io.debezium.util.HexConverter;
 
 /**
  * A {@link StreamingChangeEventSource} for Oracle.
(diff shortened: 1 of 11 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
```diff
@@ -143,26 +150,178 @@ protected void determineSnapshotOffset(RelationalSnapshotContext<OraclePartition
...
+                                        snapshotPendingTransactions.put(transactionId, Scn.valueOf(rs.getString("SCN")).subtract(Scn.ONE));
+                                    }
+                                    else {
+                                        // Unable to resolve when the transaction started, log warning
+                                        LOGGER.warn("Unable to resolve start SCN for transaction {}, transaction will not be included.", transactionId);
+                                    }
+                                }
+                            }
+                            else {
+                                LOGGER.warn("Failed to find entry for current SCN {}, associated transaction will likely not be included.", currentScnValue);
```
Naros, line 260: This seems to be logged in almost all tests, need to see if there is a better way to avoid this.  This could be because the current SCN may not have been flushed to the redo logs and if so, we may need to find a more appropriate way to just resolve the snapshot scn entirely than based solely on the current SCN.

Naros:
Hi @chanetd I may need an extra set of eyes on this. If I recall you introduced the contribution for in-progress transaction tracking and it's been brought to our attention that there are corner cases where the original implementation did not accurately capture all transactions while the connector transitioned from snapshot to streaming.

In particular, if the current SCN points to a `START` or ` COMMIT` record, it's highly probable that Oracle has not yet written the transaction metadata to `V$TRANSACTION` or has removed it from this table and therefore would not be seen as a pending transaction and as a result, would be omitted from the streamed events.  

In this PR I aimed to use a quick and simple LogMiner query to resolve what the current SCN points to in terms of operation type and to make a decision about whether the current SCN's transaction should be amended to the pending transaction list or whether it was safe to exclude it.  There are most definitely some issues with my changes here I am still working through but I'm curious if you have any other ideas that might be more reliable than what I am already doing.

chanetd:
What we really need for this to work fool-proof is two things:
- an SCN to start snapshotting from (snapshot_scn)
- an SCN to start streaming from, that is far enough before snapshot_scn that we're guaranteed to capture the beginning of all transactions that have start_scn < snapshot_scn and commit_scn > snapshot_scn. (stream_start_scn)

stream_start_scn needn't be exact; if we mine a bit too much, we'll just do some useless work, but we'll drop it if we see that the corresponding commit_scn < snapshot_scn. Ideally, you want to do as little useless work as possible, of course.

My original approach assumed (apparently incorrectly) that `v$transaction` was accurate at all times, but from what I saw in the Jira discussion that is not necessarily the case; there may be a little lag there, causing it to not show a transaction that started at, say, CURRENT_SCN-1.

**first idea (hacky but it may work)**
One approach would be to assume that the `v$transaction` lag is always low, say less than 1000 SCNs behind CURRENT_SCN. If that is the case, we could probably get away with the following logic:
- if the combined query (CURRENT_SCN + v$transaction) yields non-empty results (i.e. there are open transactions), take the start_scn of the oldest transaction in the results as stream_start_scn.
- if the combined query does _not_ return any transactions, take CURRENT_SCN - 1000 as stream_start_scn.
In both cases, use CURRENT_SCN as the snapshot_scn.

I think that, if v$transaction is non-empty, we're always good: by rewinding to the start of the oldest transaction, we'll catch every transaction that started after that, even ones that aren't in v$transaction. If v$transaction is empty, we play it safe and rewind a "reasonable" amount to cater for the potential lag in v$transaction contents.

**second idea (more complicated but probably more fool-proof)**
1. Start a transaction, and figure out the TXID
2. Poll v$transaction until the transaction we started in step 1 shows up in the view (with the combined query)
3. take the oldest start_scn from the query in step 2 as stream_start_scn and the CURRENT_SCN as snapshot_scn
4. roll back the transaction started in step 1

This basically ensures that v$transaction is never empty when we decide on the snapshot_scn, and my assumption is still that there is no problem with a non-empty v$transaction.

Naros:
Hi @chanetd, thanks for weighing in, a few comments:

> One approach would be to assume that the v$transaction lag is always low, say less than 1000 SCNs behind CURRENT_SCN.

I'm definitely not a fan of this as you probably would have guessed.  On Oracle 19, I believe Oracle is allowed to assign up to ~96k SCN values per second, so even a 1000 range wouldn't be adequate for a really high volume system.  We'd probably need something more like 100k to 300k to account for the potential lag not only of `V$TRANSACTION` but also the fact that redo buffer flushes happen on a 3s interval in the slowest of systems.  But even if we put the ~96k/sec SCN allocation, it's still just an arbitrary number that could fail given a specific set of circumstances and that's honestly not really any better than where we are right now using `V$TRANSACTION`.

> Start a transaction, and figure out the TXID

I actually considered this but the transaction must actually do something to cause a transaction to be created.  Just opening a connection and issuing a `BEGIN` isn't sufficient, we actually need to perform some operation over JDBC that triggers the actual creation of the database transaction context so that it is logged.

But I held back from doing this because I also thought about future support for logical/physical standby databases.  With logical, we can't create any transactions nor write to any tables when the database is open.  With physical, I believe it's possible for the DBA to assign per-table permissions to where some tables are locked and effectively read-only for SQL queries and others that are open and updatable because they're not part of the data replication pipeline from the primary.  I'm not a DBA but that's at least my assessment from the documentation.

So if we did create a transaction for this, that would eliminate logical standby support but I'm not necessarily sure that's really a problem since the connector can't really do anything while the database is unmounted and being used for shipping and applying primary database logs anyway.  You wouldn't be applying logs while it's open for read-only applications anyway, so there is no concept of in-progress transactions that would concern the connector in this environment anyway.  Furthermore, I'm not even sure if you could use LogMiner if the database was opened anyway in a read-only sense with physical standbys.

Do you see any concerns with my logical/physical standby assessment?  If it's accurate then perhaps this more complicated yet more stable solution is the right way forward after-all.

chanetd:
I can't comment on the logical/physical standby thing; that's way beyond my limited understanding of Oracle :-)
I agree that my first option is iffy, just wanted to put it on the table for completeness, but I agree with you that it's unsatisfactory.
I still like my second option, but the challenge is that you'll have to write something to the source database to trigger the transaction (even if you roll it back later). That means you need expanded permissions, and you need to figure out which table you can (fake-)write to. Doesn't seem so easy after all...

So if we want to avoid creating our own transaction, which reasonable lower bounds on the stream_start_scn are left?
- oldest scn in the logs: this will always work, but it's overkill -> you may need to trawl through thousands of logs for naught
- the scn of the last DDL statement (which is retrieved in OracleSnapshotChangeEventSource anyway): I _think_ this may work, but it may also be 1000s of SCNs in the past
- `TIMESTAMP_TO_SCN(NOW() - 10 minutes)` -> this is hardly better than my option 1 from above

We could also just wait for the v$transaction view to be non-empty. But that may cause long startup delays in a low-activity DB.

Or you could do a combination:
- wait for max 1 minute for v$transaction to be non-empty
- if v$transaction is still empty, set stream_start_scn to `TIMESTAMP_TO_SCN(NOW() - 10 minutes)`
It's not 100% fool-proof, but the window in which things could still go wrong is getting _very_ small here.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
```diff
@@ -143,26 +150,178 @@ protected void determineSnapshotOffset(RelationalSnapshotContext<OraclePartition
         }
 
         Optional<Scn> latestTableDdlScn = getLatestTableDdlScn(ctx);
-        Scn currentScn;
+        if (connectorConfig.getAdapter().getType().equals(LogMinerAdapter.TYPE)) {
```
jpechane, line 153: Can we have this moved to `AbstractStreamingAdapter` and its subclasses?

jpechane:
@Naros We ar eintroducing the heartbeat table support into Oracle connector. So how about combining both approaches? If heartbeat table is present/enabled then you have something to write to and create the necessary context. And if not then you'll continue with the current implementation.
The documentation can then describe advantages/disadvantages.

jpechane (changes requested):
Small refactoring and one question/comment asked

jpechane:
@Naros Have you seen my comments please?

Naros:
So @jpechane I think the write option will still cause issues because of how Oracle handles SCN assignment, where technically if the system is under load, multiple operations can be assigned the same SCN in a very small time window.  

So what I decided to do was to explore was this suggestion:
> oldest scn in the logs: this will always work, but it's overkill

But I decided to do this in an optimized way by working from the most current log and working backward to the oldest SCN.  Ultimately the logic is basically to try and find when does the Current SCN's transaction start.  This requires that we start a small mining session and iteratively walk back in time if and only if necessary to find the starting point.

In most cases, we should only ever have to walk back 1 log, which would be necessary when a transaction spans across a log switch boundary, but there is the corner case around long-running transactions that might require going back more.  With this approach, we support both situations and if we traverse all the way back and never find the starting SCN for such a transaction, it's reported in the logs as a warning.

I also opted to add a `log.mining.query.logs.for.snapshot.offset` option that is `true` by default as an internal option so that we could disable this experimental behavior for a user if it becomes problematic.  I also used this option, set to `false`, in the test suite to avoid any unnecessary time added to the tests other than those that are meant to test this use case of transactions spanning across snapshot boundaries.

I've run the two tests added for DBZ-5085 nearly 5 dozen times each and neither randomly fail.  @chanetd if you get a moment, feel free to give these changes a final review, but I'm fairly confident this is fine to be merged @jpechane if there are not outstanding code concerns you find.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerAdapter.java
```diff
@@ -74,4 +98,257 @@ public StreamingChangeEventSource<OraclePartition, OracleOffsetContext> getSourc
...
+            // While this should never be the case, this is to guard against NPE or other errors that could
+            // result from below if there is some race condition/corner case not considered
+            if (logFiles.isEmpty()) {
+                throw new DebeziumException("Failed to get log files from Oracle");
+            }
+
+            // Locate the index in the log files where we should begin
+            // This is the log where the current SCN exists
+            int logIndex = 0;
+            for (int i = 0; i < logFiles.size(); ++i) {
```
jpechane, line 193: @Naros Is it possible that the log would not exist?

PR #3457, opened 2022-05-05
DBZ-5092 Introduce a new extension api for query debezium version

harveyyue:
https://issues.redhat.com/browse/DBZ-5092


jpechane:
@harveyyue Applied, thanks

PR #3458, opened 2022-05-05
DBZ-5060 Upgrade postgres driver to version 42.3.5

ani-sha:
https://issues.redhat.com/browse/DBZ-5060


jpechane:
@ani-sha Applied, thanks

PR #3459, opened 2022-05-05
DBZ-5097: Add two new retryable postgresql error messages

markallanson:
These two error messages can occur when running inside docker and the
postgres container is either killed (sigtermed) or brought down
using `docker stop`.

In both cases these are retryable as the postgres server can be
restarted shortly thereafter.


markallanson:
@jpechane any chance you could approve the workflow again for me - I added some extra tests.

Naros (approved):
LGTM, thanks @markallanson

Naros:

PR #3461, opened 2022-05-06
DBZ-5090 Undo events across multiple mining sessions

Naros:
https://issues.redhat.com/browse/DBZ-5090


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
```diff
@@ -965,6 +970,38 @@ private class ParsedLobWriteSql {
...
+     * 4294967295 (aka -1 or 0xFFFFFFFF).
+     *
+     * This method explicitly checks if the provided transaction id has the no sequence sentinel
+     * value and if so, returns {@code true}; otherwise returns {@code false}.
+     *
+     * @param transactionId the transaction identifier to check, should not be {@code null}
+     * @return true if the transaction has no sequence reference, false if it does
+     */
+    protected boolean isTransactionIdWithNoSequence(String transactionId) {
+        return transactionId != null && transactionId.endsWith(NO_SEQUENCE_TRX_ID_SUFFIX);
```
jpechane, line 998: @Naros Is it possible that `transactionId` is `null`? If yes then I'd say the condition is not correct as in that case if will return `false` even if it is with no sequence.

jpechane:
@Naros LGTM, I've left one question. If this is a non-issue then just feel free to merge the PR yourself.

PR #3462, opened 2022-05-06
DBZ-5071 Allow configuration for custom columns for incremental snapshots

PlugaruT:


PlugaruT:
@Naros Here is a draft PR with the changes I did for this. Now, it needs a couple of improvements, but this is more of a starting point. Let me know what do you think and how this should be handled since you guys know better the codebase and practices. Thanks

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -616,4 +617,14 @@ protected Table refreshTableSchema(Table table) throws SQLException {
     private KeyMapper getKeyMapper() {
         return connectorConfig.getKeyMapper() == null ? table -> table.primaryKeyColumns() : connectorConfig.getKeyMapper();
     }
+
+    private List<Column> getQueryColumns(Table table) {
+        if (connectorConfig.getIncrementalSnapshotChunkKeyColumns() != null){
+            String cfg  = connectorConfig.getIncrementalSnapshotChunkKeyColumns();
+            return CustomKeyMapper.getInstance(cfg, x -> x.schema() + "." + x.table()).getKeyKolumns(table);
```
PlugaruT, line 624: The table mapper is specific to PostgreSQL, still thinking how to get this one to be database agnostic.

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -414,6 +414,16 @@ public static SchemaNameAdjustmentMode parse(String value) {
...
+    public static final Field INCREMENTAL_SNAPSHOT_CHUNK_KEY_COLUMNS = Field.create("incremental.snapshot.chunk.key.columns")
+            .withDisplayName("List of columns to be used by query during incremental snapshot.")
+            .withType(Type.STRING)
+            .withWidth(Width.LONG)
+            .withImportance(Importance.LOW)
+            .withValidation(Field::isListOfRegex)
+            .withDescription("This settings can be set to specify a list of columns for specific tables that would be used " +
+                    "to perform filtering when running an incremental snapshot.")
+            .withDefault(Boolean.FALSE);
+
```
Naros, line 426: This is probably the one aspect I'm a bit torn on.  I'm wondering if it would be better if this was something specified as a part of the incremental snapshot signal rather than a connector configuration proper.  I can possibly see an argument either way, so perhaps @jpechane can be our deciding vote :).

Naros (commented):
Overall, this looks really good, just one point for discussion.

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: Sorry but maybe I haven't had enough coffee, could you explain?

PlugaruT: The `getInstance` method needs an implementation for `TableIdToStringMapper` interface. I've noticed that for SQL Server and PG it's always used `x -> x.schema() + "." + x.table()`. Since this is an abstract class, I was thinking that this interface implementation should be sent when this class is being initialized maybe and not "hardcoded" since it might not apply for all implementations.

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
PlugaruT: Went with this way since it's a configuration that is not going to change that often, high probability that if it's being set, it won't change since users would like to have the queries every time they run an incremental snapshot. 
Anyway, open to suggestions.

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: No need for that, simply have a method that returns the lambda that most connectors use and then override that method for connectors that need a different implementation as we've done with other methods such as `preReadChunk` or `init` to name a few.

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
jpechane: @PlugaruT Please make this part of the signal payload. @Naros is right that this belongs to it.

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
jpechane: Just to make sure we are on the same page. If both key mapping and surrogate incremental snapshot column is used the the surrogate column is used for generating the SQL statements but the key mapper will be used to create keys in data messages, right?

PlugaruT: So, the usage here of `CustomKeyMapper` is a bit of cheating from my side, I needed all the logic from [here](https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/relational/Key.java#L112) in order to get the list of `Column` objects, from a string config, that would be used to build the SQL statements. Probably I should just create a new method for the same logic just to not be that confusing? This code does not change how the key mapper is used to create keys in data messages.

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
PlugaruT: Gotcha, I'll see how I should do that. Thanks for the input.

PlugaruT: @jpechane Looked at the code a bit, by adding this new config to the signal, it will mean that we would change the structure of the signal. Will that be ok?

jpechane: @PlugaruT If this will be done in backward compatible way (like new optional field) then there is no issue in that.

PlugaruT: @jpechane I will bother you with a few questions of how to implement this change as the signalling approach is a bit more tricky. 
From my understanding, this is what happens when an incremental snapshot get's triggered:
 1. Signal get's processed by the EventDispatcher that contains needed information for the snapshot, including the `data-collections` which represents the list of tables to be snapshoted.
 2. A `IncrementalSnapshotContext` object get's instantiated that holds the information about the snapshot, like max keys and other configurations for the snapshot. From my understanding, this context get's also stored into the signalling table because this context is re-used in case the connector is re-started while the snapshot is in progress. If not the entire context, then again, information about the chunk sizes and the limits.
 
From my understanding, this new configuration should also be stored inside the signalling table because it should be persistent for the entire duration of the snapshot as the signal that triggers the snapshot is only processed once. 

Did I get it right? Am I looking into the right direction stil?

I will do my best to implement this, but it will take a while until I get my head around of how the signalling architecture is implemented 😅 
Thanks

jpechane: @PlugaruT You've it almost right. The context *is* persisted but not in signalling table but in Kafka Connect offsets. See e.g. `io.debezium.connector.postgresql.PostgresOffsetContext.getOffset()`. That provides a an offset associated with each message sent to Kafka Connect which will then persist it.

PlugaruT:
I will close as I am  planning to open another PR when I will have the time to work on it. This is stale anyway.

PR #3463, opened 2022-05-06
DBZ-5099 Allow zero-length dates

vjuranek:
When date is zero and binary protocol use comperssion, date has zero
length [1]:

    if year, month, day, hour, minutes, seconds and micro_seconds are all 0, length is 0 and no other field is sent

Allow zero-lenght date and return it as a date which has all components
zero

[1] https://dev.mysql.com/doc/internals/en/binary-protocol-value.html#packet-ProtocolBinary::MYSQL_TYPE_TIME

https://issues.redhat.com/browse/DBZ-5099

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlBinaryProtocolFieldReader.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlBinaryProtocolFieldReader.java
@@ -8,6 +8,7 @@ package io.debezium.connector.mysql;
 import java.sql.Blob;
 import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.time.LocalDate;
 import java.util.Calendar;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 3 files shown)


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlBinaryProtocolFieldReader.java
```diff
@@ -74,6 +74,10 @@ protected Object readDateField(ResultSet rs, int columnIndex, Column column, Tab
         if (b == null) {
             return null; // Don't continue parsing date field if it is null
         }
+        else if (b.length() == 0) {
+            // Zero date has zero length when binary protocol uses compression.
+            return MySqlValueConverters.stringToLocalDate("0000-00-00", column, table);
```
jpechane, line 79: @vjuranek IMHO the result of this call will always be null. I suppose we should apply the same behaviour as in `io.debezium.connector.mysql.MySqlDefaultValueConverter.convertToLocalDate(Column, String)` and return Epoch time.

jpechane:
@vjuranek Applied, thanks

PR #3465, opened 2022-05-06
DBZ-5101 Adjust schema generator for MySQL for database.server.id

Naros:
https://issues.redhat.com/browse/DBZ-5101

* Removes default value assignment from field
* Removes auto-generation blurb from description

NOTE: This specifically targets 1.9 branch because we only want this behavior there.  For Debezium 2.0, this will be defaulted by the new `Field` definition introduced in DBZ-5100, so these two are mutually exclusive changes on two differing branches.


jpechane:
@Naros Applied, thanks

PR #3467, opened 2022-05-07
DBZ-5077 Support DROP FOREIGN KEY IF EXISTS grammar

AlexMiroshnikov:
Please see https://issues.redhat.com/browse/DBZ-5077 for problem description.


jpechane:
@AlexMiroshnikov Applied, thanks

PR #3468, opened 2022-05-08
DBZ-4780 Introduce support of LATERAL operator

AlexMiroshnikov:
Please see https://issues.redhat.com/browse/DBZ-4780 for problem statement.


jpechane:
@AlexMiroshnikov Can `LATERAL` be used as the name of a table or column? If yes, the please check rule `keywordsCanBeId`

Also for non-MariaDB specific PRs, could you please open PR also in upstream project - https://github.com/antlr/grammars-v4/tree/master/sql/mysql/Positive-Technologies

Thanks a lot for all the work done so far!

AlexMiroshnikov:
Great points! Thank you, I'm on it.

AlexMiroshnikov:
@jpechane So here's what we get in mysql 8:

```
mysql> create table lateral (lateral tinyint unsigned default null);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'lateral (lateral tinyint unsigned default null)' at line 1
mysql> create table `lateral` (lateral tinyint unsigned default null);
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'lateral tinyint unsigned default null)' at line 1

mysql> create table `lateral` (`lateral` tinyint unsigned default null);
Query OK, 0 rows affected (0.04 sec)

mysql> show tables;
+--------------+
| Tables_in_db |
+--------------+
| lateral      |
+--------------+
1 row in set (0.00 sec)

mysql> desc lateral;
ERROR 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'lateral' at line 1

mysql> desc `lateral`;
+---------+------------------+------+-----+---------+-------+
| Field   | Type             | Null | Key | Default | Extra |
+---------+------------------+------+-----+---------+-------+
| lateral | tinyint unsigned | YES  |     | NULL    |       |
+---------+------------------+------+-----+---------+-------+
1 row in set (0.01 sec)
```

TL;dr - `create table `lateral` (`lateral` tinyint unsigned default null);` allows to create a table named `lateral` with a field `lateral`, so I'm adding it to the check rule `keywordsCanBeId`.

Naros:
Hi @AlexMiroshnikov could you clean-up the commits to avoid a merge commit?  We don't particularly care for those in the history, so ideally there should be just the two commits here associated with DBZ-4780.

jpechane:
@AlexMiroshnikov Applied, thanks a lot!

PR #3470, opened 2022-05-09
oracle g4 file support case sensitive

saxisuer:
some oracle version like 11g or 10g  may have some sql not all uppercase chars when using logminer ,antlr4 will throw exception


Naros:
Hi @saxisuer, could you please provide an example that doesn't parse with the DDL parser with lower case?

From what I can see, we explicitly enable the `CaseChangingCharStream` pipeline stage for Oracle, which means we return all lexicons in upper-case which would match the grammar which is also written in upper-case.  Therefore, there should not be a situation where case impacts the parsing.  In fact, if you look at some of the examples in the parser's test suite for oracle, such as in `test\resources\oracle.examples\ddl_create.sql`, you will see some are written in lower case and pass just fine.

That's not to say there isn't a parser bug or issue with a specific parser statement, but I'm not necessarily convinced, at least not without a valid example, that there is a problem specifically with grammar case intepretation.

saxisuer:
It seems my own problem ..

PR #3473, opened 2022-05-09
DBZ-5108 Add ENABLED to the regular_id rule

Naros:
https://issues.redhat.com/browse/DBZ-5108

NOTE: Since DBZ-4641 wasn't pushed upstream, this isn't an issue in the upstream project.


jpechane:
@Naros Thanks

PR #3479, opened 2022-05-10
DBZ-5102 Server side database and collection filtering on MongoDB

luciocatinelli:


Naros:
Hi @luciocatinelli in case there is some confusion, the Bot is asking that you associate the same email address on your commits with the email you have associated with your GItHub account.  Could you do that for us?

jpechane:
@luciocatinelli Also could you please check the test failures? Thanks

luciocatinelli:
Thank you @jpechane! I'm not so experienced as java developer, I'm coming from the data engineering side using another languages so it's good all the feedback and help.
I need some time to understand those failures in the checks. 
But also I was thinking in a better idea depending on include/exclude list regex and **not depending on which database or collection exists at the moment when job starts**. The solution uses [regex ](https://www.mongodb.com/docs/manual/reference/operator/query/regex/)operator.
This is the code:
   
        String dbIncludeList = taskContext.filters().getDbIncludeList();
        String dbExcludeList = taskContext.filters().getDbExcludeList();
        String collectionIncludeList = taskContext.filters().getCollectionIncludeList();
        String collectionExcludeList = taskContext.filters().getCollectionExcludeList();

        Bson filters = Filters.in("operationType", getChangeStreamSkippedOperationsFilter());
        Bson collFilters;
        if (collectionIncludeList != null && !collectionIncludeList.trim().isEmpty()) {
            String collectionIncludeRegex = "("+collectionIncludeList.replace(",","|")+")";
            collFilters = Filters.and(filters,Filters.regex("dbcoll",collectionIncludeRegex));
        }
        else if (collectionExcludeList != null && !collectionExcludeList.trim().isEmpty()) {
            String collectionExcludeRegex = "("+collectionExcludeList.replace(",","|")+")";
            collFilters = Filters.and(filters,Filters.not(Filters.regex("dbcoll",collectionExcludeRegex)));
        }
        else {
            collFilters = filters;
        }

        Bson finalFilters;
        if (dbIncludeList != null && !dbIncludeList.trim().isEmpty()) {
            String dbIncludeRegex = "("+dbIncludeList.replace(",","|")+")";
            finalFilters = Filters.and(collFilters,Filters.regex("ns.db",dbIncludeRegex));
        }
        else if (dbExcludeList != null && !dbExcludeList.trim().isEmpty()) {
            String dbExcludeRegex = "("+dbExcludeList.replace(",","|")+")";
            finalFilters = Filters.and(collFilters,Filters.not(Filters.regex("ns.db",dbExcludeRegex)));
        }
        else {
            finalFilters =collFilters;
        }

        if (rsOffsetContext.lastResumeToken() == null) {
            // After snapshot the oplogStart points to the last change snapshotted
            // It must be filtered-out
            filters = Filters.and(finalFilters, Filters.ne("clusterTime", oplogStart));
        }

Do you agree that something like this is a better idea?

btiernay:
👋 Wanted to check in on the status of this PR to try to understand why it didn't end up being merged. This would be a great addition for both performance and cost reasons. When working with MongoDB Atlas and PrivateLink, there is the notion of [capacity units](https://www.mongodb.com/docs/atlas/billing/additional-services/#service-private-endpoint-capacity-units) where are a measure of traffic per hour for VPC activity.  This can _really_ add up, especially when you multiple connectors running in parallel 😅.

btiernay:
Starting a new PR here:
https://github.com/debezium/debezium/pull/4301

jcechace:
Closing this PR as the active work moved to #4301

PR #3482, opened 2022-05-11
DBZ-5105 Allow empty DBZ server config value

vjuranek:
An example when empty config value is meaningful is
`debezium.sink.kafka.producer.ssl.endpoint.identification.algorithm`
which means that server host name verification is turned off when
leave empty [1].

[1] https://docs.confluent.io/platform/current/kafka/authentication_ssl.html#optional-settings

https://issues.redhat.com/browse/DBZ-5105


vjuranek:
Failing Redis test is IMHO not related to this change, it fails for me locally even without this patch.

vjuranek:
Redis tests should be fixed by PR #3483

jpechane:
@vjuranek Applied, thanks

PR #3486, opened 2022-05-11
DBZ-3921: fix postgres existing publication is not updated with the n…

troeselereos:
…ew table

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -359,6 +359,7 @@ Thomas Aregger
 Thomas Deblock
 Thomas Prelle
 Thomas Thornton
+Timo Roeseler
 Tin Nguyen
 Tom Bentley
 Tom Billiet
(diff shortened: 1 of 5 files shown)


troeselereos:
Hi Chris, thank you for your feedback. I will also update the documentatin.
Timo

troeselereos:
Hi, my git skills are a bit poor. Is there any nice way to fix the commit messages?

Naros:
Hi @troeselereos you should be able to do an interactive rebase using `git rebase -i HEAD~3` and then select the commit with the message you want to reword and change `pick` to `r` for reword.  Once you save that screen it'll step you through the reapply process of each commit & allow you to change the commit message for those you specified.  Keep in mind this re-writes the commit history for these 3 commits, so you will need to force push afterward to your remote branch that this PR is based upon.  Let us know if you have any other questions.

jpechane:
@troeselereos Applied, thanks a lot!

mrMigles:
Hi,
@jpechane, when we can see this fix in one of stable 1.9 release? We faced the same issue and really need it for our production.

t0nem1n:
We have the same problem with 1.9 version, it would be easier if we have this fix in version 1.9

jpechane:
@t0nem1n Hi, we don't intend to do any other 1.9 release

PR #3491, opened 2022-05-12
DBZ-5124 moved to MySQL 5.7 in CI

jcechace:
[DBZ-5124](https://issues.redhat.com/browse/DBZ-5124)


jpechane:
@jcechace 5.7 is tested on github so I'd drop it as well.

PR #3492, opened 2022-05-12
DBZ-4251 Support stopping incremental snapshots

Naros:
https://issues.redhat.com/browse/DBZ-4251

@jpechane I haven't fully tested this completely yet, hence why it's a draft but I would love some early feedback on this.  I'm going to let the full CI run against this on GH to verify whether I've introduced any potential regressions so far.

- [x] Document new signal
- [x] Add tests for Db2


Naros:
Looks like MySQL and MSSQL failures are related to my two new tests, taking a look.

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotContext.java
```diff
@@ -175,6 +181,32 @@ public List<T> addDataCollectionNamesToSnapshot(List<String> dataCollectionIds)
...
+    }
+
+    @Override
+    public void applyPendingDataCollectionIdRemovals() {
+        if (!dataCollectionsToBeRemoved.isEmpty()) {
+            Iterator<T> iterator = dataCollectionsToBeRemoved.iterator();
+            while (iterator.hasNext()) {
+                final T dataCollectionId = iterator.next();
+                LOGGER.info("Removed '{}' from incremental snapshot collection list.", dataCollectionId);
+                dataCollectionsToSnapshot.remove(dataCollectionId);
```
jpechane, line 204: Here should be a log message that a data colection for which snapshot should be stopped is not present.

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotContext.java
```diff
@@ -78,6 +80,10 @@
 
     private boolean schemaVerificationPassed;
 
+    private boolean stopRequested;
```
jpechane, line 83: When is this flag cleared?

jpechane:
@Naros LGTM, thanks. I left two comments. This will require extensive test though. I'd like to see these scenarios

- Full stop and new snapshot executed after full stop
- Removing collection not snapshotted
- Removing collection currently snapshotted
- Removing all collections and new snapshot executed after it

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotContext.java
Naros: Great catch; I think adding a reset back to `false` inside `addDataCollectionNamesToSnapshot` is sufficient.  Would you agree?

Naros: I reworked this entirely.

Naros:
Hi @jpechane so after some investigation I was able to:

* Remove the boolean & the collection state variables I had originally introduced, and instead I now process the state upon the receipt of the stop signal.
* Introduced the complex tests requested for relational connectors & MongoDB.

I still need to send in a PR based on these changes for DB2; I'll do that Monday.

roldanbob (changes requested):

jpechane:
@Naros LGTM, but I'd wait with this after https://github.com/debezium/debezium/pull/3493 is merged as I'd say it would make sense to uses regex on the stop message too. WDYT?

Naros:
Associated DB2 connector PR: https://github.com/debezium/debezium-connector-db2/pull/52

Naros:
> I'd say it would make sense to uses regex on the stop message too. WDYT?

That's fine with me @jpechane.

Naros:
Test failures for SQL Server seem unrelated, and the DB2 failure is related to needing the DBZ-4251 PR merged in its repo.

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -434,6 +426,55 @@ public void addDataCollectionNamesToSnapshot(P partition, List<String> dataColle
...
+            else {
+                final List<String> expandedDataCollectionIds = expandDataCollectionIds(dataCollectionIds);
+                LOGGER.info("Removing '{}' collections from incremental snapshot", expandedDataCollectionIds);
+                for (String dataCollectionId : expandedDataCollectionIds) {
+                    final TableId collectionId = TableId.parse(dataCollectionId);
+                    if (currentTable != null && currentTable.id().equals(collectionId)) {
+                        window.clear();
+                        LOGGER.info("Removed '{}' from incremental snapshot collection list.", collectionId);
+                        tableScanCompleted(partition);
+                        nextDataCollection(partition);
```
jpechane, line 460: @Naros This will theoretically trigger the snapshot of the next table which can be afterward cancelled.

I wonder if the algortihm should work this way

1) Remove all tables for the list of snaphotted tables save the currently snapshotted
2) Stop the snapshot of the current table if the stopis requested
3) Stop the snapshot completely if no more tables are to be snapshotted

Naros: That does seem more deterministic. I'll make the adjustment.

Naros:
Adjustment to the order of operations when data collections are provided has been made, @jpechane.

documentation/modules/ROOT/pages/configuration/signalling.adoc
```diff
@@ -243,6 +244,47 @@ endif::community[]
...
+
+|data
+|`{"type":"INCREMENTAL", "data-collections": ["public.MyFirstTable"]}`
+
+|===
+
+You must specify the `type` of the signal.
+The `data-collections` field is optional.
+Leave the `data-collections` field blank to request the connector to stop all activity in the current snapshot.
+If you want the incremental snapshot to proceed, but you want to exclude specific collections from the snapshot, provide a comma-separated list of the names of the collections to exclude.
```
jpechane, line 286: @Naros The code now supports regexes, not only list. Please feel free to merge both PRs yourself when the doc is updated.

jpechane:
@Naros OK, now I see the problem in the docs more clearly. It is necessary to update also file https://github.com/debezium/debezium/blob/main/documentation/modules/ROOT/partials/modules/all-connectors/con-connector-incremental-snapshot.adoc that is inlined into each connector text.

jpechane:
@Naros There is a section `Triggering an incremental snapshot`. I think there should be a newsection after called `Stopping an incremental snapshot` that would deal with the details - so removing tables from the list, stopping everything etc.

Naros:
@jpechane added documentation section as requested.

jpechane:
@Naros Applied, thanks a lot!

PR #3493, opened 2022-05-12
DBZ-5123: Interpret data-collections signal as regex for ad-hoc snapshotting

Chrisss93:
https://issues.redhat.com/browse/DBZ-5123

This is a quick  attempt at redefining the data-collections-ids of a Debezium signal for ad-hoc snapshot to be a comma-separated list of regular expressions (similar to the`table.include.list` configuration parameter) instead of a comma-separated list of fully-qualified-names. Since this ad-hoc snapshotting signal and the `table.include.list` property appear to be complements, there seems to be value, at least to me, in bringing these concepts closer together by allowing them to be defined and interpreted in the same way. 

A few consequences of this:
1. the ad-hoc snapshotting no longer accepts _quoted_ fully-qualified-names (the `IncrementalSnapshotIT` class in the debezium-mysql-connector relied on this). Quoted fully-qualified-names do not work on the `table.include.list` property anyways.
2. While in most cases, users can continue to use the `data-collections` field for ad-hoc snapshotting as if it were still a fully-qualified name, this change might break existing behaviour for tables whose names are invalid regex (ie. a mysql table can be called `(` or a postgres table called `foo(`). In these cases, the offending regex characters must be escaped.

I'll add some stuff to the relevant docs in the MR once someone has given it a look over and tells me if I'm on the right track.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -71,6 +71,7 @@ Chris Baumbauer
 Chris Collingwood
 Chris Cranford
 Chris Egerton
+Chris Lee
 Chris Riccomini
 Christian Posta
 Christian Stein
--- a/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/IncrementalSnapshotIT.java
+++ b/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/IncrementalSnapshotIT.java
@@ -90,17 +90,34 @@ public class IncrementalSnapshotIT extends AbstractIncrementalSnapshotWithSchema
 
     @Override
     protected String tableName() {
-        return TableId.parse(DATABASE.qualifiedTableName("a")).toQuotedString('`');
+        return tableNameId().toQuotedString('`');
     }
 
     @Override
     protected String signalTableName() {
-        return TableId.parse(DATABASE.qualifiedTableName("debezium_signal")).toQuotedString('`');
+        return tableNameId("debezium_signal").toQuotedString('`');
     }
 
     @Override
     protected String tableName(String table) {
-        return TableId.parse(DATABASE.qualifiedTableName(table)).toQuotedString('`');
+        return tableNameId(table).toQuotedString('`');
+    }
+
+    @Override
+    protected String tableDataCollectionId() {
+        return tableNameId().toString();
+    }
+
+    private String dataCollectionName(String table) {
+        return tableNameId(table).toString();
+    }
+
+    private TableId tableNameId() {
+        return tableNameId("a");
+    }
+
+    private TableId tableNameId(String table) {
+        return TableId.parse(DATABASE.qualifiedTableName(table));
     }
 
     @Override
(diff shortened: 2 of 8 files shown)


jpechane:
@Chrisss93 Thanks for the PR. It is ddefinitely good start. We cannot get rid of quote support due to https://issues.redhat.com/browse/DBZ-4244hat also means that MySQL test database name should stay as is to verify exactly this scenario. So it is necesary to either keep the support or provide a very simple substitute for it.

jpechane:
@Chrisss93 Thanks for the PR. It is ddefinitely good start. We cannot get rid of quote support due to https://issues.redhat.com/browse/DBZ-4244hat also means that MySQL test database name should stay as is to verify exactly this scenario. So it is necesary to either keep the support or provide a very simple substitute for it.

Chrisss93:
> @Chrisss93 Thanks for the PR. It is ddefinitely good start. We cannot get rid of quote support due to https://issues.redhat.com/browse/DBZ-4244 That also means that MySQL test database name should stay as is to verify exactly this scenario. So it is necesary to either keep the support or provide a very simple substitute for it.

Thanks for the advice @jpechane , I've restored the test-case where database names may contain dashes and showed how one can still target that database with an ad-hoc snapshot signal in the new regex-based data-collections.

jpechane:
@Chrisss93 LGTM, could you please rebase the PR against the latest main and undraft it? Thanks

jpechane:
@Chrisss93 Tests are filing locally too - try `IncrementalSnapshotIT`. Could you please take a look at it?

jpechane (changes requested):
Waiting for tests to pass

Naros:
So the issue here is with the changes made to MySQL's `IncrementalSnapshotIT` with the rework of the return values for the various methods.  The tests are no longer getting the same expected table names in the right formats and that is creating the test failures.  @jpechane is there a specific reason why MySQL required the use of the double-quoted table name format?

Naros:
Test failures fixed; @Chrisss93 please take a look at my changes and let me know if you see any issues with what I changed.

chrisVia:
> Test failures fixed; @Chrisss93 please take a look at my changes and let me know if you see any issues with what I changed.

Changes look good to me - at the end of the day this is more about a bit of mysql-connector test quirkiness than the functional behaviour of this ticket. Sorry for the failures, I was using the -nsu flag with Maven so I didn't catch all of them locally after some iterations.

Naros:
No worries at all, I'll let @jpechane give this one additional pass from his perspective since I know he wanted to merge this before my stop signal work.  @jpechane regarding the stop-signal PR #3492, once this is merged I'll rebase and reverify.  There is a few refactor changes I can make in my PR that tidy up both my work with the work done here.

Naros:
@Chrisss93 would you mind giving this PR a rebase against `main` if you have moment before the weekend?

jpechane:
@Chrisss93 @Naros Applied, thanks guys!

Chrisss93:
@jpechane Apologies, I could not locally reproduce the SQLServer connector test failures.

PR #3495, opened 2022-05-13
DBZ-5129: Use buffered reader when reading history from file

morozov:


morozov:
The test failure on MySQL seems unrelated. `IncrementalSnapshotIT` passes locally, and the entire test suite passed during the [previous run](https://github.com/debezium/debezium/runs/6416111096?check_suite_focus=true).

jpechane:
@morozov Applied, thanks!

PR #3498, opened 2022-05-13
DBZ-5126 Clear tableMapEventByTableId on rotate event

Jiabao-Sun:
Clear tableMapEventByTableId on rotate event to prevent it from growing indefinitely.


jpechane:
@Jiabao-Sun Applied, thanks!

PR #3499, opened 2022-05-13
Move DB name logging support from SQL Server into core

vjuranek:
https://issues.redhat.com/browse/DBZ-5042


Naros:
@vjuranek is this one ready for review or is there anymore work needed here?

vjuranek:
@Naros it should be ready.

vjuranek:
Thanks for noticing it @Naros ! I completely missed it. 

Now I found out that oracle connector is by default not included and one has to activate `oracle` profile, otherwise Idea ignores whole module in class search.

Naros:
Hi, @vjuranek that's correct, it's still an opt-in connector.  We should probably consider changing that in 2.0; I'll add a todo for that.

Naros:
Hi @vjuranek I pushed one small commit to favor PDB name when set rather than the CDB name.  When CI passes, I'll merge.

Naros (approved):

PR #3501, opened 2022-05-16
DBZ-5131 Use fetch size only in tests which need it

vjuranek:
Cuncurrent reading and modification of MySQL table can result into MySQL
JDBC driver ArrayIndexOutOfBoundsException when it tried to merge table
fields. This is result fo caching parepared statements while table
schema is modified. It seems that caching of prepared statements is
triggered by using fetch size. Use it only in test which needs this
option to work properly and which don't do any schema changes.

https://issues.redhat.com/browse/DBZ-5131


vjuranek:
In PR #3499 failed several times in a row, now passed 5 times without any issue (for the first time it failed with unrelated `io.debezium.connector.mysql.zzz.ZZZGtidSetIT#shouldProcessPurgedLogsWhenDownAndSnapshotNeeded`) so looks like this fixes the issue.

jpechane:
@vjuranek Applied, thanks for finding the root cause

PR #3502, opened 2022-05-17
DBZ-5137: MilliSecondsBehindSource is not reported by SQL Server connector

ramanenka:
Call StreamingMeter.onEvent for every event so that it reports the
metrics correctly.


jpechane:
@ramanenka Applied, thanks!

PR #3504, opened 2022-05-17
DBZ-4867 Do not set `database.dbname` for SQL Server

Naros:
https://issues.redhat.com/browse/DBZ-4867

@jcechace can you review and merge if this looks okay to you?
@jpechane should we consider backporting this to 1.9?


jpechane:
@Naros Applied and yes I'll backport it

PR #3505, opened 2022-05-17
DBZ-5089: Add partition to outbox event router

connorszczepaniak-wk:
[DBZ-5089](https://issues.redhat.com/browse/DBZ-5089)

In some use cases, it would be useful to be able to specify on which partition an outbox event should be produced. This is currently not possible as `null` is hardcoded into the record constructor - let's add it to the config and plumb it down.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -82,6 +82,7 @@ Claus Guttesen
 Clément Loiselet
 Cliff Wheadon
 Collin Van Dyck
+Connor Szczepaniak
 Cory Harper
 Cyprien Etienne
 Cyril Scetbon
--- a/debezium-core/src/main/java/io/debezium/transforms/outbox/EventRouterConfigDefinition.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/outbox/EventRouterConfigDefinition.java
@@ -65,7 +65,8 @@ public class EventRouterConfigDefinition {
 
     public enum AdditionalFieldPlacement implements EnumeratedValue {
         HEADER("header"),
-        ENVELOPE("envelope");
+        ENVELOPE("envelope"),
+        PARTITION("partition");
 
         private final String value;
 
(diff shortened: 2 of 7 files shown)


Naros:
Hi @connorszczepaniak-wk there appear to be some test failures related to this change.  Can you take a look?

jpechane:
@connorszczepaniak-wk The code LGTM. I've triggered another test run. Could you please also prepare a docs change? Also I think you can squash the commits into a singla one. Thanks a lot.

jpechane:
@connorszczepaniak-wk I've just re-read documentation and have a second thought about the implementation. How about using https://debezium.io/documentation/reference/1.9/transformations/outbox-event-router.html#emitting-messages-with-additional-fields and introduce `partition` placement? It would be better aligned with the other configs.

connorszczepaniak-wk:
@jpechane I've put the partition into additional field placement. Much simpler now; if it looks good to you, can we kick off a CI run?

jpechane:
@connorszczepaniak-wk Good job, the code is done! I rebased your PR on the latest main and added a commit that uses `AtomicReference` to make it consistent with other instances of the similar code in the codebase. Beware! I force pushed into your branch.

Could you please just update docs in https://github.com/debezium/debezium/blob/main/documentation/modules/ROOT/pages/transformations/outbox-event-router.adoc and https://github.com/debezium/debezium/blob/main/documentation/modules/ROOT/pages/transformations/mongodb-outbox-event-router.adoc

After that the PR can be merged.

connorszczepaniak-wk:
@jpechane I've added the documentation change as well. Quick question about Debezium releases -- will this change get back-ported to older releases, or will it only be available in 2.0?

jpechane:
@connorszczepaniak-wk Applied, thanks a lot. Givent the scope of the change and longer 2.0 dev cycle I am going to backport this to 1.9 too

PR #3507, opened 2022-05-18
DBZ-5141 Should be sleep with REGISTRATION_RETRY_DELAY when registry MBean failed

harveyyue:
https://issues.redhat.com/browse/DBZ-5141

### DIFF
--- a/debezium-core/src/main/java/io/debezium/metrics/Metrics.java
+++ b/debezium-core/src/main/java/io/debezium/metrics/Metrics.java
@@ -25,7 +25,9 @@ import org.slf4j.LoggerFactory;
 import io.debezium.annotation.ThreadSafe;
 import io.debezium.config.CommonConnectorConfig;
 import io.debezium.connector.common.CdcSourceTaskContext;
+import io.debezium.util.Clock;
 import io.debezium.util.Collect;
+import io.debezium.util.Metronome;
 
 /**
  * Base for metrics implementations.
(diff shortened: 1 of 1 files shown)


jpechane:
@harveyyue Good catch! I had to be out of my mind yesterday :-(.

debezium-core/src/main/java/io/debezium/metrics/Metrics.java
```diff
@@ -86,21 +89,23 @@ public synchronized void register() {
...
-                    if (attempt <= REGISTRATION_RETRIES) {
+                    if (attempt < REGISTRATION_RETRIES) {
                         LOGGER.warn(
                                 "Unable to register metrics as an old set with the same name exists, retrying in {} (attempt {} out of {})",
                                 REGISTRATION_RETRY_DELAY, attempt, REGISTRATION_RETRIES);
+                        metronome.pause();
                     }
                     else {
-                        LOGGER.error("Failed to register metrics MBean, metrics will not be available");
+                        LOGGER.error("Failed to register metrics MBean, metrics will not be available, retrying in {} (attempt {} out of {})",
```
jpechane, line 99: PLease return back the original message. The retry should not happen after {{else}} branch.

debezium-core/src/main/java/io/debezium/metrics/Metrics.java
```diff
@@ -72,6 +74,7 @@ protected Metrics(CommonConnectorConfig connectorConfig, String contextName, boo
      */
     public synchronized void register() {
         try {
+            final Metronome metronome = Metronome.sleeper(REGISTRATION_RETRY_DELAY, Clock.system());
```
jpechane, line 77: Could you please resturcture the code to mimic https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java#L345 ?

debezium-core/src/main/java/io/debezium/metrics/Metrics.java
harveyyue: @jpechane, welcome. Got it.

debezium-core/src/main/java/io/debezium/metrics/Metrics.java
harveyyue: Done.

jpechane:
@harveyyue Could you please return back the `<` you used in the condition? I belive you were correct on that one change.

harveyyue:
> @harveyyue Could you please return back the `<` you used in the condition? I belive you were correct on that one change.

Had returned to `<`

PR #3510, opened 2022-05-18
Create SqsChangeConsumer in Debezium server

ichasepucks:


Naros:
Hi @ichasepucks a couple of things:

1. Please raise a jira issue if one does not already exist for this and then update the commit message with `DBZ-xxx` as a prefix.
2. We accept commits mostly to `main` and never to legacy branches.  Couldy you please target `main` rather than `1.9`?

ichasepucks:
@Naros so sorry!  I am not even close to ready to open this PR back to the upstream.  I meant to create this on my fork so I could share with my team.  If/when we get to the point of wanting to contribute back I'll look to follow all the right steps.  Thanks!

PR #3514, opened 2022-05-18
DBZ-4713 Update example, remove callouts as text isn't meant for copying

roldanbob:
[DBZ-4713](https://issues.redhat.com/browse/DBZ-4713)

Updates schema change event example in MySQL connector doc, based on the output that was generated on my local deployment. 

I've removed the callout comments (`//  <_n_>`) in the example, because this content isn't intended for copy/paste, so I don't think we have to worry about the extra characters causing problems if readers paste the example code into a code editor.

@jpechane Per our earlier exchange in DBZ-3390, I think that the changes here also satisfy that issue, but let me know if you think anything is still outstanding.

This change should be backported to 1.9.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/mysql.adoc
+++ b/documentation/modules/ROOT/pages/connectors/mysql.adoc
@@ -179,107 +179,105 @@ The message contains a logical representation of the table schema.
 [source,json,indent=0,subs="+attributes"]
 ----
 {
-  "schema": {
-  ...
-  },
+  "schema": { },
   "payload": {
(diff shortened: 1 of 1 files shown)


Naros:
These changes look fine to me @roldanbob.  I let @jpechane respond to your comment before merging.

jpechane:
@roldanbob Applied, thanks a lot!

jpechane:
@roldanbob This cannot be applied to 1.9 cleanly. Could you please take a look? Thanks

PR #3516, opened 2022-05-19
[docs] Fixed default value of debezium-source-offset-storage

Himanshu-LT:
Default value for "[debezium.source.offset.storage]" was mentioned in "[debezium.source.offset.storage.file.filename]". I have fixed the value.


Naros:
Excellent catch @Himanshu-LT, thanks!

Naros (approved):

PR #3519, opened 2022-05-20
DBZ-5136: Add last in data collection flag

ajunwalker:
https://issues.redhat.com/browse/DBZ-5136

The PR adds a new a new `LAST_IN_DATA_COLLECTION` in the source information block, specifically for snapshot mode.
To achieve this, I leveraged the `snapshotContext.lastRecordInTable` field to check whether the record is the last in table, and created a new method within the `OffsetContext` interface which updates the snapshot context to set `snapshot=LAST_IN_DATA_COLLECTION`.


jpechane:
@ajunwalker Thanks for the PR! Have you please read my comment on Jira proposing additional snapshot state flags?

Also I wonder if it makes sense to refactor {{OffsetContext}} classes and introduce a common parent class to avoid the code duplication. It seems to me that it would make sense not onyl for the new code but also for some of the existing one.

ajunwalker:
@jpechane I have made the requested changes, would appreciate another look! I refactored some existing code as well, specifically in the `RelationalSnapshotChangeEventSource` class.

debezium-core/src/main/java/io/debezium/relational/RelationalOffsetContext.java
```diff
@@ -0,0 +1,43 @@
+/*
```
ajunwalker, line 1: This class consolidates all the duplicated code from before.

debezium-core/src/main/java/io/debezium/relational/RelationalOffsetContext.java
```diff
@@ -0,0 +1,43 @@
...
+ */
+package io.debezium.relational;
+
+import io.debezium.connector.SnapshotRecord;
+import io.debezium.connector.common.BaseSourceInfo;
+import io.debezium.pipeline.spi.OffsetContext;
+
+public abstract class RelationalOffsetContext implements OffsetContext {
+
+    public abstract BaseSourceInfo getSourceInfoObject();
```
ajunwalker, line 14: `SourceInfo` was taken and was specific to each connector so I had to create a common interface to extract it from their respective implementation classes.

jpechane:
@ajunwalker Very nice work! Would it be possible to introduce `CommonOffsetContext` from which both MongoDB and relational offset contexts would inherit? If yes then it would be aligned with connector config hierachy too.

ajunwalker:
@jpechane Thank you for the swift reviews! I ended up just renaming `RelationalContextOffset` to `CommonOffsetContext` just to simply things.

debezium-core/src/main/java/io/debezium/relational/CommonOffsetContext.java
```diff
@@ -0,0 +1,43 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.relational;
```
jpechane, line 6: This does not belong to the `relational` package.

jpechane:
@ajunwalker I've just moved the `CommonOffsetContext` into a different package in the new commit. Please take a look at db2/vitess repos and prepare PRs that will make them compatible with this change. I'll then merge all PRs at the same time. Good work!

alwaysbemark:
Heya @jpechane ! I'm working with Andrew on this PR (we're part of the same team!) and I made some changes that hopefully streamline this PR even further. I removed the need for duplication/extra classes and appeased all of the ITs.

Moreover, here are the PRs for DB2 and Vitess:
https://github.com/debezium/debezium-connector-db2/pull/54
https://github.com/debezium/debezium-connector-vitess/pull/82

Would love another glance at the changes and a merge! Thank you again so much!

alwaysbemark:
Heya @jpechane  ! I'm working with Andrew on this PR (we're part of the same team!) and I made some changes that hopefully streamline this PR even further. I removed the need for duplication/extra classes and appeased all of the ITs. There's a few places in the ITs where this logic is validated which should also increase confidence in the change!

Moreover, here are the PRs for DB2 and Vitess:
https://github.com/debezium/debezium-connector-db2/pull/54
https://github.com/debezium/debezium-connector-vitess/pull/82

Would love another glance at the changes and a merge! Thank you again so much!

alwaysbemark:
@jpechane Absolutely! Brought back the common class and all the "easy" common functionality into it. I will definitely commit to doing a second pass and refactoring even more functionality into it, but the changes I'd like to make touch a few other files and I would prefer to not overload this PR with even more changes.

I updated DB2 and Vitess with these latest changes as well:
https://github.com/debezium/debezium-connector-db2/pull/54
https://github.com/debezium/debezium-connector-vitess/pull/82

Thanks once again!

alwaysbemark:
Thank you!

PR #3522, opened 2022-05-20
DBZ-4877 Upgrade Oracle InstantClient to 21.5.0.0

Naros:
https://issues.redhat.com/browse/DBZ-4877


Naros:
There is a subsequent website PR https://github.com/debezium/debezium.github.io/pull/801 that should be merged after the next release to align the versions of the supported Oracle drivers & Instant clients.

Naros:
@jcechace just wanted to ping you so that you are aware if we have the Oracle instant client cached on our internal Jenkins CI we'll need to account for that upgrade as well as a part of this change too.

PR #3523, opened 2022-05-20
DBZ-5011 Register MongoDB metrics with task id

Naros:
https://issues.redhat.com/browse/DBZ-5011

When starting the MongoDB connector with more than one task, the connector fails to start the additional tasks due to the MBean registration failure due to clashing names.  This PR adds a `task` scope to each registered MBean to include the associated Kafka task identifier assignment for each running task.


Naros:
Test failure is related to https://issues.redhat.com/browse/DBZ-4004.

Naros:
@Bonndan can you share your connector setup on the associated Jira and let me take a look to see if there might be a gap in the SQL Server code as well.

bonndan:
Et voila. The files have been anonymized. Hope this is helpful.

[my_connector2.json.txt](https://github.com/debezium/debezium/files/8771910/my_connector2.json.txt)
[connect-distributed.properties.txt](https://github.com/debezium/debezium/files/8771911/connect-distributed.properties.txt)
[my_connector1.json.txt](https://github.com/debezium/debezium/files/8771912/my_connector1.json.txt)

Naros:
Hi @bonndan it would seem based on the connector configurations you're using the same value for `database.server.name` in both connectors, which is not a valid configuration.  The `database.server.name` must be unique across all connectors.

bonndan:
HI @Naros thank you very much!

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
```diff
@@ -551,6 +551,11 @@ public boolean isFullUpdate() {
             .withValidation(Field::isInteger)
             .withInvisibleRecommender();
 
+    public static final Field MAX_TASKS = Field.createInternal("mongodb.max.tasks")
```
jpechane: @Naros I am a bit confused, why do we need e new config option?

jpechane:
@Naros Could you please also update the docs?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
Naros: Technically yes, but we don't explicitly set `tasks.max` anywhere in the test suite, and it's explicitly passed as a hardcoded value inside of `EmbeddedEngine#run`.  I can most definitely look for `tasks.max` and if not found assume `1` rather than passing the value this way from the task configuration method if you'd prefer.

Naros: I reworked the first commit to using `tasks.max` instead, defaulting to `1` if the property isn't passed.

jpechane:
@Naros Applied, thanks!

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/metrics/MongoDbSnapshotChangeEventSourceMetrics.java
```diff
@@ -27,11 +27,9 @@ public class MongoDbSnapshotChangeEventSourceMetrics extends DefaultSnapshotChan
     private AtomicLong numberOfDisconnects = new AtomicLong();
 
     public <T extends CdcSourceTaskContext> MongoDbSnapshotChangeEventSourceMetrics(T taskContext, ChangeEventQueueMetrics changeEventQueueMetrics,
-                                                                                    EventMetadataProvider metadataProvider, int maxTasks) {
+                                                                                    EventMetadataProvider metadataProvider) {
         super(taskContext, changeEventQueueMetrics, metadataProvider,
-                maxTasks > 1
-                        ? Collect.linkMapOf("context", "snapshot", "server", taskContext.getConnectorName(), "task", taskContext.getTaskId())
-                        : Collect.linkMapOf("context", "snapshot", "server", taskContext.getConnectorName()));
+                Collect.linkMapOf("context", "snapshot", "server", taskContext.getConnectorName(), "task", taskContext.getTaskId()));
```
jchipmunk, line 32: Hi, @Naros 

thanks for tackling this issue.

I'm concerned that I don't see some changes related with task id initialization.
From my perspective, `taskContext.getTaskId()` always returns `0` because `MongoDbTaskContext` class calls superclass constructor without specifying the task id: 
```
public MongoDbTaskContext(Configuration config) {
    super(Module.contextName(), config.getString(MongoDbConnectorConfig.LOGICAL_NAME), Collections::emptySet);
```
```
public CdcSourceTaskContext(String connectorType, String connectorName, Supplier<Collection<? extends DataCollectionId>> collectionsSupplier) {
    this(connectorType, connectorName, "0", collectionsSupplier);
}
```
`MongoDbConnectorConfig.TASK_ID` can be used to get the task id from configuration.

PR #3525, opened 2022-05-21
DBZ-3865 Debezium MySql connector does not capture floating point numbers with the right precision

harveyyue:
https://issues.redhat.com/browse/DBZ-3865


jpechane:
@harveyyue Hi, thanks for the PR. As we are already in 2.0 development, would it make sense to change the default mapping as suggested in Jira? In that case a docs update would also be required.

harveyyue:
> a docs

Hi @jpechane , `For Debezium 2.0 (whenever that'll be), we might revisit the default mapping used here`, how does revisit the default mapping? Please help to indicate it, Thanks.

jpechane:
Right now `MySqlValueConverter` follows the rules of https://debezium.io/documentation/reference/1.9/connectors/mysql.html#mysql-basic-types so `FLOAT` is mapped to `double`.
PLease see https://issues.redhat.com/browse/DBZ-3865?focusedCommentId=18831757&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-18831757 for more details
In general this PR is intended for 2.0 and should change the default behaviour instead of providing a converter that could be enabled as needed.

harveyyue:
Hi @jpechane , Got it, thanks for clarifying it. Users need to handle this incompatible case from double to float in schema registry.

jpechane:
@harveyyue Yes, I am aware of that but given the major version change we can be a bit forceful.

documentation/modules/ROOT/pages/connectors/mysql.adoc
```diff
@@ -1381,9 +1381,9 @@ a|_n/a_
 |`FLOAT32`
 a|_n/a_
 
-|`FLOAT[(M,D)]`
-|`FLOAT64`
-a|_n/a_
+|`FLOAT(P)` or `FLOAT[(M,D)]`
+|`FLOAT32` or `FLOAT64`
+a|The precision `P` from 0 to 23 results to `FLOAT32`, from 24 to 53 results to `FLOAT64`. As of MySQL 8.0.17, the nonstandard FLOAT(M,D) and DOUBLE(M,D) syntax is deprecated, and should expect support for it be removed in a future version of MySQL, so set to `FLOAT64` as default.
```
jpechane, line 1386: Could you please take a look at https://debezium.io/documentation/reference/1.9/connectors/oracle.html#oracle-numeric-types and format/descibe it in a similar way so it is consistent accross different connectors?

jpechane (approved):
@harveyyue Nice work, thanks! I've just left a small comment on the docs. When addressed the issue can be merged.

documentation/modules/ROOT/pages/connectors/mysql.adoc
harveyyue: done

Naros:
Thanks, @harveyyue and @jpechane, applied to `main` (=2.0).

PR #3526, opened 2022-05-22
DBZ-5152 LogMiner catch-up efficiency improvement

elgca:
If startscn is 30min delayed from the current time. 
Only the archive logs are loaded, during which a restart of LogMiner is not triggered until these archive logs are processed

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
@@ -12,6 +12,7 @@ import java.sql.SQLException;
 import java.text.DecimalFormat;
 import java.time.Duration;
 import java.time.Instant;
+import java.time.LocalDateTime;
 import java.time.OffsetDateTime;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
@@ -22,6 +23,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.TimeUnit;
 import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
@@ -84,6 +86,7 @@ public class LogMinerStreamingChangeEventSource implements StreamingChangeEventS
     private Scn snapshotScn;
     private List<LogFile> currentLogFiles;
     private List<BigInteger> currentRedoLogSequences;
+    private boolean catchingUp = false;
 
     public LogMinerStreamingChangeEventSource(OracleConnectorConfig connectorConfig,
                                               OracleConnection jdbcConnection, EventDispatcher<OraclePartition, TableId> dispatcher,
@@ -172,6 +175,15 @@ public class LogMinerStreamingChangeEventSource implements StreamingChangeEventS
                             continue;
                         }
 
+                        // If endScn exceeds the maximum scn of the current log, give up catch-up
+                        // If catchingUp is set to false then hasLogSwitchOccurred must be true, will reload the log file to logminer
+                        if (catchingUp) {
+                            Scn maxArchiveLogScn = getMaxArchiveLogScn(currentLogFiles);
+                            if (endScn.compareTo(maxArchiveLogScn) >= 0) {
+                                catchingUp = false;
+                            }
+                        }
+
                         flushStrategy.flush(jdbcConnection.getCurrentScn());
 
                         boolean restartRequired = false;
@@ -187,7 +199,7 @@ public class LogMinerStreamingChangeEventSource implements StreamingChangeEventS
                             }
                         }
 
-                        if (restartRequired || hasLogSwitchOccurred()) {
+                        if (restartRequired || (!catchingUp && hasLogSwitchOccurred())) {
                             // This is the way to mitigate PGA leaks.
                             // With one mining session, it grows and maybe there is another way to flush PGA.
                             // At this point we use a new mining session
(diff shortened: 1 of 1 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -328,25 +340,48 @@ private void initializeRedoLogsForMining(OracleConnection connection, boolean po
...
+     * @param connection database connection, should not be {@code null}
+     * @param startScn mining session's starting system change number (exclusive), should not be {@code null}
+     * @throws SQLException if a database exception occurred
+     */
+    private void addLogFileForMining(OracleConnection connection, Scn startScn) throws SQLException {
+        // If the time of startscn is 30 minutes later than the current time.
+        // The catch-up mode will be used. No need to load online redo log
+        final long onlineWindow = TimeUnit.MINUTES.toMillis(30);
+        LocalDateTime currentLocalTime = LocalDateTime.now();
+        Optional<LocalDateTime> startTime = connection.getScnToTimestamp(startScn).map(OffsetDateTime::toLocalDateTime);
```
Naros, line 370: I'm a bit concerned with this usage here.

The default `undo_retention` database parameter is 900 seconds (15 minutes).  In my experience, I don't rely on users having changed many of these default values unless they've hit a situation where they need to, so on a basic installation of Oracle, you'll never hit a situation where this "catch-up" logic will fire.  This is because after the 15 minutes, the SCN has aged out of the Flashback Area and the `Optional` value will not be present.

I wonder if this "30" minute value should be defined as an internal configuration property.  This way we don't need to document the property but gives us a way to tune the behavior if someone's environment needs some adjustments due to the introduction of this feature.

Also, did you happen to test this scenario specifically using `log.mining.strategy=online_catalog` or did you also test this using the default mining strategy?  I'm specifically concerned when using the default strategy that if none of the archive logs being mined has the data dictionary, will Oracle LogMiner fail?

Naros (commented):
Hi @elgca, overall this looks good, but I have one main comment inline, see below.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
Naros: Hi @elgca, just a gentle ping if you can answer my question above.

elgca: I am using `log.mining.strategy=online_catalog` for testing，not fail. In my case, scn is 3 hours behind. It was running in catch up until I caught up with the scn in the last 30 minutes.

30 minutes is not a strict number, just the number of archivelogs in 30 minutes will not waste too much time in add_logfile, for me, so I tend to keep the original logic.

elgca: @Naros

elgca: image

In fact, my archivelog interval is much less than 30min

Naros: Hi @elgca its not so much about the archive log interval rate but what is your environment's undo retention.  You could check this in your environment by running this query:
```
SELECT name, value from v$parameter where name = 'undo_retention';
```

Naros (changes requested):
This change is now compatible with the default `log.mining.strategy`.  It would seem that we are now generating a data dictionary dump on each iteration which is unacceptable.  I'm looking at ways to rework this so that the default strategy can also benefit from this without the negative side effects.

elgca:
@Naros 
My modification has been running for a few days now. the connector will have a data loss problem when running on a rac cluster. I think it is caused by `getMaxArchiveLogScn` . Since my oracle log has two relatively independent archivelog threads and crossed scn ranges, This method, `getMaxArchiveLogScn`,  will cause the one of the log files to be lost a scn range. like this:

|NAME                                       |FIRST_CHANGE#|FIRST_TIME             |NEXT_CHANGE# |NEXT_TIME              |THREAD#|SEQ |
|:-------------------------------------------:|:-----------:|:----------------------:|:------------:|:----------------------:|:------:|:----:|
|+DATA/prod/archivelog/2_4460_1104612962.dbf|9273075693039|2022-05-29 23:59:31.000|9273075866240|2022-05-30 00:00:22.000|      2|4460|
|+DATA/prod/archivelog/2_4459_1104612962.dbf|9273075628805|2022-05-29 23:58:34.000|9273075693039|2022-05-29 23:59:31.000|      2|4459|
|+DATA/prod/archivelog/1_7504_1104612962.dbf|9273075533298|2022-05-29 23:57:31.000|9273075876853|2022-05-30 00:00:25.000|      1|7504|
|+DATA/prod/archivelog/2_4458_1104612962.dbf|9273075527660|2022-05-29 23:57:28.000|9273075628805|2022-05-29 23:58:34.000|      2|4458|

Naros:
Hi @elgca so in a RAC environment, it's normal for the various threads to have overlapping SCN values.  This is because the SCN assignment is global across all nodes in the cluster so one node may get SCN 1, another SCN 2, and then the previous node gets SCN 3.  This really all depends entirely on where the work is being performed on the cluster.  

What I don't understand is the data loss comment, we have users using Oracle RAC today with no problem so I'm a bit confused on how you deduced that there will be some level of data loss?

Naros:
Hi @elgca, can you check 1.9.7.Final and see whether or not this is still an issue?  We made some improvements with Oracle RAC and I'd like to know whether we need anything else regarding this?

Naros:

Hi @elgca, please sign off all commits with:

```
git commit -s
```

Naros:
Hi @elgca, this has been reworked in Debezium 3.5 and 3.4. The new logic for catch-up scopes the logs to the mining window only and is applied only when using `online_catalog` or `hybrid`. The new approach renders these changes moot, but thank you for the PR.

PR #3527, opened 2022-05-23
Filter event and Error event don't need to trigger 'updateCommonEventMetrics' method

camellibby:
Maybe updateCommonEventMetrics should be triggered when updating 'table.include.list' in configuration. Users don't care about excluding table metrics.


jpechane:
@camellibby Hi, could you please also remove the merge commit? Thanks

PR #3532, opened 2022-05-24
DBZ-5119 Move Heartbeat creation to ConnectorConfig & descendants

Naros:
https://issues.redhat.com/browse/DBZ-5119

@jpechane I reused the same issue number as the prior PR's work since we haven't released this yet.  If you'd rather this be a new Jira issue entirely I can certainly add a new follow-up and adjust this PR accordingly.


Naros:
So there is clearly some stability issues with this change, I'm looking into it.

PR #3537, opened 2022-05-25
DBZ-5052 Don't send tombstones when consumer doesn't support them

vjuranek:
https://issues.redhat.com/browse/DBZ-5052


jpechane:
@vjuranek LGTM, could you please reformat the code?

vjuranek:
@jpechane should be ready now

vjuranek:
btw: current version adds one more small change in debezium server when embedded engine is constructed (it's a pity GH still haven't any function which would show the diff between PR versions - already reviewed vs. new version)

jpechane:
@vjuranek Applied, thanks!

PR #3540, opened 2022-05-25
DBZ-5147 Add `event.processing.failure.handling.mode` for Oracle

Naros:
https://issues.redhat.com/browse/DBZ-5147

- [x] ~Implement behavior for Xstreams~
- [x] ~Documentation~

@jpechane wrt Xstream, this would be mostly to handle LcrHandler callback errors and prefer to log/skip if the processing mode is set to warn/skip or fail as we do now by default.  Do you think it's warranted or should this mostly focus on LogMiner?

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
@@ -50,6 +50,7 @@ import io.debezium.pipeline.source.spi.ChangeEventSource.ChangeEventSourceContex
 import io.debezium.relational.Table;
 import io.debezium.relational.TableId;
 import io.debezium.util.Clock;
+import io.debezium.util.Strings;
 
 /**
  * An abstract implementation of {@link LogMinerEventProcessor} that all processors should extend.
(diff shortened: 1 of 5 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
```diff
@@ -719,6 +719,22 @@ protected void handleDataEvent(LogMinerEventRow row) throws SQLException, Interr
...
+            switch (connectorConfig.getEventProcessingFailureHandlingMode()) {
+                case FAIL:
+                    LOGGER.error("Oracle LogMiner is unable to re-construct the SQL for '{}'", row);
+                    throw new DebeziumException("Oracle failed to re-construct redo SQL '" + row.getRedoSql() + "'");
+                case WARN:
+                    LOGGER.warn("Oracle LogMiner event '{}' cannot be parsed. This event will be ignored and skipped.", row);
+                    return;
+                default:
+                    // In this case, we explicitly log the situation in "trace" only and not as an error/warn.
+                    LOGGER.trace("Oracle LogMiner event '{}' cannot be parsed. This event will be ignored and skipped.", row);
```
jpechane, line 733: IMHO we are using `debug` elsewhere.

jpechane:
@Naros I'd keep focused on LogMiner. IMHO this is not a pressing issue for XStream.

Naros:
It looks like `event.processing.failure.handling.mode` is already documented for the Oracle connector to my surprise, so I'll adjust the log level of that one line and merge.

PR #3541, opened 2022-05-26
DBZ 5165 Run system testsuite inside OpenShift 

MartinMedek:


ani-sha:
Hi @MartinMedek, I think you have referenced the wrong issue key in the commit message, it's DBZ-5165, not DBZ-5168.

MartinMedek:
> Hi @MartinMedek, I think you have referenced the wrong issue key in the commit message, it's DBZ-5165, not DBZ-5168.

I see, thanks!

PR #3543, opened 2022-05-26
DBZ-5168 Add debezium-server-bom dependency to jdk workflow

ani-sha:
https://issues.redhat.com/browse/DBZ-5168


Naros (approved):
LGTM, thanks @ani-sha

PR #3544, opened 2022-05-26
DBZ-5166 Fixed ExtractNewRecordState SMT to pass all columns' data as is

rotem-ad:
DBZ-5166 Fixed ExtractNewRecordState SMT to pass all columns' data "as is", i.e: without potential unwanted values manipulation.

Prior to this fix, when inserting an explicit NULL value to a nullable column which has a default value, resulted in a record which contains the corresponding field with the default value (instead of NULL) after applying ExtractNewRecordState transformation.
A new unit test was added as well.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -313,6 +313,7 @@ Renato Mefi
 René Kerner
 Robert Hana
 Roman Kuchar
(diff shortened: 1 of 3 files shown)


jpechane:
@rotem-ad Applied, thanks a lot!

PR #3546, opened 2022-05-27
DBZ-5172 Expose more useful exception info with building the field default value schema

harveyyue:
https://issues.redhat.com/browse/DBZ-5172


jpechane:
@harveyyue Good improvement, thanks!

PR #3549, opened 2022-05-29
DBZ-5139 Abandon transaction who has too many events according to new configuration 'log.mining.buffer.memory.transaction.events.threshold'

zouxinyu0328:


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -343,6 +345,15 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
...
 
+    public static final Field LOG_MINING_BUFFER_MEMORY_TRANSACTION_EVENTS_THRESHOLD = Field.create("log.mining.buffer.memory.transaction.events.threshold")
+            .withDisplayName("Transaction events threshold when use log.mining.buffer.type='memory'.")
+            .withType(Type.LONG)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withDefault(DEFAULT_TRANSACTION_EVENTS_THRESHOLD)
+            .withDescription("The transaction events threshold when use log.mining.buffer.type='memory'. " +
+                    "If events count exceeds this threshold the transaction will be abandoned.");
+
```
Naros, line 356: I don't think we necessarily need to limit this to just the memory buffer type.  There is no harm in this just being named `log.mining.buffer.transaction.events.threshold` because the default is `0` (there is no limit).  This would allow either implementation to take advantage of this if the user enabled it, giving the most flexibility.  You should be able to pull the logic up into the abstract class for both to use.  Don't forget to change variable names and java docs accordinglly.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/memory/MemoryLogMinerEventProcessor.java
```diff
@@ -230,6 +230,15 @@ protected void addToTransaction(String transactionId, LogMinerEventRow row, Supp
...
 
+            if (getConfig().getLogMiningBufferMemoryTransactionEventsThreshold() > 0
+                    && transaction.getEvents().size() >= getConfig().getLogMiningBufferMemoryTransactionEventsThreshold()) {
+                LOGGER.error("Transaction {} is oversized. Will abandon it.", transactionId);
+                // Discard the transaction from TransactionCache and to make it releasable.
+                getAndRemoveTransactionFromCache(transactionId).getEvents().clear();
+                abandonedTransactionsCache.add(transactionId);
+                metrics.incrementOversizedTransactions();
+                return;
+            }
```
Naros, line 241: Let's pull this up to the abstract parent class & hook into that behavior in all impls of the processors.

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -3242,6 +3246,10 @@ See <<oracle-property-log-mining-transaction-retention-hours, `log.mining.transa
 |`long`
 |The number of committed transactions in the transaction buffer.
 
+|[[oracle-streaming-metrics-number-of-oversized-transactions]]<<oracle-streaming-metrics-number-of-oversized-transactions, `+NumberOfOversizedTransactions+`>>
+|`long`
+|The number of oversized transactions in the transaction buffer.
```
Naros, line 3251:
```suggestion
|The number of transactions that were discarded because their size exceeded <<oracle-property-log-mining-buffer-memory-transaction-events-threshold, `log.mining.buffer.memory.transaction.events.threshold`>>.
```
Note that I use the existing property name as described in the documentation from your original commit but this should be aligned with the suggestion above that we unify this to be implementation agnostic.

Naros (changes requested):
Hi, @zouxinyu0328 this LGTM.  I've made a few comments inline, thanks!

Naros:
Hi @zouxinyu0328 @EthanZ328 I added a commit & rebased against `main` to avoid conflicts.  Once all checks pass green, I'll proceed to merge this.  Since this is technically a new feature, I've officially added it to 2.0.0.alpha2.  

@jpechane if you want this backported, feel free to do so, otherwise I'm fine keeping this in 2.0 only.

zouxinyu0328:
Thanks very much @Naros !

PR #3550, opened 2022-05-29
DBZ-5173 Add some user case data for unsigned bigint converter

harveyyue:
https://issues.redhat.com/browse/DBZ-5173
There is no change for unsigin bigint converter, just add some user case data for unsigned bigint


jpechane:
@harveyyue Hi, it seems to me that this PR is unnecessary. Am I right?

PR #3552, opened 2022-05-31
DBZ-5179 Use BoundedLinkedHashSet to avoid high memory usage

Naros:
https://issues.redhat.com/browse/DBZ-5179

So we previously used an unbounded `Set` implementation in the Oracle metrics to track the rollback and abandoned transaction identifiers.  In some environments with high numbers of rollback or abandoned transactions, this could cause this object's memory footprint to grow unbounded.  This PR is meant to cap the number of elements in these collections to a maximum number of elements defined by `OracleStreamingChangeEventSourceMetrics#TRANSACTION_ID_SET_SIZE`.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleStreamingChangeEventSourceMetrics.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleStreamingChangeEventSourceMetrics.java
@@ -11,7 +11,6 @@ import java.time.Instant;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/util/BoundedLinkedHashSet.java
```diff
@@ -0,0 +1,49 @@
+/*
```
jpechane, line 1: I think instead of a new class we can re-use `org.apache.kafka.common.cache.LRUCache<K, V>` as we do in other parts of the codebase.  The `LinkedHashSet` uses `LinkedHashMap` interanally anyway so there will not be any performance loss.

PR #3555, opened 2022-06-01
DBZ-5170 Mysql Commit Timestamp

harveyyue:
https://issues.redhat.com/browse/DBZ-5170
I will update to all connectors doc later.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/mysql.adoc
+++ b/documentation/modules/ROOT/pages/connectors/mysql.adoc
@@ -582,6 +582,7 @@ Transaction boundary events contain the following fields:
 
 `status`:: `BEGIN` or `END`.
 `id`:: String representation of the unique transaction identifier.
+`ts_ms`:: The time of `BEGIN` or `END` transaction.
 `event_count` (for `END` events):: Total number of events emitted by the transaction.
 `data_collections` (for `END` events):: An array of pairs of `data_collection` and `event_count` elements.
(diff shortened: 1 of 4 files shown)


jingtaozhang18:
Hi @harveyyue. Here is some information I found in binlog that maybe help you. The timestamps in Xid events have difference between auto-commit is enabled and disabled. Here is an example. If I execute SQLs as shown below.

ex7_1

Then binlog file is as show below.

ex7_2

So , the timestamps of transaction's Xid events maybe not ordered.

jingtaozhang18:
Hi @harveyyue, the example you showed above is exactly what I was trying to say. And I think the timestamp of session 2 can't relfect the tranasaction commit time. The transaction commit time I understand is when the change of the transaction can be seen. Do you have any good idea to solve this problem?

harveyyue:
> Hi @harveyyue, the example you showed above is exactly what I was trying to say. And I think the timestamp of session 2 can't relfect the tranasaction commit time. The transaction commit time I understand is when the change of the transaction can be seen. Do you have any good idea to solve this problem?

Regarding the binlog event data which I have posted, the session2 transaction commit time is 1654101009000, the session1 is 1654101018000, I think we just to consume the binlog data without any data changes.

jpechane:
@harveyyue @jingtaozhang18 Hi guys, when do we stand on this PR then? Does it provide the expected data albeit they might be a bit difficult to correctl interpret?

jpechane:

PR #3561, opened 2022-06-02
DBZ-5153 Update MBean name references for SQL Server

Naros:
https://issues.redhat.com/browse/DBZ-5153


jpechane:
@Naros Thanks!

PR #3566, opened 2022-06-03
DBZ-5197 Add the Maven wrapper files generated by mvn wrapper:wrapper

alwaysbemark:
In a project that uses Maven and often changes the required version, it might be easier to use Maven Wrapper (https://maven.apache.org/wrapper/) which allows you to pin the Maven version and lets users run Maven commands as usual by using ./mvnw which will download the right version of Maven automatically and execute with it.

We use the Gradle wrapper in all of our projects to great effect, it's nice that Maven has this option officially supported as well.

The relevant files can be generated with the command:
`mvn wrapper:wrapper -Dmaven=3.8.4`

Please close and disregard the ticket/PR if there is no intention to use it, but it can definitely make contributors' lives easier.


jpechane:
@alwaysbemark Thanks for the proposal. I, personally, do not use that but that does not mean it can be usefule for other contributors.

PR #3454, opened 2022-05-04
DBZ-5074 Reduce number of Postgres database connections

jpechane:
https://issues.redhat.com/browse/DBZ-5074

PR #3570, opened 2022-06-04
DBZ-5204 Avoid incorrect connection close in PgOutputMessageDecoder

benesch:
Adding a test for this seems like it would be straightforward for someone who is more familiar with the PostgreSQL integration test setup than I. @jpechane would you be able to help with that, maybe?

----

Since #3454, the same PostgreSQL connection is used for repeated
PgOutputMessageDecoder instantiations. Therefore the
PgOutputMessageDecoder#close method should not close this connection,
because it belongs to the instantiator. (Previously each instantiation
of PgOutputMessageDecoder created its own PostgreSQL connection, and so
the close method did have the responsbility of closing this connection.)

The observable symptom of this bug was a Debezium crash in the following
scenario:

  1. Create Debezium connector.

  2. Add a new type:
     CREATE TYPE enum1 AS ENUM ('val1', 'val2');

  3. Add a new table:
     CREATE TABLE enum_type (f1 enum1);

  4. Wait for Debezium to crash because the metadata connection it would
     use to fetch the metadata about `enum1` is closed.

Note that it is critical that the connector be created _before_ the type. If the type is created first, then the type metadata is cached up front, and Debezium doesn't attempt to use the closed metadata connection.

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java
@@ -761,8 +761,6 @@ public class PgOutputMessageDecoder extends AbstractMessageDecoder {
 
     @Override
     public void close() {
-        if (connection != null) {
-            connection.close();
-        }
+        // Connection is owned by instantiator so we should not close it.
     }
 }


Naros:
Hi @benesch could you please add an integration test to the test the scenario please?

- Start connector, capturing changes for table X
- After streaming has started add the new custom type 
- Alter the table X and add a column that uses this custom type
- Proceed to insert a record into table X

That should be adequate to reproduce the failure & to test your fix solves the problem.

Naros:
So @jpechane I added a test case and decided to dig deeper into the code as I wasn't convinced that this was the right fix and it turns out the error that @benesch noticed stems from a call inside `PostgresStreamingChangeEventSource#execute` where we reconnect the `ReplicationConnection`.

This is problematic because when we create the decoder, we pass the underlying `PostgresConnection` to the decoder but with the call to `reconnect`, that underlying connection is closed and re-opened.  So applying the fix here would essentially lead to at least one connection being left open until the java process ends, which is not ideal and defeats the purpose of the change made in the aforementioned PR and associated Jira.

I think the play here is actually to leave the code as-is and instead of passing the raw `JdbcConnection` to the decoder, we pass a lambda that can point to a method on the `ReplicationConnection` to always provide the latest connection object, so that when calling `reconnect`, it would have no negative impact on the decoder.  

Do you see some other way that isn't obvious?

jpechane:
Closing in favour of https://github.com/debezium/debezium/pull/3580

PR #3572, opened 2022-06-05
DBZ-5207 change workflows to use the Maven wrapper instead of hardcoding the version

alwaysbemark:
In continuation of merging in the Maven wrapper, I wanted to further clean up some hardcoding of the Maven version around the project. This change means that the next time the Maven version is to be changed (for contributors, CI/CD, everything) then only the following needs to be invoked and commited: 

`./mvnw wrapper -Dmaven=3.x.x`

This flow is already used by the Debezium UI today.

Optionally the Maven wrapper PR https://github.com/debezium/debezium/pull/3566 can be backported to the active release branches - I'd be happy to do the work if it helps at all. 

As before, no worries to decline and close if there's no intention to use it at this stage :). Thanks!

### DIFF
--- a/.github/workflows/debezium-workflow.yml
+++ b/.github/workflows/debezium-workflow.yml
@@ -181,11 +181,6 @@ jobs:
           distribution: 'temurin'
           java-version: 17
 
-      - name: Set up Maven
-        uses: stCarolas/setup-maven@v4.3
-        with:
-          maven-version: 3.8.4
-
       - name: Cache Maven Repository
         id: maven-cache-check
         uses: actions/cache@v2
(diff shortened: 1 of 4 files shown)


jpechane:
@jcechace ping

jcechace:
@jpechane LGTM

jpechane:
@alwaysbemark ok, thanks we are going to do that. Could you please do the same for `debezium-connector-*` repositories too? And athem the missing `mvnw` script?

jpechane:
@alwaysbemark I think there should be a set of separate PRs that will add `mvnw` to other repositories so they are uncoupled.

alwaysbemark:
@jpechane I agree and can do that as well, just thought that it is actually "cleaner" to have the other repos built with core's mvnw - this ensures that the same version of Maven is used universally and would prevent nasty surprises down the line when the incubating connectors are merged into the main repo.

I'll gladly raise the ./mvnw PRs in the incubating repos, but it might be more future proof to always compile with ./mvnw (or ./core/mvnw in case of cloning to multiple different dirs).

jpechane:
@alwaysbemark Thanks. And yes you are right in using single `mvnw` for all builds

jpechane:
@alwaysbemark All applied, thanks a lot!

PR #162, opened 2017-01-11
DBZ-177: Providing an alternative way to create JDBC connection based …

rareddy:
…on the configured JDBC driver class name and supplied classloader. The loading/creating the JDBC connections is not reliable when driver libraries in a different classloader than the DriverManager. This also preserves previous behavior intact.

PR #3574, opened 2022-06-06
DBZ-5202 Add retry/delay to offset transaction id resolution

Naros:
https://issues.redhat.com/browse/DBZ-5202


Naros:
See internal Jenkins Oracle job #162 for test output, all pass.

jpechane:
@Naros Applied after a small refactoring, thanks!

PR #3579, opened 2022-06-07
DBZ-5201 Debezium does NOT support unix_timestamp() as DEFAULT value

harveyyue:
https://issues.redhat.com/browse/DBZ-5201


harveyyue:
@jpechane , Thanks for handle the property `hasDefaultValue` value in column class.

jpechane:
@harveyyue Applied, thanks

PR #3580, opened 2022-06-07
DBZ-5204 Do not cache connection details in `TypeRegistry`.

Naros:
https://issues.redhat.com/browse/DBZ-5204


jpechane:
@Naros Looks alike a safe approach now, thanks! I just made sure that `org.postgresql.core.BaseConnection.getTypeInfo()` is a cheap call and in JDBC driver code it seems so.

benesch:
Thanks, @jpechane and @Naros! Much appreciated!

PR #3583, opened 2022-06-08
DBZ-5221 Conditionalize entry in metrics table as upstream content

roldanbob:
[DBZ-5221](https://issues.redhat.com/browse/DBZ-5221)

Add `community` conditional to `NumberOfOversizedTransactions` metric in the Oracle connector documentation.


Naros (approved):
@roldanbob this LGTM as it is.  Let us know how it goes in the downstream build.

roldanbob:
> @roldanbob this LGTM as it is. Let us know how it goes in the downstream build.

Thanks, @Naros. I tested this in a downstream build and can confirm that the conditionalized  content is not shown, so this is all set now.

Naros:
Thanks @roldanbob, applied to main only.

PR #3584, opened 2022-06-08
DBZ-5222 move files into the API package

alwaysbemark:
https://issues.redhat.com/browse/DBZ-5222

This moves the data classes (such as `VariableScaleDecimal`) to the API package.
The only potentially "problematic" change is to include the Kafka Connect API in the Debezium API library, I am assuming that it was excluded for a reason, but I can't think of a reason to avoid it unless the Debezium API is trying to stay completely agnostic of Kafka?


debezium-api/pom.xml
```diff
@@ -15,6 +15,10 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
+        <dependency>
```
alwaysbemark, line 18: The Kafka Connect API is now in the Debezium API library, which is the only change that's somewhat significant. Though I don't know what someone would do with just the Debezium API without the Kafka one.

debezium-api/src/main/java/io/debezium/jdbc/DecimalMode.java
```diff
@@ -0,0 +1,12 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.jdbc;
```
alwaysbemark, line 6: I had to extract this out since this was user all over the shop (incl. in some of the logical types). Seems common enough that it deserves to be a first-class enum in the API. Some other things from the JDBC package can be moved, but I didn't want to disturb too much.

jpechane:
I am not fully sold on having `debezium-api` depend on `connect-api`. The original plan was to completely decouple the API user from Connect - hence the reason for `io.debezium.engine.format.SerializationFormat<SourceRecord>` being in `debezium-embedded`.
I need to make up my mind if another module would be a good solution.

alwaysbemark:
@jpechane Perhaps the Kafka Connect API can be included as a `provided` scope dep, and usage of the relevant classes would include having the Kafka Connect API on your classpath - and you'd be OK to ignore it if not using any Connect-related classes.

PR #3587, opened 2022-06-08
DBZ-5045 Use correct buffer type value after deprecated `infinispan` …

vjuranek:
…was removed

### DIFF
--- a/debezium-connector-oracle/pom.xml
+++ b/debezium-connector-oracle/pom.xml
@@ -401,7 +401,7 @@
         <profile>
             <id>infinispan-buffer</id>
             <properties>
-                <log.mining.buffer.type.name>infinispan</log.mining.buffer.type.name>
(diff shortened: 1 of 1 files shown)


Naros:
That's a nice catch @vjuranek, sorry I missed that when we added the new remote flavor.

vjuranek:
No problem @Naros , but we may think about adjusting CI workflows to spot this issue before the merge (I noticed that the tests in `main` fail after the merge by chance). We run different workflows for PRs and different after merge, so CI under the PR was green while it actually caused the failure.

jpechane:
@vjuranek Unfortunately we can't as the image is not public, we need to use secrets and secrets are not allowed on PRs

vjuranek:
Thanks for explanation @jpechane , I thought that there would be probably the reason why they are different but this doesn't come to my mind.

PR #3593, opened 2022-06-09
Add support for table ID delimiter characters

vjuranek:
https://issues.redhat.com/browse/DBZ-5198


jpechane:
@vjuranek Applied, thanks!

PR #3594, opened 2022-06-09
DBZ-5232 Remove duplicate scripting jars from the  dir

vjuranek:
https://issues.redhat.com/browse/DBZ-5232


Naros (approved):
This LGTM, the jars are no longer duplicated in both `lib_opt` and `lib`.
Thanks @vjuranek !

PR #3595, opened 2022-06-09
DBZ-5225 Align `source.scn` with the event's scn

Naros:
https://issues.redhat.com/browse/DBZ-5225


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSourceInfoStructMaker.java
```diff
@@ -35,14 +35,18 @@ public Schema schema() {
...
-        final String scn = sourceInfo.getScn() == null ? null : sourceInfo.getScn().toString();
         final String commitScn = sourceInfo.getCommitScn() == null ? null : sourceInfo.getCommitScn().toString();
+        final String eventScn = sourceInfo.getEventScn() == null ? null : sourceInfo.getEventScn().toString();
 
         final Struct ret = super.commonStruct(sourceInfo)
                 .put(SourceInfo.SCHEMA_NAME_KEY, sourceInfo.tableSchema())
                 .put(SourceInfo.TABLE_NAME_KEY, sourceInfo.table())
                 .put(SourceInfo.TXID_KEY, sourceInfo.getTransactionId())
-                .put(SourceInfo.SCN_KEY, scn);
+                // While we reuse the same SCN_KEY here as the one in the connector's offsets, the value we
```
jpechane, line 45: @Naros Could you please create TWO keys even if the anme will be the same and move the comment there and have use the appropriate key variable name at correct places? I think it would be cleaner solution.

jpechane:
@Naros Applied, thanks!

PR #3596, opened 2022-06-09
DBZ-5230 Extend syntax support for ALTER TABLE TRUNCATE PARTITION

Naros:
https://issues.redhat.com/browse/DBZ-5230

Applied a broader upstream PR which required some other partition features we hadn't contributed.  The upstream pull request can be found: https://github.com/antlr/grammars-v4/pull/2649


jpechane:
@Naros Applied, thanks

PR #3158, opened 2022-01-24
DBZ-4451 Correctly recover database history after rename

Naros:
https://issues.redhat.com/browse/DBZ-4451

PR #3598, opened 2022-06-09
DBZ-4451 Store previous table id in alter schema change events

Naros:
https://issues.redhat.com/browse/DBZ-4451

The originally reported issue above discussed an error that was raised by the connector when tables were renamed.  This error was due to the fact that the in-memory schema would contain both the previous table with the old name and the new table with the new name after the `ALTER TABLE abc RENAME TO xyz` DDL statement.

We've relaxed the checks in the streaming source so that we no longer hard fault on this condition but one of the remaining tasks was to address the contents of the in-memory schema.

This change introduces a new field to the `tableChanges` document called `previousId`.  This field is designed to carry the prior table's identifier when an ALTER event occurs that renames the table.  This allows us to automatically remove the old table from the in-memory schema and leave only the current representation during recovery.

Because this field is new and is expected to be optional, it should be safe & backward compatible.

This PR supersedes #3158


debezium-core/src/main/java/io/debezium/schema/SchemaChangeEvent.java
```diff
@@ -37,13 +38,13 @@ public class SchemaChangeEvent {
     private TableChanges tableChanges = new TableChanges();
 
     public SchemaChangeEvent(Map<String, ?> partition, Map<String, ?> offset, Struct source, String database, String schema, String ddl, Table table,
-                             SchemaChangeEventType type, boolean isFromSnapshot) {
+                             SchemaChangeEventType type, boolean isFromSnapshot, TableId previousTableId) {
```
jpechane, line 41: @Naros Is it possible to rename table and at the same time change its structure? If not then we should probably handle ranme as a distinct operation and either introduce `RENAME` change type or instead of having constructors we should provide a set of factory methods.

The code as is could probably go into 1.9 but as we have more flexibility for 2.0 we should re-consider this from architectural PoV.

debezium-connector-oracle/src/test/java/KafkaDatabaseHistoryTest.java
```diff
@@ -0,0 +1,222 @@
...
+import io.debezium.util.Testing;
+
+import oracle.jdbc.OracleTypes;
+
+/**
+ * Unit tests for Oracle's database history.
+ * 
+ * @author Chris Cranford
+ */
+public class KafkaDatabaseHistoryTest {
```
jpechane, line 59: Do we need Kafka database history? Why it would not work with MemoryDatabaseHistory? Also the functionality is for the core code so maybe the test should go to the core as I can imagine other connectors like MySQL to support  storing previous table id - see question regarding `RENAME` event.

debezium-core/src/main/java/io/debezium/schema/SchemaChangeEvent.java
Naros: So I know specifically with Oracle there are 2 ways to rename a table, the legacy way is a `RENAME ...` statement and the other is to use an `ALTER TABLE RENAME ...` statement.  I've never seen any documentation that indicates that you can adjust the table's structure while also renaming it with Oracle.  In fact, I just looked at the DDL syntax generated in the docs and it seems to indicate they're mutually exclusive.

I think it would be nice to have a `RENAME` type, but I'm concerned about backward compatibility.  I think we wanted to tackle this change so that if a user upgraded and then downgraded the records would still be consumed without any problems.

debezium-connector-oracle/src/test/java/KafkaDatabaseHistoryTest.java
Naros: It should work with any history implementation, I just mirrored my work off a similar setup in MySQL.  I'll look at pulling the core part of the test up and leaving the Oracle-bits as a concrete test class as we've done for Incremental Snapshots, etc.

debezium-core/src/main/java/io/debezium/schema/SchemaChangeEvent.java
jpechane: @Naros Hence what I am saying - this is fine for 1.9 - keep compatibility and make it better for 2.0 even at cost of compatibility break.

Naros:
Added Db2 sibling PR to align code changes made to `SchemaChangeEvent`, see https://github.com/debezium/debezium-connector-db2/pull/60

jpechane:
@Naros Applied, thanks!

PR #3599, opened 2022-06-10
DBZ-5233 Use fragments to specify MBean names vs. ifeval conditionals

roldanbob:
[DBZ-5233](https://issues.redhat.com/browse/DBZ-5233)

This change removes the `ifeval` conditions that were introduced in DBZ-5153 to specify the connector snapshot and streaming MBean names in the Monitoring chapters of the connector docs. In their place it adds a single shared fragment with tagged regions that are called via `include` statements in the Snapshot metrics and Streaming metrics sections in the different connector docs.  
Ideally, a future update to can automate correct processing of the conditional statements when content is fetched for downstream use.   
Tested in a local Antora build and downstream build.

This change should be backported to 1.9.

### DIFF
--- a/documentation/modules/ROOT/partials/modules/all-connectors/frag-common-mbean-name.adoc
+++ b/documentation/modules/ROOT/partials/modules/all-connectors/frag-common-mbean-name.adoc
@@ -1,0 +1,28 @@
+// Common
+tag::common-snapshot[]
+The *MBean* is `debezium.{mbean-name}:type=connector-metrics,context=snapshot,server=_<{context}.server.name>_`.
+end::common-snapshot[]
+
+tag::common-streaming[]
+The *MBean* is `debezium.{mbean-name}:type=connector-metrics,context=streaming,server=_<{context}.server.name>_`.
+end::common-streaming[]
+
+
+// MongoDB
+tag::mongodb-snapshot[]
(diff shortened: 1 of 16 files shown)


documentation/modules/ROOT/partials/modules/all-connectors/frag-common-mbean-name.adoc
```diff
@@ -0,0 +1,28 @@
...
+
+
+// SQL Server
+tag::sqlserver-snapshot[]
+The *MBean* is `debezium.{mbean-name}:type=connector-metrics,server=_<{context}.server.name>_,task=_<task.id>_,context=snapshot`.
+end::sqlserver-snapshot[]
+
+tag::sqlserver-streaming[]
+The *MBean* is `debezium.{mbean-name}:type=connector-metrics,server=_<{context}.server.name>_,task=_<task.id>_,context=streaming`.
+end::sqlserver-streaming[]
```
roldanbob, line 28: Rather than use separate files to hold each of the 6 MBean name formats, I realized that I could _tag_ regions within a single file to specify the naming format. The `include` statement in the hosting connector file then specifies the region from which to pull the name format.

jpechane:
@roldanbob There are conflicts when the PR is applied to 1.9. Could you please take a look? Thanks

roldanbob:
> @roldanbob There are conflicts when the PR is applied to 1.9. Could you please take a look? Thanks
Done. 
See PR #3608.

PR #3601, opened 2022-06-10
DBZ-5235 read Debezium metrics from consumer

ismailsimsek:
https://issues.redhat.com/browse/DBZ-5235


ismailsimsek:
@jpechane its ready for review

debezium-server/debezium-server-core/src/test/java/io/debezium/server/DebeziumServerIT.java
```diff
@@ -66,4 +69,22 @@ public void testPostgresWithJson() throws Exception {
...
+
+        Awaitility.await().atMost(Duration.ofSeconds(TestConfigSource.waitForSeconds())).until(() -> {
+            try {
+                // snapshot process finished
+                // and consuming events finished!
+                return metrics.snapshotCompleted()
+                        && metrics.streamingQueueCurrentSize() == 0;
+            }
+            catch (Exception e) {
+                e.printStackTrace();
```
jpechane, line 85: Could you please use a logger here? Thanks

debezium-server/debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
```diff
@@ -0,0 +1,130 @@
...
+ * NOTE: calls for reading metrics should be made after debezium connector initialized,
+ * after connector registers metrics, otherwise it will throw `Debezium Mbean not found` error
+ *
+ * @author Ismail Simsek
+ */
+@Dependent
+public class DebeziumMetrics {
+    protected static final Logger LOGGER = LoggerFactory.getLogger(DebeziumMetrics.class);
+    public static final MBeanServer mbeanServer = ManagementFactory.getPlatformMBeanServer();
+    @ConfigProperty(name = "debezium.source.max.queue.size", defaultValue = CommonConnectorConfig.DEFAULT_MAX_QUEUE_SIZE + "")
```
jpechane, line 34: PLease insert a newline above this line.
Alsso why is there the DefaultValue caluclated like plus empty String?

debezium-server/debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
```diff
@@ -0,0 +1,130 @@
...
+    int maxQueueSize;
+    private ObjectName snapshotMetricsObjectName;
```
jpechane, line 36: These values are effectively constant. Can't they be calculated in advance in a `@PostConstruct` method instead of lazy initialization in getter methods?

debezium-server/debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
```diff
@@ -0,0 +1,130 @@
...
+        try {
+            return (long) mbeanServer.getAttribute(getStreamingMetricsObjectName(), "MilliSecondsBehindSource");
+        }
+        catch (Exception e) {
+            throw new DebeziumException(e);
+        }
+    }
+
+    public void logMetrics() {
+        LOGGER.info("Debezium Metrics:" +
```
jpechane, line 122: Could you please use `{}` placeholders in logging message instead of constructing the log message as a string?

jpechane (changes requested):
@ismailsimsek Thanks a lot. Functionally the code looks good. I just left a couple of comments related to best practices we usually follow.

debezium-server/debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
ismailsimsek: it should be possible to initialize them using some kind of factory method. but i didn't see a good way to it them core package.

As is current method (`getDebeziumMbean`)  is not working under `@PostConstruct` because at that point of execution debezium mbeans are not registered to mbeanServer.

what do you think about something like this
```java
@ConfigProperty(name = "debezium.source.connector.class")
String connectorClass;
@ConfigProperty(name = "debezium.source.database.server.name")
String serverName;

if connectorClass.contains("Mysql")
    snapshotMetricsObjectName = "debezium.mysql:type=connector-metrics,context=snapshot,server=" + serverName
else if connectorClass.contains("Postgresql")
    snapshotMetricsObjectName = debezium.postgresql:type=connector-metrics,context=snapshot,server=+ serverName
else if connectorClass.contains("SQLserver")
...
```

debezium-server/debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
ismailsimsek: defaultValue only accepting string value, used "" to convert int to string

debezium-server/debezium-server-core/src/main/java/io/debezium/server/DebeziumMetrics.java
ismailsimsek: done

debezium-server/debezium-server-core/src/test/java/io/debezium/server/DebeziumServerIT.java
ismailsimsek: removed the line

jpechane:
@ismailsimsek Applied, thanks a lot!

PR #3602, opened 2022-06-10
DBZ-5236 Cannot convert field type tinyint(1) unsigned to boolean

harveyyue:
https://issues.redhat.com/browse/DBZ-5236


Naros:
hI @harveyyue there appears to be some test failures, could you take a look? Also, you could please find all references to `TinyIntOneToBooleanConverter` in the documentation and please make sure that the documentation is aligned with this change as well.  Thanks!

harveyyue:
There is no related test fairue with this changes. BTW, I have update this`TINYINT(1) UNSIGNED` type to mysql doc.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/converters/TinyIntOneToBooleanConverter.java
```diff
@@ -46,7 +50,7 @@ public void configure(Properties props) {
 
     @Override
     public void converterFor(RelationalColumn field, ConverterRegistration<SchemaBuilder> registration) {
-        if (!"TINYINT".equalsIgnoreCase(field.typeName()) || field.length().orElse(-1) != 1 || !selector.test(field)) {
+        if (!TINYINT_FAMILY.contains(field.typeName()) || field.length().orElse(-1) != 1 || !selector.test(field)) {
```
jpechane, line 53: @harveyyue Is not here introduced case sensitive comaprison instead of case insesnitive?

harveyyue: Actually, we convert the data type name to upper case, however, I fix it in case some code changes.
https://github.com/debezium/debezium/blob/main/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/listener/ColumnDefinitionParserListener.java#:~:text=String%20dataTypeName%20%3D%20dataType.name().toUpperCase()%3B

jpechane:
@harveyyue Applied, thanks!

PR #3604, opened 2022-06-10
DBZ-5237 Allow LOB storage cause to use quoted string identifiers

Naros:
https://issues.redhat.com/browse/DBZ-5237


Naros:
Upstream PR https://github.com/antlr/grammars-v4/pull/2650

PR #3607, opened 2022-06-13
DBZ-5069 Upgarde maven-core & maven-plugin-api to 3.8.4

ani-sha:
https://issues.redhat.com/browse/DBZ-5069

### DIFF
--- a/debezium-schema-generator/pom.xml
+++ b/debezium-schema-generator/pom.xml
@@ -38,7 +38,7 @@
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-plugin-api</artifactId>
-      <version>3.6.3</version>
+      <version>3.8.4</version>
       <scope>provided</scope>
     </dependency>
     <dependency>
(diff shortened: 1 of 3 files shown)


debezium-schema-generator/pom.xml
```diff
@@ -38,7 +38,7 @@
     <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-plugin-api</artifactId>
-      <version>3.6.3</version>
+      <version>[3.8.4]</version>
```
jpechane, line 41: I don't think the brackets should be used here. Also as was mentioned in the Jira comment, could you please introduce a new property like `maven.version` that would be used by the enforcer plugin and scheme generator dependencies so it set in a single place?

jpechane (changes requested):

ani-sha:
Db2 failure tracked via https://issues.redhat.com/browse/DBZ-5252

jpechane:
@ani-sha Applied, thanks

PR #3609, opened 2022-06-14
DBZ-4890 BigDecimal scale match Field when column uses negative scale

Naros:
https://issues.redhat.com/browse/DBZ-4890


jpechane:
@Naros Applied, thanks! As we don't have more info it is unsure that it solves the reported issue but it is definitely good safepoint.

PR #3614, opened 2022-06-16
DBZ-5260 Filter out unavailable replicaset members

tim-patterson:
Unavailable replica set members are still returned in the `clusterDescription.getServerDescriptions`
If the first in the list is unavailable then we fail to get the replica set name which later causes a "Unable to start MongoDB connector task since no replica sets were found"

I believe we should filter out all the unavailable members before creating the `addressStr` as later on in https://github.com/debezium/debezium/blob/3fb5bf133947e3a447a1e9ba5822333ca561d3ec/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ConnectionContext.java#L142
We simply choose the first address to connect to the replicaset

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -369,6 +369,7 @@ Timo Roeseler
 Tin Nguyen
 Tom Bentley
 Tom Billiet
+Tim Patterson
 Tomaz Lemos Fernandes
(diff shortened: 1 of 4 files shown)


jcechace:
@jpechane LGTM.

PR #3615, opened 2022-06-16
DBZ-5256 Add `internal.log.mining.transaction.snapshot.boundary.mode`option

Naros:
https://issues.redhat.com/browse/DBZ-5256

This internal option is meant to replace the previous internal option `log.mining.query.logs.for.snapshot.offset`. This now enum-based setting provides much more flexibility by being able to completely disable the in-progress transaction check, only grab the transactions that are in-progress based on V$TRANSACTION, or to gather all in-progress transactions by scanning the redo logs and inspecting V$TRANSACTION.  

In addition, for the sake of downstream, I've opted to temporarily enforce SKIP as the default.  The option is _internal_ so that if users want to explore this feature they can opt-in to it, but this seems like the most logical way to keep the connector behavior stable.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -454,14 +454,15 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             .withDescription(
                     "The maximum number of milliseconds that a LogMiner session lives for before being restarted. Defaults to 0 (indefinite until a log switch occurs)");
 
-    public static final Field LOG_MINING_QUERY_LOGS_FOR_SNAPSHOT_OFFSET = Field.createInternal("log.mining.query.logs.for.snapshot.offset")
-            .withDisplayName("Specifies whether to query transaction logs for snapshot offset position")
-            .withType(Type.BOOLEAN)
+    public static final Field LOG_MINING_TRANSACTION_SNAPSHOT_BOUNDARY_MODE = Field.createInternal("log.mining.transaction.snapshot.boundary.mode")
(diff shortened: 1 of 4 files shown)


jpechane:
@Naros Applied, very reasonable design, thanks!

aludwigSVA:
@Naros  @jpechane 

Hey guys, we are currently testing the oracle debezium connector for one of our customers. They are using Conduktor as an UI and after installing the debezium connector we get an Error in the UI when selecting it for configuration:

"Failed to deserialize, DecodingFailure at .configs[97].definition.display_name: Got value 'null' with wrong type, expecting string"

When looking at all config settings, I found exactly this one from this PR with a nullValue for DisplayName. 

"definition":{
            "name":"internal.log.mining.transaction.snapshot.boundary.mode",
            "type":"STRING",
            "required":false,
            "default_value":"skip",
            "importance":"LOW",
            "documentation":"Specifies how in-progress transactions are to be handled when resolving the snapshot SCN. \nall - Captures in-progress transactions from both V$TRANSACTION and starting a LogMiner session near the snapshot SCN.\ntransaction_view_only - Captures in-progress transactions based on data in V$TRANSACTION only. Recently committed transactions near the flashback query SCN won't be included in the snapshot nor streaming.\nskip - Skips gathering any in-progress transactions.",
            "group":"Connector",
            "width":"SHORT",
            "display_name":null,
            "dependents":[
               
            ],
            "order":65
         }


As it is the only config setting with a null DisplayName, I would propose to add one here as well.

maybe "Specifies how in-progress transactions are to be handled when resolving the snapshot SCN" ?

Cheers,

Axel

Naros:
Thanks @aludwigSVA, I will add a display name.

@mfvitale / @indraraj, we likely should fall back to displaying the `"name"` attribute if the `"display_name"` field is not populated, rather than the UI having an error. Would you both agree? A quick check over the repository and there is at least one other connector property that also does not have a display name in the core repository alone.

PR #3617, opened 2022-06-16
DBZ-5070 Don't reset lsn when lastCommitLsn is null

vjuranek:
Currently we update first `lsn` and then update `lastCommitLsn`.
However, when `lastCommitLsn` is for whatever reason `null`, previously
updated `lsn` is reset to `null` as `updateLastCommit()` updates also
`lsn`. This can have unwanted consequences like streaming again records
which were already streamed. To prevent this, update `lastCommitLsn`
only when it's not `null`.

Unfortunately, I wasn't able to figure out any reasonable test for it. Still looking on it, but I'm already little bit skeptical I'll come up with something reasonable:-(. So far tested only manually with Debezium server.

https://issues.redhat.com/browse/DBZ-5070

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/SourceInfo.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/SourceInfo.java
@@ -150,7 +150,9 @@ public final class SourceInfo extends BaseSourceInfo {
      */
     protected SourceInfo updateLastCommit(Lsn lsn) {
         this.lastCommitLsn = lsn;
-        this.lsn = lsn;
+        if (lsn != null) {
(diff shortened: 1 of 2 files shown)


vjuranek:
Adding at least basic unit test.

jpechane:
@vjuranek I agree with the analysis but I ma not sure this should be handled on the caller's side as I can imagine we might in fr future do the same somewhere else. IMHO either `updateLastCommit` should gurad against that or there should be another update method.

vjuranek:
Right @jpechane , it should be checked in `updateLastCommit()`, unless there is legitimate use case when we want to reset `lastCommitLsn` as well as `lsn`. I didn't find any, so moving the check to `updateLastCommit()`.

PR #3618, opened 2022-06-16
Update features.adoc

ProofOfPizza:
Tiny typo


jpechane:
@ProofOfPizza Could you please Prefix the commit message with string `[docs]`, rebase the PR against `main` and share your name so we can add you to the list of contributors? Thanks a lot

PR #3624, opened 2022-06-17
DBZ-5227 Specify use of JAR artifact for script engine language

roldanbob:
[DBZ-5227](issues.redhat.com/browse/DBZ-5227)

Per @jcechace's findings,  when using Streams to deploy Debezium to Kafka Connect, the CR must download script language archives from Maven as JAR files. This changes updates the downstream shared deployment instructions with that information.
Upon further discussion, it also came out that, as written, the instructions for deploying the scripting SMT were incomplete, so I addressed that as well. , and made changes to the content-based routing, and filtering SMT topics.  

I'd like to revisit this next quarter to apply further conditionalization to eliminate the need for multiple versions of `*kafka-connect-yaml.adoc`. This should be doable by using a single shared file with tagged regions. I spent some time trying to  implement that, but in the end backed out of it so that I could focus on the content update.

I know that we're finalizing 1.9.4, but if possible, we should backport this to 1.9. Or if that's going to raise too many ugly conflicts, I'll raise a separate PR to do that. 

### DIFF
--- a/documentation/antora.yml
+++ b/documentation/antora.yml
@@ -13,6 +13,7 @@ asciidoc:
     debezium-kafka-version: '3.1.0'
     debezium-docker-label: '1.9'
     DockerKafkaConnect: registry.redhat.io/amq7/amq-streams-kafka-28-rhel8:1.8.0
+    groovy-version: '3.0.11'
     install-version: '1.9'
     assemblies: '../assemblies'
     modules: '../../modules'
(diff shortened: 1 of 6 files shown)


jpechane (approved):
Good for me. It seems it applies cleanly on 1.9 too. @jcechace is out today so he'll take a look tomorrow.
As this is docs only there is no need to sync with releases as docs is built out of branch not tag.

jpechane:
@jcechace Could you please take a look?

jcechace (approved):
LGTM

PR #3627, opened 2022-06-20
DBZ-5271 Support set role statement for mysql8

harveyyue:
https://issues.redhat.com/browse/DBZ-5271


jpechane:
@harveyyue Applied, thanks. Could you please create a PR in upstream project too?

harveyyue:
> 

@jpechane Ok, I will do it.

PR #3628, opened 2022-06-20
DBZ-5273 Support using variables in KILL command

vjuranek:
https://issues.redhat.com/browse/DBZ-5273

Upstream PR: https://github.com/antlr/grammars-v4/pull/2669


jpechane:
@vjuranek Applied, thanks!

PR #3629, opened 2022-06-20
DBZ-5272 Notify all threads in case of mutliple writers

jpechane:
https://issues.redhat.com/browse/DBZ-5272

### DIFF
--- a/debezium-core/src/main/java/io/debezium/connector/base/ChangeEventQueue.java
+++ b/debezium-core/src/main/java/io/debezium/connector/base/ChangeEventQueue.java
@@ -199,7 +199,7 @@ public class ChangeEventQueue<T> implements ChangeEventQueueMetrics {
         synchronized (this) {
             while (queue.size() >= maxQueueSize || (maxQueueSizeInBytes > 0 && currentQueueSizeInBytes >= maxQueueSizeInBytes)) {
                 // notify poll() to drain queue
-                this.notify();
+                this.notifyAll();
                 // queue size or queue sizeInBytes threshold reached, so wait a bit
                 this.wait(pollInterval.toMillis());
             }
@@ -214,7 +214,7 @@ public class ChangeEventQueue<T> implements ChangeEventQueueMetrics {
 
             if (queue.size() >= maxBatchSize || (maxQueueSizeInBytes > 0 && currentQueueSizeInBytes >= maxQueueSizeInBytes)) {
                 // notify poll() to start draining queue and do not wait
-                this.notify();
+                this.notifyAll();
             }
         }
     }


debezium-core/src/main/java/io/debezium/connector/base/ChangeEventQueue.java
```diff
@@ -199,7 +199,7 @@ protected void doEnqueue(T record) throws InterruptedException {
         synchronized (this) {
             while (queue.size() >= maxQueueSize || (maxQueueSizeInBytes > 0 && currentQueueSizeInBytes >= maxQueueSizeInBytes)) {
                 // notify poll() to drain queue
-                this.notify();
+                this.notifyAll();
```
gunnarmorling, line 202: Worth talking to @krnaveen14 about this, who did this change originally. Note it used to be even `notifyAll()` [initially](https://github.com/debezium/debezium/commit/62045f6430796bc0d3bee69ba1a8ba1ea5abcf1d). Tbh. I'd prefer to use any of the more modern synchronisation primitives, I haven't seen wait/notify being used out in the wild for ages.

jpechane: @gunnarmorling Yes, we need two locks for that but I'd like to get a fix out first and we can rewrite the code later.

krnaveen14: @gunnarmorling @jpechane FYI, It was initially defined as `notifyAll()`, but was suggested to use `notify()` during PR Review as it's not a Best Practice to use the former. Also since my knowledge with PostgreSQL CDC (pgoutput plugin) is that it's doing sequential WAL decode and so assumed that there would be only one writer.

I can strongly confirm that deadlock / locking issues would arise if there is more than one writer thread calling the same **Instance** of `ChangeEventQueue`. So it should be changed to `notifyAll()` and there shouldn't be any impact due to this.

krnaveen14: @jpechane Similar `notifyAll()` changes required in `poll()` method of `ChangeEventQueue`. Otherwise it would cause `deadlock` in different scenario.

vjuranek:
While changing to `nofityAll()` IMHO shouldn't cause any regression, should prevent the issue described in DBZ-5272 and also should make deadlock less likely as it moves reader thread from WAITING state to BLOCKING state and thus makes possible for reader thread to obtain the lock. But I don't think it fully fixes prevents the deadlock. Consider following case:

1. Reader `thread0` waits for notify and is in WAITING state.
2. Writer `thread1` obtained the lock and enter `doEnqueue()`.
3. Writer `thread2` wants to obtain the lock, but as it's already hold by `thread1`, it's blocked and moves to the BLOCKED state.
4. `thread1` calls `notifyAll()` which moves `thread0` to BLOCKED state as well.
5. `thread1` releases the lock and moves to WAITING state.
6. `thread0` and `thread2` compete who obtains the lock.
7. `thread2` wins and obtains the lock.
8. `thread2` calls `nofityAll()` and moves `thread1` from WAITING state to BLOCKED
9. `thread2` moves to WAITING state and  releases the lock.
10. `thread0` competes with `thread1` who obtains the lock.
11. `thread1` wins and this can loop forever causing the deadlock.

There is better change now that reader thread wins the lock, but deadlock is IMHO still possible.

vjuranek:
>  should prevent the issue described in DBZ-5272

Actually, after looking more on DBZ-5272 I'm even not sure it will fix the case described there - the thread dump shows only one thread in waiting state. If I'm not mistaken, to `notifyAll()` has some effect, there have to be more threads in waiting state (which is possible, but not the case described in DBZ-5272).

krnaveen14:
> > should prevent the issue described in DBZ-5272
> 
> Actually, after looking more on DBZ-5272 I'm even not sure it will fix the case described there - the thread dump shows only one thread in waiting state. If I'm not mistaken, to `notifyAll()` has some effect, there have to be more threads in waiting state (which is possible, but not the case described in DBZ-5272).

Yes, the current `wait() / notify()` approach is completely under the assumption that there would be atmost one writer thread. To support multiple writer threads, the Queueing mechanism should be revisited. Simply changing it to `notifyAll()` won't completely solve the Thread starvation / deadlock.

tim-patterson:
The ArrayBlockingQueue in the jdk solves the same problem.
You shouldn't need two locks, but instead two condvars(`Condition`).
This would require swapping out the synchronised blocks for a `ReentrantLock` and then creating the two `Condition`s and using them instead of `this.wait()`/`this.notify()`.

https://github.com/JetBrains/jdk8u_jdk/blob/master/src/share/classes/java/util/concurrent/ArrayBlockingQueue.java

Another oddity I spotted is that it seems we only signal the reader thread to wake up when the queue is full rather than anytime we add an event(like the ArrayBlockingQueue does), not sure if there was a reason behind this for not?

jpechane:
I raised the number of threads to 16 on the reproducer and the deadlock happenend in a scale of seconds. After applying the PR the test was able to run 2 hours continuously with no issues. I am going to apply this as a "good enough" solution and there is https://issues.redhat.com/browse/DBZ-5277 asking for major changes that should reflect your comments/recommendations.

@tim-patterson The reasoning was that we will exchange not single messages but ideally blocks that are close to {{max.batch.size}} value.

PR #3632, opened 2022-06-21
DBZ-5277 ChangeEventQueue - Support Multiple Writer (doEnqueue) and Reader (poll) Threads

krnaveen14:
- ReentrantLock based write and read coordination
- Condition based Inter Thread communication for isFull and isNotFull conditions

### DIFF
--- a/debezium-core/src/main/java/io/debezium/connector/base/ChangeEventQueue.java
+++ b/debezium-core/src/main/java/io/debezium/connector/base/ChangeEventQueue.java
@@ -11,6 +11,10 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
 import java.util.Queue;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.locks.Condition;
+import java.util.concurrent.locks.Lock;
+import java.util.concurrent.locks.ReentrantLock;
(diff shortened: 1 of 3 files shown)


jpechane:
@krnaveen14 Would it make sense to wake up either only readers or writers depending on whether the queue was written to or read from?

krnaveen14:
> @krnaveen14 Would it make sense to wake up either only readers or writers depending on whether the queue was written to or read from?

Yes, that how it is done.

**doEnqueue (writer)**
- If maxQueueSize or maxQueueSizeInBytes is reached, it wakes up `poll()` by `isFull.signalAll()` and waits for signal on `isNotFull.await(...)`
- if maxBatchSize or maxQueueSizeInBytes is reached, it wakes up `poll()` by `isFull.signalAll()` and releases the lock

**poll (reader)**
- If there is no records available or maxBatchSize not reached, it wakes up `doEnqueue()` by `isNotFull.signalAll()` and waits for signal on `isFull.await(...)`
- After removing records from queue, it wakes up `doEnqueue()` by `isNotFull.signalAll()` and releases the lock

Essentially, Writer can only wake up the Reader and Reader can only wake up the Writer.

This way Deadlock  and Thread Starvation can never occur.

PR #3637, opened 2022-06-22
DBZ-5288 Upgrade MySQL driver to 8.0.29

Naros:
https://issues.redhat.com/browse/DBZ-5288


jpechane:
@Naros Thanks for all the updates. What was the trigger to bump all of them up?

PR #3639, opened 2022-06-22
DBZ-5290 Upgrade SQL Server driver to 10.2.1.jre8

Naros:
https://issues.redhat.com/browse/DBZ-5290

- [x] Update documentation
- [x] Update `deebzium-examples` repository to explicitly disable encryption. 
https://github.com/debezium/debezium-examples/pull/267


jpechane:
@Naros Looking at the errors, is it possible it uses SSL by default now?

Naros:
Hi @jpechane it does.  It looks like in 10.2, quite a lot of breaking changes:
https://docs.microsoft.com/en-us/sql/connect/jdbc/release-notes-for-the-jdbc-driver?view=sql-server-ver16#changes-in-102

Specifically to this problem, `encrypt` is enabled by default whereas in 9.4 it was not.  Should we hold off on this upgrade then or what other changes do you suggest we do to align with the latest?

jpechane:
I think we should bite a bullt and get this resolved, yet I don't see it as a priority so somebody else but you could take a look.

Naros:
So @jpechane, I simply explicitly disabled the encryption for our test suite.  

According to this link:
https://techcommunity.microsoft.com/t5/azure-database-support-blog/pkix-path-building-failed-unable-to-find-valid-certification/ba-p/2591304

We will need to download and import the TLS1 and TLS2 certifications using `keytool` and this would need to be done on the host that runs the build as I'm not sure we necessarily wanted to do this, I opted for using the `encrypt=false` on the connection URL for the test suite for now.  Let me know if you would rather we take a different approach.

Naros:
@vjuranek since Jiri is on PTO, do you have any preference or opinion on this PR?

vjuranek:
I'm completely fine with disabling encryption in the testsuite, but given that this is the default JDBC drive setting, won't it break user deployments after upgrade?

Naros:
It would be an incompatibility that would require connector reconfiguration by adding `database.encrypt=false` if they don't use encryption but with the volume of SQL Server users with issues with TLSv1 being unavailable and requiring legacy cyphers, I'm wondering whether having the latest driver would help rather than one that is older, idk.

Naros:
So here's a thought to maintain backward compatibility.

In prior versions, if you wanted to support encryption, `database.encrypt` and all the driver bits had to be provided by the user.  If the encrpyt option wasn't provided, encryption wasn't enabled.  

To remain backward compatible, how about we check whether `database.encrypt` is even provided by the user configuration.  If it is, we simply pass that through without making any changes.  If we find the user does not supply `database.encrypt`, we inject this configuration with a value of `false`, simulating the old driver behavior as the default.

@vjuranek @jpechane what's your thoughts?

vjuranek:
Sounds good to me @Naros . Actually I'm fine with both options (breaking backward compatibility as well as keeping it) - as this is for 2.0 it's fine to break backward compatibility. I was just little bit worried that it may affect quite a lot of users (from my experience, lots of users don't read release notes, upgrade manuals etc.). Of course this comes with costs developing code for keeping it backward compatible and maintaining it. I'm too short with Debezium to have any relevant opinion on it (especially not knowing the user community well enough, have no idea how many users will actually be affected etc.). So completely up to you @Naros and @jpechane .

jpechane:
@Naros @vjuranek I'd vote for breaking backward compatibility. But how difficult would be then for a new user to start with Debezium? If too many steps are required then let's follow the path of adding {{database.encrypt=false}} by default as the user can always override it.

Naros:
In order to use encryption with the docker container, it would appear we need to follow these steps:
https://docs.microsoft.com/en-us/sql/linux/sql-server-linux-docker-container-security?view=sql-server-ver16

This basically involves:
- Creating `.key` and `.pem` files
- Supplying a custom `mssql.conf` file that needs to be mounted
- Start `mssql/server` and mount the `.key`, `.pem`, and `.conf` files into the container

I think in order to simplify using Debezium + SQL Server in both the main repository as well as in the examples repository, we may want to think about creating a pre-built container with these files mounted in the right locations so that we don't need to worry about adding all the mount details in multiple places.  Then we document that if you don't use encryption, the user would need to explicitly provide `database.encrypt=false` in their connector details.

vjuranek:
> **UPDATE** Actually, I think what might work is we mount the files as part of the test suite in the main repository and then we use the toggle to disable encryption for the tutorial as that would give us a way to test both scenarios and avoids a complicated set up on the tutorial and avoids the creation of a new container, wdyt?

sounds good to me

jpechane:
@Naros :+1:  on the proposal

Naros:
Added PR https://github.com/debezium/debezium-examples/pull/267 to address the examples repository.  The examples PR can be merged without this PR since `database.encrypt=false` is the default behavior with the current driver but will be necessary when the driver update is merged in this repository.

documentation/modules/ROOT/pages/connectors/sqlserver.adoc
```diff
@@ -1478,6 +1478,12 @@ a|_n/a_
 For {prodname} to capture change events from SQL Server tables, a SQL Server administrator with the necessary privileges must first run a query to enable CDC on the database.
 The administrator must then enable CDC for each table that you want Debezium to capture.
 
+[NOTE]
+====
+The Microsoft JDBC SQL server uses SSL encrypted connections by default.
```
roldanbob, line 1483:
```suggestion
By default, JDBC connections to Microsoft SQL Server are protected by SSL encryption.
```

documentation/modules/ROOT/pages/connectors/sqlserver.adoc
```diff
@@ -1478,6 +1478,12 @@ a|_n/a_
...
+If the SQL Server database does not have SSL enabled, or you wish to connect without using SSL, please specify the `database.encrypt=false` option in the connector configuration to disable SSL support.
```
roldanbob, line 1484:
```suggestion
If SSL is not enabled for a SQL Server database, or if you want to connect to the database without using SSL, you can disable SSL by setting the value of  the `database.encrypt` property in connector configuration to `false`.
```

documentation/modules/ROOT/pages/connectors/sqlserver.adoc
```diff
@@ -1918,6 +1926,12 @@ spec:
 |11
 |The name of the database history topic where the connector will write and recover DDL statements. This topic is for internal use only and should not be used by consumers.
 
+|12
+|The SSL truststore where the server certificates are stored. This is required when `database.encrypt=false` is not specified.
```
roldanbob, line 1930:
```suggestion
|The path to the SSL truststore that stores the server's signer certificates. 
This property is required unless database encryption is disabled (`database.encrypt=false`).
```

documentation/modules/ROOT/pages/connectors/sqlserver.adoc
```diff
@@ -1918,6 +1926,12 @@ spec:
...
+
+|13
+|The SSL truststore password. This is required when `database.encrypt=false` is not specified.
```
roldanbob, line 1933:
```suggestion
|The SSL truststore password. 
This property is required unless database encryption is disabled (`database.encrypt=false`).
```

documentation/modules/ROOT/pages/connectors/sqlserver.adoc
```diff
@@ -1976,7 +1992,8 @@ Optionally, you can ignore, mask, or truncate columns that contain sensitive dat
 <9> A list of all tables whose changes {prodname} should capture.
 <10> The list of Kafka brokers that this connector will use to write and recover DDL statements to the database history topic.
 <11> The name of the database history topic where the connector will write and recover DDL statements. This topic is for internal use only and should not be used by consumers.
-
+<12> The SSL truststore where the server certificates are stored. This is required when `database.encrypt=false` is not specified.
+<13> The SSL truststore password. This is required when `database.encrypt=false` is not specified.
```
roldanbob, line 1996:
```suggestion
<12> The path to the SSL truststore that stores the server's signer certificates. 
This property is required unless database encryption is disabled (`database.encrypt=false`).
<13> The SSL truststore password. 
This property is required unless database encryption is disabled (`database.encrypt=false`).
```

roldanbob (changes requested):

Naros:
I have been looking at the random SQL Server test failures and they don't seem to be related to the driver change.  I've been able to reproduce the failure with `testTableExcludeList` only once and even then I wasn't able to get much detail and haven't been able to reproduce it since.  The other also appears to be random as it passes locally.

jpechane:
@Naros Good solution, thanks!

PR #3641, opened 2022-06-22
DBZ-5291: Fix binlog filename comparison

morozov:
See [DBZ-5291](https://issues.redhat.com/browse/DBZ-5291).

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlHistoryRecordComparator.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlHistoryRecordComparator.java
@@ -97,12 +97,10 @@ final class MySqlHistoryRecordComparator extends HistoryRecordComparator {
             return recordedTimestamp <= desiredTimestamp;
         }
 
-        // First compare the MySQL binlog filenames that include the numeric suffix and therefore are lexicographically
-        // comparable ...
-        String recordedFilename = recorded.getString(SourceInfo.BINLOG_FILENAME_OFFSET_KEY);
-        String desiredFilename = desired.getString(SourceInfo.BINLOG_FILENAME_OFFSET_KEY);
-        assert recordedFilename != null;
-        int diff = recordedFilename.compareToIgnoreCase(desiredFilename);
+        // First compare the MySQL binlog filenames
+        BinlogFilename recordedFilename = BinlogFilename.of(recorded.getString(SourceInfo.BINLOG_FILENAME_OFFSET_KEY));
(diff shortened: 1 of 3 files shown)


debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlDatabaseSchemaTest.java
```diff
@@ -92,7 +92,7 @@ public void shouldApplyDdlStatementsAndRecover() throws InterruptedException {
         final MySqlOffsetContext offset = initializeOffset(connectorConfig);
 
         // Set up the server ...
-        offset.setBinlogStartPoint("binlog-001", 400);
```
morozov, line 95: The original binglog filename is invalid according to the [documentation](https://dev.mysql.com/doc/refman/5.7/en/mysqld.html) and the [source code](https://github.com/mysql/mysql-server/blob/mysql-5.7.35/sql/binlog.cc#L3624-L3629):
> **mysqld** appends a numeric extension to the binary log base name to generate binary log file names

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlHistoryRecordComparator.java
```diff
@@ -143,4 +141,47 @@ else if (recordedGtidSetStr != null) {
...
+
+        @Override
+        public String toString() {
+            return "BinlogFilename [baseName=" + baseName + ", extension=" + extension + "]";
+        }
+
+        @Override
+        public int compareTo(BinlogFilename other) {
+            if (!baseName.equals(other.baseName)) {
+                throw new IllegalArgumentException("Can't compare binlog filenames with different base names");
```
morozov, line 182: From the documentation:

> If no **base_name** value is given, the default name is the value of the `--pid-file` option (which by default is the name of host machine) followed by `-bin`.

So technically it is possible that if the server wasn't configured with an explicit base name and got its host renamed, the new binlog files will have a different name.

Although, the documentation also says:

> It is recommended that you specify a base name explicitly rather than using the default of the host name.

So it should be quite a rare case if existent at all.

jpechane: @morozov Unfortunately we cannot exlude this possibility. So I don't havy any problem toto it on the `main` branch I am not eager to push it to `1.9`

morozov: This is not a problem. We will run a patched version until this change is released.

jpechane: @morozov Cool, thanks for that!

PR #3642, opened 2022-06-23
DBZ-5296 Correct documentation of Event Router in expand.json mode

drozd0:
Hi, want to improve documentation in the main branch.


jpechane:
@drozd0 Applied, thanks!

PR #3643, opened 2022-06-23
DBZ-5299 Allow the maven version is greater than the current 3.8.4

harveyyue:
https://issues.redhat.com/browse/DBZ-5299

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -316,7 +316,7 @@
                         <configuration>
                             <rules>
                                 <requireMavenVersion>
-                                    <version>[${version.maven}]</version>
+                                    <version>[${version.maven},3.8.5),[3.8.6,)</version>
                                 </requireMavenVersion>
                             </rules>
                         </configuration>


jpechane:
@harveyyue It seems to me, that Debezium Server works when built with 3.8.6. Could you please double check it?
If it is the case then we still need to get rid of 3.8.5.
I'd then recommend to change the reul to something like `[${version.maven},3.8.5),[3.8.6,)` - I am not sure if I got the syntax correctly.
The default will still be `3.8.4` but the new ones but the buggy one could be used.

harveyyue:
@jpechane I have built using maven 3.8.5 and 3.8.6, and deploy mysql connector in connect, there are works fine. BTW, the 
  syntax `[${version.maven},3.8.5),[3.8.6,)` is also fine.

jpechane:
No, it is Debezium Server that is broken with 3.8.5. It starts JVM but no service

harveyyue:
> No, it is Debezium Server that is broken with 3.8.5. It starts JVM but no service

Yes, the Debezium Server built by maven 3.8.5 is not work.

2022-06-26 00:29:22,594 WARN  [io.qua.config] (main) Unrecognized configuration key "quarkus.log.console.json" was provided; it will be ignored; verify that the dependency extension for this configuration is set or that you did not make a typo
2022-06-26 00:29:24,165 INFO  [io.quarkus] (main) debezium-server-dist 2.0.0-SNAPSHOT on JVM (powered by Quarkus 2.10.0.Final) started in 7.101s. Listening on: http://0.0.0.0:8080
2022-06-26 00:29:24,166 INFO  [io.quarkus] (main) Profile prod activated.
2022-06-26 00:29:24,166 INFO  [io.quarkus] (main) Installed features: [cdi, resteasy, resteasy-jackson, smallrye-context-propagation, smallrye-health, vertx]
^C2022-06-26 00:30:32,994 INFO  [io.quarkus] (Shutdown thread) debezium-server-dist stopped in 0.098s

jpechane:
@harveyyue Applied, thanks

PR #3652, opened 2022-06-24
DBZ-5285 Fix store.only.captured.tables.ddl to work during streaming

Naros:
https://issues.redhat.com/browse/DBZ-5285


jpechane:
@Naros Applied, thanks

PR #3661, opened 2022-06-28
DBZ-5245 Track commit scn per redo thread

Naros:
https://issues.redhat.com/browse/DBZ-5245

@jpechane So this is forward compatible but isn't backward compatible.  Do we think we need to be backward compatible and if not we can scope this change to 2.0 rather than including 1.9.  We can discuss this tomorrow.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleOffsetContext.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleOffsetContext.java
@@ -52,7 +52,7 @@ public class OracleOffsetContext extends CommonOffsetContext<SourceInfo> {
      */
     private boolean snapshotCompleted;
 
-    public OracleOffsetContext(OracleConnectorConfig connectorConfig, Scn scn, Scn commitScn, String lcrPosition,
+    public OracleOffsetContext(OracleConnectorConfig connectorConfig, Scn scn, CommitScn commitScn, String lcrPosition,
                                Scn snapshotScn, Map<String, Scn> snapshotPendingTransactions,
                                boolean snapshot, boolean snapshotCompleted, TransactionContext transactionContext,
                                IncrementalSnapshotContext<TableId> incrementalSnapshotContext) {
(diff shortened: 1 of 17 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleOffsetContext.java
```diff
@@ -184,9 +185,9 @@ public static Builder create() {
             }
             else {
                 final Scn scn = sourceInfo.getScn();
-                final Scn commitScn = sourceInfo.getCommitScn();
+                final String commitScn = sourceInfo.getCommitScn().toString();
```
jpechane, line 188: @Naros Please don't use toString but a structural approach. So mutliple keys with a common prefix like we do in `TransactionContext`

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/SourceInfo.java
```diff
@@ -26,7 +26,7 @@ public class SourceInfo extends BaseSourceInfo {
     public static final String SNAPSHOT_KEY = "snapshot";
```
jpechane, line 26: I wonder if source info schema should change too and use array of strings.

jpechane (commented):
@Naros Generally LGTM, I'd just ask for two minor structural changes

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/SourceInfo.java
Naros: So the string-based implementation just writes the complete contents of the "commit_scn" value-set to the source info field however, thinking about this more, I don't believe that's exactly what we want.  

A single transaction is only ever associated with a given Oracle RAC node, and therefore the source info block should only refer to a single commit SCN data point, which is the node the transaction happened on.  I'll adjust that behavior first in the string-based implementation by tracking the Oracle RAC node's redo thread in `SourceInfo`.

Once that is done, the schema change would in fact simply be an array of fields as you described.

Naros:
Tested both Xstream and LogMiner, the CloudEventsConverter error is the only issue I see atm.

Naros:
As mentioned on Zulip, I opted to use a flattened approach rather than the nested struct approach to avoid complications with the CloudEvents converter since CE only supports a subset of data types.  This continues to expose the necessary attributes to users for consumption while staying relatively simple and not overly complex.

PR #3662, opened 2022-06-29
DBZ-5044 Centralize creation of Message Schemas

ani-sha:
https://issues.redhat.com/browse/DBZ-5044

Corresponding non-core connector changes
- [x] Vitess Connector - https://github.com/debezium/debezium-connector-vitess/pull/99
- [x] Cassandra Connector - https://github.com/debezium/debezium-connector-cassandra/pull/70
- [x] Db2 Connector - https://github.com/debezium/debezium-connector-db2/pull/72

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/LogicalDecodingMessageMonitor.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/LogicalDecodingMessageMonitor.java
@@ -64,7 +64,7 @@ public class LogicalDecodingMessageMonitor {
     private final Schema blockSchema;
     private final Schema valueSchema;
 
-    private static final PostgresSchemaFactory postgresSchemaFactoryObject = new PostgresSchemaFactory();
+    private static final PostgresSchemaFactory postgresSchemaFactoryObject = PostgresSchemaFactory.get();
 
     public LogicalDecodingMessageMonitor(PostgresConnectorConfig connectorConfig, BlockingConsumer<SourceRecord> sender) {
         this.schemaNameAdjuster = connectorConfig.schemaNameAdjustmentMode().createAdjuster();
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/LogicalDecodingMessageMonitor.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/LogicalDecodingMessageMonitor.java
@@ -12,7 +12,6 @@ import java.util.Base64;
 import java.util.Base64.Encoder;
 
 import org.apache.kafka.connect.data.Schema;
-import org.apache.kafka.connect.data.SchemaBuilder;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.source.SourceRecord;
 
@@ -65,6 +64,8 @@ public class LogicalDecodingMessageMonitor {
     private final Schema blockSchema;
     private final Schema valueSchema;
 
+    private static final PostgresSchemaFactory postgresSchemaFactoryObject = new PostgresSchemaFactory();
+
     public LogicalDecodingMessageMonitor(PostgresConnectorConfig connectorConfig, BlockingConsumer<SourceRecord> sender) {
         this.schemaNameAdjuster = connectorConfig.schemaNameAdjustmentMode().createAdjuster();
         this.sender = sender;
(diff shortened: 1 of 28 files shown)


debezium-core/src/main/java/io/debezium/schema/SchemaBuilderFactory.java
```diff
@@ -0,0 +1,88 @@
...
+ * A factory for creating {@link SchemaBuilder} structs.
+ *
+ * @author Anisha Mohanty
+ */
+
+public class SchemaBuilderFactory {
+
+    private static final int SCHEMA_VERSION = 1;
+
+    private static final String SERVER_NAME_CLASS = "io.debezium.connector.common.ServerNameKey";
```
jpechane, line 26: It si not class name but schema name for heartbeat so please use name like `HEARTBEAT_KEY_SCHEMA_NAME`.
The other schemas should follow the same naming pattern.

debezium-core/src/main/java/io/debezium/schema/SchemaBuilderFactory.java
```diff
@@ -0,0 +1,88 @@
...
+    private static final String HEARTBEAT_CLASS = "io.debezium.connector.common.Heartbeat";
+
+    private static final String TRANSACTION_BLOCK_NAME = "event.block";
+    private static final String TRANSACTION_EVENT_COUNT_COLLECTION_NAME = "event.collection";
+    private static final String TRANSACTION_METADATA_KEY_CLASS = "io.debezium.connector.common.TransactionMetadataKey";
+    private static final String TRANSACTION_METADATA_VALUE_CLASS = "io.debezium.connector.common.TransactionMetadataValue";
+
+    public static Schema heartbeatKeySchema(SchemaNameAdjuster adjuster) {
```
jpechane, line 34: Please don't use static methods but instance methods and convert the class into a singleton.
That way we could easily intorduce subclasses and use inheritance.

debezium-core/src/main/java/io/debezium/schema/SchemaBuilderFactory.java
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * A factory for creating {@link SchemaBuilder} structs.
+ *
+ * @author Anisha Mohanty
+ */
+
+public class SchemaBuilderFactory {
+
+    private static final int SCHEMA_VERSION = 1;
```
jpechane, line 24: Every schema will have its own version number so please introduce `HEARTBEAT_KEY_SCHEMA_VERSION` etc.

jpechane (commented):
@ani-sha That's exactly the approach I was looking for!
I left few comments to slightly change the structure.

debezium-core/src/main/java/io/debezium/schema/SchemaBuilderFactory.java
```diff
@@ -0,0 +1,109 @@
...
+    private static final int TRANSACTION_BLOCK_SCHEMA_VERSION = 1;
+
+    private static final String TRANSACTION_EVENT_COUNT_COLLECTION_SCHEMA_NAME = "event.collection";
+    private static final int TRANSACTION_EVENT_COUNT_COLLECTION_SCHEMA_VERSION = 1;
+
+    private SchemaBuilderFactory() {
+    }
+
+    public static SchemaBuilderFactory getInstance() {
+        return new SchemaBuilderFactory();
```
jpechane, line 52: This is not singleton, it returns new instance every call.

jpechane (commented):
@ani-sha I pushed one commitchangin naming and ordering of class attributes.
Also the class is not singleton, it creates new instance with every call. Could you please fix it?

debezium-core/src/main/java/io/debezium/schema/SchemaBuilderFactory.java
```diff
@@ -0,0 +1,109 @@
...
+import io.debezium.pipeline.txmetadata.TransactionMonitor;
+import io.debezium.util.SchemaNameAdjuster;
+
+/**
+ * A factory for creating {@link SchemaBuilder} structs.
+ *
+ * @author Anisha Mohanty
+ */
+
+public class SchemaBuilderFactory {
```
jpechane, line 22: `SchemaFactory` would be a better name.

jpechane:
@ani-sha LGTM. The next will be `ConnectTableChangeSerializer`

jpechane:
@ani-sha Partly please see `io.debezium.data.VerifyRecord.isValid(SourceRecord, boolean)`. There version explicitly ignored and we set it.  The comparison shold ignore the version numer at least now.

jpechane:
@ani-sha I've push few commits related to test fixing

ani-sha:
Thanks for the fixes, @jpechane Any other schemas to work on in this PR?

jpechane:
@ani-sha Now please convert files in `debezium-core/src/main/java/io/debezium/data/` save `Envelope` and after that we will try one of the connectors.

debezium-core/src/main/java/io/debezium/schema/SchemaFactory.java
```diff
@@ -0,0 +1,293 @@
...
+    private static final int SCHEMA_HISTORY_CHANGE_SCHEMA_VERSION = 1;
+
+    /*
+     * Datatype schemas
+     */
+
+    /*
+     * Bits schema
+     */
+    private static final int BITS_SCHEMA_VERSION = 1;
```
jpechane, line 85: @ani-sha I'd probably place the `VERSION` side by side with the schema name in the class. Also if we keep the schema, fields, version etc. in the separate class then it would make sense to keep the simple namse so just `SCHEMA_NAME`, `VERSION` etc. without adding the type name as there is no name conflict.

ani-sha: Yes, makes sense. Will add the change.

jpechane:
@ani-sha So, please take a look at `io.debezium.data.Envelope`. Generally only `Envelope` schema will be versioned. `source` and `transaction` is already handled. `after` and `before` will not be versioned as they are dynamic.

Please go through postgres connector code and try to finish the implementation for it.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/LogicalDecodingMessageMonitor.java
```diff
@@ -64,35 +63,24 @@ public class LogicalDecodingMessageMonitor {
     private final Schema blockSchema;
     private final Schema valueSchema;
 
+    private static final PostgresSchemaFactory postgresSchemaFactoryObject = PostgresSchemaFactory.get();
```
jpechane, line 66: @ani-sha Please remove this attribute

jpechane:
@ani-sha LGTM, thanks. IIRC only few tasks are remaining

- [ ] Complete all core connectors
- [ ] Get rid of few remaining SchemaFactory instances
- [ ] Implement the functionality for non-core connectors

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/LogicalDecodingMessageMonitor.java
ani-sha: @jpechane if we get rid of that how are we calling its related schemas in `PostgresSchemaFactory` ie `logicalDecodingMessageMonitorKeySchema` etc?

ani-sha:
@jpechane Need to include `MongoDbSchema#schemaFor` key and value schema in the `MongoDbSchemaFactory`?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/LogicalDecodingMessageMonitor.java
jpechane: @ani-sha Just inline the `get()` call, no need for variable.

jpechane:
> @jpechane Need to include `MongoDbSchema#schemaFor` key and value schema in the `MongoDbSchemaFactory`?

I am afraid I don't understand the question.

ani-sha:
> I am afraid I don't understand the question.

@jpechane I meant to ask if we need to include the key & value schema present in `MongoDbSchema#schemaFor` inside our Schema Factory.

ani-sha:
@jpechane Added the non-core connector changes.

jpechane:
@ani-sha Applied, thanks for the huge piece of work!

PR #3664, opened 2022-06-29
DBZ-5281 Use separate connection for snapshot-based LogMiner queries

Naros:
https://issues.redhat.com/browse/DBZ-5281


jpechane:
@Naros Applied, thanks!

PR #3666, opened 2022-06-30
DBZ-5334 Add ts_ms to txn boundary event examples; update description

roldanbob:
[DBZ-5334](https://issues.redhat.com/browse/DBZ-5334)

Updates the transaction boundary event section in several connector docs with a description of the ts_ms field, and adds the field to the associated example.

When this change is merged, it should also be applied to 1.9.


Naros:
Hi @roldanbob, the `ts_ms` field applies to all connectors, including Db2 and SQL Server.

roldanbob:
> Hi @roldanbob, the `ts_ms` field applies to all connectors, including Db2 and SQL Server.

Ok, cool, I thought it must. I'll push my other changes then. Thanks, @Naros!

jpechane:
@roldanbob Applied, thanks

PR #3667, opened 2022-06-30
DBZ-5328 field.exclude.list doesn't accept dashes (-) in database name

harveyyue:
https://issues.redhat.com/browse/DBZ-5328


jpechane:
@harveyyue Applied, thanks. Very nice test coverage.

PR #3671, opened 2022-06-30
DBZ-4846 MongoConnector's field exclusion configuration does not work with fields with the same name but from different collections

harveyyue:
https://issues.redhat.com/browse/DBZ-4846


harveyyue:
@jpechane Fix the duplicate key issue.  The useless pathsByAddress map which key is Path::toString cause following duplicate key issue. But the method toString from Path only use the field name as identity, so remove this useless map will fix it.
In my test case, have four Paths as show below, will have the same key with "name" and "active" in pathsByAddress.
*.c1.name,*.c1.active,*.c2.name,*.c2.active

java.lang.IllegalStateException: Duplicate key name (attempted merging values name and name)

PR #3677, opened 2022-07-01
DBZ-5349 Run Oracle connector tests on pushes

Naros:
https://issues.redhat.com/browse/DBZ-5349


Naros:
Applied to main (=2.0) and branch 1.9

PR #3680, opened 2022-07-05
DBZ-5336 Add note about escaping dot (.) characters

Naros:
https://issues.redhat.com/browse/DBZ-5336

Updated incremental snapshot documentation to note that tables that have a dot (.) in their name should be escaped with double-quotes.

### DIFF
--- a/documentation/modules/ROOT/partials/modules/all-connectors/con-connector-incremental-snapshot.adoc
+++ b/documentation/modules/ROOT/partials/modules/all-connectors/con-connector-incremental-snapshot.adoc
@@ -70,6 +70,13 @@ To specify the {data-collection}s to include in the snapshot, provide a `data-co
 The `data-collections` array for an incremental snapshot signal has no default value.
 If the `data-collections` array is empty, {prodname} detects that no action is required and does not perform a snapshot.
 
+[NOTE]
+====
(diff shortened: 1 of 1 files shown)


documentation/modules/ROOT/partials/modules/all-connectors/con-connector-incremental-snapshot.adoc
```diff
@@ -70,6 +70,12 @@ To specify the {data-collection}s to include in the snapshot, provide a `data-co
 The `data-collections` array for an incremental snapshot signal has no default value.
 If the `data-collections` array is empty, {prodname} detects that no action is required and does not perform a snapshot.
 
+[NOTE]
+----
+If one of the {data-collection}s being snapshot contains a dot in the database, schema, or table name, the {data-collection} parts should be escaped with double quotes.
+For example, a table with the name `"My.Table"` in schema `public` should be written as `"public"."My.Table"`.
+----
```
roldanbob, line 77:
```suggestion
====
If the name of a {data-collection} that you want to include in a snapshot contains a dot (`.`) in the name of the database, schema, or table, to add the name to the `data-collections` array, you must escape each part of the name in double quotes. +
For example, to include a table that exists in the `public` schema and that has the name `_My.Table_`, use the following format: `"public"."My.Table"`. 
====
```

Naros:
Applied to 1.9 and main (=2.0)

PR #3682, opened 2022-07-06
DBZ-5351 Add warning for LogMiner rows with unsupported op type

Naros:
https://issues.redhat.com/browse/DBZ-5351


Naros:
Applied to main (=2.0) and branch 1.9

PR #3683, opened 2022-07-06
DBZ-5340 Allow schema prefix on default value function calls

Naros:
https://issues.redhat.com/browse/DBZ-5340


jpechane:
@Naros Are you sure that this is really the right fix? I tried to reproduce the issue using `PostgresSchemaIT.shouldProperlyGetDefaultColumnValues` but I am always getting proper function name without the schema name and no warning. Could you please try to provide a reproducer test for the problem?

Naros:
@jpechane I was able to construct a test case for this, I'll be updating this PR shortly.

jpechane:
@Naros Applied, thanks!

PR #3686, opened 2022-07-06
DBZ-5371 Use initial snapshot position if "$natural:-1" is not supported

Naros:
https://issues.redhat.com/browse/DBZ-5371


jpechane:
@denco37 Please refer to the linked Jira, there is still an issue to be solved.

Naros:
@jpechane I've commented on the issue regarding the capped collection problem.  After talking with a user today who has helped test this particular fix and offered their feedback, we concluded that capped collections are a separate problem from what we're trying to solve here.  More specifically, capped collections are circular buffer-like collections that may simply not be compatible with change streams at all, whereas a non-capped collection is.

Sage-Pierce (approved):
LGTM

Sage-Pierce:
Thanks for such a quick turnaround on this, @Naros!

Naros:

PR #3690, opened 2022-07-07
DBZ-4726: Remove SQL Server single-partition mode

morozov:
**Breaking changes**:

1. The `database.dbname` configuration parameter for the SQL Server connector is no longer supported.
2. The message keys in the Kafka Connect offsets topic are now expected to contain the source database name.
3. The database history is now expected to contain the database name in the source metadata of its messages.
4. The topic and the schema names now include the database name: `server1.dbo.MyTable` → `server1.testDB.dbo.MyTable`.
5. Snapshot SELECT override statements should now include the database name. Not sure how this is going to work with multiple databases. See: https://github.com/debezium/debezium/blob/f465f5ad8672ac4689cfc033e1c105d2e0901fe7/debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SnapshotWithSelectOverridesIT.java#L112-L125 The closest open issue in Jira is [DBZ-4799](https://issues.redhat.com/browse/DBZ-4799) but there's no issue namely about overriding SELECT statements.
6. The connector metrics will be exposed in a different format. See https://github.com/debezium/debezium/pull/2946 for details and examples.

**TODO**:
- [x] Update documentation.

Documenting the upgrade path is tracked in [DBZ-4827](https://issues.redhat.com/browse/DBZ-4827).

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
@@ -103,7 +103,6 @@ public class SqlServerConnection extends JdbcConnection {
 
     private static final String URL_PATTERN = "jdbc:sqlserver://${" + JdbcConfiguration.HOSTNAME + "}:${" + JdbcConfiguration.PORT + "}";
 
-    private final boolean multiPartitionMode;
     private final String getAllChangesForTable;
     private final int queryFetchSize;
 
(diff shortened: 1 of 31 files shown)


jpechane:
@morozov I'll take a look at the PR later but yes breaking changes are acceptable for 2.0.

morozov:
IIRC, if the signaling table is in the same database as the one being incrementally snapshotted, it should be all fine. The existing integration tests should cover this.

The issue is in fact about not being able to incrementally snapshot multiple databases (for which the multi-partition mode is required), not about the impact of switching to the multi-partition mode as such.

morozov:
We haven't tried that. It's possible to incrementally snapshot a single database in the multi-partition mode according to the tests. How does one incrementally snapshot multiple databases or not seems irrelevant to this patch and is the job for DBZ-4799.

jpechane:
@morozov Applied, thanks!

PR #3691, opened 2022-07-07
DBZ-5295 Reselect LOB columns on Primary Key changes

Naros:
https://issues.redhat.com/browse/DBZ-5295

This PR primarily focuses on two critical points:

1. Reselection of LOB column values when an UPDATE changes the row's primary key
2. Performance improvements by caching the LOB column state in `OracleDatabaseSchema`.

The latter point is similar to the TOAST column handling done with PostgreSQL where instead of needing to iterate the relational `Table` model on each change event, we iterate the model once at registration and cache information about whether the specified table has any LOB columns and use this cache during event processing.  This should improve performance, particularly for tables with larger numbers of columns than for those with fewer columns.

While I was unable to reproduce the same behavior with the PG connector, I did add a test case to the PG connector suite in case just so we have coverage for this going forward to avoid regressions.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/BaseChangeRecordEmitter.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/BaseChangeRecordEmitter.java
@@ -5,7 +5,6 @@
  */
 package io.debezium.connector.oracle;
 
-import java.nio.ByteBuffer;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
@@ -38,8 +37,6 @@ public abstract class BaseChangeRecordEmitter<T> extends RelationalChangeRecordE
     private static final Logger LOGGER = LoggerFactory.getLogger(BaseChangeRecordEmitter.class);
 
     private final OracleConnectorConfig connectorConfig;
-    private final ByteBuffer unavailableValuePlaceholderBinary;
-    private final String unavailableValuePlaceholderString;
     private final Object[] oldColumnValues;
     private final Object[] newColumnValues;
     private final OracleDatabaseSchema schema;
@@ -51,8 +48,6 @@ public abstract class BaseChangeRecordEmitter<T> extends RelationalChangeRecordE
         super(partition, offset, clock);
         this.connectorConfig = connectorConfig;
         this.schema = schema;
-        this.unavailableValuePlaceholderBinary = ByteBuffer.wrap(connectorConfig.getUnavailableValuePlaceholder());
-        this.unavailableValuePlaceholderString = new String(connectorConfig.getUnavailableValuePlaceholder());
         this.oldColumnValues = oldColumnValues;
         this.newColumnValues = newColumnValues;
         this.table = table;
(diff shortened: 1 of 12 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/BaseChangeRecordEmitter.java
```diff
@@ -5,31 +5,189 @@
...
+                .append(reselectColumns.stream().map(c -> '"' + c.name() + '"').collect(Collectors.joining(", ")))
+                .append(" FROM ")
+                .append(id.toDoubleQuotedString())
+                .append(" WHERE ");
+
+        for (int i = 0; i < table.primaryKeyColumnNames().size(); ++i) {
+            if (i > 0) {
+                query.append(" AND ");
+            }
+            query.append('"').append(table.primaryKeyColumnNames().get(i)).append("\"=?");
```
jpechane, line 158: Same as above?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/BaseChangeRecordEmitter.java
```diff
@@ -5,31 +5,189 @@
...
+                            ps -> prepareReselectQueryStatement(ps, table, newKey),
+                            rs -> updateNewValuesFromReselectQueryResults(rs, reselectColumns));
+
+                    // newColumnValues have been updated via re-select, re-create the event's value
+                    //
+                    // NOTE: The conversion of the column data must occur within the current connection's context.
+                    // This is because the converters for LOB may make specific callbacks to the underlying database
+                    // and if the converter is called outside the scope of the current LOB object, the call will
+                    // fail due to database connection being unavailable.
+                    newValue = tableSchema.valueFromColumnData(newColumnValues);
```
jpechane, line 103: What will happen if

* the search row has been deleted in between?
* a DDL was executed between the change and this query?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDatabaseSchema.java
```diff
@@ -106,4 +117,63 @@ public boolean isStorageInitializationExecuted() {
...
+     * @return a list of LOB columns, may be empty if the table has no LOB columns
+     */
+    public List<Column> getLobColumnsForTable(TableId id) {
+        return lobColumnsByTableId.getOrDefault(id, Collections.emptyList());
+    }
+
+    /**
+     * Returns whether the provided relational column model is a CLOB or NCLOB data type.
+     */
+    public static boolean isClobColumn(Column column) {
```
jpechane, line 150: Do we handle BLOBs and CLOBs separately somewhere? If not then `isLobColumn` method would be better and simplify the flow.

jpechane (commented):
@Naros Thanks for the change, it is definitely improvement. Still I am a bit concerned how unhappy path - deletes and DDLs are handled.
I there even a way how we could write tests to cover these scenarios?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDatabaseSchema.java
Naros: We do this specifically in the `BaseChangeRecordEmitter#getReselectColumns` as we need to know if it's CLOB or BLOB to resolve which of the unavailable placeholder values types we should compare against to know whether we should reselect a column or not.  I could add an `isLobColumn` to satisfy the check at `LogMinerDmlParser` line 674.

Naros: So I have an idea, at least to encapsulate this and re-use some behavior from `OracleValueConverters`.  This will allow me to change the visibility on `isClobColumn` and `isBlobColumn` to private.  Let me know what you think.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/BaseChangeRecordEmitter.java
jpechane: @Naros I think we should detect the generally erroneous situation, end in a hard-stop with poper explanation but at the same time make it conditional based on `event.processing.failure.handling.mode`. In that case there will be a clear sign that something is wrong and data might be off but you'd be able to get unblocked.

jpechane: I see, it is ok then.

jpechane:
@Naros Applied, thanks a lot!

PR #3693, opened 2022-07-08
DBZ-5358 added name of user who executed transaction to source info b…

nathan-smit-1:
…lock


jpechane:
@Naros This LGTM. I don't think we should backport it to 1.9. Is it needed? Also I saw your discussion in Jira related to XStream. I'd probably keep it as logminer exclusive feature as having xstream yes and no for some of types would be quite confusing.

Naros:
Hi @nathan-smit-1, I took the liberty to rename the field to `user_name` to be consistent with most of the other recently added fields to use snake rather than camel case.  I've also added a small commit to update the documentation with the changes.

Hi @jpechane, I am fine scoping this to 2.0.  I also added a small tip in the documentation to clarify that this field is only populated in the case of LogMiner as we discussed.  If this passes CI, then we can merge this.

PR #3696, opened 2022-07-09
DBZ-5386 Missing "previousId" property with parsing the rename statement in kafka history topic

harveyyue:
https://issues.redhat.com/browse/DBZ-5386


jpechane:
@harveyyue Applied, nice example of defensive coding!

PR #3697, opened 2022-07-11
DBZ-5381 Add a pretty loggable format for commit_scn

Naros:
https://issues.redhat.com/browse/DBZ-5381


jpechane:
@Naros Applied, thanks

PR #3698, opened 2022-07-11
DBZ-5396 Add support for connector-specific relational model attributes

Naros:
https://issues.redhat.com/browse/DBZ-5396

So this is an early draft of the connector-specific relational model attribute support I need for DBZ-3401.  I'd prefer to get some early feedback on the approach in case there are some suggestions or preferences we'd like to take before I get too far into the details and incorporate this into DBZ-3401.

**Why is this needed**
I want to move to a situation where `log.mining.strategy` is no longer a requirement.  Instead, the Oracle connector will use the existing `online_catalog` as the default mode of operation, reducing the overhead on the database and increasing the performance for all Oracle users.  

For this mode to work reliably, we need to be able to support resolving the table/column names for LogMiner redo records when LogMiner fails to do this on its own.  We will do this by matching the redo record's numeric object id with the object id reference we have stored in the connector's database schema relational model.

**What's changed**
This PR introduces a new field called `attributes` both in the JSON-based schema history record and also as a part of the relational model domain object, `Table`.  These attributes are effectively a string-baed key/value tuple.  The purpose of these tuples is entirely driven by the connector and can effectively store anything.

During `JdbcConnection#readSchema`, I've added a hook that allows the connector the ability to resolve a set of attributes for the given `TableId`, and those will be added to the relational model when the `Table` domain object is constructed.  This satisfies the initial set of these from the JDBC schema call.  In addition, the JSON-based schema history serializer has been adjusted to both read and write these values to the schema change events.  This will allow the persistence of these values across restarts and hydrate these at restart time based on the current known table structures and mappings.

- [x] Update other connector repositories based on API changes
- [x] Update documentation

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/listener/CreateTableParserListener.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/listener/CreateTableParserListener.java
@@ -81,7 +81,7 @@ public class CreateTableParserListener extends MySqlParserBaseListener {
         TableId originalTableId = parser.parseQualifiedTableId(ctx.tableName(1).fullId());
         Table original = parser.databaseTables().forTable(originalTableId);
         if (original != null) {
-            parser.databaseTables().overwriteTable(tableId, original.columns(), original.primaryKeyColumnNames(), original.defaultCharsetName());
+            parser.databaseTables().overwriteTable(tableId, original.columns(), original.primaryKeyColumnNames(), original.defaultCharsetName(), original.attributes());
             parser.signalCreateTable(tableId, ctx);
         }
         super.exitCopyCreateTable(ctx);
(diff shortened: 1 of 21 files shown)


Naros:
I do think having some type of typed methods so that the conversion is abstracted away and leaving it up to the connector to use the right conversion methods are most definitely a good call, would keep the code cleaner at the call sites for sure.  But you don't have any issue with the storage of said values being string-based in the schema history record, right?

jpechane:
@Naros No, no issue in this regard, let's handle it just as serialization to a string representation.

Naros:
Hi @jpechane, I'm fine either way on the merge perspective.  I can rebase my Oracle changes on my DBZ-3401 branch against this and pull it in or we can merge this out-right; no real preference.  I'm not sure if any other development is waiting on this effort.

For `Optional<...>`, I'm torn because I think it would help in terms of keeping the API cleaner, but I'm concerned about the performance cost that might come with it when used on a hot path, such as a tight loop during stream processing.  Would the creation of an Optional combined with the use of an `ifPresent` lambda call overhead generate too many temporary objects leading to excessive GC?

Naros:
After some discussion, we agreed to keep the `Attribute` handling as is for now, so I'll proceed and merge.

PR #3704, opened 2022-07-12
[docs] Translate the README document into Korean

giljae:
Translate the README document into Korean.


jpechane:
@giljae Applied, thanks a lot.

PR #3705, opened 2022-07-12
DBZ-5384 org.postgresql.util.PSQLException: Bad value for type timestamp/date/time: CURRENT_TIMESTAMP

harveyyue:
https://issues.redhat.com/browse/DBZ-5384


jpechane:
@harveyyue Applied, thanks!

PR #3707, opened 2022-07-13
DBZ-2766 Remove Guava dependency

jpechane:
https://issues.redhat.com/browse/DBZ-2766


jpechane:
@kgalieva Hi, could you please take a look at this PR? We want to get indendpendent of Guava but there is a code in read-only snapshotting that depends on it. Would it be possible to refactor it so it is no longer dependent? Thanks a lot

kgalieva:
I've open the PR https://github.com/jpechane/debezium/pull/4

kgalieva:
I've found a bug in my change. I'll fix it tomorrow

kgalieva:
Added the fix https://github.com/jpechane/debezium/pull/5

jpechane:
@Naros I've updated the PR and provided an additional one on https://github.com/debezium/debezium-connector-cassandra/pull/68

PR #3713, opened 2022-07-13
DBZ-5404 Restore build compatibility with Kafka 2.x and 3.0.x

Naros:
https://issues.redhat.com/browse/DBZ-5404

### DIFF
--- a/debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleClobDataTypeIT.java
+++ b/debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleClobDataTypeIT.java
@@ -5,8 +5,8 @@
  */
 package io.debezium.connector.oracle;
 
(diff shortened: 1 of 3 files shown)


Naros:
The internal CI job still failed against Kafka 2.x with some references to other missing classes in the Oracle connector and the REST extension.

Naros:
@ani-sha @rk3rn3r I opted to disable the REST extension from the Kafka compatibility job for the time being.  We make use of the generics signature that was added in Kafka 3.x but 2.x didn't use these.  Since I'm not sure what impacts this could have against the `debezium-ui` repository, I opt'ed to take the more simplistic route and disable the module from the check for the time being.  Seeing as the code is explicitly only compatible with Kafka 3.x, it seemed appropriate and I think this likely won't have any impact.

@jpechane Job 41 is what uncovered the last two commit changes on internal CI.  I'm waiting for Job 42 to conclude and if it passes, I think this is done from my side.

jpechane:
@Naros Applied, thanks!

PR #3715, opened 2022-07-14
Cleanup setting of the last snapshot record

vjuranek:
This PR doesn't replace setting of the last snapshot record with doing so in heartbeat message, but only removes setting of the last snapshot event in MySQL connector where it's done twice. Setting last snapshot record once snapshot is finished doesn't seems to be possible without bigger changes as we send the event immediately. If we decide to proceed further, the most easy way seems to be to do in the same way as MySQL - buffer the events and change the last one once we are done. This is also suggested in [DBZ-3113](https://issues.redhat.com/browse/DBZ-3113).

https://issues.redhat.com/browse/DBZ-5047


jpechane:
@vjuranek Could you please repurpose the Jira then? Also could you please check if `OracleSnapshotChangeEventSource` requires the same changes?

vjuranek:
> @vjuranek Could you please repurpose the Jira then? 

done

> Also could you please check if `OracleSnapshotChangeEventSource` requires the same changes?

no, it doesn't do any caching of the last events, so we have to keep same behavior as in `RelationalSnapshotChangeEventSource` (`OracleSnapshotChangeEventSource` overrides the `createSchemaChangeEventsForTables` method from `RelationalSnapshotChangeEventSource`).

jpechane:
@vjuranek Applied, thanks for the analysis

PR #3716, opened 2022-07-14
DBZ-5408 Make sure streaming is fully running before dropping connection

vjuranek:
https://issues.redhat.com/browse/DBZ-5408


Naros (approved):
I can no longer reproduce the failure on `main` with the test fix, great work @vjuranek.

PR #3718, opened 2022-07-14
DBZ-5398 The column is referenced as PRIMARY KEY, but a matching column is not defined in table

harveyyue:
https://issues.redhat.com/browse/DBZ-5398
@jpechane @Naros I couldn't find a graceful solution to parse the column name from index definition including a function. Currently, the regex expression just support the function including parameters which split by comma, and the nested function also not be supported as below.
coalesce(coalesce(col1,'1'),'xxx')

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
@@ -16,6 +16,7 @@ import java.time.Duration;
 import java.util.Objects;
 import java.util.Optional;
 import java.util.concurrent.atomic.AtomicLong;
+import java.util.regex.Pattern;
 
 import org.apache.kafka.connect.errors.ConnectException;
 import org.postgresql.core.BaseConnection;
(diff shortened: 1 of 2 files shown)


Naros:
I can't help but wonder if this is over-engineered and could lead to other problems.  This may not apply to PG but at least for other databases, there are situations where the PK consists of a function call that wraps a synthetic column that isn't naturally available in the JDBC metadata, i.e. ROWID or some other "hidden" column.  I believe PG has similar things like CTID.

So I'm inclined to question whether a better solution for this would have been to flag that the row tuple we identified should be excluded entirely as a unique index or primary key mapping like we do for Oracle.

harveyyue:
@Naros You means we can exclude the entire index columns if including function expression according method isTableUniqueIndexIncluded. I know the following index columns are not database internal or "hidden" column. 
However, I will imeplement the solution as you sugguest.

Create Index Statement:
create unique index counter_ukey on counter(campaign_id, COALESCE(group_id, ''::text));

jpechane:
@harveyyue  I am fine with this soution but will it capture siatuatilons like index being defined as for example `col1 + col2` as I suppose there can be not only function call but arbitrary expression.

jpechane:
Could you please add a test cae similar to https://github.com/debezium/debezium/pull/3728 ?

harveyyue:
> Could you please add a test cae similar to #3728 ?

Ok, will do it.

harveyyue:
@jpechane Add a expression regex including these operators `+ - * / < > = ~ ! @ # % ^ & | ` ?`,  also add a expression test case.
https://www.postgresql.org/docs/current/sql-syntax-lexical.html#SQL-SYNTAX-OPERATORS

Naros (approved):
Overall this LGTM @harveyyue, thanks a lot for this!

Naros:
Applied to main (=2.0).

PR #3720, opened 2022-07-15
Oracle big transational downgrade

mydq:
oracle cdc:   
     big   transactional  downgrade  
     
     explanation:   because oracle logminer uses memory to cache transaction data when mining data, it is easy to cause oom when big things come, so a downgrade scheme is developed.
   
     scope:   only changed to 1.5
   
      config:
            log.mining.init.scn :          Recovery from a certain scn of oracle
            log.mining.big.transactional.cache.path  :  cache file  path       default:  user.dir + file.separator   +  oracle_big_transactional
            log.mining.big.transactional.limit.count  :   big transaction are judged on the basis of   default:10000
            log.mining.big.transactional.skip.id :  big transaction  id that needs to be skipped      
            log.mining.big.transactional.shutdown.limit.count ： big transaction fail outright      default:100000

       add dependency:
             com.alibaba.fastjson2: fastjson2 :  2.0.9


Naros:
Hi @mydq, I'm sorry for Debezium 1.5 is no longer maintained.  A lot of the configuration options you've described have been implemented in various other ways in later releases of the connector.

Debezium 1.7 - Introduced Infinispan off-heap cache support
Debezium 1.8 - Introduced Infinispan cluster cache support
Debezium 1.9 - Introduced `log.mining.buffer.transaction.events.threshold` to control discarding very large transactions

So I would highly suggest giving the latest 1.9.5.Final a try and see whether you believe any of these changes are still applicable.  If they are, please base them against `main` rather than legacy branches as well as open a Jira issue for the change as well as all major code changes must be accompanied by a Jira issue.

Naros:
I also noticed that you added _commited-data-only_ to the connector options, but did you run tests to verify that if a transaction spans across multiple mining sessions whether this works.  From my own testing this never worked as it requires all transactions mined to be mined within the scope of a single mining session rather than across iterative mining loops.

mydq:
Ok, we are based on the transformation of 1.5, later versions may not apply

Naros:
Okay, well unfortunately since branch 1.5 is no longer maintained, we aren't accepting contributions for that version.  You're welcome to use your changes locally if they work for you but unfortunately, we won't be able to integrate these changes as they are and against the branch for which you've opened the PR against.  I hope you understand.

mydq:
ok thank you

PR #3721, opened 2022-07-17
DBZ-5413 Update the topic naming strategy doc to all connectors

harveyyue:
https://issues.redhat.com/browse/DBZ-5413


Naros:
Applied to main (=2.0).

PR #3726, opened 2022-07-19
DBZ-5402: debezium-storage-s3 module

blcksrx:
This implementation provides caching `HistoryRecord` on the main memory, and it acts like  `MemoryDatabaseHistory`
 in the case of recovering records.

 Since S3 does not support Append operation on the object level, `S3DatabaseHistory#start()` fetches history log
 from S3 first and store the `HistoryRecord`'s  on the main memory. 

Also `S3DatabaseHistory#storeRecord(HistoryRecord)` creates new history object every time invokes on S3.

S3 Object keys created names after the `Thread` that creates `DatabaseHistory` object.

### DIFF
--- a/debezium-connector-mysql/src/test/java/io/debezium/relational/history/S3DatabaseHistoryTest.java
+++ b/debezium-connector-mysql/src/test/java/io/debezium/relational/history/S3DatabaseHistoryTest.java
@@ -1,0 +1,125 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.relational.history;
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.URI;
+import java.nio.charset.StandardCharsets;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.stream.Collectors;
+
+import org.apache.http.entity.ContentType;
+import org.junit.AfterClass;
+import org.junit.Assert;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import com.adobe.testing.s3mock.testcontainers.S3MockContainer;
+
+import io.debezium.config.Configuration;
+import io.debezium.document.DocumentReader;
+import io.debezium.storage.s3.history.S3SchemaHistory;
+
+import software.amazon.awssdk.auth.credentials.AnonymousCredentialsProvider;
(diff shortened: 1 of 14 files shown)


debezium-connector-mysql/src/test/java/io/debezium/relational/history/S3DatabaseHistoryTest.java
```diff
@@ -0,0 +1,124 @@
...
+import software.amazon.awssdk.services.s3.S3Client;
+import software.amazon.awssdk.services.s3.model.Bucket;
+import software.amazon.awssdk.services.s3.model.DeleteBucketRequest;
+import software.amazon.awssdk.services.s3.model.DeleteObjectRequest;
+import software.amazon.awssdk.services.s3.model.GetObjectRequest;
+import software.amazon.awssdk.services.s3.model.ListObjectsRequest;
+import software.amazon.awssdk.services.s3.model.S3Object;
+
+public class S3DatabaseHistoryTest extends AbstractDatabaseHistoryTest {
+    final public static String IMAGE_TAG = "2.4.14";
```
Naros, line 41: I would adjust the MySQL pom to pass the `version.s3mock` property into the test suite as a system property and then use `System#getProperty` to resolve this value rather than using a hard-coded string here.  This has the advantage that we adjust the version reference in one spot and that if multiple tests are written across multiple connectors, we again source that version value from one logical place.

Naros (commented):
Hi @blcksrx, overall this LGTM.  I'm going to defer to @jpechane to give his review as well when he returns next week, but in the meantime, I did leave one small comment below regarding the usage of the tag version, if we could address that it'd be great.

blcksrx:
@Naros Thanks a lot for your comments, I'll  wait for @jpechane comments and after that I will do all the changes.

jpechane:
@blcksrx Generally, the PR LGTM. Also what is correct is that the `s3` module is not automaticall packaged into connector distro files.
Still I believe we miss a way how to distribute the storage module. Would you please add one more assmebly that will create distribution files for staorage modules and S3 will be the one to use it? In that case the user will just download the zip with storage JAR and all dependencies and will unpack it into the plugin repository on the Kafka Connect installation.

blcksrx:
@jpechane Thank you very much about your note. I totally understand your concerns about the distribution package, I also believe that we could make it better. I do my best to provide a better approach

blcksrx:
@jpechane Thank you very much. 
Indeed I believe that we could do better by adjusting the `DatabaseHistory` to make connectors independent of the storage module and also move tests from MySql connectors. what is your opinion?

jpechane:
@blcksrx +1 to moving tests. Could you please elaborate a bit more on `DatabaseHistory` proposed change?

PR #3735, opened 2022-07-21
DBZ-5249 KafkaDatabaseHistory without check database history topic create result caused UnknowTopicOrPartitionException

smallYellowCat:
https://issues.redhat.com/browse/DBZ-5249

KafkaDatabaseHistory without check database history topic create result caused UnknowTopicOrPartitionException

### DIFF
--- a/debezium-storage/debezium-storage-kafka/src/main/java/io/debezium/storage/kafka/history/KafkaDatabaseHistory.java
+++ b/debezium-storage/debezium-storage-kafka/src/main/java/io/debezium/storage/kafka/history/KafkaDatabaseHistory.java
@@ -24,6 +24,7 @@ import java.util.function.Consumer;
 import org.apache.kafka.clients.admin.AdminClient;
 import org.apache.kafka.clients.admin.AdminClientConfig;
 import org.apache.kafka.clients.admin.Config;
+import org.apache.kafka.clients.admin.CreateTopicsResult;
 import org.apache.kafka.clients.admin.DescribeTopicsResult;
 import org.apache.kafka.clients.admin.NewTopic;
 import org.apache.kafka.clients.admin.TopicDescription;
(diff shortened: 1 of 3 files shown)


debezium-storage/debezium-storage-kafka/src/main/java/io/debezium/storage/kafka/history/KafkaDatabaseHistory.java
```diff
@@ -532,7 +533,8 @@ public void initializeStorage() {
 
             topic.configs(Collect.hashMapOf(CLEANUP_POLICY_NAME, CLEANUP_POLICY_VALUE, RETENTION_MS_NAME, Long.toString(RETENTION_MS_MAX), RETENTION_BYTES_NAME,
                     Long.toString(UNLIMITED_VALUE)));
-            admin.createTopics(Collections.singleton(topic));
+            CreateTopicsResult result = admin.createTopics(Collections.singleton(topic));
+            result.all().get(30L, TimeUnit.SECONDS);
```
Naros, line 537: Could we define a `Field` called `kafka.topic.create.timeout.ms` that behaves similarly to the existing field called `kafka.query.timeout.ms`?  I think the 30-second delay is acceptable, but I'd prefer this to be something a user can tweak on an as-needed basis. In addition, the `ref-connector-configuration-database-history-properties.adoc` documentation partial should be updated to include this new property.

Naros (commented):
Hi @smallYellowCat, I've added a comment inline.

debezium-storage/debezium-storage-kafka/src/main/java/io/debezium/storage/kafka/history/KafkaDatabaseHistory.java
smallYellowCat: ok, thank u for your sugestions, i will fix it.

Naros (approved):

Naros:
Thanks, @smallYellowCat, applied to main (=2.0).

PR #3742, opened 2022-07-25
DBZ-5448 Add PostgreSQL 15 to jenkins test matrix

Naros:
https://issues.redhat.com/browse/DBZ-5448


Naros:
This depends on https://github.com/debezium/container-images/pull/298, which itself has some components that are awaiting merge and the availability of the postgis component for 15 which I could not locate.

jpechane:
Incorporated in https://github.com/debezium/debezium/pull/3971

PR #3744, opened 2022-07-26
DBZ-5333 Caused by: java.io.EOFException: Failed to read next byte from position 2005308603

harveyyue:
https://issues.redhat.com/browse/DBZ-5333

This reading byte issue should happen occasionally, the `MySqlErrorHandler` handle following exception message to restart connector is good approach.

`Caused by: java.io.EOFException: Failed to read next byte from position 2005308603`


Naros (approved):
LGTM, thanks @harveyyue.

Naros:

PR #3747, opened 2022-07-27
DBZ-5422 Update link format and multiple edits

roldanbob:
[DBZ-5422](https://issues.redhat.com/browse/DBZ-5422)

While testing the downstream documentation for Debezium 1.9, the build failed because a cross-reference in the upstream `con-introduction.debezium-adoc` file, which is used as-is in the downstream build, uses a format that is incompatible with the downstream build for two reasons:
* The link uses the `xref` macro to link to a book that in the downstream docs is external to the current context. Downstream, such links require use of the `link` prefix. 
* The link target specifies only a general context, but omits use of a specific anchor id, which is required to target content downstream.
I manually fixed this link in the released downstream documentation. This change is necessary to ensure that the link isn't reverted when fetching upstream files in the future. 

In addition to updating the link format, this change includes updates to improve readability and clarity, and new links to the deployment instructions for each connector.

Tested the upstream and downstream documentation in local builds.

This should be backported to 1.9, JIC, a refresh is required over the next few weeks.

### DIFF
--- a/documentation/modules/ROOT/partials/modules/tutorial/con-introduction-debezium.adoc
+++ b/documentation/modules/ROOT/partials/modules/tutorial/con-introduction-debezium.adoc
@@ -14,5 +14,5 @@ By taking advantage of Kafka's reliable streaming platform, {prodname} makes it 
 Even if your application stops unexpectedly, or loses its connection, it does not miss events that occur during the outage.
 After the application restarts, it resumes reading from the log from the point where it left off.
 
-{prodname} includes multiple connectors, and each of them share similar capabilities.
+{prodname} includes multiple connectors, which share similar capabilities.
 The tutorial that follows focuses on the {link-prefix}:{link-mysql-connector}#debezium-connector-for-mysql[MySQL connector].
--- a/documentation/modules/ROOT/partials/modules/tutorial/con-introduction-debezium.adoc
+++ b/documentation/modules/ROOT/partials/modules/tutorial/con-introduction-debezium.adoc
@@ -1,18 +1,30 @@
-// Metadata created by nebel
-//
-// UserStory:
-
 [id="introduction-debezium"]
 = Introduction to {prodname}
 
 {prodname} is a distributed platform that converts information from your existing databases into event streams, enabling applications to detect, and immediately respond to row-level changes in the databases.
 
-{prodname} is built on top of http://kafka.apache.org[Apache Kafka] and provides {link-kafka-docs}.html#connect[Kafka Connect] compatible connectors that capture changes from database management systems.
-{prodname} records the history of the data changes that occur in a database in Kafka logs.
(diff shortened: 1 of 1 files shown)


documentation/modules/ROOT/partials/modules/tutorial/con-introduction-debezium.adoc
```diff
@@ -1,19 +1,30 @@
...
 = Introduction to {prodname}
 
-{prodname} is a distributed platform that turns your existing databases into event streams,
-so applications can see and respond immediately to each row-level change in the databases.
+{prodname} is a distributed platform that converts information from your existing databases into event streams, enabling applications to detect, and immediately respond to row-level changes in the databases.
+
+{prodname} is built on top of http://kafka.apache.org[Apache Kafka] and provides a set of {link-kafka-docs}.html#connect[Kafka Connect] compatible connectors.
+Each of the connectors works with a specific database management system (DBMS).
+Connectors record the history of data changes in the DBMS by detecting changes as they occur, and streaming a record of each change event to the Kafka logs.
+Consuming applications can then read the resulting event records from the log.
```
Naros, line 9: I wonder if using the term `topics` rather than `logs` would be more appropriate in this context.

documentation/modules/ROOT/partials/modules/tutorial/con-introduction-debezium.adoc
```diff
@@ -1,19 +1,30 @@
...
+
+By taking advantage of Kafka's reliable streaming platform, {prodname} makes it possible for applications to consume changes that occur in a database correctly and completely.
+Even if your application stops unexpectedly, or loses its connection, it does not miss events that occur during the outage.
+After the application restarts, it resumes reading from the log from the point where it left off.
```
Naros, line 13: See my comment above about `topic` vs `log`.

Naros (commented):
Hi @roldanbob just a few questions below.

documentation/modules/ROOT/partials/modules/tutorial/con-introduction-debezium.adoc
roldanbob: Ha, funny that you should say that. I had changed it, but reverted it to `logs`, because I thought that maybe the original writer intentionally used that language for some reason. I'm very happy to go with `topics`!

Naros: I think from a technical perspective someone really close to Kafka likely might refer to it as `logs`, but when you move just beyond that to the common lay person, the `topics` term seems more appropriate and clearer to me.  It also avoids any confusion in context as we often refer to "transaction logs" from a CDC perspective and this keeps that distinction crystal clear to me.

Naros (approved):

Naros:

PR #3748, opened 2022-07-27
DBZ-5455  Fixed debezium parse default value bug in JdbcValueConverters convertBigInt method

youyangkou:
issue link: https://issues.redhat.com/browse/DBZ-5455
DBZ-5455 When there are spaces in the Default value of the numeric type in a new table DDL under the same Mysql Host; MySQL can be executed successfully,But debezium return exception Bug Fixed.


Naros (changes requested):
Since this is related to default value handling, this change should actually be done inside the `MySqlDefaultValueConverter` class instead.

Naros:
Also @youyangkou, can you please rework the commits and avoid the merge commit please, thanks.

youyangkou:
@Naros ok,i will commit it again and close this PR.

PR #3753, opened 2022-07-28
DBZ-5455 Fixed debezium parse binlog default value bug in JdbcValueCo…

youyangkou:
issue link: https://issues.redhat.com/browse/DBZ-5455

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -433,3 +433,4 @@ Mark Allanson
 Rahul Khanna
 Pengwei Dou
 Zhongqiang Gong
+Youyang Kou
--- a/debezium-core/src/main/java/io/debezium/jdbc/JdbcValueConverters.java
+++ b/debezium-core/src/main/java/io/debezium/jdbc/JdbcValueConverters.java
@@ -925,7 +925,7 @@ public class JdbcValueConverters implements ValueConverterProvider {
                 r.deliver(NumberConversions.getLong((Boolean) data));
             }
             else if (data instanceof String) {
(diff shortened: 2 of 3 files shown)


youyangkou:
@Naros yesterday，i see your comment on PR3748，it should be be done inside the MySqlDefaultValueConverter class instead theoretically，But this Exception link to JdbcValueConverters class , so i will fix this bug firstly。secondly i will do it in MySqlDefaultValueConverter class after this PR,thanks.

Naros:
Hi @youyangkou so the `MySqlDefaultValueConverter` itself delegates to the `JdbcValueConverter` after it has parsed the default value either from a given DDL statement or from the JDBC metadata.  So if you trim the value inside the default value converter, you should have no reason to trim the value inside the other class, which is shared by all connectors.  Does that make sense?

Naros:
Hi @youyangkou, I took a look and this was already solved as a part of [DBZ-4150](https://issues.redhat.com/browse/DBZ-4150).  While you have opened this against the `main` repository, your Jira issue says this is an issue with 1.6 and I cannot reproduce this on 1.9 nor main.  Can you confirm this is still a problem against `main` rather than just 1.6?

youyangkou:
> 
@Naros Sorry, this is indeed a bug in version 1.6

Naros:
Thanks for confirming, so unfortunately branch 1.6 is no longer maintained.  I would suggest you upgrade to 1.9.5.Final, which is the current latest stable version and this fix would be unnecessary.  I'm afraid since DBZ-4150 fixes this on `main`, this PR is superflous, ergo I am going to close it.

PR #3755, opened 2022-07-28
Retry failed events when sending them to HTTP sink

vjuranek:
https://issues.redhat.com/browse/DBZ-5307

### DIFF
--- a/debezium-server/debezium-server-http/src/main/java/io/debezium/server/http/HttpChangeConsumer.java
+++ b/debezium-server/debezium-server-http/src/main/java/io/debezium/server/http/HttpChangeConsumer.java
@@ -99,27 +99,36 @@ public class HttpChangeConsumer extends BaseChangeConsumer implements DebeziumEn
             LOGGER.trace("Received event '{}'", record);
 
             if (record.value() != null) {
-                String value = (String) record.value();
-                HttpResponse r;
-
-                HttpRequest request = requestBuilder.POST(HttpRequest.BodyPublishers.ofString(value)).build();
-
-                try {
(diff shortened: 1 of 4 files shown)


debezium-server/debezium-server-http/src/main/java/io/debezium/server/http/HttpChangeConsumer.java
```diff
@@ -99,27 +110,42 @@ public void handleBatch(List<ChangeEvent<Object, Object>> records, DebeziumEngin
...
-                }
-                else {
-                    LOGGER.info("Failed to publish event: " + r.body());
+                int attempts = 0;
+                while (!recordSent(record)) {
+                    attempts++;
+                    if (attempts >= retries) {
+                        throw new DebeziumException("Exceeded maximum number of attempts to publish event " + record);
+                    }
+                    Thread.sleep(retryInterval);
```
Naros, line 119: Would it make sense to use `Metronome` here rather than the explicit `Thread#sleep` method call?

Naros (commented):
Just one comment inline @vjuranek otherwise LGTM.

debezium-server/debezium-server-http/src/main/java/io/debezium/server/http/HttpChangeConsumer.java
vjuranek: Yes, definitely makes sense. I vaguely remembered Debezium has some utility for it and wanted to use it, but failed to remember/find it. Thanks!

Naros:
Applied to main (=2.0) and branch 1.9.  Thanks @vjuranek.

PR #3758, opened 2022-07-30
DBZ-4727 Use larger chunk size for pause test

Naros:
https://issues.redhat.com/browse/DBZ-4727


vjuranek (approved):

Naros:
Checked Oracle both locally and on CI, the use of `50` works fine.  Once CI finishes here, we can merge.

Naros:
The SQL Server failure is unrelated, I believe we have a Jira issue tracking that one and trying to find a resolution since it seems to be random when it fails.

Naros:
Squashed and merged to `main`, thanks @vjuranek !

PR #3759, opened 2022-07-31
DBZ-5434 Translation from mongodb document to kafka connect schema fails when nested arrays contain no elements

harveyyue:
https://issues.redhat.com/browse/DBZ-5434


Naros:

PR #3763, opened 2022-08-02
DBZ-4631 Fix "No maximum LSN recorded" log message can be spammed on low-activity databases

ggaborg:


Naros:
Hi @ggaborg can you please run the check style check using:
```
./mnw clean install -DskipITs -DskipTests
```

Also, thanks for the contribution, but there was also a discussion on the associated Jira by @morozov and @zalmane about the use of a programmatic way to also perform this check.  Do you think the three of you could come to a consensus on what you'd like to see adjusted here and how that would work?

ggaborg:
Yes, sorry for that. I also use 'mvn process-sources'.
This commit addresses the log spamming only. We'll need another issue for addressing the programmatic check of the Agent status.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerStreamingChangeEventSource.java
```diff
@@ -154,9 +156,16 @@ public boolean executeIteration(ChangeEventSourceContext context, SqlServerParti
 
                 // Shouldn't happen if the agent is running, but it is better to guard against such situation
                 if (!toLsn.isAvailable()) {
-                    LOGGER.warn("No maximum LSN recorded in the database; please ensure that the SQL Server Agent is running");
+                    if (!lsnNotAvailable) {
+                        lsnNotAvailable = true;
+                        LOGGER.warn(
+                                "No maximum LSN recorded in the database; this may happen if there are no changes recorded in the change table yet or low activity database where the cdc clean up job periodically clears entries from the cdc tables. Otherwise, this may be an indication that the SQL Server Agent is not running.");
```
morozov, line 162: IMO, this is a half-measure: it reduces the spam but will still lead to a false-positive warning. My personal preference is to have this logic removed entirely and have the way to check if the agent is running described in the documentation. Requesting certain privileges for the connector just in order to allow it to produce warnings to the log looks like overkill to me.

If a connector is expected to always produce changes, this should be monitored outside of the connector (e.g. at the level of the corresponding Kafka topic(s)). Otherwise, there may be a variety of other reasons why the connector doesn't produce changes (e.g. because it was not deployed).

ggaborg: This change is about reducing the spam and making the message more clear.
I think we can remove from the log the part about the SQLServer Agent after we'll have another alternative in place, either in documentation or programmatic. At that moment, this log can be put on DEBUG.
WDYT?

morozov: Yes, the message is clearer but it still puts the responsibility on the user to analyze the root cause. It is not helpful.

> [...] after we'll have another alternative in place, either in documentation or programmatic.

Does it really need to be done as a separate step? I don't think we should be making any intermediate code changes before we agree on the path forward.

> At that moment, this log can be put on DEBUG. WDYT?

This would make the message effectively a no-op in production environments (which is actually what we do, we have a patch that replaces `warn` with `debug`). In this case, why not just get rid of it?

ggaborg: Hi,
I think we all agree that the programmatic check of Agent status is the correct answer, but I think that is a long shot.
Until then, this change has many advantages over the current implementation:
- it stops the spam
- it is clearer
- even if we document the Agent and the Agent is started at debezium startup, there may be situations when the Agent is crashing, stopped for maintenance purposes, etc. - in this case, the log is still helpful
So I suggest we take it in steps: 1. address the current issue with this change, 2. open issue to document the Agent, 3. open issue for programmatic check of Agent status.
Thanks

zalmane: @jpechane Please direct us how to proceed. We are fine with removing altogether, or separating the second part (querying the SQLAgent status) into a separate ticket.

Naros: Hi @zalmane , as I indicated in Zulip, the same PR can be used for both but this can be done with two separate commits.  This allows us to track the change in behavior to this entire bit of code by 1 issue but 2 commits allow us to see the evolution of the change as it took place.  Is there really a reason to delegate the JDBC bit to a second issue and task entirely?

zalmane: No reason other than saving time as I saw several users baffled by this message and the attached PR is a quick fix while the other will take a bit more time. Let me know what you decide and we will move forward accordingly.

Naros: Well, I think we all have concluded that the programmatic check is the ideal path.  What I don't have a feel for is what is the effort in doing that check.  If it's something that would take a few days to sort out and test, then I'd much rather we do that than the alternative with follow-up tasks, particularly if those follow-up tasks linger or are never addressed. 

> even if we document the Agent and the Agent is started at debezium startup, there may be situations when the Agent is crashing, stopped for maintenance purposes, etc. - in this case, the log is still helpful

From what I understand, the ideal approach is to supplement the `toLsn.isAvailable()` check with something like this:
```
if (!dataConnection.isAgentRunning()) {
  // log message about SQL not available
  return false;
}
else if (!toLsn.isAvailable()) {
  // log nothing here, simply return
  return false;
}
```
This prevents the false-positive, logs the message immediately when we detect an unavailable agent, and if it's just the connector waiting on a specific state to become available we don't log anything.  

Furthermore, to avoid the possibility of an extended period where the else-if branch fires too many times, we could look at logging a specifically different error message because we know the agent is up but we aren't getting data in a reasonable time period.  

What do you guys think?

ggaborg: Hi,

agent status check can be done only if no lsn, to reduce number of checks. The final implementation could look like this:

if (!toLsn.isAvailable()) {
  String status = dataConnection.fetchAgentStatus();
  if ("running".equals(status)) {
    // all good, just return; could also log debug
    return false;
  }
  if (status == null) {
    // when db admin did not give or does not want to give required permissions to the sql user
	// log message about missing permissions
	// this can spam, so we can reduce spam logging only first time in an uninterrupted serries of nulls
	return false;
  }
  // log error problem because agent status different than 'running'
  return false;
}

morozov: > ```java
> // when db admin did not give or does not want to give required permissions to the sql user
> // log message about missing permissions
> ```

In this case, the method that attempts to fetch the agent status should be allowed to throw an exemption. Otherwise, the connector will not be able to log the root cause of the failure to fetch the status.

I'd keep the method named as `boolean isAgentRunning()`. There's no need to expose the knowledge about the status and its expected value from the connection to the connector.

jpechane:
Hi,

I think we should

- Slightly change the wording of the message
- Ensure it is emitted only once
-- The flag variable should be removed

This will be fix for both 1.9 and 2.0 branch

For 2.0there should be another Jira/PR that will check if the config is correc but at the same time we need to avoid repeat of https://issues.redhat.com/browse/DBZ-5496 so the check will be optional to pass

ggaborg:
pushed entire rebase

PR #3765, opened 2022-08-02
DBZ-4329 Don't send fileds with obsolete values

vjuranek:
During incremental snapshot fields `xmin`, `lsn` and `txId` are not
updated and therefore we stream obsolete values from previous records
in `source` struct. Don't include these obsolete value in the `source`
struct during incremental snapshot.

https://issues.redhat.com/browse/DBZ-4329

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresEventMetadataProvider.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresEventMetadataProvider.java
@@ -45,10 +45,14 @@ class PostgresEventMetadataProvider implements EventMetadataProvider {
         if (source == null) {
             return null;
         }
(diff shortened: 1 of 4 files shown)


Naros (commented):
Overall this LGTM @vjuranek but I'm wondering whether this also applies to other connectors during their incremental snapshot execution or is this only related to PG?

vjuranek:
> Overall this LGTM @vjuranek but I'm wondering whether this also applies to other connectors during their incremental snapshot execution or is this only related to PG?

Ops, this was mentioned also in the Jira, but I forgot to check it. Thanks for the reminder, will update here with my findings.

vjuranek:
@Naros so I believe only Postgres is affected. MySQL and Oracle has `tableEvent()` which does update `SourceInfo` without updating other items, but these are not used during incremental snapshot AFAICT.

Naros:

PR #3769, opened 2022-08-03
DBZ-2663 | Update mysql binlog connector library and TRANSACTION_PAYLOAD event support

debjeetsarkar:
More Ref:
https://debezium.zulipchat.com/#narrow/stream/302533-dev/topic/Handling.20Transaction.20Payload.20Events.20for.20Mysql/near/289650440
https://github.com/osheroff/mysql-binlog-connector-java/pull/76#issuecomment-1185629635


pom.xml
```diff
@@ -120,7 +120,7 @@
         <!-- Database drivers, should align with databases -->
         <version.postgresql.driver>42.4.0</version.postgresql.driver>
         <version.mysql.driver>8.0.29</version.mysql.driver>
-        <version.mysql.binlog>0.25.6</version.mysql.binlog>
+        <version.mysql.binlog>0.26.0</version.mysql.binlog>
```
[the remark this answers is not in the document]
debjeetsarkar: Done, updated.

Naros:
Hi @debjeetsarkar so first off I don't have any issues upgrading the binlog client version in `main`, I'd prefer we avoid this in the `1.9` branch unless there is an explicit reason for this to be backported.  

That said, I have one question regarding whether it makes sense to use this same PR for the remaining work for DBZ-2663, to implement the compression binlog support.  We prefer to avoid situations where a Jira issue spans multiple releases and we'll likely be looking at a Beta2 release in the near future, so I think we have two ways forward:

1. Log a new Jira specifically for upgrading the binlog client to `0.26.1` and reword the commit message with that Jira.
2. Add additional commits here for DBZ-2663 that introduce the compression support and we merge the two as one PR.

I don't have a preference either way, its really entirely what you want to do, but I do want to avoid DBZ-2663 spanning multiple versions if possible.

debjeetsarkar:
@Naros , makes sense to have all the changes in a single PR. Have added the support for TRANSACTION_PAYLOAD event type. Cheers !!

support/checkstyle/src/main/resources/io/debezium/checkstyle/message.properties
```diff
@@ -1,4 +1,4 @@
 import.unused=Unused import - {0}.
 import.invalidJavaDocTag=Invalid JavaDoc tag - {0}
 header.missing=Missing a header - not enough lines in file. Every source file must contain the standard header; see 'checkstyle/src/main/resources/debezium.header'.
-header.mismatch=Line does not match expected header line of ''{0}''. Every source file must contain the standard header; see 'checkstyle/src/main/resources/debezium.header'.
\ No newline at end of file
+header.mismatch=Line does not match expected header line of ''{0}''. Every source file must contain the standard header; see 'checkstyle/src/main/resources/debezium.header'.
```
Naros, line 4: This looks superfluous, could you revert this line ending modification?

Naros (commented):
Overall LGTM, just one comment @debjeetsarkar.  Otherwise waiting on CI.

support/checkstyle/src/main/resources/io/debezium/checkstyle/message.properties
debjeetsarkar: Sure, done.

jpechane:
@debjeetsarkar Applied, thanks a lot!

PR #3771, opened 2022-08-04
DBZ-5472 Support CREATE INDEX using TABLESPACE with quoted identifiers

Naros:
https://issues.redhat.com/browse/DBZ-5472

I confirmed that the upstream ANTLR grammar already supports this syntax, so this change is based on the upstream project.


Naros:

PR #3772, opened 2022-08-04
DBZ-5476 Fix record accumulator during statistics log entry

Naros:
https://issues.redhat.com/browse/DBZ-5476


Naros:

PR #3774, opened 2022-08-04
DBZ-2193 Investigate the replacement of LB services with port redirects in OpenShift test-suite

MartinMedek:


debezium-testing/debezium-testing-system/src/main/java/io/debezium/testing/system/tools/databases/AbstractOcpDatabaseDeployer.java
```diff
@@ -81,6 +80,13 @@ public T deploy() {
 
         this.services = svcs;
 
+        String dbName = deployment.getMetadata().getLabels().get("app");
+        ServiceResource<Service> serviceResource = ocp.services().inNamespace(project).withName(dbName);
```
jcechace, line 84: Can this be typed as ``Service``?

debezium-testing/debezium-testing-system/src/main/java/io/debezium/testing/system/tools/databases/AbstractOcpDatabaseDeployer.java
```diff
@@ -81,6 +80,13 @@ public T deploy() {
...
+        int servicePort = serviceResource.get().getSpec().getPorts().get(0).getPort();
```
jcechace, line 85: I would forward all the port exposed by the service.

debezium-testing/debezium-testing-system/src/main/java/io/debezium/testing/system/tools/databases/AbstractOcpDatabaseController.java
```diff
@@ -102,21 +85,11 @@ public String getPublicDatabaseHostname() {
             LOGGER.info("Running from OCP, using internal database hostname");
             return getDatabaseHostname();
         }
-        awaitIngress();
-        return getLoadBalancedService().getStatus().getLoadBalancer()
-                .getIngress().get(0).getHostname();
+        return "localhost";
```
jcechace, line 88: Make this a constant. Something like ``FORWARDED_HOST``

jcechace:
Good job as a PoC. Two things which I would like to see


1) Move the port forwarding into Controller. Ideally into a standalone method which would be called from initialisation method. 
2) The port forwarding should be stopped at the end of the test class -- a teardown method of Fixture class comes to mind as the first place where to put it.

jcechace (changes requested):

jcechace:
@MartinMedek  pleas rebase this. It's about time we get this merged. Apology for the delay.

obabec:
@jcechace Martin will also add kafka external listener with route and ssl to this PR.

debezium-testing/debezium-testing-system/src/test/java/io/debezium/testing/system/fixtures/kafka/DockerKafka.java
```diff
@@ -33,7 +35,7 @@ public DockerKafka(@NotNull ExtensionContext.Store store) {
     }
 
     @Override
-    public void setup() {
+    public void setup() throws IOException {
```
jcechace, line 38: This needs to be there I suppose coz of a sibling class throwing this...

debezium-testing/debezium-testing-system/src/test/java/io/debezium/testing/system/tests/MongoConnectorTest.java
```diff
@@ -0,0 +1,18 @@
...
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.testing.system.tests;
+
+import io.debezium.testing.system.assertions.KafkaAssertions;
+import io.debezium.testing.system.tools.kafka.ConnectorConfigBuilder;
+import io.debezium.testing.system.tools.kafka.KafkaConnectController;
+import io.debezium.testing.system.tools.kafka.KafkaController;
+
+public class MongoConnectorTest extends ConnectorTest {
```
jcechace, line 13: I'm likely overlooking something. How is this different from `SqlConnectorTest` and why we need these 2 classes?

debezium-testing/debezium-testing-system/src/main/java/io/debezium/testing/system/tools/databases/AbstractOcpDatabaseController.java
```diff
@@ -92,36 +89,89 @@ public String getDatabaseHostname() {
...
+            if (isLocalPortFree(portNum)) {
+                return portNum;
+            }
+        }
+        throw new IllegalStateException("Couldn't find free port for forwarding");
+    }
+
+    private boolean isLocalPortFree(int port) {
+        try {
+            new ServerSocket(port).close();
```
jcechace, line 170: try with resources works here.

Also note on this approach port randomization. In this particular case you are likely safe as you are using the port from the same process where this socket was previously open. In general however, this line can leave the TCP socket in state `wait_time` and make it unusable for about 2 minutes. 

Also since we are actually not using the socket but rather close it and use the formal port number, it is a subject to race possible race condition. However there is no better way unless you can pass over the socket instance directly :)

jcechace: Also -- considering this approach works here -- you can do 

```java
try( var socket = new ServerSocket(0) {
   return socket.getLocalPort();
}
```


To get a random free port assigned by the os, effectively getting rid of the random number and attempt checking generation and saving few lines of code.

debezium-testing/debezium-testing-system/src/main/java/io/debezium/testing/system/tools/kafka/OcpKafkaController.java
```diff
@@ -125,4 +142,38 @@ private NonNamespaceOperation<Kafka, KafkaList, Resource<Kafka>> kafkaOperation(
...
+        }
+
+        // download and decode certificate
+        String cert = secret.getData().get(KAFKA_CERT_FILENAME);
+        byte[] decodedBytes = Base64.getDecoder().decode(cert);
+        cert = new String(decodedBytes);
+
+        // save to local file
+        File crtFile = Files.createTempFile("kafka-cert-", null).toFile();
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(crtFile, UTF_8))) {
```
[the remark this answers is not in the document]
obabec: Agree

debezium-testing/debezium-testing-system/src/main/java/io/debezium/testing/system/tools/kafka/KafkaController.java
```diff
@@ -46,7 +47,7 @@ public interface KafkaController {
     /**
      * @return default kafka consumer configuration
      */
-    default Properties getDefaultConsumerProperties() {
+    default Properties getDefaultConsumerProperties() throws IOException {
```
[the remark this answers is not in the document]
obabec: Agree

debezium-testing/debezium-testing-system/src/main/java/io/debezium/testing/system/tools/databases/AbstractOcpDatabaseController.java
MartinMedek: This is awesome, thanks

obabec (approved):
LGTM

PR #3777, opened 2022-08-05
DBZ-5479 Retry MongoNodeIsRecoveringException when failover

nicholas-fwang:
https://issues.redhat.com/browse/DBZ-5479

## Bug report
Debezium connector task is not retried when failover in mongodb 5

I tested mongo debezium connector for failover in mongo 4 and 5.

In case of 4, when failover, the task is retried successfully with MongoSocketOpenException on tail

but, in case of 5, the task is not retried with MongoNodeIsRecoveringException on tail

it looks mongo didn't close socket when primary failover in 5 version like recovery mode.

follow [this doc about the exception](https://mongodb.github.io/mongo-java-driver/3.6/javadoc/com/mongodb/MongoNodeIsRecoveringException.html)

then, we need to retry the exception

## What Debezium connector do you use and what version?
v1.8.0.Final

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -433,3 +433,4 @@ Mark Allanson
 Rahul Khanna
 Pengwei Dou
 Zhongqiang Gong
+Inki Hwang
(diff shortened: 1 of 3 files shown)


Naros (approved):
LGTM @nicholas-fwang, I'll merge once CI has concluded.

Naros:
Applied to main (=2.0) and brach 1.9.

PR #3781, opened 2022-08-08
DBZ-5489 add column comment to schema 

jaegwonseo:
see https://issues.redhat.com/browse/DBZ-5489

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -345,6 +345,7 @@ Scofield Xu
 Sara Fonseca
 Sean Rooney
 Sebastian Bruckner
+Seo Jae-kwon
 Sergei Morozov
 Shane Paul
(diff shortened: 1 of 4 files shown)


Naros:
Hi @jaegwonseo there is an outstanding question on the Jira issue, if you could answer that.

jaegwonseo:
Hi @Naros  i'm sorry for the late reply
accoding to avroschema compare method if the comment in DB(mysql, postgresql, ...) is changed, new schema version gonna registry in schema registry
https://github.com/confluentinc/schema-registry/blob/master/client/src/main/java/io/confluent/kafka/schemaregistry/avro/AvroSchema.java#L212

jpechane:
@jaegwonseo Applied, thanks a lot

PR #3783, opened 2022-08-08
DBZ-5488 Support undocumented DATAPUMP DDL clauses

Naros:
https://issues.redhat.com/browse/DBZ-5488


ani-sha (approved):
LGTM!

jpechane:
@Naros Applied, thanks. Without docs we cannot do any better :-(

PR #3787, opened 2022-08-09
DBZ-0000 Simple test for UI build

Naros:


Naros:
Replaced officially with PR #3792.

PR #3789, opened 2022-08-10
DBZ-2534 Synchronize on engine start and stop

vjuranek:
Current solution when we synchronize on `runningThread` is not
sufficient as e.g. sequence of start/stop/start in different threads
can result in state when the first thread runs `start()`, second calls
`stop()` and sets `runningThread` to `null`, but before the first thread
notices that `runningThread` is `null`, another thread calls `start()`
and as `runningThread` is `null` by that time, it will continue with
`start()` and therefore there will be two threads running `start()`
method on the same engine instance.

To prevent it, synchronize threads when setting `runningThread` and
also block during the whole `stop()` method to make sure no other thread
will call `start()` until stop finishes.

This doens't fix all possible issues, there is still possibility that
`stop()` will be called before `run()` is actually executed, but this
has to be coordinated by the upper level.

https://issues.redhat.com/browse/DBZ-2534

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
@@ -78,7 +78,9 @@ import io.debezium.util.VariableLatch;
  * <p>
  * Embedded connectors are designed to be submitted to an {@link Executor} or {@link ExecutorService} for execution by a single
  * thread, and a running connector can be stopped either by calling {@link #stop()} from another thread or by interrupting
- * the running thread (e.g., as is the case with {@link ExecutorService#shutdownNow()}).
+ * the running thread (e.g., as is the case with {@link ExecutorService#shutdownNow()}). If the engine is stopped by calling
+ * {@link #stop()} method, the application that runs the engine is responsible for sufficient coordination of the threads to make
+ * sure that {@link #stop()} is not called before the engine is actually running.
  *
  * @author Randall Hauch
  */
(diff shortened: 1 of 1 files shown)


vjuranek:
Split into two commits to make the review more easy.

jpechane:
@vjuranek Hi, I wonder, can we avoid synchronized blocks etc. and introduce a `Lock` that will ensure that `stop()` will be executed only after a critical part of `run()` was completed and at the same time the `run()` will not be started if `stop()` was executed before it?

vjuranek:
>  I wonder, can we avoid synchronized blocks etc. and introduce a `Lock` 

Not sure if it would help much, as `synchronized` is basically a shortcut for `Lock` (using current instance as a lock object)

> that will ensure that `stop()` will be executed only after a critical part of `run()` was completed 

We can make bigger part of `run()` synchronized (or use dedicated lock for it). The issue is how to make sure `stop()` is not called before `run()`. We can adjust `stop()` to e.g. throw an exception when engine is not running, but in such case the exception has to be handled on the upper level (same if add dedicated flag and wait for `start()` to set this flag - if wait times out before this flag is set, upper level has to handle it).

> and at the same time the `run()` will not be started if `stop()` was executed before it?

Not sure how to do this. IMHO even if we have dedicated flag for it, how would we distinguish ordinary `run()`, `stop()`, `run()` executed in this order from `run()`, `stop()` executed in `stop(), run()` order? If we add e.g. some `EngineState` enum, in the first case it will be `NOT_INITIALIZED` -> (`run()`) -> `RUNNING` -> (`stop()`) -> `STOPPED`, second `run()` will check that that the state is in `[NOT_INITIALIZED, STOPPED]` and will proceed. In the second case it will be `NOT_INITIALIZED` -> (`stop()`) -> `STOPPED`, subsequent `run()` will check that that the state is in `[NOT_INITIALIZED, STOPPED]` and will proceed. Here the answer may be that `stop()` must not be called when the engine is in state `NOT_INITIALIZED` (possibly only when engine is in the state `RUNNING`), but this leads to the previous question: what to do in such case.

We already return `false` from the `stop()` method when it's called but the engine is not running. It can be used by upper level code to make sure it's called in proper sequence. We may add some convenient method like `ensureRunning(int timeout)` to make it more easy for the users, when adding some upper level code coordination. But throwing an exception from `stop()` would require coordination on upper level as well.

Or any other ideas how to fix it?

vjuranek:
This issue was addressed in new async engine.

PR #3791, opened 2022-08-10
DBZ-5500 Use org.apache.kafka.connect.converters.ByteArrayConverter

romilpunetha:
Mongodb stores byte[] as a base64-encoded string. When using `io.debezium.converters.ByteArrayConverter`, the output I was getting was that of a byte[] for the base64 encoded string. What I needed was the byte[] for the base64 decoded value, ie, the original byte[]. `org.apache.kafka.connect.converters.ByteArrayConverter` converter helped me get the desired output.


Naros:
Hi @romilpunetha I'm curious did you check to see whether the `BinaryDataConverter` or the `ByteBufferConverter`, also in the `io.debezium.converters` package, worked rather than explicitly relying on the one from Connect?

romilpunetha:
Oddly enough, I can't seem to reproduce it. Spent 2 days trying to fix it and now I can't reproduce the original issue. 
I tried all 3, `ByteArrayConverter`,  `BinaryDataConverter` and the apache one that I'm using. I'm getting the exact data that I need to deserialize the avro object. Closing this issue.

PR #3795, opened 2022-08-11
DBZ-5504 Removes references to `MONEY` and `SMALLMONEY` data types

roldanbob:
[DBZ-5504](https://issues.redhat.com/browse/DBZ-5504)

Removes invalid entries from table in **Data type mappings** section.

Tested in local Antora build.


Naros:
Thanks for the quick turn around @roldanbob, I may have a few more, specifically `DATETIME2` and `SMALLDATETIME`.  I'm not sure where these came from but they're not allowed in CREATE TABLE statements from what I can tell on 11.5.0.  Let me verify before we commit this.

roldanbob:
> Thanks for the quick turn around @roldanbob, I may have a few more, specifically `DATETIME2` and `SMALLDATETIME`. I'm not sure where these came from but they're not allowed in CREATE TABLE statements from what I can tell on 11.5.0. Let me verify before we commit this.

@Naros It's likely that I pulled in these entries when I created or tried to synch the Db2 (and earlier, the Oracle) connector _Data types_ sections, copying them from the other connector docs in my early days on the project, not knowing that data types were not universally supported. :\

Naros:
I'm fairly certain when I looked at the committers on these bits, it appeared to me these (at least some) were from the Db2 folks, so I just wanted to be clear.  But based on the internal email thread, let's remove the following:

- MONEY
- SMALLMONEY
- DATETIME2
- SMALLDATETIME

roldanbob:
@Naros The table of data types in the Db2 documentation includes multiple entires for DATETIME2:
>
|`DATETIME2(0)`, `DATETIME2(1)`, `DATETIME2(2)`, `DATETIME2(3)`
|`INT64`
|`io.debezium.time.Timestamp` +
 +
Represents the number of milliseconds since the epoch, and does not include timezone information.

|`DATETIME2(4)`, `DATETIME2(5)`, `DATETIME2(6)`
|`INT64`
|`io.debezium.time.MicroTimestamp` +
 +
Represents the number of microseconds since the epoch, and does not include timezone information.

|`DATETIME2(7)`
|`INT64`
|`io.debezium.time.NanoTimestamp` + 


Should I remove all of these?

roldanbob:
Also, further on, the _Timestamp types_ section includes the following example:

> ... the `DATETIME2` value "2018-06-20 15:13:16.945104" is represented by an `io.debezium.time.MicroTimestamp` with the value "1529507596945104"

If I modify that to reference the `DATETIME` type and `io.debezium.time.Timestamp`, do I have to also modify the values? That is, do the values in the following sentence still make sense?: 

_... the `DATETIME` value "2018-06-20 15:13:16.945104" is represented by an `io.debezium.time.Timestamp` with the value "1529507596945104"_

Naros:
Hi @roldanbob probably not, but I am still verifying that bit as well.  One thing I noticed is that we mention specific Nano, Micro, etc in the documentation but the implementation doesn't seem to adhere to that at all.  I've hit a small snag in testing the behavior with Db2 so once I can get over that hurdle I'll let you know.

PR #3797, opened 2022-08-12
DBZ-5035 Redis Sink config properties are not passed to DB history

ggaborg:


jpechane:
@ggaborg Applied, thanks!

PR #3799, opened 2022-08-13
DBZ-5511 Keep consistence with method serialize of Connect/Json Table Change Serializer

harveyyue:
https://issues.redhat.com/browse/DBZ-5511

### DIFF
--- a/debezium-core/src/main/java/io/debezium/relational/history/ConnectTableChangeSerializer.java
+++ b/debezium-core/src/main/java/io/debezium/relational/history/ConnectTableChangeSerializer.java
@@ -45,6 +45,8 @@ public class ConnectTableChangeSerializer implements TableChanges.TableChangesSe
     public static final String AUTO_INCREMENTED_KEY = "autoIncremented";
     public static final String GENERATED_KEY = "generated";
     public static final String COMMENT_KEY = "comment";
+    public static final String DEFAULT_VALUE_EXPRESSION = "defaultValueExpression";
+    public static final String ENUM_VALUES = "enumValues";
 
     private final Schema columnSchema;
     private final Schema tableSchema;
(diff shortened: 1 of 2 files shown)


debezium-core/src/main/java/io/debezium/relational/history/ConnectTableChangeSerializer.java
```diff
@@ -45,6 +46,9 @@ public class ConnectTableChangeSerializer implements TableChanges.TableChangesSe
     public static final String AUTO_INCREMENTED_KEY = "autoIncremented";
     public static final String GENERATED_KEY = "generated";
     public static final String COMMENT_KEY = "comment";
+    public static final String HAS_DEFAULT_VALUE = "hasDefaultValue";
```
jpechane, line 49: Hi, I wonder if we need this one. In case of default value not being present `defaultValueExpression` would just be omitted. If present but `null` then it will be set to `null`.

debezium-core/src/main/java/io/debezium/relational/history/ConnectTableChangeSerializer.java
```diff
@@ -66,6 +70,9 @@ public ConnectTableChangeSerializer(SchemaNameAdjuster schemaNameAdjuster) {
                 .field(AUTO_INCREMENTED_KEY, Schema.OPTIONAL_BOOLEAN_SCHEMA)
                 .field(GENERATED_KEY, Schema.OPTIONAL_BOOLEAN_SCHEMA)
                 .field(COMMENT_KEY, Schema.OPTIONAL_STRING_SCHEMA)
+                .field(HAS_DEFAULT_VALUE, Schema.BOOLEAN_SCHEMA)
+                .field(DEFAULT_VALUE_EXPRESSION, Schema.OPTIONAL_STRING_SCHEMA)
+                .field(ENUM_VALUES, SchemaBuilder.array(Schema.STRING_SCHEMA).build())
```
jpechane, line 75: IMHO this should be `optional()` too.

harveyyue: fixed it.

jpechane:
@harveyyue Applied, thanks!

PR #3804, opened 2022-08-17
DBZ-5043 Change namespace of pass-through parameters

vjuranek:
Changes in this PR:
* Change namespace from `database.history` to `schema.history.internal`.
* Move driver pass-through parameters into `driver` namespace.
* Use `topic.prefix` as the connector logical name.
* Replace `database.server.name` with `topic.prefix`.
* Replace `mongodb.name` with `topic.prefix`.

The changes were discussed in the Jira, but this PR contains two differences:
* Replacing `mongodb.name` with `topic.prefix` was not discuss there at all. I included it for consistency as `mongodb.name` plays the same role in MongoDB connector as `database.server.name` in other connectors, so the prefix is always determined by `topic.prefix`.
* In the Jira I suggested to use connector name as the connector logical name. However, it turned out that it defaults to `database.server.name` and by changing it we would break JXM and lots of tests. So I decided to use `topic.prefix` as the logical name to avoid further refactoring. Still applies that use cannot set logical name directly. If we decide to remove logical name in the future, this refactoring would be just a vaste of time now, if we decide to introduce new config property for this logical name, we would have to do this refactoring again.


vjuranek:
@jpechane please review

SQL server failure is AFAICT random and not related to this PR, other builds fail due to missing changes in respective repos - will send PRs once this one is reviewed.

jpechane:
@vjuranek @ani-sha @Naros We also need to think about the impact on UI. This unfortunately needs to go hand-in-hand.

jpechane:
Overall LGTM

@morozov Could you please check it conforms what we agreed upon.

Still I've identified one inconsistency. In documentation we speak about schema history topic and database history topic.

Schema history topic is public facing and conforms to the topic prefix. Then we have database history topic with the DDL history recorded. The problem now is that we use prefix `schema.history` to configure database history which can be a bit confusing.

How about doing a partial step back and use prefix `history.database.`?

vjuranek:
> I'd propose to name the parameter for the internal topic as "internal.schema.history" but IIRC the "internal." prefix has a special meaning.

Yes, `internal` is already used for parameters which are not intended to be used by the users and IMHO this is good as the name suggests that it's internal parameter and user probably shouldn't touch the internals (unless fully understand what it does). What about `schema.internal.history`? `history.database` LGTM as well.

When in it, currently public schema history is named according to `database.server.name` (or `topic.prefix` with this PR), which is not very intuitive and cannot be configured without changing all other topics. Shouldn't we add separate option for configuring public (user) schema history?

> 
> Besides inconsistency in the documentation, I see some inconsistency in the code:
> 
> https://github.com/debezium/debezium/blob/6dcf6481013adaae4efb58833834409d045086e9/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java#L739-L748
> 
> Is the above code related to the internal schema history topic and should be modified?

It's internal schema history. What exactly do you see as inconsistency? Shall we name `DATABASE_HISTORY` constant, [DatabaseHistory](https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/relational/history/DatabaseHistory.java) interface etc?
 
> As for the separation of JDBC parameters into two namespaces: "database." and "driver.", it looks like the latter is always merged with the former, so technically, users will be able to pass "driver.xyz" as "database.xyz", and it will work.

Yes, we create one big config, so in the end we merge everything together.

> Besides that, if alongside "database.host" they will specify "driver.host", the former may get overwritten which might look like a bug.

`database.*` should take precedence and `driver.*` will be added only when there isn't such config key.

> I haven't seen any documentation updates that would describe the usage of the "driver." parameters, so my understanding may be wrong.

ops, I forgot update documentation for this - I'll will fix it.

morozov:
> What about `schema.internal.history`? `history.database` LGTM as well.

The former looks good to me given the limitations. I don't really like the latter. The topic tracks _schema_ changes. The "database" in the name might be misleading because a database has a schema and stores data but the topics track only schema (not data) changes.

> What exactly do you see as inconsistency?

The fact that the variable named "database history" is assigned a value from the parameter named "schema history". Then the display name states: "Database history class", the same is about the description.

> When in it, currently public schema history is named according to `database.server.name` (or `topic.prefix` with this PR), which is not very intuitive and cannot be configured without changing all other topics. Shouldn't we add separate option for configuring public (user) schema history?

If I were designing it from scratch, I'd use some the defaults like `${topic.prefix}.schema.history` and `${topic.prefix}.schema.history.internal` with the ability to override either of them. So yeah, if the former cannot be currently overridden, adding a parameter might be a good idea, especially if we will change the default (which is not very intuitive).

jpechane:
Includes fix for https://issues.redhat.com/browse/DBZ-5582

vjuranek:
> > What about `schema.internal.history`? `history.database` LGTM as well.
> 
> The former looks good to me given the limitations. I don't really like the latter. The topic tracks _schema_ changes. The "database" in the name might be misleading because a database has a schema and stores data but the topics track only schema (not data) changes.

any opinion @jpechane ?

 
> > What exactly do you see as inconsistency?
> 
> The fact that the variable named "database history" is assigned a value from the parameter named "schema history". Then the display name states: "Database history class", the same is about the description.

this should be addressed now @morozov , I renamed (hopefully) everything
 
> > When in it, currently public schema history is named according to `database.server.name` (or `topic.prefix` with this PR), which is not very intuitive and cannot be configured without changing all other topics. Shouldn't we add separate option for configuring public (user) schema history?
> 
> If I were designing it from scratch, I'd use some the defaults like `${topic.prefix}.schema.history` and `${topic.prefix}.schema.history.internal` with the ability to override either of them. So yeah, if the former cannot be currently overridden, adding a parameter might be a good idea, especially if we will change the default (which is not very intuitive).

I missed that there actually *is* an option for it ([database.history.kafka.topic](https://debezium.io/documentation/reference/1.9/connectors/mysql.html#mysql-property-database-history-kafka-topic)), but not available for all connectors (no idea why e.g. Postgres is missing it).

jpechane:
`schema.internal.history` - this sounds like a good solution
`database.history.kafka.topic` should become `schema.internal.history.kafka.topic`

Only a subset of connecotrs need to retain schema history, postgres for example gets actual schema with every single message coming from database so there is no need for it.

Right now I don't intend to make the internal topic dependent on the `topic.prefix`. That should be intended as public facing topic namespace.

Let's go back again to managed Debezium - in that case the user might not need to configure internal schema history at all and a new topic in his namespace could be a bit confusing.

vjuranek:
> `schema.internal.history` - this sounds like a good solution

Done, changed `schema.history` to `schema.internal.history`. The change is in separate commit as there was quite a lot of conflicts when I tried to squash it with original change of `database.history` to `schema.history`.

> `database.history.kafka.topic` should become `schema.internal.history.kafka.topic`

Yes, it's done in this PR (I used `database.*` as I was referring existing documentation).
 
> Only a subset of connecotrs need to retain schema history, postgres for example gets actual schema with every single message coming from database so there is no need for it.

ah, right, thanks (I'm constantly forgetting that postgres sends schema with every message :-( )

vjuranek:
I think this is finally ready for review @jpechane and @morozov . Thanks!

morozov:
> `schema.internal.history` - this sounds like a good solution

If it's not too late, what do you think if we make it `schema.history.internal` instead? "schema history" is an established combination, so it might be better if we keep it intact instead of breaking apart with the "internal" qualifier. The qualifier belongs to the topic, so it should go either before the "schema history" (which we cannot do) or after (which we can).

In the latter case, the parameter `schema.internal.history.kafka.topic` will turn into `schema.history.internal.kafka.topic` which also makes sense because it's an "internal Kafka topic", not "internal history".

vjuranek:
No, it's not too late, sound good to me and if @jpechane agrees with that, it's actually quite easy to change as I didn't squash these commits.

jpechane:
@morozov @vjuranek Just in time, I wanted to merge the PR. So yes, it makes sense please do the final adjustment and I'll get it in. As is said variable naming is the hardest part of computer science :-).

vjuranek:
Changed to `schema.history.internal`. Should be ready @jpechane . SQL server test failure is unrelated/random.

jpechane:
@vjuranek Applied, thanks a lot for this huge task. Please monitor debezium-testing module tests tomorrow if they are passign with the most recent nightly image.

jpechane:
@vjuranek Could you please take a look at https://debezium.io/documentation/reference/1.9/connectors/cassandra.html#cassandra-topic-names ? I think there should be a PR for Cassandra that will use the same config option name. This should be also reflected in the docs.

vjuranek:
@jpechane Cassandra PR is https://github.com/debezium/debezium-connector-cassandra/pull/71 and documentation was (probably by accident) already changed in https://github.com/debezium/debezium/commit/daf329bc06

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
```diff
@@ -736,19 +733,19 @@ public static SecureConnectionMode parse(String value, String defaultValue) {
...
             .withWidth(Width.LONG)
             .withImportance(Importance.LOW)
             .withInvisibleRecommender()
-            .withDescription("The name of the DatabaseHistory class that should be used to store and recover database schema changes. "
+            .withDescription("The name of the SchemaHistory class that should be used to store and recover database schema changes. "
                     + "The configuration properties for the history are prefixed with the '"
-                    + DatabaseHistory.CONFIGURATION_FIELD_PREFIX_STRING + "' string.")
-            .withDefault(KafkaDatabaseHistory.class.getName());
+                    + SchemaHistory.CONFIGURATION_FIELD_PREFIX_STRING + "' string.")
+            .withDefault(KafkaSchemaHistory.class.getName());
```
obelix74, line 748: @vjuranek this is breaking setting up a kafka connect configuration using the REST API.

Running 

```
curl -i -X POST -H "Accept:application/json" -H "Content-Type:application/json" localhost:8083/connectors/ -d '{ "name": "inventory-connector", "config": { "connector.class": "io.debezium.connector.mysql.MySqlConnector", "tasks.max": "1", "database.hostname": "mysql", "database.port": "3306", "database.user": "mysqluser", "database.password": "mysqlpw", "database.server.id": "223344", "database.server.name": "dbserver1", "database.include.list": "inventory", "database.history.kafka.bootstrap.servers": "kafka:9092", "database.history.kafka.topic": "dbhistory.inventory" } }'
```

fails with 

```
2022-09-26 21:39:21,455 WARN   ||  unhandled due to prior sendError   [org.eclipse.jetty.server.HttpChannelState]
javax.servlet.ServletException: org.glassfish.jersey.server.ContainerException: java.lang.NoClassDefFoundError: Could not initialize class io.debezium.connector.mysql.MySqlConnectorConfig
	at org.glassfish.jersey.servlet.WebComponent.serviceImpl(WebComponent.java:410)
	at org.glassfish.jersey.servlet.WebComponent.service(WebComponent.java:346)
	at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:366)
	at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:319)
	at org.glassfish.jersey.servlet.ServletContainer.service(ServletContainer.java:205)
	at org.eclipse.jetty.servlet.ServletHolder.handle(ServletHolder.java:799)
	at org.eclipse.jetty.servlet.ServletHandler.doHandle(ServletHandler.java:550)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:233)
	at org.eclipse.jetty.server.session.SessionHandler.doHandle(SessionHandler.java:1624)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextHandle(ScopedHandler.java:233)
	at org.eclipse.jetty.server.handler.ContextHandler.doHandle(ContextHandler.java:1434)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:188)
	at org.eclipse.jetty.servlet.ServletHandler.doScope(ServletHandler.java:501)
	at org.eclipse.jetty.server.session.SessionHandler.doScope(SessionHandler.java:1594)
	at org.eclipse.jetty.server.handler.ScopedHandler.nextScope(ScopedHandler.java:186)
	at org.eclipse.jetty.server.handler.ContextHandler.doScope(ContextHandler.java:1349)
	at org.eclipse.jetty.server.handler.ScopedHandler.handle(ScopedHandler.java:141)
	at org.eclipse.jetty.server.handler.ContextHandlerCollection.handle(ContextHandlerCollection.java:234)
	at org.eclipse.jetty.server.handler.StatisticsHandler.handle(StatisticsHandler.java:179)
	at org.eclipse.jetty.server.handler.HandlerWrapper.handle(HandlerWrapper.java:127)
	at org.eclipse.jetty.server.Server.handle(Server.java:516)
	at org.eclipse.jetty.server.HttpChannel.lambda$handle$1(HttpChannel.java:400)
	at org.eclipse.jetty.server.HttpChannel.dispatch(HttpChannel.java:645)
	at org.eclipse.jetty.server.HttpChannel.handle(HttpChannel.java:392)
	at org.eclipse.jetty.server.HttpConnection.onFillable(HttpConnection.java:277)
	at org.eclipse.jetty.io.AbstractConnection$ReadCallback.succeeded(AbstractConnection.java:311)
	at org.eclipse.jetty.io.FillInterest.fillable(FillInterest.java:105)
	at org.eclipse.jetty.io.ChannelEndPoint$1.run(ChannelEndPoint.java:104)
	at org.eclipse.jetty.util.thread.QueuedThreadPool.runJob(QueuedThreadPool.java:883)
	at org.eclipse.jetty.util.thread.QueuedThreadPool$Runner.run(QueuedThreadPool.java:1034)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: org.glassfish.jersey.server.ContainerException: java.lang.NoClassDefFoundError: Could not initialize class io.debezium.connector.mysql.MySqlConnectorConfig
	at org.glassfish.jersey.servlet.internal.ResponseWriter.rethrow(ResponseWriter.java:254)
	at org.glassfish.jersey.servlet.internal.ResponseWriter.failure(ResponseWriter.java:236)
	at org.glassfish.jersey.server.ServerRuntime$Responder.process(ServerRuntime.java:438)
	at org.glassfish.jersey.server.ServerRuntime$1.run(ServerRuntime.java:263)
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:248)
	at org.glassfish.jersey.internal.Errors$1.call(Errors.java:244)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:292)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:274)
	at org.glassfish.jersey.internal.Errors.process(Errors.java:244)
	at org.glassfish.jersey.process.internal.RequestScope.runInScope(RequestScope.java:265)
	at org.glassfish.jersey.server.ServerRuntime.process(ServerRuntime.java:234)
	at org.glassfish.jersey.server.ApplicationHandler.handle(ApplicationHandler.java:680)
	at org.glassfish.jersey.servlet.WebComponent.serviceImpl(WebComponent.java:394)
	... 30 more
Caused by: java.lang.NoClassDefFoundError: Could not initialize class io.debezium.connector.mysql.MySqlConnectorConfig
	at io.debezium.connector.mysql.MySqlConnector.config(MySqlConnector.java:76)
	at org.apache.kafka.connect.runtime.AbstractHerder.validateConnectorConfig(AbstractHerder.java:462)
	at org.apache.kafka.connect.runtime.AbstractHerder.lambda$validateConnectorConfig$2(AbstractHerder.java:367)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	... 1 more
```

Has the curl command / REST interface to configure connectors changed because of this PR?

vjuranek: hi @obelix74 , yes, this is breaking change. New major release 2.0 will contain breaking changes (therefore included into new major version) and this is one of these breaking changes, i.e. it is expected. You have to replace `database.history` with `schema.history.internal` and `database.server.name` with `topic.prefix` (see PR summary for all change in this PR). With this change it should work.

PR #3807, opened 2022-08-18
DBZ-5526 Support JSON-based CHECK conditions

Naros:
https://issues.redhat.com/browse/DBZ-5526

Upstream PR: https://github.com/antlr/grammars-v4/pull/2773


jpechane:
@Naros Applied, thanks!

PR #3811, opened 2022-08-19
DBZ-5439 Add transaction id associate with commit scn

Naros:
https://issues.redhat.com/browse/DBZ-5439

Hi @thangdc94 I've reworked your PR by addressing the following outstanding items.  To avoid conflicts with your current branch, I've elected to open up this new PR based on my fork instead but with your commits cherry-picked.

1. Fixed your commit with your missing GH email address.
2. Used `TreeSet` rather than `HashSet` to guarantee output order of the formatted string
3. Removed `RS_ID` and `SSN` from the `CommitScn`, they're still part of the source info block, but not in the offsets.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -439,3 +439,4 @@ Zhongqiang Gong
 Inki Hwang
 崔世杰
 合龙 张
+Phạm Ngọc Thắng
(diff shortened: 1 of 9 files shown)


thangdc94:
thanks @Naros

jpechane:
@thangdc94 @Naros Applied, thanks!

PR #3812, opened 2022-08-19
DBZ-5516: Fix examples in SQL Server connector documentation

morozov:


jpechane:
@morozov Applied, thanks

PR #3813, opened 2022-08-20
DBZ-5535 Disable preferring DDL before logical schema in history recovery

harveyyue:
https://issues.redhat.com/browse/DBZ-5535


jpechane:
@harveyyue Applied, thanks!

PR #3819, opened 2022-08-23
DBZ-5484 Better error reporting in SKIP/WARN cases

ani-sha:
https://issues.redhat.com/browse/DBZ-5484

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
@@ -248,12 +248,12 @@ public class EventDispatcher<P extends Partition, T extends DataCollectionId> im
                 case WARN:
                     LOGGER.warn(
                             "Error while processing event at offset {}",
-                            changeRecordEmitter.getOffset().getOffset());
+                            changeRecordEmitter.getOffset().getOffset(), e);
                     break;
                 case SKIP:
                     LOGGER.debug(
                             "Error while processing event at offset {}",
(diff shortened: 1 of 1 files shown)


debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
```diff
@@ -249,12 +249,12 @@ public void changeRecord(P partition,
                     LOGGER.warn(
                             "Error while processing event at offset {}",
                             changeRecordEmitter.getOffset().getOffset());
-                    break;
+                    throw new ConnectException("Error while processing event at offset " + changeRecordEmitter.getOffset().getOffset(), e);
```
jpechane, line 252: We don't want to throw new exception. We just want to modify this logging message and the one below to contain the full exception.

debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
```diff
@@ -248,13 +248,11 @@ public void changeRecord(P partition,
                 case WARN:
                     LOGGER.warn(
                             "Error while processing event at offset {}",
-                            changeRecordEmitter.getOffset().getOffset());
-                    break;
+                            changeRecordEmitter.getOffset().getOffset(), e);
```
jpechane, line 251: You should not remove the `break` statements. Also could you please try to reproduce the situation and post here from the log how the new error message looks like? Thanks a lot

Naros (approved):
LGTM now, once CI finishes I'll merge.  Thanks @ani-sha.

PR #3820, opened 2022-08-23
DBZ-5546 Add missing value format in debezium-server doc

BetaCat0:
Add missing value format type `eventclouds`  ([code ref](https://github.com/debezium/debezium/blob/539d9a39588bfecf75513433b2e51f5eaf81571c/debezium-server/debezium-server-core/src/main/java/io/debezium/server/DebeziumServer.java#L186)) for debezium-server doc.


jpechane:
@BetaCat0 Applied, thanks!

PR #3821, opened 2022-08-23
DBZ-5122 Include Oracle Debezium Connector in Debezium Server distribution

ggaborg:
DBZ-5122 Include Oracle Debezium Connector in Debezium Server distribution

also checked the lib folder of Debezium Server distribution for duplicate libraries with different version and there is none

### DIFF
--- a/debezium-server/debezium-server-dist/pom.xml
+++ b/debezium-server/debezium-server-dist/pom.xml
@@ -64,14 +64,18 @@
                     <groupId>io.debezium</groupId>
                     <artifactId>debezium-connector-sqlserver</artifactId>
                 </dependency>
-		<dependency>
+                <dependency>
+                    <groupId>io.debezium</groupId>
+                    <artifactId>debezium-connector-oracle</artifactId>
+                </dependency>
+                <dependency>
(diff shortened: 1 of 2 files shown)


Naros:
I'm not sure whether this will be adequate.  The Oracle connector doesn't ship with the driver due to licensing restrictions, won't we need to explicitly depend on the Oracle `odjcb8.jar` in the pom, excluding that dependency in the distribution archive so that all the Quarkus compile-time and classpath bindings are satisfied?  @jpechane wdyt?

jpechane:
Exactly neither oracle JDBC driver nor xstream JAR can be present in the distribution.

ggaborg:
@Naros, @jpechane The change only includes the Oracle Connector without Oracle driver and xtream. We should also add to the Debezium Server documentation a link to the "Obtaining the Oracle JDBC driver and XStream API files" section of the Oracle Connector documentation

jpechane:
OK, I can confirm the forbidden dependencies are not present as they are marked as provided.
@ggaborg Could you please provide the docs update and we can merge the PR? Thanks

jpechane:
@ggaborg Applied, thanks!

PR #3822, opened 2022-08-23
DBZ-5550: Add EMPTY as a keyword that can be used as name

morozov:


jpechane:
@morozov Applied, thanks!

PR #3823, opened 2022-08-23
DBZ-5542 Edits property description; populates empty table cells.

roldanbob:
[DBZ-5542](https://issues.redhat.com/browse/DBZ-5542)

Updates description of mongodb.connection.string property to remove typo, improve clarity, and apply style guidelines.
Also populates empty cells in property table entries with `No default` string. 

Tested in local Antora build.


jpechane:
@jcechace Could you please take a look? The wording LGTM.

documentation/modules/ROOT/pages/connectors/mongodb.adoc
```diff
@@ -1378,17 +1378,25 @@ This limitation will be removed in the next {prodname} release.
...
  +
 [NOTE]
 ====
-At the moment the connection string is used only for the initial replica set discovery. Other
-configuration properties are still used when connecting directly to primary rs members -- with the exception of credential which (when present) is extracted from given connection string.
+The connector uses this connection string only during the initial replica set discovery process.
+During this discovery process, the connector ignores connection values that are specified in other properties (`mongodb.user`, `mongodb.password`, `mongodb.authsource`, SSL configuration properties, and so forth).
+After the discovery process completes, when the connector attempts to establish a direct connection to a primary replica set member, the connector then returns to using the standard connection properties. +
+ +
+When establishing a connection, the connector generally ignores this connection string.
```
jcechace, line 1392: This and the consequent lines implies that in some cases, the connector might use properties from connection string (unless provided directly with the regular connector properties). However at the moment this is true strictly for credential properties.

jcechace (changes requested):

documentation/modules/ROOT/pages/connectors/mongodb.adoc
roldanbob: @jcechace Ok, I think that we're on the same page, but I need to reword this. What I mean is that, except during the discovery phase, the connector ignores the connection string, unless it doesn't find credential information elsewhere in the configuration. In that case, it will extract the credentials from the connection string.

roldanbob: @jcechace Can you confirm whether it is correct that, in establishing a connection to a replica set member, the connector will extract credentials from the connection string if the credentials are not otherwise present in the configuration?

roldanbob: @jcechace I went ahead and made an update based on how I interpreted your feedback from a couple weeks ago.  Please take a look when you get a chance and let me know if any further changes are necessary.

jcechace (approved):
LGTM, merging. 

Thanks @roldanbob

PR #3824, opened 2022-08-24
DBZ-5244 Retry all communication exceptions by default

jpechane:
https://issues.jboss.org/browse/DBZ-5244

We've originally decided to retry everything by default. Thinking about that let's try to limit it to set of exceptions that are usually related to communications like IOException, SqlException etc.


jpechane:
One test failure unrelated https://issues.redhat.com/browse/DBZ-5552

Other caused by this PR

Naros (commented):
Overall this LGTM, I did notice that you left the `OracleErrorHandler` untouched, was that intended?

jpechane:
@Naros Yes, it is. IIRC we agreed on not touching Oracle is it is now quite fine tuned,

Naros:
Thanks @jpechane, applied to main (=2.0).

PR #3828, opened 2022-08-26
DBZ-5557: Supply partition when comiting offsets with source database

ramanenka:
Keep track of partitions offsets being committed belong to in
`BaseSourceTask`. Supply the partition along with the offset in calls to
`commitOffset`.

### DIFF
--- a/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
+++ b/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
@@ -9,6 +9,8 @@ import java.time.Duration;
 import java.time.Instant;
 import java.time.temporal.ChronoUnit;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -74,11 +76,11 @@ public abstract class BaseSourceTask<P extends Partition, O extends OffsetContex
     private ChangeEventSourceCoordinator<P, O> coordinator;
 
     /**
-     * The latest offset that has been acknowledged by the Kafka producer. Will be
+     * The latest offsets that have been acknowledged by the Kafka producer. Will be
      * acknowledged with the source database in {@link BaseSourceTask#commit()}
      * (which may be a no-op depending on the connector).
      */
-    private volatile Map<String, ?> lastOffset;
+    private final Map<Map<String, ?>, Map<String, ?>> lastOffsets = new HashMap<>();
 
     private Duration retriableRestartWait;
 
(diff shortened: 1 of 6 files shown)


debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -178,20 +180,26 @@ void logStatistics(final List<SourceRecord> records) {
...
             SourceRecord lastRecord = records.get(batchSize - 1);
-            lastOffset = lastRecord.sourceOffset();
+            updateLastOffset(lastRecord.sourcePartition(), lastRecord.sourceOffset());
             previousOutputBatchSize += batchSize;
             if (pollOutputDelay.hasElapsed()) {
                 // We want to record the status ...
                 final Instant currentTime = clock.currentTime();
                 LOGGER.info("{} records sent during previous {}, last recorded offset: {}", previousOutputBatchSize,
-                        Strings.duration(Duration.between(previousOutputInstant, currentTime).toMillis()), lastOffset);
+                        Strings.duration(Duration.between(previousOutputInstant, currentTime).toMillis()), lastRecord.sourceOffset());
```
jpechane, line 189: It probably makes sense to log partition here too.

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -282,8 +290,22 @@ public void commit() throws InterruptedException {
...
+                    while (iterator.hasNext()) {
+                        Map.Entry<Map<String, ?>, Map<String, ?>> entry = iterator.next();
+                        Map<String, ?> partition = entry.getKey();
+                        Map<String, ?> lastOffset = entry.getValue();
+
+                        try {
+                            coordinator.commitOffset(partition, lastOffset);
+                            iterator.remove();
+                        }
+                        catch (RuntimeException e) {
```
jpechane, line 304: Why only RuntimeException? And why should be unsuccessfull commit ignored?

jpechane (commented):
@ramanenka LGTM thanks, I've left two comments to address.

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
ramanenka: I added the partition to the log message.

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
ramanenka: > Why only RuntimeException? 

There are no checked exceptions according to `coordinator.commitOffset` method signature. Which leaves us with  runtime exceptions only.

> And why should be unsuccessfull commit ignored?

I was thinking that an unsuccessful commit shouldn't fail the connector as it will have a chance to retry on the next commit attempt. In the use case we're trying to implement (to delete consumed capture instances) it makes sense to  retry until the commit is successful. Alternatively we could just let the exception bubble up and fail the connector which is also a valid option to choose. Which of the two approaches make more sense to you?

jpechane: @ramanenka I'd keep the behaviour as is now. Looking at `commitOffsets()` I've a feeling that in that case in an unpredicted state.

morozov: I agree with @jpechane. Let's focus on introducing the necessary API changes with just as little implementation as necessary.

Once we finalize the implementation for the SQL Server connector, we'll have a better understanding of the needed error handling. If any changes in the implementation of this method will become necessary, we will be able to articulate them better.

ramanenka: I've removed the `try-catch` here. Also I've simplified the code a bit by getting the key set iterator as opposed to the entry set iterator.

jpechane:
@ramanenka Applied, thanks

tooptoop4:
@ramanenka is this a breaking change when upgrading existing versions?

ramanenka:
@tooptoop4 unless you have custom code that overrides `StreamingChangeEventSource.commitOffsert` the is not a breaking change.

PR #3829, opened 2022-08-26
DBZ-5229 Restore Kafka-based database history options in SchemaGenerator

Naros:
https://issues.redhat.com/browse/DBZ-5229

So rather than meddle with what was done in `debezium-core` and `debezium-storage-xxxx` modules, I've opted to simply check whether the connector exposes a `database.history` property and if so, we inject the Kafka-based history fields back into the schema directly.

While this approach is not compatible with any other storage modules, I also think we have a much larger architecture problem we need to think about and design.  Additionally, at the very least, this mitigates the existing problem for components that are based on the schema generator module for now by restoring previous before as it was prior to DBZ-5229.

### DIFF
--- a/debezium-schema-generator/pom.xml
+++ b/debezium-schema-generator/pom.xml
@@ -19,6 +19,10 @@
       <artifactId>debezium-core</artifactId>
     </dependency>
     <dependency>
+      <groupId>io.debezium</groupId>
+      <artifactId>debezium-storage-kafka</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.apache.kafka</groupId>
       <artifactId>kafka-clients</artifactId>
     </dependency>
--- a/debezium-connector-oracle/pom.xml
+++ b/debezium-connector-oracle/pom.xml
@@ -229,6 +229,20 @@
                     </systemPropertyVariables>
                 </configuration>
             </plugin>
+            <plugin>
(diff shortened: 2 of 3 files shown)


debezium-schema-generator/pom.xml
```diff
@@ -18,6 +18,10 @@
       <groupId>io.debezium</groupId>
       <artifactId>debezium-core</artifactId>
     </dependency>
+    <dependency>
+      <groupId>io.debezium</groupId>
+      <artifactId>debezium-storage-kafka</artifactId>
```
rk3rn3r, line 23: Should we add a comment here that this is a temporary fix?

rk3rn3r:
LGTM

rk3rn3r (approved):

debezium-schema-generator/pom.xml
Naros: I don't think so, but we do need to log a Jira issue tho as a follow-up task to think about the architecture used and to see whether we can find a better, more robust way to deal with the database history configuration.  I know that @blcksrx is looking into the validation concerns, so we just don't want to lose sight of it.

Naros: https://issues.redhat.com/browse/DBZ-5564

Naros:
Applied to main (=2.0)

PR #3831, opened 2022-08-29
DBZ-3007: limit 'Streaming requested from LSN' logging

winklerm:
https://issues.redhat.com/browse/DBZ-3007

Introduced the "rate limit" for logging this particular log message at most once in 10 seconds.

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/AbstractMessageDecoder.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/AbstractMessageDecoder.java
@@ -7,12 +7,16 @@ package io.debezium.connector.postgresql.connection;
 
 import java.nio.ByteBuffer;
 import java.sql.SQLException;
+import java.time.Duration;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.debezium.connector.postgresql.TypeRegistry;
 import io.debezium.connector.postgresql.connection.ReplicationStream.ReplicationMessageProcessor;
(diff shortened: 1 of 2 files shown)


debezium-core/src/main/java/io/debezium/util/TimeBasedRateLimiter.java
```diff
@@ -0,0 +1,70 @@
...
+ *   ...
+ *   Time: +5s limiter.allows() == true // out of 5 sec interval, the action should NOT be limited, next interval starts
+ *   Time: +6s limiter.allows() == false // within 5 sec interval, the action should be limited
+ *   ...
+ *   Time: +10s limiter.allows() == true // out of 5 sec interval, the action should NOT be limited, next interval starts
+ *   ...
+ * }
+ * </pre>
+ */
+@NotThreadSafe
```
winklerm, line 37: The class is not designed to be thread safe - if the `AbstractMessageDecoder` needs thread safety of this class, please let me know.

Naros (changes requested):
Hi @winklerm take a look at `RelationalSnapshotChangeEventSource`, there is a method in that class called `getTableScanLogTimer`.  In effect, the new class you've introduced already exists as a `Timer` implementation.  You should be able to use the snapshot event source as a guide for how to make the change in the decoder without the new class implementation.

winklerm:
Hi @Naros thanks for the hint, will take a look!

winklerm:
@Naros Hi, I have updated the PR with your suggested changes. Please take a look, thanks!

Naros (approved):
LGTM @winklerm, if CI passes I'll merge.

PR #3833, opened 2022-08-29
Use TCCL to load configuration classes by default

aloubyansky:
https://issues.redhat.com/browse/DBZ-5561

Unless there is a specific reason to use getClass().getClassLoader() as the default class loader, Thread.currentThread().getContextClassLoader() would be a safer default from the integration perspective with environments that define their own classloading policies.

The original issue was reported in Quarkus [quarkusio/quarkus#27565](https://github.com/quarkusio/quarkus/issues/27565). When a project contains a custom converter implementation, the CommonConnectorConfig in the Debezium core (which is an external project dependency) loading converter classes may fail with a class loading exception. What happens in in this case is Quarkus loads the local project dependencies in a classloader that is higher in the hierarchy than the classloader loading the core Debezium classes, making the CommonConnectorConfig to look for converter implementations among the core Debezium classes and the system classloader, which may lead to a failure.

Any objections? Thanks.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -15,6 +15,7 @@ Alex Soto
 Alexander Iskuskov
 Alexander Kovryga
 Alexander Schwartz
+Alexey Loubyansky
 Alexey Miroshnikov
 Alfusainey Jallow
 Alisa Houskova
--- a/debezium-core/src/main/java/io/debezium/config/Configuration.java
+++ b/debezium-core/src/main/java/io/debezium/config/Configuration.java
@@ -1440,7 +1440,7 @@ public interface Configuration {
      *         configuration but the value could not be converted to an existing class with a zero-argument constructor
      */
     default <T> T getInstance(String key, Class<T> type) {
-        return getInstance(key, type, () -> getClass().getClassLoader());
+        return getInstance(key, type, () -> Thread.currentThread().getContextClassLoader());
     }
 
     /**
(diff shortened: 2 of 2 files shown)


Naros:
Hi @aloubyansky, I believe we likely also use `getClass().getClassLoader()` in some other places as well, particularly when we create connection class implementations for each connector too.  We probably need to revisit all call sites that use this to be safe.  

Could you log a Jira enhancement at https://issues.redhat.com and update the commit message with the DBZ-XXXX issue key?

aloubyansky:
Sure, done @Naros

jpechane:
I'd recommend to create a new util class in `io.debezium.util` that would be responsible for classloading and delegate classloading to it.

jpechane:
How about using this alternative solution? https://github.com/debezium/debezium/pull/3846

aloubyansky:
I find the suggested alternative worse because while it appears the API gives a caller a way to control the class loader to be used to load classes, in fact it uses it as a fallback. Now, while theoretically possible, it's highly unlikely you'll ever run into a situation when the context classloader is null. Which means the caller, practically, doesn't have a chance to pass in a specific classloader to load classes. I would suggest keeping it simple: if a user cares about passing in a specific classloader, the framework should respect that and use that classloader, if a user doesn't provide any classloader then use the default one associated with the current thread.

jpechane:
@aloubyansky Looking at https://cwiki.apache.org/confluence/display/KAFKA/KIP-146+-+Classloading+Isolation+in+Connect Kafka Connect sets TCCL too. So how about getting rid of that parameter completely?

aloubyansky:
This is something that is worth discussing in the team. Unfortunately, I am not familiar enough with the current use-cases to suggest such (api breaking) changes.

jpechane:
@aloubyansky Applied, thanks. I logged DBZ-5585 as a follow-up.

aloubyansky:
Thanks @jpechane

PR #3836, opened 2022-08-30
DBZ-3129 Add OCP cluster on AWS provisioning job

novotnyJiri:
https://issues.redhat.com/browse/DBZ-3129


obabec (approved):
Thanks for the PR. I think job is in pretty good shape. Maybe I would also add to this PR job for deletion of the cluster. WDYT ?

jcechace:
@obabec Agreed. @novotnyJiri  can you make sure we also have the means to tear down the clean and cleanup?

obabec (approved):
Seems that Jiri is going for little different approach and he created one job that deploys and also teardown cluster based on the boolean parameter. I missed this little thing first time, I think we can merge this. WDYT @jcechace ?

PR #3838, opened 2022-08-30
DBZ-5565 Formatting characters no longer render in published doc

roldanbob:
[DBZ-5565](https://issues.redhat.com/browse/DBZ-5565)

Fixes bad formatting of names and examples in several of the connector properties table entries.
Prior to the update, instead of the correct formatting being applied in the published documentation, formatting characters were rendered. 
Tested in local Antora build.


jpechane:
@roldanbob Applied, thanks

PR #3840, opened 2022-08-31
DBZ-5569 Oracle connector's max.queue.size.in.bytes is invalid

harveyyue:
https://issues.redhat.com/browse/DBZ-5569


Naros:
Thanks @harveyyue, applied to main (=2.0) and backported just the task-change to branch 1.9.

PR #3842, opened 2022-09-01
DBZ-5573 fix source positional attributes

ahus1:
This fixes the language type for the source code listings in this page.

I did a search for the full documentation project, and I didn't find any other.


jpechane:
@ahus1 Applied, thanks!

PR #3844, opened 2022-09-01
DBZ-5576 Initial .asciidoctorconfig file to improve editing experience

ahus1:
This sets some attributes for the preview to make the gap between the IDE and the real site smaller, and to reduce the number of warnings shown in the editor as those attributes are now known to the IDE.

The changes in the `antora.yml` make the `community` attribute a soft-set attribute so that it can be overwritten when writing the docs as needed, for example in an `.asciidoctorconfig` file.

Before: 

image

After: 

image

### DIFF
--- a/documentation/.asciidoctorconfig
+++ b/documentation/.asciidoctorconfig
@@ -1,0 +1,10 @@
+// This file provides hints for an IDE how to render the preview, for example, in IntelliJ.
+// This is mostly intended to define attributes.
+// Read more at: https://intellij-asciidoc-plugin.ahus1.de/docs/users-guide/features/advanced/asciidoctorconfig-file.html
+
+// Duplicate values for the preview that are defined in another repository in the Antora playbook
+:prodname: Debezium
+:jira-url: 'https://issues.redhat.com'
+
+// Define attributes that are used downstream so that an editor known that they exist and doesn't show them in warnings.
+:product!:


documentation/antora.yml
```diff
@@ -20,7 +20,7 @@ asciidoc:
     apicurio-version: '2.2.5.Final'
     db2-version: '11.5.0.0'
     ojdbc8-version: '21.6.0.0'
-    community: true
+    community: '@'
```
Naros, line 23: Hi @ahus1 I must admit, I'm not familiar with this syntax.  What exactly does `'@'` imply?  Is this in the documentation somewhere that explains this?

ahus1: The '@' at the end of the value implies a "soft set" attribute, see https://docs.antora.org/antora/latest/component-attributes/#soft-set for more information. The `@` is the marker for the "soft set", and will be stripped / will not be part of the value of the attribute.

This would allow you to override the attribute `community` in the `.asciidoctorconfig` file in a later step. Theoretically it would allow values to overwritten within a page as well, but that wasn't intended here. 

To be honest, this will not be needed unless one would want to overwrite the setting of `community` in the `.asciidoctorconfig` file. 

I could remove this from the PR, and it could be part of a future PR when there is a more obvious need for it. 

WDYT?

ahus1: I removed the changes to `antora.yml` in the PR.

Naros: Hi @ahus1, great, thanks for the explanation, I agree that leaving `antora.yml` as is, for now, makes the most sense given that it doesn't have any impact.  If we find that Bob wants to use this toggle we can think about adding it back.

Naros:
Hi @ahus1 so two quick questions, not directly related to this PR but more so to the IntelliJ plugin itself.  

1. Is there a way to control the rendered viewport's text width?  It appears to me that when the editor is first opened there is a default width of the right pane and the text gets rendered specifically based on that width.  If you then shrink the left editor pane and increase the size of the right rendered pane, the rendered document isn't reactive to the new pane size.
2. Is there a possibility in the future to provide multiple layouts for the rendered viewport?  Right now it always renders to the right, but I could also see it useful if the render viewport could be positioned to the bottom or top of the editor window.  That might be a limitation with IntelliJ but I just thought I'd ask.

ahus1:
@Naros - these are good questions. 

Good news: The IntelliJ plugin supports both. One is stylesheet thing that can be configured, the other one is a user's configuration.

The rendered viewport you see is the standard text width that Antora's default UI uses. They have made some UX choice made there to have a limited number of character, so this is fine. In a future PR, I will prepare configuration change to use the Debezium UI CSS in the preview. I can make that next on the list. See more [here in the docs of the plugin](https://intellij-asciidoc-plugin.ahus1.de/docs/users-guide/features/advanced/stylesheets.html). 

About splitting the preview: There is an option to split the preview horizontally, please have a look at the plugin's settings, see Editor and Preview Panel Layout (screenshot below). 

image

gunnarmorling:
@ahus1, there was a public holiday in the US yesterday, I'm sure @Naros will get back to this today.

ahus1:
No worries, take it one step at a time.

Naros:
Thanks for the valuable feedback @ahus1.  I had no idea those settings actually existed if I am to be honest, so I learned something super cool and new today, kudos!

Naros (approved):
LGTM, awesome work @ahus1.

Naros:

PR #3847, opened 2022-09-02
DBZ-5578 Add description to connector metadata

Naros:
https://issues.redhat.com/browse/DBZ-5578

I'm still waiting on details internally about additional changes that may be needed.  Once confirmed that we have all changes needed I'll convert from draft to ready.

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/metadata/MongoDbConnectorMetadata.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/metadata/MongoDbConnectorMetadata.java
@@ -16,7 +16,8 @@ public class MongoDbConnectorMetadata implements ConnectorMetadata {
 
     @Override
     public ConnectorDescriptor getConnectorDescriptor() {
-        return new ConnectorDescriptor("mongodb", "Debezium MongoDB Connector", MongoDbConnector.class.getName(), Module.version());
+        return new ConnectorDescriptor("mongodb", "Debezium MongoDB Connector", MongoDbConnector.class.getName(),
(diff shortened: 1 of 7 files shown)


rk3rn3r:
@Naros I suggest doing this with the "additional fields" pattern that we are already using for the config options. So having a more flexible way to add additional fields to the metadata block? Wdyt?

jpechane:
@rk3rn3r @Naros Do you mean builder pattern instead of constructor?
I'd say it depends on the scale/urgency? If this needs to be done fast then let's make it done as is now and just log a follow-up Jira. If not we can talk about an API for that.

Naros:
I'll happily change this if you want @rk3rn3r but I tend to agree with @jpechane just given the current workload and resource constraints.  wdyt?

rk3rn3r:
@Naros @jpechane I am sorry your reply got lost over my PTO back then.
What I meant was, when I remember correct what I had in mind back then lol, that instead of doing this here upstream (adding metadata that are only needed downstream for MCS) that we only do it here with additional properties that are statically rendered to the descriptors: https://github.com/bf2fc6cc711aee1a0c2a/cos-fleet-catalog-debezium/blob/main/src/main/java/io/debezium/mcs/envelope/ConnectorTypesEnvelope.java#L100
And add a way to add additional fields to the connector type descriptor (metadata) block.
I don't see a reason to add this to upstream (and later on require upstream releases or build to update downstream).

jpechane:
@rk3rn3r It sounds logical, just who will do the call and provide those additional metadata?

rk3rn3r:
Not needed anymore, see Jira issue for the details.

PR #3853, opened 2022-09-06
DBZ-5587 add alter add column test with COLLATE

ruanhang1993:
https://issues.redhat.com/browse/DBZ-5587
This PR add a test to reproduce an error when using COLLATE.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -160,6 +160,7 @@ Jiabao Sun
 Juan Fiallo
 Jun Zhao
 Hady Willi
+Hang Ruan
 Hans-Peter Grahsl
 Harvey Yue
 Henry Cai
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/listener/DefaultValueParserListener.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/listener/DefaultValueParserListener.java
@@ -44,7 +44,8 @@ public class DefaultValueParserListener extends MySqlParserBaseListener {
             if (ctx.constant().stringLiteral() != null) {
                 if (ctx.constant().stringLiteral().COLLATE() == null) {
                     columnEditor.defaultValueExpression(sign + unquote(ctx.constant().stringLiteral().getText()));
-                } else {
+                }
+                else {
                     columnEditor.defaultValueExpression(
                             sign + unquote(ctx.constant().stringLiteral().STRING_LITERAL(0).getText()));
                 }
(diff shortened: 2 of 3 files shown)


harveyyue:
@ruanhang1993 
`alter table t add column `collate_test` INT DEFAULT ' 29 ' COLLATE 'utf8_general_ci'; `
Anltr module will set `29 'COLLATE'utf8_general_ci` as defaultValueExpression in column object, it will cause the number format exception during apply the latest table schema, maybe need to check the MySqlParser.g4 to fix this issue.

ruanhang1993:
It seems that the stringLiteral of the default value contains the part of `COLLATE`.
```
stringLiteral
    : (
        STRING_CHARSET_NAME? STRING_LITERAL
        | START_NATIONAL_STRING_LITERAL
      ) STRING_LITERAL+
    | (
        STRING_CHARSET_NAME? STRING_LITERAL
        | START_NATIONAL_STRING_LITERAL
      ) (COLLATE collationName)?
    ;
```

harveyyue:
I have checked the `columnConstraint` which contained the rule `COLLATE collationName`, maybe it is not neccessary in `stringLiteral`.

columnConstraint
    : nullNotnull                                                   #nullColumnConstraint
    | DEFAULT defaultValue                                          #defaultColumnConstraint
    | (VISIBLE | INVISIBLE)                                         #invisibleColumnConstraint
    | (AUTO_INCREMENT | ON UPDATE currentTimestamp)                 #autoIncrementColumnConstraint
    | PRIMARY? KEY                                                  #primaryKeyColumnConstraint
    | UNIQUE KEY?                                                   #uniqueKeyColumnConstraint
    | COMMENT STRING_LITERAL                                        #commentColumnConstraint
    | COLUMN_FORMAT colformat=(FIXED | DYNAMIC | DEFAULT)           #formatColumnConstraint
    | STORAGE storageval=(DISK | MEMORY | DEFAULT)                  #storageColumnConstraint
    | referenceDefinition                                           #referenceColumnConstraint
    | COLLATE collationName                                         #collateColumnConstraint
    | (GENERATED ALWAYS)? AS '(' expression ')' (VIRTUAL | STORED | PERSISTENT)? #generatedColumnConstraint
    | SERIAL DEFAULT VALUE                                          #serialDefaultColumnConstraint
    | (CONSTRAINT name=uid?)?
      CHECK '(' expression ')'                                      #checkColumnConstraint
    ;

jpechane:
Unfortunately it is necessary - see https://dev.mysql.com/doc/refman/8.0/en/string-literals.html

So yes it makes sense as you can set collation separately per string constant and per column. The question is how MySQL itself parses this.

jpechane:
@ruanhang1993 Could you please reformat the source code before pushing?

jpechane:
For the record `COLLATE` in this case is used for setting the column collation. I think that for Debezium purposes thi is satisfactory solution a I wouln't complicate the grammar as it would  probably require two distinct string literal rules - one with COLLATION and one without.

jpechane:
@ruanhang1993 Applied, thanks!

ruanhang1993:
@jpechane Thanks for your help. There are some problems in my own development environment. So I do not reformat codes right. ;)

PR #3854, opened 2022-09-07
DBZ-5588 Field validation errors are misleading for positive, non-zero expectations

harveyyue:
https://issues.redhat.com/browse/DBZ-5588


jpechane:
@harveyyue Applied, thanks

PR #3855, opened 2022-09-07
DBZ-5591 Snapshot process setting source.ts_ms

ismailsimsek:
https://issues.redhat.com/browse/DBZ-5591
[zulip thread](https://debezium.zulipchat.com/#narrow/stream/302529-users/topic/source.2Ets_ms)

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
@@ -9,6 +9,7 @@ import java.sql.Clob;
 import java.sql.DatabaseMetaData;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.sql.Timestamp;
 import java.time.OffsetDateTime;
 import java.util.HashSet;
 import java.util.List;
@@ -227,6 +228,12 @@ public class OracleConnection extends JdbcConnection {
     @Override
     public List<String> readTableUniqueIndices(DatabaseMetaData metadata, TableId id) throws SQLException {
         return super.readTableUniqueIndices(metadata, id.toDoubleQuoted());
+    }
+
+    @Override
+    public Optional<Timestamp> getCurrentTimestamp() throws SQLException {
+        return queryAndMap("SELECT CURRENT_TIMESTAMP FROM DUAL",
+                rs -> rs.next() ? Optional.of(rs.getTimestamp(1)) : Optional.empty());
     }
 
     @Override
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
@@ -8,12 +8,15 @@ package io.debezium.connector.oracle;
 import java.sql.SQLException;
 import java.sql.Savepoint;
 import java.sql.Statement;
+import java.time.Instant;
+import java.time.OffsetDateTime;
 import java.util.Iterator;
 import java.util.List;
 import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.apache.kafka.connect.errors.ConnectException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -233,6 +236,21 @@ public class OracleSnapshotChangeEventSource extends RelationalSnapshotChangeEve
                 jdbcConnection.getTableMetadataDdl(table.id()),
                 table,
                 true);
+    }
+
+    @Override
+    protected Instant getSnapshotSourceTimestamp(RelationalSnapshotContext<OraclePartition, OracleOffsetContext> snapshotContext, TableId tableId) {
+        try {
+            Optional<OffsetDateTime> snapshotTs = jdbcConnection.getScnToTimestamp(snapshotContext.offset.getScn());
+            if (snapshotTs.isEmpty()) {
+                throw new ConnectException("Failed reading SCN timestamp from source database");
+            }
+
+            return snapshotTs.get().toInstant();
+        }
+        catch (SQLException e) {
+            throw new ConnectException("Failed reading SCN timestamp from source database", e);
+        }
     }
 
     /**
--- a/debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
+++ b/debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
@@ -22,6 +22,7 @@ import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
 import java.sql.Statement;
+import java.sql.Timestamp;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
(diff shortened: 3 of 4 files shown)


debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -13,34 +13,11 @@
...
-import java.sql.SQLException;
-import java.sql.Statement;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
+import java.sql.*;
```
Naros, line 16: Let's not use the package `.*` pattern, can you restore the original imports.

debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -96,6 +67,21 @@ public void onEntryChosenForEviction(PreparedStatement statement) {
                 }
             });
 
+    public Instant getCurrentTimestamp() throws SQLException {
```
Naros, line 70: For connectors like Oracle, if we're going to resolve this, it will need to be provided the offsets as connectors like Oracle will need to call a function based on the flashback SCN since the time across all table selects is identical for this connector.

Naros (commented):
A few comments inline.

debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
ismailsimsek: checking it...

debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
```diff
@@ -328,6 +329,23 @@ protected void tryStartingSnapshot(RelationalSnapshotContext<P, O> snapshotConte
...
+            // to get TS from the database. To avoid any mismatch between jvm time and DB time.
+            return jdbcConnection.getCurrentTimestamp();
+        }
+        catch (SQLException e) {
+            throw new ConnectException("Getting source timestamp failed", e);
+        }
+        // other options using JVM timestamp
+        // final Instant sourceTableSnapshotTimestamp = getClock().currentTimeAsInstant();
+        // final Instant sourceTableSnapshotTimestamp0 = Instant.ofEpochSecond(0);
+        // return sourceTableSnapshotTimestamp;
```
ismailsimsek, line 346: other options, simpler when using jvm based TS. commented out its just for displaying the alternatives

Naros: It is easier, but let's consider a hypothetical case.  

Let's say you need to snapshot a large data set, and it will take several hours to complete.  If you use the JVM approach, the issue is that the timestamp advances for each table, but the point in time for which the flashback query is based is static across all tables.  

I understand we have this problem even now, but if we're going to go through the exercise to change the behavior, I think it wise to do it so we are as accurate as possible.

I just don't see using the current clock time any better than what we are doing now.  I'd rather we use `0` or we go through the motions to resolve that value per connector based on whatever logic each connector requires.  For MySQL that may be `SELECT CURRENT_TIMESTAMP` but for Oracle that may be `SELECT SCN_TO_TIMESTAMP(?) FROM DUAL`, etc.

ismailsimsek: Tested changes with Mysql and Postgresql its possible to get TS reliably because snapshot process is done using setAutoCommit=false

- MySQL: somehow with auto commit OFF CURRENT_TIMESTAMP is still increasing. i believe only single table select is transactional.
- PostgreSQL: Full snapshot process is transactional. and CURRENT_TIMESTAMP returns static transaction TS for all tables. 
- Oracle:  Full snapshot process is transactional and getting TS from SCN makes it consistent.
- SQL server:  according [to documentation](https://docs.microsoft.com/en-us/sql/connect/jdbc/reference/setautocommit-method-sqlserverconnection?view=sql-server-ver16#remarks) it should be consistent similar behavior like Postgresql.
- Mongodb:  its seems already consistent no change necessary.

so only Mysql has special case for other databases CURRENT_TIMESTAMP is same as transaction timestamp.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
```diff
@@ -235,6 +238,21 @@ protected SchemaChangeEvent getCreateTableEvent(RelationalSnapshotContext<Oracle
                 true);
     }
 
+    @Override
+    protected Instant getSnapshotSourceTimestamp(RelationalSnapshotContext<OraclePartition, OracleOffsetContext> snapshotContext, TableId tableId) {
+        try {
+            Optional<OffsetDateTime> snapshotTs = jdbcConnection.getScnToTimestamp(snapshotContext.offset.getScn());
```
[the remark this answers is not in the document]
Naros: I believe that looks accurate to me.

Naros:
it looks like the tests maybe are unrelated but we need to dig into why they're happening, maybe some recent changes in `main`.  Could you do a rebase on `main` to pull in any pending changes that might have fixed these checks @ismailsimsek ?

ismailsimsek:
> it looks like the tests maybe are unrelated but we need to dig into why they're happening, maybe some recent changes in `main`. Could you do a rebase on `main` to pull in any pending changes that might have fixed these checks @ismailsimsek ?

seems like DB2 and UI errors are related to renamed class in https://github.com/debezium/debezium/pull/3804

once class not found fixed then it fails with :
org.apache.kafka.common.config.ConfigException: Invalid value io.debezium.storage.kafka.history.KafkaSchemaHistory for configuration schema.history.internal: Class io.debezium.storage.kafka.history.KafkaSchemaHistory could not be found.

ismailsimsek:
@Naros any chance to get this in for next release 1.9.6

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
```diff
@@ -230,6 +231,11 @@ public List<String> readTableUniqueIndices(DatabaseMetaData metadata, TableId id
         return super.readTableUniqueIndices(metadata, id.toDoubleQuoted());
     }
 
+    public Optional<Timestamp> getCurrentTimestamp() throws SQLException {
+        return queryAndMap("SELECT CURRENT_TIMESTAMP FROM DUAL",
+                rs -> rs.next() ? Optional.of(rs.getTimestamp(1)) : Optional.empty());
+    }
+
```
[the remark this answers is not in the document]
ismailsimsek: https://github.com/debezium/debezium-connector-db2/pull/74

Naros:
> any chance to get this in for next release 1.9.6

That should be reasonable, looking at this now.

Naros:
Reran SQL Server locally, tests passed.  Going to merge locally and retest SQL Server and Db2 and if those pass; I'll push this upstream @ismailsimsek.

Naros (approved):

Naros:

PR #3856, opened 2022-09-07
DBZ-5589 Mysql connector can't handle the case sensitive of rename/change column statement

harveyyue:
https://issues.redhat.com/browse/DBZ-5589


jpechane:
@harveyyue Applied, thanks

PR #3268, opened 2022-03-01
DBZ-4629 Retry mechanism on Kafka's ConnectException

spicy-sauce:

PR #3857, opened 2022-09-08
DBZ-4629 Debezium server crashes when no connection to the source db

ggaborg:
This is a re-make of this older pull request: https://github.com/debezium/debezium/pull/3268

I have one comment, related to the errors.max.retries property. Value of 0 means the mechanism is disabled and it is equivalent to the old behavior, which was swallowing the RetriableException in EmbeddedEngine.

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
@@ -62,6 +62,7 @@ import io.debezium.engine.StopEngineException;
 import io.debezium.engine.spi.OffsetCommitPolicy;
 import io.debezium.pipeline.ChangeEventSourceCoordinator;
 import io.debezium.util.Clock;
+import io.debezium.util.DelayStrategy;
 import io.debezium.util.VariableLatch;
 
 /**
@@ -190,6 +191,34 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
             .withDescription("Optional list of single message transformations applied on the messages. "
                     + "The transforms are defined using '<transform.prefix>.type' config option and configured using options '<transform.prefix>.<option>'");
 
+    private static final Field ERRORS_MAX_RETRIES = Field.create("errors.max.retries")
+            .withDisplayName("The maximum number of retries")
+            .withType(Type.INT)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.MEDIUM)
+            .withDefault(0)
+            .withValidation(Field::isInteger)
+            .withDescription("The maximum number of retries on connection errors before failing (-1 = no limit, 0 = disabled, > 0 = num of retries).");
+
+    private static final Field ERRORS_RETRY_DELAY_INITIAL_MS = Field.create("errors.retry.delay.initial.ms")
+            .withDisplayName("Initial delay for retries")
+            .withType(Type.INT)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.MEDIUM)
+            .withDefault(300)
+            .withValidation(Field::isPositiveInteger)
+            .withDescription("Initial delay (in ms) for retries when encountering connection errors."
+                    + " This value will be doubled upon every retry but won't exceed 'errors.retry.delay.max.ms'.");
+
+    private static final Field ERRORS_RETRY_DELAY_MAX_MS = Field.create("errors.retry.delay.max.ms")
+            .withDisplayName("Max delay between retries")
+            .withType(Type.INT)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.MEDIUM)
+            .withDefault(10000)
+            .withValidation(Field::isPositiveInteger)
+            .withDescription("Max delay (in ms) between retries when encountering connection errors.");
+
     /**
      * The array of fields that are required by each connectors.
      */
@@ -199,7 +228,8 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
      * The array of all exposed fields.
      */
     protected static final Field.Set ALL_FIELDS = CONNECTOR_FIELDS.with(OFFSET_STORAGE, OFFSET_STORAGE_FILE_FILENAME,
-            OFFSET_FLUSH_INTERVAL_MS, OFFSET_COMMIT_TIMEOUT_MS);
+            OFFSET_FLUSH_INTERVAL_MS, OFFSET_COMMIT_TIMEOUT_MS,
+            ERRORS_MAX_RETRIES, ERRORS_RETRY_DELAY_INITIAL_MS, ERRORS_RETRY_DELAY_MAX_MS);
 
     /**
      * How long we wait before forcefully stopping the connector thread when
@@ -808,10 +838,32 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
                                 break;
                             }
                             catch (RetriableException e) {
-                                LOGGER.info("Retrieable exception thrown, connector will be restarted", e);
-                                // Retriable exception should be ignored by the engine
-                                // and no change records delivered.
-                                // The retry is handled in io.debezium.connector.common.BaseSourceTask.poll()
+                                int maxRetries = getErrorsMaxRetries();
+                                LOGGER.info("Retriable exception thrown, connector will be restarted; errors.max.retries={}", maxRetries, e);
+                                if (maxRetries != 0) {
+                                    DelayStrategy delayStrategy = delayStrategy(config);
+                                    int totalRetries = 0;
+                                    boolean startedSuccessfully = false;
+                                    while (!startedSuccessfully) {
+                                        try {
+                                            totalRetries++;
+                                            LOGGER.info("Starting connector, attempt {}", totalRetries);
+                                            task.stop();
+                                            task.start(taskConfigs.get(0));
+                                            startedSuccessfully = true;
+                                        }
+                                        catch (Exception ex) {
+                                            if (totalRetries == maxRetries) {
+                                                LOGGER.error("Can't start the connector, max retries to connect exceeded; stopping connector...", ex);
+                                                throw ex;
+                                            }
+                                            else {
+                                                LOGGER.error("Can't start the connector, will retry later...", ex);
+                                            }
+                                        }
+                                        delayStrategy.sleepWhen(!startedSuccessfully);
+                                    }
+                                }
                             }
                             try {
                                 if (changeRecords != null && !changeRecords.isEmpty()) {
@@ -899,6 +951,11 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
                 completionCallback.handle(completionResult.success(), completionResult.message(), completionResult.error());
             }
         }
+    }
+
+    private int getErrorsMaxRetries() {
+        int maxRetries = config.getInteger(ERRORS_MAX_RETRIES);
+        return maxRetries;
     }
 
     /**
@@ -1100,6 +1157,11 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
 
     public void runWithTask(Consumer<SourceTask> consumer) {
         consumer.accept(task);
+    }
+
+    private DelayStrategy delayStrategy(Configuration config) {
+        return DelayStrategy.exponential(Duration.ofMillis(config.getInteger(ERRORS_RETRY_DELAY_INITIAL_MS)),
+                Duration.ofMillis(config.getInteger(ERRORS_RETRY_DELAY_MAX_MS)));
     }
 
     protected static class EmbeddedConfig extends WorkerConfig {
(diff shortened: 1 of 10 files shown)


ggaborg:
@jpechane @Naros @gunnarmorling Hi, could you give your input on this? Thanks!

jpechane (commented):
@ggaborg Thanks for the PR, I've left couple of comments.
But I am now really not sure if we are doing the right thing.
Recently we've merged the by-default retry PR into Debezium.
So I believe that that use case when connection to the source database is lost is covered.
Could you please give it a test ride? Also I double checked the `org.apache.kafka.connect.runtime.WorkerSourceTask.poll()` and it conforms to the current Debezium Engine behaviour.
Also this does not solve the sender part as it is the one in Debezium Server, not engine. WDYT?

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
```diff
@@ -813,10 +843,30 @@ public Map<String, String> configs() {
                                 break;
                             }
                             catch (RetriableException e) {
-                                LOGGER.info("Retrieable exception thrown, connector will be restarted", e);
-                                // Retriable exception should be ignored by the engine
-                                // and no change records delivered.
-                                // The retry is handled in io.debezium.connector.common.BaseSourceTask.poll()
+                                int maxRetries = config.getInteger(ERRORS_MAX_RETRIES);
+                                LOGGER.info("Retriable exception thrown, connector will be restarted; errors.max.retries={}", maxRetries, e);
```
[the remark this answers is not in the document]
ggaborg: I've let the same behavior as today

ggaborg:
@jpechane, I have checked the behavior prior to this implementation, including the "by-default retry PR" and when I stop the database the DebeziumServer stops. Is there any documentation or configuration I can look into? Maybe I have missed something. Also, like I said, I left default configuration with 0 retries which is basically disabling the mechanism and preserves the old behavior.

jpechane:
@ggaborg Have you use the snapshot version? It as not release yet. I am sorry for not telling that.

ggaborg:
I have used the up-to-date master branch

jpechane:
@ggaborg It should be https://issues.redhat.com/browse/DBZ-5244

ggaborg:
Yes, I have it in the master branch. And it should work be default right, no configuration needed? Stopping the database after successful startup should not terminate DebeziumServer, but it does. With this PR it doesn't.

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
```diff
@@ -813,10 +843,30 @@ public Map<String, String> configs() {
...
+                                    while (!startedSuccessfully) {
+                                        try {
+                                            totalRetries++;
+                                            LOGGER.trace("Starting connector, attempt {}", totalRetries);
+                                            task.stop();
+                                            task.start(taskConfigs.get(0));
+                                            startedSuccessfully = true;
+                                        }
+                                        catch (Exception ex) {
+                                            LOGGER.error("Can't start the connector:", ex);
```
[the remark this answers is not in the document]
ggaborg: the "max retries exceeded" log is in the lines below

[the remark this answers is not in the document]
ggaborg: You meant something like this:
                                            if (totalRetries == maxRetries) {
                                                LOGGER.error("Can't start the connector, max retries to connect exceeded; stopping connector...", ex);
                                                throw ex;
                                            } else {
                                                LOGGER.error("Can't start the connector, will retry later...", ex);
                                            }

ggaborg:
@jpechane @spicy-sauce Hi, I have retested Debezium Server again with latest master with both sqlserver and mysql connectors and it looks like the connectors are restarted if database is stopped and started again after a while. The recording of changes is also working as expected. Should we close this PR then and the jira issue https://issues.redhat.com/browse/DBZ-4629?

spicy-sauce:
@ggaborg so this now works as result of changing the behavior so by default a RetriableException will be thrown by the connectors? Does it mean that Debezium Server catches this exception and keeps trying to connect it based on the backoff mechanism? So no further changes are needed? What are the properties used for this backoff?

ggaborg:
@spicy-sauce I tested without playing with any properties, don't know which properties can be configured. Asked about them in a comment above. It seems that no change is required anymore, the current backoff mechanism seems to be working and keeps restarting the connector.

spicy-sauce:
OK, so this is great, just need to see how can it be configured (if at all - if not, we may want to support this), so we won't try to get a connection too instantly and bloat to log as result.

ggaborg:
@jpechane @spicy-sauce Hi, tested also with the postgres connector by stopping the db and it only works if the db restarts before the first retry; if the db is still down on first retry, dbz stops. 
With sqlserver it keeps retrying indefinitely and reconnects when db is restarted.
With mysql it looks like the mechanism does not kick in (no such messages "Going to restart connector after 10 sec. after a retriable exception") but there is another mechanism of the binary log client keepalive "INFO  [com.git.shy.mys.bin.BinaryLogClient] (blc-keepalive-localhost:3306) Keepalive: Trying to restore lost connection to localhost:3306"

jpechane:
@ggaborg @spicy-sauce I am sorry guys we had a busy week. Hopefully I'll get to this one tomorrow

spicy-sauce:
@jpechane, it seems that the current behavior is not consistent for the assortment of connectors as tested by @ggaborg.  I think this is not the desired behavior, the same backoff mechanism should be used regardless the connector in Debezium Server. WDYT?

jpechane:
@spicy-sauce You are right, sorry. I completely forgot about that :-(. It makes sense to get this in then. @ggaborg Could you please rebase the PR against the latest main? Thanks

ggaborg:
@jpechane 
Hi. I would like to check with you if my understanding of what the by-default retry mechanism is: when exception is thrown by the producer the poll() method of the BaseSourceTask is catching it. if it is a retriable it will stop the task and rethrow the error. Now the EmbeddedEngine which is doing the poll() is catching the exception and if it is retriable it ignores it and continues to poll(). In the next poll(), the task will startIfNeededAndPossible. But now it depends on each connector what it does in its start() method. My guess is that some of them are only creating a connection but they are not actually using it until doPoll() so the task is able to start, but others like postgres, are trying to use the connection and if they fail (e.g. when the db is stopped) they throw exception which stops the engine. 

Now, what we implemented in this PR is that we catch the retriable from producer and keep retrying to start the task no matter what the start exception is.

I'm thinking that in fact we don't need this PR but instead we could check for retriable exceptions also on startIfNeededAndPossible and use something similar to the ErrorHandler which solves retriable exceptions.

WDYT?

Bellow exception from postgres on first retry to start task which ends the engine:

11:11:49.183 [pool-7-thread-1] INFO io.debezium.server.ConnectorLifecycle - Connector completed: success = 'false', message = 'Error while trying to run connector class 'io.debezium.connector.postgresql.PostgresConnector'', error = '{}'
io.debezium.DebeziumException: Couldn't obtain encoding for database postgres
        at io.debezium.connector.postgresql.connection.PostgresConnection.getDatabaseCharset(PostgresConnection.java:488)
        at io.debezium.connector.postgresql.PostgresConnectorTask.start(PostgresConnectorTask.java:75)
        at io.debezium.connector.common.BaseSourceTask.start(BaseSourceTask.java:135)
        at io.debezium.connector.common.BaseSourceTask.startIfNeededAndPossible(BaseSourceTask.java:221)
        at io.debezium.connector.common.BaseSourceTask.poll(BaseSourceTask.java:153)
        at io.debezium.embedded.EmbeddedEngine.run(EmbeddedEngine.java:790)
        at io.debezium.embedded.ConvertingEngineBuilder$2.run(ConvertingEngineBuilder.java:192)
        at io.debezium.server.DebeziumServer.lambda$start$1(DebeziumServer.java:151)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at java.base/java.lang.Thread.run(Thread.java:833)
Caused by: org.postgresql.util.PSQLException: Connection to localhost:5432 refused. Check that the hostname and port are correct and that the postmaster is accepting TCP/IP connections.
        at org.postgresql.core.v3.ConnectionFactoryImpl.openConnectionImpl(ConnectionFactoryImpl.java:319)
        at org.postgresql.core.ConnectionFactory.openConnection(ConnectionFactory.java:49)
        at org.postgresql.jdbc.PgConnection.<init>(PgConnection.java:247)
        at org.postgresql.Driver.makeConnection(Driver.java:434)
        at org.postgresql.Driver.connect(Driver.java:291)
        at io.debezium.jdbc.JdbcConnection.lambda$patternBasedFactory$1(JdbcConnection.java:243)
        at io.debezium.jdbc.JdbcConnection.connection(JdbcConnection.java:882)
        at io.debezium.jdbc.JdbcConnection.connection(JdbcConnection.java:877)
        at io.debezium.connector.postgresql.connection.PostgresConnection.getDatabaseCharset(PostgresConnection.java:485)
        ... 10 common frames omitted
Caused by: java.net.ConnectException: Connection refused: no further information
        at java.base/sun.nio.ch.Net.pollConnect(Native Method)
        at java.base/sun.nio.ch.Net.pollConnectNow(Net.java:672)
        at java.base/sun.nio.ch.NioSocketImpl.timedFinishConnect(NioSocketImpl.java:542)
        at java.base/sun.nio.ch.NioSocketImpl.connect(NioSocketImpl.java:597)
        at java.base/java.net.SocksSocketImpl.connect(SocksSocketImpl.java:327)
        at java.base/java.net.Socket.connect(Socket.java:633)
        at org.postgresql.core.PGStream.createSocket(PGStream.java:241)
        at org.postgresql.core.PGStream.<init>(PGStream.java:98)
        at org.postgresql.core.v3.ConnectionFactoryImpl.tryConnect(ConnectionFactoryImpl.java:109)
        at org.postgresql.core.v3.ConnectionFactoryImpl.openConnectionImpl(ConnectionFactoryImpl.java:235)
        ... 18 common frames omitted

ggaborg:
@jpechane On the other hand, if we merge this PR, can we put the default to -1 for errors.max.retries? Currently it is 0 which disables the mechanism. See bellow the adoc:
|`errors.max.retries`
|`0`
|The maximum number of retries on connection errors before failing (-1 = no limit, 0 = disabled, > 0 = num of retries).

alwaysbemark:
Hey all, will this make its way into CR1? Thank you!

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
```diff
@@ -808,10 +838,32 @@ public Map<String, String> configs() {
                                 break;
                             }
                             catch (RetriableException e) {
-                                LOGGER.info("Retrieable exception thrown, connector will be restarted", e);
-                                // Retriable exception should be ignored by the engine
-                                // and no change records delivered.
-                                // The retry is handled in io.debezium.connector.common.BaseSourceTask.poll()
+                                int maxRetries = getErrorsMaxRetries();
+                                LOGGER.info("Retriable exception thrown, connector will be restarted; errors.max.retries={}", maxRetries, e);
+                                if (maxRetries != 0) {
```
poonam-meghnani, line 843: @jpechane In case the maxRetries is 0, it does not go thorough the retry logic, however it does not break the while loop either and keeps falling back as RetriableException. I believe we need a else part to this condition:

```
else {
    handlerError = e;
    break;
}
```

poonam-meghnani: Even when maxRetries is non-zero, say `errors.max.retries=1`, it keeps looping over and resets totalRetries to 0 each time, resulting in retrying forever.

ggaborg: @poonam-meghnani Hi. This mechanism is complementary to the one inside the poll() method of the BaseSourceTask. In the poll() method, the first line calls startIfNeededAndPossible() which calls start() method which for some connectors can throw non retriable exceptions or wrapped retriable exceptions. In that case the engine was stopping. Now with the new mechanism, the connector will try to restart no matter what the exception in the start() method is. 
Now back to your questions, when max retries is 0, the new mechanism is disabled and the behavior is exactly as it was before. You could check the comments above to understand the different behavior of different connectors without this new mechanism.

alwaysbemark: The behaviour that we're observing is that even with max retries 0, the connector still retries in an infinite loop. Is this an expected behaviour? It could be that it was broken in this way previously?

ggaborg: @alwaysbemark like I said, 0 max retries disables only this new mechanism, but not the one implemented by this issue: https://issues.redhat.com/browse/DBZ-5244. Currently both of the mechanisms are in place. Depending on the connector, DBZ-5244 is able to retry by itself to restart the task or not. Where this is not enough, the new mechanism does the job.

PR #3860, opened 2022-09-08
DBZ-5581 Apply unavailable value placeholder on when parsing SELECT_LOB_LOCATOR event types

Naros:
https://issues.redhat.com/browse/DBZ-5581


jpechane:
@Naros One question, what will happen if you set the value of the LOB column to `null`?

Naros:
Hi @jpechane I updated the test cases to include a NULL-assignment use case.  As expected and to answer your question, when the column is explicitly assigned NULL, it will be passed in the event explicitly as a nulled value rather than the placeholder.

jpechane:
@Naros Applied, thanks!

PR #3861, opened 2022-09-08
DBZ-5593 Provide ability to define SQL Server connection string without default port included

kyleyj:
When using the sqlserver connector, the "database.port" property is a required field with the default value of 1433.  This is problematic when you have a named instance and the port is not configured with the default port.  There needs to be a way via the properties to indicate that the "database.port" property is not provided and to not default to the default port (1433).

These change define a new property ("database.noport") that will create a ConnectionFactory with a URL Pattern that excludes the port configuration.  The config properties as passed to the method to create the ConnectionFactory to determine which type of ConnectionFactory to create: with port in the url pattern or without the port in the url pattern.

### DIFF
--- a/debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SqlServerConnectorConfigWithNoportIT.java
+++ b/debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SqlServerConnectorConfigWithNoportIT.java
@@ -20,7 +20,7 @@ import io.debezium.embedded.AbstractConnectorTest;
 /**
  * Integration test for the Debezium SQL Server connector.
  *
- * @author Jiri Pechanec
+ * @author Kyley Jex
  */
 public class SqlServerConnectorConfigWithNoportIT extends AbstractConnectorTest {
 


kyleyj:
What is the process to get this PR reviewed & completed?

kyleyj:
> @kyleyj Hi, tanks for the PR, I'll review it tomorrow. Could you please push the commits using your email associated with your GH account so it is properly attributed to you? Thanks

I have updated the commits to have the correct author information.  Please review and let me know if there is anything else that needs to be corrected.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
```diff
@@ -221,6 +221,15 @@ public static SnapshotIsolationMode parse(String value, String defaultValue) {
     public static final Field PORT = RelationalDatabaseConnectorConfig.PORT
             .withDefault(DEFAULT_PORT);
 
+    public static final Field NOPORT = Field.create(DATABASE_CONFIG_PREFIX + SqlServerConnection.NOPORT_NAME)
```
jpechane, line 224: @kyleyj I am not happy with this option. I'd prefer to modify `PORT` locally to be optional instead of required. In that case it would be just enough to check if it is present or not and there will be no need for new config option.

jpechane:
@kyleyj Thanks for the PR - i've left one comment related to the implementation. Unfortunately the Github UI still shows the commits not linked to your account. Are you sure you've used the proper configuration?

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
kyleyj: I would have preferred that option too, but that would have introduced a breaking change for anyone that was expecting the default port (1433) to be used when not provided.  In fact, the implementation of defaulting to default port when not provided doesn't following the specification for JDBC.  So, I would prefer to modify `PORT` to be optional but others would need to weigh in on how impactful that change would be.

jpechane: @kyleyj That's very good point. But how about using a magical umber like `0`? If set to `0` then use named instance approach?

kyleyj: Using a magic number (0, -1, etc) would work, but still requires knowledge of the intended behavior like the `noport` property.  How have breaking changes been communicated in the past?  I've noticed that with v19/v20, there was a change with some of the other properties (i.e. database.server.name, database.names, etc.).  Would we just change/communicate the change of defaulting to 1433 when `PORT` is not provided?

jpechane: @kyleyj We are swiching a major version so brekaing changes are allowed. Still I believe that your original argument has a merit. Let's stay with `1433` as default for two reasons

- it is more frequent scenario
- it is consistent with behaviour of other connectors

From user PoV knowledge there is no difference between using new option or magic number. In both cases the docs need to be read. Yet I'd personally prefer to use magical number as the new option is introduced to disable another option which is a bit strange.

Thinking about that there is even one more solution - a generic one that we provide for Oracle - https://debezium.io/documentation/reference/1.9/connectors/oracle.html#oracle-property-database-url

jpechane: @kyleyj And the thing I completely forget about originally - please see https://issues.redhat.com/browse/DBZ-4001

We can provide OOTB alternative connection factory that would just not use the port setting.

kyleyj: That looks like a great option and is probably the right approach.  I'll be testing this out tomorrow.  However, I'm not able to see any documentation that would have led me to this option in the first place.  Other than looking at the example from the ticket, where is the documentation to support this feature?

jpechane: Good point, we probably forgot to add it to the docs. The in-code description is loated in https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/jdbc/JdbcConfiguration.java#L64

PR #3862, opened 2022-09-08
DBZ-5043 Move TOPIC_PREFIX into CommonConnectorConfig

vjuranek:
`TOPIC_PREFIX` is now mandatory to all connectors therefore it make sense to have it in common config. Beside that, it also makes it more easy to use it in Debezium UI without any workarounds - if the field is not member of the given connector config, the field has to be explicitely added into known fields otherwise is invisible for UI.


jpechane:
@vjuranek Applied, thanks

PR #3867, opened 2022-09-12
DBZ-5601 Switch Debezium test containers to nightly

vjuranek:
https://issues.redhat.com/browse/DBZ-5601


jpechane:
@vjuranek Applied, thanks

PR #3870, opened 2022-09-13
DBZ-5606 Include storage modules into the build for DB2 workflow

vjuranek:
https://issues.redhat.com/browse/DBZ-5606


jpechane:
@vjuranek Please be careful. What will happen for `1.9` branch?

vjuranek:
> What will happen for `1.9` branch?

@jpechane there are already some differences between `main` and `1.9` - e.g. `main` uses `./mvnw` which is not present in `1.9`. Based on this I assume that `1.9` uses workflow from `1.9` branch and we are safe to do any changes in `main` without affecting `1.9`.

jpechane:
@vjuranek Applied, thanks

PR #3871, opened 2022-09-13
DBZ-5043 Make topic prefix field manadatory and with proper group

vjuranek:


ani-sha (commented):
Hi @vjuranek This looks good! I think this is something that was blocking the changes to be reflected in the UI. I have one more point to add if you look [here](https://github.com/debezium/debezium-ui/blob/7a38067880231595be5a6fd2ccbfbd10b578988c/backend/src/main/java/io/debezium/configserver/service/mongodb/MongoDbConnectorIntegrator.java#L159) all the properties are taken from `CONFIG_DEFINITION`. So you also need to add `TOPIC_PREFIX` [here](https://github.com/debezium/debezium/blob/a0daef37e2584c36d14c057103a7922af4e827b4/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java#L491) as well for all connectors.

vjuranek:
Thanks @ani-sha for the hints! Looking on it.

PR #3872, opened 2022-09-14
DBZ-5605 Support DEFAULT ON NULL in COLUMN_DEFINITION clause

ani-sha:
https://issues.redhat.com/browse/DBZ-5605


ani-sha:
Hi @Naros, One thing that is bothering me here is that the checks pass properly in CI and locally. But when I see it in the IDEA ANTLR tool preview this DDL shows an error for the functions inside the `CHECK` clause.
```
CONSTRAINT "FOFF_MESU_CK2" CHECK ( coalesce( date_end_reflection_period, date_purchase ) >= date_purchase ) ENABLE
CONSTRAINT "FOFF_MESU_CK1" CHECK ( date_end >= date_start ) ENABLE
CONSTRAINT "FOFF_MESU_CK3" CHECK ( pay_transaction_success_flg in ( 'Y', 'N' ) ) ENABLE, SUPPLEMENTAL LOG DATA (ALL) COLUMNS
```
That's strange, right? :thinking:

Naros:
So two things, contrary to my recollection, we actually do not push the DDL through the DDL parser during the snapshot phase since we source the relational model from the JDBC result-set and therefore this explains why we only see the DDL failure during the stream phase and not also during the snapshot phase.  Just good to know going forward.

To your point about the ANTLR preview, I've found the preview doesn't work with the out of box lexing because it leaves the case as is.  You can adjust this behavior by going to "Settings => Language & Frameworks => ANTLR" and changing the option called "Case transformation in the Preview window" to use "Transform to uppercase when lexing" and that solves part of the issue with the preview.

Naros (commented):
LGTM, thanks @ani-sha

Naros:

PR #3874, opened 2022-09-14
DBZ-5496: Improve compatibility with Azure SQL

morozov:
If the SQL Server connector is configured to capture changes from a single database, pass it to the JDBC connection string in order to bypass the limitation of Azure SQL which does not allow switching between databases within a single connection.

See the [documentation](https://docs.microsoft.com/en-us/sql/t-sql/language-elements/use-transact-sql?view=sql-server-ver15#arguments) for more details.

All the tests that use the default configuration (the majority of them) will now use the single-database connection, so no additional tests should be necessary.

UPD: we can actually test that if the connection is passed a single database, it is the current database. Otherwise, `master` should be the current database (which will hold true unless overridden for the Debezium login, i.e. reliable enough).


jpechane:
@morozov Applied, thanks

PR #3876, opened 2022-09-15
DBZ-5612 Remove records from being logged at all levels

rajdangwal:
### Problem
Looks like the connector logs records at certain points which may contain sensitive user information. 

### Solution
This PR aims to prevent records from being logged at all logging levels.

### Testing
No tests required.

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/signal/KafkaSignalThread.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/signal/KafkaSignalThread.java
@@ -136,7 +136,7 @@ public class KafkaSignalThread<T extends DataCollectionId> {
                     return;
                 }
                 catch (final Exception e) {
-                    Throwables.logErrorAndTraceRecord(LOGGER, "Skipped signal due to an error", e, record);
+                    Throwables.logErrorAndTraceRecord(LOGGER, record, "Skipped signal due to an error", e);
                 }
             }
         }
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/signal/KafkaSignalThread.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/signal/KafkaSignalThread.java
@@ -135,7 +135,7 @@ public class KafkaSignalThread<T extends DataCollectionId> {
                     return;
                 }
                 catch (final Exception e) {
-                    LOGGER.error("Skipped signal due to an error '{}'", record, e);
+                    LOGGER.error("Skipped signal due to an error", e);
                 }
             }
         }
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/signal/KafkaSignalThread.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/signal/KafkaSignalThread.java
@@ -37,6 +37,7 @@ import io.debezium.pipeline.signal.StopSnapshot;
 import io.debezium.spi.schema.DataCollectionId;
 import io.debezium.util.Collect;
 import io.debezium.util.Threads;
+import io.debezium.util.Throwables;
 
 /**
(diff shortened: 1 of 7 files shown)


jpechane:
@rajdangwal Hi, thanks for the PR. I am really not convinced about this change. As you noted all the logs were done at `TRACE` level that is almost never enabled only in case of troubleshooting.
But when we do the troubleshooting it is frequently important to see the message to know what is wrong.

From where comes this requirement? Why is the `TRACE` level limitation not enough?

jpechane (changes requested):
@rajdangwal Could you please take a look at my comments?

debezium-core/src/main/java/io/debezium/connector/base/ChangeEventQueue.java
```diff
@@ -206,7 +206,7 @@ public void disableBuffering() {
 
     protected void doEnqueue(T record) throws InterruptedException {
         if (LOGGER.isDebugEnabled()) {
-            LOGGER.debug("Enqueuing source record '{}'", record);
+            LOGGER.debug("Enqueuing source record");
```
jpechane, line 209: Let's keep the message but switch to `trace`.

debezium-core/src/test/java/io/debezium/kafka/KafkaCluster.java
```diff
@@ -780,7 +780,7 @@ public <K, V> void produce(String producerName, int messageCount,
                         ProducerRecord<K, V> record = messageSupplier.get();
                         producer.send(record);
                         producer.flush();
-                        LOGGER.debug("Producer {}: sent message {}", producerName, record);
+                        LOGGER.debug("Producer {}: sent message", producerName);
```
jpechane, line 783: This is code for test, I think it can be kept.

debezium-core/src/test/java/io/debezium/kafka/KafkaCluster.java
```diff
@@ -926,7 +926,7 @@ public <K, V> void consume(String groupId, String clientId, OffsetResetStrategy
                     consumer.subscribe(new ArrayList<>(topics));
                     while (continuation.getAsBoolean()) {
                         consumer.poll(10).forEach(record -> {
-                            LOGGER.debug("Consumer {}: consuming message {}", clientId, record);
+                            LOGGER.debug("Consumer {}: consuming message", clientId);
```
jpechane, line 929: Again, this is test code.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/signal/KafkaSignalThread.java
```diff
@@ -135,7 +135,7 @@ private void monitorSignals() {
                     return;
                 }
                 catch (final Exception e) {
-                    LOGGER.error("Skipped signal due to an error '{}'", record, e);
+                    LOGGER.error("Skipped signal due to an error", e);
```
jpechane, line 138: For handling errors (here and below) - please add a new method to {{io.debezium.util.Throwables}}. The method will accept mutliple arguments

- logger
- error message
- record/data

The metod will log the error message at error level using passed logger and the there will be a {{TRACE}} level message like `Source of error is '{}'` using a new logger defined in `Throwables` class.

jpechane (changes requested):
@rajdangwal Thanks for the explanation! I understand that and I agree that the remaining cases might be an issue. Still I think it might make sense to log them under certain conditions. Please take a look at the comments.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/signal/KafkaSignalThread.java
rajdangwal: Thanks for the suggestion. Makes sense.

jpechane:
@rajdangwal Applied, thanks!

rajdangwal:
@jpechane Thanks for applying the changes and merging.

PR #3878, opened 2022-09-15
Dbz 3327 populate empty default value cells in connector properties tables

roldanbob:
[DBZ-3327](https://issues.redhat.com/browse/DBZ-3327) 

Low impact change to certain connector configuration properties tables that inserts the string `No default` when a cell in the **Default**  column of an entry is empty.


Naros:
Hi @roldanbob, can you rebase this against the latest `main`, thanks.

Naros (approved):
LGTM, once it's rebased we can merge this.  I assume this is safe to keep also for Debezium 2.0, no need to push these changes back to Debezium 1.9, correct @roldanbob ?

roldanbob:
> LGTM, once it's rebased we can merge this. I assume this is safe to keep also for Debezium 2.0, no need to push these changes back to Debezium 1.9, correct @roldanbob ?

Should be all set now, @Naros. And I agree: no need to push these changes to 1.9.

jpechane:
@roldanbob Applied, thanks

PR #3881, opened 2022-09-16
DBZ-5585 Modify the Instantiator to not require classloader

harveyyue:
[DBZ-5585 ](https://issues.redhat.com/browse/DBZ-5585)


jpechane:
@harveyyue Applied, thanks

PR #3883, opened 2022-09-16
DBZ-5620 Upgrade mysql-binlog-connector-java library version

jpechane:
https://issues.redhat.com/browse/DBZ-5620


tooptoop4 (changes requested):
0.27.2

PR #3886, opened 2022-09-16
[docs] Fix `max.queue.size` default value in postgres connector documentation

gmouss:
[docs] Fixed the `max.queue.size` default value in the docs for the postgres connector.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -448,3 +448,4 @@ Inki Hwang
 崔世杰
 合龙 张
 Phạm Ngọc Thắng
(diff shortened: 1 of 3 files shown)


gmouss:
Hello, it states in the CONTRIBUTE.md file that:

>  As an exception, commits for trivial documentation changes which don't warrant the creation of a Jira issue can be prefixed with [docs], for example:
> [docs] Typo fix in Postgres connector documentation
but the workflow is failing. What can I do to unblock it ?

jpechane:
Merging via https://github.com/debezium/debezium/pull/3892

PR #3887, opened 2022-09-16
DBZ-4213 Use UTF-8 encoding when creating new Strings

jeremy-l-ford:
Running this test with the system property set -Dfile.encoding=ASCII will still succeed.
Running this test with the system property set -Dfile.encoding=ASCII without the fix you can see it fail.


jpechane:

PR #3889, opened 2022-09-17
DBZ-5618 fix broken anchors in the docs

ahus1:
This approach updated the `xref:[]`s with the best matching anchor I found. Alternatively, the anchor in the target could have been changed, and I can update this PR if that is preferred. 

After the change, the IntelliJ AsciiDoc plugin doesn't report any more broken anchors.

To test his, I used version 0.37.49 of the IntelliJ AsciiDoc plugin.


jpechane:
@ahus1 Applied, thanks!

PR #3894, opened 2022-09-20
DBZ-5628 Added support for Mongo pre-image in change stream

xinbinhuang:
https://issues.redhat.com/browse/DBZ-5628

- [x] Haven't added any reasonable testing yet. Any pointers on where to add test cases?
- [ ] add documentation

### DIFF
--- a/.github/workflows/debezium-workflow.yml
+++ b/.github/workflows/debezium-workflow.yml
@@ -272,7 +272,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        version-mongo-server: ["4.0", "4.4", "5.0"]
+        version-mongo-server: ["4.0", "4.4", "5.0", "6.0"]
     steps:
       - name: Checkout Action
         uses: actions/checkout@v2
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -416,6 +416,7 @@ Willie Zhu
 Wout Scheepers
 Xiao Fu
 Xiaopu Zhu
+Xinbin Huang
 Xinquan She
 Xuan Shen
 Yang Wu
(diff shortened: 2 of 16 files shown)


jpechane:
@jcechace This is one for you to keep eye on
@xinbinhuang Good job! I'd recommend to add a completely new class that would be responsible for testing.
The test itself will be executed only when run against MongoDB 6.0. For that it is possible to use `SkipWhenDatabaseVersion` and `DatabaseVersionResolver`. See `MySqlDatabaseVersionResolver` for one of the existing implementations.
It is also important to add MongoDB 6.0 to CI workflow defined in `.github/workflows/debezium-workflow.yml`

xinbinhuang:
> @jcechace This is one for you to keep eye on @xinbinhuang Good job! I'd recommend to add a completely new class that would be responsible for testing. The test itself will be executed only when run against MongoDB 6.0. For that it is possible to use `SkipWhenDatabaseVersion` and `DatabaseVersionResolver`. See `MySqlDatabaseVersionResolver` for one of the existing implementations. It is also important to add MongoDB 6.0 to CI workflow defined in `.github/workflows/debezium-workflow.yml`

thx for the info! I'll take a look and try to implement it in the next a few days.

jpechane:
@xinbinhuang I've merged (and pushed) the 6.0 container image so it should be testable now. Also I've push to your branch a commit with CI job configured to run against 6.0 too so please make sure to pull it to your local branch.

xinbinhuang:
@jpechane I just added the test and I think the PR is ready for review! PTAL. thank you!

jpechane:
@xinbinhuang Applied, thanks! Could you please send a separate PR with documentation? Thanks a lot

xinbinhuang:
> @xinbinhuang Applied, thanks! Could you please send a separate PR with documentation? Thanks a lot

Awesome! I'll follow up with the documentation in a separate PR.

PR #3897, opened 2022-09-20
DBZ-5630: Unquote double quotes from default values on MySQL

morozov:
Besides single quotes for string literals, MySQL allows double quotes.


jpechane:
@morozov Applied, thanks!

PR #3900, opened 2022-09-21
SDBZ-4417 Add information about customizing basic SMT configuration

roldanbob:
[DBZ-4417](https://issues.redhat.com/browse/DBZ-4417)

Adds paragraph to emphasize that the basic configuration generally requires customizaiton through the use of  SMT predicates.


jpechane:
@roldanbob Applied, thanks!

PR #3905, opened 2022-09-21
DBZ-5635 Reduce undo trx log verbosity when LOB enabled

Naros:
https://issues.redhat.com/browse/DBZ-5635


jpechane:
@Naros Applied, thanks

PR #3906, opened 2022-09-21
DBZ-5636 support for using any expression in kill statements

nirolevy:
https://issues.redhat.com/browse/DBZ-5636

Upstream pr:
https://github.com/antlr/grammars-v4/pull/2836


jpechane:
@nirolevy Applied, thanks!

PR #3907, opened 2022-09-21
DBZ-4027 implement external signaling source, FileSignalChannel

ismailsimsek:
cc @jpechane  @Naros is this looks any good?

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/signal/channels/FileSignalChannel.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/signal/channels/FileSignalChannel.java
@@ -1,0 +1,132 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.pipeline.signal.channels;
+
+import java.io.File;
+import java.io.FileWriter;
+import java.nio.file.Files;
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.kafka.common.config.ConfigDef;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+import io.debezium.DebeziumException;
+import io.debezium.config.CommonConnectorConfig;
+import io.debezium.config.Configuration;
+import io.debezium.config.Field;
+import io.debezium.pipeline.signal.SignalRecord;
+
+/**
+ * The class responsible for processing of signals delivered to Debezium via a file.
+ * The signal message must have the following structure, formatted as json line:
+ * <ul>
+ * <li>{@code id STRING} - the unique identifier of the signal sent, usually UUID, can be used for deduplication</li>
+ * <li>{@code type STRING} - the unique logical name of the code executing the signal</li>
+ * <li>{@code data STRING} - the data in JSON format that are passed to the signal code
+ * </ul>
+ *
+ * @author Ismail Simsek
+ *
+ */
+public class FileSignalChannel implements SignalChannelReader {
+    public static final String CONFIGURATION_FIELD_PREFIX_STRING = "signal.";
+    public static final Field SIGNAL_FILE = Field.create(CONFIGURATION_FIELD_PREFIX_STRING + "file")
+            .withDisplayName("Signal file name")
+            .withType(ConfigDef.Type.STRING)
+            .withWidth(ConfigDef.Width.LONG)
+            .withImportance(ConfigDef.Importance.HIGH)
+            .withDescription("The name of the file for the signals to the connector")
+            .withValidation(Field::isRequired);
+    public static final String CHANNEL_NAME = "file";
+    private static final Logger LOGGER = LoggerFactory.getLogger(FileSignalChannel.class);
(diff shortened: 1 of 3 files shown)


jpechane:
@ismailsimsek Hi, thanks for the PR. I'll take a look into it a bit later but right now I am not convinced I agree with some refactorings.

Generally there is a different problem - I belive we should not have all the channels running. The signalling source should be configurable as connector option and based on that database, or read-only database or file or memory will be used.

ismailsimsek:
@jpechane 

> I believe we should not have all the channels running. The signalling source should be configurable as connector option and based on that database, or Kafka or file or memory will be used.

that part was draft.. dynamic loading with config is implemented now. its ready for another look. 

> IMHO we need to decouple signalling channel from incrementalsnapshot bookeeping.

not sure which part you mean here?

its still in progress for me and i might be slow to progress. anyone interested please feel free to commit/work on it.

jpechane:
@ismailsimsek Even if non-database channel is used for signalling then it still necessary fro incremental snapshot to work to use database as the place where watermarking events are written as they needs to be recorded in transaction log.

jpechane:
@ismailsimsek Hi, I've merged @mfvitale's PR that intorduces signalling channel SPI. Would you be willing to rework the PR to use it?
Thanks

debezium-core/src/main/java/io/debezium/pipeline/signal/channels/MemorySignalChannel.java
```diff
@@ -0,0 +1,64 @@
...
+/**
+ * The class responsible for processing of signals delivered to Debezium via a JVM memory. 
+ * The signal message must have the following structure:
+ * <ul>
+ * <li>{@code id STRING} - the unique identifier of the signal sent, usually UUID, can be used for deduplication</li>
+ * <li>{@code type STRING} - the unique logical name of the code executing the signal</li>
+ * <li>{@code data STRING} - the data in JSON format that are passed to the signal code
+ * </ul>
+ */
+public class MemorySignalChannel implements SignalChannelReader{
```
jpechane, line 27: @ismailsimsek One question - how do you envision usage of this channel? In DebeziumEngine?

ismailsimsek: @jpechane I was thinking for Debezium Server, Using to trigger Read only Snapshot Signals 
something like:

1. User fills form and submits data from web UI of Debezium Server + Quarks 
2. Debezium Server Backend adds this to MemorySignalChannel
3. Debezium process processes it

what do you think? is it  reasonable feature, doable?

jpechane: I think for that we plan to use https://issues.redhat.com/browse/DBZ-6424
We intend to expose this via Jolokia so it will be available using REST endpoint.

ismailsimsek: Thats great, then i will remove it.

debezium-core/src/main/java/io/debezium/pipeline/signal/channels/FileSignalChannel.java
```diff
@@ -0,0 +1,152 @@
...
+            while (lineIterator.hasNext()) {
+                String signalLine = lineIterator.next();
+                if (signalLine == null || signalLine.isBlank()) {
+                    LOGGER.debug("Ignoring empty signal line: `{}`", signalLine);
+                    lineIterator.remove();
+                    continue;
+                }
+                try {
+                    LOGGER.trace("Processing signal line: {}", signalLine);
+                    JsonNode signalJson = mapper.readTree(signalLine);
```
mfvitale, line 103: What do you think extracting line 103-110 in a method to improve readability?

ismailsimsek: updated

jpechane:
@ismailsimsek One question for you. How it is ensured that the same signals from the are not emitted during the next channel polling iteration?

ismailsimsek:
> @ismailsimsek One question for you. How it is ensured that the same signals from the are not emitted during the next channel polling iteration?

@jpechane once it reads the signal file it removes the file content/signals by overriding the file. Line 89
only in case the signal line is not parsed then it adds it back to file

debezium-core/src/main/java/io/debezium/pipeline/signal/channels/FileSignalChannel.java
```diff
@@ -0,0 +1,141 @@
...
+                try {
+                    SignalRecord signal = readSignalString(signalLine);
+                    signals.add(signal);
+                    LOGGER.info("Processing signal: {}, {}, {}, {}", signal.getId(), signal.getType(), signal.getData(), signal.getAdditionalData());
+                }
+                catch (final Exception e) {
+                    LOGGER.warn("Skipped signal due to an error '{}'", signalLine, e);
+                    try {
+                        Files.writeString(signalFile.toPath(), signalLine + System.lineSeparator(), StandardOpenOption.APPEND);
+                    }
```
ismailsimsek, line 108: @jpechane @mfvitale Not sure if we should log it as warning/error and skip the signal or add it back to file? 

here if given line is not parsable or buggy it adds back to the file. which will create WARN message with each read

mfvitale: I think that if the line in not well formed (not a valid JSON) you should warn and skip without put it back again. I don't know if there is a special error for which you need to put it again on the file.

jpechane: I agree, we do the same with db signals as well.

mfvitale (approved):

PR #3908, opened 2022-09-21
Dbz 5504 remove invalid datatypes from db2 connector doc

roldanbob:
[DBZ-5504](https://issues.redhat.com/browse/DBZ-5504)

Replaces #3795, because the original branch had grown too stale. 

Removes invalid `MONEY`, `SMALLMONEY`, `DATETIME2`, and `SMALLDATETIME` entries from the table in _Data type mappings_ section, including the following  `DATETIME2` entries:

>
|`DATETIME2(0)`, `DATETIME2(1)`, `DATETIME2(2)`, `DATETIME2(3)`
|`INT64`
|`io.debezium.time.Timestamp` +
 +
Represents the number of milliseconds since the epoch, and does not include timezone information.

|`DATETIME2(4)`, `DATETIME2(5)`, `DATETIME2(6)`
|`INT64`
|`io.debezium.time.MicroTimestamp` +
 +
Represents the number of microseconds since the epoch, and does not include timezone information.

|`DATETIME2(7)`
|`INT64`
|`io.debezium.time.NanoTimestamp` + 

In the paragraph in the _Timestamp types_ section, I removed references to the invalid SMALLDATETIME` and `DATETIME2 types so that the sentence now references only the `DATETIME` type and the semantic type `io.debezium.time.Timestamp`.  Per [@Naros' comment in the original PR](https://github.com/debezium/debezium/pull/3795#issuecomment-1218041496), the value `1529507596945104` that was used with the `io.debezium.time.MicroTimestamp` semantic type in the original version probably does not make sense with the `io.debezium.time.Timestamp` semantic type. Need to verify this and get input on a valid replacement value.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/db2.adoc
+++ b/documentation/modules/ROOT/pages/connectors/db2.adoc
@@ -1327,7 +1327,7 @@ Represents the number of milliseconds since the epoch, and does not include time
 
 The `DATETIME` type represents a timestamp without time zone information.
 Such columns are converted into an equivalent Kafka Connect value based on UTC.
-For example, the `DATETIME` value "2018-06-20 15:13:16.945104" is represented by an `io.debezium.time.Timestamp` with the value "1529507596945104".
+For example, the `DATETIME` value "2018-06-20 15:13:16.945104" is represented by an `io.debezium.time.Timestamp` with the value "1529507596000".
 
 The timezone of the JVM running Kafka Connect and {prodname} does not affect this conversion.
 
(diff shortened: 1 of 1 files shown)


Naros:
Hi @roldanbob this is marked "waits for another PR", is there a specific PR this is waiting on other than the reference to the question from the other PR you had?  

In terms of the value, I believe it should read `1529507596000` since it's in milliseconds rather than micro or nano -seconds.

roldanbob:
> this is marked "waits for another PR", is there a specific PR this is waiting on

Hey, @Naros. Actually, when I closed the other PR because that branch had grown too stale to salvage, I just copied that from the original PR to preserve the metadata. I think Jiri had originally applied that label? Anyway, I don't have anything that I was waiting on. I know that in #3795 you had mentioned that [you'd hit a snag in the Db2 testing](https://github.com/debezium/debezium/pull/3795#issuecomment-1218041496).

jpechane:
@Naros @roldanbob Yes, that's true. I added the label as that was the way how I interpreted the last @Naros comment on the obsoleted PR.

roldanbob:
Hi @Naros 
> In terms of the value, I believe it should read `1529507596000` since it's in milliseconds rather than micro or nano -seconds.

I updated this to revise the value in the Timestamps example per your comment. Is there anything further needed here?

Naros (approved):
LGTM, thanks @roldanbob

PR #3912, opened 2022-09-22
[docs] update mysql setup docs - binlog_row_value_options

ezerk:
with regards to conversation on [zulip](https://debezium.zulipchat.com/#narrow/stream/302529-users/topic/Debezium.20missing.20update.20events)

`binlog_row_value_options=PARTIAL_JSON` might cause missing UPDATE events on mysql connector in such cases user should validate variable is unset


jpechane:
@ezerk Applied, thanks!

PR #3913, opened 2022-09-22
DBZ-5645 Skip rollback trx cache when LOB is disabled

Naros:
https://issues.redhat.com/browse/DBZ-5645


Naros:

PR #3918, opened 2022-09-23
DBZ-5648 Use a combined REGEXP_LIKE predicate for schema/table filters

Naros:
https://issues.redhat.com/browse/DBZ-5648


jpechane:
@Naros What tests do you intend to add?

Naros:
Hi @jpechane so I had a user apply this locally and this didn't have the impact we thought.  I'm going to move this back to draft for now until we can determine the root-cause.  But the label was to verify against internal CI tests.

jpechane:
@Naros I see, sorry for the confusion. IMHO we roginally use dthis label (as `pending docs`) to denote that tests are missing. IMHO we should make this consistent and for testing there should be two labels - for missing tests and tests needed to be executed.

Naros:
That's fair @jpechane, I'll avoid using this specific label for that going forward.  In either case, this PR isn't necessary and instead we've identified that we need to eliminate the use of `REGEXP_LIKE` entirely.  This seems to perform extremely poorly on a hiigh volume Oracle database and we need to consider an alternative instead.

Naros:
Closed as per above.

PR #3919, opened 2022-09-23
DBZ-5625 Specifies that props use anchored regex; consistency edits

roldanbob:
[DBZ-5625](https://issues.redhat.com/browse/DBZ-5625)

Adds information to property descriptions to explain that the regular expressions that the connector uses to match various identifiers (table names, schema names, column names, etc) are evaluated as anchored regular expressions.
Also edits property descriptions across connectors for consistency.
  
After reviewing a test build of the changes, I noticed a few other properties that take lists of regular expressions:

- {{column.truncate.to.length.chars}} 
- {{column.mask.with.length.chars}} 
- {{column.propagate.source.type}} 
- {{datatype.propagate.source.type}}

I assume that these all behave similarly?

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -293,6 +293,7 @@ Nenad Stojanovikj
 Nick Murray
 Niels Pardon
 Nikhil Benesch
+Niro Levy
 Nishant Singh
 Nitin Agarwal
 Nitin Chhabra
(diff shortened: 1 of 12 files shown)


jpechane:
@roldanbob Hi, it seems to me that there are mixed commits in thi PR. Could you please doublecheck it?

roldanbob:
> @roldanbob Hi, it seems to me that there are mixed commits in thi PR. Could you please doublecheck it?

Arg. Yeah, sorry. I rebased and it looks like I inadvertently pulled in these other commits. :\  I can close this PR and then cherry-pick my commits onto a new branch.

PR #3920, opened 2022-09-25
DBZ-5650 Support set statement in mariadb

nirolevy:
[DBZ-5650](https://issues.redhat.com/browse/DBZ-5650)

From MariaDB 10.1.2, pre-query variables are supported
src: https://mariadb.com/kb/en/set-statement/

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -293,7 +293,7 @@ Nenad Stojanovikj
 Nick Murray
 Niels Pardon
 Nikhil Benesch
-Niro Levy
+Nir Levy
 Nishant Singh
 Nitin Agarwal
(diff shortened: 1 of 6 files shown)


jpechane:
@nirolevy Applied, thanks

jpechane:
@nirolevy Just to doblecheck - can you please check if `STATEMENT` keyword can be used as table/column name? If yes it has to be added to `keywordsCanBeId` rule.

nirolevy:
> @nirolevy Just to doblecheck - can you please check if `STATEMENT` keyword can be used as table/column name? If yes it has to be added to `keywordsCanBeId` rule.

https://issues.redhat.com/browse/DBZ-5662

PR #3921, opened 2022-09-25
DBZ-5651 More reliable wait strategy for DebeziumContainer 

jerrinot:
Reasoning: There original strategy is scrapping logs, but there is a race between "Session key updated" being logged and Kafka Connect REST interface being up. Hence, on slow machines the KC REST `/connectors` interface was occasionally replying with HTTP 404.

### DIFF
--- a/debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/DebeziumContainer.java
+++ b/debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/DebeziumContainer.java
@@ -5,8 +5,6 @@
  */
 package io.debezium.testing.testcontainers;
 
-import static java.time.temporal.ChronoUnit.SECONDS;
-
 import java.io.IOException;
 import java.time.Duration;
 import java.util.Collections;
@@ -19,7 +17,7 @@ import org.testcontainers.containers.GenericContainer;
 import org.testcontainers.containers.KafkaContainer;
 import org.testcontainers.containers.Network;
 import org.testcontainers.containers.PostgreSQLContainer;
-import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
+import org.testcontainers.containers.wait.strategy.HttpWaitStrategy;
 import org.testcontainers.utility.DockerImageName;
 
 import com.fasterxml.jackson.core.type.TypeReference;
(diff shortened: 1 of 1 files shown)


gunnarmorling:
Hey @jerrinot, thanks a lot, that's a very nice improvement! The change looks good, could you only log a quick issue at https://issues.redhat.com/browse/DBZ and rework the commit message here, so that it begins with the issue key: "DBZ-xyz More reliable wait strategy..."? Thanks again!

jerrinot:
@gunnarmorling thanks for a super fast feedback. I force-pushed a new commit message. Is it better now?

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/DebeziumContainer.java
```diff
@@ -46,6 +44,7 @@ public class DebeziumContainer extends GenericContainer<DebeziumContainer> {
     private static final String DEBEZIUM_NIGHTLY_TAG = "nightly";
 
     private static final int KAFKA_CONNECT_PORT = 8083;
+    private static final Duration DEBEZIUM_CONTAINER_STARTUP_TIMEOUT = Duration.ofMinutes(5);
```
jerrinot, line 47: just to highlight it: this increases the timeout quite significantly. the old code waits only up to 1 minute.

rk3rn3r: Is this high value really needed? This code is used for PR tests which might die late with such a high value. Would it be possible for your case to change the timeout to maybe 2 or 3 mins instead?

rk3rn3r:
Thx for that addition! It's a very good enhancement indeed.
LGTM beside the minor important comment above.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/DebeziumContainer.java
jpechane: I agree that 5 mins is a bit too much. I experimented with it and it seems that pull time is not counted to it. In that case I'd say 1 min should be enough. Instead of hard-coded value it is now configurable via globa wait timeout sysproperty.

jerrinot:
hi, sorry, I was AFK. Many thanks to @jpechane for reducing the timeout!

PR #3922, opened 2022-09-27
Bump docker maven plugin version and allow to use external properties

vjuranek:
Allow docker maven plugin external properties to override default or ones specified in `pom.xml`. Also bump docker maven plugin to avoid bug related to multiple images one pom file.

https://issues.redhat.com/browse/DBZ-5657
https://issues.redhat.com/browse/DBZ-5658


jpechane:
@vjuranek Applied, thanks

PR #3924, opened 2022-09-27
Dbz 5625 specify use of anchored regex for include and exclude properties

roldanbob:
[DBZ-5625](https://issues.redhat.com/browse/DBZ-5625)

Adds information to property descriptions to explain that the regular expressions that the connector uses to match various identifiers (table names, schema names, column names, etc) are evaluated as anchored regular expressions.
Also edits property descriptions across connectors for consistency.


jpechane:
@roldanbob Applied, thanks!

PR #3930, opened 2022-10-03
[docs] Mention that Mongo change stream mode is not supported on 3.x

vjuranek:


jpechane:
@vjuranek Applied, thanks

PR #3931, opened 2022-10-03
[docs] Remove oplog capture mode from MongoDB documentation

vjuranek:
`oplog` option was removed in 707eeab3d8c4a3c9 and is not available in 2.x any more. Remove it from Mongo connector documentation.


jpechane:
@vjuranek Applied, thanks for the cleanup!

PR #3933, opened 2022-10-04
DBZ-2385 Snapshot refactoring

harveyyue:
https://issues.redhat.com/browse/DBZ-2385


jpechane:
@harveyyue Hi, thanks for taking the initiative! It is a bit late for 2.0 so I think it make sense to process this after 2.0 is done. In the meantime there is an old PR available - https://github.com/debezium/debezium/pull/2367

It probably makes sense to take a look at that one too?

harveyyue:
@jpechane Adding this feature in 2.1 sounds like a good idea. 
Ok, I will look this pr https://github.com/debezium/debezium/pull/2367.

harveyyue:
There is a postgres it test faiure, should be these changes caused, I will fix it.

jpechane:
Closing as the snapshotting is now refactored and modes unified.

PR #3938, opened 2022-10-05
DBZ-5648 Move LogMiner RegEx handling into Java

Naros:
https://issues.redhat.com/browse/DBZ-5648

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryBuilder.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryBuilder.java
@@ -6,13 +6,8 @@
 package io.debezium.connector.oracle.logminer;
 
 import java.util.Iterator;
-import java.util.List;
-import java.util.regex.Pattern;
 
 import io.debezium.connector.oracle.OracleConnectorConfig;
-import io.debezium.connector.oracle.OracleDatabaseSchema;
-import io.debezium.connector.oracle.logminer.logwriter.LogWriterFlushStrategy;
-import io.debezium.relational.TableId;
 import io.debezium.util.Strings;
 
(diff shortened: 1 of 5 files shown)


Naros:
Currently running the tests against the internal Jenkins CI across all database versions/configs.  I've also asked the user to give this PR a test drive to see whether it yields any better performance.

Naros:
There were two errors on the internal CI:
* `shouldIgnoreAllTablesInExcludedSchemas` - this is just a timing issue, CI didn't perform 10 iterations in 3 minutes likely due to being overloaded on PSI with multiple Oracle test matrix executions
* `shouldSnapshotAndStreamAllRecordsThatSpanAcrossSnapshotStreamingBoundaryLargeTrxs` - this is a known issue and one that I should have a fix for in another Jira related to Oracle RAC before the Final.

Looking at the test suite, I didn't notice any major change in execution time with our setup, so it would be interesting if we could hear back from Ronaldo on whether this helps with their performance or not.  @rlanhellas, can you give this a test drive?

rlanhellas:
Hi @Naros , sure I will download your branch make some tests with our data and give you some feedback.

Naros:
Hi @rlanhellas any update?

Naros:
Hi @jpechane let's proceed with merging this into 2.1.0.Alpha1 and looking to backport this to 2.0 in the future weeks?

rlanhellas:
> Hi @rlanhellas any update?

Sorry for the delay in answering you, but this month was crazy here and I didn't have time to validate this. I will do it by the end of this week and return here with the results.

jpechane:
@Naros Applied, thanks!

jpechane:
@Naros Applied, could you please prepare 1.9/2.0 PR(s) that would be merged later? This PR does not apply cleanly and also we want to have this tested a bit. Thanks

PR #3942, opened 2022-10-06
DBZ-5276 ORA-01289: cannot add duplicate logfile

thangdc94:
https://issues.redhat.com/browse/DBZ-5276


Naros:
Hi @thangdc94 did you test this against Oracle RAC as I'm seeing a `GROUP BY` failure with the new `THREAD#` column being added as a selectable column but not in the GROUP BY clause.

thangdc94:
@Naros I tested it on RAC mode. 
I checked my code again about `Group By` and fixed it.
The reason is that I copied the change from my project to debezium project and forgot updating some part of SQL.

Naros (approved):
Tested locally & on Oracle RAC CI, looks good.  Thanks @thangdc94 !

jpechane:
@thangdc94 Applied, thanks a lot!

PR #3945, opened 2022-10-07
Add LSN into Postregs transaction ID

vjuranek:
https://issues.redhat.com/browse/DBZ-5329


jpechane:
@vjuranek Applied, thanks!

PR #3950, opened 2022-10-10
DBZ-5213 Support for multiarch image build

jcechace:
https://issues.redhat.com/browse/DBZ-5213

related to https://github.com/debezium/container-images/pull/302

### DIFF
--- a/jenkins-jobs/pipelines/release-pipeline.groovy
+++ b/jenkins-jobs/pipelines/release-pipeline.groovy
@@ -525,21 +525,30 @@ node('Slave') {
                 }
             }
             dir("$IMAGES_DIR") {
-                modifyFile('build-all.sh') {
+                modifyFile('build-all-multiplatform.sh') {
                     it.replaceFirst('DEBEZIUM_VERSION=\"\\S+\"', "DEBEZIUM_VERSION=\"$IMAGE_TAG\"")
                 }
             }
             dir(IMAGES_DIR) {
-                sh "env SKIP_UI=true ./build-all.sh"
+                script {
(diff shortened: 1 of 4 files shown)


jcechace:
UI doesn't build on ARM, looking into the issue

jenkins-jobs/pipelines/release-pipeline.groovy
```diff
@@ -525,12 +525,16 @@ node('Slave') {
...
-                modifyFile('build-all.sh') {
+                modifyFile('build-all-multiplatform.sh') {
                     it.replaceFirst('DEBEZIUM_VERSION=\"\\S+\"', "DEBEZIUM_VERSION=\"$IMAGE_TAG\"")
                 }
             }
             dir(IMAGES_DIR) {
-                sh "env SKIP_UI=true ./build-all.sh"
+                sh """
+                    docker run --privileged --rm tonistiigi/binfmt --install all
+                    ./setup-local-builder.sh
```
jpechane, line 535: @jcechace Please make sure that release process does not depend on ARM images being built

jcechace:
@jpechane  

1) deploy-docker-images now builds all platforms and pushes the images directly to both quay and docker hub
2) release-upstream build only linux/amd64 images and pushes them to local repository (a push to repository is required by buildx*)
3) Nightly images are still deployed only for amd64 for now as the job will need more modification in standalone PR

* Technically it doesn't have to push, however it means that only single platform MUST be specified and a different set of cmd flags should be used. IMHO it's better to be uniform and simply run a local registry.

jpechane:
@jcechace Applied, thanks

PR #3952, opened 2022-10-11
DBZ-5710 Remove whitelist/blacklist from log messages

vjuranek:
https://issues.redhat.com/browse/DBZ-5710


jpechane:
@vjuranek Applied, thanks

PR #3954, opened 2022-10-12
DBZ-5715: `Emitting messages with additional fields`

almartino:
For the MongoDB Outbox Event router the property:
`transforms.outbox.table.fields.additional.placement` doesn't works because the Mongo event router looks to the property `transforms.outbox.collection.fields.additional.placement` to expose additional fields.


jpechane:
@almartino Thanks a lot. Merging via https://github.com/debezium/debezium/pull/3956

PR #3957, opened 2022-10-13
DBZ-5724 Add `PRIMARY` as keyword to MySQL grammar to fix index hints

joschi:
Index hints using the primary index (`USE INDEX(PRIMARY)`) are currently failing because `PRIMARY` is missing from the list of keywords in the ANTLR-based MySQL parser.

According to https://dev.mysql.com/doc/refman/8.0/en/keywords.html#keywords-8-0-detailed-P, `PRIMARY` is a reserved keyword.

Index hints: https://dev.mysql.com/doc/refman/8.0/en/index-hints.html


jpechane:
@joschi Applie, thanks! Could you please also prepare for the upstream project https://github.com/antlr/grammars-v4/tree/master/sql/mysql/Positive-Technologies ?

PR #3959, opened 2022-10-14
DBZ-5728 MySqlSchemaMigrationIT runs failed

harveyyue:
https://issues.redhat.com/browse/DBZ-5728


jpechane:
@harveyyue Applied, thanks!

PR #3960, opened 2022-10-14
DBZ-5727 Columns are not excluded when doing incremental snapshots

harveyyue:
https://issues.redhat.com/browse/DBZ-5727


jpechane:

enzo-cappa:
Beautiful! Thanks so much.

PR #3965, opened 2022-10-14
DBZ-5714 tasks.max description now reflects support for multiple task

roldanbob:
[DBZ-5714](https://issues.redhat.com/browse/DBZ-5714)

The ability to support multiple tasks was added to the SQL Server connector, but the documentation was not updated. The behavior of the connector in using multiple tasks is similar to the behavior of the MongoDB connector.


jpechane (changes requested):
@roldanbob I think this is misunderstanding. The task assignment is driven by `database.names`. In SQL Server each database instances can hold multiple logical databases and one or more logical databases are captured by ech task. So the text should not be talking about single/clustered databases as this is a different concept.

roldanbob:
> @roldanbob I think this is misunderstanding. The task assignment is driven by `database.names`. In SQL Server each database instances can hold multiple logical databases and one or more logical databases are captured by ech task. So the text should not be talking about single/clustered databases as this is a different concept.

Ah, got it. Thanks for the clarification, @jpechane. It was not clear about the extent to which the MongoDB description applied in this context. I've updated the description to reflect your comments.

documentation/modules/ROOT/pages/connectors/sqlserver.adoc
```diff
@@ -2096,8 +2096,12 @@ The following configuration properties are _required_ unless a default value is
 
 |[[sqlserver-property-tasks-max]]<<sqlserver-property-tasks-max, `+tasks.max+`>>
 |`1`
-|The maximum number of tasks that should be created for this connector. The SQL Server connector always uses a single task and therefore does not use this value, so the default is always acceptable.
+|Specifies the maximum number of tasks to create for this connector.
+The default value is acceptable when you use the connector with a single database.
+If xref:sqlserver-property-database-names[`database.names`] is configured with multiple entries, so that the connector captures data from more than one logical database, increase the `tasks.max` value to enable the connector to distribute work.
+Set the value to a number that is greater than the number of logical databases to permit the connector to launch a unique task to read from each logical database.
```
jpechane, line 2102:
```suggestion
Set the value to a number that is equal or less than the number of logical databases to permit the connector to launch a unique tasks each to read from a subset of logical databases.
```

rk3rn3r: Why should it be `equal or less`? It must be >1 to distribute work, right? Shouldn't we recommend to set it to a value that equals the number of databases? What happens when the number does not equal the number of databases (greater or lower)?

jpechane: @rk3rn3r Easy answer, it is always about resource management. Ho big your cluster is? Are you the single connector running? How many CPUs? In case of 2 CPUs and 10 databases it might be better to use 2 tasks instead of 10.

Good point on more tasks than databases. @morozov I think this is not a valid config and we should refuse it but just in case do you see any valid use case when it would make sense?

rk3rn3r: Is it possible to add new captured databases dynamically? (Like with incremental snapshots?)
If yes, someone could enable 10 tasks, but only have 9 databases captured with 9 tasks, but add the 10th database dynamically later? The 10th task should not be instantiated until the 10th database was added.

I am not even sure if this dynamic is possible with Kafka Connect as this might require a re-instantiation of the Connector instance? I am not sure about Kafka Connect's capabilities in this regard, but maybe it's worth checking?

morozov: > I think this is not a valid config and we should refuse it but just in case do you see any valid use case when it would make sense?

The number of tasks is the _maximum_ number of tasks that the connector is recommended to use. The connector should not spawn more tasks than it needs (the number of captured databases in this case). Why artificially restrict the maximum at the configuration level?

> I think the code does not yet handle this case, but we should not instantiate tasks that don't have databases assigned?

It doesn't: https://github.com/debezium/debezium/blob/ad2733eb1a29e8afaf1cdc62a21014df9e6f7829/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnector.java#L76-L79

But it should, which would be an easy fix (replace `maxTasks` with `min(maxTasks, databaseNames.size())`).

> I am not even sure if this dynamic is possible with Kafka Connect as this might require a re-instantiation of the Connector instance?

Debezium currently doesn't support this but Kafka Connect does (see `ConnectorContext ::requestTaskReconfiguration()`). In multi-tenant cases, instead of hard-coding a list of databases, it would make perfect sense to have the connector watch the list of databases to be captured and update task configurations accordingly.

roldanbob: Based on the discussion, and given the current code, if we focus on the fact that the property sets a maximum value, it sounds like the solution that @jpechane proposes is reasonable. 

That is, remove the statement about the default value being acceptable only when the connector is configured for a single logical database, and then go on to suggest capping the value at the number of configured logical databases. I've got the following update ready to push:

```
Specifies the maximum number of tasks that the connector uses to capture data from the database instance.
If the value of xref:sqlserver-property-database-names[`database.names`] enables the connector to capture data from multiple logical databases, you can increase number of tasks that the connector uses to read data.
Set the value to a number that is equal to or less than the number of logical databases. 
```

I guess that the open question is whether it's detrimental to performance to exceed a specific value. WDYT?

morozov: > I guess that the open question is whether it's detrimental to performance to exceed a specific value.

It depends on a variety of factors, most of which are outside of the responsibility of the connector (Connect cluster performance, SQL Server performance, network performance, etc.). I'd focus on documenting just the logic of the connector.

Personally, I would omit any performance considerations in this kind of document.

roldanbob: @morozov OK.

> Personally, I would omit any performance considerations in this kind of document.

Do you think that my last rewrite accomplishes that? Or would it improve matters if I remove any suggested contingency on the value of `database.names` and just leave it at something to the effect that user should increment the value to allow the connector to use more tasks"?

morozov: > Do you think that my last rewrite accomplishes that?

I believe so. A few more considerations on the wording:

> Specifies the maximum number of tasks that the connector uses to capture data from the database instance.

I'd replace "uses" with "is allowed to use". The connector won't necessarily use all of them, but it's advised not to use more.

> If the value of xref:sqlserver-property-database-names[`database.names`] enables the connector to capture data from multiple logical databases

IMO, this is too indirect (how does the user know if the value does or does not enable the connector to capture data?) I'd simplify it to "If the `database.names` list contains more than one element".

> you can increase number of tasks that the connector uses to read data.

I'd just add "up to the number of elements in the `database.names` list" and remove the last sentence.

roldanbob: @morozov Revised in the latest commits to incorporate your feedback. Thanks!

morozov: > If the `database.names` list contains more than one element, increase the value of this property

I'm sorry but this is misleading. One task can process an arbitrary number of databases. "[...] you can increase the value [...]" would be more correct.

roldanbob: Got it. Gettin' it. Updated.

jpechane:
@roldanbob Applied, thanks! One sentence and such a big discussion :-)

PR #3972, opened 2022-10-18
DBZ-5682 Allow unique index NULL values in trx reconciliation

Naros:
https://issues.redhat.com/browse/DBZ-5682


Ppojin:
Thank you, @Naros 👍

PR #3974, opened 2022-10-18
DBZ-5352 Clarify table eligibility rules for schema history topic

Naros:
https://issues.redhat.com/browse/DBZ-5352

Note when backporting to 1.9, the `schema.history.internal.` prefix needs to be changed to `database.history.`.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/oracle.adoc
+++ b/documentation/modules/ROOT/pages/connectors/oracle.adoc
@@ -198,6 +198,14 @@ Because the structured representation presents data in JSON or Avro format, cons
 
 [IMPORTANT]
 ====
+By default, the connector uses the `ALL_TABLES` database view to resolve the table names that should be stored in the schema history topic based on the user account's granted permissions.
+
+To adjust which tables are stored in the schema history topic,
+the user account's permissions can be altered, controlling the tables that are visible in the `ALL_TABLES` view or by setting the connector property link:#oracle-property-database-history-store-only-captured-tables-ddl[`schema.history.internal.store.only.captured.tables.ddl`] to `true`.
+====
+
+[IMPORTANT]
+====
 When the connector is configured to capture a table, it stores the history of the table's schema changes not only in the schema change topic, but also in an internal database schema history topic.
 The internal database schema history topic is for connector use only and it is not intended for direct use by consuming applications.
 Ensure that applications that require notifications about schema changes consume that information only from the schema change topic.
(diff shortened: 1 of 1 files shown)


documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -196,6 +196,14 @@ The value of `databaseName` serves as the message key.
 The `tableChanges` field contains an array that includes entries for each column of the table.
 Because the structured representation presents data in JSON or Avro format, consumers can easily read messages without first processing them through a DDL parser.
 
+[IMPORTANT]
+====
+By default, the connector uses the `ALL_TABLES` database view to resolve the table names that should be stored in the schema history topic based on the user account's granted permissions.
```
roldanbob, line 201:
```suggestion
By default, the connector uses the `ALL_TABLES` database view to identify the table names to store in the schema history topic.
Within that view, the connector can access data only from tables that are available to the user account through which it connects to the database.
```
There's a lot packed into this sentence. I think we can make it clearer by breaking out some of the content.
I'm not sure how to interpret `resolve` in this context. I took it to mean _to determine_ or _identify_. But I'm not sure that's right.
When you say `based on the user account's granted permissions` I _think_ that's referring to the permissions granted to the account through which Debezium accesses the database?

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -196,6 +196,14 @@ The value of `databaseName` serves as the message key.
...
+
+To adjust which tables are stored in the schema history topic,
+the user account's permissions can be altered, controlling the tables that are visible in the `ALL_TABLES` view or by setting the connector property link:#oracle-property-database-history-store-only-captured-tables-ddl[`schema.history.internal.store.only.captured.tables.ddl`] to `true`.
```
roldanbob, line 204:
```suggestion
You can modify settings so that the schema history topic stores a different subset of tables.
Use one of the following methods to alter the set of tables that the topic stores:

* Change the permissions of the account that {prodname} uses to access the database so that a different set of tables are visible in the `ALL_TABLES` view.
* Set the connector property link:#oracle-property-database-history-store-only-captured-tables-ddl[`schema.history.internal.store.only.captured.tables.ddl`] to `true`.
```

roldanbob (changes requested):
Thanks for running this by me. 
The original text is pretty dense, which made it hard for me to parse out. I've tried to relax it a bit and spread out the thoughts, but  I'm not sure whether my interpretation of the original text is correct.

jpechane:
@Naros Applied, thanks
@roldanbob Thanks for the review!

jpechane:
@Naros Does this resolve the linked Jira issue?

Naros:
It does @jpechane.

PR #3977, opened 2022-10-20
DBZ-5744 Add strimzi operator name as param

obabec:


jpechane:
@obabec Applied, thanks!

PR #3978, opened 2022-10-20
DBZ-5629 Correctly setting task id for MongoDbTaskContext

jcechace:
https://issues.redhat.com/browse/DBZ-5629

Root cause of the issue seems to be the use of default task ID. This in turn leads to the same name of the mbean for each task.


jcechace:
Tested with simple unit test. The initial idea after discussing with @Naros was to introduce an ``AbstractCdcCdcSourceTaskContext`` in embedded module, however that base class would have to be either heavily generic or it would need to store all the attributes passed to constructor of ``CdcSourceTaskContext``, since there is no generic way to verify the correct values. Alternatively there would be ``CdcSourceTaskContextTest`` in embedded / core module which would verify the functionality of ``CdcSourceTaskContext``... however that seems like testing getters.

Naros (approved):
LGTM, this seems like a reasonable way to test this.

jpechane:
@jcechace Probably the easiest solution now, thanks!

PR #3981, opened 2022-10-21
DBZ-5748 Remove note that refers to incremental snapshots as TP feature

roldanbob:
[DBZ-5748](https://issues.redhat.com/browse/DBZ-5748)

Deletes from the shared snapshot metrics file an incremental snapshots Tech Preview note that was intended for removal when the feature was promoted to GA in the previous 2022.Q3 downstream release .  A[ separate commit in the downstream docs repo ](https://gitlab.cee.redhat.com/red-hat-integration-documentation/integration/-/merge_requests/993)removes the note from the published downstream documentation.


jpechane:
@roldanbob Applied, thanks

PR #3985, opened 2022-10-22
DBZ-4720 Validate Debezium Server configuration properties

jeremy-l-ford:
Valication connector before attempting to initialize

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/connector/simple/SimpleSourceConnector.java
+++ b/debezium-embedded/src/main/java/io/debezium/connector/simple/SimpleSourceConnector.java
@@ -100,7 +100,7 @@ public class SimpleSourceConnector extends SourceConnector {
 
     @Override
     public ConfigDef config() {
-        return null;
+        return new ConfigDef();
     }
 
     public static class SimpleConnectorTask extends SourceTask {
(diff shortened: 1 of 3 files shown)


spicy-sauce:
@ggaborg, please review this PR (this issue was opened by me at the time).

ggaborg (approved):

ggaborg (commented):
I think there should be some tests for each connector

ggaborg:
Also, this change only touches the validation for the connector properties, not other configuration dependent components like offset storages, schema history, etc.

jeremy-l-ford:
For SchemaHistory, each implementation is already validating itself in the configuration method.

For the OffsetBackingStore, perhaps as part of DBZ-1971, validation support could be added.

In terms of tests for each connector, are you thinking that MySqlConnectorConfig as an example needs to be unit tested?

ggaborg:
@jeremy-l-ford, hi
1. "For SchemaHistory, each implementation is already validating itself in the configuration method." - this only validates required properties and is not similar with the connector validation.
2. "For the OffsetBackingStore, perhaps as part of DBZ-1971, validation support could be added." - I was thinking about a unified mechanism to validate configuration, see 1.
3. "In terms of tests for each connector, are you thinking that MySqlConnectorConfig as an example needs to be unit tested?" yes

PR #3958, opened 2022-10-13
DBZ-5661 Fix for Embedded Engine retrying indefinitely even when max retries is 0

poonam-meghnani:
## External Links

[DBZ-5661](https://issues.redhat.com/browse/DBZ-5661)

## Description
The changes is to break the indefinite retries upon RetriableException in EmbeddedEngine, when the max retries is set to 0. 

## Notes for Reviewer

As mentioned in the ticket DBZ-5661, the EmbeddedEngine does not respect the property `errors.max.retries`. It is required to stop retrying to restart the task if the property is set to 0.

PR #3986, opened 2022-10-24
DBZ-5661 Fix for Embedded Engine retrying indefinitely even when max retries is 0

poonam-meghnani:
## External Links

[DBZ-5661](https://issues.redhat.com/browse/DBZ-5661)

## Description
The changes is to break the indefinite retries upon RetriableException in EmbeddedEngine, when the max retries is set to 0. 

## Notes for Reviewer
This PR is a preliminary fix to the problem running replication using EmbeddedEngine and retrying endlessly on failure. This will break the infinite looping over RetriableException and let the engine exit gracefully when user choose not to retry through below config property set to zero:
`errors.max.retries`
@jpechane, while I will continue to work on [Original PR ](https://github.com/debezium/debezium/pull/3958) to achieve more consistent behaviour for RetriableExceptions between Kafka connectors and EmbeddedEngine, this simple solution unblocks  from the retry loop and enables to use EmbeddedEngine with the above mentioned config property to escape retries in the meantime.

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
@@ -864,6 +864,10 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
                                         delayStrategy.sleepWhen(!startedSuccessfully);
                                     }
                                 }
+                                else {
+                                    handlerError = e;
+                                    break;
+                                }
                             }
                             try {
                                 if (changeRecords != null && !changeRecords.isEmpty()) {
--- a/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
@@ -865,8 +865,7 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
                                     }
                                 }
                                 else {
-                                    handlerError = e;
-                                    break;
+                                    throw e;
                                 }
                             }
                             try {


spicy-sauce:
@ggaborg, please review this PR.

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
```diff
@@ -864,6 +864,10 @@ public Map<String, String> configs() {
                                         delayStrategy.sleepWhen(!startedSuccessfully);
                                     }
                                 }
+                                else {
+                                    handlerError = e;
```
ggaborg, line 868: handlerError means an error in the handler which is not the case here. I believe you can obtain the desired behavior by just re-throwing the caught RetriableException:
else {
    throw e;
}

ggaborg: @jpechane I don't think it is ok to change the behavior for max retries, or at least we need to have a specific value which 'disables' the mechanism in a way that it falls back to the previous behavior (prior to  https://github.com/debezium/debezium/pull/3857). This was the intention with max retries 0 and it is important to have it for connectors like mysql for example which do not need this retry mechanism because it works out of the box without it. 
I think it is better to introduce another value for max retries e.g. -1 = stop the connector, -2 or less = infinite retries. 
But even with this, I think that this specific case when CDC is not enabled for SQL Server it should be caught when starting the connector and should not reach the poll() loop; or define this use case as non retriable exception.

ggaborg: @poonam-meghnani just break is even better because it allows for offset commit, so:
else {
    break;
}

poonam-meghnani: > @poonam-meghnani Hi, handlerError means an error in the handler which is not the case here. I believe you can obtain the desired behavior by just re-throwing the caught RetriableException: else { throw e; }

@ggaborg Thanks for the review and suggestion. I agree that handlerError is specific to the handle. I feel the best workaround would be to define a retryError similar to handlerError. 

Re-throwing the retriable error works as well:
```
else {
throw e;
}
```
however in finally, it marks the connector to have successfully completed at first and then fails in a later catch, which does the trick.

poonam-meghnani: > @poonam-meghnani just break is even better because it allows for offset commit, so: else { break; }

This will break the loop, but not report the error.

ggaborg: @poonam-meghnani, correct. So probably retryError is the way to go. But again, if we do this when max retries is 0, then we don't have a value to disable this retries mechanism and obtain the old behavior.

poonam-meghnani: This works as an immediate fix for the problem in hand, I agree that the old behaviour needs careful handling and we have a [PR](https://github.com/debezium/debezium/pull/3958) up for this and marked 'needs more thinking'

poonam-meghnani: @ggaborg, @jpechane I have put up another PR for the suggestions from this one. Both works well as a primary fix to the problem. Let me know if we could atleast get the EmbeddedEngine working first and then re-think of how to get the old behaviour and max-retries work together? Thanks!

poonam-meghnani:
Agreed on suggestion in [PR](https://github.com/debezium/debezium/pull/4001), hence closing this!

PR #3654, opened 2022-06-27
DBZ-5181 Specify a taskId and maxTasks to run multiple EmbeddedEngines

vrajat:
https://issues.redhat.com/browse/DBZ-5181

A Debezium Server implementation may run multiple engines to parallelize
work between them. To enable multi-tasking an engine should be assigned
an ID and the provide total number of tasks. Using this information, it
can identify the slice of work from the source connector.

PR #3989, opened 2022-10-24
DBZ-5181: Enable EmbeddedEngine to be used in a multi-threaded server

jcechace:
https://issues.redhat.com/browse/DBZ-5181

Continues work from https://github.com/debezium/debezium/pull/3654

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
@@ -659,15 +659,15 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
         embeddedConfig.put(WorkerConfig.KEY_CONVERTER_CLASS_CONFIG, JsonConverter.class.getName());
         embeddedConfig.put(WorkerConfig.VALUE_CONVERTER_CLASS_CONFIG, JsonConverter.class.getName());
 
-        //If multiple engines are started in a server, ensure that the offset
-        //storage does not clash with each other. 
+        // If multiple engines are started in a server, ensure that the offset
+        // storage does not clash with each other.
         if (this.maxTasks != null && this.taskId != null && this.maxTasks > 0) {
             if (embeddedConfig.containsKey(OFFSET_STORAGE_FILE_FILENAME.name())) {
-                embeddedConfig.replace(OFFSET_STORAGE_FILE_FILENAME.name(), 
+                embeddedConfig.replace(OFFSET_STORAGE_FILE_FILENAME.name(),
                         String.format(embeddedConfig.get(OFFSET_STORAGE_FILE_FILENAME.name()), this.taskId));
-            } 
+            }
             else if (embeddedConfig.containsKey(OFFSET_STORAGE_KAFKA_TOPIC.name())) {
-                embeddedConfig.replace(OFFSET_STORAGE_KAFKA_TOPIC.name(), 
+                embeddedConfig.replace(OFFSET_STORAGE_KAFKA_TOPIC.name(),
                         String.format(embeddedConfig.get(OFFSET_STORAGE_KAFKA_TOPIC.name()), this.taskId));
             }
         }
@@ -725,7 +725,6 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
 
             final String engineName = config.getString(ENGINE_NAME);
             final String connectorClassName = config.getString(CONNECTOR_CLASS);
-
 
             LOGGER.info("Setting up Task name: {}, id: {}, maxTasks: {}", engineName, taskId, maxTasks);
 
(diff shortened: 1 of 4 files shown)


jcechace:
I will switch from draft once I have at least some basic test-coverage for this.

debezium-embedded/src/test/java/io/debezium/embedded/EmbeddedEngineTest.java
```diff
@@ -260,6 +263,54 @@ public void interruptedOffsetCommitShutsDown() throws Exception {
...
+                        int id = key.getInt32("id");
+                        actual.add(id);
+                        latch.countDown();
+                        committer.markProcessed(event);
+                    }
+                    committer.markBatchFinished();
+                })
+                .using(this.getClass().getClassLoader());
+
+        try (var engine = engineBuilder.build()) {
```
jcechace, line 301: What would be the recommended way to speedup the shutwodn? By default the engine will wait before stopping by the amount specified in ``EmbeddedEngine#WAIT_FOR_COMPLETION_BEFORE_INTERRUPT_PROP`` which is by default 5 minutes.

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
```diff
@@ -775,7 +816,7 @@ public OffsetStorageReader offsetStorageReader() {
                     // Start the connector with the given properties and get the task configurations ...
                     connector.start(workerConfig.originalsStrings());
                     connectorCallback.ifPresent(DebeziumEngine.ConnectorCallback::connectorStarted);
-                    List<Map<String, String>> taskConfigs = connector.taskConfigs(1);
+                    List<Map<String, String>> taskConfigs = connector.taskConfigs(maxTasks);
```
jcechace, line 819: After discussing with @Naros came to the conclusion that this is potentially problematic. When running server with multiple tasks each embedded engine will perform its own call to ``connector.taskConfigs(maxTasks)``. However, there is no guarantee that the result of these calls won't be different.  Current approach works just fort connectors with static task configurations and for example Debezium's MongoDb connector would be error prone as it retrieves the current replicaset configuration (which could change  between server starting multiple engines), furthermore the connector requests task reconfiguration in case of change to replica sets. 

A partial solution (not dealing with the reconfiguration requests) would be to push the for loop from DebeziumServer down to EmbeddedEngine directly, having Engine run multiple tasks, each in its own thread.

jcechace: @jpechane  WDY think? 

@vrajat  Are you still interested in working on this? Depending on how the change goes it might be above my currently feasible time allocation.

debezium-embedded/src/test/java/io/debezium/embedded/EmbeddedEngineTest.java
jcechace: The reason is that the `SimpleSourceConnector#poll()` purposefully blocks forever. This will have to be changed or made configurable.

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
vrajat: What is a good medium to discuss and write a design doc? I agree that if `taskConfigs` doesn't return a stable set, then this approach will not work. Is it practical to add a new API that should return stable results ? A connector can choose to implement it or not.

One advantage of multiple engines is that the complete pipeline is parallel including the transformers and sink connectors.

jpechane: IMHO it makes sense to push the multithreading down to Engine. The Debezium Engine API will be extended with an option to pass `ThreadFactory`.

When working it makes sense to look at reconfiguration.

vrajat: If I understand you correctly, the first part of `EmbeddedEngine::run()` will get the configuration and setup the tasks. Then there will be potentially multiple threads executing the [while loop](https://github.com/debezium/debezium/blob/main/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java#L833) using the task configs assigned to them?

jcechace: @vrajat  Correct. Apology for the late answer, I was out sick.

axeh:
Any updates on this? This PR is what I need on my project :)

jcechace:
@axeh Not yet. As mentioned earlier the original designed proved to be flawed. Currently this is a bit further away in our roadmap, however we would be more than happy to accept a PR (review and assistance included of course).

jcechace:
Closing due to recent engine rewrite

axeh:
So, this feature is not going to implement or? can you just share "recent engine rewrite" PRs

vjuranek:
hi @axeh the draft of the design document is [here](https://github.com/debezium/debezium-design-documents/pull/8) and the draft of implementation [here](https://github.com/debezium/debezium/pull/5172)

PR #3990, opened 2022-10-24
DBZ-5763 disabling connection encryption for SqlServer connections 

jcechace:
https://issues.redhat.com/browse/DBZ-5763

This is fix for ocp connections, for test-containers it needs further investigation


obabec (approved):
LGTM

PR #3993, opened 2022-10-24
DBZ-5615 Mask additional credentials during log events

Naros:
https://issues.redhat.com/browse/DBZ-5615


jpechane:
@Naros Applied, thanks!

PR #3995, opened 2022-10-25
DBZ-5743 Example Change the validFullname replacement rule

SimonChou12138:
https://issues.redhat.com/browse/DBZ-5743

It is uniquely encoded through Unicode's hexadecimal to decimal counterpart, preserving the topic language while conforming to the kafka naming conventions. Avoid the situation that topic names are the same when special characters of the same length are replaced with _

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -461,3 +461,4 @@ Inki Hwang
 合龙 张
 Phạm Ngọc Thắng
 Moustapha Mahfoud
+周民帅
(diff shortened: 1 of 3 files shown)


SimonChou12138:
@jpechane 
Thank you for your reply. I think your suggestions are very valuable and accurate. Looking forward to the next version of this feature

harveyyue:
@jpechane Yes, I think adding a new topic naming strategy is good way. 
I'm not sure whether we only need to add new topic strategy to process the non-alpha chars, since schema name adjuster do the naming checks after topic stragegy.

jpechane:
Closing as a larger design change is discussed. The comments were copied to Jira.

SimonChou12138:
> Closing as a larger design change is discussed. The comments were copied to Jira.

Hi jpechane, will this new topic naming policy idea be implemented in 2.1? Or does closure mean the idea will be abandoned

PR #3998, opened 2022-10-25
DBZ-5760 Only log full row data at TRACE level

Naros:
https://issues.redhat.com/browse/DBZ-5760


jpechane:
@Naros Applied, thanks!

PR #4000, opened 2022-10-25
DBZ-5759 Improve data logged by the Oracle connector

Naros:
https://issues.redhat.com/browse/DBZ-5759

I've started a discussion on the Jira issue about any other logging that might be useful.  I've added the two bits that I know of that I wanted to have added here, but I wanted some clarification on the others mentioned in the Jira before we merge.


jpechane:
@Naros Where do we stand with this PR? I commented on the linked Jira.

Naros:
HI @jpechane this should be good to go for 2.1.0.Alpha1 and preferably backported if possible.

jpechane:
@Naros Nice improvement, thanks!

PR #4001, opened 2022-10-26
DBZ-5661 Fix for Embedded Engine retrying indefinitely even when max retries is 0

poonam-meghnani:
## External Links

[DBZ-5661](https://issues.redhat.com/browse/DBZ-5661)

## Description
The changes is to break the indefinite retries upon RetriableException in EmbeddedEngine, when the max retries is set to 0. 

## Notes for Reviewer
This PR is a preliminary fix to the problem running replication using EmbeddedEngine and retrying endlessly on failure. This will break the infinite looping over RetriableException and let the engine exit gracefully when user choose not to retry through below config property set to zero:
`errors.max.retries`
@jpechane, while I will continue to work on [Original PR ](https://github.com/debezium/debezium/pull/3958) to achieve more consistent behaviour for RetriableExceptions between Kafka connectors and EmbeddedEngine, this simple solution unblocks  from the retry loop and enables to use EmbeddedEngine with the above mentioned config property to escape retries in the meantime.

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
@@ -826,7 +826,7 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
                     }
 
                     recordsSinceLastCommit = 0;
-                    Throwable handlerError = null;
+                    Throwable handlerError = null, retryError = null;
                     try {
                         timeOfLastCommitMillis = clock.currentTimeInMillis();
                         RecordCommitter committer = buildRecordCommitter(offsetWriter, task, commitTimeout);
@@ -866,6 +866,7 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
                                         catch (Exception ex) {
                                             if (totalRetries == maxRetries) {
                                                 LOGGER.error("Can't start the connector, max retries to connect exceeded; stopping connector...", ex);
+                                                retryError = ex;
                                                 throw ex;
                                             }
                                             else {
@@ -874,6 +875,11 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
                                         }
                                         delayStrategy.sleepWhen(!startedSuccessfully);
                                     }
+                                }
+                                else {
+                                    LOGGER.error("Can't start the connector, max retries to connect exceeded; stopping connector...", e);
+                                    retryError = e;
+                                    throw e;
                                 }
                             }
                             try {
@@ -912,6 +918,10 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
                             fail("Stopping connector after error in the application's handler method: " + handlerError.getMessage(),
                                     handlerError);
                         }
+                        if (retryError != null) {
+                            fail("Stopping connector after max retry limit exceeded: " + retryError.getMessage(),
+                                    retryError);
+                        }
                         try {
                             // First stop the task ...
                             LOGGER.info("Stopping the task and engine");
@@ -919,7 +929,7 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
                             connectorCallback.ifPresent(DebeziumEngine.ConnectorCallback::taskStopped);
                             // Always commit offsets that were captured from the source records we actually processed ...
                             commitOffsets(offsetWriter, commitTimeout, task);
-                            if (handlerError == null) {
+                            if (handlerError == null && retryError == null) {
                                 // We stopped normally ...
                                 succeed("Connector '" + connectorClassName + "' completed normally.");
                             }


poonam-meghnani:
@ggaborg, @jpechane I have put up this PR for the suggestions from [this one](https://github.com/debezium/debezium/pull/3986). Both works well as a primary fix to the problem. Let me know if we could atleast get the EmbeddedEngine working first and then re-think of how to get the old behaviour and max-retries work together? Thanks!

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
```diff
@@ -875,6 +876,11 @@ public Map<String, String> configs() {
                                         delayStrategy.sleepWhen(!startedSuccessfully);
                                     }
                                 }
+                                else {
+                                    LOGGER.error("Can't start the connector, max retries to connect exceeded; stopping connector...", e);
```
ggaborg, line 880: maybe change this message since there are no retries

ggaborg:
@jpechane @poonam-meghnani I still think that is not ok to change the behavior for max retries, or at least we need to have a specific value which 'disables' the mechanism in a way that it falls back to the previous behavior (prior to https://github.com/debezium/debezium/pull/3857). This was the intention with max retries 0 and it is important to have it for connectors like mysql for example which do not need this retry mechanism because it works out of the box without it.
I think it is better to introduce another value for max retries e.g. -1 = stop the connector, -2 or less = infinite retries.

jpechane:
@ggaborg You are right but I am really not in favour of adding more magic values. How urgent /critical it is? Can we intorudced it as undocumented magic value to unblock and provide the final solution in the other PR?

ggaborg:
@jpechane I believe this is critical for @poonam-meghnani. We can introduce value -2 as undocumented, yes. All I was pointing out was that we cannot change the behavior for value 0 sine it is already commited.

poonam-meghnani:
> @ggaborg You are right but I am really not in favour of adding more magic values. How urgent /critical it is? Can we intorudced it as undocumented magic value to unblock and provide the final solution in the other PR?

@jpechane,
We have this issue where we run into infinite retries which is blocking us, as a solution we could: 
1. add a magic value as suggested above
2. Add a new property (RETRIABLE_MAX_RESTART :: similar to RETRIABLE_RESTART_WAIT) to `BaseSouceTask` to control restart on retriable exception when falls back to old behaviour. 
3. Refine `SqlServerErrorHandler` class to filter out certain non-recoverable errors from RetriableError similar to `OracleErrorHandler`. 

Which one would be suitable? 
We are willing to do the work but we’d love to be able to merge a fix as it’s somewhat of a blocker for us.

alwaysbemark:
Heya @ggaborg @jpechane , thanks for all the help so far. We would really love some guidance on which solution would be the most suitable as we're currently blocked (the embedded engine enters an infinite loop of retries which we cannot exit out of). Thank you!

ggaborg:
@alwaysbemark @poonam-meghnani @jpechane Hi, I suggest the current change but with an undocumented value of -2 in order to not change the behavior when 0

jpechane:
@poonam-meghnani I am sorry for procrastinating on this one for that long :-(.

So the way forward - I see there are mutliple PRs. Please close all no longer relevant and switch the one with long term solution into draft state to keep the table clean.

PLease intorduce the `-2` undocumented option to move forward this one so we can merge it.

poonam-meghnani:
> @poonam-meghnani I am sorry for procrastinating on this one for that long :-(.
> 
> So the way forward - I see there are mutliple PRs. Please close all no longer relevant and switch the one with long term solution into draft state to keep the table clean.
> 
> PLease intorduce the `-2` undocumented option to move forward this one so we can merge it.

Thanks @jpechane, @ggaborg,
I will action the suggested soon.

poonam-meghnani:
> @poonam-meghnani I am sorry for procrastinating on this one for that long :-(.
> 
> So the way forward - I see there are mutliple PRs. Please close all no longer relevant and switch the one with long term solution into draft state to keep the table clean.
> 
> PLease intorduce the `-2` undocumented option to move forward this one so we can merge it.

@jpechane @ggaborg, 
I have made changes to stop connector if errors.max.retries is less than DEFAULT value which is -1. I have tested the changes for below cases:

- when `errors.max.retries=0` and connector fails for retriable error: falls back on to the previous behaviour and retries after a set interval `retriable.restart.connector.wait.ms`
- when `errors.max.retries=-1` and connector fails for retriable error: keeps retrying to start task indefinitely.
- when `errors.max.retries=-2` and connector fails for retriable error:  _stops the connector without retrying._
- when `errors.max.retries=2` and connector fails for retriable error: keeps retrying to start task indefinitely, I am not sure if thats expected behaviour. Please note this behaviour is same without the above changes.

Please let me know if value less than -1 sounds to be good candidate for stopping connector. 
Thank you!

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
```diff
@@ -852,6 +854,11 @@ public Map<String, String> configs() {
                                 int maxRetries = getErrorsMaxRetries();
                                 LOGGER.info("Retriable exception thrown, connector will be restarted; errors.max.retries={}", maxRetries, e);
                                 if (maxRetries != 0) {
+                                    if (maxRetries < DEFAULT_ERROR_MAX_RETRIES) {
```
ggaborg, line 857: maybe move outside the if you introduced:
if (maxRetries < DEFAULT_ERROR_MAX_RETRIES) {
...
}
else if (maxRetries != 0) {
...

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
```diff
@@ -852,6 +854,11 @@ public Map<String, String> configs() {
...
+                                        LOGGER.error("Can't start the connector, no attempts made to retry; stopping connector...", e);
```
ggaborg, line 858: the message is misleading, there is no attempt on starting the connector. I don't think we should log anything here since it will be logged when testing the retryError handler

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
```diff
@@ -912,14 +920,18 @@ public Map<String, String> configs() {
                             fail("Stopping connector after error in the application's handler method: " + handlerError.getMessage(),
                                     handlerError);
                         }
+                        if (retryError != null) {
```
ggaborg, line 923: else if since the errors can't be both not null

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
```diff
@@ -912,14 +920,18 @@ public Map<String, String> configs() {
...
+                        }
                         try {
                             // First stop the task ...
                             LOGGER.info("Stopping the task and engine");
                             task.stop();
                             connectorCallback.ifPresent(DebeziumEngine.ConnectorCallback::taskStopped);
                             // Always commit offsets that were captured from the source records we actually processed ...
                             commitOffsets(offsetWriter, commitTimeout, task);
-                            if (handlerError == null) {
+                            if (handlerError == null && retryError == null) {
```
ggaborg, line 934: maybe this if should be moved upper entirely like this:
if (handlerError != null)  {
...
}
else if (retryError != null) {
...
}
else {
// We stopped normally ...
    succeed("Stopping connector '" + connectorClassName + "' after completed normally.");
}

poonam-meghnani:
@ggaborg, thanks for the suggestions. Actually, I was trying to make minimal changes, but your suggestions totally makes sense. I made changes and tested for above mentioned test cases, please have a look again!

ggaborg (approved):

jpechane:
@poonam-meghnani There is still an issue in Debezium Server tests - looks it is locked in an infinite loop.

ggaborg:
@jpechane @poonam-meghnani This issue is random, I've seen it in many failed builds. The test is pausing the container but then gets stuck on one of connection.execute() or connection.close() and is unable to unpause it. I'm basing my logic on the fact that I can see in the logs the "Creating new redis_test table and inserting 5 records to it" message but not the "Sleeping for 2 seconds to flush records" and the repeating of log "Can't connect to Redis: io.debezium.storage.redis.RedisClientConnectionException: redis.clients.jedis.exceptions.JedisConnectionException: java.net.SocketTimeoutException: Read timed out".

Here is the piece of code in the test:

        // pause container
        Testing.print("Pausing container");
        RedisTestResourceLifecycleManager.pause();

        final MySqlConnection connection = getMySqlConnection();
        connection.connect();
        Testing.print("Creating new redis_test table and inserting 5 records to it");
        connection.execute("CREATE TABLE inventory.redis_test (id INT PRIMARY KEY)");
        connection.close();

        Testing.print("Sleeping for 2 seconds to flush records");
        Thread.sleep(2000);
        Testing.print("Unpausing container");
        RedisTestResourceLifecycleManager.unpause();

poonam-meghnani:
> @poonam-meghnani There is still an issue in Debezium Server tests - looks it is locked in an infinite loop.

@jpechane, I have rebased my changes and Debezium Server tests fails for different error now:
```
Caused by: redis.clients.jedis.exceptions.JedisConnectionException: Unknown reply: J
	at redis.clients.jedis.Protocol.process(Protocol.java:140)
	at redis.clients.jedis.Protocol.read(Protocol.java:192)
	at redis.clients.jedis.Connection.readProtocolWithCheckingBroken(Connection.java:316)
	at redis.clients.jedis.Connection.getStatusCodeReply(Connection.java:243)
	at redis.clients.jedis.Jedis.ping(Jedis.java:354)
	at io.debezium.storage.redis.RedisConnection.getRedisClient(RedisConnection.java:83)
```
It does not seem to be related to the changes, could you please guide through this?

jpechane:
@poonam-meghnani Applied, thanks a lot, it took a lot of back and forth discussion!
Now it is probably about time to create a new separate Jira and we can do design discussion first for how to architecture the error handling from now on.

poonam-meghnani:
Thanks @jpechane and @ggaborg! I will create a new ticket for a long term solution.

PR #4004, opened 2022-10-26
DBZ-5774 Upgrade to Quarkus 2.14.CR1

Naros:
https://issues.redhat.com/browse/DBZ-5774

- [x] Source jandex version from Quarkus BOM

In regards to the upstream PR requirements of https://github.com/quarkusio/quarkus-platform/pull/673

### DIFF
--- a/debezium-quarkus-outbox/runtime/pom.xml
+++ b/debezium-quarkus-outbox/runtime/pom.xml
@@ -30,9 +30,9 @@
         </dependency>
         <!-- Needed specifically for AdditionalJaxbMappingProducerImpl usage -->
         <dependency>
-            <groupId>org.jboss</groupId>
+            <groupId>io.smallrye</groupId>
             <artifactId>jandex</artifactId>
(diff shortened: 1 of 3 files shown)


Naros:
So @jpechane it would appear that our Build Parent POM specifies the Netty version as `4.1.78.Final` and currently the Pravega module specifies Netty as `4.1.52.Final` and yet Quarkus has upgraded to `4.1.82.Final`, which makes keeping these versions aligned quite complicated.

I tested using `4.1.78.Final` in the Pravega module and this continued to exhibit the same build failure.  Once I modified both the Build Parent POM and Pravega to both use `4.1.82.Final`, which is what is aligned with Quarkus 2.14.CR1, then the build error went away, which seems to indicate there is an incompatibility between 4.1.78 and 4.1.82 of Netty.

What do you suggest we do here?

Naros:
So I ended up explicitly aligning the netty dependency in the Pravega module with Quarkus.

PR #4005, opened 2022-10-27
DBZ-5402: debezium-storage-s3 module

blcksrx:
This implementation provides caching `HistoryRecord` on the main memory.
Since S3 does not support Append operation on the object level, `S3SchemaHistory#start()` fetches history log
from S3 first and store the `HistoryRecord`'s  on the main memory. 
Also `S3SchemaHistory#storeRecord(HistoryRecord)` creates new history object every time invokes on S3.
S3 Object keys created names after the `Thread` that creates `DatabaseHistory` object.

### DIFF
--- a/debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/history/S3SchemaHistory.java
+++ b/debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/history/S3SchemaHistory.java
@@ -75,6 +75,7 @@ public class S3SchemaHistory extends AbstractSchemaHistory {
     public static final String SECRET_ACCESS_KEY_CONFIG = "s3.secret.access.key";
     public static final String REGION_CONFIG = CONFIGURATION_FIELD_PREFIX_STRING + "s3.region.name";
     public static final String BUCKET_CONFIG = CONFIGURATION_FIELD_PREFIX_STRING + "s3.bucket.name";
+    public static final String OBJECT_NAME_CONFIG = CONFIGURATION_FIELD_PREFIX_STRING + "s3.object.name";
     public static final String ENDPOINT_CONFIG = CONFIGURATION_FIELD_PREFIX_STRING + "s3.endpoint";
     public static final String OBJECT_CONTENT_TYPE = "text/plain";
 
@@ -101,6 +102,12 @@ public class S3SchemaHistory extends AbstractSchemaHistory {
             .withType(Type.STRING)
             .withImportance(Importance.HIGH);
 
+    public static final Field OBJECT_NAME = Field.create(OBJECT_NAME_CONFIG)
+            .withDisplayName("S3 Object name")
+            .withType(Type.STRING)
+            .withImportance(Importance.HIGH)
+            .withDescription("The name of the object under which the history is stored.");
+
     public static final Field ENDPOINT = Field.create(ENDPOINT_CONFIG)
             .withDisplayName("S3 endpoint")
             .withType(Type.STRING)
@@ -112,9 +119,9 @@ public class S3SchemaHistory extends AbstractSchemaHistory {
     private final FunctionalReadWriteLock lock = FunctionalReadWriteLock.reentrant();
     private final DocumentWriter writer = DocumentWriter.defaultWriter();
     private final DocumentReader reader = DocumentReader.defaultReader();
-    private final String objectName = String.format("db-history-%s.log", Thread.currentThread().getName());
 
     private String bucket = null;
+    private String objectName = null;
     private Region region = null;
     private URI endpoint = null;
     private AwsCredentialsProvider credentialsProvider = null;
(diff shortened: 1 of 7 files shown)


blcksrx:
@jpechane  I'm really sorry that I got back too late to fix my commits. I appreciate a lot the works you all did about it.

jpechane:
@blcksrx Also please add the new modules to https://github.com/debezium/debezium/blob/main/.github/workflows/debezium-workflow.yml#L715

blcksrx:
@jpechane Test didn't run in the CI, shall I add `install` execution in the `debezium-storage-testing`

jpechane:
@blcksrx Good point about CI. Let's make one step back. Could you please remove the change in the workflow you did related to adding the S3 module abd instead of it add a new job in the that will build and test storage module in the same way as Debezium Server is? That's probably most univesral and scalable solution that would solve extending Debeizum with future storage modules. Thanks

jpechane:
@blcksrx It seems the tests are failing and also in the logs there it looks like there is an issue with dependencies,

blcksrx:
@jpechane Yeah, I need to add `debezium-storage-s3` to `build_db` and `build_debeizum-server`.
In addition theres a problem that i'm facing that when I run maven with `-amd` it runs tests for all the dependencies such as MySQL and when I use `-am` it doesn’t run any tests

blcksrx:
@jpechane I might need help on this. the `debezium-storage` has a lots of dependency with other modules of debezium that on building it would run tests for these modules:
```shell
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Build Order:
[INFO]
[INFO] Debezium Storage Module                                            [pom]
[INFO] Debezium Storage Kafka Module                                      [jar]
[INFO] Debezium Storage File Module                                       [jar]
[INFO] Debezium Schema Generator                                 [maven-plugin]
[INFO] Debezium Connector for MySQL                                       [jar]
[INFO] Debezium Connector for PostgreSQL                                  [jar]
[INFO] Debezium Connector for MongoDB                                     [jar]
[INFO] Debezium Connector for MS SQL Server                               [jar]
[INFO] Debezium Connector for Oracle                                      [jar]
[INFO] Debezium Microbenchmark Tests                                      [jar]
[INFO] Debezium Oracle Connector Microbenchmark Tests                     [jar]
[INFO] Debezium Server Parent                                             [pom]
[INFO] Debezium Server Core                                               [jar]
[INFO] Debezium Server Amazon Kinesis Sink Adapter                        [jar]
[INFO] Debezium Server Google Cloud Pub/Sub Sink Adapter                  [jar]
[INFO] Debezium Server Pulsar Sink Adapter                                [jar]
[INFO] Debezium Server Azure Event Hubs Adapter                           [jar]
[INFO] Debezium Server HTTP Webhook Sink Adapter                          [jar]
[INFO] Debezium Storage Redis Module                                      [jar]
[INFO] Debezium Server Redis Sink Adapter                                 [jar]
[INFO] Debezium Server Distribution                                       [jar]
[INFO] Debezium Server Kafka Sink Adapter                                 [jar]
[INFO] Debezium Server Pravega Sink Adapter                               [jar]
[INFO] Debezium Server NATS Streaming Sink Adapter                        [jar]
[INFO] Debezium Server NATS JetStream Sink Adapter                        [jar]
[INFO] Debezium Storage S3 Module                                         [jar]
[INFO] Debezium Storage Tests Module                                      [jar]
[INFO]
[INFO] --------------------< io.debezium:debezium-storage >--------------------
[INFO] Building Debezium Storage Module 2.1.0-SNAPSHOT                   [1/27]
[INFO] --------------------------------[ pom ]---------------------------------
```

jpechane:
@blcksrx The PR should run the tests and also build zip/tar.gz.

As the final step please build the module locally and take our tutorial. Drop the storage into the mysql directory and configure the plguin to use S3 so we can confirm it works as it should.

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/history/S3SchemaHistory.java
```diff
@@ -0,0 +1,291 @@
...
+        endpoint = URI.create(config.getString(ENDPOINT));
+
+        if (config.getString(ACCESS_KEY_ID).isEmpty() && config.getString(SECRET_ACCESS_KEY).isEmpty()) {
+            credentialsProvider = AnonymousCredentialsProvider.create();
+        }
+        else {
+            AwsCredentials credentials = AwsBasicCredentials.create(config.getString(ACCESS_KEY_ID), config.getString(SECRET_ACCESS_KEY));
+            credentialsProvider = StaticCredentialsProvider.create(credentials);
+        }
+
```
ismailsimsek, line 144: is it possible to use [DefaultCredentialsProvider](https://docs.aws.amazon.com/sdk-for-java/latest/developer-guide/credentials.html#credentials-chain) or allow usage of other credential providers like InstanceProfileCredentialsProvider ?

it will enable using issuance profile credentials. for security user might not want to generate access,secret key

jpechane: Reading the docs this sounds like a good idea. I can even imagine keeping those auth related config options as optional ones. If present then they'll be converted and set as system properties so they'll be picked-up by `DefaultCredentialsProvider` too. And all other ways how to configure authentication will still be available.

blcksrx: I just changed from `AnonymousCredentialsProvider` to `DefaultCredentialsProvider`. In the case that `access_key` and `private_key` the credentials would be `StaticCredentialsProvider`.
At the first thought, my Idea was to provide a field such as `s3.credentialsProvider.class` to explicitly get the `CredentialsProvider` from user but because of the AWS SDK Builder pattern, I didn't.

jpechane:
@blcksrx Applied, thanks! I just aded some config check and simplified the code a bit.

blcksrx:
@jpechane Thank you very much for all your help

PR #4007, opened 2022-10-30
DBZ-3642 CloudEventConverter can also work with source records

rkudryashov:
I started this to make combination of Outbox SMT and CloudEventConverter working. Maybe it also can be used for other use cases when the converter receives records after some SMT was applied.

This change works when a payload contains just data of a "business-event" along with type of an event: the new test uses such JSON: `"{\"email\": \"gh@mefi.in\", \"type\": \"SomeType\"}",`

It is not the final version; it is needed:

- [ ] to pass in runtime a record instead of `record.value()` to `CloudEventsConverters.fromConnectData`
- [x] ~~to get rid of `data->payload` nesting~~ Probably it is not required because final `CloudEvent`.`data` contains `payload` object along with `schema` object
- [x] to change `CloudEventsMaker.ceId` implementation not to generate id but get it from a record's header
- [ ] to include `SourceRecordCloudEventsProvider` not only to Postgres connector


rkudryashov:
@Naros 

done

PR #4008, opened 2022-10-31
DBZ-5770: Update documentation regarding the Binary Data Converter

enzo-cappa:
Updated all references from `ByteBufferConverter` to `BinaryDataConverter`.

Also added special note about changing the settings before upgrading when running in Kafka Connect (otherwise Kakfa Connect will fail to start).


jpechane:
@enzo-cappa Applied, thanks a lot!

PR #4010, opened 2022-11-01
Run tests with Apicurio

vjuranek:
Run tests with Apicurio registry. Create separate profile which, if activated, starts Apicurio registry container. If the Apicurio registry is available, use it in the tests instead of in-memory Confluent registry. Otherwise fall back to Confluent registry.

https://issues.redhat.com/browse/DBZ-2131

### DIFF
--- a/debezium-connector-mongodb/pom.xml
+++ b/debezium-connector-mongodb/pom.xml
@@ -363,6 +363,19 @@
                     <name>apicurio</name>
                 </property>
             </activation>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-failsafe-plugin</artifactId>
+                        <configuration>
+                            <systemPropertyVariables>
+                                <use.apicurio>true</use.apicurio>
+                            </systemPropertyVariables>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
             <properties>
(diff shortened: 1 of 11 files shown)


vjuranek:
Tests failures are unrelated.

jpechane:
@vjuranek LGTM, thanks! Few questions though

- [ ] Could you please rebase on the latest main? There is a small conflict now.
- [ ] Could you please prepare additional PRs for other connectors - Db2, Vitess, Cassandra?
- [ ] I am realy not sure I like the idea of detecting Apicurio based on running port. What if there is misconfiguration or registry fails to start? Then we will run the tests against unwanted registry.

vjuranek:
Rebased and for checking if use Apicurio or not i used system property which is set in `apicurio` profile. Test failure is not related.
Related PRs:
* https://github.com/debezium/debezium-connector-db2/pull/81
* https://github.com/debezium/debezium-connector-vitess/pull/105
For Cassandra it's not needed as we don't do any registry checks in the tests (which is fortunate, as adding it into Cassanda won't be very easy as Cassandra tests don't use surefire maven pluing:-)

jpechane:
@vjuranek Applied, thanks a lot! Could you please reconfigure jobs on our internal CI to use Apicurio profile?

vjuranek:
@jpechane done in https://github.com/debezium/debezium/pull/4049

jpechane:
@vjuranek Applied, thanks!

@obabec FYI - just in case jobs will start failing.

PR #4014, opened 2022-11-02
DBZ-5783 Suppress logging of undetermined optionality for explicitly excluded columns

harveyyue:
https://issues.redhat.com/browse/DBZ-5783


jpechane:
@harveyyue Thanks for the PR. I believe the fix should be done in `io.debezium.connector.postgresql.connection.pgoutput.PgOutputMessageDecoder.handleRelationMessage(ByteBuffer, TypeRegistry)` directly in logging condition. We are doing out-of-bands reads so we might end up with column not read. Yet I believe it makes sense not to obtain metadata for excluded columns. IMHO the logging condition check should check if `columnName` matches the column filter and print the message only if it is captured.

Naros (approved):
LGTM @harveyyue.  Will wait on CI and we'll merge.  Thanks!

Naros:
Checked the test failures, they're unrelated.

Naros:
Applied to main (=2.1).

PR #4017, opened 2022-11-02
Add snapshot mode ALWAYS into Oracle connector

vjuranek:

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -729,6 +729,11 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
     public enum SnapshotMode implements EnumeratedValue {
 
         /**
+         * Performs a snapshot of data and schema upon each connector start.
+         */
+        ALWAYS("always", true, true, false),
+
+        /**
          * Perform a snapshot of data and schema upon initial startup of a connector.
          */
         INITIAL("initial", true, true, false),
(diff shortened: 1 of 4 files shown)


jpechane:
@vjuranek Applied, thanks!

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -728,6 +728,11 @@ public static IntervalHandlingMode parse(String value, String defaultValue) {
      */
     public enum SnapshotMode implements EnumeratedValue {
 
+        /**
+         * Performs a snapshot of data and schema upon each connector start.
+         */
+        ALWAYS("always", true, true, false),
```
jpechane, line 734: @vjuranek Thinking about that, should not the snapshot be executed also in case of schema history error? The snapshot should be executed under all condiitons.

vjuranek: @jpechane my understanding is that `ALYWAS` should be the same as `INITIAL`, but run on each connector start, therefore I use same as for `INITIAL`. Of course do know if I get it right and @Naros meant it in [DBZ-5626](https://issues.redhat.com/browse/DBZ-5626) in this way. @Naros , any thoughts?

Naros: Hi @vjuranek I agree with @jpechane here, this ultimately should mirror MySQL's _always_ implementation or any other connector that currently supports _always_.

vjuranek: fixed in https://github.com/debezium/debezium/pull/4022

PR #4021, opened 2022-11-03
DBZ-5790 Enhance the ability to sanitize topic name

harveyyue:
https://issues.redhat.com/browse/DBZ-5790


jpechane:
@harveyyue Applied, thanks!

PR #4022, opened 2022-11-03
DBZ-5626 Do the snapshot also in case of schema history errors

vjuranek:
`ALWAYS` should mean always and also this should be in line with Postgres `ALWAYS` option.


jpechane:
@vjuranek Applied, thanks!

PR #4026, opened 2022-11-04
DBZ-5794 Run database.initial.statements when creating and updating t…

aheuermann:
…he postgres publication

https://issues.redhat.com/browse/DBZ-5794


jpechane:
@aheuermann Hi, thanks for the PR. This is a bit sensitive area as we had issues on RDS with having an orphaned query. Before merging please make sure in RDS montioring that the behaviour is the same before and after the applied change.

aheuermann:
Closing this for now, should have time to get back to this in a few weeks in which case I'll reopen.

PR #4027, opened 2022-11-05
DBZ-4669 Remove duplicate plug-in version properties from debezium-parent POM

dude0001:
The removed plug-in version properties are duplicates of the same properties in parent POM debezium-build-parent and should not be needed.

There are some plug-in version properties still in debezium-parent that are duplicated by name from debezium-build-parent, but differ in version. I wasn't comfortable removing those without further direction as it would change the version of the plug-in in debezium-parent. I also checked all other POMs that are parented to debezium-build-parent and did not find any other duplicates like this. Local `./mvnw clean verify -DskipITs` build is passing. I'm having issues running integration tests locally and still troubleshooting that.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -262,8 +262,9 @@ Marc Zbyszynski
 Marek Winkler
 Mario Mueller
 Mariusz Strzelecki
-Mark Drilling
 Mark Bereznitsky
+Mark Drilling
+Mark Lambert
 Martin Medek
 Martin Sillence
 Matt Beary
--- a/debezium-parent/pom.xml
+++ b/debezium-parent/pom.xml
@@ -16,14 +16,9 @@
     <properties>
 
         <!-- Maven Plugins -->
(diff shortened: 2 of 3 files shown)


dude0001:
There are some plug-in version properties still in debezium-parent that are duplicated by name from debezium-build-parent, but differ in version. I wasn't comfortable removing those without further direction as it would change the version of the plug-in in debezium-parent. I also checked all other POMs that are parented to debezium-build-parent and did not find any other duplicates like this. Local `./mvnw clean verify -DskipITs` build is passing. I'm having issues running integration tests locally and still troubleshooting that.

jpechane:
@dude0001 Warning, I've rebased on the latest main and force-pushed to bypass the CI Postgres test failure

jpechane:
@dude0001 Applied, thanks! COuld you please prepare an additional PR that would get rid of all remaining duplicates - try to use the newer version of plugins so we can see the result.

PR #4029, opened 2022-11-06
DBZ-5800 ReplicationConnectionIT test fails

harveyyue:
https://issues.redhat.com/browse/DBZ-5800


harveyyue:
@Naros I'm not sure this fix is appropriate since this it is work fine before.

jpechane (changes requested):
I don't think this is the correct solution. Looking at the comment it is expected that the first `UPDATE` is for `REPLICA IDENTITY DEFAULT`.

We should probably step back, see what messages were emitted before the test started to fail and compare it with what it emits now.

jpechane:
This is not a proper fix and the issue exhibits only on CI. To unblock CI the test is disabled for dome time via https://github.com/debezium/debezium/pull/4034

PR #4032, opened 2022-11-06
DBZ-5796 For outbox transformation, when 'table.expand.json.payload' is set to true null values are not correctly deserialized

harveyyue:
https://issues.redhat.com/browse/DBZ-5796


jpechane:
@harveyyue Warning, I've rebased on the latest main and force-pushed to bypass the CI Postgres test failure

jpechane:
@harveyyue Thanks for the PR! I think the solution is fine but IMHO this is something that should be controlled by a config option. In certain appliactions it might be undesirable to have field to change its schema during runtime. So the default will be the behaviour as is and the config option would enable the behaviour in the PR.

jpechane:
@harveyyue Applied, thanks!

PR #4038, opened 2022-11-08
DBZ-5626 Don't check ofsset for AWLAYS mode

vjuranek:
While `ALWAYS` mode allows also snapshot on schema error, it allows snapshot also on other ocassions. Don't throw exception when offset is missing and schema is set to `ALWAYS`.


Naros:
Thanks @vjuranek, applied to main (=2.1)

PR #4039, opened 2022-11-08
DBZ-5785 Modify ocp system tests to archive test results and logs

MartinMedek:
Changed OCP testsuite job to pod. 
Added readiness probe to the pod description to indicate testsuite is finished. 
Modified the testsuite script to save logs and trigger the readiness probe once finished. 
Modified the OCP testsuite jenkins job to wait for the readiness probe and copy the testsuite logs.

### DIFF
--- a/jenkins-jobs/docker/debezium-testing-system/testsuite.sh
+++ b/jenkins-jobs/docker/debezium-testing-system/testsuite.sh
@@ -1,40 +1,55 @@
 #!/bin/bash
+{
+  source /root/.sdkman/bin/sdkman-init.sh
 
-source /root/.sdkman/bin/sdkman-init.sh
+  set -x
 
-set -x
+  DEBEZIUM_LOCATION="/testsuite/debezium"
 
-DEBEZIUM_LOCATION="/testsuite/debezium"
+  # copy secret to debezium project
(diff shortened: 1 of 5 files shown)


jenkins-jobs/docker/debezium-testing-system/testsuite.sh
```diff
@@ -1,40 +1,55 @@
...
+                      -Dimage.as="${DBZ_ARTIFACT_SERVER_IMAGE}" \
+                      -Das.apicurio.version="${DBZ_APICURIO_VERSION}" \
+                      -Dgroups="${DBZ_GROUPS_ARG}" \
+                      "${OPTIONAL_ARGS[@]}"
+
+  pushd debezium-testing/debezium-testing-system/target/failsafe-reports || exit 1
+  zip artifacts ./*.xml
+  mv  artifacts.zip /testsuite
+  popd || exit 1
+} 2>&1 | tee /tmp/testsuite_log
```
obabec, line 46: Please get rid of all paths that start in root /... This job might run in a container in the future, and these paths would be a problem.

obabec (changes requested):
Get rid of all absolute paths ideally. Everything should be created under user dir and not roots.

obabec (approved):
LGTM

PR #4041, opened 2022-11-09
DBZ-5791 Make ORA-01089 retriable based on its message

vjuranek:
https://issues.redhat.com/browse/DBZ-5791


Naros (approved):
LGTM, thanks @vjuranek

PR #4043, opened 2022-11-09
DBZ-5816 Introduce FAQ section; document ORA-01555

Naros:
https://issues.redhat.com/browse/DBZ-5816

As we had discussed regarding the internal FAQ document, I added those to this FAQ section and merged the previous "When things go wrong" section into the FAQ as well.  I'm not sure whether the formatting is acceptable for everyone, but it's a start.


jpechane:
@Naros Applied, this is super-useful!

PR #4044, opened 2022-11-09
DBZ-4669 Use the newer version of plugins to consolidate remaining duplicates plugin versions

dude0001:
Additional consolidation duplicate plugin versions properties per request of @jpechane in #4027 (https://github.com/debezium/debezium/pull/4027#issuecomment-1307223030).

### DIFF
--- a/debezium-api/pom.xml
+++ b/debezium-api/pom.xml
@@ -4,7 +4,7 @@
         <groupId>io.debezium</groupId>
         <artifactId>debezium-parent</artifactId>
         <version>2.1.0-SNAPSHOT</version>
-	<relativePath>../debezium-parent/pom.xml</relativePath>
+	    <relativePath>../debezium-parent/pom.xml</relativePath>
     </parent>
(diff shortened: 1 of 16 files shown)


jpechane (changes requested):
@dude0001 It seems that a reformatting is required. Could you please take a look and check if it is possible to configure formatter in a way that reformatting is not necessary.
If not possible then reformatting will be necessary and will/should be done hand-in-hand with https://github.com/debezium/debezium/pull/4045

dude0001:
@jpechane I was able to duplicate the formatting check errors on my other branch `DBZ-2439`, so I went ahead and applied them in PR #4045 since that is addressing the other formatting/style changes. I'm not clear if that is what you wanted, or if I should revert that change in `DBZ-2439` and add it here.

I also went ahead and rebased this branch. The PR is still failing for Checkstyle and Formatting, and would need rebasing again once #4045 is merged.

dude0001:
I have combined this PR with #4045

PR #4045, opened 2022-11-09
DBZ-2439 Apply Checkstyle check RedundantModifier

dude0001:


dude0001:
The issue called for "Remove redundant modifiers of members for interface fields" while the RedundantModifier check (https://checkstyle.sourceforge.io/config_modifier.html#RedundantModifier) validates several other types of modifiers by default e.g. constructors, resource variable declarations, enums, etc. We could apply the `tokens` property with only the tokens we wish to validate redundant modifiers on but applying the default "all" tokens seemed in the spirit of the issue request. Let me know if I should rework this to narrow the scope we want to apply the check, or if there is another check or formater I should consider to address the issue raised in DBZ-2439.

From the documentation linked, other examples of modifier usages that this check applies to

```
* Interface and annotation definitions.
* Final modifier on methods of final and anonymous classes.
* Type declarations nested under interfaces that are declared as public or static.
* Class constructors.
* Nested enum definitions that are declared as static.
* Record definitions that are declared as final and nested record definitions that are declared as static.
```

jpechane:
@dude0001 Thanks a lot! I am going to postpone the merging of this PR till late of 2.1 release cycle. The reason is that this change will probably complicate backporting to 2.0 branch. If we keep it till befor the release we will have to ackport from 2.2 to 2.1 and it will be on already reformatted code.

dude0001:
Rebased on latest and applied auto reformatting to debezium-core.

dude0001:
It looks like Vitess might be failing for a flakey test. I do not think I have permission to rerun the job.

dude0001:
Combined commits from #4044 (DBZ-4669) per https://github.com/debezium/debezium/pull/4044#issuecomment-1324584201 and rebased on latest.

debezium-api/pom.xml
```diff
@@ -4,7 +4,7 @@
         <groupId>io.debezium</groupId>
         <artifactId>debezium-parent</artifactId>
         <version>2.1.0-SNAPSHOT</version>
-	<relativePath>../debezium-parent/pom.xml</relativePath>
+        <relativePath>../debezium-parent/pom.xml</relativePath>
```
rk3rn3r, line 7: What happened here?

jpechane: @rk3rn3r tab vs spaces

debezium-assembly-descriptors/pom.xml
```diff
@@ -4,7 +4,7 @@
         <groupId>io.debezium</groupId>
         <artifactId>debezium-parent</artifactId>
         <version>2.1.0-SNAPSHOT</version>
-	<relativePath>../debezium-parent/pom.xml</relativePath>
+        <relativePath>../debezium-parent/pom.xml</relativePath>
```
rk3rn3r, line 7: ...and here?

debezium-bom/pom.xml
```diff
@@ -4,7 +4,7 @@
         <groupId>io.debezium</groupId>
         <artifactId>debezium-build-parent</artifactId>
         <version>2.1.0-SNAPSHOT</version>
-	<relativePath>../pom.xml</relativePath>
+        <relativePath>../pom.xml</relativePath>
```
rk3rn3r, line 7: and there?

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/StreamingSourceIT.java
```diff
@@ -578,10 +578,10 @@ private void inconsistentSchema(EventProcessingFailureHandlingMode mode) throws
         start(MySqlConnector.class, config, (success, message, error) -> exception.set(error));
 
         try (
-                final MySqlTestConnection db = MySqlTestConnection.forTestDatabase(DATABASE.getDatabaseName());
-                final JdbcConnection connection = db.connect();
-                final Connection jdbc = connection.connection();
-                final Statement statement = jdbc.createStatement()) {
+                MySqlTestConnection db = MySqlTestConnection.forTestDatabase(DATABASE.getDatabaseName());
```
rk3rn3r, line 581: Why is final obsolete here?

jpechane: I was unsure too so I checked in the past https://docs.oracle.com/javase/specs/jls/se8/html/jls-14.html#jls-14.20.3

> A variable declared in a resource specification is implicitly declared final if it is not explicitly declared final.

debezium-api/pom.xml
rk3rn3r: Oh, I think I tried to mark the text in front and it showed me whitespaces on both sides. Maybe I was somehow too fast... now I see tabs. Sorry! :hand_over_mouth:

PR #4046, opened 2022-11-09
DBZ-5822 Change visibility of BaseSourceTask#logStatistics method to …

adasari:
Changed method visibility from default to protected to implement DBZ-5822.


jpechane:
@adasari Applied, thanks

PR #4048, opened 2022-11-10
[docs] Updating default poll.intervall.ms in PG connector docs

gunnarmorling:
Hey @jpechane, just noticed this inconsistency in the docs. The actual default is [500ms indeed](https://github.com/debezium/debezium/blob/c781832f41b34461c601879f22e8b0089cce7067/debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java#L310).


Naros:
Applied to main (=2.1) and branches 2.0 and 1.9.  Branch 2.1 will be fast-forwarded when synchronizing with main during next 2.1 pre-release.

gunnarmorling:
Thanks, Chris!

PR #4054, opened 2022-11-10
DBZ-5825 Use Java 19 in Java Outreach action

Naros:
https://issues.redhat.com/browse/DBZ-5825

### DIFF
--- a/.github/workflows/jdk-outreach-workflow.yml
+++ b/.github/workflows/jdk-outreach-workflow.yml
@@ -14,7 +14,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        feature: [ 19 ]
+        feature: [ 20 ]
     name: SQL Server - Java ${{ matrix.feature }}
     steps:
       - uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


jpechane:
@Naros Why is this draft only?

jpechane:
Force pushed with

* update to the latest main
* test for Spanner connector
* Server build in separate repo

jpechane:
@Naros Looking at outrech page we should probably jump directly to Java 20 EA

jpechane:
The Spanner build failure tracked via https://issues.redhat.com/browse/DBZ-6065

Naros:
> @Naros Looking at outrech page we should probably jump directly to Java 20 EA

That's a good point.  I think I created this before my holiday PTO.  I've added this to my list for this week.

jpechane:
@Naros I think we are good to go with this one. Spanner is not working and Cassandra seems to be affected by the Cassandra lib code, so IMHO we can take the results and publish them. Do you know who is the current contact on our side? Still Gunnar?

Naros:
Most likely @jpechane, I'm not aware of there being anyone else.

PR #4056, opened 2022-11-10
DBZ-5770 Update documentation regarding the Binary Data Converter

enzo-cappa:
I introduced an error on the documentation by not closing a NOTE block. Fixing it.


jpechane:
@enzo-cappa Applied, thanks!

PR #4057, opened 2022-11-11
AbstractLogMinerEventProcessor DEBUG to TRACE loglevel

tooptoop4:
https://issues.redhat.com/browse/DBZ-5833

Current issue is that it logs tons and tons of log data at DEBUG level when knowing that other tables had changes is not helpful.
I need debug level on this class to get other useful data


jpechane:
@Naros This make sense to me, @tooptoop4 Could you please create a Jira issue for the change?

Naros:
I've logged https://issues.redhat.com/browse/DBZ-5833, I'll adjust the commit message.  Thanks @tooptoop4

Naros:
Applied to main (=2.1).

jpechane:
@tooptoop4 Could you please share your name so we can create record in COPYRIGHT file? Thanks

PR #4015, opened 2022-11-02
DBZ-4303 Replace md5 sum of connect extension in Dockerfile during release

jcechace:
https://issues.redhat.com/browse/DBZ-4303

Depends on: 
https://github.com/debezium/container-images/pull/308
https://github.com/debezium/debezium/pull/4060

PR #4060, opened 2022-11-14
DBZ-4303 Added assembly profile to debezium-connect-rest-extension

jcechace:
https://issues.redhat.com/browse/DBZ-4303

Related to: https://github.com/debezium/container-images/pull/308
Should be merged before: https://github.com/debezium/debezium/pull/4015

### DIFF
--- a/debezium-assembly-descriptors/src/main/resources/assemblies/rest-distribution.xml
+++ b/debezium-assembly-descriptors/src/main/resources/assemblies/rest-distribution.xml
@@ -1,0 +1,63 @@
+<?xml version="1.0"?>
+<assembly xmlns="http://maven.apache.org/plugins/maven-assembly-plugin/assembly/1.1.2" 
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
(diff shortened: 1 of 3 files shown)


jcechace:
@jpechane  that's actually something I wanted to ask about...  The core (and transitively api) modules are runtime dependencies so I assume those are right. SLF4J seems to be just a compile dependency of connect-runtime which has provided so I'm not sure if that one should be excluded. 

as for the location. I assume this behaves as a KC plugin and is placed in its own subdir under `$KAFKA_CONNECT_PLUGINS`. Or more precisely it is [linked there](https://github.com/debezium/container-images/pull/308/files#diff-cf7084030cee71c45a2a15695b26209d3ba0e0d595cb46341519049f1ad6079a)

jpechane:
slf4j should not be present as it is in Kafka `libs` directory. I'd ask you to consider dependency on Debezium core. There is only a single instance when it is really used and that's reading one property file. I am really unsure it makes sense to ahve it as a dependency.

jcechace:
@jpechane  I will remove the slf4j from assembly. As for the core module -- yes the dependency is used just for that single method call. Is it OK if I remove the dependency a copy that single method over to the extension module? This should lead to a tar containing just the extension jar (giving us compatibility with other plugins and possibility to add dependencies in future if we need them).

jpechane:
@jcechace Applied, thanks!

PR #4070, opened 2022-11-19
DBZ-5843 Invalid Java object for schema with type FLOAT64: class java.lang.Float

harveyyue:
https://issues.redhat.com/browse/DBZ-5843


Naros (approved):
LGTM, thanks @harveyyue

PR #4071, opened 2022-11-19
DBZ-4339 MongoDB connector to prefer secondaries

jcechace:
https://issues.redhat.com/browse/DBZ-4339

So far this PR does several things

- Generalises `MongoPrimary` to `MongoPreferredNode`
- Oplog positions are read from primary to get latest data (However this can easily be configurable as oplog[ can be read from secondary](https://www.mongodb.com/docs/manual/core/replica-set-oplog/)) 
- Collection names are read from primary as well 
- The actual data reading (event construction) in snapshot is performed against secondary
- Change stream is opened against secondary 

What is missing
- `MongoDbIncrementalSnapshotChangeEventSource` still uses primary 
-  When secondary which is being used to stream changes is promoted to primary the event stream is not recreated
 
This also needs to be tested somehow against a real replica set (what we have in tests is not sufficient).

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbSnapshotChangeEventSource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbSnapshotChangeEventSource.java
@@ -27,12 +27,13 @@ import org.bson.conversions.Bson;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.mongodb.ReadPreference;
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.MongoCursor;
 import com.mongodb.client.MongoDatabase;
 
 import io.debezium.connector.SnapshotRecord;
-import io.debezium.connector.mongodb.ConnectionContext.MongoPrimary;
+import io.debezium.connector.mongodb.ConnectionContext.MongoPreferredNode;
 import io.debezium.pipeline.ErrorHandler;
 import io.debezium.pipeline.EventDispatcher;
 import io.debezium.pipeline.EventDispatcher.SnapshotReceiver;
@@ -181,17 +182,17 @@ public class MongoDbSnapshotChangeEventSource extends AbstractSnapshotChangeEven
         final MongoDbOffsetContext offsetContext = (MongoDbOffsetContext) previousOffset;
         try {
             replicaSets.onEachReplicaSet(replicaSet -> {
-                MongoPrimary primary = null;
+                MongoPreferredNode mongo = null;
                 try {
-                    primary = establishConnectionToPrimary(partition, replicaSet);
+                    mongo = establishConnection(partition, replicaSet, ReadPreference.primaryPreferred());
                     final ReplicaSetOffsetContext rsOffsetContext = offsetContext.getReplicaSetOffsetContext(replicaSet);
-                    if (primary != null && isSnapshotExpected(primary, rsOffsetContext)) {
+                    if (mongo != null && isSnapshotExpected(mongo, rsOffsetContext)) {
                         replicaSetSnapshots.add(replicaSet);
                     }
                 }
                 finally {
-                    if (primary != null) {
-                        primary.stop();
+                    if (mongo != null) {
+                        mongo.stop();
                     }
                 }
             });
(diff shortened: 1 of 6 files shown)


jcechace:
Regarding the relocation of change streams... On one side it seems as a useful feature, on the other I'm not sure if this should be connectors responsibility.

An alternative approach could be to offer an option to run the change stream against a specifically tagged server. It would then be the responsibility of dba to configure this node with sufficiently low priority so that it is not elected primary.
This could be also implemented by choosing the secondary with the lowest priority (however that means querying `local.system.replse``).

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -299,16 +300,16 @@ protected MongoDbOffsetContext initializeOffsets(MongoDbConnectorConfig connecto
         final Map<ReplicaSet, BsonDocument> positions = new LinkedHashMap<>();
         replicaSets.onEachReplicaSet(replicaSet -> {
             LOGGER.info("Determine Snapshot Offset for replica-set {}", replicaSet.replicaSetName());
-            MongoPrimary primaryClient = establishConnectionToPrimary(partition, replicaSet);
-            if (primaryClient != null) {
+            MongoPreferredNode client = establishConnectionToPreferred(partition, replicaSet, ReadPreference.primaryPreferred());
```
btiernay, line 303: Should we be calling such references "node" or "client"? I think we are reserving "client" for MongoClient elsewhere.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/MongoUtilIT.java
```diff
@@ -23,7 +24,7 @@ public void testGetPrimaryAddress() {
                 .build());
 
         primary.execute("shouldConnect", mongo -> {
-            ServerAddress primaryAddress = MongoUtil.getPrimaryAddress(mongo);
+            ServerAddress primaryAddress = MongoUtil.getPreferredAddress(mongo, ReadPreference.primary());
```
btiernay, line 27: Preferred address?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbSnapshotChangeEventSource.java
```diff
@@ -377,7 +378,7 @@ private void createDataEventsForReplicaSet(ChangeEventSourceContext sourceContex
                                     snapshotReceiver,
                                     replicaSet,
                                     id,
-                                    primaryClient);
+                                    nodeClient);
```
btiernay, line 381: Maybe just "client" like elsewhere and reserve "node" for PreferredNode / PrimaryNode?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
jcechace: Yea the naming here is a bit flaky.. However it is a wrapper client essentially. I will try to go through the names one more time and put some consistency in.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/MongoUtilIT.java
jcechace: Returns an address of node adhering to given Read Preference type. I'm open to better name  suggestions:)

btiernay: Apologies, was referring to the variable name. Method name is clear and consistent.

jcechace: Test is working with primary connection and the address retrieved here is indeed a primary address. I don't see a problem problem here

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
jcechace: Variables/parameters/attributes of ``MongoPreferredNode`` type are now named simply ``mongo`` whereas the actual clients (e.g.  lambda parameters of ``MongoPreferredNode#execute``) are named ``client``.

jcechace:
``MongoDbIncrementalSnapshotChangeEventSource`` still uses strictly primary as it also needs a write access. ``MongoDbIncrementalSnapshotChangeEventSource`` holds a direct connection to single server in the related replica set thus the ``ReadPreference`` actually dictates the type of the server not only what server is selected for reads. 

I see two possible solutions to this 

- hold two connections (primary and secondary), using primary for signal writes and secondary for reads (this approach is likely fastest to implement)
- Rework the way we handle connections to actually connect to the replica set, properly using the capability of ``MongoClient`` to route requests based on type so that when using ``ReadPreference.secondaryPreferred()`` reads are routed to secondary (if available) and writes still to primary.  This is likely the correct approach, however this will require a large overhaul of how we handle connections the entire connector.

jcechace:
@Naros  @jpechane  Please have a quick look to asses if this is going in the right direction.

btiernay:
> `MongoDbIncrementalSnapshotChangeEventSource` still uses strictly primary as it also needs a write access. `

Something to consider is that the client supplied read preference is never consulted by the driver when doing writes (it is a read preference after all). This is why the integration tests still work when using a connection string which has a secondary node specified.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/MongoUtilIT.java
btiernay: Wow, 0/2, I’m not batting very high here 😅. I missed the fact that the primary was requested as it was off screen.

btiernay:
> Regarding the relocation of change streams. On one side it seems as a useful feature, on the other I'm not sure if this should be connectors responsibility.

While the monitoring feature does seem like just a nice to have, in practice it’s actually quite important. The reason is that it’s hard to detect this from outside of Debezium and trying to actuate any changes externally requires a lot of monitoring and tooling overhead.

jcechace:
> > Regarding the relocation of change streams. On one side it seems as a useful feature, on the other I'm not sure if this should be connectors responsibility.
> 
> While the monitoring feature does seem like just a nice to have, in practice it’s actually quite important. The reason is that it’s hard to detect this from outside of Debezium and trying to actuate any changes externally requires a lot of monitoring and tooling overhead.

Yes, however I think the entire idea of "trying to actuate changes" is not the right approach. Hence suggesting the idea of supporting targeted read based on secondary tags. It would then be database  configuration responsibility to provide a node meant for CDC. However as stated, this is up for discussion and I'm only waiting for other opinions.

btiernay:
> Yes, however I think the entire idea of "trying to actuate changes" is not the right approach. Hence suggesting the idea of supporting targeted read based on secondary tags. It would then be database configuration responsibility to provide a node meant for CDC. However as stated, this is up for discussion and I'm only waiting for other opinions.

While I can see how the tags-based approach can help with some of the management, but I don’t see how it addresses the issue when reading from a cursor. One must periodically check that the current node is the “preferred” since a server address is cached upon receiving the first result set, never to be consulted again. Thus, internal detection and self-actuation is required for recovery.

btiernay:
> The reason why integration tests work is that the database used in ITs is a single server replica -- meaning it only ever has a single primary with no secondaries -- and the code is using `ReadPreference.secondaryPreferred()` in order to use `primary` in case when no `secondary` is available.

This is not accurate for my case as I was testing with a connection string that specifies a secondary read preference against a 3 node MongoDB Atlas cluster. The drop connection in the test setup will always go to a primary. You can validate this easily with a debugger and see where the driver will force this.

jcechace:
> > The reason why integration tests work is that the database used in ITs is a single server replica -- meaning it only ever has a single primary with no secondaries -- and the code is using `ReadPreference.secondaryPreferred()` in order to use `primary` in case when no `secondary` is available.
> 
> This is not accurate for my case as I was testing with a connection string that specifies a secondary read preference against a 3 node MongoDB Atlas cluster. The drop collection command in the test setup will always go to a primary. You can validate this easily with a debugger and see where the driver will force this.
> 
> > BTW: this comment would be completely valid if the connection was established against RS as a whole with ReadPreference specified. That is not currently the case
> 
> This was the case for the test I was describing above.

If you look closely the connection string is only ever used to read the cluster topology (look what happens inside `ReplicaSetsDiscovery`) from that point onwards what I described applies.

jpechane:
@jcechace Applied, thanks a lot!

btiernay:
Great work on this @jcechace, I know you did much toiling and futzing to get this to work and I really appreciate all your effort. 💯 🙇 ❤️

jcechace:
@btiernay  you are welcome. Give it a test run and report any issues found please :).

btiernay:
FYI: Created https://issues.redhat.com/browse/DBZ-5953 to address 1 of the 2 lingering items as mentioned in the PR description (i.e. adding support for honouring MongoDB read preference in change stream after promotion)

PR #4073, opened 2022-11-21
DBZ-5845 Added fallback source for offsets retrieval

kgoodies:
Added fallback mechanism to retrieve offsets for partition in case of planned change

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerPartition.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerPartition.java
@@ -24,7 +24,6 @@ public class SqlServerPartition extends AbstractPartition implements Partition {
     private final Map<String, String> fallbackPartition;
     private final int hashCode;
 
-
     public SqlServerPartition(String serverName, String databaseName) {
         this(serverName, databaseName, true);
 
@@ -37,7 +36,7 @@ public class SqlServerPartition extends AbstractPartition implements Partition {
 
         this.sourcePartition = Collect.hashMapOf(SERVER_PARTITION_KEY, serverName, DATABASE_PARTITION_KEY, databaseName);
 
-        //backward compatibility needs to be enhanced for single-partition mode only
+        // backward compatibility needs to be enhanced for single-partition mode only
         this.fallbackPartition = multiPartitionMode ? null : Collect.hashMapOf(SERVER_PARTITION_KEY, serverName);
 
         this.hashCode = Objects.hash(serverName, databaseName);
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerPartition.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerPartition.java
@@ -20,13 +20,25 @@ public class SqlServerPartition extends AbstractPartition implements Partition {
 
     private final String serverName;
     private final Map<String, String> sourcePartition;
+
+    private final Map<String, String> fallbackPartition;
     private final int hashCode;
 
+
     public SqlServerPartition(String serverName, String databaseName) {
+        this(serverName, databaseName, true);
+
+    }
+
+    public SqlServerPartition(String serverName, String databaseName, boolean multiPartitionMode) {
+
         super(databaseName);
         this.serverName = serverName;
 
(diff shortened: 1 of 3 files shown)


kgoodies:
Hi,

I understand your concern, and was also taking that into consideration.

This is a solution for the generic problem related to backward
compatibility, hence it is defensible to put it in core/spi.
The problem occurred for SQL Server, but equally well could be for other
connectors in the future.
What is specific to SQL Server - specific old format of partition id, and
conditioning - is in SQLServerParition class.
This change in SQLServerParition is for particular upgrade requireements,
so it could be removed in ie Debezium 3.0, as per your BC policy.

There is another simple solution possible, and limited to SQL Server only:
if dbnames contains only 1 name - use the old format of partition id..
But it has a number of disadvantages that I discarded that option
- would not work as an upgrade, the connector would stick to the old format
of single-partition offsets forever.
- hence all future versions of Debezium would have to deal with that
alternate format
- the single-partition connector would have to remain single-partitioned
forever, the dbname*s* would have to contain a single dbname forever. - so
it apparently would be better to keep the old dbname config param.

Another alternative is to pre-read offsets in SQLServerPartition, and
return either the new or old format.
This is not that bad, but that is not ideal either - would require multiple
reads of the offsets and complicate offsets reading by spreading that into
multiple places.
I wasn't sure if you as the maintenainer would like that direction.

Another alternative is to introduce a temporary parameter for SQL Server,
like  upgradefrom19=true
but not sure if that fits your maintenance policy.

Maybe the most correct would be to re-introduce the old dbname parameter,
deprecate it properly, use for upgrade purposes, and remove only when BC is
no longer needed.
But that could be messy, and confusing, and still some changes in the core
would be needed

So from these alternatives I chose the one that seemed most
straightforward, did not introduce any config changes, and that could be
useful for future needs.

Kind Regards,
Krzysztof

jpechane:
@kgoodies Thanks for the feedback. I like the idea of
> There is another simple solution possible, and limited to SQL Server only:
if dbnames contains only 1 name - use the old format of partition id

Could this be used for reading offsets only but not for writing? So you'd try to read old format and if not present try reading in new format. Writing will be always in the new format.

I don't want to make larger API changes as I'd like to introduce in future versioned offsets. In that case it would be possible to have connector understand different version of offsets that connector produced in the past and would be truly universal solution.

kgoodies:
> Could this be used for reading offsets only but not for writing? So you'd
try to read old format and if not present try reading in new format.
Writing will be always in the new format.

That's basically what the current PR solution does, with only change that
it starts reading  from the new format, and falls back to the old one. That
is more correct.

> I'd like to introduce in future versioned offsets. In that case it would
be possible to have connector understand different version of offsets that
connector produced in the past and would be truly universal solution.

Yep, perfect example of forthcoming change in partition format
The PR solution is almost ready to support this. I can generalize it easily
to check multiple older versions, starting from current (i.e. 4.0) and then
trying previous versions (i.e. 3.0, 2.0, no-version)  etc.. until something
is found.
Should I modify PR and go directly to that?


.

kgoodies:
Committed to PR generalized mechanism with support for multiple formats,
that can be immediately used for versioning

It will possible to specify current format (that one will be used for
writing the offsets)

getSourcePartition() {
    return {"server":"db","database":"Db_Production", "version":"4.0"};
}

but also all formats that should be supported (to read the offsets), ie

getSupportedFormats(){
    return [
         {"server":"db","database":"Db_Production", "version":"4.0"},
         {"server":"db","database":"Db_Production", "version":"3.0"},
         {"server":"db","database":"Db_Production", "version":"2.0"},
         {"server":"db","database":"Db_Production"}];
}


Would you be happy with that generic solution?

The changed code is using streams - it was like that in the OffsetReader
before. However for better readability it can be changed to more
traditional style, if needed

kgoodies:
Note: Merge on github page produces commit message like `Merge branch 'main' into DBZ-5845`, but this is conflicting with Debezium commit message format check `Commit has no DBZ-xxx prefiix`

jpechane:
@kgoodies Could you please remove the merge commits? We prefer linear history. Whats for understanding.

cobolbaby:
After upgrading the debezium-connector-sqlserver, this disruptive change caused a large amount of duplicate data to be pulled. Do we currently have some scripting tools or compatibility code to circumvent this problem?

jpechane:
@kgoodies Nice work, applied!

PR #4074, opened 2022-11-21
DBZ-5855: IllegalStateException is thrown if task is recovering while other tasks are running

mikekamornikov:
https://issues.redhat.com/browse/DBZ-5855

- The PR only removes "end offset was modified" check as it doesn't fit multi task model and may cause issues. 
- The PR doesn't introduce / suggest any alternative


jpechane:
@mikekamornikov Thanks for the PR. It probably makes sense to not fail the connector. Still I'd prefer to keep there at least a log message INFO/WARN when this situation is detect just in case we missed a corner case. Then it would be useful to have logged that the sitation happened.

mikekamornikov:
> @mikekamornikov Thanks for the PR. It probably makes sense to not fail the connector. Still I'd prefer to keep there at least a log message INFO/WARN when this situation is detect just in case we missed a corner case. Then it would be useful to have logged that the sitation happened.

@jpechane I've changed it to `warn`.

jpechane:
@mikekamornikov Applied, thanks!

PR #4075, opened 2022-11-21
DBZ-5852:Snapshotter#snapshotCompleted is invoked regardless of snaps…

adasari:
Added two lifecycle methods `completed(...)` and `aborted(...)` to AbstractSnapshotChangeEventSource.java to perform any cleanup required specific to snapshot complete or aborted. 

Also, Override `completed` method in `PostgresSnapshotChangeEventSource.java` as complete is calling `snapshotter.snapshotCompleted()`

Common cleanup regardless of snapshot result should be done in `complete(...)` method (like in oracle or sql server connector)

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/source/AbstractSnapshotChangeEventSource.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/source/AbstractSnapshotChangeEventSource.java
@@ -33,7 +33,7 @@ import io.debezium.util.Threads;
  *
  * @author Chris Cranford
  */
-public abstract class AbstractSnapshotChangeEventSource<P extends Partition, O extends OffsetContext> implements SnapshotChangeEventSource<P, O> {
+public abstract class AbstractSnapshotChangeEventSource<P extends Partition, O extends OffsetContext> implements SnapshotChangeEventSource<P, O>, AutoCloseable {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AbstractSnapshotChangeEventSource.class);
 
@@ -86,14 +86,16 @@ public abstract class AbstractSnapshotChangeEventSource<P extends Partition, O e
         }
         finally {
             LOGGER.info("Snapshot - Final stage");
-            complete(ctx);
+            close();
 
             if (completedSuccessfully) {
                 LOGGER.info("Snapshot completed");
+                completed(ctx);
                 snapshotProgressListener.snapshotCompleted(partition);
             }
             else {
                 LOGGER.warn("Snapshot was not completed successfully, it will be re-executed upon connector restart");
+                aborted(ctx);
                 snapshotProgressListener.snapshotAborted(partition);
             }
         }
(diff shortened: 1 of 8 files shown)


jpechane (changes requested):
@adasari Hi, thanks for the PR. Could you please
1) Squash commits to a single one
2) Run the build locally at tyour computer to make sure the code is properly formatted
3) Make sure `DBZ-5852` prefix is present at the commit message
4) Are you sure that the `aborted` method is called?
5) Would it be possible to add a test?

Thanks a lot

adasari:
@jpechane I am working on writing unit tests. is there a way to interrupt or fail the snapshot programmatically ?

adasari:
> 
@jpechane  - will squash all commit before the merge and will make sure DBZ-5852 prefix is present in final commit message.

I believe `snapshotProgressListener.snapshotAborted(partition)` functionality is working in the current code base. So aborted method is called in the same failure flow. I am still understanding the other pieces of the code to write a unit/IT test.

adasari:
@jpechane Added IT test and all test cases are successful except DB2. DB2 testcase failures are not related to the PR changes.

debezium-core/src/main/java/io/debezium/pipeline/source/AbstractSnapshotChangeEventSource.java
```diff
@@ -169,6 +171,24 @@ protected abstract SnapshotResult<O> doExecute(ChangeEventSourceContext context,
     protected void complete(SnapshotContext<P, O> snapshotContext) {
```
jpechane, line 171: What is the difference between `complete` and `completed` methods?

adasari: `complete` method is intended take actions regardless of snapshot result. `completed` method is intended for taking actions after snapshot is successful. 
In a way, `complete` is redundant. I didn't touch that method since it's used in both Oracle and SQL connectors and it's ok to use it to take the same actions in both snapshot successful and failed cases.

adasari:
Ping.

debezium-core/src/main/java/io/debezium/pipeline/source/AbstractSnapshotChangeEventSource.java
jpechane: @adasari I am sorry for the delay, I commented on this one already but probably forgot to hit the send button :-(.

Generally I understand what you are saying but having two methods one named complete and the other completed is very/very confusing. Could you please think about better naming for the trio of complete/completed/aborted so it is clear cut which serves to what so there is no confusion?

adasari: I think it would be a good idea to replace complete with completed/aborted to avoid confusion. thoughts ?

jpechane: @adasari How about adding `AutoCloseable` interface and change the `complete` to `close` as it should do the resource cleanup - according to the JavaDoc. So there will be completed/aborted to act upon snapshot result and close for resource cleanup?

adasari: Sounds good. will update the code. thanks.

adasari:
Added AutoCloseable implementation changes. Failed tests are not related to the PR. Could you please review ? thanks.

adasari:
ping @jpechane . any feedback is appreciated. thanks.

jpechane:
@adasari Rebased to the latest main to make sure the tests passes. Db2 and Vitess PRs needed to get their code aligned to the new interface.

adasari:
> @adasari Rebased to the latest main to make sure the tests passes. Db2 and Vitess PRs needed to get their code aligned to the new interface.

DB2 PR : https://github.com/debezium/debezium-connector-db2/pull/92
Vitess PR: https://github.com/debezium/debezium-connector-vitess/pull/120

PR #4078, opened 2022-11-22
Remove Unreachable if statement

erdinctaskin:
if statement never will be true


jpechane:
@erdinctaskin Applied, thanks

PR #4080, opened 2022-11-23
DBZ-5758 Refactoring the quarkus-outbox extension to support reactive

ingmarfjolla:
https://issues.redhat.com/browse/DBZ-5758


jpechane:
@Naros Where do we stand with this PR now?

Naros (commented):
Hi @jpechane, can you give this a review after I've added my own set of changes?

jpechane:
@Naros Overall I am happy with the refactoring. I have one comment though. Is it really wise to use the same classname including package name in the same codebase? I can imagine it can be a problem for tooling. WDYT?

Naros:
Hi @jpechane, that's a fair point I hadn't considered.  In light of that, I have repackaged the reactive code into a package path of `io.debezium.outbox.reactive`, which also aligns with other reactive-based Quarkus extensions.

jpechane:
@Naros @ingmarfjolla Applied, thanks a lot! Could you please update outbox documentation too?

ingmarfjolla:
hey @jpechane thanks for merging. do you know how long it takes usually for the artifacts to be available? I tried to use the extension now but it couldn't find it

jpechane:
@ingmarfjolla They are already up - please see https://oss.sonatype.org/content/repositories/snapshots/io/debezium/debezium-quarkus-outbox-reactive-parent/

PR #4081, opened 2022-11-23
DBZ-5874 Making sure trace message gets logged

gunnarmorling:
With the separate `forEach` loop used for logging before, we'll only get to see the columns in the log if they all could be created. If something goes wrong before and an exception is raised, no information will be logged at all, also for those columns which could be created before.

https://issues.redhat.com/browse/DBZ-5874


jpechane:
@gunnarmorling Applied, thanks

PR #4082, opened 2022-11-24
DBZ-5743 Introduce a new unicode topic naming strategy

harveyyue:
https://issues.redhat.com/browse/DBZ-5743


harveyyue:
@jpechane Could you please help to reivew this pr?

debezium-api/src/main/java/io/debezium/spi/topic/TopicNamingStrategy.java
```diff
@@ -43,17 +43,17 @@
      *
      * @link https://github.com/apache/kafka/blob/trunk/clients/src/main/java/org/apache/kafka/common/internals/Topic.java
      */
-    default String sanitizedTopicName(String topicName) {
+    default String sanitizedTopicName(String topicName, ReplacementFunction replacement) {
```
jpechane, line 46: I don't think we should change the signature and add the replacement function here.

IMHO it is about time to move `sanitizedTopicName` implementation into `AbstractTopicNamingStrategy` and pass replacement function via factory method or constructor.

debezium-core/src/main/java/io/debezium/schema/AbstractUnicodeTopicNamingStrategy.java
```diff
@@ -0,0 +1,45 @@
...
+        super(props);
+    }
+
+    @Override
+    public String sanitizedTopicName(String topicName) {
+        return sanitizedTopicName(topicName, UNICODE_REPLACEMENT);
+    }
+
+    @Override
+    public boolean isValidKafkaTopicCharacter(char c) {
```
jpechane, line 42: The function name is not clear. What is valid Kafka topic character is defined by Kafka itself and we reduce the set here.

debezium-core/src/main/java/io/debezium/schema/SchemaRegexTopicNamingStrategy.java
```diff
@@ -19,12 +20,23 @@
 @Incubating
 public class SchemaRegexTopicNamingStrategy extends AbstractRegexTopicNamingStrategy {
 
+    private final boolean multiPartitionMode;
+
     public SchemaRegexTopicNamingStrategy(Properties props) {
         super(props);
+        this.multiPartitionMode = props.get(CommonConnectorConfig.MULTI_PARTITION_MODE) == null ? false
```
jpechane, line 27: This should go to a common parent.

jpechane (changes requested):
@harveyyue Thanks a lot, that's great start! I've left few comments related to code structure.
Also could you please provide an alternative implementation of `io.debezium.util.SchemaNameAdjuster.AVRO` like `io.debezium.util.SchemaNameAdjuster.AVRO_UNICODE` that would follow the similar rules jst with Avro limitations?

When the code is completed the docs change will be needed too.

harveyyue:
@jpechane Done. Will add `io.debezium.util.SchemaNameAdjuster.AVRO_UNICODE`, also update to doc.

jpechane (approved):
@harveyyue Nice, I think we are good to go! I'll wait with merge for the schema adjustment as those two changes would ideally go hand-in-hand.

harveyyue:
@jpechane I notice that the class `FieldNameSelector` also occur the same replacement `_` issue in ticket DBZ-5743. Should we apply avro_unicode mode to it?

harveyyue:
It's seems that is irrelevant with RecordsStreamProducerIT failed.
I have created a jira issue to track it.
https://issues.redhat.com/browse/DBZ-5895

jpechane:
@harveyyue Good point! Yes please, that's unfortunate by-product of non-centrilizing some of the code.

harveyyue:
@jpechane Ok, I'm working on refactor these adjustment name codes.

jpechane:
@harveyyue Rebased on the latest main to have tests passing

harveyyue:
@jpechane Jiri, I do some code changes as below, I believe that the schema name adjuster should be improved. Feel free to review these changes.
1. Expose the `ReplacementFunction` interface to api mode, and add several implementations for avro and avro_unicode mode. 
2. Adding the no_op/avro/ avro_unicode modes to `FieldNameSelector`, I had to add a new property`field.name.adjustment.mode` instead of `sanitize.field.names`, but it looks like a duplicate property with `schema.name.adjustment.mode`.  Do we consider to merge them, and rename SchemaNameAdjuster to a appropriate name.

harveyyue:
@jpechane Jiri, I do some code changes as below, I believe that the schema name adjuster should be improved. Feel free to review these changes.
1. Expose the `ReplacementFunction` interface to api mode, and add several implementations for avro and avro_unicode mode. 
2. Adding the no_op/avro/ avro_unicode modes to `FieldNameSelector`, I had to add a new property`field.name.adjustment.mode` instead of `sanitize.field.names`, but it looks like a duplicate property with `schema.name.adjustment.mode`.  Do we consider to merge them, and rename SchemaNameAdjuster to a appropriate name.

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -275,10 +280,25 @@ public String getValue() {
...
+                case AVRO:
+                    return SchemaNameAdjuster.AVRO;
+                case AVRO_UNICODE:
+                    return SchemaNameAdjuster.AVRO_UNICODE;
+                default:
+                    return SchemaNameAdjuster.NO_OP;
+            }
+        }
+
+        public SchemaNameAdjuster createFieldNameAdjuster() {
```
jpechane, line 293: @harveyyue 
Should not this return just `FieldNamer`? It seems to me now that we are mixing two things together in the `SchemaNameAdjuster`. This is not wrond as such but the name and JavaDoc of the class implies soemthing different and in fact only half is always used.

What is IMHO correct is to have a `SchemaAdjuster`. The instance of it will be created in `CommonConnectorConfig` constructor form both options schema adjustment mode and fild name adjustment mode. And then there will be a single method `nameAdjuster()` that will return this instance and be used both for schema and field name mapping? WDYT?

jpechane:
@harveyyue OK, thinking about that I believe we are good to go even with current naming.
I've rebased and force-pushed the PR on the latest main and modified JavaDoc a bit to reflect current state of affairs.

Could you please

- [ ] provide a PR to other connectors (including Spanner, I need to add it to the CI here too)
- [ ] prove a doc for `field.name.adjustment.mode` config option?

harveyyue:
@jpechane Ok, I will complete it in next few days.

harveyyue:
Related pr:
Db2: https://github.com/debezium/debezium-connector-db2/pull/94
Cassandra: https://github.com/debezium/debezium-connector-cassandra/pull/103
Vitess: https://github.com/debezium/debezium-connector-vitess/pull/125
Spanner: https://github.com/debezium/debezium-connector-spanner/pull/16

harveyyue:
@jpechane Done, and remove the  property "sanitize.field.names" from all connector docs

jpechane:
@harveyyue Great job, thanks a lot! Applied.

PR #4084, opened 2022-11-25
DBZ-4339 Add MongoDB Testcontainers support for functional testing

btiernay:
## Description

[DBZ-4339](https://issues.redhat.com/browse/DBZ-5857): Adds MongoDB Testcontainers support to the `debezium-testing-testcontainers` module to aid in more comprehensive functional testing scenarios and conditions. Specifically, this PR adds support for testing both multi-node replica set clusters as well as full multi-shard clusters running config server and router tiers.

Much care was taken to support both Docker on Linux and Docker Desktop on Mac  / Windows. While Linux support is completely seamless, users running Docker Desktop will require additional work to configure their `/etc/hosts` to be able to interact with the running containers from a functional test that uses a MongoDB client driver:

```bash
# Unfortunately, the following is required so that multi-node replica sets can properly resolve both within cluster using
# the user-defined Docker bridge network and from the host. While it is possible to use IPs in the `rs.initialize`
# call, this is not possible on Docker Desktop for Mac as the default (`docker0`) and user-defined bridge networks 
# are not exposed. Thus, local testing on these systems requires adding these host names to the local `/etc/hosts.
for host in test-mongo{1..3} test-mongos{1..3} test-mongo-configdb{1..3} test-mongo-shard{1..3}-replica{1..3}; do 
  echo -e "127.0.0.1\t$host" | sudo tee -a /etc/hosts; 
done
```

The following versions of MongDB have been tested and pass via CI (while in `debezium-connector-mongodb`, prior to the move to `debezium-testing-testcontainers`) :
- 4.0
- 4.4
- 5.0
- 6.0

## Summary

Change set includes:
- `MongoDbContainer`: A single, configurable MongoDB container process (either `mongod` or `mongos`)
- `MongoDbReplicaSet`: A collection of `MongoDbNodes` containers
- `MongoDbShardedCluster`: A collection of data-bearing`MongoDbReplicaSet`s configured as shared, a config server `MongoDbReplicaSet` and one or more router nodes (`mongos`es)
- `MongoDbReplicaSetIT`: A simple functional test to assert on basic functionality of `MongoDbReplicaSet`
- `MongoDbShardedClusterIT`: A simple functional test to assert on basic functionality of `MongoDbShardedCluster`

**Note:** Please note that no `debezium-connector-mongodb`-specific functionality has been tested with these new facilities. This should be scheduled for follow-up work. (If removing `docker-maven-plugin`, consider using the "[singleton container](https://www.testcontainers.org/test_framework_integration/manual_lifecycle_control/#singleton-containers)" pattern for reuse across the test suite)

## Class Design

The basic structure of the Testcontainer classes and tests is shown below:
image

## References

While Testcontainers does have a [MongoDB module](https://www.testcontainers.org/modules/databases/mongodb/), it is fairly limited in what it currently supports: a single node replica set.

A much more robust, yet complex solution exists as described in [Run MongoDB Atlas locally for testing ](https://habr.com/en/post/544460/) and [silaev/mongodb-replica-set](https://github.com/silaev/mongodb-replica-set) repo on GitHub. What's interesting about this approach is that it has very elaborate support for chaos testing through Toxiproxi. While certainly useful for many needs, it doesn't have support for sharded clusters and thus is still limited in the Debezium context, which has a significant amount of functionality tied to the comings and goings of shard replica sets. Thus, this PR was born :).


.github/workflows/debezium-workflow.yml
```diff
@@ -299,6 +299,12 @@ jobs:
           restore-keys: |
             maven-debezium-test-build-${{ hashFiles('**/pom.xml') }}
 
+      - name: Add MongoDB test container hosts to /etc/hosts
+        run: >
+          for host in test-mongo{1..3} test-mongos{1..3} test-mongo-configdb{1..3} test-shard{1..3}-replica{1..3}; do 
+            echo -e "127.0.0.1\t$host" | sudo tee -a /etc/hosts; 
+          done
```
btiernay, line 306: Unfortunately, this is required so that multi-node replica sets can properly resolve. While it is possible to use IPs in the `rs.initialize` call, this is not possible on Docker Desktop for Mac as the `docker0` bridge network is not exposed. Note that this is not currently a problem for single node replica sets in CI / local using the `docker-maven-plugin` since `localhost` is valid in this setup across Mac and Linux. While there are other techniques that can be used to make this more transparent (e.g. [JVM DNS hijacking](https://github.com/CorfuDB/corfudb-cloud/blob/9e14d15a7119ef22767d0946d12faa9bfa541e30/universe/universe-core/src/main/java/org/corfudb/universe/infrastructure/docker/universe/FakeDns.java), this seemed the easiest and most consistent / reliable. However, for local testing it does require adding these host names to the local `/etc/hosts`.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/cluster/MongoDbShardedClusterIT.java
```diff
@@ -0,0 +1,60 @@
...
+                }
+
+                for (var doc : collection.find()) {
+                    logger.info("Doc: {}", doc);
+                }
+
+                logger.info("Connected to cluster: {}", client.getClusterDescription());
+                cluster.addShard();
+                logger.info("Connected to cluster: {}", client.getClusterDescription());
+                cluster.removeShard();
```
btiernay, line 55: Examples of how Testcontainers affords useful facilities for dynamically affecting the topology to test certain behaviors.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/cluster/MongoDbReplicaSetIT.java
```diff
@@ -0,0 +1,128 @@
...
+            // Wait until the cursor address doesn't match the read preference (primary)
+            await().atMost(30, SECONDS)
+                    .pollInterval(1, SECONDS)
+                    .until(() -> cluster.tryPrimary()
+                            .map(node -> !node.getAddress().equals(cursor.getServerAddress()))
+                            .orElse(false));
+
+            // Ensure it's invalid
+            if (!MongoDbNode.IMAGE_VERSION.equals("4.0")) {
+                assertThat(isSelectedReadPreference(client, collection, cursor)).isFalse();
```
btiernay, line 75: I'm not sure if this was a client behavior that has since been changed, but it acts differently compared to the later versions.

jcechace:
Pretty nice, I'm just thinking whether this should be part of mongodb module or if we should put this into the testcontainers module and then just use it from mongodb.

jcechace:
Actually ... thinking about it, this would be also useful to QE in cases when Debezium is running inside a container. Can we move this into `debezium-testing-testcontainers` module?

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/cluster/MongoDbShardedClusterIT.java
jcechace: Would be nice to turn this actually into tests which asserts taht the topology is correct instead of simply logging.

.github/workflows/debezium-workflow.yml
jcechace: It actually si a problem for the current container run by maven as it is initialised with IP :). But it should be possible to change.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/cluster/MongoDbShardedClusterIT.java
```diff
@@ -0,0 +1,60 @@
...
+            cluster.start();
+
+            String readPreference = "primary";
+            var connectionString = new ConnectionString(cluster.getConnectionString() + "/?readPreference=" + readPreference);
+            logger.info("Connecting to cluster: {}", connectionString);
+            try (var client = MongoClients.create(connectionString)) {
+                logger.info("Connected to cluster: {}", client.getClusterDescription());
+
+                var databaseName = "test";
+                cluster.enableSharding(databaseName); // Only needed in 5.0
```
jcechace, line 38: could the be perhaps made universal in regards to database version? Meaning calling this method would be a no-op on versions where explicit enablement is not required for a database.

btiernay:
> Actually ... thinking about it, this would be also useful to QE in cases when Debezium is running inside a container. Can we move this into `debezium-testing-testcontainers` module?

I was thinking about that as well, but was't sure what it would do to the classpath if used from `debezium-connector-mongodb`. I can certainly give that a go if that's the preference.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/cluster/MongoDbShardedClusterIT.java
btiernay: Turns out this is a no-op in 6.0. Do you think it's worth the effort to perform version checking to future proof against backwards incompatibility?

jcechace:
I was also thinking that it might be the best to have the option whether to use the bridge network IPs or hostnames in ``rs.initialize`` call configurable. The default would be IPs in order not to bother linux users with the pitfalls of Mac solution.

btiernay:
> I was also thinking that it might be the best to have the option whether to use the bridge network IPs or hostnames in ``rs.initialize`` call configurable. The default would be IPs in order not to bother linux users with the pitfalls of Mac solution. 

Yep, that's doable, but will complicate the implementation somewhat. I was heading down that path, but figured this was a good first iteration. Will take a stab at that if deemed valuable.

btiernay:
> I was also thinking that it might be the best to have the option whether to use the bridge network IPs or hostnames in `rs.initialize` call configurable. The default would be IPs in order not to bother linux users with the pitfalls of Mac solution.

Done. It tries to be smart about it and detect Docker Desktop and only uses names when it's found. On Linux, it will automatically use the bridge network IPs. Clean CI run on my fork:
image

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/cluster/MongoDbShardedClusterIT.java
btiernay: Done!

debezium-connector-mongodb/pom.xml
```diff
@@ -241,6 +242,7 @@
                         <connector.mongodb.hosts>${mongo.data.replicaset.name}/${docker.host.address}:27017</connector.mongodb.hosts>
                         <connector.mongodb.members.auto.discover>false</connector.mongodb.members.auto.discover>
                         <connector.mongodb.name>mongo1</connector.mongodb.name>
+                        <version.mongo.server>${version.mongo.server}</version.mongo.server>
```
btiernay, line 245: Going to leave this here for the benefit of future tests.

debezium-testing/debezium-testing-testcontainers/pom.xml
```diff
@@ -70,6 +70,11 @@
       <artifactId>postgresql</artifactId>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.mongodb</groupId>
+      <artifactId>mongodb-driver-sync</artifactId>
+      <scope>test</scope>
```
btiernay, line 76: For testing the test container infra only (no downstream dependency introduced in clients).

jcechace:
@jpechane  WDY think about this? IMHO this will be a much better replacement for our current mongo instance the connector tests. If you are OK with adding this to the module I would merge this.

.collections.json
```diff
@@ -0,0 +1,5 @@
+Error: couldn't connect to server 127.0.0.1:27017, connection attempt failed: SocketException: Error connecting to 127.0.0.1:27017 :: caused by :: Connection refused :
```
jpechane, line 1: Is this file intentional?

jpechane (commented):
LGTM, @jcechace I left one question, when ddressed this can be merged.

I've one general question regarding writign tests. Would it be necessary to write separate tests for replica set and sharded cluster or can there be a way that test is written only once and the MongoDB cluster is "injected" based on test configuration?

.collections.json
btiernay: Certainly not 😅. Will Remove.

jpechane:
@btiernay Applied, thanks a lot. I merged it to get it in the release.
@jcechace Could you still answer the question I asked? Thanks

jcechace:
@jpechane I'm not actually sure I understand the question.  It's definitively possible  for example to run the test with varius mongo versions. It's should be possible to write the tests in such a way that e.g. based on a property  a sharded cluster or single replica is used. It might be good to have a common interface over Node/ReplicaSet/Cluster which would be able to provide connection string. Right now I think that would have to be handled a differently -- still possible though.

jcechace:
Yes definitively, a common Interface such as `MongoDbInstance` with method `String getConnectionString` could be a small usability improvement.  However it wouldn't be a lot of a difference with just that method.  Mayhap also some lifecycle methods such as "stop" / "pause" could be added or whatever common functionality we need in future. Also it would allow for single tests attribute to be able to hold these...

jcechace:
@obabec  The ``DockerMongo`` in testsuite fixture could benefit from this.

PR #4086, opened 2022-11-25
Remove Unnecessary null checks

erdinctaskin:
Remove Unnecessary null checks


erdinctaskin:
Remove unnecessary null checks

PR #4088, opened 2022-11-28
DBZ-5875 Verify the correct MySQL privileges

MaoHJ-Official:
Verify 'RELOAD' privilegeinstead of 'LOCK TABLES' privilege


MaoHJ-Official:
https://issues.redhat.com/projects/DBZ/issues/DBZ-5875

jpechane:
@MaoHJ-Official Befor applyiing this, please let's go back to the Jira and find out this is a real issue. PLease share logs there as I state that `LOCK TABLES` is necessary if global locking fails.

PR #4095, opened 2022-11-29
DBZ-5739 Handle permission error from RDS Postgres

prburgu:
Certain versions of RDS Postgres (e.g. RDS Postgres 12.3) are returning the following error when using the `pg_replication_slot_advance()` even though the user has superuser or replication roles. Since we can't use the function on these versions, catch and ignore these errors when validating the replication slot. This function works on RDS Postgres 12.5 for the same user.

```
ERROR:  must be superuser or replication role to use replication slots 
```

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
@@ -369,7 +369,7 @@ public class PostgresReplicationConnection extends JdbcConnection implements Rep
                 LOGGER.info("Postgres server doesn't support the command pg_replication_slot_advance(). Not seeking to last known offset.");
             }
             else if (e.getMessage().matches("ERROR: must be superuser or replication role to use replication slots(.|\\n)*")) {
-                LOGGER.warn("Unable to use pg_replication_slot_advance() function. The Postgres server is likely on an old RDS version: " + e.getMessage());
+                LOGGER.warn("Unable to use pg_replication_slot_advance() function. The Postgres server is likely on an old RDS version", e);
             }
             else if (e.getMessage().matches("ERROR: cannot advance replication slot to.*")) {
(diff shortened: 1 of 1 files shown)


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
```diff
@@ -368,6 +368,9 @@ protected void validateSlotIsInExpectedState(WalPositionLocator walPosition) thr
             if (e.getMessage().matches("ERROR: function pg_replication_slot_advance.*does not exist(.|\\n)*")) {
                 LOGGER.info("Postgres server doesn't support the command pg_replication_slot_advance(). Not seeking to last known offset.");
             }
+            else if (e.getMessage().matches("ERROR: must be superuser or replication role to use replication slots(.|\\n)*")) {
+                LOGGER.warn("Unable to use pg_replication_slot_advance() function. The Postgres server is likely on an old RDS version: " + e.getMessage());
```
jpechane, line 372: Please pass exception as method argument to have properly logged stacktrace.

jpechane:
@prburgu Applied, thanks!

mitchbregs:
@prburgu @jpechane  This function was added in Postgres 11 (which we are using), but I am running into this exact issue and have no idea how to resolve.

prburgu:
@mitchbregs What's the exact version of Postgres 11? I believe this function is available from 11.9 onwards. Btw, this is just a warning and the error is ignored and connector should continue working.

mitchbregs:
@prburgu got it. We are using 11.6. The warning is understandable and error being ignored is fine, but does this function seek the replication slot?

According to docs on the function:

> The slot cannot be moved backwards or moved beyond the current insert location.
> Any WAL being retained due to the slot's previous position will be removed at the next checkpoint.
> An error is raised if the slot cannot be advanced. 

Will we just keep retaining WAL? We are seeing a non-stop climb in replication slot lag..

PR #4100, opened 2022-11-30
DBZ-5885 Update documentation for Debezium Server with Cassandra Connector

ggaborg:


jpechane:
@ggaborg Applied, thanks

PR #4102, opened 2022-11-30
DBZ-5739 Simplify slot validation logic

prburgu:
The previous logic relied on `pg_replication_slot_advance()` which was not supported in all the Postgres versions and certain RDS version. This PR simplifies the logic by comparing the last_flushed_lsn from the server to the last known offset. This logic works on all the Postgres versions.


jpechane (changes requested):
@prburgu As the ald issues has been closed and released a new Jira is necessary for this.
But there is one serious problem in this PR - LSN values in replication slot are not totally ordered. Replicatin slot guarantees you to have LSN in order in a single transaction and for all commit LSNs. But you cannot take two arbitrary LSNs and compare them as the order is not defined.

prburgu:
Got it. I don't have a better solution for this problem then. Closing this PR.

PR #4103, opened 2022-12-01
DBZ-5882 Fix Debezium scripting archive url in downstream deploy example

roldanbob:
[DBZ-5883](https://issues.redhat.com/browse/DBZ-5883)

Appends missing build number string to Maven URL for the Debezium scripting archive to YAML example in the  instructions for using Streams to deploy connectors.
Also implements new downstream attributes for assigning the  build number for the Debezium release and for the Apicurio registry release.

This change is targeted solely at the downstream product version of the documentation and has no effect on the community version of the documentation.

A separate change to implement this correction was merged to the downstream documentation repository.


jpechane:
@roldanbob Applied, thanks

PR #4104, opened 2022-12-01
DBZ-5890 Fix artefact server container

novotnyJiri:
https://issues.redhat.com/browse/DBZ-5890


jpechane:
@obabec ping, Could you please review?

PR #4111, opened 2022-12-05
DBZ-5857 Migrated MongoDb integration tests to test-containers

jcechace:
https://issues.redhat.com/browse/DBZ-5857

WIP -- the following needs to be addressed

- [x] Repeatable RS startup on Docker Desktop 
- [x] Oplog size settings when starting replica sets (does not seem to be required by the tests at the moment)
- [x] Ability to set readConcern majority on replica set (This is always true since mongo 5.0 and enabled by default prior, so nothing we need to tweak currently)
- [x] Investigate and fix failures in `debezium-testing-testcontainers`

The items above should be address as part of `debezium-testing/debezium-testing-testcontainres` modul rather than in mongodb. 

This PR should be processed before #4071  in order to ensure proper test coverage as current approach with single node replica set is not sufficient.

### DIFF
--- a/debezium-connector-mongodb/pom.xml
+++ b/debezium-connector-mongodb/pom.xml
@@ -115,6 +115,7 @@
         <apicurio.init.timeout>60000</apicurio.init.timeout> <!-- 60 seconds -->
         <!-- Mongo test containers -->
         <mongodb.docker.desktop.ports>27017,27018,27019</mongodb.docker.desktop.ports>
+        <mongodb.replica.size>1</mongodb.replica.size>
     </properties>
     <build>
         <plugins>
@@ -198,6 +199,7 @@
                         <!-- Make these available to the tests via system properties -->
                         <version.mongo.server>${version.mongo.server}</version.mongo.server>
                         <mongodb.docker.desktop.ports>${mongodb.docker.desktop.ports}</mongodb.docker.desktop.ports>
+                        <mongodb.replica.size>${mongodb.replica.size}</mongodb.replica.size>
                         <skipLongRunningTests>${skipLongRunningTests}</skipLongRunningTests>
                     </systemPropertyVariables>
                     <runOrder>${runOrder}</runOrder>
--- a/debezium-connector-mongodb/pom.xml
+++ b/debezium-connector-mongodb/pom.xml
@@ -54,6 +54,11 @@
             <scope>test</scope>
         </dependency>
         <dependency>
+            <groupId>io.debezium</groupId>
+            <artifactId>debezium-testing-testcontainers</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
             <groupId>ch.qos.logback</groupId>
             <artifactId>logback-classic</artifactId>
             <scope>test</scope>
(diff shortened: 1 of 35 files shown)


jcechace:
@jpechane  Regarding the tc runtime on Docker Desktop (mac, widnows). 

What is the problem: Because of the way how docker desktop handles port forwarding it is difficult to find a free port in advance. Currently employed approach with `ServerSocker(0)` is problematic as (besides the possible race condition) opening and closing the socket puts it into a TIME_WAIT state and this prevents docker desktop to bind to the port (SO_REUSEADDR has no effect here). 

The possible options to work around this are (mind you this is really only an issue when running the tests locally on Mac or Window):

- Simply generate random ports and hope for the best (as restarting the entire cascade of containers in non trivial)
- Since using a static set of ports works* we can introducing a property `mongo.container.ports` with the ability to provide either
  - comma separated list of usable ports
  - a port range. E.g. 62200:62220
  - an open port range with starting port E.g. 27017: (this would also be the default value)
  
* Since ports bound and then released by docker seems to be reusable just find, there is some (to me mysterious) difference in how docker binds ports and how opening/closing ports via `ServerSocket` handles the same.

debezium-connector-mongodb/pom.xml
```diff
@@ -238,13 +191,12 @@
...
                         <!-- Make these available to the tests via system properties -->
-                        <connector.mongodb.hosts>${mongo.data.replicaset.name}/${docker.host.address}:27017</connector.mongodb.hosts>
-                        <connector.mongodb.members.auto.discover>false</connector.mongodb.members.auto.discover>
-                        <connector.mongodb.name>mongo1</connector.mongodb.name>
                         <version.mongo.server>${version.mongo.server}</version.mongo.server>
+                        <mongodb.docker.desktop.ports>${mongodb.docker.desktop.ports}</mongodb.docker.desktop.ports>
                         <skipLongRunningTests>${skipLongRunningTests}</skipLongRunningTests>
                     </systemPropertyVariables>
                     <runOrder>${runOrder}</runOrder>
+                    <argLine>--add-opens java.base/java.net=ALL-UNNAMED</argLine>
```
jcechace, line 199: @jpechane this is required for `FakeDns` to work reliably on all JDK 11+ (actually 9+ but that doesn't concern us).  Alternatively `FakeDns` can be disabled using `docker.desktop.disable.fake.dns`property, however then users with Docker Desktop will have to modify `/etc/hosts` in such case our test containers will log a warning with instructions  (to the best extend).

debezium-connector-mongodb/pom.xml
```diff
@@ -85,25 +100,16 @@
...
         -->
         <docker.skip>false</docker.skip>
         <docker.exposeContainerInfo>docker.container</docker.exposeContainerInfo>
-        <docker.dbs>mongo:${version.mongo.server},debezium/mongo-initiator:${version.mongo.server}</docker.dbs>
-        <docker.filter>${docker.dbs}</docker.filter>
         <!-- Apicurion container properties -->
         <apicurio.port>8080</apicurio.port>
         <apicurio.init.timeout>60000</apicurio.init.timeout> <!-- 60 seconds -->
+        <!-- Mongo test containers -->
+        <mongodb.docker.desktop.ports>27017,27018,27019</mongodb.docker.desktop.ports>
```
jcechace, line 112: @jpechane  This configures the port pool available for MongoDB containers. Two patterns are supported

- Port list `27017,27018,27019`
- Port range `27017:27019`

This also means that currently the tests can run at most 3 Mongo containers in parallel.

debezium-connector-mongodb/pom.xml
jcechace: Unfortunately JDK9 to JDK17 don't provide a way to provide custom implementation of `NameService` without reflective access to private members. A proper SPI to do this was added in JDK18

jcechace:
Loggs are showing the following error 

```
"Migration Results for the last 24 hours":{"1":"Failed with error 'aborted', from shard2 to shard1"}}
``

This can happen for multitude of reasons  -- network communication, hardware resources and so on. 

I'm going to disable the sharding test for now, create a Jira and re-enable it later.

File: debezium-connector-mongodb/pom.xml [hunk 3]
```diff
@@ -90,25 +105,17 @@
...
+        <mongodb.replica.size>1</mongodb.replica.size>
```
jcechace, line 118: @jpechane  what should be the dealt here? We can either have `1` which would mean that locally one will have to think about it or we can keep the `3` and set the property in GH action definition.

jpechane:
@jcechace Great job! applied

PR #4112, opened 2022-12-06
DBZ-5772 Add missing dependency to debezium-server-dist/pom.xml

Skezzowski:


jpechane:
@Skezzowski Applied, thanks

PR #4113, opened 2022-12-06
DBZ-5901 Migrate debezium-server-nats-jetstream to AssertJ

harveyyue:
https://issues.redhat.com/browse/DBZ-5901


jpechane:
@harveyyue Applied, thanks.

PR #4120, opened 2022-12-07
DBZ-5908 ConnectorLifecycle is not logging anymore the exception stac…

ggaborg:
…ktrace when startup fails


jpechane:
@ggaborg Applied, thanks

PR #4122, opened 2022-12-07
DBZ-5904 Suppport INSERT INTO statements with dots in column names

nirolevy:
https://issues.redhat.com/browse/DBZ-5904

Upstream pr:
https://github.com/antlr/grammars-v4/pull/2945


jpechane:
@nirolevy Applied, thanks

PR #4125, opened 2022-12-07
DBZ-5906 Correct additional-conditions default in adhoc snapshots doc

roldanbob:
[DBZ-5906](https://issues.redhat.com/browse/DBZ-5906)

This change corrects the `additional-conditions` argument that was listed as the default value in the documentation for ad hoc incremental snapshots and MySQL ad hoc read-only incremental snapshots.


jpechane:
@roldanbob Applied, thanks

PR #4126, opened 2022-12-08
DBZ-5895 Don't hard-code plugin name into the test 

vjuranek:
https://issues.redhat.com/browse/DBZ-5895


vjuranek:
Tests failures are unrelated:-) Failed due to [DBZ-5914](https://issues.redhat.com/browse/DBZ-5914) which very likely failed also before, but it was hidden due to previous failure of `RecordsStreamProducerIT`

vjuranek:
should be fixed in #4128

jpechane:
@vjuranek Thanks a lot for finding that, merged via https://github.com/debezium/debezium/pull/4121

PR #4128, opened 2022-12-08
DBZ-5811,DBZ-5914 Don't manipulate with LSN when LSN flushing LSN is …

vjuranek:
…disabled

Befre we start streamig, we adjust `confirmed_flush_lsn` in Postgres to the position where we determined that the streaming should start. When user disables LSN flushing, which is not desired as it's the user reponsibility to flush LSN in Postgres.

Also adjust related tests, as when LSN flushing is not disabled, flushed LSN in Postgres may change (as we e.g. determined start of streaming from offset or from `xlogpos`) and can result in the test failure. When flushing is enabled, only LSNs after start of streaming can be compared.

https://issues.redhat.com/browse/DBZ-5811
https://issues.redhat.com/browse/DBZ-5914


jpechane:
@vjuranek I am really not sure about this change. Yes we try to advance the slot but does it mean that it moves the flush LSN too? As if not then it is pretty safe to be used even in non-flushing scenario.

jpechane:
@vjuranek Thanks for the explanation. It makes sense then why I did not see it locally. Merged via https://github.com/debezium/debezium/pull/4121

PR #4131, opened 2022-12-11
DBZ-5911 Debezium Server stops with NPE when Redis does not report th…

ggaborg:
…e "maxmemory" field in "info memory" command


debezium-server/debezium-server-redis/src/main/java/io/debezium/server/redis/RedisMemoryThreshold.java
```diff
@@ -0,0 +1,120 @@
...
+    private long memoryLimit;
+
+    private Supplier<Boolean> isMemoryOk;
+
+    public RedisMemoryThreshold(RedisClient client, int memoryThreshold, int memoryLimitMb) {
+        this.client = client;
+        this.memoryThreshold = memoryThreshold;
+        this.memoryLimit = 1024L * 1024 * memoryLimitMb;
+        Tuple2<Long, Long> memoryTuple;
+        if (memoryThreshold == 0 || (memoryTuple = memoryTuple(memoryLimit)) == null || memoryTuple.getItem2() == -1) {
```
jpechane, line 45: @ggaborg Please unwind this condition. I am not a great fan of assignments in ifs as they tend to be error-prone in long term run from maintenance PoV.

debezium-server/debezium-server-redis/src/main/java/io/debezium/server/redis/RedisMemoryThreshold.java
```diff
@@ -0,0 +1,120 @@
...
+import java.util.function.Supplier;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.debezium.storage.redis.RedisClient;
+import io.debezium.util.IoUtil;
+import io.smallrye.mutiny.tuples.Tuple2;
+
+public class RedisMemoryThreshold implements Supplier<Boolean> {
```
jpechane, line 22: As `Supplier` is encapsulated in this class I don't think the class itself should be `Supplier` as it is always instantiated and used. There is no alternative computation.

debezium-server/debezium-server-redis/src/main/java/io/debezium/server/redis/RedisMemoryThreshold.java
```diff
@@ -0,0 +1,120 @@
...
+
+    private RedisClient client;
+
+    private int memoryThreshold;
+
+    private long memoryLimit;
+
+    private Supplier<Boolean> isMemoryOk;
+
+    public RedisMemoryThreshold(RedisClient client, int memoryThreshold, int memoryLimitMb) {
```
jpechane, line 40: I'd pass the full config object, not just two memory settings as it can be obtained form that and would be easier to exand ion in future if additional settings are present.

jpechane (changes requested):
@ggaborg LGTM, thanks! I left few comments mostly related to refactoring, no functional changes.

ggaborg:
@jpechane Hi, made the changes according to the CR. The tests are passing locally, here seems a problem connecting to the Redis container. Can you give it another go? Thanks!

jpechane:
@ggaborg Applied, thanks

PR #4132, opened 2022-12-12
DBZ-3576 Implement window function in MySQL parser

uurl:
This PR closes [DBA-3576](https://issues.redhat.com/browse/DBZ-3576) Implement window function in MySQL parser
Shows that Window Function with Over clause are already implemented.


jpechane:
@uurl Applied, thanks!

PR #4135, opened 2022-12-13
DBZ-5928 Fix required Java version in Install docs

vjuranek:
https://issues.redhat.com/browse/DBZ-5928


jpechane:
@vjuranek Applied, thanks

PR #4137, opened 2022-12-14
DBZ-5936 Handle toasted int/bigint json/jsonb array columns

prburgu:
This PR addresses https://issues.redhat.com/browse/DBZ-5936 for int/bigint and json/jsonb array columns.

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
@@ -473,6 +473,8 @@ public class PostgresConnectorConfig extends RelationalDatabaseConnectorConfig {
     public static final Field PORT = RelationalDatabaseConnectorConfig.PORT
             .withDefault(DEFAULT_PORT);
 
+    public static final String DEFAULT_UNAVAILABLE_NUMBER_ARRAY_PLACEHOLDER = "28694827,62948267,94852637,62847362,94837264,73486259,29483726,43827619,84692735,29873462";
+
     public static final Field PLUGIN_NAME = Field.create("plugin.name")
             .withDisplayName("Plugin")
             .withGroup(Field.createGroupEntry(Field.Group.CONNECTION_ADVANCED_REPLICATION, 0))
@@ -833,6 +835,16 @@ public class PostgresConnectorConfig extends RelationalDatabaseConnectorConfig {
                     "the original value is a toasted value not provided by the database. " +
                     "If starts with 'hex:' prefix it is expected that the rest of the string represents hexadecimal encoded octets.");
 
+    public static final Field UNAVAILABLE_NUMBER_ARRAY_PLACEHOLDER = Field.create("unavailable.number.array.placeholder")
+            .withDisplayName("Unavailable number array placeholder")
+            .withType(Type.STRING)
+            .withGroup(Field.createGroupEntry(Field.Group.CONNECTOR_ADVANCED, 20))
+            .withWidth(Width.MEDIUM)
+            .withDefault(DEFAULT_UNAVAILABLE_NUMBER_ARRAY_PLACEHOLDER)
+            .withImportance(Importance.MEDIUM)
+            .withDescription("Specify the numbers separated by comma that will be provided by Debezium to indicate that " +
+                    "the original value is int/bigint array and not provided by the database.");
+
     public static final Field MONEY_FRACTION_DIGITS = Field.create("money.fraction.digits")
             .withDisplayName("Money fraction digits")
             .withType(Type.SHORT)
@@ -974,6 +986,10 @@ public class PostgresConnectorConfig extends RelationalDatabaseConnectorConfig {
             return Strings.hexStringToByteArray(placeholder.substring(4));
         }
         return placeholder.getBytes();
+    }
+
+    public String getUnavailableNumberArrayPlaceholder() {
+        return getConfig().getString(UNAVAILABLE_NUMBER_ARRAY_PLACEHOLDER);
     }
 
     protected int moneyFractionDigits() {
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
@@ -473,8 +473,6 @@ public class PostgresConnectorConfig extends RelationalDatabaseConnectorConfig {
     public static final Field PORT = RelationalDatabaseConnectorConfig.PORT
             .withDefault(DEFAULT_PORT);
 
-    public static final String DEFAULT_UNAVAILABLE_NUMBER_ARRAY_PLACEHOLDER = "28694827,62948267,94852637,62847362,94837264,73486259,29483726,43827619,84692735,29873462";
-
     public static final Field PLUGIN_NAME = Field.create("plugin.name")
             .withDisplayName("Plugin")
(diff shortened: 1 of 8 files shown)


prburgu:
@jpechane This PR handles TOASTed integer and bigint array columns for https://issues.redhat.com/browse/DBZ-5936. Let me know if you have any feedback on the approach. As you can see we'll need to come up with placeholder values for each type.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/UnchangedToastedReplicationMessageColumn.java
```diff
@@ -24,13 +24,23 @@ public class UnchangedToastedReplicationMessageColumn extends AbstractReplicatio
...
-    private boolean isArrayColumn = false;
+    private boolean isStringArrayColumn = false;
+    private boolean isIntArrayColumn = false;
+    private boolean isBigintArrayColumn = false;
 
     public UnchangedToastedReplicationMessageColumn(String columnName, PostgresType type, String typeWithModifiers, boolean optional) {
         super(columnName, type, typeWithModifiers, optional);
         if (typeWithModifiers.equals("text[]") || typeWithModifiers.equals("_text")) {
-            isArrayColumn = true;
+            isStringArrayColumn = true;
```
jpechane, line 37: Looking at the code as is know. Doe sit make sense to set a flag as instance attribute and then return a values based on that flag? Maybe the value returned should be set directly to an attribute in constructor and just returned in `getValue`?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java
```diff
@@ -173,14 +174,15 @@ public static PostgresValueConverter of(PostgresConnectorConfig connectorConfig,
...
+                connectorConfig.getUnavailableNumberArrayPlaceholder(),
                 connectorConfig.moneyFractionDigits());
     }
 
     protected PostgresValueConverter(Charset databaseCharset, DecimalMode decimalMode,
                                      TemporalPrecisionMode temporalPrecisionMode, ZoneOffset defaultOffset,
                                      BigIntUnsignedMode bigIntUnsignedMode, boolean includeUnknownDatatypes, TypeRegistry typeRegistry,
                                      HStoreHandlingMode hStoreMode, BinaryHandlingMode binaryMode, IntervalHandlingMode intervalMode,
-                                     byte[] toastPlaceholder, int moneyFractionDigits) {
+                                     byte[] toastPlaceholder, String toastPlaceholderNumberArrayString, int moneyFractionDigits) {
```
jpechane, line 185: If there will be additional placeholders it is not good to pass them as extended list of params.
In that case we should introduce a class that will manage all placeholder types and puss just that one.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
```diff
@@ -473,6 +473,8 @@ public static SchemaRefreshMode parse(String value) {
     public static final Field PORT = RelationalDatabaseConnectorConfig.PORT
             .withDefault(DEFAULT_PORT);
 
+    public static final String DEFAULT_UNAVAILABLE_NUMBER_ARRAY_PLACEHOLDER = "28694827,62948267,94852637,62847362,94837264,73486259,29483726,43827619,84692735,29873462";
```
jpechane, line 476: What does this number sequence mean?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/UnchangedToastedArrayPlaceholder.java
```diff
@@ -0,0 +1,37 @@
...
+ */
+package io.debezium.connector.postgresql;
+
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+public class UnchangedToastedArrayPlaceholder {
```
jpechane, line 14: Would it be possible to rename this class to `UnchangedToastedPlaceholder` and move `String` and `byte[]` handling there so it is all in the same place?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/UnchangedToastedArrayPlaceholder.java
```diff
@@ -0,0 +1,37 @@
...
+                        .map(Integer::parseInt)
+                        .collect(Collectors.toList()));
+        placeholderValues.put(UnchangedToastedReplicationMessageColumn.UNCHANGED_BIGINT_ARRAY_TOAST_VALUE,
+                Stream.of(toastPlaceholderNumberArrayString.split(","))
+                        .map(String::trim)
+                        .map(Long::parseLong)
+                        .collect(Collectors.toList()));
+    }
+
+    public Object getValue(Object obj) {
```
jpechane, line 34: Let's use Optional<Object> here instead of returning `null`.

jpechane (commented):
@prburgu I think this is moving in the right direction! Still could you please take a look at my comments and move further steps forward to unify non-array and array placeholder handling?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
prburgu: There is no specific meaning. It's random sequence of numbers. Happy to replace it with something else.

jpechane: @prburgu ok, thinking about that, do we really need a separate config option then? Can't we just call `getUnavailableValuePlaceholder()` returning `byte[]` and convert it into appropriate array type?

prburgu: Yeah, we can do that. Do we always convert 4 bytes into one integer, or do we convert to integer depending on the input type (I.e. int vs bigint)?

jpechane: @prburgu I think it would be unnecessary overkill. I added a commit that just converts byte[] to int[] or long[].

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/UnchangedToastedPlaceholder.java
```diff
@@ -0,0 +1,48 @@
...
+package io.debezium.connector.postgresql;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+public class UnchangedToastedPlaceholder {
```
jpechane, line 15: COuld you please add a JavaDoc to this class? It is last piece missing and we are then good to go.

jpechane:
@prburgu I am sorry, I completely forgot. Could you please add a test for JSON/JSONB array too? The JSON datatypes have a specific schema type and I just want to make sure that it is correctly applied in this case too.

prburgu:
@jpechane The JSON/JSONB columns are represented as text and it works without an issue. The json[]/jsonb[] (arrays) are correctly represented as String arrays.

jpechane:
@prburgu Yes, but their schema name should be `io.debezium.data.Json` so I wonder if it is the case.

jpechane:
@prburgu Applied, great job!

PR #4138, opened 2022-12-15
DBZ-5940 Support predicate parameters in Debezium Server

jeremy-l-ford:
Add predicate configuration support


jpechane:
@jeremy-l-ford Thanks a lot! Would it be possible add a test for it? Just a positive use case, nothing else.

jeremy-l-ford:
@jpechane - updated PR

jpechane:
@jeremy-l-ford Applied, thanks

PR #4141, opened 2022-12-15
Implement converte to zero scale decimals

vjuranek:
https://issues.redhat.com/browse/DBZ-5880


jpechane:
@vjuranek Applied, thanks!

PR #4142, opened 2022-12-15
Dbz 5922 remove cassandra incubating notes

roldanbob:
[DBZ-5922](https://issues.redhat.com/browse/DBZ-5922)

Removes note stating that the Cassandra connector is in an incubating state. 
Also includes minor edits for formatting and language.


jpechane:
@roldanbob Applied, thanks

PR #4143, opened 2022-12-15
DBZ-5944 Toasted varchar array is not correctly processed

prburgu:
Similar to https://github.com/debezium/debezium/pull/4136 but handles `character varying[]` type


jpechane:
Applied, thanks

PR #4148, opened 2022-12-17
DBZ-5818 field.exclude.list in MongoDB Connector v2.0 doesn't accept * as a wildcard for collectionName

harveyyue:
https://issues.redhat.com/browse/DBZ-5818


jpechane:
@harveyyue Applied, thanks!

PR #4149, opened 2022-12-17
DBZ-5946 Print the correct data class name in JdbcValueConverters.handleUnknownData

harveyyue:
https://issues.redhat.com/browse/DBZ-5946


jpechane:
@harveyyue Applied, thanks

PR #4150, opened 2022-12-18
DBZ-5926 Add support for Connect Headers to Debezium Server

jeremy-l-ford:
Add headers support to ChangeEvent
Add SerializationFormat support headers

### DIFF
--- a/documentation/modules/ROOT/pages/development/engine.adoc
+++ b/documentation/modules/ROOT/pages/development/engine.adoc
@@ -259,8 +259,15 @@ Allowed values are:
 
 * `Connect.class` - the output value is change event wrapping Kafka Connect's `SourceRecord`
 * `Json.class` - the output value is a pair of key and value encoded as `JSON` strings
+* `JsonByteArray.class` - the output value is a pair of key and value formatted as `JSON` and encoded as UTF-8 byte arrays
 * `Avro.class` - the output value is a pair of key and value encoded as Avro serialized records (see xref:{link-avro-serialization}[Avro Serialization] for more details)
 * `CloudEvents.class` - the output value is a pair of key and value encoded as xref:{link-cloud-events}[Cloud Events] messages
+
+The header format can also be specified when calling `DebeziumEngine#create()`.
+Allowed values are:
+
+* `Json.class` - the header values are encoded as `JSON` strings
+* `JsonByteArray.class` - the header values are formatted as `JSON` and encoded as UTF-8 byte arrays
 
 Internally, the engine uses the appropriate Kafka Connect converter implementation to which the conversion is delegated.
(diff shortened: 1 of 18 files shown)


documentation/modules/ROOT/pages/development/engine.adoc
```diff
@@ -259,9 +259,18 @@ Allowed values are:
...
+* `JsonByteArray.class` - the output value is a pair of key and value formatted as `JSON` and encoded as UTF-8 byte arrays
 * `Avro.class` - the output value is a pair of key and value encoded as Avro serialized records (see xref:{link-avro-serialization}[Avro Serialization] for more details)
 * `CloudEvents.class` - the output value is a pair of key and value encoded as xref:{link-cloud-events}[Cloud Events] messages
 
+The header format can also be specified when calling `DebeziumEngine#create()`.
+Allowed values are:
+
+* `Json.class` - the header values are encoded as `JSON` strings
+* `JsonByteArray.class` - the header values are formatted as `JSON` and encoded as UTF-8 byte arrays
+* `Simple.class` - the header values are encoded as strings
```
jpechane, line 271: I must admit that I am not really happy about these two options. It binds Debezium Server tightly to the way how the Connect serializes them. Maybe the plain `toString()` would suffice for most of data types?

jeremy-l-ford: Just double checking that you're referring to the Simple format?  That uses the SimpleHeaderConverter from Kafka which should boil down to 
`
StringBuilder sb = new StringBuilder();
                    append(sb, value, false);
                    return sb.toString();
`

jpechane: @jeremy-l-ford Yes, that's the case. If we decide to retain that format then we should do it mostly ourselves. Let' me ask, how is for example number and string represented with Json?

jeremy-l-ford: When using the JsonConverter and the schema option is not enabled, for the given inputs boolean true, int 100, and string myTest, the output is

true
100
"myTest"

jeremy-l-ford: I think apicurio 2.4.1 is causing the tests to fail for some reason.  I manually set a test locally to use 2.3.1 and it passed without issue.

jpechane: It probably relates to https://github.com/Apicurio/apicurio-registry/issues/3024

jpechane:
@jeremy-l-ford Applied, thanks a lot!

jpechane:
I also logged https://issues.redhat.com/browse/DBZ-6017

PR #4151, opened 2022-12-19
DBZ-5344 Add complex json array test case to EventRouterTest

harveyyue:
https://issues.redhat.com/browse/DBZ-5344


jpechane:
@harveyyue Applied, thanks!

PR #4152, opened 2022-12-19
DBZ-5915: correctly determine the startStreamingLsn

rajdangwal:
Addressing https://issues.redhat.com/browse/DBZ-5915

The LSN of the `BEGIN` message is same as the message following it.
If the `lastEventStoredLsn` corresponds to the `BEGIN` message, we are skipping the message following it because the LSN is same. This is causing data loss if the following message corresponds to any DML statement.

Proposed solution:
While looking for `lsnAfterLastEventStoredLsn`, instead of skipping the LSN which is same as the `lastEventStoredLsn` we set the `startStreamingLsn` as `txStartLsn` and return it. This will prevent the data loss.
In some cases, where the `lastEventStoredLsn` belongs to the message following the `BEGIN` message, it could lead to duplication of the message following the `BEGIN` message, which I think is fine.

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
@@ -71,6 +71,12 @@ public class WalPositionLocator {
             // We can resume streaming from it
             if (currentLsn.equals(lastEventStoredLsn)) {
                 // BEGIN and first message after change have the same LSN
+                if (txStartLsn != null) {
+                    // start from the BEGIN tx; prevent skipping of unprocessed event after BEGIN
+                    // may reprocess the event after the BEGIN
+                    startStreamingLsn = txStartLsn;
+                    return Optional.of(startStreamingLsn);
+                }
                 return Optional.empty();
             }
             lsnAfterLastEventStoredLsn = currentLsn;


rajdangwal:
Test `io.debezium.connector.postgresql.RecordsStreamProducerIT#restartConnectorInTheMiddleOfUncommittedTx` is failing as it doesn't expect the event after BEGIN to be reprocessed. But with current implementation we may see the duplicate events in case the `lastEventStoredLsn` is of the event that follows the BEGIN event.

@jpechane if the re-processing seems okay, I will go ahead and update the test.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
```diff
@@ -71,6 +71,12 @@ public Optional<Lsn> resumeFromLsn(Lsn currentLsn, ReplicationMessage message) {
             // We can resume streaming from it
             if (currentLsn.equals(lastEventStoredLsn)) {
                 // BEGIN and first message after change have the same LSN
+                if (txStartLsn != null) {
+                    // start from the BEGIN tx; prevent skipping of unprocessed event after BEGIN
```
jpechane, line 75: @rajdangwal Could we avoid the reprocessing if we add the message type to offsets? In that case we will see if it was tx begin or change and in that case we could choose the right branch.
Also for upgrade if the message type is not in offsets then choose the safe path causing potential reprocessing.
WDYT?

rajdangwal: Thanks @jpechane 
Sounds great. I will work on the same.

rajdangwal: Made the changes, but I am not able to add the respective tests because can't find a way to restart the connector after committing only the BEGIN message.

jpechane:
@rajdangwal LGTM, applied! Thanks.
For the testing, would you be willing to take care of https://issues.redhat.com/browse/DBZ-5978 ?
The approach could be verified on this change.

PR #4154, opened 2022-12-19
DBZ-5949 Support partitioned tables when updating PG publication

jchipmunk:
`PostgresConnection.getAllTableIds(String)` method was added to unify getting all tables in database catalog, including partitioned tables.

https://issues.redhat.com/browse/DBZ-5949


jpechane:
@jchipmunk Applied, thanks!

PR #4161, opened 2022-12-27
DBZ-5962:Sink adapter for Apache RocketMQ

sunxiaojian:
Add apache RocketMQ sink adapter for debezium server


debezium-server/debezium-server-rocketmq/src/test/java/io/debezium/server/rocketmq/RocketMQContainer.java
```diff
@@ -0,0 +1,74 @@
...
+    @Override
+    protected void configure() {
+        String command = "#!/bin/bash\n";
+        command += "./mqnamesrv &\n";
+        command += "./mqbroker -n localhost:" + NAMESRV_PORT;
+        withCommand("sh", "-c", command);
+    }
+
+    @Override
+    @SneakyThrows
```
jpechane, line 45: We don't use Lombok so I think this cen be removed.

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMQChangeConsumer.java
```diff
@@ -0,0 +1,145 @@
...
+import io.debezium.engine.ChangeEvent;
+import io.debezium.engine.DebeziumEngine;
+import io.debezium.server.BaseChangeConsumer;
+
+/**
+ * rocketmq change consumer
+ */
+@Named("rocketmq")
+@Dependent
+public class RocketMQChangeConsumer extends BaseChangeConsumer implements DebeziumEngine.ChangeConsumer<ChangeEvent<Object, Object>> {
```
jpechane, line 43: Please use `RocketMqChangeConsumer` name.

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMQChangeConsumer.java
```diff
@@ -0,0 +1,145 @@
...
+    }
+
+    @Override
+    public void handleBatch(List<ChangeEvent<Object, Object>> records, DebeziumEngine.RecordCommitter<ChangeEvent<Object, Object>> committer)
+            throws InterruptedException {
+        final CountDownLatch latch = new CountDownLatch(records.size());
+        for (ChangeEvent<Object, Object> record : records) {
+            try {
+                LOGGER.trace("Received event '{}'", record);
+                mqProducer.send(new Message(record.destination(), null, getString(record.key()), getBytes(record.value())),
```
jpechane, line 122: You should support configurable stream name mapping.

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMQChangeConsumer.java
```diff
@@ -0,0 +1,145 @@
...
+    // common config
+    private static final String PROP_PRODUCER_NAME_SRV_ADDR = PROP_PRODUCER_PREFIX + "name.srv.addr";
+    private static final String PROP_PRODUCER_GROUP = PROP_PRODUCER_PREFIX + "group";
+    private static final String PROP_PRODUCER_MAX_MESSAGE_SIZE = PROP_PRODUCER_PREFIX + "max.message.size";
+    private static final String PROP_PRODUCER_SEND_MSG_TIMEOUT = PROP_PRODUCER_PREFIX + "send.msg.timeout";
+
+    private DefaultMQProducer mqProducer;
+
+    @PostConstruct
+    void connect() {
```
jpechane, line 63: Could you please add support for `@CustomConsumerBuilder` like e.g. in `io.debezium.server.kinesis.KinesisChangeConsumer`?

jpechane (changes requested):
Thanks a lot, this is a great start!
I left couple of comments.
Could you lease also squash all commits into a single one and make sure the commit message it is prefixed with `DBZ-5962`?
Also what is your name so we can add you to the copyright list?

sunxiaojian:
> Thanks a lot, this is a great start! I left couple of comments. Could you lease also squash all commits into a single one and make sure the commit message it is prefixed with `DBZ-5962`? Also what is your name so we can add you to the copyright list?

Thanks for the review. My name is  Xiaojian Sun ( @sunxiaojian )，I have made changes according to the suggestions given

jpechane:
@sunxiaojian Could you please squas all commits to a single one and expecially emove the merge commit? Thanks.

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
```diff
@@ -0,0 +1,164 @@
...
+                                throw new DebeziumException(throwable);
+                            }
+                        });
+                committer.markProcessed(record);
+            }
+            catch (Exception e) {
+                throw new DebeziumException(e);
+            }
+        }
+        latch.await();
```
vjuranek, line 161: This can potentially get stuck forever. IMHO it would be better to provide (configurable) time out and throw and exception when timeout is reached. The thing is that waiting here forever may be not noticed and meanwhile cause issue somewhere else and as a result make debugging of such issue much harder. Therefore is IMHO better to be clear about the issue and if the message is not delivered in timely manner, finish immediately.

debezium-server/debezium-server-rocketmq/src/test/java/io/debezium/server/rocketmq/RocketMqTestResourceLifecycleManager.java
```diff
@@ -0,0 +1,41 @@
...
+        return new HashMap<>();
+    }
+
+    @Override
+    public void stop() {
+        rocketMQContainer.stop();
+    }
+
+    public static String getNamesrvAddr() {
+        rocketMQContainer.start();
```
vjuranek, line 34: Is this needed? Start should be called automatically by `QuarkusTestResource`.

vjuranek (changes requested):
There is couple of small issue/comment. Beside that, please squash you commit and also please update the server documentation `debezium-server.adoc`.

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
```diff
@@ -0,0 +1,164 @@
...
+                                latch.countDown();
+                            }
+
+                            @Override
+                            public void onException(Throwable throwable) {
+                                LOGGER.error("Failed to send record to {}:", record.destination(), throwable);
+                                throw new DebeziumException(throwable);
+                            }
+                        });
+                committer.markProcessed(record);
```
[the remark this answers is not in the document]
sunxiaojian: > Messages are sent asynchronously, so marking messages as processed here is not correct, it should be IMHO in `onSuccess()` callback.

Calling onSuccess() may cause the offset submission to be out of order

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
sunxiaojian: This situation is caused by the blocking or failure of sending messages. Now, if the message sending timeout or the sending return fails, an exception will be thrown and the program will terminate immediately. Therefore, I think that under normal circumstances, you should always wait for the sending to succeed

PR #4163, opened 2023-01-02
DBZ-5282 Debezium is not working with apicurio and custom truststores

jeremy-l-ford:
Update to latest version of apicurio
Force apicurio docker image to 2.3.1 temporarily

### DIFF
--- a/debezium-testing/debezium-testing-system/pom.xml
+++ b/debezium-testing/debezium-testing-system/pom.xml
@@ -20,7 +20,7 @@
     <version.strimzi.kafka>${version.kafka}</version.strimzi.kafka>
     <version.junit5>5.7.2</version.junit5>
     <version.assertj>3.11.1</version.assertj>
-    <version.apicurio>2.2.0.Final</version.apicurio>
+    <version.apicurio>2.4.1.Final</version.apicurio>
     <version.apicurio.model>2.2.0.Final</version.apicurio.model>
     <version.fixture5>0.1.0</version.fixture5>
 
--- a/debezium-testing/debezium-testing-system/pom.xml
+++ b/debezium-testing/debezium-testing-system/pom.xml
@@ -20,7 +20,7 @@
     <version.strimzi.kafka>${version.kafka}</version.strimzi.kafka>
     <version.junit5>5.7.2</version.junit5>
     <version.assertj>3.11.1</version.assertj>
-    <version.apicurio>2.4.1.Final</version.apicurio>
+    <version.apicurio>2.2.0.Final</version.apicurio>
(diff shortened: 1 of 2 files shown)


debezium-server/pom.xml
```diff
@@ -109,6 +109,21 @@
             <groupId>io.apicurio</groupId>
             <artifactId>apicurio-registry-utils-converter</artifactId>
         </dependency>
+        <dependency>
```
jpechane, line 112: @jeremy-l-ford Why are the dependencies necessary to be explcitily defined? Also if they are in the test deps block should not they be marked as test scope?

jeremy-l-ford: @jpechane - Locally I was seeing 2.3.1 being pulled in as a transitive for these jars which continued to cause issues.    Since apicurio-registry-utils-converter is not marked as a test scoped dependency, I thought that these additional jars should also not be test scoped.

jpechane: @jeremy-l-ford Thanks for the explanation. I ned to check if this is an error in scope or just a position in pom.
I am not going to merge this PR until https://issues.redhat.com/browse/DBZ-5977 is resolved. In that case the changes here should be only relevant to the issue and not handle anything else.

jeremy-l-ford: @jpechane From my reading of [DBZ-5282](https://issues.redhat.com/browse/DBZ-5282), it boils down to just needing the apicurio dependencies updated.

jpechane: @jeremy-l-ford Apicuri dependencies has been updated and the new nightly image was published. Could you take a look if it is possible to write a test for the SSL setup?

debezium-testing/debezium-testing-system/pom.xml
```diff
@@ -20,7 +20,7 @@
     <version.strimzi.kafka>${version.kafka}</version.strimzi.kafka>
     <version.junit5>5.7.2</version.junit5>
     <version.assertj>3.11.1</version.assertj>
-    <version.apicurio>2.2.0.Final</version.apicurio>
+    <version.apicurio>2.4.1.Final</version.apicurio>
```
jpechane, line 23: @jcechace Could we do the change here? I think this is downstream not upstream version, right?

jeremy-l-ford: @jpechane - Looking at this closer, it appears that this property does not actually control an apicurio dependency.   Were you wanting me to revert this?

jpechane: Reverted by me.

jpechane:
@jeremy-l-ford Applied, thanks a lot.

PR #4165, opened 2023-01-03
DBZ-2296 Change handling of GTID_MODE=OFF_PERMISSIVE and ON_PERMISSIVE

echatman-ias:
return false instead of true from isGtidModeEnabled for GTID_MODEs of OFF_PERMISSIVE and ON_PERMISSIVE


jpechane:
@echatman-ias Applied, thanks

PR #4166, opened 2023-01-03
Dbz 5976

erdinctaskin:
ReplaceAll use regex under the hood. As a nature of regex, it may slow. We can  use "replace" method rather than "replaceAll" as much as possible in order to do better perform


jpechane:
@erdinctaskin Applied, thanks!

PR #4167, opened 2023-01-04
DBZ-5967: Fix negative partition number in ComputePartition SMT

mfvitale:
- Fix negative partition number in ComputePartition SMT
- Improved check on configuration consistency to avoid 0 as the partition number.


mfvitale:
@jpechane

jpechane:
@mfvitale Applied, thanks!

PR #4168, opened 2023-01-04
DBZ-5974 Use `catchError` instead of try/finall block

vjuranek:
Also show the build status directly in email subject

https://issues.redhat.com/browse/DBZ-5974


jpechane:
@vjuranek LGTM, applied. Thanks!

PR #4170, opened 2023-01-04
DBZ-3866 Add internal read-only LogMiner support

Naros:
https://issues.redhat.com/browse/DBZ-3866

This explicitly removes the writing of the flush table to support read-only logical standby connections with LogMiner only.  I have configured the option for triggering read-only to be internal, `internal.log.mining.read.only`, as this is highly experimental.


jpechane:
@Naros Applied, thanks!

PR #4171, opened 2023-01-05
DBZ-5798 Data type conversion failed for mysql bigint

harveyyue:
https://issues.redhat.com/browse/DBZ-5798

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlDatabaseSchema.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlDatabaseSchema.java
@@ -305,7 +305,7 @@ public class MySqlDatabaseSchema extends HistorizedRelationalDatabaseSchema {
                     sanitizedDbName,
                     null,
                     event.statement(),
-                    tableId != null ? tableFor(tableId) : null,
+                    tableId != null ? tables().forTable(tableId) : null,
                     ((TableAlteredEvent) event).previousTableId());
         }
         else {
@@ -316,7 +316,7 @@ public class MySqlDatabaseSchema extends HistorizedRelationalDatabaseSchema {
                     sanitizedDbName,
                     null,
                     event.statement(),
-                    tableId != null ? tableFor(tableId) : null,
+                    tableId != null ? tables().forTable(tableId) : null,
                     snapshot);
         }
         schemaChangeEvents.add(schemaChangeEvent);


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/listener/CreateTableParserListener.java
```diff
@@ -31,7 +31,10 @@ public CreateTableParserListener(MySqlAntlrDdlParser parser, List<ParseTreeListe
     @Override
     public void enterColumnCreateTable(MySqlParser.ColumnCreateTableContext ctx) {
         TableId tableId = parser.parseQualifiedTableId(ctx.tableName().fullId());
-        if (parser.databaseTables().forTable(tableId) == null) {
+        Table existedTable = parser.databaseTables().forTable(tableId);
```
jpechane, line 34: @harveyyue I am really not sure if this is correct solution. How it is possible to have two {{CREATE TABLE}} for the same table? It sounds like a hack/workaround to a different problem.

harveyyue: @jpechane Yes, there are more  {{CREATE TABLE}} statements for same table.
Our dba use precona tool to alter table schema, when alter table schema, will create the table `_some_table_new` first, in other word there are some `_some_table_new` created statements in history topic, and we add the regex expression to exclude these intermediate tables to avoid capturing duplicated data to intermediate topic, this filter will set null value to property `tableChanges` of HistoryRecord, so antlr has to parse same table for many times during the restarted connector phase.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlDatabaseSchema.java
```diff
@@ -305,7 +305,7 @@ private void emitChangeEvent(MySqlPartition partition, MySqlOffsetContext offset
                     sanitizedDbName,
                     null,
                     event.statement(),
-                    tableId != null ? tableFor(tableId) : null,
+                    tableId != null ? tables().forTable(tableId) : null,
```
harveyyue, line 308: To avoid the re-parse the ddl statement when specify some excluded table filter in restarting connector phase, I guess re-parse the ddl and deserialize the table info to cause the existed table which not be removed.

harveyyue: @jpechane Check the internal branch and found missing this commit [DBZ-4451 Store previous table id in alter schema change events
](https://github.com/debezium/debezium/commit/863abcd0814b297cd473fbaf5fc7fdb0ab87648b#diff-68e27c3497cd7a11057780c3bef9d3358f7630f67e8e18c75a7d8ff1505542eb) which casue the keep the related table in object Tables.

jpechane: @harveyyue Could you please explain this in more detail? Given the your last comment I don;t fully understand the change.

harveyyue: FYI:
**Connector config:**
"table.exclude.list": ".*\\._+.*,.*\\..*_old$,.*\\..*_new$"
**Internal branch:**
Currently, We didn't merge DBZ-4451 commits to internal branch which means will not remove existing `_some_table_new` from `Tables` as below code snippet.
AbstractSchemaHistory.recover(...)
`
                        if (entry.getType() == TableChangeType.CREATE) {
                            schema.overwriteTable(entry.getTable());
                        }
                        else if (entry.getType() == TableChangeType.ALTER) {
                            if (entry.getPreviousId() != null) {
                                schema.removeTable(entry.getPreviousId());
                            }
                            schema.overwriteTable(entry.getTable());
                        }
`

DBA add column to table `_some_table_new` for some times from business request, will produce some series of following ddl statements. 
When restart the connector and recovery history schema
1. create table _some_table_new(id bigint not null AUTO_INCREMENT, c1 varchar(255), PRIMARY KEY (id));
-- Antrl will parse this statement becasue tableChanges is null, but table `_some_table_new` has been existed in `Tables`, so antrl will skip this parsing.
2. alter table _some_table_new table add column c2 bigint not null default 0;
-- Add column `c2` to existing table `_some_table_new`(this table schema info maybe stale)
3. rename table some_table to _some_table_old, _some_table_new to some_table;
-- Will set stale `_some_table_new` to normal table `some_table`
This is why we are missing some columns and casue the column mismatch in build table schema phase.

jpechane: @harveyyue ok, I undrestand but this leads me to two questions

1) Is this the root cause of the referred issue? If not then there should be probably a new one
2) In the current state of affairs - is it really bug in Debezium or not?

jpechane: @harveyyue Any opinion on the quesitons above?

harveyyue: @jpechane Sorry about the late response.
Point 1:  Yes, it is the root cause.
Point 2: Not a bug in current release version.
This pr will help to complete the property `tableChanges` for excluded tables.

jpechane:
@harveyyue Applied, thanks

PR #4175, opened 2023-01-05
DBZ-5966 Truncate records incompatible with ExtractNewRecordState

harveyyue:
https://issues.redhat.com/browse/DBZ-5966


jpechane:
@harveyyue Thanks for the PR. TBH I don't think we need a new config option. If the user is not interested in truncates then it can disable them at connector level are use a filter SMT. So just being able to ignore them should be enough.

harveyyue:
> 
Agree, drop it now.

jpechane:
@harveyyue Applied, thanks a lot.

PR #4176, opened 2023-01-06
DBZ-4663 Remove option for specifying driver class from MySQL Connector

harveyyue:
https://issues.redhat.com/browse/DBZ-4663

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnection.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnection.java
@@ -23,6 +23,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.mysql.cj.CharsetMapping;
+import com.mysql.cj.jdbc.Driver;
 
 import io.debezium.DebeziumException;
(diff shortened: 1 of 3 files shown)


Naros (approved):
I do not see that `database.jdbc.driver` is referenced in the documentation anywhere, so this LGTM.  Thanks @harveyyue!

Naros:
Applied to `main` (=2.2).

rk3rn3r:
Awesome! Thx a lot @harveyyue!

som-design (commented):
debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlConnectorTest.java

gnought:
@Naros @harveyyue could we revert the change to keep the `database.jdbc.driver` option or provide another way to let users specify other JDBC drivers?
We are using v1.6.4 in production that requires a AWS JDBC driver https://github.com/awslabs/aws-mysql-jdbc for IAM authentication connection.

jpechane:
@gnought Thanks for raising this up! We need to revert it ifn that case.

One question, do you know if the same functionality might be needed for other connectors?

jpechane:
@gnought Reverting via https://github.com/debezium/debezium/pull/4247

gnought:
@jpechane perfect! Thanks.

PR #4177, opened 2023-01-06
[docs] Update Infinispan config example to ISPN 14.0

vjuranek:


Naros:
Applied to `main` and branch `2.1`, where ISPN 14 was applied.

PR #4178, opened 2023-01-06
DBZ-5907 Fix Oracle transaction event undo

Naros:
https://issues.redhat.com/browse/DBZ-5907

This PR correctly performs transaction event undo by only reversing the last event in the transaction event queue that matches the row-id, rather than undoing all events with the same row-id.

@jpechane I would suggest we consider backporting this to 2.0 and/or 2.1.


jpechane:
@Naros Applied, thanks.

jpechane:
@Naros 2.1 is the new stable so I backported it only to that version.

PR #4179, opened 2023-01-07
DBZ-5984 Remove redundant Java keywords

erdinctaskin:


jpechane:
Thanks for the PR. There is https://github.com/debezium/debezium/pull/4045 that should incorporate this one.

PR #4180, opened 2023-01-09
[docs] Update the postgres container initialisation example

jakzal:
The container needs to be marked as a compatible substitute for the official postgres image. 

Otherwise, with recent versions of testcontainers, the postgres container will fail to initialise with: 

```Failed to verify that image 'debezium/postgres:11' is a compatible substitute for 'postgres'.```

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -162,6 +162,7 @@ Guy Korland
 Guy Pascarella
 Grzegorz Kołakowski
 Jacob Gminder
+Jakub Zalas
 Jan Doms
 Jan Hendrik Dolling
 Jan Werner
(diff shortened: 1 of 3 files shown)


jpechane:
@jakzal Thanks for the PR.
@obabec Could you please try to upgrade TestContainers in our project and check if the source code change is needed? If yes we need to create a new Jira and make a fix to make sure this requirement is not lost.

jakzal:
@jpechane no problem. Same PR or a new one?

jpechane:
@jakzal Thanks for checking, applied! I just linked the version update commit into its own Jira - https://issues.redhat.com/browse/DBZ-5990

PR #4181, opened 2023-01-09
DBZ-5986 Add server sink adapter for Infinispan

vjuranek:
https://issues.redhat.com/browse/DBZ-5986

### DIFF
--- a/debezium-server/debezium-server-infinispan/src/main/java/io/debezium/server/infinispan/InfinispanSinkConsumer.java
+++ b/debezium-server/debezium-server-infinispan/src/main/java/io/debezium/server/infinispan/InfinispanSinkConsumer.java
@@ -1,0 +1,128 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.server.infinispan;
+
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import javax.annotation.PostConstruct;
+import javax.annotation.PreDestroy;
+import javax.enterprise.context.Dependent;
+import javax.enterprise.inject.Instance;
+import javax.inject.Inject;
+import javax.inject.Named;
+
+import org.eclipse.microprofile.config.Config;
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.infinispan.client.hotrod.RemoteCache;
+import org.infinispan.client.hotrod.RemoteCacheManager;
+import org.infinispan.client.hotrod.configuration.ConfigurationBuilder;
+import org.infinispan.client.hotrod.impl.ConfigurationProperties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.debezium.DebeziumException;
+import io.debezium.engine.ChangeEvent;
+import io.debezium.engine.DebeziumEngine;
+import io.debezium.server.BaseChangeConsumer;
+import io.debezium.server.CustomConsumerBuilder;
+
+/**
+ * An implementation of the {@link DebeziumEngine.ChangeConsumer} interface that publishes change event messages to predefined Infinispan cache.
+ *
+ * @author vjuranek
+ */
+@Named("infinispan")
+@Dependent
+public class InfinispanSinkConsumer extends BaseChangeConsumer implements DebeziumEngine.ChangeConsumer<ChangeEvent<Object, Object>> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(InfinispanSinkConsumer.class);
+
+    private static final String CONF_PREFIX = "debezium.sink.infinispan.";
+    private static final String SERVER_HOST = CONF_PREFIX + "server.host";
+    private static final String SERVER_PORT = CONF_PREFIX + "server.port";
+    private static final String CACHE_NAME = CONF_PREFIX + "cache";
+    private static final String USER_NAME = CONF_PREFIX + "user";
+    private static final String PASSWORD = CONF_PREFIX + "password";
+
+    private RemoteCacheManager remoteCacheManager;
+    private RemoteCache cache;
+
+    @Inject
+    @CustomConsumerBuilder
+    Instance<RemoteCache> customCache;
+
+    @PostConstruct
+    void connect() {
+        if (customCache.isResolvable()) {
+            cache = customCache.get();
+            LOGGER.info("Obtained custom cache with configuration '{}'", cache.getRemoteCacheContainer().getConfiguration());
+            return;
+        }
+
+        final Config config = ConfigProvider.getConfig();
+        final String serverHost = config.getValue(SERVER_HOST, String.class);
+        final String cacheName = config.getValue(CACHE_NAME, String.class);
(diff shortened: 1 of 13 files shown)


debezium-server/debezium-server-infinispan/src/main/java/io/debezium/server/infinispan/InfinispanSinkConsumer.java
```diff
@@ -0,0 +1,128 @@
...
+    private static final String CACHE_NAME = CONF_PREFIX + "cache";
+    private static final String USER_NAME = CONF_PREFIX + "user";
+    private static final String PASSWORD = CONF_PREFIX + "password";
+
+    private RemoteCacheManager remoteCacheManager;
+    private RemoteCache cache;
+
+    @Inject
+    @CustomConsumerBuilder
+    Instance<RemoteCache> customCache;
```
jpechane, line 58: Could you please check other sinks? I am a bit unsure if this should provide `RemoteCache` or `RemoteCacheManager` is more logical.

debezium-server/debezium-server-infinispan/src/main/java/io/debezium/server/infinispan/InfinispanSinkConsumer.java
```diff
@@ -0,0 +1,128 @@
...
+        }
+    }
+
+    @Override
+    public void handleBatch(List<ChangeEvent<Object, Object>> records, DebeziumEngine.RecordCommitter<ChangeEvent<Object, Object>> committer)
+            throws InterruptedException {
+        Map<Object, Object> entries = new HashMap<>(records.size());
+        for (ChangeEvent<Object, Object> record : records) {
+            if (record.value() != null) {
+                LOGGER.debug("Received event {} = '{}'", getString(record.key()), getString(record.value()));
```
jpechane, line 110: Should be `trace` level

debezium-server/debezium-server-infinispan/src/test/java/io/debezium/server/infinispan/InfinispanTestResourceLifecycleManager.java
```diff
@@ -0,0 +1,70 @@
...
+
+import org.infinispan.client.hotrod.impl.ConfigurationProperties;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.testcontainers.containers.BindMode;
+import org.testcontainers.containers.GenericContainer;
+
+import io.quarkus.test.common.QuarkusTestResourceLifecycleManager;
+
+public class InfinispanTestResourceLifecycleManager implements QuarkusTestResourceLifecycleManager {
```
jpechane, line 20: One question - is it better to use container ot start the server in-process?
Also could be the tag in the image name synchronized with Infinispan version?

documentation/modules/ROOT/pages/operations/debezium-server.adoc
```diff
@@ -1142,6 +1142,67 @@ By default the same name is used.
...
+|CDI classifier
+|Description
+
+|[[infinispan-ext-hotrod-cache]]<<infinispan-ext-hotrod-cache, `org.infinispan.client.hotrod.RemoteCache`>>
+|`@CustomConsumerBuilder`
+|Custom instance of https://docs.jboss.org/infinispan/14.0/apidocs/org/infinispan/client/hotrod/RemoteCache.html[Hot Rod cache] which will be used for connecting and sending events to the Infinspan cluster.
+
+|[[infinispan-ext-stream-name-mapper]]<<infinispan-ext-stream-name-mapper, `io.debezium.server.StreamNameMapper`>>
+|
+|Custom implementation maps the planned destination (topic) name into a physical Kinesis stream name.
```
jpechane, line 1201: Not Kinesis

debezium-server/debezium-server-infinispan/src/main/java/io/debezium/server/infinispan/InfinispanSinkConsumer.java
```diff
@@ -0,0 +1,128 @@
...
+    void connect() {
+        if (customCache.isResolvable()) {
+            cache = customCache.get();
+            LOGGER.info("Obtained custom cache with configuration '{}'", cache.getRemoteCacheContainer().getConfiguration());
+            return;
+        }
+
+        final Config config = ConfigProvider.getConfig();
+        final String serverHost = config.getValue(SERVER_HOST, String.class);
+        final String cacheName = config.getValue(CACHE_NAME, String.class);
```
jpechane, line 70: cache name is mandatory but what if it is made optional and when not provided then topic name is used as the cache name so you can feed multiple caches? Alo in that case the custom stream name mapper can/should be used as well.

jpechane (changes requested):
@vjuranek Thanks a lot. I left few comments related to just a cosmetic change and one larger related to the cache use - single vs multiple.

debezium-server/debezium-server-infinispan/src/main/java/io/debezium/server/infinispan/InfinispanSinkConsumer.java
vjuranek: This would depend if we want to support multiple caches or not. In case of multiple caches `RemoteCacheManager` should be use, in case of a single cache (current state), it's IMHO better to use `RemoteCache`. `RemoteCacheManager` could be used also in this case, but would make the configuration more complicated as the user would have to provide also cache name.

debezium-server/debezium-server-infinispan/src/main/java/io/debezium/server/infinispan/InfinispanSinkConsumer.java
vjuranek: done

debezium-server/debezium-server-infinispan/src/test/java/io/debezium/server/infinispan/InfinispanTestResourceLifecycleManager.java
vjuranek: > One question - is it better to use container ot start the server in-process? 

IMHO better to use the container - it's more close to real use case and also I'm actually not sure if connecting to Infinispan started in library mode via hotrod is supported (or even possible).

> Also could be the tag in the image name synchronized with Infinispan version?

ops, fixed

documentation/modules/ROOT/pages/operations/debezium-server.adoc
vjuranek: fixed -  `StreamNameMapper` actually shouldn't be there at all

debezium-server/debezium-server-infinispan/src/main/java/io/debezium/server/infinispan/InfinispanSinkConsumer.java
vjuranek: Most of the other sink have a notion of stream or topic where which you can specify along with the records and it's handled on the server. This is not the case of Infinispan - you have to select the cache where you want to send the records and there is no possibility to just send the the records along with the cache names and tell Infinispan to sort it out before storing. If we want to support it, we would need to create multiple `RemoteCache`s, define the mapping and sort and send the records to respective cache ourselves. I'd say more typical Infinispan usage is to use one/few big caches rather than many small caches, so I'd support only one cache for now. We can extend it in the future if there will be any demand for it.

jpechane:
@vjuranek Applied, thanks a lot! I just unified the Infinispan version management between Oracle and Server sink.

PR #4182, opened 2023-01-10
DBZ-5988:Impose null check on schema value prior to any operation

akanimesh7:
This fixes the NPE which arises when we try to send a `execute-snapshot` signal with an invalid table name (rather than schema.table only give table). This only happens in the case of using `exclude.tables` config.

The fix for this is to check that the schema name and table name are not null before applying the lowercase operation on them.


jpechane:
@akanimesh7 Applied, thanks

PR #4184, opened 2023-01-10
DBZ-5993 Remove superfluous placeholder in log message (MySqlDatabaseSchema)

joschi:
The warning message emitted when the Debezium Connector for MySQL encounters an unparseable DDL statement contains a superfluous placeholder.

Refs https://issues.redhat.com/browse/DBZ-5993

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlDatabaseSchema.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlDatabaseSchema.java
@@ -214,7 +214,7 @@ public class MySqlDatabaseSchema extends HistorizedRelationalDatabaseSchema {
         }
         catch (ParsingException | MultipleParsingExceptions e) {
             if (schemaHistory.skipUnparseableDdlStatements()) {
-                LOGGER.warn("Ignoring unparseable DDL statement '{}': {}", ddlStatements, e);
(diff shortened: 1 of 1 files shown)


jpechane:
@joschi Applied, thanks a lot

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlDatabaseSchema.java
```diff
@@ -214,7 +214,7 @@ private List<SchemaChangeEvent> parseDdl(MySqlPartition partition, String ddlSta
         }
         catch (ParsingException | MultipleParsingExceptions e) {
             if (schemaHistory.skipUnparseableDdlStatements()) {
-                LOGGER.warn("Ignoring unparseable DDL statement '{}': {}", ddlStatements, e);
+                LOGGER.warn("Ignoring unparseable DDL statement '{}'", ddlStatements, e);
```
rk3rn3r, line 217: @jpechane didn't this remove printing the exception? :thinking:

joschi: > didn't this remove printing the exception?

@rk3rn3r No. If the last parameter is a `Throwable`, SLF4J (and Logback) will use it to print the exception cause/stack trace.

> The SLF4J API supports parametrization (of 1 or more objects) in the presence of an exception, assuming the exception is the last parameter.

https://www.slf4j.org/faq.html#paramException

PR #4185, opened 2023-01-10
DBZ-5985 Disable to print invalid estimated row count when config "snapshot.select.statement.overrides"

harveyyue:
https://issues.redhat.com/browse/DBZ-5985


jpechane:
@harveyyue Applied, thanks!

PR #4186, opened 2023-01-10
DBZ-5996 Debezium truncating micro/nanosecond part if it is all zeros

akshanshjain95:
# Bug report

### What Debezium connector do you use and what version?
Connector: MySqlSourceConnector
Version: 1.9.7.FINAL

### What is the connector configuration?
time.precision.mode is default.

### What is the captured database version and mode of deployment?
MySql 8.0.30 Source distribution
AWS RDS.

### What behaviour do you see and expect?
Database has field with type as timestamp(6). When debezium produces the field's value to kafka, it truncates any leading zeros there might be in the nano/microsecond part. If the value contains only zeros in the nano/microsecond part, then the whole part is excluded.

For ex.

2023-01-11 12:30:00.123123 --> 2023-01-11T12:30:00.123123Z

2023-01-11 12:30:00.123000 --> 2023-01-11T12:30:00.123Z

2023-01-11 12:30:00.000000 --> 2023-01-11T12:30:00Z

This breaks the sink connectors which expect the date time type to be in the format yyyy-MM-dd'T'HH:mm:ss.SSSSSS'Z'. 

Expected was that the leading zeros should not get truncated, so that all the datetimes have consistent format.

2023-01-11 12:30:00.123123 --> 2023-01-11T12:30:00.123123Z

2023-01-11 12:30:00.123000 --> 2023-01-11T12:30:00.123000Z

2023-01-11 12:30:00.000000 --> 2023-01-11T12:30:00000Z

### Do you see the same behaviour using the latest relesead Debezium version?
Yes.

Issue link: https://issues.redhat.com/browse/DBZ-5996

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -8,6 +8,7 @@ Adham
 Adrian Kreuziger
 Ahmed Eljami
 Aidas
+Akshansh Jain
 Akshath Patkar
 Alberto Martino
 Aleksejs Sibanovs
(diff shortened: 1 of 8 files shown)


jpechane:
@akshanshjain95 Applied, thanks!

PR #4189, opened 2023-01-12
DBZ-5962 Sink adapter for Apache RocketMQ

sunxiaojian:
Continues work from #4161


debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
```diff
@@ -0,0 +1,163 @@
...
+                                throw new DebeziumException(throwable);
+                            }
+                        });
+                committer.markProcessed(record);
+            }
+            catch (Exception e) {
+                throw new DebeziumException(e);
+            }
+        }
+        latch.await();
```
vjuranek, line 160: I'd consider to place a comment here that it cannot block forever as there is a timeout defined for message delivery.

vjuranek (changes requested):
The is still one issue with marking messages as processed which IMHO needs to be fixed, otherwise LGTM.
If you do any changes, please it in this PR and don't close this PR and open new one, otherwise the conversation will be lost/hard to find.

vjuranek:
For the reference, part of the conversation is under PR #4161

vjuranek:
Most of the test failures are unrelated, but `RocketMqIT` also [fails](https://github.com/debezium/debezium/actions/runs/3899997180/jobs/6663444115)

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
```diff
@@ -0,0 +1,163 @@
...
+                                latch.countDown();
+                            }
+
+                            @Override
+                            public void onException(Throwable throwable) {
+                                LOGGER.error("Failed to send record to {}:", record.destination(), throwable);
+                                throw new DebeziumException(throwable);
+                            }
+                        });
+                committer.markProcessed(record);
```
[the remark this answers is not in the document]
sunxiaojian: > > Calling onSuccess() may cause the offset submission to be out of order
> 
> right, calling it in `onSuccess` may result into out of order offsets. However, calling it here is also wrong as it will mark as processed messages that are not delivered yet and the delivery may fail. So you probably need special loop over the records after `latch.await();` where you marked them as processed once you are sure that they were really delivered.

I have modified it

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
sunxiaojian: > I'd consider to place a comment here that it cannot block forever as there is a timeout defined for message delivery.

Producer has a timeout. Producer will throw an exception if it sends a timeout or fails. This will cause the task to terminate. If no exception is thrown, then everything is normal. I think we need to wait until we succeed

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
```diff
@@ -0,0 +1,191 @@
...
+        if (mqProducer != null) {
+            mqProducer.shutdown();
+        }
+    }
+
+    @Override
+    public void handleBatch(List<ChangeEvent<Object, Object>> records, DebeziumEngine.RecordCommitter<ChangeEvent<Object, Object>> committer)
+            throws InterruptedException {
+        final CountDownLatch latch = new CountDownLatch(records.size());
+        Deque<SubmittedRecord> submittedRecords = new LinkedList<>();
```
vjuranek, line 138: Is this needed? You already have a list of records in `records`.

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
```diff
@@ -0,0 +1,191 @@
...
+            LOGGER.warn("No records to mark processed!");
+            return;
+        }
+        long markProcessedCount = 0;
+        while (submittedRecords.peek() != null && submittedRecords.peek().getAcked().get()) {
+            committer.markProcessed(submittedRecords.poll().getRecord());
+            ++markProcessedCount;
+        }
+
+        if (submittedRecords.peek() != null && !submittedRecords.peek().getAcked().get()) {
```
vjuranek, line 184: If you decide to wait for all event, this IMHO cannot happen as message is either delivered (ack is true) or exception is throws before you get to this point (you will be still waiting in `latch.await();`). Or am I missing something?

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
```diff
@@ -0,0 +1,191 @@
...
+                submittedRecords.add(submittedRecord);
+            } catch (Exception e) {
+                // Ensure that the data has been sent successfully and is marked as processed
+                internalMarkProcessed(submittedRecords, committer);
+                throw new DebeziumException(e);
+            }
+        }
+        latch.await();
+        // Mark processed
+        internalMarkProcessed(submittedRecords, committer);
```
vjuranek, line 166: why not simply
```
for (ChangeEvent<Object, Object> record : records) {
    committer.markProcessed(record);
}
```

vjuranek (changes requested):

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
sunxiaojian: SubmittedRecord is to ensure the order

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
sunxiaojian: > why not simply
> 
> ```
> for (ChangeEvent<Object, Object> record : records) {
>     committer.markProcessed(record);
> }
> ```

Of course, I just reuse the same method

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
sunxiaojian: > If you decide to wait for all event, this IMHO cannot happen as message is either delivered (ack is true) or exception is throws before you get to this point (you will be still waiting in `latch.await();`). Or am I missing something?

I think it is difficult to control the timeout. So I think it's enough to throw an exception and let us observe it

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
vjuranek: you have already this order in `records` list, no need to repeat it

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
vjuranek: I'm really no sure we need it - your original version was just fine (besides wrong place of `committer.markProcessed()`), I was missing the point that async message has default timeout for sending, so that `latch.await()` won't block forever. So IMHO this would work well:

```
// messages have set default timeout set to send.msg.timeout, so this cannot block forever
latch.await()

for (ChangeEvent<Object, Object> record : records) {
    committer.markProcessed(record);
}
```

and there's no need to copy the records to `SubmittedRecord` and the code more complicated.

vjuranek (changes requested):

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
sunxiaojian: > I'm really no sure we need it - your original version was just fine (besides wrong place of `committer.markProcessed()`), I was missing the point that async message has default timeout for sending, so that `latch.await()` won't block forever. So IMHO this would work well:
> 
> ```
> // messages have set default timeout set to send.msg.timeout, so this cannot block forever
> latch.await()
> 
> for (ChangeEvent<Object, Object> record : records) {
>     committer.markProcessed(record);
> }
> ```
> 
> and there's no need to copy the records to `SubmittedRecord` and the code more complicated.

If there are [1,2,3,4] 4 records, we have successfully sent 1,2,4, and 3 has failed to submit, do we just need to mark that it has been processed to 2? The new queue is to ensure that the data of 3 will not be missed in the next startup

vjuranek: AFAICT `markProcessed()` just updates in-memory state, actual flush is done in `markBatchFinished()`. So if you fail to deliver message number 3, you will throw an exception before flushing the offset to the disk, in-memory state is lost and next time will start from 1 again (stored offset), so nothing is lost. You get few duplicated messages, but this is in line with Debezium at-least-one delivery promise. I.e. if you fail, it IMHO doesn't matter much if you called `markProcessed()` or not (but I'd be very carefull when it's called in success case - it should be called only when we are sure message was delivered).

sunxiaojian: > AFAICT `markProcessed()` just updates in-memory state, actual flush is done in `markBatchFinished()`. So if you fail to deliver message number 3, you will throw an exception before flushing the offset to the disk, in-memory state is lost and next time will start from 1 again (stored offset), so nothing is lost. You get few duplicated messages, but this is in line with Debezium at-least-one delivery promise. I.e. if you fail, it IMHO doesn't matter much if you called `markProcessed()` or not (but I'd be very carefull when it's called in success case - it should be called only when we are sure message was delivered).

According to this logic, my original logic is correct and does not need to be modified; The current way is to narrow the scope of duplicate data

sunxiaojian: > AFAICT `markProcessed()` just updates in-memory state, actual flush is done in `markBatchFinished()`. So if you fail to deliver message number 3, you will throw an exception before flushing the offset to the disk, in-memory state is lost and next time will start from 1 again (stored offset), so nothing is lost. You get few duplicated messages, but this is in line with Debezium at-least-one delivery promise. I.e. if you fail, it IMHO doesn't matter much if you called `markProcessed()` or not (but I'd be very carefull when it's called in success case - it should be called only when we are sure message was delivered).

markProcessed() is only updated in memory status, but it also needs to be updated in order, because the same partition will be overwritten in memory. Moving markProcessed() to onSuccess() will cause disorder, and the final offset may not be what we want

vjuranek: I already mentioned, that moving `markProcessed()` to `onSuccess()` was wrong proposal and you are right that it will mess up the order. But I still believe that you should `markProcessed()` only when you are sure that the messages was delivered, i.e. after `latch.await()`

debezium-server/debezium-server-rocketmq/src/main/java/io/debezium/server/rocketmq/RocketMqChangeConsumer.java
```diff
@@ -0,0 +1,160 @@
...
+                        latch.countDown();
+                    }
+
+                    @Override
+                    public void onException(Throwable throwable) {
+                        LOGGER.error("Failed to send record to {}:", record.destination(), throwable);
+                        throw new DebeziumException(throwable);
+                    }
+                });
+                committer.markProcessed(record);
```
vjuranek, line 151: >> I already mentioned, that moving markProcessed() to onSuccess() was wrong proposal and you are right that >> it will mess up the order. But I still believe that you should markProcessed() only when you are sure that the >> messages was delivered, i.e. after latch.await()

> OK, I have rollbacked to the original version

as mentioned above, I still believe that you should call `markProcessed()` only when you are sure that the messages was delivered, i.e. after `latch.await()`. If you want to do it this way, are you sure it cannot cause any issue while waiting for `latch.await()` (i.e. during this time Kafka will think that these events are processed)?

sunxiaojian: > it cannot cause any issue

You are right. Although the logic is simple, it may still cause some problems

sunxiaojian: > > > I already mentioned, that moving markProcessed() to onSuccess() was wrong proposal and you are right that >> it will mess up the order. But I still believe that you should markProcessed() only when you are sure that the >> messages was delivered, i.e. after latch.await()
> 
> > OK, I have rollbacked to the original version
> 
> as mentioned above, I still believe that you should call `markProcessed()` only when you are sure that the messages was delivered, i.e. after `latch.await()`. If you want to do it this way, are you sure it cannot cause any issue while waiting for `latch.await()` (i.e. during this time Kafka will think that these events are processed)?

You are right. Although the logic is simple, it may still cause some problems

vjuranek (approved):

debezium-server/debezium-server-bom/pom.xml
```diff
@@ -203,13 +208,24 @@
...
             </dependency>
+            <dependency>
+                <groupId>org.apache.rocketmq</groupId>
+                <artifactId>rocketmq-client</artifactId>
+                <version>${version.rocketmq}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.rocketmq</groupId>
+                <artifactId>rocketmq-tools</artifactId>
+                <version>${version.rocketmq}</version>
```
vjuranek, line 219: IMHO you need to put exclusion of `commons-cli` here - there is a clash with quarkus dependencies
```
                <exclusions>
                    <exclusion>
                        <groupId>commons-cli</groupId>
                        <artifactId>commons-cli</artifactId>
                    </exclusion>
                </exclusions>
```

vjuranek (changes requested):
There is a clash with quarkus dependencies, so I guess you have to remove `commons-cli` transitive dependency. Also, when I run it locally, the test failed as rocketmq seems to forbid `.` in topics name, so `testc.inventory.customers` is not allowed topics name, I got this exception:
```
org.apache.rocketmq.client.exception.MQClientException: The specified topic[testc.inventory.customers] contains illegal characters, allowing only ^[%|a-zA-Z0-9_-]+$
For more information, please visit the url, http://rocketmq.apache.org/docs/faq/
        at org.apache.rocketmq.client.Validators.checkTopic(Validators.java:90)
        at org.apache.rocketmq.client.Validators.checkMessage(Validators.java:61)
        at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl.sendSelectImpl(DefaultMQProducerImpl.java:1122)
        at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl.access$600(DefaultMQProducerImpl.java:98)
        at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl$5.run(DefaultMQProducerImpl.java:1189)
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
        at java.base/java.lang.Thread.run(Thread.java:833)
```
so I think you either need to enable it (if it's possible), or use [topics re-routing](https://debezium.io/documentation/reference/stable/transformations/topic-routing.html)

sunxiaojian:
> There is a clash with quarkus dependencies, so I guess you have to remove `commons-cli` transitive dependency. Also, when I run it locally, the test failed as rocketmq seems to forbid `.` in topics name, so `testc.inventory.customers` is not allowed topics name, I got this exception:
> 
> ```
> org.apache.rocketmq.client.exception.MQClientException: The specified topic[testc.inventory.customers] contains illegal characters, allowing only ^[%|a-zA-Z0-9_-]+$
> For more information, please visit the url, http://rocketmq.apache.org/docs/faq/
>         at org.apache.rocketmq.client.Validators.checkTopic(Validators.java:90)
>         at org.apache.rocketmq.client.Validators.checkMessage(Validators.java:61)
>         at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl.sendSelectImpl(DefaultMQProducerImpl.java:1122)
>         at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl.access$600(DefaultMQProducerImpl.java:98)
>         at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl$5.run(DefaultMQProducerImpl.java:1189)
>         at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
>         at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
>         at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
>         at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
>         at java.base/java.lang.Thread.run(Thread.java:833)
> ```
> 
> so I think you either need to enable it (if it's possible), or use [topics re-routing](https://debezium.io/documentation/reference/stable/transformations/topic-routing.html)

resolved

sunxiaojian:
> There is a clash with quarkus dependencies, so I guess you have to remove `commons-cli` transitive dependency. Also, when I run it locally, the test failed as rocketmq seems to forbid `.` in topics name, so `testc.inventory.customers` is not allowed topics name, I got this exception:
> 
> ```
> org.apache.rocketmq.client.exception.MQClientException: The specified topic[testc.inventory.customers] contains illegal characters, allowing only ^[%|a-zA-Z0-9_-]+$
> For more information, please visit the url, http://rocketmq.apache.org/docs/faq/
>         at org.apache.rocketmq.client.Validators.checkTopic(Validators.java:90)
>         at org.apache.rocketmq.client.Validators.checkMessage(Validators.java:61)
>         at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl.sendSelectImpl(DefaultMQProducerImpl.java:1122)
>         at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl.access$600(DefaultMQProducerImpl.java:98)
>         at org.apache.rocketmq.client.impl.producer.DefaultMQProducerImpl$5.run(DefaultMQProducerImpl.java:1189)
>         at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:539)
>         at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:264)
>         at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1136)
>         at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:635)
>         at java.base/java.lang.Thread.run(Thread.java:833)
> ```
> 
> so I think you either need to enable it (if it's possible), or use [topics re-routing](https://debezium.io/documentation/reference/stable/transformations/topic-routing.html)

resolved

vjuranek (approved):
CI failures are unrelated

vjuranek:
Merged. Thanks a lot for your contribution and also for your patience with my reviews:-)

PR #4193, opened 2023-01-12
add local time zone for convert local date time to unix time

xie-shujian:
Debezium convert local date time to the UTC unix time. It is not right when databse use local timezone, like mysql use chinese time.
Add an environment variable called LOCAL_TIMEZONE for convet the local date time to local unix time.
You can set LOCAL_TIMEZONE="+08:00", then the long value of date time is the right chinese time.
If you not set anything, it use UTC.


jpechane:
@xie-shujian Thanks for the PR. Unfrotunately I don't think this should be done this way. Could you please

* create a Jira issue for design discussion
* change approach to use connector configuration option instead of env variable
* provide a test case
* make sure that works for all connectors

Now there is an open question if time datatype should be also affected etc.

Naros:
@xie-shujian when you configure Debezium, are you setting `database.connectionTimeZone` (or `driver.connectionTimeZone` when using Debezium 2.x+)?

PR #4195, opened 2023-01-12
DBZ-6003 Nullable columns marked with "optional: false" in DDL events

harveyyue:
https://issues.redhat.com/browse/DBZ-6003

### DIFF
--- a/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlAntlrDdlParserTest.java
+++ b/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlAntlrDdlParserTest.java
@@ -87,6 +87,102 @@ public class MySqlAntlrDdlParserTest {
     }
 
     @Test
+    @FixFor("DBZ-6003")
+    public void shouldProcessCreateUniqueBeforePrimaryKeyDefinitions() {
+        String ddl = "CREATE TABLE `dbz_6003_1` (\n"
+                + "`id` INT ( 11 ) NOT NULL,\n"
+                + "`name` VARCHAR ( 255 ),\n"
+                + "UNIQUE KEY `test` ( `name` ),\n"
+                + "PRIMARY KEY ( `id` )\n"
+                + ") ENGINE = INNODB;";
+        parser.parse(ddl, tables);
+        assertThat(((MySqlAntlrDdlParser) parser).getParsingExceptionsFromWalker().size()).isEqualTo(0);
+        assertThat(tables.size()).isEqualTo(1);
+        Table table = tables.forTable(null, null, "dbz_6003_1");
+        assertThat(table.primaryKeyColumnNames().size()).isEqualTo(1);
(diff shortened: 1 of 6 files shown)


jpechane:
@harveyyue LGTM, thanks. Stil I'd like to see wider test coverage for `ALTER TABLE` too.
Ideally there should be cases tested (if I read grammer correctly each case should be achievable with a single `ALTER`)

* added `PRIMARY KEY` and `UNIQUE KEY`
* added `UNIQUE KEY` and `PRIMARY KEY`
* add `UNIQUE KEY`, drop `PRIMARY KEY`
* drop `PRIMARY KEY`, add `UNIQUE KEY`
* drop `UNIQUE KEY`, add `PRIMARY KEY`
* add `PRIMARY KEY`, drop `UNIQUE KEY`
* drop both `PRIMARY` and `UNIQUE` keys

harveyyue:
@jpechane Ok, will add these test cases.

harveyyue:
@jpechane Done with these test cases. BTW, I add a separated method to parse the unique key columns, this method will set the required unique key columns to primary keys if there is no primary keys. It will handle following case:
`
ddl 1: create table dbz_6003_2 (id int not null, name varchar(255));
ddl 2: alter table dbz_6003_2 add unique key dbz_6003_2_uk(name);
ddl 3: alter table dbz_6003_2 add primary key(id);
`
Parse result:
primary key is id, the column name is optional.

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlAntlrDdlParserTest.java
```diff
@@ -86,6 +86,102 @@ public void beforeEach() {
...
+        assertThat(table.primaryKeyColumnNames().get(0)).isEqualTo("id");
+        assertThat(table.columnWithName("name").isOptional()).isTrue();
+
+        // add UNIQUE KEY, drop PRIMARY KEY
+        ddl = "create table dbz_6003_4 (id int not null, name varchar(255), primary key(id));"
+                + "alter table dbz_6003_4 add unique key dbz_6003_4_uk(name);"
+                + "alter table dbz_6003_4 drop primary key;";
+        parser.parse(ddl, tables);
+        table = tables.forTable(null, null, "dbz_6003_4");
+        assertThat(table.primaryKeyColumnNames().size()).isEqualTo(0);
```
jpechane, line 129: @harveyyue  And here we get into an interesting problem, if you check the next case it is the same just different order of execution yet the result here is there is no key and in the following there is. General behaviour should be that if there is any change to either primary key or unique key then

- if PK exists it should be used
- if unique key exists it should be used

harveyyue:
> Or it i not used only because `name` is not `NOT NULL`?
Yes, the unique key need the column is not null in the separated method `parseUniqueIndexColumnNames`.

jpechane:
@harveyyue Applied, thanks

PR #4197, opened 2023-01-12
Upgrade to Quarkus 2.16.0.Final

Naros:
https://issues.redhat.com/browse/DBZ-6005

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -112,7 +112,7 @@
 
         <!-- Quarkus -->
         <!-- Must be aligned with Apicurio version below -->
-        <quarkus.version>2.16.0.CR1</quarkus.version>
+        <quarkus.version>2.16.0.Final</quarkus.version>
 
         <!-- Apicurio -->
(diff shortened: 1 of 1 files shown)


jpechane:
@Naros Looking at https://debezium.zulipchat.com/#narrow/stream/302529-users/topic/Debezium.20embedded.20error Does it make sense to align antlr too?

Naros:
> @Naros Looking at [debezium.zulipchat.com/#narrow/stream/302529-users/topic/Debezium.20embedded.20error](https://debezium.zulipchat.com/#narrow/stream/302529-users/topic/Debezium.20embedded.20error) Does it make sense to align antlr too?

I amended the upgrade to 2.16.0.CR1 to include the ANTLR bump to 4.10.1.  I've also added a commit that explicitly bumps to the 2.16.0.Final artifacts of Quarkus; so if the CI passes, I think this is now good to go.

jpechane:
@Naros Applied, thanks!

PR #4198, opened 2023-01-12
DBZ-5998 Use github context instead of hard coded values in CI

ani-sha:
https://issues.redhat.com/browse/DBZ-5998


jpechane:
@ani-sha Excellent!

PR #4199, opened 2023-01-12
DBZ-5975 Remove note that indicates SMT is under active development.

roldanbob:
[DBZ-5975](https://issues.redhat.com/browse/DBZ-5975)

Brings the MongoDB version of the SMT level with the SQL-based version by removing the statement that implies that the SMT is still incubating. 


jpechane:
@roldanbob Applied, thanks.

PR #4202, opened 2023-01-14
DBZ-5879 Support retrying database connection failures during connector start

jeremy-l-ford:
Reworked when task requests start from subclasses
Add support for restarting


jpechane (approved):
@jeremy-l-ford LGTM, thanks! I added a commit with additional logging and removed dead code. When CI passes the PR can be merged.

PR #4203, opened 2023-01-16
DBZ-5071 Add support for providing a surrogate key when triggering incremental snapshots

PlugaruT:
This PR is the follow up of #3462 

For tables that have a composite Primary Key, SQL queries for incremental snaphots perform bad, see Jira ticket. This change allows to provide a surrogate key that will be used for the queries. The key should be unique in order for the snapshot to guarantee consistency.

Notes:
 - MongoDB won't support this feature
 - MySQL Read-Only snapshots don't need support for this? Not sure tho. Let me know if I should add support for it. 
 - How the configuration should look when in the same signal, multiple tables are provided? I know right now the configuration will be taken into account for all tables, but maybe we want per-table configuration? Or in that case, should the user just push multiple signals?

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbIncrementalSnapshotChangeEventSource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbIncrementalSnapshotChangeEventSource.java
@@ -334,11 +334,16 @@ public class MongoDbIncrementalSnapshotChangeEventSource
     @Override
     @SuppressWarnings("unchecked")
     public void addDataCollectionNamesToSnapshot(MongoDbPartition partition, List<String> dataCollectionIds,
-                                                 Optional<String> additionalCondition, OffsetContext offsetContext)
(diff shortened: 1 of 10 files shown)


PlugaruT:
@jpechane @Naros guys, let me know what you think of the changes and if the direction is correct. If yes, I will continue adding tests and fixing what is broken. Thanks.

jpechane:
@PlugaruT I think this is definitely good feature. I'd prefer to have a separate Jira create for it as it is not a general solution for the original Jira yet can be a useful feature.
For your questions

* MongoDB - agree with that
* MySQL read-only - it should support it too. As the mechanism is the same it will sruggle with the same problem
* Let's keep it in the same way as `additional-condition`, it is also not per-table. When needed we can in future introduce a support for complex mapping with per-table `additional-condition` and `surrogate-key`

PlugaruT:
> I'd prefer to have a separate Jira create for it as it is not a general solution for the original Jira yet can be a useful feature.

I will then create a new Jira issue and link the two together and will edit the commit message and everything then. Thanks for the remarks, I'll continue working on it then.

PlugaruT:
Closing this in favour of the new PR and new Jira ticket.

PR #4206, opened 2023-01-18
DBZ-5917 Fix querying column metadata for tables with back slash in t…

vjuranek:
…he name

If the table contains back slash, which is ANSI SQL escape chracter, in its name, querying column metadata would fail in some cases (*) which would result into NPE. Fix table name before the query and escape escape character if the table name contains it.

Default backslash works for all currently supported databases as it's ANSI SQL standard, but in case of need it escape  methods can be overridded respective subclasses.

(*) everything works when we collect column metadata for all tables, i.e. when there are no excluded tables

https://issues.redhat.com/browse/DBZ-5917

### DIFF
--- a/debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
+++ b/debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
@@ -80,6 +80,7 @@ public class JdbcConnection implements AutoCloseable {
 
     private static final int WAIT_FOR_CLOSE_SECONDS = 10;
     private static final char STATEMENT_DELIMITER = ';';
+    private static final String ESCAPE_CHAR = "\\";
     private static final int STATEMENT_CACHE_CAPACITY = 10_000;
     private final static Logger LOGGER = LoggerFactory.getLogger(JdbcConnection.class);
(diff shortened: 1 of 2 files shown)


debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -1223,11 +1224,23 @@ protected String resolveCatalogName(String catalogName) {
         return catalogName;
     }
 
+    protected static CharSequence getEscapeSequence() {
```
jpechane, line 1227: @vjuranek Is not this a bit overkill? Cant' we just use `String` constant `DEFAULT_ESCAPE_CHAR`?

vjuranek: It's here for the purpose we can override it in DB specific sub-classes if there is any other escape char. But given that `\` is SQL standard, it's questionable if we will ever needed to use it. So yes, let's use simple approach and add possibility to override it when we really need it.

jpechane (approved):
@vjuranek Could you please resolve the conflict? Feel free to merge the PR yourself after that.

vjuranek:
Test failures are not related.

PR #4207, opened 2023-01-18
DBZ-6023 Add support for providing a surrogate key when triggering incremental snapshots

PlugaruT:
Moved the PR to a separate Jira issue. Original PR is #4203 
**Context**
Incremental Snapshot uses the PK of the table to keep track of the progress and guarantee ordering of records. Because some tables might have composite PK, the SQL queries used by the snapshot are slow, especially on big tables with hundreds of millions of rows.
By making it possible to provide a surrogate key, users will be able to set which column to be used for the queries. Note that the column should be unique for the table in order to guarantee consistency. The new configuration will be used solely for the snaphotting queries and will have no impact on the records key for example.

### DIFF
--- a/debezium-embedded/src/test/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotTest.java
+++ b/debezium-embedded/src/test/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotTest.java
@@ -982,7 +982,7 @@ public abstract class AbstractIncrementalSnapshotTest<T extends SourceConnector>
         populateTable();
         startConnector();
 
-        sendAdHocSnapshotSignalWithAdditionalConditionWithSurrogateKey(Optional.empty(), Optional.of("\"pk\""), tableDataCollectionId());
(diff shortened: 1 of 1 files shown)


PlugaruT:
@jpechane can you please have a look over this PR, I'm wondering if added tests are enough or I should add more?

PlugaruT:
> @PlugaruT Thanks a lot, I reviewed the PR and pushed some small changes. There is only one issue with your integration tests - you use the primary key as the surrogate key so it is not real test representtion. Would it be possible to
> 
> * Use column `aa` as surrogate
> * Assert that the key is `aa` column and has correct value?

I will try that, I also did not feel good about that, but it was a bit confusing about how those tests are being set up 😅

jpechane:
@PlugaruT Also could you please open a separate PR with a doc for this feature? @roldanbob Will review it.

As it is in this shared fragment we need to make sure that this option is not visible in MongoDB docs.

jpechane:
@PlugaruT Applied, thanks a lot! Just the docs PR is missing as I mentioned in the comment above.

PlugaruT:
> @PlugaruT Applied, thanks a lot! Just the docs PR is missing as I mentioned in the comment above.

PR with the docs is coming by the end of the week.

PR #4209, opened 2023-01-21
DBZ-6033 dbz-server-pulsar support non-default tenant and namespace

zzzming:
To address Jira https://issues.redhat.com/browse/DBZ-6033

This PR supports non-default Pulsar tenant and namespace in Debezium Server. 

In Pulsar, a topic without tenant/namespace implicitly defaults the public tenant and the default namespace. This is what the current implementation is. This feature is required for Pulsar's multiple tenant deployment, so that the debezium server can produce messages to non-default tenant and namespaces.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -299,6 +299,7 @@ Mike Graham
 Mike Kamornikov
 Mikhail Dubrovin
 Mincong Huang
+Ming Luo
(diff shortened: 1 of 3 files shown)


jpechane:
@zzzming Applied, thanks. I just added a commit with rudimentary doc update.

PR #4214, opened 2023-01-23
DBZ-6031 Support compression/logging clauses after LOB storage clauses

Naros:
https://issues.redhat.com/browse/DBZ-6031


jpechane:
@Naros Applied, thanks

PR #4215, opened 2023-01-24
DBZ-4829 : Update mysql.adoc

lokesh1729:
The configuration `event.processing.failure.handling.mode` property is not present in the documentation but I see it is present in the common configuration in code - https://sourcegraph.com/github.com/debezium/debezium/-/blob/debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java?L400

This PR adds the configuration to the documentation


jpechane:
@lokesh1729 Applied, thanks

PR #4218, opened 2023-01-24
DBZ-6039: Reset history recovery attempt counter upon success

morozov:
Additionally, update the documentation to reflect the actual default value: https://github.com/debezium/debezium/blob/48136e08c3f23257523c97dcec668e69463a715e/debezium-storage/debezium-storage-kafka/src/main/java/io/debezium/storage/kafka/history/KafkaSchemaHistory.java#L145-L147


jpechane:
@morozov Applied, thanks

PR #4222, opened 2023-01-25
DBZ-6016 Add CLUSTERING KEY constraint for TOKUDB

ani-sha:
https://issues.redhat.com/browse/DBZ-6016


jpechane:
@ani-sha Applied, thanks

PR #4223, opened 2023-01-25
DBZ-6035 Fix race condition in SQL server tests

vjuranek:
https://issues.redhat.com/browse/DBZ-6035


jpechane:
@vjuranek Applied, thanks! COuld you please check Db2 if there is a similar issue present?

vjuranek:
@jpechane yes, looking

PR #4226, opened 2023-01-26
Note support for NATS JetStream in description

bruth:


jpechane:
@bruth Thanks a lot, merged via https://github.com/debezium/debezium/pull/4227 with updated commit message.

bruth:
Thanks @jpechane! I did take a look at the contributing guide (after the fact) and will respect that process going forward.

jpechane:
@bruth Hi, not a problem. First time is always bumpy :-). I did it this way just because it was a negligible change and I felt it is much mor eproductive to fix it myself instead of pinging back and forth.

PR #4230, opened 2023-01-27
[CI] update changed-files version and setup dependency bot

vjuranek:
* Updating `changed-files` version will hopefully fix issue with `Argument list too long` (e.g. [here](https://github.com/debezium/debezium/actions/runs/4017138191/jobs/6901095978)).
* Setup dependency bot for github actions as current `changed-files` is very old and nobody probably noticed it

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,7 @@
+version: 2
+updates:
+  # Updates for GitHub Actions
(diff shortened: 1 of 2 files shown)


jpechane:
@vjuranek Applied, thanks! Hopefully this will get rid of the issue.

PR #4233, opened 2023-01-28
DBZ-5814 Add "when_needed" option to "snapshot.mode" in MongoDB connector

harveyyue:
https://issues.redhat.com/browse/DBZ-5814

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
@@ -62,19 +62,40 @@ public class MongoDbConnectorConfig extends CommonConnectorConfig {
         /**
          * Always perform an initial snapshot when starting.
          */
-        INITIAL("initial", true),
+        INITIAL("initial", true, false),
 
         /**
          * Never perform a snapshot and only receive new data changes.
          */
-        NEVER("never", false);
+        NEVER("never", false, false),
+
+        /**
+         * Perform a snapshot when it is needed.
+         */
+        WHEN_NEEDED("when_needed", true, false);
 
         private final String value;
         private final boolean includeData;
+        private final boolean shouldSnapshotOnDataError;
 
-        SnapshotMode(String value, boolean includeData) {
+        SnapshotMode(String value, boolean includeData, boolean shouldSnapshotOnDataError) {
             this.value = value;
             this.includeData = includeData;
+            this.shouldSnapshotOnDataError = shouldSnapshotOnDataError;
+        }
+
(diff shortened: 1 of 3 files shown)


jpechane:
@harveyyue The approach LGTM! Unfortunately the large event is detected only during streaming so there has to be an additional mechanism that will set the flag that this happened and when connector is restarted via automated retry mechanism the flag should be detected and snapshot triggered.

harveyyue:
@jpechane Ok, I need to check the related streaming code of mongodb.

harveyyue:
@jpechane IIUC, mongodb snapshotter could read this docment size larger than 16m, right?

jpechane:
@harveyyue IMHO this is a hard MongoDB limit. But it will not allow you to write document larger than would be readable by snapshotter.

jpechane:
@jcechace Could you please review this PR? Is it now applicable with full switch to change streams?

jcechace:
@harveyyue  can you please rebase the code? There will be some changes but nothing major in this area i reckon.

@jpechane Technically yes, unifying this with other connectors still makes sense.

jpechane:
@jcechace Caould you please take care of the PR till merge?
Related to my question - if you look at it it still operates on the assumption of stored oplog timestamp which I suppose is no longer the case.

jcechace:
@jpechane That's why I asked whether @harveyyue can rebase and update it. 

and yes, I will see it through (kinda got that from the first message ;).

Naros:
@jcechace Could we just rebase it ourselves and open a new PR based on the rebase if necessary?  I think at this point its either fallen through the cracks or Harvey is backlogged on other things.

harveyyue:
@jcechace Sorry for late response, I found that there are many code changes in Mongodb Connector, and I'm trying to rebase the code, but some method had been removed, I'm afraid I can't complete it

Naros:
> @jcechace Sorry for late response, I found that there are many code changes in Mongodb Connector, and I'm trying to rebase the code, but some method had been removed, I'm afraid I can't complete it

No problem @harveyyue.

jcechace:
@Naros  I will look into the rebase sometimes next week. However just so that I understand correctly. This is an attempt to distinguish between `initial` a `when_needed` in a sense that `initial` is not able to recover from invalid token? Because currently `initial` actually behaves like `when_needed`.

jpechane:
@jcechace Yes, that's correct. It should work as in case of RDBC connectors. So initial should be done only once. `when_needed` should be done when there is too large event during streaming and connector needs to recover.

jcechace:
@jpechane  in that case, do you agree to have `when_needed` as the default?

jpechane:
Yes, `when_needed` is the default - like MySQL

PR #4239, opened 2023-01-30
DBZ-5973 Support for more types as key in mongodb incremental snapshot implementation

jcechace:
https://issues.redhat.com/browse/DBZ-5973

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbIncrementalSnapshotChangeEventSource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbIncrementalSnapshotChangeEventSource.java
@@ -501,7 +501,31 @@ public class MongoDbIncrementalSnapshotChangeEventSource
         if (row == null) {
             return null;
         }
-        return new Object[]{ ((BsonDocument) row[0]).getInt32(DOCUMENT_ID).getValue() };
+        var documentId = ((BsonDocument) row[0]).get(DOCUMENT_ID);
+
+        Object key;
+
+        switch (documentId.getBsonType()) {
+            case DOUBLE:
+                key = documentId.asDouble().getValue();
+                break;
+            case INT32:
+                key = documentId.asInt32().getValue();
+                break;
+            case INT64:
+                key = documentId.asInt64().getValue();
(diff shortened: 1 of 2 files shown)


jcechace:
Test failures are cause by DECIMAL128 deserialisation... Same as with ObjectId, I need to figure out how to deserialise these

jcechace:
@jpechane  Provided tests pass this can be merged. ObjectId now works as well as Decimal128. Date Time is failing as it seams to lead to empty records -- I wonder if that is because the value of DateTime is in fact a long.... Thus for now I would stick with supporting numeric types and ObjectId.

jpechane:
@jcechace Applied, good job!

xinbinhuang:
@jpechane @jcechace thx for the great work! curious on if it's possible to support string?

jpechane:
@xinbinhuang Thanks for spotting!

@jcechace I completely overlookd - we talked abot string and dates too.

jcechace:
@jpechane  @xinbinhuang  as mentioned in the initial assessment. The keys do need a stable sorting. Strings are problematic as encodings and special characters can lead to unexpected behaviour.  

@jpechane Regarding Date, I think I mentioned this to you... I was unable to make this one work (tests were failing with "Too many empty events"). Internally in BSON this type is implemented as a wrapper around Long (number of seconds from start of the epoch) and getting the value from the BSON type returns this long value. I suspect there were some problems when deserialising the value -- as it was simply interpreted as Long not a Date. I have a note here to create a follow up Jira, however I think that in this case we will have to change the current behaviour so that not only the key itself but also the type is returned by `keyFromRow`

xinbinhuang:
> re: string

@jcechace What would be the potential consequence of these unexpected behavior e.g. data loss? My feeling is that these unexpected behaviours are rare, and certain less critical workload can probably tolerate this type of unexpected behaviour. 

How do you think about allowing `STRING` but call out the limitation in the doc so user can use it with their own risk? Or create a config flag to enforce this contract?

jpechane:
@jcechace Could you please create the follow-up Jira for Date and another for String?

I think it is fine to add support for String as well just with large warning and the link you referred too. IMHO it sounds to me like the guys are hedging against a potential issue - not that there is a real issue.

PR #4241, opened 2023-01-30
DBZ-6035 Await transactions also in IncrementalSnapshotWithRecompileIT

vjuranek:
Override no-op `waitForCdcTransactionPropagation()` also in `IncrementalSnapshotWithRecompileIT` which also fails very often.

https://issues.redhat.com/browse/DBZ-6035


jpechane:
@vjuranek Applied, thanks

PR #4246, opened 2023-01-31
[ci] Configure depenecy bot to use [ci] prefix for PRs

vjuranek:
Once reviewed and merged, I'll push the same change to other repos as well (let me know if you merge it, but don't want me to push it to other repos).


jpechane:
@vjuranek Good idea! Please feel free to merge it here and for other repos as well.

PR #4249, opened 2023-01-31
Variable name clarified

Tideri-Tim2:
It is necessary to query for server_id instead of server-id, which is being used on the config file.


jpechane:
Nice hint, thanks! I am merging this to `main` via https://github.com/debezium/debezium/pull/4251

PR #4253, opened 2023-02-01
DBZ-823 Optionally parallelize initial snapshots

ggaborg:


ggaborg:
@jpechane Hi, I have locally the changes for Db2 and Vitess once this is merged. Debezium Server I suppose is failing after the move to its own repository.

debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
```diff
@@ -297,34 +348,96 @@ protected abstract SchemaChangeEvent getCreateTableEvent(RelationalSnapshotConte
...
+            else {
+                LOGGER.warn("For table '{}' the select statement was not provided, skipping table", tableId);
+                snapshotProgressListener.dataCollectionSnapshotCompleted(snapshotContext.partition, tableId, 0);
             }
+        }
 
-            LOGGER.debug("Snapshotting table {}", tableId);
+        int snapshotOrderByRowCount = connectorConfig.snapshotOrderByRowCount();
+        if (snapshotOrderByRowCount != 0) {
+            LOGGER.info("Sort tables by row count {}", snapshotOrderByRowCount > 0 ? "asc" : "desc");
```
jpechane, line 382: If you really want to implement this then I'd recommend to run `SELECT COUNT(` for tables for which the faster row count was unavailable so it is enabled for all connectors.

debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
```diff
@@ -117,6 +132,11 @@ public SnapshotResult<O> doExecute(ChangeEventSourceContext context, O previousO
             LOGGER.info("Snapshot step 5 - Reading structure of captured tables");
             readTableStructure(context, ctx, previousOffset);
 
+            if (snapshottingTask.snapshotData()) {
+                LOGGER.info("Snapshot step 5.a - Creating connection pool");
```
jpechane, line 136: Please replace ardodced numbers in the log messages with somehing like `step {}` and have a variable stepCounter, that will be passed to the log message and incremented so step numbers are dynamically assigned and there is no need for suffixes.

debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
```diff
@@ -145,10 +165,38 @@ public SnapshotResult<O> doExecute(ChangeEventSourceContext context, O previousO
             return SnapshotResult.completed(ctx.offset);
         }
         finally {
+            if (connectionPool != null) {
+                for (JdbcConnection conn : connectionPool) {
+                    if (!jdbcConnection.equals(conn)) {
```
jpechane, line 170: IMHO here should be identity comparison.

debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
```diff
@@ -297,34 +348,96 @@ protected abstract SchemaChangeEvent getCreateTableEvent(RelationalSnapshotConte
...
-            final TableId tableId = tableIdIterator.next();
-            snapshotContext.firstTable = tableOrder == 1;
-            snapshotContext.lastTable = !tableIdIterator.hasNext();
+        SnapshotReceiver<P> snapshotReceiver = dispatcher.getSnapshotChangeEventReceiver();
 
-            if (!sourceContext.isRunning()) {
-                throw new InterruptedException("Interrupted while snapshotting table " + tableId);
+        int snapshotMaxThreads = connectionPool.size();
+        LOGGER.info("Creating snapshot worker pool with {} worker thread(s)", snapshotMaxThreads);
+        ExecutorService executorService = Executors.newFixedThreadPool(snapshotMaxThreads);
```
jpechane, line 360: Please use `Threads` class so we get rpoper thread names.

debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
```diff
@@ -297,34 +348,96 @@ protected abstract SchemaChangeEvent getCreateTableEvent(RelationalSnapshotConte
...
+
+            Exception firstError = null;
+            for (int i = 0; i < tableCount; i++) {
+                Exception error = completionService.take().get();
+                if (error != null) {
+                    if (firstError == null) {
+                        firstError = error;
+                    }
+                    else {
+                        LOGGER.debug("Snapshot task failed", error);
```
jpechane, line 416: This should be at least `INFO`. Also the first error should be logged.

debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
```diff
@@ -297,34 +348,96 @@ protected abstract SchemaChangeEvent getCreateTableEvent(RelationalSnapshotConte
...
+                OptionalLong rowCount = rowCountTables.get(tableId);
+
+                Callable<Exception> callable = createDataEventsForTableCallable(sourceContext, snapshotContext, snapshotReceiver,
+                        snapshotContext.tables.forTable(tableId), firstTable, lastTable, tableOrder++, tableCount, selectStatement, rowCount, connectionPool, offsets);
+                completionService.submit(callable);
+            }
+
+            Exception firstError = null;
+            for (int i = 0; i < tableCount; i++) {
+                Exception error = completionService.take().get();
```
jpechane, line 410: Do we need to return the exception? Can't we use `ExecutionException` mechanism? Also when the first error is hit then all running tasks should be imediatelly cancelled not waiting for their completion.

debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
```diff
@@ -297,34 +348,96 @@ protected abstract SchemaChangeEvent getCreateTableEvent(RelationalSnapshotConte
...
 
         releaseDataSnapshotLocks(snapshotContext);
-        snapshotContext.offset.preSnapshotCompletion();
+        for (O offset : offsets) {
+            offset.preSnapshotCompletion();
+        }
         snapshotReceiver.completeSnapshot();
-        snapshotContext.offset.postSnapshotCompletion();
+        for (O offset : offsets) {
+            offset.postSnapshotCompletion();
```
jpechane, line 435: The called method manipulates sourceInfo. Is not it lost in that case?

debezium-core/src/main/java/io/debezium/connector/base/ChangeEventQueue.java
```diff
@@ -84,8 +85,7 @@
     @SingleThreadAccess("producer thread")
```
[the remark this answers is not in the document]
ggaborg: @jpechane no, but is only read in parallel by the snapshot mechanism and is not modified at that time

debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
```diff
@@ -437,7 +438,7 @@ public void changeRecord(P partition,
 
     private final class BufferingSnapshotChangeRecordReceiver implements SnapshotReceiver<P> {
 
-        private Supplier<DataChangeEvent> bufferedEvent;
+        private AtomicReference<BufferedDataChangeEvent> bufferedEventRef = new AtomicReference<>(new BufferedDataChangeEvent());
```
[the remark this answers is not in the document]
ggaborg: @jpechane 
to avoid a null check here:
queue.enqueue(bufferedEventRef.getAndSet(nextBufferedEvent).dataChangeEvent);

The queue does another null check in the enqueue method.

debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
ggaborg: @jpechane the equals method is the default one which does identity comparison. should we not trust the contract of the equals method?

debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
ggaborg: @jpechane can you explain? what is supposed to be lost? each offset has its own sourceInfo.

debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
ggaborg: @jpechane the first error is rethrown and is logged by the engine

jpechane:
@ggaborg @zalmane I pushed few changes I believe were necessary for the PR. If CI passes then it can be merged. Could you please take a look and prepared the additional PRs in other repos.
I am sorry for the delay, it really fell throught the cracks :-(.

debezium-core/src/main/java/io/debezium/jdbc/MainConnectionProvidingConnectionFactory.java
```diff
@@ -5,7 +5,7 @@
  */
 package io.debezium.jdbc;
 
-public interface MainConnectionFactory<T> extends ConnectionFactory<T> {
+public interface MainConnectionProvidingConnectionFactory<T> extends ConnectionFactory<T> {
```
ggabora: maybe you intended to name it MainConnectionProvidingFactory?

jpechane: @ggabora As strange is it might sound this was intentional. I want to keep the `ConnectionFactory` suffix. It is a it bit awkard but I believe it is the most descriptive I was able to come up with.

ggabora: @jpechane got it, thanks

ggabora: @jpechane Please merge so I can create PR's for Db2 and Vitess. Thanks!

jpechane: @ggabora Done!

roldanbob (commented):
@jpechane Are we targeting `snapshot.max.threads` for  productization? Or should the property table entries be conditionalized for `community` use?

jpechane:
@roldanbob I'd preferably keep it as is with incubation note. Maybe the conditionalization shoudl be made in a way that wording for product is that this is either dev or tech preview.

PR #4257, opened 2023-02-05
DBZ-6075 Related DBZ-5561 Use TCCL to load configuration classes by default

ismailsimsek:
https://issues.redhat.com/browse/DBZ-6075

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
@@ -337,7 +337,7 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord> {
         @Override
         public EmbeddedEngine build() {
             if (classLoader == null) {
-                classLoader = getClass().getClassLoader();
+                classLoader = Instantiator.getClassLoader();
             }
             if (clock == null) {
                 clock = Clock.system();
--- a/debezium-core/src/main/java/io/debezium/config/Instantiator.java
+++ b/debezium-core/src/main/java/io/debezium/config/Instantiator.java
@@ -15,6 +15,15 @@ import java.util.Properties;
  */
 public class Instantiator {
 
+    public static ClassLoader getClassLoader() {
+        ClassLoader classloader = Thread.currentThread().getContextClassLoader();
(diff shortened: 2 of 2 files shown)


jpechane:
@ismailsimsek Hi, thanks for the PR.Please check the current `Configuration` and `Instantiator` classes to see how the classloader is resolved. I'd recommoned to refactor `Instantiator` to extract `getClassLoader` method that would be used by the class and also by `EmbeddedEngine`.

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
```diff
@@ -605,12 +602,11 @@ public static Builder create() {
     private SourceTask task;
     private final Transformations transformations;
 
-    private EmbeddedEngine(Configuration config, ClassLoader classLoader, Clock clock, DebeziumEngine.ChangeConsumer<SourceRecord> handler,
+    private EmbeddedEngine(Configuration config, Clock clock, DebeziumEngine.ChangeConsumer<SourceRecord> handler,
```
ismailsimsek, line 605: Not sure what is the best way to remove `ClassLoader classLoader,` from argument.. its not necessary anymore

ismailsimsek:
> @ismailsimsek Hi, thanks for the PR.Please check the current `Configuration` and `Instantiator` classes to see how the classloader is resolved. I'd recommoned to refactor `Instantiator` to extract `getClassLoader` method that would be used by the class and also by `EmbeddedEngine`.

@jpechane used following code, what do you think? 

`offsetStore = Instantiator.getInstance(offsetStoreClassName);`

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
jpechane: @ismailsimsek I tend to disagree in this one. Tha classloaader should still be allowed to be set externally. Just if not set, it will be obtained from Instantiator class.

ismailsimsek: ok, updated

jpechane:
@ismailsimsek Applied, thanks

PR #4258, opened 2023-02-06
DBZ-6076: Add support for custom SourceInfoStructMaker for adding new fields to source field

adasari:
Added support to use custom connector config using custom task.

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
@@ -331,6 +331,7 @@ public class SqlServerConnectorConfig extends HistorizedRelationalDatabaseConnec
                     INCREMENTAL_SNAPSHOT_CHUNK_SIZE,
                     INCREMENTAL_SNAPSHOT_ALLOW_SCHEMA_CHANGES,
                     MAX_RETRIES_ON_ERROR)
+            .events(SOURCE_INFO_STRUCT_MAKER)
             .excluding(
                     SCHEMA_INCLUDE_LIST,
                     SCHEMA_EXCLUDE_LIST)
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
@@ -18,6 +18,7 @@ import org.apache.kafka.common.config.ConfigDef.Width;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import io.debezium.config.CommonConnectorConfig;
 import io.debezium.config.ConfigDefinition;
 import io.debezium.config.Configuration;
 import io.debezium.config.EnumeratedValue;
@@ -308,6 +309,9 @@ public class SqlServerConnectorConfig extends HistorizedRelationalDatabaseConnec
             .withDescription("Add OPTION(RECOMPILE) on each SELECT statement during the incremental snapshot process. "
                     + "This prevents parameter sniffing but can cause CPU pressure on the source database.");
 
+    public static final Field SOURCE_INFO_STRUCT_MAKER = CommonConnectorConfig.SOURCE_INFO_STRUCT_MAKER
+            .withDefault(SqlServerSourceInfoStructMaker.class.getName());
+
     private static final ConfigDefinition CONFIG_DEFINITION = HistorizedRelationalDatabaseConnectorConfig.CONFIG_DEFINITION.edit()
             .name("SQL Server")
             .type(
@@ -456,7 +460,7 @@ public class SqlServerConnectorConfig extends HistorizedRelationalDatabaseConnec
 
     @Override
     protected SourceInfoStructMaker<? extends AbstractSourceInfo> getSourceInfoStructMaker(Version version) {
-        return new SqlServerSourceInfoStructMaker(Module.name(), Module.version(), this);
+        return getSourceInfoStructMaker(SqlServerConnectorConfig.SOURCE_INFO_STRUCT_MAKER, Module.name(), Module.version(), this);
     }
 
     private static class SystemTablesPredicate implements TableFilter {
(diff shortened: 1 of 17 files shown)


jpechane:
@adasari Hi, I converted the PR to the draft as I suppose this is just a first step, not the actual implementation. Am I right?

jpechane:
@adasari OK, in this case it would be better to add a new config option that would containe the class name of the custom structmaker and if present then instantiate it and pass the connector config to it.
It would be similar to what we do with `topic.naming.strategy`.

adasari:
@jpechane Please review the changes. i will raise PRs to other repos when these changes are ok.

adasari:
Ping

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -693,7 +700,8 @@ public static FieldNameAdjustmentMode parse(String value) {
                     SIGNAL_ENABLED_CHANNELS,
                     TOPIC_NAMING_STRATEGY,
                     NOTIFICATION_ENABLED_CHANNELS,
-                    SinkNotificationChannel.NOTIFICATION_TOPIC)
+                    SinkNotificationChannel.NOTIFICATION_TOPIC,
+                    SOURCE_INFO_STRUCT_MAKER)
```
jpechane, line 704: @adasari This should be included in per-connector config class as the definition changes. Othrwise the tools will not see the default value in config definition.

jpechane (commented):
@adasari LGTM, I kept only one comment.
Please add a test to at least one connector to verify that this works as expected.
At the same time I beleive you can prepr PR's to non-core repo connectors as I don't expect any further changes.

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
adasari: Left a comment - https://github.com/debezium/debezium/pull/4258/files#r1213185645. thanks.

adasari:
Added test to postgres connector.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
```diff
@@ -456,7 +460,7 @@ protected boolean supportsSchemaChangesDuringIncrementalSnapshot() {
 
     @Override
     protected SourceInfoStructMaker<? extends AbstractSourceInfo> getSourceInfoStructMaker(Version version) {
-        return new SqlServerSourceInfoStructMaker(Module.name(), Module.version(), this);
+        return getSourceInfoStructMaker(SqlServerConnectorConfig.SOURCE_INFO_STRUCT_MAKER, Module.name(), Module.version(), this);
```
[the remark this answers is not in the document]
jpechane: @adasari I understand your point. The issue is that every connector has/should have a different default value and tolling like UI should be able to show that default value. And as the Field is static I don't see any way how to configure it per-connector to have a distinct default value and consume it in CommonConnectorConfig.

jpechane:
@adasari Applied, thanks a lot!

zackary-naas:
> [@jpechane](/jpechane) [jpechane](/jpechane) [force-pushed](/debezium/debezium/compare/28d04306a0805e69a501d61ffe9b3466e6aad5c4..f5b91dd701e40954e5ee875cb3c40cd4b2048a82) the DBZ-6076 branch 2 times, most recently from [ `28d0430` ](/debezium/debezium/commit/28d04306a0805e69a501d61ffe9b3466e6aad5c4) to [ `f5b91dd` ](/debezium/debezium/commit/f5b91dd701e40954e5ee875cb3c40cd4b2048a82) [ Compare ](/debezium/debezium/compare/28d04306a0805e69a501d61ffe9b3466e6aad5c4..f5b91dd701e40954e5ee875cb3c40cd4b2048a82) [ June 14, 2023 23:11 ](#event-9536053121)
> [ ](/adasari)

@adasari how did you implement writing to this custom SourceInfo field as the connector handles events?

adasari:
> > [@jpechane](/jpechane) [jpechane](/jpechane) [force-pushed](/debezium/debezium/compare/28d04306a0805e69a501d61ffe9b3466e6aad5c4..f5b91dd701e40954e5ee875cb3c40cd4b2048a82) the DBZ-6076 branch 2 times, most recently from [ `28d0430` ](/debezium/debezium/commit/28d04306a0805e69a501d61ffe9b3466e6aad5c4) to [ `f5b91dd` ](/debezium/debezium/commit/f5b91dd701e40954e5ee875cb3c40cd4b2048a82) [ Compare ](/debezium/debezium/compare/28d04306a0805e69a501d61ffe9b3466e6aad5c4..f5b91dd701e40954e5ee875cb3c40cd4b2048a82) [ June 14, 2023 23:11 ](#event-9536053121)
> > [ ](/adasari)
> 
> @adasari how did you implement writing to this custom SourceInfo field as the connector handles events?

By custom SourceInfoStructMaker. Reference implementation is https://github.com/debezium/debezium/blob/7d6b4af1dea3e7b0cb6e9abbc590427952814de6/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/CustomPostgresSourceInfoStructMaker.java#L19

zackary-naas:
Apologies, maybe I misunderstood what you were using this for. To explain a bit more what I would like to do - there's an additional field that shows up in the MySQL binlog data, and I'd like to extract that into the SourceInfo at the time of handling an event. As far as I can tell, there is not a way to do that.

To confirm and say it another way - what you are showing with the reference implementation is how to add a static field to the source info correct?

adasari:
You can use it to add static field or dynamic field based on other SourceInfo fields. PostgresSourceInfoStructMaker#struct has SourceInfo only.

adasari:
> Apologies, maybe I misunderstood what you were using this for. To explain a bit more what I would like to do - there's an additional field that shows up in the MySQL binlog data, and I'd like to extract that into the SourceInfo at the time of handling an event. As far as I can tell, there is not a way to do that.
> 
> To confirm and say it another way - what you are showing with the reference implementation is how to add a static field to the source info correct?

Q:: is bin log data is present in SourceRecord ?  
Implementation ideas:
1. Use custom SourceInfoStructMaker if new field of source info is either static or dynamic based on other source info fields.
2. Use SMT if it is  new field is already part of SourceRecord
3.  SMT can be overkill if you regenerate the schema. If SMT is a overkill in your case, Go with hybrid approach.
    a. Use custom SourceInfoStructMaker to add the field to source info schema and default value (like -1 etc)
    b. Use SMT or custom ConnectorTask of doPoll () implementation that fills new field from SourceRecord fields.

Please evaluate all the options and select that is suitable for your use case.

PR #4262, opened 2023-02-07
postgresql doc: fix link to PG List

chtitux:
https://www.postgresql.org/message-id/CAJ3gD9fE%3D0w50sRagcs%2BjrktBXuJAWGZQdSTMa57CCY%2BDh-xbg%40mail.gmail.com is encoded and works.

The link is currently broken because `@` generate a `mailto:` link. See page https://debezium.io/documentation/reference/2.1/connectors/postgresql.html#postgresql-when-things-go-wrong

image

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/postgresql.adoc
+++ b/documentation/modules/ROOT/pages/connectors/postgresql.adoc
@@ -3408,7 +3408,7 @@ ifdef::community[]
 
 [NOTE]
 ====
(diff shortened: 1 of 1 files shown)


jpechane:
@chtitux Applied, thanks! I'll port this change to the `main` as well. Please in future if possible open the PRs against the main instead of branches as we prefer doing the process the opposite way.

PR #4264, opened 2023-02-08
Backport DBZ-6075 Use Instantiator to load ClassLoader

ismailsimsek:


jpechane:
@ismailsimsek The current stable branch is 2.1. TBH I'd like to avoid any backporting to earlier branches esp. 1.x.

PR #4266, opened 2023-02-08
DBZ-6072 NPE when setting schema.history.internal.store.only.captured.tables.ddl=true

harveyyue:
https://issues.redhat.com/browse/DBZ-6072


vjuranek (approved):

jpechane:
@harveyyue I don't think we can do this. Unfortunately in case of databases that supports case sensitive table names there can be an issue in cases of tables differing in letter case only. And yes unfortunately it happens :-(.

harveyyue:
@jpechane Revert the method `isSignalDataCollection` and refactor the method `RelationalSnapshotChangeEventSource.sort` to fix this issue.

PR #4271, opened 2023-02-10
DBZ-6038 Expose walltime from mongo record parser

y5w:
https://issues.redhat.com/browse/DBZ-6038

Follow up to get wallTime accessible


jpechane:
@y5w Applied, thanks.

PR #4273, opened 2023-02-11
DBZ-6070 DDL events not stored in schema history topic for excluded tables

harveyyue:
https://issues.redhat.com/browse/DBZ-6070

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
@@ -858,13 +858,17 @@ public class MySqlConnectorConfig extends HistorizedRelationalDatabaseConnectorC
             .withImportance(Importance.LOW)
             .withDescription("Switched connector to use alternative methods to deliver signals to Debezium instead of writing to signaling table");
 
+    public static final Field STORE_ONLY_CAPTURED_DATABASES_DDL = HistorizedRelationalDatabaseConnectorConfig.STORE_ONLY_CAPTURED_DATABASES_DDL
+            .withDefault(true);
+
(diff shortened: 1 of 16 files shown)


jpechane (changes requested):
@harveyyue Thanks for the PR. It looks good for MySQL. Still my question is, does it work in the same way for Oracle/SQL Server/Db2? IMHO the method is exposed but used only by MySQL, not elswhere.

harveyyue:
> @harveyyue Thanks for the PR. It looks good for MySQL. Still my question is, does it work in the same way for Oracle/SQL Server/Db2? IMHO the method is exposed but used only by MySQL, not elswhere.

Yes, you are correct. I'm not sure whether to apply this option to Oracle/SQL Server/Db2 connectors since it will break the current captured schema behavior. This option's default is true means it only captures the tables from specified schema/database. 
Ok, I will apply this option to the others.

jpechane:
@harveyyue It is ok if default value is different for connectors. Please see `io.debezium.connector.db2.Db2ConnectorConfig.QUERY_FETCH_SIZE` as an example. I'd recommend to create the common option that sets the default corresponding to Db2/Oracle/SQL Server and then override it in MySQL.

harveyyue:
@jpechane Got it.

harveyyue:
Related pr:
Db2: https://github.com/debezium/debezium-connector-db2/pull/97

Naros:
Hi @harveyyue did we check whether the `schema.exclude.list` or `table.exclude.list` impacted this behavior similar to the `database.exclude.list`?

harveyyue:
> Hi @harveyyue did we check whether the `schema.exclude.list` or `table.exclude.list` impacted this behavior similar to the `database.exclude.list`?
No, these filters only will work on enabling the property `schema.history.internal.store.only.captured.tables.ddl`, otherwise will capture all non-system table schemas.

harveyyue:
@jpechane I will rebase it.

jpechane:
@harveyyue Applied, thanks!

PR #4278, opened 2023-02-13
DBZ-6091 Fix SSN NumericOverflow exception

Naros:
https://issues.redhat.com/browse/DBZ-6091

This PR supersedes PR #4274 and should only be applied to `main`.  There will be a separate PR opened that specifically targets the behavior as it relates to branch `2.1`.


jpechane:
@Naros Applied, thanks

PR #4281, opened 2023-02-13
DBZ-6079 Change query.fetch.size default from 0 to 2000.

Naros:
https://issues.redhat.com/browse/DBZ-6079


jpechane (changes requested):
@Naros I don't think we can easily do it. The `query.fetch.size` IS used by SQL Server and Db2. Db2 provides its own default. So IMHO Oracle should do the same. Please see `io.debezium.connector.db2.Db2ConnectorConfig.QUERY_FETCH_SIZE`.

Naros:
> The query.fetch.size IS used by SQL Server and Db2

I had looked at the usage of the accessor on the `CommonConnectorConfig` class and didn't think to check whether the actual property was being directly used in the config-map like is done with SQL Server.  I'll add that back and make the appropriate code suggestions.

jpechane:
@Naros Applied, thanks!

PR #4282, opened 2023-02-13
DBZ-6094 Reduce frequency of "Transaction already processed" debug message

Naros:
https://issues.redhat.com/browse/DBZ-6094

It's possible the connector sees "empty transactions" where we have a START and a matching COMMIT/ROLLBACK, but all the DML operations within the transaction were filtered out due to the include/exclude filters.  In this case, there is no reason to log information about that transaction being skipped and this should help reduce the frequency of this message after a connector restart where there is a large number of transactions with no relevant DML events.


jpechane:
@Naros Applied, thanks

PR #4286, opened 2023-02-15
DBZ-6107 Fix bulk updates when LOB is enabled

Naros:
https://issues.redhat.com/browse/DBZ-6107

This PR now enforces additional checks in `TransactionCommitConsumer` when two UPDATE events are read from the event stream for the same table within the same transaction.  We can determine whether this subsequent UPDATE is synthetic by examining the columns manipulated in the `SET` clause.  If none are LOB columns, the event is treated like a standalone UPDATE event.  If they're LOB columns, the event is considered mergeable and is handled as such.

This PR includes a test where `lob.enabled=true` is set, and we do bulk updates on a table that has no LOB columns, as well as another test where again `lob.enabled=true` and we are doing various update operations against this table, including setting the LOB column to `NULL`.


jpechane:
@Naros Looks quite clever! Applied, thanks.

PR #4287, opened 2023-02-15
DBZ-6113: Fix negative remaining reconnect attempts

EugeneAbramchuk:
This change respects current max attempts but avoids possible negative attempts (current behaviour): 

`Error while attempting to connect to primaryPreferred node of '###' after attempt #17 (-1 remaining)`

Also adjusts incorrect max reconnection attempts (was [actual max attempts]+1)


jpechane:
@EugeneAbramchuk Applied, thanks.

PR #4291, opened 2023-02-17
DBZ-6120 Correctly handle quoting table/schema names

Naros:
https://issues.redhat.com/browse/DBZ-6120


jpechane:
@Naros Applied, thanks!

PR #4297, opened 2023-02-20
DBZ-6115 Oracle CI on pull request events

ani-sha:
https://issues.redhat.com/browse/DBZ-6115


Naros (commented):
Overall this LGTM.  I think the only concern I have is if someone accidentally commits a PR that includes an unexpected change to `OracleAuthors.txt`.  

I am curious, would https://github.com/tspascoal/get-user-teams-membership not work and we would check whether the user is part of the Debezium organization?

ani-sha:
@Naros We can take that approach, but we need a create a team first inside the organization, since it makes no sense to give all members access to this build. Also, I'll go through this action in detail if you need anything else.

Naros:
I'll defer to @jpechane and @gunnarmorling here as I'm unaware of any implications of creating a team within the org.

jpechane:
This sounds like a good idea!
@ani-sha I created [oracle-pr-executors](https://github.com/orgs/debezium/teams/oracle-pr-executors/) team

ani-sha:
We also need a `GITHUB_TOKEN: # Personal access token used to query github (Requires scope: `read:org`)` @jpechane

jpechane:
@ani-sha But that would ,ean that PR CI will not work agains as we cannot inject it?
How about just checking if the token needed to ccess the quay.io is present?

ani-sha:
@jpechane Yes that token in needed. I just pushed the changes and you can see the current failure.

PR #4298, opened 2023-02-20
DBZ-6125 Fix potential stale offsets when using PDBs

Naros:
https://issues.redhat.com/browse/DBZ-6125

**Description:**
DBZ-5648 introduced a regression where transaction start, commit, and rollback events were only being read from within the scope of the configured PDB that the connector was capturing changes instead of the entire Oracle database. This can lead to situations where the offsets may not be advanced as quickly in a low traffic PDB environment, potentially causing stale offsets.


jpechane:
@Naros Applied, thanks

PR #4301, opened 2023-02-20
DBZ-5201: Add support for server-side filtering of replica set change streams

btiernay:
### [DBZ-5201](https://issues.redhat.com/browse/DBZ-5102): Add support for server-side filtering of replica set change streams

This change migrates from doing client-side filtering of streaming change events to performing it server-side via MongoDB's support for aggregation pipelines. This can drastically cut down on the amount of data sent between the client and server. In addition, a new approach to heartbeating has been added which relies on the `postBatchResumeToken` to advance the the cursor even when no changes occur on the pipeline.

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ChangeStreamPipelineFactory.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ChangeStreamPipelineFactory.java
@@ -1,0 +1,229 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.mongodb;
+
+import static java.util.stream.Collectors.toList;
+
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Optional;
+import java.util.stream.Stream;
+
+import org.bson.conversions.Bson;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.mongodb.BasicDBObject;
+import com.mongodb.client.model.Aggregates;
+import com.mongodb.client.model.Field;
+import com.mongodb.client.model.Filters;
+import com.mongodb.client.model.changestream.OperationType;
+
+import io.debezium.connector.mongodb.Filters.FilterConfig;
+import io.debezium.data.Envelope;
+
+/**
+ * A factory to produce a MongoDB change stream pipeline expression.
+ */
+class ChangeStreamPipelineFactory {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ChangeStreamPipelineFactory.class);
+
+    private final ReplicaSetOffsetContext rsOffsetContext;
+    private final MongoDbConnectorConfig connectorConfig;
+    private final FilterConfig filterConfig;
+
+    ChangeStreamPipelineFactory(ReplicaSetOffsetContext rsOffsetContext, MongoDbConnectorConfig connectorConfig, FilterConfig filterConfig) {
+        this.rsOffsetContext = rsOffsetContext;
+        this.connectorConfig = connectorConfig;
+        this.filterConfig = filterConfig;
+    }
+
+    List<Bson> create() {
+        // Resolve the leaf filters
+        var filters = Stream
+                .of(
+                        createCollectionFilter(filterConfig),
+                        createOperationTypeFilter(connectorConfig),
+                        createClusterTimeFilter(rsOffsetContext))
+                .flatMap(Optional::stream)
+                .collect(toList());
+
+        // Combine
+        var andFilter = Filters.and(filters);
+        var matchFilter = Aggregates.match(andFilter);
+
+        // Pipeline
+        // Note that change streams cannot use indexes:
+        // - https://www.mongodb.com/docs/manual/administration/change-streams-production-recommendations/#indexes-and-performance
+        // Note that `$addFields` must be used over `$set`/ `$unset` to support MongoDB 4.0 which doesn't support these operators:
+        // - https://www.mongodb.com/docs/manual/changeStreams/#modify-change-stream-output
+        var pipeline = List.of(
+                // Materialize a "namespace" field so that we can do qualified collection name matching per
+                // the configuration requirements
+                // Note that per the docs, if `$ns` doesn't exist, `$concat` will return `null`
+                addFields("namespace", concat("$ns.db", ".", "$ns.coll")),
+
+                // Filter the documents
+                matchFilter,
+
+                // This is required to prevent driver `ChangeStreamDocument` deserialization issues:
+                // > Caused by: org.bson.codecs.configuration.CodecConfigurationException:
+                // > Failed to decode 'ChangeStreamDocument'. Decoding 'namespace' errored with:
+                // > readStartDocument can only be called when CurrentBSONType is DOCUMENT, not when CurrentBSONType is STRING.
+                addFields("namespace", "$$REMOVE"));
+
+        LOGGER.info("Change stream pipeline: {}", new BasicDBObject("pipeline", pipeline).toBsonDocument().toJson());
+        return pipeline;
+    }
+
+    private static Optional<Bson> createCollectionFilter(FilterConfig filterConfig) {
+        // Database filters
+        var dbFilters = Optional.<Bson> empty();
+        if (filterConfig.getDbIncludeList() != null) {
+            dbFilters = Optional.of(Filters.regex("ns.db", filterConfig.getDbIncludeList().replaceAll(",", "|"), "i"));
+        }
+        else if (filterConfig.getDbExcludeList() != null) {
+            dbFilters = Optional.of(Filters.regex("ns.db", "(?!" + filterConfig.getDbExcludeList().replaceAll(",", "|") + ")", "i"));
+        }
+
+        // TODO: This should be removed since, from https://www.mongodb.com/docs/manual/changeStreams/:
+        // > non-system collections across all databases except for `admin`, `local`, and `config`
+        var excludedBuiltInDbs = Optional.<Bson> empty();
+        if (filterConfig.getBuiltInDbNames() != null) {
+            excludedBuiltInDbs = Optional.of(Filters.nin("ns.db", filterConfig.getBuiltInDbNames()));
+        }
+
+        // Collection filters
+        var collectionsFilters = Optional.<Bson> empty();
+        if (filterConfig.getCollectionIncludeList() != null) {
+            collectionsFilters = Optional
+                    .of(Filters.regex("namespace", filterConfig.getCollectionIncludeList().replaceAll(",", "|"), "i"));
+        }
+        else if (filterConfig.getCollectionExcludeList() != null) {
+            collectionsFilters = Optional
+                    .of(Filters.regex("namespace", "(?!" + filterConfig.getCollectionExcludeList().replaceAll(",", "|") + ")", "i"));
+        }
+        var includedSignalCollectionFilters = Optional.<Bson> empty();
+        if (filterConfig.getSignalDataCollection() != null) {
+            includedSignalCollectionFilters = Optional.of(Filters.eq("namespace", filterConfig.getSignalDataCollection()));
+        }
+
+        // Combined filters
+        return andFilters(
+                excludedBuiltInDbs,
+                dbFilters,
+                orFilters(
+                        includedSignalCollectionFilters,
+                        collectionsFilters));
+    }
+
+    private static Optional<Bson> createOperationTypeFilter(MongoDbConnectorConfig connectorConfig) {
+        // Per https://debezium.io/documentation/reference/stable/connectors/mongodb.html#mongodb-property-skipped-operations
+        // > The supported operations include:
+        // > - 'c' for inserts/create
+        // > - 'u' for updates/replace,
+        // > - 'd' for deletes,
+        // > - 't' for truncates, and
+        // > - 'none' to not skip any operations.
+        // > By default, 'truncate' operations are skipped (not emitted by this connector).
+        // However, 'truncate' is not supported since it doesn't exist as a
+        // [MongoDB change type](https://www.mongodb.com/docs/manual/reference/change-events/). Also note that
+        // support for 'none' effectively implies 'c', 'u', 'd'
+
+        // First, begin by including all the supported Debezium change events
+        var includedOperations = new ArrayList<OperationType>();
+        includedOperations.add(OperationType.INSERT);
+        includedOperations.add(OperationType.UPDATE);
+        includedOperations.add(OperationType.REPLACE);
+        includedOperations.add(OperationType.DELETE);
+
+        // Next, remove any implied by the configuration
+        var skippedOperations = connectorConfig.getSkippedOperations();
+        if (skippedOperations.contains(Envelope.Operation.CREATE)) {
+            includedOperations.remove(OperationType.INSERT);
+        }
+        if (skippedOperations.contains(Envelope.Operation.UPDATE)) {
+            includedOperations.remove(OperationType.UPDATE);
+            includedOperations.remove(OperationType.REPLACE);
+        }
+        if (skippedOperations.contains(Envelope.Operation.DELETE)) {
+            includedOperations.remove(OperationType.DELETE);
+        }
+
+        return Optional.of(Filters.in("operationType", includedOperations.stream()
+                .map(OperationType::getValue)
+                .collect(toList())));
+    }
+
+    private static Optional<Bson> createClusterTimeFilter(ReplicaSetOffsetContext rsOffsetContext) {
+        if (rsOffsetContext.lastResumeToken() != null) {
+            return Optional.empty();
+        }
+
+        // After snapshot the oplogStart points to the last change snapshotted, so it must filtered-out to
+        // prevent duplicates
+        return Optional.of(Filters.ne("clusterTime", rsOffsetContext.lastOffsetTimestamp()));
+    }
+
+    @SafeVarargs
+    private static Optional<Bson> andFilters(Optional<Bson>... filters) {
+        var resolved = resolveFilters(filters);
+        if (resolved.isEmpty()) {
+            return Optional.empty();
+        }
+        else if (resolved.size() == 1) {
+            return Optional.of(resolved.get(0));
+        }
+        else {
+            return Optional.of(Filters.and(resolved));
+        }
+    }
+
+    @SafeVarargs
+    private static Optional<Bson> orFilters(Optional<Bson>... filters) {
+        var resolved = resolveFilters(filters);
+        if (resolved.isEmpty()) {
+            return Optional.empty();
+        }
+        else if (resolved.size() == 1) {
+            return Optional.of(resolved.get(0));
+        }
+        else {
+            return Optional.of(Filters.or(resolved));
+        }
+    }
+
+    @SafeVarargs
+    private static List<Bson> resolveFilters(Optional<Bson>... filters) {
+        return Stream.of(filters)
+                .flatMap(Optional::stream)
+                .collect(toList());
+    }
+
+    private static Bson regexMatch(String regex, String options, Bson input) {
+        return new BasicDBObject("$regexMatch", new BasicDBObject()
+                .append("input", input)
+                .append("regex", regex)
+                .append("options", options));
+    }
+
+    private static Bson concat(Object... expressions) {
+        return new BasicDBObject("$concat", List.of(expressions));
+    }
+
+    private static Bson set(String name, Bson expression) {
+        return Aggregates.set(new Field<>(name, expression));
+    }
+
+    private static Bson unset(String name) {
+        return new BasicDBObject("$unset", List.of(name));
+    }
+
+    private static Bson addFields(String name, Object expression) {
+        return new BasicDBObject("$addFields", new BasicDBObject(name, expression));
+    }
+}
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
@@ -6,12 +6,9 @@
 package io.debezium.connector.mongodb;
 
 import java.time.Duration;
-import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Set;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.TimeUnit;
@@ -28,17 +25,14 @@ import org.slf4j.LoggerFactory;
 import com.mongodb.ReadPreference;
 import com.mongodb.ServerAddress;
 import com.mongodb.client.ChangeStreamIterable;
+import com.mongodb.client.MongoChangeStreamCursor;
 import com.mongodb.client.MongoClient;
-import com.mongodb.client.MongoCursor;
-import com.mongodb.client.model.Aggregates;
-import com.mongodb.client.model.Filters;
 import com.mongodb.client.model.changestream.ChangeStreamDocument;
 import com.mongodb.client.model.changestream.FullDocument;
 import com.mongodb.client.model.changestream.FullDocumentBeforeChange;
 
 import io.debezium.connector.mongodb.ConnectionContext.MongoPreferredNode;
 import io.debezium.connector.mongodb.recordemitter.MongoDbChangeRecordEmitter;
-import io.debezium.data.Envelope.Operation;
 import io.debezium.pipeline.ErrorHandler;
 import io.debezium.pipeline.EventDispatcher;
 import io.debezium.pipeline.source.spi.StreamingChangeEventSource;
@@ -173,25 +167,6 @@ public class MongoDbStreamingChangeEventSource implements StreamingChangeEventSo
         });
     }
 
-    private List<String> getChangeStreamSkippedOperationsFilter() {
-        final Set<Operation> skippedOperations = taskContext.getConnectorConfig().getSkippedOperations();
-        final List<String> includedOperations = new ArrayList<>();
-
-        if (!skippedOperations.contains(Operation.CREATE)) {
-            includedOperations.add("insert");
-        }
-
-        if (!skippedOperations.contains(Operation.UPDATE)) {
-            // TODO Check that replace is tested
-            includedOperations.add("update");
-            includedOperations.add("replace");
-        }
-        if (!skippedOperations.contains(Operation.DELETE)) {
-            includedOperations.add("delete");
-        }
-        return includedOperations;
-    }
-
     private void readChangeStream(MongoClient client, MongoPreferredNode mongo, ReplicaSet replicaSet, ChangeEventSourceContext context,
                                   MongoDbOffsetContext offsetContext) {
         final ReplicaSetPartition rsPartition = offsetContext.getReplicaSetPartition(replicaSet);
@@ -204,14 +179,8 @@ public class MongoDbStreamingChangeEventSource implements StreamingChangeEventSo
         final ServerAddress nodeAddress = MongoUtil.getPreferredAddress(client, mongo.getPreference());
         LOGGER.info("Reading change stream for '{}'/{} from {} starting at {}", replicaSet, mongo.getPreference().getName(), nodeAddress, oplogStart);
 
-        Bson filters = Filters.in("operationType", getChangeStreamSkippedOperationsFilter());
-        if (rsOffsetContext.lastResumeToken() == null) {
-            // After snapshot the oplogStart points to the last change snapshotted
-            // It must be filtered-out
-            filters = Filters.and(filters, Filters.ne("clusterTime", oplogStart));
-        }
-        final ChangeStreamIterable<BsonDocument> rsChangeStream = client.watch(
-                Arrays.asList(Aggregates.match(filters)), BsonDocument.class);
+        final List<Bson> pipeline = new ChangeStreamPipelineFactory(rsOffsetContext, taskContext.getConnectorConfig(), taskContext.filters().getConfig()).create();
+        final ChangeStreamIterable<BsonDocument> rsChangeStream = client.watch(pipeline, BsonDocument.class);
         if (taskContext.getCaptureMode().isFullUpdate()) {
             rsChangeStream.fullDocument(FullDocument.UPDATE_LOOKUP);
         }
@@ -234,7 +203,7 @@ public class MongoDbStreamingChangeEventSource implements StreamingChangeEventSo
             rsChangeStream.maxAwaitTime(connectorConfig.getCursorMaxAwaitTime(), TimeUnit.MILLISECONDS);
         }
 
-        try (MongoCursor<ChangeStreamDocument<BsonDocument>> cursor = rsChangeStream.iterator()) {
+        try (MongoChangeStreamCursor<ChangeStreamDocument<BsonDocument>> cursor = rsChangeStream.cursor()) {
             // In Replicator, this used cursor.hasNext() but this is a blocking call and I observed that this can
             // delay the shutdown of the connector by up to 15 seconds or longer. By introducing a Metronome, we
             // can respond to the stop request much faster and without much overhead.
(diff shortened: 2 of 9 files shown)


btiernay:
WIP! (sorry meant to be draft PR).

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -245,6 +224,7 @@ else if (oplogStart.getTime() > 0) {
                 if (event != null) {
                     LOGGER.trace("Arrived Change Stream event: {}", event);
 
+                    // TODO: Should this be removed as part of DBZ-5102?
```
btiernay, line 227: This now appears superfluous. However, there could be some edge cases for which the collection name doesn't show up in the change event. For example, the [dropDatabase]( https://www.mongodb.com/docs/manual/reference/change-events/dropDatabase/#mongodb-data-dropDatabase) event doesn't contain a `ns.coll` field, which would not be excluded by the new server-side filter.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -318,6 +299,89 @@ protected MongoDbOffsetContext initializeOffsets(MongoDbConnectorConfig connecto
...
+        }
+
+        private static Optional<Bson> createOperationTypeFilter(MongoDbConnectorConfig connectorConfig) {
+            var skippedOperations = connectorConfig.getSkippedOperations();
+            if (skippedOperations.isEmpty()) {
+                // Include all
+                return Optional.empty();
+            }
+
+            // TODO: What about the rest of these? https://www.mongodb.com/docs/manual/reference/change-events/
```
btiernay, line 356: I was scanning through the above link and noticed that not all events are represented. I'm wondering if it makes more sense to convert this expression a `$nin` filter instead?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -318,6 +299,89 @@ protected MongoDbOffsetContext initializeOffsets(MongoDbConnectorConfig connecto
...
+            // Combine
+            var andFilter = Filters.and(filters);
+            var matchFilter = Aggregates.match(andFilter);
+            var pipeline = Collections.singletonList(matchFilter);
+            LOGGER.debug("Change stream pipeline: {}", pipeline);
+
+            return pipeline;
+        }
+
+        private static Optional<Bson> createCollectionFilter(MongoPreferredNode mongo) {
```
btiernay, line 336: The main change.

.github/workflows/sanity-check.yml
```diff
@@ -51,4 +51,4 @@ jobs:
         continue-on-error: false
         with:
           script: |
-            throw new Error('Commit has no DBZ-xxx prefiix')
+            throw new Error('Commit has no DBZ-XXXX message prefix')
```
jcechace, line 54: While I agree this is a bit better wording, please let's not mix it in here.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
jcechace: Yes, I would say so. The pipeline setup now guarantees that only events on desired databases are reported by database.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -256,6 +236,7 @@ else if (oplogStart.getTime() > 0) {
                                 event.getNamespace().getDatabaseName(),
                                 event.getNamespace().getCollectionName());
 
+                        // TODO: Should this be removed as part of DBZ-5102?
```
jcechace, line 239: Same as with database filters.. it's part of the pipeline now.

jcechace (changes requested):
So far this looks like a good direction to me. Let me know if you have issues addressing those failing tests.

.github/workflows/sanity-check.yml
btiernay: Fixed.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
btiernay: Removed.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -318,6 +291,94 @@ protected MongoDbOffsetContext initializeOffsets(MongoDbConnectorConfig connecto
...
+            // > The support operations include:
+            // > - 'c' for inserts/create, '
+            // > - 'u' for updates,
+            // > - 'd' for deletes,
+            // > - 't' for truncates, and
+            // > - 'none' to not skip any operations.
+            // > By default, 'truncate' operations are skipped (not emitted by this connector).
+            // However, 'truncate' is not supported since it doesn't exist as a MongoDB change type. And support for 'none'
+            // seems to effectively imply 'c', 'u', 'd'
+            // TODO: What about the rest of these when 'none' is selected? https://www.mongodb.com/docs/manual/reference/change-events/
```
btiernay, line 352: Concern cited below with respect to docs. 

As an aside, to make the code more readable / explicit, I was thinking of defining the allowed operations explicitly in a set, then removing based on the config.

documentation/modules/ROOT/pages/connectors/mongodb.adoc
```diff
@@ -1628,7 +1628,7 @@ Disabled by default.
 |[[mongodb-property-skipped-operations]]<<mongodb-property-skipped-operations, `+skipped.operations+`>>
 |`t`
 |A comma-separated list of operation types that will be skipped during streaming.
-The operations include: `c` for inserts/create, `u` for updates, `d` for deletes, `t` for truncates, and `none` to not skip any operations.
+The operations include: `c` for inserts/create, `u` for updates/replace, `d` for deletes, `t` for truncates, and `none` to not skip any operations.
```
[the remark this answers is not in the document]
jcechace: Dropping the mention of `t` might be good enough here. Although it is indeed the default value, it is then filtered out in the following code snippet. 

```java
private List<String> getChangeStreamSkippedOperationsFilter() {
        final Set<Operation> skippedOperations = taskContext.getConnectorConfig().getSkippedOperations();
        final List<String> includedOperations = new ArrayList<>();

        if (!skippedOperations.contains(Operation.CREATE)) {
            includedOperations.add("insert");
        }

        if (!skippedOperations.contains(Operation.UPDATE)) {
            // TODO Check that replace is tested
            includedOperations.add("update");
            includedOperations.add("replace");
        }
        if (!skippedOperations.contains(Operation.DELETE)) {
            includedOperations.add("delete");
        }
        return includedOperations;
    }
```

This is a bit misleading as the filed is part of the common connector configuration -- which is a bit relational centric. There is no change event equivalent to `TRUNCATE TABLE` for mongodb, thus it is ignored. Only `c`, `u`, `d`, and none (which is the effective default) make sense for mongo.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
jcechace: Sounds good

documentation/modules/ROOT/pages/connectors/mongodb.adoc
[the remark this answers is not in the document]
btiernay: Updated.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
btiernay: Updated.

btiernay:
@jcechace @jpechane  With respect to the tests, the first issue I'm hitting appears to be a race condition between when the collection is initialized / created in the test and when the change stream is initialized. If the collection is not created first, then the change stream will materialize a listing that doesn't include the test db / collections, and this will be frozen for the rest of the test.

There are at least a couple of ways to fix this:

1. Eagerly initialize all collections before the test (not sure what the ripple is here in terms of changes)
2. Perform late evaluation of the collection / database names by making the pipeline use regexes when filtering `ns` elements.

What do you think? My take is that 1. seems like it would be generally fine in practice (i.e. the databases and collections must be created before starting the connector), but it would represent a change in (undocumented) behavior so I would like to hear what you think about losing that level of dynamism. Not sure what the performance impact of 2. would be on the server in terms of CPU load, but likely not too bad.

Will wait for you to respond before continuing further.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -245,33 +225,26 @@ else if (oplogStart.getTime() > 0) {
...
-                        oplogContext.getOffset().changeStreamEvent(event);
-                        oplogContext.getOffset().getOffset();
-                        CollectionId collectionId = new CollectionId(
-                                replicaSet.replicaSetName(),
-                                event.getNamespace().getDatabaseName(),
-                                event.getNamespace().getCollectionName());
-
-                        if (taskContext.filters().collectionFilter().test(collectionId)) {
-                            try {
-                                dispatcher.dispatchDataChangeEvent(
```
btiernay, line 261: Removing client-side database filtering.

jcechace:
> @jcechace @jpechane With respect to the tests, the first issue I'm hitting appears to be a race condition between when the collection is initialized / created in the test and when the change stream is initialized. If the collection is not created first, then the change stream will materialize a listing that doesn't include the test db / collections, and this will be frozen for the rest of the test.
> 
> There are at least a couple of ways to fix this:
> 
> 1. Eagerly initialize all collections before the test (not sure what the ripple is here in terms of changes)
> 2. Perform late evaluation of the collection / database names by making the pipeline use regexes when filtering `ns` elements.
> 
> What do you think? My take is that 1. seems like it would be generally fine in practice (i.e. the databases and collections must be created before starting the connector), but it would represent a change in (undocumented) behavior so I would like to hear what you think about losing that level of dynamism. Not sure what the performance impact of 2. would be on the server in terms of CPU load, but likely not too bad.
> 
> Will wait for you to respond before continuing further.

The issue with tests actually shows that this PR is changing the behavour...

Currently all change events are streamed and ONLY THEN filtered client side based on the include lists. This means that if a new collection (which passes the name test) is created during the stream phase, the events will be picked up by Debezium. 

Setting up the pipeline with server-side the way it is done in this PR at the moment means using specific database and collection names at the time of pipeline constructions -- this for the lifespan of the change stream, only events to these will be streamed. This means that operations on the newly created collection will be ignored. 

Thus the server-side filtering needs to use regexes. The question is whether this will have any significant impact on the database.

btiernay:
Okay, so I pretty much ported all the client-side filtering logic to a change stream pipeline filter that executes on the server side. But I just came to learn that this is more complicated than I expected [since](https://debezium.io/documentation/reference/stable/connectors/mongodb.html#mongodb-property-collection-include-list):
image

I didn't realize initially that the collection pattern matches both the collection _and database_. This means that there much be a concatenation step the runs on the server to create the effective dotted "namespace".

jcechace:
Ah.. I see... the collection filters are applied against `CollectionId` not the name of the collection... 

I'm almost inclined to "fork" the filtering logic and have the server side filtering separately while also keeping the client side one.

A way to make this work would be to use the [$replaceWith](https://www.mongodb.com/docs/manual/reference/operator/aggregation/replaceWith/#mongodb-pipeline-pipe.-replaceWith) aggregation. However this is already starting look like a decent amount of an overhead. 

@jpechane What would you recommend here?

btiernay:
Was able to get this working by using ` $set` and `$unset` in the pipeline to materialize a `namespace` field for filtering. Looks like it works. Resulting pipeline:
```json
{
   "pipeline":[
      {
         "$set":{
            "namespace":{
               "$concat":[
                  "$ns.db",
                  ".",
                  "$ns.coll"
               ]
            }
         }
      },
      {
         "$match":{
            "$and":[
               {
                  "$and":[
                     {
                        "ns.db":{
                           "$nin":[
                              "admin",
                              "config",
                              "local"
                           ]
                        }
                     },
                     {
                        "namespace":{
                           "$regularExpression":{
                              "pattern":"dbit.*",
                              "options":"i"
                           }
                        }
                     }
                  ]
               },
               {
                  "operationType":{
                     "$in":[
                        "insert",
                        "update",
                        "replace",
                        "delete"
                     ]
                  }
               },
               {
                  "clusterTime":{
                     "$ne":{
                        "$timestamp":{
                           "t":1677018164,
                           "i":17
                        }
                     }
                  }
               }
            ]
         }
      },
      {
         "$unset":[
            "namespace"
         ]
      }
   ]
}
```

This can be simplified, and will need to be reordered for correctness (e.g. can't `$concat` fields that don't exist).

btiernay:
All `MongoDbConnectorIT` tests seem to be passing except `shouldEmitHeartbeatMessages`. Could this be because we have much less change events flowing to the client? Not as familiar with this feature.

jcechace:
Heartbeats esentially has two purposes 

1) It works as a liveness check that there are still changes received by the connector from change stream, however these are not on monitored collections
2) It reduces the need to reexecute snapshot. This could happen when the last monitored change was pushed out of oplog by changes to not monitored collections --so we now have a resume token which is deemed too old (older than begining of oplog) which indicates possible data loss and thus snapshot execution. 

@jpechane so any server side filtering actually breaks this. So this probably has to be an optional feature, right?

btiernay:
It seems as though it may be possible to get a heartbeat even when the change stream does not match a pipeline. This is the approach that the MongoDB Kafka Connector take's in it's [HeartbeatManager](https://github.com/mongodb/mongo-kafka/blob/41d7f8760d6785c488fba408736c2e480ccddfa5/src/main/java/com/mongodb/kafka/connect/source/heartbeat/HeartbeatManager.java#L77) in concert with [the source task](https://github.com/mongodb/mongo-kafka/blob/41d7f8760d6785c488fba408736c2e480ccddfa5/src/main/java/com/mongodb/kafka/connect/source/StartedMongoSourceTask.java#L270-L282). To perform something similar in Debezium, we would need to change [this](https://github.com/debezium/debezium/pull/4301/files#diff-77af13386f47731de66109b5de074aa237920346f92f0958c96631dbbad4e740R205):
```java
try (MongoCursor<ChangeStreamDocument<BsonDocument>> cursor = rsChangeStream.iterator()) {
```

to be:
```java
try (MongoChangeStreamCursor<ChangeStreamDocument<BsonDocument>> cursor = rsChangeStream.cursor()) {
```

Then, the `cursor.getResumeToken()` method would be available, [which](https://github.com/mongodb/mongo-java-driver/blob/master/driver-sync/src/main/com/mongodb/client/MongoChangeStreamCursor.java#L49):

> Returns the resume token. If a batch has been iterated to the last change stream document in the batch and a postBatchResumeToken is included in the document, the postBatchResumeToken will be returned. Otherwise, the resume token contained in the last change stream document will be returned. 

The nice benefit about using this approach is that it potentially wouldn't depend on the activity of the replica set. Thoughts?

jcechace:
This would solve the liveness, what about the part where we get truly latest resume token? Meaning if I have a change stream open, 

1. consume monitored event and store resume token T1 (emit change event)
2. consume umonitored event and store resume token T2 (emit heartbeat event)
3. consume unmonitored event and store resume token T3 (emit heartbeat event)

With current implementation after restart we know it is fine to resume from T3 even when the size of the oplog was 2 and T1 is no longer valid.

Would this still work with the proposed solution? I think in this case T1 would be reported in place of T2 and T3. Meaning we might have issues resuming with the same oplogize size since T1 is no longer valid.

btiernay:
> Would this still work with the proposed solution? I think in this case T1 would be reported in place of T2 and T3. Meaning we might have issues resuming with the same oplogize size since T1 is no longer valid.

I don't believe this to be the case. From the [docs](https://www.mongodb.com/docs/manual/changeStreams/#resume-tokens) (and looking a the driver source), the resume token is updated on every [getMore](https://www.mongodb.com/docs/manual/reference/command/getMore/#mongodb-dbcommand-dbcmd.getMore) call.  Additionally, from [the spec](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#exposing-all-resume-tokens):

> Drivers MUST expose a mechanism to retrieve the same resume token that would be used to automatically resume. It MUST be possible to use this mechanism after iterating every document. **It MUST be possible for users to use this mechanism periodically even when no documents are getting returned** (i.e. getMore has returned empty batches). 

This would explain why the MongoDB Kafka Connect source uses the same approach. From their docs on [heartbeat.interval.ms](https://www.mongodb.com/docs/kafka-connector/current/source-connector/configuration-properties/error-handling/#settings):

> The connector sends heartbeat messages when source records are not published in the specified interval.
Heartbeat messages contain a postBatchResumeToken data field. The value of this field contains the MongoDB server oplog entry that the connector last read from the change stream. This mechanism improves resumability of the connector for low volume namespaces.

jcechace:
Sweet, so I think we have a solution to this problem.

btiernay:
@jcechace The biggest question I have remaining now is how to integrate this approach into the current logic. Specifically [this is the area from `MongoDbStreamingChangeEventSource`](https://github.com/debezium/debezium/blob/9ef465908db7aca800631b9203d947d17bc84436/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java#L217-L240) that I will need some hand holding to understand:
```java
    oplogContext.getOffset().changeStreamEvent(event);
    oplogContext.getOffset().getOffset();
    CollectionId collectionId = new CollectionId(
            replicaSet.replicaSetName(),
            event.getNamespace().getDatabaseName(),
            event.getNamespace().getCollectionName());

   // ...

    try {
        dispatcher.dispatchHeartbeatEvent(oplogContext.getPartition(), oplogContext.getOffset());
    }
        LOGGER.info("Replicator thread is interrupted");
        Thread.currentThread().interrupt();
        return;
    }
```

The points of clarification / guidance that would be useful:
-  Is the call to `oplogContext.getOffset().changeStreamEvent(event)` required for heart beating? Should we introduce a new `oplogContext.getOffset().resumeToken(resumeToken)` method?
-  The strange call to the getter `oplogContext.getOffset().getOffset();` seem to have side effects. What is it's precondition and post condition?
- The call to `dispatcher.dispatchHeartbeatEvent(oplogContext.getPartition(), oplogContext.getOffset());` will likely need to change. Not sure if the right approach is to affect `oplogContext` or to pass in a new argument.

@jpechane Since you've worked pretty closely in this area, I would really appreciate your input. Thank you! 🙏

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -291,6 +244,16 @@ else if (oplogStart.getTime() > 0) {
                         break;
                     }
                 }
+
+                try {
+                    oplogContext.getOffset().cursorEvent(cursor);
```
btiernay, line 249: My naive attempt at getting heartbeats to work from the cursor `resumeToken`. Also note that I moved this to happen even when `event` is undefined. It probably should be guarded to only first the heartbeat event wen the cursor has a non-null `resumeToken`.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/SourceInfo.java
```diff
@@ -299,6 +300,20 @@ public void initialPosition(String replicaSetName, BsonDocument oplogEvent) {
...
+            return;
+        }
+
+        String namespace = "";
+        long wallTime = 0L;
+        BsonTimestamp ts = ResumeTokens.getTimestamp(cursor.getResumeToken());
+        Position position = Position.changeStreamPosition(ts, cursor.getResumeToken().getString("_data").getValue(), null);
+        positionsByReplicaSetName.put(replicaSetName, position);
+
+        onEvent(replicaSetName, CollectionId.parse(replicaSetName, namespace), position, wallTime);
```
btiernay, line 314: ⚠️ I could really use a second pair of eyes here to help verify the impact of setting some of these fields thusly. Had a hard time understanding all the implications of the various state elements, their invariants, etc.

btiernay:
[So close](https://github.com/debezium/debezium/actions/runs/4247683129/jobs/7386034421) to a clean bill of health, but failing on Mongo 4.0 due to the lack of support of `$set`:
```
Caused by: com.mongodb.MongoCommandException: Command failed with error 40324 (Location40324): 'Unrecognized pipeline stage name: '$set'' on server 172.23.0.2:27017. The full response is {"operationTime": {"$timestamp": {"t": 1677106955, "i": 1}}, "ok": 0.0, "errmsg": "Unrecognized pipeline stage name: '$set'", "code": 40324, "codeName": "Location40324", "$clusterTime": {"clusterTime": {"$timestamp": {"t": 1677106955, "i": 1}}, "signature": {"hash": {"$binary": {"base64": "AAAAAAAAAAAAAAAAAAAAAAAAAAA=", "subType": "00"}}, "keyId": 0}}}
```

This limitation is documented [here](https://www.mongodb.com/docs/manual/changeStreams/#modify-change-stream-output):
> image

I believe this may be possible using [`$addFields`](https://www.mongodb.com/docs/manual/reference/operator/aggregation/addFields/#mongodb-pipeline-pipe.-addFields), so will do some experimenting and report back.

jcechace:
[$addFields](https://www.mongodb.com/docs/manual/reference/operator/aggregation/addFields/#mongodb-pipeline-pipe.-addFields) should work and I think it doesn't have to be removed in the consequent pipeline step.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
jcechace: I'm not sure I understand this comment... Shouldn't the cursor always have a non-null token (unless it was closed)?  Also what do you mean by "undefined" event?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ResumeTokens.java
```diff
@@ -0,0 +1,75 @@
...
+            // Mongo 3.6: (and potentially some 4.0?) servers encode _data as binary
+            return data.asBinary().getData();
+        }
+        else {
+            throw new IllegalArgumentException(
+                    "Expected binary or string for _data field in resume token but found " + data.getBsonType());
+        }
+    }
+
+    private static byte[] parseHex(final String hexString) {
```
jpechane, line 61: Please check `io.debezium.util.HexConverter` and extend/re-use.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ResumeTokens.java
```diff
@@ -0,0 +1,75 @@
...
+
+    public static BsonValue getData(BsonDocument resumeToken) {
+        if (!resumeToken.containsKey("_data")) {
+            throw new IllegalArgumentException("Expected _data field in resume token");
+        }
+
+        return resumeToken.get("_data");
+    }
+
+    private static byte[] getBytes(BsonValue data) {
```
jpechane, line 45: This probably belongs to `MongoUtils`

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/SourceInfo.java
```diff
@@ -299,6 +300,20 @@ public void initialPosition(String replicaSetName, BsonDocument oplogEvent) {
         onEvent(replicaSetName, CollectionId.parse(replicaSetName, namespace), position, 0L);
     }
 
+    public void cursorEvent(String replicaSetName, MongoChangeStreamCursor<?> cursor) {
```
jpechane, line 303: Please rename to `noEvent` or similar. The logic inside LGTM.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ResumeTokens.java
jcechace: @jpechane  disagree. MongoUtils is slowly becoming a mess of about bilion semi related utility methods. I'd like to keep it here at least until we need it also somewhere else.

jpechane: Then let's make a separate util class for it like `BsonUtil`.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ChangeStreamPipelineFactory.java
```diff
@@ -0,0 +1,219 @@
...
+
+        // Pipeline
+        // Note that change streams cannot use indexes:
+        // - https://www.mongodb.com/docs/manual/administration/change-streams-production-recommendations/#indexes-and-performance
+        // Note that `$addFields` must be used over `$set`/ `$unset` to support MongoDB 4.0 which doesn't support these operators:
+        // - https://www.mongodb.com/docs/manual/changeStreams/#modify-change-stream-output
+        var pipeline = List.of(
+                addFields("namespace", concat("$ns.db", ".", "$ns.coll")), // TODO : This needs to come after the change type filtering since some events don't have these fields!
+                matchFilter,
+                addFields("namespace", "$$REMOVE"));
```
btiernay, line 67: FYI: Not sure where the comment went, but this remove step is indeed required because without it, t driver `ChangeStreamDocument` deserialization fails with:
```
Caused by: org.bson.codecs.configuration.CodecConfigurationException: Failed to decode 'ChangeStreamDocument'. Decoding 'namespace' errored with: readStartDocument can only be called when CurrentBSONType is DOCUMENT, not when CurrentBSONType is STRING.
	at org.bson.codecs.pojo.PojoCodecImpl.setPropertyValue(PojoCodecImpl.java:248)
	at org.bson.codecs.pojo.PojoCodecImpl.decodePropertyModel(PojoCodecImpl.java:214)
	at org.bson.codecs.pojo.PojoCodecImpl.decodeProperties(PojoCodecImpl.java:203)
	at org.bson.codecs.pojo.PojoCodecImpl.decode(PojoCodecImpl.java:112)
	at org.bson.codecs.pojo.PojoCodecImpl.decode(PojoCodecImpl.java:116)
	at com.mongodb.client.model.changestream.ChangeStreamDocumentCodec.decode(ChangeStreamDocumentCodec.java:64)
	at com.mongodb.client.model.changestream.ChangeStreamDocumentCodec.decode(ChangeStreamDocumentCodec.java:35)
	at org.bson.RawBsonDocument.decode(RawBsonDocument.java:172)
	at com.mongodb.client.internal.MongoChangeStreamCursorImpl.getNextInBatch(MongoChangeStreamCursorImpl.java:123)
	at com.mongodb.client.internal.MongoChangeStreamCursorImpl.tryNext(MongoChangeStreamCursorImpl.java:96)
```

Will add a comment.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ResumeTokens.java
jcechace: At closer look this method really makes sense only in context of the `_data` field from resume token. Making it public for anything else than this class doesn't really make sense. Though I would rename it to `getDataByte` or something similar.

btiernay: > Making it public for anything else than this class doesn't really make sense

This method is private, does that change your opinion?

> Though I would rename it to getDataByte or something similar.

Makes sense.

btiernay: > @jpechane disagree. MongoUtils is slowly becoming a mess of about bilion semi related utility methods. I'd like to keep it here at least until we need it also somewhere else.

Imho, I agree with @jcechace here. In my experience, `<X>Utils` tend to become a broad dumping ground that only shares the weakest form of cohesion: [coincidental](https://en.wikipedia.org/wiki/Cohesion_(computer_science)#Types_of_cohesion). While just a convention, the more modern `<X>s` style promoted by libraries such as Guava and the JDK tend to be more focused and akin to "companion classes" in other languages like Scala or Kotlin.

That said, happy to follow the `<X>Utils` convention if that is more common in the code base. Though I would rather not widen scope to `BsonUtils` for the same reasons discussed above.  There is opportunity for reuse / encapsulate in other call sites as well. For example:
- https://github.com/debezium/debezium/blob/fcb0e77bbb0a3f5c7873b89e066f20d08dfc2949/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java#L224-L226
- https://github.com/debezium/debezium/blob/fcb0e77bbb0a3f5c7873b89e066f20d08dfc2949/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/SourceInfo.java#L308

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ChangeStreamPipelineFactory.java
jcechace: I removed about 3 minutes after sending it because I realised just that :)

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
btiernay: [This part of the spec](https://github.com/mongodb/specifications/blob/master/source/change-streams/change-streams.rst#updating-the-cached-resume-token) makes me believe it can still be `null`:
> If an empty batch was returned and a `postBatchResumeToken` was included, cache it.
> [...]
> If it's the last document in the batch and a `postBatchResumeToken` is included, cache it.

Additional context [here](https://www.mongodb.com/docs/manual/changeStreams/#resume-tokens) and [here](https://www.mongodb.com/docs/manual/reference/command/getMore/#mongodb-dbcommand-dbcmd.getMore).

Furthermore, since we support so many versions of Mongo, we cannot rule out things like [SERVER-63772: Post Batch Resume token not sent on initial batch from per shard cursor change stream](https://jira.mongodb.org/browse/SERVER-63772)

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/MongoDbConnectorIT.java
```diff
@@ -1549,8 +1549,9 @@ public void shouldEmitHeartbeatMessages() throws InterruptedException, IOExcepti
         final Integer monitoredOrd = (Integer) monitoredOffset.get(SourceInfo.ORDER);
         assertThat(records.recordsForTopic("__debezium-heartbeat.mongo")).hasSize(1);
         final Map<String, ?> hbAfterMonitoredOffset = records.recordsForTopic("__debezium-heartbeat.mongo").get(0).sourceOffset();
-        assertThat(monitoredTs).isEqualTo((Integer) hbAfterMonitoredOffset.get(SourceInfo.TIMESTAMP));
-        assertThat(monitoredOrd).isEqualTo((Integer) hbAfterMonitoredOffset.get(SourceInfo.ORDER));
+
+        // Change events are sent on empty cursor `getMore` batches. The first empty batch happens prior to the first monitored event
+        assertThat(monitoredTs).isGreaterThanOrEqualTo((Integer) hbAfterMonitoredOffset.get(SourceInfo.TIMESTAMP));
```
btiernay, line 1554: For clarity, I instrumented some logging to show what this test produces:
```
1677166852631 : dispatching "no even"t heart beat with `resumeToken` {"_data": "8263F78903000000022B0229296E04"}
1677166852638 : heart beat forced

1677166852714 : dispatching event ChangeStreamDocument{ operationType=insert, resumeToken={"_data": "8263F78904000000012B022C0100296E5A1004C4C6562927AD41D2800B420ED949E21F46645F6964006463F78904B089C96ADFB60AF90004"}, namespace=dbit.mhb, destinationNamespace=null, fullDocument={"_id": {"$oid": "63f78904b089c96adfb60af9"}, "a": 2, "b": 2}, fullDocumentBeforeChange=null, documentKey={"_id": {"$oid": "63f78904b089c96adfb60af9"}}, clusterTime=Timestamp{value=7203376779275272193, seconds=1677166852, inc=1}, updateDescription=null, txnNumber=null, lsid=null, wallTime=BsonDateTime{value=1677166852705}}
1677166852716 : "change event" heart beat
1677166852717 : heart beat forced

1677166852717 : dispatching no event heart beat with `resumeToken` {"_data": "8263F78904000000012B022C0100296E5A1004C4C6562927AD41D2800B420ED949E21F46645F6964006463F78904B089C96ADFB60AF90004"}
1677166852717 : heart beat skipped due to throttling

2023-02-23 10:40:52,752 INFO   MongoDbConnectorIT||test  Stopping the connector   

1677166853721 : dispatching no event heart beat with `resumeToken` {"_data": "8263F78904000000032B0229296E04"}
1677166853721 : heart beat forced
```

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ChangeStreamPipelineFactory.java
```diff
@@ -0,0 +1,224 @@
...
+        var andFilter = Filters.and(filters);
+        var matchFilter = Aggregates.match(andFilter);
+
+        // Pipeline
+        // Note that change streams cannot use indexes:
+        // - https://www.mongodb.com/docs/manual/administration/change-streams-production-recommendations/#indexes-and-performance
+        // Note that `$addFields` must be used over `$set`/ `$unset` to support MongoDB 4.0 which doesn't support these operators:
+        // - https://www.mongodb.com/docs/manual/changeStreams/#modify-change-stream-output
+        var pipeline = List.of(
+                addFields("namespace", concat("$ns.db", ".", "$ns.coll")), // TODO : This needs to come after the change type filtering since some events don't have these fields!
```
btiernay, line 65: Per the [docs](https://www.mongodb.com/docs/manual/reference/operator/aggregation/concat/#mongodb-expression-exp.-concat):

image

This is also true of [4.0](https://www.mongodb.com/docs/v4.0/reference/operator/aggregation/concat/), so we should be good 👍 .

jcechace (approved):

jcechace (approved):

jcechace:
Great work! Thanks for your contribution.

PR #4305, opened 2023-02-22
DBZ-6143 Unify Oracle TIMESTAMP WITH TIME ZONE format

Naros:
https://issues.redhat.com/browse/DBZ-6143


jpechane:
@Naros Applied, thanks!

PR #4310, opened 2023-02-23
DBZ-6131: Support change stream filtering using MongoDBs aggregation pipeline step.

btiernay:
### [DBZ-6131](https://issues.redhat.com/browse/DBZ-6131): Support change stream filtering using MongoDB's aggregation pipeline step

MongoDB Kafka Source Connector's [pipeline](https://www.mongodb.com/docs/kafka-connector/current/source-connector/configuration-properties/change-stream/#std-label-source-configuration-change-stream) property allows a user to configure:

> An array of aggregation pipelines to run in your change stream.

For example:

```json
[{"$match": { "$and": [{"operationType": "insert"}, {"fullDocument.eventId": 1404 }] } }]
```

This PR adds this capability to the Debezium MongoDB connector for change stream mode. New docs:
image

### Resources
Some related resources:

- https://www.mongodb.com/docs/kafka-connector/current/source-connector/configuration-properties/change-stream/#std-label-source-configuration-change-stream
- https://github.com/mongodb/mongo-kafka/blob/41d7f8760d6785c488fba408736c2e480ccddfa5/src/main/java/com/mongodb/kafka/connect/source/MongoCopyDataManager.java#L187

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
@@ -509,6 +509,17 @@ public class MongoDbConnectorConfig extends CommonConnectorConfig {
             .withImportance(Importance.LOW)
             .withDescription("The maximum processing time in milliseconds to wait for the oplog cursor to process a single poll request");
 
+    public static final Field CURSOR_PIPELINE = Field.create("cursor.pipeline")
+            .withDisplayName("Pipeline expression apply to the change stream cursor")
+            .withType(Type.STRING)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withValidation(MongoDbConnectorConfig::validateChangeStreamPipeline)
+            .withDescription("Applies processing to change events as part of the the standard MongoDB aggregation stream pipeline. " +
+                    "A pipeline is a MongoDB aggregation pipeline composed of instructions to the database to filter or transform data. " +
+                    "This can be used customize the data that the connector consumes. " +
+                    "Note that this comes after the internal pipelines used to support the connector (e.g. filtering database and collection names).");
+
     public static final Field TOPIC_NAMING_STRATEGY = Field.create("topic.naming.strategy")
             .withDisplayName("Topic naming strategy class")
             .withType(Type.CLASS)
@@ -609,6 +620,19 @@ public class MongoDbConnectorConfig extends CommonConnectorConfig {
         }
         catch (Exception e) {
             problems.accept(field, value, "Connection string is invalid");
+            return 1;
+        }
+        return 0;
+    }
+
+    private static int validateChangeStreamPipeline(Configuration config, Field field, ValidationOutput problems) {
+        String value = config.getString(field);
+
+        try {
+            new ChangeStreamPipeline(value);
+        }
+        catch (Exception e) {
+            problems.accept(field, value, "Change stream pipeline JSON is invalid: " + e.getMessage());
             return 1;
         }
         return 0;
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ChangeStreamPipeline.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ChangeStreamPipeline.java
@@ -1,0 +1,77 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.mongodb;
+
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
+import org.bson.Document;
+import org.bson.conversions.Bson;
+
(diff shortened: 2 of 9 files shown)


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
```diff
@@ -509,6 +510,17 @@ public boolean isIncludePreImage() {
...
+    public static final Field CURSOR_PIPELINE = Field.create("cursor.pipeline")
+            .withDisplayName("Pipeline expression apply to the change stream cursor")
+            .withType(Type.STRING)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withValidation(MongoDbConnectorConfig::validateAggregationPipeline)
+            .withDescription("Applies processing to change events as part of the the standard MongoDB aggregation stream pipeline. " +
+                    "A pipeline is a MongoDB aggregation pipeline composed of instructions to the database to filter or transform data. " +
+                    "This can be used customize the data that the connector consumes. " +
+                    "Note that this comes after the internal pipelines used to support the connector (e.g. filtering database and collection names).");
```
btiernay, line 522: Curious how this is used vs the AsciiDoc documents. Why have 2?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
```diff
@@ -614,6 +626,25 @@ private static int validateConnectionString(Configuration config, Field field, V
         return 0;
     }
 
+    private static int validateAggregationPipeline(Configuration config, Field field, ValidationOutput problems) {
+        String value = config.getString(field);
+
+        try {
+            if (value != null && !value.isEmpty()) {
+                // Top-level for `parse` must be a document not a list
```
btiernay, line 634: Same approach used by MongoDB:
https://github.com/mongodb/mongo-kafka/blob/41d7f8760d6785c488fba408736c2e480ccddfa5/src/main/java/com/mongodb/kafka/connect/util/ConfigHelper.java#L80

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
jcechace: This is used in by Debezium UI I believe.

rk3rn3r: Yes, it is. All Field/s are. ;)

jcechace: Well that's the idea... they don't always much exactly as these are written mainly by engineering and "real" docs are often edited by documentation people.

jcechace:
@btiernay  we should think about how to approach the counterpart to this for snapshotting. Right Now the snapshot would capture everything. Note that snapshot is implemented essentially as calling `find()` on on each collection. 

It should be generally OK to merge this and add support for snapshots in later PR if you prefer to test this in your staging.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
Naros: Just a comment. Please disregard.

btiernay:
> @btiernay we should think about how to approach the counterpart to this for snapshotting. Right Now the snapshot would capture everything. Note that snapshot is implemented essentially as calling find() on on each collection.

I actually considered this for the PR but noticed we already had [this](https://debezium.io/documentation/reference/stable/connectors/mongodb.html#mongodb-property-snapshot-collection-filter-overrides):

image

jcechace:
> > @btiernay we should think about how to approach the counterpart to this for snapshotting. Right Now the snapshot would capture everything. Note that snapshot is implemented essentially as calling find() on on each collection.
> 
> I actually considered this for the PR but noticed we already had [this](https://debezium.io/documentation/reference/stable/connectors/mongodb.html#mongodb-property-snapshot-collection-filter-overrides):
> 
> image

Hmm... we should definitively keep this property/properties working however I would like to see the approach unified. Ideally the snapshot would also use this `ChangeStreamPipeline` and `ChangeStreamPipelineFactory` with the ability to incorporate these existing properties. 

These properties also support only collection filter whereas pipeline can do much more.

jcechace:
Merged, thanks!

btiernay:
> Hmm... we should definitively keep this property/properties working however I would like to see the approach unified. Ideally the snapshot would also use this ChangeStreamPipeline and ChangeStreamPipelineFactory with the ability to incorporate these existing properties.
> 
> These properties also support only collection filter whereas pipeline can do much more.

Agreed, although there is a slight difference that is material. `db.watch` is a wrapper around `db.aggregate` whereas `db.find` does not support aggregation operators and the expressions are thus different. That said, they could probably be combined somehow. In general, I think we could rationalize a lot of the configuration to work more ergonomically with the driver, but that would require a few major releases and removing some of the EOL Mongo versions.

PR #3901, opened 2022-09-21
DBZ-5632 Incremental snapshot completion notifications

kgalieva:
Added an optional configuration property to switch between logging snapshot completion status (`LogSnapshotStatusNotifications` - default) and sending the notification into a Kafka topic (`KafkaSnapshotStatusNotifications`).

A Kafka notification format:

Key = `test_connector` 
Value = `{"id" : "1b8d98f4-53ec-4beb-aa14-73ea35f185e7", "type":"incremental-snapshot", "data-collections": "schema1.table1", "additional-condition" : "updated_at>='2022-09-20'", "status" : "SUCCEEDED"}`

Snapshot completion statuses:
```
EMPTY,
NO_PRIMARY_KEY,
SQL_EXCEPTION,
STOPPED,
SUCCEEDED,
UNKNOWN_SCHEMA
```

Closes: https://issues.redhat.com/browse/DBZ-5632

PR #4314, opened 2023-02-26
Add stubbed snapshot status notification

xinbinhuang:
Inspired by this PR: [DBZ-5632](https://github.com/debezium/debezium/pull/3901)
Related to: https://issues.redhat.com/browse/DBZ-5464

Add a stubbed version of the rough API interface for sending snapshot status update through kafka connect to a metadata topic. This's a bit far from a generic central metadata topic that can receive different types of messages, but I think we can refactor the `SnapshotStatusNotification` easily if the overall mechanism/initial direction seems reasonable. The overall direction that I'm thinking is to:
- have a handler in `EventDispatcher` that dispatch events similar to how it dispatch heartbeat events 
- Then other pieces of the code and use `EventDispatcher` to dispatch different types of events that implement an `Interface`.
  - Question: what information metadata should be provide?
    - I'm thinking it'll be generic information (s.g. `SourceParititon`) and context specific information (e.g. `SnapshotStatus`)

I hope to get some early feedback to make sure I'm on the right track before going deeper.

@jcechace @jpechane

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbIncrementalSnapshotChangeEventSource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbIncrementalSnapshotChangeEventSource.java
@@ -29,11 +29,8 @@ import io.debezium.annotation.NotThreadSafe;
 import io.debezium.connector.mongodb.ConnectionContext.MongoPreferredNode;
 import io.debezium.pipeline.EventDispatcher;
 import io.debezium.pipeline.source.AbstractSnapshotChangeEventSource;
-import io.debezium.pipeline.source.snapshot.incremental.CloseIncrementalSnapshotWindow;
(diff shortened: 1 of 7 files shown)


jpechane:
@xinbinhuang Thank you very much for the PR. If possible I'd propose to move the work further/deeper.
I'd like to make this more general. In the same way there is the generic `Signal` facitily to send commands to Debezium there should be a counterpat `Notification` that will be for sending mssages from Debezium.
The first notification will be the snpashot status notification.
Then there will be a notion of `Channel` that will be used as destination for the `Notification`. The initial implementation will be a `System.out` and kafka topic channels. Channels will be enabled via connector configuration - a config option that will enumaret enabled channels and a config option set per channel.

jpechane:
Closing in favour of https://github.com/debezium/debezium/pull/4515
@xinbinhuang Any feedback on the other PR welcome.
Thanks for your time and effort!

PR #4211, opened 2023-01-22
DBZ-6032 Connection management refactoring

jcechace:
Docs will be updated in a separate PR later today / tomorrow

PR #4315, opened 2023-02-26
DBZ-5953: Add support for honouring MongoDB read preference in change stream after promotion

btiernay:
### [DBZ-5953](https://issues.redhat.com/browse/DBZ-5953): Add support for honouring MongoDB read preference in change stream after promotion

Docs:
image


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -262,8 +245,62 @@ else if (oplogStart.getTime() > 0) {
...
+        var candidates = new ReadPreferenceServerSelector(readPreference)
+                .select(client.getClusterDescription());
+
+        // Determine if the cursor matches any one of these candidates
+        // TODO: In general, the host names returned between commands can be different (DNS or IP potentially).
+        //       For example, we see test-mongo1:27017 (defined in rs.init()) and e253ef34bd93:27017 (the container’s hostname).
+        //       Thus, canonicalization must be performed when making comparison.
+        return candidates.stream()
+                .map(ServerDescription::getAddress)
+                .anyMatch(address -> address.equals(cursor.getServerCursor() == null ? null : cursor.getServerCursor().getAddress()));
```
btiernay, line 303: This is my biggest concern with the approach as it could lead to false positives in certain configurations / topologies.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -262,8 +245,62 @@ else if (oplogStart.getTime() > 0) {
...
                 }
 
+                if (taskContext.getConnectorConfig().isCursorReadPreferenceMonitoringEnabled() && preferenceRateLimiter.tryConsume(1)
+                        && !isSelectedReadPreference(client, mongo.getPreference(), cursor)) {
+                    LOGGER.info("Closing and recreating cursor due to election invalidating read preference");
+                    cursor.close();
+                    // Note: There is an edge case currently unaccounted for when an election happens during the first iteration of the cursor.
+                    // In this case, we need to set resume token in rsOffsetContext for this to skip the initial snapshot last timestamp.
+                    // But this is currently immutable.
+                    rsChangeStream = createChangeStream(client, rsOffsetContext);
```
btiernay, line 255: @jcechace  Will this _always_ result in a reevaluation of the read preference?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -179,31 +181,12 @@ private void readChangeStream(MongoClient client, MongoPreferredNode mongo, Repl
...
-
-        if (connectorConfig.getCursorMaxAwaitTime() > 0) {
-            rsChangeStream.maxAwaitTime(connectorConfig.getCursorMaxAwaitTime(), TimeUnit.MILLISECONDS);
-        }
+        var rsChangeStream = createChangeStream(client, rsOffsetContext);
+        var cursor = rsChangeStream.cursor();
 
-        try (MongoChangeStreamCursor<ChangeStreamDocument<BsonDocument>> cursor = rsChangeStream.cursor()) {
+        // Once every 10 seconds
+        var preferenceRateLimiter = new RateLimiter(1, 0.1F);
```
btiernay, line 188: Should this be configurable?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -262,8 +245,65 @@ else if (oplogStart.getTime() > 0) {
...
+
+        // Determine if the cursor matches any one of these candidates
+        return candidates.stream()
+                .map(ServerDescription::getAddress)
+                .anyMatch(preferredAddress -> {
+                    // TODO: In general, the host names returned between commands can be different (DNS or IP potentially).
+                    // For example, we see test-mongo1:27017 (defined in rs.init()) and e253ef34bd93:27017 (the container’s hostname).
+                    // Thus, canonicalization must be performed when making comparison.
+                    var cursorAddress = cursor.getServerCursor() == null ? null : cursor.getServerCursor().getAddress();
+                    return preferredAddress.equals(cursorAddress);
```
btiernay, line 305: This is my biggest concern with the approach as it could lead to false positives in certain configurations / topologies. Not sure if there is a better way unfortunately, but open to ideas.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
jcechace: I believe we discussed this yesterday and the answer was no?

btiernay: The answer is yes, but subject to the current constraints that the connection is "direct" and single node, thus it can never work like this until we can relax that constraint.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
jcechace: Though I do like this as it is the standard leaky bucket algorithm, there already is `io.debezium.util.DelayStrategy`, look whether that can be used please

btiernay: Turns out I don't really need to use it after all, but this is good to know. Thanks!

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
btiernay: Reworked this area, so resolving.

btiernay:
Closing since https://github.com/debezium/debezium/pull/4211 obviates the need for this PR.

PR #4316, opened 2023-02-26
DBZ-5181 Multi-threaded Debezium Engine

jeremy-l-ford:
relates to https://issues.redhat.com/browse/DBZ-5181


debezium-api/src/main/java/io/debezium/engine/spi/OffsetCommitPolicy.java
```diff
@@ -24,6 +24,9 @@ public interface OffsetCommitPolicy {
      */
     class AlwaysCommitOffsetPolicy implements OffsetCommitPolicy {
 
+        public AlwaysCommitOffsetPolicy(Properties properties) {
+        }
+
```
Naros, line 29: If the provided `properties` object isn't being used, then I believe this is mostly a superfluous change and we can simply not pass it as a ctor argument.

debezium-api/src/main/java/io/debezium/engine/spi/OffsetCommitPolicy.java
```diff
@@ -50,7 +53,7 @@ public boolean performCommit(long numberOfMessagesSinceLastCommit, Duration time
     }
 
     static OffsetCommitPolicy always() {
-        return new AlwaysCommitOffsetPolicy();
+        return new AlwaysCommitOffsetPolicy(null);
```
Naros, line 56: Adjust based on the comment about ctor argument.

debezium-embedded/src/main/java/io/debezium/embedded/DefaultOffsetManager.java
```diff
@@ -0,0 +1,93 @@
...
+
+import io.debezium.config.Configuration;
+import io.debezium.config.Field;
+import io.debezium.config.Instantiator;
+
+/**
+ * Default implementation of {@link OffsetManager} that uses Kafka's OffsetStorageWriter to commit offsets.
+ * This class is meant to be used in a thread confined manner and is not thread safe.
+ */
+public class DefaultOffsetManager implements OffsetManager {
```
Naros, line 30: I'm inclined that we rename this from `DefaultOffsetManager` to `KafkaOffsetManager`, following a similar naming that we use for storage-based implementations as well as the `OffsetBackingStore` implementations.  

Also, annotate this class with `@NotThreadSafe` if its expressly not thread-safe.

debezium-embedded/src/main/java/io/debezium/embedded/OffsetManager.java
```diff
@@ -0,0 +1,38 @@
...
+import org.apache.kafka.connect.storage.OffsetStorageReader;
+import org.apache.kafka.connect.storage.OffsetStorageWriter;
+
+import io.debezium.config.Configuration;
+import io.debezium.config.Field;
+
+/**
+ * Interface for committing offsets.
+ */
+public interface OffsetManager {
```
Naros, line 19: Let's be sure to document methods on interfaces for clarity.

debezium-embedded/src/main/java/io/debezium/embedded/TaskOffsetManager.java
```diff
@@ -0,0 +1,60 @@
...
+package io.debezium.embedded;
+
+import org.apache.kafka.connect.source.SourceRecord;
+import org.apache.kafka.connect.storage.OffsetStorageReader;
+
+import io.debezium.config.Configuration;
+import io.debezium.config.Field;
+import io.debezium.engine.spi.OffsetCommitPolicy;
+
+public interface TaskOffsetManager {
```
Naros, line 15: Again, document methods on interfaces.

Naros (changes requested):
Hi @jeremy-l-ford this is a quick first pass at a glance, I'll be taking a closer look at the details later today and the weekend.

debezium-api/src/main/java/io/debezium/engine/spi/OffsetCommitPolicy.java
jeremy-l-ford: This change stems from the DefaultTaskOffsetManager dynamically creating the OffsetCommitPolicy.  It's assuming all implementations will have a constructor with a Properties parameter.

jeremy-l-ford:
@Naros - I was looking at this again over the weekend.  Based on how Kafka Connect works with standalone mode,  I feel like the OffsetBackingStore should be owned by the engine and each TaskWorker should own its offset reader and writer.  I made a change moving towards that, but there's more work needed there still.

PR #4317, opened 2023-02-27
Just for test oracle ci

GOODBOY008:


Naros:
Hi @GOODBOY008, is this PR to address something specific?  It seems the issue tag and the changes are superfluous.

GOODBOY008:
> Hi @GOODBOY008, is this PR to address something specific? It seems the issue tag and the changes are superfluous.

I just want to locate oracle cdc problem ,which column name is larger 30 char with dbz.

Naros:
Yep, Oracle LogMiner (despite the database being configured with Oracle 12+ long object name support) will still have issues when the column name or table name is larger than 30 characters.

PR #4318, opened 2023-02-27
DBZ-6163 Properly handle zoned timestamp within PostgreSQL

Naros:
https://issues.redhat.com/browse/DBZ-6163


jpechane:
@Naros Applied, thanks

PR #4319, opened 2023-02-27
DBZ-6130 Remove references to using properties files to configure SMTs

roldanbob:
[DBZ-6130](https://issues.redhat.com/browse/DBZ-6130)

Updates SMT docs to refer to adding SMTs settings via the connector configuration rather than via `.properties` files.


Naros:
Applied to main (=2.2) and branch `2.1`.

PR #4320, opened 2023-02-27
DBZ-6073 Move note text and eliminate embedded community conditional

roldanbob:
[DBZ-6073](https://issues.redhat.com/browse/DBZ-6073)

Fixes overlapping/embedded conditionals in the Oracle connector deployment doc and removes a redundant reference in the community content to obtaining the Oracle JDBC driver. The duplicate content was inserted in error, because it appeared to have been omitted. In fact, the original text was present, but it was contained in a `community` conditional that was embedded within a `product` conditional, preventing it from rendering upstream.
This change preserves a note that was contained in the deleted duplicate section about how Xstream user can obtain the `ojdbc8` driver file. The note now appears in the context of Step 3 of the upstream deployment procedure.
  

jpechane:
@roldanbob Applied, thanks

PR #4321, opened 2023-03-01
DBZ-6124 Modify CREATE and ALTER TABLE rules

ani-sha:
https://issues.redhat.com/browse/DBZ-6124


debezium-ddl-parser/src/test/resources/mysql/examples/ddl_alter.sql
```diff
@@ -44,6 +44,15 @@ alter table add_test drop foreign key if exists fk;
 alter table add_test drop constraint if exists cons;
 alter table add_test wait 100 add column col1 int not null;
 alter table default.task add column xxxx varchar(200) comment 'cdc test';
+
+ALTER TABLE T1 ADD FOREIGN KEY ( I )  REFERENCES TT ( I ) ON DELETE SET DEFAULT;
+ALTER TABLE T1 ADD FOREIGN KEY ( I ) REFERENCES TT ( I ) ON UPDATE SET DEFAULT;
+ALTER TABLE T1 ADD CHECK (id + 6 > 10) ENFORCED;
+ALTER TABLE T1 ADD CHECK (ID + 6 > 10) NOT ENFORCED;
+--ALTER TABLE T1 ALTER CHECK C_CONS ENFORCED;
+--ALTER TABLE T1 ALTER CHECK C_CONS NOT ENFORCED;
```
ani-sha, line 53: @Naros Can you suggest what would be the right syntax for these DDLs?

Naros: So the failure is related to the fact this portion was removed from the `dataType` rule:
```
    | NATIONAL typeName=(CHAR | CHARACTER) VARYING
      lengthOneDimension? BINARY?                                   #nationalVaryingStringDataType
```
The `ColumnDefinitionParserListener` had several blocks that depended on this varying string data type that got merged into the national string data type.  I wonder if this could have been fixed by adjusting the order of the rules.

Naros: I pushed 1 additional commit, let's see if that resolves the issue.

ani-sha: @Naros Ah, I see, yes certainly the order of the rule matters. Thanks for correcting it.

Btw the other two `ALTER` DDLs, what would be the correct rule for them? I couldn't seem to fit them in anywhere.

ani-sha:
@jpechane Rebased to latest main, should be good to merge as well!

jpechane:
@ani-sha Applied, thanks!

ani-sha:
Upstream PR - https://github.com/antlr/grammars-v4/pull/3263

PR #4322, opened 2023-03-01
DBZ-6155 Adjust LogMiner batch size based on the default batch size rather than current batch size.

smallYellowCat:
Using current batch size for comparison is wrong in case that when currentScn is topScn as described in DBZ-6155.  
In the other case (when topScn is behind currentScn)  can eventually lead to this situation.
when topScn would fall behind currentScn more and more as we would compare currentScn - topScn with bigger and bigger number (current batch size).

Comapring with logMiningBatchSizeMin could result in very small window which would mean we will send many small queries instead of several bigger ones during mining.
Therefore reverting back to do the adjustments based on the defaultBatchSize.

### DIFF
--- a/debezium-core/src/main/java/io/debezium/relational/HistorizedRelationalDatabaseConnectorConfig.java
+++ b/debezium-core/src/main/java/io/debezium/relational/HistorizedRelationalDatabaseConnectorConfig.java
@@ -106,6 +106,7 @@ public abstract class HistorizedRelationalDatabaseConnectorConfig extends Relati
         // Do not remove the prefix from the subset of config properties ...
         Configuration schemaHistoryConfig = config.subset(SchemaHistory.CONFIGURATION_FIELD_PREFIX_STRING, false)
                 .edit()
+                .with(config.subset(Field.INTERNAL_PREFIX + SchemaHistory.CONFIGURATION_FIELD_PREFIX_STRING, false))
                 .withDefault(SchemaHistory.NAME, getLogicalName() + "-schemahistory")
                 .withDefault(AbstractSchemaHistory.INTERNAL_CONNECTOR_CLASS, connectorClass.getName())
                 .withDefault(AbstractSchemaHistory.INTERNAL_CONNECTOR_ID, logicalName)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -652,11 +652,12 @@ private Scn calculateEndScn(OracleConnection connection, Scn startScn, Scn prevE
 
         // Control adjusting batch size
         boolean topMiningScnInFarFuture = false;
-        if (topScnToMine.subtract(currentScn).compareTo(currentBatchSizeScn) > 0) {
+        final Scn minBatchScn = Scn.valueOf(connectorConfig.getLogMiningBatchSizeMin());
```
Naros, line 655: Why min-batch-size rather than default-batch-size?

smallYellowCat: > Why min-batch-size rather than default-batch-size?

because the unit of every time increment or reduce batchSizeScn is minBatchSize.

e.g: default = 20, min = 5, max = 100;
currentScn = 10; start = 5. currentBath = 15;

so.
topScn = start + currentBatch = 20
topScn - current = 10; if we compare with default, it's small than default. will not reduce currentBatchSize.

increment scene is same.

vjuranek: @smallYellowCat yes, these are the facts, but what is the actual reason to base the sliding window on minimum batch size? If we do so, it would result in very small window around current SCN position so we would need to send many small queries for mining instead of several bigger ones. Won't it negatively impact the performance?  Also, default batch size is let's say desired batch size specified by the user so IMHO it's more natural to aim for the window size similar to the default batch size than the batch size close to the minimum batch size - which is the lower bound for the batch size set by user. Therefore I think we should use default batch size as a measure - which I did in PR #4327 (as I already did yesterday, and noticed this PR when I sent it today).

smallYellowCat: @vjuranek  yes, i checked your code and analysied this method again.   for increment batchSize scene, i thinke we shoudle compare with currentBatchSize.  for reduce scene we need to compare with defaultScn.
 the reason is:

1. for increment batchSize scene, image below scene:

  _______|_________5_________|__________________5 or 10__________|
          StartScn                 TopScn                             CurrentScn

currentBatchSize = 5, if  currentScn - TopScn = 5,  we can assume next time TopScn will arrive CurrentScn position. so We can tolerate not expanding the sliding window in this case.  but if cuurentScn - topScn > currentBatchSize, we should inreament batchSize.


2.  for reduce batchSize scene, image below Scene:

_________|___________________|________________________________|
           startScn               currentScn                                 topScn

if topScn > currentScn, normally we should reduce batchSize, but as u said, if slide window is to small,  and source db have a large of CDC event, it will occured mall small  range query, could be have performence impact.


So  i will change my code as soon as possible,  thank u suggestion.

tooptoop4: @smallYellowCat does this PR fix potential data loss (related to some SCNs not being mined) or is it just a performance fix?

smallYellowCat: @tooptoop4  performance fix

smallYellowCat: @Naros hi naros, could u pls review my code again? thank u .

vjuranek: I'd stick with default batch size also in the case when `topScn` is behind `currentScn` because `currentScn` would very likely move as well and it's definitely not guarantee `topScn` will catch up `currentScn` in the next iteration. Actually it can make things even worse as the current batch size would grow to catch `currentScn` you would compare `currentScn-topScn` with bigger and bigger number which effectively would mean that the distance between `currentScn` and `topScn` could increase over the time (if `currentScn` would grow fast enough - if not, this is not the issue, `topScn` would catch up with `currectScn` and we will start to decrease the batch size). Minor reason is that having constant number (default batch size) makes reasoning about the behavior more easy than the case when limits are based on constantly changing number (current batch size). So unless we have very good reasons (like performance comparison showing default batch size is not a good measure) I'd stick with default batch size in both cases.

smallYellowCat: @vjuranek Yes, what you said is no problem. It is only a trade-off between default and current. It is true that default is more reasonable in terms of logic.

smallYellowCat: @vjuranek so can u change this PR and merge it?

vjuranek: @smallYellowCat the final review and merge need to be done by @Naros . One more thing: could you please update commit message explaining, so that when someone in the future wonder why it was change in this way, he can easily find the reasoning in the commit message? Thanks!

smallYellowCat: @vjuranek ok, i will change commit message and explain why  need to do this change.

vjuranek: @smallYellowCat please let me know once you are done. Please include the summary of this discussion - why current batch size is wrong, why it's better to use default batch size than the current batch size etc. Thanks!

vjuranek: @smallYellowCat what about summary like this (feel free to adjust as you like):

Using current batch size for comparison is wrong in case that when
`currentScn` is `topScn` as described in  DBZ-6155. In the other case (when
`topScn` is behind `currentScn`) is can eventually lead to situation
when `topScn` would fall behind `currentScn` more and more as we would
compare `currentScn - topScn` with bigger and bigger number (current
batch size).
Comapring with `logMiningBatchSizeMin` could result in very small window which
would mean we will send many small queries instead of several bigger
ones during mining.
Therefore reverting back to do the adjustments based on the `defaultBatchSize`.

Please also note that you updated description here on the GH, but not the commit message itself.

smallYellowCat: @vjuranek thank u  vjuranek,  i will change it when i free.

smallYellowCat: @vjuranek both  GH commit and PR commit sumary have changed, could u check it.

smallYellowCat:
@vjuranek oh, ok,  don't worry.   merge your PR is ok.

PR #4323, opened 2023-03-01
Dbz 6112 - PostgreSQL: Set Replica Identity when the connector starts

miguelbirdie:
Adding a new configuration parameter called `replica.autoset.type` to be able to set `REPLICA IDENTITY` values for all tables captured in the connector.


benw-at-birdie (changes requested):
Mostly soft suggestions. Just two actual changes requested.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
```diff
@@ -596,6 +597,26 @@ public static AutoCreateMode parse(String value, String defaultValue) {
                             "the current filter configuration (see table/database include/exclude list properties). If the publication already" +
                             " exists, it will be used. i.e CREATE PUBLICATION <publication_name> FOR TABLE <tbl1, tbl2, etc>");
 
+    public static final Field REPLICA_IDENTITY_AUTOSET_VALUES = Field.create("replica.identity.autoset.values")
```
jpechane, line 600: Would it be possible to introduce an `enum` that will represent allowd values? In case of `INDEX` the index name would be provided via a method.
Or `io.debezium.connector.postgresql.connection.ServerInfo.ReplicaIdentity` should be re-used - maybe extract into a standalone class.

documentation/modules/ROOT/pages/connectors/postgresql.adoc
```diff
@@ -2947,6 +2947,24 @@ For example: `CREATE PUBLICATION <publication_name> FOR TABLE <tbl1, tbl2, tbl3>
 If the publication exists, the connector updates the publication for tables that match the current filter configuration.
 For example: `ALTER PUBLICATION <publication_name> SET TABLE <tbl1, tbl2, tbl3>`.
 
+|[[postgresql-replica-autoset-type]]<<postgresql-replica-autoset-type, `+replica.identity.autoset.values+`>>
+|_empty string_
+|Applies only when streaming changes by using link:https://www.postgresql.org/docs/current/sql-createpublication.html[the `pgoutput` plug-in].
```
jpechane, line 2952: Why it should be alloed only for `pgoutput`?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
```diff
@@ -764,6 +825,13 @@ public ReplicationConnectionBuilder withPlugin(final PostgresConnectorConfig.Log
             return this;
         }
 
+        @Override
+        public ReplicationConnectionBuilder withReplicaIdentity(ReplicaIdentityMapper replicaIdentityMapper) {
```
jpechane, line 829: This should be unnecessary as `PostgresConnectorConfig` is passed to the class constructor.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/ReplicaIdentityMapper.java
```diff
@@ -0,0 +1,110 @@
...
+import io.debezium.relational.TableId;
+
+/**
+ * Class that records Replica Identity information for the {@link PostgresConnector}
+ * @author Ben White, Miguel Sotomayor
+ */
+@Immutable
+public class ReplicaIdentityMapper {
+
+    public static class ReplicaIdentityValue {
```
jpechane, line 25: Again the same `enum` could be used.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/ReplicaIdentityMapper.java
```diff
@@ -0,0 +1,110 @@
...
+                .collect(Collectors.toMap(
+                        t -> {
+                            String[] tableName = t.group("tablename").split("\\.");
+                            return new TableId(databaseName, tableName[0], tableName[1]);
+                        },
+                        t -> new ReplicaIdentityValue(
+                                ServerInfo.ReplicaIdentity.valueOf(t.group("replicaidentity").replaceAll("\\s.*", "")), t.group("indexname"))));
+    }
+
+    public Map<TableId, ReplicaIdentityValue> getMapper() {
```
jpechane, line 106: I think the naming really needs to be improved. The whole class that manages mapping is named mapper and this return map which is an internal data. Cannot this class provide the business logic needed so it is not necessary to expaose the map?

jpechane (changes requested):
@miguelbirdie Thanks for the PR. I left few comments related to the code organization.

documentation/modules/ROOT/pages/connectors/postgresql.adoc
miguelbirdie: I don't actually have a reason to use the Replica Identity only for `pgoutput`. So, let's do it for `decoderbufs` as well.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/ReplicaIdentityMapper.java
benw-at-birdie: Thanks @jpechane we think this makes sense. So just to check what you are proposing is: 
- Convert `io.debezium.connector.postgresql.connection.ServerInfo.ReplicaIdentity` to a standard class which includes an enum for the different replica identity options (`FULL`, `DEFAULT`, etc) and a string representing the index field for the replica identity when the option is `INDEX`
- Migrate to this class the current business logic in `ReplicaIdentityMapper` which will remove the need to expose the `Map<TableId, ReplicaIdentityValue>` map.

Let us know if that's right in the right direction and we'll update the PR accordingly.

cc @miguelbirdie

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
miguelbirdie: >Would it be possible to introduce an enum that will represent allowd values?

@jpechane Keep in mind that this property still needs to be a `String` as users will need to set up `REPLICA IDENTITY` for each table captured (at least for a subset of these tables).

jpechane: ok, makes sense. Just please use comma as separator, not semicolon to make it consistent with `table.include.list` etc.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/ReplicaIdentityMapper.java
jpechane: Yes, I beleive we are on the same page.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
miguelbirdie: Removing `withReplicaIdentity` means that `replicaIdentityMapper` is always null. 
`publicationAutocreateMode` is also passed to the class constructor and `withPublicationAutocreateMode` exists

miguelbirdie:
We have made the following changes:
- The `ReplicaIdentityValue` class has been removed, and we are now working directly with the `ReplicaIdentity` enumerator that existed previously.
- The `ReplicaIdentityMapper` class is still used to manage the `replica.identity.autoset.values` property, but the map is no longer exposed outside of this class.
- The `replica.identity.autoset.values` property now uses a comma instead of a semicolon.

miguelbirdie:
@jpechane We've introduced some changes to allow regular expressions in the new property `replica.identity.autoset.type`
Now, we can specify something like this:
```
schema1.*:FULL,schema2.table2:NOTHING,schema2.table3:INDEX idx_name
```

miguelbirdie:
@jpechane Can you execute all checks for this Pull Request please?
I've built the artifact (including test) for PostgreSQL connector in my local machine and it's working
```
mvn clean install -B -pl debezium-connector-postgres -am -Passembly,postgres-15,pgoutput-decoder -Ddebezium.test.records.waittime=5 -Dcheckstyle.skip=true -Dformat.skip=true -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120

```
image

Thanks!

jpechane:
@miguelbirdie Applied via https://github.com/debezium/debezium/pull/4504, thanks!

PR #4324, opened 2023-03-02
[docs] Add additional properites for  http sink

jeremy-l-ford:


jpechane:
@jeremy-l-ford Applied, thanks

PR #4325, opened 2023-03-02
DBZ-6115 Use correct format for output

ani-sha:
https://issues.redhat.com/browse/DBZ-6115

### DIFF
--- a/.github/workflows/oracle-workflow-test.yml
+++ b/.github/workflows/oracle-workflow-test.yml
@@ -40,24 +40,24 @@ jobs:
   check:
     runs-on: ubuntu-latest
     outputs:
-      SKIP_CHECK: ${{ steps.check.outputs.SKIP_CHECK }}
+      SKIP_CHECK: ${{ steps.skip.outputs.SKIP_CHECK }}
     steps:
       - name: Checkout Action
         uses: actions/checkout@v3
 
       - name: Check if author has Quay.io credentials
+        id: skip
         env:
           QUAY_IO_USERNAME: ${{ secrets.QUAY_IO_USERNAME }}
           QUAY_IO_PASSWORD: ${{ secrets.QUAY_IO_PASSWORD }}
         run: |
-          if [[ $QUAY_IO_USERNAME != '' && $QUAY_IO_PASSWORD != '' ]]; then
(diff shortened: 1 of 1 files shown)


ani-sha:
Hi @Naros. I have set some dummy data for the credentials. Clearly the data isn't exposed through env variables. But I am failing to get through why the `build` step is not being triggered after the `check` step. Could you please look through and see in case I have missed anything. Thanks.

.github/workflows/oracle-workflow-test.yml
```diff
@@ -40,24 +40,25 @@ jobs:
...
-            echo "SKIP_CHECK=false" >> $GITHUB_OUTPUT
+          if [[ $QUAY_IO_USERNAME == '' && $QUAY_IO_PASSWORD == '' ]]; then
+            echo "SKIP_CHECK=true" >> $GITHUB_OUTPUT
           fi
 
   build:
     runs-on: ubuntu-latest
     needs: [check]
-    if: ${{ needs.check.outputs.SKIP_CHECK == 'false' || github.event_name == 'push' }}
+    if: ${{ needs.check.outputs.SKIP_CHECK != 'true' || github.event_name == 'push' }}
```
vjuranek, line 61: Maybe better to use `github.event_name == 'pull_request'` so that the tests are run for PRs ... or event better to trigger it for both (can be written also as `contains(fromJSON('["push", "pull_request"]'), github.event_name)`)?

ani-sha: @vjuranek So any changes to this workflow can be put to test only after it's in the `main` since it is using `pull_request_target`. Any other way to test it?

vjuranek: @ani-sha just add you fork, debug/test it and once done remove it

ani-sha:
Hi @vjuranek, tested this on my fork and it works both for `push` and `pull request`. It's good to merge IMO. Please let me know your suggestions.

vjuranek (approved):
LGTM

jpechane:
@ani-sha Applied, thanks!
@vjuranek Could you please try to create a PR with an Oracle dummy change to double check it works? Thanks

PR #4327, opened 2023-03-02
DBZ-6155 Adjust LogMiner batch size based on the default batch size

vjuranek:
https://issues.redhat.com/browse/DBZ-6155

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
@@ -646,17 +646,16 @@ public class LogMinerStreamingChangeEventSource implements StreamingChangeEventS
                 : connection.getCurrentScn();
         streamingMetrics.setCurrentScn(currentScn);
 
-        // Add the current batch size to the starting system change number
-        final Scn currentBatchSizeScn = Scn.valueOf(streamingMetrics.getBatchSize());
-        Scn topScnToMine = startScn.add(currentBatchSizeScn);
+        final Scn defaultBatchSize = Scn.valueOf(connectorConfig.getLogMiningBatchSizeDefault());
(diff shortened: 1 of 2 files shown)


jpechane:
@Naros Could you please take a look and review this one? It seems that it partly reverts DBZ-5005.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -647,16 +647,16 @@ private Scn calculateEndScn(OracleConnection connection, Scn startScn, Scn prevE
         streamingMetrics.setCurrentScn(currentScn);
 
         // Add the current batch size to the starting system change number
```
jpechane, line 649: @vjuranek Does the comment still reflect the code?

jpechane:
@vjuranek I see there is https://github.com/debezium/debezium/pull/4322
Does it make sense to keep this one open?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
vjuranek: yes, it still applies, but let's remove it as it just describes the code instead of explaining the code, so it's completely useless

vjuranek:
@jpechane well, this was created somehow parallel to #4322 and until #4322 is correct (which I believe it's not  as it was and still is) and more importantly until #4322 has proper commit message explaining the change (*), I'd prefer to merge this PR.

(*) DBZ-5005 as well as the corresponding commit don't explain why the change was done and what was the reasoning behind it, which makes later on (or for people not directly involved in the change) very hard to figure out the reason of this change. In case that the change was wrong (which at least partially seems to be the case of DBZ-5005) is even impossible and result in quite a lot of wasted time (like in this case when I though I'm missing something and tried to find case when `currentScn` is behind `startScn`).

PR #4328, opened 2023-03-02
[docs] Removing redundant TX ids from CloudEvents docs

gunnarmorling:
@jpechane, @Naros, a small fix to the CE docs. The id is there another time on line 86.


jpechane:
@gunnarmorling Good sight :-)

PR #4329, opened 2023-03-02
DBZ-6171 Github workflows not working for Cassandra job (step Build D…

ggaborg:
…ebezium Connector Cassandra)


jpechane:
@ggaborg Applied, thanks

PR #4330, opened 2023-03-03
DBZ-4488: Allow SQL Server connectors to set a maximum number of retriable exception restarts

sugarcrm-jgminder:
This reuses the `errors.max.retries` configuration parameter name from the `EmbeddedEngine` for SQL Server connectors to configure the maximum number of retries.

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerChangeEventSourceCoordinator.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerChangeEventSourceCoordinator.java
@@ -8,6 +8,7 @@ package io.debezium.connector.sqlserver;
 import java.time.Duration;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.kafka.connect.source.SourceConnector;
@@ -41,6 +42,8 @@ public class SqlServerChangeEventSourceCoordinator extends ChangeEventSourceCoor
     private final Clock clock;
     private final Duration pollInterval;
 
+    private AtomicBoolean firstStreamingIterationCompletedSuccessfully;
+
     public SqlServerChangeEventSourceCoordinator(Offsets<SqlServerPartition, SqlServerOffsetContext> previousOffsets, ErrorHandler errorHandler,
                                                  Class<? extends SourceConnector> connectorType,
                                                  CommonConnectorConfig connectorConfig,
@@ -51,8 +54,13 @@ public class SqlServerChangeEventSourceCoordinator extends ChangeEventSourceCoor
                                                  Clock clock) {
         super(previousOffsets, errorHandler, connectorType, connectorConfig, changeEventSourceFactory,
                 changeEventSourceMetricsFactory, eventDispatcher, schema);
+        this.firstStreamingIterationCompletedSuccessfully = new AtomicBoolean(false);
         this.clock = clock;
         this.pollInterval = connectorConfig.getPollInterval();
+    }
+
+    public boolean firstStreamingIterationCompletedSuccessfully() {
+        return firstStreamingIterationCompletedSuccessfully.get();
     }
 
     @Override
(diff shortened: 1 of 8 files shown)


debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerChangeEventSourceCoordinator.java
```diff
@@ -41,6 +41,8 @@ public class SqlServerChangeEventSourceCoordinator extends ChangeEventSourceCoor
     private final Clock clock;
     private final Duration pollInterval;
 
+    private AtomicReference<Boolean> streamingIterationCompleted;
```
jpechane, line 44: Is it possible to use `AtomicBoolean` here?

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -213,6 +223,10 @@ private void updateLastOffset(Map<String, ?> partition, Map<String, ?> lastOffse
      */
     protected abstract List<SourceRecord> doPoll() throws InterruptedException;
 
+    public boolean shouldRestartOnRetriableException() {
```
jpechane, line 226: Would it be possible to move the method to `ErrorHandler`? It would probably belong logically there.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
```diff
@@ -254,6 +256,18 @@ public static SnapshotIsolationMode parse(String value, String defaultValue) {
             .withValidation(Field::isNonNegativeInteger)
             .withDescription("This property can be used to reduce the connector memory usage footprint when changes are streamed from multiple tables per database.");
 
+    public static final Field MAX_RETRIABLE_RESTARTS = Field.create(MAX_RETRIABLE_RESTARTS_CONFIG_NAME)
```
jpechane, line 259: Please check `io.debezium.embedded.EmbeddedEngine.ERRORS_MAX_RETRIES`. We should unify it using the same option or option name.

jpechane (changes requested):
@sugarcrm-jgminder Thanks for the PR. I left couple of comments. Could you please tak a look and refactor the code a bit?

jpechane:
@sugarcrm-jgminder I add a commit to https://github.com/debezium/debezium/pull/4359

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorTask.java
```diff
@@ -97,7 +97,8 @@ public ChangeEventSourceCoordinator<SqlServerPartition, SqlServerOffsetContext>
                 .loggingContextSupplier(() -> taskContext.configureLoggingContext(CONTEXT_NAME))
                 .build();
 
-        errorHandler = new SqlServerErrorHandler(connectorConfig, queue);
+        int retries = (errorHandler == null) ? 0 : errorHandler.getRetries();
+        errorHandler = new SqlServerErrorHandler(connectorConfig, queue, retries, connectorConfig.getMaxRetries());
```
jpechane, line 101: Why si this necessary? Can't you just reuse the exising handler? Then you will not need to pass `retries`.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
sugarcrm-jgminder: @jpechane I would have liked to initially. The reason I didn't is SQLServer connector has the concept of a streaming iteration and that gives us a good place to perform a reset of the retries. I'm not knowledgable enough about the other connectors to determine where a retries reset should occur in their respective logic. Perhaps we could set up a meeting to discuss it in more detail to see what could be done to accomplish this.

sugarcrm-jgminder:
@jpechane The error in PostgreSQL looks like a faulty git checkout.

jpechane:
Merged via https://github.com/debezium/debezium/pull/4440
@sugarcrm-jgminder Could you please ppreare a docs PR with the new config option added? Thanks

PR #4332, opened 2023-03-03
DBZ-6153 Refactor Streams deployment content to fix errors

roldanbob:
[DBZ-6153](https://issues.redhat.com/browse/DBZ-6153)

This change fixes errors in the connector YAML file that's included in the downstream documentation for using Streams to deploy connectors. Earlier, I did a band-aid fix on the published 2022.Q4/DBZ 1.9.7 version of the  downstream docs, but this change fixes the errors for the 2.x stream.

Due to differences among the connectors, the YAML example requires small tweaks depending on the context. The differences are not complex, but because of a shortcoming in the docs toolchain, when fetching content from upstream, `ifeval` conditionals are not supported. So to get the required results, I had to refactor the docs. It's a bit of a hack, but it mirrors the structure currently in place for creating the YAML to deploy Kafka Connect. 

Tested in a local downstream build. 
The change applies only to the downstream version of the docs. There are no effects on the upstream content.

Please backport to 2.1.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/db2.adoc
+++ b/documentation/modules/ROOT/pages/connectors/db2.adoc
@@ -5,10 +5,12 @@
 
 :context: db2
 :data-collection: table
+:database-port: 50000
 :mbean-name: {context}
 :connector-file: {context}
 :connector-class: Db2Connector
 :connector-name: Db2
+:include-list-example: public.inventory
 ifdef::community[]
 
 :toc:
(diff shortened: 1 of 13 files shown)


roldanbob:
@jpechane Is it ok to merge this? The content is _downstream-only_ and is already released as a one-time fix in the product doc. The purpose of this change is to correct the source files so that the update is persistent. As noted, I've tested the change in a downstream build and can confirm that everything plays nicely with the docs toolchain.

jpechane:
@roldanbob Done and backported. Thanks

PR #4333, opened 2023-03-04
DBZ-6179 Fix swallowing of exception cause of finally block in RelationalSnapshotChangeEventSource

subodh1810:
Issue : https://issues.redhat.com/browse/DBZ-6179


subodh1810:
@jpechane please take a look

jpechane:
@subodh1810 LGTM, applied, thanks!

PR #4334, opened 2023-03-06
DBZ-6112 REPLICA IDENTITY setting

jpechane:


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
```diff
@@ -998,7 +998,7 @@ public byte[] getUnavailableValuePlaceholder() {
         return placeholder.getBytes();
     }
 
-    public ReplicaIdentityMapper getTableToReplicaIdentityMap() {
+    public ReplicaIdentityMapper replicaIdentityMapper() {
```
miguelbirdie, line 1001: @jpechane Happy to include this change in our branch 🙂

miguelbirdie:
@jpechane I see that you are trying to remove `withReplicaIdentity`. Do you think that tests are failing because of that?

jpechane:
@miguelbirdie I cahnged the code and hopefully the failures will be fixed. Still I have one more proposal for an imporvement. Right now you use table names. This should be switch to regular expression based mapping in the same way like table include list etc. work. The reason for it is that with regex you could easily set identify for all tables, per schema tables etc.

miguelbirdie:
> @miguelbirdie I cahnged the code and hopefully the failures will be fixed. Still I have one more proposal for an imporvement. Right now you use table names. This should be switch to regular expression based mapping in the same way like table include list etc. work. The reason for it is that with regex you could easily set identify for all tables, per schema tables etc.

Ok, let me know when you fix these failures to rebase my branch with yours. @jpechane regarding regular expression, I think it's a good idea. Is there a function in the code to translate a regular expression to a list of tables?

jpechane:
@miguelbirdie PLease take a look at `io.debezium.function.Predicates` and `io.debezium.util.Strings.setOfRegex(String)` as starting point.

jpechane:
@miguelbirdie You can take it over from here.

PR #4335, opened 2023-03-06
DBZ-6180: debezium-storage-azure-blob

blcksrx:


blcksrx:
@jpechane I have no idea why DB2 fails, could you please give it a look?

blcksrx:
@jpechane could you please check this out? it's like in the `debezium-server-redis` it uses the depricated  `schema.history` and it goes to `socketTimeout`

jpechane:
@blcksrx Applied, thanks!

PR #4336, opened 2023-03-06
DBZ-6186: Add support for GRANT DELETE HISTORY in mariadb

nirolevy:


jpechane:
@nirolevy Applied, thanks

PR #4341, opened 2023-03-06
DBZ-6187 Reading column metadata for each table only once

vivek-chopra:
DBZ-6187 Reading column metadata for each table only once in case of multi schema setup when an inclusion list for tables is configured.

WHAT : Support for multi schema table setups where schemaNamePattern is null for the table inclusion list defined in the properties to narrow down the column retrieval from so many schemas for the same table to rather fetch the column metadata only once for multi schema databases.

WHY : Reduce engine startup -> Ready to listen/stream to a greater extent for DBs with multiple schemas setup. Rather than looping over all schemas for the same table, just to fetch table metadata once.


jpechane:
@vivek-chopra Hi, thanks for the PR. Unfortunately I am really unsure I understand the purpose. What will happen if youhave mutliple schemas with same tables but each table will have a different schema?

vivek-chopra:
> @vivek-chopra Hi, thanks for the PR. Unfortunately I am really unsure I understand the purpose. What will happen if youhave mutliple schemas with same tables but each table will have a different schema?

@jpechane - Thanks for the comment. Yes, in such a case where different schemas could have different schemas for even same tables then I can think of introducing a new property which would be `false` and not trigger this condition unless a multi schema setup exists with same table definitions. What are your thoughts on the same ?

Why we are proposing this is because with `debezium.engine.table.include.list` property being defined, the engine takes a lot of time iterating so many schemas for the same table scheme which can be reduced to a great extent with this change.

jpechane (changes requested):
A generic approach to the problem is necessary.

vivek-chopra:
> @vivek-chopra I understand where you are heading but I am afraid this is not the correct solution. We ae talking about a spcial case here you'd insert into a generic codebase. We need to think and find a different solution that might work for other users and their special needs too.

@jpechane - Yes, I agree. On that front, We were thinking if a new property could be introduced which would mean than this solution is executed only for the specific use cases.

vivek-chopra:
> > @vivek-chopra I understand where you are heading but I am afraid this is not the correct solution. We ae talking about a spcial case here you'd insert into a generic codebase. We need to think and find a different solution that might work for other users and their special needs too.
> 
> @jpechane - Yes, I agree. On that front, We were thinking if a new property could be introduced which would mean than this solution is executed only for the specific use cases.

@jpechane - Please let us know what are your thoughts on this approach. Accordingly, we could try and plan out as this issue is a major blocker for us due to the slowness it introduces to the engine's start up/run time.

jpechane:
@vivek-chopra Unfortunately the property does not solve the generic problem, we'd just hard-code a solutioon to a specific deployment.

Naros:

Hi @vivek-chopra, please sign off all commits with:

```
git commit -s
```

Naros:

PR #4343, opened 2023-03-08
DBZ-6190: Fix support for ddl filter

nirolevy:


nirolevy:
@jpechane @Naros I see the DB2 test fails with the same error in other PRs... Is it known to be flaky? If so, can you please re-run it?

jpechane:
@nirolevy Applied, thanks!

PR #4345, opened 2023-03-08
DBZ-6191 Do not fail if slot cannot be advanced

jpechane:
https://issues.redhat.com/browse/DBZ-6191


jpechane:
Intentionally applied against 2.1 only

yoniiny:
I've run into another issue related to this section of code - calling `SELECT pg_replication_slot_advance('%s', '%s')` is sometimes very slow in some databases, taking up to an hour to complete. Currently we're running the query but only getting a warning out of it anyways. I wonder if it wouldn't be worthwhile to either remove the query entirely or add a timeout so it doesn't slow down replication startup significantly?

Liuxing778:
if slot cannot be advanced，will it build new issues?

PR #4352, opened 2023-03-09
DBZ-5490 Document the case when `message.key.columns` is set and `REPLICA IDENTITY` of the tables is not `FULL`.

PlugaruT:
It breaks tombstone events to have the proper key created because non-primary key columns will be null, and their values in the key will also be null.


jpechane:
@PlugaruT Applied, thanks

PR #4353, opened 2023-03-10
DBZ-6197 Remove duplicated createDdlFilter method from historized connector config

harveyyue:
https://issues.redhat.com/browse/DBZ-6197


jpechane:
@harveyyue Applied, thanks!

PR #4354, opened 2023-03-12
DBZ-5987 Refactor snapshotting to use change streams instead of oplog

jcechace:
https://issues.redhat.com/browse/DBZ-5987

This is just a draft for now -- although tests seem to be passing just fine, I still have some doubts.

Depending on our timeline for 2.2 release I would also like to follow up on this and finished some long overdue clean ups in `ReplicaSetOffsetContext` / `SourceInfo` classes (dates way back when the connector was ported to the common framework).

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbSnapshotChangeEventSource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbSnapshotChangeEventSource.java
@@ -5,12 +5,9 @@
  */
 package io.debezium.connector.mongodb;
 
-import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
-import java.util.LinkedHashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.CountDownLatch;
@@ -20,16 +17,19 @@ import java.util.concurrent.atomic.AtomicInteger;
 import java.util.stream.Collectors;
 
 import org.bson.BsonDocument;
-import org.bson.BsonTimestamp;
 import org.bson.Document;
 import org.bson.conversions.Bson;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.mongodb.MongoChangeStreamException;
 import com.mongodb.ReadPreference;
+import com.mongodb.client.ChangeStreamIterable;
+import com.mongodb.client.MongoChangeStreamCursor;
 import com.mongodb.client.MongoCollection;
 import com.mongodb.client.MongoCursor;
 import com.mongodb.client.MongoDatabase;
+import com.mongodb.client.model.changestream.ChangeStreamDocument;
 
 import io.debezium.DebeziumException;
 import io.debezium.connector.SnapshotRecord;
@@ -66,7 +66,7 @@ public class MongoDbSnapshotChangeEventSource extends AbstractSnapshotChangeEven
     private final ConnectionContext connectionContext;
     private final ReplicaSets replicaSets;
     private final EventDispatcher<MongoDbPartition, CollectionId> dispatcher;
-    protected final Clock clock;
+    private final Clock clock;
     private final SnapshotProgressListener<MongoDbPartition> snapshotProgressListener;
     private final ErrorHandler errorHandler;
     private AtomicBoolean aborted = new AtomicBoolean(false);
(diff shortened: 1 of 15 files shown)


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/SourceInfo.java
```diff
@@ -320,8 +295,8 @@ public void changeStreamEvent(String replicaSetName, ChangeStreamDocument<BsonDo
         long wallTime = 0L;
         if (changeStreamEvent != null) {
             BsonTimestamp ts = changeStreamEvent.getClusterTime();
```
jcechace, line 297: Here we get the timestamp from the change stream event. However based on my observation this timestamp usually has a 1 lower increment compared to the timestamp encoded in the resume token's `_data` field. 

Heart events on the other hand are producing the `ts` field which is extracted directly from the resume token.

Does this pose any problems? Technically the timestamp is not used anywhere anymore. Since the TS is part of the offset I assume we cannot remove it as that would change the  schema.

jcechace:
Right now before the snapshot starts we try to obtain a resume token by invoking 

`cursor.tryNext()`

This returns either a 
1) change stream, from which we extract the resume token
2) null, which however sets the postBastchResumeToken which can be obtained from the cursor itself

Now the official mongodb connector does the same, however in the case of 1 it additionally also caches this change event (it is then processed at the end of copying -- in our connector this would likely have to be at the start of streaming). 

However I wonder if this is really necessary as if this particular change is either insert of update it will be reflected by simply performing the consecutive snapshot. If it is a deleted, then does it really matter? If the connector were started a second later, this event would't have been captured and the snapshot would have ended with exactly the same result. 

Consequently this means that we might be actually OK with just the resume token returned from the cursor -- effectively discarding the change event. 

@jpechane  @btiernay  WDYT?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/SourceInfo.java
jpechane: Timestamp cab be removed form offsets when not needed if the connector is prepared to next expect it.
It should be kept in `source` info block as `ts_ms` as it represents event time of the event.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbSnapshotChangeEventSource.java
```diff
@@ -157,34 +147,30 @@ protected SnapshotResult<MongoDbOffsetContext> doExecute(ChangeEventSourceContex
...
             try (MongoDbConnection mongo = establishConnection(partition, replicaSet, ReadPreference.primaryPreferred())) {
-                final ReplicaSetOffsetContext rsOffsetContext = previousOffset.getReplicaSetOffsetContext(replicaSet);
+                final ReplicaSetOffsetContext rsOffsetContext = offsetContext.getReplicaSetOffsetContext(replicaSet);
                 if (mongo != null && isSnapshotExpected(mongo, rsOffsetContext)) {
                     replicaSetSnapshots.add(replicaSet);
                 }
             }
             catch (InterruptedException e) {
-                throw new DebeziumException("Interrupted while creating snapshotting task", e);
+                throw new RuntimeException(e);
```
jpechane, line 173: Why `RuntimeException`?

jpechane:
@jcechace You are right that this is probably unnecessary. This would make sense if the snapshot would be executed with proper transaction isolation level (like for relation databases) that would ensure there are no duplicates after snapshot/streaming switch. This is not the case of MongoDB so IMHO the event in question can be safely ignored.
IIRC there'd be only one difference in behaviour. If the snapshot is completed then that flag is set in first streaming event. So if we ignore that first streaming event and then there is a significant gap before another one comes there is higher risk of having the snapshot beiing rexecuted after connector is restarted even if snapshot was completed successfully.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbSnapshotChangeEventSource.java
jcechace: Good catch, I wonder how this got in there...  Likely incorrect rebase

jcechace:
> @jcechace You are right that this is probably unnecessary. This would make sense if the snapshot would be executed with proper transaction isolation level (like for relation databases) that would ensure there are no duplicates after snapshot/streaming switch. This is not the case of MongoDB so IMHO the event in question can be safely ignored. IIRC there'd be only one difference in behaviour. If the snapshot is completed then that flag is set in first streaming event. So if we ignore that first streaming event and then there is a significant gap before another one comes there is higher risk of having the snapshot being rexecuted after connector is restarted even if snapshot was completed successfully.

The offset is still saved for that event. And it is questionable whether we should use this one as really "first streaming event" as technically this event happens BEFORE the snapshot is executed (if there actually is an event).

PR #4357, opened 2023-03-13
DBZ-6204 Avoid skipping record when determining resume LSN

vjuranek:
When the last operation before connector stop was tx BEGIN, subsequent INSERT done after connector stop can have same LSN as BEGIN and thus it was possible to skip this INSERT when we search for the resume LSN, This issue was described and fixed in DBZ-5915. However, the same LSN can have also previous tx, so the WAL can look like this:

    LSN | operation
    ------------------
    123 | COMMIT
    123 | BEGIN
    123 | INSERT
    124 | COMMIT

When last stored LSN is 123 for COMMIT operation, we skip also BEGIN and INSERT operations with LSN 123 and resume from COMMIT with LSN 124.

Adjust condition introduced in DBZ-5915 to consider also the case when the last processed operation is COMMIT.

https://issues.redhat.com/browse/DBZ-6204

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
@@ -75,8 +75,9 @@ public class WalPositionLocator {
             // We can resume streaming from it
             if (currentLsn.equals(lastEventStoredLsn)) {
                 // BEGIN and first message after change have the same LSN
-                if (txStartLsn != null && (lastProcessedMessageType == null || lastProcessedMessageType == Operation.BEGIN)) {
-                    // start from the BEGIN tx; prevent skipping of unprocessed event after BEGIN
+                if (txStartLsn != null
+                        && (lastProcessedMessageType == null || lastProcessedMessageType == Operation.BEGIN || lastProcessedMessageType == Operation.COMMIT)) {
+                    // start from the BEGIN tx; prevent skipping of unprocessed event after BEGIN or previsou tx COMMIT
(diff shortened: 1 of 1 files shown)


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
```diff
@@ -75,8 +75,9 @@ public Optional<Lsn> resumeFromLsn(Lsn currentLsn, ReplicationMessage message) {
             // We can resume streaming from it
             if (currentLsn.equals(lastEventStoredLsn)) {
                 // BEGIN and first message after change have the same LSN
-                if (txStartLsn != null && (lastProcessedMessageType == null || lastProcessedMessageType == Operation.BEGIN)) {
-                    // start from the BEGIN tx; prevent skipping of unprocessed event after BEGIN
+                if (txStartLsn != null
+                        && (lastProcessedMessageType == null || lastProcessedMessageType == Operation.BEGIN || lastProcessedMessageType == Operation.COMMIT)) {
+                    // start from the BEGIN tx; prevent skipping of unprocessed event after BEGIN or previsou tx COMMIT
```
jpechane, line 80:
```suggestion
                    // start from the BEGIN tx; prevent skipping of unprocessed event after BEGIN or previous tx COMMIT
```

jpechane:
@vjuranek Could you please check if this is the same for both supported plugins?

vjuranek:
@jpechane yes, I was able to reproduce also with `decodebufs` plugin (which is IMHO expected). Marking it as ready - I'll try to deliver some tests for it in [DBZ-5978](https://issues.redhat.com/browse/DBZ-5978)

mfvitale (approved):

jpechane:
@vjuranek Applied, thanks

PR #4361, opened 2023-03-15
DBZ-5919: Refactor ComputePartition SMT to be more extensible on source info parsing

mfvitale:
Improve ComputePartition SMT source info parsing:

- allow using fields from source info block

- replace switch with the strategy pattern


mfvitale:
Even if lot of changes, core modification are on ComputePartition class.

debezium-core/src/test/java/io/debezium/transforms/SMTTestFramework.java
```diff
@@ -0,0 +1,79 @@
...
+import java.util.HashMap;
+
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.SchemaBuilder;
+import org.apache.kafka.connect.data.Struct;
+import org.apache.kafka.connect.source.SourceRecord;
+
+import io.debezium.data.Envelope;
+
+public class SMTTestFramework {
```
jpechane, line 18: Please use `Smt` not `SMT` we don' keep capitalized chars for acronyms.

debezium-core/src/main/java/io/debezium/transforms/spi/QualifiedTableNameResolver.java
```diff
@@ -0,0 +1,24 @@
...
+package io.debezium.transforms.spi;
+
+import io.debezium.converters.spi.RecordParser;
+
+/**
+ * A {@link java.util.ServiceLoader} interface that connectors should implement to resolve fully qualified table name.
+ *
+ * @author Mario Fiore Vitale
+ */
+public interface QualifiedTableNameResolver extends RecordParserProvider {
```
jpechane, line 15: Why do we need two separate interfaces?

Also please use name `QualifiedDataCollectionNameResolver` to make it datsource independent.

mfvitale: This interface is to explicitly separate the concern of the two interface. One is for parsing the record and the other one define the resolve of a full data collection name.

jpechane: Then should one inherit from another? The class can use multpile interfaces.

mfvitale:
Closing since this is not more necessary due to this changes https://issues.redhat.com/browse/DBZ-5902

PR #4363, opened 2023-03-15
Keep timezone information

bjoernhaeuser:
Instead of converting everything to UTC, which effectively drop the timezone information, continue to use the timezone which is read from the socket.

pgjdbc will set the timezone of the connection to the VMs timezone by default, therefore this has the negative side effect that the output result depends on the VMs timezone.

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
@@ -175,6 +175,39 @@ public class PostgresConnectorConfig extends RelationalDatabaseConnectorConfig {
         }
     }
 
+    public enum TimezoneHandlingMode implements EnumeratedValue {
+        CONVERT_TO_UTC("CONVERT_TO_UTC"),
+
+        /**
+         * Represents interval as ISO 8601 time interval
+         */
+        USE_VM_TIMEZONE("USE_VM_TIMEZONE");
+
+        private final String value;
+
+        TimezoneHandlingMode(String value) {
+            this.value = value;
+        }
+
+        public static TimezoneHandlingMode parse(String value) {
+            if (value == null) {
+                return null;
+            }
+            value = value.trim();
+            for (TimezoneHandlingMode option : TimezoneHandlingMode.values()) {
+                if (option.getValue().equalsIgnoreCase(value)) {
+                    return option;
+                }
+            }
+            return null;
+        }
+
+        @Override
+        public String getValue() {
+            return value;
+        }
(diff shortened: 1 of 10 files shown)


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
```diff
@@ -175,6 +175,39 @@ public static IntervalHandlingMode parse(String value, String defaultValue) {
         }
     }
 
+    public enum TimezoneHandlingMode implements EnumeratedValue {
+        CONVERT_TO_UTC("CONVERT_TO_UTC"),
+
+        /**
+         * Represents interval as ISO 8601 time interval
+         */
+        USE_VM_TIMEZONE("USE_VM_TIMEZONE");
```
gunnarmorling, line 184: The timezone of the VM running Debezium should never impact how temporal types are emitted, as that's typically opaque to consumers of change streams. Instead of baking this into Debezium (and thus further increasing its configuration surface), can't you implement the required conversion on your end, e.g. using an SMT?

bjoernhaeuser: Do you mean something like this? https://debezium.io/documentation/reference/stable/development/converters.html

Yes, it would be possible, I think. But it feels a bit redundant to read the values in the VM timezone (that's already happening), convert it to UTC to then converting it back to the target timezone (e.g. Europe/Berlin).

I understand your arguments why this is not a good idea. Though for me it would not make a difference if the values which is stored in Kafka is UTC or another timezone, which I can control. 


Thank you for your quick response on this!

jpechane: @bjoernhaeuser Hi, I think both of you have have metis in your argument.
@gunnarmorling Was referring to https://cwiki.apache.org/confluence/display/KAFKA/KIP-66%3A+Single+Message+Transforms+for+Kafka+Connect

But your idea using custom converter is even better. Do you think you could implement a custom converter that would become a part of Debezium codebase - for starter only for Postgres. The converter will accept one config parameter  -  either VM TZ will be used or an arbitrary  TZ could be set.

WDYT?

gunnarmorling: Ah, yes, completely forgot about converters. Agreed that this is the preferable solution! No strong opinion on whether it should be packaged with Debezium or not, but in any case it will allow you to solve this problem efficiently on your end, @bjoernhaeuser.

PR #4366, opened 2023-03-16
DBZ-6170 Connect and stream from sharded clusters through mongos instances

jcechace:
https://issues.redhat.com/browse/DBZ-6170

This PR brings support for connecting to sharded clusters as a single connection via monos instance(s). 

In order to remain backwards compatible a new config property `mongodb.connection.mode` is introduced. This property defaults to `replica_set` which keeps the current behaviour when the connector processes each shard as individual replica set. 

When the value is set to `sharded` then the connector will internally create a single instance of `ReplicaSet` with artificial name `cluster` which results in single connector where all connections are processed via the initially supplied connection string. 

Testing is not done yet -- I will have to introduce the very first tests for sharded cluster which is unlikely to work in GH actions. Additionally when this feature is enabled then incremental snapshotting should work also for sharded clusters and thus the incremental snapshot tests will have to be adapted as well.

### DIFF
--- a/debezium-connector-mongodb/pom.xml
+++ b/debezium-connector-mongodb/pom.xml
@@ -114,8 +114,12 @@
         <apicurio.port>8080</apicurio.port>
         <apicurio.init.timeout>60000</apicurio.init.timeout> <!-- 60 seconds -->
         <!-- Mongo test containers -->
-        <mongodb.docker.desktop.ports>27017,27018,27019</mongodb.docker.desktop.ports>
+        <!-- port range used to allocate ports -->
+        <mongodb.docker.desktop.ports>27017:27117</mongodb.docker.desktop.ports>
         <mongodb.replica.size>1</mongodb.replica.size>
+        <mongodb.shard.size>2</mongodb.shard.size>
+        <mongodb.shard.replica.size>${mongodb.replica.size}</mongodb.shard.replica.size>
+
     </properties>
     <build>
         <plugins>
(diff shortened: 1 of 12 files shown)


jcechace:
@jpechane The failure on MongoDB 4.0 is consistent. It looks like something we do with the change stream pipeline fails on 4.0 when an update event is encountered. 

Strangely enough this happens only with the original `replica_set` connection mode. With the newly introduced `sharded` mode it looks everything works. 

Since we are deprecating the support for 4.0 in 2.2 I wouldn't worry about this. However I'd like to propose porting this test to 2.1.

jcechace:
So this is definitively caused by the use of aggregation pipeline on change streams introduced for DBZ-6131. Still considering everything seems to work with 4.2+ I would propose ignoring this error.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/connection/ConnectionMode.java
```diff
@@ -0,0 +1,72 @@
...
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.mongodb.connection;
+
+import io.debezium.config.EnumeratedValue;
+
+/**
+ * The set of predefined MongoDbConnectionMode options or aliases.
+ */
+public enum ConnectionMode implements EnumeratedValue {
```
jpechane, line 13: Why is not this enum inner in MongoDbConnectorConfig as the others? Is there anything special about it?

jcechace:
@jpechane should be completely ready for merge now.

PR #4367, opened 2023-03-16
Dbz 823 conditionalize incubation note as tp for downstream

roldanbob:
[DBZ-823](https://issues.redhat.com/browse/DBZ-823)

This change updates the description for `snapshot.max.threads` that was modified in DBZ-823, by conditionalizing the existing note about parallel snapshots being in an incubating state so that it renders only in the community version of the docs. 
It also adds an equivalent Tech Preview statement for the product version.

At present, in the MongoDB documentation, the property description does not include any caveats. If that is ok, then this can be merged as-is. Let me know if I need to add a similar note for the MongoDB connector.


jpechane:
@roldanbob Applied, thanks

PR #4369, opened 2023-03-18
DBZ-6223 Add metadata; additional edits

roldanbob:
[DBZ-6223 ](https://issues.redhat.com/browse/DBZ-6223)Adds anchor id and downstream metadata, and edits to change streams content.


jpechane:
@roldanbob Applied, thanks

PR #4371, opened 2023-03-18
DBZ-6225 Treat BOOLEAN as TINYINT(1) consistently

Naros:
https://issues.redhat.com/browse/DBZ-6225

The snapshot phase uses the JDBC driver metadata to resolve the column mappings; however, during streaming, if a CREATE TABLE ddl statement is used, the column mapping is based on that of the DDL, which was not treating a BOOLEAN column as TINYINT(1); instead, it treated the column as BOOLEAN, which is incorrect for MySQL.


Naros:
About the test changes let's take the test method `paseBooleanDefaultValue`.  If we create such a table using the following SQL and then check the contents of the table we'll find that the column `a` contains `9` rather than `true`.
```
CREATE TABLE test (a boolean not null default '9');
INSERT INTO test values ();
```
The test changes effectively mirror what we expect in the table and mirror what we ultimately would be emitted as part of the change event since BOOLEAN is technically a synonym for TINYINT(1), which isn't necessarily 0/1 or false/true.

Naros:
Superseded by referenced PR.

PR #4372, opened 2023-03-18
DBZ-6226 Treat REAL as DOUBLE consistently

Naros:
https://issues.redhat.com/browse/DBZ-6226

The snapshot phase uses the JDBC driver metadata to resolve the column mappings; however during streaming if a CREATE TABLE ddl statement is used, the column mapping is based on that of the DDL, which was not treating a REAL column as DOUBLE; instead it treated the column as FLOAT32 data type, which is technically incorrect.


jpechane:
@Naros Unfortunately there is also https://dev.mysql.com/doc/refman/8.0/en/sql-mode.html#sqlmode_real_as_float
We need to decide how to incorporate it correctly.

PR #4373, opened 2023-03-19
DBZ-6013: Data will miss when use archive log only mode with Oracle RAC

LangYuanzh:
[DBZ-6013](https://issues.redhat.com/browse/DBZ-6013)

Data will miss when use archive log only mode with Oracle RAC

The cause of the problem is that there is an overlap between the SCN range contained in the redo logs of the two threads。
 
We should check the SCN range of online redo log and the current  startScn  and endScn of the connector.

  1). If the startScn is contained both  in a archived log and online redo log, we should pause the task and wait untill the online redo log is archived. 

  2). We need limit endScn to no more than the smallest scn in all online redo logs when calculate endScn.


Naros:
Hi @LangYuanzh, could you please run the following command locally and commit the format changes:
```
./mvnw clean install -DskipITs -DskipTests -Passembly,oracle-all,oracle-tests
```

PR #4374, opened 2023-03-19
DBZ-6231 Treat NCHAR/NVARCHAR consistently

Naros:

PR #4376, opened 2023-03-20
DBZ-6225 DBZ-6226 DBZ-6231 Introduce CustomConverter for JDBC sink consistency

Naros:
https://issues.redhat.com/browse/DBZ-6225
https://issues.redhat.com/browse/DBZ-6226
https://issues.redhat.com/browse/DBZ-6231

Supersedes previous PRs:
#4371 
#4372 
#4374 

Regarding the MySQL treatment of REAL as FLOAT vs. DOUBLE, I've incorporated that as part of a configurable option for the custom converter.  By default, it uses DOUBLE to be aligned with MySQL, but it gives users a way to influence this via the converter if they need that behavior.

- [ ] Add documentation


mfvitale (approved):

jpechane:
@Naros Applied, thanks. Please the docs in a separate PR. Also I applied to 2.2 only as I don't think it is necessary for 2.1.

Naros:
Created https://issues.redhat.com/browse/DBZ-6242 as a follow-up.

PR #4378, opened 2023-03-20
DBZ-2979 -  Feat: Add Config skip.messages.without.change to skip updates where there is no change in whitelisted columns

ironakj:
If config is enabled and If `column.include.list/column.exclude.list` is used and the target table receives an update for the excluded (or not included) column - such events should be ignored

Addresses/Closes DBZ-2979
https://issues.redhat.com/browse/DBZ-2979

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlChangeRecordEmitter.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlChangeRecordEmitter.java
@@ -27,14 +27,16 @@ public class MySqlChangeRecordEmitter extends RelationalChangeRecordEmitter<MySq
     private final OffsetContext offset;
     private final Object[] before;
     private final Object[] after;
+    private final boolean skipMessagesWithoutChange;
 
-    public MySqlChangeRecordEmitter(MySqlPartition partition, OffsetContext offset, Clock clock, Envelope.Operation operation, Serializable[] before,
-                                    Serializable[] after) {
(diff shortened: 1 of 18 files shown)


ironakj:
@jpechane I've made changes in Mysql and SqlServer as well, but I am blocked at writing test cases for SQL server as the test container is not booting up, I'll try once again when I am in a position to. 

Please find the conversation here if you can help 
https://debezium.zulipchat.com/#narrow/stream/302533-dev/topic/DBZ-2979

Naros:
Hi @ironakj, one thing I noticed just skimming the tests is that it appears we're only testing one scenario, and we don't have a test where the update is changing both excluded and non-excluded columns.  Let's ensure the test exists for all connector implementations because we should ensure that the event gets emitted.

jpechane:
@ironakj I don't want to delay the PR even longer so I am going to merge it when CI run completes. Could you please create a follow-up Jira and prepare a PR that will modify the code a bit?

- `CommonConnectorConfig` will get passed to `AbstractChangeRecordEmitter`
- `RelationalDatabaseConnectorConfig` will get passed to `RelationalDatabaseChangeRecordEmitter`
- `skipMessagesWithoutChange` method could be pulled-up to `AbstractChangeRecordEmitter`
- `connectorConfig.skipMessagesWithoutChange()` will not be passed to change record emitters

As a result

* The only changes to record emitters (unfortunately now to all) will be passing the connector config to parent class
* It will enable seemless improvements for future changes/configs that will be common for all emitters
* The feature can be enable for MongoDB and other connectors as well

jpechane:
@indraraj Applied, thanks!

PR #4380, opened 2023-03-21
DBZ-6013: Data will miss when use archive only mode with Oracle RAC

LangYuanzh:
[DBZ-6013](https://issues.redhat.com/browse/DBZ-6013?attachmentOrder=asc)

Data will miss when use archive log only mode with Oracle RAC

The cause of the problem is that there is an overlap between the SCN range contained in the redo logs of the two threads。

We should check the SCN range of online redo log and the current startScn and endScn of the connector.

1). If the startScn is contained both in a archived log and online redo log, we should pause the task and wait untill the online redo log is archived.

2). We need limit endScn to no more than the smallest scn in all online redo logs when calculate endScn.


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -418,6 +424,23 @@ private Scn getMaxArchiveLogScn(List<LogFile> logFiles) {
             }
         }
 
+        if (currentRedoLogFiles.isEmpty()) {
+            throw new DebeziumException("Cannot get current redo log SCN as no redo logs are present.");
```
vjuranek, line 428: AFAICT `currentRedoLogFiles` is initialized only when archive mode is on and thus `isStartScnInArchiveLogs()` is called, so if we don't use archive mode, we would fail here?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -897,9 +917,31 @@ private boolean waitForStartScnInArchiveLogs(ChangeEventSourceContext context, S
...
+
+        if (Objects.isNull(currentRedoLogFiles)) {
+            currentRedoLogFiles = new ArrayList<>();
+        }
+        currentRedoLogFiles.clear();
+        for (LogFile file : logs) {
+
+            if (file.getType().equals(LogFile.Type.ARCHIVE)
+                    && file.getFirstScn().compareTo(startScn) <= 0 && file.getNextScn().compareTo(startScn) > 0) {
+                LOGGER.info("startScn {} in archive log {} with range {},{}", startScn, file.getFileName(), file.getFirstScn(), file.getNextScn());
```
vjuranek, line 932: probably should be debug level?

vjuranek (changes requested):
Thanks for the PR and sorry for the delay with review!

Over LGTM, but I have couple of questions regarding the changes.

Also please add your name into `COPYRIGHT.txt` and `jenkins-jobs/scripts/config/Aliases.txt` in order to Contributor check would pass. Thanks!

vjuranek:
> Also please add your name into `COPYRIGHT.txt` and `jenkins-jobs/scripts/config/Aliases.txt` in order to Contributor check would pass. Thanks!

Sorry, as mentioned in github-action comments, this should be done by reviewer, so I already did that (and also rebased).

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -360,7 +361,9 @@ private void initializeRedoLogsForMining(OracleConnection connection, boolean po
             if (!isContinuousMining) {
                 currentLogFiles = setLogFilesForMining(connection, startScn, archiveLogRetention, archiveLogOnlyMode,
                         archiveDestinationName, logFileQueryMaxRetries, initialDelay, maxDelay);
-                currentRedoLogSequences = getCurrentLogFileSequences(currentLogFiles);
+                if (!archiveLogOnlyMode) {
```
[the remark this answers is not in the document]
Naros: To add to what Voljtech said, the log switch calculation is essential even for archive log mode because we need to know when a new archive log has been generated. We only know this when we detect log switches, so this step is critical regardless of the mode used.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -370,7 +373,9 @@ private void initializeRedoLogsForMining(OracleConnection connection, boolean po
                 }
                 currentLogFiles = setLogFilesForMining(connection, startScn, archiveLogRetention, archiveLogOnlyMode,
                         archiveDestinationName, logFileQueryMaxRetries, initialDelay, maxDelay);
-                currentRedoLogSequences = getCurrentLogFileSequences(currentLogFiles);
+                if (!archiveLogOnlyMode) {
+                    currentRedoLogSequences = getCurrentLogFileSequences(currentLogFiles);
+                }
```
Naros, line 378: Same as above, please remove the if-condition as its unnecessary.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
Naros: I'm really concerned about the use of `currentRedoLogFiles`, even beyond Vojtech's comments above.  This is a collection we calculate as a one-off in `isStartScnInArchiveLogs`, and its safe to use those because we're trying to determine where the SCN in question is in the archive and redo logs.

However, calculating the max SCN should be based solely on the logs that are part of the mining session, which is calculated elsewhere when we populated the `currentLogFiles` collection, and I think ultimately that should be the method call that we base this explicit calculation upon so that we don't run the risk of any type of race condition with data sets.

Looking at the code more closely, I'm inclined to say that the call to `setLogFilesForMining` when archive log mode is enabled should return all the current logs (archive and redo). This will allow the collection of the current redo log sequences to be calculated properly. Then you could use that full collection here in this calculation without relying on the `currentRedoLogFiles` collection that is calculated elsewhere based on a one-off query that could be based on a completely different state.

If you two concur, then I would suggest the following changes:

1. Introduce a new class called `LogFiles` or similar, and it acts as a parameter object, and it will expose two critical pieces of information (a) all log files for a given SCN, (b) all log files that are to be mined.
2. `LogMinerHelper#getLogFilesForOffsetScn` would return this `LogFiles` object, populated accordingly.
3. `LogMinerHelper#setLogFilesForMining` will return this `LogFiles` object.

This will then allow us inside this method to have access to both ALL and Mined log files, and we can use the ALL collection here in this method so that the list of logs is consistent with those that were added to the session without there being any risk of de-synced data sets.

Naros (changes requested):
I left a few comments inline.

Naros:
Hi @LangYuanzh, is there any planned progress on this?

Naros:

Hi @LangYuanzh, please sign off all commits with:

```
git commit -s
```

Naros:
Superseded by #7170

PR #4381, opened 2023-03-21
DBZ-6239 Use pull_request instead of pull_request_target

ani-sha:
https://issues.redhat.com/browse/DBZ-6239


vjuranek (approved):
Good catch @ani-sha , LGTM.

jpechane:
@ani-sha Applied, thanks

PR #4385, opened 2023-03-21
DBZ-6240 Change stream supports filtering schema change event

sunxiaojian:
Change stream supports filtering schema change event

### DIFF
--- a/debezium-core/src/main/java/io/debezium/transforms/SchemaChangeEventFilter.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/SchemaChangeEventFilter.java
@@ -33,8 +33,8 @@ import io.debezium.schema.SchemaChangeEvent;
 public class SchemaChangeEventFilter<R extends ConnectRecord<R>> implements Transformation<R> {
     private static final Logger LOGGER = LoggerFactory.getLogger(SchemaChangeEventFilter.class);
 
-    private static final Field SCHEMA_CHANGE_EVENT_INCLUDE_LIST = Field.create("schema.change.event.include.list")
-            .withDisplayName("Schema change event include list")
+    private static final Field SCHEMA_CHANGE_EVENT_EXCLUDE_LIST = Field.create("schema.change.event.exclude.list")
+            .withDisplayName("Schema change event exclude list")
             .withType(ConfigDef.Type.STRING)
             .withWidth(ConfigDef.Width.SHORT)
             .withImportance(ConfigDef.Importance.LOW)
@@ -49,8 +49,8 @@ public class SchemaChangeEventFilter<R extends ConnectRecord<R>> implements Tran
     public void configure(Map<String, ?> configs) {
         final Configuration config = Configuration.from(configs);
         smtManager = new SmtManager<>(config);
-        smtManager.validate(config, Field.setOf(SCHEMA_CHANGE_EVENT_INCLUDE_LIST));
-        final String includeSchemaChangeEvents = config.getString(SCHEMA_CHANGE_EVENT_INCLUDE_LIST);
+        smtManager.validate(config, Field.setOf(SCHEMA_CHANGE_EVENT_EXCLUDE_LIST));
+        final String includeSchemaChangeEvents = config.getString(SCHEMA_CHANGE_EVENT_EXCLUDE_LIST);
         this.includeSchemaChangeEvents = Arrays.stream(includeSchemaChangeEvents.split(",")).map(typeName -> SchemaChangeEvent.SchemaChangeEventType.valueOf(typeName))
                 .collect(Collectors.toSet());
 
@@ -85,7 +85,7 @@ public class SchemaChangeEventFilter<R extends ConnectRecord<R>> implements Tran
     @Override
     public ConfigDef config() {
         final ConfigDef config = new ConfigDef();
-        Field.group(config, null, SCHEMA_CHANGE_EVENT_INCLUDE_LIST);
+        Field.group(config, null, SCHEMA_CHANGE_EVENT_EXCLUDE_LIST);
         return config;
     }
 
--- a/debezium-core/src/main/java/io/debezium/transforms/SchemaChangeEventFilter.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/SchemaChangeEventFilter.java
@@ -42,7 +42,7 @@ public class SchemaChangeEventFilter<R extends ConnectRecord<R>> implements Tran
                     "Support filtering during DDL synchronization")
             .required();
 
-    private Set<SchemaChangeEvent.SchemaChangeEventType> includeSchemaChangeEvents;
+    private Set<SchemaChangeEvent.SchemaChangeEventType> excludeSchemaChangeEvents;
     private SmtManager<R> smtManager;
 
     @Override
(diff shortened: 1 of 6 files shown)


jpechane:
@sunxiaojian Hi, thanks for the PR. I am not fully convinced we want to add such functionality directly to the Debezium internals. Is this something that could be achieve via an SMT? That would be used when needed and be configured to filter out requested DDLs. At the same time it would be applicble at any connector.

sunxiaojian:
> @sunxiaojian Hi, thanks for the PR. I am not fully convinced we want to add such functionality directly to the Debezium internals. Is this something that could be achieve via an SMT? That would be used when needed and be configured to filter out requested DDLs. At the same time it would be applicble at any connector.

It's ok add an SMT to filter, but I think it should be used with the include.schema.changes configuration

sunxiaojian:
> We tend to avoid too many config options if possible. Generally only if a performance gain can be obtained which is not the case of DDL filtering.

Okay, I'll add an SMT to filter

Naros (commented):
Overall this LGTM @sunxiaojian, there is just one outstanding task: adding a reference to this new transformation in the documentation.

sunxiaojian:
> Overall this LGTM @sunxiaojian, there is just one outstanding task: adding a reference to this new transformation in the documentation.

Okay, let me add it

debezium-core/src/main/java/io/debezium/transforms/SchemaChangeEventFilter.java
```diff
@@ -0,0 +1,102 @@
...
+public class SchemaChangeEventFilter<R extends ConnectRecord<R>> implements Transformation<R> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(SchemaChangeEventFilter.class);
+
+    private static final Field SCHEMA_CHANGE_EVENT_INCLUDE_LIST = Field.create("schema.change.event.include.list")
+            .withDisplayName("Schema change event include list")
+            .withType(ConfigDef.Type.LIST)
+            .withWidth(ConfigDef.Width.SHORT)
+            .withImportance(ConfigDef.Importance.LOW)
+            .withDescription(
+                    "Support filtering during DDL synchronization!");
```
jpechane, line 43:
```suggestion
                    "Support filtering during DDL synchronization");
```

debezium-core/src/main/java/io/debezium/transforms/SchemaChangeEventFilter.java
```diff
@@ -0,0 +1,102 @@
...
+    private Set<SchemaChangeEvent.SchemaChangeEventType> includeSchemaChangeEvents;
+    private SmtManager<R> smtManager;
+
+    @Override
+    public void configure(Map<String, ?> configs) {
+        final Configuration config = Configuration.from(configs);
+        smtManager = new SmtManager<>(config);
+
+        final String includeSchemaChangeEvents = config.getString(SCHEMA_CHANGE_EVENT_INCLUDE_LIST);
+        if (includeSchemaChangeEvents == null) {
```
jpechane, line 54: Please take a look at `io.debezium.transforms.HeaderToValue.HEADERS_FIELD` for better way how to handle field validation.

debezium-core/src/main/java/io/debezium/transforms/SchemaChangeEventFilter.java
```diff
@@ -0,0 +1,102 @@
...
+            // throw exception
+            throw new IllegalArgumentException("The configuration [schema.change.event.include.list] cannot be empty");
+        }
+        this.includeSchemaChangeEvents = (includeSchemaChangeEvents == null)
```
jpechane, line 58: IIUC there lines above ensure that `includeSchemaChangeEvents` will be alway non-null.

jpechane (changes requested):
@sunxiaojian Thanks for that. I left two more minor comments to address.

sunxiaojian:
@jpechane @Naros PTAL

jpechane (changes requested):
@sunxiaojian Overall, the code LGTM!
I did few changes in docs. Alos I've one request for you - the name `schema.change.event.include.list` is very confusing as it implies the list of DDL types that should be kept in the stream but IIUC they are in fact removed. So I propose renaming it to either `schema.change.event.exclude.list` or `schema.change.event.filter.list`.

@roldanbob Could you please review the docs as well? I am unsure about the formatting a bit.

sunxiaojian:
> @sunxiaojian Overall, the code LGTM! I did few changes in docs. Alos I've one request for you - the name `schema.change.event.include.list` is very confusing as it implies the list of DDL types that should be kept in the stream but IIUC they are in fact removed. So I propose renaming it to either `schema.change.event.exclude.list` or `schema.change.event.filter.list`.
> 
> @roldanbob Could you please review the docs as well? I am unsure about the formatting a bit.

@jpechane   Okay , I have modified it to 'schema.change.event.exclude.list'

roldanbob:
> @roldanbob Could you please review the docs as well? I am unsure about the formatting a bit.

@jpechane Should I approach this as if it's going to be exposed downstream (that is, add the necessary Nebel annotations)?

roldanbob (changes requested):
@sunxiaojian  Thank you for providing this new documentation. I've suggested some language, style, and formatting changes. 
I tested the formatting changes in a local version of the file, but to guard against the possibility that I missed anything, I'd like to review the changes again after you make updates. 
I've also posted a couple of questions related to fields in the properties table.

Let me know if you have any concerns or questions about any edits.

sunxiaojian:
> @sunxiaojian Thank you for providing this new documentation. I've suggested some language, style, and formatting changes. I tested the formatting changes in a local version of the file, but to guard against the possibility that I missed anything, I'd like to review the changes again after you make updates. I've also posted a couple of questions related to fields in the properties table.
> 
> Let me know if you have any concerns or questions about any edits.

@roldanbob done

documentation/modules/ROOT/pages/transformations/schema-change-event-filter.adoc
```diff
@@ -0,0 +1,150 @@
...
+== Effect of applying the `SchemaChangeEventFilter` SMT
+After you configure the SMT for a connector, whenever the connector captures `DROP` or `TRUNCATE` events, it resets the values for these events to `null`.
+The connector does not send events with the resulting `null` values to Kafka.
+The following examples show how the transformation changes the values of event records that you designate for exclusion.
+.Schema change event record before it is processed by the `SchemaChangeEventFilter` transformation
+====
+Schema change event record before it is processed by the `SchemaChangeEventFilter` transformation::
+
+Schema change event record before the SMT processes the event record:::
++
```
roldanbob, line 44:
```suggestion
The connector does not send events with the resulting `null` values to Kafka.

The following examples shows a xref:schema-change-event-filter-example-before-processing[schema change event record before it is processed] by the SMT, followed by the xref:schema-change-event-filter-example-after-processing[output after the SMT filters the event].

[[schema-change-event-filter-example-before-processing]]
.Schema change event record before it is processed by the `SchemaChangeEventFilter` transformation
====
```

documentation/modules/ROOT/pages/transformations/schema-change-event-filter.adoc
```diff
@@ -0,0 +1,150 @@
...
+.Schema change event record after processing by the SMT
+====
+
+[source]
+----
+null
+----
+====
+
+
```
roldanbob, line 122:
```suggestion

// Type: reference
```

documentation/modules/ROOT/pages/transformations/schema-change-event-filter.adoc
```diff
@@ -0,0 +1,150 @@
...
+                "type":"DROP",
+                "id":"test.table"
+            }
+        ]
+    }
+}
+----
+
+====
+.Schema change event record after processing by the SMT
```
roldanbob, line 113:
```suggestion

Based on the earlier xref:example-configuration-schema-change-event-filter[configuration example], after the SMT processes the xref:schema-change-event-filter-example-before-processing[original schema change event], it resets the values for the `DROP` event to `null`, as shown in the following example:

[[schema-change-event-filter-example-after-processing]]
.Schema change event record after processing by the SMT
```

roldanbob (changes requested):
Added some internal cross-references, removed obsolete formatting, and inserted downstream metadata comments.

jpechane:
@sunxiaojian Applied, thanks!

PR #4386, opened 2023-03-22
DBZ-6235 Remove incubation notice from Debezium Server page

harveyyue:
https://issues.redhat.com/browse/DBZ-6235


jpechane:
@harveyyue Applied, thanks!

PR #4390, opened 2023-03-22
DBZ-6243: Fix MySql parser to permit alias in single delete statements

mfvitale:
https://issues.redhat.com/browse/DBZ-6243


jpechane:
@mfvitale Applied, thanks! Could you please prepare an [upstream](https://github.com/antlr/grammars-v4) PR and link it here?

Could you please double check if similar issue cannot affect other DMLs?

Also for the next time please don't use colon after DBZ-xxx: in the commit message. It is not strictly forbidden but we usually avoid it.

mfvitale:
Opened PR on upstream https://github.com/antlr/grammars-v4/pull/3297

> Could you please double check if similar issue cannot affect other DMLs?

Seems that this was the only part affected

PR #4395, opened 2023-03-23
DBZ-6249 Fix error thrown when snapshot.custom_class=custom and no snapshot.custom.class

nirolevy:


jpechane:
@nirolevy Applied, thanks!

PR #4396, opened 2023-03-23
DBZ-6248 Expose EmbeddedEngine configuration fields

nirolevy:


jpechane:
@nirolevy Applied, thanks!

PR #4399, opened 2023-03-24
DBZ-6251 Ignore Debezium internal transactions during fallback to restart_lsn in Postgres connector if confirmed_flush_lsn is NULL

AnatolyPopov:
A follow-up for https://github.com/debezium/debezium/pull/4265

When  Debezium tries to fetch a replication slot with null confirmed_flush_lsn from Postgres it checks for any long-running transactions before fallback to restart_lsn and if there are any it needs to wait for them to be finished. However, during start-up Debezium connector creates this kind of transaction itself and it needs to be ignored during this check.


jpechane:
@AnatolyPopov Applied, thanks

PR #4413, opened 2023-03-30
DBZ-6265 skip the wal level check if no cdc is to occur

alwaysbemark:
https://issues.redhat.com/browse/DBZ-6265 
Refactored the Postgres connector class to extract the individual connection check methods, changed the WAL_LEVEL test to only run if the snapshot mode involves streaming.

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnector.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnector.java
@@ -141,7 +141,7 @@ public class PostgresConnector extends RelationalBaseSourceConnector {
                     "SHOW wal_level",
                     connection.singleResultMapper(rs -> rs.getString("wal_level"), "Could not fetch wal_level"));
             if (!"logical".equals(walLevel)) {
-                throw new SQLException("Postgres server wal_level property must be \\\"logical\\\" but is: \" + walLevel");
+                throw new SQLException("Postgres server wal_level property must be 'logical' but is: '" + walLevel + "'");
             }
         }
         else {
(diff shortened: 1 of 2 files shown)


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnector.java
```diff
@@ -142,6 +100,63 @@ protected void validateConnection(Map<String, ConfigValue> configValues, Configu
         }
     }
 
+    private static void checkLoginReplicationRoles(PostgresConnection connection) throws SQLException {
```
alwaysbemark, line 103: All methods extracted as is from the large validation method above.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnector.java
```diff
@@ -142,6 +100,63 @@ protected void validateConnection(Map<String, ConfigValue> configValues, Configu
...
+                                || rs.getBoolean("aws_replication")),
+                        "Could not fetch roles"))) {
+            final String errorMessage = "Postgres roles LOGIN and REPLICATION are not assigned to user: " + connection.username();
+            LOGGER.error(errorMessage);
+        }
+    }
+
+    private static void checkWalLevel(ConfigValue hostnameValue, PostgresConnection connection, PostgresConnectorConfig config) throws SQLException {
+        // Logical WAL_LEVEL is only necessary for CDC snapshotting
+        if (config.getSnapshotter().shouldStream()) {
```
alwaysbemark, line 139: This is the only business logic change - this check will skip for `initial_only` as it's not needed in this case.

jpechane:
Let's discuss at https://issues.redhat.com/browse/DBZ-6271

alwaysbemark:
> Let's discuss at https://issues.redhat.com/browse/DBZ-6271

Thanks for reviewing! This change should be safe as only checking the WAL level is skipped?

jpechane:
@alwaysbemark Let's use the same approach as with MySQL PR. Throw error by default, log `WARN` for schema only.

jpechane:
@alwaysbemark Applied, thanks

PR #4418, opened 2023-03-31
DBZ-6271 allow binlog to be absent when doing a snapshot only

alwaysbemark:
https://issues.redhat.com/browse/DBZ-6271
If we're in the initial_only mode of the snapshotter and don't require streaming CDC, we don't need the bin log position to be retrieved and can safely skip it and start the snapshot from scratch.

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
@@ -270,7 +270,17 @@ public class MySqlSnapshotChangeEventSource extends RelationalSnapshotChangeEven
         }
         final MySqlOffsetContext offsetContext = MySqlOffsetContext.initial(connectorConfig);
         ctx.offset = offsetContext;
-        LOGGER.info("Read binlog position of MySQL primary server");
+        if (connectorConfig.getSnapshotMode().shouldStream()) {
+            LOGGER.info("Read binlog position of MySQL primary server");
+            setBinlogPosition(offsetContext);
+        }
+        else {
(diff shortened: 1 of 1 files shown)


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
```diff
@@ -288,6 +288,9 @@ protected void determineSnapshotOffset(RelationalSnapshotContext<MySqlPartition,
                     LOGGER.info("\t using binlog '{}' at position '{}'", binlogFilename, binlogPosition);
                 }
             }
+            else if (!connectorConfig.getSnapshotMode().shouldStream()) {
```
mfvitale, line 291: What do you think instead of logging here adding a condition to line 263 and avoid doing the query at all?

alwaysbemark: @mfvitale Agree - done!

jpechane:
Let's discuss at https://issues.redhat.com/browse/DBZ-6271

alwaysbemark:
> Let's discuss at https://issues.redhat.com/browse/DBZ-6271

Thanks for having a look Jiri! I reverted to the original state of the PR which allows for quiet failure of fetching the bin log position.

jpechane:
Let's not use theat quiet :-). Please use {{WARN}} level together with stactrace.

alwaysbemark:
> Let's not use theat quiet :-). Please use {{WARN}} level together with stactrace.

Made the required change - changed to using WARN and displaying a stacktrace as well. Thanks for reviewing promptly!

jpechane:
@alwaysbemark Applied, thanks

PR #4425, opened 2023-04-03
DBZ-6276 remove the archive log check for Oracle

alwaysbemark:
https://issues.redhat.com/browse/DBZ-6276

Skip the archive log check in case we're in INITIAL_ONLY snapshot mode, which doesn't require the redo log.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
@@ -61,7 +61,7 @@ public class OracleConnectorTask extends BaseSourceTask<OraclePartition, OracleO
                 () -> new OracleConnection(jdbcConfig));
         jdbcConnection = connectionFactory.mainConnection();
 
-        validateRedoLogConfiguration();
+        validateRedoLogConfiguration(connectorConfig);
 
         OracleValueConverters valueConverters = new OracleValueConverters(connectorConfig, jdbcConnection);
         OracleDefaultValueConverter defaultValueConverter = new OracleDefaultValueConverter(valueConverters, jdbcConnection);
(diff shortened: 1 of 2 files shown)


Naros (changes requested):
Hi @alwaysbemark, I don't think this will be sufficient for Oracle.  Please look at the configuration option `log.mining.transaction.snapshot.boundary.mode`.  If the option is configured with `all` we also consult the redo logs.

So if the user uses `initial_only` + snapshot boundary of `all`, the validate method should still be called since the snapshot offset will be based on some redo log analysis.

jpechane:
Also I'd chime with the opinion it should be the same as for MySQL and PostgreSQL PRs. The work should be done, just in case of failures there will be only warning in case of snapshot only.

alwaysbemark:
Thanks for the recommendations - I'll work on the Oracle change this week, it would be great if the Postgres and MySQL ones could be merged separately. Thank you for the prompt reviews!

Naros:
Hi @alwaysbemark to be clear, the MySQL and PostgreSQL PRs should also be adjusted per Jiri's recommendation where the checks are performed regardless of the `snapshot.mode` set, the difference is how the outcome of the check is handled.  In the case of `initial_only`, we only write a warning message to the logs.  In other cases, we keep the current behavior to fault the connector.

alwaysbemark:
> Hi @alwaysbemark to be clear, the MySQL and PostgreSQL PRs should also be adjusted per Jiri's recommendation where the checks are performed regardless of the `snapshot.mode` set, the difference is how the outcome of the check is handled. In the case of `initial_only`, we only write a warning message to the logs. In other cases, we keep the current behavior to fault the connector.

Thanks for the heads up - the MySQL change implements that already, I'll adjust the Postgres implementation alongside the Oracle change to do just that.

alwaysbemark:
Heya @jpechane @Naros , I've made the required changes. The UI backend build seems to be flaky or fail due to an unrelated issue. Would love a second look at this, thank you!

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
```diff
@@ -171,14 +171,28 @@ protected Iterable<Field> getAllConfigurationFields() {
...
+                        + "required for this connector to work properly. Change the Oracle configuration to use a "
+                        + "LOG_MODE=ARCHIVELOG and restart the connector.");
+            }
+            else {
+                LOGGER.warn("Failed the archive log check but continuing as redo log isn't strictly required");
+            }
         }
+        else {
+            LOGGER.info("Successfully validated the archive LOG_MODE");
+        }
```
Naros, line 189: I think if archive log mode is enabled, there is no need to log anything.

Naros (changes requested):

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
alwaysbemark: Fixed! Thank you.

Naros (approved):
Hi @alwaysbemark and @jpechane, I made a slight change to the test for this specific issue because our default Oracle images are preconfigured with archive log mode enabled, and this isn't something we can just turn off in the test suite proper, so I disabled the specified test in the test suite execution.  

This LGTM and its safe to merge.

jpechane:
@alwaysbemark Applied, thanks!

PR #4431, opened 2023-04-04
DBZ-6306 Use connection string configuration in ConnectorConfiguration#forMongoDbContainer

ani-sha:
https://issues.redhat.com/browse/DBZ-6306

### DIFF
--- a/debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/ConnectorConfiguration.java
+++ b/debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/ConnectorConfiguration.java
@@ -8,7 +8,6 @@ package io.debezium.testing.testcontainers;
 import java.util.List;
 
 import org.testcontainers.containers.JdbcDatabaseContainer;
-import org.testcontainers.containers.MongoDBContainer;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
@@ -39,12 +38,11 @@ public class ConnectorConfiguration {
 
     private static final String CONNECTOR = "connector.class";
     private static final String HOSTNAME = "database.hostname";
-    private static final String HOSTS = "mongodb.hosts";
+    private static final String CONNECTION_STRING = "mongodb.connection.string";
     private static final String PORT = "database.port";
     private static final String USER = "database.user";
     private static final String PASSWORD = "database.password";
     private static final String DBNAME = "database.dbname";
-    private static final String AUTO_DISCOVER_MEMBERS = "mongodb.members.auto.discover";
 
     public static ConnectorConfiguration forJdbcContainer(JdbcDatabaseContainer<?> jdbcDatabaseContainer) {
         ConnectorConfiguration configuration = new ConnectorConfiguration();
@@ -68,14 +66,10 @@ public class ConnectorConfiguration {
         return configuration;
     }
 
-    public static ConnectorConfiguration forMongoDbContainer(MongoDBContainer mongoDbContainer) {
-        final List<Integer> exposedPorts = mongoDbContainer.getExposedPorts();
-
+    public static ConnectorConfiguration forMongoDbContainer(MongoDbContainer mongoDbContainer) {
         ConnectorConfiguration configuration = new ConnectorConfiguration();
         configuration.with(CONNECTOR, "io.debezium.connector.mongodb.MongoDbConnector")
-                .with(HOSTS, "rs0/" + mongoDbContainer.getContainerInfo().getConfig().getHostName()
-                        + ":" + exposedPorts.get(0))
-                .with(AUTO_DISCOVER_MEMBERS, false);
+                .with(CONNECTION_STRING, "mongodb://" + mongoDbContainer.getNamedAddress());
 
         return configuration;
     }


Naros (approved):
These seem okay to me, but I'll defer to Jakub as a final say.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/ConnectorConfiguration.java
```diff
@@ -68,14 +66,13 @@ public static ConnectorConfiguration forJdbcContainer(JdbcDatabaseContainer<?> j
...
+    public static ConnectorConfiguration forMongoDbContainer(MongoDbContainer mongoDbContainer) {
         final List<Integer> exposedPorts = mongoDbContainer.getExposedPorts();
 
         ConnectorConfiguration configuration = new ConnectorConfiguration();
         configuration.with(CONNECTOR, "io.debezium.connector.mongodb.MongoDbConnector")
-                .with(HOSTS, "rs0/" + mongoDbContainer.getContainerInfo().getConfig().getHostName()
-                        + ":" + exposedPorts.get(0))
-                .with(AUTO_DISCOVER_MEMBERS, false);
+                .with(CONNECTION_STRING, "mongodb://" + mongoDbContainer.getContainerInfo().getConfig().getHostName()
+                        + ":" + exposedPorts.get(0));
```
rk3rn3r, line 75: Does this return the port inside the container or the outside port? We should make sure we use the right port here. That might have changed as before inside and outside port were equal and not automatically mapped.

ani-sha: I believe the inside port is returned here.

jpechane:
@jcechace Could you please review and merge this one? Thanks

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/ConnectorConfiguration.java
jcechace: I see this also switches to our contaienr. Apology for not noticing

jcechace:
@ani-sha Depending on what you need to do there are two options

1) Use `MongoDbReplicaSet` instead -- this is a class representing Mongo RS and technically it is the minimal viable unit intended to be run. This class implements `MongoDbDeployment` which directly provides `getConnectionString()` method. 

https://github.com/debezium/debezium/pull/4430# would have to be extended to allow specifying the image when creating `MongoDbReplicaSet` or `MongoDbShardedCluster` (for completeness sake).

2) If you really need to access `MongoDbContainer` directly then I would say we should have the container implement `MongoDbDeployment`

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/ConnectorConfiguration.java
jcechace: @ani-sha  also this is not the correct way to get the address here.

If you need an address which is addressable from client use `MongoDbContainer#getClientAddress()`. If ( as seems to be case here) you only need an address which is reachable within docker network, use `MongoDbContaienr#getNamedAddress()`

jcechace:
> @ani-sha Depending on what you need to do there are two options
> 
> 1. Use `MongoDbReplicaSet` instead -- this is a class representing Mongo RS and technically it is the minimal viable unit intended to be run. This class implements `MongoDbDeployment` which directly provides `getConnectionString()` method. This would be preferable
> 
> #4430 would have to be extended to allow specifying the image when creating `MongoDbReplicaSet` or `MongoDbShardedCluster` (for completeness sake).
> 
> 2. If you really need to access `MongoDbContainer` directly then I would say we should have the container implement `MongoDbDeployment`

Hmm thinking about it... If you need an internal connectionString than this interface won't be helpful (as that is intended to provide client connectionString).  So I would say for now, just use the `MongoDbContaienr#getNamedAddress()` as mentioned above.

rk3rn3r:
My 2cents: I agree we should use `MongoDbReplicaSet` in the UI and enable the functionality there too, if this is not inherited down.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/ConnectorConfiguration.java
```diff
@@ -68,14 +66,10 @@ public static ConnectorConfiguration forJdbcContainer(JdbcDatabaseContainer<?> j
...
-    public static ConnectorConfiguration forMongoDbContainer(MongoDBContainer mongoDbContainer) {
-        final List<Integer> exposedPorts = mongoDbContainer.getExposedPorts();
-
+    public static ConnectorConfiguration forMongoDbContainer(MongoDbContainer mongoDbContainer) {
         ConnectorConfiguration configuration = new ConnectorConfiguration();
         configuration.with(CONNECTOR, "io.debezium.connector.mongodb.MongoDbConnector")
-                .with(HOSTS, "rs0/" + mongoDbContainer.getContainerInfo().getConfig().getHostName()
-                        + ":" + exposedPorts.get(0))
-                .with(AUTO_DISCOVER_MEMBERS, false);
+                .with(CONNECTION_STRING, mongoDbContainer.getNamedAddress().toString());
```
jcechace, line 72: this returns a `MongoDbContainer.Address` instance and the `toString()` call resolves to `"<host>:<port>"`, which is not a valid connection string.

PR #4432, opened 2023-04-04
DBZ-6198 Update Oracle permissions to include statistics tables

Naros:
https://issues.redhat.com/browse/DBZ-6198


jpechane:
@Naros Applied, thanks

PR #4433, opened 2023-04-05
DBZ-6310 Document JDBC sink connector

Naros:
https://issues.redhat.com/browse/DBZ-6310

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -12,21 +12,21 @@ toc::[]
 
 [NOTE]
 ====
-This connector is currently in incubating state, i.e. exact semantics, configuration options etc. may change in future revisions, based on the feedback we receive.
+This connector is currently in an incubating state. In future versions, we might change the exact semantics, configuration options, and so forth, depending on the feedback that we receive.
 Please let us know if you encounter any problems.
 ====
 
 [[jdbc-overview]]
 == Overview
 
-{prodname}'s JDBC connector is a Kafka Connect sink connector implementation that allows you to consume from any number of source topics and write the events from those topics to any relational database using a JDBC driver.
-This connector supports a wide variety of database dialects, including Db2, MySQL, PostgreSQL, Oracle, and SQL Server.
+The {prodname} JDBC connector is a Kafka Connect sink connector implementation that can consume events from multiple source topics, and then write those events to a relational database by using a JDBC driver.
+This connector supports a wide variety of database dialects, including Db2, MySQL, Oracle, PostgreSQL, and SQL Server.
 
 [[how-the-jdbc-connector-works]]
 == How the JDBC connector works
 
 The {prodname} JDBC connector is a Kafka Connect sink connector, and therefore requires the Kafka Connect runtime.
-The JDBC connector periodically polls the Kafka topics that it is subscribed, and consumes events from those topics and writes them to the configured relational database.
+The connector periodically polls the Kafka topics that it subscribes to, consumes events from those topics, and then writes the events to the configured relational database.
 The connector supports idempotent write operations by using upsert semantics and basic schema evolution.
 
 The following features are supported:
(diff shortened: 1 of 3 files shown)


documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -0,0 +1,660 @@
...
+====
+
+[[jdbc-deployment]]
+== Deployment
+
+To deploy a {prodname} JDBC connector, you install the {prodname} JDBC connector archive, configure the connector, and start the connector by adding its configuration to Kafka Connect.
+
+.Prerequisites
+* link:https://zookeeper.apache.org/[Apache ZooKeeper], link:http://kafka.apache.org/[Apache Kafka], and link:{link-kafka-docs}.html#connect[Kafka Connect] are installed.
+* Destiniation database is installed and configured to accept JDBC connections.
```
jpechane, line 437:
```suggestion
* Destination database is installed and configured to accept JDBC connections.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -0,0 +1,660 @@
...
+
+.Procedure
+
+. Download the {prodname} https://repo1.maven.org/maven2/io/debezium/debezium-connector-jdbc/{debezium-version}/debezium-connector-oracle-{debezium-version}-plugin.tar.gz[JDBC connector plug-in archive].
+. Extract the files into your Kafka Connect environment.
+. Optionally download the JDBC driver from Maven Central and extract the downloaded driver file to the directory that contains the JDBC sink connector JAR file.
++
+NOTE: Drivers for Oracle and Db2 are not automatically shipped with the JDBC sink connector and must be manually installed.
+. Add the directory with the JAR files to {link-kafka-docs}/#connectconfigs[Kafka Connect's `plugin.path`].
```
jpechane, line 446: I am not sure about this wording. We usually don't add plug-in directory to the `plugin.path` but the directory that contains plugin directory.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -0,0 +1,660 @@
...
+* Replace the `${topic}` placeholder in the `table.name.format` configuration property with the event's topic.
+* Sanitize the table name by replacing dots (`.`) with underscores (`_`).
+
+|===
+
+[[jdbc-faq]]
+== Frequently asked questions
+
+* *Is the* `ExtractNewRecordState` *single message transformation required?* +
+No, that is actual one of the differentiating factors of the {prodname} JDBC connector from its competitors.
```
jpechane, line 635: Do we want to use term `competitor`? Maybe `other solution` would be better?

jpechane (commented):
@Naros LGTM, I think it is nice and well explained. I left very few typo/basic comments.

@roldanbob Could you please take a look at it?

documentation/modules/ROOT/pages/connectors/jdbc.adoc
Naros: I'll adjust this wording, but this was taken from the Oracle documentation, so @roldanbob we may need to address this in our other connector docs as well and unify the language.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -0,0 +1,661 @@
...
+The {prodname} JDBC connector uses a sophisticated type system to resolve a column's data type.
+You can refer to the xref:#jdbc-data-and-type-mappings[] section for details on how this type system resolves a specific field's schema definition to a JDBC type.
+If you desire a different data type mapping, you should define the table manually so you explicitly control the desired column type you need.
+
+* *How do you specify a prefix or a suffix to the table name without changing the Kafka topic name?* +
+In order to add a prefix or a suffix to the destination table  name, simply adjust the xref:#jdbc-property-table-name-format[table.name.format] connector configuration property and prepend or append the desired prefixed or suffix.
+For example, to prefix all table names with `jdbc_`, specify the `table.name.format` configuration property with a value of `jdbc_${topic}`.
+If such a connector is subscribed to a topic called `orders`, the resulting table will be created as `jdbc_orders`.
+
+* *Even when identifier quoting is not enabled, some columns are automatically quoted, why?* +
```
roldanbob, line 656:
```suggestion
* *Why are some columns automatically quoted, even though identifier quoting is not enabled?* +
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -0,0 +1,661 @@
...
+There are situations where specific column or table names may still be explicitly quoted even when `quote.identifiers` is not enabled.
```
roldanbob, line 657:
```suggestion
In some situations, specific column or table names might be explicitly quoted, even when `quote.identifiers` is not enabled.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -0,0 +1,661 @@
...
+This is often necessary when the column or table name starts with or uses a specific convention that would otherwise be considered illegal syntax.
+For example, when setting the xref:#jdbc-property-primary-key-mode[primary.key.mode] to `kafka`, certain databases will not allow a column's name to begin with an underscore unless the column's name is quoted.
```
roldanbob, line 659:
```suggestion
For example, when the xref:jdbc-property-primary-key-mode[primary.key.mode] is set to `kafka`, some databases only permit column names to begin with an underscore if the column's name is quoted.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -0,0 +1,661 @@
...
+This behavior is dialect specific and varies depending on the database used.
```
roldanbob, line 660:
```suggestion
Quoting behavior is dialect-specific, and varies among different types of database.
```

roldanbob (changes requested):
@Naros I know that this is a lot of changes to resolve using this format, but I started in doing the review in GH, and was pretty far along before I realized how many edits I'd be proposing.   :\

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -0,0 +1,660 @@
...
+When schema evolution is enabled, an `ALTER TABLE` SQL statement will be prepared and executed by the connector before applying the DML event to the destination table.
+The changing of column data types, dropping of columns, and adjustments to primary keys can be considered dangerous, and therefore none of these operations are performed by the connector.
+
+Whether a column is `NULL` or `NOT NULL` is derived directly from the optionality defined in the field's schema, and default values are also resolved from the field schema as well.
+If the connector attempts to create a table with an undesired nullability setting or default value, the table will either need to be created manually ahead of time or the schema of the associated field adjusted prior to be processing by the sink connector.
+This can be accomplished by introducing a custom single message transformation that adjusts these in the pipeline or modifying the column state in the source database.
```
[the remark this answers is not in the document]
Naros: What I am trying to convey is that if the source connector emits a field with an undesirable schema type, the user can adjust this by adding a single message transformation (SMT) to the pipeline either on the source before going into Kafka or the sink as the JDBC sink connector reads the data from Kafka, they can add a `CustomConverter` on the source connector, or they can alter the default value or nullability at the table level in the source database.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -0,0 +1,660 @@
...
+The {prodname} JDBC sink connector can delete rows in the destination database when a `DELETE` or _tombstone_ event is consumed.
+By default, the JDBC sink connector does not enable delete mode, therefore, it must be enabled explicitly in the connector configuration if you wish to support removing rows.
+
+Deletes can be enabled by setting `delete.enabled=true` in the connector configuration, but this mode requires that the `primary.key.mode` be set to a value that is not `none`.
+This is because deletes are executed based on the primary key mapping, and therefore is a destination table has no primary key mapping, rows cannot be deleted by the connector.
+
+[[jdbc-idempotent-writes]]
+=== Idempotent writes
+
+The {prodname} JDBC sink connector supports idempotent writes, allowing the same records to be replayed repeatedly and the final database state remains consistent.
```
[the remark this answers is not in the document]
roldanbob:
```suggestion
The {prodname} JDBC sink connector supports idempotent writes, so that even if the same records are applied multiple times, the final database state is not changed after the first application.
```
Ok, I thnk I get it. Does this ^ work?

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -0,0 +1,661 @@
...
+    "type": "INT8",
+    "parameters": {
+      "__debezium.source.column.type": "TINYINT",
+      "__debezium.source.column.length": "1"
+    }
+  }
+}
+----
+
+In the above example, without the schema parameters, the {prodname} JDBC sink connector would map this field to a column type of `Types.SMALLINT`, which will have varying logical database types depending on the dialect but for MySQL this would be a `TINYINT` column type without any length.
```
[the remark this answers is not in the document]
Naros: I think _no specified length_ makes logical sense.  I'll apply that wording instead.

Naros:
@jpechane @roldanbob edits applied.

jpechane:
@Naros @roldanbob Applied, thanks!

PR #4435, opened 2023-04-05
DBZ-6314: Add revoke statement for Oracle parser

mfvitale:
Fix for https://issues.redhat.com/browse/DBZ-6314

### DIFF
--- a/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/oracle/generated/PlSqlParser.g4
+++ b/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/oracle/generated/PlSqlParser.g4
@@ -95,7 +95,7 @@ unit_statement
     | drop_index
 
     | rename_object
-
+    | revoke_statement
     | comment_on_column
     | comment_on_table
     | comment_on_materialized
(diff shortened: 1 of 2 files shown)


Naros (commented):
Hi @mfvitale this LGTM.  

Would you please open a PR in the upstream repository, https://github.com/antlr/grammars-v4/tree/master/sql/plsql, including these changes so we keep the upstream repository aligned as well?

mfvitale:
> Hi @mfvitale this LGTM.
> 
> Would you please open a PR in the upstream repository, https://github.com/antlr/grammars-v4/tree/master/sql/plsql, including these changes so we keep the upstream repository aligned as well?

I took the changes from the upstream. I saw that we are a little bit disaligned, maybe we can create a task to align it. WDYT?

Naros:
Oh, if these are directly from upstream, then that's fine.  

Also, there have been several smaller deviations, and we certainly haven't kept the grammar synchronized from upstream like we probably should.  There is https://issues.redhat.com/browse/DBZ-2071 that I created some time ago to perform this alignment, but I would say that it's not necessary to spend any additional cycles on that at this time.  I believe there are much more pressing concerns that we should focus on right now.

PR #4436, opened 2023-04-05
DBZ-6216 Update testcontainers to use quay.io images

vjuranek:
When needed update also version.
Update also leftovers in readme files.

https://issues.redhat.com/browse/DBZ-6216


obabec (approved):
Lgtm

jpechane (changes requested):
@vjuranek I tried locally the test-containers and it fails too so it is probably related.

jpechane:
@vjuranek Applied thanks!

PR #4439, opened 2023-04-06
DBZ-6216 Fix container-based tests fetching from quay.io

joschi:
https://issues.redhat.com/browse/DBZ-6216


joschi:
Relates to #4436 but takes a different approach.

jpechane:
@joschi Thanks a lot, at the end I opted for a slightly different approach. Please see an extra commit on the linked PR.

PR #4440, opened 2023-04-06
DBZ-4488: Allow SQL Server connectors to set a maximum number of retries

jpechane:


jpechane:
@sugarcrm-jgminder Applied, thanks!

PR #4441, opened 2023-04-10
DBZ-5395 Improve Oracle offset management when LOB enabled

Naros:
https://issues.redhat.com/browse/DBZ-5395


Naros:
@jpechane can you review and merge this if it seems okay to you?

jpechane:
@Naros Applied, thanks

PR #4442, opened 2023-04-10
DBZ-6319 Update docs about async producer support for Pulsar sink

henkosch:
Proposing a new `async` option for the pulsar consumer in the following PR: 
https://github.com/debezium/debezium-server/pull/15

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -178,6 +178,7 @@ Hang Ruan
 Hans-Peter Grahsl
 Harvey Yue
 Helong Zhang
+Henrik Schnell
 Henry Cai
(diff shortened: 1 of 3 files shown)


jpechane:
@henkosch Could you pleas eopen the PR against `main` instead of `2.1`? Thanks

jpechane (changes requested):
@henkosch Could you please create the PR against main and update it to reflect the changed code? So no longer needed to talk about async much but the new timeout option needs to be added. Thanks

henkosch:
@jpechane Here is the new PR: https://github.com/debezium/debezium/pull/4462

PR #4448, opened 2023-04-11
DBZ-6265 always check the WAL_LEVEL but fail if streaming is required

alwaysbemark:
This is to address this comment: https://github.com/debezium/debezium/pull/4425#issuecomment-1501691602
This means that the WAL_LEVEL check always happens and an appropriate warning message will be given if it's not configured properly but the connector can function without it (initial_only mode).

Note: [Oracle](https://github.com/debezium/debezium/pull/4425/files) and MySQL have the desired behaviour already.

### DIFF
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresConnectorIT.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresConnectorIT.java
@@ -265,19 +265,6 @@ public class PostgresConnectorIT extends AbstractConnectorTest {
     }
 
     @Test
-    public void shouldSkipWalLevelCheckIfNoCdc() throws Exception {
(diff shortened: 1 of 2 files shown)


debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresConnectorIT.java
```diff
@@ -264,19 +264,6 @@ public void shouldValidateReplicationSlotName() throws Exception {
         assertConfigurationErrors(validatedConfig, PostgresConnectorConfig.SLOT_NAME, 1);
     }
 
-    @Test
```
alwaysbemark, line 267: I had to delete the test as there was no way to easily toggle the WAL_LEVEL in the integration test and anything that I'd add would probably introduce more complexity and flakiness than a change like this is worth.

PR #4449, opened 2023-04-12
DBZ-6324 Provide @Flaky annotation for CI unstable tests

jpechane:
https://issues.redhat.com/browse/DBZ-6324

### DIFF
--- a/debezium-core/src/test/java/io/debezium/junit/ConditionalFail.java
+++ b/debezium-core/src/test/java/io/debezium/junit/ConditionalFail.java
@@ -9,8 +9,11 @@ import java.lang.reflect.InvocationTargetException;
 import java.util.function.Supplier;
 
 import org.junit.Assert;
+import org.junit.Assume;
 import org.junit.runner.Description;
 import org.junit.runners.model.Statement;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 /**
  * JUnit rule that inspects the presence of the {@link ShouldFailWhen} annotation on a test method.
@@ -21,12 +24,26 @@ import org.junit.runners.model.Statement;
  */
 public class ConditionalFail extends AnnotationBasedTestRule {
 
+    private static final Logger FLAKY_LOGGER = LoggerFactory.getLogger(Flaky.class);
+
+    private static final String JIRA_BASE_URL = "https://issues.jboss.org/browse/";
+
     @Override
     public Statement apply(final Statement base, final Description description) {
         final ShouldFailWhen conditionClass = hasAnnotation(description, ShouldFailWhen.class);
-        if (conditionClass == null) {
-            return base;
+        if (conditionClass != null) {
+            return failOnCondition(base, description, conditionClass);
         }
+
+        final Flaky flakyClass = hasAnnotation(description, Flaky.class);
+        if (flakyClass != null) {
+            return ignoreFlakyFailure(base, description, flakyClass);
+        }
+        return base;
+    }
+
+    private Statement failOnCondition(final Statement base, final Description description,
+                                      final ShouldFailWhen conditionClass) {
         try {
             Supplier<Boolean> condition = conditionClass.value().getDeclaredConstructor().newInstance();
             return new Statement() {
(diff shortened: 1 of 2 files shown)


jpechane:
@Naros @vjuranek @jcechace Could I ask for the feedback? Generally the test method is annotated with a Jira id.
If the system property `ignoreFlakyFailures` is set to true then any failure is logged and the test result is changed to skipped.

I suppose we'd apply it only at PR workflows so the contributors will not be detracted and the regular branch builds will be still executed as are now as it is only us who investigate them.

rk3rn3r:
See my comment in the Jira task, what about using https://junit-pioneer.org/docs/retrying-test/ ?

jpechane:
@rk3rn3r IMHO it is JUnit 5 extension, we are on combination of 4 and 5.

vjuranek (approved):

debezium-core/src/test/java/io/debezium/junit/ConditionalFail.java
```diff
@@ -21,12 +24,26 @@
  */
 public class ConditionalFail extends AnnotationBasedTestRule {
 
+    private static final Logger FLAKY_LOGGER = LoggerFactory.getLogger(Flaky.class);
+
+    private static final String JIRA_BASE_URL = "https://issues.jboss.org/browse/";
```
Naros, line 29: I know it's nit-pick but we probably should use the new domain `issues.redhat.com` here rather than the old redirect address, who knows how long that redirect will last for.

debezium-core/src/test/java/io/debezium/junit/ConditionalFail.java
```diff
@@ -55,4 +72,37 @@ else if (failure != null) {
...
     }
+
+    private Statement ignoreFlakyFailure(final Statement base, final Description description,
+                                         final Flaky flakyClass) {
+
+        final String flakyFailuresProperty = System.getProperty(Flaky.IGNORE_FLAKY_FAILURES_PROPERTY);
+        if (flakyFailuresProperty == null || !Boolean.valueOf(flakyFailuresProperty)) {
+            return base;
+        }
+        final String flakyAttemptsProperty = System.getProperty(Flaky.FLAKY_ATTEMPTS_FAILURES_PROPERTY, "1");
```
Naros, line 83: I like this idea, and it raises a question for me.

Should we consider adding a workflow that we can manually trigger where we start a CI build for the specific connector module and be able to specify the test class and/or method name to re-run the test on the GitHub CI infrastructure?  If so, adding this to that workflow might be neat so we have a better way to combat testing on GitHub CI with race conditions rather than waiting 30-60 minutes for an entire test suite build.  What do others think?

Naros (commented):
Just a few inline comments.

debezium-core/src/test/java/io/debezium/junit/ConditionalFail.java
vjuranek: Yes, such workflow would be definitely useful.

debezium-core/src/test/java/io/debezium/junit/ConditionalFail.java
jpechane: Thanks, done. I copied it for very old history, not sure why :-)

debezium-core/src/test/java/io/debezium/junit/ConditionalFail.java
jpechane: Good proposal, logged via https://issues.redhat.com/browse/DBZ-6333

debezium-core/src/test/java/io/debezium/junit/Flaky.java
```diff
@@ -0,0 +1,30 @@
...
+ * failures of tests know to be unstable on CI, using the {@code ignoreFlakyFailures} system property.
+ *
+ * @author Jiri Pechanec
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD, ElementType.TYPE })
+public @interface Flaky {
+
+    String IGNORE_FLAKY_FAILURES_PROPERTY = "ignoreFlakyFailures";
+    String FLAKY_ATTEMPTS_FAILURES_PROPERTY = "flaky.attempts";
```
rk3rn3r, line 24: Same here. I think this is the number of retries? Why is the above option lowerCamelCase and this one a.dotted.name? A more descriptive name, like `flaky.tests.retries` seems to be closer to how such options are usually named.

debezium-core/src/test/java/io/debezium/junit/ConditionalFail.java
```diff
@@ -55,4 +72,37 @@ else if (failure != null) {
             throw new IllegalStateException(e);
         }
     }
+
+    private Statement ignoreFlakyFailure(final Statement base, final Description description,
+                                         final Flaky flakyClass) {
+
+        final String flakyFailuresProperty = System.getProperty(Flaky.IGNORE_FLAKY_FAILURES_PROPERTY);
```
rk3rn3r, line 79: According to the renaming proposed in my other 2 nit-pick comments I would rename those consts accordingly.

rk3rn3r:
Added just some nit-pick comments. LGTM else

jpechane:
@rk3rn3r Changed and merged

PR #4454, opened 2023-04-13
 DBZ-6260 documentation for RabbitMQ sink

olivierboudet:
This PR adds documentation for RabbitMQ sink added via https://github.com/debezium/debezium-server/pull/13


jpechane:
@olivierboudet Applied, thanks!

PR #4458, opened 2023-04-14
Support withstatement parsing

caicancai:
After mysql5.7, with cte analysis is supported, but MySqlParser.g4 does not currently support with keyword analysis.
Now improve MySqlParser.g4 to support with cte parsing

### DIFF
--- a/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlParser.g4
+++ b/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlParser.g4
@@ -74,7 +74,7 @@ dmlStatement
     : selectStatement | insertStatement | updateStatement
     | deleteStatement | replaceStatement | callStatement
     | loadDataStatement | loadXmlStatement | doStatement
-    | handlerStatement | valuesStatement |  | withStatement
+    | handlerStatement | valuesStatement | withStatement
     ;
 
 transactionStatement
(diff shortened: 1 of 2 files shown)


mfvitale:
Hi @caicancai, thanks for your contribution. 
Please add an example query into debezium-ddl-parser/src/test/resources/mysql/examples/dml_select.sql

jpechane:
@caicancai Thanks, also could you please

- [ ] create a new Jira that describes the issue
- [ ] modify the commit messge to start with BZ-xxx of the Jira
- [ ] port the fix o the [upstream grammar](https://github.com/antlr/grammars-v4/tree/master/sql/mysql/Positive-Technologies) too

debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlParser.g4
```diff
@@ -74,7 +74,7 @@ dmlStatement
     : selectStatement | insertStatement | updateStatement
     | deleteStatement | replaceStatement | callStatement
     | loadDataStatement | loadXmlStatement | doStatement
-    | handlerStatement | valuesStatement
+    | handlerStatement | valuesStatement |  | withStatement
```
jpechane, line 77: Are those two pipes with nothing in between correct?

caicancai: Sorry, there is an extra pipe here. It has been modified now

caicancai:
> @caicancai Thanks, also could you please
> 
> * [ ]  create a new Jira that describes the issue
> * [ ]  modify the commit messge to start with BZ-xxx of the Jira
> * [ ]  port the fix o the [upstream grammar](https://github.com/antlr/grammars-v4/tree/master/sql/mysql/Positive-Technologies) too
thanks for your reminder

PR #4459, opened 2023-04-15
DBZ-6336 MySql in debezium-parser-ddl does not support with keyword p…

caicancai:
After mysql5.7, with cte analysis is supported, but MySqlParser.g4 does not currently support with keyword analysis.
Now improve MySqlParser.g4 to support with cte parsing

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -508,3 +508,4 @@ Phạm Ngọc Thắng
 Moustapha Mahfoud
 Đỗ Ngọc Sơn
 RJ Nowling
+蔡灿材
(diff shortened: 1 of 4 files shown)


jpechane:
@caicancai Applied, thanks a lot! Could you please prepeare a PR for https://github.com/antlr/grammars-v4/tree/master/sql/mysql/Positive-Technologies and add a comment here with the link to it to make sure it is contributed back to upstream?

caicancai:
@jpechane  Ok,thank you for your reminder，i have given prepeare a PR for https://github.com/antlr/grammars-v4/tree/master/sql/mysql/Positive-Technologies

PR #4460, opened 2023-04-15
DBZ-6340 Add support to the Postgres connector for allow and prefer SSL modes  (with prefer as the new default value)

fredlegaucher:
The supported values for database.sslmode are disable, require, verify-ca and verify-full. In In addition to those values, the Postgres JDBC [driver](https://jdbc.postgresql.org/documentation/publicapi/org/postgresql/jdbc/SslMode.html) supports 2 more values:
* allow: try to connect to Postgres without encryption and, failing that, try encrypted 
* prefer: try to connect to Postgres with encryption and, failing that, try unencrypted

Let's add support for both of those modes and then make the prefer value the default mode.  

[DBZ-6340](https://issues.redhat.com/browse/DBZ-6340)

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/postgresql.adoc
+++ b/documentation/modules/ROOT/pages/connectors/postgresql.adoc
@@ -2788,10 +2788,14 @@ For example: `P1Y2M3DT4H5M6.78S`.
 For more information, see xref:postgresql-basic-types[PostgreSQL basic types].
 
 |[[postgresql-property-database-sslmode]]<<postgresql-property-database-sslmode, `+database.sslmode+`>>
-|`disable`
+|`prefer`
 |Whether to use an encrypted connection to the PostgreSQL server. Options include: +
  +
 `disable` uses an unencrypted connection. +
+ +
+`allow` attempts to use an unencrypted connection first and, failing that, a secure (encrypted) connection. +
+ +
+`prefer` attempts to use a secure (encrypted) connection first and, failing that, an unencrypted connection. +
  +
 `require` uses a secure (encrypted) connection, and fails if one cannot be established. +
  +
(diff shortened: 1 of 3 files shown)


jpechane:
@fredlegaucher Hi, thanks for the PR. This is definitely good change but as it changes the default behaviour we will include it in 2.3 as 2.2 is almost done. This change needs to be propely documented in release notes.

jpechane:
@fredlegaucher Applied, thanks

documentation/modules/ROOT/pages/connectors/postgresql.adoc
```diff
@@ -2788,11 +2788,15 @@ For example: `P1Y2M3DT4H5M6.78S`.
 For more information, see xref:postgresql-basic-types[PostgreSQL basic types].
 
 |[[postgresql-property-database-sslmode]]<<postgresql-property-database-sslmode, `+database.sslmode+`>>
-|`disable`
+|`prefer`
```
rk3rn3r, line 2791: Looking at those PRs for PG and MySQL.... instead of using the database's specific pass-through parameters and options... wouldn't it make more sense to provide a semantic option like `use.secure.connections` which can have 3 or more options, eg. `auto/prefer` --> tries to connect via secure connection but falls back to insecure when no secure connection can be established ; `enforce` --> tries secure connection and fails when secure connection cannot be established ; `disable` --> don't use secure connections at all ; `strict` --> added later to enforce cert validation ; 
Wdyt? @jpechane @Naros

rk3rn3r: I filed https://issues.redhat.com/browse/DBZ-6397 for this

PR #4461, opened 2023-04-16
DBZ-4427 basic test for incremental snapshot support with sharded MongoDB deployments

jcechace:
https://issues.redhat.com/browse/DBZ-4427

Few noteworthy comments 

1) At this point tests in MongoDB module would benefit from some cleanup and code deduplication 
2) currently the test uses a chunk size of 100. This is a change compared to non sharded variant of the tests due to yet unknown performance issue. When running the tests against sharded MongoDB incremental snapshot is almost 100x slower when the chunk size is extremely low (e.g. chunk size 10 for 10000 entries). This will be documented for the next release and the support marked as incubating. I will continue with the investigation in the meantime.

### DIFF
--- a/debezium-connector-mongodb/src/test/resources/logback-test.xml
+++ b/debezium-connector-mongodb/src/test/resources/logback-test.xml
@@ -22,6 +22,31 @@
         <appender-ref ref="CONSOLE" />
     </logger>
     <logger
+            name="io.debezium.connector.mongodb.ReplicaSetDiscovery"
+            level="warn" additivity="false">
+        <appender-ref ref="CONSOLE" />
+    </logger>
+    <logger
+            name="io.debezium.connector.mongodb.MongoDbIncrementalSnapshotChangeEventSource"
+            level="TRACE" additivity="false">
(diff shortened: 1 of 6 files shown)


jcechace:
I have to look into these test failures

jcechace:
@jpechane this can be merged I think.

debezium-connector-mongodb/src/test/resources/logback-test.xml
```diff
@@ -21,6 +21,31 @@
         level="warn" additivity="false">
         <appender-ref ref="CONSOLE" />
     </logger>
+    <logger
+            name="io.debezium.connector.mongodb.ReplicaSetDiscovery"
+            level="warn" additivity="false">
+        <appender-ref ref="CONSOLE" />
+    </logger>
+    <logger
```
jpechane, line 29: @jcechace Do we want to keep `TRACE` level loggers?

jpechane:
@jcechace If the logging level is not desirable then please create a follow-up PR.

PR #4464, opened 2023-04-18
DBZ-6325 Describe example configuration

rnowling-memphis:
Also, be consistent about using {prodname} instead of the constant string "Debezium".

This PR is for [DBZ-6325](https://issues.redhat.com/browse/DBZ-6325).


rnowling-memphis:
I'm not sure how to change the commit messages.  I propose squashing the commits on merge and setting the issue name in the squashed commit message.

jpechane:
@rnowling-memphis Applied, thanks!

PR #4467, opened 2023-04-18
DBZ-6331 Add support for WITH SYSTEM VERSIONING table option

ani-sha:
https://issues.redhat.com/browse/DBZ-6331


jpechane:
@ani-sha Please denote MariaDB specific keywords, constructs as such. Thanks

ani-sha:
@jpechane I believe these keywords can be considered as `keywordsCanBeId` in general, not mariadb specific keywords.

PR #4481, opened 2023-04-20
DBZ-6368 Align version of Antlr runtime

slknijnenburg:
Ensure all modules use antlr-runtime version 4.10.1. There is a breaking change in [Antlr 4.10 ](https://github.com/antlr/antlr4/releases/tag/4.10) that makes parsers from older versions incompatible.

Link: https://issues.redhat.com/browse/DBZ-6368

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -107,7 +107,7 @@
 
         <!-- ANTLR -->
         <!-- Align with Antlr runtime version pulled in via Quarkus -->
-        <version.antlr>4.8</version.antlr>
+        <version.antlr>4.10.1</version.antlr>
         <version.antlr4test.plugin>1.18</version.antlr4test.plugin>
 
         <!-- Quarkus -->


Naros:
Hi, @slknijnenburg I believe this is already in Debezium 2.2, released today.  Is there a need to backport this to 2.1 since we likely won't be making many major changes to 2.1 moving forward?

slknijnenburg:
> Hi, @slknijnenburg I believe this is already in Debezium 2.2, released today. Is there a need to backport this to 2.1 since we likely won't be making many major changes to 2.1 moving forward?

I wasn't aware 2.2 final was coming out so soon, that's good news :) For us it's no problem to upgrade as we're still in POC phase, but if people for any reason need to stay on 2.1.X it might still be worthwhile to merge as the MySQLConnector is broken on 2.1.3/2.1.4 when used in debezium-server.

I'll leave that decision to you, we can move forward either way :+1:

jpechane:
@slknijnenburg Thanks for the PR. I merged it. Albeit we don't plan to do another 2.1.x release it is still worthy for those whou'd build it from source.

PR #4485, opened 2023-04-21
DBZ-6360: Add job for build jdbc connector

mfvitale:
https://issues.redhat.com/browse/DBZ-6360

### DIFF
--- a/.github/workflows/debezium-workflow.yml
+++ b/.github/workflows/debezium-workflow.yml
@@ -894,6 +894,7 @@ jobs:
       - name: Build Debezium Connector JDBC
         run: >
           ./core/mvnw clean install -f jdbc/pom.xml -Passembly
+          -DskipITs=true
           -Dcheckstyle.skip=true
           -Dformat.skip=true
           -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
--- a/.github/workflows/debezium-workflow.yml
+++ b/.github/workflows/debezium-workflow.yml
@@ -843,6 +843,63 @@ jobs:
           -Dmaven.wagon.http.pool=false
           -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
 
+  build_jdbc:
+    needs: [ check_style, file_changes ]
+    if: ${{ needs.file_changes.outputs.common-changed == 'true'}}
+    name: "JDBC"
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


.github/workflows/debezium-workflow.yml
```diff
@@ -843,6 +843,63 @@ jobs:
...
+          -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
+
+      - name: Build Debezium Connector JDBC
+        run: >
+          ./core/mvnw clean install -f jdbc/pom.xml -Passembly
+          -Dcheckstyle.skip=true
+          -Dformat.skip=true
+          -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
+          -Dmaven.wagon.http.pool=false
+          -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
```
Naros, line 901: So in the JDBC connector repository, there are two types of tests:

**Integration tests**
These tests leverage a factory to construct monk `SinkRecord` instances, and the JDBC connector consumes these mock records.  Therefore connector behavior changes in the main repository won't be reflected in these tests.

**End to End tests**
These tests rely on spinning up a `debezium/connect:nightly` image to run the actual connector inside Kafka Connect and to verify full end-to-end behavior from source to sink connector pipelines.  Since the nightly images require code to be merged to the `main` branch, PR changes won't be reflected in the image build and therefore we won't see any impact of the PR changes until some future time when the image is baked.

Overall, I'm inclined to suggest we shouldn't run any type of integration tests and this should specify the `-DskipITs` toggle, but I'd be interested in what others think.

mfvitale: So it's just a check that changes on core doesn't breaks the compilation on jdbc sink

Naros: Agreed, although even that list is quite tiny, mostly dependency versions from the BOM, `testcontainers` artifact, and some bits around configuration. :grin:

jpechane: ok, let's skip ITs right now. As mentioned above the approach using locally built image would bve better but I'd recommend to log it as a separate follow-up Jira.

mfvitale: CI changed to skip IT and opened this follow-up https://issues.redhat.com/browse/DBZ-6383

jpechane:
@mfvitale Applied, thanks!

PR #4486, opened 2023-04-21
DBZ-6353 Use SQL state to identify missing function

jpechane:
https://issues.redhat.com/browse/DBZ-6353


jpechane:
Applied, verified by a community member.

PR #4487, opened 2023-04-21
DBZ-6379: Fix handling of toasted hstore

bpaquet:


jpechane:
@bpaquet Applied, thanks

PR #4488, opened 2023-04-21
DBZ-6329 Make sure Oracle source rs_id field populated

Naros:
https://issues.redhat.com/browse/DBZ-6329


jpechane:
@Naros Applied, thanks

PR #4490, opened 2023-04-24
DBZ-6384: Fix snapshotting with hstore in Map mode

bpaquet:


jpechane:
@bpaquet Applied, thanks!

PR #4491, opened 2023-04-24
DBZ-6355 retention last offset not sysdate

eizners:


Naros:
Superseded by PR #4492

PR #4492, opened 2023-04-24
Dbz 6355 retention last offset not sysdate update

eizners:


Naros (approved):
With the latest changes, I believe this is now set and LGTM.

PR #4494, opened 2023-04-25
DBZ-5182 Fetch Postgres types by ascending order

aristofanisl:
…ostgres types first. Does not override types that are already registered (e.g does not allow orafce module types to be registered and override default postgres types)

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
@@ -131,7 +131,12 @@ public class TypeRegistry {
 
     private void addType(PostgresType type) {
         oidToType.put(type.getOid(), type);
-        nameToType.put(type.getName(), type);
+        if (!nameToType.containsKey(type.getName())) {
+            nameToType.put(type.getName(), type);
+        }
+        else {
+            LOGGER.warn("Type [oid:{}, name:{}] is already mapped", type.getOid(), type.getName());
+        }
 
         if (TYPE_NAME_GEOMETRY.equals(type.getName())) {
             geometryOid = type.getOid();


Naros:
Hi @aristofanisl I'm curious if there is a way to fetch the default types first, purposely excluded any other types that may overwrite or extend the current types, and once the registry is primed with the defaults we then fetch the others with a separate query as an alternative to the `ORDER BY` clause?

aristofanisl:
Hi @Naros I think you are correct here. The delay builders should be sufficient. I checked the pg_type table there is only a unique constrain on oid and order by doesn't assure us that default type will get loaded first. I will remove it.
Should I add a log message of warning level when a type is not inserted in Map so the users will know when a duplicate type is found and not added in registry?

Naros:
> Should I add a log message of warning level when a type is not inserted in Map so the users will know when a duplicate type is found and not added in registry?

I'd prefer that to be the last resort if we can't find a way to handle this directly.  

With these orafce types, is there a potential issue here with n-level parent relationships being loaded incorrectly?  For example, the code as it is written handles the top-level correctly, but what happens if you have some type relationship like this:

root_type -> child_type -> grandchild_type

Is it possible these orafce types fall into this grandchild_type scenario?

aristofanisl:
I see what you mean. In my local environment with orafce installed searching for types that have parent this is what I see
image

The last column is the parent type name
My query is the following 
`SELECT t.oid AS oid, t.typname AS name, t.typelem AS element, t.typbasetype AS parentoid, t.typtypmod as modifiers, t.typcategory as category, e.values as enum_values , pt.typname as parent_type_name
	FROM pg_catalog.pg_type t JOIN pg_catalog.pg_namespace n ON (t.typnamespace = n.oid) 
LEFT JOIN  (SELECT t.enumtypid as id, array_agg(t.enumlabel) as values FROM pg_catalog.pg_enum t GROUP BY id ) e ON (t.oid = e.id)
left join pg_catalog.pg_type pt on t.typbasetype = pt.oid WHERE n.nspname != 'pg_toast' and t.typbasetype > 0;`

aristofanisl:
Hi @Naros I really dont know if it is possible and i dont see why not but with different type names.
The problem is what happens when 2 types have the same name e.g Date. From the looks of the code the type resolution is done by type name. When the TypeRegistry holds a map of name and type it will eventually override the prime type with the custom one.
This could be a bad thing that orafce has , i dont know. We have a lot of time with orafce installed on our environments with no issue though. Postgres seems to work with its native types and not the custom ones from modules unless the column has different type other than the native one.

jpechane:
@aristofanisl @Naros Hi, I was thinking hard about this probem and I came to the conclusion

- We need to provide a hothix in this PR. The hotfix will be executing two queries - first one to obtain the types stored in pg_catalog schema. These are the core ones and should be read first. Then the rest will be read and no overwrite will be in place.
- The final solution would require introducing schema anme into type registry and make sure that both pgoutput and pgproto decoding plugins would be able to provide them and work with them

aristofanisl:
@jpechane 
1. I believe this already happens with the exception that overwrites the map that holds the schema. My code fixes this issue but i dont know how general is my solution

jpechane:
@aristofanisl Yes, but I beleive to have core types loaded separately will make it even more robust solution.

Naros:
Hi @aristofanisl can you split the query for types as Jiri had suggested?

aristofanisl:
Hi all,
@jpechane , @Naros To recap, the problem here is when a type is deriving from another type and **has the same name with another type**. If a type is deriving from another one is indicated in postgres when the column **typbasetype** is not 0 (the number here is the oid of the parent) on pg_type table.
This makes the current code to overwrite any existing one in the nameToType Map, see here io.debezium.connector.postgresql.TypeRegistry#addType . 
I think even if we break down the type query into 2 queries (one with typbasetype = 0 and the other with typbasetype > 0) won't solve the issue since the code will still overwrite the one in the list.

So I think we need also the map key check prior to adding the type in the nameToType map.
image

Am I correct or do I miss something?

jpechane:
@aristofanisl The intent is not to allow a custom type to overwrite a core type. Can we be sure with the current query that the core types will be read and processed first and the custome type after them?

aristofanisl:
Hi all,
After using this in production (custom build) for 1 year we did not have any issues at all. The original query results, are sorted by the oid which in turn always has the native types first (nspname -> pg_catalog).

What are your thoughts on this? 
It is getting hard to maintain the fix every time DBZ gets a new version :)

jpechane (approved):
@aristofanisl In that case it sounds like a proved and tested solution :-)

I squased the commits and rebased on the latest main. If CI passes the PR can be merged.

jpechane:
@aristofanisl Applied, thanks

aristofanisl:
@jpechane is this included in any branch? I cannot see it anywhere

jpechane:
@aristofanisl It is in `main`. It will be added to `2.7` after next `2.7` release - at that time 2.7 branch is rebased on main

PR #4498, opened 2023-04-26
[docs] Wrong download link for jdbc connector

mfvitale:


Naros:
Applied to main(=2.3) and 2.2.  Thanks @mfvitale.

PR #4502, opened 2023-04-27
DBZ-6396 Honor slot.max.retries when committing LSN before starting replication

nirolevy:


jpechane:
@nirolevy LGTM, applied, thanks.

PR #4403, opened 2023-03-27
DBZ-6046 Add instructions for upgrading the PG database used by Debezium

roldanbob:
[DBZ-6046](https://issues.redhat.com/browse/DBZ-6046)

Adds steps for upgrading the PostgreSQL database that Debezium connects to.

Tested in a local Antora build.

I've embedded some questions as comments (L2295, L2298, L2314)

PR #4503, opened 2023-04-27
DBZ-6046 Add instructions for upgrading the PG database used by Debezium

roldanbob:
[DBZ-6046](https://issues.redhat.com/browse/DBZ-6046)

Replaces PR #4403. I've incorporated review comments that @jpechane submitted for that PR. 

This change adds a section to the PostgreSQL connector documentation to provide guidance about how to upgrade PostgreSQL databases that Debezium captures from.
Tested in a local Antora build and a local downstream build.


jpechane:
@roldanbob I think we are good to go. I left just one comment on one sentence.

documentation/modules/ROOT/pages/connectors/postgresql.adoc
```diff
@@ -2262,6 +2263,108 @@ Again, regularly emitting events solves the problem.
...
+Otherwise, {prodname} cannot capture the changes, resulting in data loss.
+
+ifdef::product[]
++
+For information about setting up a replication slot, see xref:configuring-a-replication-slot-for-the-debezium-pgoutput-plug-in[].
+endif::product[]
+
+14. Verify that the publication that defines the tables for {prodname} to capture is still present after the upgrade.
+If the publication is not available, connect to the database as a PostgreSQL administrator to create a new publication.
+
```
roldanbob, line 2340: @jpechane Reading up on PostgreSQL I learned that a replication slot is created automatically when you create a publication. So if a user finds that the publication is not preserved during the upgrade and it must be recreated, then we can assume that the replication slot is created. Should I reverse the order of steps 13 and 14?

Naros: I'm not aware of a publication creating a replication slot automatically.  I do know that a _subscription_ will create a remote replication slot on the observed PostgreSQL instance automatically, but a subscription and a publication are two very different things.  

A subscription in this context is a replica telling a primary "hey, I want you to send me data for tables A, B, and C".  

A publication on the other hand is a filter that is optionally associated with a replication slot (decoderbufs doesn't use these), which controls what data the slot is permitted to emit and how the data is to be emitted.  For example, you may have the REPLICA IDENTITY set to FULL for all tables, but a publication that includes tables A and C only means that changes for table B will be dropped by PostgreSQL and never sent to Debezium.

PR #4505, opened 2023-04-28
DBZ-6401 MySql in debezium-parser-ddl :The inserted sql statement rep…

caicancai:
https://issues.redhat.com/browse/DBZ-6401

Fix the insert statement, the form of INSERT ... ON DUPLICATE supports as as an alias, and this syntax is supported on the official website of mysql8.0
https://dev.mysql.com/doc/refman/8.0/en/insert.html


Naros:
Hi @caicancai, thanks for applying the change upstream. At one point in time, there were plans to introduce a separate grammar upstream for MySQL 8.0, but as you've seen, KvanTTT typically accepts 8.0 variants.

Naros:
Applied to main (=2.3) and branch 2.2.  Thanks @caicancai !

caicancai:
@Naros Thank you for your reply, if I want to participate in the syntax support of mysql8.0

PR #4506, opened 2023-04-28
DBZ-6402 Using MongoCommandException to detect invalid change stream …

jcechace:
…resume token in order to perform snapshot properly

https://issues.redhat.com/browse/DBZ-6402


Naros (approved):
LGTM

Naros:
Applied to main (=2.3) and branch 2.2.  

_SSO is currently down so I, so I cannot update the Jira, so lets make sure not to avoid updating the issue next week_.

PR #4512, opened 2023-05-02
DBZ-6411 Fix HeaderToValue schema update cache.

mfvitale:
https://issues.redhat.com/browse/DBZ-6411

When the connector starts for the first time with initial snapshot, records with read operation will be created. In this phase the ExtractChangedRecordState will produce no headers and since HeaderToValue was not skipping records without headers the cache of the new schema will be created only with the old fields.


jpechane:
@mfvitale Applied, thanks!

PR #4515, opened 2023-05-03
DBZ-1973: Enable Debezium to send notifications about it's status

mfvitale:
https://issues.redhat.com/browse/DBZ-1973

This is an example of the notification sent to Kafka topic

```json
{
  "schema": {
    "type": "struct",
    "fields": [
      {
        "type": "string",
        "optional": false,
        "field": "id"
      },
      {
        "type": "string",
        "optional": false,
        "field": "type"
      },
      {
        "type": "string",
        "optional": false,
        "field": "aggregate_type"
      },
      {
        "type": "map",
        "keys": {
          "type": "string",
          "optional": false
        },
        "values": {
          "type": "string",
          "optional": false
        },
        "optional": true,
        "field": "additional_data"
      }
    ],
    "optional": false,
    "name": "io.debezium.connector.common.Notification",
    "version": 1
  },
  "payload": {
    "id": "1a8a17d7-276f-4a16-9365-98178ec0e6e4",
    "type": "Status COMPLETED",
    "aggregate_type": "Initial Snapshot",
    "additional_data": {
      "name": "dbserver1"
    }
  }
}
```

### DIFF
--- a/debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
+++ b/debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
@@ -20,6 +20,7 @@ import java.util.function.Supplier;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
+import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigDef.Importance;
 import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.config.ConfigDef.Width;
@@ -39,6 +40,7 @@ import io.debezium.heartbeat.HeartbeatConnectionProvider;
 import io.debezium.heartbeat.HeartbeatErrorHandler;
 import io.debezium.heartbeat.HeartbeatImpl;
 import io.debezium.pipeline.ErrorHandler;
+import io.debezium.pipeline.notification.channels.SinkNotificationChannel;
 import io.debezium.relational.CustomConverterRegistry;
 import io.debezium.schema.SchemaNameAdjuster;
 import io.debezium.schema.SchemaTopicNamingStrategy;
@@ -385,6 +387,7 @@ public abstract class CommonConnectorConfig {
     private static final String CONVERTER_TYPE_SUFFIX = ".type";
     public static final long DEFAULT_RETRIABLE_RESTART_WAIT = 10000L;
     public static final long DEFAULT_MAX_QUEUE_SIZE_IN_BYTES = 0; // In case we don't want to pass max.queue.size.in.bytes;
+    public static final String NOTIFICATION_CONFIGURATION_FIELD_PREFIX_STRING = "notification.";
 
     public static final Field TOPIC_PREFIX = Field.create("topic.prefix")
             .withDisplayName("Topic prefix")
@@ -658,6 +661,13 @@ public abstract class CommonConnectorConfig {
                     + " matched against this regular expression. If matched the error is changed to retriable.")
             .withDefault(false);
 
+    public static final Field NOTIFICATION_ENABLED_CHANNELS = Field.create(NOTIFICATION_CONFIGURATION_FIELD_PREFIX_STRING + "enabled.channels")
+            .withDisplayName("Enabled notification channels names")
+            .withType(Type.LIST)
+            .withWidth(ConfigDef.Width.LONG)
+            .withImportance(Importance.MEDIUM)
+            .withDescription("List of notification channels names that is enabled.");
+
     protected static final ConfigDefinition CONFIG_DEFINITION = ConfigDefinition.editor()
             .connector(
                     EVENT_PROCESSING_FAILURE_HANDLING_MODE,
@@ -681,7 +691,9 @@ public abstract class CommonConnectorConfig {
                     SIGNAL_DATA_COLLECTION,
                     SIGNAL_POLL_INTERVAL_MS,
                     SIGNAL_ENABLED_CHANNELS,
-                    TOPIC_NAMING_STRATEGY)
+                    TOPIC_NAMING_STRATEGY,
+                    NOTIFICATION_ENABLED_CHANNELS,
+                    SinkNotificationChannel.NOTIFICATION_TOPIC)
             .create();
 
     private final Configuration config;
@@ -715,6 +727,9 @@ public abstract class CommonConnectorConfig {
     private final EnumSet<Operation> skippedOperations;
     private final String taskId;
 
+    private final String notificationTopicName;
+    private final List<String> enabledNotificationChannels;
+
     protected CommonConnectorConfig(Configuration config, int defaultSnapshotFetchSize) {
         this.config = config;
         this.emitTombstoneOnDelete = config.getBoolean(CommonConnectorConfig.TOMBSTONES_ON_DELETE);
@@ -744,6 +759,8 @@ public abstract class CommonConnectorConfig {
         this.signalEnabledChannels = getSignalEnabledChannels(config);
         this.skippedOperations = determineSkippedOperations(config);
         this.taskId = config.getString(TASK_ID);
+        this.notificationTopicName = config.getString(SinkNotificationChannel.NOTIFICATION_TOPIC);
+        this.enabledNotificationChannels = config.getList(NOTIFICATION_ENABLED_CHANNELS);
     }
 
     private static List<String> getSignalEnabledChannels(Configuration config) {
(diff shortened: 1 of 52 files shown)


debezium-core/src/main/java/io/debezium/pipeline/notification/channels/KafkaNotificationChannel.java
```diff
@@ -0,0 +1,77 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.debezium.config.CommonConnectorConfig;
+import io.debezium.function.BlockingConsumer;
+import io.debezium.pipeline.notification.Notification;
+import io.debezium.schema.SchemaFactory;
+import io.debezium.schema.SchemaNameAdjuster;
+
+public class KafkaNotificationChannel implements NotificationChannel, ConnectChannel {
```
jpechane, line 22: Could you please rename this class to `SinkNotificationChannel` and updated related code to it?

mfvitale: Done.

mfvitale:
Let me move the SPI files only to the core.

debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
```diff
@@ -83,6 +88,9 @@ public ChangeEventSourceCoordinator(Offsets<P, O> previousOffsets, ErrorHandler
         this.executor = Threads.newSingleThreadExecutor(connectorType, connectorConfig.getLogicalName(), "change-event-source-coordinator");
         this.eventDispatcher = eventDispatcher;
         this.schema = schema;
+
+        notificationService = NotificationService.getInstance();
```
jpechane, line 92: I think we should pass the `NotificationService` in the same way as we pass `SignalProcessor` and all the initialization should be done at connector startup so we can use it at an arbitrary place.

debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
```diff
@@ -136,6 +144,14 @@ protected void executeChangeEventSources(CdcSourceTaskContext taskContext, Snaps
         previousLogContext.set(taskContext.configureLoggingContext("snapshot", partition));
         SnapshotResult<O> snapshotResult = doSnapshot(snapshotSource, context, partition, previousOffset);
 
+        notificationService.initConnectChannels(Map.of(), Map.of(), SchemaFactory.get(), eventDispatcher::enqueueNotification);
```
jpechane, line 147: See above

jpechane (commented):
@mfvitale LGTM, thanks. The main chnage recommended is the wiring of the notification service and dispatcher.

debezium-connector-sqlserver/src/test/resources/META-INF/services/io.debezium.pipeline.notification.channels.NotificationChannel
```diff
@@ -0,0 +1,2 @@
+io.debezium.pipeline.notification.channels.SinkNotificationChannel
```
jpechane, line 1: @mfvitale This can be probably also removed

debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
mfvitale: I was in doubt between the two approach until the end. But definitely maybe the other one is better. A Dependency Injection framework would be useful. 

I'll do the changes.

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -637,6 +640,20 @@ public static FieldNameAdjustmentMode parse(String value) {
                     + " If set a stacktrace of non-retriable exception is traversed and messages are"
                     + " matched against this regular expression. If matched the error is changed to retriable.")
             .withDefault(false);
+    public static final Field NOTIFICATION_TOPIC = Field.create(NOTIFICATION_CONFIGURATION_FIELD_PREFIX_STRING + "topic.name")
```
jpechane, line 643: Please change the prefix to `notification.sink` for this option.

As the follow-up - the notification channel specific field definiitons should be moved to the channel if possible.

debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
jpechane: Yes, DI would be useful :-(. Generally there is nothing wrong with your approach but I'd prefer to keep the code consistent with other services too.

debezium-connector-sqlserver/src/test/resources/META-INF/services/io.debezium.pipeline.notification.channels.NotificationChannel
mfvitale: Are you refer to the file or the specific implementation?

jpechane:
@mfvitale Could you please address remaining comments (and remove test service loader files) and rebase as the signalling channel PR was merged. I'd like to get in Alpha1. Thanks!

mfvitale:
> @mfvitale Could you please address remaining comments (and remove test service loader files) and rebase as the signalling channel PR was merged. I'd like to get in Alpha1. Thanks!

Quite finished. I just want to add `NotificationIT` for Oracle, MySQL and MongoDB

debezium-core/src/test/resources/META-INF/services/io.debezium.pipeline.notification.channels.NotificationChannel
```diff
@@ -0,0 +1,2 @@
+io.debezium.pipeline.notification.channels.SinkNotificationChannel
```
jpechane, line 1: I think this file can be removed

jpechane:
@mfvitale I think we are almost there. I don't expect any changes in the API/SPI, so could you please

- [ ] Squash the existing commits
- [ ] Prepare PRs for non-core connector repos
- [ ] Add notifications for incremental snapshot

mfvitale:
The failure on `debezium-storage` is related to an issue on the CI workflow. I have opened https://issues.redhat.com/browse/DBZ-6443 to track it

mfvitale:
There are two open point:

- Initial snapshot message not work with MongoDB due to how partition are managed. I'll fix it with https://issues.redhat.com/browse/DBZ-6416
- There was a request from community to correlate `id` of incremental snapshot notifications with the one from `execute-snapshot`. What do you think to track it with a follow-up PR?

jpechane:
@mfvitale Please let's make a follow-up Jira for it so we can push this into the field

jpechane:
@mfvitale Applied, great job, thanks!

debezium-core/src/main/java/io/debezium/pipeline/notification/channels/KafkaNotificationChannel.java
rk3rn3r: Why did you name it "**Sink**NotificationChannel"? I feel this makes an assumption (where it's hard to make an assumption) about the target, but the event has info from the source (source incr snapshot state). Furthermore, maybe some stream processing apps use this information, those aren't sinks, they transform or aggregate data in motion and those aren't sinks. Same goes for monitoring systems that use those events to show the status "realtime". Could this be used for other events too in the future? I feel the name is not the best considering what is inside those messages and how it can be used (yet and in the future).
Wdyt?

jpechane: @rk3rn3r Hi, the sink name was chosen to reflect the situation that the notofication goes out via Kafka Connect API, nothing more nothing less.

mfvitale: This should be the name of the `output` channel. With the connect deployment this is Kafka but for debezium server this can be other message broker the we usually call `sink`.  I agree that it can be to generic, maybe `MessageBrokerNotificationChannel` can fit well?

jpechane: @mfvitale IMHO MessageBroker would be the same issue. And I want to avoud using `Connect*` based names as well. The only other generic name I can come up with is `DestinationNotificationChannel` but for me it is the same as sink.

PR #4516, opened 2023-05-03
[docs] resolve grammatical error with note

sclarkson-zoomcare:
Property `skip.messages.without.change` in Postgresql Connector documentation has a note with a grammatical error.


sclarkson-zoomcare:
Can I get an issue in JIRA created for me? I don't see the issue listed and this is a grammatical fix.

jpechane:
@sclarkson-zoomcare Squashed, fixed commit messages and applied. Thanks!

PR #4520, opened 2023-05-05
DBZ-6435 MySql in debezium-parser-ddl does not support table keyword parsing

caicancai:


caicancai:
https://issues.redhat.com/browse/DBZ-6435

PR #4521, opened 2023-05-05
DBZ-6436 Improve Oracle LogMiner failure logging

Naros:
https://issues.redhat.com/browse/DBZ-6436

This change explicitly captures the state of two critical tables when the connector fails to start a LogMiner session, namely V$LOGMNR_LOGS and V$LOGMNR_PARAMETERS. While these were previously captured by the LogMinerDatabaseStateWriter, the old code required DEBUG log level and that isn't always feasible, so this change will capture these in ERROR log level regardless moving forward.

@jpechane I'd prefer we backport this as far back as you're comfortable, preferably 2.1, as it shouldn't be an intrusive change.


jpechane:
@Naros Applied, thanks

PR #4522, opened 2023-05-05
DBZ-6438 Add config to reset mysql binlog offset

kgalieva:
https://issues.redhat.com/browse/DBZ-6438

Allow reseting MySQL connector's offset

### Which use case/requirement will be addressed by the proposed feature?
Add a config to MySQL connector that allows overriding committed offset. Replaying binlog is useful in case of a data corruption incident or data loss incidents. For example an SMT could have an issue in filtering or transformation logic when events get lost or corrupted.

### Implementation
Add an optional config that takes an offset to force reset to. The format is the same as for committed offsets, so that the desired offset could be copied from the connector's offsets topic.
The previous offset gets overridden with the value from config if config is provided.
The schema history will be recovered up to the provided offset, because there could be schema changes between provided and the latest committed offset.

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
@@ -5,6 +5,7 @@
  */
 package io.debezium.connector.mysql;
 
+import java.io.IOException;
 import java.math.BigDecimal;
 import java.time.Duration;
 import java.util.Set;
@@ -14,6 +15,7 @@ import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigDef.Importance;
 import org.apache.kafka.common.config.ConfigDef.Type;
 import org.apache.kafka.common.config.ConfigDef.Width;
+import org.apache.kafka.connect.errors.ConnectException;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
@@ -24,6 +26,8 @@ import io.debezium.config.Field;
 import io.debezium.config.Field.ValidationOutput;
 import io.debezium.connector.AbstractSourceInfo;
 import io.debezium.connector.SourceInfoStructMaker;
+import io.debezium.document.Document;
+import io.debezium.document.DocumentReader;
 import io.debezium.function.Predicates;
 import io.debezium.jdbc.JdbcValueConverters.BigIntUnsignedMode;
 import io.debezium.jdbc.TemporalPrecisionMode;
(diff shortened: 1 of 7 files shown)


ironakj:
Hi @kgalieva , even I was working on the same since this is pretty much a very common scenario. Wanted to know, if taken from config, how do you ensure the config is only honored once? i.e. let's say if the connector restarts how would make sure whether to reset position or not

kgalieva:
Hi @ironakj, great question. The plan is to revert the `force.reset.offset` config soon after the connector starts running and committing the new offsets. An alternative or additional config option could allow retesting only an initial offset, to use it for an existing connector reset one would need to delete committed offsets before restarting a connector.

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/StreamingSourceIT.java
```diff
@@ -582,9 +582,6 @@ private void inconsistentSchema(EventProcessingFailureHandlingMode mode) throws
                 JdbcConnection connection = db.connect();
                 Connection jdbc = connection.connection();
                 Statement statement = jdbc.createStatement()) {
-            if (mode == null) {
```
jpechane, line 585: Why is this change necessary? It ensured that the DML was executed in the streaming phase.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
```diff
@@ -1123,6 +1135,18 @@ public int bufferSizeForStreamingChangeEventSource() {
         return config.getInteger(MySqlConnectorConfig.BUFFER_SIZE_FOR_BINLOG_READER);
     }
 
+    public Document getForceResetOffset() {
```
jpechane, line 1138: Please use `Optional`

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
```diff
@@ -729,6 +733,14 @@ public static SecureConnectionMode parse(String value, String defaultValue) {
             .withDefault(DEFAULT_BINLOG_BUFFER_SIZE)
             .withValidation(Field::isNonNegativeInteger);
 
+    public static final Field FORCE_RESET_OFFSET = Field.create("force.reset.offset")
```
jpechane, line 736: PLease use `Field.createInternal`

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
```diff
@@ -1123,6 +1135,18 @@ public int bufferSizeForStreamingChangeEventSource() {
...
     }
 
+    public Document getForceResetOffset() {
+        String value = config.getString(MySqlConnectorConfig.FORCE_RESET_OFFSET);
+        try {
+            return (value == null || value.isEmpty()) ? null
+                    : DocumentReader.defaultReader().read(value);
+        }
+        catch (IOException e) {
+            throw new ConnectException(String.format("Unable to parse the %s config json value %s",
```
jpechane, line 1145: PLease use `DebeziumException`

jpechane (changes requested):
@kgalieva Thanks for the PR. I must admit I am not really fan of the feature as is at it opens potential can of worms. Also having it public means that we would need to commit to a published offset format which we want to avoid.
But I am willing to compromise on it so we can gather a limited experience/feedback using it.
I proposed in one change to create an internal field out of it. It is an undocumented property then that is not fully supported and will give us quite a leeway. Is it ok for you?

Also one more question - if we'd introduce an offset editor/manipulator, would it be better/mor robust/systematic solution?

kgalieva:
> if we'd introduce an offset editor/manipulator, would it be better/mor robust/systematic solution?

I guess we could stop a connector, commit an offset we want to reset to manually and then restart the connector. I can test it. Is this what you are suggesting by an offset editor/manipulator?

jpechane:
@kgalieva Yes, that's what I am proposing.

kgalieva:
I've tried deleting a connector, overriding the offset and  then recreating the connector and it worked. I can close this PR so that upstream can change offset format in the future without worrying about breaking the internal API

jpechane:
@kgalieva Yes please, we really need to come up with a solution that will be robust, would apply only once and be reusable by all connectors.

PR #4523, opened 2023-05-06
DBZ-6366: Code Improvements for skip.messages.without.change

ironakj:
Addresses/Closes DBZ-6366

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbIncrementalSnapshotChangeEventSource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbIncrementalSnapshotChangeEventSource.java
@@ -170,7 +170,7 @@ public class MongoDbIncrementalSnapshotChangeEventSource
      */
     protected ChangeRecordEmitter<MongoDbPartition> getChangeRecordEmitter(MongoDbPartition partition,
                                                                            OffsetContext offsetContext, Object[] row) {
-        return new MongoDbSnapshotRecordEmitter(partition, offsetContext, clock, (BsonDocument) row[0]);
+        return new MongoDbSnapshotRecordEmitter(partition, offsetContext, clock, (BsonDocument) row[0], connectorConfig);
     }
 
     protected void deduplicateWindow(DataCollectionId dataCollectionId, Object key) {
@@ -465,7 +465,7 @@ public class MongoDbIncrementalSnapshotChangeEventSource
             Object[] firstRow = null;
 
             for (BsonDocument doc : collection.find(predicate).sort(new Document(DOCUMENT_ID, 1))
-                    .limit(connectorConfig.getIncrementalSnashotChunkSize())) {
+                    .limit(connectorConfig.getIncrementalSnapshotChunkSize())) {
                 rows++;
                 final Object[] row = new Object[]{ doc };
                 if (firstRow == null) {
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlChangeRecordEmitter.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlChangeRecordEmitter.java
@@ -27,16 +27,14 @@ public class MySqlChangeRecordEmitter extends RelationalChangeRecordEmitter<MySq
     private final OffsetContext offset;
     private final Object[] before;
     private final Object[] after;
-    private final boolean skipMessagesWithoutChange;
 
     public MySqlChangeRecordEmitter(MySqlPartition partition, OffsetContext offset, Clock clock, Operation operation, Serializable[] before,
-                                    Serializable[] after, boolean skipMessagesWithoutChange) {
-        super(partition, offset, clock);
+                                    Serializable[] after, MySqlConnectorConfig connectorConfig) {
+        super(partition, offset, clock, connectorConfig);
         this.offset = offset;
         this.operation = operation;
         this.before = before;
         this.after = after;
-        this.skipMessagesWithoutChange = skipMessagesWithoutChange;
     }
 
     @Override
@@ -63,10 +61,5 @@ public class MySqlChangeRecordEmitter extends RelationalChangeRecordEmitter<MySq
     protected void emitTruncateRecord(Receiver receiver, TableSchema tableSchema) throws InterruptedException {
         Struct envelope = tableSchema.getEnvelopeSchema().truncate(getOffset().getSourceInfo(), getClock().currentTimeAsInstant());
         receiver.changeRecord(getPartition(), tableSchema, Operation.TRUNCATE, null, envelope, getOffset(), null);
-    }
-
-    @Override
-    public boolean skipMessagesWithoutChange() {
-        return skipMessagesWithoutChange;
     }
 }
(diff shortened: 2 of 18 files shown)


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbIncrementalSnapshotChangeEventSource.java
```diff
@@ -460,7 +460,7 @@ private void createDataEventsForDataCollection(MongoDbPartition partition) throw
             Object[] firstRow = null;
 
             for (BsonDocument doc : collection.find(predicate).sort(new Document(DOCUMENT_ID, 1))
-                    .limit(connectorConfig.getIncrementalSnashotChunkSize())) {
+                    .limit(connectorConfig.getIncrementalSnapshotChunkSize())) {
```
ironakj, line 463: Typo in method name fix

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -176,7 +176,7 @@ protected void sendEvent(P partition, EventDispatcher<P, T> dispatcher, OffsetCo
      */
     protected ChangeRecordEmitter<P> getChangeRecordEmitter(P partition, T dataCollectionId,
                                                             OffsetContext offsetContext, Object[] row) {
-        return new SnapshotChangeRecordEmitter<>(partition, offsetContext, row, clock);
+        return new SnapshotChangeRecordEmitter<>(partition, offsetContext, row, clock, connectorConfig);
```
ironakj, line 179: Not really necessary here, should I create a different constructor instead? @Naros @jpechane

debezium-core/src/main/java/io/debezium/pipeline/AbstractChangeRecordEmitter.java
```diff
@@ -24,10 +25,13 @@
     private final OffsetContext offsetContext;
     private final Clock clock;
 
-    public AbstractChangeRecordEmitter(P partition, OffsetContext offsetContext, Clock clock) {
+    private final CommonConnectorConfig connectorConfig;
+
+    public AbstractChangeRecordEmitter(P partition, OffsetContext offsetContext, Clock clock, CommonConnectorConfig connectorConfig) {
```
Naros, line 30: I'm not a fan of pushing `CommonConnectorConfig` and the `skipMessagesWithoutChange` method up this far into the class hierarchy as it forces MongoDB to explicitly return false, making most of these changes at this level in the class hierarchy a bit superfluous.

I'd prefer you only pass `CommonConnectorConfig` up to the `RelationalChangeRecordEmitter` and simply expose the `skipMessagesWithoutChange` there, which gives a clear indication that this is a relational connector-only feature as it isn't exposed at the top of the class hierarchy that both relational and non-relational connectors are based from.

Naros (changes requested):
One comment inline.

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: I think it's OK to add the constructor argument that way, but I made a suggestion below regarding how far up the class hierarchy this should be passed.

debezium-core/src/main/java/io/debezium/pipeline/AbstractChangeRecordEmitter.java
Naros: After reviewing the Jira issue, it seems we may plan to extend this to MongoDB, and in that case, this makes sense.

debezium-core/src/main/java/io/debezium/relational/RelationalChangeRecordEmitter.java
```diff
@@ -33,8 +34,8 @@
     public static final String PK_UPDATE_OLDKEY_FIELD = "__debezium.oldkey";
     public static final String PK_UPDATE_NEWKEY_FIELD = "__debezium.newkey";
 
-    public RelationalChangeRecordEmitter(P partition, OffsetContext offsetContext, Clock clock) {
-        super(partition, offsetContext, clock);
+    public RelationalChangeRecordEmitter(P partition, OffsetContext offsetContext, Clock clock, CommonConnectorConfig connectorConfig) {
```
jpechane, line 37: @ironakj Can we use `RelationalDatabaseConnectorConfig` here?

debezium-core/src/main/java/io/debezium/relational/SnapshotChangeRecordEmitter.java
```diff
@@ -19,8 +20,8 @@
 
     private final Object[] row;
 
-    public SnapshotChangeRecordEmitter(P partition, OffsetContext offset, Object[] row, Clock clock) {
-        super(partition, offset, clock);
+    public SnapshotChangeRecordEmitter(P partition, OffsetContext offset, Object[] row, Clock clock, CommonConnectorConfig connectorConfig) {
```
jpechane, line 23: @ironakj Can we use RelationalDatabaseConnectorConfig here?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlChangeRecordEmitter.java
```diff
@@ -27,16 +28,14 @@ public class MySqlChangeRecordEmitter extends RelationalChangeRecordEmitter<MySq
     private final OffsetContext offset;
     private final Object[] before;
     private final Object[] after;
-    private final boolean skipMessagesWithoutChange;
 
     public MySqlChangeRecordEmitter(MySqlPartition partition, OffsetContext offset, Clock clock, Operation operation, Serializable[] before,
-                                    Serializable[] after, boolean skipMessagesWithoutChange) {
-        super(partition, offset, clock);
+                                    Serializable[] after, CommonConnectorConfig connectorConfig) {
```
jpechane, line 33: Is it possible to use `MySqlConnectorConfig` here and the appropriate config types at other places too use the most specific type?

jpechane (changes requested):
@ironakj Thanks a lot. I left comments related the useage of specific types. Could you please also prepare PRs for non-core connector repos?

debezium-core/src/main/java/io/debezium/relational/RelationalChangeRecordEmitter.java
ironakj: Done

debezium-core/src/main/java/io/debezium/pipeline/AbstractChangeRecordEmitter.java
ironakj: Marking this as resolved

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
ironakj: Ok

PR #4526, opened 2023-05-09
DBZ-6435 MySql in debezium-parser-ddl does not support table keyword parsing

caicancai:
Support table keyword parsing in mysql8.0：https://dev.mysql.com/doc/refman/8.0/en/table.html


jpechane:
@caicancai Thanks for the PR. It seems the grammar is still not correct. Also if possible I'd recommend to provide a test at the integration test level to see how this statement behaves in the binlog, binlog parser and streaming processor. It looks like the command is effectively `SELECT` statement so we have to check if it is processed as such.

mfvitale:
@jpechane I think it can be merged. Integration test passes.

jpechane:
@caicancai @mfvitale Applied, thanks a lot!

jpechane:
@caicancai Could you please prepare a PR to upstream project as well? https://github.com/antlr/grammars-v4/tree/master/sql/mysql

Thanks

caicancai:
@jpechane thanks, I've reflowed upstream

PR #4528, opened 2023-05-09
DBZ-6442 Support FORCE-clause in DROP TABLE

Naros:
https://issues.redhat.com/browse/DBZ-6442


jpechane:
@Naros Applied, thanks

PR #4529, opened 2023-05-09
DBZ-6421 Support MongoDB Connector with external cloud databases

jcechace:
- Test are runnable against external databases
- Tests are runnable against DocumentDB specifically

In terms of DocumentDB compatibility 

- Unsupported `$addField` operator in change stream pipeline was replaced with `$replaceRoot`
- Connector now properly handles change stream cursors which are not returning a resume token when iterating over empty change stream. In such cases streaming is started from operation time. Without this change the connector was losing changes between snapshot end and streaming start and potentially even some changes during snapshot runtime (e.g. changes to already snapshotted documents). 
- To clean few things up we are now setting timestamp for offsets only when resume token is not available.


jcechace:
There are still several tests failing for DocumentDB due to

- different description of changed nested fields / arrays compared to MongoDB
- inability to set transaction document lock timeout

See https://issues.redhat.com/browse/DBZ-6419 for detail

jcechace:
@jpechane should we port these to 2.2? I don't think there is anything backwards incompatible

jpechane:
@jcechace No, backporting is unnecessary

obabec (approved):
LGTM, after a long struggle, is finally in a working state. I am OK with merging it now or we can wait and explore what these tests will do in an environment that does not require change of /etc/hosts.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/SourceInfoTest.java
```diff
@@ -122,16 +122,16 @@ public void assertSourceInfoContents(SourceInfo source,
         assertThat(source.hasOffset(REPLICA_SET_NAME)).isEqualTo(hasOffset);
 
         Map<String, ?> offset = source.lastOffset(REPLICA_SET_NAME);
-        assertThat(offset.get(SourceInfo.TIMESTAMP)).isEqualTo(timestamp.getTime());
-        assertThat(offset.get(SourceInfo.ORDER)).isEqualTo(timestamp.getInc());
+        assertThat(offset.get(SourceInfo.TIMESTAMP)).isEqualTo((timestamp != null) ? timestamp.getTime() : 0);
```
jpechane, line 125: Is it possible to introduce something like `io.debezium.connector.postgresql.DecoderDifferences` ?

jcechace: Likely yes -- at it will be useful for tests that include update descriptions. However I would rather do it as a follow up PR.

jpechane:
@jcechace Applied, thanks

btiernay:
> @jcechace No, backporting is unnecessary

I think this would help solve the issue we are seeing in 2.2.1 on startup where we cannot initialize the streaming change source because the `resumeToken` is missing.

PR #4531, opened 2023-05-10
DBZ-6443 Add step to build Debezium Connector MySQL in store build job

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6443

### DIFF
--- a/.github/workflows/debezium-workflow.yml
+++ b/.github/workflows/debezium-workflow.yml
@@ -616,6 +616,17 @@ jobs:
           -Dmaven.wagon.http.pool=false 
           -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
 
+      - name: Build Debezium Connector MySQL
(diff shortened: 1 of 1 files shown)


.github/workflows/debezium-workflow.yml
```diff
@@ -616,6 +616,17 @@ jobs:
           -Dmaven.wagon.http.pool=false 
           -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
 
+      - name: Build Debezium Connector MySQL
```
jpechane, line 619: @mfvitale Can this just be solved by changing the name of build core step and adding MySQL to the list of projects?

mfvitale: I separated to avoid the `am` flag and avoid to build dependencies on MySQL connector (i.e debezium-storage-file)

PR #4533, opened 2023-05-10
DBZ-6355 Fix test failure - race condition

Naros:
https://issues.redhat.com/browse/DBZ-6355


jpechane:
@Naros Applied, thanks

PR #4539, opened 2023-05-11
DBZ-6439 Don't scan all the tables when loading schema

vjuranek:
* Enable set log level in tests (https://issues.redhat.com/browse/DBZ-6460)
* Don't scan all the tables when loading schema (https://issues.redhat.com/browse/DBZ-6439)


jpechane:
@vjuranek Applied, thanks!

PR #4541, opened 2023-05-11
DBZ-6462 Schedule SignalProcessor task with a more relaxed scheduling

mfvitale:
Removes parallelStreams while processing signalChannelReaders


jpechane:
@mfvitale Applied, thanks!

PR #4545, opened 2023-05-16
DBZ-6470 Signaling data collection document update

angsdey2:
DBZ-6470 While creating signaling table we need to have privileges to create table in source database not target


jpechane:
Applied via https://github.com/debezium/debezium/pull/4547, thanks!

PR #4548, opened 2023-05-16
DBZ-6439 Filter tables only when user want to capture schema for incl…

vjuranek:
…uded tables only

https://issues.redhat.com/browse/DBZ-6439


Naros (approved):
LGTM, thanks @vjuranek

Naros:
Applied to main (=2.3).

PR #4549, opened 2023-05-17
DBZ-6457 Use PDB for parallel snapshot connections

Naros:
https://issues.redhat.com/browse/DBZ-6457


jpechane:
@Naros Applied, thanks!

PR #4551, opened 2023-05-18
[DBZ-6476] Change encoding of JDBC offset data from UTF-16 to UTF-8

subkanthi:
- Change encoding of JDBC offset data from UTF-16 to UTF-8


jpechane:
@subkanthi Could you please squash all the commits (and especially remove the merge one) into a single one with message
`DBZ-6476 Change encoding of JDBC offset data from UTF-16 to UTF-8`?

Thanks

jpechane:
@subkanthi Applied, thanks!

PR #4552, opened 2023-05-19
DBZ-3202 Testsuite should deploy PostgreSQL with Primary-Secondary streaming replication

MartinMedek:


obabec (approved):
LGTM, good job investigation the replication in PostgreSQL

PR #4553, opened 2023-05-19
DBZ-6447 Correlate incremental snapshot notifications ids with execute signal

mfvitale:
Incremental snapshot notification ids are now equal to the id sent in the `execute-snapshot` signal.

closes: https://issues.redhat.com/browse/DBZ-6447


jpechane:
@mfvitale Applied, good job!

PR #4556, opened 2023-05-23
DBZ-6490 Convert the field BigDecimal value to bytes failure of ApproximateStructSizeCalculator

harveyyue:
https://issues.redhat.com/browse/DBZ-6490


jpechane:
@harveyyue Applied, thanks!

PR #4560, opened 2023-05-24
DBZ-6502 Fix Oracle 23.x JDBC driver compatibility

Naros:
https://issues.redhat.com/browse/DBZ-6502

Solves ORA-18176 error "{} not in any time zone.DATE" error.


jpechane:
@Naros Applied, thanks

PR #4530, opened 2023-05-09
WIP: DBZ-6219 Add doc that clarifies how connector snapshots populate schema history

roldanbob:
[DBZ-6219](https://issues.redhat.com/browse/DBZ-6219)

Enhance the connector documentation to clarify how initial snapshots capture the schemas of database tables.
[x]  Update  description of `schema.history.internal.store.only.captured.tables.ddl` property.
[ ]  Update Db2 snapshots topic
[x]  Update MySQL snapshots topic
[ ]  Update Oracle snapshots topic
[ ]  Update SQL Server snapshots topic
[ ] Add subtopic to schema history section that explains why snapshots do not honor the  `table.include.list` settings when capturing the table schema

PR #4561, opened 2023-05-24
DBZ-6219 Doc update to explain default snapshot behavior for capturing schema history

roldanbob:
[DBZ-6219](https://issues.redhat.com/browse/DBZ-6219)

Replaces PR #4530 
This version incorporates feedback received for the original PR.

Enhance the connector documentation to describe default schema capture behavior during initial snapshots, how to modify the defaults, and how to capture tables whose schema were not captured:

[x] Update description of schema.history.internal.store.only.captured.tables.ddl property in shared file (ref-connector-configuration-database-history-properties.adoc).
[x] Update Db2 snapshots topic
[x] Update MySQL snapshots topic
[x] Update Oracle snapshots topic
[x] Update SQL Server snapshots topic
[x] Add subtopic to Snapshots section that explains why snapshots do not honor the table.include.list settings when capturing the table schema
- [x] MySQL
- [x] Db2
- [x] Oracle
- [x] SQL Server

[ ] Add steps for adding tables excluded from initial snapshot
- [x] Db2
- [x] MySQL
- [x] Oracle
- [x] SQL Server 

Please backport to 2.3.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/db2.adoc
+++ b/documentation/modules/ROOT/pages/connectors/db2.adoc
@@ -144,8 +144,9 @@ If you configure a different snapshot mode, the connector completes the snapshot
 1. Establish a connection to the database.
 
 2. Determine which tables are in capture mode and should be included in the snapshot.
-  By default, the connector captures all non-system tables.
-  To have the connector capture a subset of tables or table elements, you can set a number of `include` and `exclude` properties to filter the data, for example, xref:db2-property-table-include-list[`table.include.list`] or xref:db2-property-table-exclude-list[`table.exclude.list`].
+   By default, the connector captures the data for all non-system tables.
+   After the snapshot completes, the connector continues to stream data for the specified tables.
+   If you want the connector to capture data only from specific tables you can direct the connector to capture the data for only a subset of tables or table elements by setting properties such as xref:{context}-property-table-include-list[`table.include.list`] or xref:{context}-property-table-exclude-list[`table.exclude.list`].
 
 3. Obtain a lock on each of the tables in capture mode.
   This lock ensures that no schema changes can occur in those tables until the snapshot completes.
@@ -153,18 +154,16 @@ If you configure a different snapshot mode, the connector completes the snapshot
 
 4. Read the highest (most recent) LSN position in the server's transaction log.
 
-5. Capture the schema of all tables that are in capture mode.
-The connector persists this information in its internal database schema history topic.
+5. Capture the schema of all tables or all tables that are designated for capture.
+The connector persists schema information in its internal database schema history topic.
 The schema history provides information about the structure that is in effect when a change event occurs. +
 +
 [NOTE]
 ====
 By default, the connector captures the schema of every table in the database that is in capture mode, including tables that are not configured for capture.
-However, for tables that are not configured for capture, the initial snapshot captures only their structure; it does not capture any table data.
+If tables are not configured for capture, the initial snapshot captures only their structure; it does not capture any table data.
 
-There is a benefit to persisting schema information for tables that you did not include in the initial snapshot.
-When {prodname} has information about the structure of all existing tables, it can readily capture events from any existing table should you later decide that is needed.
-If you're certain that it won't be necessary to capture data from other tables in the future, you can modify the default behavior by changing the value of the xref:{context}-property-database-history-store-only-captured-tables-ddl[`schema.history.internal.store.only.captured.tables.ddl`] property to `true`.
+For more information about why snapshots persist schema information for tables that you did not include in the initial snapshot, see xref:understanding-why-initial-snapshots-capture-the-schema-history-for-all-tables[Understanding why initial snapshots capture the schema for all tables].
 ====
 
 6. Release any locks obtained in Step 3.
@@ -204,7 +203,7 @@ If the connector stops again for any reason, after it restarts, it resumes strea
 The initial snapshot that a connector runs captures two types of information:
 
 Table data::
-Information about `INSERT`, `UPDATE`, and `DELETE` operations in tables that are named in the connector's xref:{context}-property-snapshot-include-collection-list[`snapshot.include.collection.list`] property.
+Information about `INSERT`, `UPDATE`, and `DELETE` operations in tables that are named in the connector's xref:{context}-property-table-include-list[`table.include.list`] property.
 Schema data::
 DDL statements that describe the structural changes that are applied to tables.
 Schema data is persisted to both the internal schema history topic, and to the connector's schema change topic, if one is configured.
@@ -213,18 +212,17 @@ After you run an initial snapshot, you might notice that the snapshot captures s
 By default, initial snapshots are designed to capture schema information for every table that is present in the database, not only from tables that are designated for capture.
 Connectors require that the table's schema is present in the schema history topic before they can capture a table.
 By enabling the initial snapshot to capture schema data for tables that are not part of the original capture set, {prodname} prepares the connector to readily capture event data from these tables should that later become necessary.
-
-For tables that are designated for capture in the `snapshot.include.collection.list`, the snapshot captures only the table structure.
-It does not capture table data.
+If the initial snapshot does not capture a table's schema, you must add the schema to the history topic before the connector can capture data from the table.
 
 In some cases, you might want to limit schema capture in the initial snapshot.
 This can be useful when you want to reduce the time required to complete a snapshot.
-But if the initial snapshot does not capture a table's schema, you must add the schema to the history topic before the connector can capture data from the table.
+Or when {prodname} connects to the database instance through a user account that has access to multiple logical databases, but you want the connector to capture changes only from tables in a specific logic database.
 
 .Additional information
 * xref:{context}-capturing-data-from-tables-not-captured-by-the-initial-snapshot[Capturing data from tables not captured by the initial snapshot (no schema change)]
 * xref:{context}-capturing-data-from-new-tables-with-schema-changes[Capturing data from tables not captured by the initial snapshot (schema change)]
-* xref:{context}-property-database-history-store-only-captured-tables-ddl[Using the `schema.history.internal.store.only.captured.tables.ddl` property to configure schema capture in an initial snapshot.  ].
+* Setting the xref:{context}-property-database-history-store-only-captured-tables-ddl[`schema.history.internal.store.only.captured.tables.ddl`] property to specify the tables from which to capture schema information.
+* Setting the xref:{context}-property-database-history-store-only-captured-databases-ddl[`schema.history.internal.store.only.captured.databases.ddl`] property to specify the logical databases from which to capture schema changes.
 
 // Type: procedure
 [id="db2-capturing-data-from-tables-not-captured-by-the-initial-snapshot"]
@@ -297,7 +295,7 @@ You can add the schema by running a new schema snapshot, or by running an initia
 .Procedure
 
 Initial snapshot captured the schema for all tables (`store.only.captured.tables.ddl` was set to `false`)::
-1. Edit the xref:{context}-property-table-include-list[`table.include.list`] or xref:{context}-property-snapshot-include-collection-list[`snapshot.include.collection.list`] property to specify the tables that you want to capture.
+1. Edit the xref:{context}-property-table-include-list[`table.include.list`] property to specify the tables that you want to capture.
 2. Restart the connector.
 3. Initiate an xref:{context}-incremental-snapshots[incremental snapshot] if you want to capture existing data from the newly added tables.
 
--- a/documentation/modules/ROOT/pages/connectors/db2.adoc
+++ b/documentation/modules/ROOT/pages/connectors/db2.adoc
@@ -137,7 +137,7 @@ For each change that the snapshot captures, the connector emits a `read` event t
 The following workflow lists the steps that {prodname} takes to create a snapshot.
 These steps describe the process for a snapshot when the xref:db2-property-snapshot-mode[`snapshot.mode`] configuration property is set to its default value, which is `initial`.
 You can customize the way that the connector creates snapshots by changing the value of the `snapshot.mode` property.
-If you configure a different snapshot mode, the completes the snapshot by using a modified version of this workflow.
+If you configure a different snapshot mode, the connector completes the snapshot by using a modified version of this workflow.
 
 .Default workflow that the {prodname} Db2 connector uses to perform an initial snapshot
 
@@ -439,28 +439,18 @@ However, the database schema can be changed at any time, which means that the co
 Also, a connector cannot necessarily apply the current schema to every event.
 If an event is relatively old, it's possible that it was recorded before the current schema was applied.
 
-////
-{prodname}requires the schema history so that it can apply the table structure in effect at the time an event occurs.
-When the connector is stopped and restarted if the database metadata has changed so that the current structure of the table does not match the structure that was in effect at the point in time we're reading changes from; which is why the topic is needed.
+To ensure correct processing of events that occur after a schema change, the {prodname} Db2 connector stores a snapshot of the new schema based on the structures of the Db2 change data tables, which mirror the structures of their associated data tables.
+The connector stores the table schema information, together with the LSN of operations the result in schema changes, in the database schema history Kafka topic.
+The connector uses the stored schema representation to produce change events that correctly mirror the structure of tables at the time of each insert, update, or delete operation.
 
-The connector reads offsets at start-up if they exist.
-It then uses the position to recover the schema history topic, but only up to the position specified by the offsets.
+When the connector restarts after either a crash or a graceful stop, it resumes reading entries in the Db2 change data tables from the last position that it read.
+Based on the schema information that the connector reads from the database schema history topic, the connector applies the table structures that existed at the position where the connector restarts.
 
-If you apply schema changes to a table after the position of the last event that is recorded in the offsets, when the connector restarts, it starts from its last known position, which, according to the offsets file, excludes the most recent schema data in the dbhistory file
-It ignores the subsequent schema changes, skipping over them without reporting their presence.
-So the connector tries to process the event at the next offset without having the schema data.
+If you update the schema of a Db2 table that is in capture mode, it's important that you also update the schema of the corresponding change table.
+You must be a Db2 database administrator with elevated privileges to update database schema.
+For more information about how to update Db2 database schema in {prodname} environments, see xref:db2-schema-evolution[Schema history eveolution].
 
-////
-
-To ensure correct processing of changes that occur after a schema change, Db2 includes in the transaction log not only the row-level changes that affect the data, but also the DDL statements that are applied to the database.
-As the connector reads the transaction log and comes across these DDL statements, it parses them and updates an in-memory representation of each table’s schema.
-The connector uses this schema representation to identify the structure of the tables at the time of each insert, update, or delete operation and to produce the appropriate change event.
-In a separate database schema history Kafka topic, the connector records all DDL statements along with the position in the transaction log where each DDL statement appeared.
-
-When the connector restarts after either a crash or a graceful stop, it starts reading the transaction log from a specific position, that is, from a specific point in time.
-The connector rebuilds the table structures that existed at this point in time by reading the database schema history Kafka topic and parsing all DDL statements up to the point in the transaction log where the connector is starting.
-
-This database schema history topic is for connector use only.
+The database schema history topic is for internal connector use only.
 Optionally, the connector can also xref:about-the-debezium-db2-connector-schema-change-topic[emit schema change events to a different topic that is intended for consumer applications].
 
 .Additional resources
--- a/documentation/modules/ROOT/pages/connectors/db2.adoc
+++ b/documentation/modules/ROOT/pages/connectors/db2.adoc
@@ -275,6 +275,34 @@ To configure custom topic names, you specify regular expressions in the logical 
 For more information about using the logical topic routing SMT to customize topic naming, see {link-prefix}:{link-topic-routing}#topic-routing[Topic routing].
 
 // Type: concept
+// ModuleID: how-debezium-db2-connectors-handle-database-schema-changes
+// Title: How {prodname} Db2 connectors handle database schema changes
+[[db2-schema-history-topic]]
+=== Schema history topic
+
+When a database client queries a database, the client uses the database’s current schema.
+However, the database schema can be changed at any time, which means that the connector must be able to identify what the schema was at the time each insert, update, or delete operation was recorded.
+Also, a connector cannot necessarily apply the current schema to every event.
+If an event is relatively old, it's possible that it was recorded before the current schema was applied.
+
+To ensure correct processing of changes that occur after a schema change, Db2 includes in the binlog not only the row-level changes that affect the data, but also the DDL statements that are applied to the database.
+As the connector reads the binlog and comes across these DDL statements, it parses them and updates an in-memory representation of each table’s schema.
+The connector uses this schema representation to identify the structure of the tables at the time of each insert, update, or delete operation and to produce the appropriate change event.
+In a separate database schema history Kafka topic, the connector records all DDL statements along with the position in the binlog where each DDL statement appeared.
+
+When the connector restarts after either a crash or a graceful stop, it starts reading the binlog from a specific position, that is, from a specific point in time.
+The connector rebuilds the table structures that existed at this point in time by reading the database schema history Kafka topic and parsing all DDL statements up to the point in the binlog where the connector is starting.
+
+This database schema history topic is for connector use only.
+Optionally, the connector can also xref:about-the-debezium-db2-connector-schema-change-topic[emit schema change events to a different topic that is intended for consumer applications].
+
+When the Db2 connector captures changes in a table to which a schema change tool such as `gh-ost` or `pt-online-schema-change` is applied, there are helper tables created during the migration process.
+The connector needs to be configured to capture change to these helper tables.
+If consumers do not need the records generated for helper tables, then a single message transform can be applied to filter them out.
+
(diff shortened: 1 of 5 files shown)


documentation/modules/ROOT/pages/connectors/mysql.adoc
```diff
@@ -373,54 +373,118 @@ For more information, see xref:mysql-schema-history-topic[schema history topic].
...
-+
-.Workflow for performing an initial snapshot with table-level locks
+After the snapshot process begins, if the process is interrupted due to connector failure, rebalancing, or other reasons, the process restarts after the connector restarts.
+
+After the connector completes the initial snapshot, it continues streaming from the position that it read in Step 5 so that it does not miss any updates.
+
+If the connector stops again for any reason, after it restarts, it resumes streaming changes from where it previously left off.
+
+If the connector stops for long enough, the binlog files that record the connector's position can expire.
+If the database purges the expired binlog files, the connector's position is lost, which causes the connector to revert to the _initial snapshot_ for its starting position.
```
Naros, line 473: I feel this sentence is a bit misleading.  By default, if the position in the logs points a place that is no longer available because the logs have been pruned, then the connector will fail to start throwing an error that a new snapshot must be taken.  Setting `snapshot.mode` to `when_needed` can circumvent this failure, but requires being explicitly configured.

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -83,20 +83,61 @@ Typically, the redo logs on an Oracle server are configured to not retain the co
...
-2. Obtains a `ROW SHARE MODE` lock on each of the captured tables to prevent structural changes from occurring during creation of the snapshot. {prodname} holds the locks for only a short time.
-3. Reads the current system change number (SCN) position from the server's redo log.
-4. Captures the structure of all relevant tables.
-5. Releases the locks obtained in Step 2.
-6. Scans all of the relevant database tables and schemas as valid at the SCN position that was read in Step 3 (`SELECT * FROM ... AS OF SCN 123`), generates a `READ` event for each row, and then writes the event records to the table-specific Kafka topic.
-7. Records the successful completion of the snapshot in the connector offsets.
+1. Establish a connection to the database.
+
+2. Determine the tables to be captured.
+By default, the connector captures all non-system tables.
```
Naros, line 98: I think we may want to use the phrase "By default, the connector captures all tables except those excluded from capture" and then provide a link to https://debezium.io/documentation/reference/2.3/connectors/oracle.html#schemas-that-the-debezium-oracle-connector-excludes-when-capturing-change-events.

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -83,20 +83,61 @@ Typically, the redo logs on an Oracle server are configured to not retain the co
...
+By default, the connector captures the schema of every table in the database that is in capture mode, including tables that are not configured for capture.
+However, for tables that are not configured for capture, the initial snapshot captures only their structure; it does not capture any table data.
+
+There is a benefit to persisting schema information for tables that you did not include in the initial snapshot.
+When {prodname} has information about the structure of all existing tables, it can readily capture events from any existing table should you later decide that is needed.
+If you're certain that it won't be necessary to capture data from other tables in the future, you can modify the default behavior by changing the value of the xref:{context}-property-database-history-store-only-captured-tables-ddl[`schema.history.internal.store.only.captured.tables.ddl`] property to `true`.
+====
+
+6. Releases the locks obtained in Step 3.
+Other database clients can now write to the database.
```
Naros, line 121: We don't actually do any database level locks, so I think it may be better to use the following phrase:
```suggestion
Other database clients can now write to the previously locked table(s).
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -83,20 +83,61 @@ Typically, the redo logs on an Oracle server are configured to not retain the co
...
+
+7. At the SCN position that was read in Step 4, the connectors scans the tables that are designated for capture (`SELECT * FROM ... AS OF SCN 123`).
+During the scan, the connector completes the following tasks:
```
Naros, line 124: I wonder if we shouldn't consider adding a small note about `UNDO_RETENTION` and the snapshot duration and the potential for ORA-01555; see the Oracle FAQ at the end about "What’s the cause for ORA-01555 and how to handle it?".

Naros (commented):
A few comments inline @roldanbob !

documentation/modules/ROOT/pages/connectors/oracle.adoc
roldanbob: > We don't actually do any database level locks, so I think it may be better to use the following phrase:

I assume that this statement is true for the Oracle connector only, yah?

documentation/modules/ROOT/pages/connectors/db2.adoc
```diff
@@ -217,6 +306,34 @@ If the default topic name do not meet your requirements, you can configure custo
...
+
+When the connector restarts after either a crash or a graceful stop, it starts reading the binlog from a specific position, that is, from a specific point in time.
+The connector rebuilds the table structures that existed at this point in time by reading the database schema history Kafka topic and parsing all DDL statements up to the point in the binlog where the connector is starting.
+
+This database schema history topic is for connector use only.
+Optionally, the connector can also xref:about-the-debezium-db2-connector-schema-change-topic[emit schema change events to a different topic that is intended for consumer applications].
+
+When the Db2 connector captures changes in a table to which a schema change tool such as `gh-ost` or `pt-online-schema-change` is applied, there are helper tables created during the migration process.
+The connector needs to be configured to capture change to these helper tables.
+If consumers do not need the records generated for helper tables, then a single message transform can be applied to filter them out.
```
roldanbob, line 333: This content appears in the MySQL connector schema history topic. Does it also apply to Db2, Oracle, and SQL Server?

documentation/modules/ROOT/pages/connectors/mysql.adoc
```diff
@@ -373,90 +374,216 @@ For more information, see xref:mysql-schema-history-topic[schema history topic].
...
+
+Users might be surprised to see that an intial snapshot captured schema information for non-captured tables, and might assume that an error occurred.
+But inital snapshots capture schema information for all tables by design.
+By persisting schema data for tables that are not part of the original capture set, the connector prepares itself to readily capture event data from these tables should that later become necessary.
+
+Because the database schema can change at any time, {prodname} must be able to identify the schema that is in effect for a table at the time that any event occurs.
+Then, if the connector later stops, when it resumes reading from the binlog, it can correlate event information with information from the schema history topic to ensure that it applies the correct schema to each event.
+
+If you exclude schema capture for non-captured tables, and you later want to capture data from such a table, an exception results due to the unknown schema.
+Additional manual steps are required to add the new table's schema to the schema history topic.
```
roldanbob, line 581: I found some references in Zulip to how to add the schema for a table when the initial snapshot didn't capture it  (e.g., Delete the existing history topic, and then use schema_only_recovery to repopulate it) , but it isn't clear to me whether we want to include them in the documentation, and whether the steps might be consistent across connectors.

roldanbob:
I think that this is now all set, pending a final review. Let me know if further changes are needed.

jpechane:
@Naros Could you please do the final review?

documentation/modules/ROOT/pages/connectors/db2.adoc
Naros: These are only MySQL specific.

documentation/modules/ROOT/pages/connectors/mysql.adoc
Naros: These are consistent across all connectors sans PostgreSQL which doesn't have a schema history implementation; however, the caveat here remains that in order to recover the schema history, there cannot have been any schema changes made to the database between the time position of the transaction log offset and now.  In other words, when the connector re-reads the schemas using the JDBC metadata, the table structures must be identical to the expected table structures that all remaining transaction log entries are based on.  Make sense?

documentation/modules/ROOT/pages/connectors/db2.adoc
roldanbob: > These are only MySQL specific.

Ah, ok. Is the use of such tools in general also specific to MySQL? That is, for another DB, might a user apply a different schema change tool that would also generate files that the connector should capture?
Or should I remove these 3 lines from the doc for the other connectors?

Naros: I'm not aware of any, so I would recommend removing these from the other connectors.

rk3rn3r:
@Naros Here's something for after your holidays.

roldanbob:
@Naros I think this is now ready for you to take a look. The scope of this actually grew quite a bit as I dug into the details. IMO it wasn't enough to simply explain how to configure the schema history topic to store schema data only for intended tables, but it was important to explain the logic behind the default behavior, the consequences of changing the default, and how users could "recover" if they want to capture a table whose schema is not present in the history topic.

documentation/modules/ROOT/pages/connectors/db2.adoc
```diff
@@ -122,21 +129,225 @@ endif::product[]
...
+Schema data::
+DDL statements that describe the structural changes that are applied to tables.
+Schema data is persisted to both the internal schema history topic, and to the connector's schema change topic, if one is configured.
+
+After you run an initial snapshot, you might notice that the snapshot captures schema information for tables that are not designated for capture.
+By default, initial snapshots are designed to capture schema information for every table that is present in the database, not only from tables that are designated for capture.
+Connectors require that the table's schema is present in the schema history topic before they can capture a table.
+By enabling the initial snapshot to capture schema data for tables that are not part of the original capture set, {prodname} prepares the connector to readily capture event data from these tables should that later become necessary.
+
+For tables that are designated for capture in the `snapshot.include.collection.list`, the snapshot captures only the table structure.
```
Naros, line 217: I am somewhat concerned about our usage of `snapshot.include.collection.list`.  This information isn't wrong, but it's not complete either.  A user does not have to provide this property and when they do not, we derive the list of tables that are to be snapshot based on the include/exclude list configurations.  This is done by gathering all tables the user can access and then applying the filters to generate a finite list of "capturable" tables.  

A user may opt to specify a `collection.include.collections.list` property if they want to take a snapshot but perhaps they want to capture changes for tables A and B but only want to capture the initial data for B.

If this phrasing is used in other docs, we should be consistent there too.

documentation/modules/ROOT/pages/connectors/db2.adoc
```diff
@@ -122,21 +129,225 @@ endif::product[]
...
+.Prerequisites
+* You want to capture data from a table with a schema that the connector did not capture during the initial snapshot.
+* A schema change was applied to the table so that the records to be captured do not have a uniform structure.
+
+.Procedure
+
+Initial snapshot captured the schema for all tables (`store.only.captured.tables.ddl` was set to `false`)::
+1. Edit the xref:{context}-property-table-include-list[`table.include.list`] or xref:{context}-property-snapshot-include-collection-list[`snapshot.include.collection.list`] property to specify the tables that you want to capture.
+2. Restart the connector.
+3. Initiate an xref:{context}-incremental-snapshots[incremental snapshot] if you want to capture existing data from the newly added tables.
```
Naros, line 302: We're missing the reset on the offset/schema history topics.  Without that, the change in the configuration and connector restart won't impact the ability to perform an incremental snapshot.  

If this is used in other docs, we should be consistent there too.

documentation/modules/ROOT/pages/connectors/db2.adoc
```diff
@@ -217,12 +428,51 @@ If the default topic name do not meet your requirements, you can configure custo
...
+It then uses the position to recover the schema history topic, but only up to the position specified by the offsets.
+
+If you apply schema changes to a table after the position of the last event that is recorded in the offsets, when the connector restarts, it starts from its last known position, which, according to the offsets file, excludes the most recent schema data in the dbhistory file
+It ignores the subsequent schema changes, skipping over them without reporting their presence.
+So the connector tries to process the event at the next offset without having the schema data.
+
+////
+
+To ensure correct processing of changes that occur after a schema change, Db2 includes in the transaction log not only the row-level changes that affect the data, but also the DDL statements that are applied to the database.
+As the connector reads the transaction log and comes across these DDL statements, it parses them and updates an in-memory representation of each table’s schema.
```
Naros, line 456: For Db2, this isn't the case.  Db2 works in a similar way to SQL Server with an offline/online schema evolution procedure.  The actual raw DDL isn't captured, nor is it directly available, but instead, the connector stores a snapshot of the new schema based on the capture table and uses this in the same way that other connectors use the raw DDL.

documentation/modules/ROOT/pages/connectors/db2.adoc
```diff
@@ -217,12 +428,51 @@ If the default topic name do not meet your requirements, you can configure custo
...
+The connector uses this schema representation to identify the structure of the tables at the time of each insert, update, or delete operation and to produce the appropriate change event.
+In a separate database schema history Kafka topic, the connector records all DDL statements along with the position in the transaction log where each DDL statement appeared.
+
+When the connector restarts after either a crash or a graceful stop, it starts reading the transaction log from a specific position, that is, from a specific point in time.
+The connector rebuilds the table structures that existed at this point in time by reading the database schema history Kafka topic and parsing all DDL statements up to the point in the transaction log where the connector is starting.
+
```
Naros, line 462:
```suggestion
The connector rebuilds the table structures that existed at this point in time by reading the database schema history Kafka topic up to the point in the transaction log where the connector is starting.

```
We likely want to avoid the use of parsing here because that's not necessarily accurate.  For Db2 specifically, we use the previously cached schema representation that does not include any DDL at all.  For SQL Server we use the same strategy.  I'd probably just be consistent with removing this across all conenctors.

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -83,20 +83,67 @@ Typically, the redo logs on an Oracle server are configured to not retain the co
 As a result, the {prodname} Oracle connector cannot retrieve the entire history of the database from the logs.
 To enable the connector to establish a baseline for the current state of the database, the first time that the connector starts, it performs an initial _consistent snapshot_ of the database.
 
-You can customize the way that the connector creates snapshots by setting the value of the xref:oracle-property-snapshot-mode[`snapshot.mode`] connector configuration property.
-By default, the connector's snapshot mode is set to `initial`.
+The following workflow lists the steps that {prodname} takes to create a snapshot.
+These steps describe the process for a snapshot when the xref:{context}-property-snapshot-mode[`snapshot.mode`]  configuration property is set to its default value, which is `initial`.
+You can customize the way that the connector creates snapshots by changig the value of the `snapshot.mode` property.
+If you configure a different snapshot mode, the completes the snapshot by using a modified version of this workflow.
```
Naros, line 89:
```suggestion
If you configure a different snapshot mode, the connector completes the snapshot by using a modified version of this workflow.
```

documentation/modules/ROOT/pages/connectors/sqlserver.adoc
```diff
@@ -122,28 +122,221 @@ endif::product[]
...
+The initial snapshot captures the structure and data of the tables in the database.
 
-You can configure how the connector creates snapshots.
-By default, the connector's snapshot mode is set to `initial`.
-Based on this `initial` snapshot mode, the first time that the connector starts, it performs an initial _consistent snapshot_ of the database.
-This initial snapshot captures the structure and data for any tables that match the criteria defined by the `include` and `exclude` properties that are configured for the connector (for example, `table.include.list`, `column.include.list`, `table.exclude.list`, and so forth).
+The following workflow lists the steps that {prodname} takes to create a snapshot.
+These steps describe the process for a snapshot when the xref:{context}-property-snapshot-mode[`snapshot.mode`] configuration property is set to its default value, which is `initial`.
+You can customize the way that the connector creates snapshots by changing the value of the `snapshot.mode` property.
+If you configure a different snapshot mode, the completes the snapshot by using a modified version of this workflow.
```
Naros, line 131:
```suggestion
If you configure a different snapshot mode, the connector completes the snapshot by using a modified version of this workflow.
```

documentation/modules/ROOT/partials/modules/all-connectors/ref-connector-configuration-database-history-properties.adoc
```diff
@@ -39,16 +39,23 @@ Skipping should be used only with care as it can lead to data loss or mangling w
 |[[{context}-property-database-history-store-only-captured-tables-ddl]]<<{context}-property-database-history-store-only-captured-tables-ddl, `+schema.history.internal.store.only.captured.tables.ddl+`>>
 |`false`
 |A Boolean value that specifies whether the connector should record all DDL statements from specified schema or database +
+Specify one of the following values:
 
-`true` records only those DDL statements that are relevant to tables whose changes are being captured by {prodname}. Set to `true` with care because missing data might become necessary if you change which tables have their changes captured. +
+`false` (default):: During a database snapshot, the connector collects the schema data collected for all non-system tables in the database, including tables that are not designated for capture.
```
Naros, line 44:
```suggestion
`false` (default):: During a database snapshot, the connector collects the schema data for all non-system tables in the database, including tables that are not designated for capture.
```

documentation/modules/ROOT/partials/modules/all-connectors/ref-connector-configuration-database-history-properties.adoc
```diff
@@ -39,16 +39,23 @@ Skipping should be used only with care as it can lead to data loss or mangling w
...
+It's best to retain the default setting.
+If you later decide to capture changes from tables that you did not originally designate for capture, the connector can easily begin to capture data from those tables, because their schema structure is already stored in the schema history topic.
+{prodname} requires the schema history of a table so that it can identify the structure that was present at the time that a change event occurred.
 
-The safe default is `false`.
+`true`:: The snapshot records only those DDL statements that are relevant to the tables from which {prodname} captures change events.
```
Naros, line 49:
```suggestion
`true`:: The snapshot records only the table schemas that are relevant to the tables from which {prodname} captures change events.
```

documentation/modules/ROOT/partials/modules/all-connectors/ref-connector-configuration-database-history-properties.adoc
```diff
@@ -39,16 +39,23 @@ Skipping should be used only with care as it can lead to data loss or mangling w
...
+If you change the default value, and you later configure the connector to capture data from other tables in the database, the connector lacks the schema information that it requires to capture change events from the tables. +
 
 |[[{context}-property-database-history-store-only-captured-databases-ddl]]<<{context}-property-database-history-store-only-captured-databases-ddl, `+schema.history.internal.store.only.captured.databases.ddl+`>>
 |`false`
 |A Boolean value that specifies whether the connector should record all DDL statements +
+Specify one of the following values:
 
-`true` records only those DDL statements that are relevant to database/schema's tables whose changes are being captured by {prodname}. `false` will store all incoming DDL statements. +
+`true`:: Records only those DDL statements that are relevant to database/schema's tables whose changes are being captured by {prodname}.
```
Naros, line 57:
```suggestion
`true`:: Records only those schema structures that are relevant to database/schema's tables whose changes are being captured by {prodname}.
```

documentation/modules/ROOT/partials/modules/all-connectors/ref-connector-configuration-database-history-properties.adoc
```diff
@@ -39,16 +39,23 @@ Skipping should be used only with care as it can lead to data loss or mangling w
...
+`false`:: The schema history topics stores incoming DDL statements for all databases. +
```
Naros, line 58:
```suggestion
`false`:: The schema history topics store incoming schema changes for all databases. +
```

Naros (commented):
Hi @roldanbob, just a few things inline.  

Also I noticed we referenced the `store.only.captured.tables.ddl` within the connector docs sections on schema capture, but with the new `store.only.captured.databases.ddl` option, not sure if we want to include that at all or make a reference to the property.  It's fairly useful for connectors like MySQL and perhaps others in the future.

documentation/modules/ROOT/pages/connectors/db2.adoc
```diff
@@ -122,21 +129,225 @@ endif::product[]
...
-.. Produces a _read_ event for each row in each table that is in capture mode. All _read_ events contain the same LSN position, which is the LSN position that was obtained in step 3.
-.. Emits each _read_ event to the Kafka topic that has the same name as the table.
-. Records the successful completion of the snapshot in the connector offsets.
+.Default workflow that the {prodname} Db2 connector uses to perform an initial snapshot
+
+1. Establish a connection to the database.
+
+2. Determine which tables are in capture mode and should be included in the snapshot.
+  By default, the connector captures all non-system tables.
+  To have the connector capture a subset of tables or table elements, you can set a number of `include` and `exclude` properties to filter the data, for example, xref:db2-property-table-include-list[`table.include.list`] or xref:db2-property-table-exclude-list[`table.exclude.list`].
```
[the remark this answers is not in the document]
roldanbob: Well, yes, although, actually, it's both right? By default. 
I'll add a few sentences to clarify that initial snapshots capture data for all tables, unless you set up include/exclude lists and captures schema for all tables unless you configure `store.only.captured.tables.ddl` or `store.only.captured.tables.ddl`.

documentation/modules/ROOT/pages/connectors/db2.adoc
roldanbob: OK, point taken. I had initially mentioned the include/exclude list configurations, but then I realized that _technically_ the list of tables specified in those lists could be overridden by the `include.collections.list`.  So, in your view, in this context, would it be best to keep it simple and not mention the `include.collections.list`?

documentation/modules/ROOT/pages/connectors/db2.adoc
roldanbob: I thought that the steps to remove the history topic and clear the offsets applied only if the schema of a table was not captured by the initial snapshot? After a schema change are these steps required even for tables whose schema are saved in the schema history topic?

documentation/modules/ROOT/pages/connectors/db2.adoc
roldanbob: OK, right. When I added the schema history topics for Db2 and SQL Server, I took the content directly from the existing MySQL topic. I'll make the changes.

documentation/modules/ROOT/pages/connectors/db2.adoc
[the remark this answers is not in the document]
Naros: > Re: # 1, does the property affect the recording of schema structures both during an initial snapshot and at runtime? I.e., once Debezium starts recording schema data for a table, it continues to track future changes to the schema?

Yes that's correct, both of these settings affect the capture of schema and schema changes across both phases.

> Does the property exist only to enable schema capture for tables or databases that are named in exclude lists?
> ...
> Re: # 2 , I'm not sure what to make of "relevant to database/schema's tables ..." I think that is trying to say that when true, the connector captures schema only for databases or tables from which the connector captures data?


No.

Several databases, such as MySQL, Oracle, and SQL Server allow you to install the database engine once and then you can use that engine to manage multiple logical databases.  In this type of multi-tenancy deployment, we then usually refer to the engine as the "instance" and each logical database as a database.  In MySQL, these databases are also referred to as catalogs, and in Oracle they're referred to as pluggable databases or PDBs.

In databases like Oracle and MySQL, user accounts are shared across these databases within the instance and this means that a connector user may have visibility to select against any number of tables across multiple databases within that instance.  Most often, users are interested in capturing changes for a single databases within the instance, but since these users can access and have visibility to tables in other databases, you may want to prevent the schema history topic from containing references to these other tables in other databases that are of no interest.

For MySQL, a typical deployment configuration may look like:
```
database.include.list=inventory
table.include.list=inventory.tab1,inventory.tab2,inventory.tab2
schema.history.internal.store.only.captured.tables.ddl=false
schema.history.internal.store.only.captured.databases.ddl=true
```

In this example, if there is an `inventory` and `fulfillment` database on the MySQL instance, then the Debezium connector will not capture schema changes from the `fulfillment` database at all because I have specified the restriction on the `databases.ddl` as `true`; however the connector will successfully capture all schema changes related to all tables inside the `inventory` database because that database is included.

Hopefully, that makes that clearer.

> As a reader, it's also a little strange to be reading the Db2, Oracle, or SQL Server connector docs and see the description mention the default value for the MySQL connector. A relic of pulling the content from a shared file, but maybe we can do something to exclude that where it's not needed.

Are you referring to the describe of the database property?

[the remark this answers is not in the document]
roldanbob: Yes, exactly: https://github.com/debezium/debezium/blame/main/documentation/modules/ROOT/partials/modules/all-connectors/ref-connector-configuration-database-history-properties.adoc#L53

documentation/modules/ROOT/pages/connectors/db2.adoc
Naros: Probably.  Given a configuration as follows:
```
table.include.list=inventory.table1,inventory.table2
snapshot.include.collection.list=inventory.table2
```
In this case, only `table2` would have its data initially snapshot.  If this `snapshot.include.*` property wasn't defined, then the connector snapshots the data of both included tables.  So if we want to include something about this snapshot prefixed property, perhaps a small admonition would suffice?

documentation/modules/ROOT/pages/connectors/db2.adoc
roldanbob: @Naros For Db2 and SQL Server, after the schema evolution process, I assume that the connector stores the snapshot of the new schema in the history topic as for other connectors?

Naros (approved):
LGTM, thanks for the work on this @roldanbob.

Naros:

PR #4566, opened 2023-05-31
DBZ-6511 Restore Kafka signal offset only when KafkaSignalChannel is enabled

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6511

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlReadOnlyIncrementalSnapshotChangeEventSource.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlReadOnlyIncrementalSnapshotChangeEventSource.java
@@ -105,7 +105,11 @@ public class MySqlReadOnlyIncrementalSnapshotChangeEventSource<T extends DataCol
 
     private void restoreOffset() {
 
-        dispatcher.getSignalProcessor().restoreKafkaOffset(getContext().getSignalOffset());
+        KafkaSignalChannel kafkaSignal = dispatcher.getSignalProcessor().getSignalChannel(KafkaSignalChannel.class);
+
+        if (getContext().getSignalOffset() != null) {
+            kafkaSignal.reset(getContext().getSignalOffset());
+        }
     }
 
     private boolean isKafkaChannelEnabled() {
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlReadOnlyIncrementalSnapshotChangeEventSource.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlReadOnlyIncrementalSnapshotChangeEventSource.java
@@ -81,7 +81,6 @@ public class MySqlReadOnlyIncrementalSnapshotChangeEventSource<T extends DataCol
 
     private static final Logger LOGGER = LoggerFactory.getLogger(MySqlReadOnlyIncrementalSnapshotChangeEventSource.class);
     private final String showMasterStmt = "SHOW MASTER STATUS";
-    private final KafkaSignalChannel kafkaSignal;
 
     public MySqlReadOnlyIncrementalSnapshotChangeEventSource(RelationalDatabaseConnectorConfig config,
                                                              JdbcConnection jdbcConnection,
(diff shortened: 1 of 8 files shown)


jpechane:
@mfvitale Can MySQL's read-only incremental snapshot work without Kafka singal channel?
If not then IMHO the right soultion is just fail the connector validation if read-only snapshotting is started but Kafka channel not present.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlReadOnlyIncrementalSnapshotChangeEventSource.java
```diff
@@ -92,20 +91,27 @@ public MySqlReadOnlyIncrementalSnapshotChangeEventSource(RelationalDatabaseConne
...
-            kafkaSignal.seek(signalOffset);
+        if (isKafkaChannelEnabled()) {
+
+            restoreOffset();
         }
     }
 
+    private void restoreOffset() {
+
+        dispatcher.getSignalProcessor().restoreKafkaOffset(getContext().getSignalOffset());
```
jpechane, line 108: Instead of adding `restoreKafkaOffset` to `SignalProcessor` would it be possible to extend `SignalChannelReader` with a method with a generic type that could be overridden by implementations. In case of `KafkaSignalChannel` it will be type `long`.
This method would just ask for `kafka` channel to be provided by `SignalProcessor`. If available the method will be called and ignored otherwise.

mfvitale:
> @mfvitale Can MySQL's read-only incremental snapshot work without Kafka singal channel? If not then IMHO the right soultion is just fail the connector validation if read-only snapshotting is started but Kafka channel not present.

Yes, it do. Now signaling can be sent also with FileSignalChannel. 

I have added a specific test so this will be more clear.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlReadOnlyIncrementalSnapshotChangeEventSource.java
mfvitale: @jpechane Can you please better explain this sentence?
> This method would just ask for kafka channel to be provided by SignalProcessor. If available the method will be called and ignored otherwise.

jpechane: @mfvitale Sorry, it will call `getSignalChannel` to get the channel instance and call the method on it.

mfvitale: So if I understood well your proposal is to:

- add a method in `SignalChannelReader` let's call it `restoreOffset(T offset)` maybe with an empty default implementation
- implement this method in the  `KafkaSignalChannel` with Long type
- ySqlReadOnlyIncrementalSnapshotChangeEventSource will call getSignalChannel and the call the the method for restoring the offset

rk3rn3r: Wow `MySqlReadOnlyIncrementalSnapshotChangeEventSource` is a very long class name. Up for Scrabble anyone?

jpechane: @mfvitale Exactly, just I'd like to use different name, like `init`/`reset` or similar to make it generic and re-usable if needed elswhere.

mfvitale: Even if this is not strictly related to this PR, honestly I don't think it's to long. It tells exactly what it is. Do you think there is a better name?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlReadOnlyIncrementalSnapshotChangeEventSource.java
```diff
@@ -97,25 +97,21 @@ public MySqlReadOnlyIncrementalSnapshotChangeEventSource(RelationalDatabaseConne
...
-            restoreOffset();
-        }
+        restoreOffset();
     }
 
     private void restoreOffset() {
-
+        // Restore Kafka signal offsets if needed
         KafkaSignalChannel kafkaSignal = dispatcher.getSignalProcessor().getSignalChannel(KafkaSignalChannel.class);
+        if (kafkaSignal == null) {
```
mfvitale, line 106: This code is not equivalent. The`getSignalChannel` method only return registered channels but channel can be not enabled by user configuration.

jpechane: @mfvitale Thanks for explanation. Could you please remove my commit and just add a JavaDoc to the method? Then feel free to merge yourself.

PR #4570, opened 2023-06-01
DBZ-6518 [mongodb] use multipe threads to query chunks for incremental snapshot

xidui:
https://issues.redhat.com/browse/DBZ-6518

Previously, the mongodb debezium connector it's single threaded when querying chunks, which can be very slow if we want to snapshot a very large collection. Adding a multi-thread support in the mongo connector so that it can query mongo in parallel for different chunks.

On a high level here is what it does:
* Use `Projection` to get only the `_id` fields of the next `chunk * threads` docs as the `_id` field is always indexed by default.
* Iterate through all the ids and group them into multiple chunks for different threads to query
* Wait until all threads to finish and update the last chunk end so that the next run of the connector can pick up where it left off.

### DIFF
--- a/debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/MultiThreadIncrementalSnapshotIT.java
+++ b/debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/MultiThreadIncrementalSnapshotIT.java
@@ -12,7 +12,6 @@ import java.util.Map;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
-import java.util.function.Function;
 
 import org.junit.Test;
 


xidui:
@jpechane could you take a look when you have time, ty!

jpechane (changes requested):
@xidui Thanks for the PR. I rebased th code on the latest main and added few small changes.
The test is failing for MongoDB 4.4. Could you please take a look?

Thanks

xidui:
@jpechane 

I reset your refactors and checked in again to see if this will trigger the build again, but seems not.

I had some trouble building and testing this test case in my local laptop by executing the following
```
cd debezium-connector-mongodb
mvn clean verify -Dquick
```
I do this this dep is defined in code but it's still faining, I'm not sure how to fix this dependency issue.
image

By looking into the previous failure test logs on [github](https://github.com/debezium/debezium/actions/runs/5201323351/jobs/9381368785?pr=4570), seems like the connector was failed in the middle while executing:

xidui:
I was able to test it on my Intellij, but not command line, which is successful.
image

xidui:
@jpechane gentle ping on this

jpechane:
@xidui Hi, I am sorry for the delay. I'll be OOO now but will take care of it as soon as I return back. This PR/feature was prioritized  for 2.4.

xidui:
@jpechane may I get some eye on this PR pls?

Naros:
Hi @xidui, I just cleaned up the commit history and removed the merge commit and ran the test suite and everything passes.  I also cleaned up the checkstyle failures, so once CI passes we should be good for a final review.

jpechane:
@xidui Thanks for coming back! I've yet one more question - have you tried to run `IncrementalSnapshotIT` with multithreading enabled? I'd ideally create a new test `MultiThreadedIncrementalSnapshotIT` theat will extend from `IncrementalSnapshotIT` and just change the default config to use multiple threads. Ideally the test results should be the same.

xidui:
@jpechane 
> have you tried to run IncrementalSnapshotIT with multithreading enabled? I'd ideally create a new test MultiThreadedIncrementalSnapshotIT theat will extend from IncrementalSnapshotIT and just change the default config to use multiple threads. Ideally the test results should be the same.

Sure, I will do it this week.

xidui:
@jpechane I've updated the comment. Separated the multi-thread test cases into a different file

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbIncrementalSnapshotChangeEventSource.java
```diff
@@ -512,52 +521,49 @@ private void createDataEventsForDataCollection(MongoDbPartition partition) throw
...
+                addChunkToExecutor(collection, lastRow, futureList, lastChunkKey);
+            }
+
+            try {
+                for (Future<?> future : futureList) {
+                    future.get(); // Wait for the tasks to complete
+                }
+            }
+            catch (InterruptedException | ExecutionException e) {
+                throw new InterruptedException(e.toString());
```
jpechane, line 564: There is no need fo catching `InterruptedException` it can be propagagated directly from the method.
The `ExecutionException` should be handled like `throw new DebeziumException("Error while processing chunk", e)`

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/MultiThreadIncrementalSnapshotIT.java
```diff
@@ -0,0 +1,82 @@
...
+import org.junit.Test;
+
+import io.debezium.config.Configuration;
+
+/**
+ * Test to verify multi-thread incremental snapshotting for MongoDB.
+ *
+ * @author Yue Wang
+ */
+public class MultiThreadIncrementalSnapshotIT extends IncrementalSnapshotIT {
```
jpechane, line 25: You should override `io.debezium.connector.mongodb.IncrementalSnapshotIT.config()` here and set `MongoDbConnectorConfig.SNAPSHOT_MAX_THREADS`.
Ideally the inherited tests will just pass with it.

jpechane (changes requested):
@xidui Thanks, could you please take a look at the last two comments? One is just a minor one and the other is about more extensive testing.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbIncrementalSnapshotChangeEventSource.java
xidui: done

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/MultiThreadIncrementalSnapshotIT.java
xidui: done

xidui:
@jpechane updated code based on comments

jpechane:
@xidui Applied, thanks!

PR #4571, opened 2023-06-02
DBZ-6519 Parse mysql table name failed which ending with backslash

harveyyue:
https://issues.redhat.com/browse/DBZ-6519


jpechane:
@harveyyue This is realtively large change - not size but in impact. It LGTM. But could you please prepare upstream PR first and get it in? My reasoning is that the upstream guys might have different ideas on the grammar changes so we would then change back and forth to stay relatively aligned.

jpechane:
@harveyyue Applied, thanks

PR #4572, opened 2023-06-04
DBZ-6472 MySqlSnapshotChangeEventSource parallel execute createSchemaEventsForTables

harveyyue:
https://issues.redhat.com/browse/DBZ-6472

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
@@ -12,13 +12,20 @@ import java.sql.Statement;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.OptionalLong;
+import java.util.Queue;
 import java.util.Set;
+import java.util.concurrent.Callable;
+import java.util.concurrent.CompletionService;
+import java.util.concurrent.ExecutorCompletionService;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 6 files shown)


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
```diff
@@ -333,38 +346,77 @@ protected void readTableStructure(ChangeEventSourceContext sourceContext,
...
+            throws Exception {
+        CompletionService<Map<TableId, String>> completionService = new ExecutorCompletionService<>(executorService);
+        List<TableId> realTablesToRead;
+        if (firstPhase) {
+            realTablesToRead = tablesToRead.stream()
+                    .filter(id -> !delayedSchemaSnapshotTables.contains(id))
+                    .collect(Collectors.toList());
+        }
+        else {
+            realTablesToRead = new ArrayList<>(tablesToRead);
```
mfvitale, line 390: Maybe you can simply this removing the else and move initialization to line 383.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
```diff
@@ -333,38 +346,77 @@ protected void readTableStructure(ChangeEventSourceContext sourceContext,
...
+        }
+
+        if (realTablesToRead.size() > 0) {
+            for (TableId tableId : realTablesToRead) {
+                completionService.submit(() -> {
+                    JdbcConnection connection = connectionPool.poll();
+                    assert connection != null;
+                    try {
+                        return connection.queryAndMap("SHOW CREATE TABLE " + quote(tableId), rs -> {
```
mfvitale, line 399: WDYT to wrap this into a method to improve readability?

mfvitale:
Can this changes be generic and not MySQL specific?

harveyyue:
> Can this changes be generic and not MySQL specific?

I think it's not generic, since only the MySQL connector use the statement `show create table table_name` to get the schema per table, the parallel reading mechanism will improve this step.

harveyyue:
> @harveyyue I don't think we can do it this way. MySQL by default tries to obtain global read lock to capture schema snapshot. DOn't we end up with serial execution anyway as each of the threads will wait for getting the lock one by one?

Yes, you are right. I didn't consider this lock scenario, maybe it can be used to `none` locking mode.

jpechane:
@harveyyue Yes, in that case it could be separated. Could you please refactor the code to reflect that fact?

harveyyue:
> @harveyyue Yes, in that case it could be separated. Could you please refactor the code to reflect that fact?

Okay

harveyyue:
> @harveyyue I don't think we can do it this way. MySQL by default tries to obtain global read lock to capture schema snapshot. DOn't we end up with serial execution anyway as each of the threads will wait for getting the lock one by one?

Done.

jpechane:
@harveyyue LGTM! Could you please add test(s) to it too?

harveyyue:
> @harveyyue LGTM! Could you please add test(s) to it too?

OK

vjuranek (approved):

jpechane:
@harveyyue Applied, thanks!

PR #4579, opened 2023-06-06
DBZ-6529 Use Murmur hash for PartitionRouting SMT

vjuranek:
Use better hash function than default Java Object::hash function to get better hashes which would be more equally spred over the hash space and thus more equally over the Kafka partitions.

Use Murmur3 hash function for computing hashes of the fields. Murmur3 implementation is taken from from Infinispan project code base.

https://issues.redhat.com/browse/DBZ-6529

### DIFF
--- a/debezium-core/src/test/java/io/debezium/transforms/partitions/PartitionRoutingTest.java
+++ b/debezium-core/src/test/java/io/debezium/transforms/partitions/PartitionRoutingTest.java
@@ -184,6 +184,35 @@ public class PartitionRoutingTest {
         assertThat(transformedUpdateRecord).isNotEqualTo(updateRecord);
     }
 
+    @Test
+    public void byDefaultJavaHashIsUsed() {
+
+        partitionRoutingTransformation.configure(Map.of(
+                "partition.payload.fields", "change.id, change.product",
+                "partition.topic.num", 100));
+
+        final SourceRecord eventRecord = buildSourceRecord(productRow(1L, 1.0F, "orange"), CREATE);
+
+        SourceRecord transformed = partitionRoutingTransformation.apply(eventRecord);
+
+        assertThat(transformed.kafkaPartition()).isEqualTo(39);
+    }
+
+    @Test
+    public void murmurHashWillBeUsed() {
+
+        partitionRoutingTransformation.configure(Map.of(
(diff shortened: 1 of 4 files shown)


jpechane:
@vjuranek Can we make this configurable? The current users would end-up with incorrect partitioning after the upgrade.

vjuranek:
@jpechane sure, do you mean just a config option (with default to current implementation) or also SPI interface to eventually add user-defined hash function?

jpechane:
@vjuranek Just a config option. SMT itself is kind of SMT. If you make `computePartition` protected then would be possible to just override it in a subclass.

debezium-core/src/test/java/io/debezium/transforms/partitions/PartitionRoutingTest.java
```diff
@@ -134,7 +134,7 @@ public void onlyFieldThatExistForCurrentEventWillBeUsedForPartitionComputation()
 
         SourceRecord transformed = partitionRoutingTransformation.apply(eventRecord);
 
-        assertThat(transformed.kafkaPartition()).isOne();
+        assertThat(transformed.kafkaPartition()).isZero();
```
jpechane, line 137: @vjuranek Could you please change the test data in a way that has is non-zero? Zero could be default value so let's make the test more robust.

documentation/modules/ROOT/pages/transformations/partition-routing.adoc
```diff
@@ -267,6 +268,18 @@ For example, `after.name,source.table,change.name`
...
+| `java`
+|Hash function to be used when computing hash of the fields which would determine number of the destination partition.
+Possible values are: +
+ +
+`java` - standard Java `Object::hashCode` function +
+ +
+`murmur` - latest version of http://en.wikipedia.org/wiki/MurmurHash[MurmurHash] function, MurmurHash3 +
+ +
+This configuration is optional.
+If not specified or invalid value is used, the default value will be used.
```
mfvitale, line 282: Am I wrong or there is no mention about what the default value is?

mfvitale (approved):

jpechane:
@vjuranek Applied, thanks!

PR #4580, opened 2023-06-07
DBZ-6533: fix table order on partial match when snapshotting

isacandrei:
Fixes https://issues.redhat.com/browse/DBZ-6533


isacandrei:
@jpechane, thanks for the tip.

I'm having little trouble running the tests on a M1 Mac. The mysql container does not start, most likely because the image is not available on this platform.
Screenshot 2023-06-07 at 23 17 23

PR #4585, opened 2023-06-08
DBZ-6539 Adding operator release modifications to the pipeline

jcechace:
https://issues.redhat.com/browse/DBZ-6539


jcechace:
@jpechane  this should be the required change without image support

PR #4587, opened 2023-06-08
DBZ-6547 Declare exactly-once semantic support for Postgres connector

vjuranek:
https://issues.redhat.com/browse/DBZ-6547

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnector.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnector.java
@@ -15,6 +15,7 @@ import java.util.Map;
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigValue;
 import org.apache.kafka.connect.connector.Task;
+import org.apache.kafka.connect.source.ExactlyOnceSupport;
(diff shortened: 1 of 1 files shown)


jpechane:
@vjuranek Could you please create a new issue for posgres connector and attribute the PR/commit to it? I don't think we'll declare full EoS support in 2.3. Thanks

vjuranek:
sure @jpechane , updated

gunnarmorling:
Hey @vjuranek, @jpechane, out of curiosity: what will happen in terms of EO in case a connector gets restarted while an (initial or incremental) snapshot is running?

vjuranek:
hi @gunnarmorling , this is EOS support only for streaming. If the initial snapshot is not finished successfully, next time it's started again and therefore would result into duplicates. Probably the only way how to ensure EOS for initial snapshot is to run it in one huge transaction, which doesn't seem to be reasonable. On the other hand, for incremental snapshots sending the snapshot chunks in transaction seems to be doable, but the implementation itself is still TBD. See [this Zulip topic](https://debezium.zulipchat.com/#narrow/stream/302533-dev/topic/exactly.20once.20semantics) for more details. Any suggestions/ideas are welcome!

jpechane:
@vjuranek Applied, thanks for the investigation.

PR #4588, opened 2023-06-08
DBZ-6537 Allow "national" as identifier in MySql

tommyk-gears:
https://issues.redhat.com/browse/DBZ-6537


jpechane:
@tommyk-gears Applied, thanks!

PR #4594, opened 2023-06-09
DBZ-6543 Fix NPE due to optional fields in PartitionRouting SMT

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6543


jpechane:
@mfvitale Applied, thanks

PR #4595, opened 2023-06-09
DBZ-6499 Support restarting JDBC connections

Naros:
https://issues.redhat.com/browse/DBZ-6499

### Changes

Introduces a new configuration option, `log.mining.restart.connection`, which closes and re-opens the JDBC connection when an Oracle Log switch occurs or when the optionally configured log mining session max lifetime is reached.

@jpechane I'm open to renaming the option if you think a better name is useful.


jpechane:
@Naros Applied, thanks!

PR #4596, opened 2023-06-09
[docs] add initial to table snapshot.mode in postgresql.adoc

TechIsCool:
This adds a bit more clarity into the `initial` mode since it was very confusing as a new user to understand that `initial` was the default value but didn't exist in the table.


jpechane:
@TechIsCool Thanks, applied via https://github.com/debezium/debezium/pull/4600

PR #4602, opened 2023-06-13
DBZ-6543 Unhandled NullPointerException in PartitionRouting will crash the whole connect plugin

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6543


jpechane:
@mfvitale Applied, thanks

PR #4608, opened 2023-06-14
DBZ-6534 Update MySQL test image from 5.7 to 8.0

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6534

### DIFF
--- a/.github/workflows/debezium-workflow.yml
+++ b/.github/workflows/debezium-workflow.yml
@@ -310,7 +310,41 @@ jobs:
           -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
           -Dmaven.wagon.http.pool=false
           -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
-            
+
+  ## Add this as until 5.7 will be EOL in October 2023
+  build_mysql_5_7:
+    needs: [ check_style, file_changes ]
+    if: ${{ needs.file_changes.outputs.common-changed == 'true' || needs.file_changes.outputs.mysql-changed == 'true' || needs.file_changes.outputs.mysql-ddl-parser-changed == 'true' || needs.file_changes.outputs.schema-generator-changed == 'true' }}
+    name: "MySQL 5.7"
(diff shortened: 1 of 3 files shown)


jpechane:
@mfvitale It might be good to add testing on MySQL 5.7 as part of GH workflow for the next 3 months - just in case

mfvitale:
@Naros Switching to MySQL 8.0 causes this test `MySqlJdbcSinkDataTypeConverterIT.testNationalizedCharacterDataTypeMappings:235` to fail wih `expected:<"utf8[]"> but was:<"utf8[mb3]">`

From [MySQL 8.0 docs](https://dev.mysql.com/doc/refman/8.0/en/charset-unicode-sets.html) 

> The utf8mb3 character set is deprecated and you should expect it to be removed in a future MySQL release. Please use utf8mb4 instead. utf8 is currently an alias for utf8mb3, but it is now deprecated as such, and utf8 is expected subsequently to become a reference to utf8mb4. Beginning with MySQL 8.0.28, utf8mb3 is also displayed in place of utf8 in columns of Information Schema tables, and in the output of SQL SHOW statements.
To avoid ambiguity about the meaning of utf8, consider specifying utf8mb4 explicitly for character set references.

It's strange to see this behavior since the table used during test has this charset declaration `CHARSET=utf8mb4;`
On tests with MySQL 5.7 the returned value is utf8 even if declared with `utf8mb4` and with MySQL 8.0 is utf8mb3.

Do you have any suggetion?

Naros:
So on MySQL 5.7, the DDL is constructed with a character set name of `utf8`, while on MySQL 8.0, the DDL refers to the character set `utf8mb3`.  

I considered whether we should account for this in the `ColumnDefinitionListener` of the DDL parser, but that wouldn't ultimately be wise.  Instead, I believe the right path forward should be to adjust the test so that we check that the character set starts with `utf8`, which would match 5.7 (utf8), 8.0.32 (utf8mb3), and the future re-aliasing to utf8mb4 and preserves existing behavior for users.

jpechane:
@mfvitale Applied, thanks

PR #4612, opened 2023-06-15
DBZ-6559 Restore backward compatibility for field that stars with a digit

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6559

### DIFF
--- a/debezium-core/src/main/java/io/debezium/schema/SchemaNameAdjuster.java
+++ b/debezium-core/src/main/java/io/debezium/schema/SchemaNameAdjuster.java
@@ -275,7 +275,12 @@ public interface SchemaNameAdjuster {
         }
         else {
             sb.append(replacement.replace(c));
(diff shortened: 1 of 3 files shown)


jpechane:
@mfvitale I pushed a commit that handles the keeping of the first char as an exception not as rule.

jpechane:
@mfvitale Could you please doble check the behaviour of the schema names too? If a topic prefix has value `3prefix`, what is and what was the name of the key/value schemas of messages in the topic? Do they start with `_prefix` or with `_3prefix`?

mfvitale:
> @mfvitale Could you please doble check the behaviour of the schema names too? If a topic prefix has value `3prefix`, what is and what was the name of the key/value schemas of messages in the topic? Do they start with `_prefix` or with `_3prefix`?

Tested and the prefix will be `_3prefix` is it correct?

jpechane:
@mfvitale Yes, that's correct - it is consistent

mfvitale:
> @mfvitale Yes, that's correct - it is consistent

Yes. Let me push this other test.

jpechane:
@mfvitale Applied, thanks

PR #4613, opened 2023-06-15
DBZ-6513 Fix the computation of Oracle's negative interval

vjuranek:
https://issues.redhat.com/browse/DBZ-6513


jpechane:
@vjuranek Applied, thanks

PR #4614, opened 2023-06-16
DBZ-6416 Add initial snapshot notifications

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6416 https://issues.redhat.com/browse/DBZ-5464

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/notification/InitialSnapshotNotificationService.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/notification/InitialSnapshotNotificationService.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.pipeline.notification;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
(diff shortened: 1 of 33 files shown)


debezium-core/src/main/java/io/debezium/pipeline/notification/InitialSnapshotNotificationService.java
```diff
@@ -0,0 +1,81 @@
...
+import java.util.UUID;
+
+import io.debezium.config.CommonConnectorConfig;
+import io.debezium.pipeline.spi.OffsetContext;
+import io.debezium.pipeline.spi.Offsets;
+import io.debezium.pipeline.spi.Partition;
+import io.debezium.pipeline.spi.SnapshotResult;
+import io.debezium.spi.schema.DataCollectionId;
+
+public class InitialSnapshotNotificationService<P extends Partition, O extends OffsetContext> {
```
jpechane, line 19: Given that we are going to introduce the blocking snapshot during runtime I think we should come with a different class and aggregate name. It is OK now but we need to think about that.

jpechane:
@mfvitale Applied, great job, thanks!

debezium-core/src/main/java/io/debezium/pipeline/notification/InitialSnapshotNotificationService.java
mfvitale: Yes, I agree.

PR #4615, opened 2023-06-16
DBZ-3605 Add support for several Oracle data types

Naros:
https://issues.redhat.com/browse/DBZ-3605

Data Types:
- [x] XMLTYPE
- [x] RAW
- [ ] ~~LONG - Not likely supported due to complications with LogMiner.~~
- [ ] ~~LONG RAW - Not likely supported due to the same complications with LogMiner as LONG types.~~
- [x] Documentation
- [x] Fix dependencies on XDB and SAX environment variables


Naros:
So `LONG` (replaced by CLOB) and `LONG RAW` (replaced by BLOB) are data types provided by LogMiner in two different ways depending on the data length in those columns.  The use case where this becomes problematic is when these data types are updated with a new value.

If the length of the character data (`LONG`) or byte data (`LONG RAW`) is short enough, Oracle inlines the column values in the re-constructed SQL so the SQL statements appear as follows:

```
UPDATE table SET long_col = 'the long text' WHERE long_col = 'the old long text value';
UPDATE table SET long_raw_col = HEXTORAW('new hex value') WHERE long_raw_col = HEXTORAW('old hex value');
```

This works well, but unfortunately, once the character or binary data length reaches a specific threshold, Oracle will re-construct the SQL differently, using what is called `LONG DATA` operations (or opcode 24) event types.  The event stream becomes the following:

```
UPDATE table set long_col = 'LONG DATA' WHERE long_col = 'LONG DATA';
[a series of LONG DATA events]
```
or
```
UPDATE table set long_raw_col = 'LONG DATA' WHERE long_raw_col = 'LONG DATA';
[a series of LONG DATA events]
```

These `LONG DATA` operation events are a continuous stream of character or binary data representing both the old and new values.  For `LONG RAW`, we can only determine the breakpoint between the old/new values because the stream is the concatenation of `HEXTORAW('...')HEXTORAW('...')` where the former function call is old, and the latter is new; however when working with `LONG` data types, this distinction doesn't exist.

The only reason that `RAW` works is that it is explicitly restricted to a maximum length of 2000, which allows Oracle to inline the value, and therefore its always directly available in the re-constructed SQL without any secondary event types that require merge semantics.

We can safely support `LONG` and `LONG RAW` via Xstream because GoldenGate can differentiate the break point in the `LONG DATA` redo entries, but unfortunately LogMiner does not.  This means that the OpenLogReplicator implementation should be more than capable of doing precisely what GoldenGate does for Xstream, so if we wanted to pursue this support, it would be non-LogMiner only for these two.

Naros:
@roldanbob could you take a peak at the recent commit for the documentation changes, and suggest or add whatever changes are needed for the product side around the new XDB library dependency?

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -1415,7 +1415,7 @@ The following table describes how the connector maps binary and character large
 
 |`RAW`
 |n/a
-|_This data type is not supported._
+|Either the raw bytes (the default), a base64-encoded String, or a base64-url-safe-encoded String, or a hex-encoded String, based on the xref:oracle-property-binary-handling-mode[`binary.handling.mode`] connector configuration property setting.
```
roldanbob, line 1418:
```suggestion
|Depending on the setting of the xref:oracle-property-binary-handling-mode[`binary.handling.mode`] property in the connector configuration, the connector maps LOB values of this type to one of the following the semantic types:

* Raw bytes (the default)
* A base64-encoded string
* A base64-url-safe-encoded string
* A hex-encoded string.
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -1741,6 +1741,39 @@ endif::product[]
 
 |===
 
+[id="oracle-xml-types"]
+=== XML type
```
roldanbob, line 1745:
```suggestion

[id="oracle-xml-types"]
=== XML types
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -1741,6 +1741,39 @@ endif::product[]
...
+ifdef::community[]
+[NOTE]
+====
+Support for `XMLTYPE` is currently in incubating state, that is, the exact semantics, configuration options and so forth might change in future revisions, based on feedback we receive.
```
roldanbob, line 1749:
```suggestion
Support for `XMLTYPE` is currently in an incubating state.
Depending on the feedback that we receive, the exact semantics, configuration options, and so forth might change in future revisions.
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -1741,6 +1741,39 @@ endif::product[]
...
+Please let us know if you encounter any problems while using these data types.
+====
+endif::community[]
+ifdef::product[]
+[IMPORTANT]
+====
+Use of the `XMLTYpE` with the {prodname} Oracle connector is a Technology Preview feature only.
```
roldanbob, line 1756:
```suggestion
Use of the `XMLTYPE` with the {prodname} Oracle connector is a Technology Preview feature only.
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -2118,6 +2151,7 @@ To deploy a {prodname} Oracle connector, you install the {prodname} Oracle conne
 +
 NOTE: If you use the {prodname} Oracle connector with Oracle XStream, obtain the JDBC driver as part of the Oracle Instant Client package.
 For more information, see xref:obtaining-oracle-jdbc-driver-and-xstreams-api-files[].
+. Download the link:https://repo1.maven.org/maven2/com/oracle/database/xml/xdb/{ojdbc8-version}/xdb-{ojdbc8-version}.jar[XDB library for Oracle] from Maven Central and extract the downloaded file to the directory that contains the {prodname} Oracle connector JAR file.
```
roldanbob, line 2154: In your review request, you refer to changes that might be "needed for the product side", which I take to mean for use downstream. But this content is only in the Deployment topic that's conditionalized for community use. 

Do we need to include a reference to obtaining the XDB library in `/partials/modules/all-connectors/ref-deploy-oracle-kafka-connect-yaml.adoc` ([L10](https://github.com/debezium/debezium/blame/main/documentation/modules/ROOT/partials/modules/all-connectors/ref-deploy-oracle-kafka-connect-yaml.adoc#L10), and [L42](https://github.com/debezium/debezium/blame/main/documentation/modules/ROOT/partials/modules/all-connectors/ref-deploy-oracle-kafka-connect-yaml.adoc#L42))?

And probably a short topic in the Deployment overview in `oracle.adoc`, analogous to (and adjacent to)  _[Obtaining the Oracle JDBC driver_,](https://github.com/debezium/debezium/blame/main/documentation/modules/ROOT/pages/connectors/oracle.adoc#L2151) that explains why you might want to obtain the XDB library.

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -1741,6 +1741,39 @@ endif::product[]
...
+Red Hat does not recommend using them in production. These features provide early access to upcoming product features, enabling customers to test functionality and provide feedback during the development process.
+For more information about the support scope of Red Hat Technology Preview features, see link:https://access.redhat.com/support/offerings/techpreview[https://access.redhat.com/support/offerings/techpreview].
+====
+endif::product[]
+The following table describes how the connector maps XMLTYPE data types.
+
+.Mappings for Oracle XMLTYPE data types
+[cols="20%a,15%a,55%a",options="header"]
+|===
+|Oracle Data Type
```
roldanbob, line 1767:
```suggestion
|Oracle data type
```
We use sentence-style capitalization in table headings downstream. I know that this isn't consistently applied upstream, so we can either let it go here for now, or make the change and give me the task to do a thorough scrubbing of the doc set to apply the formatting.

documentation/modules/ROOT/pages/connectors/oracle.adoc
roldanbob: We could apply a similar change to the description on L1398.

roldanbob (changes requested):

documentation/modules/ROOT/pages/connectors/oracle.adoc
Naros: That's correct @roldanbob.  Additionally the user will need to obtain the `xmlparserv2` dependency at 
```
<dependency>
  <groupId>com.oracle.database.xml</groupId>
  <artifactId>xmlparserv2</artifactId>
  <version>21.6.0.0</version>
</dependency>
```
I also added a blurb about the `-Djavax.xml.parsers.SAXParserFactory` argument required for XML support with Oracle, so we may need to rework how to include this elsewhere when enabling XML support.

roldanbob: Agreed. A procedure module, e.g., _Enabling support for Oracle XML data types_.

As far as adding content to the downstream deployment instructions, how shall we handle that.  Want me to open a separate PR against DBZ-3605 to make those changes? Or I could wait until you merge this and then add the downstream version of the content?

jpechane:
@Naros Applied, thanks!

Naros:
@roldanbob I'll follow-up the documentation changes in a follow PR later this week.

PR #4616, opened 2023-06-19
DBZ-6449 Use debezium/kafka image in zk-less mode

ani-sha:


jpechane:
@ani-sha What is missing in our image that we cannot use `withKraft` as defined in https://java.testcontainers.org/modules/kafka/ ?

ani-sha:
@jpechane I haven't tried as it was available in 1.18.0 version, and we are using testcontainers version 1.17.6

jpechane:
@ani-sha OK, could you try to upgrade it and use if it is ok?

ani-sha:
@jpechane using `withKraft` won't be possible with our `debezium/kafka` image, as it explicitly checks for confluent's kafka image version.

jpechane:
@ani-sha Applied, thanks for checking.

PR #4621, opened 2023-06-19
DBZ-6577 Don't try to alter publication created for all tables when c…

AnatolyPopov:
…onfiguration changes to 'filtered'


jpechane (changes requested):
Requested change to connector failure, see Jira discussion

jpechane:
@AnatolyPopov Applied, thanks!

PR #4623, opened 2023-06-20
DBZ-6580 Support retries for reconnect

jpechane:
https://issues.redhat.com/browse/DBZ-6580


debezium-storage/debezium-storage-redis/src/main/java/io/debezium/storage/redis/history/RedisSchemaHistory.java
```diff
@@ -120,6 +119,19 @@ protected void storeRecord(HistoryRecord record) throws SchemaHistoryException {
         }
     }
 
+    private void reconnect() {
```
vjuranek, line 122: Won't it be better to call it `disconnect()` given that the only thing done here is disconnection the client? Same for the log message.

jpechane: @vjuranek Should be resolved via https://github.com/debezium/debezium/pull/4632

PR #4624, opened 2023-06-20
DBZ-6578 Debezium should honor read preference from connection string

jcechace:
https://issues.redhat.com/browse/DBZ-6578


jcechace:
@jpechane  this gets rid of all hardcoded read preferences so that all configuration from connection string is honored. 

A new `mongodb.connection.string.shard.params` property was introduce in order allow configuration of connection string parameters when connecting to individual shards (that is when `mongodb.connection.mode=replica_set`). I will document this in a separate PR. 

Also some smaller code cleanup around the changed code was done.

jpechane:
@jcechace Applied, thanks

PR #4627, opened 2023-06-20
DBZ-6585 Support dropping multiple Oracle partitions

vjuranek:
https://issues.redhat.com/browse/DBZ-6585

Upstream PR: https://github.com/antlr/grammars-v4/pull/3542


jpechane:
@vjuranek Applied, thanks

PR #4629, opened 2023-06-21
DBZ-6590 Support MySQL CAST AS DEC

vjuranek:
https://issues.redhat.com/browse/DBZ-6590

Upstream PR: https://github.com/antlr/grammars-v4/pull/3545

CLI example:

    > SELECT CAST('1' AS DEC);
    +------------------+
    | CAST('1' AS DEC) |
    +------------------+
    |                1 |
    +------------------+


vjuranek:
Add also support for `CAST (UN)SIGNED INT`

jpechane:
@vjuranek Applied, thanks

PR #4630, opened 2023-06-21
DBZ-6588 Skip tombstone record on HeaderToValue SMT

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6588


jpechane:
@mfvitale Applied, thanks

PR #4631, opened 2023-06-21
DBZ-6182 Allowing mongodb connector to run with database restricted privileges

jcechace:
https://issues.redhat.com/browse/DBZ-6182


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
```diff
@@ -508,6 +572,26 @@ public static ConnectionMode parse(String value, String defaultValue) {
...
+    public static final Field CAPTURE_SCOPE = Field.create("capture.scope")
+            .withDisplayName("Capture scope")
+            .withEnum(CaptureScope.class, CaptureScope.DEPLOYMENT)
+            .withGroup(Field.createGroupEntry(Field.Group.CONNECTOR_ADVANCED, 2))
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.MEDIUM)
+            .withDescription("The scope of captured changes. "
+                    + "Options include: "
+                    + "'deployment' to capture changes from the entire MongoDB deployment; "
+                    + "'database' (the default) to capture changes from a specific MongoDB database");
```
jpechane, line 584: I think `deployment` is the default

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
```diff
@@ -217,6 +217,70 @@ public boolean isIncludePreImage() {
         }
     }
 
+    /**
+     * The set of predefined CaptureScope options or aliases.
+     */
+    public enum CaptureScope implements EnumeratedValue {
+        /**
```
jpechane, line 224: Please add the impact on reuqired privileges in per-option JavaDoc

jpechane (commented):
@jcechace Overall LGTM. Just one note, there are right now only two places the branches based  on the difference between database/deployment. If more will be added then it might be good to intorduce a strategy pattern interface and classes and put the code there.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
jcechace: I don't think this will be needed elsewhere at the moment. But definitively I will keep that in mind.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/connection/MongoDbConnection.java
```diff
@@ -175,6 +174,19 @@ public List<CollectionId> collections() throws InterruptedException {
...
 
+    /**
+     * Executes the ping command (<a href="https://www.mongodb.com/docs/manual/reference/command/ping/">...</a>) using
+     * the first available database
+     *
+     * @return timestamp of the executed operation
+     */
+    public BsonTimestamp ping() throws InterruptedException {
+        return execute("ping on first available database", client -> {
+            var dbName = databaseNames().stream().findFirst().orElse("admin");
```
jcechace, line 185: This is required as system databases are filtered out of `databaseNames()` call so we might get an enpty collection. 

This won't happen when `capture.scope` is set to `database` as in such case the method return the value specified in `capture.target` propertery. However executing the ping command is possible even when the user doesn't have the read permissions for the "admin" database.

vjuranek (approved):

vjuranek:
Merged, thanks!

PR #4635, opened 2023-06-22
DBZ-6417 Provide extensible signal actions mechanism through SPI

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6417

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/signal/actions/SignalActionProvider.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/signal/actions/SignalActionProvider.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.pipeline.signal.actions;
+
+import java.util.Map;
+
+import io.debezium.config.CommonConnectorConfig;
+import io.debezium.pipeline.EventDispatcher;
+import io.debezium.pipeline.spi.Partition;
+import io.debezium.spi.schema.DataCollectionId;
+
+/**
+ * This interface is used to provide custom signal actions:
+ * Implementations must:
+ *
+ * providea map of signal action in the {@link #createActions(EventDispatcher, CommonConnectorConfig)} method.
+ *
+ * @author Mario Fiore Vitale
+ */
+public interface SignalActionProvider {
+
+    /**
+     * Create a map of signal action where the key is the name of the action.
+     *
+     * @param dispatcher the event dispatcher instance
+     * @param connectorConfig the connector config
+     * @return a concrete action
+     */
+
+    <P extends Partition> Map<String, SignalAction<P>> createActions(EventDispatcher<P, ? extends DataCollectionId> dispatcher, CommonConnectorConfig connectorConfig);
+}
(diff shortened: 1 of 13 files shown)


debezium-core/src/main/java/io/debezium/pipeline/signal/actions/SignalActionProvider.java
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * This interface is used to provide custom signal actions:
+ * Implementations must:
+ * define the name of the action in {@link #getName()},
+ * provide the implementation of the action in the {@link #createAction(EventDispatcher, CommonConnectorConfig)} method.
+ *
+ * @author Mario Fiore Vitale
+ */
+public interface SignalActionProvider {
```
jpechane, line 21: @mfvitale Why do we need provider classes? Can't implementations of class `Action` be service loaded and the `getName()` method added to them? I don't see an advatage of using indirect apprach in this case.

jpechane: OTOH, yes let's make it consistent with other providers we use. Still do we need per-action provider? How about having a provider that will look like

```java
interface SignalActionProvider {

    // returns map of name of signal, action
    Map<String, <P extends Partition> SignalAction<P>> createActions(EventDispatcher<P, ? extends DataCollectionId> dispatcher, CommonConnectorConfig connectorConfig);
}
```

mfvitale: The actions requires the dispatcher instance, so in this case direct instantiation with ServiceLoader is not possible. 

> @mfvitale Why do we need provider classes? Can't implementations of class `Action` be service loaded and the `getName()` method added to them? I don't see an advatage of using indirect apprach in this case.

The actions requires the dispatcher instance, so in this case direct instantiation with ServiceLoader is not possible.

mfvitale: > OTOH, yes let's make it consistent with other providers we use. Still do we need per-action provider? How about having a provider that will look like
> 
> ```java
> interface SignalActionProvider {
> 
>     // returns map of name of signal, action
>     Map<String, <P extends Partition> SignalAction<P>> createActions(EventDispatcher<P, ? extends DataCollectionId> dispatcher, CommonConnectorConfig connectorConfig);
> }
> ```

I think it can be a good alternative even in case of user extension seem a good interface.

jpechane: @mfvitale Yes, please do.

jpechane: > The actions requires the dispatcher instance, so in this case direct instantiation with ServiceLoader is not possible.
> 
> > @mfvitale Why do we need provider classes? Can't implementations of class `Action` be service loaded and the `getName()` method added to them? I don't see an advatage of using indirect apprach in this case.
> 
> The actions requires the dispatcher instance, so in this case direct instantiation with ServiceLoader is not possible.

This would be solvable by executing late initializtion - in the same way as we use `configure` methods. Still it is unnecessary as the other option I talked about seems like best of both worlds.

jpechane:
@mfvitale Perfect, I really like it this way. Please add a test and modify non-core connectors so they can compile and I belive we are good to go!

mfvitale:
> @mfvitale Perfect, I really like it this way. Please add a test and modify non-core connectors so they can compile and I belive we are good to go!

Test that covers this parts already present: `SignalIT`and `IncrementalSnapshotIT`. Don't think we need more. 

Db2 changes: https://github.com/debezium/debezium-connector-db2/pull/112

jpechane:
@mfvitale I mean test with custom signal.

mfvitale:
> @mfvitale I mean test with custom signal.

Added.

jpechane:
@mfvitale Applied, thanks! Could you please add a docs for the feature in a separate PR? Thanks

mfvitale:
> @mfvitale Applied, thanks! Could you please add a docs for the feature in a separate PR? Thanks

Doc PR https://github.com/debezium/debezium/pull/4640

PR #4637, opened 2023-06-23
DBZ-6605 Fix DataCollections for snapshot completion notification

bdbene:
https://issues.redhat.com/browse/DBZ-6605

This is a fix to see only the table name for which the table scan has completed. The notification should be per table so that the it can be acted on correctly.

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/notification/IncrementalSnapshotNotificationService.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/notification/IncrementalSnapshotNotificationService.java
@@ -23,6 +23,7 @@ public class IncrementalSnapshotNotificationService<P extends Partition, O exten
 
     public static final String INCREMENTAL_SNAPSHOT = "Incremental Snapshot";
     public static final String DATA_COLLECTIONS = "data_collections";
+    public static final String SCANNED_COLLECTION = "scanned_collection";
     public static final String CURRENT_COLLECTION_IN_PROGRESS = "current_collection_in_progress";
     public static final String MAXIMUM_KEY = "maximum_key";
     public static final String LAST_PROCESSED_KEY = "last_processed_key";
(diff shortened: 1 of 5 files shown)


jpechane:
@mfvitale Could you please review the PR? The proposal seems logical to me.

debezium-core/src/main/java/io/debezium/pipeline/notification/IncrementalSnapshotNotificationService.java
```diff
@@ -99,9 +99,7 @@ public <T extends DataCollectionId> void notifyAborted(IncrementalSnapshotContex
     public <T extends DataCollectionId> void notifyTableScanCompleted(IncrementalSnapshotContext<T> incrementalSnapshotContext, P partition, OffsetContext offsetContext,
                                                                       long totalRowsScanned, TableScanCompletionStatus status) {
 
-        String dataCollections = incrementalSnapshotContext.getDataCollections().stream().map(DataCollection::getId)
-                .map(DataCollectionId::identifier)
-                .collect(Collectors.joining(LIST_DELIMITER));
```
mfvitale, line 104: What do you think to add a new field (maybe `SCANNED_COLLECTION`) as for in progress notification (`CURRENT_COLLECTION_IN_PROGRESS`) and leave `DATA_COLLECTIONS` unchanged?

jpechane: Sounds good!

mfvitale:
@bdbene The proposal is good! Please

- [ ] address the suggested change
- [ ] change documentation accordingly (https://github.com/debezium/debezium/blob/fa78e777977e49c8f54260609099b3ccb1ed352e/documentation/modules/ROOT/pages/configuration/notification.adoc?plain=1#L151)

debezium-core/src/main/java/io/debezium/pipeline/notification/IncrementalSnapshotNotificationService.java
bdbene: Thanks for the quick feedback! I'll make the changes.

bdbene:
I've made the changes; added a new field to indicate for which table the snapshot has completed. And I've updated the docs to indicate that the existing field is a list of the remaining tables, and the new field has the completed table.

mfvitale (approved):

mfvitale:
@bdbene good work! Thank you.

PR #4643, opened 2023-06-27
DBZ-6610 Fix retrying nested ORA-01327 exception

Naros:
https://issues.redhat.com/browse/DBZ-6610


vjuranek:
Applied to main and 2.3 branches. Thanks!

PR #4655, opened 2023-07-04
DBZ-6632 Improve properties description on JDBC connector documentation

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6632

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -521,7 +521,7 @@ For more information, see the xref:jdbc-property-primary-key-mode[primary.key.mo
 <10> Enables the connector to evolve the destination database's schema.
 For more information, see the xref:jdbc-property-schema-evolution[schema.evolution] configuration property.
 <11> Specifies the timezone used when writing temporal field types.
-<12> List of topics to consume, separated by commas
+<12> List of topics to consume, separated by commas.
 
 For a complete list of configuration properties that you can set for the {prodname} JDBC connector, see xref:jdbc-connector-properties[JDBC connector properties].
 
@@ -552,11 +552,14 @@ Information about the properties is organized as follows:
 
 |[[jdbc-property-connection-name]]<<jdbc-property-connection-name, `+name+`>>
 |No default
-|Unique name for the connector. Attempting to register again with the same name will fail. This property is required by all Kafka Connect connectors.
+|Unique name for the connector. 
+A failure results if you attempt to reuse this name when registering a connector.   
+This property is required by all Kafka Connect connectors.
 
 |[[jdbc-property-connection-class]]<<jdbc-property-connection-class, `+connector.class+`>>
 |No default
-|The name of the Java class for the connector. Always use a value of `io.debezium.connector.jdbc.JdbcSinkConnector` for the JDBC connector.
+|The name of the Java class for the connector. 
+For the {prodname} JDBC connector, specify the value `io.debezium.connector.jdbc.JdbcSinkConnector`.
 
 |[[jdbc-property-connection-task]]<<jdbc-property-connection-task, `+tasks.max+`>>
 |1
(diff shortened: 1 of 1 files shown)


documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -537,10 +539,40 @@ The {prodname} JDBC sink connector has several configuration properties that you
...
+|===
+|Property |Default |Description
+
+|[[jdbc-property-connection-name]]<<jdbc-property-connection-name, `+name+`>>
+|No default
+|Unique name for the connector. Attempting to register again with the same name will fail. This property is required by all Kafka Connect connectors.
+
+|[[jdbc-property-connection-class]]<<jdbc-property-connection-class, `+connector.class+`>>
+|No default
+|The name of the Java class for the connector. Always use a value of `io.debezium.connector.jdbc.JdbcSinkConnector` for the JDBC connector.
```
roldanbob, line 559:
```suggestion
|The name of the Java class for the connector. 
For the {prodname} JDBC connector, specify the value `io.debezium.connector.jdbc.JdbcSinkConnector`.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -537,10 +539,40 @@ The {prodname} JDBC sink connector has several configuration properties that you
...
+
+|[[jdbc-property-connection-task]]<<jdbc-property-connection-task, `+tasks.max+`>>
+|1
+|Maximum number of tasks to use for this connector.
+
+|[[jdbc-property-connection-topics]]<<jdbc-property-connection-topics, `+topics+`>>
+|No default
+|List of topics to consume, separated by commas.
```
roldanbob, line 567:
```suggestion
|List of topics to consume, separated by commas.
Do not use this property in combination with the xref:jdbc-property-connection-topics-regex[`topics.regex`] property.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -520,6 +521,7 @@ For more information, see the xref:jdbc-property-primary-key-mode[primary.key.mo
 <10> Enables the connector to evolve the destination database's schema.
 For more information, see the xref:jdbc-property-schema-evolution[schema.evolution] configuration property.
 <11> Specifies the timezone used when writing temporal field types.
+<12> List of topics to consume, separated by commas
```
roldanbob, line 524:
```suggestion
<12> List of topics to consume, separated by commas.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -537,10 +539,40 @@ The {prodname} JDBC sink connector has several configuration properties that you
 Many properties have default values.
 Information about the properties is organized as follows:
 
+* xref:jdbc-connector-properties-generic[]
 * xref:jdbc-connector-properties-connection[]
 * xref:jdbc-connector-properties-runtime[]
 * xref:jdbc-connector-properties-extendable[]
```
roldanbob, line 546: I missed this earlier, but these link lists that appear in the documentation for the other connectors are designed for downstream use (conditionalized as `product`]) , where the specified IDs match anchors applied to the downstream version of the docs. I forget now if and when we're planning to expose the JDBC connector downstream, but if we do, we'll have to make sure that we retain the upstream anchor IDs vs. using the `ModuleID` comment to modify the IDs for downstream. Otherwise the empty link text causes problems.
 Just a comment for now ... no action required.

roldanbob (changes requested):
Hi @mfvitale A few minor edits, otherwise LGTM.
I also made one comment about how we generally use the "mini TOCs" at the beginning of a section, but as I state, it's just something to note; there's no need to change anything related to that comment.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
Naros: We likely will at some future point, I can't say when @roldanbob, so if it makes less work for later to fix this to be compliant for what we may need, I'm all for it.

mfvitale: No issue! I'll merge.

PR #4656, opened 2023-07-04
DBZ-3925 Add support for FLUSH FIREWALL_RULES 

mfvitale:
Special syntax for Azure Database for MySQL Single Server instance

closes: https://issues.redhat.com/browse/DBZ-3925


mfvitale:
I don't think we want to port this back to upstream since it's a very specific case that seems will be dropped with further version as reported in the issue.

ani-sha (approved):

Naros:
> Special syntax for Azure Database for MySQL Single Server instance
> 
> closes: [issues.redhat.com/browse/DBZ-3925](https://issues.redhat.com/browse/DBZ-3925)

That seems reasonable to me, great work and thanks for the quick review @ani-sha.

Naros:
Applyied to main (=2.4)

PR #4661, opened 2023-07-07
DBZ-6640 Keep exception details when a SQLException is caught whilst reading the maximum key for a table

fredlegaucher:
In  AbstractIncrementalSnapshotChangeEventSource#readChunk, the signature of org.sl4j.logger.error currently used for
`LOGGER.error("Failed to read maximum key for table {}", currentTableId, e);` 
is:
`void error(String var1, Object var2, Object var3);`

As there are no placeholders for the 2nd variable, that 2nd variable is unused. Here the 2nd variable is the exception. As a result the exception details are lost.

This PR switches to the following signature:
`void error(String var1, Throwable var2);`


fredlegaucher:
A colleague just pointed out that snippet from [Baeldung](https://www.baeldung.com/slf4j-log-exceptions):
> In presence of multiple parameters, if the last argument in a logging statement is an exception, then SLF4J will presume that the user wants the last argument to be treated as an exception instead of a simple parameter

And indeed the SLF4J documentation indicates - https://www.slf4j.org/faq.html#paramException 

I will close!

PR #4663, opened 2023-07-07
DBZ-6628 Correct formatting in `additional-condition` examples

roldanbob:
[DBZ-6628](https://issues.redhat.com/browse/DBZ-6628)

Updates formatting of string values in `additional-condition` examples.
The original examples incorrectly omitted single quotes around string values for the parameter.

This bug fix should be backported to 2.3


jpechane:
@roldanbob Applied, thanks

PR #4669, opened 2023-07-12
DBZ-6651 Avoid confusing label and IPv6 address

vjuranek:
Some label combined with some keywords (when written without a space in between, which is valid for MySQL) can look like beginning of an IPv6 address, e.g.

    aaa:BEGIN

is currectly prased as IPv6 address `aaa:be` and leads to a parsing failure.

Make parsing of IPv6 address more robust and require at least two `:` to make sure there is no confusion with labels.

Upstream PR: https://github.com/antlr/grammars-v4/pull/3606


mfvitale (approved):

jpechane:
@vjuranek Applied, thanks

PR #4670, opened 2023-07-13
DBZ-6650 Avoid ConcurrentModificationException during dispatch

Naros:
https://issues.redhat.com/browse/DBZ-6650


jpechane:
@Naros Applied, thanks

PR #4671, opened 2023-07-13
DBZ-6659 Fixing junit dependencies

jcechace:
https://issues.redhat.com/browse/DBZ-6659

A proper test would be to switch to junit vintage and get rid of these issue once and for all. 

For what it's worth I tried running the test and replacing `junit:junit` with `org.junit.vintage:junit-vintage-engine` seems to be enough. I am opened to try replacing it in all modules and seeing what happens.

### DIFF
--- a/debezium-bom/pom.xml
+++ b/debezium-bom/pom.xml
@@ -51,6 +51,8 @@
 
         <!-- Testing -->
         <version.junit>4.13.1</version.junit>
+        <version.junit5>5.7.2</version.junit5>
+        <version.junit.platform.launcher>1.7.2</version.junit.platform.launcher>
         <version.fest>1.4</version.fest>
         <version.assertj>3.11.1</version.assertj>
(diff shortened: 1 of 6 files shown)


vjuranek (approved):

jpechane:
@jcechace Applied, thanks.
WRT your proposal, we are early in 2.4 cycle so it is either now or wait for 2.5. I'll let you to decide when you want to handle it.
Also if we follow this path it is important also to check if there is any impact on IDEs - at least Eclipse and IDEA.

jcechace:
@jpechane  IDE's are fine. This doesn't concern them really - the launcher is the same for the entire jupiter platform. Since you can run Junit5 tests just fine you can run junit4 too if you have vintage engine available. 

I will give this a spin on Monday. It should be fairly easy switch (the long part is rerunning all the tests and check if everything works).

PR #4673, opened 2023-07-13
DBZ-6648 Use String fallback when interval handling mode is String

vjuranek:
The root cause of DBZ-6648 is using long number as a fallback value, while String is expected and thus falling conversion and returning `null`.

Use String fallback when interval handling mode is String.

https://issues.redhat.com/browse/DBZ-6648


jpechane:
@vjuranek Applied, thanks

PR #4677, opened 2023-07-14
DBZ-6660 Introduce internal.log.mining.max.scn.deviation.ms`

Naros:
https://issues.redhat.com/browse/DBZ-6660

I also added a small inclusion of the redo thread id in the abandoned transaction log message based on feedback from Oren that could help identify which RAC node is predominately seeing abandoned transactions on RAC.


Naros:
It seems Db2 errors are unrelated @jpechane.

zalmane:
Looking at the RAC Flush strategy, will we get the same effect if we were to use the Metronome pause there (as happens if, for example, the flush itself fails)? Since we are interested in delaying the mining for X seconds until the flush happens?

jpechane:
@Naros Applied, thanks

PR #4680, opened 2023-07-17
DBZ-6677 Support altering an Oracle table's column visibility

Naros:
https://issues.redhat.com/browse/DBZ-6677


jpechane:
@Naros Applied, thanks

PR #4681, opened 2023-07-17
DBZ-6676 Oracle XML/RAW documentation updates

Naros:
https://issues.redhat.com/browse/DBZ-6676

@roldanbob This is a follow-up from DBZ-3605 (PR #4615).


Naros:
@roldanbob I addressed everything but this one discussion referenced here:
https://github.com/debezium/debezium/pull/4615#discussion_r1245794743

roldanbob (approved):
Hey, @Naros. I started to correct base64 => Base64, but I think that in the context it's correct to use lc.  Other than that, I think that this is GTG. Thanks!

PR #4682, opened 2023-07-17
DBZ-6673 Fix flaky Oracle test

Naros:
https://issues.redhat.com/browse/DBZ-6673

The shouldCaptureChangesForTransactionsAcrossSnapshotBoundaryWithoutReemittingDDLChanges test only expects the tables created by the entire test to exist but tables from other tests not, and it would appear this commonly happens when another test fails to cleanup after itself.

This fix is to guarantee that the Oracle database state is set properly so that tests from within this class are executed with the right number of tables expected to exist.


jpechane:
@Naros Applied, thanks

PR #4689, opened 2023-07-21
DBZ-6669 Snapshot will not capture data when signal.data.collection is present without table.include.list

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6669

### DIFF
--- a/debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
+++ b/debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
@@ -74,6 +74,7 @@ public abstract class RelationalSnapshotChangeEventSource<P extends Partition, O
     private static final Logger LOGGER = LoggerFactory.getLogger(RelationalSnapshotChangeEventSource.class);
 
     public static final Pattern SELECT_ALL_PATTERN = Pattern.compile("\\*");
+    public static final Pattern MATCH_ALL_PATTERN = Pattern.compile(".*");
 
     private final RelationalDatabaseConnectorConfig connectorConfig;
     private final JdbcConnection jdbcConnection;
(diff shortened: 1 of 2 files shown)


debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
```diff
@@ -241,13 +242,16 @@ private Stream<TableId> toTableIds(Set<TableId> tableIds, Pattern pattern) {
...
         String tableIncludeList = connectorConfig.tableIncludeList();
         String signalingDataCollection = connectorConfig.getSignalingDataCollectionId();
         List<Pattern> captureTablePatterns = new ArrayList<>();
         if (!Strings.isNullOrBlank(tableIncludeList)) {
             captureTablePatterns.addAll(Strings.listOfRegex(tableIncludeList, Pattern.CASE_INSENSITIVE));
         }
+        else {
+            captureTablePatterns.add(MATCH_ALL_PATTERN);
+        }
         if (!Strings.isNullOrBlank(signalingDataCollection)) {
```
Naros, line 255: So this larger block:
```
        if (!Strings.isNullOrBlank(tableIncludeList)) {
            captureTablePatterns.addAll(Strings.listOfRegex(tableIncludeList, Pattern.CASE_INSENSITIVE));
        }
        else {
            captureTablePatterns.add(MATCH_ALL_PATTERN);
        }
        if (!Strings.isNullOrBlank(signalingDataCollection)) {
            captureTablePatterns.addAll(getSignalDataCollectionPattern(signalingDataCollection));
        }
        if (captureTablePatterns.size() > 0) {
            return captureTablePatterns
                    .stream()
                    .flatMap(pattern -> toTableIds(capturedTables, pattern))
                    .collect(Collectors.toCollection(LinkedHashSet::new));
        }
```
I believe this larger code will always go into the `captureTablePatterns.size() > 0` branch now, right?

Naros (commented):
Just one comment inline @mfvitale.

debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
mfvitale: Yes, good point.

Naros (approved):

Naros:
Applied to main (=2.4), thanks @mfvitale.

PR #4691, opened 2023-07-23
DBZ-4843 Add large schema optimization option

ajdub508:


ajdub508:
I welcome any feedback but, in particular, wondering if we want to have a configuration option for this feature. I included one but wanted to wait to confirm that this change should be optional before adding all the integration tests that would be required for the option.

For what it is worth, I am running a version of the connector (based on v1.8.1) in production for my organization where it is not optional, and the integration tests pass with the change in place.

In my opinion and as far as I am aware, it's possible someone might choose to not use this option if they are confident that each polling cycle will include changes for a very large percentage of the whitelisted tables. Otherwise, the option will result in far less load on source DB and server.

jpechane:
@ajdub508 Hi, thanks for the PR. Generally it looks good to me and I feel it is something that should be used by default.
Would it be possible to make the optimization default behaviour and rename the config option to something like `query.all.tables` that is disabled by default and have description in docs saying when it is good to enable it?

ajdub508:
Thanks @jpechane - Good call, I will make those changes. I will also fix the issue I introduced when I rebased - repeating the max retries config. Thanks!

ajdub508:
@jpechane - Suggested changes have been made

ajdub508:
@jpechane - Is it worth kicking off the SQL Server build again? One of the checks failed but some have seen some transient failures of tests in the past.

mwbayley:
Hi @ajdub508, I'm curious what this change looked like and why it never made it in?

PR #4692, opened 2023-07-24
[jenkins-jobs] Disable arm64 build for UI due to build failure

jpechane:


jm-nab:
I was looking into why this would be, and found these, for anyone else wondering why arm64 isn't available on 2.5 quay/dockerhub, this might help provide some context:

- https://github.com/debezium/container-images/pull/302
- https://github.com/debezium/container-images/pull/301
- https://issues.redhat.com/browse/DBZ-5213

PR #4696, opened 2023-07-25
Refactor EmbeddedEngine::run method

vjuranek:
https://issues.redhat.com/browse/DBZ-6715


mfvitale (approved):

jpechane (approved):
@vjuranek Nice job! Could you take it one more step further and extract the poll logic into a separate method too so the run will be mostly flow of methods/states?

Feel free to merge it yourself after that.

PR #4698, opened 2023-07-25
DBZ-6716 Support Oracle `DROP USER` statements

Naros:
https://issues.redhat.com/browse/DBZ-6716

Compared with upstream grammar, rule already exists.


Naros:

PR #4700, opened 2023-07-26
DBZ-6655 Document not supported schema change records on JDBC connector

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6655

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -53,7 +53,8 @@ When a delete event is consumed by the sink connector, no part of the event's pa
 
 [IMPORTANT]
 ====
-This connector is not meant to process `schema change` records. If your source connector is configured to capture schema changes then configure `topics` or `topics.regex` JDBC connector properties to exclude `schema change` topic.
+The {prodname} JDBC sink connector is not designed to read from schema change topics. 
+If your source connector is configured to capture schema changes, in the JDBC connector configuration, set the xref:jdbc-property-connection-topics[`topics`] or xref:jdbc-property-connection-topics-regex[`topics.regex`] properties so that the connector does not consume from schema change topics.
 ====
 
 [[jdbc-at-least-once-delivery]]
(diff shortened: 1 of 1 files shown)


roldanbob (changes requested):
Hi @mfvitale, I made some preliminary suggestions, but I was confused about a couple of things, so before you implement any changes, please take a look at those. After you have a chance to respond I can take take another pass. Thanks!

mfvitale:
> You refer to excluding schema change topics. Do you mean excluding schema change event records? I made a change assuming that you intended to refer to records vs. topics.

No, I mean schema change topic.

> It's not clear what you mean when you refer to configuring JDBC connector properties to exclude schema change topics (or event records). Pending discussion with you, for now I left the reference to JDBC connector properties, but I am not sure that is correct.
The Debezium JDBC connector doc doesn't refer to either of the properties that you mention. If possible, we should provide a link to the properties that you mention, or to another document that describes how to configure these properties.
I know that Debezium provides SMTs that include a topic.regex option (no s) for [topic routing](https://debezium.io/documentation/reference/stable/transformations/topic-routing.html#by-logical-table-router-topic-regex) and [message filtering](https://debezium.io/documentation/reference/stable/transformations/filtering.html#filter-topic-regex), but I don't find a topics.regex or topic property in any of our connector documentation.

The properties are generic and are default for every Kafka sink connector https://debezium.io/documentation/reference/nightly/connectors/jdbc.html#jdbc-connector-properties. The aim of this documentation change is to state that the schema changes topic is not supported by this sink connector and, in case you enabled `include.schema.changes` in the source connector, this topic must not be included in the topics consumed by the JDBC connector. This can be achieved using the `topics` or `topics.regex` properties. 

Let me know if it's more clear.

roldanbob:
> Let me know if it's more clear.

Thanks, @mfvitale! Yes, having the link to the generic Kafka connector properties clarifies things.

roldanbob (changes requested):
@mfvitale OK, I've reworked my suggestions based on your input. Thanks!

Naros (approved):

Naros:

PR #4705, opened 2023-07-27
DBZ-6213 Adjust connector restart

vjuranek:
* Try to start the connector in BaseSourceTask::start(). This is the root cause of DBZ-6213 as EmbeddedEngine calls start() and assumes it really tries to start the connector.
* Simplify the code by removing unnecessary checks and restartDelay.hasElapsed() call.

https://issues.redhat.com/browse/DBZ-6213

### DIFF
--- a/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
+++ b/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
@@ -138,6 +138,16 @@ public abstract class BaseSourceTask<P extends Partition, O extends OffsetContex
                     LOGGER.info("   {} = {}", propName, propValue);
                 });
             }
+            try {
+                this.coordinator = start(config);
+                setTaskState(State.RUNNING);
+            }
+            catch (RetriableException e) {
+                LOGGER.warn("Failed to start connector, will re-attempt during polling.", e);
+                restartDelay = ElapsedTimeStrategy.constant(Clock.system(), retriableRestartWait);
+                restartDelay.hasElapsed();
+                setTaskState(State.RESTARTING);
+            }
         }
         finally {
             stateLock.unlock();
@@ -166,14 +176,8 @@ public abstract class BaseSourceTask<P extends Partition, O extends OffsetContex
     public final List<SourceRecord> poll() throws InterruptedException {
 
         try {
-            boolean started = startIfNeededAndPossible();
-
-            // in backoff period after a retriable exception
-            if (!started) {
-                // WorkerSourceTask calls us immediately after we return the empty list.
-                // This turns into a throttling so we need to make a pause before we return
-                // the control back.
-                Metronome.parker(Duration.of(2, ChronoUnit.SECONDS), Clock.SYSTEM).pause();
+            // in we fail to start, return empty list and try to start next poll() method call
+            if (!startIfNeededAndPossible()) {
                 return Collections.emptyList();
             }
 
@@ -233,10 +237,10 @@ public abstract class BaseSourceTask<P extends Partition, O extends OffsetContex
      * Starts this connector in case it has been stopped after a retriable error,
      * and the backoff period has passed.
      */
-    private boolean startIfNeededAndPossible() {
+    private boolean startIfNeededAndPossible() throws InterruptedException {
         stateLock.lock();
 
-        boolean result;
+        boolean result = false;
         try {
             State currentState = getTaskState();
             if (currentState == State.RUNNING) {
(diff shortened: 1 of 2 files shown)


debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -138,6 +138,20 @@ public final void start(Map<String, String> props) {
                     LOGGER.info("   {} = {}", propName, propValue);
                 });
             }
+            try {
+                this.coordinator = start(config);
+                setTaskState(State.RUNNING);
+            }
+            catch (Exception e) {
+                if (e instanceof RetriableException) {
```
jeremy-l-ford, line 146: Could just catch the RetriableException and avoid the instanceof check.

jeremy-l-ford (approved):

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
vjuranek: right, thanks for noting it! Fixed.

jeremy-l-ford (approved):

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -248,29 +251,15 @@ else if (currentState == State.RESTARTING) {
...
                     this.coordinator = start(config);
                     LOGGER.info("Successfully restarted task");
+                    restartDelay = null;
+                    setTaskState(State.RUNNING);
                     result = true;
                 }
                 else {
                     LOGGER.info("Awaiting end of restart backoff period after a retriable error");
-                    result = false;
+                    Metronome.parker(Duration.of(retriableRestartWait.toMillis(), ChronoUnit.MILLIS), Clock.SYSTEM).pause();
```
jpechane, line 260: Aren't you converting `Duration` to `Duration` here?

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -312,7 +301,6 @@ private void stop(boolean restart) {
                 setTaskState(State.RESTARTING);
                 if (restartDelay == null) {
                     restartDelay = ElapsedTimeStrategy.constant(Clock.system(), retriableRestartWait.toMillis());
-                    restartDelay.hasElapsed();
```
jpechane, line 315: IMHO by removing it you ae removing the pause.

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -138,6 +138,15 @@ public final void start(Map<String, String> props) {
                     LOGGER.info("   {} = {}", propName, propValue);
                 });
             }
+            try {
+                this.coordinator = start(config);
+                setTaskState(State.RUNNING);
+            }
+            catch (RetriableException e) {
+                LOGGER.warn("Failed to start connector, will re-attempt during polling.", e);
+                restartDelay = ElapsedTimeStrategy.constant(Clock.system(), retriableRestartWait.toMillis());
```
jpechane, line 147: I think there is a method that takes Duration as parameter.

jpechane (changes requested):
@vjuranek LGTM. I left few minor comments.

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
vjuranek: right, fixed

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
vjuranek: ops, yes:-( Fixed.

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
vjuranek: Right, but only the first one pause:-) [This method](https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/util/ElapsedTimeStrategy.java#L50) says, if the delay interval has already elapsed or not. As a side effect it recomputes next timestamp for waiting. This method is very confusing (it's description, side effects and initialization) and IMHO should be also refactored.

However, you are right that removing it is wrong - due to its side effect if this call is removed, the elapsed interval won't be initialized and actually won't do any wait when the strategy is used for the first time. So basically this has to be called every time right after `ElapsedTimeStrategy` is created.

vjuranek (commented):
@jpechane all the issues should be fixed now, thanks for spotting them, especially the one with removing `restartDelay.hasElapsed()`.

jpechane (approved):
Waiting for CI to complete before merging.

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
jpechane: @vjuranek I agree on the refactoring

jpechane:
@vjuranek Applied, thanks a lot

PR #4713, opened 2023-07-28
DBZ-6727 use correct config for driver

gurpiarbassi:


debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlConnectorIT.java
```diff
@@ -121,6 +123,24 @@ public void shouldNotStartWithInvalidConfiguration() {
...
+    @Test
+    public void shouldNotStartWithUnknownJdbcDriver() {
+        config = DATABASE.defaultConfig()
+                .with(MySqlConnectorConfig.JDBC_DRIVER, "foo.bar")
+                .build();
+
+        final AtomicBoolean successResult = new AtomicBoolean();
+        final AtomicReference<String> message = new AtomicReference<>();
+        start(MySqlConnector.class, config, (success, msg, error) -> {
+            successResult.set(success);
```
gurpiarbassi, line 135: Tested using Atomic references here otherwise the test still passes regardless of the assertions within the callback failing. There is another test in this class `shouldNotStartWithInvalidConfiguration` that needs the same treatment. It asserts `assertThat(error).isNotNull();` however this assertion actually fails but the test still passes.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnection.java
```diff
@@ -532,7 +532,9 @@ public MySqlConnectionConfiguration(Configuration config) {
                     .without("queryInterceptors");
 
             this.jdbcConfig = JdbcConfiguration.adapt(jdbcConfigBuilder.build());
-            String driverClassName = this.jdbcConfig.getString(MySqlConnectorConfig.JDBC_DRIVER);
+            // The jdbcConfig has unprefixed values so you have to access it without the prefix.
```
jpechane, line 535: @gurpiarbassi You should be able to obtain the driver class from `config` variable instead of hard-coding the `jdbc.driver` string.

gurpiarbassi: @jpechane thanks, I've corrected this.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
```diff
@@ -624,6 +624,15 @@ public static SecureConnectionMode parse(String value, String defaultValue) {
             .withValidation(Field::isClassName)
             .withDescription("JDBC Driver class name used to connect to the MySQL database server.");
 
+    public static final Field JDBC_PROTOCOL = Field.create(DATABASE_CONFIG_PREFIX + "protocol")
+            .withDisplayName("Jdbc Protocol")
+            .withType(Type.STRING)
+            .withGroup(Field.createGroupEntry(Field.Group.CONNECTION, 41))
+            .withWidth(Width.MEDIUM)
```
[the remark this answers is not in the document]
gurpiarbassi: also, how to decide on position In group?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
```diff
@@ -624,6 +624,15 @@ public static SecureConnectionMode parse(String value, String defaultValue) {
             .withValidation(Field::isClassName)
             .withDescription("JDBC Driver class name used to connect to the MySQL database server.");
 
+    public static final Field JDBC_PROTOCOL = Field.create(DATABASE_CONFIG_PREFIX + "protocol")
+            .withDisplayName("Jdbc Protocol")
```
jpechane, line 628:
```suggestion
            .withDisplayName("JDBC Protocol")
```

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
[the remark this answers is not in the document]
jpechane: importance is low, width is medium. For the position, please add it as the last one.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
gurpiarbassi: done

gurpiarbassi:
@jpechane can we target this for 2.3 as well as main?

jpechane:
@gurpiarbassi Applied thanks! The 2.3 branch is the stable one and we try to limit backports to the bugfixes only so I'd prefer to keep it on the main only.

PR #4718, opened 2023-07-31
DBZ-6730 Exit consumption loop if engine has stopped

Naros:
https://issues.redhat.com/browse/DBZ-6730


mfvitale (approved):

jpechane:
@Naros Applied, thanks

PR #4720, opened 2023-07-31
DBZ-6733 Skip mining distance calc. if upper bounds not within distance

Naros:
https://issues.redhat.com/browse/DBZ-6733


jpechane:
@Naros Applied, thanks

PR #4728, opened 2023-08-01
Dbz 6704 config properties inconsistently classified as advanced vs required

roldanbob:
[DBZ-6704](https://issues.redhat.com/browse/DBZ-6704)

For consistency, this change moves properties in the MongoDB and MySQL connector documentation from the Required connector properties table to the Advanced connector properties table. 

In addressing this change it became apparent that the MongoDB connector documentation does not reference the `converters` configuration property, but the Custom converters documentation does not explicitly state that you cannot create custom converters for MongoDB. 

To complete this work, one of the following further updates is required, depending on whether the content about custom converters applies to  MongoDB:

~~* Add the `converters` property to the table of Advanced MongoDB connector configuration properties.~~
* Add a note to the Custom converters documentation, to explicitly state that you cannot use custom converters with the MongoDB connector.  **DONE**

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/mongodb.adoc
+++ b/documentation/modules/ROOT/pages/connectors/mongodb.adoc
@@ -1582,17 +1582,6 @@ To match the name of a namespace, {prodname} applies the regular expression that
 That is, the specified expression is matched against the entire name string of the namespace; it does not match substrings that might be present in a database name. +
 If you include this property in the configuration, do not set the `collection.include.list` property.
 
-|[[mongodb-property-snapshot-mode]]<<mongodb-property-snapshot-mode, `+snapshot.mode+`>>
-|`initial`
-|Specifies the criteria for performing a snapshot when the connector starts.
-Set the property to one of the following values:
-
-`initial`::
-When the connector starts, if it does not detect a value in its offsets topic, it performs a snapshot of the database.
-
-`never`::
-When the connector starts, it skips the snapshot process and immediately begins to stream change events for operations that the database records to the oplog.
-
 |[[mongodb-property-capture-mode]]<<mongodb-property-capture-mode, `+capture.mode+`>>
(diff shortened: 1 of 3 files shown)


Naros:
Hi @roldanbob, AFAIK the `CustomConverter` contract and implementation is explicitly used solely in the relational-based connectors as the hook is inside `TableSchemaBuilder`, which isn't used by MongoDB.  

So I think it's safe to say that there is no need to make references to it in the MongoDB documentation and it makes sense to add a disclaimer about it not being applicable to MongoDB.

That said, I did notice that registration of such converters is part of the `CommonConnectorConfig`, @jpechane in light of this, should we consider moving this to the `RelationalConnectorConfig` instead to add clarity to which connectors it applies to?

roldanbob:
> ... I think it's safe to say that there is no need to make references to it in the MongoDB documentation and it makes sense to add a disclaimer about it not being applicable to MongoDB.

@Naros OK, I went ahead and added a note to the Converters doc to indicate that the content does not apply to use with the MongoDB connector.

Naros (commented):
So far this all looks good, so all that remains are the Oracle changes I believe based on the comments in the Jira  I'll add a comment on the Jira as I have an idea with regard to the Oracle property splits.

roldanbob:
> ... I'll add a comment on the Jira as I have an idea with regard to the Oracle property splits.

@Naros  Let me know if you have input wrt how to classify the Oracle properties as Required/Advanced, or if I should just do a pass for you to review.

Naros:
I think lets stick to what we've already done for now @roldanbob.

roldanbob:
> I think lets stick to what we've already done for now @roldanbob.

Ok. If you're good with that for now, since that was the only outstanding issue, then I think this is complete?

Naros:
Applied to main (=2.4).  Thanks @roldanbob and @obabec .

PR #4731, opened 2023-08-02
DBZ-6743 Remove duplicate query.fetch.size definition

Naros:
https://issues.redhat.com/browse/DBZ-6743

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorConfig.java
@@ -314,8 +314,7 @@ public class SqlServerConnectorConfig extends HistorizedRelationalDatabaseConnec
                     SCHEMA_NAME_ADJUSTMENT_MODE,
                     INCREMENTAL_SNAPSHOT_OPTION_RECOMPILE,
                     INCREMENTAL_SNAPSHOT_CHUNK_SIZE,
-                    INCREMENTAL_SNAPSHOT_ALLOW_SCHEMA_CHANGES,
-                    QUERY_FETCH_SIZE)
+                    INCREMENTAL_SNAPSHOT_ALLOW_SCHEMA_CHANGES)
             .events(SOURCE_INFO_STRUCT_MAKER)
             .excluding(
                     SCHEMA_INCLUDE_LIST,


debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -719,7 +719,6 @@ public static FieldNameAdjustmentMode parse(String value) {
                     SNAPSHOT_FETCH_SIZE,
                     SNAPSHOT_MAX_THREADS,
                     RETRIABLE_RESTART_WAIT,
-                    QUERY_FETCH_SIZE,
```
vjuranek, line 722: it should be removed from `SqlServerConnecotrConfig` where it was accidentally added [here](https://github.com/debezium/debezium/commit/d3ad2bf318c9bc3188171ba4638759f18c9ef847#diff-eb175c0416ed060a1a98046c5daeb07d397d7d009833629c3f7e11b5e6651f28L333), not from Common config.

vjuranek (changes requested):

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
vjuranek: @Naros I'd preffer not main and 2.3 branch to diverge as less as possible and the difference creates more space for other conflicts for subsequence backports. Not if you feel this is a better solution, I'd not object any further:-).

Naros: Personally, I'm fine with either. This was the approach @jpechane and I discussed on Zulip, but I'm also fine doing the other way if we believe that is the lesser of the two evils in terms of conflicts on 2.3 moving forward.

jpechane: Let's prefer the option that is more maintainable - less conflicts. @vjuranek Thanks for pointing that out!

Naros: Ok,I'll adjust the PR accordingly.

Naros: Btw @vjuranek I too like this better, after some sleep I realized I misspoke about Oracle, it had the exclusion defined that we eventually did for SQL Server on main, so really doing this directly in SQL Server is the best play for conflicts.  Great call!

Naros:
@jpechane / @vjuranek - failure is unrelated to this change.

jpechane:
@Naros Applied
@vjuranek Thanks for the advices!

PR #4735, opened 2023-08-02
Update sqlserver.adoc - fix typo

paul-cheung:


jpechane:
@paul-cheung Applied, thanks

PR #4740, opened 2023-08-03
DBZ-6480 fix mysql timezone

xie-shujian:
mysql datetime not right in debezium, just hard code UTC
use connectionTimezone parameter to determin the right timezone


Naros:
Hi @xie-shujian, can you share an example of what you believe is to be wrong with MySQL's timestamp/time details?

jpechane:
@xie-shujian Also @ani-sha is now working on SMT that will allow you to configure and change the timestamp values to arbitrary TZ for all connectors.

PR #4741, opened 2023-08-03
DBZ-6731 Blocking snapshot takes configuration from signal payload

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6731

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbSnapshotChangeEventSource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbSnapshotChangeEventSource.java
@@ -8,12 +8,16 @@ package io.debezium.connector.mongodb;
 import java.util.Collections;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Map;
+import java.util.Optional;
 import java.util.Queue;
+import java.util.Set;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
+import java.util.regex.Pattern;
 import java.util.stream.Collectors;
 
 import org.bson.BsonDocument;
@@ -40,7 +44,10 @@ import io.debezium.pipeline.ErrorHandler;
 import io.debezium.pipeline.EventDispatcher;
 import io.debezium.pipeline.EventDispatcher.SnapshotReceiver;
 import io.debezium.pipeline.notification.NotificationService;
+import io.debezium.pipeline.signal.actions.snapshotting.AdditionalCondition;
+import io.debezium.pipeline.signal.actions.snapshotting.SnapshotConfiguration;
 import io.debezium.pipeline.source.AbstractSnapshotChangeEventSource;
+import io.debezium.pipeline.source.SnapshottingTask;
 import io.debezium.pipeline.source.spi.SnapshotChangeEventSource;
 import io.debezium.pipeline.source.spi.SnapshotProgressListener;
 import io.debezium.pipeline.spi.ChangeRecordEmitter;
@@ -145,7 +152,7 @@ public class MongoDbSnapshotChangeEventSource extends AbstractSnapshotChangeEven
             executor.submit(() -> {
                 try {
                     taskContext.configureLoggingContext(replicaSet.replicaSetName());
-                    snapshotReplicaSet(context, mongoDbSnapshotContext, replicaSet);
+                    snapshotReplicaSet(context, mongoDbSnapshotContext, replicaSet, snapshottingTask);
                 }
                 catch (Throwable t) {
                     LOGGER.error("Snapshot for replica set {} failed", replicaSet.replicaSetName(), t);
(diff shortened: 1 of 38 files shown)


debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -1482,25 +1482,25 @@ public static <T> T querySingleValue(Connection connection, String queryString,
         }
     }
 
-    public String buildSelectWithRowLimits(TableId tableId, int limit, String projection, Optional<String> condition,
-                                           Optional<String> additionalCondition, String orderBy) {
+    public String buildSelectWithRowLimits(TableId tableId, int limit, String projection, String condition,
```
jpechane, line 1485: @mfvitale Why is this change needed?

mfvitale: Just removed the `Optional` as parameters since it's not a good practice. Now, in case of missing value is simply empty `""`

jpechane: I don't think I agree on that one. We use it in the codebase elsewhere as well. The `Optional` is a way how to conway the message in code that value might be present. Even if originally it was told that Optional should be mostly return value there are differing opinions on that rule too.

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/DataCollection.java
```diff
@@ -39,11 +38,11 @@ public T getId() {
         return id;
     }
 
-    public Optional<String> getAdditionalCondition() {
+    public String getAdditionalCondition() {
```
jpechane, line 41: In these cases the `Optional` is valid even in the strict sense

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbSnapshotChangeEventSource.java
```diff
@@ -177,39 +184,49 @@ protected SnapshotResult<MongoDbOffsetContext> doExecute(ChangeEventSourceContex
...
+                .collect(Collectors.toMap(k -> TableId.parse(k.getDataCollection().toString()), AdditionalCondition::getFilter));
+
+        return new MongoDbSnapshottingTask(replicaSets.all(), snapshotConfiguration.getDataCollections(), filtersByTable);
+    }
+
+    @Override
+    public SnapshottingTask getSnapshottingTask(MongoDbPartition partition, MongoDbOffsetContext offsetContext) {
+
+        Set<Pattern> dataCollectionsToBeSnapshotted = connectorConfig.getDataCollectionsToBeSnapshotted();
+        Map<TableId, String> snapshotSelectOverridesByTable = Map.of();
```
jpechane, line 199: Please use better name as Mongo does not use neither select nor tables.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbSnapshotChangeEventSource.java
```diff
@@ -177,39 +184,49 @@ protected SnapshotResult<MongoDbOffsetContext> doExecute(ChangeEventSourceContex
     }
 
     @Override
-    protected SnapshottingTask getSnapshottingTask(MongoDbPartition partition, MongoDbOffsetContext offsetContext, boolean isBlockingSnapshot) {
+    public SnapshottingTask getBlockingSnapshottingTask(MongoDbPartition partition, MongoDbOffsetContext previousOffset, SnapshotConfiguration snapshotConfiguration) {
 
-        if (isBlockingSnapshot) {
-            return new MongoDbSnapshottingTask(replicaSets.all());
-        }
+        Map<TableId, String> filtersByTable = snapshotConfiguration.getAdditionalConditions().stream()
```
jpechane, line 189: `TableId` is relational, it should not leak into MongoDB.

debezium-core/src/main/java/io/debezium/pipeline/signal/actions/AbstractSnapshotSignal.java
```diff
@@ -20,9 +24,20 @@ public abstract class AbstractSnapshotSignal<P extends Partition> implements Sig
...
     protected static final String FIELD_DATA_COLLECTIONS = "data-collections";
+    protected static final String FIELD_DATA_COLLECTION = "data-collection";
     protected static final String FIELD_TYPE = "type";
+    @Deprecated
     protected static final String FIELD_ADDITIONAL_CONDITION = "additional-condition";
+    protected static final String FIELD_ADDITIONAL_CONDITIONS = "additional-conditions";
     protected static final String FIELD_SURROGATE_KEY = "surrogate-key";
+    protected static final String FIELD_FILTER = "filter";
+
+    protected static Set<Pattern> getDataCollectionPatterns(List<String> dataCollections) {
```
jpechane, line 35: Doe it mean that the user cannot prescribe order in which tables will be snapshotted?

jpechane (changes requested):
@mfvitale Overall LGTM. My main comments are

* Removal of `Optional` - we can discuss it later if'd wish but IMHO it is out of scope of this PR and make it quite hard to review. If we decide to drop it then it should be a separate PR
* MongoDB - it seems that relational aspects are leaking to it
* Plase make sure the user can prescribe the order of tables for snapshot

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbSnapshotChangeEventSource.java
mfvitale: Fortunately, your comment revealed a missing. I was convinced that additional conditions was not supported at all but this is not the case. It is unsupported for incremental snapshot but is supported for initial snapshot.

debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
mfvitale: The main "issue" (it's more about clean code) with Optional in the parameter is that it causes conditional logic inside methods, like passing boolean. This means that the method will have two different flow inside and so doing to much things. Leveraging on  method overloading or method with different name can be better. 

But it is just a different opinion on that. We can discuss it apart.

debezium-core/src/main/java/io/debezium/pipeline/signal/actions/AbstractSnapshotSignal.java
jpechane: Could you please dig a bit deeper into initial snapshot? It seems to me that the order is driven by the `table.include.list` property.
For the increment snapshot  - yes deduplication of matchin tables should be done - could you please create a se parate Jira for it?

mfvitale: > Could you please dig a bit deeper into initial snapshot? It seems to me that the order is driven by the `table.include.list` property. 

```java
private void determineCapturedTables(RelationalSnapshotContext<P, O> ctx, Set<Pattern> dataCollectionsToBeSnapshotted) throws Exception {
        Set<TableId> allTableIds = getAllTableIds(ctx);
        Set<TableId> snapshottedTableIds = determineDataCollectionsToBeSnapshotted(allTableIds, dataCollectionsToBeSnapshotted).collect(Collectors.toSet());

        Set<TableId> capturedTables = new HashSet<>();
        Set<TableId> capturedSchemaTables = new HashSet<>();

        for (TableId tableId : allTableIds) {
            if (connectorConfig.getTableFilters().eligibleForSchemaDataCollectionFilter().isIncluded(tableId)) {
                LOGGER.info("Adding table {} to the list of capture schema tables", tableId);
                capturedSchemaTables.add(tableId);
            }
        }

        for (TableId tableId : snapshottedTableIds) {
            if (connectorConfig.getTableFilters().dataCollectionFilter().isIncluded(tableId)) {
                LOGGER.trace("Adding table {} to the list of captured tables for which the data will be snapshotted", tableId);
                capturedTables.add(tableId);
            }
            else {
                LOGGER.trace("Ignoring table {} for data snapshotting as it's not included in the filter configuration", tableId);
            }
        }

        ctx.capturedTables = addSignalingCollectionAndSort(capturedTables);
        ctx.capturedSchemaTables = capturedSchemaTables
                .stream()
                .sorted()
                .collect(Collectors.toCollection(LinkedHashSet::new));
    }
```

All tables are read from the database and then filtered by the pattern specified in the `snapshot.include.collection.list` properties if provided, otherwise all table will be returned. After that only tables that are includes in `table.include.list` will be used for the snapshot. 

>For the increment snapshot - yes deduplication of matchin tables should be done - could you please create a se parate Jira for it?

Yes.

jpechane:
@mfvitale Applied, great work, thanks!

PR #4745, opened 2023-08-04
DBZ-4395 Add connector specific Debezium Connect REST Extension/s and move logic from UI backend to the Debezium Connect REST Extension and DBZ-6761 Add "validate connection" endpoints in connector specific Connect REST extensions

rk3rn3r:
* DBZ-4395 Add connector specific Debezium Connect REST Extension/s and move logic from UI backend to the Debezium Connect REST Extension (part 1 of the re-architecturing/re-factoring)
* DBZ-6761 Add "validate connection" endpoints in connector specific Connect REST extensions

closes to https://issues.redhat.com/browse/DBZ-4395
closes to https://issues.redhat.com/browse/DBZ-6761


rk3rn3r:
@indraraj FYI this is my branch with the initial implementation for the Connect REST Extension based architecture for the UI.

@Naros @jpechane @jcechace You might want to take a small glimpse at it as well?

rk3rn3r:
If you all agree, then we could merge this PR after review and we will handle the open todos in separate JIRAs?

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/rest/DebeziumSqlServerConnectorResource.java
```diff
@@ -0,0 +1,61 @@
...
+
+    @javax.ws.rs.core.Context
+    private ServletContext context;
+
+    public DebeziumSqlServerConnectorResource(ConnectClusterState clusterState, Map<String, ?> config) {
+        this.config = config;
+    }
+
+    @GET
+    @Path("/version")
```
mfvitale, line 44: Do you think  `getConnectorVersion` and  `getConnectorSchema` can be moved to an interface with default implementation or to an abstract class and let the implementation to provide just the name of the json?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/rest/DebeziumMySqlConnectorResource.java
```diff
@@ -0,0 +1,61 @@
...
+    @GET
+    @Path("/version")
+    public String getConnectorVersion() {
+        return Module.version();
+    }
+
+    @GET
+    @Path("/schema")
+    public JsonNode getConnectorSchema() {
+        ObjectMapper mapper = new ObjectMapper();
```
mfvitale, line 52: Can this be passed on the constructor and instantiated on `*ConnectRestExtension`?

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/rest/DebeziumSqlServerConnectorResource.java
rk3rn3r: About `getConnectorVersion`: I don't see what this would safe us? We would need to have this method here anyway for the `@Path` annotation and we need to pass the right instance of the `Module` class, which has no interface at all. So this would not work imho. Or do you have another solution?

About `getConnectorSchema`: I was thinking about it earlier too, but I decided against it as we won't safe a lot of duplicate code but we loose independence (the name of that file can be customized on the connector's `pom.xml` level, so if for whatever reason we want to change the name, then we can simply change it in the local implementation. maybe I should even load it from the `Module` class via a System Property eventually so there is only one place to update the name and it is inherited everywhere?) If we would change the name inside the connector's pom.xml it would require us to change the code and add logic for it inside that shared package and release debezium-core (or whatever shared package we use) and with this all connectors that are using that package (even only a single connector might require the change).

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/rest/DebeziumMySqlConnectorResource.java
rk3rn3r: A very good point, this is a copy and paste leftover from the first naive implementation.

rk3rn3r: What do you think about this approach in general @mfvitale ? I could also just return the content of the file. The only reason why I did it that way was to ensure valid JSON is coming from that file.

mfvitale: This can be an option but is still not easily testable. Passing it to the constructor you can easy tests the class passing also a mocked `ObjectMapper`. If this serves just for validating the json isn't it better to fail on the `register` method?

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/rest/DebeziumSqlServerConnectorResource.java
mfvitale: > About `getConnectorVersion`: I don't see what this would safe us? We would need to have this method here anyway for the `@Path` annotation and we need to pass the right instance of the `Module` class, which has no interface at all. So this would not work imho. Or do you have another solution?

Yes, `Module` is a problem. The only solution I see is to have the `getConnectorVersion` method on the abstract class the calls the abstract method that must be implemented by each connector with the `Module.version()'. 

Regarding the REST annotation this should be feasible (If I am not wrong) https://stackoverflow.com/questions/53667894/jersey-jax-rs-annotating-interface-not-the-implementation-class-results-in-error 

> About `getConnectorSchema`: I was thinking about it earlier too (about having a common abstract class for it), but I decided against it as we won't safe a lot of duplicate code but we loose independence (the name of that file can be customized on the connector's `pom.xml` level, so if for whatever reason we want to change the name or how it is formed, then we can simply change it in the local implementation. maybe I should even load it from the `Module` class via a System Property eventually so there is only one place to update the name and it is inherited everywhere?) If we would change the name inside the connector's pom.xml it would require us to change the code and add logic for it inside that shared package and release debezium-core (or whatever shared package we use) and with this all connectors that are using that package (even only a single connector might require the change).

I was thinking to just push in the abstract class the main code and let the path coming from an abstract method that must be implemented by each connectors. Is it feasible?

Naros:
Hi @rk3rn3r I haven't gotten through all of this PR just yet, but one thing struck me when looking at the connector-specific `ConnectRestExtension` implementations, would it be possible to abstract the dependency on `connect-runtime` away entirely for each connector type and solely depend on the Debezium rest extension?  

For example, the REST extension module would expose:
```
public class DebeziumRestExtension<T> implements ConnectRestExtension {
    @Override
    default void register(ConnectRestExtensionContext registerPluginContext) {
      try {
        restPluginContext.configurable().register(getConnectorClass().newInstance(config));
      }
      catch (Exception e) {
        throw new ConnectException("Failed to instantiate connector rest extension", e);
      }
    }

    @Override
    default String version() {
      return AppInfoParser.getVersion();
    }

    Class<T> getConnectorClass();
}
```

I also think if you'd rather implement an `AbstractDebeziumExtension` implementation inside the REST extension and then utilize this as an SPI for each connector module, that would be okay too.  But doing this, we should be able to eliminate the `connect-runtime` dependency based on what you've shared thus far from each connector module and only rely on the REST extension module proper.  wdyt?

rk3rn3r:
Hey Chris. Thank you for the feedback. 
Can you explain me why you want to avoid that dependency? I think in any of the Debezium backends/runtimes we already have a (currently explicit) dependency on `connect-runtime` so it should already be available everywhere, no? Embedded engine (which is used by Debezium Server) uses it and Kafka Connect provides it obviously. Why should we encapsulate it and build facades when the only foreseeable backends/runtimes are based on that connect-runtime dependency?

The SPI that we re-use is the one from Kafka Connect and I think we can simply re-use the same for Debezium Server later or make a copy to a custom namespaced SPI?
In general my idea was to explicitly avoid coupling of different modules so that a change inside that parent repo/module would not require changes to basically not-affected connector modules. but this is a personal choice and I think someone prefers either the higher coupled or the lesser-coupled one.

I would personally prefer if I could leave some refactoring for the moment when we make those endpoints available to Debezium Server and there's urge to avoid some dependencies and we are really required to add some facades in the architecture.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/rest/DebeziumMySqlConnectorResource.java
rk3rn3r: Hummm, what logic exactly would you test except that the right file & content is being returned or that exception being thrown in case of failure? Especially when I compare it to the alternative solution (return just the file contents as String without further JSON parsing and "validation"). Using that ObjectMapper re-conversion is just some maybe unnecessary safety net.

mfvitale: Should you catch `DataBindException` for "validation" error right? In the case of `IOException` you are just checking if able to read the file. In any case. do what you think is the best.

Naros:
> Can you explain why you want to avoid that dependency? I think in every of the Debezium backends/runtimes we already have a (currently explicit) dependency on connect-runtime so it should already be available everywhere, no? 

Unfortunately, no.  The connectors themselves only ever depend on `connect-api` for runtime.  There is a dependency on the embedded engine's test jar that directly depends on the `connect-runtime`, but its a test-only dependency and not runtime.

My concern with directly being dependent on the `connect-runtime` is that Kafka imposes different forward/backward compatibility rules for their runtime module than they do for the API module, and this can have an adverse effect on the broad range of Kafka versions that Debezium can be used on.  We've already seen some obscure changes made to several contracts as-is with some of the latest versions of Kafka that have made being cross-version compatible challenging and we should try to avoid adding any other potential contributor to that mix if possible.

> Why should we encapsulate it and build facades when the only foreseeable backends/runtimes are based on that connect-runtime dependency?

Well if the `connect-runtime` is needed as you've demonstrated, if you mark it as a compile-only dependency of the REST extension, you avoid this dependency being pulled into each connector.  To me this allows connectors to easily be compatible across a larger range of Kafka Connect environments and we avoid any potential misuse of `connect-runtime` dependencies in connector runtime modules.  

> I would personally prefer if I could leave some refactoring for the moment when we make those endpoints available to Debezium Server and there's urge to avoid some dependencies and we are really required to add some facades in the architecture.

I'm definitely not saying you need to do the refactor at this moment, it can certainly be something that is left at the end of the development of this feature; that way you can focus on the work at hand and clean-up at the end.

rk3rn3r:
@Naros please read my sentence again. I was not talking about connector dependencies but dependencies of the runtimes/backends. They all depend on `connect-runtime` and that way it is provided (by Kafka Connect and Debezium Server), which should be similar to the compile-time dependency that you want t have, right?

debezium-connector-mongodb/pom.xml
```diff
@@ -34,6 +34,21 @@
             <groupId>org.mongodb</groupId>
             <artifactId>mongodb-driver-sync</artifactId>
         </dependency>
+        <dependency><!--// For Debezium MongoDB Connect REST Extension, @see DebeziumMongoDbConnectRestExtension //-->
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>connect-runtime</artifactId>
+            <scope>provided</scope>
```
rk3rn3r, line 40: @Naros isn't this what you asked for? Provided scope is similar to compile-time deps, isn't it?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/rest/DebeziumMySqlConnectorResource.java
rk3rn3r: @mfvitale Yes you are right DataBindException must be catched, but DataBindException is an IOException (subclass/extended class of IOException), so it is already catched with the IOException.

jcechace:
> 

Nevertheless this has been something we were trying to gradually remove (or minimise) for a while. We should avoid further coupling with KC as much as possible

debezium-connector-mongodb/pom.xml
jcechace: AFAIK provieded is more similar to runtime actually with the exception of not pulling down the dependency by the current maven reactor during the build but rather it is deferred to the user/runtime platform to provide the dependency during runtime

rk3rn3r:
@jpechane This PR is still waiting for you ;)

jpechane (commented):
@rk3rn3r Thanks for the PR, it is very good start. I left couple of comments.
We just really need to handle the `connect-runtime` dependency.

- I really don't see any reason for `connect-runtime` dependency for connectors as it seems to me they really need `connect-api` only
- `debezium-connect-rest-extension` right now really needs the `connect-runtime` dependency but could you refactor the code in a way that there is a separate package with class(es) interacting exclusively with Connect Runtime and then the rest? En example is `PredicateDefinition`, instead of leakign connect runtime class to it via the factory method there would be the factory class creating all definitions the will have localized all Connect runtime dependent code. That way we can easily replace it with another one for Debezium Server implementation

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/DebeziumResource.java
```diff
@@ -48,21 +46,19 @@
 @Consumes(MediaType.APPLICATION_JSON)
 public class DebeziumResource {
 
-    private static final Logger LOGGER = LoggerFactory.getLogger(DebeziumResource.class);
+    private static final List<String> SUPPORTED_CONNECTORS = Arrays.asList(
```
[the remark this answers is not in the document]
rk3rn3r: Why? It is only used inside this class, isn't it? I think they serve different purpose.

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/DebeziumResource.java
```diff
@@ -107,40 +83,47 @@ else if (m.lookingAt()) {
...
-    @GET
-    @Path("/topic-creation")
-    public boolean getTopicCreationEnabled() {
-        return this.isTopicCreationEnabled;
-    }
-
-    private synchronized Boolean isTopicCreationEnabled() {
+    private synchronized Boolean isTopicCreationEnabled(Map<String, ?> config) {
         Version kafkaConnectVersion = parseVersion(AppInfoParser.getVersion());
         String topicCreationProperty = (String) config.get("topic.creation.enable");
```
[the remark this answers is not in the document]
rk3rn3r: Good point

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/entities/PredicateDefinition.java
```diff
@@ -0,0 +1,47 @@
...
+import org.apache.kafka.connect.runtime.isolation.PluginDesc;
+import org.apache.kafka.connect.transforms.predicates.Predicate;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+/**
+ * JSON model that describes a Single Message Transform (SMT) entry.
```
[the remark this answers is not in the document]
rk3rn3r: yeah, c&p issue :scissors: when I added it recently with a first refactoring :face_exhaling:

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/DebeziumResource.java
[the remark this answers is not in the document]
jpechane: I suppose those two lists or should be kept in sync, shouldn't it?

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/entities/PredicateDefinition.java
```diff
@@ -0,0 +1,47 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(PredicateDefinition.class);
+
+    @JsonCreator
+    public PredicateDefinition(String className, ConfigDef config) {
+        super(className, config);
+    }
+
+    @JsonCreator
+    public static PredicateDefinition fromPluginDesc(PluginDesc<Predicate<?>> predicate) {
```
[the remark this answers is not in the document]
rk3rn3r: `PluginDesc` is what we receive from Kafka Connect and I would personally say that changing it to avoid some depenendencies is an effort not really balanced with the improvements. We don't really have a dependency on connect-runtime we only consume a datastructure from there. That means we rely on data (what we do anyway) and not behavior. We would copy that datastructure 1:1 into a Debezium package. Is this really worth looking at the solid compatibility of Kafka packages? I see that for several reasons in some situations it is good to be decoupled from some external dependencies, but I do not know what is such a reason for Kafka/Kafka Connect deps (especially as a `provided` scope dependency)? I would also say that this architecture would increase the complexity of this PR from M to L/XL, wdyt? Is this really worth the effort now (in this PR)? @jpechane

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/entities/PredicateDefinition.java
```diff
@@ -0,0 +1,47 @@
...
+        String className = predicate.pluginClass().getName();
+        LOGGER.info("Loading config for PREDICATE: " + className + "...");
+        try {
+            return new PredicateDefinition(className, predicate.pluginClass().getDeclaredConstructor().newInstance().config());
+        }
+        catch (ReflectiveOperationException e) {
+            LOGGER.error("Unable to load PREDICATE: " + className + "\n\t Reason: " + e);
+            throw new RuntimeException(e);
```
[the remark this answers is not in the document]
rk3rn3r: Just for "the sake of uniformity" or does it have a technical reason? The REST extension runs outside the scope of connectors and the request fetches registered predicates/transforms/etc from the runtime (not really related to Debezium).
Maybe I should even drop the exception at all... Do you think there's a reason to re-throw any exception in the first place?

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/entities/TransformDefinition.java
```diff
@@ -0,0 +1,53 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(TransformDefinition.class);
+
+    @JsonCreator
+    public TransformDefinition(String className, ConfigDef config) {
+        super(className, config);
+    }
+
+    @JsonCreator
+    public static TransformDefinition fromPluginDesc(PluginDesc<Transformation<?>> transformPlugin) {
```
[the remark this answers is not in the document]
rk3rn3r: See my other comment. Isn't a change like this too huge for this PR and for doing it now? I am not convinced that it is good architecture (although it can be a good architecture) to adapter (more or less 1:1) to Kafka classes when I look at the stability and compatibility over time. It should be worth to introduce that complexity and effort, but I am not yet convinced (also I generally understand and promote such architecture - when it is worth the effort). But change my mind! :wink:

debezium-connect-rest-extension/src/test/java/io/debezium/kcrestextension/TestHelper.java
```diff
@@ -47,18 +47,16 @@ public static DebeziumContainer getDebeziumContainer() {
     }
 
     public static void setupDebeziumContainer(String debeziumContainerImageVersion) {
-        DEBEZIUM_CONTAINER = new DebeziumContainer(DockerImageName.parse("quay.io/debezium/connect:" + debeziumContainerImageVersion))
+
+        DEBEZIUM_CONTAINER = new DebeziumContainer(DockerImageName.parse("debezium/connect-base:" + debeziumContainerImageVersion))
```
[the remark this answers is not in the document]
rk3rn3r: Because for the tests where Kafka Connect does not support topic auto-creation (prior 2.6.0) we are using ancient Debezium 1.2 which is not available on `quay.io`. This code here even changed drastically in my workspace (not pushed yet) to enable some more tests and needs to be reviewed later again after I finished adding those additional tests.

debezium-connector-mongodb/pom.xml
```diff
@@ -34,6 +34,21 @@
             <groupId>org.mongodb</groupId>
             <artifactId>mongodb-driver-sync</artifactId>
         </dependency>
+        <dependency><!--// For Debezium MongoDB Connect REST Extension, @see DebeziumMongoDbConnectRestExtension //-->
```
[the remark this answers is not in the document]
rk3rn3r: Dependencies to the Kafka Connect `herder` which will be required for the other feature that are not yet migrated. The REST extension only has very limited visibility / access to Kafka Connect data.

debezium-connector-mongodb/pom.xml
rk3rn3r: Ah yes, I think it is required as we need access the Kafka Connect `Herder` and consume some Kafka Connect datastructures that are in the `connect-runtime` package.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/rest/DebeziumSqlServerConnectorResource.java
rk3rn3r: I am not convinced because we have connectors that are not in the main repo which are required to upgrade when we upgrade the centralized code. I understand the idea of DRY but didn't we also learn that in some distributed environments copying is better than DRY because of lower coupling? And in this case we are talking about 1 line of active code (plus try - catch). Instead of this 1 single line we need that line plus 1 additional method that returns the filename/path in every implementation, plus the coupling introduced to that central abstract implementation that holds the shared code.

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/DebeziumResource.java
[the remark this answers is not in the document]
rk3rn3r: It looks like that but I don't agree. The Debezium Schema Generator (where this class was made for) can support other connectors than the UI/REST endpoints. For example look at projects like RHOC.

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/DebeziumResource.java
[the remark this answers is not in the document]
Naros: So `Map<String, ?>` is just another way to write `Map<String, ? extends Object>`; therefore you can simply recast the map to `Map<String, Object>` temporarily to use the `getOrDefault` method.

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/entities/PredicateDefinition.java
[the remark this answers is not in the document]
Naros: Hi @rk3rn3r and @jpechane, I do think there will be spots in the REST extension where we're inevitably going to have dependencies on the `connect-runtime` bits, for example, using the `Herder#plugins()` method to get the loaded transformations (unless we do our own classpath scanning of course).  

That said, as an alternative, what if we passed the `Class<?>` reference obtained from the `PluginDesc` as the argument?  That would remove the runtime dependency in this class's scope at the very least.

Obviously this all based on how we interact with `connect-runtime` in various spots, but if we can effectively centralize that into one or two classes at most, we guard ourselves from the module's wider view of code and if we face some compatibility concern in the future, the changes needed are far more isolated and targetable.  wdyt?

debezium-connect-rest-extension/pom.xml
```diff
@@ -119,7 +123,7 @@
     <profile>
       <id>assembly</id>
       <activation>
-        <activeByDefault>false</activeByDefault>
+        <activeByDefault>true</activeByDefault>
```
Naros, line 126: Is this necessary and intended?

rk3rn3r: Yes it is! The REST extension now shares code from `debezium-core` and this requires installing it with its dependencies in a "plugin-dir".

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/rest/DebeziumSqlServerConnectorResource.java
rk3rn3r: I have centralized this as an interface now. Please have another look and resolve this discussion if it fits. @mfvitale @jpechane

debezium-connector-mongodb/pom.xml
rk3rn3r: I have removed this dependency from the connector modules in a kind of elegant way, I assume. :crossed_fingers: Thx for pushing me, after some time I found a hopefully good solution.

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/entities/PredicateDefinition.java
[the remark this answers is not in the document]
rk3rn3r: I have removed the dependency from the connector modules. I have to keep it for this (the) general KC REST extension because I depend on the herder. I think this is okay because this code is purely made for KC. When we provide similar endpoints for DBZ Server in the future we won't have to re-use this. If we still want to re-use the same logic for DBZ Server we can rework this small part in the future. wdyt @Naros  @jpechane ?

debezium-connector-mongodb/pom.xml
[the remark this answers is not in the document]
rk3rn3r: removed the dependency

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResourceIT.java
```diff
@@ -0,0 +1,124 @@
...
+    @BeforeClass
+    public static void checkCondition() {
+        Assume.assumeThat("Skipping DebeziumMongoDbConnectorResourceIT tests when assembly profile is not active!",
+                System.getProperty("isAssemblyProfileActive", "false"),
+                is("true"));
+    }
+
+    @Before
+    public void start() {
+        RestExtensionTestInfrastructure.setupDebeziumContainer(Module.version(), DebeziumMongoDbConnectRestExtension.class.getName());
```
jcechace, line 47: Why not use `MongoDbDatabaseProvider`? That way we can assure that the initial container infrastructure is consistent across tests as is the imagehandling, port resolver (important for MacOS) and so on.

debezium-connector-mongodb/src/test/resources/Dockerfile.rest.test
```diff
@@ -0,0 +1,29 @@
...
+    curl --silent -fSL -o /tmp/package.tar.gz "${MAVEN_REPOSITORY}/debezium-${PACKAGE}/${DEBEZIUM_VERSION}/debezium-${PACKAGE}-${CONNECTOR_VERSION}.tar.gz" && \
+    echo "Extracting debezium-${PACKAGE}-${CONNECTOR_VERSION}.tar.gz ..." && \
+    tar -xzf /tmp/package.tar.gz -C $EXTERNAL_LIBS_DIR && \
+    echo "Successfully installed debezium-${PACKAGE}-${CONNECTOR_VERSION}!" ; \
+    rm -f /tmp/package.tar.gz ; \
+done
+
+COPY --chown=kafka:kafka debezium-connector-${CONNECTOR}-${DEBEZIUM_VERSION}-plugin.tar.gz /tmp/plugin.tar.gz
+
+RUN tar -xvzf /tmp/plugin.tar.gz -C ${KAFKA_CONNECT_PLUGINS_DIR}/ ; rm -f /tmp/plugin.tar.gz
```
jcechace, line 29: Don't we have a support fro extending the KC container image in our TC wrapper already? I can't image anyone will keep these dockerfiles in sync.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
```diff
@@ -280,7 +280,6 @@ public void start() {
     public void stop() {
         LOGGER.info("[{}] Stopping...", name);
         MoreStartables.deepStopSync(members.stream());
-        network.close();
```
jcechace, line 283: @rk3rn3r  Though I likely  aagree with this (the reason for this was that initially network was not configurable and this was likely forgotten) this is exactly the type of change which should have it's own commit clarifying what the isue was here.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResourceIT.java
rk3rn3r: Why didn't you do this in the Debezium when you was working with Anisha? This task was about moving what we have in the UI backend to connector-specific Connect REST extensions. If we can improve then contributions are welcome. ;)

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
rk3rn3r: For me it was just wrong that a single component (of a test stack) closes a network meant for testing a whole stack. I didn't feel this needs discussion, a separate task or commit.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/RestExtensionTestInfrastructure.java
```diff
@@ -0,0 +1,257 @@
...
+import org.testcontainers.utility.DockerImageName;
+
+import io.debezium.testing.testcontainers.Connector;
+import io.debezium.testing.testcontainers.ConnectorConfiguration;
+import io.debezium.testing.testcontainers.ConnectorResolver;
+import io.debezium.testing.testcontainers.DebeziumContainer;
+import io.debezium.testing.testcontainers.MongoDbReplicaSet;
+import io.debezium.testing.testcontainers.util.MoreStartables;
+
+public class RestExtensionTestInfrastructure {
```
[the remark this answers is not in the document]
rk3rn3r: Why should this interfere? The only touch point can be TCP ports (which are/should be dynamically managed by Docker) or components which use broadcasts to discover the cluster (should not be possible with Docker network isolation until you not expose everything to the local physical network).
Like I said above, I think at the moment we have to accept that we have a mix of old an new things for historical reasons and no time to spent the effort to migrate everything now. But I am also against keeping the legacy way of testing because then we make the mountain higher and higher that we have to walk eventually.

rk3rn3r:
@jpechane @ani-sha There was something wrong with the commit check. My commit has a `DBZ-` task id, it even has 2 in it and it starts with one. The bot should not have commented, should it?

ani-sha:
@rk3rn3r I saw the actions log and found a parsing error with `jq` while attempting to retrieve the commit message.  This issue could potentially be linked to a failure in the API request.

debezium-connector-mongodb/src/test/resources/Dockerfile.rest.test
jcechace: Are we using KC in such a way though? 

@Naros ?

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
jcechace: Yes for you and for me it makes sense as well (since I know what is going on). The prupose is for somebody else to be able to track why this changed

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/RestExtensionTestInfrastructure.java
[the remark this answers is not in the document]
jcechace: So what added value you get with TC databases compared to what we have? Which cannot be gradually upgraded with the rest (say connector by connector -- the same way as it was done with MongoDB, for which there were additional reasons to switch in order to support sharded clusters and proper replica sets).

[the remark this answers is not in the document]
jcechace: Also what about the complexity and the fact that I don't think this class shoudl exists within the testcontainers module.

rk3rn3r:
@ani-sha maybe we can retry 2-3 times in case of failure? do we check the exit code of the API call?

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/DebeziumResource.java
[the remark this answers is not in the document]
rk3rn3r: Isn't it more performant to cast the single retrieved value of type `?` to `String` and compare it like this instead of casting the whole map to <String, Object> only to retrieve a single entry that even might not be set in most cases?

rk3rn3r:
@jcechace @jpechane @Naros I have splitted the changes into 3 commits (1 for the main task, 1 for the minor MongoDB bugfix, 1 again for the main task and the "validate connection" endpoint task)

rk3rn3r:
The DB2 test errors are unrelated to the code change, it seems. I filed https://issues.redhat.com/browse/DBZ-6849 for those.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/RestExtensionTestInfrastructure.java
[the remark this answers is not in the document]
rk3rn3r: Should the class be named "TestStackForRestExtensions"  or something similar? It is a template or a composition for some testcontainers meant for testing the generic and connector-specific REST extensions.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
rk3rn3r: I split the PR in 3 commits now to reflect and isolate the bugfix.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResourceIT.java
obabec: I agree with Jakub on this one. If you are adding new tests (whatever) tests, it should be consistent with the previous ones. The Mongo connector tests are the only tests that are stable and I am not willing to trade this because of this.

debezium-connector-mongodb/src/test/resources/Dockerfile.rest.test
obabec: I don't agree that we have to accept anything. We should address these things now and not just accept that it is gonna be bad.

obabec (commented):
In general, I do not think that this approach works. I would rather split this into separate PRs and address each connector individually.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResourceIT.java
jcechace: 1) I wasn't wroking with anisha. I was hotfixing a code which should have been gradually upgraded (and never was) and which was using MongoDB without the udnerstanding how the database networking works
2) This provider is meant for the mongo connector module tests -- though it acts as a factory for the varius configurations used by these tests it is not trying to contain configuration required by all other modules
3) Porting something from one codebase to another is not just a matter of simply moving the code.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
```diff
@@ -280,7 +280,7 @@ public void start() {
...
```
jcechace, line 283: I'd agree that the code was broken since it closed a potentially shared network when the database shut down. However the code is now broken in exactly the oposite way -- the code now doesn't close a network implicitely created by the builder, which is likely to be used when the user wants just the databases.

If we are fixing this it should be done properly. The network should be closed only if it was provided explicitely. In case the network is implicitly created by the builder (since you don't want to manually manage the only network in cases when database is all you need) then it should also be implicitly closed.

rk3rn3r: I would agree that this can be causing issues, although I don't see that in the current test. Can you point me to an error message in the CI runs?
Personally I am still convinced, that if there's only one instance then the network would automatically been shutdown/closed at the end of the test / test suites run by testcontainers, no?
I like your idea but currently working tests show that it works that way and testcontainers might automatically remove the network at the end, doesn't it? And the added complexity might not be needed, is it  really required? I never saw such behavior anywhere else. Doesn't the GenericContainer create a network when none was configured? 
Furthermore, shouldn't another component exist that coordinates the whole test infrastructure/ test stack and for example connects the Mongo DB network to other components in the test stack? That components should manage this. I would personally remove the `Network.newNetwork()` code from the container code and let another component manage this.

rk3rn3r:
@obabec all tests are green and most of the code is simple copy and paste and using other connector-specific config options. I think it'll work this way.
@jpechane?

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
jcechace: For all the tests but rest extension the entire "test stack" is just the databse.

Regardless if there the network is garbage collected -- if we are fixing something we should not break it in a different way. That's just good practice. I'm fine if you just open a Jira for this in order not to lose track and it can be fixed afterwards.

jcechace:
> @obabec all tests are green and most of the code is simple copy and paste and using other connector-specific config options. I think it'll work this way. @jpechane?

I think Ondrej is looking at it form a Quality perspective. I think we can pretty much agree that the code show that this was mostly coppied from another project. This also means that extendability wasn't likely taken into account. How is this goign to work with anything but Kafka Connect (e.g. debezium server). We now that there is at least one group of users who would like to have a UI for deploying DS.

debezium-connect-rest-extension/pom.xml
jcechace: It's just for the rest-extension module, right? So it won't activate the assembly profile for the entire build (Provided I'm building the entire project)?

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/entities/PredicateDefinition.java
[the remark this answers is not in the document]
jpechane: The rest-extension depends on debezium-core so I don't see any reason why we should not use DebeziumException for sake of uniformity.

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/entities/PredicateDefinition.java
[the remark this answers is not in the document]
jpechane: The rest-extension depends on debezium-core so I don't see any reason why we should not use DebeziumException for sake of uniformity.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/RestExtensionTestInfrastructure.java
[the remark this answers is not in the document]
jpechane: This class should not really belong in `debezium-testing`

jpechane:
Let me express my opinion - this PR should not ensure we get everything in plae for Debezium Server support. The purpose of this PR is

* get rid of Debezium UI backend
* create REST extension compatible with Debezium UI frontend by moving the code
* create battery of tests to make sure everyhing that everything works

This is a stepping stone to further refactoring that would make the code compatible with Debezium Server. As we will have a baseline and test battery we can be reasonably safe with larger code changes.

debezium-core/src/main/java/io/debezium/metadata/ConnectorDescriptor.java
```diff
@@ -33,4 +41,38 @@ public String getClassName() {
...
+            case "io.debezium.connector.oracle.OracleConnector":
+                return "oracle";
+            case "io.debezium.connector.postgresql.PostgresConnector":
+                return "postgres";
+            case "io.debezium.connector.sqlserver.SqlServerConnector":
+                return "sqlserver";
+            default:
+                throw new RuntimeException("Unsupported connector type with className: \"" + className + "\"");
+        }
+    }
```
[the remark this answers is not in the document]
rk3rn3r: @jcechace Can you explain why this is a codesmell for you? There must be a place where we map from the Connector classname to a unique identifier for the UI. This is shared code used for a) the schema genarator, b) the Connect REST extensions, etc. We won't have the classes / classnames until we make up a dependency on all possible connector modules neither is there any other better place to describe that mapping from a classname to a connector type id. It belongs to what a ConnectorDescriptor defines / produces, so it should belong here until you suggest a better place for it. Without providing an alternative complaining is pointless, isn't it?

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/RestExtensionTestInfrastructure.java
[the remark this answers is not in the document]
rk3rn3r: @jpechane  Can you suggest a better place? What's wrong with my previous point of view about that this describes a tech stack for containers in `debezium-testing-testcontainers`? `debezium-testing-testcontainers` can produce the containers but also full test-stacks, no?

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
rk3rn3r: That code line breaks tests that reuse the same MongoDB testcontainers like you would normally do when you are not coming from ancient old test stack of an aged project. So this line must go now and not later based on a Jira.

rk3rn3r:
@jpechane 

> IMHO we need a common abstract test template class that defines all scenarios and is implemented for each connector - think in terms of IncrementalSnapshotIT

While `IncrementalSnapshotIT` depends on a common / shared behavior those tests for the UI don't. The tests for the UI are specific to the underlying database/connector implementation. Yes, we definitely could make this equal for all connectors, but then we loose a lot of flexibility in the UI to react specifically according to the connector because we need to settle on a common set of behaviors, config options, etc. Looking at the legacy UI code it showed that we were quite limited by doing it like that.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
jcechace: Where did I state the oposite? I said that the fix is not complete and leaves a "TODO" if you will and as such should be captured in a followup jira. So please create one.

debezium-core/src/main/java/io/debezium/metadata/ConnectorDescriptor.java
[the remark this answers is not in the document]
jcechace: There are methods to do it cleanly no? E.g. you extract this from conenctor config or via some declarative metadata such as annotation. You can even employ a design pattern such as strategy or whatnot.  

I understand why this is here -- due to it being a shared code. I'm saying that easiest to do (and incorrect) approach was taken here. I'm fine with this being a temporary solution but it is a code smell.

jcechace:
> @jpechane
> 
> > IMHO we need a common abstract test template class that defines all scenarios and is implemented for each connector - think in terms of IncrementalSnapshotIT
> 
> While `IncrementalSnapshotIT` depends on a common / shared behavior those tests for the UI don't. The tests for the UI are specific to the underlying database/connector implementation. Yes, we definitely could make this equal for all connectors, but then we loose a lot of flexibility in the UI to react specifically according to the connector because we need to settle on a common set of behaviors, config options, etc. Looking at the legacy UI code it showed that we were quite limited by doing it like that.

Isn't there a common subset to that behaviour though? And isn't the setup for each connector structurally exactly the same? Jiri's proposal by no means suggests that the connector specific test classes can't extend the shared test base.

debezium-core/src/main/java/io/debezium/metadata/ConnectorDescriptor.java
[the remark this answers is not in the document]
Naros: Hi @rk3rn3r, when I typically see a large if-block or a switch statement, the first thing I usually ask is whether or not this could be refactored to take advantage of OOO design principles.

Something to consider might be whether you could take advantage of the ServiceLoader pattern and have a contract that each connector implements that exposes the connector class, its display name, and other metadata.  If you are deploying the REST extension alongside the connector as it seems you're doing with the connector-specific classes, this would work as there would only ever be a single REST implementation per connector plugin, no?

This would also work to avoid the need to deal with passing class names for connectors that are not supported and therefore having to throw exceptions like the reference Jakub shared.

Not saying this needs to change, but it's really something to consider to make code fundamentally more reliable and less more resilient to potential corner case failures.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
rk3rn3r: Never mind. I misunderstood your previous comment then.

debezium-core/src/main/java/io/debezium/metadata/ConnectorDescriptor.java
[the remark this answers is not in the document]
Naros: > Are you sure this will work in the way Kafka Connect REST Extensions are loaded? KC uses an isolated ClassLoader that should not be capable of finding artifacts in the connect plugin dirs of KC. I think this restriction then applies to ServiceLoader as well, doesn't it?

Well, it depends on how we intend to bundle the REST extension.  For example, if the extension is bundled with each connector plug-in, then the ClassLoader would only find the single implementation of the service for that specific connector, which I believe should satisfy API calls into the connector-specific REST endpoints. 

> Furthermore, this does not solve the issue that I described above, that the UI needs a stable identifier for connectors. If a connector can decide to change its "connector type id" then this is not stable. I think this code belongs into debezium-core or a similar central package in a way that defines the name globally instead of decentralized in the connector modules. wdyt?

> Furthermore, this does not solve the issue that I described above, that the UI needs a stable identifier for connectors. If a connector can decide to change its "connector type id" then this is not stable. I think this code belongs into debezium-core or a similar central package in a way that defines the name globally instead of decentralized in the connector modules. wdyt?

I disagree, I actually think something like this probably makes sense in each connector's `Module` implementation.  If you take a look at the connector configuration classes, you'll see that we have two methods that get exposed, `getContextName()` and `getConnectorName()`, and both of these are delegating to `Module`.  It's implied that we should not change these values as it could have implications since these are used in metrics bean names, the source info block, and so on.  

I could imagine `Module` being adopted as follows and perhaps one of the REST classes you load via ServiceLoader could then simply act like the configuration class and delegate the call to the local connector's `Module`.
```
public static <T> Class<T> getConnectorClass() {
  return OracleConnector.class;
}

public static String getConnectorClassName() {
  return getConnectorClass().getName();
}
```

> this applies in the same way to the SchemaGenerator which requires a stable connector type id for connectors
> The definition of these ids should be closer to the UI and SchemaGenerator than to the connector modules.

I don't see how that precludes anything that was suggested.  In fact, you already use ServiceLoader to fetch the specific `ConnectorMetadata` for the connector in question, so that same instance could have methods that delegate to the `Module` within each connector module to return that specific detail without the need for the switch statement.

In fact, if `ConnectorMetadata` delegated to `Module`, you would replace the hard-coded strings you have inside each of the metadata classes with a singular string reference that is used for a variety of reasons.  This means that if there is ever a reason to change the identifier in the future, this is managed in a singular place rather than across a multitude of files littered throughout the repository.

What do you think of this?  @jpechane @jcechace any takes?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnector.java
```diff
@@ -216,17 +216,18 @@ public Config validate(Map<String, String> connectorConfigs) {
...
             try (MongoClient client = connContext.connect()) {
-                client.listDatabaseNames();
+                client.listDatabaseNames().first(); // only when we try to fetch results a connection gets established
             }
             catch (MongoException e) {
-                connectionStringValue.addErrorMessage("Unable to connect: " + e.getMessage());
+                if (connectorConfigs.containsKey(MongoDbConnectorConfig.CONNECTION_STRING.name())) {
+                    connectionStringValue.addErrorMessage("Unable to connect: " + e.getMessage());
+                    results.put(MongoDbConnectorConfig.CONNECTION_STRING.name(), connectionStringValue);
+                }
```
jcechace, line 228: If this is something that needs to be done then this is likely an issue at many other places (e.g MysqlConnector, PostgresConnector)

jcechace, line 228: If this is something that needs to be done then this is likely an issue at many other places (e.g MysqlConnector, PostgresConnector)

rk3rn3r: You are definitely right! @jcechace

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/entities/PredicateDefinition.java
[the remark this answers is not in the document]
rk3rn3r: This modules name is `kcrestextension`. It is a KC specific implementation that adds some behavior which is missing from KC. Debezium Server (DS) might implement this separately and I don't see that we need to have this module compatible with DS.

debezium-connect-rest-extension/pom.xml
rk3rn3r: Yes, this only activates the assembly profile for the `debezium-connect-rest-extension` build.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResourceIT.java
rk3rn3r: Filed https://issues.redhat.com/browse/DBZ-6946 to evaluate switching to `MongoDbDatabaseProvider` eventually.

debezium-connector-mongodb/src/test/resources/Dockerfile.rest.test
rk3rn3r: @jcechace I think we can only influence runtime config but we can't re-configure the image structure / content. Especially what we can't do imo is using artifacts that we just built inside those containers. That's the gap I am filling here.
Please file a Jira if there's something that we should do or evaluate to come closer to the solution you are having in mind. Please attach it as sub-task to this parent task/epic: https://issues.redhat.com/browse/DBZ-4395

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/RestExtensionTestInfrastructure.java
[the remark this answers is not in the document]
rk3rn3r: I filed https://issues.redhat.com/browse/DBZ-6947 for tackling those concerns. @jpechane @jcechace  Could you please add some description to the issue about what exactly you would like to see/achieve?

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
rk3rn3r: I have logged https://issues.redhat.com/browse/DBZ-6948 to address this issue and also taking care of managing the `enableFakeDnsIfRequired()` related code. It should only be activated inside the container startup methods on a system environment that requires it instead of having this enabled somewhere outside. Alternatively this could be done with JUnit5 env-specific configuration on Mac.

debezium-core/src/main/java/io/debezium/metadata/ConnectorDescriptor.java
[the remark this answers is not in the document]
rk3rn3r: ConnectorMetadata should be removed eventually. It can only be used in the SchemaGenerator use-case but not from the REST extension side. We should migrate to this mapping here instead to produce stable identifiers for the UI frontend / API side of things.

I filed https://issues.redhat.com/browse/DBZ-6950 for this to follow-up on the topic.

rk3rn3r:
@jpechane I think I filed all the concerns as sub-tasks in https://issues.redhat.com/browse/DBZ-4395 and re-based this PR.

Everyone please have a look, add descriptions or comments in those sub-tasks.
@ani-sha @Naros @mfvitale @jcechace @obabec

jpechane:
@rk3rn3r Applied, thanks!

PR #4748, opened 2023-08-07
DBZ-6767 Add missing type name

jpechane:
https://issues.redhat.com/browse/DBZ-6767

### DIFF
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TestHelper.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TestHelper.java
@@ -124,7 +124,7 @@ public final class TestHelper {
      */
     public static PostgresConnectorConfig.LogicalDecoder decoderPlugin() {
         final String s = System.getProperty(PostgresConnectorConfig.PLUGIN_NAME.name());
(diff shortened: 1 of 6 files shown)


novotnyJiri (commented):
The changes look good to me. But I would somehow resolve the the duplicity in the tests. Probably fix the ByteArray test so it actually tests byte array and not date array.

jpechane:
@novotnyJiri Second round ready

novotnyJiri (approved):
Other than probably a naming error LGTM.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TestHelper.java
```diff
@@ -353,6 +353,10 @@ protected static void dropPublication() {
         dropPublication(ReplicationConnection.Builder.DEFAULT_PUBLICATION_NAME);
     }
 
+    protected static void createPublicationForAllTables() {
+        dropPublication(ReplicationConnection.Builder.DEFAULT_PUBLICATION_NAME);
```
novotnyJiri, line 357: Is it correct, that method `createPublicationForAllTables()` actually does not create any publications and just drops default publication?

jpechane:
@novotnyJiri Thanks for the review!

novotnyJiri:
Glad I could help!

PR #4753, opened 2023-08-08
DBZ-6777 Fix wrong static notification and signals lists that lead to be wrongly shared between connectors instances

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6777


jpechane:
@mfvitale Applied, thanks

PR #4754, opened 2023-08-08
DBZ-6779 Fix version parsing of non-numerical chars

novotnyJiri:
https://issues.redhat.com/browse/DBZ-6779


jpechane:
@novotnyJiri I am not sure this is correct solution. It seems to me that with unfortunate settings the net result will just be that the tests will use `0` in the version for matching and will just ignore the execution of them. The correct solution should make sure that if version is not parseable then it should expect all tests should be run against it.
Or even better if the version scheme is know then just the `-<suffix>` should be removed and parsing done as it should

novotnyJiri:
@jpechane Second round ready. I checked the version naming of the mongo containers that we use and are taken from the `version.mongo.server` and they all follow the same version scheme just as you said.

jpechane:
@novotnyJiri Applied, thanks

novotnyJiri:
Thanks for the review!

PR #4755, opened 2023-08-08
DBZ-6782 Properly solve XMLTYPE data types during DDL changes

Naros:
https://issues.redhat.com/browse/DBZ-6782


jpechane:
@Naros Applied, thanks

PR #4757, opened 2023-08-08
DBZ-6784 Updated Redis (Stream) documentaiton with new parameter description

mfortunat:


jpechane:
@mfortunat Applied, thanks

PR #4759, opened 2023-08-09
DBZ-6786 Use custom row deserializers in case of binlog compression

rolevinks:


jpechane (changes requested):
@rolevinks Thanks, the solution LGTM! Unfortunately we need to test it as well.

Would it be possible to

- [ ] Create a test that fails with disable binlog compession
- [ ] Create a special annotation and rule to mark that test so it is executed only under condition
- [ ] Add a new profile to run test for enabled compression with MySQL compression enabled container
- [ ] Add that execution to `assembly` profile

That way we'd be able to verify the code works and make it part of our regular CI.

rolevinks:
Hi @jpechane 
Thanks for the feedback!

There already is a test that uses binlog compression, the `TransactionPayloadIT`, so I added some checks to that test. I'm not sure how to create a test that fails with the binlog compression disabled, because then the flow just would not go through the `TransactionPayloadDeserializer `, but it would go through the custom `RowDeserializers` and succeed.

Please let me know if this is enough as far as testing goes.

jpechane:
@rolevinks According to the Jira the issue is visible for UUID, so maybe adding a UUID column to the test would revela the problem?

rolevinks:
@jpechane Yes, that's what I did. I added a `binary(16)` column to the `TransactionPayloadIT` and checked that the record contains the correct value.

jpechane:
So, if the test is used on non-fixed branch and run against compression enabled MySQL then wil it fail?

rolevinks:
That is correct.

PR #4761, opened 2023-08-09
DBZ-6760 Outbox transformation can cause connector to crash

harveyyue:
https://issues.redhat.com/browse/DBZ-6760


jpechane:
@harveyyue Applied, thanks!

PR #4764, opened 2023-08-09
DBZ-6758 When using pgoutput in postgres connector, (+/-)Infinity is not supported in decimal values

harveyyue:
https://issues.redhat.com/browse/DBZ-6758


jpechane:
@harveyyue Thanks for the PR. The statement `io.debezium.connector.postgresql.AbstractRecordsProducerTest.INSERT_NUMERIC_TYPES_STMT` was expected to test infinite values and was passing before the change. Could you please check why it was the case? If this scenario is not cevered, could you please add another test?

harveyyue:
@jpechane OK

harveyyue:
@jpechane The IT `AbstractRecordsProducerTest ` not match following properties:
```
plugin.name=pgoutput
decimal.handling.mode=string
```

jpechane:
@harveyyue Makes sense. Apllied, thanks a lot!

PR #4765, opened 2023-08-09
DBZ-6176 Include records with lower bound SCN into result set

vjuranek:
Several records may share same SCN and when we exclude records with lower bound SCN from the result set, we may miss some records upon restart. Therefore, we should include also records with lower bound SCN.

https://issues.redhat.com/browse/DBZ-6778


vjuranek:
@Naros I not sure if we should include the test as it is now - I wasn't able to reproduce the issue with it. If you have any idea how to improve it, please let me know.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryBuilder.java
```diff
@@ -72,7 +72,7 @@ public static String build(OracleConnectorConfig connectorConfig) {
         query.append("FROM ").append(LOGMNR_CONTENTS_VIEW).append(" ");
 
         // These bind parameters will be bound when the query is executed by the caller.
-        query.append("WHERE SCN > ? AND SCN <= ?");
+        query.append("WHERE SCN >= ? AND SCN <= ?");
```
Naros, line 75: This has the potential to cause duplicate events to be emitted, I don't recommend doing this.  At least one of the bounds must always be exclusive.

vjuranek: Yes, I was thinking about it (with no clear conclusion) and as all the tests have passed I though it's fine. Now, looking into the test, we actually have only very few tests which check if there are any duplicates, so now I'm trying to write a test which would produce some duplicates (without much success so far).

Naros: So what I might recommend is in the test, start a transaction and write an insert without committing.  Then have the test wit for X mining loops using the `LogInterceptor` and after you've observed 5 or so, perform a final insert and commit.  From here you should be able to check that you only got 2 DML events rather than potentially 3 or more.

vjuranek: Sorry for the delay @Naros . I played with it little bit and wasn't able to reproduced the duplicates. I added the test so you can review it. However, I agree that the interval should be exclusive on one of the boundaries, so I moved inclusive interval into the second commit with the test for duplicates. If you are fine with the first commit, I'll remove the second commit (I include the second commit with the test just to share it with you, as I wonder if test is correct or not).

Naros: Let's plan to make this swap for 2.5.0.Alpha1 where we apply the inclusion on the lower end and exclusion on the upper-end (i.e. reverse the current logic) and test that throughout 2.5.  That way we get plenty of time to let the community validate the behavior.  wdyt?

vjuranek:
as noted by the user `guosc` in [this Zulip discussion](https://debezium.zulipchat.com/#narrow/stream/348250-community-oracle/topic/the.20duplicate.20scn.20cause.20data.20missing):

>  I have tried adjusting the code according to PR #4765, but it seems to lead to the generation of duplicate data. If a query interval segment has no data and only the data corresponding to startScn, it will cause a loop query and keep querying the repeated interval segment.

which seems to be good point. Moving to draft.

Naros:
@vjuranek I think we can close this, as I resolved this a part of the DBZ-8060 transaction sequence tracking work.

PR #4767, opened 2023-08-09
DBZ-6780 Relax lexicon rule for MySQL `REAL_LITERAL`

Naros:
https://issues.redhat.com/browse/DBZ-6780


jpechane:
@Naros Applied, thanks! Could you please prepare the upstream PR too?

PR #4770, opened 2023-08-11
DBZ-6689 make kafka signal consumer group configurable

vsantona:
Closes: https://issues.redhat.com/browse/DBZ-6689


jpechane:
@vsantona Applied, thanks!

PR #4776, opened 2023-08-12
DBZ-6808 Allow the embedded infinispan global configuration to be configurable

jeremy-l-ford:


jpechane:

PR #4778, opened 2023-08-13
DBZ-6720: handle unchanged toasted uuid array

rajdangwal:
Addresses https://issues.redhat.com/browse/DBZ-6720

The designated format for an unchanged toasted UUID array is a singleton array containing a UUID element that represents the encoded UUID of the placeholder string.
The uuid encoding of the default placeholder value `__debezium_unavailable_value` is `b68a35a7-17ad-35b3-af2a-ae46edb4545a`.


jpechane:
@rajdangwal Applied, thanks!

PR #4779, opened 2023-08-14
DBZ-6635 Send heartbeats also before processing first event

vjuranek:
We should send heartbeats no matter if there is already any DB record processed or not. This prevents situation when after the start there is no new record in the Db and Debezium is not sending neither DB events nor heartbeats.

https://issues.redhat.com/browse/DBZ-6635


vjuranek:
not sure about the fix, please see https://debezium.zulipchat.com/#narrow/stream/302533-dev/topic/Heartbeats

jpechane:
@vjuranek Applied, thanks. It seems to me from the very pas history that the condiiton related to `wal2json` plugin tha is no loner in place.

PR #4780, opened 2023-08-14
DBZ-6811 Send heartbeats when there are no updates in the DB

vjuranek:
Heartbeats should be sent when Debezium is not sending any events obatined from the DB. Send heartbeat before finishing itteration which doesn't contain any updates from the DB.

https://issues.redhat.com/browse/DBZ-6811


vjuranek:
not sure if this is actually the issue at all or intended behaviour, please see https://debezium.zulipchat.com/#narrow/stream/302533-dev/topic/Heartbeats

PR #238, opened 2017-05-31
DBZ-265 Initial attempt to use Arquillian Cube instead of Fabric8

jpechane:

PR #4781, opened 2023-08-14
DBZ-6567 SMT for handling timezone based fields

ani-sha:
https://issues.redhat.com/browse/DBZ-6567

To Do

- [x] Docs - https://github.com/debezium/debezium/pull/4819
- [x] KC timestamp field
- [x] Test cases

### DIFF
--- a/debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
@@ -94,7 +94,7 @@ public class TimezoneConverter<R extends ConnectRecord<R>> implements Transforma
     private List<String> excludeList;
     private static final String SOURCE = "source";
     private static final String TOPIC = "topic";
-    private static final Pattern TIMEZONE_PATTERN = Pattern.compile("^[+-]\\d{2}:\\d{2}$");
+    private static final Pattern TIMEZONE_OFFSET_PATTERN = Pattern.compile("^[+-]\\d{2}:\\d{2}(:\\d{2})?$");
     private static final Pattern LIST_PATTERN = Pattern.compile("^\\[(source|topic|[\".\\w\\s_]+):([\".\\w\\s_]+(?::[\".\\w\\s_]+)?(?:,|]$))+$");
     private final Map<String, Set<String>> topicFieldsMap = new HashMap<>();
     private final Map<String, Set<String>> tableFieldsMap = new HashMap<>();
@@ -226,7 +226,7 @@ public class TimezoneConverter<R extends ConnectRecord<R>> implements Transforma
     }
 
     private boolean validateTimezoneString() {
-        if (TIMEZONE_PATTERN.matcher(convertedTimezone).matches()) {
+        if (TIMEZONE_OFFSET_PATTERN.matcher(convertedTimezone).matches()) {
             return true;
         }
         else if (ZoneId.getAvailableZoneIds().contains(convertedTimezone)) {
@@ -268,14 +268,14 @@ public class TimezoneConverter<R extends ConnectRecord<R>> implements Transforma
                 Instant microInstant = Instant.ofEpochSecond(microTimestamp / 1_000_000, (microTimestamp % 1_000_000) * 1_000);
                 LocalDateTime microLocalDateTime = microInstant.atOffset(ZoneOffset.UTC).toLocalDateTime();
                 zoneOffset = zoneId.getRules().getOffset(microLocalDateTime);
-                updatedFieldValue = microLocalDateTime.atOffset(zoneOffset).toInstant().toEpochMilli() * 1_000;
+                updatedFieldValue = microLocalDateTime.toEpochSecond(zoneOffset) * 1_000_000 + microLocalDateTime.getNano() / 1_000;
                 break;
             case NanoTimestamp.SCHEMA_NAME:
                 long nanoTimestamp = (long) fieldValue;
                 Instant nanoInstant = Instant.ofEpochSecond(nanoTimestamp / 1_000_000_000, (nanoTimestamp % 1_000_000_000));
                 LocalDateTime nanoLocalDateTime = nanoInstant.atOffset(ZoneOffset.UTC).toLocalDateTime();
                 zoneOffset = zoneId.getRules().getOffset(nanoLocalDateTime);
-                updatedFieldValue = nanoLocalDateTime.atOffset(zoneOffset).toInstant().toEpochMilli() * 1_000_000;
+                updatedFieldValue = nanoLocalDateTime.toEpochSecond(zoneOffset) * 1_000_000_000 + nanoLocalDateTime.getNano();
                 break;
             case Timestamp.SCHEMA_NAME:
                 Instant instant = Instant.ofEpochMilli((long) fieldValue);
--- a/debezium-core/src/test/java/io/debezium/transforms/TimezoneConverterTest.java
+++ b/debezium-core/src/test/java/io/debezium/transforms/TimezoneConverterTest.java
@@ -62,7 +62,7 @@ public class TimezoneConverterTest {
         before.put("id", (byte) 1);
         before.put("name", "Srikanth");
         before.put("order_date_micros", 1529507596945104L);
-        before.put("order_date_nanos", 1531481025340000000L);
+        before.put("order_date_nanos", 1531481025340000104L);
         before.put("order_date_timestamp", 1514908810123L);
         before.put("order_date_zoned_timestamp", "2018-01-02T11:15:30.123456789+00:00");
         before.put("order_date_zoned_time", "11:15:30.123456789+00:00");
@@ -91,8 +91,8 @@ public class TimezoneConverterTest {
         final Struct transformedValue = (Struct) transformedRecord.value();
         final Struct transformedAfter = transformedValue.getStruct(Envelope.FieldName.AFTER);
 
-        assertThat(transformedAfter.get("order_date_micros")).isEqualTo(1529487796945000L);
-        assertThat(transformedAfter.get("order_date_nanos")).isEqualTo(1531461225340000000L);
+        assertThat(transformedAfter.get("order_date_micros")).isEqualTo(1529487796945104L);
+        assertThat(transformedAfter.get("order_date_nanos")).isEqualTo(1531461225340000104L);
         assertThat(transformedAfter.get("order_date_timestamp")).isEqualTo(1514889010123L);
         assertThat(transformedAfter.get("order_date_zoned_timestamp")).isEqualTo("2018-01-02T16:45:30.123456789+05:30");
         assertThat(transformedAfter.get("order_date_zoned_time")).isEqualTo("16:45:30.123456789+05:30");
@@ -457,7 +457,7 @@ public class TimezoneConverterTest {
     @Test
     public void testExcludeListWithNoPrefix() {
         final Map<String, String> props = new HashMap<>();
-        props.put("converted.timezone", "+08:00");
+        props.put("converted.timezone", "+08:00:00");
         props.put("exclude.list", "db.server1.table1:order_time");
         converter.configure(props);
 
@@ -794,8 +794,8 @@ public class TimezoneConverterTest {
 
     @Test
     public void testDayLightSavings() {
-        String withDayLightSavingsInstant = "2014-12-22T10:15:30Z"; // DST is in effect
-        String withoutDayLightSavingsInstant = "2015-06-22T10:15:30Z"; // DST is not in effect
+        String withDayLightSavingsInstant = "2014-06-22T10:15:30Z"; // DST is in effect
+        String withoutDayLightSavingsInstant = "2014-12-22T10:15:30Z"; // DST is not in effect
         Clock clockWithDayLightSavings = Clock.fixed(Instant.parse(withDayLightSavingsInstant), ZoneId.of("UTC"));
         Clock clockWithoutDayLightSavings = Clock.fixed(Instant.parse(withoutDayLightSavingsInstant), ZoneId.of("UTC"));
 
@@ -842,7 +842,7 @@ public class TimezoneConverterTest {
         final Struct transformedValue = (Struct) transformedRecord.value();
         final Struct transformedAfter = transformedValue.getStruct(Envelope.FieldName.AFTER);
 
-        assertThat(transformedAfter.get("order_date")).isEqualTo("2014-12-22T11:15:30+01:00");
-        assertThat(transformedAfter.get("shipping_date")).isEqualTo("2015-06-22T12:15:30+02:00");
+        assertThat(transformedAfter.get("order_date")).isEqualTo("2014-06-22T12:15:30+02:00");
+        assertThat(transformedAfter.get("shipping_date")).isEqualTo("2014-12-22T11:15:30+01:00");
     }
 }


debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -0,0 +1,388 @@
...
+                ZonedDateTime zdtWithZone = zdt.withZoneSameInstant(zoneId);
+                updatedFieldValue = ZonedTimestamp.toIsoString(zdtWithZone, null, 0);
+                break;
+            case ZonedTime.SCHEMA_NAME:
+                OffsetTime offsetTime = OffsetTime.parse((String) fieldValue, DateTimeFormatter.ISO_OFFSET_TIME);
+                OffsetTime offsetTimeWithZone = offsetTime.withOffsetSameInstant(zoneId.getRules().getOffset(Instant.now()));
+                updatedFieldValue = ZonedTime.toIsoString(offsetTimeWithZone, null);
+                break;
+            case MicroTimestamp.SCHEMA_NAME:
+                Instant microInstant = Instant.ofEpochMilli((long) fieldValue / 1_000);
```
jpechane, line 208: Precision can be lost, please use two parameter `ofEpochMilli` to provide micros and nanos.

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -0,0 +1,388 @@
...
+                LocalDateTime microLocalDateTime = microInstant.atOffset(ZoneOffset.UTC).toLocalDateTime();
+                updatedFieldValue = microLocalDateTime.atZone(zoneId).toInstant().toEpochMilli() * 1_000;
+                break;
+            case NanoTimestamp.SCHEMA_NAME:
+                Instant nanoInstant = Instant.ofEpochMilli((long) fieldValue / 1_000_000);
```
jpechane, line 213: Precision can be lost, please use two parameter `ofEpochMilli` to provide micros and nanos.

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -0,0 +1,388 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(TimezoneConverter.class);
+
+    private static final Field CONVERTED_TIMEZONE = Field.create("converted.timezone")
+            .withDisplayName("Converted Timezone")
+            .withType(ConfigDef.Type.STRING)
+            .withWidth(ConfigDef.Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.HIGH)
+            .withDescription("The timezone to convert to. " +
+                    "If not specified, the connector will use the default UTC timezone.")
+            .withDefault(ZoneOffset.UTC.toString());
```
jpechane, line 52: The original value should be in UTC so converting to UTC does not make sense, Maybe there should be no default value.

jpechane: Thinking about that a bit more, I'd really like to avoid using timezones completely and use timeshifts.
The parameter will accept value in format postiove or negative `hh:mm`. The value will be parsed and method `plus` on Instant/ZonedTimestamp etc. will be called to offset the time to appropriate value.

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -0,0 +1,388 @@
...
+    @Override
+    public ConfigDef config() {
+        final ConfigDef config = new ConfigDef();
+        Field.group(config, null, CONVERTED_TIMEZONE, INCLUDE_LIST, EXCLUDE_LIST);
+        return config;
+    }
+
+    @Override
+    public R apply(R record) {
+        if (record.value() == null || !(record.value() instanceof Struct)) {
```
jpechane, line 105: Please see `io.debezium.transforms.SmtManager.isValidEnvelope`

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -0,0 +1,388 @@
...
+        }
+        if (after != null) {
+            handleValueForFields(Envelope.FieldName.AFTER, after, listType, matchName, fieldName);
+        }
+    }
+
+    private void handleValueForFields(String structType, Struct structValue, Type listType, String matchName, String fieldName) {
+        Schema schema = structValue.schema();
+        for (org.apache.kafka.connect.data.Field field : schema.fields()) {
+            if ((listType == Type.INCLUDE && field.name().equals(fieldName))
```
jpechane, line 252: Please reorder the condition to have `listType == Type.ALL` first as fast resolution.

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -0,0 +1,388 @@
...
+                matchName = parts[0];
+                fieldName = parts[1];
+            }
+        }
+        return new FieldItem(prefix, matchName, fieldName);
+    }
+
+    private void handleInclude(Struct value) {
+        for (String item : includeList) {
+            FieldItem includeItem = parseItem(item);
```
jpechane, line 359: Is it possible to do all parsing in advance in config method and prepare filtering logic/structures in advance?

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -0,0 +1,388 @@
...
+                    + "`<matchname>:<fieldnames>`:: Uses a heuristic matching algorithm to matches the source information block table name if the source information block exists, otherwise matches against the topic name. The conversion is applied only to to the specified field name. ");
+
+    private static final int CACHE_SIZE = 64;
+    private final BoundedConcurrentHashMap<String, Map<String, Object>> cache = new BoundedConcurrentHashMap<>(CACHE_SIZE);
+    private String convertedTimezone;
+    private List<String> includeList;
+    private List<String> excludeList;
+    private static final String SOURCE = "source";
+    private static final String TOPIC = "topic";
+    private final List<String> topicsAndTables = new ArrayList<>();
```
jpechane, line 87: This should be `Set`. Also is not it safer to have separate attribute for tables and topics?

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -0,0 +1,388 @@
...
+                    + "`source:<tablename>`:: Matches only Debezium change events with a source information block with the specified table name. All time-based fields will be excluded. "
+                    + "`source:<tablename>:<fieldnames>`:: Matches only Debezium change events with a source information with the specified table name. Only the specified field name will be excluded. "
+                    + "`topic:<topicname>`:: Matches the specified topic name, excluding all time-based fields. "
+                    + "`topic:<topicname>:<fieldnames>`:: Matches the specified topic name, excluding only the specified field name. "
+                    + "`<matchname>:<fieldnames>`:: Uses a heuristic matching algorithm to matches the source information block table name if the source information block exists, otherwise matches against the topic name. The conversion is applied only to to the specified field name. ");
+
+    private static final int CACHE_SIZE = 64;
+    private final BoundedConcurrentHashMap<String, Map<String, Object>> cache = new BoundedConcurrentHashMap<>(CACHE_SIZE);
```
jpechane, line 81: Is the cache really needed?

jpechane (commented):
@ani-sha Thanks for the PR. I left few comments asking some question and proposing additional improvements that might lead to few simplifications.

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -0,0 +1,470 @@
...
+
+    private enum Type {
+        ALL,
+        INCLUDE,
+        EXCLUDE
+    }
+
+    private Object getTimestampWithTimezone(String schemaName, Object fieldValue) {
+        Object updatedFieldValue = fieldValue;
+        ZoneId zoneId = ZoneId.of(convertedTimezone);
```
mfvitale, line 229: How this behave with the DST? Is it correctly managed?

mfvitale: For example in Italy we usually have the UTC+1 offeset but from the 25 March to 30 October we move to DST that is UTC+2. I am wondering if using a fixed offeset is right. https://www.baeldung.com/java-zone-offset

mfvitale: I think yes since without think about DST can lead to wrong date conversion. 

You are using ZoneId.of("+01:00") and as for Java [docs](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneId.html#of-java.lang.String-) in this case a [ZoneOffset.of(String)](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneOffset.html#of-java.lang.String-) is used to parse the Zone ID. 

From ZoneOffset docs
>Different parts of the world have different time-zone offsets. The rules for how offsets vary by place and time of year are captured in the [ZoneId](https://docs.oracle.com/javase/8/docs/api/java/time/ZoneId.html) class.
For example, Paris is one hour ahead of Greenwich/UTC in winter and two hours ahead in summer. The ZoneId 
instance for Paris will reference two ZoneOffset instances - a +01:00 instance for winter, and a +02:00 instance for 
summer.

I think thath you need to use the same approach used in line #238 
```java
zoneId.getRules().getOffset(Instant.now())
```
also for the other case in the switch. 

I suggest to create specific test to see if it works.

jpechane: @mfvitale Thanks for raising the concern but I'd prefer to keep it off this PR and make a subseqent one.
We need to evaluate first if there is any impact on DST on values as theu come from Debezium core. If it is already accounted fro then it would be unnecessary to handle it here.

mfvitale: How we store date in Debezium events? In case we are storing in Debezium zoned date consider the case when you want to move from a timezone that doesn't have the DST to another one that has it. In this case this cannot work.

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -0,0 +1,470 @@
...
+                LocalDateTime nanoLocalDateTime = nanoInstant.atOffset(ZoneOffset.UTC).toLocalDateTime();
+                updatedFieldValue = nanoLocalDateTime.atZone(zoneId).toInstant().toEpochMilli() * 1_000_000;
+                break;
+            case Timestamp.SCHEMA_NAME:
+                Instant instant = Instant.ofEpochMilli((long) fieldValue);
+                LocalDateTime localDateTime = instant.atOffset(ZoneOffset.UTC).toLocalDateTime();
+                updatedFieldValue = localDateTime.atZone(zoneId).toInstant().toEpochMilli();
+                break;
+            case org.apache.kafka.connect.data.Timestamp.LOGICAL_NAME:
+                // to-do
```
Naros, line 259: Let's not forget this implementation.

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -0,0 +1,470 @@
...
+public class TimezoneConverter<R extends ConnectRecord<R>> implements Transformation<R> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(TimezoneConverter.class);
+
+    private static final Field CONVERTED_TIMEZONE = Field.create("converted.timezone")
+            .withDisplayName("Converted Timezone")
+            .withType(ConfigDef.Type.STRING)
+            .withWidth(ConfigDef.Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.HIGH)
+            .withValidation(Field::isRequired)
+            .withDescription("A string of time shift value representing timezone in the format of +/-hh:mm, e.g. -08:00 or +05:30.");
```
Naros, line 53: I've alluded to this in my other comment; however, I'm concerned here that the use of `+/-HH:MM` has the potential to be confusing to users who deal with daylight savings time.  For example, if I am configuring this during the summer months and live in Central Europe, do I configure this as `+01:00` still, which represents my offset during the winter, or do I use `+02:00`, which is the current summer offset, or vice versa.

This seems ambiguous, so we'll need to be explicit in the documentation.   My concern is that no matter how clear we make this in the documentation, it will likely be a detail that is overlooked by users and could lead to a series of questions every time DST starts.

Naros (commented):
I've added some comments.

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
ani-sha: Thanks @mfvitale and @Naros for you inputs and concerns regarding DST. As discussed with @jpechane, I would be adding the support for both timezone names and timeshift values `+/-HH:MM`. It would be best suited for users and avoid unnecessary confusion.

ani-sha: @mfvitale  @Naros How do we handle the tests when the DST is in effect?

debezium-core/src/test/java/io/debezium/transforms/TimezoneConverterTest.java
```diff
@@ -788,6 +790,59 @@ public void testExcludeListWithMultipleFields() {
         assertThat(transformedInventoryAfter.get("name")).isEqualTo("Amy Rose");
         assertThat(transformedInventoryAfter.get("shipping_time")).isEqualTo("22:19:25+03:00");
         assertThat(transformedInventoryAfter.get("order_time")).isEqualTo("10:19:25+00:00");
+    }
+
+    @Test
+    public void testDayLightSavings() {
```
ani-sha, line 796: @mfvitale Added this test case. Please check it.

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -0,0 +1,502 @@
...
+                OffsetTime offsetTime = OffsetTime.parse((String) fieldValue, DateTimeFormatter.ISO_OFFSET_TIME);
+                OffsetTime offsetTimeWithZone = offsetTime.withOffsetSameInstant(zoneOffset);
+                updatedFieldValue = ZonedTime.toIsoString(offsetTimeWithZone, null);
+                break;
+            case MicroTimestamp.SCHEMA_NAME:
+                long microTimestamp = (long) fieldValue;
+                Instant microInstant = Instant.ofEpochSecond(microTimestamp / 1_000_000, (microTimestamp % 1_000_000) * 1_000);
+                LocalDateTime microLocalDateTime = microInstant.atOffset(ZoneOffset.UTC).toLocalDateTime();
+                zoneOffset = zoneId.getRules().getOffset(microLocalDateTime);
+                updatedFieldValue = microLocalDateTime.atOffset(zoneOffset).toInstant().toEpochMilli() * 1_000;
```
jpechane, line 271: @ani-sha You should use combination of `getEpochSecond` and `getNano` otherwise you are losing precision.

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -0,0 +1,502 @@
...
+                break;
+            case NanoTimestamp.SCHEMA_NAME:
+                long nanoTimestamp = (long) fieldValue;
+                Instant nanoInstant = Instant.ofEpochSecond(nanoTimestamp / 1_000_000_000, (nanoTimestamp % 1_000_000_000));
+                LocalDateTime nanoLocalDateTime = nanoInstant.atOffset(ZoneOffset.UTC).toLocalDateTime();
+                zoneOffset = zoneId.getRules().getOffset(nanoLocalDateTime);
+                updatedFieldValue = nanoLocalDateTime.atOffset(zoneOffset).toInstant().toEpochMilli() * 1_000_000;
```
jpechane, line 278: Same as above

debezium-core/src/test/java/io/debezium/transforms/TimezoneConverterTest.java
```diff
@@ -0,0 +1,848 @@
...
+                .field("order_date_zoned_time", ZonedTime.builder().optional().build())
+                .build();
+
+        final Struct before = new Struct(recordSchema);
+        final Struct source = new Struct(sourceSchema);
+
+        before.put("id", (byte) 1);
+        before.put("name", "Srikanth");
+        before.put("order_date_micros", 1529507596945104L);
+        before.put("order_date_nanos", 1531481025340000000L);
```
jpechane, line 65: Do not se rounded value as it masks loss precision.

debezium-core/src/test/java/io/debezium/transforms/TimezoneConverterTest.java
```diff
@@ -0,0 +1,848 @@
...
+                "db.server1.table1",
+                envelope.schema(),
+                payload);
+
+        final SourceRecord transformedRecord = converter.apply(record);
+
+        final Struct transformedValue = (Struct) transformedRecord.value();
+        final Struct transformedAfter = transformedValue.getStruct(Envelope.FieldName.AFTER);
+
+        assertThat(transformedAfter.get("order_date_micros")).isEqualTo(1529487796945000L);
```
jpechane, line 94: The assert should not ignore microsecond difference.

debezium-core/src/test/java/io/debezium/transforms/TimezoneConverterTest.java
```diff
@@ -0,0 +1,848 @@
...
+        final Struct transformedInventoryValue = (Struct) transformedInventoryRecord.value();
+        final Struct transformedInventoryAfter = transformedInventoryValue.getStruct(Envelope.FieldName.AFTER);
+        assertThat(transformedInventoryAfter.get("name")).isEqualTo("Amy Rose");
+        assertThat(transformedInventoryAfter.get("shipping_time")).isEqualTo("22:19:25+03:00");
+        assertThat(transformedInventoryAfter.get("order_time")).isEqualTo("10:19:25+00:00");
+    }
+
+    @Test
+    public void testDayLightSavings() {
+        String withDayLightSavingsInstant = "2014-12-22T10:15:30Z"; // DST is in effect
```
jpechane, line 797: These two values are strange - december is definitely not DST and June is DST.

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -0,0 +1,502 @@
...
+                    + "`topic:<topicname>:<fieldnames>`:: Matches the specified topic name, excluding only the specified field name. "
+                    + "`<matchname>:<fieldnames>`:: Uses a heuristic matching algorithm to matches the source information block table name if the source information block exists, otherwise matches against the topic name. The conversion is applied only to to the specified field name. ");
+
+    private SmtManager<R> smtManager;
+    private String convertedTimezone;
+    private List<String> includeList;
+    private List<String> excludeList;
+    private static final String SOURCE = "source";
+    private static final String TOPIC = "topic";
+    private static final Pattern TIMEZONE_PATTERN = Pattern.compile("^[+-]\\d{2}:\\d{2}$");
```
jpechane, line 97: Please check JavaDoc for `java.time.ZoneOffset.of(String)` to see what are allowed values. From my PoV we can accept simiplified rules where colons are mandatory, hours and minutes are mandatory and there can be optional seconds.

jpechane (changes requested):

jpechane:
@ani-sha We are almost there! Please make sure there is no data loss for sub-milliseconds transformations. Add support for seconds in the offset and double check the DST test case. When address this is good to go. This SMT is really going to be useful.

debezium-core/src/test/java/io/debezium/transforms/TimezoneConverterTest.java
ani-sha: Sorry for the confusion, values got interchanged, otherwise works fine! :)

PR #4782, opened 2023-08-14
DBZ-6798 Expose SCN-based metrics as `BigInteger`

Naros:
https://issues.redhat.com/browse/DBZ-6798

The SCN data types were previously exposed as `String` types, which is not consumable by Grafana and Prometheus. By using `BigInteger`, we can now make these accessible on dashboards.


jpechane:
@Naros Applied, thanks!

PR #4783, opened 2023-08-14
DBZ-6801 Introduce JDBC sink option `dialect.sqlserver.identity.insert`

Naros:
https://issues.redhat.com/browse/DBZ-6801

Documents https://github.com/debezium/debezium-connector-jdbc/pull/27

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -703,8 +703,8 @@ The default is `public`; however, if the PostGIS extension was installed in anot
 
 |[[jdbc-property-dialect-sqlserver-identity-insert]]<<jdbc-property-dialect-sqlserver-identity-insert, `+dialect.sqlserver.identity.insert+`>>
 |`false`
-|Specifies whether the connector automatically enables identity inserts before an _insert_ or _upsert_ operation and is disabled immediately afterward.
-The default is `false`, meaning an insert or upsert into a SQL Server table that has an `IDENTITY` column throws a SQL exception.
+|Specifies whether the connector automatically sets an `IDENTITY_INSERT` before an `INSERT` or `UPSERT` operation into the identity column of SQL Server tables, and then unsets it immediately after the operation.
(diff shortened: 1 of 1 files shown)


documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -701,6 +701,10 @@ The placeholder, `${topic}`, is replaced by the topic name.
 |Specifies the schema name where the PostgreSQL PostGIS extension is installed.
 The default is `public`; however, if the PostGIS extension was installed in another schema, this property should be used to specify the alternate schema name.
 
+|[[jdbc-property-dialect-sqlserver-identity-insert]]<<jdbc-property-dialect-sqlserver-identity-insert, `+dialect.sqlserver.identity.insert+`>>
+|`false`
+|Specifies whether the connector automatically enables identity inserts before an _insert_ or _upsert_ operation and is disabled immediately afterward.
```
roldanbob, line 706:
```suggestion
|Specifies whether the connector automatically sets an `IDENTITY_INSERT` before an `INSERT` or `UPSERT` operation into the identity column of SQL Server tables , and then unsets it immediately after the operation.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -701,6 +701,10 @@ The placeholder, `${topic}`, is replaced by the topic name.
...
+The default is `false`, meaning an insert or upsert into a SQL Server table that has an `IDENTITY` column throws a SQL exception.
```
roldanbob, line 707:
```suggestion
When the default setting (`false`) is in effect, an `INSERT` or `UPSERT` operation into the `IDENTITY` column of a table results in a SQL exception.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
roldanbob: The convention in the SQL Server documentation seems to be to capitalize SQL commands, so I followed that model.

roldanbob (changes requested):

Naros:
Applied to main (=2.4)

PR #4784, opened 2023-08-15
DBZ-6774 MongoDB New Document State Extraction: nonexistent field for add.headers

harveyyue:
https://issues.redhat.com/browse/DBZ-6774

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/ExtractNewDocumentState.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/ExtractNewDocumentState.java
@@ -531,7 +531,7 @@ public class ExtractNewDocumentState<R extends ConnectRecord<R>> implements Tran
             return SchemaUtil.copySchemaBasics(schemaField.schema()).optional().build();
         }
 
-        boolean exists(Schema originalRecordSchema) {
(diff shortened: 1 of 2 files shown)


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/ExtractNewDocumentState.java
```diff
@@ -352,8 +352,19 @@ private Headers makeHeaders(List<FieldReference> additionalHeaders, Struct origi
...
             }
-            headers.add(fieldReference.getNewFieldName(), fieldReference.getValue(originalRecordValue),
-                    fieldReference.getSchema(originalRecordValue.schema()));
+            try {
+                Schema fieldSchema = fieldReference.getSchema(originalRecordValue.schema());
+                headers.add(fieldReference.getNewFieldName(), fieldReference.getValue(originalRecordValue), fieldSchema);
+            }
+            catch (Exception e) {
+                // skip unknown field from origin record value struct
+                if (e.getMessage().startsWith("Unexpected field name:")) {
```
jpechane, line 361: @harveyyue I don't think this is correct. If the field is not supported it should be clear from the schema of the struct so if the field is not in schema it should be just ignored?

jpechane:
@harveyyue Applied, thanks!

PR #4786, opened 2023-08-15
Improve DB2 example commands; Update db2.adoc

andreas-ibm:
The Procedure for compiling the capture agent didn't show how to correctly invoke `bldrtn` as it assumed it was in the same directory as the Debezium management UDFs, which is isnt.  I've added instructions for getting it on PATH and also an example later on for how to invoke SQL commands from the command line once they're already logged on.  (I didn't update the following two SQL statements, as I'd assume that an intelligent viewer could learn from the first example!)


jpechane:
@andreas-ibm Thanks, merging via https://github.com/debezium/debezium/pull/4787

PR #4790, opened 2023-08-15
DBZ-6615 Add OldestScnAgeInMilliseconds LogMiner based metric.

Naros:
https://issues.redhat.com/browse/DBZ-6615


jpechane (changes requested):
@Naros It seems the test failure is related - in unit tests

jpechane:
@Naros Applied, thanks

PR #4791, opened 2023-08-16
DBZ-5337 Add Oracle XMLTYPE table support

Naros:
https://issues.redhat.com/browse/DBZ-5337

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
@@ -227,7 +227,10 @@ public class OracleConnection extends JdbcConnection {
                 // filter nested tables
                 "and nested = 'NO'" +
                 // filter parent tables of nested tables
-                "and table_name not in (select PARENT_TABLE_NAME from ALL_NESTED_TABLES)";
+                "and table_name not in (select PARENT_TABLE_NAME from ALL_NESTED_TABLES)" +
+                "union " +
+                "select owner, table_name from all_xml_tables " +
+                "where table_name not like 'XDB$%' and table_name not like 'BIN$%'";
 
         Set<TableId> tableIds = new HashSet<>();
         query(query, (rs) -> {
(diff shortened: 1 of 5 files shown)


Naros:
So @jpechane, I'm unsure whether this will be feasible for Debezium 2.4. Overall, support for `XMLTYPE`-based tables is quite complex after digging into the details.  

Immediately at the start, XMLTYPE tables are represented by two columns:

1. `SYS_NC_OID$` (hidden column)
2. `SYS_NC_ROWINFO$` (data)

The hidden column is the main problem right now because of how LogMiner materializes DML operations.  For an insert into a normal table, it would be constructed as follows:
```
INSERT INTO "SCHEMA"."TABLE" ("COL1", "COL2") VALUES ('V1', 'V2');
```
However, for XMLTYPE-based tables with CLOB storage, they use the format:
```
INSERT INTO "SCHEMA"."TABLE" VALUES (EMPTY_CLOB());
INSERT INTO "SCHEMA"."TABLE" VALUES (XMLTYPE('...'));
```

We can immediately see that CLOB storage XML tables follow a similar sequence to LOB-based columns and so we have to add logic to merge these two inserts into a single operation.  Now, if we table that aspect for now and focus on the SQL details, we can see that LogMiner only provides the XML data here, the data that is written into `SYS_NC_ROWINFO$`.

If we then look at updates for example, they're constructed as follows:
```
UPDATE "SCHEMA".TABLE" a SET a."SYS_NC_ROWINFO$" = XMLTYPE('...') WHERE a."SYS_NC_OID$" = RAWTOHEX('...');
```

As we can see the `SYS_NC_OID$` field is included here, which is a hidden column but it acts as the table's primary key as far as Oracle is concerned.  This presents a bit of a conundrum because:

1. Inserts do not provide this column value, there is no way to resolve it from the provided LogMiner data.
2. Updates and deletes rely on `SYS_NC_OID$` to target the specific row with the reconstructed SQL.

There is frankly too much inconsistency here between the redo reconstruction, the JDBC metadata, and what a consumer logically would need in order to handle effectively working with updates and deletes in my opinion.  Ideally, I'd prefer to include this `SYS_NC_OID$` column as the primary key; however, given that the insert operations don't included it nor would we have a reliable way to fetch it -- I'm not sure whether that's enough to say this is just not something we can reliably support.  wdyt?

jpechane:
@Naros Do I understan it correctly that for `INSERT` we are able to get the row PK but for updates we get only `SYS_NC_OID$` and there is now way how to map them together without going out-of-band metadata reads?

Naros:

```
git commit -s
```

PR #4793, opened 2023-08-17
DBZ-6724 Modify JOIN clause to support unstructured joins

ani-sha:
https://issues.redhat.com/browse/DBZ-6724

Upstream - https://github.com/antlr/grammars-v4/pull/3668


ani-sha:
@Naros Also added an example for `STRAIGHT_JOIN`. Need any other examples?

jpechane:
@ani-sha Applied, thanks!

PR #4804, opened 2023-08-23
DBZ-6824 trim default value for the BIGINT and SMALLINT types when parsing MySQL DDL

ruanhang1993:
[Fix DBZ-6824.](https://issues.redhat.com/browse/DBZ-6824)

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlDefaultValueConverter.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlDefaultValueConverter.java
@@ -59,10 +59,10 @@ public class MySqlDefaultValueConverter implements DefaultValueConverter {
 
     private static final Pattern CHARSET_INTRODUCER_PATTERN = Pattern.compile("^_[A-Za-z0-9]+'(.*)'$");
 
+    // Default values of these data types and number data types need to be trimmed.
     @Immutable
-    private static final Set<Integer> TRIM_DATA_TYPES = Collect.unmodifiableSet(Types.TINYINT, Types.INTEGER,
-            Types.DATE, Types.TIMESTAMP, Types.TIMESTAMP_WITH_TIMEZONE, Types.TIME, Types.BOOLEAN, Types.BIT,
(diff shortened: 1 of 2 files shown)


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlDefaultValueConverter.java
```diff
@@ -62,7 +62,7 @@ public class MySqlDefaultValueConverter implements DefaultValueConverter {
     @Immutable
     private static final Set<Integer> TRIM_DATA_TYPES = Collect.unmodifiableSet(Types.TINYINT, Types.INTEGER,
             Types.DATE, Types.TIMESTAMP, Types.TIMESTAMP_WITH_TIMEZONE, Types.TIME, Types.BOOLEAN, Types.BIT,
-            Types.NUMERIC, Types.DECIMAL, Types.FLOAT, Types.DOUBLE, Types.REAL);
+            Types.NUMERIC, Types.DECIMAL, Types.FLOAT, Types.DOUBLE, Types.REAL, Types.BIGINT, Types.SMALLINT);
```
jpechane, line 65: Is now `TRIM_DATA_TYPES` a supreset of complete `NUMBER_DATA_TYPES`? If it is it probably makes sense to not repeat all the values but instead either make `TRIM_DATA_TYPES` of `NUMBER_DATA_TYPES` and the additional values or remove numeric types from this var and modify `convert` method to use both attributes.

jpechane:
@ruanhang1993 Thanks for the PR, I left a recommendation for a small change reducing code duplication.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlDefaultValueConverter.java
ruanhang1993: Thanks for the suggestions. I have updated the PR and enriched the test.

Naros (approved):
Thanks, @ruanhang1993, the code changes look much better!

Naros:

PR #4807, opened 2023-08-24
DBZ-6839 Fix PostgresConnectorIT#shouldAddNewFieldToSourceInfo

vjuranek:
PostgresConnectorIT#shouldAddNewFieldToSourceInfo fails only when run together with other tests and the failure is random. It seem there is a caching issue in Apicuro, when `test_server.s1.a-value` artifact references `io.debezium.connector.postgresql.Source` version 1, which hasnt't `newField` field and this reference is used also in `shouldAddNewFieldToSourceInfo` where artifact with version 2 should be used. Using dedicated table and thus creating new artifact in Apucurio should fix this issue.

Also remove unused variable from `CustomPostgresSourceInfoStructMaker`.

https://issues.redhat.com/browse/DBZ-6839


obabec (approved):

jpechane:
@vjuranek Applied, thanks

PR #4808, opened 2023-08-28
DBZ-6841 Fix randomly failing BaseSourceTaskTest#verifyTaskRestartsSuccessfully

vjuranek:
Test was passing locally and in the CI as the start was very fast and after the first start retriable wait condition wasn't met, so we actually sleep and while the sleep we doesn't increment the counter and the state is still `RESTARTING`. On slower machines when start takes longer time and we don't wait, the test is actualy wrong. The correct test either should remove last assert for `RESTARTING` state or increase number of failures and add a sleep time after initial start, so that there is no retry wait time.
This PR implements the later option.

https://issues.redhat.com/browse/DBZ-6841


rk3rn3r:
Thx a lot for the quick evaluation of the root cause and the quick fix.

PR #4810, opened 2023-08-29
DBZ-6840 Wrong filtered comments

harveyyue:
https://issues.redhat.com/browse/DBZ-6840


jpechane:
@harveyyue @Naros Applied, thanks

PR #4811, opened 2023-08-29
DBZ-6742 change skipped change event log to debug

vsantona:
Closes: https://issues.redhat.com/browse/DBZ-6843


Naros:
@vsantona can you please adjust the commit message to use issue key DBZ-6843 rather than DBZ-6742 as the latter seems to be unrelated to your change.  Thanks!

PR #4812, opened 2023-08-29
DBZ-6843 change skipped change event log to debug

vsantona:
Closes: https://issues.redhat.com/browse/DBZ-6843


jpechane:
@vsantona Applied, thanks

PR #4815, opened 2023-08-29
DBZ-6844 Add truncate column config to Vitess connector docs

twthorn:
Add docs on this config since it's fully supported for vitess connector.


jpechane:
@twthorn Applied, thanks!

PR #4824, opened 2023-09-01
DBZ-6828 Ad-hoc blocking snaps trigger emits schema changes of all tables

mfvitale:
With these changes the blocking snapshot will generate schema events only for signaled tables

closes: https://issues.redhat.com/browse/DBZ-6828


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
```diff
@@ -350,8 +350,10 @@ protected void readTableStructure(ChangeEventSourceContext sourceContext,
                 .collect(Collectors.groupingBy(TableId::catalog, LinkedHashMap::new, Collectors.toList()));
         final Set<String> databases = tablesToRead.keySet();
 
-        // Record default charset
-        addSchemaEvent(snapshotContext, "", connection.setStatementFor(connection.readMySqlCharsetSystemVariables()));
+        if (!snapshottingTask.isBlocking()) {
+            // Record default charset
+            addSchemaEvent(snapshotContext, "", connection.setStatementFor(connection.readMySqlCharsetSystemVariables()));
```
mfvitale, line 355: @jpechane I noted that for Charset statement no database name is provided. Is it right?

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/BlockingSnapshotIT.java
```diff
@@ -146,4 +149,20 @@ private TableId tableNameId(String table) {
...
 
+    @Override
+    protected int expectedDdlsCount() {
+        return 12;
+    }
+
+    @Override
+    protected void assertDdl(List<String> schemaChangesDdls) {
+
+        assertThat(schemaChangesDdls.get(schemaChangesDdls.size() - 2)).isEqualTo("DROP TABLE IF EXISTS `blocking_snapshot_test_1`.`b`");
```
mfvitale, line 160: Currently this will fail because for the `DROP` events we don't have information about the table. See https://github.com/debezium/debezium/pull/4824/files#diff-63840528f25c3335e78d8b91c2ea67602750af012c1620d9981f11cb476b480aR642

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
```diff
@@ -635,6 +639,10 @@ protected void createSchemaChangeEventsForTables(ChangeEventSourceContext source
             LOGGER.debug("Processing schema event {}", event);
 
             final TableId tableId = event.getTables().isEmpty() ? null : event.getTables().iterator().next().id();
+            if (snapshottingTask.isBlocking() && !snapshotContext.capturedTables.contains(tableId)) {
```
mfvitale, line 642: @jpechane Here I need to avoid to resend already processed events and process the only ones requested in bloking snapshot signal. Is there a way to check the same for the `DROP` statements for which `tableId` is null? 

I mean something more robust then checking the name of the table in the `ddl` field.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
jpechane: Yes, IIRC it does not depnd on database

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
jpechane: @mfvitale Under which conditions the `DROP` even will not contain `tableId`?

mfvitale: I think in any DROP events 

image

mfvitale: @jpechane Seems the missing of the `tableChanges` depends on this code

image

where the table is added to changes but removed from `tablesByTableId` and then whit will cause 

image

`tables().forTable(tableId) to return `null`.

Should `tablesByTableId` be aligned with `changes`?

mfvitale: @jpechane I think I have found an acceptable solution. Please check commit f478c7e1af0597f0ed7bf6902aabf791e7cdd9e5

jpechane: @mfvitale It is good workaround. But I'd argu that the missing table change is bug in this case. Would you be able to dig into it a bit deeper? If not we can merge this PR and create another Jira just for that problem.

mfvitale: I have the same feeling but not 100%. I think in this case is better to create a separate Jira issue to investigate it.

jpechane:
@mfvitale Applied, thanks! Please create the follow-up Jira for `DROP` events

mfvitale:
> @mfvitale Applied, thanks! Please create the follow-up Jira for `DROP` events

https://issues.redhat.com/browse/DBZ-6945

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
```diff
@@ -319,8 +321,13 @@ else if (!connectorConfig.getSnapshotMode().shouldStream()) {
...
     private void addSchemaEvent(RelationalSnapshotContext<MySqlPartition, MySqlOffsetContext> snapshotContext,
                                 String database, String ddl) {
-        schemaEvents.addAll(databaseSchema.parseSnapshotDdl(snapshotContext.partition, ddl, database,
-                snapshotContext.offset, clock.currentTimeAsInstant()));
+
+        List<SchemaChangeEvent> schemaChangeEvents = databaseSchema.parseSnapshotDdl(snapshotContext.partition, ddl, database,
+                snapshotContext.offset, clock.currentTimeAsInstant());
+        List<SchemaChangeEvent> missingSchemaChangeEvents = schemaChangeEvents.stream()
+                .filter(not(schemaEvents::contains))
+                .collect(Collectors.toList());
```
minchowang, line 329: @mfvitale Hi here have a performance bottleneck.  time complexity is `O(databaseSchemaNumber²)`, We have 140+ databases on single instance, here  process need 2 hour+.

mfvitale: Hi @minchowang can you please open a separate issue? Thanks.

drew-vz: https://issues.redhat.com/browse/DBZ-7608

PR #4825, opened 2023-09-01
DBZ-3642 CloudEventsConverter can retrieve metadata from headers

rkudryashov:
This makes Outbox SMT (that removes source info) + CloudEventsConverter working (by getting source info from headers), that is to publish events from `outbox` table converted to CloudEvents format.

 This works when the connector is configured in appropriate way, particularly:
* filter unwanted messages, such as, heartbeats and tombstones (this is only needed for `HeaderFrom` transform)
* add `source`, `op`, `transaction` headers
* enable schemas for header converter

The following is an example of a configured connector:

```
{
  "name": "book-db-source-connector",
  "config": {
    "plugin.name": "pgoutput",
    "database.hostname": "book-db",
    "database.port": "5432",
    "database.user": "postgres",
    "database.password": "password",
    "database.dbname": "book",
    "database.server.name": "book",
    "topic.prefix": "book",
    "topic.creation.default.replication.factor": 1,
    "topic.creation.default.partitions": 6,
    "table.include.list": "public.outbox",
    "heartbeat.interval.ms": "5000",
    "slot.name": "book_debezium",
    "publication.name": "book_publication",
    "tombstones.on.delete": false,
    "transforms": "addSourceHeaders,outbox",
    "predicates": "isHeartbeat",
    "predicates.isHeartbeat.type": "org.apache.kafka.connect.transforms.predicates.TopicNameMatches",
    "predicates.isHeartbeat.pattern": "__debezium-heartbeat.*",
    "transforms.addSourceHeaders.type": "org.apache.kafka.connect.transforms.HeaderFrom$Value",
    "transforms.addSourceHeaders.fields": "source,op,transaction",
    "transforms.addSourceHeaders.headers": "source,op,transaction",
    "transforms.addSourceHeaders.operation": "copy",
    "transforms.addSourceHeaders.predicate": "isHeartbeat",
    "transforms.addSourceHeaders.negate": true,
    "transforms.outbox.type": "io.debezium.transforms.outbox.EventRouter",
    "transforms.outbox.table.field.event.key": "event_key",
    "transforms.outbox.table.expand.json.payload": true,
    "transforms.outbox.route.by.field": "aggregate_type",
    "transforms.outbox.route.topic.replacement": "outbox.event.library",
    "value.converter": "io.debezium.converters.CloudEventsConverter",
    "value.converter.source.location": "header",
    "header.converter": "org.apache.kafka.connect.json.JsonConverter",
    "header.converter.schemas.enable": true,
    "key.converter": "org.apache.kafka.connect.json.JsonConverter",
    "key.converter.schemas.enable": false
  }
}
```

An example of a message:

```
{
	"id": "name:book;lsn:27733536;txId:751;sequence:[\"27733312\",\"27733536\"]",
	"source": "/debezium/postgresql/book",
	"specversion": "1.0",
	"type": "io.debezium.postgresql.datachangeevent",
	"time": "2023-09-03T17:15:09.228Z",
	"datacontenttype": "application/json",
	"iodebeziumop": "c",
	"iodebeziumversion": "2.4.0-SNAPSHOT",
	"iodebeziumconnector": "postgresql",
	"iodebeziumname": "book",
	"iodebeziumtsms": "1693761309228",
	"iodebeziumsnapshot": "false",
	"iodebeziumdb": "book",
	"iodebeziumsequence": "[\"27733312\",\"27733536\"]",
	"iodebeziumschema": "public",
	"iodebeziumtable": "outbox",
	"iodebeziumtxId": "751",
	"iodebeziumlsn": "27733536",
	"iodebeziumxmin": null,
	"iodebeziumtxid": null,
	"iodebeziumtxtotalorder": null,
	"iodebeziumtxdatacollectionorder": null,
	"data": {
		"schema": {
			"type": "struct",
			"fields": [
				{
					"type": "int32",
					"optional": true,
					"field": "id"
				},
				{
					"type": "string",
					"optional": true,
					"field": "name"
				},
				{
					"type": "int32",
					"optional": true,
					"field": "publicationYear"
				}
			],
			"optional": true,
			"name": "payload"
		},
		"payload": {
			"id": 11,
			"name": "a book",
			"publicationYear": 1824
		}
	}
}
```

Side effect: `"header.converter.schemas.enable": true,` causes all other headers to be added with their schemas.

UPD: added transaction header parsing

### DIFF
--- a/debezium-core/src/main/java/io/debezium/converters/CloudEventsConverter.java
+++ b/debezium-core/src/main/java/io/debezium/converters/CloudEventsConverter.java
@@ -21,6 +21,8 @@ import java.util.ServiceLoader;
 import java.util.function.Function;
 
 import org.apache.kafka.common.errors.SerializationException;
+import org.apache.kafka.common.header.Header;
+import org.apache.kafka.common.header.Headers;
 import org.apache.kafka.connect.data.Field;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.Schema.Type;
@@ -46,6 +48,10 @@ import io.debezium.annotation.VisibleForTesting;
 import io.debezium.config.Configuration;
 import io.debezium.config.Instantiator;
 import io.debezium.connector.AbstractSourceInfo;
+import io.debezium.converters.CloudEventsConverterConfig.MetadataLocation;
+import io.debezium.converters.recordandmetadata.RecordAndMetadata;
+import io.debezium.converters.recordandmetadata.RecordAndMetadataBaseImpl;
+import io.debezium.converters.recordandmetadata.RecordAndMetadataHeaderImpl;
 import io.debezium.converters.spi.CloudEventsMaker;
 import io.debezium.converters.spi.CloudEventsProvider;
 import io.debezium.converters.spi.RecordParser;
@@ -135,6 +141,8 @@ public class CloudEventsConverter implements Converter {
     private final JsonConverter jsonCloudEventsConverter = new JsonConverter();
     private JsonConverterConfig jsonCloudEventsConverterConfig = null;
 
+    private JsonConverter jsonHeaderConverter = new JsonConverter();
+
     private final JsonConverter jsonDataConverter = new JsonConverter();
 
     private boolean enableJsonSchemas;
@@ -143,6 +151,8 @@ public class CloudEventsConverter implements Converter {
     private Converter avroConverter;
     private List<String> schemaRegistryUrls;
     private SchemaNameAdjuster schemaNameAdjuster;
+
+    private MetadataLocation metadataLocation;
 
     public CloudEventsConverter() {
         this(null);
@@ -163,6 +173,12 @@ public class CloudEventsConverter implements Converter {
         ceSerializerType = ceConfig.cloudeventsSerializerType();
         dataSerializerType = ceConfig.cloudeventsDataSerializerTypeConfig();
         schemaNameAdjuster = ceConfig.schemaNameAdjustmentMode().createAdjuster();
+        metadataLocation = ceConfig.metadataLocation();
+
+        Map<String, Object> jsonHeaderConverterConfig = new HashMap<>();
+        jsonHeaderConverterConfig.put(JsonConverterConfig.SCHEMAS_ENABLE_CONFIG, true);
+        jsonHeaderConverterConfig.put(JsonConverterConfig.TYPE_CONFIG, "header");
+        jsonHeaderConverter.configure(jsonHeaderConverterConfig);
 
         boolean usingAvro = false;
 
@@ -220,21 +236,45 @@ public class CloudEventsConverter implements Converter {
 
     @Override
     public byte[] fromConnectData(String topic, Schema schema, Object value) {
+        return this.fromConnectData(topic, null, schema, value);
+    }
+
+    @Override
+    public byte[] fromConnectData(String topic, Headers headers, Schema schema, Object value) {
         if (schema == null || value == null) {
             return null;
         }
-        if (!Envelope.isEnvelopeSchema(schema)) {
-            // TODO Handling of non-data messages like schema change or transaction metadata
-            return null;
+
+        if (this.metadataLocation == MetadataLocation.VALUE) {
+            if (!Envelope.isEnvelopeSchema(schema)) {
+                // TODO Handling of non-data messages like schema change or transaction metadata
+                return null;
+            }
+        }
+        else {
+            if (headers.lastHeader(Envelope.FieldName.SOURCE) == null || headers.lastHeader(Envelope.FieldName.OPERATION) == null) {
+                return null;
+            }
         }
         if (schema.type() != STRUCT) {
             throw new DataException("Mismatching schema");
         }
 
         Struct record = requireStruct(value, "CloudEvents converter");
-        CloudEventsProvider provider = lookupCloudEventsProvider(record);
+        Struct source = getSource(record, headers);
 
-        RecordParser parser = provider.createParser(schema, record);
+        CloudEventsProvider provider = lookupCloudEventsProvider(source);
+
+        RecordAndMetadata recordAndMetadata;
+        if (this.metadataLocation == MetadataLocation.VALUE) {
+            recordAndMetadata = new RecordAndMetadataBaseImpl(record, schema);
+        }
+        else {
+            recordAndMetadata = new RecordAndMetadataHeaderImpl(record, schema, headers, jsonHeaderConverter);
+        }
+
+        RecordParser parser = provider.createParser(recordAndMetadata);
+
         CloudEventsMaker maker = provider.createMaker(parser, dataSerializerType,
                 (schemaRegistryUrls == null) ? null : String.join(",", schemaRegistryUrls));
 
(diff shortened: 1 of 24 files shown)


debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -377,6 +377,52 @@ public static FieldNameAdjustmentMode parse(String value) {
         }
     }
 
+    /**
+     * The set of predefined SourceLocation options
+     */
+    public enum SourceLocation implements EnumeratedValue {
```
Naros, line 383: Please move this to the `CloudEventsConverterConfig` class.  The `CommonConnectorConfig` should only contain things that are specific to actual connectors while this is specific to CloudEvents only.

debezium-core/src/main/java/io/debezium/converters/spi/CloudEventsProvider.java
```diff
@@ -31,6 +32,10 @@ public interface CloudEventsProvider {
      */
     RecordParser createParser(Schema schema, Struct record);
 
+    default RecordParser createParser(Schema schema, Struct record, Struct source, SchemaAndValue operation, Struct transaction) {
+        return createParser(schema, record);
+    }
```
Naros, line 37: Can you explain why we explicitly only override this for PostgreSQL but don't do this for other connectors?

Naros (commented):
A few comments inline @rkudryashov.

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
rkudryashov: done

debezium-core/src/main/java/io/debezium/converters/spi/CloudEventsProvider.java
rkudryashov: This is because I could only test Debezium with Postgres in my project. But apparently a new test should be enough for other databases, so I'm going to add it

debezium-core/src/main/java/io/debezium/converters/CloudEventsConverter.java
```diff
@@ -220,21 +233,44 @@ protected Map<String, String> configureConverterType(boolean isKey, Map<String,
...
+
+        if (this.sourceLocation == SourceLocation.VALUE) {
+            if (!Envelope.isEnvelopeSchema(schema)) {
+                // TODO Handling of non-data messages like schema change or transaction metadata
+                return null;
+            }
+        }
+        else {
+            if (headers.lastHeader(Envelope.FieldName.SOURCE) == null || headers.lastHeader(Envelope.FieldName.OPERATION) == null
+                    || headers.lastHeader(Envelope.FieldName.TRANSACTION) == null) {
```
jpechane, line 253: `transaction` metadata are optional and require transactio metadata to be enabled. Is it intentional that it is mandatory here?

debezium-core/src/main/java/io/debezium/converters/CloudEventsConverterConfig.java
```diff
@@ -32,6 +33,10 @@ public class CloudEventsConverterConfig extends ConverterConfig {
             + "'avro' replaces the characters that cannot be used in the Avro type name with underscore (default)"
             + "'none' does not apply any adjustment";
 
+    public static final String CLOUDEVENTS_SOURCE_LOCATION_CONFIG = "source.location";
```
jpechane, line 36: IMHO this is misleading property name. You are setting location not only for `source` info block but for other metadata as well, so maybe it should be something like `metadata.location`?

debezium-core/src/main/java/io/debezium/converters/spi/RecordParser.java
```diff
@@ -51,8 +52,21 @@ protected RecordParser(Schema schema, Struct record, String... dataFields) {
         this.dataSchema = getDataSchema(schema, connectorType, dataFields);
     }
 
+    protected RecordParser(Schema schema, Struct record, Struct source, SchemaAndValue operation, Struct transaction) {
```
jpechane, line 55: Instead of adding a new constructor I'd propose to change the existing one to with one parameter of type interface like
```
interface RecordAndMetadata {
  Struct getRecord();
  Struct source();
  String operation();
  Struct transaction();
}
```
there will be two implementations for this interface, one operating over `source` info block and the other over headers.
WDYT?

debezium-core/src/main/java/io/debezium/converters/CloudEventsConverterConfig.java
rkudryashov: done

debezium-core/src/main/java/io/debezium/converters/CloudEventsConverter.java
rkudryashov: removed that check

debezium-core/src/main/java/io/debezium/converters/spi/RecordParser.java
rkudryashov: done

debezium-core/src/main/java/io/debezium/converters/spi/RecordParser.java
```diff
@@ -39,20 +41,31 @@ public abstract class RecordParser {
...
-        this.connectorType = source.getString(AbstractSourceInfo.DEBEZIUM_CONNECTOR_KEY);
-        this.dataSchema = getDataSchema(schema, connectorType, dataFields);
+    protected RecordParser(RecordAndMetadata recordAndMetadata, String... dataFields) {
+        this.record = recordAndMetadata.record();
+        this.source = recordAndMetadata.source();
+        this.transaction = recordAndMetadata.transaction();
+        this.op = recordAndMetadata.operation();
+        this.opSchema = Schema.STRING_SCHEMA;
+
+        if (recordAndMetadata instanceof RecordAndMetadataBaseImpl) {
```
jpechane, line 51: Do we need `instanceof` here? Is not something that can be delegated to `RecordAndMetadata` implementation?

debezium-core/src/test/java/io/debezium/converters/CloudEventsConverterTest.java
```diff
@@ -317,6 +324,94 @@ public static void shouldConvertToCloudEventsInAvro(SourceRecord record, String
...
+            Testing.print("  value: " + SchemaUtil.asString(record.value()));
+            Testing.print("  value deserialized from CloudEvents in JSON: " + prettyJson(valueJson));
+            if (t instanceof AssertionError) {
+                throw t;
+            }
+            fail("error " + msg + ": " + t.getMessage());
+        }
+    }
+
+    // TODO: do I need this to convert `connect` package -> `common` package headers?
```
jpechane, line 390: Why TODO?

debezium-core/src/test/java/io/debezium/converters/CloudEventsConverterTest.java
```diff
@@ -317,6 +324,94 @@ public static void shouldConvertToCloudEventsInAvro(SourceRecord record, String
...
+                String topic = record.topic();
+                for (Header header : headers) {
+                    String key = header.key();
+                    byte[] rawHeader = jsonHeaderConverter.fromConnectHeader(topic, key, header.schema(), header.value());
+                    result.add(key, rawHeader);
+                }
+            }
+            return result;
+        }
+        catch (IOException e) {
```
jpechane, line 410: There is no need for `catch`, just propagate the exception.

jpechane (commented):
@rkudryashov Thanks, I left 3 minor comments, other than that I blieve we are good to go.

jpechane:
@rkudryashov Applied, thanks a lot!

PR #4827, opened 2023-09-03
DBZ-6865 Avoid getting NPE when executing the arrived method in ExecuteSnapshot

harveyyue:
https://issues.redhat.com/browse/DBZ-6865


jpechane:
@harveyyue Applied, thanks

PR #4829, opened 2023-09-04
DBZ-6867 streaming pipeline should match events from collection include list as well as signal collection

jcechace:
https://issues.redhat.com/browse/DBZ-6867


jpechane:
@jcechace Applied, thanks

PR #4833, opened 2023-09-06
DBZ-6773 Adding support for new field/header mappings to ExtractNewDocumentState

jcechace:
https://issues.redhat.com/browse/DBZ-6773


jpechane:
@jcechace Applied, thanks

PR #4834, opened 2023-09-06
DBZ-6872 Ability to switch the order of user defined and internal aggregation pipeline

jcechace:
https://issues.redhat.com/browse/DBZ-6872

prepares the ground for https://issues.redhat.com/browse/DBZ-6871

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ChangeStreamPipelineFactory.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ChangeStreamPipelineFactory.java
@@ -22,6 +22,7 @@ import com.mongodb.client.model.Aggregates;
 import com.mongodb.client.model.Filters;
 import com.mongodb.client.model.changestream.OperationType;
 
+import io.debezium.DebeziumException;
 import io.debezium.connector.mongodb.Filters.FilterConfig;
 import io.debezium.data.Envelope;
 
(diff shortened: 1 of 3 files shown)


btiernay:
I don't think this should break anything but I could see it causing some performance regressions for expensive user filters that would otherwise be avoided with the preceding op filter that now moves to the end.

jpechane:
@jcechace This is valid concern. Do you think we can introduce an internal property that could affect the order of evaluation with default kept as is in this PR?

jcechace:
That's not a problem, however I see this as yet another complexity. If performance is a problem, then the user can always do some prefiltering as part of their pipeline no?

jpechane:
IIUC it would be to effectivelly repeat the pipeline as defined in Debezium. This can be quite error prone. We don't know if this is a real concern but the internal option will give us a wy how to quickly workaround it if necessary.

jcechace:
In that case I have to ask what is meant by internal? 

As for temporary -- I didn't intend to reverse this for 2.4.

Naros:
> In that case I have to ask what is meant by internal?

This generally means `Field.create` vs `Field.createInternal` properties.

jpechane:
It does not mean it is going to be reverted, just that the option can be safely removed in future.

jcechace:
@jpechane I kinda don't like this idea of having a switch for the order. At the same time I unerstand that the fact that the input data set for user pipeline changes based on the order is  problematic.

What about we keep the default as is (so internal - user) and the switch is used just as sort of an official workaround (where we state that there are valid usecaases but there migh be a performance impact).

jpechane:
@jcechace :+1:

jcechace:
@jpechane this shoudl be ready for merge. The FAQ PR was also updated

jpechane:
@jcechace Applied, thanks!

PR #4838, opened 2023-09-07
DBZ-6871 MongoDB connector is now able to skip events for oversized documents

jcechace:
https://issues.redhat.com/browse/DBZ-6871

Should be merged only after https://github.com/debezium/debezium/pull/4834

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
@@ -425,6 +425,70 @@ public class MongoDbConnectorConfig extends CommonConnectorConfig {
         }
     }
 
+    /**
+     * The set of predefined OversizeHandlingMode options or aliases.
+     */
+    public enum OversizeHandlingMode implements EnumeratedValue {
+        /**
+         * Connect individually to each replica set
+         */
+        FAIL("fail"),
+
+        /**
+         * Connect to sharded cluster with single connection via mongos
+         */
+        SKIP("skip");
+
+        private String value;
+
+        OversizeHandlingMode(String value) {
+            this.value = value;
+        }
+
+        @Override
+        public String getValue() {
+            return value;
+        }
+
+        /**
+         * Determine if the supplied value is one of the predefined options.
+         *
+         * @param value the configuration property value; may not be null
+         * @return the matching option, or null if no match is found
+         */
+        public static OversizeHandlingMode parse(String value) {
+            if (value == null) {
+                return null;
+            }
+            value = value.trim();
+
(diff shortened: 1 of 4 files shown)


Naros:
@jcechace do we need to exclude the new tests from older versions of MongoDB?

jcechace:
@Naros Looking into it...I guess since this feature has to be explicitly enabled it's OK to condition just the tests -- we will simply document that this shoudl be enabled only for MongoDB > 4.4

jcechace:
@jpechane @Naros do we want to still support Mongo 4.4 with 2.4? It's EOL since February and I would like to kepe the policy of supporting just one EOLed version (5.0 goes EOL in October)

jpechane:
@jcechace Jira issue and kill it please

jpechane:
@jcechace Could you please rebase/updated the PR?

jpechane:
@jcechace Applied, thanks a lot! Please prepare a new docs PR with all the new config options. One commit for 2.3/2.4 changes and one for 2.4 only

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
```diff
@@ -738,6 +802,27 @@ public static CursorPipelineOrder parse(String value, String defaultValue) {
...
+                    + "'fail' (the default) the connector fails if the total size of change event exceed the maximum BSON size"
+                    + "'skip' any change events for documents exceeding the maximum size will be ignored ");
+
+    public static final Field CURSOR_OVERSIZE_SKIP_THRESHOLD = Field.create("cursor.oversize.skip.threshold")
+            .withDisplayName("Oversize document skip threshold")
+            .withType(Type.INT)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withDefault(0)
+            .withValidation(MongoDbConnectorConfig::validateOversizeSkipThreshold)
```
sharanrudresh, line 821: @jcechace @jpechane These two configs were not added to the list in `CONFIG_DEFINITION` (line 867 below), and thus this validation does not run.
Curious - was this a miss, or was it intentional. I see there's more such configs in the latest `main` branch as well - https://github.com/debezium/debezium/blob/aef8b3eac54bcbe8c07587270e1ec55200eac7b5/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java#L1019

jpechane: @sharanrudresh This is most probably oversight, could you pease add the missing options?

sharanrudresh: Sure, thankyou!

PR #4839, opened 2023-09-07
DBZ-6876 Add key/drop index clause to all drop constraints

Naros:
https://issues.redhat.com/browse/DBZ-6876


jpechane:
@Naros Applied, thanks

PR #4840, opened 2023-09-07
[docs] Add primary key updates Kafka headers section to Vitess connector

twthorn:
The same behavior & headers as other debezium connectors (e.g., [mysql](https://github.com/debezium/debezium/blob/main/documentation/modules/ROOT/pages/connectors/mysql.adoc?plain=1#L1447-L1455)) is done for primary key updates with the vitess connector. The docs make no mention of this so update them to do so.


jpechane:
@twthorn Applied, thanks

PR #4841, opened 2023-09-08
[chore] Fix grammar issue in README

tisonkun:
TRIVIAL AS IS


tisonkun:
It seems too complex to finish such a trivial change. I'd suggest you advice what exactly I should do or just pick it up and push by yourself.

jpechane:
@tisonkun Applied, thanks

PR #4842, opened 2023-09-08
DBZ-6880 Move some Oracle log messages to DEBUG

Naros:
https://issues.redhat.com/browse/DBZ-6880

@zalmane here's the PR I talked about.  It should be a direct back-port to Debezium 2.3 branch.  Let me know what you think about these log changes,.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
@@ -418,7 +418,7 @@ public abstract class AbstractLogMinerEventProcessor<T extends AbstractTransacti
         final T transaction = getAndRemoveTransactionFromCache(transactionId);
         if (transaction == null) {
             handleCommitNotFoundInBuffer(row);
-            LOGGER.trace("Transaction {} not found, commit skipped.", transactionId);
+            LOGGER.debug("Transaction {} not found, commit skipped.", transactionId);
             return;
         }
 
@@ -626,7 +626,7 @@ public abstract class AbstractLogMinerEventProcessor<T extends AbstractTransacti
                 return false;
             }
             else if (connectorConfig.getLogMiningUsernameExcludes().contains(transaction.getUserName())) {
-                LOGGER.trace("Skipped transaction with excluded username {}", transaction);
+                LOGGER.debug("Skipped transaction with excluded username {}", transaction);
                 return true;
             }
         }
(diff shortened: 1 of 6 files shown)


zalmane:
Do we also want to include details about the "unsupported op" in DEBUG mode? Will help understand why those happen.

Naros:
> Do we also want to include details about the "unsupported op" in DEBUG mode? Will help understand why those happen.

Added.

Naros:
I still need to run the tests locally for LogMiner to make sure that this doesn't impair any of our tests that check log messages, so we'll hold off merging this into `main` until I can confirm test suite passes.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
```diff
@@ -990,7 +990,7 @@ protected void handleDataEvent(LogMinerEventRow row) throws SQLException, Interr
             return;
         }
 
-        LOGGER.trace("DML: {}", row);
+        LOGGER.debug("DML: {}", row);
```
jpechane, line 993: @Naros Is not this exactly the type of messages that makes the privacy concerns?

zalmane:
Correct. We should omit the payload of the actual row but keep the opcode
and Metadata

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
jpechane: Feel free to merge yourself when this and the similar cases are addressed

Naros: No, I modified the `LogMinerEventRow` `toString()` method to explicitly exclude the `redoSql` unless the log level was at least `TRACE`; otherwise, it replaces the SQL with `<omitted>` but keeps the remainder of the output identical so we get the other metadata and attributes about the row which isn't sensitive.

Naros:
> Correct. We should omit the payload of the actual row but keep the opcode and Metadata

If this looks good @zalmane I'll go back and merge & backport.  Just let me know.   I didn't confirm setting DEBUG the DML outputs are as follows:
```
DML: LogMinerEventRow{scn=4602630, tableId='ORCLPDB1.DEBEZIUM.CUSTOMER', tableName='CUSTOMER', tablespaceName='DEBEZIUM', eventType=INSERT, changeTime=2023-09-08T17:43:44Z, transactionId='0a0007000a070000', operation='INSERT', userName='DEBEZIUM', rowId='AAASPFAAMAAAACFAAB', rollbackFlag=false, rsId=0x00001f.0001e868.0010, ssn=1, redoSql='<omitted>'}
```

zalmane:
Looks good!

Naros:

PR #4843, opened 2023-09-08
DBZ-6481 Support Postgres enum as key in incremental snapshot

jpechane:
https://issues.redhat.com/browse/DBZ-6481


Naros (commented):
LGTM @jpechane, can you update the commit message to say "Postgres Enum" to describe the behavior change better?

jpechane:
@Naros Done, thanks for the review

PR #4848, opened 2023-09-11
Update sqlserver.adoc

sami2ahmed:
#10 does not make sense, I think it got miscopied from item #11


Naros:
Hi @sami2ahmed I actually think the description as it is reflects the correct intent.  The list of Kafka brokers where we will read and write DDL schema history to.

sami2ahmed:
Thanks @Naros -- I guess its a bit confusing because this is also the field that the connector uses to simply make contact with kafka at all right? It's not just used for reading/writing DDL schema history

Naros:
> Thanks @Naros -- I guess its a bit confusing because this is also the field that the connector uses to simply make contact with kafka at all right? It's not just used for reading/writing DDL schema history

Anything configured under the `schema.history.internal.*` namespace is specifically for reading and writing the DDL to the schema history topic only.  Either way, can you please adjust the base branch to `main` rather than `2.3.`?

jpechane:
@sami2ahmed Thanks for the PR but I am closing it. The descritpion is correct as this setting is used really for internal schema history store only. The general Kafka Connect <-> Kafka Borker connection config is handled separately at Kafka Connect level and this config option is unrelated.

PR #4852, opened 2023-09-12
DBZ-6892 Removing all code related to previously deprecated MongoDB conenctor configuration

jcechace:
https://issues.redhat.com/browse/DBZ-6892


jpechane:
@jcechace Applied, thanks!

PR #4853, opened 2023-09-13
DBZ-6901 ExtractNewRecordState's schema cache is not updated with arrival of the ddl change event

harveyyue:
https://issues.redhat.com/browse/DBZ-6901

### DIFF
--- a/debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordState.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordState.java
@@ -21,6 +21,7 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Optional;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordState.java
```diff
@@ -318,7 +318,7 @@ private static String buildCacheKey(Struct value, Struct originalRecordValue) {
         // But since ExtractChangedRecordState generates changed field with `update` or `delete` operation and then eventually copied to the payload with HeaderToValue SMT,
         // the schema in that case will never be updated since cached on the first `read` operation.
         // Using also the operation in the cache key will solve the problem.
-        return value.schema() + originalRecordValue.getString(OPERATION);
+        return value.schema().hashCode() + originalRecordValue.getString(OPERATION);
```
jpechane, line 321: Don't you think it would bet better to use value.schema() + value.schema().hashCode() + originalRecordValue.getString(OPERATION) to avoid potential hash collisions with unrelated schemas?

jpechane (commented):
@harveyyue Thanks for the PR, I think it is good solution. I am just a bit uneasy about potential hash collisons, please see the other comment.

debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordState.java
harveyyue: I prefer to use topic name instead of same value of `value.schema()`

jpechane: @harveyyue There is still an issue of having routing SMT applied on the source side and sending multiple tables to a single topic. In that case you'll end up with the collisions again. WDYT?

harveyyue: @jpechane Yes, you are right, I just want to reduce the probability of hash collisions. However, we should add a static class like `NewRecordValueMetadata` to fix it.

jpechane:
@harveyyue Applied, thanks

PR #4856, opened 2023-09-13
DBZ-6903 Use distinct dir for protobuf generated sources

jpechane:
https://issues.redhat.com/browse/DBZ-6903


Naros (approved):
Where have I seen this change before, it's like deja vu :grin: 

LGTM.

PR #4859, opened 2023-09-14
DBZ-6895 Fix OpenLogReplicator confirmation

Naros:
https://issues.redhat.com/browse/DBZ-6895


Naros:
Hi @jpechane I've run the tests for OLR locally, I've kicked off the XStream and LogMiner ones via the testing farm.  I believe this is ready barring your review as long as XStream and LogMiner pass CI.

Naros:
Looks like a testing farm problem, I'll kick the tests off locally and let them run over-night instead.

jpechane:
@Naros LGTM, but did not we say that we will emit forcefully heartbeat message when a checkpoint event is received?

Naros:
> @Naros LGTM, but did not we say that we will emit forcefully heartbeat message when a checkpoint event is received?

We did and I completely forgot.  I'll add that and double-check that it doesn't create any problems with the tests.

PR #4860, opened 2023-09-14
DBZ-7194 Enable running Oracle 23 Free Edition

baabgai:
match only to the first occurrence of `Release ` substring in Oracle Banner. For certain versions of oracle databases the current regular expression fails to parse the version.

Old Regexp:
Oracle Database 12c Enterprise Edition Release 12.2.0.1.0 - 64bit Production => ✅ => 12.2.0.1.0 is correctly parsed
Oracle Database 23c Free, Release 23.0.0.0.0 - Developer-Release Version 23.2.0.0.0 => ❌ => failed to parse version

new Regexp only matches until the first occurrence of `Release `:
Oracle Database 12c Enterprise Edition Release 12.2.0.1.0 - 64bit Production => ✅ => 12.2.0.1.0 is correctly parsed
Oracle Database 23c Free, Release 23.0.0.0.0 - Developer-Release Version 23.2.0.0.0 => ✅ => 23.2.0.0.0 is correctly parsed

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDatabaseVersion.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDatabaseVersion.java
@@ -15,7 +15,7 @@ import java.util.regex.Pattern;
  */
 public class OracleDatabaseVersion {
     private final static Pattern VERSION_PATTERN = Pattern
-            .compile("(?:.*)(?:Release )([0-9]+)\\.([0-9]+)\\.([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:.*)");
+            .compile("(?:.*)(?:\bRelease \b)([0-9]+)\\.([0-9]+)\\.([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:.*)");
     private final static Pattern VERSION_18_1_PATTERN = Pattern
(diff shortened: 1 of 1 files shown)


jpechane:
@baabgai Thank you very much for the PR. Could you please create a Jira for the issue and change commit message to
`DBZ-xxx Refine regexp for parsing oracle versions` where xxx is Jira number? It is always good to have it handy to track history.

Also I see you have multiple vrsion Banners available. Would it be possible to create a unit test that would test different version banners so we'd be safe in case of future changes?

Naros (commented):
I don' think we should add `Developer-Release` to the banner.

Oracle has decided to rebrand XE as "Free" (aka Developer-Release) and as we've seen in the past, these versions often do not ship with all the Enterprise edition features required such as supplemental logging (which Oracle 18 XE explicitly disabled).

Additionally, the developer release itself is constrained to the number of cpus/cores, database size, and so on; some of which will impact the performance of the database.  We discussed this internally and we opted not to try and build any type of image on Oracle 23 DR because of these reasons, hence why the banner does not honor this release type.

Personally, I like the idea that the banner forces the connector to stop.  We also haven't updated our test matrices to include Oracle 23 either due to no EE build available, so I'm inclined to think we should reject this change.  @jpechane  wdyt?

jpechane:
@Naros But is it technically possible to use that version with some effort? If yes then I wonder if we could allow the user to override the version check if necessary?

Naros:
Hi @baabgai, we'll look at adding this in Debezium 2.5.0.Alpha1 in a couple of weeks.  That will allow me to review all the Oracle 23 compatibility requirements, despite the performance concerns, before we open this up.

Naros:
Closed  in favor of PR #5050

PR #4863, opened 2023-09-16
DBZ-6907 Introduce a new handling deleted record option "delete.tombstone.handling.mode"

harveyyue:
https://issues.redhat.com/browse/DBZ-6907

### DIFF
--- a/debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordState.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordState.java
@@ -102,33 +102,34 @@ public class ExtractNewRecordState<R extends ConnectRecord<R>> extends AbstractE
 
     @Override
     public R doApply(final R record) {
+        // Add headers if needed
+        if (!additionalHeaders.isEmpty()) {
+            Headers headersToAdd = makeHeaders(additionalHeaders, (Struct) record.value());
+            headersToAdd.forEach(h -> record.headers().add(h));
+        }
+
+        // Handling tombstone record
         if (record.value() == null) {
-            if (dropTombstones) {
-                LOGGER.trace("Tombstone {} arrived and requested to be dropped", record.key());
-                return null;
-            }
-            if (!additionalHeaders.isEmpty()) {
-                Headers headersToAdd = makeHeaders(additionalHeaders, (Struct) record.value());
-                headersToAdd.forEach(h -> record.headers().add(h));
-            }
-            return record;
+            return extractRecordStrategy.handleTruncateRecord(record);
         }
 
         if (!smtManager.isValidEnvelope(record)) {
             return record;
         }
 
-        if (!additionalHeaders.isEmpty()) {
-            Headers headersToAdd = makeHeaders(additionalHeaders, (Struct) record.value());
-            headersToAdd.forEach(h -> record.headers().add(h));
-        }
-
-        R newRecord = afterDelegate.apply(record);
-        if (newRecord.value() == null && beforeDelegate.apply(record).value() == null) {
+        R newRecord = extractRecordStrategy.afterDelegate().apply(record);
+        if (newRecord.value() == null && extractRecordStrategy.beforeDelegate().apply(record).value() == null) {
             LOGGER.trace("Truncate event arrived and requested to be dropped");
             return null;
         }
+
         if (newRecord.value() == null) {
+            // Handling delete records
+            newRecord = extractRecordStrategy.handleDeleteRecord(record);
+            if (newRecord == null) {
+                return null;
+            }
+
             if (routeByField != null) {
                 Struct recordValue = requireStruct(record.value(), "Read record to set topic routing for DELETE");
                 String newTopicName = recordValue.getStruct("before").getString(routeByField);
@@ -139,22 +140,13 @@ public class ExtractNewRecordState<R extends ConnectRecord<R>> extends AbstractE
                 newRecord = dropFields(newRecord);
             }
 
-            // Handling delete records
-            switch (handleDeletes) {
-                case DROP:
-                    LOGGER.trace("Delete message {} requested to be dropped", record.key());
-                    return null;
-                case REWRITE:
-                    LOGGER.trace("Delete message {} requested to be rewritten", record.key());
-                    R oldRecord = beforeDelegate.apply(record);
-                    oldRecord = addFields(additionalFields, record, oldRecord);
-
-                    return removedDelegate.apply(oldRecord);
-                default:
-                    return newRecord;
+            if (newRecord.value() != null) {
+                newRecord = addFields(additionalFields, record, newRecord);
             }
         }
         else {
+            // Handling insert and update records
+            newRecord = extractRecordStrategy.handleRecord(record);
             // Add on any requested source fields from the original record to the new unwrapped record
             if (routeByField != null) {
                 Struct recordValue = requireStruct(newRecord.value(), "Read record to set topic routing for CREATE / UPDATE");
@@ -167,16 +159,8 @@ public class ExtractNewRecordState<R extends ConnectRecord<R>> extends AbstractE
             if (!Strings.isNullOrEmpty(dropFieldsHeaderName)) {
                 newRecord = dropFields(newRecord);
             }
-
-            // Handling insert and update records
-            switch (handleDeletes) {
-                case REWRITE:
-                    LOGGER.trace("Insert/update message {} requested to be rewritten", record.key());
-                    return updatedDelegate.apply(newRecord);
-                default:
-                    return newRecord;
-            }
         }
+        return newRecord;
     }
 
     @Override
--- a/debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordState.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordState.java
@@ -6,47 +6,27 @@
 package io.debezium.transforms;
 
 import static io.debezium.data.Envelope.FieldName.OPERATION;
-import static io.debezium.data.Envelope.FieldName.SOURCE;
-import static io.debezium.data.Envelope.FieldName.TIMESTAMP;
-import static io.debezium.data.Envelope.FieldName.TRANSACTION;
-import static io.debezium.pipeline.txmetadata.TransactionMonitor.DEBEZIUM_TRANSACTION_DATA_COLLECTION_ORDER_KEY;
-import static io.debezium.pipeline.txmetadata.TransactionMonitor.DEBEZIUM_TRANSACTION_ID_KEY;
-import static io.debezium.pipeline.txmetadata.TransactionMonitor.DEBEZIUM_TRANSACTION_TOTAL_ORDER_KEY;
+import static io.debezium.transforms.ExtractNewRecordStateConfigDefinition.CONFIG_FIELDS;
 import static org.apache.kafka.connect.transforms.util.Requirements.requireStruct;
 
 import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
 import java.util.Optional;
-import java.util.regex.Pattern;
-import java.util.stream.Collectors;
 
 import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.connect.connector.ConnectRecord;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.SchemaBuilder;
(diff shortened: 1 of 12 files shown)


Naros:
Looks like some test failures related to the change @harveyyue, would you mind taking a look?

harveyyue:
> Looks like some test failures related to the change @harveyyue, would you mind taking a look?

OK

jpechane:
We are now on 2.5 so we can merge this when ready.

rk3rn3r:
@jpechane Didn't we once talk about unifying `drop.tombstones` and `delete.handling.mode` because having both configs is quite confusing?
I remember always hitting this topic when I setup a replication using Debezium with Kafka and with a JDBC sink. wdyt?

jpechane:
@rk3rn3r That's good reminder. We just need to keep in mind that the tombostone might still be required in the stream. So there would be 4 states instaed of 3.
`drop` - will remove both delete and tombstone
`none` - will convert delete to tombstone and remove tombstone
`rewrite` - will convert delete and remove tombstone
`rewrite-with-tombstone` - will convert delete and keep tombstone

WDYT?

Naros:
Hi @rk3rn3r and @jpechane, I think there is one state we need that needs to be mentioned.  For example, you may need both tombstone and delete, where the former is primarily for topic compaction, and the latter is what you consume on the consumer side of the pipeline.  

I believe the options should likely be:

* `drop` - removes both tombstones and deletes events from the stream.
* `drop-keep-tombstone` - removes deletes but retains tombstones for topic compaction.
* `none` - removes nothing from the stream, passing tombstones and deletes.
* `rewrite` - rewrite deletes and removes tombstones.
* `rewrite-keep-tombstone` - rewrite deletes and retains tombstones.

The main difference here is that `none` doesn't act on either delete or tombstones, leaving them untouched in the stream, while there is a variant of `drop-keep-tombstone` or `drop-with-tombstone` that drops the delete event but retains the tombstone for compaction purposes.

That should retain all currently possible configuration possibilities.

jpechane:
@Naros `none` converts delete into tombstone, see the docs hence I don't think it is not necessary to keep original tombstones. Please see the docs. In context of `ExtractNewDocumentState` there is no way how you could keep delete.

rk3rn3r:
Should we maybe call the one that does not keep the delete event but the tombstone call `tombstone` ? And honestly, from that list I don't know which one it is... the `none`? I feel the brief descriptions could be worded better. :man_shrugging:

jpechane:
Sounds good. The convert delete means that it takes the `d` mesage and makes tombostone (by setting the value to `null`) out of it.

harveyyue:
Guys, I will implement these change to ExtractNewRecordState and ExtractNewDocumentState transformations

jpechane:
@harveyyue Thanks a lot! Could you please take a look at the changes earlier if possible? As it is breaking change it would be nice to have it done at the start of release cycle. Also we will need to intorduce a deprecation path. So either a new config option replacing the older two will be introduced or `delete.handling.mode` would need to support both old and new option names.
When old config option/value is used a warning should be logged. Also if possible it would be nice to structure the code in such a way that the old branch could be deleted easily after deprecation pahse.

harveyyue:
> @harveyyue Thanks a lot! Could you please take a look at the changes earlier if possible? As it is breaking change it would be nice to have it done at the start of release cycle. Also we will need to intorduce a deprecation path. So either a new config option replacing the older two will be introduced or `delete.handling.mode` would need to support both old and new option names. When old config option/value is used a warning should be logged. Also if possible it would be nice to structure the code in such a way that the old branch could be deleted easily after deprecation pahse.

@jpechane I prefer to introduce a new config option and easy to be compatible with older config `delete.handling.mode`.

debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordState.java
```diff
@@ -105,80 +83,56 @@ public class ExtractNewRecordState<R extends ConnectRecord<R>> implements Transf
...
 
     @Override
-    public R apply(final R record) {
+    public R doApply(final R record) {
+        // Handling tombstone record
         if (record.value() == null) {
-            if (dropTombstones) {
-                LOGGER.trace("Tombstone {} arrived and requested to be dropped", record.key());
-                return null;
+            if (enableDeleteHandlingV2) {
```
jpechane, line 108: Can those `if (enableDeleteHandlingV2)` be converted into methods and placed into strategy interface that would be implemented by configuration enums or similar approach like having a strategy class created based on the config field used? Or maybe even better have the switches removed too if possible?

jpechane (commented):
@harveyyue Thanks I think we are on a good path. I commented on the config option name (and probably variable names too) and I'd recommend additiona code refactoring if possible.

harveyyue:
@jpechane I will add related IT tests for new option `delete.tombstone.handling.mode`

debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordStateConfigDefinition.java
```diff
@@ -67,6 +67,71 @@ public static DeleteHandling parse(String value, String defaultValue) {
...
+
+    public static final Field HANDLE_DELETES_V2 = Field.create("delete.handling.mode.v2")
+            .withDisplayName("Handle delete records")
+            .withEnum(DeleteHandlingV2.class)
+            .withWidth(ConfigDef.Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.MEDIUM)
+            .optional()
+            .withDescription("How to handle delete records. Options are: "
+                    + "drop - will remove both delete and tombstone,"
+                    + "tombstone - will convert delete to tombstone and remove tombstone (should be the default),"
```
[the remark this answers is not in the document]
rk3rn3r: What's the difference between the first tombstone and the second tombstone word here?
I assume they don't mean the same event,right?
I think it is a technical description about what happens under the hood, is it?

I assume that option means that Debezium drops the delete event and only emits a tombstone? Why not explaining it like this, for the user it should not matter if technically a delete event is converted into a tombstone and then gets emitted and the original tombstone is dropped (in order to only emit a single tombstone instead of 2).

debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordStateConfigDefinition.java
```diff
@@ -67,6 +67,71 @@ public static DeleteHandling parse(String value, String defaultValue) {
...
+    }
+
+    public static final Field HANDLE_TOMBSTONE_DELETES = Field.create("delete.tombstone.handling.mode")
+            .withDisplayName("Handle delete records")
+            .withEnum(DeleteTombstoneHandling.class)
+            .withWidth(ConfigDef.Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.MEDIUM)
+            .optional()
+            .withDescription("How to handle delete records. Options are: "
+                    + "drop - will remove both delete and tombstone,"
```
rk3rn3r, line 129:
```suggestion
                    + "drop - will remove both, the delete event and tombstone, from the stream,"
```

debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordStateConfigDefinition.java
```diff
@@ -67,6 +67,71 @@ public static DeleteHandling parse(String value, String defaultValue) {
...
+                    + "tombstone - will convert delete to tombstone and remove tombstone (the default),"
```
rk3rn3r, line 130:
```suggestion
                    + "tombstone - will only leave a tombstone in the stream (the default),"
```
I am not happy with that explanation but from a user's perspective I feel that explains more what the user will get (instead of technically describing that the delete event is converted into a tombstone and the original tombstone is dropped from the stream).
wdyt?

debezium-core/src/main/java/io/debezium/transforms/strategy/DeleteTombstoneExtractRecordStrategy.java
```diff
@@ -0,0 +1,80 @@
...
+
+import io.debezium.DebeziumException;
+import io.debezium.transforms.ExtractNewRecordStateConfigDefinition.DeleteTombstoneHandling;
+
+/**
+ * A default implementation of {@link AbstractExtractRecordStrategy}
+ *
+ * @author Harvey Yue
+ */
+public class DeleteTombstoneExtractRecordStrategy<R extends ConnectRecord<R>> extends AbstractExtractRecordStrategy<R> {
```
rk3rn3r, line 20: Isn't this more a type of `DeleteHandlingStrategy` not a entire `ExtractRecordStrategy`? E.g. `DropTombstoneDeleteHandlingStrategy` here?

debezium-core/src/main/java/io/debezium/transforms/strategy/DeleteExtractRecordStrategy.java
```diff
@@ -0,0 +1,78 @@
...
+import io.debezium.DebeziumException;
+import io.debezium.transforms.ExtractNewRecordStateConfigDefinition.DeleteHandling;
+
+/**
+ * Deprecated, use {@link DeleteTombstoneExtractRecordStrategy} instead
+ *
+ * @author Harvey Yue
+ */
+@Deprecated
+public class DeleteExtractRecordStrategy<R extends ConnectRecord<R>> extends AbstractExtractRecordStrategy<R> {
```
rk3rn3r, line 22: Shouldn't this be a Strategy below the `ExtractNewRecordState` layer and then be named something like `...DeleteHandlingStrategy` instead of `...ExtractRecordStrategy`?
Maybe I am mistaken here though... (see my other comment about the strategy name as well)

debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordStateConfigDefinition.java
```diff
@@ -67,6 +67,71 @@ public static DeleteHandling parse(String value, String defaultValue) {
...
+                    + "rewrite - will convert delete and remove tombstone, __deleted field is added to records,"
```
rk3rn3r, line 131:
```suggestion
                    + "rewrite - add a `__deleted"` field with the value `true` to the record and remove tombstone and delete event from the stream,"
```
`ExtractNewRecordState` also makes sense to use on the sink side as well, right? Then this option should be marked as only use-able on the source side, no? Or can it be used on the sink side as well? I don't think it works there in the same way as on the sink side, does it? It can't combine multiple messages into one there, right?


When I was writing this I was also confused by the difference between the physical thing happening - the delete operation of a record in the database and the name of the message that Debezium produces to reflect that delete operation in the stream, which we also call a `DELETE event`. I would call the operation happening in the database the `delete event`. But we use `delete event` here as a name for  the delete message that Debezium emits to reflect that event. That makes it hard for me to understand and precisely identify when we mean what.
@Naros  @jpechane @roldanbob wdyt?
Should we maybe change the wording and call the operation `delete event` and the Debezium delete record as `delete message` or something? Maybe it is just me but I find this easy to confuse. :man_facepalming:

debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordStateConfigDefinition.java
```diff
@@ -67,6 +67,71 @@ public static DeleteHandling parse(String value, String defaultValue) {
...
+                    + "rewrite-with-tombstone - will convert delete and keep tombstone, __deleted field is added to records.");
```
rk3rn3r, line 132: I feel that `convert delete` is not clear enough about what's happening. Similar to my concerns above.

debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordStateConfigDefinition.java
roldanbob: In my way of thinking, we need to use language to precisely differentiate between events and event records. 
That is, the original operation that is applied to a table is an event (`delete` event). 
When Debezium captures such an event from a table, it then emits a corresponding event record (i.e., a `delete event record`).  
I believe this matches up with the terminology in most of the documentation.

jpechane:
@harveyyue Thanks a lot!. I renamed the strategy classes and package. When CI passes the PR will be merged.

Could you please create a separate PR with docs update and check with @rk3rn3r and @roldanbob on best wording/description? The description in the code could be modified as part of the docs PR.

harveyyue:
> @harveyyue Thanks a lot!. I renamed the strategy classes and package. When CI passes the PR will be merged.
> 
> Could you please create a separate PR with docs update and check with @rk3rn3r and @roldanbob on best wording/description? The description in the code could be modified as part of the docs PR.

OK

jpechane:
@harveyyue Applied, thanks a lot! Please provide the docs PR when you have a time.

PR #4868, opened 2023-09-18
DBZ-6935 fix logger named

wuzhenhua01:
fix logger named
image
it should be PostgresOffsetContext


jpechane:
@wuzhenhua01 Applied, thanks

PR #4870, opened 2023-09-18
DBZ-6938 Make retry scn-log-check per redo thread

Naros:
https://issues.redhat.com/browse/DBZ-6938


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
```diff
@@ -150,6 +150,7 @@ public static String allMinableLogsQuery(Scn scn, Duration archiveLogRetention,
             sb.append("LEFT JOIN ").append(ARCHIVED_LOG_VIEW).append(" A ");
             sb.append("ON A.FIRST_CHANGE# = L.FIRST_CHANGE# AND A.NEXT_CHANGE# = L.NEXT_CHANGE# ");
             sb.append("WHERE (A.STATUS <> 'A' OR A.FIRST_CHANGE# IS NULL) ");
+            sb.append("AND L.STATUS != 'UNUSED' ");
```
zalmane, line 153: Is the "UNUSED" status related to the same phenomena?

jpechane (approved):
@Naros LGTM, thanks. Please check @zalmane question and we can merge the PR then.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
Naros: Not directly, which is why I added it as a separate commit.

In effect, this use case can lead to an unexpected log entry in the logs stating that we found a redo log with a range of 0-0 with a sequence of 0 that is being excluded.  We are simply lucky that the logic in `LogMinerHelper` is very explicit about the criteria to add to the online redo logs collection :) .

Naros:
If we're okay with my response, I'll merge this today.  @zalmane ?

zalmane:
Great

Naros:

PR #4874, opened 2023-09-20
DBZ-6910 Values being omitted from list of JSON object

harveyyue:
https://issues.redhat.com/browse/DBZ-6910


jpechane:
@harveyyue Applied, thanks

PR #4876, opened 2023-09-20
DBZ-6945 Remove Table object for DROP schema event

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6945


jpechane:
@mfvitale Applied, thanks!

PR #4881, opened 2023-09-21
DBZ-6787: Dedupe table list while incremental snapshot

akanimesh7:


debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -620,6 +621,9 @@ private List<String> expandDataCollectionIds(List<String> dataCollectionIds) {
                     return ids.isEmpty() ? Stream.of(x) : ids.stream();
                 })
                 .collect(Collectors.toList());
+        HashSet<String> uniqueDataCollections = new HashSet<>();
```
jpechane, line 624: @akanimesh7 Thanks for the PR. Unfortunately I am afraid that the result of this operation will be the loss of order intended to execute the snapshot. The deduplication should keep the order of tables prescribed.

akanimesh7: @jpechane Sorry, i missed that. Changed it to LinkedHashSet to preserve the oder.

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -620,6 +622,8 @@ private List<String> expandDataCollectionIds(List<String> dataCollectionIds) {
                     return ids.isEmpty() ? Stream.of(x) : ids.stream();
                 })
                 .collect(Collectors.toList());
```
mfvitale, line 624: I think you can simplify it using `collect( Collectors.toCollection( LinkedHashSet::new ));`

Can also try to add a test?

akanimesh7: @mfvitale distinct seems to be even better solution, since it preserves the insertion order

akanimesh7:
@jpechane @mfvitale What would be the correct way to test this ? I could only see integration tests for this class and also thre unit tests are there but they are there only for the `buildChunkQuery` part.

akanimesh7:
@mfvitale / @jpechane 
It is right now not feasible to add test for this case. The reason being if we add a test like this -- 
```
@Test
    public void snapshotWithDuplicateDataCollections() throws Exception {
        populateTable();
        startConnector();
        sendAdHocSnapshotSignal(tableDataCollectionId(), tableDataCollectionId());

        final int expectedRecordCount = 2*ROW_COUNT;
        Testing.Debug.enable();
        final Map<Integer, Integer> dbChanges = consumeMixedWithIncrementalSnapshot(expectedRecordCount);
        for (int i = 0; i < expectedRecordCount; i++) {
            assertThat(dbChanges).contains(entry(i + 1, i));
        }
    }
```
in AbstractIncrementalSnapshotTest.java

Then the problem is the function `consumeMixedWithIncrementalSnapshot` stores the consumed entries in a map. So, even thought we receive double the entries here, we are not able to consume 2*ROW_COUNT entries.

This needs to be refactored to fix this.

mfvitale:
> @mfvitale / @jpechane It is right now not feasible to add test for this case. The reason being if we add a test like this --
> 
> ```
> @Test
>     public void snapshotWithDuplicateDataCollections() throws Exception {
>         populateTable();
>         startConnector();
>         sendAdHocSnapshotSignal(tableDataCollectionId(), tableDataCollectionId());
> 
>         final int expectedRecordCount = 2*ROW_COUNT;
>         Testing.Debug.enable();
>         final Map<Integer, Integer> dbChanges = consumeMixedWithIncrementalSnapshot(expectedRecordCount);
>         for (int i = 0; i < expectedRecordCount; i++) {
>             assertThat(dbChanges).contains(entry(i + 1, i));
>         }
>     }
> ```
> 
> in AbstractIncrementalSnapshotTest.java
> 
> Then the problem is the function `consumeMixedWithIncrementalSnapshot` stores the consumed entries in a map. So, even thought we receive double the entries here, we are not able to consume 2*ROW_COUNT entries.
> 
> This needs to be refactored to fix this.

Try to look to more low lever method in `AbstractConnectorTest` like `consumeRecords`

akanimesh7:
@mfvitale Thanks for that. I have added the required test.

mfvitale (approved):

akanimesh7:
@jpechane Can you take a pass ?

jpechane:
@akanimesh7 Applied, thanks!

Enigma25:
Nice!

PR #4896, opened 2023-09-30
DBZ-6982 Provide configuration option to exclude extension attributes from a CloudEvent

rkudryashov:


jpechane:
@rkudryashov Applied, thanks a lot! Could you please provide a separate PR with docs update? Thanks

rkudryashov:
@jpechane thanks, I'll update docs later this week

PR #4897, opened 2023-10-02
DBZ-6983 Add the ability to sanitize field name when inferencing json schema

harveyyue:
https://issues.redhat.com/browse/DBZ-6983


jpechane:
@harveyyue Applied thanks! I'd like to ask you for a favour. This PR was quite difficult to review as there were many functionally unrelated changes done. Could you please next time separate the relavant and non-relevnt code changes into separate commits? It would make it easier for me. Thanks for understanding

PR #4903, opened 2023-10-02
DBZ-6987 Create schema in Postgres test if not exists

vjuranek:
https://issues.redhat.com/browse/DBZ-6987


jpechane:
@vjuranek Applied, thanks

PR #4909, opened 2023-10-03
Dbz 6989 refactor content in incremental snapshots partials files

roldanbob:
[DBZ-6989](https://issues.redhat.com/browse/DBZ-6989)

This change refactors the content to remove the recursive `include` directives from files in the `/partials` directory. It also removes a Head4 from the included content. Partials files should not contain `include` directives or headers.

In the downstream docs repo,  the `{partialsdir}` attribute that defines the path to the `/partials` directory is set to a relative path, rather than to an absolute path as in the upstream repo. Before this change, `include` directives in the parent connector files were used to reference shared files in the `/partials` directory, which, in turn contained other `include` directives to pull in content from another shared file in the directory. The resulting recursive references to the `/partials `directory caused the downstream build to fail.

After this change is merged, these commits should be forward ported to `main`.


jpechane:
@roldanbob Thanks, unfortunately it cannot be forward ported to main/2.4 without changes. Could you please prepare another PR for that when you''l have a time? Thanks

PR #4910, opened 2023-10-04
DBZ-6778 Refactor ElapsedTimeStrategy

vjuranek:
https://issues.redhat.com/browse/DBZ-6778


jpechane:
@vjuranek Applied, thanks. Could you please check vitess connector as I believe there is an extra call in tests that can be removed too.

vjuranek:
@jpechane you are right, fixed in https://github.com/debezium/debezium-connector-vitess/pull/161

PR #4911, opened 2023-10-04
DBZ-6996 Updating system tests to properly use MongoDB in a container

jcechace:
https://issues.redhat.com/browse/DBZ-6996

@obabec  @novotnyJiri  This ought to be cleaned up a bit and more importantly other implementations of mongo controllers should be addapted. Either merge this with a follow up or (ideally) just finish the work here before merging


jcechace:
@obabec skimmed through it and so far the changes look good. Let me know if you need a more in depth review once finished ;).

jcechace:
@obabec this can be closed, right?

novotnyJiri:
I am not Ondrej but we already merged the needed changes in https://github.com/debezium/debezium/pull/4950. Therefore yes, it can be closed.

PR #4917, opened 2023-10-06
DBZ-6763 Add integration tests for the Debezium Connect REST extension for the Oracle connector / Oracle Connect REST extension

rk3rn3r:
closes to https://issues.redhat.com/browse/DBZ-6763

### DIFF
--- a/.github/workflows/oracle-workflow-test.yml
+++ b/.github/workflows/oracle-workflow-test.yml
@@ -54,7 +54,6 @@ jobs:
         run: |
           if [[ $QUAY_IO_USERNAME == '' ]] && [[ $QUAY_IO_PASSWORD == '' ]]; then
             echo "HAS_QUAY_IO=false" >> $GITHUB_OUTPUT
-            exit -1
           else
             echo "HAS_QUAY_IO=true" >> $GITHUB_OUTPUT
           fi
--- a/.github/workflows/oracle-workflow-test.yml
+++ b/.github/workflows/oracle-workflow-test.yml
@@ -38,6 +38,7 @@ on:
 
 jobs:
   check:
+    name: Check Pull-Secrets
     runs-on: ubuntu-latest
     outputs:
       HAS_QUAY_IO: ${{ steps.skip.outputs.HAS_QUAY_IO }}
@@ -53,12 +54,13 @@ jobs:
         run: |
           if [[ $QUAY_IO_USERNAME == '' ]] && [[ $QUAY_IO_PASSWORD == '' ]]; then
             echo "HAS_QUAY_IO=false" >> $GITHUB_OUTPUT
+            exit -1
           else
             echo "HAS_QUAY_IO=true" >> $GITHUB_OUTPUT
           fi
 
-
-  build:
+  test-connector:
+    name: Oracle Connector
     runs-on: ubuntu-latest
     needs: [check]
     if: ${{ needs.check.outputs.HAS_QUAY_IO == 'true' }}
@@ -108,9 +110,67 @@ jobs:
           QUAY_IO_PASSWORD: ${{ secrets.QUAY_IO_PASSWORD }}
         run: >
           ./mvnw clean install -B -pl debezium-connector-oracle -am -Pinfinispan-buffer,oracle-docker,oracle-tests
-          -Ddocker.username="$QUAY_IO_USERNAME" 
-          -Ddocker.password="$QUAY_IO_PASSWORD" 
+          -Ddocker.username="$QUAY_IO_USERNAME"
+          -Ddocker.password="$QUAY_IO_PASSWORD"
+          -Dformat.formatter.goal=validate
+          -Dformat.imports.goal=check
+          -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
+          -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
+
+  test-rest-extension:
+    name: Oracle Connect REST Extension
+    runs-on: ubuntu-latest
+    needs: [check]
+    if: ${{ needs.check.outputs.HAS_QUAY_IO == 'true' }}
+    steps:
+      - name: Checkout Action
+        uses: actions/checkout@v4
+
+      - name: Get Date
+        id: get-date
+        run: |
+          echo "date=$(/bin/date -u "+%Y-%m")" >> $GITHUB_OUTPUT
+        shell: bash
+
+      - name: Set up Java 17
+        uses: actions/setup-java@v3
+        with:
+          distribution: 'temurin'
+          java-version: 17
+
+      # This workflow uses its own dependency cache rather than the main debezium workflow cache because
+      # we explicitly want to trigger this build on pushes separate from the other workflow.
+      - name: Cache Maven Repository
+        uses: actions/cache@v3
+        with:
+          path: ~/.m2/repository
+          # refresh cache every month to avoid unlimited growth
+          # we do not explicitly base this on POM changes since we are explicitly targeting a single connector
+          # and the likelihood that lots of dependencies will change that impacts Oracle is significantly low
+          # compared to other parts of Debezium that rely on frameworks such as Quarkus.
+          key: maven-oracle-${{ steps.get-date.outputs.date }}
+          restore-keys: |
+            maven-oracle-${{ steps.get-date.outputs.date }}
+
+      - name: Build and Install Debezium dependencies
+        run: >
+          ./mvnw clean install -pl debezium-bom,debezium-core,:debezium-ide-configs,:debezium-checkstyle,:debezium-revapi -am
+          -DskipTests=true 
+          -DskipITs=true 
           -Dformat.formatter.goal=validate 
           -Dformat.imports.goal=check 
-          -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn 
+          -Dhttp.keepAlive=false 
+          -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
+
+      - name: Build and Test Debezium Connector Oracle REST Extension
+        env:
+          QUAY_IO_USERNAME: ${{ secrets.QUAY_IO_USERNAME }}
+          QUAY_IO_PASSWORD: ${{ secrets.QUAY_IO_PASSWORD }}
+        run: >
+          ./mvnw clean install -B -pl debezium-connector-oracle -am -Poracle-tests,rest-assembly
+          -Dregistry.username="$QUAY_IO_USERNAME"
+          -Dregistry.password="$QUAY_IO_PASSWORD"
+          -Dformat.formatter.goal=validate
+          -Dformat.imports.goal=check
+          -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
           -Dmaven.wagon.http.pool=false -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
--- a/debezium-connector-oracle/pom.xml
+++ b/debezium-connector-oracle/pom.xml
@@ -125,6 +125,26 @@
             <artifactId>mockito-core</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.debezium</groupId>
+            <artifactId>debezium-testing-testcontainers</artifactId>
+            <scope>test</scope>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.junit.platform</groupId>
+                    <artifactId>junit-platform-launcher</artifactId>
+                </exclusion>
+                <exclusion>
+                    <groupId>org.junit.jupiter</groupId>
+                    <artifactId>junit-jupiter</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>io.rest-assured</groupId>
+            <artifactId>rest-assured</artifactId>
+            <scope>test</scope>
+        </dependency>
 
         <!-- Required by VerifyRecord -->
         <dependency>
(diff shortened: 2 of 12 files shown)


rk3rn3r:
@Naros if you look here, the error is back

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/OracleContainer.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    public String getDriverClassName() {
+        return "oracle.jdbc.OracleDriver";
+    }
+
+    protected void configure() {
+        super.configure();
+        withEnv("ORACLE_PWD", getPassword());
+        withEnv("ORACLE_SID", "ORCLCDB");
```
Naros, line 67: Could you use `getDatabaseName()` here so we only have one reference for that string in `preconfigure()`?

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/RestExtensionTestInfrastructure.java
```diff
@@ -155,8 +161,8 @@ public static void setupDebeziumContainer(String connectorVersion, String restEx
         final String debeziumVersion = debeziumContainerImageVersion.startsWith("1.2") ? "1.2.5.Final" : connectorVersion;
         String baseImageName = registry + "debezium/connect-base:" + debeziumContainerImageVersion;
         DEBEZIUM_CONTAINER = new DebeziumContainer(new ImageFromDockerfile("debezium/connect-rest-test:" + debeziumVersion)
-                .withFileFromPath(".", Paths.get("target/"))
-                .withFileFromPath("Dockerfile", Paths.get("src/test/resources/Dockerfile.rest.test"))
+                .withFileFromPath(".", Paths.get(System.getProperty("user.dir") + "/target/"))
+                .withFileFromPath("Dockerfile", Paths.get(System.getProperty("user.dir") + "/src/test/resources/Dockerfile.rest.test"))
```
Naros, line 165: IIRC `user.dir` assumes the directory from which the process starts, which may not necessarily be what you expect.  Would it be better to have maven (if it isn't already passed) to provide the project directory and we use that instead so that regardless of wherever the build is started from we have the right directory reference to those resources?

rk3rn3r: Yes, but: The code before was working for all connectors but not for the Oracle connector, I would like to get back the previous code that was supposed to work relatively to the module's base dir. But it was not working for the Oracle connector. Ideas?

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/OracleContainer.java
rk3rn3r: I might even not have to set this at all. I need to re-validate when tests are running.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/RestExtensionTestInfrastructure.java
Naros: Nope, there is absolutely nothing in the `pom.xml` that I see that would influence or change that behavior that is any different from other connectors.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/OracleContainer.java
rk3rn3r: I was able to remove it. It is not needed!

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/RestExtensionTestInfrastructure.java
rk3rn3r: I am using `project.basedir` and `project.build.directory` now. Fixed.

rk3rn3r:
Ready for a final review @Naros and @jpechane :partying_face:

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/OracleContainer.java
```diff
@@ -0,0 +1,65 @@
...
+        preconfigure();
+    }
+
+    public OracleContainer(Future<String> dockerImageName) {
+        super(dockerImageName);
+        preconfigure();
+    }
+
+    private void preconfigure() {
+        withUsername("c##dbzuser").withPassword("dbz").withDatabaseName("ORCLCDB");
```
Naros, line 46: We are specifying the `version.oracle.server` so that we can technically run the container with any flavor of Oracle. This means we could technically point this against an instance that doesn't support multitenancy.  In this case, the user `c##dbzuser` wouldn't be allowed since common-user designators are not permitted when there is no multi-tenancy.  I would perhaps recommend we source these from system properties much like the `DEFAULT_TAG` above so that this can be easily toggled.  wdyt?

Naros (commented):
Overall good work, just a few inline comments.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/OracleContainer.java
Naros: This also may mean that any test (which I think you have one) that explicitly looks for `ORCLPDB1` as the response would fail if we did point to non-CDB images as the only database there is the `ORCLCDB` one fwiw.

debezium-connector-oracle/pom.xml
```diff
@@ -470,6 +482,80 @@
...
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-failsafe-plugin</artifactId>
+                        <version>${version.failsafe.plugin}</version>
+                        <configuration>
+                            <systemPropertyVariables>
+                                <isAssemblyProfileActive>true</isAssemblyProfileActive>
+                            </systemPropertyVariables>
+                        </configuration>
+                    </plugin>
```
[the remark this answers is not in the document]
rk3rn3r: No, it's not a leftover. When building the assembly package at the time of the release the REST Extension tests should be run. This requires enabling the tests with that system property. I am not sure if we can tests this (on the Test Farm, as I suggested on the chat) because it won't work on GHA here because of too limited resources to run Testcontainers Oracle DB instance and the Maven Docker Plugin one at the same time.

debezium-parent/pom.xml
```diff
@@ -77,6 +77,10 @@
                     <configuration>
                         <argLine>${debug.argline} ${modules.argline} ${test.argline}</argLine>
                         <useSystemClassLoader>${useSystemClassLoader}</useSystemClassLoader>
+                        <systemPropertyVariables combine.children="append">
+                            <project.basedir>${project.basedir}</project.basedir>
+                            <project.build.directory>${project.build.directory}</project.build.directory>
+                        </systemPropertyVariables>
```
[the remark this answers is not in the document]
rk3rn3r: Sorry, wdym? This is the pom.xml of module `debezium-parent` and it's inheriting those system properties down. But in order to access those Maven properties in the tests it's mandatory to specify them in the maven-failsafe-plugin definition so they get exposed as env properties.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/OracleContainer.java
rk3rn3r: I was looking at the `pom.xml` of the Oracle connector it looks like we are only passing the version number like `<version.oracle.server>19.3.0</version.oracle.server>` there and it looked like that only the version is speficied there, and nowhere it was including the xs or noncdb "flavors"? The `xtreams` (and oracle-all) profiles don't seem to define another image using this system property, nor I found it in the release config (dir jenkins-jobs) or the debezium-testing-system module. Do you have a pointer to where we set it to something else? Is it only used for development? Maybe we can split this into the version number (that I expect alone in that property) and an optional "flavor" extension for other profiles?

A related question: Should very different images like the noncdb or xs images be just another tag on the same image, or should those be in different repos? Like: `rh_integration/dbz-oracle`, `rh_integration/dbz-oracle-xs`, `rh_integration/dbz-oracle-noncdb`? That would avoid such confusion. When those flavors are not compatible they should maybe not be different tags in the same repo but in separate repos?

debezium-parent/pom.xml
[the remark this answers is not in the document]
Naros: Ok, that's what I was trying to grok, thanks.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/OracleContainer.java
Naros: So by default, `version.oracle.server` points to the `19.3.0` image, which is the LogMiner multi-tenancy image; however, one could use `-Dversion.oracle.server=19.3.0-noncdb` and use the LogMiner image that does not include multitenancy.  In this case, the `c##dbzuser` won't be valid because users cannot be prefixed with `c##` in those builds; that's only permissible with multitenancy.

This specifically has to do with how the database is configured.  The `19.3.0` image with CDB enabled is essentially a single instance that contains a ROOT database and one pluggable database (PDB), aka a tenant.  Without the CDB feature, the single instance is the only database and there is no concept of common vs local users.

The POM profiles don't get into the specifics around multi-tenancy because the only difference is the credentials and connector configuration used, which can all be controlled using `-D` command line arguments.

Where you'll find some reference is in Oracle's `TestHelper#getDatabaseConfig(String)` method.  We basically consume all systems properties that start with `database.` and those allow the CLI to override the user name, password, database name, and pluggable database names.

> Should very different images like the noncdb or xs images be just another tag on the same image, or should those be in different repos? Like: rh_integration/dbz-oracle, rh_integration/dbz-oracle-xs, rh_integration/dbz-oracle-noncdb?

It's simply a matter of preference.  It's no different than MySQL having  `8.0.34`, `8.0.34-debian`, etc.  We could certainly think about changing this if its a problem, but it's certainly not something that I'd be comfortable doing right now, given how much reliance some of our tooling is now on those specific tags.

> If you think that this is not valid or you want to keep the current structure, sure, I can use system properties for configuration!

So all I am suggesting is that rather than these be strictly hardcoded, we allow the following to override the hardcoded values from the CLI if they're provided:

* `database.user` for the username
* `database.password` for the password
* `database.dbname` for the database name (references where you are using `ORCLCDB`)
* `database.pdb.name` for the pluggable database (references where you may be using `ORCLPDB1`).

rk3rn3r: Done (except `database.pdb.name`). I need some more guidance what exactly should change in the Oracle REST Extension tests. I filed https://issues.redhat.com/browse/DBZ-7091 for this.
Could you merge the PR when you are back?

debezium-connector-oracle/pom.xml
```diff
@@ -168,9 +188,9 @@
         <skipITs>true</skipITs>
 
         <!--
-        Specify properties that will be used for setting up the integration tests' Docker container.
-        Note that the `dockerhost.ip` property is computed from the IP addrses of DOCKER_HOST, which will
-        work on all platforms.  We'll set some of these system properties during integration testing.
+        Specify properties used for setting up the Oracle DB container instance/s for integration tests.
+        Note that the `docker.host.address` property is computed from the IP address of DOCKER_HOST, which will
```
rk3rn3r, line 192: @jpechane @Naros is this comment change accurate? I was not able to find any usage of the `dockerhost.ip` parameter.

.github/workflows/oracle-workflow-test.yml
```diff
@@ -53,12 +54,13 @@ jobs:
         run: |
           if [[ $QUAY_IO_USERNAME == '' ]] && [[ $QUAY_IO_PASSWORD == '' ]]; then
             echo "HAS_QUAY_IO=false" >> $GITHUB_OUTPUT
+            exit -1
```
Naros, line 57: This will cause any Oracle-related PR to always be in a failed state and will cancel any other pending action workflows that are running per the GH Action docs.  Can we not simply leave this as it is?

debezium-connector-oracle/pom.xml
Naros: Ya, it seems fine to me, likely just an oversight or maybe a situation where a variable name was adjusted but we failed to keep the comment aligned.

.github/workflows/oracle-workflow-test.yml
Naros: I added a commit to revert this change.

Naros (approved):
I believe this LGTM now.

.github/workflows/oracle-workflow-test.yml
rk3rn3r: @Naros I don't like that "solution". I think we then could remove that check at all because now it will always passes, right?
I suggest we either change the test code to work on PRs too or at least we change the Oracle CI test to not run the entire test suite but only the compile check on PRs and add another, new workflow for tests on the main branch.
Maybe we could even do: run the proper check here. Then, depending on the result we run either a compile-only test or the whole test suite when secrets are available. That way we can have PRs on the main repo here like mine that will work and we can test the whole suite. I would personally feel much more comfortable if I do changes to such a black box like Oracle DB connector.
If you and @jpechane like on of the options I can implement one of those solutions.

rk3rn3r:
@Naros I am fine to merge this PR now including reverting that check behavior, but maybe we can agree on an alternative what we do with the CI tests and this Credentials Check step and if we act based on the result. Wdyt?

jpechane:
@rk3rn3r @Naros Applied, thanks!

@rk3rn3r PLease create a follow-up Jira but I don't think it is a priority right now.

.github/workflows/oracle-workflow-test.yml
rk3rn3r: I filed https://issues.redhat.com/browse/DBZ-7103 for this.

PR #4918, opened 2023-10-06
DBZ-7013 Remove deprecated and unused code from embedded package

vjuranek:
https://issues.redhat.com/browse/DBZ-7013


jpechane:
@vjuranek Applied, thanks

PR #4919, opened 2023-10-06
DBZ-7015 Enable replication slot advance check

prburgu:
Reverts debezium/debezium#4456

The change in https://github.com/debezium/debezium/pull/4502 should address the underlying issue.

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
@@ -523,6 +523,16 @@ public class PostgresConnectorConfig extends RelationalDatabaseConnectorConfig {
                     "Whether or not to drop the logical replication slot when the connector finishes orderly. " +
                             "By default the replication is kept so that on restart progress can resume from the last recorded location");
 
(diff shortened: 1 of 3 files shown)


prburgu:
@jpechane We have been running with this patch for a few months and did not see any issue. The retries in the referred PR definitely solved the issue for us where it complained about the slot is in use.

From [Postgres docs](https://www.postgresql.org/docs/current/view-pg-replication-slots.html): 
```
confirmed_flush_lsn pg_lsn

The address (LSN) up to which the logical slot's consumer has confirmed receiving data. Data corresponding to the transactions committed before this LSN is not available anymore. NULL for physical slots.
```

I'm able to consistently use `pg_replication_slot_advance` to advance to the `confirmed_flush_lsn`. We should at least keep this feature as experimental behind a separate config. Without this feature there can be silent data loss (when slot is dropped for some reason), which is bad for production workloads.

jpechane:
This user claims that he stil experiences the warnings https://debezium.zulipchat.com/#narrow/stream/348249-community-postgresql/topic/https.3A.2F.2Fissues.2Eredhat.2Ecom.2Fbrowse.2FDBZ-6087

I am fine with intorudction of internal (undocumented) see `io.debezium.config.Field.createInternal(String)` config option that would enable this behaviour as opt-in

prburgu:
@jpechane I've updated the PR with a config for opt-in.

jpechane:
@prburgu Applied, thanks

PR #4921, opened 2023-10-08
DBZ-7016 `id` and `type` of a CloudEvent can be retrieved from headers

rkudryashov:
https://issues.redhat.com/browse/DBZ-7016


jpechane:
@rkudryashov Thanks for the PR. I've slightly different proposal for configuration so please let me know it is something you'd consider

- `MetadataLocation` will become three-state, adding `generate`
- the value of `metadata.location` will change. It will support global setting - only `HEADER/VALUE` supported and then there will be the name of field to be taken followd by the source, so example of the config would be `metadata.location=value,id:header,type:generate`. WDYT?

rkudryashov:
@jpechane LGTM

rkudryashov:
Please squash the commits when the PR will be ready for merging

rkudryashov:
@jpechane this is ready

jpechane:
@rkudryashov I think it i good. I like the solution. My only last comment is if there is a way how to get rid of all of those `if (metadataLocation.global() == MetadataLocationValue.HEADER)`?

rkudryashov:
@jpechane done

rkudryashov:
@jpechane thanks! Moved common code to separate method

jpechane:
@rkudryashov Applied, thanks a lot! Could you please prepare the docs PR?

rkudryashov:
@jpechane sure, just want to make one pr to rename metadata.location ->
metadata.source to make it more semantically correct after adding third
option, "generate", and using id and type inside the property

Also, is it too late to ask you to squash the commits?

PR #4923, opened 2023-10-09
DBZ-7018 Allow SQL statements greater than 40kb

Naros:
https://issues.redhat.com/browse/DBZ-7018

### Description	
In some corner cases, users may have unusually large SQL statements that need to be buffered due to the number of columns paired with the data in those columns. Previously we capped this to 4000*10 or 40kb primarily to address situations with LOB operations that could lead to OOM scenarios. The new code rather logs a warning when exceeding 100kb and hard faults ony when the connector sees Integer.MAX_VALUE number of SQL lines for a single SQL buffer.


thrix:
@Naros we found the issue, it was a bit funny, but we hope the issue is resolved

thrix:
nice and green :green_heart:

Naros:
Thanks for the help @thrix

Naros:
Applied to main (=2.5) and backported to branches 2.4 and 2.3.

PR #4924, opened 2023-10-09
DBZ-7020 Use optional schema for Timezone Converter tests

ani-sha:
https://issues.redhat.com/browse/DBZ-7020


jpechane:
@ani-sha Could you please add `VerifyRecord` calls to the tests as well so we make sure there is no other issue left?

jpechane:
@ani-sha Could you please add `VerifyRecord` after the transformation too? So the record is verified both befor and after transfromation? Thanks

ani-sha:
@jpechane Could you please give this a final review?

jpechane:
@ani-sha Applied, thanks!

PR #4926, opened 2023-10-09
DBZ-6975 Correctly handle escaped single-quotes in UNISTR arguments

Naros:
https://issues.redhat.com/browse/DBZ-6975


jpechane:
@Naros Applied, thanks

PR #4933, opened 2023-10-13
DBZ-7035 MySqlStreamingChangeEventSource will effectively stop during a blocking snapshot

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7035


jpechane:
@mfvitale Is there a way how to test it?

mfvitale:
> @mfvitale Is there a way how to test it?

I have tried but its complicated. I did a "manual" test checking that event stream didn't logs the receive of the event. 

Doing the same automatically means to relay strongly on the logs.

jpechane:
@mfvitale Sounds reasonable, applied. Thanks!

PR #4937, opened 2023-10-16
DBZ-7033 Upgrade Kafka to 3.6.0

jpechane:
https://issues.redhat.com/browse/DBZ-7033


jpechane:
@rk3rn3r Hi, could you please take a look at this one? `PrediccatedTransformation` class is no longer present in `connect-runtime` so listing of the predicates does not work.

rk3rn3r:
@jpechane I have run the tests locally for the REST extension and all were green. I re-triggered the CI jobs for the failed tests. What  I saw is that the transforms and the predicates are multiple times in the list. That seems to be a very strange bug then. Let's see what the new CI build says.

rk3rn3r:
@jpechane I have added the commit to remove that class that already got removed in 3.5. But this was not causing the issue. The issue was some CI weirdness. :man_shrugging:

PR #4938, opened 2023-10-16
DBZ-7042 Use oracle container registry for MySQL images

ani-sha:
https://issues.redhat.com/browse/DBZ-7042


ani-sha:
@jpechane MySQL 5.7 docker startup is failing with this oracle images
```
[ERROR] DOCKER> I/O Error [[debezium/mysql-server-test-database:latest]: Timeout after 60017 ms while waiting on log out 'MySQL init process done. Ready for start up.(?s)(.*)mysqld: ready for connections.']
```

jpechane:
@ani-sha Can you modify the CI and pom.xml that you can pass not only the tag but also the name of the image?
We intend to drop 5.7 soon so it probably does not make sense to invest too much time into investigation of what's going wrong.
The 5.7 tests would run against Docker Hub image and the rest against Oracle Hub.

jpechane:
@ani-sha Applied, thanks

PR #4940, opened 2023-10-16
DBZ-7044 Adapted DebeziumContainer to lazily retrieve latest stable…

MarkDucommun:
…image tag when needed.

https://issues.redhat.com/browse/DBZ-7044


jpechane:
@MarkDucommun Thanks for the PR. Where do you see the method used?
I see few problems with it

* The `DEBEZIUM_STABLE_TAG` is public so if any client was using it it will be now invalid. The class should be modified to make the attibutes private and have accessor methods
* The resolved version is not cached. In case of a test suite starting the container multiple times the net result migh be longer time associated with every test/usage

@obabec Is it something that affects you?

rk3rn3r:
@jpechane I don't know if your comment relates to an older code change (but based on the conversation history I think it doesn't), but  `DEBEZIUM_STABLE_TAG` was not public, the method `latestStable()` was public and still returns the same instance. The only difference is that the resolution of the latest stable tag is deferred to when `latestStable()` is called instead of when the static class is instantiated (which in general can slow down setup).

jpechane:
@rk3rn3r Thanks for the comment, I must be blind :-(. Yes, the comment related to `public` is irrelevant. Still the comenta relate to caching of the value stands.

MarkDucommun:
Hi @jpechane, @rk3rn3r - just wanted to touch base and see if the changes that I made yesterday are acceptable for lazily caching the version tag. I'm very motivated to get this merged into main so that we can have a bit more consistency back into our test runs.

jpechane (approved):
@MarkDucommun LGTM, thanks. Will merge afte CI run is completed.

PR #4943, opened 2023-10-17
DBZ-7040 Handle extra semicolon for create table

ani-sha:
https://issues.redhat.com/browse/DBZ-7040


jpechane:
@ani-sha Would not this be a problem for other statements too, not only CREATE TABLE?

ani-sha:
@jpechane I see the `;` added for various other statements too like `create_tablespace` and `create_context`. Should we add for all others too?

jpechane:
I think we need to take a different approach to that problem. What if there will be 2 semicolons?
Is it the same problem for MySQL? If not, why?
As the robust solution probably the best approach would not be solving at the grammar level but at Debezium code level. I can envision that before we would pass the statement to parser we would

* strip all trailing whitespaces
* remove the semicolon if it exists
* repeat until there is no semicolon at the end
* pass the statement to the parser (and add 1 semicolon back if needed)

ani-sha:
@jpechane That's a good idea, closing this PR.

PR #4944, opened 2023-10-17
DBZ-7047 Infinispan Embedded Performance with Oracle

Naros:
https://issues.redhat.com/browse/DBZ-7047

Using parallel streams, this causes a severe loss of performance.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
@@ -20,6 +20,7 @@ import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -237,10 +238,12 @@ public abstract class AbstractLogMinerEventProcessor<T extends AbstractTransacti
                         metrics.getSleepTimeInMilliseconds());
 
                 if (metrics.getNumberOfActiveTransactions() > 0) {
-                    LOGGER.debug("All active transactions: {}",
-                            getTransactionCache().values().stream()
-                                    .map(t -> t.getTransactionId() + " (" + t.getStartScn() + ")")
-                                    .collect(Collectors.joining(",")));
+                    // This is wrapped in try-with-resources specifically for Infinispan performance
+                    try (Stream<T> stream = getTransactionCache().values().stream()) {
+                        LOGGER.debug("All active transactions: {}",
+                                stream.map(t -> t.getTransactionId() + " (" + t.getStartScn() + ")")
+                                        .collect(Collectors.joining(",")));
+                    }
                 }
 
                 metrics.setLastProcessedRowsCount(counters.rows);
@@ -612,6 +615,13 @@ public abstract class AbstractLogMinerEventProcessor<T extends AbstractTransacti
     protected abstract void finalizeTransactionCommit(String transactionId, Scn commitScn);
 
     /**
+     * Returns only the first transaction id in the transaction buffer.
+     *
+     * @return the first active transaction in the buffer, or {@code null} if there is none.
+     */
+    protected abstract String getFirstActiveTransactionKey();
+
+    /**
      * Check whether the supplied username associated with the specified transaction is excluded.
      *
      * @param transaction the transaction, never {@code null}
@@ -708,7 +718,7 @@ public abstract class AbstractLogMinerEventProcessor<T extends AbstractTransacti
                 advanceLowerScnBoundary = true;
             }
             else if (activeTransactions == 1) {
-                final String transactionId = getTransactionCache().keySet().iterator().next();
+                final String transactionId = getFirstActiveTransactionKey();
                 if (transactionId.equals(row.getTransactionId())) {
                     // The row's transaction is the current and only active transaction.
                     advanceLowerScnBoundary = true;
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
@@ -237,7 +237,7 @@ public abstract class AbstractLogMinerEventProcessor<T extends AbstractTransacti
                         offsetContext.getCommitScn(), metrics.getNumberOfActiveTransactions(),
                         metrics.getSleepTimeInMilliseconds());
 
-                if (metrics.getNumberOfActiveTransactions() > 0) {
+                if (metrics.getNumberOfActiveTransactions() > 0 && LOGGER.isDebugEnabled()) {
                     // This is wrapped in try-with-resources specifically for Infinispan performance
                     try (Stream<T> stream = getTransactionCache().values().stream()) {
                         LOGGER.debug("All active transactions: {}",
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/AbstractInfinispanLogMinerEventProcessor.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/AbstractInfinispanLogMinerEventProcessor.java
@@ -362,6 +362,12 @@ public abstract class AbstractInfinispanLogMinerEventProcessor extends AbstractL
     /**
      * Purge the necessary caches with all entries that occurred prior to the specified change number.
      *
+     * NOTE: This method is abstract despite the code used by both all implementations being identical.
+     * This is because the method needed {@code entrySet()} is made available on two different concrete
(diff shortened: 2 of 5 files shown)


Naros:
@vjuranek, when you are able, would you mind checking whether this gives you any performance gain?  I was able to see some significant performance gain from the `parralelStream` to the `stream` usage; however, Jeremy Ford mentioned the iterators, so I've added that too.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
```diff
@@ -237,10 +238,12 @@ public Scn process(Scn startScn, Scn endScn) throws SQLException, InterruptedExc
                         metrics.getSleepTimeInMilliseconds());
 
                 if (metrics.getNumberOfActiveTransactions() > 0) {
```
jeremy-l-ford, line 240: Could change this to 

`if (metrics.getNumberOfActiveTransactions() > 0 && LOGGER.isDebugEnabled()) {`

to avoid iterating over the cache when not needed.

Naros: Should also review the other cache usages, pointed out [here](https://debezium.zulipchat.com/#narrow/stream/348250-community-oracle/topic/Poor.20performance.20when.20using.20infinispan.20caches/near/397111756).

Naros:
Thanks for the quick review, looks like there may be a few other tweaks we should do too.

Naros:
@jeremy-l-ford This should cover all the bases for the touch points for the caches.  As I mentioned on Zulip, I did confirm with the ISPN team that only `RemoteCache` implementations are affected by the fact that they do not automatically clean up after iterators and streams when using the standard Java API.  But for completeness, I applied it to both in our code to be explicitly sure.

Naros:
Logged [DBZ-7072](https://issues.redhat.com/browse/DBZ-7072) for the unrelated error(s).

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/EmbeddedInfinispanLogMinerEventProcessor.java
```diff
@@ -180,6 +180,12 @@ protected String getFirstActiveTransactionKey() {
         return null;
     }
 
+    @Override
+    protected void purgeCache(Scn minCacheScn) {
+        removeIf(processedTransactionsCache.entrySet().iterator(), entry -> Scn.valueOf(entry.getValue()).compareTo(minCacheScn) > 0);
```
jeremy-l-ford, line 185: Seems like the logic here as changed?  Originally, the this was 

`
getProcessedTransactionsCache().entrySet().removeIf(entry -> Scn.valueOf(entry.getValue()).compareTo(minCacheScn) < 0);
            getSchemaChangesCache().entrySet().removeIf(entry -> Scn.valueOf(entry.getKey()).compareTo(minCacheScn) < 0);
`

whereas now it is > 0?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/RemoteInfinispanLogMinerEventProcessor.java
```diff
@@ -191,6 +191,12 @@ protected String getFirstActiveTransactionKey() {
         return null;
     }
 
+    @Override
+    protected void purgeCache(Scn minCacheScn) {
+        removeIf(processedTransactionsCache.entrySet().iterator(), entry -> Scn.valueOf(entry.getValue()).compareTo(minCacheScn) > 0);
```
jeremy-l-ford, line 196: Same as before, seems like logic has changed.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/AbstractInfinispanLogMinerEventProcessor.java
```diff
@@ -356,6 +359,33 @@ protected Scn calculateNewStartScn(Scn endScn, Scn maxCommittedScn) throws Inter
         }
     }
 
+    /**
+     * Purge the necessary caches with all entries that occurred prior to the specified change number.
+     *
+     * @param minCacheScn the minimum system change number to keep entries until
+     */
+    protected abstract void purgeCache(Scn minCacheScn);
```
jpechane, line 367: @Naros It seems to me that both implementations of the method are the same, why does it need to be introduced as an abstract one?

jpechane (commented):
@Naros In general LGTM. I left one question related to the need of abstracton.

Feel free to merge yourself when you feel it is useful.

vjuranek:
@Naros sorry for the long delay, I finally get to it.

I run it with our JHM [EndToEndPerf](https://github.com/debezium/debezium/blob/main/debezium-microbenchmark-oracle/src/main/java/io/debezium/performance/connector/oracle/EndToEndPerf.java) adjusted to use ISPN embedded and ISPN configured to use SIFS with passivation off and write through setup.

The results are bellow.

Given the high volatility of the results, I'm not sure we can conclude there is any performance gain. There is a signal there is *could* be perf. boost for higher number of records and online catalog, but IMHO for redo catalog we cannot do any conclusion. Maybe the only only conclusion we can make is that there isn't any *significant* perf. boost. There could be smaller one, which may be not visible due to running on very unstable environment (my laptop).


#### ISPN embedded without patch (current main branch)
* first run 
 
```
Benchmark             (dmlEvents)  (miningStrategy)  Mode  Cnt   Score   Error  Units
EndToEndPerf.capture         1000  redo_log_catalog    ss       31.106           s/op
EndToEndPerf.capture         1000    online_catalog    ss        5.084           s/op
EndToEndPerf.capture         5000  redo_log_catalog    ss       33.031           s/op
EndToEndPerf.capture         5000    online_catalog    ss        7.508           s/op
EndToEndPerf.capture        10000  redo_log_catalog    ss       38.972           s/op
EndToEndPerf.capture        10000    online_catalog    ss        9.654           s/op
```

* second run

```
EndToEndPerf.capture         1000  redo_log_catalog    ss       33.832           s/op
EndToEndPerf.capture         1000    online_catalog    ss        4.882           s/op
EndToEndPerf.capture         5000  redo_log_catalog    ss       34.361           s/op
EndToEndPerf.capture         5000    online_catalog    ss        7.038           s/op
EndToEndPerf.capture        10000  redo_log_catalog    ss       50.811           s/op
EndToEndPerf.capture        10000    online_catalog    ss       11.563           s/op
```

#### ISPN embedded with the patch
* first run
 
```
EndToEndPerf.capture         1000  redo_log_catalog    ss       31.314           s/op
EndToEndPerf.capture         1000    online_catalog    ss        4.691           s/op
EndToEndPerf.capture         5000  redo_log_catalog    ss       32.111           s/op
EndToEndPerf.capture         5000    online_catalog    ss        6.936           s/op
EndToEndPerf.capture        10000  redo_log_catalog    ss       45.673           s/op
EndToEndPerf.capture        10000    online_catalog    ss        6.468           s/op
```

* secodn run

```
EndToEndPerf.capture         1000  redo_log_catalog    ss       46.762           s/op
EndToEndPerf.capture         1000    online_catalog    ss        4.733           s/op
EndToEndPerf.capture         5000  redo_log_catalog    ss       43.959           s/op
EndToEndPerf.capture         5000    online_catalog    ss        7.274           s/op
EndToEndPerf.capture        10000  redo_log_catalog    ss       45.894           s/op
EndToEndPerf.capture        10000    online_catalog    ss        6.561           s/op
```

vjuranek (approved):
Although I don't see any significat perf. boost, the changes are definitely good.

Naros:
Hi @vjuranek, I think you're right on the money with the fact in the smaller data sets, we won't see much improvement, and I think a lot of that is dependent on the fact that the data remains resident in memory, and there may be little failover to disk that causes disk fetches.

Based on my conversations with the Infinispan team, a lot of the performance (good and bad) we see can be heavily influenced by the user's configuration.  I wonder whether it would be prudent to add a hyperlink in our documentation to the upstream ISPN docs with a TIP/NOTE that users should refer to the ISPN upstream documentation for a complete guide on all the performance and tuning options available.  wdyt?

vjuranek:
@Naros IMHO this can be two-edged sword as various config options can have various consequences for Debezium. E.g. user may configure async (write behind) storage as it obviously speed the things up, but may not realize that this could result in Debezium issues when e.g. Debezium crashes and restarts as some data may not be written to the storage. In general I think this is a good idea (to link ISPN documentation and mention possible perf. boost), but we should also provide sufficient context how it can impact Debezium. Possibly not as a part of Debezium documentation (as it may be long, not exhaustive etc.), but possibly we may write e.g. blog post where we would discuss selected ISPN options and their possible impact on Debezium and possible perf. improvements caused by tuning ISPN parameters (in some specific situations) and besides linking ISPN documentation link also this blog post for broader context? WDYT? If sounds good to you and it's fine to do it in something like ~2 months time frame, I can work on the blog post.

Naros:
That sounds like a great idea @vjuranek :heavy_check_mark:

Naros:
Applied to main (=2.5) and branch 2.4.

PR #4945, opened 2023-10-17
DBZ-2913 Preview MariaDB first-class support

Naros:
https://issues.redhat.com/browse/DBZ-2913

- [x] Get tests running with MariaDB 11.1.2 (basic deployment)
- [x] SSL-based tests fail, need to understand why
- [ ] Determine how to best handle JSON; MariaDB aliases this to LONGTEXT like old MySQL 5.x.
- [x] Provide skip checks based on version and MariaDB vs MySQL
- [x] Add profile to run MariaDB rather than MySQL
- [x] Add GitHub action workflow for basic MariaDB testing
- [ ] GTID support

### Identified differences

* MySQL `binlog_rows_query_log_events` equates to MariaDB `binlog_annotate_row_events`.
* By default, MySQL rounds time precision up, MariaDB truncates; controlled via `TIME_ROUND_FRACTIONAL`
* ~~MariaDB does not emit `QUERY` events with `BEGIN` text, which causes `EventBuffer` to spam logs.~~ Solved by making the MariaDB capability setting configurable in the Binlog Client.
* ~~There are differences in binlog compression between MariaDB and MySQL.  Specifically in the `TransactionPayloadIT`, certain events are missed when enabling compression for MariaDB that are not with MySQL, and disabling compression, the test passes for both sources.~~ Solved by setting `CHAR_AND_BINARY_AS_BYTE_ARRAY` capability on the event deserializer specifically when using MariaDB.
* ~~GTID mode is always enabled with MariaDB, so any test or check that looks for `gtid_mode` variable fails.  This affects `ReadOnlyIncrementalSnapshotIT` and `ZZZGtidSetIT`.~~ Solved by explicitly looking for `gtid_strict_mode` for MariaDB as that's the equivalent to MySQL's `gtid_mode`.
* Currently, the use of `mysql-ssl` profile fails.  Additionally, when trying to create such a profile for MariaDB, I found that the order of the init scripts was the opposite of MySQL. I have a fix for execution order that gets MariaDB to throw the same error as MySQL with this profile, which leads me to believe it's related to the certificates we install.

### Upstream Binlog Client PRs

- https://github.com/osheroff/mysql-binlog-connector-java/pull/125


Naros:
> DBZ-2913 MariaDB image has SSL out of the box

Just a note to myself to reword this commit. It should read "has no SSL".

jpechane:
@Naros Great job, applied, thanks a lot!

As a follow-up, please
* create seaparate Jiras for GTID and JSON support
* refator the code slightly by introducing strategy/adapter so the code is not littered with ` isMariaDb()` calls

Naros:
Logged https://issues.redhat.com/browse/DBZ-7082 and https://issues.redhat.com/browse/DBZ-7083.

PR #4912, opened 2023-10-05
DBZ-6763 Add integration tests for the Debezium Connect REST extension for the Oracle connector / Oracle Connect REST extension

rk3rn3r:
closes to https://issues.redhat.com/browse/DBZ-6763

PR #4947, opened 2023-10-18
DBZ-6763 Fix CI tests caused by MongoDbReplicaSet which is trying to stop not running containers almost infinitely

rk3rn3r:
relates to https://issues.redhat.com/browse/DBZ-6763

### DIFF
--- a/debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
+++ b/debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
@@ -5,7 +5,6 @@
  */
 package io.debezium.testing.testcontainers;
 
-import static io.debezium.testing.testcontainers.MongoDbContainer.node;
 import static io.debezium.testing.testcontainers.util.DockerUtils.logDockerDesktopBanner;
 import static java.util.concurrent.TimeUnit.MINUTES;
 import static java.util.concurrent.TimeUnit.SECONDS;
@@ -14,6 +13,7 @@ import static org.awaitility.Awaitility.await;
 
 import java.net.URLEncoder;
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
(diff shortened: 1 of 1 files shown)


rk3rn3r:
@Naros  @jpechane  Here's the extracted commit from #4912 that should fix REST extension test issues.

Naros:
Thanks @rk3rn3r

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
```diff
@@ -368,6 +370,11 @@ public List<String> getHostNames() {
                 .collect(Collectors.toList());
     }
 
+    public MongoDbReplicaSet withStartupTimeout(Duration startupTimeout) {
```
jcechace, line 373: The entire class is not designed as a fluent builder. I don't see a reason why a single message should behave differently. This should be in the builder. Kindly place it there and be careful about mixing various design patterns into single thing.

Thank you.

rk3rn3r: @jcechace This just goes back to your own comment here:
https://github.com/debezium/debezium/blob/main/debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/RestExtensionTestInfrastructure.java#L135-L140

The Oracle Connect REST extension PR will then use this method instead of the code linked, but I decided to add it to this bugfix PR so it can be used early before we finish that other PR. Maybe you want to log a Jira to change and refactor this part/class?

jcechace: @rk3rn3r  There is a structure to those classes. Put it into the builder, pass it from the builder to the RS class and then over to the underlying containers in a similar fashion. 

I don't see a need for Jira. The code was not refactored correctly and the PR was merged before I could review it. Please make it consistent with the rest of the class design. Thank you.

rk3rn3r: @jcechace This is not how we work and I only applied a change that you suggested in the comment that I linked above, which was a good comment. The class is `Startable` and suggested to reflect a running container. Thus, a timeout config makes sense. Feel free to fix it that it fits your individual idea of that class or provide a Jira explaining what you want to achieve.

jcechace: https://issues.redhat.com/browse/DBZ-7054

PR #4948, opened 2023-10-18
DBZ-6873 Support MySQL 8.2

ani-sha:
https://issues.redhat.com/browse/DBZ-6873


jpechane (changes requested):
@ani-sha PLease take a look at test results in closer detail. There are almost 60 failures.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
```diff
@@ -611,6 +611,15 @@ public static SecureConnectionMode parse(String value, String defaultValue) {
             .withDescription("The password for the trust store file. "
                     + "Used to check the integrity of the truststore, and unlock the truststore.");
 
+    public static final Field ALLOW_PUBLIC_KEY_RETRIEVAL = Field.create("database.allowPublicKeyRetrieval")
```
jpechane, line 614: I don't think wee need the config property for it. It is just `driver.` pass-thorugh property.

jpechane (commented):
@ani-sha Good job for geting it running. I just think the new property is not necessary, it is just JDBC driver passthrough property so it does not need to be explicitly added.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
jpechane: Thinking about that as it uses `database.` prefix which is not supported for Debezium 2.x is the config option even needed for 8.2 at all?

ani-sha: Yes, I have removed the field. The tests are building now. The test failure seems unrelated.

jpechane:
@ani-sha Applied, thanks

PR #4957, opened 2023-10-24
DBZ-7066 The DefaultDeleteHandlingStrategy couldn't add the rewrite "__deleted" field to a non-struct value

harveyyue:
https://issues.redhat.com/browse/DBZ-7066


jpechane:
@harveyyue  I think we dropped the ball on testing of this change a bit. Could you please make a copy of `ExtractNewDocumentStateTestIT`, `ExtractNewRecordStateTest` and `ExtractNewDocumentStateTest` and modify it so it uses the new config option? Ideally there should be no change in asserts, only in config. Only after that we can be sure everything has been caught.

harveyyue:
> @harveyyue I think we dropped the ball on testing of this change a bit. Could you please make a copy of `ExtractNewDocumentStateTestIT`, `ExtractNewRecordStateTest` and `ExtractNewDocumentStateTest` and modify it so it uses the new config option? Ideally there should be no change in asserts, only in config. Only after that we can be sure everything has been caught. Or more precisely make a copy that will test the legacy setting and modify the original ones.

@jpechane OK

jpechane:

PR #4958, opened 2023-10-24
DBZ-7067 Improve Oracle logging for COMMITs

Naros:
https://issues.redhat.com/browse/DBZ-7067


Naros:
Checked the XStream test locally and it passed.
As this is a trivial change, I'm going to proceed and merge this myself.

PR #4959, opened 2023-10-24
DBZ-7037 com.github.shyiko.mysql.binlog.event.deserialization.MissingTableMapEventException: No TableMapEventData has been found for table id:137

sean-k1:
https://issues.redhat.com/browse/DBZ-5891
https://issues.redhat.com/browse/DBZ-7026
https://issues.redhat.com/browse/DBZ-7037

Fake Rotates Event skip tableMapEventByTableId clear

When the connection to the Mysql database is lost, a fake Rotate Event occurs in communication with the master. Currently, even when initializing tableMapEventByTableId when a fake Rotate Event occurs, an error occurs because the corresponding tableMapEvent for the tableId cannot be found.

I think this could be the issue with this problem. What do you think?

https://mariadb.com/kb/en/fake-rotate_event/
https://github.com/go-mysql-org/go-mysql/blob/3a75f6a26d8819c2a88629e54c7de0b7698ae05c/replication/backup.go#L59

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java
@@ -255,8 +255,7 @@ public class MySqlStreamingChangeEventSource implements StreamingChangeEventSour
                     }
 
                     // DBZ-5126 Clean cache on rotate event to prevent it from growing indefinitely.
-                    if (event.getHeader().getEventType() == EventType.ROTATE
-                        && event.getHeader().getTimestamp() != 0) {
+                    if (event.getHeader().getEventType() == EventType.ROTATE && event.getHeader().getTimestamp() != 0) {
                         tableMapEventByTableId.clear();
                     }
                     return event;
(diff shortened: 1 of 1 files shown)


Naros:
Hi @sean-k1, I am curious whether it may make more logical sense for this to be something we push upstream into the Binlog client itself.  This has the benefit that it helps anyone else using the client, not only Debezium whereas this fix only addresses the problem from Debezium's perspective.  

I'd say that if we can confirm this fixes the reproducer, we should ideally push this upstream.

jpechane:
In the meantime I'd be fine to merge it as a workaround until the fixed client is available.

sean-k1:
@Naros Are you talking about handling the binlog client ignoring the fake rotateEvent?

Naros:
> @Naros Are you talking about handling the binlog client(mysql-binlog-connector-java) ignoring the fake rotateEvent?

That's correct.

sean-k1:
@Naros @jpechane 
Okay As you know, fake rotate event is not left in the binlog file.
I think we should talk to the binlog-client owner about the issue.

It took me almost a month to contribute to mysql-binlog-connector-java last time, I think the owner is very busy. Please note 
Until then, I think it's a good idea to put this in debezium as defensive logic.

jpechane:
@sean-k1 Applied, thanks!
We are in alpha phase so IMHO it is good to go to have it as workaround until binlog client is done.
Test failure seems to be transient.

PR #4961, opened 2023-10-25
DBZ-7071 Introduce internal schema change exclusion option

Naros:
https://issues.redhat.com/browse/DBZ-7071

Add a new internal `log.mining.schema_changes.username.exclude.list` to allow users to customize the default behavior for excluding the SYS and SYSTEM usernames from DDL changes.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -584,6 +584,14 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             .withValidation(OracleConnectorConfig::validateRequiredWhenUsingOpenLogReplicator)
             .withDescription("The port of the OpenLogReplicator network service");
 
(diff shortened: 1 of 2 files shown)


Naros:
Failures are unrelated to the change.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -781,6 +791,7 @@ public OracleConnectorConfig(Configuration config) {
         this.logMiningRestartConnection = config.getBoolean(LOG_MINING_RESTART_CONNECTION);
         this.logMiningMaxScnDeviation = Duration.ofMillis(config.getLong(LOG_MINING_MAX_SCN_DEVIATION_MS));
         this.logMiningInifispanGlobalConfiguration = config.getString(LOG_MINING_BUFFER_INFINISPAN_CACHE_GLOBAL);
+        this.logMiningSchemaChangesUsernameExcludes = Strings.setOf(config.getString(LOG_MINING_SCHEMA_CHANGES_USERNAME_EXCLUDE_LIST), String::new);
```
jpechane, line 794: I am going to keep it here as it is the same as in other case of the class but IMHO `Function.identity()` would better reflect the meaning.

jpechane:
@Naros Applied, thanks!

PR #4963, opened 2023-10-25
DBZ-7060 Rename `metadata.location` -> `metadata.source`

rkudryashov:
https://issues.redhat.com/browse/DBZ-7060


jpechane:
Applied, given the feature is present in codebase for very short time and relatively less used we are replacing without going through deprecation path.

jpechane:
@rkudryashov Thanks!

PR #4964, opened 2023-10-26
DBZ-6963 Do not merge events if table has no LOB columns

Naros:
https://issues.redhat.com/browse/DBZ-6963


jpechane (changes requested):
The test failures are in the new test so it is related.

Naros:
Oops, that's my mistake on that @jpechane.  Should be fixed now.

jpechane:
@Naros Applied, thanks!

PR #4970, opened 2023-10-30
DBZ-7065 Add `PrepareCloudEvent` transform

rkudryashov:
This fixes https://issues.redhat.com/browse/DBZ-7065

The transform converts a CloudEvent to a structure suitable for `JdbcSinkConnector`. It uses provided by a user mapping between a CloudEvent's fields and names of database columns. The resulting value schema has no name. A CloudEvent's `data` field is flattened if needed

### DIFF
--- a/debezium-core/src/main/java/io/debezium/transforms/PrepareCloudEvent.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/PrepareCloudEvent.java
@@ -1,0 +1,155 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.transforms;
+
+import static org.apache.kafka.connect.transforms.util.Requirements.requireStruct;
+
+import java.util.Collections;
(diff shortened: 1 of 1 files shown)


debezium-core/src/main/java/io/debezium/transforms/PrepareCloudEvent.java
```diff
@@ -0,0 +1,155 @@
...
+import io.debezium.transforms.outbox.AdditionalFieldsValidator;
+
+/**
+ * The transform converts a CloudEvent to a structure suitable for `JdbcSinkConnector`. It uses provided by a user
+ * mapping between a CloudEvent's fields and names of database columns. The resulting value schema has no name. A
+ * CloudEvent's `data` field is flattened if needed
+ *
+ * @author Roman Kudryashov
+ */
+public class PrepareCloudEvent<R extends ConnectRecord<R>> implements Transformation<R> {
```
jpechane, line 38: Hi, the SMT name is a bit too generic and it realy does not describe its purpose.
Also is it a generic SMT or it is useful for JDBC Sink connector only? If the latter then I propose to move it to that repository.

PR #4975, opened 2023-11-02
DBZ-7085 Add MariaDB driver and use with MariaDB profile

Naros:
https://issues.redhat.com/browse/DBZ-7085

### DIFF
--- a/debezium-bom/pom.xml
+++ b/debezium-bom/pom.xml
@@ -200,6 +200,13 @@
                 <version>1.0.2</version>
             </dependency>
 
+            <!-- MariaDB JDBC driver -->
+            <dependency>
+                <groupId>org.mariadb.jdbc</groupId>
+                <artifactId>mariadb-java-client</artifactId>
+                <version>${version.mariadb.driver}</version>
+            </dependency>
+
             <!-- MongoDB Java driver -->
             <dependency>
                 <groupId>org.mongodb</groupId>
--- a/debezium-connector-mysql/pom.xml
+++ b/debezium-connector-mysql/pom.xml
@@ -57,6 +57,10 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.mariadb.jdbc</groupId>
+            <artifactId>mariadb-java-client</artifactId>
+        </dependency>
 
         <!-- Testing -->
         <dependency>
@@ -172,6 +176,8 @@
         <mysql.init.timeout>60000</mysql.init.timeout> <!-- 60 seconds -->
         <apicurio.port>8080</apicurio.port>
         <apicurio.init.timeout>60000</apicurio.init.timeout> <!-- 60 seconds -->
+        <mysql.database.protocol>jdbc:mysql</mysql.database.protocol>
+        <mysql.database.jdbc.driver>com.mysql.cj.jdbc.Driver</mysql.database.jdbc.driver>
         <!--
         MariaDB specific
         -->
@@ -541,6 +547,9 @@
                         <database.password>${mysql.password}</database.password>
                         <database.replica.hostname>${docker.host.address}</database.replica.hostname>
                         <database.replica.port>${mysql.replica.port}</database.replica.port>
+                        <!-- Specifies which driver to use for the tests -->
+                        <database.protocol>${mysql.database.protocol}</database.protocol>
+                        <database.jdbc.driver>${mysql.database.jdbc.driver}</database.jdbc.driver>
                         <skipLongRunningTests>${skipLongRunningTests}</skipLongRunningTests>
                         <database.ssl.mode>disabled</database.ssl.mode>
                     </systemPropertyVariables>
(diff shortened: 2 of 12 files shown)


debezium-bom/pom.xml
```diff
@@ -200,6 +200,13 @@
                 <version>1.0.2</version>
             </dependency>
 
+            <!-- MariaDB JDBC driver -->
+            <dependency>
+                <groupId>org.mariadb.jdbc</groupId>
+                <artifactId>mariadb-java-client</artifactId>
+                <version>${version.mariadb.driver}</version>
```
rk3rn3r, line 207: I am not sure about this but maybe should drivers be `provided` scope?!

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/UniqueDatabase.java
```diff
@@ -193,6 +193,10 @@ public Configuration.Builder defaultJdbcConfigBuilder() {
         Builder builder = Configuration.create()
                 .with(MySqlConnectorConfig.HOSTNAME, System.getProperty("database.hostname", "localhost"))
                 .with(MySqlConnectorConfig.PORT, System.getProperty("database.port", "3306"))
+                .with(MySqlConnectorConfig.JDBC_PROTOCOL, System.getProperty("database.protocol",
+                        MySqlConnectorConfig.JDBC_PROTOCOL.defaultValueAsString()))
+                .with(MySqlConnectorConfig.JDBC_DRIVER, System.getProperty("database.jdbc.driver",
+                        MySqlConnectorConfig.JDBC_DRIVER.defaultValueAsString()))
```
rk3rn3r, line 199: Wouldn't it make it easier for users when they have a single switch that sets the backend to on of the supported MySQL types, like mysql, percona, mariadb, etc and we set related parameters accordingly instead of providing that full detail level?

debezium-connector-mysql/pom.xml
```diff
@@ -57,6 +57,10 @@
                 </exclusion>
             </exclusions>
         </dependency>
+        <dependency>
+            <groupId>org.mariadb.jdbc</groupId>
+            <artifactId>mariadb-java-client</artifactId>
```
rk3rn3r, line 62: I am not sure about this but maybe should drivers be `provided` scope?!

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/UniqueDatabase.java
Naros: Yes, there is a Jira to introduce a strategy pattern for these use cases.  But rather than shoehorn that into the code, we're doing this piecemeal so we see all the touch points we need to refactor more clearly over the 2.5 release cycle.

debezium-connector-mysql/pom.xml
Naros: We've always shipped all open-source drivers with the connectors to simplify the installation for users.  If a user needs or requires a specific driver, they're free to replace the driver as needed.

jpechane: Yes, both drivers hould be shipped, esp. if they can co-exist side-by-side.

Naros:
Speaking of co-existence @jpechane, see my recent commit about "fetchSize".  

> Per https://jira.mariadb.org/browse/CONJ-977, the MariaDB team does not
believe that the Integer.MIN_VALUE solution for triggering streaming is
spec compliant and as of MariaDB 3.x+ drivers, they explicitly will not
allow this.
> 
> This change explicitly sets the fetchSize to 1 to mimic the same style
of behavior that the default fetchSize uses when using MySQL. If a user
provides a custom fetchSize, that will take precedences.

dwdraju:
FYI, the link to download Mariadb connector plug-in is broken in [v2.5 release page](https://debezium.io/releases/2.5/).

Naros:
> FYI, the link to download Mariadb connector plug-in is broken in [v2.5 release page](https://debezium.io/releases/2.5/).

We'll have to adjust the logic in the UI bits to avoid adding that entry as really technically the MariaDB link should be the MySQL connector but because of how we listed the databases in the test matrix, that happened.

Naros:
@jpechane so assuming that the tests pass now (I confirmed that the failures in the previous run at least passed locally now with my recent changes), this should be ready for a review.  I added a note to the strategy Jira [DBZ-7083](https://issues.redhat.com/browse/DBZ-7083) about the protocol changes in the connection configuration class so we can find a better solution in the later refactoring we intend to do.

Naros:
So it looks like MySQL's handling of `TIMETSTAMP` default values differs from MariaDB.  I'm looking into this, but I'm unsure whether there is a viable way to handle this uniformly across both since removing `timezone=auto` fixes this particular test failure but causes other tests to fail on the actual column values themselves due to timezone handling behavior.

Naros:
After an in-depth discussion with some folks with MariaDB, we've concluded that both drivers are working correctly but reacting to the default value information differently due to how the two drivers manage time zones.

For MySQL, the `connectionTimeZone` actually tells the driver to do value manipulation at the driver level, which means that when the connector executes `SHOW CREATE TABLE`, the data returned, in this case, is simply a string. The time function on the server is unaware of the `connectionTimeZone` setting.  This is why MySQL, if a timestamp has a default value of `1970-01-01 00:00:01`, gets returned unaffected.

MariaDB takes a completely different approach. In their case, their `timezone=auto` setting changes the physical timezone in the session at the server level, meaning that not only does the server return the values to us with using the same timezone like MySQL's `connectionTimeZone` when set to `SERVER`, but it also affects any rendering of timestamps via server-side calls, like the one that generates the output for `SHOW CREATE TABLE`.

So it seems logical that we need to adjust the `MySqlDefaultValueIT` test for MariaDB to account for this, but I'd like your thoughts @jpechane if you maybe have some idea I haven't explored.

jpechane:
@Naros I don't think there is other solution. Let's just modify the test to account for it.

Naros:
Test failures are unrelated.

jpechane:
@Naros Aplied, thanks a lot

PR #4976, opened 2023-11-02
DBZ-7105 Fix exceptions that are triggered by RelationalBaseSourceConnector#validate when calling #validateConnection() with an invalid configuration

rk3rn3r:
DBZ-7105 Fix exceptions that are triggered by RelationalBaseSourceConnector#validate when calling #validateConnection() with an invalid configuration. Other, not-checked fields can cause an exception when the ConnectorConfig class is instantiated to setup the connection. Now we are only allowing connection validation when all config options are free of errors.

closes to https://issues.redhat.com/browse/DBZ-7105


jpechane:
@rk3rn3r Could you please take a look at the Spanner failure?

jpechane (approved):

rk3rn3r:
@jpechane It was unrelated or maybe an intermittent failure.

jpechane:
@rk3rn3r Applied, thanks

PR #4978, opened 2023-11-03
DBZ-7095 The MySqlParser.g4 parser has a priority problem

caicancai:
https://github.com/antlr/grammars-v4/pull/3804


caicancai:
cc @jpechane

caicancai:
cc @jpechane

jpechane:
@caicancai Squashed and applied, thanks!

PR #4979, opened 2023-11-06
DBZ-7109 Upgrade SQLServer JDBC driver to support sensitivity classification

AnatolyPopov:


jpechane:
@AnatolyPopov Applied, thanks

PR #4980, opened 2023-11-06
DBZ-7108 Switched default connection mode for sharded clusters

jcechace:
https://issues.redhat.com/browse/DBZ-7108


jcechace:
There is an unexpected side effect -- the config property was previously also indirectly used to determine the type of the cluster when constructing the internal ReplicaSet object. As a result the tests are now failing because an incorrect RS name is used (for sharded clusters with "sharded" connection mode a synthetic name `cluster` was used). I will solve this by introducing an internal property explicitly noting the type of the cluster.

jcechace:
Internal property wasn't needed in the end. Seems that i initially overcomplicated the creation the internal ReplicaSet object.

jpechane:
@jcechace Applied, thanks

PR #4983, opened 2023-11-07
DBZ-6875 ensuring root user is created before secured Mongo RS is considered started

jcechace:
https://issues.redhat.com/browse/DBZ-6875


jpechane:
@jcechace Applied, hopefully it will go away, thanks!

PR #4985, opened 2023-11-07
DBZ-6762 Add "validate filters" endpoint/s for connector-specific Connect REST Extensions

rk3rn3r:
closes to https://issues.redhat.com/browse/DBZ-6762


debezium-connector-oracle/src/test/resources/rest/data.sql
```diff
@@ -0,0 +1,44 @@
+-- Connect
+--CONN c##dbzuser/dbz@//localhost:1521/ORCLCDB
+CONN c##dbzuser/dbz@//localhost:1521/ORCLPDB1
```
Naros, line 3: Could we avoid hard-coding these credentials in the start-up script?

I'm wondering if it would be less brittle of a test setup if we instead relied on the `@BeforeClass` JUnit annotation and simply got a reference to the `OracleContainer`, opened a connection directly to the database, loaded this SQL file from the file system and executed the statements as a part of our test class.

I'm just considering testing with Oracle FREE vs non-FREE containers and CDB vs non-CDB instances. wdyt?

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
```diff
@@ -278,6 +278,15 @@ public void start() {
...
     public void stop() {
+        if (started) {
+            forceStop();
+        }
+    }
+
+    public void forceStop() {
+        LOGGER.info("[{}] Stopping...", name);
+        MoreStartables.deepStopSync(members.stream());
+        started = false;
```
jcechace, line 289: Can you elaborate on this change please?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
```diff
@@ -41,6 +59,48 @@ public Connector getConnector() {
         return new MongoDbConnector();
     }
 
+    protected MongoDbConnection primary(MongoDbTaskContext context) throws DebeziumException {
```
jcechace, line 62: This is rather misleading name... primary is a term used for the leading node in mongo RS. This method seems to be creating our internal representation of the whole RS

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
```diff
@@ -41,6 +59,48 @@ public Connector getConnector() {
...
+        else {
+            return allDataCollections.stream()
+                    .filter(dataCollectionId -> snapshotAllowedDataCollections.stream()
+                            .anyMatch(s -> s.matcher(dataCollectionId.identifier()).matches()));
+        }
+    }
+
+    @Override
+    public List<DataCollection> getMatchingCollections(Configuration configuration) {
+        MongoDbTaskContext context = new MongoDbTaskContext(configuration);
```
jcechace, line 88: Please avoid naming variables "context". I know it is still the case in test sources, however I've purposefully tried eliminating this in the primary code in order to make it more visible which context the variable holds reference to (we have too many damn contexts).

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
```diff
@@ -41,6 +59,48 @@ public Connector getConnector() {
         return new MongoDbConnector();
     }
 
+    protected MongoDbConnection primary(MongoDbTaskContext context) throws DebeziumException {
+        ReplicaSet replicaSet = new ReplicaSet(context.getConnectionContext().connectionString());
+        return context.getConnectionContext().connect(replicaSet, context.filters(), (s, throwable) -> {
+            throw new DebeziumException(s, throwable);
+        });
```
jcechace, line 66: Note that this likely behaves differently than you would expect in case of sharded cluster and `connection.mode=replica_set` as it creates a connection from the intial connection string -- a different approach than this connection mode utilises later in the connector. 

Nevertheless, I think for the purpose of validation it should be fine. However, in such case I'd avoid using our internal MongoDbConnection and ReplicaSet classes and rather use the MongoDB client for this "seed connection" via `context.getConnectionContext().connect()`

jcechace: `MongoDbConnection` exists mainly for historical purposes -- it used to be much more complex. It's been gradually trimmed down with changes to overall connection management, error handling and so on. Next stage of trimming will be when we deprecate the replica_set connection mode

jcechace: Also note that `context.getConnectionContext().connectionString()` is a bit wonky as well as it returns the connection string as provided by the user which is not necessarily the exact cs used by the connector. That is the case when user doesn't specify the rs name via `replicaSet` parameter. In such case `ReplicaSetDiscovery#getReplicaSets()` emits a warning and tries to  determine the RS name form cluster description. 

Per se this is not a problem as Java's MongoDB driver can deal with this (however other drivers can't). 

The final connection string (or multiple in case of sharded cluster and `connection.mode=replica_set`) is then passed to the connector task via `mongodb.internal.task.connection.strings`. `MongoDbTask` then takes this value and reconstructs the internal `ReplicaSet` instances out of it.

jcechace: Although your approach will work I'm nut sure how problematic is the fact that connection logic of REST extension is different from what the connector does

jcechace:
My overall feeling is that we are creating yet another layer of validation (although I do see the difference compared to what we have -- the UI needs the validation result of our config fields to be exposed).  Maybe we should take a step back and consolidate the validation before we start exposing it to the UI

jcechace:
Another point for mongo specifically. Mongo is dynamic by nature. Collection doesn't have to exists before it's written to (so non existing collection is existentially equal to an empty collection). Though I'm note sure if I understand the code properly in order to know how relevant this is.

debezium-core/src/main/java/io/debezium/config/ValidationResults.java
```diff
@@ -14,11 +14,11 @@
 
 public class ValidationResults {
```
jcechace, line 15: We really should be writing javadocs. Self descriptive code is a myth and in this case I find it a bit hard to get an exact understanting of how the code works / is intended to be used.

jpechane: Self descriptive code is not myth but public classes/interfaces esp. those that represent a contract should have at least class levele JavaDoc.

jpechane:
@rk3rn3r Thanks for the PR. It is a good start. I've one major comment though. I sense a really strong overlap with `*SnapshotChangeEventSource` classes. In both cases you are bilding and filtering the same set of tables (in fact not fully as the snapshot event source can add additional filtering). So maybe in that case should not be the logic merged and maybe moved to `JdbcConnection`? I suppose the same is for MongoDB. In that case (even if it was mentioned by @jcechace above there is potentional for removal) I'd say that `MongoDbConnection` could provide the same service?

debezium-connector-oracle/src/test/resources/rest/data.sql
rk3rn3r: @Naros till me what needs to be different for other containers / instance types, then I can do the required changes.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
rk3rn3r: I am aware of what the term is used for and it is not exclusive for Mongo DB, is it?
I just copied this code from the UI backend (and maybe from legacy Mongo DB connector code?!). I have no idea what you want to change. Feel free to add a commit with better code or setup a call to explain this in detail.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
rk3rn3r: Same as above. It was just copied from the UI backend and I did not write the code there. I think it was maybe also only copied from legacy Mongo DB connector code.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
rk3rn3r: Maybe this detail level is not what we should be concerned about in PRs as long as var names are totally off, no? I got your point and if that method would be longer or having multiple context vars I fully agree and would have named it less generic, sure.

debezium-core/src/main/java/io/debezium/config/ValidationResults.java
rk3rn3r: A very good point on that "helper" interface which is solely used to pass an anonymous method / lambda. That could really require some description here! Will add it.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
rk3rn3r: I found that there is still some cases where `Stopping [rs0] .......`  was output on the console and stop was triggered in tests where no MongoDB instance was running and fixed it in MongoDbShardedCluster. Then I realized I remove the option to force stop and fixed it there in that way so the original code is still callable via `forceStop()`. If you say it is not needed then I can revert this change here and solve it in MongoDbShardedCluster in the same way. I just tried to do it in a less intrusive way here.

rk3rn3r:
@jcechace 

> the UI needs the validation result of our config fields to be exposed

What do you mean with `our validation result`?  The REST extension only calls `Connector#validate()` for the related connector (the REST extension does not have its own validation). Still, it needs to bring the results into a structured format that can be consistently returned to the UI / REST caller. That's what the classes in `io.debezium.rest` package in debezium-core are meant for. Those are the data models used to represent the schema for the UI endpoint. I think it's a good idea to decouple those models from the implementation like TableId. wdyt?
Should that package for models `io.debezium.rest` be renamed to `io.debezium.rest.models` to make it clearer?

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
jcechace: I wouldn't worry about being intrusive in this case -- it's just an internal testing tool... Though I wonder how log output would be getting into tests when there was no mongo running

jcechace: @rk3rn3r  Looking at the comment... seems I was unabele to capture the entire snippet... This is what I found strange

```
public void forceStop() {
        LOGGER.info("[{}] Stopping...", name);
        MoreStartables.deepStopSync(members.stream());
        started = false;
        if (started) {
            LOGGER.info("[{}] Stopping...", name);
            MoreStartables.deepStopSync(members.stream());
            started = false;
        }
    }
```

should the conditioned block be there?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
jcechace: We are working in context of MongoDB so the interpretation is in the context of MongoDB :). 

And yes, I suspected this might be from legacy code -- we used to have this all over the place. The reason was that back then we were indeed connecting directly to RS primary so the name made sense back then. What about naming this "connection" or something simmilar

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
jcechace: I'm sorry, but this shouldn't be used as an argument. The code wasn't frozen a year ago.

Using `context.getConnectionContext().connect()` should be better. However, we can definitively setup a call and explore the options. Feel free to ping me and we can look into it. We should probably discuss whether this validation makes sense for mongo anyway.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
rk3rn3r: Maybe you like to switch to side-by-side diffs eventually? ;)

jcechace:
> @jcechace
> 
> > the UI needs the validation result of our config fields to be exposed
> 
> What do you mean with `our validation result`? The REST extension only calls `Connector#validate()` for the related connector (the REST extension does not have its own validation). Still, it needs to bring the results into a structured format that can be consistently returned to the UI / REST caller. That's what the classes in `io.debezium.rest` package in debezium-core are meant for. Those are the data models used to represent the schema for the UI endpoint. I think it's a good idea to decouple those models from the implementation like TableId. wdyt? Should that package for models `io.debezium.rest` be renamed to `io.debezium.rest.models` to make it clearer?

Sorry for not being clear (had a bit of a hard time wrapping my head around it)... Yes, I more or less meant what you described. So it does make sense!

My point was that maybe we should provide more universal validation "framework"/api and use its bits where needed -- it seems a bit disintegrated at the moment. However that is definitively something which we can revisit in the future and something I might be wrong about as it's mostly based on feelings and impressions.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
rk3rn3r: > Though I wonder how log output would be getting into tests when there was no mongo running

For simplicity `RestExtensionTestInfrastructure#stopContainers()` iterates over all known containers and runs `#stop()` on them. This can be improved for sure.
Normally all Container implementations return directly like here with an early return, when the underlying container is not running. And they just ignore the request to stop.
That way calling `#stop()` was idempotent and could be called mutliple times. As we didn't do this here the output was generated. Furthermore imo in some cases even the request to stop the container would have been triggered by Testcontainers and that means the request timeout to Docker deamon calls applies here. Leading to a slightly longer test execution. But that's just guessed.

rk3rn3r: I will then remove the `#forceStop` and just include the code inside the overriden `#stop()` method. Thx

jcechace: That would remove the indempotence no? Something we should keep.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
rk3rn3r: I re-added enforcing now.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
rk3rn3r: Discussed after triaging call and continued on chat. Should be good now.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
rk3rn3r: No, it keeps idempotency but removes the non-idempotent option to enforce stopping (on a non-running container).

debezium-connector-oracle/src/test/resources/rest/data.sql
Naros: > please tell me what needs to be different for other containers / instance types (aka what needs to be different for non-cdb instances and what are the conditions to realize that we are in non-cdb mode)? Then I can do the required changes. Is changing the user name the only thing that I have to do? The debezium user is available on that Oracle DB test image instances?

I wish it were that simple.

So with Oracle, there are effectively two users (a) connector user and (b) test account user.  We should never use the connector user for anything because this account has to be different depending on whether the database is deployed with or without multi-tenancy.  Oracle mandates that when multi-tenancy is enabled, the user name must be prefixed with `c##` and when multi-tenancy is not it cannot be prefixed with `c##`.  This is just how Oracle works with what they call "common" vs "local" users.

So in short, you should always use the `debezium` user unless you are explicitly configuring the connector to communicate with LogMiner, XStream, or OpenLogReplicator.

The other problem is the SID or database instance you've specified.  In a multi-tenancy scenario, we would connect and work with either `ORCLPDB1` or `FREEPDB1`, depending on whether the instance is based on our images or Oracle's.  If multi-tenancy is not enabled, we'd connect to `ORCLCDB` or `FREE` , again depending on which image is being used.

This is why I recommend that you centralize this in one place by perhaps simply loading the file as a resource, iterating the lines, and executing the statements from within Java using the `OracleContainer`'s JDBC URL to connect with as this would avoid the need to use the `CONNECT` statement here and hardcode credentials.  

> For sure the code should be resilient to using different flavors of Oracle DB and I want to make those changes. But I am not a fan of doing so much in our code as there is already a mechanism in place to load such files into the database. Still, I can imagine we introduce this into the TestHelper eventually and put the SQL code from the *IT classes into the filesystem eventually. Wdyt, should I file a Jira for this?

Sure, there is a mechanism, but which makes more sense?  Duplicate the SQL files for each type of database installation with 99% of the same data but with different connection strings or actually reducing that footprint so that our technical debt and maintenance is within reason ;) 

Overall, I think we need to decide whether we want to even consider testing the REST extension with any multi-configuration setups in the first place, is there any justification for that.  If the answer is no, then picking one and sticking with that is satisfactory.  Then we simply need to evaluate what we've done primarily around the database names to be sure those are easily managed when we decide to move from Oracle 192/1 to Oracle 23 or beyond as our image base.  But if the answer is we should test varied configurations, then we have to be judicious about usernames, database names, and schema names that are used so they're mostly centralized.  TestHelper definitely does quite a bit of this already.

Naros: Unfortunately, nothing with Oracle is ever simple, so if you want to discuss this more, feel free to schedule a time next week, and we can undoubtedly sync-up and talk about this in more detail if you think it's needed.  Anyone else on the team is welcomed to join too if they wanna ask or learn more.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
```diff
@@ -41,6 +55,33 @@ public Connector getConnector() {
         return new MongoDbConnector();
     }
 
+    private ConnectionString enforceReadPreferencePrimary(ConnectionString connectionString) {
+        URI connectionStringUrl = UriBuilder.fromUri(connectionString.getConnectionString())
+                .replaceQueryParam("readPreference", "primary")
```
jcechace, line 60: There are several users who were very explicit(e.g. Okta) about having a control over what nodes debezium interacts with. Selecting collection is not crucial to the point where we need the primary node -- If the rest of the connector itself doesn't need it, UI should not be the component mandating this. 


@jpechane @Naros  what's your take on this?

debezium-core/src/main/java/io/debezium/rest/ConnectionValidationResource.java
```diff
@@ -27,7 +27,7 @@ default ValidationResults validateConnectionProperties(Map<String, ?> properties
         ValidationResults validationResults;
         ClassLoader originalClassLoader = Thread.currentThread().getContextClassLoader();
         Thread.currentThread().setContextClassLoader(getConnector().getClass().getClassLoader());
```
jcechace, line 29: @rk3rn3r  BTW is this still required now when the actual rest resource is in the connector module? I'm just curious.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/rest/DebeziumMySqlConnectorResource.java
```diff
@@ -42,6 +53,40 @@ public Connector getConnector() {
...
 
+    @Override
+    public List<DataCollection> getMatchingCollections(Configuration configuration) {
+        final MySqlConnectorConfig config = new MySqlConnectorConfig(configuration);
+        final MySqlConnection.MySqlConnectionConfiguration connectionConfig = new MySqlConnection.MySqlConnectionConfiguration(configuration);
+        try (MySqlConnection connection = new MySqlConnection(connectionConfig)) {
+            Set<TableId> tables;
+
+            final List<String> databaseNames = new ArrayList<>();
+            connection.query("SHOW DATABASES", rs -> {
```
jcechace, line 64: It might be a good idea to move this into the connector code.

debezium-core/src/main/java/io/debezium/config/ValidationResults.java
```diff
@@ -14,11 +14,11 @@
...
 
-    public final List<ValidationResult> validationResults;
     public Status status;
+    public final List<ValidationResult> validationResults;
 
-    public ValidationResults(Connector connector, Map<String, ?> properties) {
-        this.validationResults = convertConfigToValidationResults(connector.validate(convertPropertiesToStrings(properties)));
+    public ValidationResults(ConnectorParameter connector, Map<String, ?> properties) {
+        this.validationResults = convertConfigToValidationResults(connector.get().validate(convertPropertiesToStrings(properties)));
```
jcechace, line 21: Although this was already in... I think that the result of `connector.get().validate()` should be a parameter for this class' constructor or something. I would honestly not expect a class named `ValidationResult` to internally call the validation, even less that it would be done in constructor.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
rk3rn3r: The primary node in a low consistency (or tunebale consistency) datastore like Mongo DB is the only reliable source for metadata like active tables. This was used explicitly for this reason and especially bigger, busy cluster could return unexpected metadata from my experiences.

debezium-core/src/main/java/io/debezium/rest/ConnectionValidationResource.java
rk3rn3r: The reason for this is that JAX RS resources for Connect REST extensions are loaded in an isolated way with an isolated class loader that has no access to anything else except of the exposed interfaces. I need to "break" this.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/rest/DebeziumMySqlConnectorResource.java
rk3rn3r: True. Could you please file a Jira for this? Thank you

debezium-core/src/main/java/io/debezium/config/ValidationResults.java
rk3rn3r: What else should we call it?

Validation must happen inside this class as we might have to potentially catch exceptions here too and this is shared behavior.

rk3rn3r:
@jcechace 

> My point was that maybe we should provide more universal validation "framework"/api and use its bits where needed 

Maybe we need something like this sooner or later for Debezium Server if we are going to replace usage of Kafka's ConfigDef and the related validation. I would then use whatever comes here eventually. I would personally not do this here. What do other think? Do we need a task for Debezium Server and should we add a comment to this one to reuse it here? Did you had plans for such a thing @jpechane  @Naros ? How should we proceed with that?

debezium-connector-oracle/src/test/resources/rest/data.sql
rk3rn3r: > Sure, there is a mechanism, but which makes more sense? Duplicate the SQL files for each type of database installation with 99% of the same data but with different connection strings

At the moment I am unfortunately not aware of other scenarios (for the REST extension) and their preconditions aka. how to detect them.
So I have only this one use-case in the REST Extension (mind we are in general talking about a REST endpoint that _can_ be used by the UI but in general those endpoints should provide benefit for Kafka Connect + Debezium users.

May I suggest again what I raised in the other PR:
How about introducing a new system property `<oracle.container.image.tag>${version.oracle.server}</oracle.container.image.tag>`? That is used in the maven-docker-plugin / pom.xml to select the Oracle image. In the test suite for the Oracle REST Extension I can then continue using `version.oracle.server` as the only reference to only the version number and we and QE can still customize the image used by changing that new parameter? I must say I was confused and disappointed that a variable where the name seems to only point to a version number of the oracle server is actually used to switch to another image tag.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
jcechace: Rene... I spend literary hundreds of hours working with the mongo code. Including all the comments and Jiras we historically had. I'm pretty sure that I know why things were done one way or the other in the past. I also interact with our users regarding this on daily basis and especially in busy environments they don't like their primary getting touched! 

As for the "freshness" of metadata -- I already acknowledged that you are **technically** correct. However my argument is that for the purpose of UI we don't need that level of "freshness". IMHO the case of somebody going to create a collection and then immediately jumping into Debezium UI and looking for it the for the purpose of setting up a connector is highly unlikely.

debezium-core/src/main/java/io/debezium/config/ValidationResults.java
jcechace: I just just find it unexpected that the validation is performed in an instance of something called `ValidationResult`.... would kinda expect this instance to be create only after the validation happened.

jcechace:
> @jcechace
> 
> > My point was that maybe we should provide more universal validation "framework"/api and use its bits where needed
> 
> Maybe we need something like this sooner or later for Debezium Server if we are going to replace usage of Kafka's ConfigDef and the related validation. I would then use whatever comes here eventually. I would personally not do this here. What do other think? Do we need a task for Debezium Server and should we add a comment to this one to reuse it here? Did you had plans for such a thing @jpechane @Naros ? How should we proceed with that?

Definitively agree. It is not something to be done as part of this PR.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
Naros: Wouldn't it also make sense in this context to rely on the data in the `secondary` if that's what they are connecting with rather than the primary to avoid inconsistencies? I could imagine deploying a connector and pointing to the `secondary` and realizing the collections that are of interest not being visible could be considered a warning or light bulb for the user to check their replication status, no?

In the other case, fetching this detail from the primary and deploying the connector could give the user a false sense of everything being okay. Then, when the connector is running and not capturing changes, this could be seen more as a Debezium problem rather than checking their environment, no?

Given this example, using the secondary in this case makes logical sense, as it can help the user identify potential issues before the connector is even deployed.

Naros:
> Maybe we need something like this sooner or later for Debezium Server if we are going to replace usage of Kafka's ConfigDef and the related validation. I would then use whatever comes here eventually. I would personally not do this here. What do other think? Do we need a task for Debezium Server and should we add a comment to this one to reuse it here? Did you had plans for such a thing @jpechane @Naros ? How should we proceed with that?

This entire validation hook is very rigid, particularly if we want to avoid constructing any connector config object in that method.  For databases like Oracle and MySQL, there are particular details from the connector's configuration beyond just the username, password, host, and port details that are mandatory to create a connection, and trying to strike the right balance here feels extremely limiting with our current class design.

For example, in MySQL, I need to construct a `ConnectorAdapter` to get the correct implementation of the new `AbstractConnectorConnection` class that the validation should use; this requires looking at a plethora of the configuration values and constructing the adapter hierarchy that makes these bits accessible. Creating the adapter also means building other objects the connector uses to deviate behavior between MySQL, MariaDB, and their related drivers.

I'm unsure of the right path forward in this context. Still, we need to be flexible here to allow for adequate design patterns required for such connectors to keep the overall code for connection management clean and avoid multiple code paths that could lead to inconsistencies.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
jcechace: Personally I think we are getting into details which are not really important. In practice the users is IMHO much more likely to be configuring the connector to capture data from collection which does not exist yet (e.g. user wants to capture changes to all "department" collection from inventory database. `collection.include.list=inventory\.department_.*`) rather than facing an issue where the delay between primary and secondary is so large that it is not yet visible in the secondary (a problem which is easily solvable by a presence of "refresh" button next to the collection list).

debezium-connector-oracle/src/test/resources/rest/data.sql
Naros: > How about introducing a new system property <oracle.container.image.tag>${version.oracle.server}</oracle.container.image.tag>? 
We can certainly consider this, I'm certainly not opposed to it.

> Do you have time to do a quick pair-programming-like session today? I really need to understand the differences between FREE, XS, CDB and non-CDB/PDB scenarios their conditions and differences in that Oracle DB.

Sure, just ping me and schedule something, I'm here this week and out all next week on PTO, so sooner may be ideal than waiting.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
rk3rn3r: > Wouldn't it also make sense in this context to rely on the data in the secondary if that's what they are connecting with rather than the primary to avoid inconsistencies? I could imagine deploying a connector and pointing to the secondary and realizing the collections that are of interest not being visible could be considered a warning or light bulb for the user to check their replication status, no?

I see it from the other side. The user always sees the correct collections all the time when using the correct list of collections from the primary and eventually the system (replica node/s) comes into a consistent state and the initially selected collections are streamed from the secondary. So the right set of collections is the one on the primary not the info from the  replica nodes. When using the info from a random node the user might see collections that got removed on the primary but are not yet replicated to the replica node that the user connects to. After a certain amount of time the system will be in the state of the primary, not in the state of one of the replica nodes.
Btw the default ReadPreference mode should be `primary` and that means per default the client should connect to the primary anyways, no?

Furthermore, the UI is not hammering that endpoint. Actually the user needs to configure the filter config options and then press a button to make the request to validate the inputs and get the list of collections returned. I don't think the argument about 
The only attack vector is that someone is hammering this endpoint on purpose but this would be an attack on purpose and can be handled by adding a circuit breaker in front of the API.

debezium-core/src/main/java/io/debezium/config/ValidationResults.java
rk3rn3r: I will rename those classes to `ConnectorConfigValidator`  and `FilterConfigValidator`. I think that matches better the current usage. Initially those classes were only used as a data model for the results. That's the reason for those names. But I agree, they are outdated for the different approach that encapsulates repetitive code that is used for the validation. wdyt?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
jcechace: The default is whatever the standard mongo default is. Which is indeed `primary`. I have no problem with reaching to primary if the user doesn't specify the read preference or explicitly sets it to `primary` :).

rk3rn3r:
@Naros regarding your last comment: Did you maybe miss a detail about the validation? The new code does no new / custom validation it only calls `Connector#validate()`, the rest is entirely left to the related connector class instance.
Your concerns about the connection, I think with the previous agreement to file a Jira to include the filtering code inside the connector implementations will solve the issue mid-term. Short-term I assume a MySqlConnectorConfig/MySqlConnection (and etc for the other connectors) does the required things depending on the provided config and acts more like a builder for concrete Config-/Connection-classes for MariaDB, etc. I think we are talking about a kind of internal API for those *ConnectorConfig/*Connection classes that's defined by the current implementation and could evolve from there?

Naros:
> Did you maybe miss a detail about the validation? The new code does no new / custom validation it only calls Connector#validate(), the rest is entirely left to the related connector class instance.

I was referring more to our discussion I believe in the weekly meeting where we discussed the regex concerns with that the various connector config classes validate and my understanding was you didn't want to construct something like a `MySqlConnectorConfig` in the validate method for that reason; perhaps I misunderstood the conversation?

> Another point, why not instead of bending the current implementation with that ConnectorAdapter, why not creating separate connector classes (and maybe modules) that have shared parent classes (or better compose with a class that provides the shared code)?

I believe that would have been a much larger and disruptive change over the adapter pattern, and I wanted to avoid any potential risks to the MySQL implementation.  I also believe that the adapter pattern provides a much clearer picture for code deviations.  For example, even the new `AbstractConnectorConnection` class I made was way more disruptive than I wanted, but it was unfortunately necessary.  I believe what you're asking would have been more along that scale.  

And if I were to be honest, I rarely see massive refactors end well.  You end up being a bull in a china shop, everything stays broken for extended periods of time and you can't do any type of test-driven design with that methodology.  That may work for others, but in my experience it's not the safe route when you want to preserve compatibility.

debezium-core/src/main/java/io/debezium/config/ValidationResults.java
rk3rn3r: Done

rk3rn3r:
@jpechane @jcechace I applied the changes that we discussed. I hope that addresses most of your concerns.
I will now finish the conditions for the Oracle DB container image types together with @Naros in a quick call now and I hope that we then can merge the PR soon?

debezium-connector-oracle/src/test/resources/rest/data.sql
rk3rn3r: @Naros all my changes are in now. Please have a final look. :pray:

rk3rn3r:
I tested the Oracle REST Extension tests with the `-xs` and the `-noncdb` containers too and all tests were green locally.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
jcechace: @jpechane  we obviously have a disagreement on this topic. What would be your preferred way of handling this?

jcechace: @rk3rn3r I would keep this unresolved until we have Jiri's take on this. If multiple team members think we should do it this way then I'm resting my case ;).

rk3rn3r: I am not going to change the code in a way that will lead to wrong results sooner or later.
From my best current knowledge replica nodes can have the wrong view about the active collections and other cluster metadata and thus might return results that are already outdated or might be outdated eventually.
If you can prove that those metadata are consistent I will surely change this now but I don't see any good reason to change this for the fore-mentioned reasons.
For example your point that this might cause stability issues of the primary node: like I mentioned before, those requests are only happening when the user validates the filter list and even with a very high number of collections I am not convinced this will harm the primary's stability. I am very sorry and although I see your point I just can't do this from my developer's ethical attitude.

What I could imagine is to offer an option to relax this behavior: I can file a Jira task to add a `filter.validation.from.replicas` config option that defaults to `false`. When enabled by setting it to `true` it will change the behavior to use replica nodes. If you don't find this compromise acceptable, feel free to file a Jira for your preferred solution and change it yourself after this PR but I am not putting my name on lines of code that I can't prove to be correct. Honestly: sorry again!

rk3rn3r: Okay, thanks for that comment. I think your points are valid concerns. Such scenario would all require to set `directConnection` option in the connection string, right? That means in the case that `directConnection` is in the connection string I should/must not do "adjustments" to the connection string. But in other cases when there is no `directConnection` parameter set it should be okay? Did I get this right?

jcechace: @rk3rn3r not the one with tags (that's essentially just scoping the set of usable nodes for this connection string). Relaxing to `primaryPreferred` would work for that case. 

Which once again demonstrates my point about hidden complexity and "smart adjustments".

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/rest/DebeziumMySqlConnectorResource.java
jpechane: @rk3rn3r This is vry similar comment to https://github.com/debezium/debezium/pull/4985#issuecomment-1801586596

I am fine with handling this as a separate PR but quite fastr after this one to make sure we avoid any divergence. Could you please create a Jira for it and chedule it for the next release?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
jpechane: I really don't think there should be any adujstment/optimizaion to connection setup in the REST extension. If something is broken it has to be fixed in the connector/core code. If not then REST extension should not introduce any change in default behaviour.

jcechace: @jpechane  just to be clear. This doesn't change the configuration for the connector. It's a temporary change to the connection string just for the purpose of the connection used by the REST extension.

jpechane: Yes, I understand. My take is that we ould use the same connection setup when checking the filters for UI usecase like we do for connector. As soon as we'd allow differences we might end up in strange discrepances that would be hard to debug in future.

rk3rn3r: > As soon as we'd allow differences we might end up in strange discrepances that would be hard to debug in future.

I'm sorry @jpechane , but I can't agree on this one. A cluster will eventually reach the same status as the primary node and not the other way around (the state of a replica node). That means if you are not reading the fresh metadata of the primary node then you will end up facing such kind of discrepancies that you mentioned, for example: why was this topic created (by Debezium) or shown in the UI but no data are coming anymore? Answer: the replica node is not aware of the removal of that collection on the primary node (because it lags behind) but still returns it in the list of collections.
Speaking generally for the connector I would even assume that metadata should be read fresh from the primary even when the replication happens from a replica/secondary node. And it should furthermore ignore for example collections that were removed on the primary node but are still available on that replica node.

jcechace: @rk3rn3r  what metadata would that be in our specific implementation? In a theoretical scenario where collection doesn't yet appear to be present during snapshot the event's will be streamed anyway later. 

Additionally all you are saying is extremely time dependant. I don't think complexity is worth it if the situation would be different, had the user decide to deploy the connector a minute later. Again, this is not the hot path.

jcechace: BTW how do we deal with this in case of relational databases? Things like PG replicas. Oracle Stand-By and so on? I know very little about these but I think the situation might be similar on that front.

Naros: I think we're all caught up in the data technicalities but we don't seem to focus on what I believe are more critical issues: trust, security, and expectations.

If I configure a tool to connect to a replica/secondary server and then it blatantly disregards that configuration and connects to my primary, that can be dangerous for any of those aforementioned reasons.  It's one thing if I don't specify a preference, there aren't any rules pre-established for which we use, and we connect to the primary; however, if we've established those rules, those should be followed -- regardless of the eventual consistency concerns anyone may have.  

If we begin to change the strategy in the UI that doesn't align with the connector's strategy, then you've eroded trust, you've made it even more confusing for the user, especially if they've expressly specified to use the secondary in the connection string, and could lead to a DBA putting "can't deploy to prod" order in place for new tooling because they strictly forbid primary connections -- which is more common than you might think in overly exacerbated environments where they're already working at near max capacity.

There is nothing saying we can't put a small indicator in the UI that alerts the user that we may be connecting to a secondary and that the list of collections/tables shown may not be absolutely accurate due to using the replica rather than the primary -- that solves @rk3rn3r's concerns and it allows the process to be consistent with the connector's behavior without an entirely new set of code around connection management.

rk3rn3r: @jpechane I filed https://issues.redhat.com/browse/DBZ-7165 for a suggestion task to change the behavior of the connector towards that higher consitency model that we recently gave up. imo we were always reading from the primary before, no?

I filed https://issues.redhat.com/browse/DBZ-7166 for aligning the REST extension behavior with whatever we decide as outcome of DBZ-7165 above. In order to align the behavior of the REST endpoint and the connector.

I also filed https://issues.redhat.com/browse/DBZ-7167 to move the filtering logic from the "outside" in the REST extension code into the "inside" of the connector implementations / the connector classes.

jpechane:
@rk3rn3r Applied, thanks

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
rk3rn3r: Done

PR #4986, opened 2023-11-08
DBZ-7108 Minor edit to description of mongodb.connection.mode property

roldanbob:
Suggested terminology edit to property description.


jpechane:
@roldanbob Applied, thanks

PR #4991, opened 2023-11-08
DBZ-6726 support for $changeStreamSplitLargeEvent aggregation in MongoDB connector

jcechace:
https://issues.redhat.com/browse/DBZ-6726


jcechace:
Missing tests, should be available tomorrow

jpechane:
@jcechace LGTM, thanks! I push a commit replacing LinkedList with ArrayList as the iterations is faster for it and we are iterating repeatedly over the message.

jcechace:
@jpechane OK... I guess it's to do with the coherent memory guarantee... I think that array list cant be larger then 4 items in this case, but it's good to follow good practices. 

BTW: I'm gonna also add documentation before we merge.

jpechane:
@jcechace Applied, thanks. Please add docs as a separate PR (you can use the same DBZ-xxx for it).

PR #4993, opened 2023-11-09
DBZ-6726 Documenting split handling mode for oversized change events

jcechace:
Docs for https://issues.redhat.com/browse/DBZ-6726


jpechane:
@jcechace Thanks

PR #4995, opened 2023-11-09
DBZ-6317 Document tombstone behavior

mfvitale:
Related to https://github.com/debezium/debezium-connector-jdbc/pull/43

### DIFF
--- a/debezium-core/src/main/java/io/debezium/transforms/extractnewstate/DefaultDeleteHandlingStrategy.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/extractnewstate/DefaultDeleteHandlingStrategy.java
@@ -48,6 +48,10 @@ public class DefaultDeleteHandlingStrategy<R extends ConnectRecord<R>> extends A
                 LOGGER.trace("Delete message {} requested to be dropped", record.key());
                 return null;
             case TOMBSTONE:
+                // NOTE
+                // Debezium TOMBSTONE has both value and valueSchema to null, instead here we are generating
+                // a record only with null value that by JDBC connector is treated as a flattened delete.
+                // Any change to this behavior can have impact on JDBC connector.
                 return afterDelegate.apply(record);
             case REWRITE:
             case REWRITE_WITH_TOMBSTONE:
--- a/debezium-core/src/main/java/io/debezium/transforms/extractnewstate/LegacyDeleteHandlingStrategy.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/extractnewstate/LegacyDeleteHandlingStrategy.java
@@ -46,6 +46,10 @@ public class LegacyDeleteHandlingStrategy<R extends ConnectRecord<R>> extends Ab
                 LOGGER.trace("Delete message {} requested to be dropped", record.key());
                 return null;
             case NONE:
+                // NOTE
+                // Debezium TOMBSTONE has both value and valueSchema to null, instead here we are generating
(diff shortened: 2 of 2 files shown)


debezium-core/src/main/java/io/debezium/transforms/extractnewstate/DefaultDeleteHandlingStrategy.java
```diff
@@ -48,6 +48,10 @@ public R handleDeleteRecord(R record) {
                 LOGGER.trace("Delete message {} requested to be dropped", record.key());
                 return null;
             case TOMBSTONE:
+                // NOTE
+                // Debezium TOMBSTONE has both value and valueSchema to null, instead here we are generating
```
rk3rn3r, line 52: I don't really get what is referred to as "Debezium TOMBSTONE" here and how it differs. Maybe we can word this in a better way.

Furthermore, the term "tombstone" is defined and used in Kafka. It refers to a Kafka message that has data the message key (e.g. the primary key of the field/s that got deleted) and optionally in the header too, and a message value which is null. Maybe we can point to that definition (although it is hard to find in the official Kafka docs).

Those tombstones are used for compaction purposes and their semantically meaning is a delete for that key (from the dataset relate to that message key). The JDBC Sink just re-uses those semantics.

Maybe we should also place a note about a tombstone related parameter in the docs for the Kafka topic config: [delete.retention.ms](https://kafka.apache.org/documentation/#topicconfigs_delete.retention.ms)
The default value is 1 day and that means that in cleaned log segments (segment files) all tombstones older than 1 day are being removed. That means that consumers / JDBC sinks that are off (crashed) for longer than 24 hrs might miss those deletes. I recommended users that use those tombstone semantics to tune the topic config to use `delete.retention.ms=2678400000` (31 days) or similar as a value so that it is okay to loose a sink for longer than 1 day without loosing those deletes. This also applies to topic auto-creation (see my blog post from back then).

debezium-core/src/main/java/io/debezium/transforms/extractnewstate/LegacyDeleteHandlingStrategy.java
```diff
@@ -46,6 +46,10 @@ public R handleDeleteRecord(R record) {
                 LOGGER.trace("Delete message {} requested to be dropped", record.key());
                 return null;
             case NONE:
+                // NOTE
+                // Debezium TOMBSTONE has both value and valueSchema to null, instead here we are generating
```
rk3rn3r, line 50: My comment from above applies here as well.

debezium-core/src/main/java/io/debezium/transforms/extractnewstate/DefaultDeleteHandlingStrategy.java
jpechane: @rk3rn3r Please read the discussion fully in https://github.com/debezium/debezium-connector-jdbc/pull/43#discussion_r1387749959. It is mainly about whether the tombstone has a value schema associated with or not. We are now kind of "abusing" the situtation that tombstone this SMT generates has schema while the regular tombstone useually does not. It creates a hidden dependency. This comment reflects that dependency till we will find a final solution to the problem - whcih at the end might require explicit config option on the sink side to ignore or process the tombstones while we now rely on the heuristics of having schema present.

jpechane:
@mfvitale Applied, thanks

PR #4996, opened 2023-11-09
DBZ-6942 Verify no log sequence gaps on log switch

Naros:
https://issues.redhat.com/browse/DBZ-6942

There is a corner case where it's possible the Oracle connector may query the Oracle metadata tables quicker than the ARC process can generate an archive log history record in V$ARCHIVED_LOG, and this can lead to a race condition where we may incorrectly advance the connector forward to start mining a group of logs when a log sequence gap exists in the log ranges.

For users who use the online_catalog strategy, there are some checks that LogMiner does automatically which it skips, and one is with log sequence gaps. This fix enforces that check by Debezium even for users who may use the faster online_catalog mode so that no logs are omitted and events could be missed.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerHelper.java
@@ -23,6 +23,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.debezium.DebeziumException;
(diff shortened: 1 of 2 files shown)


Naros:
If we can, @jpechane, we could even incorporate this into 2.4.x, that'd be great if you think the fix is minor enough to be backported.

Naros:
Oracle 21 - Test `shouldHandleParentChildIndexOrganizedTables` is unrelated, see [DBZ-7072](https://issues.redhat.com/browse/DBZ-7072).

jpechane:
@Naros Applied, thanks!

PR #4999, opened 2023-11-10
DBZ-6317 Document `batch.size` property for JDBC Sink Connector

mfvitale:


jpechane:
@mfvitale Applied, thanks

PR #5002, opened 2023-11-11
[docs] Docs for `ConvertCloudEventToSaveableForm`

rkudryashov:
better to merge after https://github.com/debezium/debezium-connector-jdbc/pull/46


jpechane:
I think you need to modify `documentation/modules/ROOT/nav.adoc`, and `documentation/modules/ROOT/pages/transformations/index.adoc` too.

jpechane:
@rkudryashov Applied, thanks

PR #5003, opened 2023-11-13
[docs] Mention SQL server instance name is ignored when port is speci…

vjuranek:
…fied


jpechane:
@vjuranek Applied, thanks

PR #5005, opened 2023-11-13
DBZ-7099 Provide default value for PeriodicCommitOffsetPolicy

vjuranek:
In 7b4cf1901 deprecated `io.debezium.embedded.spi.OffsetCommitPolicy`, which provided also constructor for `Configuration`, was removed. This constructor was actually used for creating `OffsetCommitPolicy`. `Configuration` provides default values for options which are not explicitly set, while `Properties` based constructor cannot do that and therefore with this switch the default value for `offset.flush.interval.ms` is now missing.

As debezium-api package has no knowledge about `Configuration` interface, which is part of debezium-core, and thus about the default values, specify default value direcly in the `PeriodicCommitOffsetPolicy` class.

https://issues.redhat.com/browse/DBZ-7099


jpechane:
@vjuranek Applied, thanks

PR #5008, opened 2023-11-14
DBZ-7083 DBZ-1482 - MySQL/MariaDB Strategy & MariaDB GTID support

Naros:
https://issues.redhat.com/browse/DBZ-7083


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/strategy/AbstractConnectionConfiguration.java
```diff
@@ -0,0 +1,167 @@
...
+import io.debezium.jdbc.JdbcConfiguration;
+import io.debezium.jdbc.JdbcConnection;
+import io.debezium.util.Strings;
+
+/**
+ * @author Chris Cranford
+ */
+public abstract class AbstractConnectionConfiguration implements ConnectionConfiguration {
+
+    public static final String URL_PATTERN = "${protocol}://${hostname}:${port}/?useInformationSchema=true&nullCatalogMeansCurrent=false&useUnicode=true&characterEncoding=UTF-8&characterSetResults=UTF-8&zeroDateTimeBehavior=CONVERT_TO_NULL&connectTimeout=${connectTimeout}";
```
jpechane, line 26: Are really all the the connection parameters relevant for both MySQL and MariaDB? If not then we should IMHO spearate connections strings at this stage too.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/strategy/mariadb/MariaDbGtidSet.java
```diff
@@ -0,0 +1,55 @@
...
+ * @author Chris Cranford
+ */
+public class MariaDbGtidSet implements GtidSet {
+
+    public MariaDbGtidSet(String gtid) {
+    }
+
+    @Override
+    public boolean isEmpty() {
+        throw new DebeziumException("NYI");
```
jpechane, line 23: If these exceptions are going to stay for a while then I'd recommend to use `UnsupportedException("Not yet implemented")` for the reader's advanatge. If it is only placeholder for the sake of this PR then it can be kept as is now.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
```diff
@@ -507,6 +511,92 @@ public static SecureConnectionMode parse(String value, String defaultValue) {
...
+            protected ConnectorAdapter getAdapter(MySqlConnectorConfig connectorConfig) {
+                LOGGER.info("Using " + MariaDbConnectorAdapter.class.getName());
+                return new MariaDbConnectorAdapter(connectorConfig);
+            }
+        },
+
+        /**
+         * Expects the target database to be MariaDB but uses the MySQL driver.
+         */
+        MARIADB_HYBRID("mariadb-hybrid") {
```
jpechane, line 544: Is there any advanatage in this mode? I thought it to be a transient one till MariaDB db + MariaDB driver is in place which already is.

jpechane:
@Naros Great and clean job!

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/strategy/mariadb/MariaDbGtidSet.java
Naros: No, these are only for my benefit. These will all be removed once I finalize the implementation.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
Naros: I'm on the fence about it.  

I imagine someone upgrading from an older version who may want to use the legacy behavior, and that's partly why I left it.  If anything, I think it would be something we would leave in for one minor release and remove to give people a chance to adapt with a buffer.  wdyt?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/strategy/AbstractConnectionConfiguration.java
Naros: Likely not, I'll leave this as a todo to clean-up.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
jpechane: Sounds reasonable. Maybe we should add to the documentation that this is not preferred mode of opertion or something like that.

Naros:
> @Naros Great and clean job!

Thanks, I'm not really pleased with the GTID bits tbh; hopefully I can find a nice refactor of that before the end of the quarter but first I'm focused primarily on passing tests.

Naros:
**NOTE TO SELF**
Make sure to tag MariaDB GTID support commits as DBZ-1482 rather than DBZ-7083.  They're being included here because the refactor made the support easier. 

@jpechane if you want any of this split, just let me know if you think it's necessary.

PR #5009, opened 2023-11-14
DBZ-7093 Check schema length to find missed DDL by SQL_BIN_LOG=OFF

nicholas-fwang:
https://issues.redhat.com/browse/DBZ-7093


jpechane:
@nicholas-fwang Applied, thanks!

PR #5011, opened 2023-11-15
DBZ-7141 Remove ComputePartition SMT related code

mfvitale:
`ComputePartition` SMT  has been deprecated in 2.2 release and replaced by `PartitionRouting`.


jpechane:
@rk3rn3r Hi, there is a failur in REST extension test. Genrally it seems that the test will get broken anytime the lis of SMTs is updated as it is not reflected yet in the published image. Is there a way how to handle such scenario in a more robust way?

jpechane:
@mfvitale Applied, thanks!

PR #5012, opened 2023-11-15
DBZ-7132 Document MySQL database.ssl.* options

vjuranek:
https://issues.redhat.com/browse/DBZ-7132


jpechane:
@vjuranek Applied, thanks!

PR #5013, opened 2023-11-15
DBZ-7142 Improved Outbox Event Router SMT to handle whitespaces inside of fields.additional.placement setting

rgibaiev:
Guys, I do understand that it is not a major improvement to the codebase, but believe me - it will significantly improve life of many people (those who use different UIs to manage connectors).
Two minor points here:
- technically there is no sense in adding ".trim()" to a "AdditionalFieldPlacement placement = AdditionalFieldPlacement.parse(parts[1]);" line (line 332) as AdditionalFieldPlacement.parse() function does call trim() under the hood. But I still decided to add it to keep the consistent for all 3 parts and not to confuse any of its future readers
- there already exists a test for this functionality, so I just slightly modified it


jpechane:
@rgibaiev It is a good QoL improvement, thanks!

PR #5015, opened 2023-11-16
DBZ-7146 Configurable inactivity pause duration in MongoDB connector

jcechace:
https://issues.redhat.com/browse/DBZ-7146

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
@@ -5,7 +5,6 @@
  */
 package io.debezium.connector.mongodb;
 
-import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
@@ -37,7 +36,7 @@ import io.debezium.pipeline.EventDispatcher;
 import io.debezium.pipeline.source.spi.StreamingChangeEventSource;
 import io.debezium.pipeline.txmetadata.TransactionContext;
 import io.debezium.util.Clock;
-import io.debezium.util.Metronome;
+import io.debezium.util.DelayStrategy;
 import io.debezium.util.Threads;
 
 /**
@@ -46,6 +45,8 @@ import io.debezium.util.Threads;
 public class MongoDbStreamingChangeEventSource implements StreamingChangeEventSource<MongoDbPartition, MongoDbOffsetContext> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(MongoDbStreamingChangeEventSource.class);
+
+    private static final int THROTTLE_NO_MESSAGE_BEFORE_PAUSE = 5;
 
     private final MongoDbConnectorConfig connectorConfig;
     private final EventDispatcher<MongoDbPartition, CollectionId> dispatcher;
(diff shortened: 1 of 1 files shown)


obabec (approved):

jpechane:
@jcechace I'd recommend to take a look at https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresStreamingChangeEventSource.java#L56 too

jcechace:
@jpechane good point. Will rephrase the Jira title and employ the same approach.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
```diff
@@ -913,6 +914,15 @@ public static OversizeHandlingMode parse(String value, String defaultValue) {
                     + "This includes both, the size before and after database operation, "
                     + "more specifically this limits the size of fullDocument and fullDocumentBeforeChange filed of MongoDB change events.");
 
+    public static final Field CURSOR_INACTIVITY_PAUSE_MS = Field.create("cursor.inactivity.pause.ms")
```
jpechane, line 917: Do need the new config option or can we re-use `io.debezium.config.CommonConnectorConfig.getPollInterval()` is is the case with PostgreSQL?

jcechace: Probably not and we could reuse. I was just confused since this option seems to be used as an await time when the queue threshold is reached.  However the documentation for it suggests that it should be used for the use case we have here :). 

So if this overloaded use case is OK I will change things.

jcechace:
> @jcechace I'd recommend to take a look at https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresStreamingChangeEventSource.java#L56 too

Looking at this closely. Do we want to change this? Both conenctors seems to utilise a bit different approach 

Postres uses`DelayStrategy#constant` to wait a `POLL_INTERVAL_MS` ms if no new message was received `THROTTLE_NO_MESSAGE_BEFORE_PAUSE` number of times. 


Mongo is using `Metronome.sleeper` to wait if no new message was received in the last `CURSOR_INACTIVITY_PAUSE_MS` ms (can be unified to also use POLL_INTERVAL_MS).

Dependning on the polling speed I actually think that the Postgres connector might wait more often. We could combine the two and have Mongo connector wait only if no new message was received `THROTTLE_NO_MESSAGE_BEFORE_PAUSE` number of times AND it's been at least `CURSOR_INACTIVITY_PAUSE_MS` since the last message arrived... However that seems as a bit premature optimisation which removes predictability and complicates tuning.

jpechane:
The problem is that the perf testing shown that there is the similar behaviour as in case of Postgres. That the data might be present but are not served by the client library so IMHO it is not a premature optimization.

jcechace:
> The problem is that the perf testing shown that there is the similar behaviour as in case of Postgres. That the data might be present but are not served by the client library so IMHO it is not a premature optimization.

I don't think that's the case.. I think it means that there really is no data...  Lets leave if for the call. I also think that my description of the metronom behaviour was not correct (or maybe that was the intended behaviour but that the sleeper metronome doesn't in fact work that way)

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -248,7 +253,10 @@ else if (rsOffsetContext.lastTimestamp() != null) {
                     }
 
                     try {
-                        pause.pause();
+                        if (noMessageIterations >= THROTTLE_NO_MESSAGE_BEFORE_PAUSE) {
```
jcechace, line 256: @jpechane  shouldn't we then emit heart beat event only when this threshold is reached?

jpechane: @jcechace It is IMHO unnecessary optimization.

PR #5018, opened 2023-11-17
DBZ-7153 in memory cache to speed up infinispan

sherpa003:
Already discussed with @Naros.
I tested it on our environment, including the cache reload from ISPN data during restart


Naros:
There are two test failures attributed to this change:
```
[ERROR]   OracleConnectorIT.shouldNotEmitEventsOnConstraintViolations:2272 
Expecting actual not to be null
[ERROR]   OracleConnectorIT.shouldUndoOnlyLastEventWithSavepoint:4668 
Expecting actual not to be null
```

Naros (approved):

Naros:
Applied to main (=2.5), thanks @sherpa003

PR #5019, opened 2023-11-18
DBZ-7157: Key serialisation for events from sharded collections with composite sharding key

jcechace:
https://issues.redhat.com/browse/DBZ-7157


jpechane:
@jcechace So now let's put aside incremental snapshot. What will be the key of the message in case of sharded collection? Will it contain only `_id` or the sharding key too? If the latter then I don't think it is correct.

jcechace:
Don't ask me why `ChangeStreamDocument#keyDocument` contains the sharding key too. I have no idea why that would be useful :) ... maybe to indicate that the document moved between shards or something.

jpechane:
@jcechace Exactly and does the PR now behaves in that way? Also (and that's topic for another PR) maybe the sharding key should end up in Connect headers, WDYT?

jpechane:
@jcechace Applied, thanks

PR #5020, opened 2023-11-18
DBZ-7158 Treat Oracle RAC redo thread log sequences independently

Naros:
https://issues.redhat.com/browse/DBZ-7158

@jpechane This fixes the discussion had on https://issues.redhat.com/browse/DBZ-6942 with Lucas.  This should also be back ported to 2.4.2.Final as 2.4.1.Final will likely not work on Oracle RAC but should be fine for Oracle Standalone.


jpechane:
@Naros Applied, thanks

PR #5021, opened 2023-11-19
DBZ-7159 Fail fast during deserialization if a value is not a CloudEvent

rkudryashov:
https://issues.redhat.com/browse/DBZ-7159

A new validator will also be used by [`ConvertCloudEventToSaveableForm`](https://github.com/debezium/debezium-connector-jdbc/blob/main/src/main/java/io/debezium/connector/jdbc/transforms/ConvertCloudEventToSaveableForm.java#L156)

### DIFF
--- a/debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/CloudEventsConverterIT.java
+++ b/debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/CloudEventsConverterIT.java
@@ -193,6 +193,48 @@ public class CloudEventsConverterIT extends AbstractMongoConnectorIT {
         insertHeader.close();
     }
 
+    @Test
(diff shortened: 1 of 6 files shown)


jpechane:
@rkudryashov Should `baseCheck` and `checkFields` throw an exception in case of not fmet preconditions?

jpechane:
@rkudryashov Yes, I understand but the aforementioned methods can throw an exception which means that `isCloudEvent` can throw an exception too. Is it desirable for your use case?

rkudryashov:
@jpechane the methods throw an exception only if they are called with
serializer type other than json or avro. I think that's ok

rkudryashov:
@jpechane changed to return false

rkudryashov:
Eventually I think an exception should be thrown in case of a serializer type other than JSON or Avro because the validator doesn't know how to check such records

PR #5022, opened 2023-11-20
DBZ-7143 Error when fail converting value with internal schema

nicholas-fwang:
https://issues.redhat.com/browse/DBZ-7143


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlValueConverters.java
```diff
@@ -319,6 +319,14 @@ public ValueConverter converter(Column column, Field fieldDefn) {
         return super.converter(column, fieldDefn);
     }
 
+    @Override
+    protected Object handleUnknownData(Column column, Field fieldDefn, Object data) {
```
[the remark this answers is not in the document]
jpechane: @nicholas-fwang Looking at it I believe the best solution is to modify `io.debezium.jdbc.JdbcValueConverters.handleUnknownData(Column, Field, Object)` instead of adding the method to MySQL only.  All handling should be ideally done in `createValueGenerator`.
It is just necessary to think about how to modify the logic to keep it backward compatible for the default setting.

debezium-core/src/main/java/io/debezium/relational/TableSchemaBuilder.java
```diff
@@ -281,17 +290,24 @@ protected StructGenerator createValueGenerator(Schema schema, TableId tableId, L
...
-                                    "Failed to properly convert data value for '{}.{}' of type {}", tableId,
-                                    col.name(), col.typeName(), e);
-                        }
                         catch (final Exception e) {
+                            // TODO(nicholas): 여기서 에러가 핸들링되면서 에러 처리되지 않는다.
                             Column col = columns.get(i);
                             Loggings.logErrorAndTraceRecord(LOGGER, row,
                                     "Failed to properly convert data value for '{}.{}' of type {}", tableId,
                                     col.name(), col.typeName(), e);
+                            if (eventConvertingFailureHandlingMode == FAIL) {
```
jpechane, line 299: `switch` would be better

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlDatabaseSchema.java
```diff
@@ -87,7 +87,8 @@ public MySqlDatabaseSchema(MySqlConnectorConfig connectorConfig, MySqlValueConve
                         connectorConfig.customConverterRegistry(),
                         connectorConfig.getSourceInfoStructMaker().schema(),
                         connectorConfig.getFieldNamer(),
-                        false),
+                        false,
+                        connectorConfig.getEventConvertingFailureHandlingMode()),
```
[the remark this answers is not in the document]
nicholas-fwang: @jpechane it seems that `TableSchemaBuilder` class doesn't have `connectorConfig`.
if this field is not passed in constructor, we have to pass it in `createValueGenerator` and `create` method.
I think it would be better that passed by constructor.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlValueConverters.java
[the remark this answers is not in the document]
nicholas-fwang: @jpechane okay, I will make throw exception in `io.debezium.jdbc.JdbcValueConverters.handleUnknownData`.
and for backward compatible, the default setting should be `SKIP` or `IGNORE`, not `FAIL`. I'll also update it.

debezium-core/src/main/java/io/debezium/relational/TableSchemaBuilder.java
```diff
@@ -75,9 +77,11 @@ public TableSchemaBuilder(ValueConverterProvider valueConverterProvider,
                               CustomConverterRegistry customConverterRegistry,
                               Schema sourceInfoSchema,
                               FieldNamer<Column> fieldNamer,
-                              boolean multiPartitionMode) {
+                              boolean multiPartitionMode,
+                              EventConvertingFailureHandlingMode eventConvertingFailureHandlingMode) {
         this(valueConverterProvider, null, schemaNameAdjuster,
-                customConverterRegistry, sourceInfoSchema, fieldNamer, multiPartitionMode);
+                customConverterRegistry, sourceInfoSchema, fieldNamer, multiPartitionMode,
+                eventConvertingFailureHandlingMode);
```
jpechane, line 84: @nicholas-fwang Let's do it using slightly different way. There should be a new constructor that will accept `EventConvertingFailureHandlingMode`. The existing constructor will just use the value that conforms to the default behaviour.
Then only `MySqlDatabaseSchema` and MySQL conncetor in generl will be affected and we can introduce the config behaviour into other connectors if/when needed. WDYT?

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MysqlDefaultValueIT.java
```diff
@@ -729,7 +729,7 @@ public void dateAndTimeTest() throws InterruptedException {
     public void timeTypeWithConnectMode() throws Exception {
         config = DATABASE.defaultConfig()
                 .with(MySqlConnectorConfig.SNAPSHOT_MODE, MySqlConnectorConfig.SnapshotMode.INITIAL)
-                .with(MySqlConnectorConfig.TABLE_INCLUDE_LIST, DATABASE.qualifiedTableName("DATE_TIME_TABLE"))
+                .with(MySqlConnectorConfig.TABLE_INCLUDE_LIST, DATABASE.qualifiedTableName("DATE_TIME_TABLE_CONNECT_MODE"))
```
nicholas-fwang, line 732: @jpechane 
conversion of value of `TIME` type is failed when connect mode if out of range from 00:00:00 to 24:00:00.
in `DATE_TIME_TABLE` table, record consumption is failed because default value is out of range.
previously, this value is replaced to null by `handleUnknownData` in JdbcValueConverters with warning log.
but, in `handleUnknownData` of MysqlValueConverters, it's always throw exception. and the exception is handled by `createValueGenerator` in TableSchemaBuilder according to `event.converting.failure.handling.mode`.
so, I have added `DATE_TIME_TABLE_CONNECT_MODE` table without out of range default value in this test case.

jpechane: I am afraid we cannot do that and we need to find a way how to get back to the previous behaviour.
The reasons is that as stange it might sound MySQL database can emit invalid date. The old behvaiour would mean that Debezium would skip over such data but now the connector will stop processing without a way how to restart it.

jpechane: If it would mean that we'd need to duplicate the code then it is the price we need to pay :-(.

nicholas-fwang: @jpechane I have tested more and I missed default value converting failure [here](https://github.com/debezium/debezium/blob/main/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlDefaultValueConverter.java#L111).
I will consider where we handle default value converting failure. :)

nicholas-fwang: If we insert out of range time value to mysql with valid default value, consumption error doesn't occur and just remain warning log because `event.converting.failure.handling.mode` is skip(default).

nicholas-fwang: @jpechane it looks complicated :(
in TableSchemaBuilder, `converter.convert(value)` is used in key and value generator and default value parser..
so, I think it's better to handle `event.converting.failure.handling.mode` in `ValueConverter` [here](https://github.com/debezium/debezium/pull/5022/commits/4d80b5210ebe04b95432c12b9f7bc8520c78cca8#diff-0fa5a64cda387e02edbc4db74f552610f349f94f1eed7a05d04e9c184e56b385R80-R102).
and MysqlValueConverters use the method in [converters](https://github.com/debezium/debezium/pull/5022/commits/4d80b5210ebe04b95432c12b9f7bc8520c78cca8#diff-23dc3f586e92bd33654f4014313421f5b93d94044e92b4d1b3337594ed3e485fR244).

then, we can use the mode only in mysql connector and this handling mode can be used in all mysql converters.
I removed `parsingErrorHandler` in MysqlValueConverters because it looks similar with the mode.

it requires many changes about converters.. I'm not sure if this is an appropriate change.

nicholas-fwang: @jpechane Add the mode handler when parse default value [here](https://github.com/debezium/debezium/pull/5022/files#diff-3526f88652b3ce58c9afb44542a2faf244ea01c0aa2eeb661f22c941d6e964a4R472)

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlValueConverters.java
```diff
@@ -386,8 +400,13 @@ protected Object convertJson(Column column, Field fieldDefn, Object data) {
                         r.deliver(JsonBinary.parseAsString((byte[]) data));
                     }
                     catch (IOException e) {
-                        parsingErrorHandler.error("Failed to parse and read a JSON value on '" + column + "' value " + Arrays.toString((byte[]) data), e);
-                        r.deliver(column.isOptional() ? null : "{}");
+                        if (eventConvertingFailureHandlingMode == FAIL) {
```
nicholas-fwang, line 403: @jpechane 
I removed `parsingErrorHandler` because it looks same with `eventConvertingFailureHandlingMode`.
but to use the mode, MysqlValueConverters needs the value of mode in constructor.

jpechane:
@nicholas-fwang I am now a bit lost. Is there any information you are asking me for or is the PR more or less done (it seems to me so)? Thanks a lot

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlValueConverters.java
```diff
@@ -134,8 +137,8 @@ else if (70 <= year && year <= 99) {
      */
     @VisibleForTesting
     public MySqlValueConverters(DecimalMode decimalMode, TemporalPrecisionMode temporalPrecisionMode, BigIntUnsignedMode bigIntUnsignedMode,
```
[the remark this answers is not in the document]
nicholas-fwang: @jpechane added a constructor, but still tests are changed because I removed `parsingErrorHandler`.

nicholas-fwang:
> @nicholas-fwang I am now a bit lost. Is there any information you are asking me for or is the PR more or less done (it seems to me so)? Thanks a lot

@jpechane yes, I'm done, please review PR focusing following points :)
there are 2 points updated different with initial.
- `event.converting.failure.handling.mode` is applied when convert the default value.
- `parsingErrorHandler` in MysqlValueConverters is replaced by value of `event.converting.failure.handling.mode`.

jpechane (commented):
@nicholas-fwang I think it is almost done. I left few comments related to additional code cleanup an additional assertions. other than that it is done. Good job!

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlValueConverters.java
```diff
@@ -386,8 +422,13 @@ protected Object convertJson(Column column, Field fieldDefn, Object data) {
                         r.deliver(JsonBinary.parseAsString((byte[]) data));
                     }
                     catch (IOException e) {
-                        parsingErrorHandler.error("Failed to parse and read a JSON value on '" + column + "' value " + Arrays.toString((byte[]) data), e);
-                        r.deliver(column.isOptional() ? null : "{}");
+                        if (eventConvertingFailureHandlingMode == FAIL) {
+                            throw new DebeziumException("Failed to parse and read a JSON value on '" + column + "' value " + Arrays.toString((byte[]) data), e);
+                        }
+                        else {
+                            logger.warn("Failed to parse and read a JSON value on '{}' value '{}'", column, Arrays.toString((byte[]) data));
```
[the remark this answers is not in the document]
nicholas-fwang: updated using `Loggings`

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlValueConverters.java
```diff
@@ -386,8 +422,13 @@ protected Object convertJson(Column column, Field fieldDefn, Object data) {
                         r.deliver(JsonBinary.parseAsString((byte[]) data));
                     }
                     catch (IOException e) {
-                        parsingErrorHandler.error("Failed to parse and read a JSON value on '" + column + "' value " + Arrays.toString((byte[]) data), e);
-                        r.deliver(column.isOptional() ? null : "{}");
+                        if (eventConvertingFailureHandlingMode == FAIL) {
+                            throw new DebeziumException("Failed to parse and read a JSON value on '" + column + "' value " + Arrays.toString((byte[]) data), e);
+                        }
+                        else {
```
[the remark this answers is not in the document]
nicholas-fwang: updated!

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlValueConverters.java
```diff
@@ -117,8 +121,8 @@ else if (70 <= year && year <= 99) {
         return temporal;
     }
 
-    private final ParsingErrorHandler parsingErrorHandler;
```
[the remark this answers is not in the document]
nicholas-fwang: removed!
https://github.com/debezium/debezium/pull/5022/commits/cb090ebd447c5ae8ad2a5a69ee09a0b6efb3aab9#diff-23dc3f586e92bd33654f4014313421f5b93d94044e92b4d1b3337594ed3e485fL78-L82

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlValueConvertersTest.java
```diff
@@ -130,25 +127,21 @@ public void testSkipInvalidJsonValues() {
         Column colA = table.columnWithName("A");
         Field fieldA = new Field(colA.name(), -1, converters.schemaBuilder(colA).optional().build());
         assertThat(converters.converter(colA, fieldA).convert(INVALID_JSON)).isEqualTo(null);
-        assertThat(errorCount.get()).isEqualTo(1);
```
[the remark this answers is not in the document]
nicholas-fwang: @jpechane Add assertions!
https://github.com/debezium/debezium/pull/5022/commits/cb090ebd447c5ae8ad2a5a69ee09a0b6efb3aab9#diff-aef62fc926d19929b497ab602be2f822525218b4fe99062dd69efb5f55b1501bR132-R142

jpechane:
@nicholas-fwang Could you please take a look at test failures? They seems to be related.

nicholas-fwang:
@jpechane I have fixed mysql TC for gtid (replica) case.
Should I fix also TC for Db2 and SQL server? I'm not sure that it's also related.

nicholas-fwang:
@jpechane could you confirm this?

jpechane:
@nicholas-fwang Restarted Db2 job - IMHO the failures are unrelated and are known flaky tests. I'll merge the PR when the run is finished.

jpechane:
The PR was rebased to the latest (2.6) main

jpechane:
@nicholas-fwang Applied, thanks! Could you please prepare an addtional PR with MySQL docs updated - the new config option?

Thanks!

jpechane:
@nicholas-fwang I am sorry, yes you are right. Nothing more is needed.

PR #5023, opened 2023-11-20
DBZ-7139 MongoDB data collection filter requires replica set specification on blocking/initial snapshot execution

mfvitale:
Use namespace instead of identifierto match collections to be snapshotted


jcechace (approved):
LGTM 

@jpechane this removes the possibility to snapshot collection on a specific shard in a sharded deployment. However

1) We are removing the `replica_set` mode in 2.6 anyway
2) I don't really see a usecase for doing it anyway. Maybe if one had some complex range sharding key -- where e.g. documents for a certain department would be directed to a specific shard. However for such cases we have other means of filtering the snapshot data IIRC.

jpechane:
@mfvitale Applied, thanks!
@jcechace Yes, let's remove it as `replica_set` is removed in the next release. I'll just highlight the change in the release notes.

PR #5027, opened 2023-11-23
Remove MySQL 5.x from TF matrix

obabec:


obabec:
\packit test

jpechane:
@obabec What should be done here?

PR #5028, opened 2023-11-23
DBZ-7179 Fix DebeziumMySqlConnectorResource not using the new MySQL adatper structure to support different MySQL flavors

rk3rn3r:
closes to https://issues.redhat.com/browse/DBZ-7179


jpechane:
@rk3rn3r Applied, thanks

PR #5029, opened 2023-11-23
DBZ-7169 - populating processed transaction cache only if lobs enabled

sherpa003:
to be in sync with memory event processor

I tested that with this change the connector runs stable, finally ( for our case). However I'm struggling to make Docker based integration test working on my machine so I did not run those.


Naros (approved):
LGTM, if CI passes, I'll merge.  Thanks @sherpa003 !

Naros:
Applied to main (=2.5).

PR #5032, opened 2023-11-24
[docs] Adding missing config

gunnarmorling:
Hey @jpechane, @Naros, quick doc fix.

### DIFF
--- a/documentation/modules/ROOT/pages/development/engine.adoc
+++ b/documentation/modules/ROOT/pages/development/engine.adoc
@@ -85,8 +85,9 @@ Here's an example of code that configures and runs an embedded xref:{link-mysql-
 [source,java,indent=0]
 ----
     // Define the configuration for the Debezium Engine with MySQL connector...
-    final Properties props = config.asProperties();
+    final Properties props = new Properties();
     props.setProperty("name", "engine");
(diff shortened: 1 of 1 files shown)


documentation/modules/ROOT/pages/development/engine.adoc
```diff
@@ -85,8 +85,9 @@ Here's an example of code that configures and runs an embedded xref:{link-mysql-
 [source,java,indent=0]
 ----
     // Define the configuration for the Debezium Engine with MySQL connector...
-    final Properties props = config.asProperties();
```
gunnarmorling, line 88: It's not clear what `config` is, seems not needed?

documentation/modules/ROOT/pages/development/engine.adoc
```diff
@@ -85,8 +85,9 @@ Here's an example of code that configures and runs an embedded xref:{link-mysql-
...
+    final Properties props = new Properties();
     props.setProperty("name", "engine");
+    props.setProperty("connector.class", "io.debezium.connector.mysql.MySqlConnector");
```
gunnarmorling, line 90: this showed up in the break-down below, but was missing here, which is in the way of copy and paste of the entire snippet.

jpechane:
@gunnarmorling Applied, thanks a lot. It is probably a leftover of old embedded engine docs.

PR #5033, opened 2023-11-24
DBZ-7181 Avoiding queries not available in recovery mode

gunnarmorling:
Fixes https://issues.redhat.com/browse/DBZ-7181


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresOffsetContext.java
```diff
@@ -232,7 +232,7 @@ public static PostgresOffsetContext initialContext(PostgresConnectorConfig conne
         try {
             LOGGER.info("Creating initial offset context");
             final Lsn lsn = Lsn.valueOf(jdbcConnection.currentXLogLocation());
-            final long txId = jdbcConnection.currentTransactionId().longValue();
+            final Long txId = jdbcConnection.currentTransactionId();
```
gunnarmorling, line 235: Code elsewhere was prepared for the TX id to be null already, so I think the change just highlighted this incorrect usage here.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
```diff
@@ -529,7 +529,7 @@ public Long currentTransactionId() throws SQLException {
     public long currentXLogLocation() throws SQLException {
         AtomicLong result = new AtomicLong(0);
         int majorVersion = connection().getMetaData().getDatabaseMajorVersion();
-        query(majorVersion >= 10 ? "select * from pg_current_wal_lsn()" : "select * from pg_current_xlog_location()", rs -> {
+        query(majorVersion >= 10 ? "select (case pg_is_in_recovery() when 't' then pg_last_wal_receive_lsn() else pg_current_wal_lsn() end) AS pg_current_wal_lsn" : "select * from pg_current_xlog_location()", rs -> {
```
gunnarmorling, line 532: This mimics similar implementations elsewhere, e.g. [here](https://github.com/mblp/postgres_exporter/commit/11cfc43fe9326e74ce46b9b9358a5b4ed87be513).

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
```diff
@@ -511,7 +511,7 @@ public synchronized void close() {
      */
     public Long currentTransactionId() throws SQLException {
         AtomicLong txId = new AtomicLong(0);
-        query("select * from txid_current()", rs -> {
+        query("select (case pg_is_in_recovery() when 't' then 0 else txid_current() end) AS pg_current_txid", rs -> {
```
gunnarmorling, line 514: I think returning `null` in the stand-by case from this method is fine, the tx id is only used for contextual info anyways. Note this only makes a difference upon start-up, actual change events do have the tx id as per the transaction from the primary.

gunnarmorling (commented):
Tested this with a primary/stand-by set-up on RDS, all looking good. I don't think setting up a replicated set-up for automated testing is really worth it, as it's quite involved. Unless someone feels up for it :)

jpechane:
@gunnarmorling Applied, thanks a lot!
We'd definitely like to do for Postgres the same thing we do for MySQL - running ITs for different topologies as part of `assembley profile. The Jira is there but work has not started yet.

PR #5035, opened 2023-11-25
DBZ-7182: MongoDB : allow array with documents with optional fields

olivierboudet:
### Which use case/requirement will be addressed by the proposed feature?
I would like to use Debezium with MongoDB. But in our use case, some documents have arrays of documents of same types BUT with optional fields.

In current implementation, it is not allowed, and this is documented : 

> 
> array encoding
> 
> If array.encoding is set to array (the default), the SMT encodes uses the array datatype to encode arrays in the original message. To ensure correct processing, all elements in an array instance must be of the same type. This option is a restricting one, but it enables downstream clients to easily process arrays.

I was wondering if at least the schema could be generated based on the document with the maximum number of fields. The restriction is always true, but a little less strict... and might allow us to use debezium. Otherwise Debezium is not usable for us


jpechane:
@olivierboudet Hi, thanks for the PR. We'd definitely like that functionality. But I think we should intorduce a separate array handling mode that will do the array union schema.

Naros:

Hi @olivierboudet, please sign off all commits with:

```
git commit -s
```

Naros:

PR #5036, opened 2023-11-25
Support MySQL 8 high resolution replication timestamps from GTID events

methodmissing:
References and depends on https://github.com/osheroff/mysql-binlog-connector-java/pull/130

## Gist of the changes

More accurate metrics and also `source.ts_ms` for `MySQL >= 8.0.1` with `GTID` replication mode enabled. At Shopify we're doubling down on lower latency business use cases for CDC and we're aware of the second resolution (int4) timestamp constraint in binlog event headers for some time.

This issue and Github pull requests to the binlog client and Debezium is an attempt to engage with the community to report higher resolution metrics and `source.ts_ms` in a backwards compatible way.

Some example graphs from a lower environment connecting to MySQL 8 writers where the higher resolution timestamps are rendered - a lot easier to spot trends and variance for the writer -> binlog client hop.

Screenshot 2023-11-26 at 00 23 13

Screenshot 2023-11-26 at 00 23 32

### Dead code removals

I could not find any references to these methods anymore:

* `SourceInfo::setBinlogTimestampSeconds`
* `SourceInfo::getBinlogTimestampSeconds`

### Alignment of event time extraction in onEvent and handleEvent

I preferred to handle event timestamp extraction in `MySqlStreamingChangeEventSource::onEvent` as both handlers are called in order by the binlog client event listeners feature. That way emitted metrics values and `source.ts_ms` key off the same `Instant` instance. 

### Event timestamp logic

MySQL 8.0.1 introduced [2 new microsecond resolution timestamps](https://webyog.com/blog/monyog/replication-performance-enhancements-mysql-8/) on GTID events to track replication lag more accurately:

* [immediate_commit_timestamp](https://github.com/percona/percona-server/blob/c610bb54c2f122729741aa39866adc758fe23b4e/libbinlogevents/include/control_events.h#L1005-L1009), microsecond resolution timestamp from the immediate master (or equal to original_commit_timestamp if connecting to master) since MySQL 8.0.1
* [original_commit_timestamp](https://github.com/percona/percona-server/blob/c610bb54c2f122729741aa39866adc758fe23b4e/libbinlogevents/include/control_events.h#L1005-L1009), microsecond resolution timestamp of the commit on the originating master (or equal to original_commit_timestamp if connecting to master) since MySQL 8.0.1

`original_commit_timestamp` is representative of the original commit time on the MySQL writer.

The order of precedence is:

* IF GTID mode is enabled and `GtidEventData::getOriginalCommitTimestamp` returns a non-zero value (MySQL >= 8.0.1), use the high resolution timestamp as set by the writer
* Otherwise fallback to the second resolution timestamps

### Other due diligence checks

* Snapshots set the current time in milliseconds for `source.ts_ms` already, I did not find any timestamp resolution issues with aligning on millisecond resolution for the streaming path as well

Jira issue: https://issues.redhat.com/projects/DBZ/issues/DBZ-7183

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java
@@ -9,6 +9,7 @@ import java.io.IOException;
 import java.sql.SQLException;
 import java.time.Duration;
 import java.time.Instant;
+import java.time.temporal.ChronoUnit;
 import java.util.EnumMap;
 import java.util.List;
 import java.util.Map;
@@ -87,6 +88,7 @@ public class MySqlStreamingChangeEventSource implements StreamingChangeEventSour
     private long initialEventsToSkip = 0L;
     private boolean skipEvent = false;
     private boolean ignoreDmlEventByGtidSource = false;
+    private final boolean isGtidModeEnabled;
     private final Predicate<String> gtidDmlSourceFilter;
     private final AtomicLong totalRecordCounter = new AtomicLong();
     private volatile Map<String, ?> lastOffset = null;
@@ -190,6 +192,7 @@ public class MySqlStreamingChangeEventSource implements StreamingChangeEventSour
         Configuration configuration = connectorConfig.getConfig();
         boolean filterDmlEventsByGtidSource = configuration.getBoolean(MySqlConnectorConfig.GTID_SOURCE_FILTER_DML_EVENTS);
         gtidDmlSourceFilter = filterDmlEventsByGtidSource ? connectorConfig.gtidSourceFilter() : null;
+        isGtidModeEnabled = connection.isGtidModeEnabled();
     }
 
     protected void onEvent(MySqlOffsetContext offsetContext, Event event) {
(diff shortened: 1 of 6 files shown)


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java
```diff
@@ -212,7 +214,25 @@ protected void onEvent(MySqlOffsetContext offsetContext, Event event) {
...
+                    eventTimestamp = Instant.ofEpochMilli(eventTs);
+                } else {
+                    // >= MySQL 8.0.1, prefer the higher resolution replication timestamp
+                    eventTimestamp = Instant.EPOCH.plus(gtidEventTs, ChronoUnit.MICROS);
+                }
+            }
+        } else {
+            // Fallback to second resolution event timestamps
+            eventTimestamp = Instant.ofEpochMilli(eventTs);
+        }
```
Naros, line 233: May I suggest refactoring this block into a separate method and having only two statements, the one that determines the ">= MySQL 8.0.1" behavior returning the higher precision timestamp or the fallback that handles all other cases?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java
```diff
@@ -87,6 +88,7 @@ public class MySqlStreamingChangeEventSource implements StreamingChangeEventSour
     private long initialEventsToSkip = 0L;
     private boolean skipEvent = false;
     private boolean ignoreDmlEventByGtidSource = false;
+    private boolean isGtidModeEnabled = false;
```
Naros, line 91: If you move this to an instance variable,  it will make more sense to initialize this value in the constructor and mark it as `final` as this shouldn't change over the runtime of the connector.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java
methodmissing: Thanks for the review Chris, on it.

Naros:
Hi @methodmissing, would you please rebase against the latest `main`? Thanks.

methodmissing:
Rebased, expected to fail on missing symbol `gtidEvent::getOriginalCommitTimestamp`:

```
[INFO] -------------------------------------------------------------
[INFO] -------------------------------------------------------------
Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/debezium/debezium/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java:[230,51] cannot find symbol
  symbol:   method getOriginalCommitTimestamp()
  location: variable gtidEvent of type com.github.shyiko.mysql.binlog.event.GtidEventData
```

Requires `0.28.4` of the binlog client released [here](https://github.com/osheroff/mysql-binlog-connector-java/pull/130) and version bump [here](https://github.com/debezium/debezium/blob/7fc7e10cc287988760efdd5bafa0190765165ec7/pom.xml#L127). Unless there's a different and preferred path without releasing `0.28.4` first.

Naros:
Waits for the upstream MySQL binlog client 0.28.4.  

Tested this locally with the upstream PR in the client, both MariaDB and MySQL passed with `mysql-ci-gtids` profile, so we'll run the full suite once the upstream client has been published.  For now, I expect the tests to fail due to the missing artifact.

Naros (approved):

Naros:
Applied, thanks @methodmissing !

zackary-naas:
Not sure where else to comment, so please send me to the right place if this is incorrect. Wondering if there is a way for us to also add the immediate_commit_timestamp to the events for MySQL >= 8.0. We'd love to be able to track the time from the replica write to when Debezium processes the binlog event. Thanks!

Naros:
Hi @zackary-naas, please raise a Jira issue for this enhancement, and reference this specific PR as a reference.  Thanks.

PR #5038, opened 2023-11-27
DBZ-7152 Allow singleUpdateStatement to use tableSources clause instead of only table name

ani-sha:
https://issues.redhat.com/browse/DBZ-7152


jpechane:
@ani-sha Applied, thanks!

PR #5039, opened 2023-11-27
DBZ-7184 Change event buffering for MongoDB connector

jcechace:
https://issues.redhat.com/browse/DBZ-7184

@jpechane  this is an initial implementation, the following needs to be addressed

- [x] Use bounded queue with configurable size (or manually control the size and block the fetcher when threshold is reached)
- [x] Tidy and structure the processing loop in streaming source (this includes the fragment handling code which I think could at this point be moved into this new cursor implementation)
- [x] ~~`Cursor#next()` should probably be blocking rather then throwing an exception~~
- [x] Probably consolidate offset handling (we shouldn't rely on the cursor anymore but rather on the `ResumableEvent#resumeToken`)

Note: heartbeat event dispatch is now affected by the no event throttling (this means that a heartbeat can only occur when empty `ResumableEvent` is polled out of the buffer queue -- that corresponds to 5 empty reads of the underlying change stream.

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
@@ -14,6 +14,7 @@ import java.util.Queue;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Semaphore;
+import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
 
@@ -118,6 +119,8 @@ public class BufferingChangeStreamCursor<TResult> implements MongoChangeStreamCu
      */
     public static final class EventFetcher<TResult> implements Runnable, Closeable {
 
+        public static final long QUEUE_OFFER_TIMEOUT_MS = 100;
+
         private final ChangeStreamIterable<TResult> stream;
         private final Semaphore capacity;
         private final Queue<ResumableChangeStreamEvent<TResult>> queue;
@@ -205,13 +208,18 @@ public class BufferingChangeStreamCursor<TResult> implements MongoChangeStreamCu
         }
 
         private void fetchEvents(MongoChangeStreamCursor<ChangeStreamDocument<TResult>> cursor) throws InterruptedException {
+            ResumableChangeStreamEvent<TResult> lastEvent = null;
+            var repeat = false;
             while (isRunning()) {
-                var maybeEvent = fetchEvent(cursor);
-                if (maybeEvent.isEmpty()) {
-                    LOGGER.warn("Resume token not available on this poll");
-                    continue;
+                if (!repeat) {
+                    var maybeEvent = fetchEvent(cursor);
+                    if (maybeEvent.isEmpty()) {
+                        LOGGER.warn("Resume token not available on this poll");
+                        continue;
+                    }
+                    lastEvent = maybeEvent.get();
                 }
-                enqueue(maybeEvent.get());
+                repeat = !enqueue(lastEvent);
             }
         }
 
@@ -240,9 +248,14 @@ public class BufferingChangeStreamCursor<TResult> implements MongoChangeStreamCu
             }
         }
 
-        private void enqueue(ResumableChangeStreamEvent<TResult> event) throws InterruptedException {
-            this.capacity.acquire();
-            queue.offer(event);
+        private boolean enqueue(ResumableChangeStreamEvent<TResult> event) throws InterruptedException {
+            var available = this.capacity.tryAcquire(QUEUE_OFFER_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+            if (!available) {
+                LOGGER.warn("Unable to acquire buffer lock, buffer queue is likely full");
+                return false;
+            }
+            // always true
+            return queue.offer(event);
         }
     }
 


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -0,0 +1,303 @@
...
+                cursorRef.set(null);
+                close();
+            }
+        }
+
+        private void fetchEvents(MongoChangeStreamCursor<ChangeStreamDocument<TResult>> cursor) throws InterruptedException {
+            while (isRunning()) {
+                var beforeEventPollTime = clock.currentTimeAsInstant();
+                var event = new ResumableChangeStreamEvent<>(cursor.next(), cursor);
+                event.document.ifPresent(doc -> LOGGER.trace("Polled Change Stream event: {}", event));
```
jcechace, line 191: this is probably not optimal as we should ensure that ResumableEvents are only enqueued when the presence of resume token is guaranteed (it might not be in some "Mongo compatible" databases and in case of bugs like SERVER-63772) 

in such case the time diff might be captured as higher. 

@jpechane  also please remind me which of the metrics we wanted to kept? poll time for event / no event, count of no events? Count of events as added recently is duplicit IMHO and should be removed.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -0,0 +1,303 @@
...
+        }
+    }
+
+    public static <TResult> BufferingChangeStreamCursor<TResult> fromIterable(
+                                                                              ChangeStreamIterable<TResult> stream,
+                                                                              MongoDbTaskContext taskContext,
+                                                                              MongoDbStreamingChangeEventSourceMetrics metrics,
+                                                                              Clock clock) {
+        var pause = DelayStrategy.constant(taskContext.getConnectorConfig().getPollInterval());
+        var fetcher = new EventFetcher<>(stream, 1000, pause, metrics, clock);
```
jcechace, line 224: Do we want to reuse some config property or introduce a new one ?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -0,0 +1,303 @@
...
+    }
+
+    @Override
+    public ResumableChangeStreamEvent<TResult> tryNext() {
+        var event = fetcher.poll();
+        if (event != null) {
+            lastResumeToken = event.resumeToken;
+        }
+        return event;
+    }
```
jcechace, line 253: @jpechane  if we want to add the throttling here, how would you envision it?  Say this method is called and no event is available. It doesn't make sense, IMHO, to try N-times and then pause before returning null.  Maybe count the number of empty polls on the class level and if the N previous calls return null we pause before trying again?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/metrics/MongoDbStreamingChangeEventSourceMetrics.java
```diff
@@ -77,13 +74,13 @@ public long getNumberOfEmptyPolls() {
...
-    public void onSourceEventPolled(ChangeStreamDocument<BsonDocument> event, Clock clock, Instant prePollTimestamp) {
+    public void onSourceEventPolled(ResumableChangeStreamEvent<?> event, Clock clock, Instant prePollTimestamp) {
         var now = clock.currentTimeAsInstant();
 
         AtomicLong counter;
         AtomicLong timer;
 
-        if (event == null) {
+        if (event.isEmpty()) {
             counter = numberOfEmptyPolls;
```
jcechace, line 84: @jpechane  please remind me which of these metrics we wanted to keep? poll time for event / no event, count of no events? Count of events as added recently is duplicit IMHO and should be removed.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
jpechane: Or maybe a progressive pause would be better here, starting with a small one? If it works nicely we could probably adopt it for the main loop too.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/metrics/MongoDbStreamingChangeEventSourceMetrics.java
jpechane: Count of no events and poll time for event. IIRC poll time for no event did not provide any info

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
jcechace: OK. I will look into it. By main loop I suppose you mean the one in the fetcher?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/metrics/MongoDbStreamingChangeEventSourceMetrics.java
jcechace: OK... I will rearrange things to the right places in this PR and remove the useless metrics in a sepparrate one

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
jcechace: What about we do introduce a new config (e.g. `cursor.max.batch.size`) defaulting either to zero or the value of `snapshot.fetch.size`  and then have the queue sized as  `min(3*cursor.max.batch.size, max.batch.size)`.  The obvious problem is that the default cursor.max.batch.size is 0 which means server default (usually as "I'l try fill 16MB")

So maybe we could just set it to `max.batch.size`. I think we should be careful here as this increases the required memory. So in a sense standalone config property defaulting to `max.batch.size` is really the optimal, however I understand the need not to increase the already overwhelming number of configuration properties we have :(

So maybe reusing the `max.batch.size` is really the way to go here

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
jpechane: Sorry, yes

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
jpechane: Well max.batech.size is always <= max.queue.size

jcechace: I see, we ensure this in validation. So a value of `max.batch.size` sounds as a good start I think.

jcechace: That also makes quite predictable what is the maximum number of records/events the connector can keep in memory during runtime. `max.queue.size` of records + `max.batch.size` resumable events. In reality this maximum is never going to be achieved unless KC consumer takes a vacation.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -0,0 +1,337 @@
...
+        boolean slept;
+        ResumableChangeStreamEvent<TResult> event;
+
+        do {
+            event = fetcher.poll();
+            slept = throttler.sleepWhen(event == null);
+        } while (slept);
+
+        return event;
+    }
```
jcechace, line 302: This seems to work nicely in combination with the exponential delay strategy -- which I took the librety of slightly modifying (added the ability to also reset the time when the maximum delay is reached).

jcechace: It is questionable whether we should use this also in the fetcher loop. 

The same approach in the fetcher would actually increase the total sleep time. Since we would exponentially sleep after each empty polls. Considering there is a already an implicit timeout on the Mongo side I think the time sleeping might actually increase quite a bit. 

We could also combine the current approach with exponential delay and instead of flat delay every 5 empty polls we would engage the exponential delay. So every 5 empty events we would progressively wait longer until the max delay is reached and then reset. While that increases the number of cursor polls I think it would still provide sufficient throttling -- which would be at the same time lower compared to the one in the prefetching cursor, which is exactly what we want (as we don't want the fetcher to throttle more than reads from the buffer queue)

jpechane: @jcechace Sunds good

jcechace:
@jpechane unless there are some unexpected failures, I think this is ready for a more in depth review. The only thing missing is the progressive throttling in fetcher loop.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -0,0 +1,358 @@
...
```
jcechace, line 323: @jpechane I asked  @Naros to also take a a look and one thing he is worried about is blocking in the main thread  as it can block connector shutdown for up to poll interval. 

What's your take on this? Maybe the queue wouldn't mind being poll constantly and we should indeed just use some microsleep intervals.

Naros: I wouldn't deviate from what you have if it works. The hard sleep in the main loop based on the poll interval and its impact on shutdown, when a user has a considerable poll time, is a rare use case and one I'm fine tackling as a separate concern as it would affect all our connectors anyway.

jpechane: @jcechace Is not this interruptible? If yes than it can either end up in `InterruptedException` or to be even more sure the loop can have additional condition relatied to `isInterrupted` or a running flag.

jcechace: @jpechane I thought it was ... @Naros  seems to think that the DelayStrategy when sleeping will postpone connector shutdown.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -0,0 +1,371 @@
...
+        private boolean enqueue(ResumableChangeStreamEvent<TResult> event) throws InterruptedException {
+            var available = this.capacity.tryAcquire(QUEUE_OFFER_TIMEOUT_MS, TimeUnit.MILLISECONDS);
+            if (!available) {
+                LOGGER.warn("Unable to acquire buffer lock, buffer queue is likely full");
+                return false;
+            }
+            // always true
+            return queue.offer(event);
+        }
+    }
```
jcechace, line 260: @jpechane I think static value for the timeout should be OK, no? We can always make it configurable in the future

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -0,0 +1,371 @@
...
+                    var maybeEvent = fetchEvent(cursor);
+                    if (maybeEvent.isEmpty()) {
+                        LOGGER.warn("Resume token not available on this poll");
+                        continue;
+                    }
+                    lastEvent = maybeEvent.get();
+                }
+                repeat = !enqueue(lastEvent);
+            }
+        }
```
jcechace, line 224: @jpechane unfortunately this looks like this now. Good enough I guess... :/

jpechane:
@jcechace Applied, thanks a lot!

PR #5040, opened 2023-11-27
DBZ-7186 Cleanup event processing loop in streaming event source of MongoDB connector

jcechace:
https://issues.redhat.com/browse/DBZ-7186

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
@@ -146,33 +146,11 @@ public class MongoDbStreamingChangeEventSource implements StreamingChangeEventSo
 
     private void readChangeStream(MongoClient client, ReplicaSet replicaSet, ChangeEventSourceContext context,
                                   MongoDbOffsetContext offsetContext) {
+        LOGGER.info("Reading change stream for '{}'", replicaSet);
         final ReplicaSetPartition rsPartition = effectiveOffset.getReplicaSetPartition(replicaSet);
         final ReplicaSetOffsetContext rsOffsetContext = effectiveOffset.getReplicaSetOffsetContext(replicaSet);
+        final ChangeStreamIterable<BsonDocument> rsChangeStream = initChangeStream(client, rsOffsetContext);
 
-        LOGGER.info("Reading change stream for '{}'", replicaSet);
-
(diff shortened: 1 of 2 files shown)


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -77,22 +73,21 @@ public MongoDbStreamingChangeEventSource(MongoDbConnectorConfig connectorConfig,
 
     @Override
     public void init(MongoDbOffsetContext offsetContext) {
-
         this.effectiveOffset = offsetContext == null ? emptyOffsets(connectorConfig) : offsetContext;
     }
 
     @Override
-    public void execute(ChangeEventSourceContext context, MongoDbPartition partition, MongoDbOffsetContext offsetContext)
+    public void execute(ChangeEventSourceContext context, MongoDbPartition partition, MongoDbOffsetContext unused)
```
jcechace, line 80: @jpechane  this was previously changed by @mfvitale  to use the `effectiveOffset` field set in the `init` method. I'm just making it cleaner that this parameter is not used.

jpechane: @jcechace Instead of renaming the var it might be better to add a JavaDoc and explain why it is not used

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/SplitEventHandler.java
```diff
@@ -0,0 +1,98 @@
...
+import java.util.function.Function;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import com.mongodb.client.model.changestream.ChangeStreamDocument;
+
+import io.debezium.DebeziumException;
+
+public class SplitEventHandler<TResult> {
```
jpechane, line 22: Is there any advanatge in parametrising the class?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
jcechace: The renaming in some IDEs helps by hinting that this is unused on purpose. However I can surely add the docs. 

BTW: I think this is a case for majority of our connectors currently.

PR #5044, opened 2023-11-27
DBZ-7188 Upgrade Outbox Extension to Quarkus 3.5.3

Naros:
https://issues.redhat.com/browse/DBZ-7188


jpechane:
@Naros Applied, thanks

PR #5045, opened 2023-11-27
DBZ-7144 Upgrade protobuf-java-util to 3.25.1

Naros:
https://issues.redhat.com/browse/DBZ-7144


jpechane:
@Naros Applied, thanks

PR #5049, opened 2023-11-28
DBZ-7193 Unchanged toasted array columns are substituted with placeholder even if  REPLICA IDENTITY == FULL

Prabhu19:
Jira Issue: https://issues.redhat.com/browse/DBZ-7193
Affected Debezium versions > 2.0

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -568,3 +568,4 @@ Vincenzo Santonastaso
 Matan Cohen
 Don Seiler
 Ahmed Rachid Hazourli
+Pavithrananda Prabhu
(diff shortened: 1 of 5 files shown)


jpechane:
@Prabhu19 Hi, thanks for the PR. The code LGTM on the first look. Could you lease add a test that would verify the fixed behaviour? ideally one that will fail before fixing and passes after.

Prabhu19:
> @Prabhu19 Hi, thanks for the PR. The code LGTM on the first look. Could you lease add a test that would verify the fixed behaviour? ideally one that will fail before fixing and passes after.

@jpechane I added testcases can you please check it once?

Prabhu19:
> > @Prabhu19 Hi, thanks for the PR. The code LGTM on the first look. Could you lease add a test that would verify the fixed behaviour? ideally one that will fail before fixing and passes after.
> 
> @jpechane I added testcases can you please check it once?

I also verified against the current main branch and all the added testcases fail
Screenshot 2023-11-30 at 12 46 55

jpechane:
@Prabhu19 Applied, thanks a lot!

PR #5055, opened 2023-11-29
Dbz 7192 ispn abandoned transactions

sherpa003:
this closes the feature gap between memory and ISPN based processors.
It also fixes 2 bugs in existing ISPN processors impl (AFAIK): 
- the transactions abandonment was already implemented in ISPN version but the events were not removed from the event cache (it was staying there forever)
- if event came for abandoned transaction, it would be processed normally (not sure what would happen exactly, but not good situation anyway)

It also fixes dubious name `removeTransactionAndEventsFromCache` which was not doing what it says (e.g. in memory version it was not removing neither transaction nor events)

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
@@ -104,7 +104,6 @@ public abstract class AbstractLogMinerEventProcessor<T extends AbstractTransacti
     private Scn lastProcessedScn = Scn.NULL;
     private boolean sequenceUnavailable = false;
 
-
     private final Set<String> abandonedTransactionsCache = new HashSet<>();
 
     public AbstractLogMinerEventProcessor(ChangeEventSourceContext context,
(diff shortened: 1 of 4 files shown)


Naros:
> It also fixes dubious name removeTransactionAndEventsFromCache which was not doing what it says (e.g. in memory version it was not removing neither transaction nor events)

Good catch; this was because previously, in one of the original implementations, a `Transaction` actually held a collection of Events, and we eventually had to split into two separate caches due to a ISPN limitation.

Naros:
Test failure is unrelated, but I've seen attach failures more recently with XStream, and they're random, but the underlying fix can be made generically to address.  -- Opened [DBZ-7207](https://issues.redhat.com/browse/DBZ-7207) to follow-up on that fix.

Naros (approved):

Naros:
Applied to main (=2.5), thanks @sherpa003

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
```diff
@@ -592,11 +611,19 @@ protected void handleRollbackNotFoundInBuffer(LogMinerEventRow row) {
...
      * @param transaction the transaction instance, should never be {@code null}
+     * @param isAbandoned whether the removal is because transaction is being abandoned
      */
-    protected abstract void removeTransactionAndEventsFromCache(T transaction);
+    protected void cleanupAfterTransactionRemovedFromCache(T transaction, boolean isAbandoned) {
+        if (isAbandoned) {
+            abandonedTransactionsCache.remove(transaction.getTransactionId());
+        }
+        else {
+            abandonedTransactionsCache.add(transaction.getTransactionId());
```
sherpa003, line 624: this is definitely stupid and wrong 
this causes huge memory grow over time.

PR #5058, opened 2023-11-30
DBZ-7207 Allow XStream attach errors to be retried

Naros:
https://issues.redhat.com/browse/DBZ-7207


jpechane:
@Naros Applied, thanks a lot.

Could you please take a look at CI jobs? It seems to me that MySQL jobs (at least names) are somehow mangled.

jpechane:
Also I wonder it if should

1) Refactor the code in future and introduce a util class that provide that kind of retriable operation. I think we have it already at multiple places in the codebase.
2) We are adding hard-coded attempts and timeouts in different part of the codebase. Maybe we should just introduce two internal properties with sensible defaults and replace the hardcode values with them to make them configurable just in case.

PR #5059, opened 2023-11-30
DBZ-7208 - implemented check for NPE

sherpa003:
trivial fix


Naros:
@sherpa003, what exactly is this static method called from?  It seems unused from Debezium's code base.

PR #5061, opened 2023-11-30
DBZ-6834 Provide INSERT/DELETE semantics for incremental snapshot watermarking

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-6834

Todo:
- [x] Add support for MongoDB
- [x] Check DB2

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
@@ -274,7 +274,7 @@ public class EventDispatcher<P extends Partition, T extends DataCollectionId> im
 
                         LOGGER.trace("Received change record {} for {} operation on key {} with context {}", value, operation, key, offset);
 
-                        if (operation == Operation.CREATE && connectorConfig.isSignalDataCollection(dataCollectionId) && sourceSignalChannel != null) {
+                        if (isASignalEventToProcess(dataCollectionId, operation) && sourceSignalChannel != null) {
                             sourceSignalChannel.process(value);
 
                             if (signalProcessor != null) {
@@ -292,6 +292,11 @@ public class EventDispatcher<P extends Partition, T extends DataCollectionId> im
                             }
                             streamingReceiver.changeRecord(partition, schema, operation, key, value, offset, headers);
                         }
+                    }
+
+                    private boolean isASignalEventToProcess(T dataCollectionId, Operation operation) {
+                        return operation == Operation.CREATE &&
+                                connectorConfig.isSignalDataCollection(dataCollectionId);
                     }
                 });
                 handled = true;
--- a/debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/SignalBasedIncrementalSnapshotChangeEventSource.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/SignalBasedIncrementalSnapshotChangeEventSource.java
@@ -74,13 +74,13 @@ public class SignalBasedIncrementalSnapshotChangeEventSource<P extends Partition
     }
 
     @Override
-    protected void emitWindowClose(Partition partition, OffsetContext offsetContext) throws SQLException {
+    protected void emitWindowClose(Partition partition, OffsetContext offsetContext) throws Exception {
 
         String signalTableName = getSignalTableName(connectorConfig.getSignalingDataCollectionId());
 
         WatermarkWindowCloser watermarkWindowCloser = getWatermarkWindowCloser(connectorConfig, jdbcConnection, signalTableName);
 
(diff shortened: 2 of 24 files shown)


debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/SignalBasedIncrementalSnapshotChangeEventSource.java
```diff
@@ -70,11 +75,20 @@ protected void emitWindowOpen() throws SQLException {
...
-        jdbcConnection.commit();
+
+        String signalTableName = getSignalTableName(connectorConfig.getSignalingDataCollectionId());
+
+        WatermarkWindowCloser watermarkWindowCloser = getWatermarkWindowCloser(connectorConfig, jdbcConnection, signalTableName);
+
+        watermarkWindowCloser.closeWindows(partition, offsetContext, context.currentChunkId());
+    }
+
+    private WatermarkWindowCloser getWatermarkWindowCloser(CommonConnectorConfig connectorConfig, JdbcConnection jdbcConnection, String signalTable) {
```
jpechane, line 86: I wonder if the `WindowCloser` can be provided directly by `SnapshotWatermarkStrategy`?

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/WatermarkWindowCloser.java
```diff
@@ -0,0 +1,16 @@
...
+package io.debezium.pipeline.source.snapshot.incremental;
+
+import java.sql.SQLException;
+
+import io.debezium.pipeline.spi.OffsetContext;
+import io.debezium.pipeline.spi.Partition;
+
+public interface WatermarkWindowCloser {
+
+    void closeWindows(Partition partition, OffsetContext offsetContext, String chunkId) throws SQLException;
```
jpechane, line 15:
```suggestion
    void closeWindow(Partition partition, OffsetContext offsetContext, String chunkId) throws SQLException;
```

debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
```diff
@@ -293,6 +293,11 @@ public void changeRecord(P partition,
                             streamingReceiver.changeRecord(partition, schema, operation, key, value, offset, headers);
                         }
                     }
+
+                    private boolean isASignalEventToProcess(T dataCollectionId, Operation operation) {
+                        return operation == Operation.CREATE &&
```
jpechane, line 298: Should not this support `DELETE` as well?

jpechane (commented):
@mfvitale I like the way the PR evolves. I left couple of preliminary comments to think about.

debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
mfvitale: I did myself the same question. Enabling `DELETE` will avoid to directly close the window because it will be then processed when the delete happens instead of close insert. I was going to this path but then I thought that this will have a side effect. If the user delete something on the signal table the event will be processed again. So i preferred to use this approach.

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/SignalBasedIncrementalSnapshotChangeEventSource.java
mfvitale: It can be an option but for now, since it only used here, maybe this approach can be fine. 

I feel the missing of `Dependency Injection` :disappointed_relieved:

mfvitale:
DB2 failures are not related.

mfvitale:
> Yes, I reverted the change already

I saw, maybe I just need to invalidate the old cache.

debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
jpechane: @mfvitale One question. What if `DELETE` is executed but the connector crashesh before the offsets are saved. What will happen then?

mfvitale: Having a look to the code again seems that since the commit is after the call to the `closeWindow` we should be fine also in case of crash before the call to it. Am I right?

jpechane:
@mfvitale Applied, thanks!

debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
jpechane: ok, I tried to come up with multiple scenarios of when the crash can happen it seems the soultion is resilient to it.

PR #5064, opened 2023-12-01
DBZ-6764 DBZ-7178 DBZ-7177 Connector metrics implementation for REST extension

ani-sha:
https://issues.redhat.com/browse/DBZ-6764
https://issues.redhat.com/browse/DBZ-7178
https://issues.redhat.com/browse/DBZ-7177


debezium-core/src/main/java/io/debezium/rest/MetricsResource.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.kafka.connect.connector.Connector;
+import org.json.simple.JSONObject;
+
+import io.debezium.rest.jolokia.JolokiaAttributes;
+import io.debezium.rest.jolokia.JolokiaClient;
+
+public interface MetricsResource {
+    Connector getConnector();
+
+    String getAttributesFilePath();
```
jpechane, line 25: Is this really necessary? Do we know for what connector we are looking for metrics? If we do then it is just possible to set the pattern the attribute name files have to follow and just calculate it.
Also is it possible to have a common file with all attributes shared by all connectors and a per-connector definition?

debezium-core/src/main/java/io/debezium/rest/MetricsResource.java
```diff
@@ -0,0 +1,43 @@
...
+    String CONNECTOR_METRICS_ENDPOINT = "/connectors/{connector-name}/metrics";
+
+    @GET
+    @Path(CONNECTOR_METRICS_ENDPOINT)
+    @Produces(MediaType.APPLICATION_JSON)
+    default List<JSONObject> getConnectorMetrics(@PathParam("connector-name") String connectorName) {
+        if (getConnector() == null) {
+            throw new RuntimeException("Unable to fetch metrics for connector " + connectorName + " as the connector is not available.");
+        }
+        JolokiaClient jolokiaClient = new JolokiaClient();
```
jpechane, line 38: Why do we need to use Jolokia? Can't we just use `MBeanServer` to query the metrics as we are running in the same JVM?

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/rest/DebeziumOracleConnectorResourceIT.java
```diff
@@ -178,4 +179,31 @@ public void testFiltersWithInvalidTableIncludeList() {
                         "The 'table.include.list' value is invalid: A comma-separated list of valid regular expressions is expected, but Dangling meta character '+' near index 0\n+\n^"));
     }
 
+    @Test
+    public void testMetricsEndpoint() {
```
jpechane, line 183: It seems to me that the test is the same for all connectors. IMHO it is about time to intorduce a common ancestor to all connector resource ITs and you could move it there.

debezium-core/src/main/java/io/debezium/rest/MetricsResource.java
jpechane: Is it even necessary to have the definition done in files? Would it be technically possible to re-use `io.debezium.metadata.ConnectorMetadata` for it?

rk3rn3r: `ConnectorMetadata` class is used for the SchemaGenerator. I would personally not cross concerns with the REST extensions. Furthermore I think it is a good idea to have the implementation like this. That way each connector can freely decide how it provides the metrics definitions. Also mind in the future the REST extension needs to provide different versions of Debezium connectors. Using something hardcoded (outside the connector modules) is imo not a good idea. That way the connector can freely decide where and how metrics attributes are returned.
But maybe we could return something more "explicitly typed" here? Like having a helper method somewhere that loads the list of attributes from a file, but this method `getAttributesFilePath()`  could return a `List<String>` or `List<MetricsAttribute>` or even a new `MetricsAttributes` instance and let each connector give multiple options how this is created (from that txt files like now, a CONST, etc). wdyt @ani-sha  @jpechane ?

debezium-core/src/main/java/io/debezium/rest/MetricsResource.java
rk3rn3r: Yes we can, but in order to quickly finish the migration from the UI backend to the REST extensions we agreed on a temporary more-or-less 1:1 migration from the UI backend. I filed [DBZ-7177](https://issues.redhat.com/browse/DBZ-7177) for this back then.

ani-sha: @rk3rn3r As suggested by @jpechane to use`MBeanServer` directly instead in this code change instead of just adding Jolokia and then removing it in future release. Also some metrics are not available and it is difficult to find the root cause with the current changes. So I am reworking on this to use the `MBeanServer` and remove Jolokia implementation. Hopefully it will help to identify the problem with SQL Server metrics.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/rest/DebeziumOracleConnectorResourceIT.java
jpechane: Well, I'd disagree in the claim of diveregence. There will be a large set of metrics that is shared by all connectors. So it makes sense to have a single method testing all common metrics and then per-connector methods to test the distinct metrics.

debezium-core/src/main/java/io/debezium/rest/MetricsResource.java
jpechane: I'd still like to see union of common config file and per-connector file as very large number of metrics is shared.

jpechane: And again if we come back to my original comment, do we really need it if the pattern is the same so the path can be just calculated?

rk3rn3r: Do you think providing this flexibility does any harm? Also think about 3rd party connectors or community connectors that should eventually add up to the Debezium UI feature set. They might require flexibility.

debezium-core/src/main/java/io/debezium/rest/metrics/MetricsDescriptor.java
```diff
@@ -0,0 +1,77 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.rest.metrics;
```
rk3rn3r, line 6: Please move it to the `io.debezium.rest.model` package.

debezium-core/src/main/java/io/debezium/rest/metrics/MetricsAttributes.java
```diff
@@ -0,0 +1,24 @@
...
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.rest.metrics;
+
+import java.util.List;
+
+/**
+ * Class containing the attributes names for the metrics exposed via REST API
+ */
+public class MetricsAttributes {
```
jpechane, line 13: PLease convert into interface and implementation.

debezium-core/src/main/java/io/debezium/rest/MetricsResource.java
```diff
@@ -0,0 +1,87 @@
...
+import javax.ws.rs.Produces;
+import javax.ws.rs.core.MediaType;
+
+import org.apache.kafka.connect.connector.Connector;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.debezium.rest.metrics.MetricsDescriptor;
+
+public interface MetricsResource {
```
jpechane, line 33: PLease add class `getMetricsAttributes` method.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/rest/DebeziumPostgresConnectorResource.java
```diff
@@ -53,6 +66,24 @@ public Connector getConnector() {
         return new PostgresConnector();
     }
 
+    @Override
+    public MetricsDescriptor getMetrics(String connectorName, MBeanServer mBeanServer) throws MalformedObjectNameException {
+        Map<String, String> connectorConfig = connectClusterState.connectorConfig(connectorName);
+        String serverName = connectorConfig.get("topic.prefix");
+        String tasksMax = connectorConfig.get("tasks.max");
+
+        ObjectName objectName = getObjectName("postgres", "streaming", serverName);
```
rk3rn3r, line 75: We should use constants for those strings, basically based on the connector's metrics definition. If we don't have constants their they should be introduced. This can be filed as new Jira and done later for sure.

debezium-core/src/main/java/io/debezium/rest/MetricsResource.java
rk3rn3r: We agreed on dropping Jolokia for several reasons, for example to keep the additional dependency out of connector modules.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectRestExtension.java
```diff
@@ -27,7 +27,7 @@ public class DebeziumMongoDbConnectRestExtension implements ConnectRestExtension
 
     @Override
     public void register(ConnectRestExtensionContext restPluginContext) {
-        restPluginContext.configurable().register(new DebeziumMongoDbConnectorResource());
+        restPluginContext.configurable().register(new DebeziumMongoDbConnectorResource(restPluginContext.clusterState()));
```
rk3rn3r, line 30: Wouldn't setter injection into a MetricsResource interface's default method make more sense? The base code is almost everywhere the same
```
        Map<String, String> connectorConfig = connectClusterState.connectorConfig(connectorName);
        String serverName = connectorConfig.get("topic.prefix");
        String tasksMax = connectorConfig.get("tasks.max");
```
That can be done inside the MetricsResource code in a default method, no?
@jpechane wdyt?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
```diff
@@ -51,6 +62,16 @@ public Connector getConnector() {
         return new MongoDbConnector();
     }
 
+    @Override
+    public MetricsDescriptor getMetrics(String connectorName, MBeanServer mBeanServer)
```
rk3rn3r, line 66: mBeanServer is never used here and in no other `Debezium*ConnectorResource#getMetrics()` call. The parameter can be removed here and from the interface. MBeanServer instance is only used inside `MetricsResource`  and I think it can be defined there, maybe even as a static const.

debezium-core/src/main/java/io/debezium/rest/MetricsResource.java
```diff
@@ -0,0 +1,132 @@
...
+
+import io.debezium.rest.model.MetricsAttributes;
+import io.debezium.rest.model.MetricsDescriptor;
+
+public interface MetricsResource extends MetricsAttributes {
+    String CONNECTOR_METRICS_ENDPOINT = "/connectors/{connector-name}/metrics";
+
+    Connector getConnector();
+
+    MetricsDescriptor getMetrics(String connectorName, MBeanServer mBeanServer)
```
rk3rn3r, line 38: I think, according to my other comment/s, the `MBeanServer` parameter can safely be removed. It is always passed to `#getMetrics()` but always only forwarded down to be used here in the MetricsResource class/interface/trait.

debezium-core/src/main/java/io/debezium/rest/MetricsResource.java
```diff
@@ -0,0 +1,132 @@
...
+    @Path(CONNECTOR_METRICS_ENDPOINT)
+    @Produces(MediaType.APPLICATION_JSON)
+    default MetricsDescriptor getConnectorMetrics(@PathParam("connector-name") String connectorName)
+            throws ReflectionException, MalformedObjectNameException, InstanceNotFoundException, IntrospectionException, AttributeNotFoundException, MBeanException {
+
+        if (getConnector() == null) {
+            throw new RuntimeException("Unable to fetch metrics for connector " + connectorName + " as the connector is not available.");
+        }
+
+        MBeanServer mBeanServer = ManagementFactory.getPlatformMBeanServer();
```
rk3rn3r, line 129: This can be moved into a private const here in the interface and can be removed as parameter on the next line.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/Module.java
```diff
@@ -35,4 +35,11 @@ public static String name() {
     public static String contextName() {
         return "Postgres";
     }
+
+    /**
+     * @return logical name of the connector used in mbean object name
+     */
+    public static String logicalName() {
```
rk3rn3r, line 42: `logicalName`  sounds very generic. The comment says it is used as MBean object name. Is it still right to use that generic name here? @jpechane

rk3rn3r:
@ani-sha The PR looks good overall. The only comments that I left are a bit about cleaning up some artifacts that I think were left from some refactoring process while writing and changing the code. Furthermore I had 2 small question for @jpechane where one is about de-duplication of some code and related dependencies, and the other is only about naming.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/Module.java
jpechane: I think the method is unnecessary and duplicates `contextName()`. If lowercase is needed for querying then just call `Module.contextName().toLowercase()`

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectRestExtension.java
jpechane: Yes, I agree if it is possible.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
jpechane: I think if you'd like to turn this into `default` method then it wuld be better to pass it as a parameter.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/Module.java
ani-sha: @jpechane That can done for `postgres`. But `contextName()` for SQL Server doesn't match. It should be `sql_server`.

jpechane: @ani-sha It can, `sql_server` is just lower case variant of `SQL_Server`

rk3rn3r: @ani-sha I would prefer addding a "metricsName()` method and put the same strings there as you did here.
```suggestion
    public static String metricsName() {
```
and leave the rest unchanged. I am not a fan of doing the lowercase call. It would be executed on every call to `#metricsName()`.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
rk3rn3r: @jpechane there is no call to MBeanServer in the implementations. @ani-sha do you think we require it as parameter for adding additional (connector-specific) metrics. Or can we add other ways to specify how addditional metrics can be added? Then it can be safely removed.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/Module.java
rk3rn3r: We agreed on that we will use `contextName().toLowerCase()` for now and I gonna file a Jira to replace all occurences of `contextName().toLowerCase()` with an additional parameter passed to the `Metrics` implementation in core with the respective name in order that with a single PR / task all metrics base on the same code.

PR #5066, opened 2023-12-03
DBZ-7216 Fix data schema name in case of metadata is in headers

rkudryashov:
https://issues.redhat.com/browse/DBZ-7216


rkudryashov:
For some reason, [ApicurioRegistryTest.java](https://github.com/debezium/debezium/pull/5066/files#diff-bf2ef0f512ce226c40fd738a68c9da0ab6135602228b80fdbd35ba66c4b60cce) started to fail so I had to update it too

jpechane:
@rkudryashov Applied, thanks!

PR #5069, opened 2023-12-04
DBZ-7193 Unchanged toasted array columns are substituted with placeholder even if REPLICA IDENTITY == FULL

Prabhu19:
Jira Issue: https://issues.redhat.com/browse/DBZ-7193
Affected Debezium versions > 2.0


jpechane:
@Prabhu19 Applied, thanks

PR #5073, opened 2023-12-05
DBZ-7167 Provide a public API from the connector implementations to retrieve the list of matching collections or tables based on the different include-/exclude lists

harveyyue:
https://issues.redhat.com/browse/DBZ-7167
Will implement the connector interface method `getMatchingCollections` to rest of connectors

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
@@ -5,23 +5,17 @@
  */
 package io.debezium.connector.mongodb.rest;
 
-import java.util.List;
-
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 
-import org.apache.kafka.connect.connector.Connector;
-
-import io.debezium.config.Configuration;
 import io.debezium.connector.mongodb.Module;
 import io.debezium.connector.mongodb.MongoDbConnector;
 import io.debezium.rest.ConnectionValidationResource;
 import io.debezium.rest.FilterValidationResource;
 import io.debezium.rest.SchemaResource;
-import io.debezium.rest.model.DataCollection;
 
 /**
  * A JAX-RS Resource class defining endpoints of the Debezium MongoDB Connect REST Extension
@@ -30,7 +24,8 @@ import io.debezium.rest.model.DataCollection;
 @Path(DebeziumMongoDbConnectorResource.BASE_PATH)
 @Produces(MediaType.APPLICATION_JSON)
 @Consumes(MediaType.APPLICATION_JSON)
-public class DebeziumMongoDbConnectorResource implements SchemaResource, ConnectionValidationResource, FilterValidationResource {
+public class DebeziumMongoDbConnectorResource
+        implements SchemaResource, ConnectionValidationResource<MongoDbConnector>, FilterValidationResource<MongoDbConnector> {
 
     public static final String BASE_PATH = "/debezium/mongodb";
     public static final String VERSION_ENDPOINT = "/version";
@@ -42,13 +37,8 @@ public class DebeziumMongoDbConnectorResource implements SchemaResource, Connect
     }
 
     @Override
-    public Connector getConnector() {
+    public MongoDbConnector getConnector() {
         return new MongoDbConnector();
-    }
-
-    @Override
-    public List<DataCollection> getMatchingCollections(Configuration configuration) {
-        return ((MongoDbConnector) getConnector()).getMatchingCollections(configuration);
     }
 
     @Override
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
@@ -5,17 +5,22 @@
  */
 package io.debezium.connector.mongodb.rest;
 
+import java.util.List;
+import java.util.stream.Collectors;
+
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
 import javax.ws.rs.Path;
 import javax.ws.rs.Produces;
 import javax.ws.rs.core.MediaType;
 
+import io.debezium.config.Configuration;
 import io.debezium.connector.mongodb.Module;
 import io.debezium.connector.mongodb.MongoDbConnector;
 import io.debezium.rest.ConnectionValidationResource;
 import io.debezium.rest.FilterValidationResource;
 import io.debezium.rest.SchemaResource;
+import io.debezium.rest.model.DataCollection;
 
 /**
  * A JAX-RS Resource class defining endpoints of the Debezium MongoDB Connect REST Extension
@@ -46,4 +51,10 @@ public class DebeziumMongoDbConnectorResource
         return "/META-INF/resources/mongodb.json";
     }
 
+    @Override
+    public List<DataCollection> getMatchingCollections(Configuration configuration) {
+        return getConnector().getMatchingCollections(configuration).stream()
+                .map(collectionId -> new DataCollection(collectionId.replicaSetName(), collectionId.dbName(), collectionId.name()))
+                .collect(Collectors.toList());
+    }
 }
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
@@ -6,7 +6,6 @@
 package io.debezium.connector.mongodb.rest;
 
 import java.util.List;
-import java.util.stream.Collectors;
 
 import javax.ws.rs.Consumes;
 import javax.ws.rs.GET;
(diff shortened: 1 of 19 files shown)


jpechane:
@harveyyue Thanks a lot for the PR. The intend of this change is twofold - the migration of the code **and** using the same code in the REST extension and snapshot/streaming interface. So there should be an overlap with changes on `io.debezium.relational.RelationalSnapshotChangeEventSource.getAllTableIds(RelationalSnapshotContext<P, O>)` and `io.debezium.pipeline.source.AbstractSnapshotChangeEventSource.determineDataCollectionsToBeSnapshotted(Collection<T>, Set<Pattern>)`. We want to make sure that the same pice of code is used both in UI provided metadata and in the connector itself to make sure the operate on the same set.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/rest/DebeziumMongoDbConnectorResource.java
```diff
@@ -51,24 +46,9 @@ public Connector getConnector() {
...
     public List<DataCollection> getMatchingCollections(Configuration configuration) {
-        try (MongoDbConnection connection = getConnection(configuration)) {
-            return connection.collections().stream()
-                    .map(collectionId -> new DataCollection(collectionId.replicaSetName(), collectionId.dbName(), collectionId.name()))
-                    .collect(Collectors.toList());
-        }
-        catch (InterruptedException e) {
-            throw new DebeziumException(e);
-        }
+        return ((MongoDbConnector) getConnector()).getMatchingCollections(configuration);
```
rk3rn3r, line 51: @jpechane Would it maybe make sense to add another interface on the connector classes/implementations and let `getConnector()` return this one here? That way we could move this code here into the FilterValidationResource class instead of having this duplicate coder everywhere and an unnecessary cast. Wdyt?

jpechane: Is the typecasting even necessary? IMHO it could be pulled-up to the parent class and just call `return getConnector().getMatchingCollections(configuration);`

jpechane: ok, I see what you mean. We don't have a common parent for all Debezium connectors. Maybe instead of adding a new interface we should add a common parent class for all connectors lakie we do for tasks so intorducing `BaseConnector` class?

rk3rn3r: Agreed!

jpechane (commented):
@harveyyue It looks very neat, thanks a lot!

I left just two monir comments related to code simplification and typing changes.

@rk3rn3r Could you please confirm there would be no issue if the resource interface would be generic and have a default method?

jpechane (approved):
@harveyyue LGTM, thanks a lot! Waiting for @rk3rn3r to comment before the merge.

harveyyue:
informix pr: https://github.com/debezium/debezium-connector-informix/pull/14

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/rest/DebeziumMySqlConnectorResource.java
```diff
@@ -49,41 +40,13 @@ public String getSchemaFilePath() {
...
-                        .map(tableId -> new DataCollection(tableId.catalog(), tableId.table()))
-                        .collect(Collectors.toList());
-                allMatchingTables.addAll(matchingTables);
-            }
-            return allMatchingTables;
-        }
-        catch (SQLException e) {
-            throw new DebeziumException(e);
-        }
+        return getConnector().getMatchingCollections(configuration);
```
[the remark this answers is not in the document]
rk3rn3r: I don't think so (if I got you right) because `getConnector()` and `getMatchingCollections()` have entirely connector specific code. There is not much in common or else we have to add a bunch of other methods to the interface to provide the diverging `*ConnectorConfig` instances `*Connection` classes etc.

debezium-core/src/main/java/io/debezium/rest/ConnectionValidationResource.java
```diff
@@ -10,13 +10,12 @@
 import javax.ws.rs.PUT;
 import javax.ws.rs.Path;
 
-import org.apache.kafka.connect.connector.Connector;
-
+import io.debezium.connector.common.BaseSourceConnector;
 import io.debezium.rest.model.ValidationResults;
 
 public interface ConnectionValidationResource {
```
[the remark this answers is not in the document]
rk3rn3r: Ah, now I see what you meant. :bulb:

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/rest/DebeziumMySqlConnectorResource.java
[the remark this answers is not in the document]
rk3rn3r: never mind, I think it looks very good now!

rk3rn3r:
LGTM. It looks like after marking some other tests as flaky now others (maybe some that did not run before because of earlier tests failing which now get skipped) are failing. I re-triggered builds for failing jobs and see how they resolve. I will merge the PR after this test run finished.
Thx a lot for a very important and valuable contribution @harveyyue!

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceConnector.java
```diff
@@ -0,0 +1,26 @@
...
+
+/**
+ * Base class for Debezium's CDC {@link SourceConnector} implementations.
+ * Provides functionality common to all CDC connectors, such as validation.
+ */
+public abstract class BaseSourceConnector extends SourceConnector {
+
+    protected abstract Map<String, ConfigValue> validateAllFields(Configuration config);
+
+    public abstract List<DataCollection> getMatchingCollections(Configuration config);
```
rk3rn3r, line 25: @jpechane  I also don't think we should use a response model from the REST extension here `List<DataCollection>`? Would TableId or something else match better the "internal connector" use-case while for the UI we then need to translate that into the DataCollection model?

harveyyue:
@rk3rn3r Have implemented db2, do you mean the other cassandra and spanner connectors?

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceConnector.java
jpechane: @rk3rn3r Well, you can't as MongoDB does not use `TableId`. What you can do it to have generic parameter `T extends DataCollectionId`.

rk3rn3r: I was not saying it must be `TableId`, just a suggestion (and I think those are new methods and if we decide that the method return `TableId` we could use a meaningful mapping to the TableId fields, this is required anyway). We could also create a value object as a copy from the `DataCollection` REST resonse model but I don't think it's a good idea to use the same entity for the connectors and for the REST response, although they are related there might be reasons in the future to diverge on the connector side implementation but we like to keep the REST response model unchanged.

rk3rn3r:
@harveyyue No, please have the look at the CI test error for Db2:

```
 Error:  COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
Error:  /home/runner/work/debezium/debezium/db2/src/main/java/io/debezium/connector/db2/Db2Connector.java:[33,8] io.debezium.connector.db2.Db2Connector is not abstract and does not override abstract method getMatchingCollections(io.debezium.config.Configuration) in io.debezium.connector.common.BaseSourceConnector
```
https://github.com/debezium/debezium/actions/runs/7139864033/job/19557779538?pr=5073#step:7:130

harveyyue:
> @harveyyue No, please have the look at the CI test error for Db2:
> 
> ```
>  Error:  COMPILATION ERROR : 
> [INFO] -------------------------------------------------------------
> Error:  /home/runner/work/debezium/debezium/db2/src/main/java/io/debezium/connector/db2/Db2Connector.java:[33,8] io.debezium.connector.db2.Db2Connector is not abstract and does not override abstract method getMatchingCollections(io.debezium.config.Configuration) in io.debezium.connector.common.BaseSourceConnector
> ```
> 
> https://github.com/debezium/debezium/actions/runs/7139864033/job/19557779538?pr=5073#step:7:130

Yes, there is another pr for db2 connector:
https://github.com/debezium/debezium-connector-db2/pull/124

jpechane:
@rk3rn3r It is a different repository, it must be different PR

rk3rn3r:
@harveyyue never mind. I am not 100% in shape today and in my main IDE workspace I have db2 repo symlinked. what make me think it is in the main repo. sorry :see_no_evil:

rk3rn3r:
But it still leaves the question what other entity we can use as value object to be returned by `List<Something> getMatchingCollections(Configuration config)`. I don't think it should be `io.debezium.rest.model.DataCollection` as this is purely meant as REST response model. @jpechane

jpechane:
Well, and here is another cofusion. We have `io.debezium.spi.schema.DataCollectionId` and IMHO it should be returned. I am sorry for not noticing the package name.

harveyyue:
> Well, and here is another cofusion. We have `io.debezium.spi.schema.DataCollectionId` and IMHO it should be returned. I am sorry for not noticing the package name. So the return value will be `T extends DataCollectionId` so releational connectors will return TableId and MongoDB will return CollectionId.

@rk3rn3r @jpechane Done.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
```diff
@@ -258,7 +253,9 @@ protected void releaseDataSnapshotLocks(RelationalSnapshotContext<MySqlPartition
                 }
```
rk3rn3r, line 253: Shouldn't similar code above ( https://github.com/debezium/debezium/pull/5073/files#diff-63840528f25c3335e78d8b91c2ea67602750af012c1620d9981f11cb476b480aR144-R154 )  not also use the new filter code? Maybe it must even be merged into the new method to handle only "readable databases". Whatever this is....

rk3rn3r:
LGTM @jpechane after those changes got applied. We could easily file Jiras for my other comment/s. wdyt?
I re-triggered tests to see if failures were intermittent or constant errors.

jpechane:
@harveyyue LGTM, applied, thanks a lot!
@rk3rn3r Please perpare the folow-up Jira. Please make sure that the task to use the new API in snapshot code is included.

PR #5075, opened 2023-12-05
DBZ-6834 Add docs for incremental.snapshot.watermarking.strategy

mfvitale:
Doc for : https://github.com/debezium/debezium/pull/5061

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/db2.adoc
+++ b/documentation/modules/ROOT/pages/connectors/db2.adoc
@@ -2752,6 +2752,14 @@ Increasing the chunk size provides greater efficiency, because the snapshot runs
 However, larger chunk sizes also require more memory to buffer the snapshot data.
 Adjust the chunk size to a value that provides the best performance in your environment.
 
+|[[db2-property-incremental-snapshot-watermarking-strategy]]<<db2-property-incremental-snapshot-watermarking-strategy, `+incremental.snapshot.watermarking.strategy+`>>
+|`insert_insert`
+|Specify the strategy used for watermarking during an incremental snapshot: +
++
+`insert_insert`: both open and close signal is written into signal data collection. +
++
+`insert_delete`: only open signal is written into signal data collection, the close one will delete the relative open signal. Useful to keep signal data collection size low. +
+
 
 |[[db2-property-topic-naming-strategy]]<<db2-property-topic-naming-strategy, `topic.naming.strategy`>>
 |`io.debezium.schema.SchemaTopicNamingStrategy`
--- a/documentation/modules/ROOT/pages/connectors/informix.adoc
+++ b/documentation/modules/ROOT/pages/connectors/informix.adoc
@@ -2391,6 +2391,14 @@ Increasing the chunk size provides greater efficiency, because the snapshot runs
 However, larger chunk sizes also require more memory to buffer the snapshot data.
 Adjust the chunk size to a value that provides the best performance in your environment.
 
+|[[informix-property-incremental-snapshot-watermarking-strategy]]<<informix-property-incremental-snapshot-watermarking-strategy, `+incremental.snapshot.watermarking.strategy+`>>
+|`insert_insert`
+|Specify the strategy used for watermarking during an incremental snapshot: +
++
+`insert_insert`: both open and close signal is written into signal data collection. +
++
+`insert_delete`: only open signal is written into signal data collection, the close one will delete the relative open signal. Useful to keep signal data collection size low. +
+
 
 |[[informix-property-topic-naming-strategy]]<<informix-property-topic-naming-strategy, `topic.naming.strategy`>>
 |`io.debezium.schema.SchemaTopicNamingStrategy`
--- a/documentation/modules/ROOT/pages/connectors/mongodb.adoc
+++ b/documentation/modules/ROOT/pages/connectors/mongodb.adoc
@@ -1967,6 +1967,17 @@ ifdef::product[]
 Incremental snapshots is a Technology Preview feature for the {prodname} MongoDB connector.
 endif::product[]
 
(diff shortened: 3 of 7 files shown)


documentation/modules/ROOT/pages/connectors/db2.adoc
```diff
@@ -2752,6 +2752,14 @@ Increasing the chunk size provides greater efficiency, because the snapshot runs
 However, larger chunk sizes also require more memory to buffer the snapshot data.
 Adjust the chunk size to a value that provides the best performance in your environment.
 
+|[[db2-property-incremental-snapshot-watermarking-strategy]]<<db2-property-incremental-snapshot-watermarking-strategy, `+incremental.snapshot.watermarking.strategy+`>>
+|`insert_insert`
+|Specify the strategy used for watermarking during an incremental snapshot: +
++
+`insert_insert`: both open and close signal is written into signal data collection. +
++
+`insert_delete`: only open signal is written into signal data collection, the close one will delete the relative open signal. Useful to keep signal data collection size low. +
```
roldanbob, line 2761:
```suggestion
`insert_insert`:: When you send a signal to initiate an incremental snapshot, both the open and close signals are written to the signaling data collection. +
+
`insert_delete`:: When the connector receives signals to initiate an incremental snapshot, it writes only the signal to open the snapshot to the signal data collection.
The signal to close the snapshot deletes the related open signal. Specify this option to limit the size of the signaling data collection. +
```
@mfvitale From what I recall, the mechanism of the watermark framework writes low and high watermark entries _to the transaction log_ to delimit the opening and closing of the window in which a snapshot captures a table chunk. Is that correct? 
I'm confused by the references to open and close signals being written to the signaling data collection vs. the log, but perhaps the mechanism has changed?

roldanbob (commented):
@mfvitale I added comments about the changes in `db2.adoc`, but because I had questions, I thought it best to await your responses before I propagate my comments to the other files.

documentation/modules/ROOT/pages/connectors/db2.adoc
mfvitale: > I recall, the mechanism of the watermark framework writes low and high watermark entries to the transaction log to delimit the opening and closing of the window in which a snapshot captures a table chunk. Is that correct?

Yes, you are correct. But to have it on the transaction log we write open and close signals into signal data collection. And with the default implementation we write the open ( low watermark) and then write the close (high watermark) for every chunk. With the `insert_delete` we just write the open and then remove it instead writing the close.

mfvitale: @roldanbob Jiri wrongly merged it. Please tell me if I answered to your doubt. I can address any changes with another PR.

roldanbob: Thanks, @mfvitale. Yes, you answered that question, but that raises another couple of questions for me.  First, it appears to me that the `incremental.snapshot.watermarking.strategy` property applies only when triggering an incremental snapshot by sending a SQL query to a signaling table vs. when using the Kafka signaling channel. I ask only because the description specifically refers to signals being written to the signaling data collection vs. to a Kafka topic.

Second, I want to clarify the origin of the open and close signals. From the user perspective, if I want to initiate an incremental snapshot, I submit a signal of type `execute-snapshot` to the signaling data collection. From the original description of the property it makes it sound as though the open` and `close` signals originate from the user. However, I assume Debezium generates these signals automatically to demarcate the snapshot window. after it receives the triggering`execute-snapshot` signal.  Is that right?

mfvitale: >  First, it appears to me that the `incremental.snapshot.watermarking.strategy` property applies only when triggering an incremental snapshot by sending a SQL query to a signaling table vs. when using the Kafka signaling channel. I ask only because the description specifically refers to signals being written to the signaling data collection vs. to a Kafka topic.

The property applies when the signal trigger an incremental snapshot that requires the `source` channel enabled since it is required for event de-duplication.  So it applies to every channel.

> Second, I want to clarify the origin of the open and close signals. From the user perspective, if I want to initiate an incremental snapshot, I submit a signal of type `execute-snapshot` to the signaling data collection. From the original description of the property it makes it sound as though the open`and`close` signals originate from the user. However, I assume Debezium generates these signals automatically to demarcate the snapshot window. after it receives the triggering`execute-snapshot` signal. Is that right?

Yes, `open` and `close` are generated by Debezium.

mfvitale: I would add that not only the close is no written but also that the open is deleted. Apart of that now it's fine.

mfvitale: > @mfvitale Dumb question maybe, but, for the `insert_delete` case, I don't understand the steps involved in deleting the entry for the signal that opens the snapshot window. Here's what I think happens:
> 
> 1. A user submits an  execute-snapshot signal to the signaling data collection.
> 2. At the opening of the snapshot windows, Debezium writes an entry to the signaling data collection to record the opening of the snapshot window.
> 3. The connector performs the snapshot.
> 4. The snapshot window closes.
> 5. Debezium generates a signal to the transaction log to mark the closing of the snapshot window. An entry to mark the close snapshot window event is propagated to the signaling data collection.
> 6. When the signaling data collection receives this close snapshot event entry, rather than adding the entry, it deletes the corresponding open snapshot event entry.
> 

Point 5 and 6 are true for the 'insert_insert' (old behavior). With the 'insert_delete' the step 5 will be the deletion of the relative open signal and then the 6 (closing the windows) is just done via code. 

> So, in the end, following a successful incremental snapshot, no records are present in the signaling data collection to mark the opening and closing of the snapshot window?

This is true with 'insert_delete', in the signal data table, only sent user sent signal will be present.

PR #5076, opened 2023-12-05
DBZ-7230 MySQL BIT Type should have a default length 1

Jiabao-Sun:
MySQL BIT Type should have a default length 1

https://dev.mysql.com/doc/refman/8.0/en/bit-type.html


Naros (approved):
Thanks for the contribution @Jiabao-Sun.

@jpechane this appears to be a safe change at this point in 2.5.  I checked the logic and when we don't set the length to `1` in the parser, the column is constructed with a default length of `-1`, and the `BIT` handler in `JdbcValueConverters` treats this as a `BOOLEAN` use case as well, so users who use schema registry should not experience any change in the output schema.

Naros (approved):
Thanks for the contribution @Jiabao-Sun.

@jpechane this appears to be a safe change at this point in 2.5.  I checked the logic and when we don't set the length to `1` in the parser, the column is constructed with a default length of `-1`, and the `BIT` handler in `JdbcValueConverters` treats this as a `BOOLEAN` use case as well, so users who use schema registry should not experience any change in the output schema.

leonardBang (approved):

jpechane:
@Jiabao-Sun Applied, thanks.
@Naros Thanks for checking. It really looks safe but I still marked the issue to be shown in breaking change just in case.

Jiabao-Sun:
Thanks @Naros and @jpechane for the quick review.

PR #5080, opened 2023-12-06
DBZ-7236 simple fix of inversed if logic

sherpa003:
https://github.com/debezium/debezium/pull/5055 introduced this unfortunate mistake


Naros (approved):
LGTM as long as tests pass.

PR #5081, opened 2023-12-06
DBZ-4321 New `PostProcessor` contract and Column Reselection

Naros:
https://issues.redhat.com/browse/DBZ-4321

Supersedes https://github.com/debezium/debezium/pull/5065

- [x] Implement Db2 
- [x] Implement Informix
- [x] Implement Vitess
- [x] Implement Spanner
- [x] Add hook in Task for MongoDB
- [x] Documentation


debezium-core/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
```diff
@@ -0,0 +1,240 @@
...
+    }
+
+    @Override
+    public void setDatabaseSchema(RelationalDatabaseSchema schema) {
+        this.schema = schema;
+    }
+
+    private List<String> getRequiredColumnSelections(TableId tableId, Struct after) {
+        final List<String> columnSelections = new ArrayList<>();
+        for (org.apache.kafka.connect.data.Field field : after.schema().fields()) {
```
jpechane, line 187: I think here the logic needs more thinking. I'd expect user to filter in/out columns depending whether they are interested in re-selection. SO maybe we need addtional config options and rename `columns.list` option?

debezium-core/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
```diff
@@ -0,0 +1,240 @@
...
+            final Object value = after.get(field);
+            if (isUnavailableValueHolder(field, value)) {
+                LOGGER.debug("Adding column {} for table {} to re-select list due to unavailable value placeholder.",
+                        field.name(), tableId);
+                columnSelections.add(field.name());
+            }
+            else {
+                final String fullyQualifiedName = jdbcConnection.getQualifiedTableName(tableId) + ":" + field.name();
```
jpechane, line 195:
```suggestion
                if (value == null) {
                    final String fullyQualifiedName = jdbcConnection.getQualifiedTableName(tableId) + ":" + field.name();
                    if (selector.test(fullyQualifiedName)) {
```

jpechane (commented):
@Naros LGTM, thanks! I left couple of minor comments. The main discussion will probably be around the injection of the deps into processors.

debezium-core/src/main/java/io/debezium/processors/spi/JdbcConnectionAware.java
```diff
@@ -0,0 +1,17 @@
...
+package io.debezium.processors.spi;
+
+import io.debezium.jdbc.JdbcConnection;
+
+/**
+ * Contract that allows injecting a {@link JdbcConnection}.
+ *
+ * @author Chris Cranford
+ */
+public interface JdbcConnectionAware {
```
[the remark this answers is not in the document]
Naros: Yep I actually just made this change to support MongoDB :)

debezium-core/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
Naros: In that case, would something as simple as `column.include.list` and `column.exclude.list` suffice?

debezium-core/src/main/java/io/debezium/processors/PostProcessorRegistry.java
```diff
@@ -0,0 +1,85 @@
...
+        return this.processors;
+    }
+
+    /**
+     * Set the optional dependencies needed by a post processor
+     *
+     * @param valueConverterProvider the value converter
+     * @param connection the JDBC connection
+     */
+    public void injectDependencies(ValueConverterProvider valueConverterProvider, JdbcConnection connection, RelationalDatabaseSchema schema,
```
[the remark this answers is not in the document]
jpechane: At the end I was thinking abbout intorducing good old service registry, WDYT?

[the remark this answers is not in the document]
Naros: I was thinking about that approach, but I wasn't sure if we wanted to go that far.  I would think it may make sense to try and have the registry as standalone as generic as possible so that not only post-processors but perhaps other bits (i.e. converters) could make use of such a mechanism too.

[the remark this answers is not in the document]
jpechane: Exactly!

debezium-core/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
jpechane: Yes, and the `column.list` should be renamed to something like `reselct.null.columns` or similar to be more descriptive.

Naros:
Let me explain my thinking with `BeanRegistry` and `ServiceRegistry` to provide some context before you review the changes @jpechane :grin: 

# BeanRegistry

So, this provides a simple object lookup concept based on a key.  I took some inspiration from Camel on this, as there may be future value in finding ways to integrate with CDI frameworks like Quarkus, Spring, etc.  The idea would be that all Debezium components would interact with our `BeanRegistry` for a universal contract. If a CDI provider were found, a primary registry would be injected into our `BeanRegistry`, and we would first resolve beans against that provider, falling back to our registry if not found. 

This could be useful for Debezium Server users who may want to grab a Quarkus bean in some Debezium component or some Spring bean if using Debezium Embedded Engine in a Spring application.  In fact, it could be utilized with any CDI provider we want to integrate with.

# ServiceRegistry (Service Locator)

So this introduces the Service Locator pattern with some simple dependency injection of services into other services. Any class in Debezium can implement the `Service` contract and define a `ServiceProvider` to be registered with the registry.  

When a Debezium connector requests a service, it will be constructed if it does not exist. This allows for complex, and straightforward construction with dependencies managed, as one service can request another and they'll be created in dependency order.  Additionally, services can also implement lifecycle callbacks to be configured, started, and stopped.  And finally, the registry manages an ordered list of the services so that during the shutdown, they'll be stopped in reverse order to avoid any issues with the dependency chain.

I ended up injecting the `BeanRegistry` as our first `Service` and I also added the new `PostProcessorRegistry` as the second service, showing the provider pattern and how that can likely help us subdivide our classes and code into more modular components.

I'll be interested in your thoughts.

debezium-core/src/main/java/io/debezium/processors/spi/JdbcConnectionAware.java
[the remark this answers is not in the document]
Naros: Removed this in factor of the `BeanRegistry`.

debezium-core/src/main/java/io/debezium/service/spi/Stoppable.java
```diff
@@ -0,0 +1,22 @@
...
+import io.debezium.common.annotation.Incubating;
+
+/**
+ * A contract that defines that the service wishes to be stopped during service shutdown.
+ * This is useful if the service has some resources that should be closed.
+ *
+ * @author Chris Cranford
+ */
+@Incubating
+public interface Stoppable {
```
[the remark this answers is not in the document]
Naros: Seems reasonable.

Naros:
@roldanbob, just an FYI that a whole new section is coming in the documentation called "Post Processors".  You can align these with transformations from a Kafka sense. Still, these operate within the Debezium runtime rather than as part of Kafka Connect, and allow access to more Debezium internals.  These are all incubating new features, but I wanted to give you a heads-up in case you had any questions.

Naros:
@jpechane I looked at Vitess and Spanner and overall I believe there the `ReselectColumnsPostProcessor` may not be applicable because they're basing their data stream with the source more like PostgreSQL's replication rather than the more traditional JDBC connection like what we have in our other relational connectors.

What I did do; however, was at least introduce the changes to spin up the `BeanRegistry` and the `ServiceRegistry`.

jpechane:
@Naros And it is done, thanks! There is just a test failure in Informix PR. When addressed please feel free to merge yourself.

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -0,0 +1,73 @@
...
+
+[NOTE]
+====
+This post-processor is supported only for the SQL database connectors.
+====
+
+== Overview
+
+There are use cases where a {prodname} connector may not include a specific column's value in the change event, at not fault of its own but rather that's simply how the source database works.
+For example, TOAST columns in PostgreSQL, LOB columns in Oracle or Extended String columns with Oracle Exadata all come to mind.
```
roldanbob, line 19:
```suggestion
For example, values for TOAST columns in PostgreSQL, LOB columns in Oracle, or Extended String columns in Oracle Exadata, might all be excluded.
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -0,0 +1,73 @@
...
+
+The `ReselectColumnsPostProcessor` provides a simple way to re-select one or more columns using an out-of-bands query to fetch the current column's state from the database table.
```
roldanbob, line 21:
```suggestion
The `ReselectColumnsPostProcessor` provides a way to re-select one or more columns from a database table and fetch the current state.
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -0,0 +1,73 @@
...
+The post processor can be configured to re-select columns that are `null`, columns that are emitted that use the configured `unavailable.value.placeholder` sentinel value.
```
roldanbob, line 22:
```suggestion
You can configure the post processor to re-select the following column types:

 * `null` columns. 
 * columns the contain the`unavailable.value.placeholder` sentinel value.
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -0,0 +1,73 @@
...
+
+== Configuration example
+
+Configure a `PostProcessor` much in the same way that you would configure a `CustomConverter` or `Transformation`.
```
roldanbob, line 26:
```suggestion
Configuring a `PostProcessor` is similar to configuring a `CustomConverter` or `Transformation`.
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -0,0 +1,73 @@
...
+  "reselector.type": "io.debezium.processors.reselect.ReselectColumnsPostProcessor", // <2>
+  "reselector.reselect.columns.include.list": "<schema_name>.<table_name>:<colA>,<schema_name>.<table_name>:<colB>", // <3>
+  "reselector.reselect.unavailable.values": "true", // <4>
+  "reselector.reselect.null.values": "true" // <5>
+----
+<1> Comma-separated list of post-processor prefixes
+<2> The fully-qualified class type name for the post-processor
+<3> Comma-separated list of columns using the format `<schema>.<table>:<column>`
+<4> Enables or disables the re-selection of columns that contain the `unavailable.value.placeholder` sentinel value
+<5> Enables or disables the re-selection of columns that are `null`
```
roldanbob, line 40:
```suggestion
<1> Comma-separated list of post-processor prefixes.
<2> The fully-qualified class type name for the post-processor.
<3> Comma-separated list of column names specified by using the following format: `<schema>.<table>:<column>`.
<4> Enables or disables the re-selection of columns that contain the `unavailable.value.placeholder` sentinel value.
<5> Enables or disables the re-selection of columns that are `null`.
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -0,0 +1,73 @@
...
+.Reselect columns post processor configuration options
+[cols="30%a,25%a,45%a"]
+|===
+|Property
+|Default
+|Description
+
+|[[reselect-columns-post-processor-property-reselect-columns-include-list]]<<reselect-columns-post-processor-property-reselect-columns-include-list, `+reselect.columns.include.list+`>>
+|No default
+|Comma-separated list of columns that should be re-selected from the source database, using the format `_<schema>_._<table>_:_<column>_` for each re-selectable column. +
```
roldanbob, line 55:
```suggestion
|Comma-separated list of column names to re-select from the source database.
Use the following format to specify column names: + `_<schema>_._<table>_:_<column>_`  +
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -0,0 +1,73 @@
...
+ +
+This is mutually exclusive with `reselect.columns.exclude.list`.
```
roldanbob, line 57:
```suggestion
Do not set this property if you set the `reselect.columns.exclude.list` property.
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -0,0 +1,73 @@
...
+
+|[[reselect-columns-post-processor-property-reselect-columns-exclude-list]]<<reselect-columns-post-processor-property-reselect-columns-exclude-list, `+reselect.columns.exclude.list+`>>
+|No default
+|Comma-separated list of columns that should not be re-selected from the source database, using the format `_<schema>_._<table>_:_<column>_`. +
```
roldanbob, line 61:
```suggestion
|Comma-separated list of column names in the source database to exclude from re-selection.
Use the following format to specify column names: + `_<schema>_._<table>_:_<column>_` +
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -0,0 +1,73 @@
...
+ +
+This is mutually exclusive with `reselect.columns.include.list`.
```
roldanbob, line 63:
```suggestion
Do not set this property if you set the `reselect.columns.include.list` property.
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -0,0 +1,73 @@
...
+
+|[[reselect-columns-post-processor-property-reselect-unavailable-values]]<<reselect-columns-post-processor-property-reselect-unavailable-values, `+reselect.unavailable.values+`>>
+|`true`
+|Specifies whether any column that matches the inclusion filter that has the value specified by the `unavailable.value.placeholder` connector configuration property will be re-selected.
```
roldanbob, line 67:
```suggestion
|Specifies whether the post processor reselects a column that matches the `reselect.columns.include.list` filter if the column value is provided by the connector's `unavailable.value.placeholder` property.
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -0,0 +1,73 @@
...
+
+|[[reselect-columns-post-processor-property-reselect-null-values]]<<reselect-columns-post-processor-property-reselect-null-values, `+reselect.null.values+`>>
+|`true`
+|Specifies whether any column that matches the inclusion filter that has a `null` value will be re-selected.
```
roldanbob, line 71:
```suggestion
|Specifies whether the post processor reselects a column that matches the `reselect.columns.include.list` filter if the column value is `null`.
```

roldanbob (changes requested):

Naros:
Applied to main (=2.5).

PR #5082, opened 2023-12-06
DBZ-7237 Align LOB behavior in snapshot and streaming

Naros:
https://issues.redhat.com/browse/DBZ-7237

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
@@ -214,11 +214,14 @@ public class OracleValueConverters extends JdbcValueConverters {
             case Types.NCHAR:
             case Types.NVARCHAR:
             case Types.STRUCT:
-            case Types.CLOB:
             case OracleTypes.ROWID:
                 return data -> convertString(column, fieldDefn, data);
+            case Types.CLOB:
+            case Types.NCLOB:
+            case Types.SQLXML:
+                return data -> convertClob(column, fieldDefn, data);
             case Types.BLOB:
-                return data -> convertBinary(column, fieldDefn, data, binaryMode);
+                return data -> convertBlob(column, fieldDefn, data, binaryMode);
             case OracleTypes.BINARY_FLOAT:
                 return data -> convertFloat(column, fieldDefn, data);
             case OracleTypes.BINARY_DOUBLE:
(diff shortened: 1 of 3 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
```diff
@@ -324,45 +337,61 @@ protected Object convertBinary(Column column, Field fieldDefn, Object data, Bina
...
-                else {
-                    Blob blob = (Blob) data;
-                    data = blob.getBytes(1, Long.valueOf(blob.length()).intValue());
-                }
-            }
             else if (data instanceof RAW) {
                 data = ((RAW) data).getBytes();
             }
-
-            if (data == UNAVAILABLE_VALUE) {
```
sherpa003, line 355: is this if statement not needed for Binary data type? I don't know, maybe not...

Naros: It was moved to the `convertBlob` method, so it was retained, just moved.

jpechane:
@Naros I really wonder if we should drop unavailable value placeholder. IMHO it is valid content in this context, WDYT?

Naros:
I believe the idea @jpechane was to basically remove any value related to these columns entirely when `lob.enabled` was not set in the configuration; however, I can see the point of maintaining that value as a clear indicator of that being the setting in the connector configs.  

I'd then say we need to tackle this in the reverse and when `lob.enabled` is set to `false` during the snapshot, we should explicitly include the unavailable value placeholder during snapshot to match the behavior of streaming.

@sherpa003 you okay with this, as long as the outcome from streaming/snapshot are aligned?

sherpa003:
> I believe the idea @jpechane was to basically remove any value related to these columns entirely when `lob.enabled` was not set in the configuration; however, I can see the point of maintaining that value as a clear indicator of that being the setting in the connector configs.
> 
> I'd then say we need to tackle this in the reverse and when `lob.enabled` is set to `false` during the snapshot, we should explicitly include the unavailable value placeholder during snapshot to match the behavior of streaming.
> 
> @sherpa003 you okay with this, as long as the outcome from streaming/snapshot are aligned?

Yes, the main goal is to have consistent behavior. So if lob.enabled=false, the placeholder value should be there also for short CLOB (treated as VARCHAR) values and for snapshot values.  Then it should be fine. Thank you

Naros:
Superseded by #5092.

PR #5083, opened 2023-12-07
DBZ-7235 Add config option to customize CloudEvent schema type name

rkudryashov:
https://issues.redhat.com/browse/DBZ-7235

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/converters/MongoDbCloudEventsMaker.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/converters/MongoDbCloudEventsMaker.java
@@ -17,8 +17,8 @@ import io.debezium.converters.spi.SerializerType;
  */
 public class MongoDbCloudEventsMaker extends CloudEventsMaker {
 
-    public MongoDbCloudEventsMaker(RecordParser parser, SerializerType contentType, String dataSchemaUriBase) {
-        super(parser, contentType, dataSchemaUriBase);
+    public MongoDbCloudEventsMaker(RecordParser parser, SerializerType contentType, String dataSchemaUriBase, String cloudEventsSchemaName) {
+        super(parser, contentType, dataSchemaUriBase, cloudEventsSchemaName);
     }
 
     @Override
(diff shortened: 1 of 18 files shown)


rkudryashov:
It seems that this is a breaking change, in a case when a user configure global setting `metadata.source` to get metadata from headers and `dataSchemaType` header is not provided

rkudryashov:
Added `schema.cloudevents.name` config [option](https://github.com/debezium/debezium/pull/5083/files#diff-6548dc2c3274efd87e5219b50cb365b2ef9cbe7831e0b9fde1e6a2be2ee0307eR44-R45), `null` by default, which means the current schema name generation algorithm will be used. Otherwise, the schema name will be that non-null value

It is more complex to preserve the current behavior - `io.debezium.connector.postgresql.Data` - while adding [config option](https://github.com/debezium/debezium/pull/5083/files#diff-6548dc2c3274efd87e5219b50cb365b2ef9cbe7831e0b9fde1e6a2be2ee0307eR49) for customization cloudEvents.data name. It cab be a problem in a case if someone has already configured `metadata.source` global setting to `header` which will be applied to the new `dataSchemaName`. It seems to me, it can't be not a breaking change without some additional property, for example, `metadata.source.dataSchemaName.header.enable`, `false` by default.

Also, the PR contains changes in CE provider/maker, that is it will require PRs for Db2/Vitess connectors

rkudryashov:
The other problem is that if `value.converter.data.serializer.type` is `json`, the `dataSchemaName` option is meaningless, because the schema won't be sent to a schema registry. But in this case, as well as in `avro` case, user will need to specify `dataSchemaName` to avoid error caused by the absence of the header.

We can avoid this problem if won't use global settings concept, but only concept of [default options](https://github.com/debezium/debezium/blob/b6924e6efcb6fecbb76bcaa5fe74ef66a15817e5/debezium-core/src/main/java/io/debezium/converters/CloudEventsConverterConfig.java#L45C38-L45C38). In the considered case, `metadata.source` can be `header`, by the default value for `metadata.source.dataSchemaName` will be `generate` and won't be overridden by the global setting

jpechane (approved):
@rkudryashov LGTM, thanks. Could you please prepare PRs for non-core connectors? We can merge it after that. Also could you please prepare a docs PR?

rkudryashov:
@jpechane 

https://github.com/debezium/debezium-connector-jdbc/pull/53
https://github.com/debezium/debezium-connector-db2/pull/125
https://github.com/debezium/debezium-connector-vitess/pull/173

jpechane:
Force push after rebase to the latest main

PR #5089, opened 2023-12-11
DBZ-7242 Add Grammar Oracle Truncate Cluster

ilyasahsan123:
https://issues.redhat.com/browse/DBZ-7242


Naros (approved):
Looks to be in the upstream as-is, thanks @ilyasahsan123 :heavy_check_mark:

PR #5092, opened 2023-12-11
DBZ-7237 Align LOB behavior in snapshot and streaming

Naros:
https://issues.redhat.com/browse/DBZ-7237

This reworks the original code in https://github.com/debezium/debezium/pull/5082 to be based on aligning snapshot output to streaming instead.


jpechane:
@Naros Applied, thanks!

PR #5095, opened 2023-12-12
DBZ-7251 Length value is not removed when changing a column's type

harveyyue:
https://issues.redhat.com/browse/DBZ-7251


jpechane:
@harveyyue Applied, thanks

PR #5097, opened 2023-12-12
DBZ-7252 Using correct offsets in table/collection notifications for MongoDB

jcechace:
https://issues.redhat.com/browse/DBZ-7252

@jpechane This also affect the notifications in upper level -- see https://issues.redhat.com/browse/DBZ-7253 However there I don't see any simple solution, so for now I'd recommend just documenting that the offset information may be incorrect in certain cases. 

//cc @mfvitale


mfvitale (approved):

jpechane:
@jcechace Applied, thanks!

PR #5103, opened 2023-12-14
DBZ-7264: Propagating exception when using multiple snapshotting threads

jcechace:
https://issues.redhat.com/browse/DBZ-7264


jcechace:
@jpechane if there is an error during snapshot, what is the expected behaviour? I'd say we set the producerThrowable and return with `snapshotResult.aborted();` no? Currently it looks that even tests are expecting that the snapshot will start

jpechane:
@jcechace Applied, thanks

PR #5110, opened 2023-12-18
DBZ-7272 Guarding against implicit offset invalidation 

jcechace:
https://issues.redhat.com/browse/DBZ-7272

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
@@ -573,6 +573,14 @@ public class MongoDbConnectorConfig extends CommonConnectorConfig {
             .withDescription("Internal use only")
             .withType(Type.LIST);
 
+    /**
+     * The {@link ReplicaSets#SEPARATOR}-separated list of connection strings
+     */
+    public static final Field ALLOW_OFFSET_INVALIDATION = Field.createInternal("mongodb.allow.offset.invalidation")
+            .withDescription("Allows offset invalidation when required by change of connection mode")
+            .withDefault(false)
+            .withType(Type.BOOLEAN);
+
     // MongoDb fields in Connection Group start from 1 (topic.prefix is 0)
     public static final Field CONNECTION_STRING = Field.create("mongodb.connection.string")
             .withDisplayName("Connection String")
@@ -931,6 +939,7 @@ public class MongoDbConnectorConfig extends CommonConnectorConfig {
                     TOPIC_PREFIX,
                     CONNECTION_STRING,
                     CONNECTION_MODE,
+                    ALLOW_OFFSET_INVALIDATION,
                     USER,
                     PASSWORD,
                     AUTH_SOURCE,
@@ -973,6 +982,7 @@ public class MongoDbConnectorConfig extends CommonConnectorConfig {
     private final CaptureScope captureScope;
     private final String captureTarget;
     private final ConnectionMode connectionMode;
+    private final boolean offsetInvalidationAllowed;
     private final int snapshotMaxThreads;
     private final int cursorMaxAwaitTimeMs;
     private final ReplicaSets replicaSets;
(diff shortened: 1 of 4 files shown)


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
```diff
@@ -573,6 +573,14 @@ public static OversizeHandlingMode parse(String value, String defaultValue) {
             .withDescription("Internal use only")
             .withType(Type.LIST);
 
+    /**
+     * The {@link ReplicaSets#SEPARATOR}-separated list of connection strings
+     */
+    public static final Field ALLOW_OFFSET_INVALIDATION = Field.createInternal("mongodb.internal.allow.offset.invalidation")
```
jpechane, line 579:
```suggestion
    public static final Field ALLOW_OFFSET_INVALIDATION = Field.createInternal("mongodb.allow.offset.invalidation")
```
It is already prefixed with `internal`

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorTask.java
```diff
@@ -196,10 +198,49 @@ private MongoDbOffsetContext getPreviousOffset(MongoDbConnectorConfig connectorC
...
+            logger.warn("Found at least one shard specific offset from previous run");
+            if (connectorConfig.isOffsetInvalidationAllowed()) {
+                logger.warn("Offset invalidation is allowed, previous shard specific offsets will be ignored");
+                return;
+            }
+
+            throw new DebeziumException("Found at least one shard specific offset from previous run." +
+                    "The default connection mode for sharded has changed to 'sharded' and previous offsets would be invalidated." +
+                    "Either explicitly set 'mongodb.connection.mode=replica_set' to postpone the migration " +
+                    "or set 'mongodb.internal.allow.offset.invalidation=true'. " +
```
jpechane, line 239:
```suggestion
                    "or set 'internal.mongodb.allow.offset.invalidation=true'. " +
```

jpechane: Please add an information that snapshot will be reexucted.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
jcechace: Hm... Then I think the following is not exactly correct either

```
public static final Field TASK_CONNECTION_STRINGS = Field.createInternal("mongodb.internal.task.connection.strings")
            .withDescription("Internal use only")
            .withType(Type.LIST);
```

BTW  wont the change result in the filed named `internal.mongodb.allow.offset.invalidation` then?

jcechace: `TASK_CONNECTION_STRINGS` is however truly internal and is of no concern to the user (as it is internally set). Additionally it's gone in 2.6 so we can safely ignore that :)

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorTask.java
```diff
@@ -196,10 +198,49 @@ private MongoDbOffsetContext getPreviousOffset(MongoDbConnectorConfig connectorC
...
+            logger.warn("Found at least one shard specific offset from previous run");
+            if (connectorConfig.isOffsetInvalidationAllowed()) {
+                logger.warn("Offset invalidation is allowed, previous shard specific offsets will be ignored and snapshot re-executed");
+                return;
+            }
+
+            throw new DebeziumException("Found at least one shard specific offset from previous run." +
+                    "The default connection mode for sharded has changed to 'sharded' and previous offsets would be invalidated." +
+                    "Either explicitly set '" + MongoDbConnectorConfig.CONNECTION_MODE.name() + "=replica_set' to postpone the migration " +
+                    "or set '" + MongoDbConnectorConfig.ALLOW_OFFSET_INVALIDATION.name() + "=true' to re-execute snapshot." +
```
jpechane, line 239:
```suggestion
                    "or set '" + MongoDbConnectorConfig.ALLOW_OFFSET_INVALIDATION.name() + "=true' to re-execute snapshot and reset offsets." +
```

jpechane:
@jcechace Applied, thanks

PR #5112, opened 2023-12-18
DBZ-7164 Support offset commit for the signal topic

vsantona:
Documentation: https://github.com/debezium/debezium/pull/5116

### DIFF
--- a/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/SignalsIT.java
+++ b/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/SignalsIT.java
@@ -1,0 +1,230 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package io.debezium.connector.mysql;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.io.File;
+import java.nio.file.Path;
+import java.sql.SQLException;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Function;
+
+import org.apache.kafka.clients.producer.KafkaProducer;
+import org.apache.kafka.clients.producer.ProducerConfig;
+import org.apache.kafka.clients.producer.ProducerRecord;
+import org.apache.kafka.common.serialization.StringSerializer;
+import org.junit.After;
+import org.junit.AfterClass;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
+
+import io.debezium.config.CommonConnectorConfig;
+import io.debezium.config.Configuration;
+import io.debezium.embedded.AbstractConnectorTest;
+import io.debezium.engine.DebeziumEngine;
+import io.debezium.junit.logging.LogInterceptor;
+import io.debezium.kafka.KafkaCluster;
+import io.debezium.pipeline.signal.actions.snapshotting.ExecuteSnapshot;
+import io.debezium.pipeline.signal.channels.KafkaSignalChannel;
+import io.debezium.util.Collect;
+import io.debezium.util.Testing;
+
+public class SignalsIT extends AbstractConnectorTest {
+    protected static final String SERVER_NAME = "is_test";
+    protected static final Path SCHEMA_HISTORY_PATH = Files.createTestingPath("file-schema-history-is.txt")
+            .toAbsolutePath();
+    protected final UniqueDatabase DATABASE = new UniqueDatabase(SERVER_NAME, "incremental_snapshot-test").withDbHistoryPath(SCHEMA_HISTORY_PATH);
+    protected static KafkaCluster kafka;
+
+    @BeforeClass
+    public static void startKafka() throws Exception {
+        File dataDir = Testing.Files.createTestingDirectory("signal_cluster");
+        Testing.Files.delete(dataDir);
+        kafka = new KafkaCluster().usingDirectory(dataDir)
+                .deleteDataPriorToStartup(true)
+                .deleteDataUponShutdown(true)
+                .addBrokers(1)
+                .withKafkaConfiguration(Collect.propertiesOf(
+                        "auto.create.topics.enable", "false",
+                        "zookeeper.session.timeout.ms", "20000"))
+                .startup();
+
+        kafka.createTopic("signals_topic", 1, 1);
+    }
+
+    @AfterClass
+    public static void stopKafka() {
+        if (kafka != null) {
+            kafka.shutdown();
+        }
+    }
+
+    @Before
+    public void before() throws SQLException {
+        stopConnector();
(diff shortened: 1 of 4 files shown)


debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/SignalsIT.java
```diff
@@ -0,0 +1,185 @@
...
+                .with(MySqlConnectorConfig.INCLUDE_SCHEMA_CHANGES, false)
+                .with(MySqlConnectorConfig.SIGNAL_DATA_COLLECTION, DATABASE.qualifiedTableName("debezium_signal"))
+                .with(CommonConnectorConfig.SIGNAL_POLL_INTERVAL_MS, 1)
+                .with(MySqlConnectorConfig.INCREMENTAL_SNAPSHOT_CHUNK_SIZE, 10)
+                .with(MySqlConnectorConfig.INCREMENTAL_SNAPSHOT_ALLOW_SCHEMA_CHANGES, true)
+                .with(CommonConnectorConfig.SCHEMA_NAME_ADJUSTMENT_MODE, CommonConnectorConfig.SchemaNameAdjustmentMode.AVRO);
+    }
+
+    @Test
+    public void givenOffsetCommitDisabledAndASignalSentWithConnectorRunning_whenConnectorComesBackUp_thenAllSignalsAreCorrectlyProcessed()
```
mfvitale, line 102: I would like to see also a test with the old behavior: signal sent during shutdown will not be processed on startup.

debezium-core/src/main/java/io/debezium/pipeline/signal/channels/KafkaSignalChannel.java
```diff
@@ -163,15 +172,20 @@ public void reset(Object reference) {
...
                 .withDefault(ConsumerConfig.GROUP_ID_CONFIG, signalConfig.getString(GROUP_ID))
                 .withDefault(ConsumerConfig.FETCH_MIN_BYTES_CONFIG, 1) // get even the smallest message
                 .withDefault(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, false)
                 .withDefault(ConsumerConfig.SESSION_TIMEOUT_MS_CONFIG, 10000) // readjusted since 0.10.1.0
                 .withDefault(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class)
-                .withDefault(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class)
+                .withDefault(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG, StringDeserializer.class);
+        if (signalConfig.getBoolean(SIGNAL_CONSUMER_OFFSET_COMMIT_ENABLED)) {
+            confBuilder.withDefault(ConsumerConfig.ENABLE_AUTO_COMMIT_CONFIG, true)
+                    .withDefault(ConsumerConfig.AUTO_OFFSET_RESET_CONFIG, "earliest");
```
mfvitale, line 186: Is this enough? Should this act only when there is no initial offset in Kafka or if an offset is out of range? What about the restart after some processed records? Should be an explicit commit or auto commit present?

mfvitale:
@vsantona thanks! I leaved couple of comments.

debezium-core/src/main/java/io/debezium/pipeline/signal/channels/KafkaSignalChannel.java
mfvitale: Sorry, just saw the auto commit setting. Don't mind mu comment.

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/SignalsIT.java
vsantona: done

mfvitale (approved):

mfvitale:
@vsantona Applied, thanks a lot!

debezium-core/src/main/java/io/debezium/pipeline/signal/channels/KafkaSignalChannel.java
rk3rn3r, line 186: I randomly hit this PR and I was interested why you picked "earliest" here? The most probably case for not having an offset is either on a newly created connector or `kafka.group.id` or a purposely renamed connector `kafka.group.id` or consumer offsets . That might also be re-using an previously used group id accidentally.
Should in this case really all signals that existed before be re-processed?
(the code now looks a little different, but the property is still set to `earliest`)

mfvitale: You can read the reason behind this in [DBZ-7164](https://issues.redhat.com/browse/DBZ-7164)

**TL;DR**

The main problem was that when a connector is stopped if you send signals during this period them was not processed once the connector starts this because there was neither autocommit nor commit. 

Now the code is different since we have the auto-commit enabled by default and we just fallback to `earliest` in case of no offset.

rk3rn3r: Thank you. Yeah, that's a good explanation about `what` happened and I saw it from the ticket and the conversation on chat.
But let me ask again the last question from above: should in this case really all signals that existed before being re-processed aka `earliest`? This is also the case when starting a new "instance" of the kafka signal channel by renaming the consumer group, etc. I get that this can be (experienced) user's intention, but this can also be a source of error for less experienced users. The bus factor of `latest` is much lower. Maybe this could be configurable option and defaults to `latest`? Wdyt?
FYI @jpechane @Naros

Naros: But @rk3rn3r, if you deploy a random sink connector in Kafka Connect, when there are no offsets, that connector defaults to `earliest` unless you explicitly override the default behavior. So I would say what we are doing maps directly to the expected behavior for any other sink deployment. The expectation is that if the user needs a different strategy, they must override that.

rk3rn3r: That's not fully accurate. A Kafka consumer group (which is used for that behavior in source connectors) defaults to that. And for source connectors it makes sense.

I think signals have a different idea on how they work: you might not want to consume all signals every created (that are probably already tombstoned later in a stream and not yet removed, for example), but you normally want to read all data of a database (respecting several filter options etc).

And my take was not about the technical process, but the significant higher bus factor when `earliest` was used, but not meant. There is only a way with additional kafka property overwriting for the consumer config options to switch to `latest`. A dbz config option would make this more visibile.

mfvitale: @rk3rn3r If you still have some doubt I suggest to open a discussion on zulip commenting the current code. Since the code in this PR doesn't fully reflect the current code.

PR #5114, opened 2023-12-19
DBZ-7276 Grant DELETE on signalling table in the incremental snapshot…

vjuranek:
… tests

Delete is needed for `insert_delete` watermarking strategy.

https://issues.redhat.com/browse/DBZ-7276


jpechane:
@vjuranek Applied, thanks

PR #5125, opened 2023-12-20
DBZ-7284 Provide config option to customize CloudEvents.data schema name

rkudryashov:
To actually retrieve data schema name from the `dataSchemaName` header, a user should specify two properties:

```
    "value.converter.metadata.source": "header",
    "value.converter.schema.data.name.source.header.enable": true,
```

The second property by default is `false` which provides backward compatibility

If it is Outbox SMT + CloudEventsConverter integration, the header can be added as follows:

```
    "transforms.outbox.table.fields.additional.placement": "type:header,aggregate_type:header:dataSchemaName",
```


jpechane:
@rkudryashov Applied, thanks! COuld you please prepare a docs PR too?

PR #5130, opened 2023-12-26
DBZ-7288 Unittest for hasCommitAlreadyBeenHandled in CommitScn Class

ilyasahsan123:
https://issues.redhat.com/browse/DBZ-7288


Naros (approved):

ilyasahsan123:
hi @Naros, the CI job is failed due to timeout issue. Is it intermittent?

image

Naros:
Sometimes they will do that @ilyasahsan123 but for this particular use case, it's unrelated.   The `CommitScn` class is only used by the LogMiner adapter and since those tasks passed, we're good to merge.  Thanks for the contributions :+1:

PR #5131, opened 2024-01-02
DBZ-7266 Use column name instead of regular id for partitioning clause

ani-sha:
https://issues.redhat.com/browse/DBZ-7266


jpechane:
@ani-sha Applied, thanks.

PR #5133, opened 2024-01-02
DBZ-7297: Implement AWS IAM Authent in debezium-connector-postgres

bpaquet:


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
```diff
@@ -67,13 +67,13 @@ public class PostgresConnection extends JdbcConnection {
 
     private static final Pattern FUNCTION_DEFAULT_PATTERN = Pattern.compile("^[(]?[A-Za-z0-9_.]+\\((?:.+(?:, ?.+)*)?\\)");
     private static final Pattern EXPRESSION_DEFAULT_PATTERN = Pattern.compile("\\(+(?:.+(?:[+ - * / < > = ~ ! @ # % ^ & | ` ?] ?.+)+)+\\)");
+    public static final String DEFAULT_PROTOCOL = "postgresql";
+    public static final String DEFAULT_DRIVER_CLASS_NAME = org.postgresql.Driver.class.getName();
     private static Logger LOGGER = LoggerFactory.getLogger(PostgresConnection.class);
 
-    private static final String URL_PATTERN = "jdbc:postgresql://${" + JdbcConfiguration.HOSTNAME + "}:${"
+    private static final String URL_PATTERN_WITHOUT_PROTOCOL = "${" + JdbcConfiguration.HOSTNAME + "}:${"
```
jpechane, line 74: Is not it possible to pass the protocol as pattern placeholder?

bpaquet: I need to change the JDBCConfiguration to do this, which is widely used, so it's simpler to me to do like this :)

jpechane:
@bpaquet I think we can sill salvage the PR even if it will not cover all the situations.

I'd recommend to add an integration test that will be using the JDBC URL passed as an config option.

Then after the merge we can think about merging Oracle, MySQL and PostgreSQL config options and allow all relational connectors to pass arbitrary JDBC connection string.

jakeyheath:
This looks great, we also have a need to leverage IAM authentication to our RDS postgres. Any updates on the status of this being merged in?

bpaquet:
Hello, I'm stucked because IAM authent is not supported by Aurora, only by
standard RDS. I'm not able to test.

https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.IAMDBAuth.html#UsingWithRDS.IAMDBAuth.Limitations

Are you using Aurora or RDS?

jakeyheath:
Ah, we are use RDS Aurora as well. This is the line you are referring to? 

> For Aurora PostgreSQL, you cannot use IAM authentication to establish a replication connection.

bpaquet:
Yes :(

sandpwr:
I'm also looking for this support. Since RDS Aurora PotgreSQL does support IAM now, any chance this can be prioritized?

bpaquet:
This is still impossible, according to documentation:

https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.IAMDBAuth.html#UsingWithRDS.IAMDBAuth.Limitations


   -

   For Aurora PostgreSQL, you cannot use IAM authentication to establish a
   replication connection.

sandpwr:
Oh I see, thanks for pointing that out.

notaninterestingusername:
Hello, @bpaquet!

I couldn't find the point you cited in the [documentation](https://docs.aws.amazon.com/AmazonRDS/latest/AuroraUserGuide/UsingWithRDS.IAMDBAuth.html#UsingWithRDS.IAMDBAuth.Limitations). Not sure if I'm looking in the right place? 
> For Aurora PostgreSQL, you cannot use IAM authentication to establish a
   replication connection

I was able to get a CDC workflow going successfully by supplying the auth token in the `database.password` field. However, having the token be periodically refreshed would be super helpful, particularly in cases where the token expires before the connector can be fully setup. Any thoughts? Can this be implemented now?

bpaquet:
Seems iam auth is finally supporting replication. I will have a look soon,
hopefully next week

notaninterestingusername:
@bpaquet, awesome, thank you!

bpaquet:
Hello @jpechane,

Almost 2 year later, AWS implemented the support of replication connection with IAM.

Here is a version which works on my machine with Aurora Postgresql 15.13.

I had two solutions for implementation:
* add multiple parameters to allow the user to tune the JDBC url to activate AWS JDBC Wrapper
* add a single parameter which configure everything (`aws_iam_authentication`)
I clearly chose the second way. Let me know if it seems okay to you.

In addition of overriding the `ConnectionFactory`, I also had to replace some cast by `unwrap`. 
I'm not sure which tests I can add.
Current tests seems green, let me know what you think. 

To use it, you have to add the AWS JDBC Wrapper (https://github.com/aws/aws-advanced-jdbc-wrapper) in your classpath.

I can invest some time to try to add it in MySQL, but I will not use it in production.

jpechane:
@bpaquet Thanks for all the effort. There is `io.debezium.jdbc.JdbcConfiguration.CONNECTION_FACTORY_CLASS` config option. Do you think that it is possible to isolate/implement the functionality as dedicated connection factory class?

bpaquet:
Yes, it's possible, but that will require to
* Accept duplication between patternFactory
* Refactor the existing factory to allow subclassing.

jpechane:
@bpaquet I think solution similar to
```java
public class PostgresAwsIamConnectionFactory implements JdbcConnection.ConnectionFactory {

    private static final String URL_PATTERN = "...";
    protected static final ConnectionFactory FACTORY = JdbcConnection.patternBasedFactory(URL_PATTERN,
            ""software.amazon.jdbc.Driver",
            PostgresConnection.class.getClassLoader(),
            JdbcConfiguration.PORT.withDefault(PostgresConnectorConfig.PORT.defaultValueAsString()));

    @Override
    public Connection connect(JdbcConfiguration config) throws SQLException {
        return FACTORY.connect(config);
    }
}
```

is quite acceptable and you'd just configure `connection.factory.class` config option.

bpaquet:
@jpechane done.

bpaquet:
I can also update the doc if we are okay with the code.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresAwsIamConnectionFactory.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ * @author Bertrand Paquet
+ */
+public class PostgresAwsIamConnectionFactory implements JdbcConnection.ConnectionFactory {
+    private static final String AWS_IAM_URL_PATTERN = "jdbc:aws-wrapper:postgresql://${" + JdbcConfiguration.HOSTNAME + "}:${"
+            + JdbcConfiguration.PORT + "}/${" + JdbcConfiguration.DATABASE + "}?wrapperPlugins=iam";
+
+    private static final JdbcConnection.ConnectionFactory FACTORY = JdbcConnection.patternBasedFactory(
+            AWS_IAM_URL_PATTERN,
+            "software.amazon.jdbc.Driver",
```
jpechane: @bpaquet Sorry for not highlighting it before, this should go into a dedicated String constant.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
```diff
@@ -83,9 +83,14 @@ public class PostgresConnection extends JdbcConnection {
...
-    protected static final ConnectionFactory FACTORY = JdbcConnection.patternBasedFactory(URL_PATTERN,
+
+    protected static final ConnectionFactory defaultConnectionFactory = JdbcConnection.patternBasedFactory(
+            URL_PATTERN,
             org.postgresql.Driver.class.getName(),
-            PostgresConnection.class.getClassLoader(), JdbcConfiguration.PORT.withDefault(PostgresConnectorConfig.PORT.defaultValueAsString()));
+            PostgresConnection.class.getClassLoader(),
+            JdbcConfiguration.PORT.withDefault(PostgresConnectorConfig.PORT.defaultValueAsString()));
+
+    protected static ConnectionFactory FACTORY = new ConnectionFactoryDecorator(defaultConnectionFactory);
```
jpechane: Is this really necessary? IMHO the decorator is added at `JdbcConnection` class so I believe there is no need for any changes in `PostgresConnection` class at all.

debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -122,11 +122,11 @@ public interface ConnectionFactory {
         Connection connect(JdbcConfiguration config) throws SQLException;
     }
 
-    private class ConnectionFactoryDecorator implements ConnectionFactory {
+    public static class ConnectionFactoryDecorator implements ConnectionFactory {
```
jpechane: The change in visibility might be unnecessary if `POstgresConnection` changes could be avoided.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
bpaquet: Hmm, you are correct, we can do without.
Pushed

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresAwsIamConnectionFactory.java
bpaquet: Done

debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
bpaquet: Removed

jpechane:
@bpaquet Applied, thanks!

PR #5136, opened 2024-01-03
[docs] Remove redundant --rm flag

overwatcheddude:
Deleted the extra --rm flag in Docker and Podman commands.


jpechane:
@overwatcheddude Applied, thanks. Could you please share our name so we can add you to the contributor list?

overwatcheddude:
> @overwatcheddude Applied, thanks. Could you please share our name so we can add you to the contributor list?

حمود سمبول

PR #5137, opened 2024-01-04
DBZ-7313: Oracle Infinispan abandoned trasactions minor enhancements

sherpa003:
https://issues.redhat.com/browse/DBZ-7313

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
@@ -1427,7 +1427,7 @@ public abstract class AbstractLogMinerEventProcessor<T extends AbstractTransacti
                                 if (first) {
                                     LOGGER.warn("All transactions with SCN <= {} will be abandoned.", thresholdScn);
                                     if (LOGGER.isDebugEnabled()) {
-                                        try (Stream s = getTransactionCache().keySet().stream()) {
+                                        try (Stream<String> s = getTransactionCache().keySet().stream()) {
                                             LOGGER.debug("List of transactions in the cache before transactions being abandoned: [{}]",
                                                     s.collect(Collectors.joining(",")));
                                         }
@@ -1452,7 +1452,7 @@ public abstract class AbstractLogMinerEventProcessor<T extends AbstractTransacti
                         }
                     }
                     if (LOGGER.isDebugEnabled()) {
-                        try (Stream s = getTransactionCache().keySet().stream()) {
+                        try (Stream<String> s = getTransactionCache().keySet().stream()) {
                             LOGGER.debug("List of transactions in the cache after transactions being abandoned: [{}]",
                                     s.collect(Collectors.joining(",")));
                         }
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
@@ -1426,9 +1426,11 @@ public abstract class AbstractLogMinerEventProcessor<T extends AbstractTransacti
                             if (entry.getValue().getStartScn().compareTo(thresholdScn) <= 0) {
                                 if (first) {
                                     LOGGER.warn("All transactions with SCN <= {} will be abandoned.", thresholdScn);
-                                    if(LOGGER.isDebugEnabled()) {
-                                        LOGGER.debug("List of transactions in the cache before transactions being abandoned: [{}]",
-                                                getTransactionCache().keySet().stream().collect(Collectors.joining(",")));
+                                    if (LOGGER.isDebugEnabled()) {
+                                        try (Stream s = getTransactionCache().keySet().stream()) {
+                                            LOGGER.debug("List of transactions in the cache before transactions being abandoned: [{}]",
+                                                    s.collect(Collectors.joining(",")));
+                                        }
                                     }
                                     first = false;
                                 }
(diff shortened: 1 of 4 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
```diff
@@ -1419,24 +1420,35 @@ public void abandonTransactions(Duration retention) throws InterruptedException
...
-                                LOGGER.warn("All transactions with SCN <= {} will be abandoned.", thresholdScn);
-                                first = false;
+                    try {
+                        while (iterator.hasNext()) {
+                            Map.Entry<String, T> entry = iterator.next();
+                            if (entry.getValue().getStartScn().compareTo(thresholdScn) <= 0) {
+                                if (first) {
+                                    LOGGER.warn("All transactions with SCN <= {} will be abandoned.", thresholdScn);
+                                    LOGGER.debug("List of transactions in the cache before transactions being abandoned: [{}]",
+                                            getTransactionCache().keySet().stream().collect(Collectors.joining(",")));
```
Naros, line 1430: Guard this with `if(LOGGER.isDebugEnabled())` to avoid the expensive keySet collection if debugging isn't enabled.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
```diff
@@ -1419,24 +1420,35 @@ public void abandonTransactions(Duration retention) throws InterruptedException
...
-                            metrics.setActiveTransactionCount(getTransactionCache().size());
                         }
                     }
+                    finally {
+                        if (iterator instanceof CloseableIterator) {
+                            ((CloseableIterator<Map.Entry<String, T>>) iterator).close();
+                        }
+                    }
+                    LOGGER.debug("List of transactions in the cache after transactions being abandoned: [{}]",
+                            getTransactionCache().keySet().stream().collect(Collectors.joining(",")));
```
Naros, line 1451: Same here, guard this with `if(LOGGER.isDebugEnabled())`.

Naros (commented):
Just a minor inline request.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
```diff
@@ -1419,24 +1420,39 @@ public void abandonTransactions(Duration retention) throws InterruptedException
...
-                                first = false;
+                    try {
+                        while (iterator.hasNext()) {
+                            Map.Entry<String, T> entry = iterator.next();
+                            if (entry.getValue().getStartScn().compareTo(thresholdScn) <= 0) {
+                                if (first) {
+                                    LOGGER.warn("All transactions with SCN <= {} will be abandoned.", thresholdScn);
+                                    if(LOGGER.isDebugEnabled()) {
+                                        LOGGER.debug("List of transactions in the cache before transactions being abandoned: [{}]",
+                                                getTransactionCache().keySet().stream().collect(Collectors.joining(",")));
```
jeremy-l-ford, line 1431: To be safe, the stream should closed here.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
```diff
@@ -1419,24 +1420,39 @@ public void abandonTransactions(Duration retention) throws InterruptedException
...
                         }
                     }
+                    finally {
+                        if (iterator instanceof CloseableIterator) {
+                            ((CloseableIterator<Map.Entry<String, T>>) iterator).close();
+                        }
+                    }
+                    if(LOGGER.isDebugEnabled()) {
+                        LOGGER.debug("List of transactions in the cache after transactions being abandoned: [{}]",
+                                getTransactionCache().keySet().stream().collect(Collectors.joining(",")));
```
jeremy-l-ford, line 1454: Stream needs to be closed here as well.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
sherpa003: good catch

sherpa003:
btw, why these debugs logs are there... With ISPN, I'm getting this error:
```Caused by: java.lang.IllegalStateException: Error reading header from 604464:808543792 | 0
    at org.infinispan.persistence.sifs.IndexNode$LeafNode.getHeaderAndKey(IndexNode.java:1144)``` 
When this happens in separate thread, Debezium tries to remove transaction from the list backed by ISPN. However that does not return error neither removes it... So it stays there and eventually is reported as abandoned. 
Then this code tries to remove it after processing abandoned transaction but fails the same way.

Naros:
> btw, why these debugs logs are there... With ISPN, I'm getting this error:
> 
> ```
> Caused by: java.lang.IllegalStateException: Error reading header from 604464:808543792 | 0
>     at org.infinispan.persistence.sifs.IndexNode$LeafNode.getHeaderAndKey(IndexNode.java:1144)
> ```

No problem, I don't mind the log entries themselves, it's just that if the log entry is going to be discarded because the user has not enabled DEBUG logging, we should guard these `LOGGER.debug` calls if the arguments perform some step that can be seen as expensive and unnecessary as they'll be evaluated by Java only for the call to be skipped when the logger API is called.

> When this happens in separate thread, Debezium tries to remove transaction from the list backed by ISPN.  However that does not return error neither removes it... So it stays there and eventually is reported as abandoned. Then this code tries to remove it after processing abandoned transaction but fails the same way.

So, to be clear, the separate thread scraping that we discussed with William that you have, the above error is what you observed.  And with the code in this PR, you are observing that same failure with these changes?  Is that accurate?

sherpa003:
> > btw, why these debugs logs are there... With ISPN, I'm getting this error:
> > ```
> > Caused by: java.lang.IllegalStateException: Error reading header from 604464:808543792 | 0
> >     at org.infinispan.persistence.sifs.IndexNode$LeafNode.getHeaderAndKey(IndexNode.java:1144)
> > ```
> 
> No problem, I don't mind the log entries themselves, it's just that if the log entry is going to be discarded because the user has not enabled DEBUG logging, we should guard these `LOGGER.debug` calls if the arguments perform some step that can be seen as expensive and unnecessary as they'll be evaluated by Java only for the call to be skipped when the logger API is called.

Understood and this is already resolved.

> > When this happens in separate thread, Debezium tries to remove transaction from the list backed by ISPN.  However that does not return error neither removes it... So it stays there and eventually is reported as abandoned. Then this code tries to remove it after processing abandoned transaction but fails the same way.
> 
> So, to be clear, the separate thread scraping that we discussed with William that you have, the above error is what you observed. And with the code in this PR, you are observing that same failure with these changes? Is that accurate?

This PR has no effect on this ISPN bug. It just helped to uncover that once ISPN exception happens (within ISPN thread so main Debezium thread does not know about it), the removal from collection does not do what it should. Sharing this only as a side info.

Anyway, so can we merge this?

Naros:
I made one small change to use parameterized types rather than the raw-types.  The only issue I don't particularly care for is the use of the Infinispan bits in this class but we can address refactoring this later since they'll be more changes in the future to address other off-heap implementations.

PR #5138, opened 2024-01-04
DBZ-7311 Permits to execute a blocking snapshot even if snapshot.mode=never

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7311

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/source/AbstractSnapshotChangeEventSource.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/source/AbstractSnapshotChangeEventSource.java
@@ -68,7 +68,7 @@ public abstract class AbstractSnapshotChangeEventSource<P extends Partition, O e
 
         final SnapshotContext<P, O> ctx;
         try {
-            ctx = prepare(partition);
+            ctx = prepare(partition, snapshottingTask.isBlocking());
         }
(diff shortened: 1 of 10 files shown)


debezium-core/src/main/java/io/debezium/pipeline/source/AbstractSnapshotChangeEventSource.java
```diff
@@ -178,7 +178,7 @@ protected abstract SnapshotResult<O> doExecute(ChangeEventSourceContext context,
     /**
      * Prepares the taking of a snapshot and returns an initial {@link SnapshotContext}.
      */
-    protected abstract SnapshotContext<P, O> prepare(P partition) throws Exception;
+    protected abstract SnapshotContext<P, O> prepare(P partition, boolean blocking) throws Exception;
```
jpechane, line 181: @mfvitale I think the parameter here and elswhere should be renamed to either `adHoc` or `onDemand` to better reflect the situation. The initial snapshot is always blocking in the sense that streaming it not running.

mfvitale: Good point.

jpechane:
@mfvitale Applied, thanks!

PR #5139, opened 2024-01-04
DBZ-7098 Converting engine should honor unsupported tombstone flag

jpechane:
https://issues.redhat.com/browse/DBZ-7098

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/ConvertingEngineBuilder.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/ConvertingEngineBuilder.java
@@ -95,34 +95,53 @@ public class ConvertingEngineBuilder<R> implements Builder<R> {
         return format1 == format2;
     }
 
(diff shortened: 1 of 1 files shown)


ArthurLERAY:
Hi @jpechane I tried to pull and build from the related branch but I'm having a hard time building debezium. The error I'm getting is `Failed to execute goal com.github.os72:protoc-jar-maven-plugin:3.8.0:run (default) on project debezium-connector-postgres: Error extracting protoc for version 3.8.0: Unsupported platform: protoc-3.8.0-osx-aarch_64.exe`.

First, I was using Java 19 but I was getting the error `Unsupported class file major version 65` so I switched to Java 21.
I'm trying to build debezium using `mvn clean install -DskipITs -DskipTests`

I'm on an ARM M1 Macbook with:
- `javac 21.0.1`
- `java 21.0.1 2023-10-17 LTS`
- `Apache Maven 3.9.6`

It seems that the 3.8.0 protobuf does not support M1 chips yet

PR #5140, opened 2024-01-04
DBZ-7315 Fix Oracle include/exclude lists with spaces

Naros:
https://issues.redhat.com/browse/DBZ-7315


jpechane:
@Naros Applied, thanks

PR #655, opened 2018-10-06
DBZ-780 | Adding RENAME column semantics to rename a column

vamossagar12:

PR #5147, opened 2024-01-10
DBZ-7342 Respect the max error retries setting

shybovycha:
Break the loop of retries after the max number of retries has been reached.


debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
```diff
@@ -673,9 +673,10 @@ else if (maxRetries < EmbeddedEngineConfig.DEFAULT_ERROR_MAX_RETRIES) {
                     startedSuccessfully = true;
                 }
                 catch (Exception ex) {
-                    if (totalRetries == maxRetries) {
+                    if (maxRetries > 0 && totalRetries >= maxRetries) {
```
vjuranek, line 676: I don't think we have to adjust this condition, it should be fine. We just need to exit from the loop, so just adding the `break` should be sufficient.

vjuranek (changes requested):
Thanks for the PR! I left a suggestion not to make the condition more complicated and minimize the fix.

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
shybovycha: the thing is: there is a documented behavior for [`errors.max.retries`](https://debezium.io/documentation/reference/stable/development/engine.html#engine-properties) when `maxRetries` is set to `-1` meaning we should _never_ stop retrying:

> The maximum number of retries on connection errors before failing (-1 = no limit, 0 = disabled, > 0 = num of retries).

vjuranek: Right, but when `maxRetries` is `-1`, you start with `totalRetries = 0` and in each iteration increment `totalRetries`, thus `totalRetries == maxRetries` condition is never met and you effectively never stop retrying. So, unless I'm missing something, condition `totalRetries == maxRetries` is fine as it is now, we just need to escape from the loop (i.e. add `break` statement) when `totalRetries == maxRetries` is reached.

shybovycha: as per the documentation, there are three scenarios to consider:

1. `maxRetries = -1` - should keep retrying (`break` should never be triggered)
2. `maxRetries = 0` - should exit on first error (`break` should be triggered on first retry)
3. `maxRetries > 0` - should exit after `maxRetries`

if we only check for `totalRetries == maxRetries` and we start with `totalRetries = 0`, on the first failure we will have `totalRetries == 1` and for each scenario:

1. the condition won't be ever met, will never `break` ✅ 
2. the condition won't be met on the first failure, will not `break` but should 🔴 
3. the condition will be met at some point, will `break` then ✅ 

so at the very least we should check for `maxRetries > 0` on top of `totalRetries == maxRetries`. having `>=` there is more of a safeguard, so agree, that bit could be left unchanged. but `maxRetries > 0` is essential to support the documented behaviour.

vjuranek: @shybovycha if I'm not mistaken, the case `maxRetries == 0` is handled in the different branch 
```
if (maxRetries == 0) {
    retryError = e;
}
```
 [currently on the line #655](https://github.com/debezium/debezium/blob/main/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java#L655), so we don't need to handle this case here.

shybovycha: yeah, you're right - all that `retryError` jazz got me confused. i guess it could make the code easier to understand if we have early return instead.

vjuranek (approved):
Yes, return is IMHO fine, LGTM now. Thanks  a lot @shybovycha !
Once CI is green or I'm sure the failures are random, I'll merge it.

vjuranek:
The failures don't seem to be random, but are not related to this PR.

vjuranek:
Merged, thanks @shybovycha !

PR #5148, opened 2024-01-10
DBZ-7022 Allow conversion of source block timestamp fields in TZ SMT

ani-sha:
https://issues.redhat.com/browse/DBZ-7022

### DIFF
--- a/debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
@@ -94,9 +94,9 @@ public class TimezoneConverter<R extends ConnectRecord<R>> implements Transforma
     private List<String> excludeList;
     private static final String SOURCE = "source";
     private static final String TOPIC = "topic";
-    private static final String FIELD_SOURCE_PREFIX = "source";
-    private static final String FIELD_BEFORE_PREFIX = "before";
-    private static final String FIELD_AFTER_PREFIX = "after";
+    private static final String FIELD_SOURCE_PREFIX = "source.";
+    private static final String FIELD_BEFORE_PREFIX = "before.";
+    private static final String FIELD_AFTER_PREFIX = "after.";
     private static final Pattern TIMEZONE_OFFSET_PATTERN = Pattern.compile("^[+-]\\d{2}:\\d{2}(:\\d{2})?$");
     private static final Pattern LIST_PATTERN = Pattern.compile("^\\[(source|topic|[\".\\w\\s_]+):([\".\\w\\s_]+(?::[\".\\w\\s_]+)?(?:,|]$))+$");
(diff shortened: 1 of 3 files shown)


debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -307,12 +308,38 @@ private void handleStructs(Struct value, Type type, String matchName, Set<String
 
         Struct before = getStruct(value, Envelope.FieldName.BEFORE);
         Struct after = getStruct(value, Envelope.FieldName.AFTER);
+        Struct source = getStruct(value, Envelope.FieldName.SOURCE);
+
+        Set<String> beforeFields = new HashSet<>();
+        Set<String> afterFields = new HashSet<>();
+        Set<String> sourceFields = new HashSet<>();
+
+        if (!fields.isEmpty() && !fields.contains(null)) {
```
Naros, line 317: In what use case would the `Set<String>` fields contain a `null` entry rather than simply being empty?

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -307,12 +308,38 @@ private void handleStructs(Struct value, Type type, String matchName, Set<String
...
+            for (String field : fields) {
+                if (field.startsWith(FIELD_SOURCE_PREFIX)) {
```
jpechane, line 319: I think you should match for `source.` includeing the dot, not only `source` as it would conflict with allowed field names.
Same for before/after below.

jpechane (changes requested):
@ani-sha Thanks, I left one comment that relates to matching. Also could you please comment on @Naros question? Thanks

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
ani-sha: @Naros Sorry I missed this comment. This check is added from here while parsing each of the include/exclude items. https://github.com/debezium/debezium/blob/c2eac1714eaa560817a00e83ba215dce1a0e9bca/debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java#L400

I have added proper null check for that.

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -94,6 +94,9 @@ public class TimezoneConverter<R extends ConnectRecord<R>> implements Transforma
     private List<String> excludeList;
     private static final String SOURCE = "source";
     private static final String TOPIC = "topic";
+    private static final String FIELD_SOURCE_PREFIX = "source.";
```
jpechane, line 97: Could you please take the constants from `io.debezium.data.Envelope.FieldName`?

PR #5149, opened 2024-01-11
[docs] Mention the need to include primary column in case of sql sele…

vjuranek:
…rv column include list


jpechane:
@vjuranek Thanks!

PR #5152, opened 2024-01-11
DBZ-7312 On demand blocking snapshot will not set transaction snapshot

mfvitale:
closes: [DBZ-7312](https://issues.redhat.com/browse/DBZ-7312)


jpechane:
@mfvitale Applied thanks

PR #5153, opened 2024-01-12
[docs] Update Podman commands in the tutorial

overwatcheddude:
I removed the sudo prefix from the podman commands for the following reasons:
1. There's no need to run podman commands with sudo. It is unnecessary.
2. Pods created with sudo won't appear in Podman Desktop.
3. Maintain consistency with the Docker commands in the tutorial (they are not prefixed with sudo).

When `dbz` pod is created, only port 8083 will be published. This is because the host will only interact with Kafka Connect (8083) via curl. The host will not interact with Apache Zookeeper (9092) or MySQL (3306), thus opening these two ports in the pod is unnecessary.


jpechane:
@overwatcheddude Applied, thanks

PR #5156, opened 2024-01-13
DBZ-7347 Initial snapshot notifications should use full identifier

wukachn:
Proposed solution to the feature request seen in DBZ-7347. Could potentially be done in a nicer way with an interface api change but can easily get around that for now.

https://issues.redhat.com/browse/DBZ-7347

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -575,3 +575,4 @@ Don Seiler
 Ahmed Rachid Hazourli
 Pavithrananda Prabhu
 حمود سمبول
+Peter Hamer
--- a/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/NotificationsIT.java
+++ b/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/NotificationsIT.java
@@ -9,6 +9,7 @@ package io.debezium.connector.mysql;
 import java.nio.file.Path;
 import java.sql.SQLException;
 import java.util.List;
+import java.util.stream.Collectors;
 
 import org.junit.After;
 import org.junit.Before;
@@ -43,7 +44,8 @@ public class NotificationsIT extends AbstractNotificationsIT<MySqlConnector> {
     }
 
     protected List<String> collections() {
-        return List.of("a", "b", "c", "a4", "a42", "a_dt", "a_date", "debezium_signal");
+        return List.of("a", "b", "c", "a4",
+                "a42", "a_dt", "a_date", "debezium_signal").stream().map(tbl -> String.format("%s.%s", DATABASE.getDatabaseName(), tbl)).collect(Collectors.toList());
     }
 
     @Override
--- a/debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/NotificationsIT.java
+++ b/debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/NotificationsIT.java
@@ -39,7 +39,7 @@ public class NotificationsIT extends AbstractNotificationsIT<OracleConnector> {
     }
 
     protected List<String> collections() {
-        return List.of("A");
+        return List.of("ORCLPDB1.DEBEZIUM.A");
     }
 
     @Override
--- a/debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/NotificationsIT.java
+++ b/debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/NotificationsIT.java
@@ -12,6 +12,7 @@ import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
+import java.util.stream.Collectors;
 
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.source.SourceRecord;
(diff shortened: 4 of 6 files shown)


debezium-core/src/test/java/io/debezium/pipeline/notification/IncrementalSnapshotNotificationServiceTest.java
```diff
@@ -131,7 +131,7 @@ public void notifyTableScanCompleted() {
         Notification expectedNotification = new Notification("12345", "Incremental Snapshot", "TABLE_SCAN_COMPLETED", Map.of(
                 "connector_name", "connector-test",
                 "data_collections", "db.inventory.product,db.inventory.customer",
-                "scanned_collection", "db.inventory.product",
+                "scanned_collection", "\"db\".\"inventory\".\"product\"",
```
mfvitale, line 134: WDYT to maintain the same format as for 'data_collections" field? Maybe here quoting is not strictly required.

mfvitale:
Hi @wukachn , thanks for the proposal. I just leaved a single comment.

debezium-core/src/test/java/io/debezium/pipeline/notification/IncrementalSnapshotNotificationServiceTest.java
wukachn: I did have the same thought. Ideally, debezium would use a double quoted identifier across all notifications, metrics etc. This removes ambiguity for the (rare) case that an identifier part includes a `.`

That could maybe be a different ticket if others agreed.

Im happy to go with a regular identifier for now. The main purpose of this ticket was to use the whole identifier for all table scan notifications.

mfvitale: This is a good point. What is not clear to me is that you have reported 
>You cannot infer which table the notification is referring to when you have 2 tables with the same name in different schemas e.g. sch1.tbl and sch2.tbl. Both will produce a table scan completed notification with the 
"scanned_collection" of "tbl".

But the current implementation with the 'identifier' method already returns the full table name. Isn't it? Also the test states this. 

So is the problem just the quoting? Can you please clarify?

wukachn: I need to update the ticket. Currently, an incremental snapshot uses the complete table identifier but for an initial snapshot only the table name is given.

At the time of ticket creation, I assumed this was a design decision but it appears to be a bug.

This PR would fix that initial snapshot bug, then add double quotes to remove some ambiguity to both types of snapshot.

jpechane: Thinking about that, should we use double quotes? This seems to be database dependent as some of the database use double qutes, then there is for example SQL server that uses `[schema].[table]` format at db level.
We have `io.debezium.jdbc.JdbcConnection.quotedTableIdString(TableId)` that handles this nuance. Maybe we should introduce a spearate per connector class that would be repsonsible for such differences (frequently taking them ofo JdbcConnection) but also get rid of `useCatalogBeforeSchema` `io.debezium.relational.TableId` etc. WDYT?

wukachn: My thoughts are that we can use double quotes for all db's here. In my eyes this is more about wrapping the parts to remove ambiguity, than trying to keep inline with the db level syntax. 

We could use a different character to wrap the parts but I feel like this should be independent of the db.

Again, if we want to remove the ambiguity from all notifications, metrics etc. Im happy for that to be its own ticket. The key issues in this PR is resolving the initial snapshot bug, ensuring we use a full identifier instead of table name.

mfvitale: I would propose to just solve the real bug with this PR and maybe open a further enhancement about the quoting to better discuss it. For example another solution can be to have a more structured data ( JSON Object) with `database`, `schema`, `table`.  

In that case you can change this issue type from `Feature Request` to `Bug`.

wukachn: Ill do that @mfvitale.

wukachn: https://issues.redhat.com/browse/DBZ-7350

mfvitale:
@wukachn Can you please fix the tests? Failures of `notificationCorrectlySentOnItsTopic` and `notificationCorrectlySentOnJmx` are related to the change. Thanks.

wukachn:
Need to update oracle/NotificationsIT but I don't have experience with how debezium builds its tableIds or running oracle ITs.

mfvitale:
> Need to update oracle/NotificationsIT but I don't have experience with how debezium builds its tableIds or running oracle ITs.

@wukachn Adding `"ORCLPDB1.DEBEZIUM.A"` to `collections` method of Oracle `NotificationsIT` test should fix it.

jpechane:
@mfvitale LGTM, please feel free to merge when its tested.

mfvitale:
Failures seems not related.

mfvitale:
@wukachn Applied, thanks!

PR #5157, opened 2024-01-15
DBZ-7300 Refactor Snapshotter interface to be an SPI through SnapshotterProvider.

mfvitale:
Closes: https://issues.redhat.com/browse/DBZ-7300


mfvitale:
Just a draft PR for discussing on the proposal. 

The current 'Snapshotter' interface define different things:

- When to create snapshot/streaming 
- How to do snapshot

The proposal aims to split the behaviors into different interface resulting in:

- `Snapshotter`:  define when the snapshot should be done or not and if continue with streaming or not. 
- `SnapshotterProvider`: SPI to provide custom `Snapshotter` implementations. The snapshotter will be choosed using the `snapshot.mode` configuration property. The `custom` option will be removed. 
- `SnapshotQuery`: define how data are queried from the database. Custom implementation will be provided specifying the implementing class through the `snapshot.query.custom.class` and loaded through the class loader. As done before for the `Snapshotter` interface.
- `SnapshotLock`: define the table lock mode used during schema snapshot. Custom implementation will be provided specifying the implementing class through the `snapshot.locking.custom.class` and loaded through the class loader.

One the proposal is accepted this PR will provide the implementation for PostgreSQL.

jpechane:
@mfvitale Hi, I really like the approach. I've two major questions

1. Do we need separate interfaces for querying and locking? IMHO that's something that can be on a single one
2. What is the advantage of having `SnapshotterProvider` and don't use service loader directly?

mfvitale:
> 1. Do we need separate interfaces for querying and locking? IMHO that's something that can be on a single one
What if a user want just customize the query and not the locking? Having a single interface the user is forced to implement also the lock part. 
> 2. What is the advantage of having `SnapshotterProvider` and don't use service loader directly?
In that way we can avoid to have the `init` method in the interface to pass all the useful dependencies, instead they are passed in  the `create(Configuration config);` and then let the implementation have the constructor.

Naros:
> In that way we can avoid to have the `init` method in the interface to pass all the useful dependencies, instead they are passed in the `create(Configuration config);` and then let the implementation have the constructor.

I would also suggest considering how you can explore using the `Service` marker contract in this context.  In this regard, you may even be able to avoid the `Configuration` entirely and instead rely on the `ServiceRegistry` to inject the `BeanRegistry` which could pass the connector's pre-created configuration object.  This would allow us to centralize all the configuration handling in this connector config object for parsing, etc.

mfvitale:
> > In that way we can avoid to have the `init` method in the interface to pass all the useful dependencies, instead they are passed in the `create(Configuration config);` and then let the implementation have the constructor.
> 
> I would also suggest considering how you can explore using the `Service` marker contract in this context. In this regard, you may even be able to avoid the `Configuration` entirely and instead rely on the `ServiceRegistry` to inject the `BeanRegistry` which could pass the connector's pre-created configuration object. This would allow us to centralize all the configuration handling in this connector config object for parsing, etc.

@Naros, good point. What is the difference between the `ServiceRegistry` and the `BeanRegistry`?

Naros:
> @Naros, good point. What is the difference between the `ServiceRegistry` and the `BeanRegistry`?

So, `ServiceRegistry` implements a service locator and dependency injection-style solution. You ask the registry for a given service by class type; whatever service is registered by that type will be returned. If no one has asked for that service but it has an initiator (registration) present, the service will be created on the fly before being returned. This helps simplify things like start/stop handlers as services are constructed in the order requested and torn down in the reverse order on connector shutdown. Additionally, a service can annotate a method with `@InjectService` to have other services dependency injected. I hope that over time, we can move legacy code to services and that new features can be designed as services.

The `BeanRegistry` itself is one service, and it's designed to simply provide a key/value store of beans by some name. This way, rather than having a constructor with a dozen arguments, we can either (a) pass the registry or (b) utilize the service locator pattern to inject the registry into the class directly to begin to clean up some of these very large ctor argument lists.

debezium-api/src/main/java/io/debezium/spi/snapshot/Snapshotter.java
```diff
@@ -0,0 +1,76 @@
...
+ * can provide more advanced functionality, such as partial snapshots
+ *
+ * Implementor's must return true for either {@link #shouldSnapshot()} or {@link #shouldStream()}
+ * or true for both.
+ */
+@Incubating
+public interface Snapshotter {
+
+    /**
+     * @return true if the snapshotter should take a snapshot
```
vjuranek, line 29: *{@code true}* instead of *true*?

debezium-api/src/main/java/io/debezium/spi/snapshot/Snapshotter.java
```diff
@@ -0,0 +1,76 @@
...
+ * Namely:
+ * - Should a snapshot occur at all
+ * - Should streaming occur
+ * - What queries should be used to snapshot
+ *
+ * While many default snapshot modes are provided with debezium (see documentation for details)
+ * a custom implementation of this interface can be provided by the implementor which
+ * can provide more advanced functionality, such as partial snapshots
+ *
+ * Implementor's must return true for either {@link #shouldSnapshot()} or {@link #shouldStream()}
```
vjuranek, line 22: *Implementor's* -> *Implementors*

debezium-api/src/main/java/io/debezium/spi/snapshot/SnapshotterProvider.java
```diff
@@ -0,0 +1,29 @@
...
+public interface SnapshotterProvider {
+
+    /**
+     * Create a map of snapshotter where the key is its name used in 'snapshot.mode' configuration.
+     *
+     * @param Configuration the connector configuration
+     *
+     * @return a map of custom snapshotter
+     */
+    Map<String, Snapshotter> create(Configuration config); // Can we move the Configuration interface from core module?
```
vjuranek, line 28: I'd be very careful with moving anything which includes Kafka stuff into the api package. Especially when it's even not a part of Kafka API package (e.g. `connect-api`). I'd prefer to keep it int core package for now.

vjuranek (commented):
I have a few nitpick comments, but LGTM. The only thing I not sure about it moving `Configuration` (with its Kafka dependency) into api package.

Edit: reading the comments, I forgot about service registry, so +1 to @Naros - would be nice to try to use it here.

vjuranek:
> > 1. Do we need separate interfaces for querying and locking? IMHO that's something that can be on a single one
> 
> What if a user want just customize the query and not the locking? With a single interface the user is forced to implement also the lock part.

Not sure how much it's likely, but I can also imagine users want to use various combinations of locking mode and snapshot query. With separate interfaces you can easily create whatever combination you like.

debezium-api/src/main/java/io/debezium/spi/snapshot/SnapshotterProvider.java
mfvitale: This `Configuration` is `io.debezium.config.Configuration` so is not Kafka related. 

In any case I'm evaluating the Chris approach and maybe in that case we don't need it.

vjuranek: > This `Configuration` is `io.debezium.config.Configuration` so is not Kafka related.

Yes, but if you look into the `io.debezium.config.Configuration` class [here](io.debezium.config.Configuration), the class itself contains dependency on Kafka `org.apache.kafka.common.config.ConfigValue`, so by moving `io.debezium.config.Configuration` into api module, the module will start to depend on Kafka, which I'm not sure we want it (at least personally I don't like it). 

> In any case I'm evaluating the Chris approach and maybe in that case we don't need it.

yes, this would be IMHO nice example where this dependency mechanism would be very useful - we can avoid loading user configuration here and thus avoid various kind of dependencies (which is always good, no matter if I like it or not:-)

mfvitale:
Wondering if `SnapshotQuery` is not a a duplicate of [snapshot.select.statement.overrides](https://debezium.io/documentation/reference/2.5/connectors/postgresql.html#postgresql-property-snapshot-select-statement-overrides).

This is the way the snapshotter query is currently used:
https://github.com/debezium/debezium/blob/c1b7e68319ee56c176ee87cd2c3a3509f708f23c/debezium-core/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java#L650C5-L668C6

It think `snapshot.select.statement.overrides' is more flexible and easy to use. @jpechane WDYT?

mfvitale:
I reviewed the implementation to use the Debezium `Service` and `ServiceProvider`. All the interfaces are load via Java SPI and then registered in our `DefaultServiceRegistry`.

Currently the provided implementations are the one available for PostgreSQL and with the subsequents PRs  the goal is to put all the possible implementation inside the `debezium-core` module. To provide custom implementation, for example for `Snapshotter` interface,  users can implement the interface and the set the `snapshot.mode` property to `custom` and the provide its name through `snapshot.mode.custom.name`. Same approach is used for `SnapshotQuery` and `SnapshotLock`.

debezium-core/src/main/java/io/debezium/bean/StandardBeanNames.java
```diff
@@ -42,4 +43,9 @@ public interface StandardBeanNames {
     // all connector-specific connection details across all connectors (including MongoDB),
     // and then replace this name key as CONNECTION_FACTORY regardless of the connector.
     String JDBC_CONNECTION = JdbcConnection.class.getName();
```
jcechace, line 45: If this is supposed to include Mongo, can we use something like `DatabaseConnection` as a name? IMHO even when documented this is going to be very unintuitive .

mfvitale: Actually this is only for relation databases as of now.

Naros: I think that's reasonable @jcechace, although I'd like to find time to explore the comment above that to see if there is some type of unification we can do across MongoDB and Relational connectors.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSnapshotChangeEventSource.java
```diff
@@ -242,20 +244,34 @@ protected void aborted(SnapshotContext<PostgresPartition, PostgresOffsetContext>
...
+             * the same snapshot from the existing exported transaction as for the initial snapshot.
+             */
+            String snapSet = String.format("SET TRANSACTION SNAPSHOT '%s';", newSlotInfo.snapshotName());
+            return "SET TRANSACTION ISOLATION LEVEL REPEATABLE READ; \n" + snapSet;
+        }
+
+        // TODO should this customizable?
+
+        // we're using the same isolation level that pg_backup uses
+        return "SET TRANSACTION ISOLATION LEVEL SERIALIZABLE, READ ONLY, DEFERRABLE;";
```
mfvitale, line 272: @jpechane Should we let customize it? This seems to be valid only for PostgresSQL and MySQL.

jpechane: No, let's keep it per connector now. We can always reconsider in future

PR #5158, opened 2024-01-15
DBZ-5071 Improve incremental snapshot query performance for PostgreSQL

james-johnston-thumbtack:
This branch builds on the suggested code changes made by @jpechane and the query changes suggested by myself in [DBZ-5071](https://issues.redhat.com/browse/DBZ-5071) to improve the incremental snapshot query performance for PostgreSQL by using row value constructors.

@jpechane and I also noted and discussed special considerations around NULL handling, and one thing I realized is that the connector doesn't do well with rows that have NULL value in their keys today.  The connector might crash, or it might simply omit them entirely.

This branch therefore has three main commits:

- Refactor some functions into the suggested ChunkQueryBuilder interface.
- Fix all the erroneous NULL handling by making the default chunk query builder NULL-aware.
- Create RowValueConstructorChunkQueryBuilder and use it for PostgreSQL.

Each is described in more detail in the respective commit message.  The changes are dependent upon each other and thus in one branch.  If the Debezium maintainers prefer, I could split this into three sequential pull requests instead (let me know).  The second commit in particular is arguably scope-creeping DBZ-5071, itself an extension of the scope-creeping discussion in DBZ-5071 about NULL handling.

It was suggested in the JIRA issue to also apply this for MySQL.  However, that is counterproductive because they don't [support taking advantage of them in the query planner](https://bugs.mysql.com/bug.php?id=111952).  In fact, with some [simple experimentation](https://onecompiler.com/mysql/3zzdsudzx) I found that MySQL is already using optimal queries using the existing queries generated by the default chunk query builder.  Therefore, I suggest we not mess with MySQL further unless there are complaints about performance that can be traced to poor query plans.  The original complaint in DBZ-5071 was about PostgreSQL.  The good news is that with this refactor, it's easy to update MySQL connector in the future to use row value constructors if/when MySQL itself can utilize them properly.

Test results on a table with a ~few million rows (subset of real production data), run on a local PostgreSQL 13 instance, local Kafka broker, local Connect instance:

- Old incremental scanning code: durations:
    - 1st run: 0.51s user 0.26s system 0% cpu 1:50.70 total (cold Connect restart)
    - 2nd run: 0.48s user 0.20s system 0% cpu 1:45.67 total
    - 3rd run: 0.50s user 0.20s system 0% cpu 1:45.64 total
- New incremental scanning code: durations:
    - 1st run: 0.60s user 0.28s system 1% cpu 1:13.49 total (cold Connect restart)
    - 2nd run: 0.48s user 0.18s system 1% cpu 1:02.84 total
    - 3rd run: 0.49s user 0.16s system 1% cpu 1:02.47 total
- Ad-hoc blocking snapshot: durations:
    - 1st run: 0.44s user 0.17s system 1% cpu 58.627 total
    - 2nd run: 0.43s user 0.17s system 1% cpu 60.000 total
    - 3rd run: 0.42s user 0.16s system 0% cpu 58.959 total

I noticed that the incremental and ad-hoc blocking snapshot times were suspiciously close together.  `docker stats` reports that PG only uses minimal CPU in the last two cases, and Connect is clearly CPU-bound.  Spot checking with VisualVM and I found multiple hotspots and potential areas for improvement, but that's another story.  The point is: the time savings with this pull request is probably even greater than the numbers above, because now it's Connect's CPU usage limiting performance instead of PostgreSQL query performance.


james-johnston-thumbtack:
The MongoDB 5.0 failure (`MultiThreadIncrementalSnapshotIT#pauseDuringSnapshot`) seems like an unrelated flaky test to me.  I tried running it myself and the test passes: `debezium/debezium-connector-mongodb$ ../mvnw clean verify -DskipTests -DskipITs=false -Dit.test=io.debezium.connector.mongodb.MultiThreadIncrementalSnapshotIT#pauseDuringSnapshot -Dversion.mongo.server=5.0`  (verified it was MongoDb 5.0 by checking Docker logs while test ran)

The MySQL percona failure (`SnapshotParallelSourceIT#shouldCreateSnapshotOfSingleDatabaseWithoutGlobalLock`) also I suspect to be unrelated flaky.  I similarly couldn't reproduce it locally /w percona docker container running.

SQL Server failures look to be a similar situation:
```
2024-01-15T14:40:06.6588694Z [ERROR]   IncrementalSnapshotWithRecompileIT.stopCurrentIncrementalSnapshotWithAllCollectionsAndTakeNewNewIncrementalSnapshotAfterRestart:176->AbstractIncrementalSnapshotTest.stopCurrentIncrementalSnapshotWithAllCollectionsAndTakeNewNewIncrementalSnapshotAfterRestart:596->AbstractSnapshotTest.startConnector:288->AbstractSnapshotTest.startConnector:279->AbstractConnectorTest.assertNoRecordsToConsume:983 expected:<[tru]e> but was:<[fals]e>
2024-01-15T14:40:06.6593135Z [ERROR]   IncrementalSnapshotWithRecompileIT>AbstractIncrementalSnapshotTest.stopCurrentIncrementalSnapshotWithoutCollectionsAndTakeNewNewIncrementalSnapshotAfterRestart:541->AbstractSnapshotTest.startConnector:288->AbstractSnapshotTest.startConnector:279->AbstractConnectorTest.assertNoRecordsToConsume:983 expected:<[tru]e> but was:<[fals]e>
2024-01-15T14:40:06.6596091Z [ERROR]   SqlServerReselectColumnsProcessorIT.afterEach:41->AbstractReselectProcessorTest.afterEach:67->AbstractConnectorTest.assertNoRecordsToConsume:983 expected:<[tru]e> but was:<[fals]e>
```
The first one is explicitly tagged Flaky, but is simply a passthrough that calls into the super class, which is not designated flaky (yet also failed).  The last one is not designated flaky, but also appears unrelated to this branch.  Given that the tests are either marked flaky or seem unrelated, I'm inclined to suspect none of the failures have anything to do with this PR.

For spanner, this test failed seemed unrelated: `io.debezium.connector.spanner.kafka.internal.BufferedPublisherTest.testBufferedPublisher_2`

james-johnston-thumbtack:
**The Db2 failure, however, is a compilation failure and raises some question how best to proceed**.  Would appreciate some guidance here, especially since that connector is in a separate repo.  The issue arises from the second commit in this branch that improves NULL handling.  The following changes in that commit seem notable that could cause test failures in derived connectors:
- The `AbstractIncrementalSnapshotTest` class had two abstract functions that provide table name, and two tests added to it related to testing incremental snapshots with NULL keys.  This was motivated due to the copy/pasted code between PostgreSQL and MySQL, and I didn't want to further copy/paste into Oracle and SQL Server.  I also didn't want to leave those connectors with a gap in test coverage (which turned out to be needed).  Derived classes have to create table schema for the test (if they weren't already), set up message key column connector setting, and override those abstract functions.  But now the Db2 connector fails because it derives its own incremental snapshot IT from this abstract class.
- The `JdbcConnector` has a new `nullsSortLast` function which is used by the default chunk builder to build the NULL-aware chunking queries.  Although the function is not abstract, it does return a tri-state `Optional<Boolean>` and the base class returns empty, meaning "I don't know" (there's no reasonable default to assume here).  If this is empty value is found by the query chunk builder, _and_ a key column is nullable, then the query chunk builder now fails fast with an error.  This is arguably an improvement over older Debezium versions which, while they didn't fail, would _silently omit records from the snapshot_ if there were NULL values.  In this PR, I override that function in every connector (PG, MySQL, Oracle, SQL Server) with the correct value based on the vendor's documentation.  But connectors outside this repo would also need updating.

One open question in my mind is what level of compatibility needs to be guaranteed.  For example, how stable are the test cases in the `AbstractIncrementalSnapshotTest` really intended to be / are they set in stone or can we make changes like I'm proposing to do here?

I am happy to try making the improvements to the Db2 (and any other) connectors as well, but would appreciate some feedback on the best way to proceed before I spend more time on it.  Here's one idea:

1.  Split this PR into three or more.  I feel more like this PR might be biting off more than it can chew.
2.  The first PR merges the QueryChunkBuilder refactor, because it's relatively straightforward, yet also touches a number of different places that might be more likely to result in a merge conflict as time passes.
3.  Other connector repos are prepared by ensuring the integration tests create tables with NOT NULL columns.  If they had duplicated the code from PG/MySQL that is being abstracted into the base test class, then those tests are deleted.  In practice we figure out what changes to make here by observing the GitHub Actions workflow failures with the next step, below.
4.  The second PR is the same as the "Correctly handle NULL values" commit from this PR, except that the new functions and tests in `AbstractIncrementalSnapshotTest` _are not abstract_.  Instead, a new (temporary) function is also added that allows derived classes to opt-in to the new nullability tests.  The default behavior is to opt-out.  The tests will be no-ops in such an opt-out case.
5.  The other connector repos are then modified to override the new `nullsSortLast` function in `JdbcConnector`.  The tests are also updated to opt-in to the new nullable key tests in the base class by overriding the appropriate functions.
6.  If we want, at this point, we can clean up the `AbstractIncrementalSnapshotTest` a bit by making the functions abstract as currently being proposed in this PR, rather than returning placeholder/invalid defaults.  The "opt-in/opt-out" function is changed to return "opt-in" by default.  This is merged, and then the connectors are updated to remove the override of the opt-in/out function.  Finally, the opt-in/out function is deleted entirely.
7.  After all this, merge the 3rd PR setting up the RowValueConstructorQueryChunkBuilder; this should also be straightforward since the PG connector is in the same repo as core packages.

A variation of the idea could also use the opportunity to make `JdbcConnection` and the new `nullsSortLast` both abstract, if we wanted to force derived classes to provide this information at compile-time, rather than waiting until runtime to fail.  I guess it comes down to: what compatibility guarantee needs to be made here...?

jpechane:
@james-johnston-thumbtack Thank you very much for the PR. Generally it LGTM but I need to take a look in more detail.

WRT how to continue, I don't think it is necessary to split the PR. Yes, it is a bit large but it is quite easy to understand. In relation to Db2 - I think it is just enough to create a PR in Db2 repo that will use the branch with the same name. In that case Db2 PR will be built and tested against this PR. Just please be aware that DB2 CI is a bit unstable in relation to incremental snapshots so please make sure you run it locally and report the results from local run too so we'll avoid back and forth pinging.

james-johnston-thumbtack:
Hi @jpechane - that sounds much simpler to proceed!  I prepared a PR as recommended for the Db2 connector: https://github.com/debezium/debezium-connector-db2/pull/131.  The incremental tests pass when run locally, as reported in my comment there.

jpechane:
@james-johnston-thumbtack I am OOO till end of the week but I'll be able to finally review the PR at the beginning of the next one. Thanks for the patience

debezium-core/src/test/java/io/debezium/pipeline/source/snapshot/incremental/SignalBasedSnapshotChangeEventSourceTest.java
```diff
@@ -1,299 +0,0 @@
-/*
```
jpechane, line 1: Are you sure the test hav nothing in common? IMHO it might make sense to have either 3 test classes or two test classes with a common parent with shared tests.

jpechane (commented):
@james-johnston-thumbtack I happy with the result, thanks a lot! Could you please just take a look at the comment about tests and rebase on the latest `main`? When done the PR can be merged immediatelly. Also a PR against Informix is probably needed.

james-johnston-thumbtack:
@jpechane here's the Informix PR: https://github.com/debezium/debezium-connector-informix/pull/30

Everything is rebased against today's main, and it looks like the test results for this PR are passing apart from expected compilation errors for Db2 and Informix since those PRs aren't merged yet.

jpechane:

jpechane:
@JordanP Thanks for verification

@james-johnston-thumbtack Great job!

james-johnston-thumbtack:
Great to hear it's already working for someone!  We have yet to deploy this version in production, but I'm looking forward to it!

PR #5160, opened 2024-01-15
DBZ-7352 Upgrade to Quarkus 3.6.5

Naros:
https://issues.redhat.com/browse/DBZ-7352


jpechane:
@Naros Applied, thanks

PR #5163, opened 2024-01-16
DBZ-7316 Send heartbeats while searching for WAL position

james-johnston-thumbtack:
If the logical replication slot has absolutely zero events in it, then the searchWalPosition could be looping for quite some time.  During this time, it wasn't sending heartbeats.

This commit fixes that function to send heartbeats.


jpechane:

PR #5165, opened 2024-01-16
DBZ-7355 Skip Avro serialization validation

vjuranek:
https://issues.redhat.com/browse/DBZ-7355


jpechane:
@vjuranek Applied, thanks

PR #5166, opened 2024-01-16
DBZ-7358 Use relational table primary key by default

Naros:
https://issues.redhat.com/browse/DBZ-7358

This fix uses the relational table primary key by default; however, as a user can define `message.key.columns` to even override the primary key configuration or to handle keyless tables, the user can override the `reselect.use.event.key` option with `true` so use the event key fields instead for the re-select so that keyless tables can also participate with the column reselection process.


jpechane:
@Naros Applied, for Alpha1 release thanks! Could ou just please prepare a follow-up PR with a test?

PR #5170, opened 2024-01-18
[docs] incremental snapshot: blocking snapshot are also supported

JordanP:
PR https://github.com/debezium/debezium/pull/4721 documented blocking snapshot added in [DBZ-6566](https://issues.redhat.com/browse/DBZ-6566). But it missed a
few spots.

Also, for consistency use "snapshot type" instead of "snapshot kind".

### DIFF
--- a/documentation/modules/ROOT/partials/modules/all-connectors/proc-stopping-an-incremental-snapshot-nosql.adoc
+++ b/documentation/modules/ROOT/partials/modules/all-connectors/proc-stopping-an-incremental-snapshot-nosql.adoc
@@ -63,7 +63,7 @@ If this component of the `data` field is omitted, the signal stops the entire in
 
 |5
 |`incremental`
-|A required component of the `data` field of a signal that specifies the kind of snapshot operation that is to be stopped. +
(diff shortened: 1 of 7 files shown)


JordanP:
There are some more places with `Currently, the only valid option is the default value, incremental`:
 * an-incremental-snapshot-nosql.adoc
 * an-incremental-snapshot.adoc

I only did the `-sql.adoc` variant because I am not sure the other variants support blocking snapshot. If so, please lemme know and I will update the doc there too.

mfvitale:
Hi @JordanP thanks for the contribution. Just few notes:

> I only did the `-sql.adoc` variant because I am not sure the other variants support blocking snapshot. If so, please lemme know and I will update the doc there too.\

Yes, it is supported by all connectors.

Just note that the stop/pause/resume is only available for the `incremental` snapshot.

roldanbob (approved):
Thanks for the updates! No major changes to the doc. LGTM!

Naros:
Will merge this once we get 2.6.0.Alpha1 out.

jpechane:
@JordanP Applied, thanks

PR #5172, opened 2024-01-19
Asynchronous implementation of DebeziumEngine

vjuranek:
https://issues.redhat.com/browse/DBZ-7024


vjuranek:
Still WIP, there are still a few `TODO` places and I need to add more tests, but the default (synchronous) implementation should pass all the MySQL tests (MySQL is switched to use new engine implementation in all the tests).

vjuranek:
New engine implementation is in [this commit](https://github.com/debezium/debezium/pull/5172/commits/8b7e8c2cfe01e7d670bd9ba693a3fa82bdca16a8), other commits are mostly minor refactoring of existing code, switching tests to new engine implementation or test fixes after the switch. Currently all MySQL/MariDB tests should run with the new engine implementation + few other tests (e.g. those inheriting from `AbstractSnapshotTest`) in other connectors.
From now on, to make review more easy, any changes will be push in separate commits.

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEngineConfig.java
```diff
@@ -0,0 +1,65 @@
...
+            .withDefault(1000L)
+            .withValidation(Field::isPositiveInteger);
+
+    /**
+     * An optional field that specifies how the records will be produced. Sequential processing (the default) means that the records will be produced in the same order
+     * as the engine obtained them from the connector. Non-sequential processing means that the records can be produced in arbitrary order, typically once the record is
+     * transformed and/or serialized.
+     */
+    Field RECORD_PROCESSING_SEQUENTIALLY = Field.create("record.processing.sequentially")
+            .withDescription("Determines how the records should be produce. Sequential processing means (setting to `true`, the default) that the records are "
```
mfvitale, line 50:
```suggestion
            .withDescription("Determines how the records should be produced. Sequential processing means (setting to `true`, the default) that the records are "
```

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEngineConfig.java
```diff
@@ -0,0 +1,65 @@
...
+
+    int AVAILABLE_CORES = Runtime.getRuntime().availableProcessors();
+
+    /**
+     * An optional field that specifies the maximum amount of time to wait for task lifecycle operation, i.e. for starting and stopping the task.
+     */
+    Field TASK_MANAGEMENT_TIMEOUT_MS = Field.create("task.management.timeout.ms")
+            .withDescription("Time to wait for task's lifecycle management operations (starting and stopping), given in milliseconds. "
+                    + "Defaults to 5 seconds (5000 ms).")
+            .withDefault(5_000L)
```
rk3rn3r, line 25: Why do we make this assumption and why is this required to be configurable? Isn't the actual process of starting a task an asynchronous process anyway and it should also be deterministic, e.g. retries retry-able errors like configured and fails with a task-specific timeout? There can be a timeout but do we really need users to configure it ans/or just allow a quite high timeout instead, at least the timeout should be higher when looking at experiences with certain environments (like in RHOC there were sometimes issues with starting up a connector took 2-5 mins or more)?

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEngineConfig.java
```diff
@@ -0,0 +1,65 @@
...
+            .withValidation(Field::isPositiveInteger);
+
+    /**
+     * An optional field that specifies the number of threads to be used for processing CDC records.
+     */
+    Field RECORD_PROCESSING_THREADS = Field.create("record.processing.threads")
```
rk3rn3r, line 31: I was not able to find the usage of this parameter (in this commit). Where is it used? Furthermore, depending on the workload and the environment over-committing (at least on bare-metal) can make sense, or limiting it using upper limits (like sum of running/scheduled/max tasks). For example environments like dedicated servers (probably bare-metal) for running the engine vs. 96-core servers running many different workloads in parallel.
What is meant by "processing CDC records"?

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEngineConfig.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * An optional field that specifies how the records will be produced. Sequential processing (the default) means that the records will be produced in the same order
+     * as the engine obtained them from the connector. Non-sequential processing means that the records can be produced in arbitrary order, typically once the record is
+     * transformed and/or serialized.
+     */
+    Field RECORD_PROCESSING_SEQUENTIALLY = Field.create("record.processing.sequentially")
+            .withDescription("Determines how the records should be produce. Sequential processing means (setting to `true`, the default) that the records are "
+                    + "produced in the same order as they were obtained from the database. Non-sequential processing means that the records can be produced in a different order "
+                    + "order than the original one, but this approach given better throughput, as the records are produced immediately once the SMTs and serialization of "
+                    + "the message is done, without waiting of other records.")
```
rk3rn3r: What's the use-case for non-sequential aka "unordered" processing?

jpechane: Well, there were users who are interested only in knowing that event happened, not about order or payload and require high performance. Yet I'd recommend to change this option a bit. It should be named `record.processing.order` with values `ordered` and `unordered` and in future a new option will be added `ordered_per_key`.
@vjuranek WDYT?

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEngineConfig.java
jpechane: I'd definitely keep this configurable but it is really a bit too techincal tuning detail.
How about changing it into `internal` parameter and raise the default to 30 s?
Based on the feedback we might make it public.

vjuranek: @rk3rn3r when you run any async operation, you should always set some timeout, unless there is very good reason for it. Otherwise you risk to get stuck waiting for the result forever. Sure, I can increase the default to whatever value you like. I'd make it internal only in case the default is quite high, as @rk3rn3r mentioned, starting time can vary quite a lot across the environments so my expectation is changing this value may happen to many users.

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEngineConfig.java
vjuranek: It's a good point to make this field `String` which would accept some predefined value and eventually parse the string as a number.
However, I'm not sure I understand why to limit number of threads to some predefined value. Is the concern here possible delay due to context switching on overloaded environments?

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEngineConfig.java
vjuranek: @rk3rn3r you may take a look on the [design document](https://github.com/debezium/debezium-design-documents/pull/8) first to get some explanation what this implementation intend to do and how.

@jpechane yes, definitely make sense. I expect this discussion happen under https://github.com/debezium/debezium-design-documents/pull/8 This is what I meant yesterday when I said the design document should be reviewed first. No problem to discussed it here, but I'd like not to exclude other people (like Jeremy, who originally came with this idea) from the discussion.

jpechane: The third option is out of scope of this work, but I want to have config ready so we don't need than to play with config options again.

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEngineConfig.java
jpechane: Well, unless we test with such "absurd" HW setup I'd avoid to large default pools. It is more precautionary approach than anything else.

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
```diff
@@ -0,0 +1,1266 @@
...
+                else {
+                    throw new InterruptedException("Time out while waiting for source task to start.");
+                }
+
+                LOGGER.debug("Started task #{} out of {} tasks.", i + 1, nTasks);
+            }
+            catch (Exception e) {
+                LOGGER.debug("Task #{} (out of {} tasks) failed to start. Failed with {}", i + 1, nTasks, e.getMessage());
+                if (LOGGER.isDebugEnabled()) {
+                    e.printStackTrace();
```
jpechane, line 363: Here and at other places it should be converted to logging the stacktrace. But even betrer, you can add it as a part of the log message as the last parameter, cant't you?

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
```diff
@@ -0,0 +1,1266 @@
...
+                }
+
+                // Store only the first error.
+                if (error == null) {
+                    error = e;
+                }
+                continue;
+            }
+            LOGGER.debug("Calling connector callback after task is started.");
+            connectorCallback.ifPresent(DebeziumEngine.ConnectorCallback::taskStarted);
```
jpechane, line 373: Future improvement - the `taskStarted` should get info which task was started, id and config?

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
```diff
@@ -0,0 +1,1266 @@
...
+            runTasksPolling(tasks);
+            // Tasks run infinite polling loop until close() is called or exception is thrown.
+        }
+        catch (Throwable t) {
+            exitError = t;
+            LOGGER.error("Engine has failed with ", exitError);
+
+            final State stateBeforeStop = getEngineState();
+            // Skip shutting down the engine if it's already being stopped.
+            if (State.STOPPING.compareTo(stateBeforeStop) > 0) {
```
jpechane, line 190: Is this safe? Maybe the enum can provide method like `canBeStopped` and called here.

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
```diff
@@ -0,0 +1,1266 @@
...
+        final ExecutorCompletionService<Void> taskCompletionService = new ExecutorCompletionService(taskService);
+        for (EngineSourceTask task : tasks) {
+            final RecordProcessor processor = selectRecordProcessor();
+            processor.initialize(recordService, transformations, new SourceRecordCommitter(task));
+            taskCompletionService.submit(new PollRecords(task, processor, state));
+        }
+
+        for (int i = 0; i < tasks.size(); i++) {
+            try {
+                taskCompletionService.take().get();
```
jpechane, line 399: I am not happy with the blocking call here. Can we change the loop so `poll()` can be used with pause between iterations?

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
```diff
@@ -0,0 +1,1266 @@
...
+
+    /**
+     * Select and instantiate {@link RecordProcessor} based on the user configuration.
+     *
+     * @return {@link RecordProcessor} instance which will be used for processing the records.
+     */
+    private RecordProcessor selectRecordProcessor() {
+        // If the change consumer is provided, it has precedence over the consumer.
+        if (handler != null && recordConverter == null) {
+            LOGGER.debug("Using {} processor", ParallelSmtBatchProcessor.class.getName());
```
jpechane, line 417: I'd like to see `INFO` log level message saying which processor is used. This should be printed only once, not for every task.

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
```diff
@@ -0,0 +1,1266 @@
...
+    /**
+     * Possible engine states.
+     * Engine state must be changed only via {@link AsyncEmbeddedEngine#setEngineState(State, State)} method.
+     */
+    private enum State {
+        // Order of the possible states is important, enum ordinal is used for state comparison.
+        CREATING, // the engine is being started, which mostly means engine object is being created or was already created, but run() method wasn't called yet
+        INITIALIZING, // initializing the connector
+        CREATING_TASKS, // creating connector tasks
+        STARING_TASKS, // starting connector tasks
```
jpechane, line 811:
```suggestion
        STARTING_TASKS, // starting connector tasks
```
it is even in the spec :-)

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
```diff
@@ -0,0 +1,1266 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.embedded;
```
jpechane, line 6: Maybe it would be better to move this class to a separate package `io.debezium.embedded.async` and split it file into multiple classes?

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
```diff
@@ -0,0 +1,1266 @@
...
+            this.processor = processor;
+            this.engineState = engineState;
+        }
+
+        @Override
+        public Void doCall() throws Exception {
+            while (engineState.get() == State.POLLING_TASKS) {
+                LOGGER.debug("Thread {} running task {} starts polling for records.", Thread.currentThread().getName(), task.connectTask());
+                final List<SourceRecord> changeRecords = task.connectTask().poll(); // blocks until there are values ...
+                LOGGER.debug("Thread {} polled {} records.", Thread.currentThread().getName(), changeRecords == null ? "no" : changeRecords.size());
```
jpechane, line 900: PLease consider if the other log messages should not be changed to `TRACE`.

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
```diff
@@ -0,0 +1,1266 @@
...
+            this.userHandler = userHandler;
+        }
+
+        @Override
+        public void processRecords(final List<SourceRecord> records) throws Exception {
+            LOGGER.debug("Thread {} is submitting {} records for processing.", Thread.currentThread().getName(), records.size());
+            final List<Future<SourceRecord>> recordFutures = new ArrayList<>(records.size());
+            records.stream().forEachOrdered(r -> recordFutures.add(recordService.submit(new TransformRecord(r, transformations))));
+
+            LOGGER.debug("Thread {} is getting source records.", Thread.currentThread().getName());
```
jpechane, line 1027: Here and other processors - again some of the messages might be better to change to `TRACE`

jpechane (commented):
@vjuranek Great job! I believe that the PR in the current state reflects DDD.

I left couple of comments that relates more to the code details or organizations but don't require any functional changes.

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEngineConfig.java
vjuranek: done

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
vjuranek: ack (actually I intended to do so before marking ready for review ... obviously I forgot:-()

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
vjuranek: The thread is not in blocking state, but it should be in waiting state as `take()` [calls](https://github.com/openjdk/jdk/blob/be7cc1c2b083ac7cbcec6b0fe77caff16f14bb60/src/java.base/share/classes/java/util/concurrent/LinkedBlockingQueue.java#L435) `Conditon.await()`. So I'm not sure if there would be any advantage if we sleep here in a loop. Actually we would be just spending cycles here most of the time by awaking and sleeping threads with no effect.

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEngineConfig.java
rk3rn3r: Thx for the change. But I can't see this is an `internal.` field yet, is it?

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEngineConfig.java
rk3rn3r: In ideal case the parallelism can be aligned to the number of partitions for sinks that use partitioned storage like Kafka with its partitions. Parallelism could be the number of (active) partitions. But it will only work for downstream systems that support and expose partitioning and the used algorithm for partitioning. Aligning this might be too complex for systems where those parameters are known and impossible for systems that hides those information.

rk3rn3r:
Impressive work so far @vjuranek! My comments were just small things that were coming to my head while looking over your work.

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEngineConfig.java
```diff
@@ -0,0 +1,85 @@
...
+    /**
+     * The array of all exposed fields.
+     */
+    Field.Set ALL_FIELDS = EmbeddedEngineConfig.ALL_FIELDS.with(
+            RECORD_PROCESSING_SHUTDOWN_TIMEOUT_MS,
+            RECORD_PROCESSING_THREADS,
+            RECORD_PROCESSING_ORDER,
+            RECORD_PROCESSING_WITH_SERIAL_CONSUMER,
+            // internal fields
+            TASK_MANAGEMENT_TIMEOUT_MS);
```
rk3rn3r, line 84: It is not always suggested to add internal fields to `ALL_FIELDS`, but sometimes we do. Is this intentional here?

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
vjuranek: I have no particular intention wrt naming. I'm fine with whatever sounds better or whatever @Naros suggests.

As for the start/stop connectors: IIRC @jpechane explicitly wanted that (at least for now) the engine should manage exactly one connector, so start/stop of the connector should be equivalent to start/stop whole engine. In the future we may reconsider this (especially running source connector and jdbc sink connector withing one engine would make a lot of sense to me), but it's not part of this work.

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEngineConfig.java
vjuranek: TBH I'm not aware about any policy or system what should or shouldn't be added in `ALL_FIELDS`. So this is the result of my lack of knowledge in this area and therefore I added everything. So I get it right `TASK_MANAGEMENT_TIMEOUT_MS` should be removed from there?

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
vjuranek: done now

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
Naros: > I only want to make sure this is intentional, because on the first read I felt like names should be TASKS_CREATING, TASKS_STARTING, TASKS_POLLING. On the second look I did not feel like this has to be changed but could, but I wanted to ask your opinion

So this really generally boils down to a question of what is the focus of this enum.  I'd say that our focus is on the state the engine is in, "starting" / "initializing" / "running" / "polling", and less so on the intermediate object that is being acted upon.  This also aligns with what Vojta chose.  If our focus was more on the object being acted upon and less so on the action being taken, then using "<noun>_<verb>" form would be more appropriate vernacular.

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEngineConfig.java
Naros: Hi @rk3rn3r, I have not looked at this PR in it's totality to know precisely how this `ALL_FIELDS` are used, so I'm going to rely on @vjuranek to help me out here :)

At least in KC land where we define a connector, we include the internal properties as part of the `ALL_FIELDS` declaration as this is used to define what the `ConfigDef` is that gets returned to KC and ultimately what KC bases it's own validation hooks upon.  So if we omit such fields from the `ALL_FIELDS` configuration, that field won't have it's validation hooks fired at least within KC -- and I'd surmise that we ideally should be operating in very much the same approach in the Embedded Engine and by Extension DS.

jpechane: In fact we do add them to `ALL_FIELDS` in connectors - at least I found most of them. It is possible that we are not really consistent. But we add `InvisibleRecommender` to them. This should be in fact done by default in `createInternal` instead of delegating it to field definitions.

debezium-embedded/src/main/java/io/debezium/embedded/AsyncEmbeddedEngine.java
jpechane: Thanks for the discussion. I must admit I don't have any preference to it either to I cowardly defer to the native speaker too :-). I beleive it can go as is now so I am merging the PR as the names can be change later.

jpechane:
@vjuranek Applied, thanks a lot for the great work!

PR #5182, opened 2024-01-23
DBZ-7376 Fix logging for schema only recovery in mysql

akanimesh7:
https://issues.redhat.com/browse/DBZ-7376


jpechane:
@akanimesh7 Applied, thanks

PR #5183, opened 2024-01-23
DBZ-7373 Zookeeper container not being deleted quick fix + added wait…

novotnyJiri:
https://issues.redhat.com/browse/DBZ-7373

### DIFF
--- a/debezium-testing/debezium-testing-system/src/main/java/io/debezium/testing/system/tools/kafka/DockerKafkaController.java
+++ b/debezium-testing/debezium-testing-system/src/main/java/io/debezium/testing/system/tools/kafka/DockerKafkaController.java
@@ -13,6 +13,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.debezium.testing.system.tools.kafka.docker.KafkaContainer;
+import io.debezium.testing.system.tools.kafka.docker.ZookeeperContainer;
 
 /**
  * This class provides control over Kafka instance deployed as DockerContainer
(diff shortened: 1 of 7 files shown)


debezium-testing/debezium-testing-system/src/main/java/io/debezium/testing/system/tools/kafka/DockerKafkaController.java
```diff
@@ -51,7 +57,8 @@ public String getTlsBootstrapAddress() {
     @Override
     public boolean undeploy() {
         container.stop();
-        return container.isRunning();
+        zookeeperContainer.stop();
+        return zookeeperContainer.isRunning();
```
jpechane, line 61:
```suggestion
        return container.isRunning() && zookeeperContainer.isRunning();
```
but I am a bit confused on te return value, it seems to me accroding to JavaDoc of the method on interface it should return negated value of `isRunning()` as I suppos that successful `undeploy` means that container is NOT running. If it is the case it should be fixed at other controllers too.

jpechane (changes requested):
@novotnyJiri Please see the comment about `undeploy` method

debezium-testing/debezium-testing-system/src/main/java/io/debezium/testing/system/tools/kafka/DockerKafkaController.java
novotnyJiri: Great catch with the JavaDoc discrepancy! I will fix all the controllers accordingly.

jpechane:
@novotnyJiri Applied, thanks!

PR #5184, opened 2024-01-23
DBZ-7380 Offset transaction id only when it's non null

akanimesh7:
https://issues.redhat.com/browse/DBZ-7380


jpechane:
@akanimesh7 Applied, thanks

PR #5185, opened 2024-01-24
[docs] Fix name of capture table DDL option for SQL server

vjuranek:


jpechane:
@vjuranek Applied, thanks.

PR #5188, opened 2024-01-24
DBZ-7374 Allow NULL as the default value for SQL server (Unicode) character strings

iankko:
https://issues.redhat.com/browse/DBZ-7374

SQL server connector -- allow the default value mapper properly to accept character strings & Unicode character strings _**n(char|varchar|text)**_, when using NULL as the default value

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -240,6 +240,7 @@ Jakub Cechacek
 James Gormley
 James Johnston
 Jan-Hendrik Dolling
+Jan Lieskovsky
 Jan Uyttenhove
 JapuDCret
 Jaromir Hamala
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerDefaultValueConverter.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerDefaultValueConverter.java
@@ -153,14 +153,14 @@ class SqlServerDefaultValueConverter implements DefaultValueConverter {
         });
 
         // Character strings
-        result.put("char", (c, v) -> v.substring(2, v.length() - 2)); // Sample value: ('aaa')
-        result.put("text", (c, v) -> v.substring(2, v.length() - 2)); // Sample value: ('aaa')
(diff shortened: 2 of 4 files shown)


iankko:
@jpechane PTAL once got a chance / a minute (I would appreciate also some pointer / examples on how to contribute a test, so I could add them)

Thanks,
Jan

Naros:
Hi @iankko, there is already a default value test in the `SqlServerConnectionIT` where a table is created, and we assert the resolved default values; see `shouldProperlyGetDefaultColumnValues`.

You can either choose to add a new column to that table definition with the right default value combinations your change is to support and then assert the column's state in that same method **or** copy that method into a new method with an appropriate name related to this PR's change, remove all the columns expect those that apply to this change and then assert on only those columns.

iankko:
@Naros

Hello Chris, 

> Hi @iankko, there is already a default value test in the `SqlServerConnectionIT` where a table is created, and we assert the resolved default values; see `shouldProperlyGetDefaultColumnValues`.

Wonderful, TYVM for the pointers! (these are exactly the ones I was searching for).

> 
> You can either choose to add a new column to that table definition with the right default value combinations your change is to support and then assert the column's state in that same method **or** copy that method into a new method with an appropriate name related to this PR's change, remove all the columns except those that apply to this change and then assert on only those columns.

Noticed, there was existing `shouldProperlyGetDefaultColumnNullValues()` test method, thus added those tests for null string (char, varchar etc.) datatype there (for all three combinations of `NULL`, `(NULL)`, and `((NULL))` we are aware might cause the problem (seems various versions of MS SQL server can return either `(NULL)` or `((NULL))`, but not exactly sure TBH which particular version is returning which variant. Should you need this list, this would need to be collected yet).

Thanks again!
Jan

jpechane:
LGTM, I triggered CI build. If passes the PR can be merged and backported to 2.5 too.

iankko:
> LGTM, I triggered CI build. If passes the PR can be merged and backported to 2.5 too.

👍

jpechane:
@iankko Applied, thanks!

iankko:
> @iankko Applied, thanks!

@jpechane Wonderful, thank you very much! Truly appreciate it (extra points for merging this against / into the `2.5` branch too! :))

PR #5189, opened 2024-01-25
DBZ-7362 modify db2 docs that add DECFLOAT datatype mapping

leoloel:


jpechane:
@leoloel Applied thanks!

PR #5191, opened 2024-01-25
DBZ-7382 Don't check record order in parallel snapshot tests

vjuranek:
https://issues.redhat.com/browse/DBZ-7382


jpechane:
@vjuranek Applied, thanks

PR #5192, opened 2024-01-25
DBZ-7299 Ability to use post images for full document update

jcechace:
https://issues.redhat.com/browse/DBZ-7299


jcechace:
@jpechane should be good to review. I will try to implement some tests if possible. It should be doable by stopping the connector and performing two consecutive updates. Regular lookup should result in the latest version in both events, post-image lookup should properly reflect the gradual changes.

jpechane:
@jcechace Applied, thanks!

PR #5122, opened 2023-12-20
DBZ-7273: Introduce mode to query capture tables directly

ramanenka:
https://issues.redhat.com/browse/DBZ-7273

PR #5194, opened 2024-01-25
DBZ-7273: Document `data.query.mode` for sqlserver connector

ramanenka:
This documents a new SQL Server connector setting introduced in #5122.


jpechane:
@ramanenka Applied, thanks!

PR #5196, opened 2024-01-26
DBZ -7301 Implement Snapshotter SPI MySQL/MariaDB

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7233

The PR is based on https://github.com/debezium/debezium/pull/5157, must be rebased after its merge. 

- [x] Implement the Snapshotter interface
- [x] Implement the SnapshotLock interface
- [x] Implement the SnapshotQuery interface 
- [x] Add test for a custom Snapshotter
- [x]  Provide PR for Db2, Spanner, Vitess to address new code change

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbChangeEventSourceFactory.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbChangeEventSourceFactory.java
@@ -23,6 +23,7 @@ import io.debezium.pipeline.source.spi.DataChangeEventListener;
 import io.debezium.pipeline.source.spi.SnapshotChangeEventSource;
 import io.debezium.pipeline.source.spi.SnapshotProgressListener;
 import io.debezium.pipeline.source.spi.StreamingChangeEventSource;
+import io.debezium.snapshot.SnapshotterService;
 import io.debezium.spi.schema.DataCollectionId;
 import io.debezium.util.Clock;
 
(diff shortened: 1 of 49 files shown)


mfvitale:
> I expect that with migrating another connector there will be a refactoring of snapshotter implementations so theere will be abstract classes for each of the mode with methods that will be overriden per-connector - mostly validate, am I right?

The initial idea is to start adding the possibility to customize the SPIs and the with https://issues.redhat.com/browse/DBZ-7308 unify the modes. So at the end I think that the snapshotter implementations will move to the core and, as you said, maybe just 'validate' will be connector specific. Let's see how evolve. 

> Also I believe it will be possible to make common SnapshotMode class re-used among connectors (or maybe two one with schema history and one without).

Yes this is the idea. I intend to do this at the end with https://issues.redhat.com/browse/DBZ-7308, since now every connectors have their way to manage it. 

> Also in that case the Percona snapshot mode should be removed and the percona users would be required to specify Percona locking implementation. WDYT?

I don't see any "Percona" snapshot mode. Do you mean locking?

In any case I am trying to refactor while adding more connector. For example you can see https://github.com/debezium/debezium/pull/5232 that I moved the all the stuff related to `SnapshotQuery` to the `debezium-core` and partially the custom part for `SnapshotLock`. The `SnapshotLockingMode` is still not in the `core` since MySQL has very special lock modes. Let's see if we can move at the end. 

Generally I am using a step by step approach to let the design emerging.

jpechane:
@mfvitale Yes, sorry, I meant locking

PR #5198, opened 2024-01-27
DBZ-7420 Modify alterSpecification Clause

ilyasahsan123:
https://issues.redhat.com/browse/DBZ-7420


jpechane:
@ilyasahsan123 Applied, thanks!

PR #5199, opened 2024-01-29
DBZ-7421 Improve incremental snapshot performance with increasing number of collections to snapshot

mfvitale:
a possibile solution for https://issues.redhat.com/browse/DBZ-7421

Before
```
Benchmark                             (numberOfTableToSnapshot)   Mode  Cnt       Score   Error  Units
IncrementalSnapshotContextPerf.store                          1  thrpt    2  962999.691          ops/s
IncrementalSnapshotContextPerf.store                          5  thrpt    2  118903.880          ops/s
IncrementalSnapshotContextPerf.store                         10  thrpt    2   31712.744          ops/s
IncrementalSnapshotContextPerf.store                         50  thrpt    2     786.632          ops/s
IncrementalSnapshotContextPerf.store                        100  thrpt    2     195.274          ops/s
```

After some code optimization
```
Benchmark                             (numberOfTableToSnapshot)   Mode  Cnt       Score   Error  Units
IncrementalSnapshotContextPerf.store                          1  thrpt    2  926594.940          ops/s
IncrementalSnapshotContextPerf.store                          5  thrpt    2  130743.643          ops/s
IncrementalSnapshotContextPerf.store                         10  thrpt    2   30547.231          ops/s
IncrementalSnapshotContextPerf.store                         50  thrpt    2     873.143          ops/s
IncrementalSnapshotContextPerf.store                        100  thrpt    2     222.739          ops/s
```

With the cache
```
Benchmark                             (numberOfTableToSnapshot)   Mode  Cnt        Score   Error  Units
IncrementalSnapshotContextPerf.store                          1  thrpt    2  1794620.143          ops/s
IncrementalSnapshotContextPerf.store                          5  thrpt    2  1984367.094          ops/s
IncrementalSnapshotContextPerf.store                         10  thrpt    2  1792400.614          ops/s
IncrementalSnapshotContextPerf.store                         50  thrpt    2  1985691.291          ops/s
IncrementalSnapshotContextPerf.store                        100  thrpt    2  1911970.545          ops/s
```

Even if improving the code that serializes the data collections to JSON, the performance was not good. Adding a cache seems the best approach here. 
With this implementation the serialize code will not affect the `store` method that is called during the event `IncrementalSnapshotChangeRecordReceiver` but will affect the methods that modify the `dataCollectionsToSnapshot` variable, but the impact in this case is limited to the number of data collections to snapshot and not at every events. 

@Naros, @jpechane @vjuranek @jcechace  WDYT?

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotContext.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotContext.java
@@ -348,7 +348,7 @@ public class AbstractIncrementalSnapshotContext<T> implements IncrementalSnapsho
                 + Arrays.toString(maximumKey) + "]";
     }
 
-    public static class SnapshotDataCollection<T> extends LinkedList<DataCollection<T>> {
+    private static class SnapshotDataCollection<T> extends LinkedList<DataCollection<T>> {
 
         public static final String DATA_COLLECTIONS_TO_SNAPSHOT_KEY = INCREMENTAL_SNAPSHOT_KEY + "_collections";
 
@@ -365,15 +365,15 @@ public class AbstractIncrementalSnapshotContext<T> implements IncrementalSnapsho
         private final Queue<DataCollection<T>> dataCollectionsToSnapshot = new LinkedList<>();
         private String dataCollectionsToSnapshotJson;
 
-        public SnapshotDataCollection() {
+        SnapshotDataCollection() {
         }
 
-        public void add(List<DataCollection<T>> dataCollectionIds) {
+        void add(List<DataCollection<T>> dataCollectionIds) {
             this.dataCollectionsToSnapshot.addAll(dataCollectionIds);
             this.dataCollectionsToSnapshotJson = jsonString();
         }
 
-        public DataCollection<T> getNext() {
+        DataCollection<T> getNext() {
             DataCollection<T> nextDataCollection = this.dataCollectionsToSnapshot.poll();
             this.dataCollectionsToSnapshotJson = jsonString();
             return nextDataCollection;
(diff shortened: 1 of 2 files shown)


jpechane:
@mfvitale I think the solution seems to be right! I'd just prefer now as both the attributes are linked together not to manipulate them directly in the methods but instead provide specific methods that will change both of them when needed. WDYT?

jpechane:
Maybe it is necessary just for clear scenario.

vjuranek (approved):
LGTM

jpechane:
@mfvitale I think it is good, thanks!
Theoretically it should be delegation (HAS-A) not inheritance (IS-A) relationship but I think it is good pragmatic solution.

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotContext.java
```diff
@@ -390,8 +343,112 @@ public void setSchemaVerificationPassed(boolean schemaVerificationPassed) {
     @Override
     public String toString() {
         return "IncrementalSnapshotContext [windowOpened=" + windowOpened + ", chunkEndPosition="
-                + Arrays.toString(chunkEndPosition) + ", dataCollectionsToSnapshot=" + dataCollectionsToSnapshot
+                + Arrays.toString(chunkEndPosition) + ", dataCollectionsToSnapshot=" + snapshotDataCollection.getDataCollectionsToSnapshot()
                 + ", lastEventKeySent=" + Arrays.toString(lastEventKeySent) + ", maximumKey="
                 + Arrays.toString(maximumKey) + "]";
     }
+
+    public static class SnapshotDataCollection<T> extends LinkedList<DataCollection<T>> {
```
jpechane, line 351:
```suggestion
    private class SnapshotDataCollection<T> extends LinkedList<DataCollection<T>> {
```

jpechane (approved):

mfvitale:
> @mfvitale I think it is good, thanks! Theoretically it should be delegation (HAS-A) not inheritance (IS-A) relationship but I think it is good pragmatic solution.

Isn't it?

Naros (approved):

mfvitale:
@jpechane @Naros seems something eligible for a backport, WDYT?

jpechane:
@mfvitale Applied, thanks!

PR #5200, opened 2024-01-29
DBZ-7379 Connector specific keystore/truststore

jcechace:
https://issues.redhat.com/browse/DBZ-7379

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
@@ -5,6 +5,9 @@
  */
 package io.debezium.connector.mongodb;
 
+import static java.util.function.Predicate.not;
+
+import java.nio.file.Path;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -30,7 +33,6 @@ import io.debezium.config.Field;
 import io.debezium.config.Field.ValidationOutput;
 import io.debezium.connector.AbstractSourceInfo;
 import io.debezium.connector.SourceInfoStructMaker;
-import io.debezium.connector.mongodb.connection.ConnectionStrings;
 import io.debezium.connector.mongodb.connection.DefaultMongoDbAuthProvider;
 import io.debezium.connector.mongodb.connection.MongoDbAuthProvider;
 import io.debezium.data.Envelope;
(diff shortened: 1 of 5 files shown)


jcechace:
@jpechane I will send a separate PR which allows testing this with external database. However from our preliminary testing with @obabec  it looks like this is good to go.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
```diff
@@ -563,6 +565,41 @@ public static OversizeHandlingMode parse(String value, String defaultValue) {
...
+
+    public static final Field SSL_TRUSTSTORE = Field.create("mongodb.ssl.truststore")
+            .withDisplayName("SSL Truststore")
+            .withType(Type.STRING)
+            .withGroup(Field.createGroupEntry(Field.Group.CONNECTION_ADVANCED_SSL, 3))
+            .withWidth(Width.LONG)
+            .withImportance(Importance.MEDIUM)
+            .withDescription("The location of the trust store file for the server certificate verification.");
+
+    public static final Field SSL_TRUSTSTORE_PASSWORD = Field.create("mongodb.ssl.truststore.password")
```
Naros, line 594: Did we intentionally prefix these with the `mongodb.ssl.*` namespace rather than just using `ssl.*`?

jcechace: Well, for MySQL these are prefixed "database.ssl.*", additionally MongoDb already contained `mongodb.ssl.enabled`. I see where you are going, however these properties are a bit all over the place across connectors. 

I've talk to @jpechane about potentially restructuring for DBZ 3.0

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/connection/client/MongoDbClientFactory.java
```diff
@@ -0,0 +1,121 @@
...
+    /**
+     * Creates keystore
+     *
+     * @param path keyfile path
+     * @param password keyfile password
+     * @return keystore with loaded keys
+     */
+    static KeyStore loadKeyStore(Path path, char[] password) {
+        try (var keys = Files.newInputStream(path)) {
+            var ks = KeyStore.getInstance("PKCS12");
```
jcechace, line 64: @jpechane @Naros  this could actually also be configurable. So user should decide whether to use `jks` or `pkcs12` or whatever else.

jpechane: @jcechace ok, let's make it configurable. I think we'll take MongoDB as test bed from SSL config PoV that will be generalized and re-used accross all connectors.

jcechace: Done

jpechane:
@jcechace Applied, thanks! BTW, it might be nice as a follow-up task to extract the SSL config into a separate class.

jcechace:
> @jcechace Applied, thanks! BTW, it might be nice as a follow-up task to extract the SSL config into a separate class.

Definitively something I want to do. Just need to have a chat with you about the design. It's a bit more complicated than I initially though and I'd like to created something that can eventually be pulled up for other connectors as well.

PR #5209, opened 2024-01-30
DBZ-7429 Resolve primary key values from after struct

Naros:
https://issues.redhat.com/browse/DBZ-7429


jpechane:
@Naros Applied, thanks!

PR #5212, opened 2024-01-31
DBZ-7432 Create systemtests for Unwrap SMT

obabec:
- Bumping apicurio model version
- Adding test for unwrap SMT


Naros (approved):
LGTM @obabec.

PR #5215, opened 2024-01-31
[docs] List required files for Confluent Avro converter

mostafaghadimi:
Required JAR files are added to properly integrate the Debezium integration with AVRO and Confluent Schema Registry.

### DIFF
--- a/documentation/modules/ROOT/pages/configuration/avro.adoc
+++ b/documentation/modules/ROOT/pages/configuration/avro.adoc
@@ -383,11 +383,32 @@ To enable the Confluent Schema Registry for a {prodname} container, install the 
 * `kafka-connect-avro-data`
 * `kafka-avro-serializer`
 * `kafka-schema-serializer`
+* `kafka-schema-converter`
 * `kafka-schema-registry-client`
 * `common-config`
 * `common-utils`
 
 You can download the preceding files from the https://packages.confluent.io/maven/[Confluent Maven repository].
+
+There are also some other JAR files required to be located into the Connect plugin directory:
+
+* `avro`
+* `commons-compress`
+* `failureaccess`
+* `guava`
+* `minimal-json`
+* `re2j`
+* `slf4j-api`
+* `snakeyaml`
+* `swagger-annotations`
+* `jackson-databind`
+* `jackson-core`
+* `jackson-annotations`
+* `jackson-dataformat-csv`
+* `logredactor`
+* `logredactor-metrics`
+
+You can download the preceding files from the https://mvnrepository.com/[Maven repository].
 
 The configuration is slightly different.
 


documentation/modules/ROOT/pages/configuration/avro.adoc
```diff
@@ -383,12 +383,33 @@ To enable the Confluent Schema Registry for a {prodname} container, install the
...
+* `re2j`
+* `slf4j-api`
+* `snakeyaml`
+* `swagger-annotations`
+* `jackson-databind`
+* `jackson-core`
+* `jackson-annotations`
+* `jackson-dataformat-csv`
+* `logredactor`
+* `logredactor-metrics`
```
Naros, line 409: First, thanks for the contribution @mostafaghadimi.  

I'm worried about explicitly listing the transitive dependencies, as the list can change from release to release and could become an unnecessary maintenance task. I wonder whether Confluent has a location that lists these by version or if they ship a distributable with all dependencies that would make more sense to reference here.  wdyt?

mostafaghadimi: Hi @Naros,

Thank you for reviewing the PR. 
Here is the required JAR files that Confluent introduced as Kafka Connect Avro Converter:
https://www.confluent.io/hub/confluentinc/kafka-connect-avro-converter

As you can download and check the JAR files (located at `lib` directory), it is indicated that these files are required.
I have also tried running the quickstart example of Debezium using AVRO serialization. If the above JAR files aren't provided, then we will receive different errors like:
https://stackoverflow.com/questions/77605649/debezium-connect-java-lang-noclassdeffounderror-com-google-common-base-ticker

I also resolved this issue and create an example for it in the following repository:
https://github.com/mostafaghadimi/debezium_avro_integration

jpechane: @mostafaghadimi Thanks for the link, I was not aware the converteor has now its own stndalone distribution. How about changing the procedure to download and unzip this file instead of enumerating the dependencies?

mostafaghadimi: Hi @jpechane ,
What you say makes sense, however, the JAR files founded in the link is even not completed! In order to resolve my issue using Avro I was to install another JAR files that I've listed in the documentation.

mostafaghadimi: @jpechane @Naros Would you please give me update on this?

jpechane: @mostafaghadimi Unofrtunately there is no best solution wo we have accept this as good enough. Thanks.

jpechane:
Please squash the commits and use commit message like `[docs] List required files for Confluent Avro converter`. Other than that we are good to go. Thanks!

jpechane:
@mostafaghadimi It seems to me the PR got dirty. There are still two commits and the first one seems to be merged with another docs commit. Was it intentional? Maybe you squashed incorrect commits?

mostafaghadimi:
It should be fixed right now. Thanks
@jpechane

jpechane:
@mostafaghadimi Applied, thanks!

PR #5221, opened 2024-01-31
DBZ-7434 Error caused by database name containing regular keywords, such as database name: test$user

jw-itq:
https://issues.redhat.com/browse/DBZ-7434


jpechane:
@jw-itq Hi, thanks for the PR. What is the current status? Do you intend to continue to work on the fix?

PR #5223, opened 2024-01-31
DBZ-7437: ReselectColumnsPostProcessor filter not use exclude predicate.

Fr0z3Nn:
The "reselect.columns.exclude.list" property is not applied when configuring the connector. Configuration without exclude+include do not processed (all passed values ignoring).
Reason:
The reselectColumnInclusions predicate never becomes null.
Build filter condition is always true.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -582,3 +582,4 @@ Peter Hamer
 Artem Shubovych
 leoloel
 Clifford Cheefoon
+Fr0z3Nn
--- a/debezium-core/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
+++ b/debezium-core/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
@@ -252,7 +252,9 @@ public class ReselectColumnsPostProcessor implements PostProcessor, BeanRegistry
             if (columnNames == null || columnNames.trim().isEmpty()) {
                 reselectColumnInclusions = null;
             }
-            reselectColumnInclusions = Predicates.includes(columnNames, Pattern.CASE_INSENSITIVE);
+            else {
+                reselectColumnInclusions = Predicates.includes(columnNames, Pattern.CASE_INSENSITIVE);
+            }
             return this;
         }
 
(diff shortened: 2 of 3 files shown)


Naros:
Hi @Fr0z3Nn there appears to be a number of formatting changes, that are superfluous.  Can you please run:
```
./mvnw clean install -DskipITs -DskipTests
```
That should format the code and then squash that change into your prior commit.  Thanks.

Naros (approved):
LGTM, just waiting on CI.  :+1:

Fr0z3Nn:
@Naros Thanks for approve, can I restart only one step of CI?
Build Debezium / PostgreSQL (assembly,postgres-16,pgoutput-decoder) (pull_request)

2024-01-31T20:33:40.0863414Z [ERROR] Errors: 
2024-01-31T20:33:40.0865826Z [ERROR]   RecordsStreamProducerIT.testEmptyChangesProducesHeartbeat:2509 » ConditionTimeout Condition with io.debezium.connector.postgresql.RecordsStreamProducerIT was not fulfilled within 50 seconds.

jpechane:
@Fr0z3Nn I triggered the build for pgoutput

Fr0z3Nn:
@jpechane Hi, can we restart failed jobs? I think issue in low tests timeouts, I will check how my changes affect this tests after jobs rerun

jpechane:
I can't reproduce the issue locally yet the Postgres test failed 3 times in a row which is quite suspicious.

Naros:
It appears the PostgreSQL failure on PG16 was released to a heartbeat test, which I believe is to be unrelated.  For sanity sake, I'm going to rebase the PR against `main` and rerun the suite one last time to be sure.

Naros:
Hi, @jpechane; I'm not convinced that the issue related to PG16 is attributable to this PR.  

I have reproduced the issue locally, setting up a run-until-failure scenario, and it's timing-related.  From what I see, the first Awaitility loop before the schema change records the same LSN as the Awaitility loop after the `CREATE SCHEMA s1`, which seems to contradict the expectations of the test.  But this is a challenging situation to reproduce; in some instances, it took upwards of 3500 iterations to see the failure, and in others, it took less than three dozen.

For curiosity's sake, I will check with PG15 now; perhaps this is specific to PG16 since we haven't seen the older PG job fail with this test case; however, that, too, could be a sheer timing coincidence.

jpechane:
@Fr0z3Nn Applied, thanks!

@Naros So this looks like a kind of behaviour in database engine that based on timing might or might not put the changes under the same LSN. So maybe we should mark the test as Flaky?

Naros:
Logged https://issues.redhat.com/browse/DBZ-7453, added https://github.com/debezium/debezium/pull/5238 to mark the test as Flaky for now.

PR #5225, opened 2024-02-01
DBZ-7359 Added setQueryColumnValue fix and supporting Integration Tests

coreillylocus:
@jpechane and @mfvitale submitting this draft pull request with suggested fix. I set up the integration tests in a separate file due to not following the same pattern of having int PKs that the other tests leverage.

The main branch tests don't appear to be stable as I've been getting this:

[ERROR] Failures: 
[ERROR]   SpecialCharsInNamesIT.shouldParseSpecialChars:159 
Expecting actual not to be null
[INFO] 
[ERROR] Tests run: 223, Failures: 1, Errors: 0, Skipped: 4

with the mssql connector integrations with our without the changes in this PR. 

Let me know your thoughts.


mfvitale:
Hi @coreillylocus , thanks for the PR. 

I am just wondering if we can just manage explicitly both `NCHAR` and `CHAR` (and the other alternatives) w/o wondering about the `sendStringParametersAsUnicode` parameter. In that case I think you will not be forced to use the `sendStringParametersAsUnicode` and then just use the right `CHAR` values during the query avoiding to use the `setObject` that will transform the value in `N'value'`

WDYT?

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/IncrementalSnapshotCollationSortOrderMismatchIT.java
```diff
@@ -0,0 +1,395 @@
...
+import io.debezium.config.Configuration;
+import io.debezium.connector.sqlserver.util.TestHelper;
+import io.debezium.jdbc.JdbcConnection;
+import io.debezium.pipeline.source.snapshot.incremental.AbstractSnapshotTest;
+import io.debezium.util.Testing;
+
+/**
+ * Tests for <a href="https://issues.redhat.com/browse/DBZ-7359">DBZ-7359</a>
+ */
+public class IncrementalSnapshotCollationSortOrderMismatchIT extends AbstractSnapshotTest<SqlServerConnector> {
```
jpechane, line 40: I don't think it is necessary to inherit from `AbstractSnapshotTest` and fit into its structure. You can make a completely standalone test out of it if it would be easier.

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/IncrementalSnapshotCollationSortOrderMismatchIT.java
```diff
@@ -0,0 +1,395 @@
...
+    private static final List<String> ALL_IDS = new ArrayList<>();
+    // Contains the IDs that get skipped when there is a sort order mismatch.
+    private static final List<String> SKIPPED_IDS = new ArrayList<>();
+
+    private SqlServerConnection connection;
+    private boolean isSendStringParametersAsUnicode;
+
+    @BeforeClass
+    public static void beforeClass() throws IOException {
+        try (InputStreamReader isr = new InputStreamReader(
```
jpechane, line 55: You can re-use `io.debezium.util.IoUtil` methods.

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/IncrementalSnapshotCollationSortOrderMismatchIT.java
```diff
@@ -0,0 +1,395 @@
...
+                        var id = ALL_IDS.get(i);
+                        // NOTE: in the case where isSendStringParametersAsUnicode is true, the 36 records will
+                        // have been skipped (i.e. you need to set it to false for it to work).
+                        if (isSendStringParametersAsUnicode && SKIPPED_IDS.contains(id)) {
+                            continue;
+                        }
+                        var val = dbChanges.get(id);
+                        if (val == null || val != i) {
+                            result = false;
+                            System.err.println(ALL_IDS.get(i) + " value is not = " + i + ", is = " + val);
```
jpechane, line 155:
```suggestion
                            Testing.Print(ALL_IDS.get(i) + " value is not = " + i + ", is = " + val);
```

jpechane: Same for the other instances

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/IncrementalSnapshotCollationSortOrderMismatchIT.java
```diff
@@ -0,0 +1,395 @@
...
+                "int",
+                (_pk, idx) -> String.format("%d", idx),
+                record -> ((Struct) record.value()).getStruct("after").getInt32(valueFieldName()),
+                dbChanges -> {
+                    boolean result = true;
+                    for (int i = 0; i < ALL_IDS.size(); i++) {
+                        var id = ALL_IDS.get(i);
+                        // NOTE: in the case where isSendStringParametersAsUnicode is true, the 36 records will
+                        // have been skipped (i.e. you need to set it to false for it to work).
+                        if (isSendStringParametersAsUnicode && SKIPPED_IDS.contains(id)) {
```
jpechane, line 149: I wonder if it is important to test cases that does not work and we expect they will not work.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
```diff
@@ -598,6 +602,61 @@ public Object getColumnValue(ResultSet rs, int columnIndex, Column column, Table
         }
     }
 
+    // NOTE: fix for DBZ-7359
+    @Override
+    public void setQueryColumnValue(PreparedStatement statement, Column column, int pos, Object value) throws SQLException {
+        boolean isColumnValueSet = false;
```
jpechane, line 608: I think instead of the flag you can intorduce `default` branch in the `switch` and `return` for the conditional

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
```diff
@@ -598,6 +602,61 @@ public Object getColumnValue(ResultSet rs, int columnIndex, Column column, Table
...
+
+        if (column.typeUsesCharset()) {
```
jpechane, line 610: Also it might be good to add a `TRACE` message here and another if we did not handle it went via default.

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/IncrementalSnapshotCollationSortOrderMismatchIT.java
coreillylocus: I tried that initially but the test leverages these 3 methods:
 AbstractConnectorTest.initializeConnectorTestFramework() 
 AbstractSnapshotTest.startConnector()
 AbstractSnapshotTest.sendAdHocSnapshotSignal() 
and I wanted to avoid cutting and pasting the code into it. Let me know if you'd like me to do this and make it completely standalone.

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/IncrementalSnapshotCollationSortOrderMismatchIT.java
coreillylocus: done.

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/IncrementalSnapshotCollationSortOrderMismatchIT.java
coreillylocus: I used Testing.printError() as I want to see the error if it occurs (it shouldn't normally). The print() method appears to need to be explicitly enabled to print something.

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/IncrementalSnapshotCollationSortOrderMismatchIT.java
coreillylocus: I think its important to call out in the tests that the expected behavior is to drop records if the default setting of `SendStringParametersAsUnicode=true` is set. What do you think of adding some documentation under [here](https://debezium.io/documentation/reference/2.5/connectors/sqlserver.html#sql-server-basic-values) that calls out the need to set this flag to true if the server being connected to is using an SQL_ collation and uses char types on its PK columns.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
coreillylocus: I removed the `isColumnValueSet` flag but the logic with the flag always set the value, so I had to add several additional else statements to get the equivalent logic. I'm not terribly keen on either approach as they are both brittle to maintenance changes. Thoughts?

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
coreillylocus: I've held off on doing this till we agree on the cleanest implementation of this method.

coreillylocus:
> Hi @coreillylocus , thanks for the PR.
> 
> I am just wondering if we can just manage explicitly both `NCHAR` and `CHAR` (and the other alternatives) w/o wondering about the `sendStringParametersAsUnicode` parameter. In that case I think you will not be forced to use the `sendStringParametersAsUnicode` and then just use the right `CHAR` values during the query avoiding to use the `setObject` that will transform the value in `N'value'`
> 
> WDYT?

coreillylocus:
> Hi @coreillylocus , thanks for the PR.
> 
> I am just wondering if we can just manage explicitly both `NCHAR` and `CHAR` (and the other alternatives) w/o wondering about the `sendStringParametersAsUnicode` parameter. In that case I think you will not be forced to use the `sendStringParametersAsUnicode` and then just use the right `CHAR` values during the query avoiding to use the `setObject` that will transform the value in `N'value'`
> 
> WDYT?

@mfvitale I think you are suggesting extending the switch statement as follows?:
```java
switch (column.jdbcType()) {
                case Types.CHAR:
                    if (value instanceof String) {
                        statement.setString(pos, (String) value);
                    }
                    else {
                        // not set, fall back on default implementation.
                        super.setQueryColumnValue(statement, column, pos, value);
                    }
                    break;
                case Types.VARCHAR:
                    if (value instanceof String) {
                        statement.setCharacterStream(pos, new StringReader((String) value));
                    }
                    else if (value instanceof Reader) {
                        statement.setCharacterStream(pos, (Reader) value);
                    }
                    else {
                        // not set, fall back on default implementation.
                        super.setQueryColumnValue(statement, column, pos, value);
                    }
                    break;
                case Types.LONGVARCHAR:
                    if (value instanceof String) {
                        // we'll fall back on nvarchar handling
                        statement.setCharacterStream(pos, new StringReader((String) value));
                    }
                    else if (value instanceof Reader) {
                        // we'll fall back on nvarchar handling
                        statement.setCharacterStream(pos, (Reader) value);
                    }
                    else if (value instanceof NClob) {
                        statement.setClob(pos, (NClob) value);
                    }
                    else {
                        // not set, fall back on default implementation.
                        super.setQueryColumnValue(statement, column, pos, value);
                    }
                    break;
                   ...
```
The test that indicate records are skipped i.e.:

orderMismatchPkCharValueIntParamsAsUnicodeTrueSkip36
orderMismatchPkVarcharValueIntParamsAsUnicodeTrueSkip36

will still pass. My understanding is when `SendStringParametersAsUnicode=true` the JDBC driver is going to send your parameters as Unicode irrespective of what method you call on its API. So when an SQL_ collation is being used and a char column is part of the primary key then you have the potential to skip data. I suggested to @jpechane in another of the comments feedbacks to add this to the documentation explicitly so that people have a better chance of becoming aware of the need to set this flag to false.

jpechane:
@coreillylocus I am sorry that I get back to the PR only now. I resolved my comments, the only issue open is the documentation. Would you please be able to finish the PR and undraft it so we can merge it?

coreillylocus:
@jpechane I'll try get to over the next few days.

coreillylocus:
> @coreillylocus I am sorry that I get back to the PR only now. I resolved my comments, the only issue open is the documentation. Would you please be able to finish the PR and undraft it so we can merge it?

Hi @jpechane sorry for the delay with getting back to this (swamped at work). I've undrafted the PR and added the warning documentation. I changed the location from the original proposed section under data types to the adhoc snapshots section as that seems more appropriate due to the issue arising in that context.

jpechane:
@coreillylocus Thanks a lot. I am closing this PR and follow in https://github.com/debezium/debezium/pull/5420
I removed the merge commits and moved the docs from common fragment to SQL Server docs. The PR will be merged as soon as CI run is finished.

coreillylocus:
> @coreillylocus Thanks a lot. I am closing this PR and follow in #5420 I removed the merge commits and moved the docs from common fragment to SQL Server docs. The PR will be merged as soon as CI run is finished.

@jpechane sounds good. I got myself in a muddle with those merge commits, thanks for cleaning up.

PR #5227, opened 2024-02-02
DBZ-7441 Postpone SignalProcessor start after streaming is initialized

mfvitale:
This will avoid, for channels that not depends on the event streaming like the 'source' channel, to start processing signals before the IncrementalSnapshotChangeEventSource is initialized.

closes: https://issues.redhat.com/browse/DBZ-7441


jpechane (approved):

mfvitale:
Test failures are note related. Applied to main (2.6) and 2.5

PR #5228, opened 2024-02-03
DBZ-7443 [docs] correct debezium.sink.pubsub.flowcontrol.* variable names

CliffordCheefoon:
Correct configuration variables case discrepancy between docs and pubsub sink connector 

https://github.com/debezium/debezium-server/blob/df953233220e5250129135fabf1d97212d77ea40/debezium-server-pubsub/src/main/java/io/debezium/server/pubsub/PubSubChangeConsumer.java#L98C3-L98C88

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -580,3 +580,4 @@ Pavithrananda Prabhu
 Peter Hamer
 Artem Shubovych
 leoloel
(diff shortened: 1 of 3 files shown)


jpechane:
Applied via https://github.com/debezium/debezium/pull/5231 Thanks!

PR #5229, opened 2024-02-04
DBZ-7445 LogMiner batch size does not increase automatically

jchipmunk:
The condition for increasing the LogMiner batch size must depend on `log.mining.batch.size.max`, and not on `log.mining.batch.size.min`.

https://issues.redhat.com/browse/DBZ-7445

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
@@ -680,19 +680,21 @@ public class LogMinerStreamingChangeEventSource implements StreamingChangeEventS
     }
 
     private void updateBatchSize(boolean increment) {
-        if (increment && currentBatchSize < connectorConfig.getLogMiningBatchSizeMin()) {
-            currentBatchSize += connectorConfig.getLogMiningBatchSizeMin();
-            if (currentBatchSize == connectorConfig.getLogMiningBatchSizeMax()) {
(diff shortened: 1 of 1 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -680,7 +680,7 @@ public void endMiningSession(OracleConnection connection, OracleOffsetContext of
     }
 
     private void updateBatchSize(boolean increment) {
-        if (increment && currentBatchSize < connectorConfig.getLogMiningBatchSizeMin()) {
+        if (increment && currentBatchSize < connectorConfig.getLogMiningBatchSizeMax()) {
             currentBatchSize += connectorConfig.getLogMiningBatchSizeMin();
```
Naros, line 684:
```suggestion
        if (increment && currentBatchSize < connectorConfig.getLogMiningBatchSizeMax()) {
            currentBatchSize += connectorConfig.getLogMiningBatchSizeMin();
```
I suggest we also use `Math.min` on the increment line and insure that the batch size never exceeds the max.  This guarantees that the log entry is always processed properly. 

Additionally, I'd look at the `else if` condition for `< LogMiningBatchSize` where we decrement the batch size and likely use `Math.max` to handle the same but opposite scenario where the `currentBatchSize` never falls below the min.

Naros:
Test failure is unrelated.

Naros:
Applied to main (=2.6) and branch 2.5, thanks @jchipmunk .

PR #5230, opened 2024-02-04
DBZ-7446 Reduce string creation during SQL_REDO column read

jchipmunk:
Oracle Connector can create a new string when reading SQL_REDO column only if CSF column value is 1.

https://issues.redhat.com/browse/DBZ-7446


Naros (approved):
LGTM, let's just wait on CI.  Thanks @jchipmunk

Naros (changes requested):
Actually appears to be some test failures :(

Naros:
Test failure is unrelated.

Naros (approved):

PR #5237, opened 2024-02-06
DBZ-7451 Support for standard BSON representation of UUID in incremental snapshots

jcechace:
https://issues.redhat.com/browse/DBZ-7451

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/snapshot/MongoDbIncrementalSnapshotChangeEventSource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/snapshot/MongoDbIncrementalSnapshotChangeEventSource.java
@@ -23,8 +23,10 @@ import java.util.concurrent.atomic.AtomicReference;
 import java.util.stream.Collectors;
 
 import org.apache.kafka.connect.data.Struct;
+import org.bson.BsonBinarySubType;
 import org.bson.BsonDocument;
 import org.bson.Document;
+import org.bson.UuidRepresentation;
 import org.slf4j.Logger;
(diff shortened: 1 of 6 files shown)


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/snapshot/MongoDbIncrementalSnapshotChangeEventSource.java
```diff
@@ -638,6 +640,19 @@ private Object[] keyFromRow(Object[] row) {
             case STRING:
                 key = documentId.asString().getValue();
                 break;
+            case BINARY:
+                var subtype = documentId.asBinary().getType();
+                if (!BsonBinarySubType.isUuid(subtype)) {
+                    throw new IllegalStateException("Unsupported type of document id");
```
jpechane, line 646: @jcechace Would it be possible to handle unrecognized binary as either `byte[]` or String-encoded byte value?

jcechace: In theory yes, however I'm not sure how well that's gonna play with our snapshotting logic (e.g. determining the max/min key values and so on).

Theoretically we could even consider ignoring UUID and just treat everything as binary

jpechane: Fair point, we cannot make any assumptions around orderin so it is better to handle it case by case.

jpechane:
@jcechace Applied, thanks

PR #5239, opened 2024-02-07
DBZ-3401 Add new hybrid mining strategy

Naros:
https://issues.redhat.com/browse/DBZ-3401

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
@@ -398,11 +398,7 @@ public class OracleValueConverters extends JdbcValueConverters {
             return ((Double) data).floatValue();
         }
         else if (data instanceof String) {
-            final String s = (String) data;
-            if (isHexToRawFunctionCall(s)) {
-                data = String.valueOf(convertHexToRawFunctionToNumber(column, s));
-            }
-            return Float.parseFloat((String) data);
+            return Float.parseFloat(toStringFromNumericHexToRawIfApplicable(column, (String) data));
         }
 
         return super.convertFloat(column, fieldDefn, data);
@@ -419,11 +415,7 @@ public class OracleValueConverters extends JdbcValueConverters {
             }
         }
         else if (data instanceof String) {
-            final String s = (String) data;
-            if (isHexToRawFunctionCall(s)) {
-                data = String.valueOf(convertHexToRawFunctionToNumber(column, s));
-            }
-            return Double.parseDouble((String) data);
+            return Double.parseDouble(toStringFromNumericHexToRawIfApplicable(column, (String) data));
         }
 
         return super.convertDouble(column, fieldDefn, data);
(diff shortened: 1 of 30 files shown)


Naros:
- Debezium Server test failure unrelated.
- PostgreSQL test failure unrelated.

Naros:
Hi @jpechane, I am afraid I may have hit a roadblock for this change.  

While looking into how to support LOB columns this weekend, I found when a LOB column's value exceeds 4000 bytes and Oracle swaps to using chunk-based events, we are not supplied any reference we can use to deduce the column the byte array stream is associated with. Unfortunately, this entire PR is contingent on Oracle providing us the `COL ##` reference. Without that detail on the `SEL_LOB_LOCATOR` operation, we cannot determine which column the provided byte array stream is associated with.

I've thought about whether it makes sense to drop `lob.enabled` entirely.  I'm not sure I'm necessarily convinced to do that yet, but doing so would simplify the connector's code if we simply relied on the re-select columns post-processor for users who want LOB data.  The only real question would be whether we can identify the one use case where there are additional `UPDATE` statements that perform LOB mutations in a non-chunk-based way.

Obviously, the downside to relying on the re-select column post-processor is that it's an additional query per change event, and if you're streaming a table with lots of changes that contain LOB columns, there could be some performance concerns.

Lets sync-up on Monday and discuss this in more detail as I had wanted to get this PR wrapped up, but this really throws a wrench into how to move forward.

Naros:
Moving this back to draft state, see comments on the related Jira.

Naros:
Looks like something with recent `main` changes before this PR, looking into it :(

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
```diff
@@ -391,6 +398,10 @@ else if (data instanceof Double) {
             return ((Double) data).floatValue();
         }
         else if (data instanceof String) {
+            final String s = (String) data;
+            if (isHexToRawFunctionCall(s)) {
```
jpechane, line 402: Is it necessary to have condition here and in other instances? Can't the methods be always called and the check and branching done inside of it?

jpechane (commented):
@Naros LGTM, thanks! I left one question related to haw to raw handling. Other than that I belive we are good to go.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
Naros: One thing I'd like to rework eventually is a different value converter per adapter.  We had talked about this with OpenLogReplicator and it would certainly allow a really good way to overlay specific conversion steps based on another or even configuration within an adapter in a more OO-way.

Naros: Looking at this more closely, I think we could consolidate many of these, but I don't think it will be universal without potentially looking at using some type of lambda which would be an extra object allocation on the stack for each column converted.  For example, in `convertIntervalYearMonth`, where if it's a hex-to-raw function, we use the return value as-is, but instead, we do some other manipulation since the value is using an alternative format.  The ones where we essentially assign the return value to `data` and pass it through the remainder of the converter method could be changed though.  Is that sufficient you think?

Naros: I added some changes in 5de55df, let me know if that suffices or if I am off base.

jpechane (approved):
@Naros that's exactly I was looking for, thanks. I agree that there is a space for additional refactoring but it is not a pressing one. Please feel free to merge yourself if the test failure is unrelated or when it is resolved.

PR #5240, opened 2024-02-08
DBZ-7456 Skip PK change LOB column reselect if column is excluded

Naros:
https://issues.redhat.com/browse/DBZ-7456

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/BaseChangeRecordEmitter.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/BaseChangeRecordEmitter.java
@@ -8,7 +8,7 @@ package io.debezium.connector.oracle;
 import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.SQLException;
-import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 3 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/BaseChangeRecordEmitter.java
```diff
@@ -119,9 +119,12 @@ protected void emitUpdateAsPrimaryKeyChangeRecord(Receiver receiver, TableSchema
     private List<Column> getReselectColumns(Struct newValue) {
         final List<Column> reselectColumns = new ArrayList<>();
```
jchipmunk, line 120: My comment is not related to this fix, but it seems to me that here is no need to create a new `ArrayList` if there are no CLOB/BLOB columns in the table. This can be easily avoided by adding the following check at the beginning:
```
final List<Column> lobColumns = schema.getLobColumnsForTable(table.id());
if (lobColumns.isEmpty()) {
    return Collections.emptyList();
}
final List<Column> reselectColumns = new ArrayList<>();
for (Column column : lobColumns) {
    ...
```

Naros: Great catch, ya this is still in the works as I need to add a test case, but I'll adjust this too; :+1:

PR #5242, opened 2024-02-08
DBZ-7416 Fix duplicate SMTs sometimes returned by Kafka Connect

rk3rn3r:
Added deduplication with a Map with a key based on the SMT's fully-qualified class name and version.

closes https://issues.redhat.com/browse/DBZ-7416

### DIFF
--- a/debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/DebeziumResource.java
+++ b/debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/DebeziumResource.java
@@ -11,6 +11,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


rk3rn3r:
@indraraj Mind testing this branch if it still returns duplicates for SMTs?

debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/DebeziumResource.java
```diff
@@ -92,20 +93,20 @@ else if (m.lookingAt()) {
...
     private synchronized void initConnectorPlugins() {
         if (null == this.availableConnectorPlugins || this.availableConnectorPlugins.isEmpty()) {
             // TODO: improve once plugins are allowed to be added/removed during runtime by Kafka Connect, @see org.apache.kafka.connect.runtime.rest.resources.ConnectorPluginsResource
-            final List<ConnectorDescriptor> connectorPlugins = new ArrayList<>();
+            final Map<String, ConnectorDescriptor> connectorPlugins = new HashMap<>();
             Herder herder = getHerder();
             addConnectorPlugins(connectorPlugins, herder.plugins().sinkConnectors());
             addConnectorPlugins(connectorPlugins, herder.plugins().sourceConnectors());
-            this.availableConnectorPlugins = Collections.unmodifiableList(connectorPlugins);
+            this.availableConnectorPlugins = Collections.unmodifiableList(new ArrayList<>(connectorPlugins.values()));
```
Naros, line 109: Would it make more sense to leverage a `Set<ConnectorDescriptor>` with a `Comparable` interface to control de-duplication rather than using a Map?

rk3rn3r: Wouldn't that require iterating the whole Set in case of adding an item that is not yet in the list? Then this temporary map is the fastest way, no?

mfvitale: It depends on the `Set` implementation used. `HashSet`, for example,  uses `HashMap` internally to store it’s elements and  offers performance of order O(1) for insertion, removal and retrieval operations. The same performance applies to `LinkedHashSet` but permits to maintain the inserted order. If you instead need to maintain a certain order defined by a `Comparator`, you need to use the `TreeSet` that has gives performance of order O(log(n)) for insertion, removal and retrieval operations. 

In that case I think you can go with `HashSet`.

Naros: Thanks @mfvitale.

Given that you're effectively using the `Map` in this case like `HashSet` does under the hood, that's precisely why I wonder why we simply didn't opt for `Set` in this case as it would have been a much cleaner way to implement the behavior without any meaningful loss in performance @rk3rn3r.

rk3rn3r: Ah, for sure. As you know from the last team call, I had only last week and just did it quickly to fix it in the allowed time frame of 1 week. I also created the new, working, slim UI container image and didn't give this not so much thought as usual here, but for sure, this is correct. I think `LinkedHashSet` is even a good way to maintain ordering and with some time now I should have chosen this one.

PR #5232, opened 2024-02-05
DBZ-7302 Implement Snapshotter SPI Oracle

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7302

The PR is based on https://github.com/debezium/debezium/pull/5196, must be rebased after its merge.

- [ ]  Implement the Snapshotter interface
- [x]  Implement the SnapshotLock interface
- [x]  Implement the SnapshotQuery interface
- [ ]  Add test for a custom Snapshotter

PR #5245, opened 2024-02-09
DBZ-7461 Align snapshot modes for PostgreSQL, MySQL, Oracle

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7461

The PR is based on https://github.com/debezium/debezium/pull/5232

Scope of the PR is starting to support all possible snapshot modes across PostgreSQL, MySQL and Oracle. 

- [x] Support `Always` for MySQL 
- [x] Support `When_needed` for Oracle
- [x] Support `When_needed` for PostgreSQL
- ~[ ] Support `Never` for Oracle~
- [x] Oracle - Add generic properties `archive.log.hours` and `archive.destination.name` used to check the availability of the stored offset on the db log. 
- [x] Oracle: Add a warn in case there are multiple valid archive log destinations for Xstream and OpenLogReplicator


mfvitale:
I'll provide pr for updating the docs.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -350,6 +342,34 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
...
                     "When not set, the connector will automatically select the first LOCAL and VALID destination.");
 
+    public static final Field ARCHIVE_DESTINATION_NAME = Field.create("archive.destination.name")
+            .withDisplayName("Name of the archive log destination to be used for reading archive logs")
+            .withType(Type.STRING)
+            .withWidth(Width.MEDIUM)
+            .withImportance(Importance.LOW)
+            .withGroup(Field.createGroupEntry(Field.Group.CONNECTION_ADVANCED, 32))
+            .withDescription("Sets the specific archive log destination as the source for reading archive logs." +
+                    "When not set, the connector will automatically select the first LOCAL and VALID destination.");
```
Naros, line 352: We probably should make sure this gets highlighted in the release notes. For XStream and OpenLogReplicator users, they will not have defined this value at all and in most production environments, this will be required to avoid potential a miscalculation.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
```diff
@@ -97,13 +97,20 @@ public ChangeEventSourceCoordinator<OraclePartition, OracleOffsetContext> start(
...
 
-        OraclePartition partition = previousOffsets.getTheOnlyPartition();
         OracleOffsetContext previousOffset = previousOffsets.getTheOnlyOffset();
 
-        validateAndLoadSchemaHistory(connectorConfig, partition, previousOffset, schema, snapshotterService);
+        validateAndLoadSchemaHistory(connectorConfig, jdbcConnection, previousOffsets, schema, snapshotterService.getSnapshotter());
 
         taskContext = new OracleTaskContext(connectorConfig, schema);
 
+        // If the binlog position is not available it is necessary to re-execute snapshot
```
Naros, line 106:
```suggestion
        // If the redo log position is not available it is necessary to re-execute snapshot
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
```diff
@@ -97,13 +97,20 @@ public ChangeEventSourceCoordinator<OraclePartition, OracleOffsetContext> start(
...
+        if (previousOffset == null) {
+            LOGGER.info("No previous offset found");
+        }
+        else {
+            LOGGER.info("Found previous offset {}", previousOffset);
+        }
```
Naros, line 112: We seem to log "No previous offset found" in a few places in the Oracle connector, are all these log references necessary?  See OracleSnapshotChangeEventSource.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -402,13 +405,15 @@ private LogMinerEventProcessor createProcessor(ChangeEventSourceContext context,
...
+
+        final String oldestFirstChangeQuery = SqlUtils.oldestFirstChangeQuery(archiveLogRetention, archiveDestinationName);
+        final String oldestScn = connection.singleOptionalValue(oldestFirstChangeQuery, rs -> rs.getString(1));
         if (oldestScn == null) {
-            throw new DebeziumException("Failed to calculate oldest SCN available in logs");
+            return Optional.empty();
         }
         LOGGER.trace("Oldest SCN in logs is '{}'", oldestScn);
-        return Scn.valueOf(oldestScn);
+        return Optional.of(Scn.valueOf(oldestScn));
```
Naros, line 416: I believe you moved this to `OracleConnection`, can this be removed?

Naros (commented):
Hi @mfvitale, just an initial review for the Oracle-related changes.

Naros:
I've pinged Ondrej about the Oracle 19 images wrt 21.11 libs and the `21.3.0-xs` image not being baked.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
mfvitale: Well, I was tempted to put it as `required` but I was in doubt. Do you think is it an option?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
mfvitale: Just removed the log on *SnapshotChangeEventSource

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
Naros: I would not set it as required because the connector is smart enough to resolve the value if only 1 VALID and LOCAL log destination is configured.  

For OpenLogReplicator and XStream users, they'll need to check their `V$ARCHIVED_DEST_STATUS` table and confirm if there are multiple rows that match `STATUS=VALID and TYPE=LOCAL`.  If only a single row, they can safely leave this option as the default. If they have multiple, they'll need to confer with the DBA to know which configured destination should be used.

mfvitale: Or should we do this check https://github.com/debezium/debezium/blob/b6e45beab1d04683d1ae11bc1879770f936b9f0d/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java#L932 also for XStream and OpenLogReplicator?

mfvitale: > We could; however, add this to the Snapshotter code as a sanity check before fetching the oldest available SCN from the logs, wdyt?

This is what I meant.

mfvitale:
Oracle XStream test failures seems related. Checking.

mfvitale:
@jpechane Failure are already marked as `flaky`. So I think we can merge after the review.

Naros (commented):
Rechecked the Oracle changes :+1: on those from my side.

PR #5251, opened 2024-02-10
Adjust CI Flaky Test Support

Naros:


jpechane:
@Naros Is this WIP or can be merged? If the latter then please feel free to do it.

Naros:
Hi @jpechane I converted it back to a draft.  I need to review the current state of all the changes around the Async Engine and our recent Db2 changes to see if any of these are still relevant.  I'll adjust accordingly.

mfvitale (approved):

Naros:
Thanks @jpechane and @mfvitale, merged.

.github/workflows/debezium-workflow.yml
```diff
@@ -813,6 +820,7 @@ jobs:
           -Dorg.slf4j.simpleLogger.log.org.apache.maven.cli.transfer.Slf4jMavenTransferListener=warn
           -Dmaven.wagon.http.pool=false
           -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
+          -Dfail.flaky.tests=false
```
jpechane, line 823: @Naros Is this correct here?

Also I am afraid we have know workflows that use `fail.flaky.tests` control flag - like Db2 and maybe others. We need to align on the same behaviour.

debezium-parent/pom.xml
```diff
@@ -207,6 +208,7 @@
                     <systemPropertyVariables combine.children="append">
                         <java.io.tmpdir>${project.build.directory}</java.io.tmpdir>
                         <skipLongRunningTests>${skipLongRunningTests}</skipLongRunningTests>
+                        <failFlakyTests>${failFlakyTests}</failFlakyTests>
```
jpechane, line 211: @Naros Should not be failsafe plugin configured too if we want to make it cetralized?

PR #5252, opened 2024-02-11
DBZ-7473 Defer transaction capture until the first DML event occurs

jchipmunk:
Do not handle start event from V$LOGMNR_CONTENTS view if CLOB/BLOB support is disabled and memory buffer is used.

https://issues.redhat.com/browse/DBZ-7473

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryBuilder.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryBuilder.java
@@ -144,7 +144,13 @@ public class LogMinerQueryBuilder {
             operationInClause.withValues(OPERATION_CODES_LOB);
         }
         else {
-            operationInClause.withValues(OPERATION_CODES_NO_LOB);
+            final List<Integer> operationCodes = new ArrayList<>(OPERATION_CODES_NO_LOB);
+            // The transaction start event needs to be handled when a persistent buffer (Infinispan) is used
+            // because it is needed to reset the event id counter when re-mining transaction events.
+            if (connectorConfig.getLogMiningBufferType() == OracleConnectorConfig.LogMiningBufferType.MEMORY) {
+                operationCodes.removeIf(operationCode -> operationCode == 6);
+            }
+            operationInClause.withValues(operationCodes);
         }
         predicate.append("(").append(operationInClause.build());
 
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
@@ -444,50 +444,45 @@ public abstract class AbstractLogMinerEventProcessor<T extends AbstractTransacti
 
         final T transaction = getAndRemoveTransactionFromCache(transactionId);
         if (transaction == null) {
+            LOGGER.debug("Transaction {} not found in cache, no events to commit.", transactionId);
             handleCommitNotFoundInBuffer(row);
-            LOGGER.debug("Transaction {} not found, commit skipped.", transactionId);
-            return;
         }
 
-        // Calculate the smallest SCN that remains in the transaction cache
-        final Optional<T> oldestTransaction = getOldestTransactionInCache();
-        final Scn smallestScn;
-        if (oldestTransaction.isPresent()) {
-            smallestScn = oldestTransaction.get().getStartScn();
-            metrics.setOldestScnDetails(smallestScn, oldestTransaction.get().getChangeTime());
-        }
-        else {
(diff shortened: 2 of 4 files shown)


Naros:
Hi @jchipmunk, thanks for the PR, but I need to be convinced this is the right approach to solve the issue.  If you have long-running transactions on tables you're not interested in; while this may help with that, as soon as you get a long-running transaction on a table that is captured, we're back to square one.

jchipmunk:
> but I need to be convinced this is the right approach to solve the issue. If you have long-running transactions on tables you're not interested in; while this may help with that, as soon as you get a long-running transaction on a table that is captured, we're back to square one.

Hi @Naros, Yes, I understand that in this case we get the same behavior as before, but then the offset SCN will only get stuck when it is really needed. In my case, long transactions are regularly running on the database that have nothing to do with the tables that interest me.

Regarding this PR, I think that the current changes are not enough and the commit event handling code also needs to be changed.

Naros:
> Hi @Naros, Yes, I understand that in this case we get the same behavior as before, but then the offset SCN will only get stuck when it is really needed. In my case, long transactions are regularly running on the database that have nothing to do with the tables that interest me.

Understandable, then I would likely suggest rephrasing the commits and likely the Jira to suggest that instead we're going to delay capturing transactions until the first DML event for a transaction rather than giving any impression that the change is addressing the "stuck SCN" issue as technically it isn't.

> Regarding this PR, I think that the current changes are not enough and the commit event handling code also needs to be changed.

That's fair; I suspect there will definitely be some logging concerns as we added some to identify when we detect a commit or a rollback without a matching transaction in the buffer, IIRC, but otherwise, I would suspect that most of the code should be fairly flexible here.  In fact, before adding LOB support, we actually started transactions lazily when we identified the first DML event.

jchipmunk:
Hi @Naros 

Some Oracle test runs failed with the following error:
```
Database deployed
++ echo 'Installing all necessary drivers'
Installing all necessary drivers
++ pushd /root/oracle-libs/21.11.0.0.0
/root/install-oracle-driver.sh: line 20: pushd: /root/oracle-libs/21.11.0.0.0: No such file or directory
```
should I merge the commit https://github.com/debezium/debezium/commit/2af17ac69c19c675dc2627b1eca961a9be7e667d to my branch?

Naros:
Hi @jchipmunk, a rebase of the PR should be sufficient I believe.

jchipmunk:
> /packit test --labels oracle

@Naros, could you please run the Oracle tests again?

Naros:
> > /packit test --labels oracle
> 
> @Naros, could you please run the Oracle tests again?

I can @jchipmunk, just waiting on the images to be re-baked.  We made some small changes and some of the images are not working correctly.  Hope to have that resolved today.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryBuilder.java
```diff
@@ -144,7 +144,13 @@ private static String getOperationCodePredicate(OracleConnectorConfig connectorC
...
         }
         else {
-            operationInClause.withValues(OPERATION_CODES_NO_LOB);
+            final List<Integer> operationCodes = new ArrayList<>(OPERATION_CODES_NO_LOB);
+            // The transaction start event needs to be handled when a persistent buffer (Infinispan) is used
+            // because it is needed to reset the event id counter when re-mining transaction events.
+            if (connectorConfig.getLogMiningBufferType() == OracleConnectorConfig.LogMiningBufferType.MEMORY) {
+                operationCodes.removeIf(operationCode -> operationCode == 6);
+            }
+            operationInClause.withValues(operationCodes);
```
Naros, line 153: So IIRC, the connector works the same regardless if Infinispan or Memory-based buffers are used, we always restart the mining process on connector restarts from the offset's `scn` value.  So theoretically, the fact that Infinispan is persistent is primarily irrelevant in that case.

This raises a good question as to whether we should technically just treat Infinispan caches as off-heap data stores that do not persist across restarts.  If we did that, this would, therefore, be identical, no?

Naros (commented):
Overall this LGTM, just have one question inline below

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerQueryBuilder.java
jchipmunk: Yes, I believe that then the behavior will be identical and there will be no need to process the transaction start event.

Naros: Ok let me check with the one user I know that heavily relies on ISPN.

Naros (approved):
Double checked both Infinispan and non-Infinispan and found a small bug that is unrelated to this change for the Infinispan Buffer, see https://github.com/debezium/debezium/pull/5352; otherwise LGTM.

PR #5253, opened 2024-02-12
DBZ-7479 PreparedStatement leak in Oracle ReselectColumnsProcessor

jchipmunk:
Each time, Oracle connector creates a new instance of PreparedStatement because value of commit SCN is added directly to SQL query to reselect column values.

https://issues.redhat.com/browse/DBZ-7479

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
@@ -573,32 +573,34 @@ public class OracleConnection extends JdbcConnection {
     }
 
     @Override
-    public String buildReselectColumnQuery(TableId tableId, List<String> columns, List<String> keyColumns, Struct source) {
-        final String commitScn = source.getString(SourceInfo.COMMIT_SCN_KEY);
+    public Map<String, Object> reselectColumns(TableId tableId, List<String> columns, List<String> keyColumns, List<Object> keyValues, Struct source)
+            throws SQLException {
         final TableId oracleTableId = new TableId(null, tableId.schema(), tableId.table());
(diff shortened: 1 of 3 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
```diff
@@ -579,16 +580,25 @@ public String buildReselectColumnQuery(TableId tableId, List<String> columns, Li
...
+        final String commitScn = source.getString(SourceInfo.COMMIT_SCN_KEY);
+        final List<Object> values;
+        if (Strings.isNullOrEmpty(commitScn)) {
+            values = bindValues;
+        }
+        else {
+            values = new ArrayList<>(bindValues.size() + 1);
+            values.add(commitScn);
+            values.addAll(bindValues);
+        }
```
Naros, line 600: Hi @jchipmunk, good call.  I suppose the only thing I particularly don't like is the notion that we use COMMIT_SCN_KEY in both methods now to so very similar things, but I'm not sure we can clean this up anymore without there being a larger refactor involved, so likely best to leave as-is for now.

jchipmunk: Hi @Naros, I agree, but did not dare to change the current methods much.

jchipmunk:
@Naros Tests that fail are not related to the Oracle connector.

Naros (approved):
Thanks @jchipmunk, LGTM.  I believe the test failures are unrelated and are being fixed in another PR due to some recent changes with the embedded engine.

Naros:
Applied to main (=2.6) and branch 2.5.  Thanks again @jchipmunk .

PR #5254, opened 2024-02-12
DBZ-7481 SnapshotterServiceProvider will check if snapshot mode class is related to the running connector.

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7481


jpechane:
@mfvitale Applied, thanks. Still I feel this is a little bit hack solution so I'd prefer to come come with something cleaner.

PR #5261, opened 2024-02-14
DBZ-7107 Introduce micro/nano second based envelope timestamps

Naros:
https://issues.redhat.com/browse/DBZ-7107

- [ ] Update documentation
- [x] Create sibling PRs for Informix, Db2, Vitess, Spanner, and Cassandra.

@jpechane once we agree on the implementation, I'll address the documentation.

### DIFF
--- a/debezium-core/src/main/java/io/debezium/schema/SchemaFactory.java
+++ b/debezium-core/src/main/java/io/debezium/schema/SchemaFactory.java
@@ -365,6 +365,8 @@ public class SchemaFactory {
             public Envelope build() {
                 builder.field(Envelope.FieldName.OPERATION, Envelope.OPERATION_REQUIRED ? Schema.STRING_SCHEMA : Schema.OPTIONAL_STRING_SCHEMA);
                 builder.field(Envelope.FieldName.TIMESTAMP, Schema.OPTIONAL_INT64_SCHEMA);
+                builder.field(Envelope.FieldName.TIMESTAMP_US, Schema.OPTIONAL_INT64_SCHEMA);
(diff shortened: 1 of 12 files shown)


debezium-core/src/main/java/io/debezium/schema/SchemaFactory.java
```diff
@@ -365,6 +365,8 @@ public Envelope.Builder withDoc(String doc) {
             public Envelope build() {
                 builder.field(Envelope.FieldName.OPERATION, Envelope.OPERATION_REQUIRED ? Schema.STRING_SCHEMA : Schema.OPTIONAL_STRING_SCHEMA);
                 builder.field(Envelope.FieldName.TIMESTAMP, Schema.OPTIONAL_INT64_SCHEMA);
+                builder.field(Envelope.FieldName.TIMESTAMP_US, Schema.OPTIONAL_INT64_SCHEMA);
```
jpechane, line 368: Should we update the envelope schema version?

jpechane (commented):
@Naros LGTM, thanks. But I wonder if we should make it symmetric with `source` info block too. Some of the sources also provide better precisision then just milliseconds.

Naros:
All good points, I'll amend. :+1:

Naros:
The test failures related to PostgreSQL, SQL Server, and Debezium Server are unrelated.  The failures related to Db2, Spanner, Informix, and Vitess have been fixed in separate pull requests.

jpechane:
@Naros Applied, thanks!

jpechane:
@Naros Please provide docs update as a separate PR

PR #5262, opened 2024-02-14
DBZ-7490 Add Informix to Debezium CI workflow

Naros:
https://issues.redhat.com/browse/DBZ-7490

### DIFF
--- a/.github/workflows/debezium-workflow.yml
+++ b/.github/workflows/debezium-workflow.yml
@@ -814,6 +814,65 @@ jobs:
           -Dmaven.wagon.http.pool=false
           -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
           
+  build_informix:
+    needs: [check_style, file_changes]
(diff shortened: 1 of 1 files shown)


Naros:
@jpechane it appears the two Informix builds on its own repository aren't yet stable, so I'm not sure we want to necessarily add this to the mix just yet in the main repository.  Have you reached out to Lars about Informix or is he aware of the test suite's current state?

jpechane:
The Informix tests were fairly stable with occasionaly single IT test failur, nothing at this scale. It is possible that the root cause for massive failuer lies in other changes. It is still important to ensure the connector is aligned with the latest main so i pushed a commit that will just disable ITs. The issue will stay open till the ITs are enable again.

nrkljo:
That does look strange - I'll have a look...

jpechane:
@nrkljo Please hang on, https://github.com/debezium/debezium/pull/5274 is probably the root cause

PR #5268, opened 2024-02-15
DBZ-7488 Skip re-selection on `r` (read) events

Naros:
https://issues.redhat.com/browse/DBZ-7488


jpechane:
@Naros  Applied, thanks!

PR #5272, opened 2024-02-15
DBZ-7489 Correctly parse XML preamble SQL ending with "IS NULL"

Naros:
https://issues.redhat.com/browse/DBZ-7489


Naros:
Test failures are all related to the Async Embedded Engine default timeouts.

jpechane (approved):
@Naros LGTM thanks! Let's wait wit merge till the management timeout issue in testsuite is solved.

jpechane:
@Naros Applied, thanks! I think is safe to go ahead with it.

PR #5274, opened 2024-02-16
DBZ-7496 Make sure completion callback is called after engine is stopped

vjuranek:
https://issues.redhat.com/browse/DBZ-7496
https://issues.redhat.com/browse/DBZ-7495

### DIFF
--- a/debezium-embedded/src/test/java/io/debezium/embedded/async/AsyncEmbeddedEngineTest.java
+++ b/debezium-embedded/src/test/java/io/debezium/embedded/async/AsyncEmbeddedEngineTest.java
@@ -559,6 +559,74 @@ public class AsyncEmbeddedEngineTest {
         runEngineBasicLifecycleWithConsumer(props);
     }
 
+    @Test
+    @FixFor("DBZ-7496")
+    public void testCompletionCallbackCalledAfterConnectorStop() throws Exception {
+        final Properties props = new Properties();
+        props.setProperty(ConnectorConfig.NAME_CONFIG, "debezium-engine");
+        props.setProperty(ConnectorConfig.TASKS_MAX_CONFIG, "1");
+        props.setProperty(ConnectorConfig.CONNECTOR_CLASS_CONFIG, FileStreamSourceConnector.class.getName());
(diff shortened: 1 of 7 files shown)


debezium-embedded/src/test/java/io/debezium/embedded/async/AsyncEmbeddedEngineTest.java
```diff
@@ -612,15 +681,15 @@ protected void waitForEngineToStart() {
         Awaitility.await()
                 .alias("Engine haven't started on time")
                 .pollInterval(TEST_TASK_MANAGEMENT_TIMEOUT_MS, TimeUnit.MILLISECONDS)
-                .atMost(1, TimeUnit.SECONDS)
+                .atMost(5 * TEST_TASK_MANAGEMENT_TIMEOUT_MS, TimeUnit.SECONDS)
```
jpechane, line 684: @vjuranek  We have {{debezium.test.records.waittime}} sys property. Maybe we can intorduce a new one or re-use it?
I can imagine that different connector will have different timout needs.

vjuranek: good idea, added `debezium. test.engine.waittime`

vjuranek:
Sorry for mixing these two issue, I'd like to see how many/which tests fail after applying both these patches.

jpechane:
@vjuranek Looks much better now, thanks!

cy-phiz:
Hi @vjuranek 

I'm sorry for bringing this up, but recently I encountered an issue with Debezium Server version 2.3.0 that might be related to this PR. I'm wondering if you could enlighten me a bit.

What I bumped into is: The server did not actually stop after logging:

```
Received request to stop the engine
Stopping the embedded engine
```

I expected another line to appear, but it did not show:

```
debezium-server-dist stopped in 0.XXXs
```
It was just hanging there instead.

I’m wondering if this PR might be the fix for my situation?
Any help would be greatly appreciated!

vjuranek:
hi @cy-phiz , if it happens again (or you can reproduce), please take a thread dump (see e.g. [here](https://www.baeldung.com/java-thread-dump)) and create an issue under https://issues.redhat.com/browse/DBZ . Thanks!

PR #5281, opened 2024-02-17
Fix MySQL 8 event timestamp resolution logic error where fallback to seconds occurs erroneously for non-GTID events

methodmissing:
Jira issue: https://issues.redhat.com/browse/DBZ-7500

Fixes: https://github.com/debezium/debezium/pull/5036

Applied to a MySQL 8 connector:

Screenshot 2024-02-16 at 23 41 22

Proof of millisecond resolution `source.ts_ms`:

```json
"source": {
    "query": "REDACTED",
    "thread": 3683566,
    "server_id": 12844,
    "version": "2.6.0-SNAPSHOT",
    "sequence": "",
    "file": "binlog.726557",
    "connector": "mysql",
    "pos": 59658237,
    "name": "REDACTED",
    "gtid": "6aafb3ca-9b99-11ee-a237-4201f300802c:187141265",
    "row": 0,
    "ts_ms": 1708126035867,
    "snapshot": "false",
    "db": "REDACTED",
    "table": "products"
  },
```

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java
@@ -229,16 +229,22 @@ public class MySqlStreamingChangeEventSource implements StreamingChangeEventSour
     private void setEventTimestamp(Event event, long eventTs) {
         // Prefer higher resolution replication timestamps from MySQL 8 GTID events, if possible
         if (isGtidModeEnabled) {
-            if (event.getHeader().getEventType() == EventType.GTID) {
-                GtidEventData gtidEvent = unwrapData(event);
-                final long gtidEventTs = gtidEvent.getOriginalCommitTimestamp();
-                if (gtidEventTs != 0) {
-                    // >= MySQL 8.0.1, prefer the higher resolution replication timestamp
-                    eventTimestamp = Instant.EPOCH.plus(gtidEventTs, ChronoUnit.MICROS);
-                }
-                else {
-                    // Fallback to second resolution event timestamps
-                    eventTimestamp = Instant.ofEpochMilli(eventTs);
+            if (connection.isMariaDb()) {
+                // Fallback to second resolution event timestamps
+                eventTimestamp = Instant.ofEpochMilli(eventTs);
+            }
+            else {
+                if (event.getHeader().getEventType() == EventType.GTID) {
+                    GtidEventData gtidEvent = unwrapData(event);
+                    final long gtidEventTs = gtidEvent.getOriginalCommitTimestamp();
+                    if (gtidEventTs != 0) {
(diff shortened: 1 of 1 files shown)


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlStreamingChangeEventSource.java
```diff
@@ -229,16 +229,22 @@ protected void onEvent(MySqlOffsetContext offsetContext, Event event) {
...
-                    // >= MySQL 8.0.1, prefer the higher resolution replication timestamp
-                    eventTimestamp = Instant.EPOCH.plus(gtidEventTs, ChronoUnit.MICROS);
-                }
-                else {
-                    // Fallback to second resolution event timestamps
-                    eventTimestamp = Instant.ofEpochMilli(eventTs);
+            if (connection.isMariaDb()) {
+                // Fallback to second resolution event timestamps
+                eventTimestamp = Instant.ofEpochMilli(eventTs);
+            }
```
Naros, line 235: I'm not extremely happy with using this here; however, since the adapter does not provide access to the unwrap method for the event data, it's probably the cleanest solution for now and keeps the change localized, so I think it's fine until I refactor this in 2.7.  :+1:

methodmissing: Agree. Also possible to use the `MARIADB_GTID` event type. I looked at metadata exposed by [MariaDB GTID events](https://mariadb.com/kb/en/gtid_event/) and its quite sparse, unfortunately.

```java
if (event.getHeader().getEventType() == EventType.MARIADB_GTID) {
    // Fallback to second resolution event timestamps
    eventTimestamp = Instant.ofEpochMilli(eventTs);
}
```

methodmissing: 💯 

> That said, I wonder if this method would be better written this way, it seems more concise in terms of logic and self documenting in my opinion, wdyt:

How do you prefer to move forward? Apply direct or want me to commit the suggestion?

Naros: Please go ahead and if you have the cycles @methodmissing

methodmissing: Applied in https://github.com/debezium/debezium/pull/5281/commits/853f39b43c1e78c022444036199c8ffca9e66d4c and piped though a lower environment - no regression:

Screenshot 2024-02-17 at 15 05 37

Naros (approved):

Naros:
Thanks @methodmissing, applied to main (=2.6) and branch 2.5.

methodmissing:
Amazing turn around and improvement proposal, thank you!

PR #5282, opened 2024-02-17
DBZ-7501 Remove incubating from Debezium documentation

nancyxu123:
Remove incubating from Debezium documentation since the connector has been launched for quite some time now.


Naros (approved):
Thanks @nancyxu123 !

PR #5284, opened 2024-02-17
DBZ-7364 Distribute Oracle ojdbc8 with Oracle connector

Naros:
https://issues.redhat.com/browse/DBZ-7364


jpechane:
@Naros Applied, thanks

PR #5285, opened 2024-02-18
DBZ-7480 Allow special characters in signal table name

harveyyue:
https://issues.redhat.com/browse/DBZ-7480


BenMotts:
Thanks for the solution @harveyyue! Looks like I'm too late to this, but is there no concern that parsing the signal table config value into a TableId will be inefficient for every record?
Would it make more sense to parse into a variable as part of class construction?

PR #5287, opened 2024-02-18
DBZ-6679 Only advance LogMiner start position when query has data

Naros:
https://issues.redhat.com/browse/DBZ-6679


jpechane:
Rebased on the latest `main` with stabilized tests

jpechane:
@Naros Applied to main and 2.5. It does not cpply cleanly on 2.3.

PR #5290, opened 2024-02-18
DBZ-7505 Support Oracle 23 SELECT without FROM clause

Naros:
https://issues.redhat.com/browse/DBZ-7505

Upstream PR - https://github.com/antlr/grammars-v4/pull/3972


jpechane:
@Naros Applied, thanks

PR #5291, opened 2024-02-18
DBZ-7506 Support Oracle 23 Annotations on CREATE/ALTER TABLE

Naros:
https://issues.redhat.com/browse/DBZ-7506


jpechane:
@Naros Applied, thanks

PR #5295, opened 2024-02-19
DBZ-7510 Add Informix to Java Outreach

Naros:
https://issues.redhat.com/browse/DBZ-7510


Naros:
@jpechane I spoke with Sanne and he suggested we treat setting the ByteBuddy flag to experimental to see whether we can get past the ByteBuddy failures on Java 23 and determine if there are any other dependencies that are not yet compatible.

Naros:
Will re-run the jobs later, GH actions are currently saturated and this can wait.

jpechane:
@Naros Informix is passing, applied, thanks!

PR #5301, opened 2024-02-20
DBZ-7508 Exit from readChunk after createDataEventsForTable if snapshot is not running anymore

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7508


mfvitale:
Postgres test failures is not related. Fixed with https://github.com/debezium/debezium/pull/5303

mfvitale:
All DB2 test are already marked as `flaky` but dunno why is still reported. @Naros Any idea?

jpechane:
It seems that the workflow in core do not have flag `-Dfail.flaky.tests=false` set.

Also I completely forgot, the default value for Flaky test reattempt is just `1`. Maybe we should change it?

jpechane:
@mfvitale Yes, please see `io.debezium.junit.ConditionalFail.ignoreFlakyFailure(Statement, Description, Flaky)`

rk3rn3r (approved):

rk3rn3r:
Thx @mfvitale for the PR. I just hit the same error and found 3 issues (one of those is this DBZ-7508) in Jira and linked them and closed the duplicate. Thx for the fix.

Naros:
Applied to main (=2.6).

PR #5302, opened 2024-02-20
DBZ-7521 Promote JDBC sink from Incubating

Naros:
https://issues.redhat.com/browse/DBZ-7521


roldanbob (approved):
LGTM. For the productized version, figuring that the DP designation was not long for this world, I only added a DP designator to the files in the downstream repo, not in the Debezium repo, so no change required here to reflect the change in status.

PR #5306, opened 2024-02-20
DBZ-7530 DBZ-7534 DBZ-7453 Stablize PG tests

Naros:
https://issues.redhat.com/browse/DBZ-7530

@jpechane here is what I came up with and it seems stable across Parallel and Non-Parallel snapshots

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
@@ -231,7 +231,7 @@ public class EventDispatcher<P extends Partition, T extends DataCollectionId> im
     }
 
     public SnapshotReceiver<P> getSnapshotChangeEventReceiver() {
-        return new BufferingSnapshotChangeRecordReceiver();
+        return new BufferingSnapshotChangeRecordReceiver(connectorConfig.getSnapshotMaxThreads() > 1);
     }
 
     public SnapshotReceiver<P> getIncrementalSnapshotChangeEventReceiver(DataChangeEventListener<P> dataListener) {
@@ -512,6 +512,11 @@ public class EventDispatcher<P extends Partition, T extends DataCollectionId> im
     private final class BufferingSnapshotChangeRecordReceiver implements SnapshotReceiver<P> {
 
         private AtomicReference<BufferedDataChangeEvent> bufferedEventRef = new AtomicReference<>(BufferedDataChangeEvent.NULL);
+        private final boolean threaded;
+
+        BufferingSnapshotChangeRecordReceiver(boolean threaded) {
+            this.threaded = threaded;
+        }
 
         @Override
         public void changeRecord(P partition,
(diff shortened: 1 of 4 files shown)


Naros:
@jpechane before we merge this, lets discuss synchronization problem I messaged you about regarding:
```java
queue.enqueue(bufferedEventRef.getAndSet(nextBufferedEvent).dataChangeEvent);
```

debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
```diff
@@ -543,7 +548,17 @@ public void changeRecord(P partition,
             nextBufferedEvent.offsetContext = offsetContext;
             nextBufferedEvent.dataChangeEvent = new DataChangeEvent(record);
 
-            queue.enqueue(bufferedEventRef.getAndSet(nextBufferedEvent).dataChangeEvent);
+            if (threaded) {
+                // This entire step needs to happen atomically when using buffering with multiple threads.
+                // This guarantees that the getAndSet and the enqueue do not cause a dispatch of out-of-order
+                // events within a single thread.
```
vjuranek, line 554: I'm not sure I understand this comment. AFAICT single thread executes instructions sequentially, so the issue described in the Jira that records from single table are delivered out of order IMHO won't be fixed by this change as single thread will update `bufferedEventRef`, enqueue the event and only after that would eventually process another event.

`bufferedEventRef` is share across the threads so it may result in arbitrary order of records from different tables, but this is expected in case snapshot runs with multiple threads.

Another concern may be around marking last snapshot event. But as we wait in for completion service for all tasks to finish (in `RelationalSnapshotChangeEventSource`) and only after that call `completeSnapshot()` and we enqueue this very last event in `completeSnapshot()` call (not here), it should ensure that the very last record is really enqueued as the last one.

So unless I'm missing something, I'm not sure the synchronization here is actually needed.

Naros: Hi @vjuranek that's for the feedback.

With a single-threaded snapshot configuration, there's no issue.  The issue is only during multi-threaded snapshots where the calls to `getAndSet` and `enqueue` happen to where the order of events that were placed into and popped out out `bufferedEventRef` get dispatched differently from how they were buffered.

Given 2 threads, each operating on a single table:
* Thread 1 adds `TableA` with PK 1 to `bufferedEventRef`, which returns NULL and nothing is enqueued.
* Thread 1 adds `TableA` with PK 2 to `bufferedEventRef`, which returns `TableA` with PK 1.
* Thread 2 adds `TableB` with PK 1 to `bufferedEventRef`, which returns `TableA` with PK 2.
* Thread 2 calls queue.enqueue - dispatches `TableA` with PK 2.
* Thread 1 calls queue.enqueue - dispatches `TableA` with PK 1.

With the synchronization we avoid this:
* Thread 1 adds `TableA` with PK 1 to `bufferedEventRef`, which returns NULL and nothing is enqueued.
* Thread 1 adds `TableA` with PK 2 to `bufferedEventRef`, which returns `TableA` with PK 1.
* Thread 1 calls queue.enqueue - dispatches `TableA` PK 1.
* Thread 2 adds `TableB` with PK 1 to `bufferedEventRef`, which returns `TableA` with PK 2.
* Thread 2 calls queue.enqueue - dispatches `TableA` PK 2.
...

The synchronization guarantees that we maintain dispatch order that matches read order from the various threads.  While the events from each thread are interwoven, the data set for a given table will always be dispatched in the order read from the result-set.

Am I wrong?

vjuranek: Ops, sorry, you are of course right @Naros . I didn't realize that due to the buffering second thread can actually enqueue values obtained by the first thread. Thanks for explanation. With this, the fix LGTM.

Naros: Thanks for the confirmation and set pair of eyes, enjoy your weekend! :+1:

PR #5314, opened 2024-02-22
DBZ-7533 Upgrade PostgreSQL driver to 42.7.4

Naros:
https://issues.redhat.com/browse/DBZ-7533

**UPDATE**
The PostgreSQL error looks like it could be related to the upgrade, looking into it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -136,7 +136,7 @@
         <version.apicurio>2.6.2.Final</version.apicurio>
 
         <!-- Database drivers, should align with databases -->
-        <version.postgresql.driver>42.7.3</version.postgresql.driver>
+        <version.postgresql.driver>42.7.4</version.postgresql.driver>
         <version.mysql.driver>9.0.0</version.mysql.driver>
         <version.mysql.binlog>0.31.0</version.mysql.binlog>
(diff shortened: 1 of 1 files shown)


Naros:
I can confirm that dropping Debezium back to 42.7.0 or 42.6.0 does not cause this problem.  Looking at the code changes in `pgjdbc`, this seems related to the new keepalive feature for flushing LSN details automatically in [this](https://github.com/pgjdbc/pgjdbc/pull/2941) PR.  I've reached out directly to Dave to see whether or not this behavior can be toggable so we can preserve this behavior.

Moving to draft until we get some feedback.

Naros:
Upstream issue https://github.com/pgjdbc/pgjdbc/issues/3138

Naros:
So it seems 42.7.3 also causes the same issue :(

ozcan-kara:
<version.postgresql.driver>42.7.4</version.postgresql.driver>

<dependency>
    <groupId>org.postgresql</groupId>
    <artifactId>postgresql</artifactId>
    <version>${version.postgresql.driver}</version>
    <scope>runtime</scope>
</dependency>

Naros:
We've already bumped the pom to 42.7.7, this is obsolete.

PR #5315, opened 2024-02-22
DBZ-7416 Fix duplicate SMTs sometimes returned by Kafka Connect. 

rk3rn3r:
Moved deduplication from Map to LinkedHashSet, based on feedback from #5242

closes https://issues.redhat.com/browse/DBZ-7416


rk3rn3r:
The MySQL CI error is not related, caused by an intermittent test failure that should be fixed with #5301

Naros (approved):
Thanks for making this change @rk3rn3r, LGTM.  Feel free to merge it yourself if I don't get to before the end of my day.

PR #5318, opened 2024-02-23
DBZ-7562 Truncate values that exceed column scale

Naros:
https://issues.redhat.com/browse/DBZ-7562

### DIFF
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresDefaultValueConverterIT.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresDefaultValueConverterIT.java
@@ -130,7 +130,8 @@ public class PostgresDefaultValueConverterIT extends AbstractConnectorTest {
                 + "CREATE TABLE s1.dbz7562 (pk SERIAL, cost numeric(12,0) not null default 0.0, PRIMARY KEY(pk));"
                 + "INSERT INTO s1.dbz7562 (cost) values (2.25);"
                 + "INSERT INTO s1.dbz7562 (cost) values (3);"
-                + "INSERT INTO s1.dbz7562 (cost) values (0);";
+                + "INSERT INTO s1.dbz7562 (cost) values (0);"
+                + "INSERT INTO s1.dbz7562 default values;";
 
         // Test Snapshot
         TestHelper.execute(ddl);
@@ -144,8 +145,9 @@ public class PostgresDefaultValueConverterIT extends AbstractConnectorTest {
         TestHelper.execute("INSERT INTO s1.dbz7562 (cost) values (5.5);");
         TestHelper.execute("INSERT INTO s1.dbz7562 (cost) values (4);");
         TestHelper.execute("INSERT INTO s1.dbz7562 (cost) values (0);");
+        TestHelper.execute("INSERT INTO s1.dbz7562 default values;");
 
-        final int expectedRecords = 6;
+        final int expectedRecords = 8;
         final SourceRecords records = consumeRecordsByTopic(expectedRecords);
         final List<SourceRecord> recordsForTopic = records.recordsForTopic(topicName("s1.dbz7562"));
         assertThat(recordsForTopic).hasSize(expectedRecords);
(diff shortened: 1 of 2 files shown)


debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresDefaultValueConverterIT.java
```diff
@@ -121,6 +122,44 @@ public void testShouldHandleDefaultValueFunctionsWithSchemaPrefixes() throws Exc
...
+        final List<SourceRecord> recordsForTopic = records.recordsForTopic(topicName("s1.dbz7562"));
+        assertThat(recordsForTopic).hasSize(expectedRecords);
+
+        for (SourceRecord record : recordsForTopic) {
+            final Struct after = ((Struct) record.value()).getStruct(Envelope.FieldName.AFTER);
+            final Schema schema = after.schema().field("cost").schema();
+            assertThat(schema.parameters()).containsEntry("connect.decimal.precision", "12");
+            assertThat(schema.parameters()).containsEntry("scale", "0");
+            assertThat(schema.defaultValue().getClass()).isEqualTo(BigDecimal.class);
+            assertThat(schema.defaultValue()).isEqualTo(new BigDecimal(0).setScale(0));
```
rk3rn3r, line 159: Until here you check if the schema is correct and has the expected types.
Shouldn't the streamed values maybe checked too? Especially in the case of BigDecimal and values like `0` and `0.0` etc. For checking the default value wouldn't it make sense to add another `INSERT` that add a record with the default value and see if it is streamed as `0` or `0.0`?

rk3rn3r:
Beside that test-case related comment the code lgtm.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresDefaultValueConverterIT.java
rk3rn3r: But we don't trigger a default value entry, do we?

Naros: @rk3rn3r no we don't but that isn't the reported problem.  The problem is with how we serialize the field's default value into `Field#defaultValue`, not into the field's value.

rk3rn3r: Yes, but this test is the other side of the same thing, no? I don't know, I feel like it should then be tested there too (with a link to this DBZ issue) or testing the values here too. But I might be to strict on this. dunno.

rk3rn3r: @jpechane wdyt?

Naros: I just checked `RecordsStreamProducerIT` method `shouldReceiveChangesForTypeConstraints` where we explicitly set type as `numeric(6,1)` and insert a value of `2123.41` and we verify that the value is truncated and rounded down.  I can certainly add that here if we absolutely think its necessary but it seems redundant.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresDefaultValueConverterIT.java
```diff
@@ -121,6 +122,46 @@ public void testShouldHandleDefaultValueFunctionsWithSchemaPrefixes() throws Exc
...
+        TestHelper.execute("INSERT INTO s1.dbz7562 (cost) values (0);");
+        TestHelper.execute("INSERT INTO s1.dbz7562 default values;");
+
+        final int expectedRecords = 8;
+        final SourceRecords records = consumeRecordsByTopic(expectedRecords);
+        final List<SourceRecord> recordsForTopic = records.recordsForTopic(topicName("s1.dbz7562"));
+        assertThat(recordsForTopic).hasSize(expectedRecords);
+
+        for (SourceRecord record : recordsForTopic) {
+            final Struct after = ((Struct) record.value()).getStruct(Envelope.FieldName.AFTER);
```
jpechane, line 156: For MySQL default valut tests we call `VerifyRecord.isValid` - it check serialization too. I recommend to add here. When done feel free to merge the PR yourself.

jpechane (commented):
@Naros LGTM, thanks! I left one comment related to one additional check. When addressed feel fre to merge yourself.

Naros:
Applied to main (=2.6) and branch 2.5.

PR #5326, opened 2024-02-27
Dbz 7258 - add reconnect DB feature to JDBC Storage

sherpa003:


sherpa003:
> Hi @sherpa003, thanks for your contribution. Please prefix the commit message(s) with the [DBZ-xxx JIRA issue key](https://github.com/debezium/debezium/blob/main/CONTRIBUTE.md#making-changes).

I don't understand why this is failing as all commits are prefixed

jpechane:
@sherpa003 No worries, it sometimes happens if you do force push and overwirte commits that contained incorrect messages.

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RecoverableConnectionHolder.java
```diff
@@ -0,0 +1,130 @@
...
+import java.sql.SQLRecoverableException;
+import java.time.Duration;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.debezium.storage.jdbc.offset.JdbcOffsetBackingStore;
+import io.debezium.util.DelayStrategy;
+
+public class RecoverableConnectionHolder implements AutoCloseable {
```
jpechane, line 20:
```suggestion
public class RetrieableConnection implements AutoCloseable {
```

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RecoverableConnectionHolder.java
```diff
@@ -0,0 +1,130 @@
...
+            }
+        }
+    }
+
+    @FunctionalInterface
+    public interface ConnectionFunction<T> {
+        T apply(Connection conn) throws SQLException, CustomException;
+    }
+
+    public static class CustomException extends RuntimeException {
```
jpechane, line 121: Lets use `org.apache.kafka.connect.errors.RetriableException` instead of a new one.

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RecoverableConnectionHolder.java
```diff
@@ -0,0 +1,130 @@
...
+import java.sql.SQLException;
+import java.sql.SQLRecoverableException;
+import java.time.Duration;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.debezium.storage.jdbc.offset.JdbcOffsetBackingStore;
+import io.debezium.util.DelayStrategy;
+
```
jpechane, line 19: Could you please add a JavaDoc?

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RecoverableConnectionHolder.java
```diff
@@ -0,0 +1,130 @@
...
+    @Override
+    public void close() throws SQLException {
+        conn.close();
+    }
+
+    public boolean isConnectionCreated() {
+        return conn != null;
+    }
+
+    public synchronized <T> T executeWithReconnect(ConnectionFunction<T> func, String name, boolean rollback)
```
jpechane, line 55:
```suggestion
    public synchronized <T> T executeWithRetry(ConnectionFunction<T> func, String name, boolean rollback)
```

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/history/JdbcSchemaHistory.java
```diff
@@ -54,8 +53,7 @@ public final class JdbcSchemaHistory extends AbstractSchemaHistory {
     private final DocumentReader reader = DocumentReader.defaultReader();
     private final AtomicBoolean running = new AtomicBoolean();
     private final AtomicInteger recordInsertSeq = new AtomicInteger(0);
-
-    private Connection conn;
+    private RecoverableConnectionHolder connHandler;
```
jpechane, line 56: Let's keep the name `conn`

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RecoverableConnectionHolder.java
```diff
@@ -0,0 +1,130 @@
...
+                        // ignore
+                    }
+                }
+                throw e;
+            }
+        }
+    }
+
+    @FunctionalInterface
+    public interface ConnectionFunction<T> {
```
jpechane, line 117: I'd like to see TWO functional interfaces. One as is now and another that will be derived from `Consumer` for calls not returning any return value. There should be two `executeWithReconnect` each taking different parameter.

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RecoverableConnectionHolder.java
```diff
@@ -0,0 +1,130 @@
...
+    private Connection conn;
+
+    public RecoverableConnectionHolder(String url, String user, String pwd, Duration waitRetryDelay, int maxRetryCount) throws SQLException {
+        this.url = url;
+        this.user = user;
+        this.pwd = pwd;
+        this.waitRetryDelay = waitRetryDelay;
+        this.maxRetryCount = maxRetryCount;
+
+        createConnection();
```
jpechane, line 38: Should it be called here? It will fail away immediatelly without tying to reconnect, is it expected?

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/history/JdbcSchemaHistory.java
```diff
@@ -184,13 +195,18 @@ protected synchronized void recoverRecords(Consumer<HistoryRecord> records) {
...
         boolean sExists = false;
         try {
-            DatabaseMetaData dbMeta = conn.getMetaData();
-            String databaseName = config.getDatabaseName();
-            ResultSet tableExists = dbMeta.getTables(databaseName,
-                    null, config.getTableName(), null);
-            if (tableExists.next()) {
-                sExists = true;
-            }
+            sExists = connHandler.executeWithReconnect(conn -> {
```
jpechane, line 198: Can direct `return` be called without variable?

jpechane (changes requested):
@sherpa003 Thanks for the PR, I think we are on a good path. I left few comments related to namind/docs and small code changes but IMHO the requested funcitonlaity is well in place.

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RecoverableConnectionHolder.java
jpechane: No there is no need for code duplication, you can implement it via a single method. What is important is not to force the caller to return dummy `null` value if nothing is going to be returned.
So the new method can just delegate to the old one or there can be a thir `private` called by the new and old.

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RecoverableConnectionHolder.java
jpechane: Are you sure? IMHO is an exception is thrown it gets propagated from the constructor.

sherpa003: oh right, thanks for the catch

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RecoverableConnectionHolder.java
sherpa003: that's where I'm not sure how to bridge between the 2 functional interfaces with different signature
Is this cast going to work?

```
public synchronized void executeWithRetry(ConnectionVoidFunction func, String name, boolean rollback)
            throws SQLException, RetriableException {
        executeWithRetry((ConnectionFunction<? extends Object>) func, name, rollback);
    }
```

I have 

```
    @FunctionalInterface
    public interface ConnectionFunction<T> {
        T apply(Connection conn) throws SQLException, RetriableException;
    }

    @FunctionalInterface
    public interface ConnectionVoidFunction {
        void apply(Connection conn) throws SQLException, RetriableException;
    }
```

jpechane: First, please mimic `java.util.function.Consumer<T>` interface for no return value.

The easiest implementation, the third private method accepts both function based and consumer based parameters. The inside of the method will be the same as the current one but will check which of those two parameters is not null and call appropriate function/consumer.

The two public methods will call the private one passing the parameter they got and `null` for the other one.

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RecoverableConnectionHolder.java
sherpa003: I opted for UncheckedIOException because the retriable exception would be really confusing here - it must not retry connection on it and it is used for further processing if action fails (see JdbcSchemaHistory.recoverRecords)

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RecoverableConnectionHolder.java
sherpa003: thanks for guidance

jpechane:
@sherpa003 Thanks, unfortunately there are failures in JDBC storage tests.

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RetriableConnection.java
```diff
@@ -0,0 +1,219 @@
...
+         * Performs this operation on the given connection.
+         *
+         * @param conn the input connection
+         * @return result of the operation
+         */
+        T accept(Connection conn) throws SQLException, UncheckedIOException;
+    }
+
+    @FunctionalInterface
+    public interface ConnectionFunctionVoid {
```
jpechane, line 189: Please don't call it `ConnectionFunction`. In Java it corresponds to `java.util.function.Consumer<T>` so it should be `ConnectionConsumer`.

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RetriableConnection.java
```diff
@@ -0,0 +1,219 @@
...
+                    }
+                    attempt++;
+                    LOGGER.debug("Waiting for reconnect for {} ms.", waitRetryDelay);
+                    DelayStrategy delayStrategy = DelayStrategy.constant(waitRetryDelay);
+                    delayStrategy.sleepWhen(true);
+                    continue;
+                }
+            }
+            try {
+                if (func == null) {
```
jpechane, line 133:
```suggestion
                if (func != null) {
```

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RetriableConnection.java
```diff
@@ -0,0 +1,219 @@
...
+     * @param rollback if set to true, the rollback will be called in case of SQLException
+     * @throws SQLException sql connection related exception
+     * @throws UncheckedIOException exception that can be thrown by code snippet
+     */
+    public synchronized <T> T executeWithRetry(ConnectionFunction<T> func, String name, boolean rollback)
+            throws SQLException, UncheckedIOException {
+        return executeWithRetry(func, null, name, rollback);
+    }
+
+    private synchronized <T> T executeWithRetry(ConnectionFunction<T> func, ConnectionFunctionVoid funcVoid, String name, boolean rollback)
```
jpechane, line 110:
```suggestion
    private synchronized <T> T executeWithRetry(ConnectionFunction<T> func, ConnectionConsumer consumer, String name, boolean rollback)
```

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RetriableConnection.java
```diff
@@ -0,0 +1,219 @@
...
+
+    /**
+     * execute code snippet where no value is returned.
+     * @param func code snippet to execute
+     * @param name name of the operation being executed (for logging purposes)
+     * @param rollback if set to true, the rollback will be called in case of SQLException
+     * @throws SQLException sql connection related exception
+     * @throws UncheckedIOException exception that can be thrown by code snippet
+     */
+    public synchronized void executeWithRetry(ConnectionFunctionVoid func, String name, boolean rollback)
```
jpechane, line 92:
```suggestion
    public synchronized void executeWithRetry(ConnectionConsumer consumer, String name, boolean rollback)
```

debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RetriableConnection.java
```diff
@@ -0,0 +1,219 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.storage.jdbc;
+
+import java.io.UncheckedIOException;
```
jpechane, line 8: Looking at the code now is it even necessary to use the different exception?

sherpa003: you are right

jpechane:
@Naros I rebased it against the latest main and cleaned the code a bit more. Could you please review it? Thanks

Naros (commented):
Hi @jpechane and @sherpa003 I've left a few comments on a few things.

jpechane:
@Naros Build failure is unrelated

Naros (approved):
LGTM, thanks @jpechane; feel free to merge yourself.

jpechane:
And it is merged. Thanks @sherpa003 for the PR!

PR #5327, opened 2024-02-27
DBZ-7569 removed lombok code from debezium-testing-system

MartinMedek:


vjuranek (approved):

MartinMedek:
Ran the tests for downstream just in case, tests passed

PR #5328, opened 2024-02-27
Use default wait time in all AsyncEmbeddedEngineTest

vjuranek:
https://issues.redhat.com/browse/DBZ-7568


jpechane:
@vjuranek Applied, thanks

PR #5335, opened 2024-02-28
DBZ-7582 Correctly serialize toasted HSTORE values when unavailable

Naros:
https://issues.redhat.com/browse/DBZ-7582


jpechane:
@Naros Applied, thanks!

PR #5338, opened 2024-02-29
DBZ-7049 Fix to use caching_sha2_password instead of mysql_native_password

Bue-von-hon:
this Resolves [DBZ-7049](https://issues.redhat.com/browse/DBZ-7049)

PTAL @Naros


Bue-von-hon:
I think a more understandable change would be to change the default_authentication_plugin property in debezium-connector-mysql/src/test/server.my.cnf for 8.0 and 8.2 versions.

I'd appreciate your feedback on which method you prefer.

jpechane:
@Bue-von-hon I'd prefer the other proposed solution, change of server config.

Naros:
Hi @Bue-von-hon looks like the test failures are related, can you please take a look?

jpechane:
RSA key needs to be provided wirh this setign too, see https://dev.mysql.com/doc/refman/8.0/en/caching-sha2-pluggable-authentication.html

njiddasalifu (commented):
Did you notice the failing Builds @Bue-von-hon

Bue-von-hon:
Oh... Thank you for your help, @jpechane @Naros

PR #5340, opened 2024-02-29
DBZ-7579 Exclude OpenLogReplicator dependencies

Naros:
https://issues.redhat.com/browse/DBZ-7579


jpechane:
@Naros Applied, thanks

PR #5346, opened 2024-03-02
Rework GitHub Actions

Naros:
# Logic

Pull requests will always consume many runners as needed.  GitHub provides us with a maximum of 20 runners, and 
these runners shouldn't be fully consumed as most of our changes should generally trend more toward specific connectors rather than common code in modules such as `debezium-core`.

Push builds require an initial single runner until the check-style job completes, at which point, the workflow will allocate a maximum of 8 runners (1 for a set of sequential execution steps outlined below and several matrix jobs that typically take between 30-40 minutes each, and a couple for the longer runs like Db2 and SQL Server).

# Overall changes:

## Pull Requests
* Renamed `debezium-workflow` to `debezium-workflow-pr`, only applies to pull requests
* Added concurrency, any push to the same branch or pull request cancels prior job automatically.
* Merged MySQL and MySQL 8 into a single matrix (8.0 and 8.2)
* Removed MongoDB 6.0 from matrix, only runs MongoDB 5.0 and 7.0.
* Reused as much action logic between pull request and push workflows as possible.
* Merged Oracle test workflow with pull request workflow since we do not run any tests.

## Pushes
* Added a new workflow `debezium-workflow-push` specifically for pushes to the repository.
* A given branch can have only 1 running push workflow
* Merged MySQL and MySQL 8 into a single matrix (8.0 and 8.2)
* Removed MongoDB 6.0 from matrix, only runs MongoDB 5.0 and 7.0.
* Uses a separate cache from pull requests to avoid thrashing.
* Always builds the entire repository, regardless of the changes (except documentation).
* MongoDB matrix is restricted to 1 worker for 2 jobs.
* MySQL matrix is restricted to 2 workers for 8 jobs.
* MariaDB matrix is restricted to 1 worker for 2 jobs.
* PostgreSQL matrix is restricted to 1 worker for 2 jobs.
* Merged Oracle test workflow with push workflow since we do not run any tests.
* The following jobs (~93 minutes) are executed sequentially (allows parallelism with matrix jobs, MSSQL, Db2 without over saturing the runner pool)
  * Oracle (~ 5 minutes)
  * Quarkus Outbox Extension (~2 minutes)
  * REST Extension (~5 minutes)
  * Schema Generator (~2 minutes)
  * Testing Module (~4 minutes)
  * Storage Module (~4 minutes)
  * Cassandra (~25 minutes)
  * Spanner (~8 minutes)
  * JDBC (~2 minutes)
  * Vitess (~20 minutes)
  * Debezium Server (~16 minutes)

image


rk3rn3r:
Nice job! I was not able to look into all the details but the change looks really good! Thx @Naros!

PR #5347, opened 2024-03-04
DBZ-7507 Configure startup timeout for MongoDb ReplicaSet

ani-sha:
https://issues.redhat.com/browse/DBZ-7507


debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
```diff
@@ -45,7 +44,9 @@
 public class MongoDbReplicaSet implements MongoDbDeployment {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(MongoDbReplicaSet.class);
-
+    private static final String STARTUP_TIMEOUT = System.getProperty("startup.timeout.seconds");
```
Naros, line 47: I wonder if it would be best to prefix this with `debezium.test.` like we do for other wait times just so that we keep some consistency with how these are named.  wdyt @ani-sha ?

.github/workflows/debezium-workflow.yml
```diff
@@ -646,6 +646,7 @@ jobs:
           -Dmaven.wagon.http.pool=false
           -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
           -DfailFlakyTests=false
+          -Dstartup.timeout.seconds=120
```
Naros, line 649: Same as above.

Naros (commented):
LGTM, just one inline comment @ani-sha :tada:

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/MongoDbReplicaSet.java
ani-sha: @Naros I had initially added the prefix, but since it was not directly used in the test classes, I removed it. I agree, it's good to have the naming consistency throughout the code. Do you have any other prefix in mind or we should go with `debezium.test.`?

mfvitale: I would add some reference to better identify `startup of what`? Something like ` debezium.test.mongo.replica.primary.startup.timeout.seconds` WDYT?

Naros: I agree @mfvitale, we should try and aim at being descriptive :100:

Naros:
@ani-sha do you mind rebasing on `main` ?

Naros:
@ani-sha you added the `debezium-workflow.yml` back, that was removed on `main` in favor of the `debezium-workflow-pr` and `debezium-workflow-push` split.  Can you please take a look?

jpechane:
@ani-sha @Naros Applied, thanks. Let's hope this is the end solution for it.

ani-sha:
@jpechane Unforunately not, the build on the main branch for Mongo 7.0 is failing with the same test failure for this PR changes :(

PR #5348, opened 2024-03-04
MySqlStreamingChangeEventSource: Convert non-full rows to full rows

shanielh:
This should fix DBZ-7591


Naros:
Hi @shanielh the issue reports this as affecting MariaDB but I do not see anything in the code that guards this change specifically to MariaDB.  In addition, could this be due to the binlog row format configuration?

shanielh:
> Hi @shanielh the issue reports this as affecting MariaDB but I do not see anything in the code that guards this change specifically to MariaDB.  In addition, could this be due to the binlog row format configuration?  

Hi @Naros, thanks for the fast response. I think this might happen in MySQL too since it's related to the MySQL binlog wire format, it might be related to the configuration but I couldn't find anything related to it. I'll try look for more information tomorrow.

Naros:
Hi @shanielh can you please add some tests for this change?

shanielh:
Hi @Naros, sure. I'll try to add tests using minimal row binlog setting. I hope to do that by Monday.

Naros:
Hi @shanielh, just wanting to follow-up and see if you have any questions or issues adding the test.

shanielh:
> Hi @shanielh, just wanting to follow-up and see if you have any questions or issues adding the test.

Yeah actually I do, should I change the format of the row to minimal, or should I somehow create another instance of a MySQL in which I'll change the configuration?

jpechane:
@shanielh I don't think it is necessary to create another long-running instance

1) It seems to me tha binlog format is a dynamic setting. So I'd first recommend to try set the different format as part of the test class setup and run tests against it. It is necessary to validate that the non-full rows were read and processed. After the test it is necessary to make sure that the format is switched back to the full one.
2) If option 1) will show to be non-feasible (like breaking the rest of test suite etc.) then the other option is to write the test to use test containers to manage its own private MySQL with requested config and run test against it.

Naros:

Hi @shanielh, please sign off all commits with:

```
git commit -s
```

PR #5360, opened 2024-03-07
DBZ-7608 Resolve performance issue during addSchemaEvent

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7608

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
@@ -5,8 +5,6 @@
  */
 package io.debezium.connector.mysql;
 
-import static java.util.function.Predicate.not;
-
 import java.sql.Connection;
 import java.sql.ResultSet;
 import java.sql.SQLException;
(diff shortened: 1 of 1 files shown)


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlSnapshotChangeEventSource.java
```diff
@@ -314,10 +312,7 @@ private void addSchemaEvent(RelationalSnapshotContext<MySqlPartition, MySqlOffse
 
         List<SchemaChangeEvent> schemaChangeEvents = databaseSchema.parseSnapshotDdl(snapshotContext.partition, ddl, database,
                 snapshotContext.offset, clock.currentTimeAsInstant());
-        List<SchemaChangeEvent> missingSchemaChangeEvents = schemaChangeEvents.stream()
-                .filter(not(schemaEvents::contains))
-                .collect(Collectors.toList());
-        schemaEvents.addAll(missingSchemaChangeEvents);
+        schemaEvents.addAll(new HashSet<>(schemaChangeEvents));
```
jpechane, line 315: If there are multiple events we'd lose the order in which they are applied.

mfvitale: Changed it with `LinkedHashSet` that maintain the same performance but with ordering.

jpechane:
@mfvitale Applied, thanks

PR #5361, opened 2024-03-07
DBZ-7592 Add OCP JDBC sink system tests

novotnyJiri:
https://issues.redhat.com/browse/DBZ-7592


obabec (approved):
Good work! Seems OK to me.

PR #5362, opened 2024-03-08
Dbz 7614 Catch exceptions during validation

andreas-ibm:
Fixes https://issues.redhat.com/browse/DBZ-7614
Adds a try-catch block around the `validate` call, which means that we can catch NoClassDef exceptions when a jar file is missing from the plugin directory

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
@@ -508,12 +508,12 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord>, Embed
     private Map<String, String> getConnectorConfig(final SourceConnector connector, final String connectorClassName) throws EmbeddedEngineRuntimeException {
         Map<String, String> connectorConfig = workerConfig.originalsStrings();
         Config validatedConnectorConfig = null;
-        try{
-          validatedConnectorConfig = connector.validate(connectorConfig);
+        try {
+            validatedConnectorConfig = connector.validate(connectorConfig);
         }
-        catch (java.lang.NoClassDefFoundError ex){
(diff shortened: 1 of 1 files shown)


debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
```diff
@@ -507,7 +507,14 @@ private SourceConnector instantiateConnector(final String connectorClassName) th
 
     private Map<String, String> getConnectorConfig(final SourceConnector connector, final String connectorClassName) throws EmbeddedEngineRuntimeException {
         Map<String, String> connectorConfig = workerConfig.originalsStrings();
-        Config validatedConnectorConfig = connector.validate(connectorConfig);
+        Config validatedConnectorConfig = null;
+        try {
+            validatedConnectorConfig = connector.validate(connectorConfig);
+        }
+        catch (java.lang.NoClassDefFoundError ex) {
```
vjuranek, line 514: I'd catch general `Exception` as `validate()` can throws also other errors like `ConnectException`.

vjuranek (changes requested):
Thanks for the PR! I left just a small suggestion for improvement.

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
andreas-ibm: thanks, thanks, I've expanded the scope, was that what you meant?

vjuranek (approved):
Let's wait for test, but LGTM now

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
vjuranek: yes, thanks for quick response

vjuranek:
Merged, thanks @andreas-ibm !

PR #5363, opened 2024-03-09
DBZ-7618 Implement Versioned interfaces in Transformation and Converter plugins

harveyyue:
https://issues.redhat.com/browse/DBZ-7618

### DIFF
--- a/debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/DebeziumConnectRestExtension.java
+++ b/debezium-connect-rest-extension/src/main/java/io/debezium/kcrestextension/DebeziumConnectRestExtension.java
@@ -7,7 +7,6 @@ package io.debezium.kcrestextension;
 
 import java.util.Map;
 
-import org.apache.kafka.common.utils.AppInfoParser;
 import org.apache.kafka.connect.rest.ConnectRestExtension;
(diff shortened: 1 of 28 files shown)


harveyyue:
jdbc: https://github.com/debezium/debezium-connector-jdbc/pull/65

jpechane:
@harveyyue Thank you very much for the PRs. While it is great to have the support for versioned interfaces I am not really sure about the other change introduced - centralization of the version source. The current implementation is able to detect real version based on the classloader used to load the connector. I think thi should astay as it is now. The SMTs in connectors should use appropriate `Module` class and the core and scripting library should intorduce its own equivalent of `Module` that will provide the version string.

harveyyue:
> @harveyyue Thank you very much for the PRs. While it is great to have the support for versioned interfaces I am not really sure about the other change introduced - centralization of the version source. The current implementation is able to detect real version based on the classloader used to load the connector. I think thi should astay as it is now. The SMTs in connectors should use appropriate `Module` class and the core and scripting library should intorduce its own equivalent of `Module` that will provide the version string.

OK, will do it.

jpechane:
@harveyyue Applied, thanks a lot. Should we do the same for connector classes?

jpechane:
Create a PR for IBMi connector - https://github.com/debezium/debezium-connector-ibmi/pull/6

PR #5371, opened 2024-03-11
DBZ-7304 Align snapshot modes for MongoDb connector

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7304

- [x] Add tests for Always, When Needed and Custom snapshot
- [x] Check support for SnapshotQuery interface

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbChangeEventSourceFactory.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbChangeEventSourceFactory.java
@@ -75,7 +75,8 @@ public class MongoDbChangeEventSourceFactory implements ChangeEventSourceFactory
                 dispatcher,
                 errorHandler,
                 clock,
-                streamingMetrics);
+                streamingMetrics,
+                snapshotterService);
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


mfvitale:
@jpechane @jcechace Wondering if the possibility to customize the `query` for the snapshot has sense for Mongo. 

For relations databases, users can implement `Optional<String> snapshotQuery(String tableId, List<String> snapshotSelectColumns);`

and have enough power/infos to manipulate the query. 
For Mongo it's more complex since they can provide just the filtering part (no info about `columns`) or, if we want to give them the full power, they should play directly with `MongoCollection`.  Am I right?

Said that I'll maintain the possibility to define advanced filtering just using [snapshot.collection.filter.overrides](https://debezium.io/documentation/reference/2.5/connectors/mongodb.html#mongodb-property-snapshot-collection-filter-overrides). WDYT?

jcechace:
@mfvitale the problem with customising the query for MongoDB is that it wouldn't be just a simple string so the API would have to be different -- instead of string a we would likely need the API to return either FindIterable or just the Filter query (Which is a Bson type)

mfvitale:
> @mfvitale the problem with customising the query for MongoDB is that it wouldn't be just a simple string so the API would have to be different -- instead of string a we would likely need the API to return either FindIterable or just the Filter query (Which is a Bson type)

Yes, and for that reason I will propose to not support the `SnapshotterQuery` interface. @jpechane WDYT?

jcechace:
I'd say that not supporting it is reasonable -- at least for now.

jpechane:
Can't we just return the filter string, parse it via `Document.parse` and use it for filtering?

jpechane:
OK, let's skip it now. Fortunately you designed separated interfaces :-). We can rethink the issue when there is a demand for it.

mfvitale:
> We can rethink the issue when there is a demand for it.

Yes, I agree. So this is ready for review ;-)

jpechane:
@mfvitale Applied, thanks!

PR #5374, opened 2024-03-11
DBZ-7611 Exclude signal table from re-selection

Naros:
https://issues.redhat.com/browse/DBZ-7611


jpechane:
@Naros Applied, thanks

PR #5379, opened 2024-03-12
[docs] Vitess connector ts_ms is seconds level precision in milliseconds units

twthorn:
It is a subtle behavior that should be documented. Although called ts_ms the precision is seconds converted to milliseconds. The examples are also misleading.


Naros (approved):
Awesome thanks for confirming @twthorn !

PR #5381, opened 2024-03-13
DBZ-5215 Parse column names and column values expressed using "col 1" and " hextoraw()", and table names that use "BIN" and "OBJ xxx".

butioy:
DBZ-5215 Parse column names and column values expressed using "col 1" and "hextoraw()", and table names that use "BIN" and "OBJ xxx".


Naros:
Hi @butioy this seems like duplicated work that was already done as part of https://issues.redhat.com/browse/DBZ-3401, can you please take a look at that work and see if this continues to make sense?  Additionally, please target `main` and not branch 2.3 as 2.3 is no longer maintained in the community.

PR #5388, opened 2024-03-14
DBZ-7643 Support numeric values without mantissas

Naros:
https://issues.redhat.com/browse/DBZ-7643


jpechane:
@Naros Applied, thanks!

PR #5389, opened 2024-03-14
DBZ-7645 Revert removal of LogMiner continuous mining

Naros:
https://issues.redhat.com/browse/DBZ-7645


jpechane:
@Naros It seems it aleast Oracle 23 failures are related

Naros:
I am going to re-run the Oracle 23 test @jpechane as it does not fail locally.

Naros:
Looks to have been a fluke with the environment @jpechane, so I think this is good.

PR #5392, opened 2024-03-14
DBZ-7647 Adds capture.mode.full.update.type entry to config props table

roldanbob:
[DBZ-7647](https://issues.redhat.com/browse/DBZ-7647)

Updates the MongoDB documentation by adding an entry for `capture.mode.full.update.type` to  the table of Advanced connector configuration properties.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/mongodb.adoc
+++ b/documentation/modules/ROOT/pages/connectors/mongodb.adoc
@@ -1818,7 +1818,7 @@ The following _advanced_ configuration properties have good defaults that will w
 |[[mongodb-property-capture-mode-full-update-type]]<<mongodb-property-capture-mode-full-update-type, `+capture.mode.full.update.type+`>>
 |`lookup`
 |Specifies how the connector looks up the full value of an updated document when the xref:mongodb-property-capture-mode[`capture.mode`] is set retrieve full documents.
-The connector retrieves full documents when `capture.mode` is set to one of the following options:
+The connector retrieves full documents when its `capture.mode` is set to one of the following options:
 
 * `change_streams_update_full`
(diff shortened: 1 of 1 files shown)


documentation/modules/ROOT/pages/connectors/mongodb.adoc
```diff
@@ -1813,6 +1815,20 @@ The following _advanced_ configuration properties have good defaults that will w
 |Default
 |Description
 
+|[[mongodb-property-capture-mode-full-update-type]]<<mongodb-property-capture-mode-full-update-type, `+capture.mode.full.update.type+`>>
+|`lookup`
+|Specifies how the connector looks up the full value of an updated document when the xref:mongodb-property-capture-mode[`capture.mode`] is set retrieve full documents.
+The connector retrieves full documents when `capture.mode` is set to one of the following options:
```
roldanbob, line 1821:
```suggestion
The connector retrieves full documents when its `capture.mode` is set to one of the following options:
```

jpechane:
@jcechace Could you please proof-read the change? Feel free to merge PR yourself if it is OK for you.

jcechace:
@roldanbob LGTM. Maybe adding an info that post-pre images need to be enable per collection would be a good idea? https://www.mongodb.com/docs/manual/changeStreams/#change-streams-with-document-pre--and-post-images

roldanbob:
> @roldanbob Applied, thanks! Is this going to be backported to 2.5 branch?

@jpechane Yes, I think that's what we want, right? I realized that this was missing when I was looking at the 2.5 requirement to enable post-image support in MongoDB.

jpechane:
Yes, we do. I just updated the fix version in Jira.

PR #5394, opened 2024-03-15
DBZ-7652 Upgrade MySQL driver to 8.3.0

Naros:
https://issues.redhat.com/browse/DBZ-7652


Naros:
See note on the Jira about MySQL connector-j 8.3.0 driver compatibility with MySQL versions prior to 8.0.

jpechane:
@Naros Applied, thanks. The limitation is acceptable.

PR #5398, opened 2024-03-17
DBZ-7389 Oracle log collector with full RAC support

Naros:
https://issues.redhat.com/browse/DBZ-7389


Naros:
@jpechane, so I left the overall error-throwing bit the same in the rework here and littered it with a ton of debug logging, but I am considering having the underlying sanity check functions throw exceptions rather than retrying true/false.  

This would allow us to bubble up from the sanity checks a more concrete root cause to the top-level exception, giving users immediate feedback on the underlying error reason in the event they do not have DEBUG logging enabled. Otherwise, I think the solution is solid from my PoV; but I did ping Lucas on the Jira to give it a review & check.

jpechane:
@Naros Do we need to throw an exception? Maybe we can create and log an exception as part of the checking to get the stacktrace in the log?

Naros:
> @Naros Do we need to throw an exception? Maybe we can create and log an exception as part of the checking to get the stacktrace in the log?

I actually think that idea is much cleaner and avoids unnecessarily over-complicating the caller methods where the loop logic is so it remains quite clean :+1:

PR #5401, opened 2024-03-18
DBZ-7308 Align all snapshot mode on all connectors

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7308

This is just a refactor after all connectors has been aligned.


jpechane:
@mfvitale Applied, thanks!

PR #5402, opened 2024-03-18
[DBZ-7649] Add missing test annotation in PostgresConnectorIT

vaibhav-yb:
This PR adds the missing test annotation `@Test` from the file `PostgresConnectorIT.java`.


jpechane:
@vaibhav-yb Applied, thanks

PR #5407, opened 2024-03-19
DBZ-3658 Oracle JDBC driver JDK8 to JDK11 (ojdbc11) 21.15.0.0

Naros:
https://issues.redhat.com/browse/DBZ-3658

@obabec your favorite task -- we'll need to align the testing farm to contain `ojdbc11` 21.11.0.0, when you have time :grinning: 

- [x] Check release bits and verify no changes needed for `ojdbc8` to `ojdbc11`.
- [x] Update driver reference used by the JDBC sink connector.

### DIFF
--- a/debezium-assembly-descriptors/src/main/resources/assemblies/connector-distribution-no-infinispan.xml
+++ b/debezium-assembly-descriptors/src/main/resources/assemblies/connector-distribution-no-infinispan.xml
@@ -36,6 +36,7 @@
 
               <!-- Exclude Oracle JDBC driver libraries -->
               <exclude>com.oracle.database.jdbc:ojdbc8:*</exclude>
+              <exclude>com.oracle.database.jdbc:ojdbc11:*</exclude>
               <exclude>com.oracle.database.nls:orai18n:*</exclude>
 
               <!-- Exclude dependencies with incorrect scope -->
--- a/debezium-bom/pom.xml
+++ b/debezium-bom/pom.xml
@@ -230,7 +230,7 @@
             <!-- Oracle JDBC driver, XML support, Oracle XStream -->
             <dependency>
                 <groupId>com.oracle.database.jdbc</groupId>
-                <artifactId>ojdbc8</artifactId>
+                <artifactId>ojdbc11</artifactId>
                 <version>${version.oracle.driver}</version>
             </dependency>
             <dependency>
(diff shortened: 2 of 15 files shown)


Naros:
@roldanbob wrt docs, I wonder if we shouldn't just change the attribute from `ojdbc8-version`, now `ojdbc11-version` to actually just reading `ojdbc-version`.  wdyt?  Would this be easy to coordinate for downstream?

roldanbob:
> @roldanbob wrt docs, I wonder if we shouldn't just change the attribute from `ojdbc8-version`, now `ojdbc11-version` to actually just reading `ojdbc-version`. wdyt? Would this be easy to coordinate for downstream?

We can easily update the attribute name from `ojdbc-version` to `ojdbc-version`, I think , and then assign the current value to it. That makes sense to me. Upstream, it looks like there are only 3 instances in the sources where we reference the attribute, so it's a pretty painless change.

Naros:
Sounds good, then I'll modify the PR based on that.  Thanks @roldanbob.

roldanbob:
@Naros Based on the Jira issue, I assume that this affects only 2.6 and later, right? Nothing to turn around for 2.5?

Naros:
> @Naros Based on the Jira issue, I assume that this affects only 2.6 and later, right? Nothing to turn around for 2.5?

That's correct, and it may even be something we hold off until 2.7; deferring that to @jpechane.
Given that we're on the brink of 2.6.CR1, I'd say that may even be the best and safest bet.

jpechane:
@Naros Where do we stand with this PR? Can it be undrafted for 3.0?

Naros:
Let me rerun the checks and get back to you on that @jpechane

roldanbob:
> @roldanbob wrt docs, I wonder if we shouldn't just change the attribute from `ojdbc8-version`, now `ojdbc11-version` to actually just reading `ojdbc-version`. wdyt? Would this be easy to coordinate for downstream?

@Naros Per our earlier convo, should we use the attribute `ojdbc-version` rather than `ojdbc11-version`?

Naros:
I've tested both XStream and LogMiner locally.  I'll need to coordinate with Ondrej so that we can update the testing farm images before we merge this PR so that we don't run into an inconsistency of dependencies, but otherwise the tests passed for both adapters using ojdbc11 21.15.0.0.

vjuranek (approved):

Naros:
Thanks for all the help on making this happen for Debezium 3 @vjuranek ❤️

PR #5408, opened 2024-03-19
DBZ-7662 Update Debezium Server to Quarkus 3.2.11.Final

Naros:
https://issues.redhat.com/browse/DBZ-7662


jpechane:
@Naros Applied, thanks

PR #5409, opened 2024-03-19
DBZ-7665 Exclude jcl-over-slf4j dependency

Naros:
https://issues.redhat.com/browse/DBZ-7665


jpechane:
@Naros Applied, thanks!

PR #5411, opened 2024-03-20
DBZ-6964 Temporary fix schema for MySQL "Geometry" types conflicting with Avro schema registries (Confluent, Apicurio).

rk3rn3r:
It seems that schema registries don't incorporate schema parameters correctly leading to that only the first row of a Geometry type is used as schema reference. If a single topic/table has multiple fields/rows of a geo type than only one schema is referenced for all fields, even when the real schema of those fields differ.

### DIFF
--- a/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlGeometryIT.java
+++ b/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlGeometryIT.java
@@ -8,8 +8,11 @@ package io.debezium.connector.mysql;
 import static io.debezium.junit.EqualityCheck.LESS_THAN;
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
 import java.nio.file.Path;
 import java.sql.SQLException;
+import java.util.concurrent.atomic.AtomicReference;
 
 import javax.xml.bind.DatatypeConverter;
 
@@ -20,8 +23,17 @@ import org.junit.Assert;
 import org.junit.Before;
 import org.junit.Test;
 
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+
+import io.apicurio.registry.rest.client.RegistryClient;
+import io.apicurio.registry.rest.client.RegistryClientFactory;
 import io.debezium.config.Configuration;
 import io.debezium.data.Envelope;
+import io.debezium.data.VerifyRecord;
+import io.debezium.doc.FixFor;
 import io.debezium.embedded.async.AbstractAsyncEngineConnectorTest;
 import io.debezium.junit.SkipWhenDatabaseVersion;
 import io.debezium.util.Testing;
@@ -36,8 +48,9 @@ import mil.nga.wkb.io.WkbGeometryReader;
 @SkipWhenDatabaseVersion(check = LESS_THAN, major = 5, minor = 6, reason = "Function ST_GeomFromText not added until MySQL 5.6")
 public class MySqlGeometryIT extends AbstractAsyncEngineConnectorTest {
 
-    private static final Path SCHEMA_HISTORY_PATH = Testing.Files.createTestingPath("file-schema-history-json.txt")
+    private static final Path SCHEMA_HISTORY_PATH = Files.createTestingPath("file-schema-history-json.txt")
             .toAbsolutePath();
+    public static final ObjectMapper OBJECT_MAPPER = new ObjectMapper();
     private UniqueDatabase DATABASE;
     private DatabaseGeoDifferences databaseDifferences;
 
@@ -53,7 +66,7 @@ public class MySqlGeometryIT extends AbstractAsyncEngineConnectorTest {
         DATABASE.createAndInitialize();
 
         initializeConnectorTestFramework();
-        Testing.Files.delete(SCHEMA_HISTORY_PATH);
+        Files.delete(SCHEMA_HISTORY_PATH);
     }
 
     @After
@@ -62,7 +75,7 @@ public class MySqlGeometryIT extends AbstractAsyncEngineConnectorTest {
             stopConnector();
         }
         finally {
-            Testing.Files.delete(SCHEMA_HISTORY_PATH);
+            Files.delete(SCHEMA_HISTORY_PATH);
         }
     }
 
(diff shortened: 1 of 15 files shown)


rk3rn3r:
@jpechane Looking at the Cassandra PR that you mentioned in the last triaging call, I felt like it could make sense to not only do this for MySQL Geometry but maybe for other fields that potentially conflict?
There are some schema parameters that are handled by the different Avro registries (feels like some copied the other) like scale and precision values. The ones that we used to differ schemas of different fields are not incorporated in the cache key and storage, leading to those issues.
Maybe we can identify those datatype for all connectors?

In general I suggest this as a temporary fix as the schema name for those field look ugly.
Maybe it is even better to use the logical coordinate (`server.schema.table.field`) as schema names until it is fixed?
Wdyt?

What also might be missing is a Apicurio client test for Postgres. Should I add one?

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlGeometryIT.java
```diff
@@ -218,7 +231,7 @@ else if (i == 2) {
         }
     }
 
-    private DatabaseGeoDifferences databaseGeoDifferences(boolean mySql5) {
+    public static DatabaseGeoDifferences databaseGeoDifferences(boolean mySql5) {
```
rk3rn3r, line 234: I actually did not change this. This interface is still private, Github shows an outdated commit content here. :shrug:

mfvitale:
> There are some schema parameters that are handled by the different Avro registries (feels like some copied the other) like scale and precision values.

Yes, issue was reported with https://issues.redhat.com/browse/DBZ-6836. Opened and issue to [Confluent](https://github.com/confluentinc/schema-registry/issues/2740) and [Apicurio](https://github.com/Apicurio/apicurio-registry/issues/3606) but no answer.

debezium-core/src/main/java/io/debezium/data/geometry/Geometry.java
```diff
@@ -33,9 +33,9 @@ public class Geometry {
      *
      * @return the schema builder
      */
-    public static SchemaBuilder builder() {
+    public static SchemaBuilder builder(String columnName) {
         return SchemaBuilder.struct()
-                .name(LOGICAL_NAME)
+                .name(LOGICAL_NAME + "__" + columnName) // temporary fix for DBZ-6964
```
rk3rn3r, line 38: This is the essence of the change. I don't feel super happy about it and looking at the Cassandra PR https://github.com/debezium/debezium-connector-cassandra/pull/121 there might be other fields affected in general.
Looking at the registry code it seems that all types that can collide, collide when there are other parameters used than things like precision and scale. For example DECIMAL types are handled here:
https://github.com/Apicurio/apicurio-registry/blob/f2d1f06cae0b709acf3f6d4edb982f5775b50fa9/utils/converter/src/main/java/io/apicurio/registry/utils/converter/avro/AvroData.java#L886-L896

rk3rn3r:
Thx @mfvitale! You can look at the linked code in my previous reply (the one after yours). It shows the code and which datatypes/fields get a "special" handling. It should be possible to fix the issue somewhere around there too to incorporate other parameters not only those "magical" ones. wdyt?

mfvitale:
> Thx @mfvitale! You can look at the linked code in my previous reply (the one after yours). It shows the code and which datatypes/fields get a "special" handling. It should be possible to fix the issue somewhere around there too to incorporate other parameters not only those "magical" ones. wdyt?

Honestly, I don't know. My first clue of the issue was about this lines. https://github.com/confluentinc/schema-registry/blob/cede655d600161767be861f34c47516c76923594/avro-data/src/main/java/io/confluent/connect/avro/AvroData.java#L1172

debezium-core/src/main/java/io/debezium/data/geometry/Geometry.java
Naros: Making logical names dynamic will break the JDBC sink connector.   If we go with this approach, we'll need to add a check for "prefixed geometry logical names" to sanitize the incoming data to correctly lookup and resolve target columns accordingly.

rk3rn3r: @Naros I agree. But let me challenge this. Is this the right way to to it? Couldn't we use `"__debezium.source.column.type" : "GEOMETRY",` schema parameter instead? With this change we can guarantee that all parameters are stored in the registry. Then we can rely on our custom fields to decide on the Debezium datatype instead of relying on the default behavior of the registry.

jpechane:
@rk3rn3r Thanks for the PR. It is definitely omething that works but I am not keen in having it in the core - given we are effectivelly patching registry bug. Also it is unnecessary for JSON. Is there a chance the implementation can be change along the lines of https://debezium.zulipchat.com/#narrow/stream/348251-community-cassandra/topic/issues.20in.20pushing.20the.20row.20if.20we.20have.20Datatype.20.3A.20set.3Ctext.3E/near/421428439 and intead of having it in the core it would be an SMT?

debezium-core/src/main/java/io/debezium/data/geometry/Geometry.java
Naros: We could for Debezium events, but don't forget the JDBC sink can be used by non-Debezium producers.

We will need to look at the schema type no matter what for the latter use case; I just want to be mindful of any changes we make here and how that could impact performance, specifically given how hard Mario worked on improving the sink's performance metrics already; no need to walk that back.

rk3rn3r: @Naros  `and resolve target columns accordingly`

I think the jdbc sink resolves the column name from `field.name()` not from the schema, does it? The issue will be with identifying the correct data type:

Looking at the code in `iGeneralDatabaseDialect#getSchemaType()`

```
      if (!Objects.isNull(schema.name())) {
            final Type type = typeRegistry.get(schema.name());
            if (!Objects.isNull(type)) {
                LOGGER.trace("Schema '{}' resolved by name from registry to type '{}'", schema.name(), type);
                return type;
            }
        }
        if (!Objects.isNull(schema.parameters())) {
            final String columnType = schema.parameters().get("__debezium.source.column.type");
            if (!Objects.isNull(columnType)) {
                final Type type = typeRegistry.get(columnType);
```
The schema is loaded via the record's `schema.name()` (which will work because the schema comes from `field.schema()`.
But in that first case `typeRegistry.get(schema.name());` we would have to match only the first name or cleanup the key in the cache. I don't know, it is an unfortunate fix.

rk3rn3r:
@jpechane There is a MySQLSchemaFactory, similar to the fix done in the Cassandra PR, but I assume that, but it is imo not used at all at the moment. Should I move the fix there? Imo other data types will be affected too.

Naros:

Hi @rk3rn3r, please sign off all commits with:

```
git commit -s
```

PR #5412, opened 2024-03-20
Dbz 7653 adjust tech preview designations for 2.5 productization

roldanbob:
[DBZ-7653](https://issues.redhat.com/browse/DBZ-7653)

Adds and removes conditionalized DP and TP boilerplate text to reflect changes in release status of MongoDB, MySQL, and PostgreSQL connector features. 

Backport to 2.5 (bc 2.6 is not targeted for downstream release, I assume that no bp is needed there, unless we have strong feelings about keep branch content in sync).


jpechane:
@roldanbob Applied, thanks!

PR #5413, opened 2024-03-21
DBZ-7690 Fix history comparisons for SERVER_ID_KEY and TIMESTAMP_KEY

Naros:
https://issues.redhat.com/browse/DBZ-7690


jpechane:
@Naros Good catch, applied!

PR #5416, opened 2024-03-21
DBZ-7695 add `_id` field to tombstone rewrite events, for mongodb delete events

ismailsimsek:


jpechane:
@ismailsimsek Thanks for the PR. Could you please add a config option that will enable this behaviour so it is not used by default?

ismailsimsek:
> @ismailsimsek Thanks for the PR. Could you please add a config option that will enable this behaviour so it is not used by default?

@jpechane which config key would you suggest? added 4 option below

```
transforms=unwrap,...
transforms.unwrap.type=io.debezium.connector.mongodb.transforms.ExtractNewDocumentState
transforms.unwrap.drop.tombstones=false
transforms.unwrap.delete.handling.mode=drop
-- options ------------
transforms.unwrap.mongodb.delete.tombstone.handling.mode.rewrite-with-id
transforms.unwrap.mongodb.delete.tombstone.handling.rewrite-with-id
-- using `includes`
transforms.unwrap.mongodb.delete.tombstone.handling.mode.rewrite-icludes-id
transforms.unwrap.mongodb.delete.tombstone.handling.rewrite-icludes-id
------------------------
```

ismailsimsek:
@jpechane if you have time.. its ready for a review.

ismailsimsek:
@Naros  @jpechane its ready for review.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/ExtractNewDocumentState.java
```diff
@@ -216,6 +226,9 @@ public R doApply(R record) {
         // add rewrite field
         if (extractRecordStrategy.isRewriteMode()) {
             valueDocument.append(DELETED_FIELD, new BsonBoolean(isDeletion));
+            if (rewriteTombstoneDeletesWithId && !valueDocument.containsKey("_id") && keyDocument.containsKey("id")) {
+                valueDocument.append("_id", keyDocument.get("id"));
```
Naros, line 230: One question: is there any specific reason we opted for `_id` with one underscore when other fields like `op` and the `deleted` field are prepended with `__op` (who underscores)?  I'd prefer consistency with `__id` unless there is a technical reason why it must be `_id` .  @ismailsimsek @jpechane wdyt?

ismailsimsek: good point, thats because with mongodb somehow id field flattened as `_id` .
can confirm that for snapshot events, example below

```
EmbeddedEngineChangeEvent [

key={"schema":{"type":"struct","fields":[{"type":"int32","optional":true,"field":"id"}],"optional":false},"payload":{"id":105}}, 

value={"schema":{"type":"struct","fields":[{"type":"int64","optional":true,"field":"_id"},{"type":"string","optional":true,"field":"name"},{"type":"string","optional":true,"field":"description"},{"type":"double","optional":true,"field":"weight"},{"type":"int32","optional":true,"field":"quantity"},{"type":"boolean","optional":true,"field":"__deleted"},{"type":"string","optional":true,"field":"__op"},{"type":"int64","optional":true,"field":"__source_ts_ms"},{"type":"string","optional":true,"field":"__db"}],"optional":false,"name":"testc.inventory.products"},"payload":{"_id":105,"name":"hammer","description":"14oz carpenter's hammer","weight":0.875,"quantity":5,"__deleted":false,"__op":"r","__source_ts_ms":0,"__db":"inventory"}}, sourceRecord=SourceRecord{sourcePartition={rs=rs0, server_id=testc}, sourceOffset={sec=0, ord=-1, initsync=true}} ConnectRecord{topic='testc.inventory.products', kafkaPartition=null, key=Struct{id=105}, keySchema=Schema{STRUCT}, value=Struct{_id=105,name=hammer,description=14oz carpenter's hammer,weight=0.875,quantity=5,__deleted=false,__op=r,__source_ts_ms=0,__db=inventory}, valueSchema=Schema{testc.inventory.products:STRUCT}, timestamp=null, headers=ConnectHeaders(headers=)}" 

]
```

jpechane: `_id` is MongoDB name for PK so it makes sense to keep it as is.

jpechane:
@ismailsimsek Applied, thanks a lot! Could you please open a separate PR with docs updated?

PR #5422, opened 2024-03-26
DBZ-7497 Add a configuration based snapshot modes configurable via connector properties

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7497

Provides an implementation of `Snapshotter` SPI that permits to configure it through connector configuration properties.


jpechane:
@mfvitale Applied, thanks a lot! Could you please just check the compatibility between this mode and blocking snapshot?

mfvitale:
> @mfvitale Applied, thanks a lot! Could you please just check the compatibility between this mode and blocking snapshot?

As how now the blocking snapshot always snapshot schema and data, so snapshotter is not used here. On the other hand the locking and query part will be used.

PR #5428, opened 2024-03-27
DBZ-7711 Adjust fakeDNS startup conditional for Max OS

obabec:
https://issues.redhat.com/browse/DBZ-7711


jcechace (approved):

jcechace:
Fix seems to work. Local (MacOS) failures are cause by a the fact that unlike docker, podman will kill connection after 10s of inactivity (and TC has a pool of connections which are reused for 3 minutes). 

See https://github.com/containers/podman/issues/17640

A workaround is to add the following to `/etc/containers/containers.conf` inside your podman machine

```
[engine]
service_timeout=0
```

PR #5436, opened 2024-03-28
DBZ-7717 Move LogPositionValidator outside JdbcConnection

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7717


jpechane:
@mfvitale Applied, thanks!

PR #5437, opened 2024-03-28
DBZ-7698 Add custom transaction metadata

twthorn:
### Changes

Add ability for customizing transaction metadata. See this PR https://github.com/debezium/debezium-connector-vitess/pull/188 on how this can work for the Vitess Connector and be used for adding ordering metdata.

Adds two new configs for overriding the transaction struct used (similar to how source info struct makers can be overridden), and for overriding the transaction context used (to allow for custom tracking of other transaction info that is not just transaction ID).

### Additional Notes

SQL server & other tests appear to be flakey, and are failing for other PRs as well. All other tests passed.


debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -710,6 +713,24 @@ public static SnapshotQueryMode parse(String value, String defaultValue) {
             .withDescription("Enables transaction metadata extraction together with event counting")
             .withDefault(Boolean.FALSE);
 
+    public static final Field TRANSACTION_STRUCT_MAKER = Field.create("transaction.struct.maker")
```
jpechane, line 716: I don't think we need these two classes to be configurable by user a least now. It is fine if the connectors can override them but I'd keep it closed now.

debezium-core/src/main/java/io/debezium/pipeline/txmetadata/TransactionStructMaker.java
```diff
@@ -0,0 +1,44 @@
...
+    String DEBEZIUM_TRANSACTION_ID_KEY = "id";
+    String DEBEZIUM_TRANSACTION_TOTAL_ORDER_KEY = "total_order";
+    String DEBEZIUM_TRANSACTION_DATA_COLLECTION_ORDER_KEY = "data_collection_order";
+    String DEBEZIUM_TRANSACTION_STATUS_KEY = "status";
+    String DEBEZIUM_TRANSACTION_EVENT_COUNT_KEY = "event_count";
+    String DEBEZIUM_TRANSACTION_COLLECTION_KEY = "data_collection";
+    String DEBEZIUM_TRANSACTION_DATA_COLLECTIONS_KEY = "data_collections";
+    String DEBEZIUM_TRANSACTION_TS_MS = "ts_ms";
+
+    void setSchemaNameAdjuster(SchemaNameAdjuster adjuster);
```
jpechane, line 27: This should not be part of the interface but the `adjuster` should be passed as constructor.

debezium-core/src/main/java/io/debezium/pipeline/txmetadata/TransactionStructMaker.java
```diff
@@ -0,0 +1,44 @@
...
+
+import java.time.Instant;
+
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.Struct;
+
+import io.debezium.pipeline.spi.OffsetContext;
+import io.debezium.schema.SchemaNameAdjuster;
+
+public interface TransactionStructMaker {
```
jpechane, line 16: The interface and methods needs to be Javadoced.

debezium-core/src/main/java/io/debezium/pipeline/txmetadata/TransactionStructMaker.java
```diff
@@ -0,0 +1,44 @@
...
+    String DEBEZIUM_TRANSACTION_DATA_COLLECTION_ORDER_KEY = "data_collection_order";
+    String DEBEZIUM_TRANSACTION_STATUS_KEY = "status";
+    String DEBEZIUM_TRANSACTION_EVENT_COUNT_KEY = "event_count";
+    String DEBEZIUM_TRANSACTION_COLLECTION_KEY = "data_collection";
+    String DEBEZIUM_TRANSACTION_DATA_COLLECTIONS_KEY = "data_collections";
+    String DEBEZIUM_TRANSACTION_TS_MS = "ts_ms";
+
+    void setSchemaNameAdjuster(SchemaNameAdjuster adjuster);
+
+    Struct prepareTxStruct(OffsetContext offsetContext, long dataCollectionEventOrder, Struct value);
```
jpechane, line 29: As this is now part of a distinct interface the name of methods could be improved.

debezium-core/src/main/java/io/debezium/pipeline/txmetadata/BasicTransactionInfo.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.pipeline.txmetadata;
+
+public class BasicTransactionInfo implements TransactionInfo {
```
jpechane, line 8: I'd call it `Default` instead if `Basic`. Same for struct maker

debezium-core/src/main/java/io/debezium/pipeline/txmetadata/TransactionInfo.java
```diff
@@ -0,0 +1,10 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.pipeline.txmetadata;
+
+public interface TransactionInfo {
```
jpechane, line 8: Please add JavaDoc

jpechane (changes requested):
@twthorn Thanks, I think the changes are OK from my PoV. I left couple of comments related to naming and documentation.

debezium-core/src/main/java/io/debezium/pipeline/txmetadata/TransactionContext.java
```diff
@@ -30,15 +30,15 @@
...
+    public static final String OFFSET_TRANSACTION_ID = TransactionStructMaker.DEBEZIUM_TRANSACTION_KEY + "_" + TransactionStructMaker.DEBEZIUM_TRANSACTION_ID_KEY;
+    private static final String OFFSET_TABLE_COUNT_PREFIX = TransactionStructMaker.DEBEZIUM_TRANSACTION_KEY + "_"
+            + TransactionStructMaker.DEBEZIUM_TRANSACTION_DATA_COLLECTION_ORDER_KEY + "_";
     private static final int OFFSET_TABLE_COUNT_PREFIX_LENGTH = OFFSET_TABLE_COUNT_PREFIX.length();
 
-    private String transactionId = null;
-    private final Map<String, Long> perTableEventCount = new HashMap<>();
-    private final Map<String, Long> viewPerTableEventCount = Collections.unmodifiableMap(perTableEventCount);
-    private long totalEventCount = 0;
+    public String transactionId = null;
```
jpechane, line 38: Please don't break the encapsulation and keep it private.

twthorn:
@jpechane Can you re-review when you get the chance? Thank you

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
jpechane: If we'd like to keep this configurable then it should be part of an SPI and only one config option hould be used instead of two. In that case it might be better to introduce a class that would be a factory for those SPI interface implementations. `io.debezium.connector.SourceInfoStructMaker` was intended more as a utility vehicle to enable `SourceInfo` migration for version 1.

Thinking loud and summoning @Naros. Maybe it is about time that we'd open `io.debezium.service.spi.ServiceRegistry` to be configurable from a connector confiuration so user could inject its own implementation?

Naros: Yep, I could imagine this being done one of two ways @jpechane (maybe others):

The first way would be a very simple approach where the user supplies a comma-separated list of services that should be spun up into the registry, i.e.:
```
services=fully.qualified.service.a,fully.qualfiied.service.b
```
The other approach would be a more traditional configuration style like `converters` and `transforms` where the user supplies a `services` configuration that defines keys, and then we fetch the specific service and perhaps some optional configuration options via lookups.

In both cases, these services would extend a `ServiceContributor` contract that ultimately acts as a thin wrapper around the construction of the service, so that we could define things like this:

```java
public interface SomeFancyStructMakerService extends Service {
}

public class UserSuppliedFancyStructMakerService implements SomeFancyStructMakerService {
}
```

It's verbose, but it keeps a solid SoC between the contributor aspect and the actual implementation, allowing Debezium to initial the registry with a default/standard implementation and the user overriding that service with their own.  wdyt?

twthorn: I refactored to a single factory spi interface/class, so we only need one config option, which can be overridden in a connector ([example](https://github.com/debezium/debezium-connector-vitess/pull/188/files#diff-1b9011e9a90b9250c8885c36b838ac867f597b3726cf037ec1fa0aefb47d8221R430)).

The shared service registry changes sound useful, but probably would be outside of the scope of this PR. Once they are added, we can refactor transaction metadata to use the shared service registry.

twthorn:
@jpechane Updated based on your feedback. Ready for re-review

jpechane:
@twthorn Hi, there was a major refactoring of MySQL/MariaDB code finished. Could you please rebase the PR on the latest main? Thanks for understanding.

WRT your proposal about the config - yes I agree let's do it in a separate PR. When the rebase on the new main is done the PR is ready to be merged.

twthorn:
Rebased onto main

jpechane:
@twthorn Applied, thanks for the patience and good work!

PR #5446, opened 2024-04-02
DBZ-7729 Fix shouldNotStreamWhenUsingSnapshotModeInitialOnly test assertion due to log message change

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7729


jpechane:
@mfvitale Applied, thanks

PR #5455, opened 2024-04-04
DBZ-7741 Avoid write-behind cache clear error

Naros:
https://issues.redhat.com/browse/DBZ-7741


jpechane:
@Naros Applied, thanks

PR #5458, opened 2024-04-05
DBZ-7740 Fixed NOAUTH authentication required error while initializing Redis connection

selman-genc-alg:
This PR aims to fix NOAUTH authentication required error thrown when redis database index property is specified in the configuration

You can see the details of the issue here:
https://issues.redhat.com/browse/DBZ-7740

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -598,3 +598,4 @@ Fr0z3Nn
 Xianming Zhou
 Akula
 Nick Golubev
+Selman Genç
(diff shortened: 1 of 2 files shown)


selman-genc-alg:
@jpechane thanks for the feedback, I will be doing these changes.
You suggest removing client.auth after Jedis is created, should I only leave the ping call there? and if ping doesn't return PONG should we try authenticate again or does ping() throw if it can't get a response?

jpechane:
IMHO the `ping` could be called regrdless of whether the auth is set or not.

selman-genc-alg:
@jpechane can you please review again?
also, any idea why am I getting an error on commit message even though I added the issue ID as prefix?

jpechane (approved):
@selman-genc-alg Thanks, I squased the commits into a single one. The PR is ready to be merged after CI run is finished.

jpechane:
@selman-genc-alg Applied, thanks a lot

PR #5460, opened 2024-04-05
DBZ-7732 Add ConnectorSpecific annotation to manage multiple SPI connector specific implementation

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7732

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/snapshot/query/SelectAllSnapshotQuery.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/snapshot/query/SelectAllSnapshotQuery.java
@@ -9,9 +9,12 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import io.debezium.annotation.ConnectorSpecific;
 import io.debezium.config.CommonConnectorConfig;
+import io.debezium.connector.mongodb.MongoDbConnector;
 import io.debezium.snapshot.spi.SnapshotQuery;
 
+@ConnectorSpecific(connector = MongoDbConnector.class)
 public class SelectAllSnapshotQuery implements SnapshotQuery {
 
     @Override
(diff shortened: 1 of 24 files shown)


jpechane:
@mfvitale I wonder if we can add a test to Debezium Server rep verifying this fix. Ideally it should have multiple connectors on the classpath and test two of them to make sure they are working. If necessary a new module would me introduced for it. WDYT?

mfvitale:
> @mfvitale I wonder if we can add a test to Debezium Server rep verifying this fix. Ideally it should have multiple connectors on the classpath and test two of them to make sure they are working. 

Yeah, I was thinking to add it too but I need to understand how tests works on Debezium Server. Any point where to start? Seems that as how now it just start PostgreSQL, isn't it?

> If necessary a new module would me introduced for it.

Can you please explain better?

jpechane:
Well, the idea is that there will be a new module added to `debezium-server` project, like `debezium-server-testing`. This one will have multiple connectors on the classpath to make sure there are clashing classes. There will be one basic test for for example PostgreSQL and on for MySQL - generally get one record from snapshot, one from streaming, check there are no errors. Such tests should exhibit the existing bug and start passing with the fix.

jpechane:
The other option is to try to add it to the core library - a new module under `debezium-testing` and test it at Debezium Engine level as again I suppose there should be the issue if there are multiple connectors on the module classpath, should not it?

mfvitale:
> The other option is to try to add it to the core library - a new module under `debezium-testing` and test it at Debezium Engine level as again I suppose there should be the issue if there are multiple connectors on the module classpath, should not it?

Yes and maybe this is a better solution. But I was looking at `debezium-testing-system` module that seems to do exact this thing but seems to be designed to run manually. Is there any place where we execute it?

jpechane:
`debezium-testing-system` is module that contains product tests and these are executed on product build candidates - @obabec can tell you more

mfvitale:
@jpechane At the end I added it to Debezium Server since it can be the base for further tests on it. 

https://github.com/debezium/debezium-server/pull/85

I have verified that without the fix the tests fails.

jpechane:
@mfvitale Applied, thanks!

PR #5461, opened 2024-04-05
DBZ-7744 RS name is not required when cluster type is not REPLICA_SET

jcechace:
https://issues.redhat.com/browse/DBZ-7744


Naros (approved):

Naros:
Applied to main (=2.7) and branch 2.6.

PR #5463, opened 2024-04-05
DBZ-7754 Removes `database.names` entry from connector yaml example

roldanbob:
[DBZ-7754](https://issues.redhat.com/browse/DBZ-7754)

During review of the downstream docs, QE noticed that the connector YAML example in the deployment instructions still included a reference to the `database.names` property, even though that property is no longer supported. This change removes that line from the example. 

This change applies to the downstream documentation only and does not affect the version that's published to the community site.


roldanbob:
2.5 PR: #5464

PR #5466, opened 2024-04-05
DBZ-7753 Oracle converter for `RAW` to `STRING` conversion

Naros:
https://issues.redhat.com/browse/DBZ-7753


jpechane:
@Naros Applied, thanks! Could you please add a docs for it too in a separate PR? Given there are already other converters as well we might need to introduce a new section for OOTB custom converters.

PR #5469, opened 2024-04-08
DBZ-7747 Add parallel maven build to testsuite pod

novotnyJiri:
https://issues.redhat.com/browse/DBZ-7747


obabec (changes requested):
I am firmly against adding this option here. This image is mainly used for debugging, so I think it would do more harm than good.

obabec (approved):

PR #5472, opened 2024-04-09
[ci] Use github.ref_name on pushes for sibling checkouts

Naros:
@ani-sha can you double-check my logic here, the idea is that on pushes, this would make sure that we check out the sibling repository with the same branch name as in the core repository, i.e. `main`, `2.6`, etc.  

@jpechane if Anisha gives this the green light, we should backport this to 2.6.


ani-sha (approved):
@Naros LGTM

jpechane:
@Naros Applied, thanks!

PR #5473, opened 2024-04-10
DBZ-7765: Use different query for validating log position for SQLServer

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7765

### DIFF
--- a/debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
+++ b/debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
@@ -1044,7 +1044,7 @@ public abstract class CommonConnectorConfig {
             .withDescription(
                     "When 'snapshot.mode' is set as configuration_based, this setting permits to specify whenever the data should be snapshotted or not in case of error.");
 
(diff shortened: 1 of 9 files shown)


debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -1043,6 +1044,16 @@ public static SnapshotQueryMode parse(String value, String defaultValue) {
             .withDescription(
                     "When 'snapshot.mode' is set as configuration_based, this setting permits to specify whenever the data should be snapshotted or not in case of error.");
 
+    public static final Field LOG_POSITION_CHECK_ENABLED = Field.create("log.position.check.enable")
```
jpechane, line 1047:
```suggestion
    public static final Field LOG_POSITION_CHECK_ENABLED = Field.createInternal("log.position.check.enable")
```

let's keep it undocumented at the start till we evaluate if it is something necessary

jpechane:
@mfvitale Applied, thanks

jpechane:
@mfvitale Could you please prepare backport to 2.6? Thanks

mfvitale:
@jpechane Done.

PR #5474, opened 2024-04-10
DBZ-7763 Remove record processing thread cap

vjuranek:
and use cached thread pool as the default.

https://issues.redhat.com/browse/DBZ-7763
https://issues.redhat.com/browse/DBZ-7764


jpechane:
@vjuranek Applied, thanks!

PR #5475, opened 2024-04-10
DBZ-7759 Ensure error handling of oversized events swallowed by BufferedChangeStreamCursor thread

ani-sha:
closes https://issues.redhat.com/browse/DBZ-7759

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
@@ -12,6 +12,7 @@ import org.bson.BsonString;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.mongodb.MongoException;
 import com.mongodb.client.ChangeStreamIterable;
 import com.mongodb.client.MongoClient;
 import com.mongodb.client.model.changestream.ChangeStreamDocument;
@@ -117,6 +118,10 @@ public class MongoDbStreamingChangeEventSource implements StreamingChangeEventSo
                     return;
                 }
             }
+        }
+        catch (MongoException e) {
+            LOGGER.error("Error while reading change stream", e);
+            errorHandler.setProducerThrowable(e);
         }
     }
 
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
@@ -22,6 +22,8 @@ import org.bson.BsonDocument;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
+import com.mongodb.MongoChangeStreamException;
+import com.mongodb.MongoException;
 import com.mongodb.ServerAddress;
 import com.mongodb.ServerCursor;
 import com.mongodb.client.ChangeStreamIterable;
@@ -225,15 +227,23 @@ public class BufferingChangeStreamCursor<TResult> implements MongoChangeStreamCu
 
         private Optional<ResumableChangeStreamEvent<TResult>> fetchEvent(MongoChangeStreamCursor<ChangeStreamDocument<TResult>> cursor) {
             var beforeEventPollTime = clock.currentTimeAsInstant();
-            var document = cursor.tryNext();
+            ChangeStreamDocument<TResult> document;
+            try {
+                document = cursor.tryNext();
+            }
+            catch (MongoException e) {
+                running.set(false);
+                throw new MongoChangeStreamException("Error while fetching change stream event" + e.getMessage());
+            }
             metrics.onSourceEventPolled(document, clock, beforeEventPollTime);
             throttleIfNeeded(document);
 
             // Only create resumable event if we have either document or cursor resume token
             // Cursor resume token may be `null` in case of issues like SERVER-63772, and situations called out in the Javadocs:
             // > resume token [...] can be null if the cursor has either not been iterated yet, or the cursor is closed.
+            ChangeStreamDocument<TResult> finalDocument = document;
             return Optional.<ResumableChangeStreamEvent<TResult>> empty()
-                    .or(() -> Optional.ofNullable(document).map(ResumableChangeStreamEvent::new))
+                    .or(() -> Optional.ofNullable(finalDocument).map(ResumableChangeStreamEvent::new))
                     .or(() -> Optional.ofNullable(cursor.getResumeToken()).map(ResumableChangeStreamEvent::new));
         }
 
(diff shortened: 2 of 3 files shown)


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -225,15 +227,23 @@ private void fetchEvents(MongoChangeStreamCursor<ChangeStreamDocument<TResult>>
...
         private Optional<ResumableChangeStreamEvent<TResult>> fetchEvent(MongoChangeStreamCursor<ChangeStreamDocument<TResult>> cursor) {
             var beforeEventPollTime = clock.currentTimeAsInstant();
-            var document = cursor.tryNext();
+            ChangeStreamDocument<TResult> document;
+            try {
+                document = cursor.tryNext();
+            }
+            catch (MongoException e) {
+                running.set(false);
+                throw new MongoChangeStreamException("Error while fetching change stream event" + e.getMessage());
```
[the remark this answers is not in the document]
jpechane: Do we need to use `MongoChangeStreamException`? If not, let's use another one so the original exception is added to the stack.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -297,8 +307,18 @@ public BufferingChangeStreamCursor<TResult> start() {
...
+                lastResumeToken = event.resumeToken;
+            }
+            else {
+                if (!fetcher.isRunning() && fetcher.isEmpty()) {
+                    throw new MongoChangeStreamException("Failed to fetch event from change stream cursor, cursor is closed");
+                }
+            }
+        }
+        catch (MongoException e) {
+            throw new MongoException(e.getMessage());
```
jpechane, line 321: Please add original exception to the stack

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/MongoDbConnectorIT.java
```diff
@@ -826,6 +826,66 @@ public void shouldConsumeAllEventsFromDatabase() throws InterruptedException, IO
...
+            var updateDocument = new Document("$set", new Document("value", afterValue));
+            updateDocument(dbName, collName, new Document("_id", docId), updateDocument);
+        }
+
+        // Consume records
+        SourceRecords records = consumeRecordsByTopic(1);
+        assertThat(records.allRecordsInOrder().size()).isEqualTo(0);
+
+        // Stop the connector and check for error message
+        stopConnector(value -> assertThat(logInterceptor.containsErrorMessage("Error while reading change stream")).isTrue());
```
jpechane, line 885: I think the error message is different

jpechane (changes requested):
@ani-sha Thanks for the PR, it is a good start but needs some polish.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/MongoDbConnectorIT.java
ani-sha: It's the error from `MongoDbStreamingChangeEventSource#readChangeStream` catch clause.

jpechane: I don't see such error message there.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbStreamingChangeEventSource.java
```diff
@@ -118,6 +119,10 @@ private void readChangeStream(MongoClient client, ChangeEventSourceContext conte
                 }
             }
         }
+        catch (MongoException e) {
```
ani-sha, line 122: @jpechane It's here ^^

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
jcechace: Definitively Could use DebeziumException as a wrapper

jcechace:
I don't like the approach where the exception is propagated via a cascade of throws up to the main streaming loop. Utilising a reference like in https://github.com/nj4x/debezium/commit/4ca19aa90d4552833c595fcf4a147b14f9b72d58 seems as a cleaner solution (except that commits forgets to mark the thread as stopped). 

Could we do something like that instead?

jcechace (changes requested):
I don't like the approach where the exception is propagated via a cascade of throws up to the main streaming loop. Utilising a reference like in https://github.com/nj4x/debezium/commit/4ca19aa90d4552833c595fcf4a147b14f9b72d58 seems as a cleaner solution (except that commits forgets to mark the thread as stopped). 

Could we do something like that instead?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -198,7 +224,8 @@ public void run() {
                 noMessageIterations = 0;
                 fetchEvents(cursor);
             }
-            catch (InterruptedException e) {
+            catch (Throwable e) {
+                error.set(e);
                 throw new DebeziumException("Fetcher thread interrupted", e);
             }
```
jcechace, line 230: This needs to stay as two separate branches. With `InterruptedException` we want to end right away. Whereas on other exceptions we want to keep the opportunity to drain the current buffer queue. 

So we should set the error and mark the fetcher as not running.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -163,14 +165,38 @@ public boolean isRunning() {
...
         public void close() {
             running.set(false);
         }
 
         public ResumableChangeStreamEvent<TResult> poll() {
             var event = queue.poll();
-            if (event != null) {
+            if (event == null) {
+                if (hasError()) {
+                    close();
```
jcechace, line 195: I think the fetter should stop itself on error.

jcechace (changes requested):
@ani-sha Almost there... left few small comments. 

BTW: when you are at it. The referenced commit from our user added some useful logs. Could you incorporate those as well please?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -199,8 +224,14 @@ public void run() {
                 fetchEvents(cursor);
             }
             catch (InterruptedException e) {
+                LOGGER.error("Fetcher thread interrupted", e);
                 throw new DebeziumException("Fetcher thread interrupted", e);
             }
+            catch (Throwable e) {
+                error.set(e);
+                LOGGER.error("Fetcher thread has failed", e);
+                running.set(false);
```
jcechace, line 233: No need, it's closed in the finally block.

jcechace (approved):
LGTM... tests improvements can be submitted in a new PR

jpechane:
@ani-sha Applied, thanks!

PR #5478, opened 2024-04-10
DBZ-6116 Rewrite to comply with style guidelines

roldanbob:
[DBZ-6116](https://issues.redhat.com/browse/DBZ-6116)

This documentation was previously merged, but never received my review/approval. 
Per style guidelines, `please` is not for use in _"technical information or in instructions. Terms of politeness are superfluous, convey the wrong tone for technical material, and are not regarded the same way in all cultures. "_
Also, updated the reference to the database to refer to its correct trade name, and made other edits intended to enhance clarify.


roldanbob:
2.5 MR: #5479

PR #5485, opened 2024-04-12
DBZ-7758 Initial implementation of EhCache buffer

PavelVrtal:
### Work that still needs/could to be done before merge
- [ ]  use explicit [Ehcache Serializers](https://www.ehcache.org/documentation/3.8/serializers-copiers.html) (instead of implementing native Java `Serializable` as of now) to avoid breaking immutable pattern on some data classes
- [ ] only disk tier from Ehcache tiering options is used, so may be good to allow to configure heap tier as well
-- Combination of heap and disk which we wanted for our use case turned out not to be working well, as the limit put on the heap was breached anyway (couldn't find out why) which led to OutOfMemoryError (we needed to limit the heap tier to less than 100 MB).
- [ ] upgrade Ehcache dependency to 3.10: I ended up using Ehcache version 3.8.1 (I've tried also the latest one: 3.10, but encountered problems with dependencies which I couldn't resolve after spending some time on it)

### FYI
- Some refactoring had to be done because up until the introduction of Ehcache, all the implementations were `Map`-based, while Ehcache's `Cache` interface doesn't provide some of `Map`'s methods.

### DIFF
--- a/debezium-connector-oracle/pom.xml
+++ b/debezium-connector-oracle/pom.xml
@@ -51,6 +51,11 @@
             <artifactId>infinispan-core-jakarta</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.ehcache</groupId>
+            <artifactId>ehcache</artifactId>
+            <version>3.8.1</version>
+        </dependency>
+        <dependency>
             <groupId>org.infinispan</groupId>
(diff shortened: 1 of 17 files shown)


jpechane:
Closing, merged via https://github.com/debezium/debezium/pull/5741

PR #5493, opened 2024-04-18
DBZ-7790 fix Debezium oracle connector lost data of the update operat…

leoloel:
…ion when the table has no primary key and contains lob type

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/TransactionCommitConsumer.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/TransactionCommitConsumer.java
@@ -158,6 +158,12 @@ public class TransactionCommitConsumer implements AutoCloseable, BlockingConsume
             return;
         }
 
+        if(table.primaryKeyColumnNames().isEmpty()){
+            LOGGER.debug("\tEvent for table {} has no primary key, dispatching.", table.id());
+            dispatchChangeEvent(event);
+            return;
+        }
+
         if (!tryMerge(accumulatorEvent, event)) {
             prepareAndDispatch(accumulatorEvent);
             if (rowId.equals(currentLobDetails.rowId)) {


Naros:
Hi @leoloel, thanks for the PR.  Can you please also add a test case for this into the `OracleClobDataTypesIT`, `OracleBlobDataTypesIT` and `OracleXmlDataTypesIT` tests where we have each of these as a part of a keyless table.

leoloel:
Yes, the test has been run. The command is  `mvn clean install -pl "debezium-connector-oracle" -am -P"oracle-tests" -D"database.pdb.name=ORCLPDB1"`. Some unrelated tests failed. I tried, before modifying, some of the tests also failed.

Naros:
Thanks @leoloel, I'll look it in more detail then on Monday 👍

Naros:
So, looking at the errors, the issue is where we are dispatching.

In the case of `OracleClobDataTypeIT`, we have a table with a single `CLOB` column.  In this example, Oracle will write several entries to the transaction logs depending on the size of the CLOB data.

### CLOB data less-than 2k

* INSERT INTO table (clob_field) values (EMPTY_CLOB());
* UPDATE table SET clob_field = '...';

### CLOB data greater-than 2k

* INSERT INTO table (clob_field) values (EMPTY_CLOB());
* SELECT_LOB_LOCATOR
* Series of LOB_WRITE
* LOB_END

Your Jira issue is focusing on just the aspect of an `UPDATE`, and in that case, yes we'd see only two events, but only in the context of where the CLOB data is small enough to be stored in the data page rather than in LOB storage when it exceeds 2k.  Therefore, what you're proposing simply won't work for `INSERT` events nor `UPDATE` events with a payload greater than 2k.

Unfortunately, LogMiner does not provide sufficient context.  In the past, I looked at `SEQUENCE#`, `RBASQN`, and `RS_ID` with `SSN` and determined these were not reliable enough for all use cases, but I can certainly take another pass to be absolutely sure.  

But what I am certain is the proposed solution isn't adequate, unfortunately.

leoloel:
So, Do you have any suggestions for modifications? I need help.

Naros:
hi @leoloel, so i didn't want you think this had fallen on deaf ears.  I've been looking at XStream and LogMiner for this to identify some common ground for handling this more efficiently.  I have identified one possibility but I am still checking to be sure this does not cause regressions.  So please stay tuned.

PR #5495, opened 2024-04-18
DBZ-7788 Support with statement/clause in binlog connectors

Naros:
https://issues.redhat.com/browse/DBZ-7788

@jpechane I separated these as only the one applicable for MySQL should be backported since the MariaDB one will only be relevant to 2.7 with the new connector grammar.


jpechane:
@Naros Applied, thanks! Just for the next time it is probably unnecessary to keep the testing query in the procedure definition. IMHO the behavour (and test) should be the same for standalone statements too.

jpechane:
BTW, thanks for splitting the commit. Backporting to 2.6 was seamless!

PR #5497, opened 2024-04-19
DBZ-7697 When new tables are added to MSSQL, its schema must be not captured.

Mon-MFEC:
The current Debezium captured the schema of newly added tables on MSSQL after initial deployment. This pull request will make that if a new tables are added on MSSQL, its schema must be not captured which be a follow the behavior as documented.


Mon-MFEC:
@jpechane  @mfvitale  @Naros 

new PR for [DBZ-7697](https://issues.redhat.com/browse/DBZ-7697)

PR #5499, opened 2024-04-19
DBZ-7797 Fail fast if surrogate key for incremental snapshot doesn't exist

joschi:
Fixes https://issues.redhat.com/browse/DBZ-7797


mfvitale:
Hi @joschi thanks for the PR. I have only one concern: snapshot context may remain in an inconsistent state because of  `snapshotDataCollection` and `correlationId` not cleared. Can you please check this scenario with an integration test?

jpechane:
@joschi Hi, would it be possible for you to provide the tests as recommend? Thanks

jpechane:
@joschi Applied, thanks! Apon further thinking if the exception is thrown the connector will fail so the state will be reset.

PR #5500, opened 2024-04-19
DBZ-7798 Fix documentation for `header` variable in filters

joschi:
Fixes https://issues.redhat.com/browse/DBZ-7798

### DIFF
--- a/documentation/modules/ROOT/pages/transformations/filtering.adoc
+++ b/documentation/modules/ROOT/pages/transformations/filtering.adoc
@@ -141,15 +141,12 @@ The following table lists the variables that {prodname} binds into the evaluatio
 |`keySchema` |Schema of the message key.|`org.apache.kafka.connect{zwsp}.data{zwsp}.Schema`
 |`valueSchema`|Schema of the message value.| `org.apache.kafka.connect{zwsp}.data{zwsp}.Schema`
 |`topic`|Name of the target topic.| String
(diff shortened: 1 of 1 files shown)


documentation/modules/ROOT/pages/transformations/filtering.adoc
```diff
@@ -141,15 +141,12 @@ The following table lists the variables that {prodname} binds into the evaluatio
 |`keySchema` |Schema of the message key.|`org.apache.kafka.connect{zwsp}.data{zwsp}.Schema`
 |`valueSchema`|Schema of the message value.| `org.apache.kafka.connect{zwsp}.data{zwsp}.Schema`
 |`topic`|Name of the target topic.| String
-|`headers`
-a|A Java map of message headers. The key field is the header name.
+|`header`|A Java map of message headers. The key field is the header name.
```
joschi, line 144: IMHO the name `headers` was better but changing it now would break existing filter expressions. 😞

joschi: https://github.com/debezium/debezium/blob/v2.6.1.Final/debezium-scripting/debezium-scripting/src/main/java/io/debezium/transforms/scripting/Jsr223Engine.java#L72

jpechane:
@joschi Applied, thanks!

PR #5503, opened 2024-04-22
DBZ-7793 Oracle 21 xstream tests doesn't finish in 5h

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7793

### DIFF
--- a/debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/util/TestHelper.java
+++ b/debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/util/TestHelper.java
@@ -5,6 +5,8 @@
  */
 package io.debezium.connector.oracle.util;
 
+import static io.debezium.relational.RelationalDatabaseConnectorConfig.QUERY_TIMEOUT_MS;
+
 import java.io.BufferedReader;
 import java.io.IOException;
 import java.io.InputStreamReader;
@@ -277,6 +279,7 @@ public class TestHelper {
         jdbcConfiguration.forEach(
                 (field, value) -> builder.with(OracleConnectorConfig.DATABASE_CONFIG_PREFIX + field, value));
 
+        builder.with(QUERY_TIMEOUT_MS, 0); // Fix Oracle 21 xstream tests
         builder.with(CommonConnectorConfig.TOPIC_PREFIX, SERVER_NAME);
         return builder;
     }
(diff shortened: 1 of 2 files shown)


Naros:
@obabec, something is awry here.

The logs for Oracle 21 Xstream show we begin the maven build at `07:29:57` and then the instance is terminated due to the 5h mark at `11:02:02`.  That's only 3.5 hours, so where is the other 90 minutes going?  Is that somehow related to the installation of the database image, downloading of some initial artifacts from maven central, or is it that we waited 90 minutes before a worker was free?

Naros:
Hi @mfvitale, talked with @obabec, and I believe this is because Oracle ARC can no longer checkpoint, see:
```
11:02:03                 out: Version 21.3.0.0.0   [io.debezium.connector.oracle.OracleConnection]
11:02:03                 out: 2024-04-23 11:02:03,019 INFO   TransactionMetadataIT||test  Loading the custom source info struct maker plugin: io.debezium.connector.oracle.OracleSourceInfoStructMaker   [io.debezium.config.CommonConnectorConfig]
12:26:58         Command returned '124'.
12:26:58         Test duration '5h' exceeded.
12:26:58         Append to file '/var/ARTIFACTS/work-oraclev5mufwqd/debezium-testing/tmt/plans/oracle/oracle/execute/data/guest/default-0/debezium-testing/tmt/tests/debezium/oracle-1/output.txt'.
12:26:59         Copy '/var/ARTIFACTS/work-oraclev5mufwqd/debezium-testing/tmt/plans/oracle/oracle/execute/data/guest/default-0/debezium-testing/tmt/tests/debezium/oracle-1' from the guest to '/'.
12:26:59         Run command: rsync -s -R -r -z --links --safe-links --protect-args -e 'ssh -oForwardX11=no -oStrictHostKeyChecking=no -oUserKnownHostsFile=/dev/null -oServerAliveInterval=60 -oServ
```

@jpechane because we continue to add more and more tests that generate excessive redo logs using XStream (because the entire database has to be supplemental logged indefinitely), while we can increase the archive destination up from 10GB, it does raise a question of at what point should we re-evaluate our tests and how they're run, and the data we actually generate within those tests.

jpechane:
@Naros Could we split the test in multiple runs and restart the database in between?

Naros:
> @Naros Could we split the test in multiple runs and restart the database in between?

That should be doable, but I think for the short-term the simple fix is to increase the volume size of the Oracle disk so that we can hold more redo logs for now.  Then once @obabec and I present and we discuss the move to Junit 5, I think that would be the ideal time to really focus on tags and organizing tests precisely for this type of scenario.  wdyt @jpechane ?

Naros:
Btw, I just opened a new PR with some other Oracle changes, lets see how the Xstream images work now that Ondrej and I have those sized a bit larger for the test suites :crossed_fingers:

jpechane:
@Naros If enlarging the volume is not an issue then yes, definitely fo for it.

PR #5504, opened 2024-04-22
DBZ-7138 Documentation for additional-conditions for MongoDb incremental snapshot

ani-sha:
https://issues.redhat.com/browse/DBZ-7138

### DIFF
--- a/documentation/modules/ROOT/partials/modules/all-connectors/proc-triggering-an-incremental-snapshot-nosql.adoc
+++ b/documentation/modules/ROOT/partials/modules/all-connectors/proc-triggering-an-incremental-snapshot-nosql.adoc
@@ -16,7 +16,7 @@ If the `data-collections` array is empty, {prodname} detects that no action is r
 [NOTE]
 ====
 If the name of a {data-collection} that you want to include in a snapshot contains a dot (`.`) in the name of the database, schema, or table, to add the {data-collection} to the `data-collections` array, you must escape each part of the name in double quotes. +
-+
+ +
 For example, to include a data collection that exists in the `*public*` database, and that has the name `*My.Collection*`, use the following format: `*"public"."My.Collection"*`.
 ====
 
@@ -88,9 +88,9 @@ If you do not specify a value, the connector runs an incremental snapshot.
 |6
 |`additional-conditions`
 | An optional array that specifies a set of additional conditions that the connector evaluates to determine the subset of records to include in a snapshot. +
-Each additional condition is an object with `data-collection` and `filter` properties.
-You can specify different filters for each data collection. +
+Each additional condition is an object with `data-collection` and `filter` properties. You can specify different filters for each data collection. +
 * The `data-collection` property is the fully-qualified name of the data collection for which the filter will be applied.
+For more information about the `additional-conditions` parameter, see xref:{context}-incremental-snapshots-additional-conditions[].
 |===
 
 The following example, shows the JSON for an incremental snapshot event that is captured by a connector.
--- a/documentation/modules/ROOT/partials/modules/all-connectors/proc-triggering-an-incremental-snapshot-nosql.adoc
+++ b/documentation/modules/ROOT/partials/modules/all-connectors/proc-triggering-an-incremental-snapshot-nosql.adoc
@@ -5,7 +5,7 @@ You submit a signal to the signaling {data-collection} by using the MongoDB `ins
 After {prodname} detects the change in the signaling {data-collection}, it reads the signal, and runs the requested snapshot operation.
 
 The query that you submit specifies the {data-collection}s to include in the snapshot, and, optionally, specifies the type of snapshot operation.
-Currently, the only valid options for snapshots operations are `incremental`and `blocking`.
+Currently, the only valid options for snapshots operations are `incremental` and `blocking`.
 
 To specify the {data-collection}s to include in the snapshot, provide a `data-collections` array that lists the {data-collection}s or an array of regular expressions used to match {data-collection}s, for example, +
 `{"data-collections": ["public.Collection1", "public.Collection2"]}` +
@@ -16,7 +16,7 @@ If the `data-collections` array is empty, {prodname} detects that no action is r
 [NOTE]
 ====
 If the name of a {data-collection} that you want to include in a snapshot contains a dot (`.`) in the name of the database, schema, or table, to add the {data-collection} to the `data-collections` array, you must escape each part of the name in double quotes. +
- +
++
 For example, to include a data collection that exists in the `*public*` database, and that has the name `*My.Collection*`, use the following format: `*"public"."My.Collection"*`.
 ====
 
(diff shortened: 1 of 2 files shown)


documentation/modules/ROOT/partials/modules/all-connectors/proc-triggering-an-incremental-snapshot-nosql.adoc
```diff
@@ -83,6 +84,13 @@ The array lists regular expressions which match {data-collection}s by their full
 |An optional `type` component of the `data` field of a signal that specifies the type of snapshot operation to run. +
 Currently supports the `incremental` and `blocking` types. +
 If you do not specify a value, the connector runs an incremental snapshot.
+
+|6
+|`additional-conditions`
+| An optional array that specifies a set of additional conditions that the connector evaluates to determine the subset of records to include in a snapshot. +
+Each additional condition is an object with `data-collection` and `filter` properties.
+You can specify different filters for each data collection. +
+* The `data-collection` property is the fully-qualified name of the data collection for which the filter will be applied.
```
roldanbob, line 93: It seems that something is missing here. The asterisk (`*`) is supposed to represent a bullet in an unordered list.  I think that this text is supposed to mirror the text in the description for `additional-conditions` that's in the file `proc-triggering-an-incremental-snapshot-kafka.adoc` (L36-44). Is that right?
That description includes a definition list with two entries, `data-collection` and `filter`. 

To modify the text in this PR to reflect that definition, it's necessary to add the `a` style operator to the specifier for the description column in the table's attributes list on L56 (or place the operator just in front of the cell separator on L90  (i.e., `a|`). 
With that change in place, we can then replace the current text in the table cell with a definition list, by adding the following:

```suggestion
For each additional condition object you can set the following parameters:

data-collection::
The fully-qualified name of the collection that the filter applies to. 
You can apply different filters to each collection. 

filter:: Specifies column values that must be present in a database record for the snapshot to include it, for example, `"color='blue'"`.
```

roldanbob (changes requested):
Hi @ani-sha 
A few suggestions and even more questions!

documentation/modules/ROOT/partials/modules/all-connectors/proc-triggering-an-incremental-snapshot-nosql.adoc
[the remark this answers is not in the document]
ani-sha: There is no such `snapshot. collection.statement.overrides` property. You are right the filter parameter accepts value similar to `snapshot.select.statement.overrides`. I will modify and add it here too.

ani-sha: > I think that this text is supposed to mirror the text in the description for additional-conditions that's in the file proc-triggering-an-incremental-snapshot-kafka.adoc (L36-44). Is that right?

I think it shouldn't match the kafka adoc. It should rather match the `additional-conditions` of `proc-triggering-an-incremental-snapshot.adoc` (L127-132).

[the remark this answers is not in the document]
ani-sha: Update : so the details on the `filter` are mentioned on the `proc-triggering-an-incremental-snapshot-kafka.adoc`. And this part of the documentation just points to it. So we don't need to re-mention it again here.

ani-sha:
@roldanbob I have addressed your questions and adjusted the docs. Please let me know if you have any other suggestions.

documentation/modules/ROOT/partials/modules/all-connectors/proc-triggering-an-incremental-snapshot-nosql.adoc
```diff
@@ -83,6 +84,13 @@ The array lists regular expressions which match {data-collection}s by their full
 |An optional `type` component of the `data` field of a signal that specifies the type of snapshot operation to run. +
 Currently supports the `incremental` and `blocking` types. +
 If you do not specify a value, the connector runs an incremental snapshot.
+
+|6
+|`additional-conditions`
+| An optional array that specifies a set of additional conditions that the connector evaluates to determine the subset of records to include in a snapshot. +
+Each additional condition is an object with `data-collection` and `filter` properties. You can specify different filters for each data collection. +
+* The `data-collection` property is the fully-qualified name of the data collection for which the filter will be applied.
```
roldanbob, line 92:
```suggestion
Each additional condition is an object with `data-collection` and `filter` properties. You can specify different filters for each data collection. +
* The `data-collection` property is the fully-qualified name of the data collection for which the filter will be applied.
```
I think that we still have a problem in this passage. L91 states that _"Each additional condition is an object with `data-collection` and `filter` properties"_.  I think that sentence is intended to introduce the unordered list that follows, but the list contains only the single entry, `data-collection`, and no entry for the `filter` property. 

If it is correct that we don't also have a `filter` property, the reference to it should be removed on L91. Or if you set the filter elsewhere, provide a reference to where and how that's done. 

If there's no `filter` property L91 should omit the mention it, and we should remove the asterisk/bullet on L92, letting it just stand as a plain sentence. Currently, we signal an unordered list, and provide an introduction that suggests the list will include two entries, but then we only provide the first item. 
Does that make sense? Let me know if you want to discuss further.

roldanbob (changes requested):
Thanks for your work, @ani-sha. I added one further comment concerning the content on L91-92.

documentation/modules/ROOT/partials/modules/all-connectors/proc-triggering-an-incremental-snapshot-nosql.adoc
roldanbob: `data-collection`:: The fully-qualified name of the data collection for which the filter will be applied. 
`filter`:: Specifies the column values that must be present in a data collection record for the snapshot to include it, for example, `"color='blue'"`.

documentation/modules/ROOT/partials/modules/all-connectors/proc-triggering-an-incremental-snapshot-nosql.adoc
```diff
@@ -88,9 +88,8 @@ If you do not specify a value, the connector runs an incremental snapshot.
 |6
 |`additional-conditions`
 | An optional array that specifies a set of additional conditions that the connector evaluates to determine the subset of records to include in a snapshot. +
-Each additional condition is an object with `data-collection` and `filter` properties. You can specify different filters for each data collection. +
-* The `data-collection` property is the fully-qualified name of the data collection for which the filter will be applied.
-For more information about the `additional-conditions` parameter, see xref:{context}-incremental-snapshots-additional-conditions[].
+`data-collection`:: The fully-qualified name of the data collection for which the filter will be applied.
```
roldanbob, line 91: Restore sentence that introduces the array keys.

```suggestion
Each element in the `additional-conditions` array is an object that includes the following keys:

`data-collection`:: The fully-qualified name of the data collection for which the filter will be applied.
```

documentation/modules/ROOT/partials/modules/all-connectors/proc-triggering-an-incremental-snapshot.adoc
```diff
@@ -127,9 +127,8 @@ If you do not specify a value, the connector runs an incremental snapshot.
 |6
 |`additional-conditions`
 | An optional array that specifies a set of additional conditions that the connector evaluates to determine the subset of records to include in a snapshot. +
-Each additional condition is an object with `data-collection` and `filter` properties. You can specify different filters for each data collection. +
-* The `data-collection` property is the fully-qualified name of the data collection for which the filter will be applied.
-For more information about the `additional-conditions` parameter, see xref:{context}-incremental-snapshots-additional-conditions[].
+`data-collection`:: The fully-qualified name of the data collection for which the filter will be applied.
```
roldanbob, line 130:
```suggestion
Each element in the `additional-conditions` array is an object that includes the following keys:

`data-collection`:: The fully-qualified name of the data collection for which the filter will be applied.
```

roldanbob (changes requested):
@ani-sha One final suggestion. Thanks!

roldanbob (approved):
LTGM! Thanks!

PR #5507, opened 2024-04-23
DBZ-7811 Reduce logging level for invalid timestamp/date values, fix edge case

twthorn:
That warn message could happen for every event processed in the worst case so change to debug.

Also there's an edge case with the year zero value that can still be valid.

https://issues.redhat.com/browse/DBZ-7811

### DIFF
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogValueConverters.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogValueConverters.java
@@ -76,6 +76,7 @@ import io.debezium.util.Strings;
 public abstract class BinlogValueConverters extends JdbcValueConverters {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(BinlogValueConverters.class);
+    private static final Logger INVALID_VALUE_LOGGER = LoggerFactory.getLogger(BinlogValueConverters.class.getName() + ".invalid_value");
 
     /**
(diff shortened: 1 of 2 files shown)


jpechane:
@twthorn Hi, you are right that the message has a potential to be triggered for any record. Yet I'd still argue that in most cases it really deservers `WARN` as it is at the end invalid data.
I'd propose an alternative solution which we already used elsewhere. Add another logger to the `BinlogValueConverters` that will be used for this messages. In this case the `WARN` can still be used and be triggered by default and when this is not reasonable for given dataset it can be disabled via logging config. WDYT?

Naros:
Hi @twthorn so this is what Jiri is recommending, at the top of the class we define two logers:
```java
private static final Logger LOGGER = LoggerFactory.getLogger(BinlogValueConverters.class);
private static final Logger INVALID_VALUE_LOGGER = LoggerFactory.getLogger(BinlogValueConverters.class.getName() + ".invalid_values");
```
Then in the two methods where we currently issue the WARN that you want to specifically control, change their references from using `LOGGER` to using `INVALID_VALUE_LOGGER` with the `warn` level, so from a user point of view the functionality doesn't change.

Then in your local configuration, you can set `io.debezium.connector.binlog.jdbc.BinlogValueConverters.invalid_values` to a logging level of ERROR to avoid the WARN messages while retaining all the other logging behavior for other methods.

Naros (approved):
LGTM from my PoV, deferring to @jpechane, in case he wants any other changes.

jpechane:
Applied, thanks guys!

PR #5508, opened 2024-04-24
DBZ-7807 added noncdb option for oracle connector tests in tmt

MartinMedek:


debezium-testing/tmt/tests/debezium/test.sh
```diff
@@ -35,6 +35,10 @@ then
     export ORACLE_CONNECTION="-Ddatabase.dbname=FREEPDB1 -Ddatabase.pdb.name=FREEPDB1"
     debezium-testing/tmt/tests/debezium/init-db.sh
   fi
+  DATABASE_USER="c##dbzuser"
+  if [[ "$ORACLE_VERSION" = *noncdb ]]; then
```
Naros, line 39: You may also have to pass `database.pdb.name` with an empty value?

MartinMedek: this is already passed in another parameter, will try. Thanks!

obabec (approved):

debezium-testing/tmt/tests/debezium/test.sh
obabec: @Naros we are not running noncdb on the TF anyway..

PR #5510, opened 2024-04-24
[docs] Add community connectors to monitoring page

Naros:


jpechane:
@Naros Applied, thanks

PR #5490, opened 2024-04-17
DBZ-7570/#133: started to add workaround using SO_LINGER with 0 timeout

acristu:
…ut if use.nongraceful.disconnect=true

PR #5517, opened 2024-04-25
Add workaround using SO_LINGER with 0 timeout

Naros:
Supersedes #5490 to rebase against `main`.


jpechane:
@acristu @Naros Applied, thanks

@acristu Could you please send a separate PR with docs update?

Naros:
@jpechane those numbers are getting a bit annoying :P

PR #5520, opened 2024-04-26
DBZ-7695 Backport to 2.6 add _id field to tombstone rewrite events, for delete events

ismailsimsek:


jpechane:
@ismailsimsek Hi, thanks for the PR (ant the other one). As this is a new feature not bugfix we prefer not to do the backports.

PR #5524, opened 2024-04-26
DBZ-4332 Add `row_id` to Oracle event's source info block

Naros:
https://issues.redhat.com/browse/DBZ-4332

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleOffsetContext.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleOffsetContext.java
@@ -305,6 +305,14 @@ public class OracleOffsetContext extends CommonOffsetContext<SourceInfo> {
         sourceInfo.setRedoSql(redoSql);
     }
 
+    public String getRowId() {
+        return sourceInfo.getRowId();
+    }
+
+    public void setRowId(String rowId) {
+        sourceInfo.setRowId(rowId);
+    }
+
     @Override
     public boolean isSnapshotRunning() {
         return sourceInfo.isSnapshot() && !snapshotCompleted;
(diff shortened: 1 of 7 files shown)


debezium-core/src/main/java/io/debezium/relational/Column.java
```diff
@@ -155,6 +155,13 @@ default boolean isRequired() {
      */
     String comment();
 
+    /**
+     * Get whether the column maps to an invisible column.
+     *
+     * @return {@code true} if the column is invisible, or {@code false} otherwise
+     */
+    boolean isInvisible();
```
jpechane, line 163: I wonder if using the name `invisible` is correct. IMHO `INVISIBLE`  columns in Oracle, MySQL etc. is something completely different and we are processing them. Is not `ROWID` pseudo (synthetic) column, not invisible?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
```diff
@@ -319,6 +320,17 @@ protected Callable<Void> createDataEventsForTableCallable(ChangeEventSourceConte
...
     }
 
+    @Override
+    protected void doPreDispatchCreateDataEventForTable(ResultSet rs, Table table, OracleOffsetContext offsetContext) throws SQLException {
+        try {
+            final String rowId = rs.getString("ROWID");
+            offsetContext.setRowId(rowId);
+        }
+        catch (SQLException e) {
+            e.printStackTrace();
```
jpechane, line 330: Please use `LOGGER`

jpechane (changes requested):
@Naros LGTM, thanks! I left two comments related naming and exception handling. When addressed please feel free to merge youself.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
Naros: Oops, that was not intended :)

jpechane (approved):
@Naros LGTM, thanks! Feel free to merge as soon as CI passes.

Naros:
Looking into the failures, they seem to be related to the changes.

Naros:
@jpechane, after investigating the failures, I've changed to focus only on streaming.

The changes needed to handle snapshots were starting to grow tendrils into the schema history topic and requiring some changes there that are better served as a design discussion before proceeding rather than to try and shoehorn a solution that may not be ideal without discussion.

jpechane:
@Naros `OracleSchemaMigrationIT` failure is unrelated?

Naros:
> @Naros `OracleSchemaMigrationIT` failure is unrelated?

I don't think so, i just reran that class locally against 21-xs and it passed all green, so I'd say it's likely timing:
image

jpechane:
@Naros Applied, thanks!

PR #5530, opened 2024-04-30
DBZ-7835 Use SEGMENT_COLUMN_ID instead of COLUMN_ID as the order of c…

butioy:
https://issues.redhat.com/browse/DBZ-7835


Naros (changes requested):
Hi @butioy, this is a great catch, but I'm afraid we won't be able to accept this PR as it is written.  The Oracle license is very clear on the following point:

_You do not cause or permit reverse engineering, disassembly or decompilation of the Programs_

The code you've provided in OracleDatabaseMetadata is a direct violation of this.  If you can engineer a PR that provides the functionality needed without violating the license, then we'd happily look to integrate it.

Naros:
Hi @butioy, could we start by simply issuing a second query against `ALL_TAB_COLUMNS` using the OWNER, TABLE_NAME, and COLUMN_NAME and get the SEGMENT_COLUMN_ID rather than trying to reimplement the entire metadata query?  We can then evaluate the performance implication of that without the license concern.

butioy:
I didn't notice the Oracle license, so I'll change it to query `ALL_TAB_COLUMNS ` again

PR #5532, opened 2024-05-01
DBZ-7816: Postgres: Potential data loss on connector restart

adasari:


Naros:
Hi @adasari there are some test failures, could you please take a look.

adasari:
@Naros I am still trying to compile the repo locally. will look into it. thanks.

adasari:
@Naros Test failures due to this [line](https://github.com/debezium/debezium/blob/cb18f662d038bad0662e0698579496935f1ba2d6/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorTask.java#L407) 
The IT tests utilize the Debezium engine, which lacks support for SourceConnectorTask#commitRecord(record, recordmetadata). As a result, the minimum timestamp map remains empty, leading to the absence of database flushes.

I have added min timestamp check to handle major DB upgrade scenarios. 

Major DB upgrades reset LSN and old kafka offsets LSN are not supposed to flushed to DB WAL. 
An alternative approach to address this scenario is to release the offsets that are more recent than the start time of the corrector or connector task. We may need to discuss Major DB upgrade scenario once all tests have passed successfully.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorTask.java
```diff
@@ -71,9 +75,13 @@ public class PostgresConnectorTask extends BaseSourceTask<PostgresPartition, Pos
     private volatile ErrorHandler errorHandler;
     private volatile PostgresSchema schema;
 
+    private Partition.Provider<PostgresPartition> partitionProvider = null;
+    private OffsetContext.Loader<PostgresOffsetContext> offsetContextLoader = null;
+
+    private final ReentrantLock stateLock = new ReentrantLock();
```
vjuranek, line 81: I think this will create another instance of the `stateLock` and it won't synchronize with blocks guarded by `BaseSourceTask#stateLock`

adasari: @vjuranek yes and is fine. I will add `PostgresConnectorTask#commitRecord(SourceRecord)` into avoid confusion for `DebeziumEngine` flow. `PostgresConnectorTask#commit` method is not using the state set by `BaseSourceTask#commitRecord`. So stateLock is used to make sure only one commit() processing is happening at any time. Let me know if you have any questions.

adasari: Done.

adasari:
> @Naros Test failures due to this [line](https://github.com/debezium/debezium/blob/cb18f662d038bad0662e0698579496935f1ba2d6/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorTask.java#L407) The IT tests utilize the Debezium engine, which lacks support for SourceConnectorTask#commitRecord(record, recordmetadata). As a result, the minimum timestamp map remains empty, leading to the absence of database flushes.
> 
> I have added min timestamp check to handle major DB upgrade scenarios.
> 
> Major DB upgrades reset LSN and old kafka offsets LSN are not supposed to flushed to DB WAL. An alternative approach to address this scenario is to release the offsets that are more recent than the start time of the corrector or connector task. We may need to discuss Major DB upgrade scenario once all tests have passed successfully.

@Naros @vjuranek Not sure if DB major upgrade case required to be handled. Alternatives to handle it without a code change, 
1. Send the new offset entry with new replication slot start LSN to offset topic. 
2. Use new kafka offset topic or empty the offset topic. 
3. Create new connector (with new name) with snapshot mode never. 

Option#1 ensure new replication slot offset always available in Kafka offset topic.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorTask.java
vjuranek: My main concern was that the commit can happen e.g. during stopping the connector. Looking more into it, it really seems harmless, we e.g. if the connection is already closed, the connector doesn't fail but skips the commit. So I'd suggest to rename this `stateLock` to e.g. `commitLock` to make clear that this has nothing to do with `stateLock` from parent class and it's intentionally used only for serializing calls of `commit()` method.

adasari: Done.

jpechane:
@adasari Applied, thanks a lot! Could you please create a separate PR with docs update?

adasari:
> @adasari Applied, thanks a lot! Could you please create a separate PR with docs update?

Sure. I'll review the documentation. Do you have a particular section in mind for updates? If not, I'll locate the relevant information.

jpechane:
@adasari I think it can go into section https://debezium.io/documentation/reference/2.6/connectors/postgresql.html#postgresql-when-things-go-wrong

PR #5533, opened 2024-05-01
DBZ-7840 Bump Infinispan to 15.0.2.Final

Naros:
https://issues.redhat.com/browse/DBZ-7840


Naros:
* The Debezium Server error is expected, there is a PR for Debezium Server to fix that dependency.  
* The test failure for Oracle XStream is unrelated and just a timing issue.
* The test failure with Oracle 23 is unrelated, its an issue with the image.

@vjuranek do you agree with these changes?

vjuranek (approved):

PR #5543, opened 2024-05-08
DBZ-7845 update COPYRIGHT.txt and Aliases.txt

gaurav7261:


Naros:
Hi @gaurav7261 can you also add to this PR and update to the `jdbc.adoc` for the new configuration option?

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -845,6 +845,16 @@ Note that if you set `_consumer.max.poll.records_` in the Connect worker propert
...
 
+|[[jdbc-property-use-reduction-buffer]]<<jdbc-property-use-reduction-buffer, `+use.reduction.buffer+`>>
+|`false`
+|Specifies whether to use the reduction buffer.
+
+[NOTE]
+====
+A reduction buffer consolidates the execution of SQL statements by primary key to reduce the SQL load on the target database. When set to false (the default), each incoming event is applied as a logical SQL change. When set to true, incoming events that refer to the same row will be reduced to a single logical change based on the most recent row state.
+For example, when using postgres, user can add `reWriteBatchedInserts` parameter in jdbc url along with enabling this property, then batch upsert queries will get fire to db
+====
```
roldanbob, line 856:
```suggestion
|Specifies whether to enable the {prodname} JDBC connector's reduction buffer.

Choose one of the following settings:

`false`:: (default) The connector writes each change event that it consumes from Kafka as a separate logical SQL change. 
`true`:: The connector uses the reduction buffer to reduce change events before it writes them to the sink database.
That is, if multiple events refer to the same primary key, the connector consolidates the SQL queries and writes only a single logical SQL change, based on the row state that is reported in the most recent offset record. +
Choose this option to reduce the SQL load on the target database.

To optimize query processing in a PostgreSQL sink database when the reduction buffer is enabled, you must also enable the database to execute the batched queries by adding the `reWriteBatchedInserts` parameter to the JDBC connection URL.
```

roldanbob (changes requested):
Rewritten to highlight the affect of choosing each option, and to move content out of the note admonition.
Let me know if anything looks out of place. Thanks!

gaurav7261:
Hi @roldanbob, changes done, please review again

PR #5544, opened 2024-05-08
DBZ-7847 Support in query filter mode with 1000+ tables

Naros:
https://issues.redhat.com/browse/DBZ-7847


Naros:
These failures are apicurio related.  Running the tests locally with apicurio 2.4.3.Final they pass, anything with 2.4.4.Final or later fail.

jpechane:
@Naros Should we override Apicurio version for Oracle test execution then?

Naros:
I spoke with the Apicurio team, and I think we have a way forward.  Let's discuss this offline.

Naros:
So only the Oracle fix needs to be backported to 2.6, the apicurio version in 2.6 is still based on 2.4.3.Final which isn't broken.

jpechane:
@Naros Applied, thanks!

PR #5545, opened 2024-05-08
[ci] Test bump Oracle TF HW

obabec:


Naros:
It's safe to close this @obabec, the issue is that apicurio is broken with 2.4.4 or later for some of our Oracle tests that perform quick succession changes to table schemas.

PR #5548, opened 2024-05-09
DBZ-7806 Skip search for WAL position when offset exist and is just after the initial snapshot

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7806

@jpechane The intent here is to skip the search on the WAL when the connector has an offset that is the result of the initial snapshot. Not sure if there is a better way to check this condition.


jpechane:
The change LGTM but can you please elaborate a bit more on how this fixes the original report?

jpechane:
@mfvitale Applied, thanks

jpechane:
@mfvitale Could you please prepare a backport for 2.6? The changes in MySQL are affected by the module split. Thanks

mfvitale:
Backported to 2.6.

PR #5549, opened 2024-05-09
Initial MySQL 8.4 compliance implementation

obabec:


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java
```diff
@@ -50,7 +53,13 @@ public boolean isGtidModeEnabled() {
     @Override
     public GtidSet knownGtidSet() {
         try {
-            return queryAndMap("SHOW MASTER STATUS", rs -> {
+            String showMasterStmt = "SHOW MASTER STATUS";
+
+            if (complianceMode.equals(MySqlConnectorConfig.ComplianceMode.MYSQL_84_PLUS)) {
+                showMasterStmt = "SHOW BINARY LOG STATUS;";
+            }
```
Naros, line 60: We don't need to necessarily do it now, but I wonder if we're going to see this type of behavior in the future with various MySQL versions if we should consider `ComplianceMode` returning an object that we can call somewhat like we do with the `StreamingAdapter` in the Oracle connector to abstract away these details of the if/else.  This way in the call sites it's simply `getComplianceMode().getShowMasterStatement()` or something like this.  This way it avoids the redundancy of this logic at each call site.  wdyt?

obabec: I think that is a great idea. TBH, this was just an initial implementation to figure out what changes I needed to make in our tests.

jpechane:
@obabec I'd see this PR should incorporate https://github.com/debezium/debezium/pull/5338 too to avoid authentication issue in future

obabec:
@jpechane that makes sense to me :)

AAljasser:
@obabec, do you have any updates on this? I'd be happy to assist if needed

jpechane:
I updated https://github.com/debezium/debezium/pull/5338 Hopefully we'll manage to run the tests with the other auth module and then this work can continue.
@AAljasser I believe that @obabec has other issues on his plate now (alse heading for vacations soon) so if you could continue on the PR it would be great.

jpechane:
Replaced with https://github.com/debezium/debezium/pull/5765

PR #5550, opened 2024-05-09
DBZ-7859: reduce enum array allocation

sullis:
reduce Java enum array allocation by caching the result of .values() method.

Reference:
https://www.gamlor.info/wordpress/2017/08/javas-enum-values-hidden-allocations/

https://issues.redhat.com/projects/DBZ/issues/DBZ-7859

### DIFF
--- a/debezium-core/src/main/java/io/debezium/data/Envelope.java
+++ b/debezium-core/src/main/java/io/debezium/data/Envelope.java
@@ -8,7 +8,10 @@ package io.debezium.data;
 import java.time.Instant;
 import java.util.Collections;
 import java.util.HashSet;
+import java.util.Map;
 import java.util.Set;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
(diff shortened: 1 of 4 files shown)


debezium-core/src/main/java/io/debezium/data/Envelope.java
```diff
@@ -57,19 +61,20 @@ public enum Operation {
          */
         MESSAGE("m");
 
+        private static final Map<String, Operation> CODE_LOOKUP = Stream.of(Operation.values()).collect(
```
mfvitale, line 64: I would add the reference as a comment here.

sullis: Done.

jpechane (approved):
@sullis LGTM, thanks! We'll merge the PR as soon as CI run is finished.

jpechane:
@sullis Applied, thanks!

PR #5561, opened 2024-05-13
DBZ-7870 default value of -1 for connection error retries interpreted as no limit

augi:
[The documentation](https://github.com/debezium/debezium/blob/main/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngineConfig.java#L184) of `errors.max.retries` states that `-1` should mean _unlimited_, but the implementation doesn't handle it.

We discovered this because we have these two messages in the log (after Postgres update), and the Debezium wasn't processing any data then.
```
Retriable exception thrown, connector will be restarted; errors.max.retries=-1
Can't start the connector, max retries to connect exceeded; stopping connector...
```

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
@@ -672,7 +672,7 @@ public final class EmbeddedEngine implements DebeziumEngine<SourceRecord>, Embed
                 startedSuccessfully = true;
             }
             catch (Exception ex) {
-                if (totalRetries >= maxRetries) {
+                if (maxRetries != EmbeddedEngineConfig.DEFAULT_ERROR_MAX_RETRIES && totalRetries >= maxRetries) {
                     LOGGER.error("Can't start the connector, max retries to connect exceeded; stopping connector...", ex);
                     return ex;
                 }
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -350,6 +350,7 @@ Meng Qiu
 Melissa Winstanley
 Michael Cizmar
 Michael Wang
+Michal Augustyn
 Mickael Maison
 Mickaël Isaert
 Mike Graham


Naros:
Hi @augi can you please create a Jira for this change and prefix the commits with the Jira issue key.
Thanks.

augi:
Did it, hopefully everything will be green now 🤞

jpechane:
@vjuranek Could you please take a look at this PR and check if this also an issue for async mebedded engine? Thanks

vjuranek (approved):

vjuranek:
@jpechane In async engine it should work correctly, or better to say we have a [test for it](https://github.com/debezium/debezium/blob/main/debezium-embedded/src/test/java/io/debezium/embedded/async/RetryingCallableTest.java#L125), this is passing.

jpechane:
@augi Applied, thanks!

debezium-embedded/src/main/java/io/debezium/embedded/EmbeddedEngine.java
```diff
@@ -672,7 +672,7 @@ private Throwable handleRetries(final RetriableException e, final List<Map<Strin
                 startedSuccessfully = true;
             }
             catch (Exception ex) {
-                if (totalRetries >= maxRetries) {
+                if (maxRetries != EmbeddedEngineConfig.DEFAULT_ERROR_MAX_RETRIES && totalRetries >= maxRetries) {
```
rk3rn3r, line 675: @jpechane @vjuranek Do you think it is a good idea to reference the const for the default value here? This condition should only yield true when the value is explicitly -1. Imagine that in a future PR the default value is going to be updated, then this line would be invalid and the code misbehaves. wdyt?

jpechane: @rk3rn3r Hi, yes, it is good idea. But the fact is that what is not correct is not the use of the constant but the name of the constant. It is both default and special value and the name should reflect it is the special value, not default value.

PR #5562, opened 2024-05-14
DBZ-7641 Publish snapshots to new infra

jpechane:


jpechane:
Tested and working - see https://s01.oss.sonatype.org/content/repositories/snapshots/io/debezium/

PR #5565, opened 2024-05-15
DBZ-7871 Change logging level for `Found previous partition offset`

rkudryashov:
https://issues.redhat.com/browse/DBZ-7871

### DIFF
--- a/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
+++ b/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
@@ -487,7 +487,7 @@ public abstract class BaseSourceTask<P extends Partition, O extends OffsetContex
 
             if (offset != null) {
                 found = true;
-                LOGGER.info("Found previous partition offset {}: {}", partition, offset.getOffset());
+                LOGGER.debug("Found previous partition offset {}: {}", partition, offset.getOffset());
             }
         }
 
--- a/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
+++ b/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
@@ -204,6 +204,13 @@ public abstract class BaseSourceTask<P extends Partition, O extends OffsetContex
 
     private final List<NotificationChannel> notificationChannels;
 
+    /**
+     * A flag to record whether the offsets stored in the offset store are loaded for the first time.
+     * This is typically used to reduce logging in case a connector like PostgreSQL reads offsets
+     * not only on connector startup but repeatedly during execution time too.
+     */
+    private boolean offsetLoadedInPast = false;
+
     protected BaseSourceTask() {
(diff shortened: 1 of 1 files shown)


mfvitale (approved):

jpechane:
I definitely don't want to remove it globally. The same method is used for initial idnetification of existing offsets by all connectors and it is IMHO important that the message is logged at `INFO` level. I'd be OK with intorducing a flag that will decide whether the logging is at INFO or DEBUG/TRACE level and intorudce a new method that would accept that flag. By default INFO will be used and his instance can override it.

mfvitale:
> I definitely don't want to remove it globally. The same method is used for initial idnetification of existing offsets by all connectors and it is IMHO important that the message is logged at `INFO` level. I'd be OK with intorducing a flag that will decide whether the logging is at INFO or DEBUG/TRACE level and intorudce a new method that would accept that flag. By default INFO will be used and his instance can override it.



> I definitely don't want to remove it globally. The same method is used for initial idnetification of existing offsets by all connectors and it is IMHO important that the message is logged at `INFO` level. I'd be OK with intorducing a flag that will decide whether the logging is at INFO or DEBUG/TRACE level and intorudce a new method that would accept that flag. By default INFO will be used and his instance can override it.



> I definitely don't want to remove it globally. The same method is used for initial idnetification of existing offsets by all connectors and it is IMHO important that the message is logged at `INFO` level. I'd be OK with intorducing a flag that will decide whether the logging is at INFO or DEBUG/TRACE level and intorudce a new method that would accept that flag. By default INFO will be used and his instance can override it.



> I definitely don't want to remove it globally. The same method is used for initial idnetification of existing offsets by all connectors and it is IMHO important that the message is logged at `INFO` level. I'd be OK with intorducing a flag that will decide whether the logging is at INFO or DEBUG/TRACE level and intorudce a new method that would accept that flag. By default INFO will be used and his instance can override it.

What do you think to remove the log `side effect` out of this method and then log it on the task? So any other use of that method will require no previous knowledge.

Naros:
While they do @mfvitale, I can see certain situations where knowing that the offset is per partition could be useful in the case of SQL Server where it stores offsets across partitions, so I wouldn't recommend removing it.  What makes it more complicated as removing the logging would break behavior for MongoDB which has a very different way of logging this phase for users.  

I think the path of least resistance here would be as Jiri mentioned, pass a flag from PG where we do this special call in the commit method to bypass logging like we do for Replica Identity and Schema Refreshes.

jpechane:
@mfvitale WDYT? https://github.com/debezium/debezium/pull/5565/commits/04372acd5ac091e12dc7fd77bbd344e462c91096

mfvitale:
> @mfvitale WDYT? [04372ac](https://github.com/debezium/debezium/commit/04372acd5ac091e12dc7fd77bbd344e462c91096)

Sounds good. Just leaved a nitpick comment.

jpechane:
@mfvitale Comment added

PR #5569, opened 2024-05-21
DBZ-7889: Multiple completed reading from a capture instance notifications

ramanenka:
There is a small chance the connector doesn't advance and re-reads the same LSN range. This happens under the following conditions:
* a new capture instance has been added in the current LSN range;
* while reading CDC changes one of existing capture instances dissapears. 

The dissapeared capture instance causes an exception which is catched and processed in `processErrorFromChangeTableQuery`. This leads to the current connector iteration to be correctly exited without advancing. On the next iteration the connector starts from the same LSN as the previous iteration and finds the same new capture instance. Although `Set` was used to track the list of tables to be removed `SqlServerChangeTable` doesn't implement `hashCode` so same table could be added multiple times to the same set.

The fix is to implement `hashCode` and `equals` methods in `ChangeTable` which is the parent class of `SqlServerChangeTable`.

Additionally a synchronisation block is needed where the tables are added to the hash map as it happens in a different thread from the one that removes the tables from the hash map.

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerStreamingChangeEventSource.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerStreamingChangeEventSource.java
@@ -369,8 +369,10 @@ public class SqlServerStreamingChangeEventSource implements StreamingChangeEvent
     private void collectChangeTablesWithKnownStopLsn(SqlServerPartition partition, SqlServerChangeTable[] tables) {
         for (SqlServerChangeTable table : tables) {
             if (table.getStopLsn().isAvailable()) {
-                LOGGER.info("The stop lsn of {} change table became known", table);
-                changeTablesWithKnownStopLsn.computeIfAbsent(partition, x -> new HashSet<>()).add(table);
+                synchronized (changeTablesWithKnownStopLsn) {
+                    LOGGER.info("The stop lsn of {} change table became known", table);
+                    changeTablesWithKnownStopLsn.computeIfAbsent(partition, x -> new HashSet<>()).add(table);
+                }
             }
         }
     }
--- a/debezium-core/src/main/java/io/debezium/relational/ChangeTable.java
+++ b/debezium-core/src/main/java/io/debezium/relational/ChangeTable.java
@@ -5,6 +5,8 @@
  */
 package io.debezium.relational;
 
+import java.util.Objects;
+
 /**
  * A logical representation of a change table containing changes for a given source table.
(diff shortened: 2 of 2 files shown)


ramanenka:
The test failures seem to be unrelated:
```
2024-05-21T14:58:43.6002631Z [INFO] Results:
2024-05-21T14:58:43.6003022Z [INFO] 
2024-05-21T14:58:43.6004777Z [ERROR] Failures: 
2024-05-21T14:58:43.6010016Z [ERROR]   IncrementalSnapshotWithRecompileIT.stopCurrentIncrementalSnapshotWithAllCollectionsAndTakeNewNewIncrementalSnapshotAfterRestart:176->AbstractIncrementalSnapshotTest.stopCurrentIncrementalSnapshotWithAllCollectionsAndTakeNewNewIncrementalSnapshotAfterRestart:659->AbstractSnapshotTest.startConnector:292->AbstractSnapshotTest.startConnector:283->AbstractConnectorTest.assertNoRecordsToConsume:991 expected:<[tru]e> but was:<[fals]e>
2024-05-21T14:58:43.6017080Z [ERROR]   IncrementalSnapshotWithRecompileIT.stopCurrentIncrementalSnapshotWithoutCollectionsAndTakeNewNewIncrementalSnapshotAfterRestart:182->AbstractIncrementalSnapshotTest.stopCurrentIncrementalSnapshotWithoutCollectionsAndTakeNewNewIncrementalSnapshotAfterRestart:604->AbstractSnapshotTest.startConnector:292->AbstractSnapshotTest.startConnector:283->AbstractConnectorTest.assertNoRecordsToConsume:991 expected:<[tru]e> but was:<[fals]e>
2024-05-21T14:58:43.6019574Z [ERROR] Errors: 
2024-05-21T14:58:43.6022255Z [ERROR]   SqlServerConnectorIT.shouldStopRetriableRestartsAtConfiguredMaximumDuringStreaming:2949->shouldStopRetriableRestartsAtConfiguredMaximum:3299 » ConditionTimeout Condition with alias 'Checking for maximum restart messages' didn't complete within 5 seconds because condition with io.debezium.connector.sqlserver.SqlServerConnectorIT was not fulfilled.
2024-05-21T14:58:43.6024354Z [INFO] 
2024-05-21T14:58:43.6024676Z [ERROR] Tests run: 246, Failures: 2, Errors: 1, Skipped: 2
```

jpechane:
@ramanenka How about adding the `hashCode` into `SqlServerChangeTable`?

ramanenka:
@jpechane That's definitely a better idea. I've updated the PR.

jpechane:
@ramanenka Applied, thanks!

PR #5570, opened 2024-05-21
DBZ-6722 Add exclude components for transaction metadata

twthorn:
[DBZ-6722](https://issues.redhat.com/browse/DBZ-6722)

Provide transaction metadata does several things:
1. sends begin/end messages to transaction topics
2. adds tx id in transaction block
3. adds order info in transaction block

Connectors may not always want all of these functionalities. Especially for connectors where the volume is high (transaction topic is massive) or the id's are very large (e.g., mysql, vitess connectors). We should provide an easy config for excluding/including components of tx metadata so unnecessary data is not produced (larger payloads in data messages, or the transaction topic).


twthorn:
I see failed tests but they appear to be flakey/unrelated.

jpechane:
@twthorn Hi, thanks for the PR. I think ihe idea is reasonable yet the more I think about it the more it seems to be it is really specific Vitess issue at least for now. For that reason I'd propose to

- Move `ExcludeTransactionComponents` to vitess connector till we find the need for it to make it global
- Get rid of changes in the core - if changes at schema needs to be done, they should be done at SMT
- Merge DBZ-6721 and DBZ-6722 into a single PR just with different commits to simplify CI

WDYT?

twthorn:
@jpechane Sounds good I will update https://github.com/debezium/debezium-connector-vitess/pull/194 to contain this. If we want to move to main it should be an easy refactor.

PR #5572, opened 2024-05-21
DBZ-7890 Updates text of PP index file and removes incubation note

roldanbob:
[DBZ-7890](https://issues.redhat.com/browse/DBZ-7890)

Aligns text of post processors `index.adoc` file with the downstream version, and removes incubation note.  

The product docs do not use this file, so this update has no affect on the downstream version.
 
Per @Naros, it's safe to apply this change to 2.6 and later.


jpechane:
@roldanbob Applied, thanks

PR #5577, opened 2024-05-23
DBZ-7898 Support parsing for interval default values

Naros:
https://issues.redhat.com/browse/DBZ-7898


jpechane:
@Naros Applied, thanks

PR #5586, opened 2024-05-30
DBZ-7886 Skip current data collection snapshot when SQLException occurs

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7886


jpechane:
@mfvitale Applied, thanks!

PR #5588, opened 2024-05-30
DBZ-7908 Add subpartition into allowed expressions when altering table

vjuranek:


jpechane:
Backported via direct cherry pick and push

PR #5589, opened 2024-05-30
DBZ-7902 Add configurable delay after snapshotting, before starting streaming

rajdangwal:
Addresses: https://issues.redhat.com/browse/DBZ-7902

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
@@ -19,6 +19,9 @@ import java.util.concurrent.locks.ReentrantLock;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
+import io.debezium.config.ConfigurationDefaults;
+import io.debezium.util.Clock;
+import io.debezium.util.Metronome;
 import org.apache.kafka.connect.source.SourceConnector;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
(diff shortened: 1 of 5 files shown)


debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
```diff
@@ -193,11 +196,37 @@ protected void executeChangeEventSources(CdcSourceTaskContext taskContext, Snaps
...
             previousLogContext.set(taskContext.configureLoggingContext("streaming", partition));
             streamEvents(context, partition, snapshotResult.getOffset());
         }
     }
 
+    /**
+     * Delays streaming execution as per the {@link CommonConnectorConfig#STREAMING_DELAY_MS} parameter.
+     */
+    protected void delayStreamingIfNeeded(ChangeEventSourceContext context) throws InterruptedException {
+        Duration streamingDelay = connectorConfig.getStreamingDelay();
```
jpechane, line 211: I believe it is necessary to check if streaming is enabled on `Snapshotter` and skip delay if it is not.

jpechane (commented):
@rajdangwal LGTM! I left one comment related to no streaming scenario. Also would it be possible to add a single test for it?

debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
```diff
@@ -193,11 +196,37 @@ protected void executeChangeEventSources(CdcSourceTaskContext taskContext, Snaps
         LOGGER.debug("Snapshot result {}", snapshotResult);
 
         if (running && snapshotResult.isCompletedOrSkipped()) {
+            if (snapshotResult.isCompleted()) {
```
mfvitale, line 199: Why do delay only in case of  `completed` snapshot? I know that the goal here is to let the connector to flush offsets and avoid a re-snapshot on restart but I think this should be more general delay.

debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
rajdangwal: Thanks for the review @jpechane 
Addressed the comments.

debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
rajdangwal: If the connector is configured to skip snapshot, then we don't want to add an unnecessary delay for streaming as there is no snapshot offset to be committed.
Also, for aborted/failed snapshots, the streaming won't start anyway, so no need to add a delay.

mfvitale: I just saw that you mentioned the behavior in the field description, so I am fine with that. Please add it also on the documentation.

debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
rajdangwal: I am unable to download the logs for failed tests.
Most likely, they should be unrelated, but wanted to confirm. Will retry downloading and analyzing later.

jpechane:
@rajdangwal Applied, thanks! Could you please provide a separate PR with doc update?

rajdangwal:
Thanks @jpechane 
I will raise a doc update PR.

PR #5590, opened 2024-05-30
DBZ-7894 Restart LOB mining from min cached LOB transaction

Naros:
https://issues.redhat.com/browse/DBZ-7894

- [ ] Still need to evaluate and apply logic for the ISPN off-heap buffer.


Naros:
Given research in the DDD-9 around `SAFE_RESUME_SCN` and how I've identified several corner cases around what we can and cannot do with restarting mid-transaction, this is not viable. Closing.

PR #5594, opened 2024-06-03
DBZ-7878 Add test for GIS types during snapshot

mfvitale:
related to: https://issues.redhat.com/browse/DBZ-7878


jpechane:
@mfvitale Applied, thanks

PR #5597, opened 2024-06-05
DBZ-7813: waitForAvailableRecords using Awaitility

nrkljo:
It appears that under some conditions the test runner thread does effectively never yield during the busy-wait loop in the original waitForAvailableRecords(). With constrained CPU resources this can cause the engine thread to not run and thus not generate records in time. Awaitility sleep-wait alleviates this.


jpechane:
@nrkljo Does it make sense to intorduce a new method? There is already `waitForAvailableRecords(...)` on the class. Maybe it would make sense to modify that one to use `Awaitility`?

jpechane:
@nrkljo Applied, thanks

PR #5601, opened 2024-06-06
DBZ-7926 Update format of `type` field values in CloudEvent records 

roldanbob:
[DBZ-7926](https://issues.redhat.com/browse/DBZ-7926)

To align with an earlier code change ([DBZ-7216](https://issues.redhat.com/browse/DBZ-7216)), this issue makes the following changes to the CloudEvents documentation:

- Updates the format that is used for values of the `type` field of a CloudEvents change record.
- Implements other related minor formatting changes.

Changes were tested in a local Antora build.


jpechane:
@roldanbob Applied, thanks

PR #5604, opened 2024-06-07
DBZ-7894 Implement opt-in log.mining.use.safe.resume.scn

Naros:
https://issues.redhat.com/browse/DBZ-7894

A different implementation based on using LogMiner `SAFE_RESUME_SCN`.  This new behavior is an opt-in feature behind the internal option `internal.log.mining.use.safe.resume.scn`.  We'll look at moving this to be default in Debezium 3.0.

This PR is intended to supersede #5590


Naros:
A new PR implementation follow-up to this is in #6341.  Based on research in the DDD-9, the `SAFE_RESUME_SCN` field is not reliable to always control where the connector should resume from, so I am closing this as obsolete.

PR #5607, opened 2024-06-10
DBZ-7937 Add exception stack trace when engine fails to commit offset

vjuranek:
https://issues.redhat.com/browse/DBZ-7937


jpechane:
@vjuranek Applied, thanks

PR #5611, opened 2024-06-11
DBZ-7917: Read-only incremental snapshots for PostgreSQL

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7917

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PgSnapshot.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PgSnapshot.java
@@ -66,10 +66,15 @@ public class PgSnapshot {
 
     @Override
     public boolean equals(Object o) {
-        if (this == o)
+
+        if (this == o) {
             return true;
-        if (o == null || getClass() != o.getClass())
+        }
+
+        if (o == null || getClass() != o.getClass()) {
             return false;
+        }
+
         PgSnapshot that = (PgSnapshot) o;
         return Objects.equals(xMin, that.xMin) && Objects.equals(xMax, that.xMax) && Objects.equals(xip, that.xip);
     }
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PgSnapshot.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PgSnapshot.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.postgresql;
+
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import io.debezium.DebeziumException;
+
+public class PgSnapshot {
+
+    private static final String SNAPSHOT_FORMAT = "(\\d+):(\\d+):((\\d+,*)+)*";
+    private static final Pattern SNAPSHOT_PATTERN = Pattern.compile(SNAPSHOT_FORMAT);
+    private static final String SEPARATOR = ",";
+
+    private final Long xMin;
+    private final Long xMax;
+    private final Set<Long> xip;
+
+    public PgSnapshot(Long xMin, Long xMax, Set<Long> xip) {
+        this.xMin = xMin;
+        this.xMax = xMax;
+        this.xip = xip;
+    }
+
+    public Long getXMin() {
+        return xMin;
+    }
+
+    public Long getXMax() {
+        return xMax;
+    }
+
+    public Set<Long> getXip() {
+        return xip;
+    }
+
+    public static PgSnapshot from(String snapshotString) {
+
+        Matcher matcher = SNAPSHOT_PATTERN.matcher(snapshotString);
+
+        if (matcher.matches()) {
+
+            Long xmin = Long.parseLong(matcher.group(1));
+            Long xmax = Long.parseLong(matcher.group(2));
+
+            Set<Long> xip = Set.of();
+            if (matcher.group(3) != null) {
+                xip = Arrays.stream(matcher.group(3).split(SEPARATOR))
+                        .map(Long::parseLong)
+                        .collect(Collectors.toSet());
+            }
+
+            return new PgSnapshot(xmin, xmax, xip);
+        }
+
+        throw new DebeziumException(String.format("Unable to parse PgCurrentSnapshot result %s.", snapshotString));
+    }
+
+    @Override
+    public boolean equals(Object o) {
+        if (this == o)
+            return true;
+        if (o == null || getClass() != o.getClass())
+            return false;
+        PgSnapshot that = (PgSnapshot) o;
+        return Objects.equals(xMin, that.xMin) && Objects.equals(xMax, that.xMax) && Objects.equals(xip, that.xip);
+    }
+
+    @Override
+    public int hashCode() {
+        return Objects.hash(xMin, xMax, xip);
+    }
+
+    @Override
+    public String toString() {
+        return "PgSnapshot{" +
+                "xMin=" + xMin +
+                ", xMax=" + xMax +
+                ", xip=" + xip +
+                '}';
+    }
+}
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PgSnapshot.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PgSnapshot.java
@@ -14,6 +14,11 @@ import java.util.stream.Collectors;
 
 import io.debezium.DebeziumException;
 
+/**
+ * This class contains the information returned by the pg_current_snapshot function.
+ *
+ * @author Mario Fiore Vitale
+ */
 public class PgSnapshot {
 
     private static final String SNAPSHOT_FORMAT = "(\\d+):(\\d+):((\\d+,*)+)*";
@@ -42,7 +47,13 @@ public class PgSnapshot {
         return xip;
     }
 
-    public static PgSnapshot from(String snapshotString) {
+    /**
+     * Returns a PgSnapshot instance representing the specified snapshot string
+     *
+     * @param snapshotString is the string returned by the pg_current_snapshot function
+     * @return a PgSnapshot representing the {@code snapshotString}
+     */
+    public static PgSnapshot valueOf(String snapshotString) {
 
         Matcher matcher = SNAPSHOT_PATTERN.matcher(snapshotString);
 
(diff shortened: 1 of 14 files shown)


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PgSnapshot.java
```diff
@@ -0,0 +1,95 @@
...
+import java.util.Arrays;
+import java.util.Objects;
+import java.util.Set;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
+import java.util.stream.Collectors;
+
+import io.debezium.DebeziumException;
+
+public class PgSnapshot {
```
jpechane, line 17: Please add JavaDoc

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PgSnapshot.java
```diff
@@ -0,0 +1,95 @@
...
+
+    public Long getXMax() {
+        return xMax;
+    }
+
+    public Set<Long> getXip() {
+        return xip;
+    }
+
+    public static PgSnapshot from(String snapshotString) {
```
jpechane, line 45: Can we use `valueOf` name as it is common pattern in Java

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
```diff
@@ -959,6 +959,16 @@ public static AutoCreateMode parse(String value, String defaultValue) {
             .withDefault(Boolean.TRUE)
             .withValidation(Field::isBoolean, PostgresConnectorConfig::validateFlushLsnSource);
 
+    public static final Field READ_ONLY_CONNECTION = Field.create("read.only")
```
jpechane, line 962: Given the option name we should as a follw-up task verify that it really works on read-only database and if not then make the necessary changes.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresStreamingChangeEventSource.java
```diff
@@ -251,27 +251,30 @@ private void processReplicationMessages(PostgresPartition partition, PostgresOff
...
-
+        LOGGER.trace("Processing replication message {}", message);
         if (message.isLastEventForLsn()) {
             lastCompletelyProcessedLsn = lsn;
         }
 
         // Tx BEGIN/END event
         if (message.isTransactionalMessage()) {
+
+            offsetContext.updateWalPosition(lsn, lastCompletelyProcessedLsn, message.getCommitTime(), toLong(message.getTransactionId()),
```
jpechane, line 262: I'd be really really careful here. Please make sure that the LSN of the commit is different form the LSN of the following message otherwise we might end up with message loss on crash.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/ReadOnlyIncrementalSnapshotIT.java
```diff
@@ -0,0 +1,223 @@
...
+import io.debezium.junit.logging.LogInterceptor;
+import io.debezium.pipeline.signal.actions.AbstractSnapshotSignal;
+import io.debezium.pipeline.signal.actions.snapshotting.ExecuteSnapshot;
+import io.debezium.pipeline.signal.actions.snapshotting.StopSnapshot;
+import io.debezium.pipeline.signal.channels.KafkaSignalChannel;
+import io.debezium.relational.RelationalDatabaseConnectorConfig;
+import io.debezium.util.Strings;
+
+@SkipWhenDatabaseVersion(check = LESS_THAN, major = 13, minor = 0, reason = "Function pg_current_snapshot() not supported until PostgreSQL 13")
+public class ReadOnlyIncrementalSnapshotIT extends IncrementalSnapshotIT {
```
jpechane, line 35: Should we introcude a common `ReadOnlyIncrementalSnapshotIT` class that would be parent for both MySQL and Postgres tests? If net is there an option for utility class used by the read-only test classes?

debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
```diff
@@ -426,6 +426,17 @@ public void dispatchHeartbeatEvent(P partition, OffsetContext offset) throws Int
                 this::enqueueHeartbeat);
     }
 
+    public void dispatchHeartbeatEventAlsoToIncrementalSnapshot(P partition, OffsetContext offset) throws InterruptedException {
```
jpechane, line 429: Do we need and extra method here? Should not it by done by default?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresReadOnlyIncrementalSnapshotChangeEventSource.java
```diff
@@ -0,0 +1,219 @@
...
+import io.debezium.pipeline.source.spi.SnapshotProgressListener;
+import io.debezium.pipeline.spi.OffsetContext;
+import io.debezium.relational.RelationalDatabaseConnectorConfig;
+import io.debezium.relational.Table;
+import io.debezium.relational.TableId;
+import io.debezium.schema.DatabaseSchema;
+import io.debezium.spi.schema.DataCollectionId;
+import io.debezium.util.Clock;
+
+public class PostgresReadOnlyIncrementalSnapshotChangeEventSource<P extends PostgresPartition>
```
jpechane, line 30: PLease add JavaDoc.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresReadOnlyIncrementalSnapshotContext.java
```diff
@@ -0,0 +1,109 @@
...
+    public void setHighWatermark(PgSnapshot highWatermark) {
+        LOGGER.trace("Setting high watermark to {}", highWatermark);
+        this.highWatermark = highWatermark;
+    }
+
+    public void updateWindowState(OffsetContext offsetContext) {
+
+        Long eventTxId = offsetContext.getSourceInfo().getInt64(SourceInfo.TXID_KEY);
+        LOGGER.trace("Received event with TxId {}", eventTxId);
+        LOGGER.trace("Updating window. Window oped: {}, low watermark {}, high watermark {}", windowOpened, lowWatermark, highWatermark);
```
jpechane, line 65:
```suggestion
        LOGGER.trace("Updating window. Window opened: {}, low watermark {}, high watermark {}", windowOpened, lowWatermark, highWatermark);
```

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresReadOnlyIncrementalSnapshotChangeEventSource.java
```diff
@@ -0,0 +1,219 @@
...
+                    LOGGER.trace("Created new transaction ID {}", rs.getString(1));
+                }
+            });
+        }
+        catch (SQLException e) {
+            throw new DebeziumException(e);
+        }
+    }
+
+    private <T> JdbcConnection.ResultSetMapper<T> singleResultMapper(JdbcConnection.ResultSetExtractor<T> extractor, String error) throws SQLException {
```
jpechane, line 208: Why can't you use `io.debezium.jdbc.JdbcConnection.singleResultMapper(ResultSetExtractor<T>, String)`?

jpechane (commented):
@mfvitale Overall, LGTM, thanks! I left a couple of minor comments. PLease take a look

debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
mfvitale: Well, that was my first try. The "problem" seems to be with MySQL since the `heartbeat` is dispatched after every received message from the binlog, this cause tests failures because it will be more fast then expected. The MySQL read only implementation was designed to process the `server heartbeat` only.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/ReadOnlyIncrementalSnapshotIT.java
mfvitale: > If net is there an option for utility class used by the read-only test classes?

I was on that path with the idea to have a simple method to send signal and the let the implementation, based on something decide how to send it, but then there was some others changes. 

I ended up with the more pragmatic solution for now and maybe we can think to it on the next `read only` implementation. WDYT?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
mfvitale: >  Postgres connector runs two queries which are not supported in recovery mode (i.e. on read replicas).

Is [DBZ-7181](https://issues.redhat.com/browse/DBZ-7181)  absolves to the doubt?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresStreamingChangeEventSource.java
mfvitale:
```sql
SELECT * FROM pg_create_logical_replication_slot('regression_slot', 'test_decoding');

begin 
	
INSERT INTO public.test
(a)
VALUES('test');
commit

	
INSERT INTO public.test
(a)
VALUES('test2');

SELECT * FROM pg_logical_slot_peek_changes('regression_slot', NULL, NULL);
```

|lsn|xid|data|
|---|---|----|
|0/176A578|558|BEGIN 558|
|0/176A578|558|table public.test: INSERT: a[character varying]:'test'|
|0/176A5E8|558|COMMIT 558|
|0/176A620|559|BEGIN 559|
|0/176A620|559|table public.test: INSERT: a[character varying]:'test2'|
|0/176A690|559|COMMIT 559|


Is this answering your question?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
jpechane: Partially. We need to add a guard condition on slot/publication creation. It should not be attempted to be created on `read.only` connection, but it should be a separate Jira issue.

debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
jpechane: OK, let's keep it as is but maybe a comment is warrented when you should use each method.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/ReadOnlyIncrementalSnapshotIT.java
jpechane: I agree

jpechane:
@mfvitale Applied, thanks! Please add the requested comment as a part of a docs PR. Also please log follow-up Jiras for the details we discussed.

debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
mfvitale: https://github.com/debezium/debezium/pull/5624

PR #5612, opened 2024-06-11
DBZ-7931 Modify out_of_line_constraint clause

ani-sha:
https://issues.redhat.com/browse/DBZ-7931

@Naros I think similar issues had come up regarding ordering of the constraints. In this particular issue, the DDL should be using the `out_of_line_constraint`, but it doesn't. Even if we did a reordering in the past for the `relational_property` so the preference would be given to `out_of_line_constraint` before using the `column_definition` clause, it didn't work as the definition of `out_of_line_constraint` goes by 
```
out_of_line_constraint
    : (CONSTRAINT constraint_name)?
```
But if DDL contains has `CONSTRAINTS`, it would go to the `column_definition` clause and end up near the `inline_clause` throwing an error. 
Adding the change in the PR works for corner cases. But it also makes me wonder why the DDL failed earlier near the `inline_clause` since it also doesn't have the `CONSTRAINTS` keyword. 🤔 

If you have any suggestions, we can discuss them.

### DIFF
--- a/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/oracle/generated/PlSqlParser.g4
+++ b/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/oracle/generated/PlSqlParser.g4
@@ -1511,7 +1511,7 @@ out_of_line_ref_constraint
     ;
 
 out_of_line_constraint
-    : ( (CONSTRAINT constraint_name)?
+    : ( ((CONSTRAINT | CONSTRAINTS) constraint_name)?
           ( UNIQUE '(' column_name (',' column_name)* ')'
           | PRIMARY KEY '(' column_name (',' column_name)* ')'
           | foreign_key_clause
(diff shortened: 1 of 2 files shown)


Naros:
Hi @ani-sha, I'll take a look and see, but I wouldn't be surprised if your approach isn't the path of least resistance.  

Just an FYI, an upstream contributor to Antlr all but rewrote many of the Oracle rules some months ago, but I've been hesitant to pull that all down for fear of regressions, so I've been letting the upstream changes percolate for a bit; and who knows the upstream rules may also suffer from this bug as well.  For Debezium 3, I had planned to pull down the new grammar and give it a test and see how it fairs, but didn't want to disrupt Debezium 2.x with such a change.

I think what's hard for us in particular is that what the Oracle documentation says about SQL rules does not always align with what LogMiner reconstructs, and so we end up in some situations where we have to make concessions to the grammar, this may be one of those too; but I'll look.

Naros (approved):

Naros:
Applied to main (=2.7) and branch 2.6

PR #5613, opened 2024-06-11
DBZ-7944 Exclude signal table from data snapshot

Naros:
https://issues.redhat.com/browse/DBZ-7944


Naros:
Errors are related to change - looking into a fix.

mfvitale:
@Naros Is this feasible? How we can produce the event without knowing the schema?

Naros:
> @Naros Is this feasible? How we can produce the event without knowing the schema?

I'm only excluding it from the data part of the snapshot, it will still have its schema captured.

jpechane (changes requested):
@Naros I am fine with this PR for removing signalling tbale from snapshots. But in relation to supplemental logging check I disagree. We are currently testing with ALL supplemental check enabled. Either the test needs to be updated or the initial check needs to stay in place.

jpechane:
@Naros Applied, thanks

PR #5614, opened 2024-06-11
DBZ-7801 Add SQL Server `heartbeat.action.query` support

Naros:
https://issues.redhat.com/browse/DBZ-7801


jpechane:
@Naros Applied, thanks!

PR #5615, opened 2024-06-11
DBZ-7640 Guard row calculation with logger `BinlogRowEstimateLogger`

Naros:
https://issues.redhat.com/browse/DBZ-7640


jpechane:
@Naros Applied, thanks

PR #5616, opened 2024-06-12
DBZ-7260 Remove no longer valid note about Mongo `tasks.max` usage

TomaszGaweda:
DBZ-7260 Removed multi-task capability, so I'm removing notice about this function from docs.


jpechane:
Thanks! Merged via https://github.com/debezium/debezium/pull/5621

PR #5617, opened 2024-06-12
DBZ-7942 Address parsing signal tables with special characters

Naros:
https://issues.redhat.com/browse/DBZ-7942


jpechane:
@Naros The Oracle's all failed on the same tests so it is better to investigate before merge.

jpechane:
@Naros Applied, thanks!

PR #5618, opened 2024-06-12
DBZ-7128 Move to PostgreSQL 12 as baseline

Naros:
https://issues.redhat.com/browse/DBZ-7128


jpechane:
@Naros Applied, thanks

PR #5619, opened 2024-06-12
DBZ-7079 Support 0 length byte buffer / string column truncation

Naros:
https://issues.redhat.com/browse/DBZ-7079


jpechane:
@Naros Applied, thanks

PR #5628, opened 2024-06-13
DBZ-7783 Remove MariaDB dependency on MySQL driver

Naros:
https://issues.redhat.com/browse/DBZ-7783

- [x] Introduce new `BinlogCharsetRegistry` contract
- [x] Introduce separate MariaDB implementation using `charset_mappings.json`
- [x] Remove MySQL driver dependency


jpechane:
@Naros LGTM, thanks! One question, does it make sense to change it to a service provided by the registry?

Naros:
> One question, does it make sense to change it to a service provided by the registry?

I don't see why we shouldn't.  I've added a new commit for this, let me know if that's sufficient.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorTask.java
```diff
@@ -78,6 +80,7 @@ protected Iterable<Field> getAllConfigurationFields() {
 
     @Override
     protected ChangeEventSourceCoordinator<MariaDbPartition, MariaDbOffsetContext> start(Configuration configuration) {
+        final MariaDbCharsetRegistry charsetRegistry = new MariaDbCharsetRegistry();
```
jpechane, line 83: Is this still needed?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogValueConverters.java
```diff
@@ -105,15 +107,18 @@ public abstract class BinlogValueConverters extends JdbcValueConverters {
...
+     * @param charsetRegistry the character set registry
      */
     public BinlogValueConverters(DecimalMode decimalMode,
                                  TemporalPrecisionMode temporalPrecisionMode,
                                  BigIntUnsignedMode bigIntUnsignedMode,
                                  BinaryHandlingMode binaryHandlingMode,
                                  TemporalAdjuster adjuster,
-                                 EventConvertingFailureHandlingMode eventConvertingFailureHandlingMode) {
+                                 EventConvertingFailureHandlingMode eventConvertingFailureHandlingMode,
+                                 BinlogCharsetRegistry charsetRegistry) {
```
jpechane, line 118: Here I'd prefer passing of the registry itself so in future we can just read from the reagistry instead of changing the signature.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorTask.java
Naros: Good catch, I must have missed that in the refactoring :+1:

jpechane:
@Naros Applied, thanks! Failure seems to be just a CI issue.

PR #5629, opened 2024-06-13
DBZ-7948 Run JDBC sink integration tests on connector changes

Naros:
https://issues.redhat.com/browse/DBZ-7948


jpechane:
@Naros Applied, thanks

PR #5631, opened 2024-06-18
DBZ-6618 Document usage of `${source.` prefix for table name formatting

Naros:
https://issues.redhat.com/browse/DBZ-6618

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -823,15 +823,15 @@ The connector adds missing columns to the table by comparing the incoming event'
 |[[jdbc-property-table-name-format]]<<jdbc-property-table-name-format, `+table.name.format+`>>
 |`+${topic}+`
 |Specifies a string that determines how the destination table name is formatted, based on the topic name of the event. +
-When the property is set to its default value, `+${topic}+`, when the connector reads an event from Kafka, it writes the event record to a destination table that is based on the name of the source topic . +
+When the property is set to its default value, `+${topic}+`, after the connector reads an event from Kafka, it writes the event record to a destination table with a name that matches the name of the source topic. +
  +
 You can also use this property to dynamically influence the names of target tables, based on values in incoming events.
 By applying this feature of the property, you can implement dynamic naming without having to write a custom single message transformation. +
  +
 To enable dynamic naming of output tables, set the value of this property to a pattern such as `+${source._field_}+`.
-This type of pattern instructs the connector to extract values from the `source` block of the {prodname} change event for use in constructing the table name.
+When you specify this type of pattern, the connector extracts values from the `source` block of the {prodname} change event, and then uses those values to construct the table name.
 For example, you might set the value of the property to the pattern `+${source.schema}_${source.table}+`.
-Based on this pattern, if the connector reads an event in which the source block contains the `schema` name value, `user`, and the `table` name value, `tab`, the connector writes the event record to a table with the name `user_tab`.
+Based on this pattern, if the connector reads an event in which the `schema` field in the source block contains the value, `user`, and the `table` field contains the value, `tab`, the connector writes the event record to a table with the name `user_tab`.
 
 |[[jdbc-property-dialect-postgres-postgis-schema]]<<jdbc-property-dialect-postgres-postgis-schema, `+dialect.postgres.postgis.schema+`>>
 |`public`
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -823,7 +823,11 @@ The connector adds missing columns to the table by comparing the incoming event'
 |[[jdbc-property-table-name-format]]<<jdbc-property-table-name-format, `+table.name.format+`>>
 |`+${topic}+`
 |Specifies a string that determines how the destination table name is formatted, based on the topic name of the event.
-The placeholder, `+${topic}+`, is replaced by the topic name.
+The placeholder, `+${topic}+`, is replaced by the topic name. +
+ +
+This configuration can also be specified using the `+${source._field_}+` pattern to use values from the Debezium change event's `source` information block for more dynamic formats.
+For example, given a schema of `user` and the table name of `tab`, a value of `+${source.schema}_${source.table}+` would use a table name mapping of `user_tab`.
+This allows each incoming event to dynamically influence the target table name without necessarily writing a custom single message transformation.
 
 |[[jdbc-property-dialect-postgres-postgis-schema]]<<jdbc-property-dialect-postgres-postgis-schema, `+dialect.postgres.postgis.schema+`>>
 |`public`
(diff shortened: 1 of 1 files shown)


documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -823,7 +823,11 @@ The connector adds missing columns to the table by comparing the incoming event'
 |[[jdbc-property-table-name-format]]<<jdbc-property-table-name-format, `+table.name.format+`>>
 |`+${topic}+`
 |Specifies a string that determines how the destination table name is formatted, based on the topic name of the event.
-The placeholder, `+${topic}+`, is replaced by the topic name.
+The placeholder, `+${topic}+`, is replaced by the topic name. +
+ +
+This configuration can also be specified using the `+${source._field_}+` pattern to use values from the Debezium change event's `source` information block for more dynamic formats.
```
roldanbob, line 828:
```suggestion
You can also use this property to dynamically influence the names of target tables, based on values in incoming events.
By applying this feature of the property, you can implement dynamic naming without having to write a custom single message transformation. +
 +
To enable dynamic naming of output tables, set the value of this property to a pattern such as `+${source._field_}+`.
This type of pattern instructs the connector to extract values from the `source` block of the {prodname} change event for use in constructing the table name.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -823,7 +823,11 @@ The connector adds missing columns to the table by comparing the incoming event'
...
+For example, given a schema of `user` and the table name of `tab`, a value of `+${source.schema}_${source.table}+` would use a table name mapping of `user_tab`.
```
roldanbob, line 829:
```suggestion
For example, you might set the value of the property to the pattern `+${source.schema}_${source.table}+`.
Based on this pattern, if the connector reads an event in which the source block contains the `schema` name value, `user`, and the `table` name value, `tab`, the connector writes the event record to a table with the name `user_tab`.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -823,7 +823,11 @@ The connector adds missing columns to the table by comparing the incoming event'
...
+This allows each incoming event to dynamically influence the target table name without necessarily writing a custom single message transformation.
```
roldanbob, line 830:
```suggestion

```
Per my earlier comment, if we cover the idea in this sentence as in the earlier paragraph (new L829), we can drop it here.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
roldanbob: I think that it makes sense to present the use case for enabling dynamic table naming via the property vs. using an SMT near the beginning of the description rather than mentioning it later. WDYT?

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -823,7 +823,11 @@ The connector adds missing columns to the table by comparing the incoming event'
 |[[jdbc-property-table-name-format]]<<jdbc-property-table-name-format, `+table.name.format+`>>
 |`+${topic}+`
 |Specifies a string that determines how the destination table name is formatted, based on the topic name of the event.
```
roldanbob, line 825:
```suggestion
|Specifies a string that determines how the destination table name is formatted, based on the topic name of the event. +
```

roldanbob (changes requested):
Some thoughts on re-presenting the information. I think that I got the formatting correct, but I think it's worth another look after you implement some of the changes.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -822,8 +822,16 @@ The connector adds missing columns to the table by comparing the incoming event'
 
 |[[jdbc-property-table-name-format]]<<jdbc-property-table-name-format, `+table.name.format+`>>
 |`+${topic}+`
-|Specifies a string that determines how the destination table name is formatted, based on the topic name of the event.
-The placeholder, `+${topic}+`, is replaced by the topic name.
+|Specifies a string that determines how the destination table name is formatted, based on the topic name of the event. +
+When the property is set to its default value, `+${topic}+`, when the connector reads an event from Kafka, it writes the event record to a destination table that is based on the name of the source topic . +
```
roldanbob, line 826:
```suggestion
When the property is set to its default value, `+${topic}+`, after the connector reads an event from Kafka, it writes the event record to a destination table with a name that matches the name of the source topic. +
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -822,8 +822,16 @@ The connector adds missing columns to the table by comparing the incoming event'
...
+ +
+You can also use this property to dynamically influence the names of target tables, based on values in incoming events.
+By applying this feature of the property, you can implement dynamic naming without having to write a custom single message transformation. +
+ +
+To enable dynamic naming of output tables, set the value of this property to a pattern such as `+${source._field_}+`.
+This type of pattern instructs the connector to extract values from the `source` block of the {prodname} change event for use in constructing the table name.
```
roldanbob, line 832:
```suggestion
When you specify this type of pattern, the connector extracts values from the `source` block of the {prodname} change event, and then uses those values to construct the table name.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -822,8 +822,16 @@ The connector adds missing columns to the table by comparing the incoming event'
...
+For example, you might set the value of the property to the pattern `+${source.schema}_${source.table}+`.
+Based on this pattern, if the connector reads an event in which the source block contains the `schema` name value, `user`, and the `table` name value, `tab`, the connector writes the event record to a table with the name `user_tab`.
```
roldanbob, line 834:
```suggestion
Based on this pattern, if the connector reads an event in which the `schema` field in the source block contains the value, `user`, and the `table` field contains the value, `tab`, the connector writes the event record to a table with the name `user_tab`.
```

roldanbob (changes requested):
Formatting was fine except for the extra space, but I realized that I'd introduced some awkward wording. Tweaked it in this version.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -822,8 +822,16 @@ The connector adds missing columns to the table by comparing the incoming event'
...
 |`+${topic}+`
-|Specifies a string that determines how the destination table name is formatted, based on the topic name of the event.
-The placeholder, `+${topic}+`, is replaced by the topic name.
+|Specifies a string that determines how the destination table name is formatted, based on the topic name of the event. +
+When the property is set to its default value, `+${topic}+`, after the connector reads an event from Kafka, it writes the event record to a destination table with a name that matches the name of the source topic. +
+ +
+You can also use this property to dynamically influence the names of target tables, based on values in incoming events.
+By applying this feature of the property, you can implement dynamic naming without having to write a custom single message transformation. +
+ +
+To enable dynamic naming of output tables, set the value of this property to a pattern such as `+${source._field_}+`.
```
roldanbob, line 831:
```suggestion
To configure the property to dynamically generate the names of destination tables, set its value to a pattern such as `+${source._field_}+`.
```

roldanbob (changes requested):
Revisited with some additional thoughts.

PR #5637, opened 2024-06-23
DBZ-7758 oracle connector cache refactor

jeremy-l-ford:
Refactor caching into common base class
Provide new LogMinerCache interface for swapping out alternative caches


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/LogMinerCache.java
```diff
@@ -0,0 +1,65 @@
...
+package io.debezium.connector.oracle.logminer.processor;
+
+import java.util.Optional;
+import java.util.function.BiConsumer;
+import java.util.function.BiFunction;
+import java.util.function.Consumer;
+import java.util.function.Predicate;
+import java.util.stream.Stream;
+
+public interface LogMinerCache<K, V> {
```
Naros, line 15: Lets document this interface and methods.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/echache/EhcacheLogMinerEventProcessor.java
```diff
@@ -0,0 +1,473 @@
...
+import io.debezium.connector.oracle.logminer.parser.LogMinerDmlEntryImpl;
+import io.debezium.connector.oracle.logminer.processor.AbstractTransactionCachingLogMinerEventProcessor;
+import io.debezium.connector.oracle.logminer.processor.LogMinerCache;
+import io.debezium.pipeline.EventDispatcher;
+import io.debezium.pipeline.source.spi.ChangeEventSource;
+import io.debezium.relational.TableId;
+
+public class EhcacheLogMinerEventProcessor extends AbstractTransactionCachingLogMinerEventProcessor<EhcacheTransaction> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(EhcacheLogMinerEventProcessor.class);
```
Naros, line 70: This doesn't seem to be used any longer.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/echache/EhcacheLogMinerEventProcessor.java
```diff
@@ -0,0 +1,473 @@
...
+    private final StatisticsService transactionCacheStats;
```
Naros, line 71: Is there a reason this needs to be an instance variable instance of a local variable in the constructor?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/echache/EhcacheTransaction.java
```diff
@@ -0,0 +1,72 @@
...
+import com.fasterxml.jackson.annotation.JsonGetter;
+import com.fasterxml.jackson.annotation.JsonProperty;
+
+import io.debezium.connector.oracle.Scn;
+import io.debezium.connector.oracle.logminer.processor.AbstractTransaction;
+import io.debezium.connector.oracle.logminer.processor.infinispan.marshalling.VisibleForMarshalling;
+
+public class EhcacheTransaction extends AbstractTransaction {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(EhcacheTransaction.class);
```
Naros, line 23: Doesn't appear to be used at this time, remove?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/echache/LogMinerObjectType.java
```diff
@@ -0,0 +1,18 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.oracle.logminer.processor.echache;
+
+public enum LogMinerObjectType {
```
Naros, line 8: Is there a reason for this to be independent of a nested enum in `LogMinerEvent`?  If we do keep this separate, we should add some javadocs here.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/AbstractInfinispanLogMinerEventProcessor.java
```diff
@@ -28,34 +13,24 @@
...
  * An implementation of {@link io.debezium.connector.oracle.logminer.processor.LogMinerEventProcessor}
  * that uses Infinispan to persist the transaction cache across restarts on disk.
  *
  * @author Chris Cranford
  */
-public abstract class AbstractInfinispanLogMinerEventProcessor extends AbstractLogMinerEventProcessor<InfinispanTransaction> implements CacheProvider {
+public abstract class AbstractInfinispanLogMinerEventProcessor extends AbstractTransactionCachingLogMinerEventProcessor<InfinispanTransaction>
+        implements CacheProvider<InfinispanTransaction> {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(AbstractInfinispanLogMinerEventProcessor.class);
```
Naros, line 33: With the refactor, it seems this is no longer used.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/memory/MemoryLogMinerEventProcessor.java
```diff
@@ -41,6 +43,7 @@
  *
  * @author Chris Cranford
  */
+// TODO: can this be a caching impl now as well, just with map
```
Naros, line 46: Is this todo still relevant?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
```diff
@@ -0,0 +1,483 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractTransactionCachingLogMinerEventProcessor.class);
+    private final OracleConnection jdbcConnection;
+    private final LogMinerStreamingChangeEventSourceMetrics metrics;
+    private final OraclePartition partition;
+    private final OracleOffsetContext offsetContext;
+    private final EventDispatcher<OraclePartition, TableId> dispatcher;
+
+    private InMemoryPendingTransactionsCache inMemoryPendingTransactionsCache = new InMemoryPendingTransactionsCache();
+
+    private static AbstractTransactionCachingLogMinerEventProcessor instance;
```
Naros, line 54: This will get us in trouble in Kafka Connect where multiple connector instances will run.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
```diff
@@ -0,0 +1,483 @@
...
+public abstract class AbstractTransactionCachingLogMinerEventProcessor<T extends Transaction> extends AbstractLogMinerEventProcessor<T> implements CacheProvider<T> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractTransactionCachingLogMinerEventProcessor.class);
+    private final OracleConnection jdbcConnection;
+    private final LogMinerStreamingChangeEventSourceMetrics metrics;
+    private final OraclePartition partition;
+    private final OracleOffsetContext offsetContext;
+    private final EventDispatcher<OraclePartition, TableId> dispatcher;
+
+    private InMemoryPendingTransactionsCache inMemoryPendingTransactionsCache = new InMemoryPendingTransactionsCache();
```
Naros, line 52: Can be made final.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
```diff
@@ -0,0 +1,483 @@
...
+        }
+        return first;
+    }
+
+    protected Optional<T> getOldestTransactionInCache() {
+        AtomicReference<T> oldestTransaction = new AtomicReference<>(null);
+
+        getTransactionCache().forEach((k, transaction) -> {
+
+            T finalTransaction = transaction;
```
Naros, line 146: This seems redundant.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
```diff
@@ -0,0 +1,483 @@
...
+    }
+
+    @Override
+    protected T getAndRemoveTransactionFromCache(String transactionId) {
+        // todo: Infinispan bug?
+        // When interacting with ISPN with a remote server configuration, the expected
+        // behavior was that calling the remove method on the cache would return the
+        // existing entry and remove it from the cache; however it always returned null.
+        //
+        // For now, we're going to use get to obtain the value and then remove it after-the-fact.
```
Naros, line 239: I believe in this case, the cache needs to be used like:
```java
Object returnValue = cache.withFlags(Flags.FORCE_RETURN_VALUE).remove(key);
```

debezium-connector-oracle/pom.xml
```diff
@@ -46,6 +46,11 @@
             <groupId>org.antlr</groupId>
             <artifactId>antlr4-runtime</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.ehcache</groupId>
+            <artifactId>ehcache</artifactId>
+            <version>3.8.1</version>
```
Naros, line 52: For now, this is fine so we can keep the build localized to just the Oracle connector; however once we near the end of this PR's lifecycle, we need to move the version to the main pom where we maintain the other versions.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/Transaction.java
```diff
@@ -66,9 +66,16 @@ public interface Transaction {
...
     /**
      * Helper method that resets the event identifier back to {@code 0}.
-     *
+     * <p>
      * This should be called when a transaction {@code START} event is detected in the event stream.
      * This is required when LOB support is enabled to facilitate the re-mining of existing events.
      */
     void start();
+
+    default String getEventId(int index) {
```
Naros, line 75: Javadoc.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/memory/MemoryLogMinerEventProcessor.java
```diff
@@ -273,33 +285,38 @@ protected Scn calculateNewStartScn(Scn endScn, Scn maxCommittedScn) throws Inter
 
     @Override
     protected Scn getTransactionCacheMinimumScn() {
-        return transactionCache.values().stream()
-                .map(MemoryTransaction::getStartScn)
-                .min(Scn::compareTo)
-                .orElse(Scn.NULL);
+        AtomicReference<Scn> result = new AtomicReference<>();
```
Naros, line 288: Mostly an FYI, I don't mind the use of atomics, my only concern is their impact on performance, especially if used within the context of a processing loop.  If it's once per iteration, then I would say those are generally okay.

Naros (commented):
I have just a few inline nit-picks at the moment as it seems the code is still a bit in flux, but overall I am really liking the refactor work :+1:

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -431,6 +434,38 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
...
 
+    public static final Field LOG_MINING_BUFFER_EHCACHE_CACHE_PATH = Field.create("log.mining.buffer.ehcache.path")
+            .withDisplayName("Ehcache storage path configuration")
+            .withType(Type.STRING)
+            .withWidth(Width.LONG)
+            .withImportance(Importance.LOW)
+            .withValidation(OracleConnectorConfig::validateLogMiningEhcacheCacheConfiguration)
+            .withDescription("Ehcache storage path configuration");
+
+    public static final Field LOG_MINING_BUFFER_EHCACHE_CACHE_TRANSACTION_SIZE_GB = Field.create("log.mining.buffer.ehcache.cache.transaction.size.gb")
```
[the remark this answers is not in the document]
jeremy-l-ford: My current plan is to split this into the refactor PR and leave the ehcache impl as another PR.  The ehcache impl is here atm as a POC to prove out the the LogMinerCache interface will work for another cache.   Will keep this in mind though.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/echache/EhcacheLogMinerEventProcessor.java
jeremy-l-ford: The ehcache impl is a minimal rough impl just to get the unit test to function.  Apologies about that.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/memory/MemoryLogMinerEventProcessor.java
jeremy-l-ford: Yes.  The previous/current impl of MemoryLogMinerEventProcessor does not extend from the new AbstractTransactionCachingLogMinerEventProcessor impl atm.  It seems like it could extend it, but wasn't sure if it would still be better to leave it as is.  I plan on trying it out and running through the test suite to verify.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
jeremy-l-ford: The instance only seems to be used by logCacheStats which has a comment about being useful for Debezium embedded mode.  Doesn't appear to be used by DBZ itself.  Should we just remove it?

jeremy-l-ford:
Made changes.  Still need to go back through and add javadoc and probably code cleanup.  Main things to note is the change to the LogMinerCache to use the streamAndReturn method to avoid use of Atomics.

Also, changed abandonTransactions in AbstractLogMinerEventProcessor to use streamAndReturn and simplify the impl.  The consequence is that the keys to remove are now always in memory (vs just during debug).  Is this a concern?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/memory/MemoryLogMinerEventProcessor.java
Naros: I think its perfectly fine to extend it; I would just suggest doing that part in a separate commit in the same PR would be the only ask, so we sort of distinguish between refactors & ehcache implementation details.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
Naros: Oh, I didn't realize that was part of the current code base :thinking: 

I wouldn't change necessarily in this work then for EhCache, let me add a Jira and take care of that on the main branch because we should definitely fix this asap.

Naros:
> Also, changed abandonTransactions in AbstractLogMinerEventProcessor to use streamAndReturn and simplify the impl. The consequence is that the keys to remove are now always in memory (vs just during debug). Is this a concern?

In general, it shouldn't be.  The default out-of-the-box configuration doesn't abandon transactions anyway, and even if users configure this option, you generally would only expect a few hundred to maybe a thousand or so transactions to be present in this list anyway at any given moment in time, depending on the system's activity.

jeremy-l-ford:
Wrote up JIRA https://issues.redhat.com/browse/DBZ-8054 to cover the refactor only.  
Do you want to continue reviewing here or in a new PR?  Any other changes or suggestions?

jeremy-l-ford:
I reworked AbstractLogMinerEventProcessor#getLastScnToAbandonFallbackByTransactionChangeTime to be stream based.  I think I translated that logic properly.  Unit tests pass at least.

jeremy-l-ford:
Opened up https://github.com/debezium/debezium/pull/5688.

Naros:
> Wrote up JIRA [issues.redhat.com/browse/DBZ-8054](https://issues.redhat.com/browse/DBZ-8054) to cover the refactor only. Do you want to continue reviewing here or in a new PR? Any other changes or suggestions?

If all the changes proposed here are in the other PR you just opened, we can then continue the discussion there I think and formerly close this draft PR if that works for you @jeremy-l-ford .

PR #5643, opened 2024-06-25
Document Postgres' Limitations

Maithem:
This patch documents possible inconsistency issues caused by replication slots.


Naros:
Thanks, and congrats on your first time contribution @Maithem :tada:

PR #5645, opened 2024-06-25
DBZ-7976 Mark EmbeddedEngine as deprecated

vjuranek:
https://issues.redhat.com/browse/DBZ-7976


Naros:
Simple and effective, thanks @vjuranek !

PR #5646, opened 2024-06-26
DBZ-7760 MongoDb collection scoped streaming

ani-sha:
https://issues.redhat.com/browse/DBZ-7760

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoUtils.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoUtils.java
@@ -220,6 +220,15 @@ public class MongoUtils {
             return client.getDatabase(database).watch(pipeline.getStages(), BsonDocument.class);
         }
 
+        // capture scope is collection
+        if (config.getCaptureScope() == MongoDbConnectorConfig.CaptureScope.COLLECTION) {
+            var captureTarget = config.getCaptureTarget().orElseThrow();
+            var database = captureTarget.split("\\.")[0];
+            var collection = captureTarget.split("\\.")[1];
+            LOGGER.info("Change stream is restricted to '{}' collection", collection);
+            return client.getDatabase(database).getCollection(collection).watch(pipeline.getStages(), BsonDocument.class);
+        }
+
         // capture scope is deployment
         return client.watch(pipeline.getStages(), BsonDocument.class);
     }
--- a/documentation/modules/ROOT/pages/connectors/mongodb.adoc
+++ b/documentation/modules/ROOT/pages/connectors/mongodb.adoc
@@ -1760,7 +1760,12 @@ To support {link-prefix}:{link-signalling}#[{prodname} signaling], if you set `c
 +
 [NOTE]
 ====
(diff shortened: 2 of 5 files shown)


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoUtils.java
```diff
@@ -220,6 +220,15 @@ public static ChangeStreamIterable<BsonDocument> openChangeStream(MongoClient cl
             return client.getDatabase(database).watch(pipeline.getStages(), BsonDocument.class);
         }
 
+        // capture scope is collection
+        if (config.getCaptureScope() == MongoDbConnectorConfig.CaptureScope.COLLECTION) {
+            var captureTarget = config.getCaptureTarget().orElseThrow();
+            var database = captureTarget.split("\\.")[0];
+            var collection = captureTarget.split("\\.")[1];
```
jcechace, line 227: Seeing as this gets repeated. I'd refactor `MongoDbConnectorConfig#getCaptureTarget` to return `Optional<CollectionId>` instead :)

jcechace: CollectionId actually requires both parts, so lets leave it like this for now.

ani-sha:
Hi @roldanbob, Could you please review the docs? Thank you!

jcechace (approved):
Once reviewed by @roldanbob  this can be merged

jcechace:
@roldanbob It really depends. It's definitively going to be part of the downstream release - I'd personally designate it as dev preview for now as we don't yet know how to deal with the limitations.  However the CDC side of things works fine, it just poses restrictions on other features.

documentation/modules/ROOT/pages/connectors/mongodb.adoc
```diff
@@ -1753,6 +1756,18 @@ Set this property to one of the following values:
...
+`collection`:: Opens a change stream cursor for a single collection to watch for changes to that collection.
++
+[NOTE]
+====
+This feature is currently in incubating state, i.e. exact semantics, configuration options etc. may change in future revisions, based on the feedback we receive.
+====
++
+[WARNING]
+====
+When `capture.scope` is set to `collection`, the connector does not support {link-prefix}:{link-signalling}#[{prodname} signaling] via the `source` channel which is enabled by default. Due to this limitation the connector also does not support incremental snapshots.
```
roldanbob, line 1768:
```suggestion
When you set the value of the `capture.scope` property to `collection`, the connector cannot process signals sent through the default `source` {link-prefix}:{link-signalling}[signaling] channel.
Because the `source` channel is required for processing incremental snapshot signals received over any channel, including the Kafka, JMX, and file channels, this limitation prevents the connector from performing incremental snapshots.
```

roldanbob (changes requested):
The main content looks good. I conditionalized incubating note and added conditionalized Dev Preview note. 
I also revised the WARNING to expressly state that not having use of the `source` channel prevents the connector from processing signals sent over other channels. I had to go back an look this up, but I think I got the details right. Let me know if something looks wrong. Thanks.

jpechane:
@ani-sha Applied, thanks a lot! Could you please create a Jira with follow-up tasks? Thanks

PR #5648, opened 2024-06-27
DBZ-7964 (quickfix) Only reset error retry counter if poll() successfully produced records

jcechace:
https://issues.redhat.com/browse/DBZ-7964


jpechane:
@jcechace Applied, thanks

PR #5649, opened 2024-06-27
DBZ-7986 Fix testcontainers Kafka container

obabec:
https://issues.redhat.com/browse/DBZ-7986

### DIFF
--- a/debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/DebeziumKafkaContainer.java
+++ b/debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/DebeziumKafkaContainer.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.testing.testcontainers;
+
+import org.testcontainers.containers.KafkaContainer;
+import org.testcontainers.containers.Network;
+import org.testcontainers.utility.DockerImageName;
+
+public class DebeziumKafkaContainer {
+    private static final String defaultImage = "confluentinc/cp-kafka:7.2.10";
+
+    public static KafkaContainer defaultKRaftContainer(Network network) {
+        try (
+                KafkaContainer kafka = new KafkaContainer(DockerImageName.parse(defaultImage))
+                        .withNetwork(network)
+                        .withKraft()) {
+            return kafka;
+        }
+        catch (Exception e) {
+            throw new RuntimeException("Cannot create KRaftContainer with default image.", e);
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/DebeziumKafkaContainer.java
```diff
@@ -0,0 +1,37 @@
...
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.testing.testcontainers;
+
+import org.testcontainers.containers.KafkaContainer;
+import org.testcontainers.containers.Network;
+import org.testcontainers.utility.DockerImageName;
+
+public class DebeziumKafkaContainer {
+    private static final String defaultImage = "confluentinc/cp-kafka:7.2.10";
```
rk3rn3r, line 13: I don't think we should use Confluent's Kafka container / Confluent Platfrom? @jpechane

obabec: @rk3rn3r We have been using it for the whole time. It was just hidden in Kafka Container default logic. Let me check if there is a possibility to easily swap to apache one.

rk3rn3r: Sure, I think so. The main reason is that confluent provides a very good configuration that properly handles docker/cloud scenarios with external and internal IPs/hostnames etc which, at least in the past, was missing from the original Apache Kafka container. I don't know if they closed this gap.

jpechane: TBH, if it would rquire a major effort then I am fine with keeping things as they are. For me it is more important for testing that we use our Connect container.

obabec: So, I have checked the Apache container for TC. It seems that we would have to upgrade TC dependency and it is still a little bit behind. For example, it does not support KRaft, which TBH significantly speeds up the deployment for us. I think we should keep this as is so we can migrate the rest of the TF images to CentOS Stream 9 and create Jira to revisit this later with proper investigation. WDYT @jpechane @rk3rn3r

obabec: Honestly, I would very much appreciate merging this, as it is the only thing that is blocking me from rebuilding the rest of the TF slaves, and we can be calm about the deprecation of repositories there...

jpechane:
@obabec Applied, thanks! Please create a follow-up Jira for the investigation. But as I told if it would require a lot of effort then just keep it as is.

PR #5653, opened 2024-06-28
DBZ-7835 Use SEGMENT_COLUMN_ID instead of COLUMN_ID as the order of c…

butioy:
DBZ-7835 Use SEGMENT_COLUMN_ID instead of COLUMN_ID as the order of columns in the table


Naros:
Closed in favor of #5665

PR #5659, opened 2024-07-03
DBZ-7224 Use JDK 21 as build/runtime target for Outbox

Naros:
https://issues.redhat.com/browse/DBZ-7224

Follows on https://github.com/debezium/debezium/pull/5654

### DIFF
--- a/debezium-quarkus-outbox-common/pom.xml
+++ b/debezium-quarkus-outbox-common/pom.xml
@@ -14,6 +14,14 @@
     <name>Debezium Quarkus :: Outbox :: Common</name>
     <packaging>pom</packaging>
 
+    <properties>
+        <!-- Java version used for Debezium specific components - Server, Operator, Outbox -->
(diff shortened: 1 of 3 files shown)


debezium-quarkus-outbox-common/pom.xml
```diff
@@ -14,6 +14,14 @@
     <name>Debezium Quarkus :: Outbox :: Common</name>
     <packaging>pom</packaging>
 
+    <properties>
+        <!-- Java version used for Debezium specific components - Server, Operator, Outbox -->
+        <debezium.java.specific.target>21</debezium.java.specific.target>
```
jpechane, line 19: @Naros This should not be overridden here and in other places. The plan is that as soon as we have Java 17 settled for everything then this property will be changed centrally via https://github.com/debezium/debezium/pull/5658

jcechace:
Should be fixed in https://github.com/debezium/debezium/pull/5660

PR #5663, opened 2024-07-08
DBZ-8027 updated type "money" fallback value creating.

GitHubSergei:
The fallback value is now been created in accordance with decimalMode that is set in the connector configuration. So it always fits the scheme.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -478,6 +478,7 @@ Seongjoon Jeong
 Sergei Morozov
 Sergey Eizner
 Sergey Ivanov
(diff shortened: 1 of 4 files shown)


Naros:
Hi @GitHubSergei, can you please add a test in the test suite for this specific use case.  Thanks.

Naros:
Hi @GitHubSergei, just a quick follow-up on whether you have the cycles to add a test for this change.

jpechane:
Rebased on the latest main, added tests and force pushed. If passes CI the PR will be merged.

PR #5677, opened 2024-07-12
DBZ-8059 Updated FakeDNS to JDK 18 InetAddressResolver SPI

jcechace:
https://issues.redhat.com/browse/DBZ-8059

The actual behaviour changed.

Previously FakeDNS was "installed" only when explicitly requested and the internal `InetAddress` state was restored when it was disabled. All of this was done with ugly reflection. 

Current behaviour is as that the `FakeDnsAddressResolver` is active at all times -- if its provider is register under `META-INF/services`. However unless `FakeDNS` is enabled, the  `FakeDnsAddressResolver`  will effectively only perform a passthrough to the default platform resolver. If `FakeDNS` is enabled then the resolutions are first attempted against its entries and if not found then the default platform resolver is used as fallback. 

Thus the effective behaviour is the same as previously even though the resolver is active at all times.


debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/util/dns/FakeDns.java
```diff
@@ -0,0 +1,94 @@
...
+
+    public void removeResolution(String hostname, InetAddress ip) {
+        registry.remove(hostname, ip);
+        LOGGER.info("Removed dns resolution: {} -> {}", hostname, ip);
+    }
+
+    public Optional<InetAddress> resolve(String hostname) {
+        return isEnabled()
+                ? resolveInternal(hostname)
+                : Optional.empty();
```
jcechace, line 53: FakeDNS only reports resolutions when it is enabled

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/util/dns/FakeDnsAddressResolverProvider.java
```diff
@@ -0,0 +1,69 @@
...
+            this.fallbackResolver = fallbackResolver;
+        }
+
+        @Override
+        public Stream<InetAddress> lookupByName(String host, LookupPolicy lookupPolicy) throws UnknownHostException {
+            var maybeAddress = fakeDns.resolve(host);
+
+            return maybeAddress.isPresent()
+                    ? maybeAddress.stream()
+                    : fallbackResolver.lookupByName(host, lookupPolicy);
```
jcechace, line 57: If FakeDNS is disabled then it will not return a resolution. In such case the `FakeDnsAddressResolver` will use fallback resolver -- which is in our case is always the default (Platform) resolver.

Naros (approved):
LGTM @jcechace, once Ondrej has given his feedback, feel free to merge.

obabec (approved):
lgtm

PR #5682, opened 2024-07-15
DBZ-8068 Fix logging for LogMiner streaming metrics

Naros:
https://issues.redhat.com/browse/DBZ-8068


Naros:
Applied to main (=3.0) and branch 2.7.

PR #5686, opened 2024-07-16
DBZ-8041 Updated k8s manifests and examples with dev version at the end of release process

jcechace:
https://issues.redhat.com/browse/DBZ-8041


Naros (commented):
Applied, thanks @jc

PR #5688, opened 2024-07-16
DBZ 8054 cache refactor

jeremy-l-ford:


Naros:
Hi @jeremy-l-ford can you take a look at the one test failure related to `shouldNotEmitLastCommittedTransactionEventsUponRestart`?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
```diff
@@ -0,0 +1,435 @@
...
+                return;
+            }
+
+            String eventKey = transaction.getEventId(transaction.getNextEventId());
+            if (!getEventCache().containsKey(eventKey)) {
+                // Add new event at eventId offset
+                LOGGER.trace("Transaction {}, adding event reference at key {}", transactionId, eventKey);
+                getEventCache().put(eventKey, eventSupplier.get());
+                metrics.calculateLagFromSource(row.getChangeTime());
+                inMemoryPendingTransactionsCache.putOrIncrement(transaction.getTransactionId());
```
Naros, line 266: So I get why this pending transaction cache exists, you split the `MemoryTransaction` to where events are no longer stored in the Transaction like we did for Infinispan, which means we no longer have a solid way to get the event cache size per transaction.  But do we know if this introduces any measurable overhead for heap-based implementations, particularly when large transactions are at play?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
```diff
@@ -0,0 +1,435 @@
...
+     *
+     * @param minCacheScn the minimum system change number to keep entries until
+     */
+    protected void purgeCache(Scn minCacheScn) {
+        getProcessedTransactionsCache().removeIf(entry -> Scn.valueOf(entry.getValue()).compareTo(minCacheScn) < 0);
+        getSchemaChangesCache().removeIf(entry -> Scn.valueOf(entry.getKey()).compareTo(minCacheScn) < 0);
+    }
+
+    @Override
+    protected Scn calculateNewStartScn(Scn endScn, Scn maxCommittedScn) throws InterruptedException {
```
Naros, line 344: You can ignore these for now, this is something I am still working through:

// todo: this logic has changed (at least from the heap), we need to verify no regression introduced.
// todo: this matches the infinispan implementation line-by-line.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
```diff
@@ -0,0 +1,435 @@
...
+                }
+            }
+            Loggings.logWarningAndTraceRecord(LOGGER, row, "Cannot undo change on table '{}' since event with row-id {} was not found.", row.getTableId(),
+                    row.getRowId());
+        }
+    }
+
+    protected List<String> getTransactionKeysWithPrefix(String prefix) {
+        AtomicReference<List<String>> result = new AtomicReference<>();
+        getEventCache().keys(stream -> {
```
Naros, line 173: I am curious: if a user has many in-progress transactions, what overhead in performance does this iteration across all events in the cache cause, particularly if I have multiple large transactions?

Naros (commented):
Here are some initial feedback points. 

Btw, the heap tests without the Memory-processor refactor all pass.  I still need to identify the root-cause of the problem with the refactor commit.

Finally, I still need to run the Infinispan buffer to see if there are any regressions.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
Naros: I also failed to mention, the `Events` array that is maintained in the `MemoryTransaction`, it would seem that tis no longer applicable, so it can be removed, yes?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
```diff
@@ -0,0 +1,435 @@
...
+            // This means that Oracle LogMiner found an event that should be undone but its corresponding
+            // undo entry was read in a prior mining session and the transaction's sequence could not be
+            // resolved. In this case, lets locate the transaction based solely on XIDUSN and XIDSLT.
+            final String transactionPrefix = getTransactionIdPrefix(row.getTransactionId());
+            LOGGER.debug("Undo change refers to a transaction that has no explicit sequence, '{}'", row.getTransactionId());
+            LOGGER.debug("Checking all transactions with prefix '{}'", transactionPrefix);
+            eventKeys = getTransactionKeysWithPrefix(transactionPrefix);
+            if (!eventKeys.isEmpty()) {
+                // Enforce that the keys are always reverse sorted.
+                eventKeys.sort(EventKeySortComparator.INSTANCE.reversed());
```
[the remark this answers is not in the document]
jeremy-l-ford: Agree that it makes sense to either extract the logic or refactor this method.  Will take a look at what I can do.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
```diff
@@ -0,0 +1,435 @@
...
+        if (isRecentlyProcessed(transactionId)) {
+            LOGGER.debug("Transaction {} has been seen by connector, skipped.", transactionId);
+            return;
+        }
+        super.processRow(partition, row);
+    }
+
+    @Override
+    protected boolean hasSchemaChangeBeenSeen(LogMinerEventRow row) {
+        return getSchemaChangesCache().containsKey(row.getScn().toString());
```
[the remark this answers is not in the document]
jeremy-l-ford: Might be nice to just have Scn as the key type, but then would need infinispan to have a scn serializer.  I think an ScnAdapter would be needed.    Also, for the LogMinerEventAdapter, would the factory change from String to Scn as well?  

perhaps something to consider for a subsequent PR?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
jeremy-l-ford: I'm not sure what overhead this will introduce.  If we remove the events from the MemoryTransaction, seems like things might balance out.  Would need to run profiling against this to know.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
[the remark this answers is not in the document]
Naros: Agreed, I don't know what the cost overhead of an additional serializer would be in this case.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
Naros: Well, the `MemoryTransaction's event array is no longer used; it's just an empty collection that serves no purpose; hence, my request to remove it.  

However, my concern is the difference between a `List#size` and `List#add`; we have this repeated map put/get+increment tuple. I suspect the latter loses here in performance.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/memory/MemoryLogMinerEventProcessor.java
```diff
@@ -85,221 +50,28 @@ protected MemoryTransaction createTransaction(LogMinerEventRow row) {
...
-                dispatcher.dispatchHeartbeatEvent(partition, offsetContext);
-            }
-            else {
-                abandonTransactions(getConfig().getLogMiningTransactionRetention());
-                final Scn minStartScn = getTransactionCacheMinimumScn();
-                if (!minStartScn.isNull()) {
-                    offsetContext.setScn(minStartScn.subtract(Scn.valueOf(1)));
-                    dispatcher.dispatchHeartbeatEvent(partition, offsetContext);
-                }
-            }
```
Naros, line 269: Found the issue, this part is missing in the calculation, which causes the test to fail.  Reviewing the code more clearly but at least I know the culprit.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
```diff
@@ -0,0 +1,430 @@
...
+            if (!getLastProcessedScn().isNull() && getLastProcessedScn().compareTo(endScn) < 0) {
+                // If the last processed SCN is before the endScn we need to use the last processed SCN as the
+                // next starting point as the LGWR buffer didn't flush all entries from memory to disk yet.
+                endScn = getLastProcessedScn();
+            }
+
+            // update offsets
+            offsetContext.setScn(endScn);
+            metrics.setOldestScnDetails(minCacheScn, minCacheScnChangeTime);
+            metrics.setOffsetScn(endScn);
```
Naros, line 373:
```suggestion
            offsetContext.setScn(minCacheScn.isNull() ? endScn : minCacheScn.subtract(Scn.valueOf(1)));
            metrics.setOldestScnDetails(minCacheScn, minCacheScnChangeTime);
            metrics.setOffsetScn(offsetContext.getScn());
```

Naros: I need to test this with Infinispan, but I'm pretty confident this should likely solve both problems.  I'm actually shocked, however, this wasn't identified previously :thinking:

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
jeremy-l-ford: The in memory impl kept the events as part of the transaction whereas infinispan used this in memory pending cache.  Should the in memory pending cache be turned into a transient count on the transaction itself?

Naros:
Hmm, these tests failed locally with ISPN:
```
[ERROR]   OracleConnectorIT.shouldCorrectlyUndoWithMultipleSavepoints:4752 expected:<"update[a]"> but was:<"update[]">
[ERROR]   OracleConnectorIT.shouldNotEmitEventsOnConstraintViolations:2282->AbstractConnectorTest.assertNoRecordsToConsume:1004 expected:<[tru]e> but was:<[fals]e>
[ERROR]   OracleConnectorIT.shouldNotEmitEventsOnConstraintViolationsAcrossSessions:2398->AbstractConnectorTest.assertNoRecordsToConsume:1004 expected:<[tru]e> but was:<[fals]e>
[ERROR]   OracleConnectorIT.shouldUndoOnlyLastEventWithSavepoint:4689->AbstractConnectorTest.assertNoRecordsToConsume:1004 expected:<[tru]e> but was:<[fals]e>
```

jeremy-l-ford:
> Hmm, these tests failed locally with ISPN:

I assume you're seeing that locally?  Went and reordered the code to make it easier to compare against the original AbstractInfinispanLogMinerEventProcessor.  Also marked methods whose impls differed from the original implementation.  

Are you using local ISPN or remote ISPN?  The remove method is using the force return flag now where previously it was fetched first before removed.  Just pointing at that difference as well.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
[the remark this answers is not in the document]
Naros: Follow-up Jira - [DBZ-8110](https://issues.redhat.com/browse/DBZ-8110).

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractTransactionCachingLogMinerEventProcessor.java
Naros: Logged [DBZ-8011](https://issues.redhat.com/browse/DBZ-8111) to document this so we can review this later and see whether there are any performance concerns with this specific method.

Naros (approved):

PR #5689, opened 2024-07-17
[ci] Rename maven global env variable because of 3.9+ changes

obabec:


obabec:
\packit test

Naros (approved):
LGTM, thanks again for this @obabec !

PR #5692, opened 2024-07-18
DBZ-8022 Handle pause/resume polling for buffer changestreams cursor

ani-sha:
https://issues.redhat.com/browse/DBZ-8022

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
@@ -17,9 +17,6 @@ import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicReference;
-import java.util.concurrent.locks.Condition;
-import java.util.concurrent.locks.Lock;
-import java.util.concurrent.locks.ReentrantLock;
 
 import org.bson.BsonDocument;
 import org.slf4j.Logger;
@@ -37,6 +34,7 @@ import io.debezium.annotation.NotThreadSafe;
 import io.debezium.connector.mongodb.MongoDbConnector;
 import io.debezium.connector.mongodb.MongoDbTaskContext;
 import io.debezium.connector.mongodb.metrics.MongoDbStreamingChangeEventSourceMetrics;
+import io.debezium.pipeline.source.spi.ChangeEventSource;
 import io.debezium.util.Clock;
 import io.debezium.util.DelayStrategy;
 import io.debezium.util.Threads;
@@ -60,9 +58,6 @@ public class BufferingChangeStreamCursor<TResult> implements MongoChangeStreamCu
     private final ExecutorService executor;
     private final DelayStrategy throttler;
     private BsonDocument lastResumeToken = null;
-    private final Lock lock = new ReentrantLock();
-    private final Condition condition = lock.newCondition();
-    private volatile boolean paused = false;
 
     /**
      * Combination of change stream event and resume token
@@ -137,10 +132,12 @@ public class BufferingChangeStreamCursor<TResult> implements MongoChangeStreamCu
         private final MongoDbStreamingChangeEventSourceMetrics metrics;
         private final Clock clock;
         private int noMessageIterations = 0;
+        private final ChangeEventSource.ChangeEventSourceContext context;
 
         public EventFetcher(ChangeStreamIterable<TResult> stream,
                             int capacity,
                             MongoDbStreamingChangeEventSourceMetrics metrics,
+                            ChangeEventSource.ChangeEventSourceContext context,
                             Clock clock,
                             DelayStrategy throttler) {
             this.stream = stream;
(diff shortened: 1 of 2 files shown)


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -367,6 +391,27 @@ private ResumableChangeStreamEvent<TResult> pollWithDelay() {
...
 
+    public void pause() {
+        lock.lock();
+        try {
+            paused = true;
+        }
+        finally {
+            lock.unlock();
+        }
+    }
```
jcechace, line 402: This only synchronises the access to `paused`. It also means that when `condition#await` is called the current thread doesn't hold the lock

jcechace (changes requested):
Left some explaining comments. As it is right now the fetcher is not really paused.  The `BufferingChangeStreamCursor`'s `pause()` and `resume()` methods provide a nice interface, however they should only call methods on the fetcher. 

Look into `ChangeEventSource` how streaming pausing / snapshot completion are actually used (there are two conditions used, the methods "cross-wait" for the other one)

jcechace:
@ani-sha  imho the best course of action here is to pass `ChangeEventSourceContext` to the cursor (and down to the fetcher) and just use the exposed methods there :)

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -268,6 +273,15 @@ private Optional<ResumableChangeStreamEvent<TResult>> fetchEvent(MongoChangeStre
                     .or(() -> Optional.ofNullable(cursor.getResumeToken()).map(ResumableChangeStreamEvent::new));
         }
 
+        private void waitWhenStreamingPaused(ChangeEventSource.ChangeEventSourceContext context) throws InterruptedException {
+            if (context.isPaused()) {
+                LOGGER.info("Buffering change stream cursor paused");
+                context.streamingPaused();
+                context.waitSnapshotCompletion();
+                LOGGER.info("Buffering change stream cursor resumed");
+            }
```
jcechace, line 282: @ani-sha (cc @Naros) I'm not really sure whether this is the right approach either. 

The reason why I liked the initial pause/resume API on the `BufferingChangeStreamCursor` was that the flow of operation remained explicit and clear. 

It allowed for this workflow

The `MongoDbStreamingChangeEventSource` checked the context, 

1) call to `cursor.pause()` (which in turn should pause the fetcher)
2) call `context.streamingPaused();` to notify that the streaming loop reached a safe waiting spot and thus the streaming is to be considered safely paused
3) await snapshot completion by calling `context.waitSnapshotCompletion()`
4) call `cursor.resume()` (which in turn should resume the fetcher)
5) return and continue with the streaming loop 


Added benefit was that there was not tight coupling between `ChangeEventSourceContext` and the `BufferingChangeStreamCursor`. 

having this exact logic in both `BufferingChangeStreamCursor` and `MongoDbStreamingChangeEventSource` makes the action flow quite confusing as it's not clear when exactly is the notification "streaming safely paused, waiting conditions can proceed" (call to `context.streamingPaused()`) invoked -- depending on timing it could be invoked in the fetcher or when `MongoDbStreamingChangeEventSource#waitWhenStreamingPaused` is called. There are two problems tied to this 

1) One of the two calls of `context.streamingPaused()` will redundant (probably safe, however I'd say that the duplicit notification is error prone)
2) Although the call `MongoDbStreamingChangeEventSource#waitWhenStreamingPaused`  is (and should be!) the first operation in the main streaming loop, the trigger inside fetcher is not visible in `MongoDbStreamingChangeEventSource` and could lead a future developer to think that prior to this the streaming is still running.

jcechace: Thus if we decide that creating the aforementioned coupling is OK, the entire "pause" logic should be moved to the `BufferingChangeStreamCursor` and `MongoDbStreamingChangeEventSource#waitWhenStreamingPaused` should be removed, replacing the call to it with a call to `BufferingChangeStreamCursor#waitWhenStreamingPaused` (which would essentially perform the procedure outlined in the previous comment). 

We would have to make sure that potential interrupts are handled correctly

jcechace:
@Naros One thing I noticed, which slipped in the error handling PR, regarding the fetcher... 

```java
@Override
        public void run() {
            try (MongoChangeStreamCursor<ChangeStreamDocument<TResult>> cursor = stream.cursor()) {
                cursorRef.compareAndSet(null, cursor);
                running.set(true);
                noMessageIterations = 0;
                fetchEvents(cursor);
            }
                LOGGER.error("Fetcher thread interrupted", e);
                throw new DebeziumException("Fetcher thread interrupted", e);
            }
```

Catching `InterruptedException` clears out the interrupt flag, since we are just trowing an exception, is there a need to reset the flag?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
ani-sha: > Added benefit was that there was not tight coupling between ChangeEventSourceContext and the BufferingChangeStreamCursor

@jcechace  I agree on this. 
I do recall you suggested this approach in one of yours comments which I can't seem to find this PR now. I will wait for Chris's review and rework accordingly.

jcechace: I think it was in chat, I said it was a possibility which has some drawbacks :)

Naros:
Hi @jcechace, this comes down to code future-proofing.

My understanding is that since the fetcher thread where you catch this use case terminates and no code looks at the current thread's interrupt status after that, not explicitly resetting the flag is fine.   However, if, in the future, some additional code is added where the flag reset becomes essential, it just means we'd need to remember to reset it.

If it were me, I'd probably reset it regardless.

jcechace:
@ani-sha  looks good. I've cleaned up few things -- mainly naming, in order to keep the API clean (from fetcher's pov the reason for pausing should be arbitrary, it doesn't have to be snapshot).

jcechace (approved):

Naros (commented):
Thanks you two, LGTM as well 👍

Naros:
Applied to main (=3.0) and branch 2.7.

PR #5694, opened 2024-07-18
DBZ-8018 Add note to JDBC documentation about the primary.key.fields

mfvitale:
related to: https://github.com/debezium/debezium-connector-jdbc/pull/89


Naros (approved):
Thanks @mfvitale, applied.

PR #5695, opened 2024-07-18
DBZ-8075 Speed-up PostgresShutdownIT test

vjuranek:
https://issues.redhat.com/browse/DBZ-8075


Naros (approved):
Thanks @vjuranek applied.

PR #5700, opened 2024-07-19
DBZ-8078 Standard Webhooks auth secret config value is not marked as PASSWORD_PATTERN

TimoWilhelm:
In https://github.com/debezium/debezium-server/pull/108 a new configuration value `debezium.sink.http.authentication.webhook.secret` has been introduced. This value is currently not marked as a password and will be logged during initialization.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -515,6 +515,7 @@ Thomas Thornton
 Tim Loes
 tisonkun
 Timo Roeseler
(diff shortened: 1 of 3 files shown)


Naros (approved):
Hi @TimoWilhelm LGTM, great catch :+1:

Naros:
Thanks @TimoWilhelm, applied.

PR #5702, opened 2024-07-20
DBZ-8082: Pass Headers to Key/Value Converters

ryanvanhuuksloot:
Part of: https://issues.redhat.com/browse/DBZ-8082

We want to be able to use headers in the key/value converter. When a configuration is set, the headers will be passed through to the converters.

My gut reaction to the increased complexity this configuration adds would be to keep the code clean and change the default behaviour. Converters by default call the current method if you pass the Headers so I think we don't need a configuration. I understand the extra `cast` will be slower - but I'm curious if that really warrants the extra complexity.

```
    /**
     * Convert a Kafka Connect data object to a native object for serialization,
     * potentially using the supplied topic and headers in the record as necessary.
     *
     * <p>Connect uses this method directly, and for backward compatibility reasons this method
     * by default will call the {@link #fromConnectData(String, Schema, Object)} method.
     * Override this method to make use of the supplied headers.</p>
     * @param topic the topic associated with the data
     * @param headers the headers associated with the data; any changes done to the headers
     *        are applied to the message sent to the broker
     * @param schema the schema for the value
     * @param value the value to convert
     * @return the serialized value
     */
    default byte[] fromConnectData(String topic, Headers headers, Schema schema, Object value) {
        return fromConnectData(topic, schema, value);
    }
```

I still need to add some testing but I'd love a first pass at the general structure.

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/ConverterBuilder.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/ConverterBuilder.java
@@ -7,13 +7,13 @@ package io.debezium.embedded;
 
 import java.nio.charset.StandardCharsets;
 import java.util.ArrayList;
-import java.util.Collections;
 import java.util.List;
 import java.util.Properties;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 import java.util.stream.StreamSupport;
 
+import org.apache.kafka.common.header.internals.RecordHeaders;
 import org.apache.kafka.connect.source.SourceRecord;
 import org.apache.kafka.connect.storage.Converter;
 import org.apache.kafka.connect.storage.ConverterConfig;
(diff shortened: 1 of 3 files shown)


Naros (approved):
At a first glance, this LGTM @ryanvanhuuksloot.

@jpechane will be back on Friday I believe, so I'll ask that he give it a cursory check from his PoV before we merge this in case there is something I may have missed.

ryanvanhuuksloot:
Perfect - thanks!

I'm still tempted to say that an implementation more akin to https://github.com/debezium/debezium/compare/main...ryanvanhuuksloot:debezium:rvh.example-record-headers may be generally simpler to maintain and basically as fast.

I'm actually using ^ in our test environment over this PR because I don't have to change configs so I can give a basic idea of profiling if there is interest.

debezium-embedded/src/main/java/io/debezium/embedded/ConverterBuilder.java
```diff
@@ -148,8 +126,48 @@ private boolean shouldConvertHeadersToString() {
...
+    }
+
+    private KeyValueHeaders convertRecord(SourceRecord record, HeaderConverter headerConverter) {
+        Converter keyConverter = createConverter(formatKey, true);
+        Converter valueConverter = createConverter(formatValue, false);
+        String topicName = Objects.requireNonNullElse(record.topic(), TOPIC_NAME);
+
+        // If your key/value converters don't require headers for read or write, we can skip the header conversion to
+        // save some CPU cycles
+        if (shouldPassHeaders()) {
```
jpechane, line 140: I believe this is unnecessary optimization. If the header converter is present then the conversion and passing of headers should be done always.
It will also simplify the code

jpechane (changes requested):
@ryanvanhuuksloot Thanks for the PR. I think it is good start. Still I feel that `convertRecord` method is a bit convoluted and IMHO it is also because of the optimization. I don't think it is necessary as IMHO it is acceptable to pass headers to converter every time there is header converter set.

debezium-embedded/src/main/java/io/debezium/embedded/ConverterBuilder.java
ryanvanhuuksloot: Hi Jiri, Thanks for the review.

Sounds perfect to me - thoughts on this kind of change instead?
https://github.com/debezium/debezium/pull/5702#issuecomment-2246719521

ryanvanhuuksloot: 👋 There is a different implementation on this branch: https://github.com/debezium/debezium/compare/main...ryanvanhuuksloot:debezium:rvh.example-record-headers

It was the first pass at implementing but I was asked to do this without always passing headers.
But it always passes the headers and I believe is more aligned with what you are asking. If you look and think it is better I can swap this PR to use something more akin to that.

ryanvanhuuksloot: @jpechane did that make more sense?

ryanvanhuuksloot:
@jpechane Change made

I don't love the test but the engine only supports SourceRecord so without a larger refactor I had to hack around using a new engine.

We don't have a great environment for building locally so I assume some spotless may be off but the functionality should be available for review now.

jpechane:
Reformatted, rebased and force-pushed. When CI finishes the PR is ready for merge.

jpechane:
@ryanvanhuuksloot Applied, thanks!

ryanvanhuuksloot:
>Reformatted, rebased and force-pushed.

Thank you! I missed that there were more style issues.

PR #5703, opened 2024-07-20
DBZ-8083 Invalidate empty value passed for history topic

akanimesh7:
https://issues.redhat.com/browse/DBZ-8083


debezium-storage/debezium-storage-kafka/src/main/java/io/debezium/storage/kafka/history/KafkaSchemaHistory.java
```diff
@@ -111,7 +111,7 @@ public class KafkaSchemaHistory extends AbstractSchemaHistory {
             .withWidth(Width.LONG)
             .withImportance(Importance.HIGH)
             .withDescription("The name of the topic for the database schema history")
-            .withValidation(KafkaSchemaHistory.forKafka(Field::isRequired));
```
Naros, line 114: If you don't intend to supply a topic name, you shouldn't use `KafkaSchemaHistory`.  This implementation does not make sense by supplying an empty topic name.  In such use cases, you should be using the file-based history implementation or another that doesn't require a topic name.

Naros (changes requested):

PR #5704, opened 2024-07-20
V2.7.0.final with jdbc flags

dpatrica:
#### Summary
This pull request adds the ability to specify additional flags in the PostgreSQL connector configuration for Debezium. These flags will be appended to the connection string, allowing for greater flexibility and customization of the database connection.

#### Changes
- Introduced a new configuration property: database.jdbc.connection.flags.
- This property allows users to specify additional flags for the JDBC connection string. For example, it can be used as follows:
 
  "database.jdbc.connection.flags": "?targetServerType=master&tcpKeepAlive=true&prepareThreshold=0"
  
#### Rationale
In my company, it is crucial to set the prepareThreshold=0 flag in the PostgreSQL connection string. Without this flag, it is impossible to create a transaction with a “prepared statement.” This enhancement provides the necessary flexibility to accommodate such specific requirements, ensuring that Debezium can be used effectively in environments with similar constraints.

#### Example Usage
"database.jdbc.connection.flags": "?targetServerType=master&tcpKeepAlive=true&prepareThreshold=0"
This will result in a connection string that includes the specified flags, allowing the connector to operate as required by the user's environment.

### Testing
- Conducted integration tests to ensure that the connector works seamlessly with the specified flags.

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
@@ -550,6 +550,7 @@ public class PostgresConnectorConfig extends RelationalDatabaseConnectorConfig {
     public static final Field PORT = RelationalDatabaseConnectorConfig.PORT
             .withDefault(DEFAULT_PORT);
 
+    public static final Field JDBC_CONNECTION_FLAGS = RelationalDatabaseConnectorConfig.JDBC_CONNECTION_FLAGS;
     public static final Field PLUGIN_NAME = Field.create("plugin.name")
             .withDisplayName("Plugin")
             .withGroup(Field.createGroupEntry(Field.Group.CONNECTION_ADVANCED_REPLICATION, 0))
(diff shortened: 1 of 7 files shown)


dpatrica:
I'm sorry, I didn't create a "issue" topic.
Just wrote the code )

Naros (commented):
Hi @dpatrica, I'm not necessarily convinced this change is necessary.  Debezium already allows you to append any number of driver-specific configuration properties by using the `driver.*` connector configuration namespace.  So to set the driver's `prepareThreshold` configuration, you'd simply add this to your configuration:

```
driver.prepareThreshold=0
```

Coming back to your example of `targetServerType=master&tcpKeepAlive=true&prepareThreshold=0`, this should already be possible by amending the connector configuration as follows:

```
driver.targetServerType=master
driver.tcpKeepAlive=true
driver.prepareThreshold=0
```

Could you try using the `driver.*` namespace with your needed properties and verify it works?
Thanks

Naros:
Hi @dpatrica, any update on my comments above?

dpatrica:
Hi @Naros !
I apologize for the long response.
Yes, it works!
But I didn't find this anywhere in the documentation :(

Naros:
Thanks @dpatrica, I see now.  It would seem that this particular section in `ref-connector-pass-through-database-driver-configuration-properties.adoc` isn't referenced in all the connector docs.  I'll file a Jira and send a PR specifically to address that.  It seems we need this for PostgreSQL, Informix, Vitess, IBMi, and Spanner.

Naros:
I've added the documentation update for PostgreSQL.
Thanks again, I'm going to close this in favor of the change in #5712.

PR #5705, opened 2024-07-22
DBZ-8086 The MySqlParser.g4 parser supports statement as field name

caicancai:
https://issues.redhat.com/browse/DBZ-8086

https://github.com/antlr/grammars-v4/pull/4167/files


caicancai:
@Naros @jpechane PATL, thank you

Naros:
Hi @caicancai does this also apply to MariaDB?

caicancai:
> Hi @caicancai does this also apply to MariaDB? If so, could you amend the MariaDB Parser & its tests, too?
Unfortunately, I don't have a MariaDB test environment.

debezium-ddl-parser/src/test/resources/mysql/examples/dml_select.sql
```diff
@@ -322,4 +322,7 @@ FROM
     your_table,
     (SELECT @sum := 0) AS init
 ORDER BY
-    some_order_column;
\ No newline at end of file
+    some_order_column;
+
+-- statement is supported as the field name
+SELECT REPLACE(statement, ' ','') as statement from your_table;
```
Naros, line 328: So I'm a bit confused, I added this to the MySQL `dml_select.sql` without modifying the grammar, and this works without an error, see this:
```
#begin
-- statement is supported as the field name
SELECT REPLACE(statement, ' ','') as statement from your_table;
#end
```
Did you check whether the DDL parser failed without making changes to the grammar?

debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/generated/MySqlParser.g4
```diff
@@ -2952,6 +2952,7 @@ functionNameBase
     | JSON_VALID | JSON_TABLE | JSON_SCHEMA_VALID | JSON_SCHEMA_VALIDATION_REPORT
     | JSON_PRETTY | JSON_STORAGE_FREE | JSON_STORAGE_SIZE | JSON_ARRAYAGG
     | JSON_OBJECTAGG
+    | STATEMENT
```
Naros, line 2955: I also don't believe this is the right place for this change, it should be part of `keywordsCanBeId` rule.

Furthermore, this `keywordsCanBeId` already includes the `STATEMENT` keyword as it was added sometime ago specifically for MariaDB, so I don't think this change is needed based on the current `main` code base.

I would again redirect you to my above inquiry, does the DDL tests pass by simply adding the SQL but no change to the grammar.

Naros (changes requested):

Naros:
Hi @caicancai, I'm following up to see if you concur with my assessment above.
Can we close this pull request?

caicancai:
> 你好@caicancai，我正在跟进，看看您是否同意我上述的评估。 我们可以关闭此拉取请求吗？

Yes，I'm sorry that I can't follow up due to my recent busy schedule. Thank you for your follow-up.

caicancai:
> 你好@caicancai，我正在跟进，看看您是否同意我上述的评估。 我们可以关闭此拉取请求吗？

Yes，I'm sorry that I can't follow up due to my recent busy schedule. Thank you for your follow-up.

PR #5714, opened 2024-07-26
DBZ-8088 Add MariaDB connector to release pipeline

Naros:
https://issues.redhat.com/browse/DBZ-8088

@jpechane this is the follow-up we discussed for https://github.com/debezium/container-images/pull/386.  I'm not sure since this also modifies the release pipeline for 2.7 if this needs to be backported, deferring to you.


jpechane:
@Naros Applied, thanks. There is no need for backport.

PR #5717, opened 2024-07-29
DBZ-8055 Cap LogMiner range to min open redo thread last scn

Naros:
https://issues.redhat.com/browse/DBZ-8055

### Premise

We have identified cases where on Oracle RAC, due to each node flushing LGWR buffers independently, we cannot rely precisely on the existing LogMiner SCN range calculation.  In short, the range's `endScn` needs to be capped based on the minimum `LAST_REDO_CHANGE#` in `V$THREAD`, which represents the data of the last redo thread flush.  This cap is only applied to open redo threads, any threads that are disabled or closed will not factor into the upper bound cap.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
@@ -34,6 +34,7 @@ import io.debezium.connector.oracle.OracleConnectorConfig.LogMiningBufferType;
 import io.debezium.connector.oracle.OracleDatabaseSchema;
 import io.debezium.connector.oracle.OracleOffsetContext;
 import io.debezium.connector.oracle.OraclePartition;
+import io.debezium.connector.oracle.RedoThreadState.RedoThread;
 import io.debezium.connector.oracle.Scn;
 import io.debezium.connector.oracle.logminer.logwriter.CommitLogWriterFlushStrategy;
 import io.debezium.connector.oracle.logminer.logwriter.LogWriterFlushStrategy;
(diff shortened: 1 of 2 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
```diff
@@ -208,6 +209,29 @@ public void execute(ChangeEventSourceContext context, OraclePartition partition,
...
+                        if (!minOpenRedoThreadLastScn.isNull()) {
+                            if (minOpenRedoThreadLastScn.compareTo(endScn) < 0) {
+                                // There are situations where on first start-up that the startScn may be higher
+                                // than the last flushed redo thread SCN, in which case we should delay by one
+                                // iteration until the startScn is before the minOpenRedoTheadLastScn
+                                if (minOpenRedoThreadLastScn.compareTo(startScn) < 0) {
+                                    pauseBetweenMiningSessions();
+                                    continue;
+                                }
+                                LOGGER.info("Minimum Redo Log Flush SCN is {}, using range {} to {} instead of {} to {}.", minOpenRedoThreadLastScn, startScn,
```
Naros, line 229: This needs to be changed to `debug`, will do so after testing this change with the tests.  I think it also makes sense to refactor this logminer range logic into a separate class, but I'll do that in a separate commit as well.

Naros:
The Oracle 19 XStream failure has been recorded in [DBZ-8102](https://issues.redhat.com/browse/DBZ-8102), and a PR has already been opened to make it retriable.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerStreamingChangeEventSource.java
jpechane: +1 for refactoring!

jpechane:
@Naros Applied, thanks! This part of the code becomes a small expert system :-).

PR #5718, opened 2024-07-29
DBZ-8035: Expose a metric for number of create, update, delete events per table

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-8035?filter=12414517

- [x] Add tests for other connectors
- [x] Provide a JMX operation to enable/disable it at runtime


Naros:
Hi @mfvitale, not to intentionally play scope-creep here, but I wonder what you think about also adding a truncate counter here to these metrics?  There are some connectors like Oracle/PostgreSQL (I think) that support truncate operations, too.

mfvitale:
> Hi @mfvitale, not to intentionally play scope-creep here, but I wonder what you think about also adding a truncate counter here to these metrics? There are some connectors like Oracle/PostgreSQL (I think) that support truncate operations, too.

@Naros good point. I'll add it in this PR.

jpechane:
@mfvitale Applied, great job! I've just moved activity classes into their own package.

PR #5719, opened 2024-07-29
DBZ-8102 Make ORA-26928 "Unable to communicate with XStream apply coordinator process" retriable

Naros:
https://issues.redhat.com/browse/DBZ-8102


jpechane:
@Naros Applied, thanks

PR #5724, opened 2024-07-31
DBZ-8105 Upgrade Kafka to 3.8.0

jpechane:
https://issues.redhat.com/browse/DBZ-8105


jpechane:
@Naros We will probably need to backport `JdbcSinkConnectorTask` into 2.7 just with the opposite logic to make 2.7 compatible with Kafka 3.8.

PR #5725, opened 2024-07-31
[docs] Fix rabbitmq config link

vjuranek:


jpechane:
@vjuranek Applied, thanks

PR #5726, opened 2024-07-31
DBZ-8109 Don't close the connection after loading metadata in Timesca…

vjuranek:
…leDb SMT

https://issues.redhat.com/browse/DBZ-8109


jpechane:
@vjuranek Applied, thanks!

PR #5727, opened 2024-07-31
DBZ-8054 Workaround bug with ISPN and toList() collection

jeremy-l-ford:


Naros (approved):
LGTM, thanks @jeremy-l-ford

PR #5731, opened 2024-08-02
DBZ-7879 Retry to connect to Redis on failure in RedisOffsetBackingStore

nguymin4:
https://issues.redhat.com/browse/DBZ-7879
> Redis timeout occured when storing offset and then it got stuck in a crash loop of NullPointerException

I noticed the similar issue (Debezium 2.7) when testing locally with `docker compose`. Stop `redis` container and start it again put us in endless loop, until the debezium-server is manually restarted.

This PR attempts to fix that issue by `connect()` on every failure instead of just `RedisClientConnectionException`
`connect()` calls `closeClient()` (Related https://github.com/debezium/debezium/commit/84ab7f7b8adc9b9c437ea5c5829ee83b86d3a761) which has `null` guard so I believe this is safe but I may miss something.

- I updated the tests and documentation so this can be tested quite easily with `mvn test`.
- I can also try to add my example to https://github.com/debezium/debezium-examples/ as well so we can test this further if needed.

### DIFF
--- a/.github/workflows/debezium-workflow-pr.yml
+++ b/.github/workflows/debezium-workflow-pr.yml
@@ -435,7 +435,7 @@ jobs:
     name: "Storage Module"
     needs: [ check_style, file_changes ]
     runs-on: ubuntu-latest
-    if: ${{ needs.file_changes.outputs.common-changed == 'true' || needs.file_changes.outputs.mysql-changed == 'true' || needs.file_changes.outputs.mysql-ddl-parser-changed == 'true' || needs.file_changes.outputs.storage-changed == 'true' }}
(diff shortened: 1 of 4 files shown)


jpechane:
@nguymin4 Hi, please be patient. It seems that for unknow reaosn your PR does not trigger tests in CI. We'll try to fix it.

jpechane:
@nguymin4 I understand your motivation but I am not convinced about the solution the means that one of the branch eliminated.
Would it be better to keep them configure retires for them as well.

@zalmane Any comment?

nguymin4:
@jpechane I understand your concern. If we don't want to retry on every failure, we can remove that commit while keeping the first two commits which I believe is an improvement. 

Also it would be nice if you can check this PR for the example as well https://github.com/debezium/debezium-examples/pull/355 it's also beneficial for testing cases like this

What do you think?

jpechane:
@nguymin4 Yes, please don't retry on every failure. We could revisit this in a separate Jira/PR.

nguymin4:
@jpechane I will close this PR in favor of another one https://github.com/debezium/debezium/pull/5789 which only focuses on improving tests. Thank you.

PR #5732, opened 2024-08-02
DBZ-8119 Seed `endScn` as `Scn.NULL`

Naros:
https://issues.redhat.com/browse/DBZ-8119


jpechane:
@Naros Aplied, thanks

PR #5740, opened 2024-08-05
DBZ-8117 Correct log file name; add example titles and formatting

roldanbob:
[DBZ-8117](https://issues.redhat.com/browse/DBZ-8117)

This change updates instances in `logging.adoc` that incorrectly refer to the log4j properties file as `log4j.properties` rather than `connect-log4j.properties`. In addition, to improve scannability, unique titles are applied to examples, and example formatting is also applied.


roldanbob:
@Naros Per my message yesterday, it's not clear to me whether references to the Log4J properties file should also be changed in the [container images section](https://debezium.io/documentation/reference/nightly/operations/logging.html#configuring-log-level-docker). If you can confirm that the same changes apply there, I'll push a new commit to make the additional fixes.

roldanbob:
@jpechane Per conversation with @Naros, this should now be ready to go, bc, any references to `log4j.properties` in the community-only section that come after L329 can remain as-is.

jpechane:
@roldanbob Applied, thanks

PR #5743, opened 2024-08-06
Adaptions for Informix integration tests (DBZ-8114)

nrkljo:
In AbstractEventRouterTest, start connector after altering schema to avoid having to stop and start it again for Informix (which does not permit altering schema during capture)


nrkljo:
In Abstract(Incremental)SnapshotTest:s, waitForAvailableRecords 1s is sometimes not enough for Informix when frequently starting and restarting the capture with. Using records.waittime (default 2s) allows one to tune this somewhat.

Naros:
@nrkljo rebased on main to get past checkstyle failures.

jpechane:
@nrkljo Applied, thanks!

PR #5746, opened 2024-08-06
DBZ-8124 Adds info about limiting the scope of FLASHBACK/SELECT grants

roldanbob:
[DBZ-8124](https://issues.redhat.com/browse/DBZ-8124)

Provides guidance about how you might reduce the scope of the permissions that you set when creating the Debezium LogMiner user.


jpechane:
@roldanbob Applied, thanks

PR #5748, opened 2024-08-07
DBZ-8113 Preserve source info during emitting PK change

vjuranek:
https://issues.redhat.com/browse/DBZ-8113

### DIFF
--- a/debezium-core/src/main/java/io/debezium/relational/RelationalChangeRecordEmitter.java
+++ b/debezium-core/src/main/java/io/debezium/relational/RelationalChangeRecordEmitter.java
@@ -171,16 +171,19 @@ public abstract class RelationalChangeRecordEmitter<P extends Partition>
     protected void emitUpdateAsPrimaryKeyChangeRecord(Receiver<P> receiver, TableSchema tableSchema, Struct oldKey,
                                                       Struct newKey, Struct oldValue, Struct newValue)
             throws InterruptedException {
+        final OffsetContext offset = getOffset();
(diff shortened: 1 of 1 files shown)


debezium-core/src/main/java/io/debezium/relational/RelationalChangeRecordEmitter.java
```diff
@@ -171,16 +171,15 @@ protected boolean skipEmptyMessages() {
     protected void emitUpdateAsPrimaryKeyChangeRecord(Receiver<P> receiver, TableSchema tableSchema, Struct oldKey,
                                                       Struct newKey, Struct oldValue, Struct newValue)
             throws InterruptedException {
+        Struct deleteEnvelope = tableSchema.getEnvelopeSchema().delete(oldValue, getOffset().getSourceInfo(), getClock().currentTimeAsInstant());
```
jpechane, line 174: @vjuranek How about being really sure and call `getOffset().getSourceInfo()` only once into a variable and re-use that variable? Also in the subsequent call there is `getOffset` called twice. Maybe it should be called only once at the top of method too  just in case?

vjuranek: Right, originally I thought `getOffset().getSourceInfo()` returns a reference to `SourceInfo` object which is mutable, but you are right, it actually returns `Struct` with current values.

jpechane:
@vjuranek Applied, thanks

PR #5749, opened 2024-08-07
DBZ-8107 Remove references to `additional-condition` signal paramter

roldanbob:
[DBZ-8107](https://issues.redhat.com/browse/DBZ-8107)

Removes entries for the `additional-condition` parameter from snapshot and signal tables that describe the fields in signal records.
Please backport to 2.7.


jpechane:
@roldanbob Applied, thanks

PR #5752, opened 2024-08-08
DBZ-8128 Don't stop engine twice

vjuranek:
https://issues.redhat.com/browse/DBZ-8128


jpechane:
@vjuranek Applied, thanks

PR #5753, opened 2024-08-08
DBZ-8125 Make ORA-00600 errors related to `krvrdGetUID` retriable

Naros:
https://issues.redhat.com/browse/DBZ-8125


jpechane:
@Naros Could you please investigate the test failures?

Naros:
As these errors are related to the Apicurio discussion from today's call, I'm going to proceed with merging as the failures are unrelated to the change and tests pass when Apicurio is not used.

PR #5754, opened 2024-08-08
DBZ-8130 Preserve ordering of tables inside batch

ashishbinu:
Fixes [DBZ-8130](https://issues.redhat.com/browse/DBZ-8130)


jpechane:
@ashishbinu Applied, thanks!

PR #5756, opened 2024-08-10
DBZ-8103 Add config option to DecodeLogicalDecodingMessageContent SMT…

rkudryashov:
… to customize including null fields


jpechane:
@rkudryashov Applied, thanks! Could you please send a docs PR too?

PR #5758, opened 2024-08-12
DBZ-8134 Fix parsing of server ID

vjuranek:
https://issues.redhat.com/browse/DBZ-8134


jpechane:
@vjuranek Applied, thanks

PR #5760, opened 2024-08-12
DBZ-8135 In process signalling for Debezium Engine

jcechace:
https://issues.redhat.com/browse/DBZ-8135

I tried to be as non-invasive towards the public API as possible.

### DIFF
--- a/debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
+++ b/debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
@@ -177,6 +177,32 @@ public interface DebeziumEngine<R> extends Runnable, Closeable {
     }
 
     /**
+     * Signaler defines the contract for sending signals to connector tasks.
+     * @param <T> signal type
+     */
+    interface Signaler<T> {
+
+        /**
+         * Initialize the signaler with the engine.
+         *
+         * <p>
+         *     Note that it is up the the implementation to decide whether it is compatible with given engine.
+         * </p>
+         *
+         * @param engine the engine instance
+         * @param <E> type of engine
+         */
+        <E extends DebeziumEngine<?>> void init(E engine);
+
+        /**
+         * Send a signal to the connector.
+         *
+         * @param signal the signal to send
+         */
+        void signal(T signal);
+    }
+
+    /**
      * A builder to set up and create {@link DebeziumEngine} instances.
      */
     interface Builder<R> {
@@ -198,6 +224,15 @@ public interface DebeziumEngine<R> extends Runnable, Closeable {
          * @return this builder object so methods can be chained together; never null
          */
         Builder<R> notifying(ChangeConsumer<R> handler);
+
+        /**
+         * Use the specified signaler to send signals to the connector.
+         * @param signaler the signaler
+         * @return this builder object so methods can be chained together; never null
+         */
+        default Builder<R> using(Signaler<?> signaler) {
+            return this;
+        };
 
         /**
          * Use the specified configuration for the connector. The configuration is assumed to already be valid.
--- a/debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
+++ b/debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
@@ -19,6 +19,7 @@ import java.util.function.Consumer;
 import org.slf4j.LoggerFactory;
 
 import io.debezium.DebeziumException;
+import io.debezium.common.annotation.Incubating;
 import io.debezium.engine.format.ChangeEventFormat;
 import io.debezium.engine.format.KeyValueChangeEventFormat;
 import io.debezium.engine.format.KeyValueHeaderChangeEventFormat;
@@ -180,6 +181,7 @@ public interface DebeziumEngine<R> extends Runnable, Closeable {
      * Signaler defines the contract for sending signals to connector tasks.
      * @param <T> signal type
      */
+    @Incubating
     interface Signaler<T> {
 
         /**
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/SignalsIT.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/SignalsIT.java
@@ -32,6 +32,7 @@ import io.debezium.connector.postgresql.PostgresConnectorConfig.SnapshotMode;
 import io.debezium.connector.postgresql.spi.CustomActionProvider;
 import io.debezium.embedded.async.AbstractAsyncEngineConnectorTest;
 import io.debezium.junit.logging.LogInterceptor;
+import io.debezium.pipeline.signal.SignalRecord;
 import io.debezium.pipeline.signal.actions.Log;
 
 public class SignalsIT extends AbstractAsyncEngineConnectorTest {
@@ -63,11 +64,20 @@ public class SignalsIT extends AbstractAsyncEngineConnectorTest {
     }
 
     @Test
+    public void signalLogViaInProcessChannel() throws InterruptedException {
+        signalLog(false, false);
+    }
+
+    @Test
     public void signalLogWithEscapedCharacter() throws InterruptedException {
         signalLog(true);
     }
 
     private void signalLog(boolean includingEscapedCharacter) throws InterruptedException {
+        signalLog(includingEscapedCharacter, true);
+    }
+
+    private void signalLog(boolean includingEscapedCharacter, boolean useSource) throws InterruptedException {
         // Testing.Print.enable();
         final LogInterceptor logInterceptor = new LogInterceptor(Log.class);
 
@@ -81,6 +91,7 @@ public class SignalsIT extends AbstractAsyncEngineConnectorTest {
                 .with(PostgresConnectorConfig.DROP_SLOT_ON_STOP, Boolean.TRUE)
                 .with(PostgresConnectorConfig.SIGNAL_DATA_COLLECTION, includingEscapedCharacter ? signalTableWithEscapedCharacter : signalTable)
                 .with(CommonConnectorConfig.SIGNAL_POLL_INTERVAL_MS, "500")
+                .with(CommonConnectorConfig.SIGNAL_ENABLED_CHANNELS, "source,in-process")
                 .build();
         start(PostgresConnector.class, config);
         assertConnectorIsRunning();
@@ -93,15 +104,23 @@ public class SignalsIT extends AbstractAsyncEngineConnectorTest {
         // insert and verify a new record
         TestHelper.execute(INSERT_STMT);
 
-        // Insert the signal record
-        String insertSql = String.format("INSERT INTO %s VALUES('1', 'log', '{\"message\": \"Signal message at offset ''{}''\"}')",
-                includingEscapedCharacter ? signalTableWithEscapedCharacter : signalTable);
-        TestHelper.execute(insertSql);
+        int expectedNumRecords;
+        if (useSource) {
+            expectedNumRecords = 2;
+            // Insert the signal record
+            String insertSql = String.format("INSERT INTO %s VALUES('1', 'log', '{\"message\": \"Signal message at offset ''{}''\"}')",
+                    includingEscapedCharacter ? signalTableWithEscapedCharacter : signalTable);
+            TestHelper.execute(insertSql);
+        }
+        else {
+            expectedNumRecords = 1;
+            signaler.signal(new SignalRecord("1", "log", "{\"message\": \"Signal message at offset ''{}''\"}", null));
+        }
 
         waitForAvailableRecords(800, TimeUnit.MILLISECONDS);
 
-        final SourceRecords records = consumeRecordsByTopic(2);
-        assertThat(records.allRecordsInOrder()).hasSize(2);
+        final SourceRecords records = consumeRecordsByTopic(expectedNumRecords);
+        assertThat(records.allRecordsInOrder()).hasSize(expectedNumRecords);
         assertThat(logInterceptor.containsMessage("Signal message at offset")).isTrue();
     }
 
--- a/debezium-core/src/main/java/io/debezium/pipeline/signal/channels/process/InProcessSignalChannel.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/signal/channels/process/InProcessSignalChannel.java
@@ -55,6 +55,16 @@ public class InProcessSignalChannel implements SignalChannelReader, SignalChanne
         return Stream.ofNullable(signals.poll()).toList();
     }
 
+    /**
+     * Close the channel and drain all the signals
+     *
+     * <p>
+     *     Note: To properly synchronise closing and reading a lock should be used.
+     *     However this is not necessary as any concurrent call to these methods is
+     *     equal to serially calling read and close (or the other way around). The
+     *     connector  handles both cases (and the queue is already thread safe).
+     * </p>
+     */
     @Override
     public void close() {
         if (open.compareAndSet(true, false)) {
--- a/debezium-core/src/main/java/io/debezium/pipeline/signal/channels/process/InProcessSignalChannel.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/signal/channels/process/InProcessSignalChannel.java
@@ -1,0 +1,75 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.pipeline.signal.channels.process;
+
+import java.util.List;
+import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.stream.Stream;
+
+import org.slf4j.Logger;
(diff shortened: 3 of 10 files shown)


debezium-core/src/main/java/io/debezium/pipeline/signal/channels/process/SignalWriter.java
```diff
@@ -0,0 +1,15 @@
...
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.pipeline.signal.channels.process;
+
+import io.debezium.pipeline.signal.SignalRecord;
+
+/**
+ * Interface for writing signals to a channel.
+ */
+public interface SignalWriter {
```
mfvitale, line 13: What do you think to have rename it to `SignalChannelWriter` to maintain consistency with the reader counterpart?

debezium-core/src/main/java/io/debezium/pipeline/signal/channels/process/InProcessSignalChannel.java
```diff
@@ -0,0 +1,75 @@
...
+ * used for sending signal from the same JVM process
+ *
+ * <p>
+ *     Mainly targeted at Debezium Engine
+ * </p>
+ */
+@ThreadSafe
+public class InProcessSignalChannel implements SignalChannelReader, SignalWriter {
+    private static final Logger LOGGER = LoggerFactory.getLogger(InProcessSignalChannel.class);
+    public static final String CHANNEL_NAME = "process";
```
mfvitale, line 34:
```suggestion
    public static final String CHANNEL_NAME = "in-process";
```

jcechace: Agreed

debezium-core/src/main/java/io/debezium/pipeline/signal/channels/process/SignalWriter.java
jcechace: Sure, done

jcechace:
One thing to note -- currently signal is always sent to all tasks. Do you think it's needed to have the ability to target specific task? Proposal how to modify this solution? 

E.g.

-- adding taks id to `Signaler#signal` 
-- Leaving this completely to signaler implementations (the id can be part of the parameter type used as a signal)

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/SignalsIT.java
```diff
@@ -81,6 +91,7 @@ private void signalLog(boolean includingEscapedCharacter) throws InterruptedExce
                 .with(PostgresConnectorConfig.DROP_SLOT_ON_STOP, Boolean.TRUE)
                 .with(PostgresConnectorConfig.SIGNAL_DATA_COLLECTION, includingEscapedCharacter ? signalTableWithEscapedCharacter : signalTable)
                 .with(CommonConnectorConfig.SIGNAL_POLL_INTERVAL_MS, "500")
+                .with(CommonConnectorConfig.SIGNAL_ENABLED_CHANNELS, "source,process")
```
Naros, line 94:
```suggestion
                .with(CommonConnectorConfig.SIGNAL_ENABLED_CHANNELS, "source,in-process")
```

Naros (commented):
Hi @jcechace , just one small issue I saw with the `process` rename, otherwise LGTM so far.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/SignalsIT.java
jcechace: Good catch, forgot about it :)

mfvitale (approved):

debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEngineSignaler.java
```diff
@@ -0,0 +1,36 @@
...
+    }
+
+    @Override
+    public void signal(SignalRecord signal) {
+        engine.tasks().stream()
+                .map(EngineSourceTask::debeziumConnectTask)
+                .flatMap(Optional::stream)
+                .map(task -> task.getSignalChannel(InProcessSignalChannel.class))
+                .flatMap(Optional::stream)
+                .forEach(task -> task.signal(signal));
```
vjuranek, line 34: `task` is IMHO little bit confusing here, `signaler` would be more explanatory

debezium-core/src/main/java/io/debezium/pipeline/signal/channels/process/InProcessSignalChannel.java
```diff
@@ -0,0 +1,75 @@
...
+
+    @Override
+    public void init(CommonConnectorConfig connectorConfig) {
+        open.compareAndSet(false, true);
+        LOGGER.info("Reading signals from {} channel", CHANNEL_NAME);
+    }
+
+    @Override
+    public List<SignalRecord> read() {
+        return Stream.ofNullable(signals.poll()).toList();
```
vjuranek, line 52: possibly also add check if the channel is open here?

vjuranek (commented):
Few small suggestions, overall LGMT.

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
```diff
@@ -176,6 +176,28 @@ default boolean supportsTombstoneEvents() {
         }
     }
 
+    interface Signaler<T> {
```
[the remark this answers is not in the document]
jcechace: +1

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
```diff
@@ -199,6 +221,15 @@ interface Builder<R> {
          */
         Builder<R> notifying(ChangeConsumer<R> handler);
 
+        /**
+         * Use the specified signaler to send signals to the connector.
+         * @param signaler teh signaler
```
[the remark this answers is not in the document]
jcechace: +1

debezium-embedded/src/main/java/io/debezium/engine/source/EngineSourceTask.java
```diff
@@ -31,4 +35,11 @@ public DebeziumSourceTaskContext context() {
     public SourceTask connectTask() {
         return connectTask;
     }
-}
\ No newline at end of file
+
+    @SuppressWarnings("unchecked")
+    public Optional<BaseSourceTask<?, ?>> debeziumConnectTask() {
```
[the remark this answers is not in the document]
jcechace: That's true only for our tasks.  The entire implementation still seams to count with the possibility of running 3rd party connectors. Signalling obviously won't work there but this ensure a no-op instead of failure should anyone attempt to use it.

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
```diff
@@ -176,6 +176,28 @@ default boolean supportsTombstoneEvents() {
...
+         * Initialize the signaler with the engine.
+         *
+         * <p>
+         *     Note that it is up the the implementation to decide whether it is compatible with given engine.
+         * </p>
+         *
+         * @param engine the engine instance
+         * @param <E> type of engine
+         */
+        <E extends DebeziumEngine<?>> void init(E engine);
```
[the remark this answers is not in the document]
jcechace: That's a possibility -- initialise with the list of tasks instead of the engine. However that brings us to the fact that on API layer there is no notion of task or anything. 

Or we could  actually initialise with the list of Signal channels, however then we will have to do few things

1) Bring signal channels (at least SignalChannelWriter) to the public API  (which they already sort of are as they are meant to be extendible anyway (nevertheless it would mean moving these contracts to debezium-api module)
2) Ensure that the write has information about the the specific tasks (or we could pass in something like `Map<task_id, signal_writer>` 

In all of these cases it means creating more detailed API. The current way essentially doesn't add anything to the API besides this interface (which doesn't expose any new information)

@jpechane  if you are up for it, I'm willing to expand the API. However in such case I would actually argue that there should also be a contract for the signal record.

debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEngineSignaler.java
```diff
@@ -0,0 +1,36 @@
...
+        this.engine = (AsyncEmbeddedEngine<?>) engine;
+    }
+
+    @Override
+    public void signal(SignalRecord signal) {
+        engine.tasks().stream()
+                .map(EngineSourceTask::debeziumConnectTask)
+                .flatMap(Optional::stream)
+                .map(task -> task.getSignalChannel(InProcessSignalChannel.class))
+                .flatMap(Optional::stream)
```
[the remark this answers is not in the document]
jcechace: Most likely, I will look into it

debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEngineSignaler.java
jcechace: it's supposed to be `channel`... the ai based autocompletion is driving me crazy :). I was fixing this 3 times already..

debezium-core/src/main/java/io/debezium/pipeline/signal/channels/process/InProcessSignalChannel.java
jcechace: Is there a point in doing that? I mean there is no harm but in case it isn't the read will just return empty list.  Which TBF looks safer (there is no danger of exception) in case we overlooked some concurrency sync point and read is called when not expected (no need to be extra careful here due to the nature of what signals are).

vjuranek: Actually it's safer to throw an exception, in case you overlook something, you are notified you overlooked something (and there may be some more side effects). Also, when you read from closed channel and wait for signal, you would be stuck there forever. Basically reading from closed channel shouldn't happen, if it happen it should be either fixed (better option) or the exception can be used as a signal to closer the signal reader (not that great option to use exception as signals, but IMHO still better than reading from closed channel without noticing it's closed).

debezium-embedded/src/main/java/io/debezium/engine/source/EngineSourceTask.java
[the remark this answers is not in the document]
vjuranek: I don't think any task running inside Debezium server doesn't inherit from `BaseSourceTask`, but why not to check it explicitly. I'd just rename it to something like `debeziumSourceTask` or something, just to avoid some connotation to Kafka Connect.

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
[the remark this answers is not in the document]
vjuranek: Maybe this is misunderstanding, I'm not definitely proposing extending the public API. I'm just proposing to change this one to `<E extends DebeziumEngine<?>> void init(E engine, String signalClassName);` or `<E extends DebeziumEngine<?>> void init(E engine, Class<S> signlaChannel);`

debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEngineSignaler.java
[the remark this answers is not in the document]
jcechace: There is always going to be just few tasks with few channels. I'm not sure whether it's worth adding caching in this case.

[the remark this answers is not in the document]
vjuranek: Depends how often you are going to send the signals. Anyway, it was just a proposal, I don't have any strong opinion about this one, so I'm fine keep it as it is if you think this is good.

debezium-core/src/main/java/io/debezium/pipeline/signal/channels/process/InProcessSignalChannel.java
jcechace: > you would be stuck there forever

That's not true, all our channels return an empty List in such case. This would be the only one which throws an exception

The "close" here could very well be a no-op here. As there really isn't any resource which needs closing. It's implemented this way only to notify the sending side that the connector is shutting down and thus the signal would not be processed.

However, thinking about it there might be a bit of an issue when close is draining the queue and read is called, so I will add the check ;)

debezium-core/src/main/java/io/debezium/pipeline/signal/channels/process/InProcessSignalChannel.java
```diff
@@ -0,0 +1,78 @@
...
+        }
+        return Stream.ofNullable(signals.poll()).toList();
+    }
+
+    @Override
+    public void close() {
+        if (open.compareAndSet(true, false)) {
+            drain();
+        }
+    }
```
jcechace, line 63: @vjuranek  @jpechane  This is not sufficient and in theory `close()` can start draining  after concurrent call to `read` already passed the `open.get()` check. So a lock would be in order to properly prevent this

However again -- I don't think there is a problem if that happens as such order of operation is equal to eider calling `read()` and then `close()` serially or the other way around -- the connector code  operating the channel must be able to handle any of these orders and thus locking here is pointless IMHO.

debezium-embedded/src/main/java/io/debezium/engine/source/EngineSourceTask.java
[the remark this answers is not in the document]
jcechace: Any taks that will not be from Debezium connector will not inherit from BaseSourceTask (as that is OUR class).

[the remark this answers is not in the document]
vjuranek: Do you envision Debezium server will be used one day for something else then running Debezium tasks? As I wrote, I have no problem with it, we should have ambitions!

debezium-core/src/main/java/io/debezium/pipeline/signal/channels/process/InProcessSignalChannel.java
vjuranek: Well, if all our signaler do so (which I checked and it's correct), let's keep it this way. That being said, I still consider reading from a closed resource as an error which shouldn't happen and throw an exception in such case.

debezium-core/src/main/java/io/debezium/pipeline/signal/channels/process/InProcessSignalChannel.java
vjuranek: I'd at least add comment here with the reasoning why locking doesn't matter.

debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEngineSignaler.java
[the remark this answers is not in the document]
jpechane: @jcechace I'd keep it as is now but I think the caching proposal deserves its own follow-up Jira

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
```diff
@@ -176,6 +176,32 @@ default boolean supportsTombstoneEvents() {
         }
     }
 
+    /**
+     * Signaler defines the contract for sending signals to connector tasks.
+     * @param <T> signal type
+     */
+    interface Signaler<T> {
```
jpechane, line 183: Please mark as `@Incubating`

jpechane (approved):
@jcechace LGTM! I left two comments. When addressed and CI passes feel free to merge yourself.

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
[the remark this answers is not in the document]
jcechace: The `String className` is imho not a good approach



```
<E extends DebeziumEngine<?>> void init(E engine, Class<S> signlaChannel);
```

Where does the S come from? The problem here is that this would require exposing the SignalChannelReader interface to public API... Actualy the SignalChannelWritter as that's the important thing here.

debezium-embedded/src/main/java/io/debezium/engine/source/EngineSourceTask.java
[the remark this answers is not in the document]
jcechace: On several occasions it was publicaly stated that it does. And it was used to do that in the past by some people. I'm not saying it should be support, I just didn't want to break it until we decide to do so.

debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEngineSignaler.java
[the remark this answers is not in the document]
jcechace: @jcechace  added the caching, less work than Jira :D

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
jcechace: Done

jpechane:
@jcechace Applied, thanks

PR #5761, opened 2024-08-12
DBZ-8103 Edit property description

roldanbob:
[DBZ-8103](https://issues.redhat.com/browse/DBZ-8103)

Edit property description


Naros (approved):
Thanks @roldanbob

PR #5764, opened 2024-08-13
DBZ-8138 Re-add check if assembly profile is active as JUnit annotation

rk3rn3r:
closes https://issues.redhat.com/browse/DBZ-8138


rk3rn3r:
@obabec this would not exclude the artifacts downstream, like discussed in the team call, but it should stop execution of the tests when the assembly profile is not set. The check was removed from the original code and is now re-added as annotation. I hope it fixes your issues!

rk3rn3r:
Test failures are unrelated.

jpechane:
@rk3rn3r Applied, thanks!

PR #5765, opened 2024-08-13
DBZ-7838 Support for MySQL 8.4

jpechane:
https://issues.redhat.com/browse/DBZ-7838

PR #5769, opened 2024-08-14
DBZ-8030 Add support for MySQL 9

jpechane:
https://issues.redhat.com/browse/DBZ-8030

Built on top of https://github.com/debezium/debezium/pull/5765

### DIFF
--- a/.github/workflows/debezium-workflow-pr.yml
+++ b/.github/workflows/debezium-workflow-pr.yml
@@ -308,7 +308,7 @@ jobs:
       # Runs each combination concurrently
       matrix:
         profile: [ "mysql-ci", "mysql-ci-gtids", "mysql-ci-percona", "mysql-ci-ssl" ]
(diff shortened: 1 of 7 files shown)


.github/workflows/debezium-workflow-pr.yml
```diff
@@ -299,7 +299,7 @@ jobs:
       # Runs each combination concurrently
       matrix:
         profile: [ "mysql-ci", "mysql-ci-gtids", "mysql-ci-percona", "mysql-ci-ssl" ]
-        version-mysql-server: [ "8.0", "8.4" ]
```
Naros, line 302: Would it be sufficient, at least on the upstream CI, to limit this to only 8.0 and 9.0?

jpechane: Well, given the scope of differences I'd prefer to have here both 8.0 and 8.4.

PR #5771, opened 2024-08-14
[ci] Fix systemtests CRB creation and bump Java for testsuite

obabec:


vjuranek (approved):
I'm not sure I fully understand the changes, but it tests are passing, so approving:-)

PR #5778, opened 2024-08-16
DBZ-8151 Allow `record_value` with no `primary.key.fields`

Naros:
https://issues.redhat.com/browse/DBZ-8151


jpechane:
@Naros Applied, thanks

PR #5779, opened 2024-08-18
DBZ-8156 Add an option for `publication.autocreate.mode` to create a publication with no tables

rkudryashov:
https://issues.redhat.com/browse/DBZ-8156


jpechane:
@rkudryashov Applied, thanks! Could you please send a docs PR update?

PR #5785, opened 2024-08-21
DBZ-8160 Update engine's signal API

jcechace:
https://issues.redhat.com/browse/DBZ-8160

THe original approach proved to be poorly usable as there was no way to ensure (or know) which implementation of signaler to use. This PR does the following

- Add Signal record to DebeziumEngine API
- Simplifies Signaler interface so that it always accepts Signal
- Moves the responsibility of creating Signaler to the engine implementations
- Exposes Signaler via DebeziumEngine API

this allows use as following 

```java
engine.getSignaler().signal(mySignal)
```

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEngineSignaler.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEngineSignaler.java
@@ -6,6 +6,7 @@
 package io.debezium.embedded.async;
 
 import java.util.List;
+import java.util.Objects;
 import java.util.Optional;
 
 import io.debezium.DebeziumException;
@@ -14,20 +15,16 @@ import io.debezium.engine.source.EngineSourceTask;
 import io.debezium.pipeline.signal.SignalRecord;
 import io.debezium.pipeline.signal.channels.process.InProcessSignalChannel;
 
-public class AsyncEngineSignaler implements DebeziumEngine.Signaler<SignalRecord> {
+public class AsyncEngineSignaler implements DebeziumEngine.Signaler {
     private AsyncEmbeddedEngine<?> engine;
     private List<InProcessSignalChannel> channels;
 
-    @Override
-    public <E extends DebeziumEngine<?>> void init(E engine) {
-        if (!(engine instanceof AsyncEmbeddedEngine)) {
-            throw new DebeziumException("AsyncEngineSignaler can only be used with AsyncEmbeddedEngine");
-        }
-        this.engine = (AsyncEmbeddedEngine<?>) engine;
+    public AsyncEngineSignaler(AsyncEmbeddedEngine<?> engine) {
+        this.engine = Objects.requireNonNull(engine);
     }
-
+    
     @Override
-    public void signal(SignalRecord signal) {
+    public void signal(DebeziumEngine.Signal signal) {
         if (channels == null) {
             channels = engine.tasks().stream()
                     .map(EngineSourceTask::debeziumConnectTask)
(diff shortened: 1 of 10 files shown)


debezium-embedded/src/test/java/io/debezium/embedded/async/AbstractAsyncEngineConnectorTest.java
```diff
@@ -17,18 +16,17 @@
...
+        return new AsyncEmbeddedEngine.AsyncEngineBuilder();
     }
 
     @Override
     protected TestingDebeziumEngine createEngine(DebeziumEngine.Builder builder) {
         return new TestingAsyncEmbeddedEngine((AsyncEmbeddedEngine) builder.build());
     }
+
+    protected DebeziumEngine.Signaler getSignaler() {
+        return (DebeziumEngine.Signaler) engine.getSignaler().get();
```
jcechace, line 30: This needs to be here as the entirety of Test infra uses generics poorly (the whole things shines with yellow underlines like a Christmas tree)  

I will try to fix it tomorrow but it might be a long way down the rabbit hole. It alls stems from the fact that the following 


```java
static class Foo<T> {
    Optional<String> getName() {
        return Optional.of("John Doe");
    }
}

public static void main(String[] args) {
    Foo foo = new Foo();
    String name = foo.getName().get();
    System.out.println(name);
}
```

will not compile. As ALL generic information is erased on `foo` thus `name` is of type `Object`. 

Our entire suite uses raw types instead of `<?>` where it doesn't care about the generic type.

debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEngineSignaler.java
```diff
@@ -6,36 +6,21 @@
...
 import io.debezium.engine.DebeziumEngine;
-import io.debezium.engine.source.EngineSourceTask;
 import io.debezium.pipeline.signal.SignalRecord;
-import io.debezium.pipeline.signal.channels.process.InProcessSignalChannel;
+import io.debezium.pipeline.signal.channels.process.SignalChannelWriter;
 
-public class AsyncEngineSignaler implements DebeziumEngine.Signaler<SignalRecord> {
-    private AsyncEmbeddedEngine<?> engine;
-    private List<InProcessSignalChannel> channels;
+public class AsyncEngineSignaler implements DebeziumEngine.Signaler {
```
jcechace, line 14: this could be now named `EmbeddedSignaler` and moved to the parent package as the implementation itself is not tied to async engine anymore

debezium-embedded/src/test/java/io/debezium/embedded/async/AbstractAsyncEngineConnectorTest.java
jcechace: Filed https://issues.redhat.com/browse/DBZ-8166

debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEngineSignaler.java
jpechane: Do you plan to do that change?

jcechace: Renamed

jcechace:
@jpechane  should be good to merge now. I made few finishing touches -- mainly simplifying the access to the writer instance (the more complex version can be reintroduced if needed, though I don't think it ever will).

jpechane:
@jcechace Applied, thanks!

PR #5795, opened 2024-08-24
DBZ-8169 Upgrade Adobe s3mock to 3.10.0

sullis:
https://issues.redhat.com/browse/DBZ-8169

upgrade to latest version of s3mock

https://github.com/adobe/S3Mock/releases


jpechane:
@sullis Applied, thanks

PR #5800, opened 2024-08-28
DBZ-8170 Error caused by database name containing regular keywords

jw-itq:
https://issues.redhat.com/browse/DBZ-8170

Error caused by database name containing regular keywords, such as database name: test$user

### DIFF
--- a/debezium-core/src/main/java/io/debezium/function/Predicates.java
+++ b/debezium-core/src/main/java/io/debezium/function/Predicates.java
@@ -15,6 +15,7 @@ import java.util.function.BiFunction;
 import java.util.function.BiPredicate;
 import java.util.function.Function;
 import java.util.function.Predicate;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
 import java.util.stream.Collectors;
@@ -227,12 +228,12 @@ public class Predicates {
      * @throws PatternSyntaxException if the string includes an invalid regular expression
      */
     public static <T> Predicate<T> includes(String regexPatterns, Function<T, String> conversion) {
-        Set<Pattern> patterns = Strings.setOfRegex(regexPatterns, Pattern.CASE_INSENSITIVE);
+        Set<Pattern> patterns = Strings.setOfRegex(regexPatterns != null ? Matcher.quoteReplacement(regexPatterns) : null, Pattern.CASE_INSENSITIVE);
         return includedInPatterns(patterns, conversion);
     }
 
     public static <T, U> BiPredicate<T, U> includes(String regexPatterns, BiFunction<T, U, String> conversion) {
-        Set<Pattern> patterns = Strings.setOfRegex(regexPatterns, Pattern.CASE_INSENSITIVE);
+        Set<Pattern> patterns = Strings.setOfRegex(regexPatterns != null ? Matcher.quoteReplacement(regexPatterns) : null, Pattern.CASE_INSENSITIVE);
         return includedInPatterns(patterns, conversion);
     }
 
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -634,3 +634,4 @@ Gaurav Miglani
 张展业
 Ashish Binu
 Mohamed El Shaer
+Shiwanming
--- a/debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
+++ b/debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
@@ -44,6 +44,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
+import java.util.regex.Matcher;
 import java.util.stream.Collectors;
 
 import org.apache.kafka.connect.data.Struct;
@@ -314,7 +315,7 @@ public class JdbcConnection implements AutoCloseable {
 
                 if (value != null) {
                     // And replace the variable ...
-                    url = url.replaceAll("\\$\\{" + name + "\\}", value);
+                    url = url.replaceAll("\\$\\{" + name + "\\}", Matcher.quoteReplacement(value));
                 }
             }
         }
--- a/jenkins-jobs/scripts/config/Aliases.txt
+++ b/jenkins-jobs/scripts/config/Aliases.txt
@@ -281,3 +281,4 @@ wltmlx,Lukas Langegger
 GitHubSergei,Sergey Kazakov
 shaer,Mohamed El Shaer
 SylvainMarty,Sylvain Marty
+jw-itq,Shiwanming


Naros:
Hi @jw-itq would it be possible to add a test not only for the use case you've described with `test$user`, but also checking what this change will do should the configuration supply the regular expression contain escaped syntax values, such as `test\$user`, does this cause such a configuration to no longer work or be interpreted correctly?

jw-itq:
> Hi @jw-itq would it be possible to add a test not only for the use case you've described with `test$user`, but also checking what this change will do should the configuration supply the regular expression contain escaped syntax values, such as `test\$user`, does this cause such a configuration to no longer work or be interpreted incorrectly?

thanks, I will attempt to add a test case. Yes, this could potentially cause test$user not to be interpreted correctly. However, on the JDBC URL, you cannot use \$ but you can use \\$. If the Matcher.quoteReplacement is used, test$user will still be effective.

Naros:
> > Hi @jw-itq would it be possible to add a test not only for the use case you've described with `test$user`, but also checking what this change will do should the configuration supply the regular expression contain escaped syntax values, such as `test\$user`, does this cause such a configuration to no longer work or be interpreted incorrectly?
> 
> thanks, I will attempt to add a test case. Yes, this could potentially cause test$user not to be interpreted correctly. However, on the JDBC URL, you cannot use $ but you can use \$. If the Matcher.quoteReplacement is used, test$user will still be effective.

But you're also adjusting the `Predicates` class which is used for many other configuration options such as the include and exclude list configurations that we've always communicated that special regex characters should be escaped if they're part of the database, schema, or table name.  So we need to consider all those variants with this change imo.

jw-itq:
> > > Hi @jw-itq would it be possible to add a test not only for the use case you've described with `test$user`, but also checking what this change will do should the configuration supply the regular expression contain escaped syntax values, such as `test\$user`, does this cause such a configuration to no longer work or be interpreted incorrectly?
> > 
> > 
> > thanks, I will attempt to add a test case. Yes, this could potentially cause test$user not to be interpreted correctly. However, on the JDBC URL, you cannot use $ but you can use $. If the Matcher.quoteReplacement is used, test$user will still be effective.
> 
> But you're also adjusting the `Predicates` class which is used for many other configuration options such as the include and exclude list configurations that we've always communicated that special regex characters should be escaped if they're part of the database, schema, or table name. So we need to consider all those variants with this change imo.

thanks, the Predicate indeed requires more thought, and I will reconsider the changes regarding the Predicate. Or do you have any better suggestions?

Ritabrata1080 (commented):
Please squash the commits to have a linear history.

Naros:

Hi @jw-itq, please sign off all commits with:

```
git commit -s
```

jw-itq:
> ❌ Developer Certificate of Origin (DCO) check failed.
> 
> Hi @jw-itq, please sign off all commits with:
> 
> ```
> git commit -s
> ```
> 
> If pull request commits are not signed off, the pull request cannot be merged. For more information about why this is required, please see [our blog](https://debezium.io/blog/2025/12/12/contribution-requirements-changing/) about contribution requirement changes.

ok, thanks. I'll handle it right away.

jw-itq:
> Please squash the commits to have a linear history.

ok!

jw-itq:
@Naros @Ritabrata1080 hi,commits squashed and DCO sign-off added.

debezium-core/src/main/java/io/debezium/function/Predicates.java
```diff
@@ -227,12 +228,12 @@ public static Predicate<String> excludes(String regexPatterns, int regexFlags) {
...
      */
     public static <T> Predicate<T> includes(String regexPatterns, Function<T, String> conversion) {
-        Set<Pattern> patterns = Strings.setOfRegex(regexPatterns, Pattern.CASE_INSENSITIVE);
+        Set<Pattern> patterns = Strings.setOfRegex(regexPatterns != null ? Matcher.quoteReplacement(regexPatterns) : null, Pattern.CASE_INSENSITIVE);
         return includedInPatterns(patterns, conversion);
     }
 
     public static <T, U> BiPredicate<T, U> includes(String regexPatterns, BiFunction<T, U, String> conversion) {
-        Set<Pattern> patterns = Strings.setOfRegex(regexPatterns, Pattern.CASE_INSENSITIVE);
+        Set<Pattern> patterns = Strings.setOfRegex(regexPatterns != null ? Matcher.quoteReplacement(regexPatterns) : null, Pattern.CASE_INSENSITIVE);
```
Ritabrata1080: What happens if patterns set tries to set regex of null value ? Please check for potential NPE crash across patch.

debezium-core/src/main/java/io/debezium/function/Predicates.java
```diff
@@ -227,12 +228,12 @@ public static Predicate<String> excludes(String regexPatterns, int regexFlags) {
      * @throws PatternSyntaxException if the string includes an invalid regular expression
      */
     public static <T> Predicate<T> includes(String regexPatterns, Function<T, String> conversion) {
-        Set<Pattern> patterns = Strings.setOfRegex(regexPatterns, Pattern.CASE_INSENSITIVE);
+        Set<Pattern> patterns = Strings.setOfRegex(regexPatterns != null ? Matcher.quoteReplacement(regexPatterns) : null, Pattern.CASE_INSENSITIVE);
```
Naros: We've always expected that if a provided include/exclude list configuration property contains regex characters that may not be interpreted as literal, they should be manually escaped in the configuration. This change here would specifically create breaking behavior in environments that already pre-escape special characters.

I'm personally not convinced we should change this behavior, as using `$` and, to a lesser degree, `?` in tablenames is quite common among certain database vendors. I am concerned about the potential impact on a longstanding expectation when handling these strings.

When it comes to the JDBC URL, we can certainly be significantly more lenient there. 
@jpechane any thoughts?

jpechane: I agree, we should not break backward compatibility in this regard.

jw-itq:
@Naros @Ritabrata1080 @jpechane  Thank you for your review. I will revert the changes made to Predicates.java

jpechane:
@jw-itq LGTM. Could you please rebase the PR on the latest main and add a test case if possible to make sure the results are expected?

jw-itq:
> @jw-itq LGTM. Could you please rebase the PR on the latest main and add a test case if possible to make sure the results are expected?

@jpechane ok.I see that the latest version has been updated. Do we not need to proceed with the current PR?

jpechane:
@jw-itq Please take a lok at the PR changes. I belive it is broken now.

jw-itq:
> @jw-itq Please take a lok at the PR changes. I belive it is broken now.

@jpechane hi. the JdbcConnection.java fix using Matcher.quoteReplacement() is already present in the main branch (commit fe5758e98d or earlier). After rebasing, my PR now only contains contributor information updates. Should I close this PR since the actual fix is already merged, or would you like me to keep it for the contributor credits?

jpechane:
@jw-itq Please keep the credits, just it is necessary to rebase on the latest main and remove the extra lines ike `<<<<<`

jw-itq:
> @jw-itq Please keep the credits, just it is necessary to rebase on the latest main and remove the extra lines ike `<<<<<`

@jpechane Thanks! Done.

PR #5801, opened 2024-08-28
DBZ-8039 Add support for `max_string_size` using extended strings

Naros:
https://issues.redhat.com/browse/DBZ-8039


Naros:
@obabec @vjuranek do either of you know why the Oracle 23 TF image is failing with credential issues?  This seems to happen from time to time but it isn't consistent and makes little to no sense.

Naros:
The Oracle 21 XStream test failure is due to wait times in the test and the CI being saturated.

jpechane (approved):
@Naros LGTM thanks! One question though. I see that `OracleExtendedStringIT` does not run on CI. Is there a plan to enable it for one of the Oracel versions?

Naros:
> @Naros LGTM thanks! One question though. I see that `OracleExtendedStringIT` does not run on CI. Is there a plan to enable it for one of the Oracel versions?

Yes, but it requires a unique build of Oracle that cannot be automated.  I'll add this to next week's team meeting to discuss our long-term testing strategy.  I have some concerns about the continuation of just building one-off images for the sole purpose of testing these types of unique configurations, as these have an overall impact on our slave sizes.

jpechane:
@Naros I wonder if we can have a single image (not for all Oracle versions) with all non-default settings and run the testsuite against it.

Naros:
@jpechane from your PoV, is this okay to merge and add a follow-up for the testing?

PR #5806, opened 2024-08-29
DBZ-8142 Fix conversion of `BigInteger` timestamp values

Naros:
https://issues.redhat.com/browse/DBZ-8142

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/olr/OpenLogReplicatorValueConverter.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/olr/OpenLogReplicatorValueConverter.java
@@ -5,6 +5,7 @@
  */
 package io.debezium.connector.oracle.olr;
 
+import java.math.BigInteger;
 import java.sql.SQLException;
 import java.time.Instant;
 import java.time.OffsetDateTime;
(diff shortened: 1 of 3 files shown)


Naros:
The XStream ORA-01466 error, while fixed in DBZ-7072, isn't used during CI tests.
Enabled feature for CI in https://github.com/debezium/debezium/pull/5811

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/olr/OpenLogReplicatorValueConverter.java
```diff
@@ -113,7 +114,7 @@ protected Object convertTimestampWithZone(Column column, Field fieldDefn, Object
 
     @Override
     protected Object convertTimestampWithLocalZone(Column column, Field fieldDefn, Object value) {
-        if (value instanceof Long) {
+        if (value instanceof Number) {
```
jpechane, line 117: This does not look safe. You check for `Number` but then you typecaset to `Long`

jpechane (changes requested):
In general it looks good, but there is one instance which I consider to be an unsafe typecasting.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/olr/OpenLogReplicatorValueConverter.java
Naros: Thanks, I'll take a look next week when back.

jpechane (approved):
@Naros LGTM, thanks. Feel free to merge it when CI passes.

PR #5807, opened 2024-08-29
DBZ-8071 Improve hybrid mining performance with object-id-to-table-id cache

Naros:
https://issues.redhat.com/browse/DBZ-8071

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
@@ -112,6 +112,7 @@ public abstract class AbstractLogMinerEventProcessor<T extends Transaction> impl
     private Instant lastProcessedScnChangeTime = null;
     private Scn lastProcessedScn = Scn.NULL;
     private boolean sequenceUnavailable = false;
+    private HashMap<Long, TableId> objectIdToTableIdCache = new HashMap<>();
 
     private final Set<String> abandonedTransactionsCache = new HashSet<>();
     private final InMemoryPendingTransactionsCache inMemoryPendingTransactionsCache = new InMemoryPendingTransactionsCache();
@@ -268,6 +269,9 @@ public abstract class AbstractLogMinerEventProcessor<T extends Transaction> impl
     @Override
     public Scn process(Scn startScn, Scn endScn) throws SQLException, InterruptedException {
         counters.reset();
+
+        // Recreate the cache on each iteration
+        objectIdToTableIdCache = new HashMap<>();
 
         try (PreparedStatement statement = createQueryStatement()) {
             LOGGER.debug("Fetching results for SCN [{}, {}]", startScn, endScn);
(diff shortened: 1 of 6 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
```diff
@@ -112,6 +112,7 @@ public abstract class AbstractLogMinerEventProcessor<T extends Transaction> impl
     private Instant lastProcessedScnChangeTime = null;
     private Scn lastProcessedScn = Scn.NULL;
     private boolean sequenceUnavailable = false;
+    private HashMap<Long, TableId> objectIdToTableIdCache = new HashMap<>();
```
jpechane, line 115:
```suggestion
    private Map<Long, TableId> objectIdToTableIdCache = new HashMap<>();
```

How large the cache can be? Do we want to use an unbounded one?

jpechane (commented):
@Naros LGTM, thanks. I left one comment on the code and also question related to cache size/bounds.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
jeremy-l-ford: I'm also curious if the cache needs to be reset for every call to processSCNs.  Can it be kept around for the life time of the processor.

Naros: @jpechane, there are some corner cases where we could see the cache be anywhere between 1x and 2x the number of captured tables; however, within a single mining step, that would be extremely rare. This cache only applies when the Hybrid strategy is enabled; it's unused otherwise.

@jeremy-l-ford, that's a good question. My concern with leaving it around indefinitely is that when a `DROP TABLE` is executed, the entry in the cache won't be evicted. Now, Oracle doesn't reuse OBJECT_IDs, so it isn't like the entry itself has a logical issue at runtime, just that the entry would consume some memory unnecessarily.

Perhaps falling back on @jpechane's suggestion with a bounded cache would be the best of both worlds. We would be able to guard against corner cases where the cache could consume valuable memory unnecessarily and have a way to evict old entries.

Are you both in agreement?

jeremy-l-ford: I think the bounded cache would work.  Maybe even making its size configurable/unbounded?  

Also curious what you would think about this information being owned by the OracleDatabaseSchema object.  It could then respond to schema events and drop a table/remove from the cache.

Naros: For now, the cache will be capped at a size of `10` but can be controlled using `internal.object.id.cache.size`, which must be greater than `0`. I also moved the code to the `OracleDatabaseSchema` class and cleaned-up the code references, including some small refactors around iteration & attribute names.

jpechane (approved):
@Naros LGTM thanks! I'd just consider increasing the default value to sometging slightly larger like 256.

Please feel free to merge when CI passes.

Naros:
@jpechane will do, thanks!

jeremy-l-ford:
@Naros - did you want to remove entries from the cache upon schema change events?  Or are you thinking LRU will just handle that given time?

Naros:
> @Naros - did you want to remove entries from the cache upon schema change events? Or are you thinking LRU will just handle that given time?

I think its reasonable to just let the cache age those out.  To remove them on schema changes, we'd have to iterate the cache value set, and don't think that's worth the overhead.  Would you concur @jeremy-l-ford ?

jeremy-l-ford:
> I think its reasonable to just let the cache age those out. To remove them on schema changes, we'd have to iterate the cache value set, and don't think that's worth the overhead or maintain an inverse cache of TableId to ObjectId. Would you concur @jeremy-l-ford ?

I concur.

jeremy-l-ford (approved):

PR #5813, opened 2024-08-29
DBZ-8185 Add `transaction` field to change event of a logical decoding message

rkudryashov:
https://issues.redhat.com/browse/DBZ-8185

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/LogicalDecodingMessageMonitor.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/LogicalDecodingMessageMonitor.java
@@ -10,6 +10,7 @@ import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
 import java.util.Base64;
 import java.util.Base64.Encoder;
+import java.util.List;
 
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.Struct;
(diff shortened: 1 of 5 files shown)


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/LogicalDecodingMessageMonitor.java
```diff
@@ -99,6 +101,9 @@ public void logicalDecodingMessageEvent(Partition partition, OffsetContext offse
         value.put(DEBEZIUM_LOGICAL_DECODING_MESSAGE_KEY, logicalMsgStruct);
         value.put(Envelope.FieldName.SOURCE, offsetContext.getSourceInfo());
 
+        TableId tableId = PostgresSchema.parse("fake");
```
rkudryashov, line 104: perhaps this should be changed somehow

jpechane: Definitely not fake :-). Ideally (if possible) a new implementation of `io.debezium.spi.schema.DataCollectionId` will be defined in this class and it will either return empty string or strin like 'LOGICAL_MESSAGE' whatver will be feasible.

rkudryashov:
Build errors seem unrelated

jpechane:
@rkudryashov Applied, thanks!

PR #5814, opened 2024-08-29
DBZ-8188 Use new connection on attach retries

Naros:
https://issues.redhat.com/browse/DBZ-8188


Naros:
Oracle 23 test failures are unrelated.

PR #5815, opened 2024-08-29
DBZ-8189 Return new byte array/buffer when truncating

twthorn:
### Description

[DBZ-8189](https://issues.redhat.com/browse/DBZ-8189)
For some converters they properly only look at the byte buffer slice. However for Kafka Connect's [default json converter](https://github.com/apache/kafka/blob/3.8.0/connect/json/src/main/java/org/apache/kafka/connect/json/JsonConverter.java#L631) it gets the underlying array which undoes the truncate logic. So we fix that here by returning a new byte buffer & new byte array that way no subsequent converter can access the original untruncated byte array.

### Verification

Add unit test to reproduce the error with the call to get the underlying array that the slice view is referencing.

We also had a case where a connector we were running ran into this issue & RecordTooLargeException and deploying this change fixed it.

### DIFF
--- a/debezium-core/src/main/java/io/debezium/relational/mapping/TruncateColumn.java
+++ b/debezium-core/src/main/java/io/debezium/relational/mapping/TruncateColumn.java
@@ -6,6 +6,7 @@
 package io.debezium.relational.mapping;
 
 import java.nio.ByteBuffer;
+import java.util.Arrays;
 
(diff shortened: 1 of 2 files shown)


twthorn:
@jpechane this is ready for review when you get the chance. Thanks!

jpechane:
@twthorn Nice, thanks!

PR #5816, opened 2024-08-30
DBZ-8187 Don't close transformations in EmbeddedEngine

vjuranek:
`EmbeddedEngine` can be stopped and started again. Closing the transfomation chain may cause the issue when engine is started again and already closed transformations are applied to stream of data.

Restore original behavior of `EmbeddedEngine` and don't close the transformation chain in this engine.

https://issues.redhat.com/browse/DBZ-8187


vjuranek:
DB2 failure is unrelated.

jpechane:
@vjuranek Applied, thanks

PR #5828, opened 2024-09-04
DBZ-7996 Move ChangeEventSink class to debezium-core and use the interface for the MongoDB sink connector

rk3rn3r:
relates to https://issues.redhat.com/browse/DBZ-7996


rk3rn3r:
@jpechane Sorry, I forgot to push my changes from last week and file that PR. I reverted temporary some future work that I already did and provide a small PR here to just move the ChangeEventSink interface and switch to that new interface.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/sink/MongoDbSinkConnectorTask.java
```diff
@@ -51,7 +51,7 @@ public void start(final Map<String, String> props) {
         try {
             this.connectionContext = new MongoDbConnectionContext(config);
```
Naros, line 52: Hey @rk3rn3r, unrelated to these changes but more for my own edification, is there a long-term goal for the `MongoDbConnectionContext` to be a member variable rather than a local variable?  It doesn't seem to be used anywhere else, so I wasn't sure if this could just be allocated and  used here locally?

PR #5840, opened 2024-09-08
DBZ-8217 fix wrong log message

zalmane:


jpechane:
@zalmane Applied, thanks!

PR #5841, opened 2024-09-09
DBZ-8219 Fix JAXB xml parsing of Ehcache configuration

Naros:
https://issues.redhat.com/browse/DBZ-8219

### DIFF
--- a/debezium-connector-oracle/pom.xml
+++ b/debezium-connector-oracle/pom.xml
@@ -92,15 +92,11 @@
             <artifactId>ehcache</artifactId>
             <version>${version.ehcache}</version>
         </dependency>
+        <!-- Ehcache JAXB implementation dependency -->
         <dependency>
-            <groupId>com.sun.xml.bind</groupId>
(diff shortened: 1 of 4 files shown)


Naros:
Test failures are unrelated.

jpechane (approved):
@Naros LGTM! In this case I'd recommend to look into the generated plugin and Debezium Server as well to make sure dependencies are OK and test the connector in thos deployments as well.

Naros:
Hi @jpechane so I've tried to align this in Debezium Server, but despite my best efforts, the `jaxb-runtime` is always pulled in as 4.0.4 rather than 2.3.1, despite what's in this PR.  Any idea what I am missing?  I've tried exclusions in quarkus and others with no success.

jpechane:
@Naros Hi, I pushed one more commit and created https://github.com/debezium/debezium-server/pull/122

Please give it a test ride to doble check that now everything is aligned and the connector and cach works. Thanks.

PR #5845, opened 2024-09-10
DBZ-8224 allow field name without topic qualifier

msillence:
make filtering match documentation


jpechane:
@Naros @vjuranek Any comments? If not please feel free to merge.

Naros (approved):
LGTM from my PoV.

vjuranek (changes requested):
The change LGTM, but when you are in it, could you please add `.strip()` to all `parts`? As you pointed out in the comment in Jira, topics with whitespace before comma are not matched, which I'd say users don't expect. Thanks!

Naros:
As I made the changes suggested by @vjuranek, to avoid creating a new PR since this one is locked, I merged these changes directly into main and backported the change to the 2.7 branch.

PR #5868, opened 2024-09-17
DBZ-8241 Use pre-defined builder factory when none is specified 

vjuranek:
Order of loading classes via SPI is not guaranteed. As DebeziumEngine.create() method is defined in debezium-api module which we don't want to depends on embedded module, possibly the only how to prevend loading async engine builder via SPI is to remove it from SPI.

https://issues.redhat.com/browse/DBZ-8241


jpechane:
@vjuranek Applied, thanks.

PR #5869, opened 2024-09-17
[ci] Use the PR base branch also for other dependecies repos

vjuranek:
Without specifying the branch default branch is checked out and in case of backport PRs (PR to 2.7 branch), build of DBZ server dependecies (an others) fails as it use 3.0.0 branch and requires Java 21.


Naros:
Great catch on the PR workflow @vjuranek :+1:

PR #5872, opened 2024-09-18
DBZ-8250 Move engine running check to avoid premature exit

Naros:
https://issues.redhat.com/browse/DBZ-8250

@jpechane as requested, this should continue to satisfy the original logic but avoid the premature loop exits when data still needs to be consumed that hasn't yet been drained from the queue. I tested this with the Db2ConnectorIT method you referred me to and it now passes on the first execution :tada:


jpechane:
@Naros LGTM, thanks! I've also increased record wait time as it seems to improve stability of other tests too.

PR #5883, opened 2024-09-23
DBZ-8260 Make Oracle NLS library provided

jpechane:
https://issues.redhat.com/browse/DBZ-8260

### DIFF
--- a/LICENSE-3rd-PARTIES.txt
+++ b/LICENSE-3rd-PARTIES.txt
@@ -3111,3 +3111,124 @@ necessary.  Here is a sample; alter the names:
   Ty Coon, President of Vice
 
 That's all there is to it!
+
+
+=====
+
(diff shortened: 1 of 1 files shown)


jpechane:
@Naros I prepared a PR for this issue. Could you please double check if NLS is redistributable or not? Thanks

Naros:
Hi, @jpechane, this is just like the JDBC driver, so we shouldn't distribute it downstream.

jpechane:
And in upstream?

Naros:
So Oracle's FUTC does not allow us to redistribute it either, so it needs to be _provided_ for all contexts it seems.

jpechane:
@Naros So now to clarify, FUTC does not allow us to redistribute only NLS library or the JDBC driver too?

Naros:
As per our offline discussion, we can redistribute both upstream.

PR #5885, opened 2024-09-23
DBZ-8263 Common binlog code does not depend on MariaDB

jpechane:
https://issues.redhat.com/browse/DBZ-8263


Naros (approved):
Great catch @jpechane :+1:

PR #5886, opened 2024-09-23
DBZ-8267 Fix test failures

Naros:
https://issues.redhat.com/browse/DBZ-8267


jpechane:
@Naros Applied, thanks

PR #5888, opened 2024-09-23
DBZ-1252 Allow configuring the postgres snapshot isolation level

cwholmes:
This change allows users to define the desired transaction isolation for the postgres snapshot.

This will allow for running the initial snapshot with a reader instance that doesn't support SERIALIZABLE.


jpechane:
@cwholmes Applied, thanks! Could you please send another PR with docs update?

PR #5890, opened 2024-09-24
DBZ-8244 Avoid blocking snapshot to remain in an inconsistent state in case of failure

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-8244

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
@@ -257,7 +257,7 @@ public class ChangeEventSourceCoordinator<P extends Partition, O extends OffsetC
 
                 }
                 catch (Exception e) {
-                    LOGGER.error("Error while executing requested blocking snapshot: {}", e.getMessage());
+                    LOGGER.warn("Error while executing requested blocking snapshot.", e);
                     resumeStreaming(partition);
                 }
             }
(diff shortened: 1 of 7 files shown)


debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
```diff
@@ -247,13 +247,18 @@ public void doBlockingSnapshot(P partition, OffsetContext offsetContext, Snapsho
...
+                        resumeStreaming(partition);
+                    }
 
-                if (running && snapshotResult.isCompletedOrSkipped()) {
-                    previousLogContext.set(taskContext.configureLoggingContext("streaming", partition));
-                    paused = false;
-                    context.resumeStreaming();
+                }
+                catch (Exception e) {
+                    LOGGER.error("Error while executing requested blocking snapshot: {}", e.getMessage());
```
jpechane, line 260: I'd prefer full stacktrace here but at `WARN` level.

debezium-embedded/src/test/java/io/debezium/pipeline/AbstractBlockingSnapshotTest.java
```diff
@@ -246,6 +247,37 @@ public void executeBlockingSnapshotWithEscapedCollectionName() throws Exception
 
     }
 
+    @Test
+    @FixFor("DBZ-8244")
+    public void anErrorDuringBlockingSnapshotShouldLeaveTheConnectorInAGoodState() throws Exception {
```
jpechane, line 252: Could we add a test case where connector is restarted right after the snapshot failure?

jpechane (commented):
@mfvitale LGTM, thanks! I left two minor comments. When addressed feel free to merge yourself.

debezium-embedded/src/test/java/io/debezium/pipeline/AbstractBlockingSnapshotTest.java
mfvitale: I suppose you are thinking to this https://issues.redhat.com/browse/DBZ-7903. So in that case I think it will try to wrongly do an initial snapshot. I'll fix that in a separate PR.

PR #5896, opened 2024-09-26
DBZ-8221 Correctly bind Geometry column values when null

Naros:
https://issues.redhat.com/browse/DBZ-8221


Naros:
Just an FYI @rk3rn3r regarding the binding issue we discussed today during and after the triaging call.  I was able to do this from within the binding itself to guarantee that for this use case, we always generate the same SQL regardless and instead forced binding 2 values when NULL rather than 1. This should be less impactful to your rework I hope.

Naros:
Once the tests all pass, I'll adjust core to use the `SchemaAndValueField` implementation with the special JDBC sink one and add the accessors to the immutable object.  I just wanted to get a quick validation within the JDBC code base before I had all connector tests run.

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/util/SinkRecordFactory.java
```diff
@@ -267,6 +268,45 @@ default SinkRecord createRecordWithSchemaValue(String topicName, byte key, List<
...
+                .topic(topicName)
+                .offset(offset)
+                .partition(0)
+                .keySchema(keySchema)
+                .recordSchema(recordSchema)
+                .sourceSchema(basicSourceSchema())
+                .source("ts_ms", (int) Instant.now().getEpochSecond());
+
+        for (SchemaAndValueField keyField : keyFields) {
+            builder.key(keyField.fieldName(), keyField.value());
```
rk3rn3r, line 299: instead of doing this here and exposing this new API in SchemaAndValueField and adding another traversal over key fields, couldn't `SchemaBuilder#field()` do that already with an additional argument `boolean isKey` to distinguish between key and value fields?

rk3rn3r:
Thx for the pointer @Naros. I saw the PR on a notification mail from GH yesterday evening after you created it and marked it to look into it today. I also thought if binding nulls to both fields would be a solution and confirmed now - it is! :))
That makes it even simpler regarding handling those new batches in a future Writer implementation.

I love it when a plan comes together

mfvitale (approved):

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/util/SinkRecordFactory.java
mfvitale: IMHO I am not a fan of methods with boolean arguments that behave differently based on it. So I'll maintain as is. 

Just for clearness I'll see a `keys` method that will get the list of key fields and do the loop. But it is just nitpick.

jpechane: I'd keep it as it is now and we might have a follow-up if needed.
@mfvitale You are right about the boolean argument yet in case of Kafka Connect it is pretty common that methods use `isKey` argument and change the behaviour based on that so at the end it would be consistent with current practice.

jpechane:
@Naros Applied, thanks!

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/util/SinkRecordFactory.java
rk3rn3r: I can relate those concerns, but I also find the performance gain is immense when you consider passing around millions and billions of messages over time. And you avoid iterating collections twice (or more).

PR #5899, opened 2024-09-27
DBZ-8278 Remove deprecated additional condition for incremental snapshot

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-8278


jpechane:
@mfvitale Applied, thanks

PR #5906, opened 2024-09-30
DBZ-8262 Align relational_table clauses to handle logminer rules

ani-sha:
https://issues.redhat.com/browse/DBZ-8262


jpechane:
@ani-sha Applied, thanks!

PR #5916, opened 2024-10-04
DBZ-7040 Fix Oracle DDL parsing to strip trailing new lines

Naros:
https://issues.redhat.com/browse/DBZ-7040


Naros:
The reselect test failures are fixed in #5920

PR #5917, opened 2024-10-04
DBZ-8292 Make ORA-01001 invalid cursor errors to be retried

Naros:
https://issues.redhat.com/browse/DBZ-8292

Leaving this in draft until the questions on the Jira are answered, in case we need to add any additional fixes.


jpechane:
@Naros I wonder if `io.debezium.config.CommonConnectorConfig.CUSTOM_RETRIABLE_EXCEPTION` could be used by the user to test the solution without waiting for the patch.

Naros:
@jpechane that's a great idea. I'll add a note to the Jira.

Naros:
Reselect column failures are fixed in #5920.

PR #5919, opened 2024-10-05
DBZ-8303 Fix race condition in stop snapshot

kgalieva:
Issue: https://issues.redhat.com/browse/DBZ-8303
[Conversation](https://groups.google.com/g/debezium/c/4Ir18BzYTiE/m/PeEzHa_cAQAJ)

### Problem: 

The `AbstractIncrementalSnapshotContext` is accessed by two threads—the main binlog thread and the `SignalProcessor` thread. A race condition occurs when the `SignalProcessor` thread calls `stopSnapshot()` while the main thread is executing `readChunk()`. This can lead to inconsistencies, causing exceptions during snapshot processing.

### Solution:

- Instead of immediately modifying the snapshot state in `stopSnapshot()`, a queue (`dataCollectionsToStop`, a `LinkedBlockingQueue<String>`) is used to manage the list of collections to stop. This queue can contain collection names or regex patterns. ".*" is used to stop all collections' snapshots.
- The method `stopSnapshot()` has been renamed to `requestSnapshotStop()` to reflect that it only queues the collections for stopping, rather than stopping the snapshot immediately.
- The `readChunk()` method has been updated to check the queue (`dataCollectionsToStop`) for any collections that should be stopped using a helper method `checkAndProcessStopFlag()`
- The method `expandAndDedupeDataCollectionIds()` for `MongoDB` has been updated to handle cases where all snapshots need to be stopped. It ensures that snapshots are stopped properly when ".*" is passed

### Edge Case:

A potential edge case exists where the stop snapshot signal might be lost if Debezium is stopped between receiving the signal and executing `readChunk()` plus one event after `readChunk()`. (One event after `readChunk()` is necessary to trigger an offset commit with the updated list of `incremental_snapshot_collections`)

### DIFF
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogReadOnlyIncrementalSnapshotChangeEventSource.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogReadOnlyIncrementalSnapshotChangeEventSource.java
@@ -146,8 +146,8 @@ public abstract class BinlogReadOnlyIncrementalSnapshotChangeEventSource<P exten
     }
 
     @Override
-    public void stopSnapshot(P partition, OffsetContext offsetContext, Map<String, Object> additionalData, List<String> dataCollectionIds) {
-        super.stopSnapshot(partition, offsetContext, additionalData, dataCollectionIds);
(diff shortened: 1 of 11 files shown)


bdbene (approved):
LGTM!

jpechane (approved):
@kgalieva LGTM, thanks!
Build failures seems unrelated. I triggered another run just in case and will merge when the run will finish.

jpechane:
@kgalieva Applied, thanks a lot!

PR #5925, opened 2024-10-07
DBZ-6984 Add multi-task support for MongoDB connectors

adityanair-stripe:
https://issues.redhat.com/browse/DBZ-6984

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
@@ -1008,6 +1008,51 @@ public class MongoDbConnectorConfig extends CommonConnectorConfig implements Sha
                     "for data change, schema change, transaction, heartbeat event etc.")
             .withDefault(DefaultTopicNamingStrategy.class.getName());
 
+    public static final Field MONGODB_MULTI_TASK_ENABLED = Field.create("mongodb.multi.task.enabled")
+            .withDisplayName("Enable Mongo DB multi-task")
+            .withType(Type.BOOLEAN)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.MEDIUM)
+            .withDefault(false)
+            .withValidation(Field::isBoolean)
+            .withDescription("Should the multi-task connector setup be enabled to the MongoDB instance, only valid if the connector manages a single replset.");
+
+    public static final Field MONGODB_MULTI_TASK_GEN = Field.create("mongodb.multi.task.gen")
+            .withDisplayName("Mongo DB multi-task generation")
+            .withType(Type.INT)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.MEDIUM)
+            .withDefault(-1)
+            .withValidation(Field::isInteger)
+            .withDescription("Define multi-task generation to be enabled on MongoDB instance");
+
+    public static final Field MONGODB_MULTI_TASK_PREV_TASKS = Field.create("mongodb.multi.prev.tasks.max")
+            .withDisplayName("Mongo DB previous multi-task generation task count")
+            .withType(Type.INT)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.MEDIUM)
+            .withDefault(1)
+            .withValidation(Field::isPositiveInteger)
+            .withDescription("Maximum number of tasks allowed in the previous generation");
+
+    public static final Field MONGODB_MULTI_TASK_PREV_GEN = Field.create("mongodb.multi.prev.gen")
+            .withDisplayName("Mongo DB previous multi-task generation")
+            .withType(Type.INT)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.MEDIUM)
+            .withDefault(-1)
+            .withValidation(Field::isInteger, MongoDbConnectorConfig::validateMultiTaskPrevGen)
+            .withDescription("Previous multi-task generation that was enabled on MongoDB instance");
+
+    public static final Field MONGODB_MULTI_TASK_HOP_SECONDS = Field.create("mongodb.multi.hop.sec")
+            .withDisplayName("Mongo DB previous multi-task hop seconds")
+            .withType(Type.INT)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.MEDIUM)
+            .withDefault(10)
+            .withValidation(Field::isPositiveInteger)
+            .withDescription("Mongo DB previous multi-task hop seconds");
+
     public static final Field SOURCE_INFO_STRUCT_MAKER = CommonConnectorConfig.SOURCE_INFO_STRUCT_MAKER
             .withDefault(MongoDbSourceInfoStructMaker.class.getName());
 
@@ -1027,7 +1072,12 @@ public class MongoDbConnectorConfig extends CommonConnectorConfig implements Sha
                     MONGODB_POLL_INTERVAL_MS,
                     SSL_ENABLED,
                     SSL_ALLOW_INVALID_HOSTNAMES,
-                    CURSOR_MAX_AWAIT_TIME_MS)
+                    CURSOR_MAX_AWAIT_TIME_MS,
+                    MONGODB_MULTI_TASK_ENABLED,
+                    MONGODB_MULTI_TASK_GEN,
+                    MONGODB_MULTI_TASK_PREV_TASKS,
+                    MONGODB_MULTI_TASK_PREV_GEN,
+                    MONGODB_MULTI_TASK_HOP_SECONDS)
             .events(
                     DATABASE_INCLUDE_LIST,
                     DATABASE_EXCLUDE_LIST,
@@ -1081,6 +1131,11 @@ public class MongoDbConnectorConfig extends CommonConnectorConfig implements Sha
     private final OversizeHandlingMode oversizeHandlingMode;
     private final FiltersMatchMode filtersMatchMode;
     private final int oversizeSkipThreshold;
+    private final boolean multiTaskEnabled;
+    private final int multiTaskGen;
+    private final int multiTaskPrevMaxTasks;
+    private final int multiTaskPrevGen;
+    private final int multiTaskHopSeconds;
 
     public MongoDbConnectorConfig(Configuration config) {
         super(config, DEFAULT_SNAPSHOT_FETCH_SIZE);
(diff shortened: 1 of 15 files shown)


debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -1210,6 +1212,7 @@ protected CommonConnectorConfig(Configuration config, int defaultSnapshotFetchSi
         this.signalEnabledChannels = getSignalEnabledChannels(config);
         this.skippedOperations = determineSkippedOperations(config);
         this.taskId = config.getString(TASK_ID);
+        this.maxTasks = config.getInteger(TASKS_MAX, 1);
```
jpechane, line 1215: Is this really guranteed that Kafka Connect will always propagate the connector infra level config to the connector configuration?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
```diff
@@ -1008,6 +1008,51 @@ public static OversizeHandlingMode parse(String value, String defaultValue) {
                     "for data change, schema change, transaction, heartbeat event etc.")
             .withDefault(DefaultTopicNamingStrategy.class.getName());
 
+    public static final Field MONGODB_MULTI_TASK_ENABLED = Field.create("mongodb.multi.task.enabled")
```
jpechane, line 1011: Do we really need to have such a config option? IMHO just the setting of `tasks.max > 1` should trigger the behaviour.
But as a safety concern (given we are going to merge it to stable branch) I'd keep this yet use `Field.createInternal`. It would be `false` by default in 3.0 and `true` in 3.1. The option could be then removed completely.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
```diff
@@ -1008,6 +1008,51 @@ public static OversizeHandlingMode parse(String value, String defaultValue) {
...
+            .withDisplayName("Enable Mongo DB multi-task")
```
jpechane, line 1012: Here and anywhere else, it is `MongoDB` not `Mongo DB`

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
```diff
@@ -1008,6 +1008,51 @@ public static OversizeHandlingMode parse(String value, String defaultValue) {
...
+    public static final Field MONGODB_MULTI_TASK_GEN = Field.create("mongodb.multi.task.gen")
+            .withDisplayName("Mongo DB multi-task generation")
+            .withType(Type.INT)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.MEDIUM)
+            .withDefault(-1)
+            .withValidation(Field::isInteger)
+            .withDescription("Define multi-task generation to be enabled on MongoDB instance");
+
+    public static final Field MONGODB_MULTI_TASK_PREV_TASKS = Field.create("mongodb.multi.prev.tasks.max")
```
jpechane, line 1029: Are the `prev` and even `gen` options really necessary? Can't we store the number of tasksin the offsets and detect the change based on comparison of value from offests with the one configured in connector config?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbPartition.java
```diff
@@ -35,13 +48,34 @@ public MongoDbPartition(String serverId) {
...
     @Override
     public Map<String, String> getSourcePartition() {
-        return replicaSetName != null
-                ? Collect.hashMapOf(SERVER_ID_KEY, serverId, REPLICA_SET_NAME, replicaSetName)
-                : Collect.hashMapOf(SERVER_ID_KEY, serverId);
+        Map<String, String> partition = new HashMap<>();
+        if (this.multiTaskEnabled) {
+            partition.put(TASK_ID_KEY, String.valueOf(this.taskId));
+            partition.put(MULTI_TASK_GEN_KEY, String.valueOf(this.multiTaskGen));
+            partition.put(TASK_COUNT_KEY, String.valueOf(taskCount));
```
jpechane, line 72: Why do we need to use taks count?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
jcechace: I would definitively keep an option that specifically enables **THIS TYPE** of multitask As in the future we might want to do multi tasking based on e.g. databases.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbPartition.java
adityanair-stripe: We included task count in the partition to guard against accidental misconfiguration. Ex: if generation 0 has 3 tasks, and generation 1 is configured with `prev_tasks=1` the connector will fail to find any offsets and exit instead of silently using the incorrect configuration.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
adityanair-stripe: `prev_gen` is here because we might not want to start at `generation - 1`. Suppose there was some problem with the current generation, we might want to start a new generation starting at the previous generation. Ex: generation 0 is fine, generation 1 is problematic, so we start generation 2 from generation 0's offsets. For this to work `gen` and `prev_gen` need to be explicitly defined.

Moving `gen` into the offset creates an issue in the above scenario because if we don't include the generation in the partition then we won't be able to get the previous generation's offsets.

jcechace:
I'm going over this. However before this gets merged I would very much appreciate at least some basic docs describing how the system works.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
ning-stripe: Agree with @jcechace on this :) There is also an existing replset based multi-tasking for Mongo, so we wanted to be very explicit about having a flag to turn the feature on/off.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/MongoDbMultiTaskConnectorIT.java
```diff
@@ -0,0 +1,241 @@
...
+import org.bson.BsonTimestamp;
+import org.bson.Document;
+import org.bson.types.ObjectId;
+import org.junit.After;
+import org.junit.Before;
+import org.junit.Test;
+
+import io.debezium.data.Envelope;
+
+public class MongoDbMultiTaskConnectorIT extends AbstractMongoConnectorIT {
```
ning-stripe, line 28: Amazing work! The tests look so much more thorough and less hacky :D

jcechace:
@adityanair-stripe  I'm slowly crunching through the PR. Would it be possible to schedule a call for the next week to discuss the concept? I know we touched on it sometime last year but a lot of things changed since then -- for one the buffering cursor did not exist. 

The way I see this being helpful is only when there is a significant write into a single collection and Debezium is not able to keep up in single task. 

If the writes were spread across databases / collections then multitasking based on databases/collections seems to make more sense. If the load is really on single collection, wouldn't it then make more sense to introduce paralel event processing into a single task?

I'm trying to understand where exactly the bottleneck was and whether this is addressing it at the right spot -- as although this is likely a feasible solution, it also adds a lot of complexity to the code.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/events/BufferingChangeStreamCursor.java
```diff
@@ -257,25 +337,29 @@ public int size() {
...
+                }
+                catch (Throwable e) {
+                    error.set(e);
+                    LOGGER.error("Fetcher thread has failed", e);
+                    close();
+                }
+                finally {
+                    cursorRef.set(null);
+                }
+            } while (isRunning());
```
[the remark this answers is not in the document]
adityanair-stripe: The outer loop is to move the multi-task window whenever we finish streaming all records from the current window. Within `fetchEvents` we call `shouldUpdateStream` on every new event and break out of the inner loop when it returns true (when we get an event outside the current window). Then we move the window, clean up the cursor, update the stream, and open a new cursor before starting to iterate over events again.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MultiTaskOffsetHandler.java
```diff
@@ -0,0 +1,132 @@
...
+ * If the offset is past the stop time, it will move the next hopt
+ *
+ * EX: 3 tasks, hop size of 5, and lastOffset of 7, will distribute the tasks based on offsetTime showing start and stop times.
+ *  * Task 0: [15->19], [30->34], ... (Skipping [0-4])
+ *  * Task 1: [7->9], [20->24], ...   (Reducing the first hop from [5->9] to [7->9]
+ *  * Task 2: [10->14], [25->29], ... (No changes)
+ *
+ * @author Anthony Noakes
+ */
+public class MultiTaskOffsetHandler {
```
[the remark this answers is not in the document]
adityanair-stripe: sure, I'll call it `MultiTaskWindowHandler`

adityanair-stripe:
@jcechace 
> However before this gets merged I would very much appreciate at least some basic docs describing how the system works.

I've collected some of our docs and combined them into [this doc](https://docs.google.com/document/d/e/2PACX-1vTL3DP8BkyAvQj3B9u3jmB8df3cmXOegaLFzCTPHLcf0WCjsQBNM-8kHDtwmlxjKUIUS3DlJsE8jKhc/pub). Let me know if there are any other details about the design that you'd like me to elaborate on.

> Would it be possible to schedule a call for the next week to discuss the concept

Sent an email, let me know what time works for you.

> The way I see this being helpful is only when there is a significant write into a single collection

Yeah, this is a common occurrence for us - one of the goals of this feature is to solve the 'hot collection' issue described here. Distributing events to tasks based on db/collection doesn't really address the issue.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorConfig.java
jcechace: @ning-stripe RS-based one was removed as it was highly problematic

jcechace:
> this doc

This seems not to be accessible. Can you either make it public or share with me directly?

Naros:

Hi @adityanair-stripe, please sign off all commits with:

```
git commit -s
```

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
adityanair-stripe: I don't think there's any documented guarantee. It is the actual behavior though: https://github.com/apache/kafka/blob/trunk/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/distributed/DistributedHerder.java#L2106

PR #5930, opened 2024-10-10
DBZ-8194 Always set binlog client GTID when GTID is enabled

vjuranek:
When GTID is enabled, binlog client tracks GTID and eventually resumes from the last GTID position. However, to enable GTID on the client, client has to set initial GTID position (can be empty). When GTID is enbled in the connector, always enable GTID also on the binlog client to make sure that it tracks GTID and its keep alive thread eventually resumes from the right position.

https://issues.redhat.com/browse/DBZ-8194

### DIFF
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
@@ -253,7 +253,18 @@ public abstract class BinlogStreamingChangeEventSource<P extends BinlogPartition
                 // We've not yet seen any GTIDs, so that means we have to start reading the binlog from the beginning ...
                 client.setBinlogFilename(effectiveOffsetContext.getSource().binlogFilename());
                 client.setBinlogPosition(effectiveOffsetContext.getSource().binlogPosition());
-                initializeGtidSet("");
+                if (purgedServerGtidSet == null || purgedServerGtidSet.isEmpty()) {
+                    LOGGER.info("No GTID stored in the offset, registering binlog reader with empty GTID set.");
+                    client.setGtidSet("");
+                    initializeGtidSet("");
+                }
+                else {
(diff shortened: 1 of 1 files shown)


debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
```diff
@@ -253,7 +253,18 @@ public void execute(ChangeEventSourceContext context, P partition, O offsetConte
                 // We've not yet seen any GTIDs, so that means we have to start reading the binlog from the beginning ...
                 client.setBinlogFilename(effectiveOffsetContext.getSource().binlogFilename());
                 client.setBinlogPosition(effectiveOffsetContext.getSource().binlogPosition());
-                initializeGtidSet("");
+                if (purgedServerGtidSet.isEmpty()) {
```
jpechane, line 256: Looking at `io.debezium.connector.mysql.jdbc.MySqlConnection.purgedGtidSet()` it seems to me that `null` is valid return value. Would the code be safe in that case?

vjuranek: I don't think we can ever get `null` from [here](https://github.com/debezium/debezium/blob/main/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java#L108-L110) (and same for [MariaDB](https://github.com/debezium/debezium/blob/main/debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/jdbc/MariaDbConnection.java#L71))

vjuranek: Just noted the logging message is misleading and needs to be fixed, so the test will run again anyway - adding also the check for `null` as it never hurts.

jpechane:
@vjuranek LGTM, applied! Thanks

PR #5932, opened 2024-10-10
DBZ-8314 Clarifies on enabling CDC on MS SQL Server table

x-0-r:
Aligns the documentation with the official MS docs for sys.sp_cdc_enable_table (https://learn.microsoft.com/en-us/sql/relational-databases/system-stored-procedures/sys-sp-cdc-enable-table-transact-sql)


jpechane:
@x-0-r Applied, thanks

PR #5936, opened 2024-10-15
DBZ-8324 Add support for MySQL 9.1

jpechane:
https://issues.redhat.com/browse/DBZ-8324


jpechane:
JDBC driver is not yet in Maven Central

PR #5938, opened 2024-10-16
DBZ-7903 Improve blocking snapshot reliability in case of restart

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-7903

Tasks:

- [X] Change on main relational connectors
- [x] Change on MongoDB
- [x] Address others connectors
- [x] Evaluate if incremental snapshot can be moved to SnapshotType enum too

### DIFF
--- a/debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbOffsetContext.java
+++ b/debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbOffsetContext.java
@@ -60,8 +60,8 @@ public class MariaDbOffsetContext extends BinlogOffsetContext<SourceInfo> {
             final long binlogPosition = longOffsetValue(offset, BinlogSourceInfo.BINLOG_POSITION_OFFSET_KEY);
 
             final MariaDbOffsetContext offsetContext = new MariaDbOffsetContext(
-                    loadSnapshot((Map<String, Object>) offset),
-                    isTrue(offset, SNAPSHOT_COMPLETED_KEY),
+                    loadSnapshot(offset).orElse(null),
+                    loadSnapshotCompleted(offset),
                     TransactionContext.load(offset),
                     connectorConfig.isReadOnlyConnection()
                             ? MariaDbReadOnlyIncrementalSnapshotContext.load(offset)
--- a/debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbOffsetContext.java
+++ b/debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbOffsetContext.java
@@ -11,6 +11,7 @@ import java.util.Map;
 
 import org.apache.kafka.connect.errors.ConnectException;
 
+import io.debezium.connector.SnapshotType;
 import io.debezium.connector.binlog.BinlogOffsetContext;
 import io.debezium.connector.binlog.BinlogSourceInfo;
 import io.debezium.pipeline.source.snapshot.incremental.IncrementalSnapshotContext;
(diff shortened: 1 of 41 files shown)


mfvitale:
For MongoDB I just added the test case checking that the current implementation is not affected by the issue since the implementation differs. As of now I think it is not worth to invest time to align it. @jpechane WDYT?

mfvitale:
@jpechane Tests failures is not related.

debezium-core/src/main/java/io/debezium/pipeline/spi/OffsetContext.java
```diff
@@ -31,6 +33,18 @@ public interface OffsetContext {
      * Implementations load a connector-specific offset context based on the offset values stored in Kafka.
      */
     interface Loader<O extends OffsetContext> {
+
+        default SnapshotType loadSnapshot(Map<String, Object> offset) {
```
jpechane, line 37:
```suggestion
        default Optional<SnapshotType> loadSnapshot(Map<String, Object> offset) {
```

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbOffsetContext.java
```diff
@@ -59,7 +60,7 @@ public MariaDbOffsetContext load(Map<String, ?> offset) {
             final long binlogPosition = longOffsetValue(offset, BinlogSourceInfo.BINLOG_POSITION_OFFSET_KEY);
 
             final MariaDbOffsetContext offsetContext = new MariaDbOffsetContext(
-                    isTrue(offset, BinlogSourceInfo.SNAPSHOT_KEY),
+                    loadSnapshot((Map<String, Object>) offset),
                     isTrue(offset, SNAPSHOT_COMPLETED_KEY),
```
jpechane, line 64: IMHO it makes sense tro introduce a new `loadSnapshotMethod` as above now when the snapshot completed key is in common class.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresOffsetContext.java
```diff
@@ -207,15 +207,18 @@ public PostgresOffsetContext load(Map<String, ?> offset) {
             final String msgType = (String) offset.getOrDefault(SourceInfo.MSG_TYPE_KEY, null);
             final Operation messageType = msgType == null ? null : Operation.valueOf(msgType);
             final Instant useconds = Conversions.toInstantFromMicros((Long) ((Map<String, Object>) offset).getOrDefault(SourceInfo.TIMESTAMP_USEC_KEY, 0L));
-            final boolean snapshot = (boolean) ((Map<String, Object>) offset).getOrDefault(SourceInfo.SNAPSHOT_KEY, Boolean.FALSE);
+            final SnapshotType snapshot = loadSnapshot((Map<String, Object>) offset);
+            boolean snapshotCompleted = Boolean.TRUE.equals(offset.get(CommonOffsetContext.SNAPSHOT_COMPLETED_KEY));
```
jpechane, line 211: How this will work in case of connector update when this field is not provided? Might be an issue for the other connectors too.

debezium-core/src/main/java/io/debezium/pipeline/CommonOffsetContext.java
```diff
@@ -5,20 +5,35 @@
...
+import io.debezium.connector.SnapshotType;
 import io.debezium.connector.common.BaseSourceInfo;
 import io.debezium.pipeline.spi.OffsetContext;
 
 public abstract class CommonOffsetContext<T extends BaseSourceInfo> implements OffsetContext {
 
+    public static final String SNAPSHOT_COMPLETED_KEY = "snapshot_completed";
+
     protected final T sourceInfo;
+    protected SnapshotType snapshot;
```
jpechane, line 22: Could we add a JavaDoc that will state what values will be for `snapshot` and `snapshotCompleted` fields in case of the following scenarios

* incomplete initial snapshot
* completed initial snapshot
* incmplete blocking snapshot
* completed blocking snapshot
* running incremental snapshot
* completed incremental snapshot

Such documentation would work as a contract against which we can assert tests and define code behaviour.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresOffsetContext.java
mfvitale: Well, before this PR the all connectors used the `snapshotCompleted` flag so they should have the flag in their offset and during the update it will correctly managed. 

Only PostgreSQL haven't the flag before and so it will be put as `false`.

mfvitale:
@jpechane can you have another look?

jpechane:
@mfvitale Applied, thanks a lot! Please keep an eye on non-core connector CI builds to make sure eferything fits together.

jpechane:
@mfvitale OK, it seems that the method rename did not get into the all non-core repos. Could you please fix it and merge it on your own? Thanks

PR #5939, opened 2024-10-16
DBZ-8325 Remove converters from mysql parsers

twthorn:
Remove converters from mysql parsers. 

A refactor to make https://github.com/debezium/debezium-connector-vitess/pull/210 possible.

I tried moving more language specific DDL parsers (eg mariadb, mysql, oracle) to the debezium-ddl-parser package but they are quite inter-tangled with other parts of the mariadb/mysql connectors (eg converters, charset registry). Additionally, the test modules (eg MySqlAntlrDdlParserTest) rely on BinlongAntlrDdlParserTest which relies on specific TableSchemaBuilders (eg MySqlTableSchemaBuilder) and others. It seems difficult to do a proper refactor without circular dependencies. I could move the AntlrDdlParsers and not their tests but it seems odd to have a class in one package and its unit tests located in another package. Based on this, I am just having the vitess-connector import the MySqlConnector. Let me know if there are other ideas for how to structure this.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -1015,7 +1015,7 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
         /**
          * Perform a snapshot of the schema but no data upon initial startup of a connector.
          *
-         * @deprecated to be removed in Debezium 3.0, replaced by {{@link #NO_DATA}}
+         * @deprecated to be removed in Debezium 3.0, replaced by {{@link #no_data}}
          */
         SCHEMA_ONLY("schema_only"),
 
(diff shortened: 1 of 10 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -1015,7 +1015,7 @@ public enum SnapshotMode implements EnumeratedValue {
         /**
          * Perform a snapshot of the schema but no data upon initial startup of a connector.
          *
-         * @deprecated to be removed in Debezium 3.0, replaced by {{@link #NO_DATA}}
+         * @deprecated to be removed in Debezium 3.0, replaced by {{@link #no_data}}
```
Naros, line 1018: The link is supposed to be to the enum `NO_DATA`, so I believe the capitalization here is accurate, although I generally use `{@link #NO_DATA}` rather than the double-brackets, but IDEA seems to support both.

Naros (commented):
@twthorn @jpechane if the Vitess changes for DDL are something you both land on as part of Debezium 3.1 (if you haven't already), I wonder if we could then simply create a separate Jira and attribute this change to that for Debezium 3?

These changes are a net zero behavior change and just remove unnecessary code that wasn't being used in the connector, so its generally harmless IMO.  wdyt?

jpechane:
@twthorn I agree with @Naros. Let's make a separate Jira for this PR so it can be merged to the `main` - current 3.0.
For the vitess we'll create a separate forward looking branch into which the PR will be merged and that branch will be merged back to main in January.

twthorn:
New PR with updated Jira https://github.com/debezium/debezium/pull/5961

PR #5945, opened 2024-10-21
[docs] Remove unordered record processing from sever docs

vjuranek:
Currently Debezium server allows to use only sinks which implements `ChangeConsumer` interface, which excludes these options. Listing these properties may be confusing for users. Remove them until we actually allow to use them in the Debezium sever.


jpechane:
@vjuranek Applied, thanks

PR #5947, opened 2024-10-23
DBZ-8348 add Cassandra 5 version to pom.xml

smiklosovic:


smiklosovic:
@jpechane @vjuranek would you guys take a look? This will be integrated with https://github.com/debezium/debezium-connector-cassandra/pull/135

smiklosovic:
@vjuranek @jpechane errors are not related

PR #5950, opened 2024-10-24
DBZ-8208 Extend error.processing.failure.handling.mode to DML parsing

Naros:
https://issues.redhat.com/browse/DBZ-8208

@jpechane backport to 2.7


jpechane:
I wonder if we should use `io.debezium.util.Loggings` here and in the other part of `io.debezium.connector.oracle.logminer.processor.AbstractLogMinerEventProcessor` to not leak source data if not required.

WDYT?

PR #5951, opened 2024-10-24
DBZ-8316 Support POSITION keyword as column name

Naros:
https://issues.redhat.com/browse/DBZ-8316

Upstream PR https://github.com/antlr/grammars-v4/pull/4287


jpechane:
@Naros Applied, thanks

PR #5952, opened 2024-10-24
DBZ-8315 Add in-memory-clause support for sub-partition lists

Naros:
https://issues.redhat.com/browse/DBZ-8315

Upstream PR https://github.com/antlr/grammars-v4/pull/4288


jpechane:
@Naros Applied, thanks

PR #5953, opened 2024-10-24
DBZ-8345 Don't fault archive log only mode when SCN is unavailable

Naros:
https://issues.redhat.com/browse/DBZ-8345


jpechane:
@Naros Applied, thanks!

jpechane:
@Naros I backported it to 2.7 too. Is it OK?

Naros:
As discussed on the call, it is @jpechane

PR #5955, opened 2024-10-24
DBZ-8305 Add SECONDARY_ENGINE table option

ani-sha:
https://issues.redhat.com/browse/DBZ-8305


jpechane:
@ani-sha Applied, thanks

PR #5956, opened 2024-10-24
DBZ-8290 Increment binlog event counter

vjuranek:
https://issues.redhat.com/browse/DBZ-8290


vjuranek:
It's little bit questionable which records we should count, as we the message is printed when binlog client is disconnecting, I'd count all the records which binlog client processed, not only records produced by Debezium.

jpechane:
@vjuranek Yes, good decision. Applied, thanks!

PR #5960, opened 2024-10-28
DBZ-8351 Add possibility to store offsets into a config map with ConfigMapOffsetStore

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-8351

The test uses the [kube-api-test](https://github.com/fabric8io/kubernetes-client/blob/main/doc/kube-api-test.md) 

I did also some e2e test on local minikube and the config map as been created 

```bash
kubectl get cm
NAME                                     DATA   AGE
debezium-offsets                         1      27m
my-debezium                              1      6h29m
```

```bash
kubectl get cm debezium-offsets -o yaml
apiVersion: v1
binaryData:
  kafka.server-inventory: eyJsYXN0X3NuYXBzaG90X3JlY29yZCI6dHJ1ZSwibHNuIjozNzkzMDM2MCwidHhJZCI6Nzc4LCJ0c191c2VjIjoxNzMwMTI4MzA4NDQ4NjM0LCJzbmFwc2hvdCI6dHJ1ZX0=
kind: ConfigMap
metadata:
  creationTimestamp: "2024-10-28T15:11:48Z"
  name: debezium-offsets
  namespace: debezium
  resourceVersion: "150819"
  uid: e05a00dc-cd0e-4aba-9d88-7edf6b7af9d2

```

```bash
echo eyJsYXN0X3NuYXBzaG90X3JlY29yZCI6dHJ1ZSwibHNuIjozNzkzMDM2MCwidHhJZCI6Nzc4LCJ0c191c2VjIjoxNzMwMTI4MzA4NDQ4NjM0LCJzbmFwc2hvdCI6dHJ1ZX0= | base64 --decode
{"last_snapshot_record":true,"lsn":37930360,"txId":778,"ts_usec":1730128308448634,"snapshot":true}
```
After killing the pod it restarts without taking the initial snapshot.


mfvitale:
Failures are related to https://github.com/debezium/debezium/pull/5963

debezium-bom/pom.xml
```diff
@@ -509,6 +512,13 @@
                 <version>${version.azure.blob}</version>
             </dependency>
 
+            <!-- K8s Config Map Storage -->
+            <dependency>
+                <groupId>io.fabric8</groupId>
+                <artifactId>kubernetes-client</artifactId>
+                <version>${version.kubernetes.client}</version>
+            </dependency>
```
jcechace, line 520: This will need to be excluded in DS as we should use the same version as Quarkus kubernetes-client extension. 

I would actually consider putting the entire module under DS project. However on the oposite side it might also be useful for use with engine...

jpechane: We've already had a debate with @mfvitale about it and at the end we opted for the core module.

debezium-storage/debezium-storage-configmap/src/main/java/io/debezium/storage/configmap/ConfigMapFormatter.java
```diff
@@ -0,0 +1,112 @@
...
+ * <p>
+ * The keys are converted to a valid config:
+ * alphanumeric characters, '-', '_' or '.' (e.g. 'key.name',  or 'KEY_NAME',  or 'key-name',
+ * regex used for validation is '[-._a-zA-Z0-9]+').
+ * <p>
+ * The values are just base64 encoded
+ *
+ * @author Mario Fiore Vitale
+ */
+public class ConfigMapFormatter {
```
jcechace, line 32: We need a better docs for this -- or better a test class for format specifically

debezium-storage/debezium-storage-configmap/src/main/java/io/debezium/storage/configmap/ConfigMapFormatter.java
```diff
@@ -0,0 +1,112 @@
...
+        String[] keyAndValue = storedEntry.split(KEY_VALUE_SEPARATOR);
+        return Map.entry(keyAndValue[0], keyAndValue[1]);
+    }
+
+    private String toStringKey(ByteBuffer key) {
+
+        try {
+            JsonNode jsonNode = objectMapper.readTree(key.array());
+
+            // a valid config key must consist of alphanumeric characters, '-', '_' or '.' (e.g. 'key.name', or 'KEY_NAME', or 'key-name', regex used for validation is '[-._a-zA-Z0-9]+').
```
jcechace, line 88: Add a note that offset key is always in a form `[namespace, partition]`. On KC `namespace` is the connector name, for engine it is the configured engine name.

jcechace: Just to be sure for engine/connector name `myengine` and a partitions 

```json
{
  "pkey1": "A1",
  "pkey2": "A2"
}
```

```json
{
  "pkey1": "B1",
  "pkey2": "B2"
}
```

and some arbitrary offsets values (say `{offset: 1}` and `{offset: 2}` the config map would have the following entires?

```text
"engine.pkey1-A1_pkey2-A2": "{offset: 1}"
"engine.pkey1-B1_pkey2-B2": "{offset: 2}" 
```

While this is fine I think we should be able to omit the name of the connector / engine completely here as each connector/engine has it's own config map (which would be actually a reasonable default for the name). This way we can simplify the used separator a bit to something like 

```text
"pkey1-A1.pkey2-A2": "{offset: 1}"
"pkey1-B1.pkey2-B2": "{offset: 2}" 
```

The name of the engine/connector can be placed into annotations.

mfvitale: > I think we should be able to omit the name of the connector / engine completely here as each connector/engine has it's own config map (which would be actually a reasonable default for the name). This way we can simplify the used separator a bit to something like

But then you still need to add it when the offset are read from the config map otherwise it will never match the original offset keys. Said that what is the real benefit of this simplification?

mfvitale: >  the config map would have the following entires?
>```text
>"engine.pkey1-A1_pkey2-A2": "{offset: 1}"
>"engine.pkey1-B1_pkey2-B2": "{offset: 2}" 
>```

Yes, but more precisely

``text
"myengine.pkey1-A1_pkey2-A2": "e29mZnNldDogMX0="
"myengine.pkey1-B1_pkey2-B2": "e29mZnNldDogMn0=" 
```

mfvitale:
@jcechace can you please give another look on that?

jcechace (approved):

jcechace:
@mfvitale Maybe one more thing.. add `@Incubating`? So that we keep the possibility of tweaking things if needed later?

PR #5961, opened 2024-10-28
DBZ-8360 Remove converters from mysql parsers

twthorn:
Remove converters from mysql parsers.

A refactor to make https://github.com/debezium/debezium-connector-vitess/pull/210 possible.

I tried moving more language specific DDL parsers (eg mariadb, mysql, oracle) to the debezium-ddl-parser package but they are quite inter-tangled with other parts of the mariadb/mysql connectors (eg converters, charset registry). Additionally, the test modules (eg MySqlAntlrDdlParserTest) rely on BinlongAntlrDdlParserTest which relies on specific TableSchemaBuilders (eg MySqlTableSchemaBuilder) and others. It seems difficult to do a proper refactor without circular dependencies. I could move the AntlrDdlParsers and not their tests but it seems odd to have a class in one package and its unit tests located in another package. Based on this, I am just having the vitess-connector import the MySqlConnector. Let me know if there are other ideas for how to structure this.

### DIFF
--- a/debezium-connector-mariadb/src/test/java/io/debezium/connector/mariadb/DefaultValueTest.java
+++ b/debezium-connector-mariadb/src/test/java/io/debezium/connector/mariadb/DefaultValueTest.java
@@ -24,7 +24,7 @@ import io.debezium.relational.RelationalDatabaseConnectorConfig;
  */
 public class DefaultValueTest extends BinlogDefaultValueTest<MariaDbValueConverters, MariaDbAntlrDdlParser> {
     @Override
-    protected MariaDbAntlrDdlParser getDdlParser(MariaDbValueConverters valueConverters) {
+    protected MariaDbAntlrDdlParser getDdlParser() {
         return new MariaDbAntlrDdlParser();
     }
 
--- a/debezium-connector-mariadb/src/test/java/io/debezium/connector/mariadb/DefaultValueTest.java
+++ b/debezium-connector-mariadb/src/test/java/io/debezium/connector/mariadb/DefaultValueTest.java
@@ -25,7 +25,7 @@ import io.debezium.relational.RelationalDatabaseConnectorConfig;
 public class DefaultValueTest extends BinlogDefaultValueTest<MariaDbValueConverters, MariaDbAntlrDdlParser> {
     @Override
     protected MariaDbAntlrDdlParser getDdlParser(MariaDbValueConverters valueConverters) {
-        return new MariaDbAntlrDdlParser(valueConverters);
+        return new MariaDbAntlrDdlParser();
     }
 
     @Override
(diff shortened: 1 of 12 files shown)


twthorn:
@Naros @jpechane created an updated pull request with separate Jira ticket (also addressed feedback on java doc comment from previous PR). Let me know if I can do anything else.

debezium-connector-mariadb/src/test/java/io/debezium/connector/mariadb/DefaultValueTest.java
```diff
@@ -25,7 +25,7 @@
 public class DefaultValueTest extends BinlogDefaultValueTest<MariaDbValueConverters, MariaDbAntlrDdlParser> {
     @Override
     protected MariaDbAntlrDdlParser getDdlParser(MariaDbValueConverters valueConverters) {
```
Naros, line 27:
```suggestion
    protected MariaDbAntlrDdlParser getDdlParser() {
```
This and any other references to `getDdlParser()` function in `BinlogDefaultValueTest` imo.

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlDefaultValueTest.java
```diff
@@ -24,7 +24,7 @@
 public class MySqlDefaultValueTest extends BinlogDefaultValueTest<MySqlValueConverters, MySqlAntlrDdlParser> {
     @Override
     protected MySqlAntlrDdlParser getDdlParser(MySqlValueConverters valueConverter) {
```
Naros, line 26:
```suggestion
    protected MySqlAntlrDdlParser getDdlParser() {
```
Same as my former comment.

Naros (commented):
Overall LGTM, just one small nit-pick @twthorn.

jpechane:
@twthorn It seems to me that `debezium-storage` module fails to compile now.

twthorn:
@jpechane I tried compiling locally and I see no such issue. Is it possible to trigger a rebuild or is there some stored state that may be affecting the test? The error
```
/home/runner/work/debezium/debezium/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlDefaultValueTest.java:[24,8] io.debezium.connector.mysql.MySqlDefaultValueTest is not abstract and does not override abstract method 
```

Seems to indicate that it has the latest code for `MySqlDefaultValueTest` but not the latest for debezium-connector-binlog so it thinks that it is not overriding all methods (ie the `protected abstract P getDdlParser();`).

jpechane:
@twthorn Hi, yes, you are right. It builds yet I get `io.debezium.storage.s3.history.S3SchemaHistoryTest.testMissingObjectName` test to fail.

jpechane:
ok, I found the reason for test failiure. It is unrelated to this PR.

jpechane:
Added fix to the test failure, rebased on the latest main and force pushed. When Ci completes the PR will be merged.

jpechane:
@twthorn Applied, thanks

PR #5964, opened 2024-10-30
DBZ-8208 Use the Loggings utility class

Naros:
https://issues.redhat.com/browse/DBZ-8208


jpechane:
@Naros Applied, thanks

PR #5966, opened 2024-10-31
[ci] Add Informix matrix strategy

nrkljo:


jpechane:
@nrkljo Applied, thanks

PR #5969, opened 2024-11-01
DBZ-8371 bump protobuf dependencies to avoid vulnerability

enzo-cappa:
The current version of the protobuf libraries (3.25.2) is affected by [SNYK-JAVA-COMGOOGLEPROTOBUF-8055227](https://security.snyk.io/vuln/SNYK-JAVA-COMGOOGLEPROTOBUF-8055227).

The recommended fix is to upgrade to 3.25.5, which should be safe as it's just a patch fix.


jpechane:
@enzo-cappa Applied, thanks for reporting!

PR #5970, opened 2024-11-01
[ci] Back-port Informix matrix strategy

nrkljo:


jpechane:
@nrkljo This should be unnecessary as GH CI as managed from the `main` based config.

nrkljo:
@jpechane Of course, thanks!

PR #5973, opened 2024-11-04
DBZ-8375 Migrate rest of Debezium testsuite to async engine

vjuranek:
https://issues.redhat.com/browse/DBZ-8375


vjuranek:
DB2 test failu is not related.

PR #5976, opened 2024-11-05
DBZ-8346 Use DebeziumSinkRecord instead of Kafka Connect's SinkRecord inside Debezium sink connectors

rk3rn3r:
- pt.1 Refactor TableId to CollectionId and related classes/variable name changes and refactoring, introduce debezium-sink package
- pt.2 Switching to DebeziumSinkRecord instead of SinkRecord
- pt.3 Add Field#withDeprecatedAliases for backward compatible aliases for config options; updated documentation
- pt.4 Re-Add deprecated implementations to not harm existing user implementations using the old API / config options (DefaultTableNamingStrategy, TableNamingStrategy), incl test scenarios

closes to https://issues.redhat.com/browse/DBZ-8346


rk3rn3r:
@jpechane looks like the failing MongoDB 7.0 test was a hiccup kind of issue from the 7.0 image. Tests worked now after another restart this morning and local tests with -Dversion.mongo.server=7.0' were green too.

I'll be adding the second commit with the `DebeziumSinkRecord` interface and related implementations later.

rk3rn3r:
@jpechane I have added the second commit to add `DebeziumSinkRecord`. Local tests were all good now. Waiting for CI now.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -205,14 +159,6 @@ public class JdbcSinkConnectorConfig {
             .withDescription("A comma-separated list of primary key field names. " +
                     "This is interpreted differently depending on " + PRIMARY_KEY_MODE + ".");
 
-    public static final Field DATABASE_TIME_ZONE_FIELD = Field.create(DATABASE_TIME_ZONE)
```
jpechane, line 208: Please make sure that no config options are renamed without deprecation period, especially  when the change is going into micro release. I guess the same problem is with `TABLE_NAME_FORMAT_FIELD`.

rk3rn3r: Sure, I had this on my list. Should we add an `alias` and/or `deprecated alias` for property names? With `deprecated alias` we could automate deprecation warnings. wdyt @jpechane

jpechane: @rk3rn3r Neat idea!

debezium-sink/pom.xml
```diff
@@ -0,0 +1,36 @@
...
+        <version>3.0.0-SNAPSHOT</version>
+        <relativePath>../debezium-parent/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>debezium-sink</artifactId>
+    <name>Debezium Sink</name>
+    <description>Debezium Sink module</description>
+
+    <properties>
+        <maven.compiler.source>17</maven.compiler.source>
```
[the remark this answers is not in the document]
rk3rn3r: I overlooked that. I think the IDE somehow sneaked it in. Will drop it.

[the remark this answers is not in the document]
rk3rn3r: Done

rk3rn3r:
@jpechane I have added a hopefully working `Field#withDeprecatedAliases()` method to allow defining deprecated alias configs. The change is deep with the changes done to the Field and Configuration classes. I hope the logic is still okay and won't harm previous code. We need to carefully check the addition in the extra 3rd commit.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
rk3rn3r: Done

mfvitale:
> The change is deep with the changes done to the Field and Configuration classes. I hope the logic is still okay and won't harm previous code. We need to carefully check the addition in the extra 3rd commit.

Can we add a test?

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/CollectionNamingStrategyTest.java
```diff
@@ -0,0 +1,136 @@
...
+    public void testCollectionNamingStrategyWithTableNameFormat() {
+        final JdbcSinkConnectorConfig config = new JdbcSinkConnectorConfig(Map.of(JdbcSinkConnectorConfig.COLLECTION_NAME_FORMAT, "kafka_${topic}"));
+        final SinkRecordFactory factory = new DebeziumSinkRecordFactory();
+        final DefaultCollectionNamingStrategy strategy = new DefaultCollectionNamingStrategy();
+        assertThat(strategy.resolveCollectionName(factory.createRecord("database.schema.table"), config.getCollectionNameFormat()))
+                .isEqualTo("kafka_database_schema_table");
+    }
+
+    @Test
+    public void testDeprecatedTableNamingStrategyWithTableNameFormat() {
```
rk3rn3r, line 50: @mfvitale You find tests for the deprecated option names here. Is that what you were looking for?

Naros:
> @rk3rn3r LGTM, thanks a lot! I left just few comments, the most important relates to config option names. When addresd the PR can be merged.

@jpechane if we're concerned with configuration option names being changed without leaving deprecated predecessors, wouldn't the same apply to the changes made to `TableNamingStrategy` and `ColumnNamingStrategy`. The contracts are completely different and the old ones are not left in-place with default methods to make the transition easy. As is, any user who customized those public hooks won't be able to upgrade without changes.

rk3rn3r:
@Naros if those things are an issue it should be easy to provide old contracts that call the new functionality, but I think we never guaranteed compatibility on the code API layer, did we?

Naros:
> @Naros if those things are an issue it should be easy to provide old contracts that call the new functionality, but I think we never guaranteed compatibility on the code API layer, did we?

Hi @rk3rn3r since these are contracts that can be supplied by user/customers, we need an adapter layer to make sure that in a micro-release that these continue to function as-is with warnings that notify the user they should consider porting their code to use the new contracts and we'll remove the adapter/deprecation layer bits in 3.1.

rk3rn3r:
@Naros do you mean config options or really the old class names?

jpechane:
@rk3rn3r Hi, generally if we providea way for the user to inject his own code (even if it is not yet matured into API module) we must not break the contract in microreleases. For that reaosn a code that was written with the old interfaces must still compile and be runnable. Ideally there will be a compatibility/legacy layer that would ensure the old contract is still available and will be removed later in one of minor releases.

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/CollectionNamingStrategyTest.java
mfvitale: I am refereeing more to the deprecation functionality it self. Checking the working message for example.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/naming/DefaultTableNamingStrategy.java
```diff
@@ -0,0 +1,26 @@
...
+import io.debezium.sink.naming.DefaultCollectionNamingStrategy;
+
+/**
+ * Default implementation of the {@link TableNamingStrategy} where the table name is driven
+ * directly from the topic name, replacing any {@code dot} characters with {@code underscore}
+ * and source field in topic.
+ *
+ * @author Chris Cranford
+ */
+public class DefaultTableNamingStrategy extends DefaultCollectionNamingStrategy implements TableNamingStrategy {
```
rk3rn3r, line 21: @jpechane @Naros I have added another commit to re-add the old API, mapping to the new implementation. Including test scenarios for using manually provided `DefaultTableNamingStrategy` as config option and make derived implementations work. Wdyt?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/naming/TableNamingStrategy.java
```diff
@@ -0,0 +1,37 @@
...
+import io.debezium.connector.jdbc.JdbcSinkConnectorConfig;
+import io.debezium.sink.SinkConnectorConfig;
+import io.debezium.sink.naming.CollectionNamingStrategy;
+
+/**
+ * A pluggable strategy contract for defining how table names are resolved from kafka records.
+ *
+ * @author Chris Cranford
+ */
+public interface TableNamingStrategy extends CollectionNamingStrategy {
```
rk3rn3r, line 22: @jpechane @Naros I have added another commit to re-add the old API, mapping to the new implementation. Wdyt?

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/CollectionNamingStrategyTest.java
```diff
@@ -28,60 +29,64 @@
...
+        assertThat(strategy.resolveCollectionName(RECORD_FACTORY.createRecord("database.schema.table"), config.getCollectionNameFormat()))
+                .isEqualTo("kafkadep_database_schema_table");
+    }
+
+    @Test
+    public void testDeprecatedDefaultTableNamingStrategyWithTableNameFormat() {
+        final JdbcSinkConnectorConfig config = new JdbcSinkConnectorConfig(Map.of(JdbcSinkConnectorConfig.DEPRECATED_TABLE_NAME_FORMAT, "kafkadep_${topic}"));
+        final DefaultCollectionNamingStrategy strategy = new DefaultTableNamingStrategy();
+        assertThat(strategy.resolveCollectionName(RECORD_FACTORY.createRecord("database.schema.table"), config.getCollectionNameFormat()))
                 .isEqualTo("kafkadep_database_schema_table");
```
rk3rn3r, line 63: @mfvitale You are right! I have added this and another test for checking the deprecated options work as expected!

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/CollectionNamingStrategyTest.java
```diff
@@ -28,60 +29,64 @@
...
         final JdbcSinkConnectorConfig config = new JdbcSinkConnectorConfig(Map.of(JdbcSinkConnectorConfig.DEPRECATED_TABLE_NAME_FORMAT, "kafkadep_${topic}"));
-        final SinkRecordFactory factory = new DebeziumSinkRecordFactory();
         final DefaultCollectionNamingStrategy strategy = new DefaultCollectionNamingStrategy();
-        assertThat(strategy.resolveCollectionName(factory.createRecord("database.schema.table"), config.getCollectionNameFormat()))
+        assertThat(strategy.resolveCollectionName(RECORD_FACTORY.createRecord("database.schema.table"), config.getCollectionNameFormat()))
+                .isEqualTo("kafkadep_database_schema_table");
+    }
+
+    @Test
+    public void testDeprecatedDefaultTableNamingStrategyWithTableNameFormat() {
```
rk3rn3r, line 59: @mfvitale You are right! I have added this and another test for checking the deprecated options work as expected! This test uses the deprecated option AND the deprecated value!

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfigTest.java
```diff
@@ -149,6 +155,32 @@ public void testPrimaryKeyRecordValueDoesNotRequirePrimaryKeyFields() {
         assertThat(config.validateAndRecord(List.of(JdbcSinkConnectorConfig.INSERT_MODE_FIELD), LOGGER::error)).isTrue();
     }
 
+    @Test
+    public void testDeprecatedTableNamingStrategy() {
```
rk3rn3r, line 159: @mfvitale @jpechane @Naros I have added a special test case here to check the deprecate behavior is working as expected, returns the proper deprecation messages and the fallback behavior is working in all directions.
let me know what you think!

mfvitale: Now we can say that it works without relying on hope :wink:

rk3rn3r: I would not go that far because the previous tests were already using the old config options and were still working like expected. After re-adding the old API which is now calling the new implementation it made sense to add this definitely more conscious test. Agreed so far! ;)

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/naming/TableNamingStrategy.java
```diff
@@ -0,0 +1,38 @@
...
+     * Resolves the logical table name from the sink record.
+     *
+     * @param config sink connector configuration, should not be {@code null}
+     * @param record Kafka sink record, should not be {@code null}
+     * @return the resolved logical table name; if {@code null} the record should not be processed
+     */
+    String resolveTableName(JdbcSinkConnectorConfig config, SinkRecord record);
+
+    default String resolveCollectionName(KafkaDebeziumSinkRecord record, String collectionNameFormat) {
+        var config = new JdbcSinkConnectorConfig(Map.of(SinkConnectorConfig.COLLECTION_NAME_FORMAT, collectionNameFormat));
```
Naros, line 35: This is close, but I worry about only passing a partial config object.

For 3.0.x, ideally, we should settle for `resolveCollectionName` with a signature that passes both the `KafkaDebeziumSinkRecord` and `SinkConnectorConfig` for now. In the default method here, we can cast it to the `JdbcSinkConnectorConfig`, and we end up with the same behavior without any potential surprises.

Then, in 3.1, we can consider reworking the signature of `CollectionNamingStrategy` only to pass the format instead if we believe we want to restrict the details the user implementation has access to.

In this case, I would mark `CollectionNamingStrategy` with the `@Incubating` annotation to identify that we can will be changing this contract in the future and that users should expect changes.

Naros (commented):
Hi @rk3rn3r this is close, just one inline observation.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/naming/TableNamingStrategy.java
rk3rn3r: I have added now a `TemporaryBackwardCompatibleCollectionNamingStrategyProxy` to handle the fallback properly. wdyt?

jpechane: @rk3rn3r Well, that's what I call self-descriptive naming :-D

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -528,49 +359,45 @@ public String getValue() {
...
         this.truncateEnabled = config.getBoolean(TRUNCATE_ENABLED_FIELD);
-        this.tableNameFormat = config.getString(TABLE_NAME_FORMAT_FIELD);
+        this.collectionNameFormat = config.getString(COLLECTION_NAME_FORMAT_FIELD);
         this.primaryKeyMode = PrimaryKeyMode.parse(config.getString(PRIMARY_KEY_MODE_FIELD));
         this.primaryKeyFields = Strings.setOf(config.getString(PRIMARY_KEY_FIELDS_FIELD), String::new);
         this.schemaEvolutionMode = SchemaEvolutionMode.parse(config.getString(SCHEMA_EVOLUTION));
         this.quoteIdentifiers = config.getBoolean(QUOTE_IDENTIFIERS_FIELD);
-        // this.dataTypeMapping = Strings.setOf(config.getString(DATA_TYPE_MAPPING_FIELD), String::new);
-        this.tableNamingStrategy = config.getInstance(TABLE_NAMING_STRATEGY_FIELD, TableNamingStrategy.class);
+        this.collectionNamingStrategy = new TemporaryBackwardCompatibleCollectionNamingStrategyProxy(
```
rk3rn3r, line 388: The proxy is simply used here.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/naming/TableNamingStrategy.java
Naros: Agreed, but definitely a more cleaner fix, bravo :tada:

jpechane:
@rk3rn3r Applied, thanks a lot!

PR #5978, opened 2024-11-06
DBZ-8386 Consume all records before starting actual snapshot test

vjuranek:
This seems to be more reliable then waiting more or less random time and assuming that all transaction were already processed by the database.

https://issues.redhat.com/browse/DBZ-8386


vjuranek:
DB2 fails with
```
Error:  Errors: 
Error:    Db2ConnectionIT.shouldEnableCdcForDatabase:24 » SQL ASNCAP server did not start.
```
which is unrelated.

PR #5979, opened 2024-11-07
DBZ-8392 Allow for non-historized schemas to dispatch schema change events

twthorn:
Allow non-historized connectors (eg vitess, postgres) to handle schema change events correctly. Change  EventDisptacher to check in case the historized schema may be null (otherwise these fail trivially with NullPointerException when trying to publish schema change events).

This is needed to support vitess-connector to emit DDL events [DBZ-8325](https://issues.redhat.com/browse/DBZ-8325). It may also be useful for postgres connector as mentioned [here](https://github.com/debezium/debezium-connector-vitess/pull/210#pullrequestreview-2410041912).

### DIFF
--- a/debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
+++ b/debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
@@ -1468,12 +1468,24 @@ public abstract class CommonConnectorConfig {
         return eventConvertingFailureHandlingMode;
     }
 
+    public static final Field INCLUDE_SCHEMA_CHANGES = Field.create("include.schema.changes")
+            .withDisplayName("Include database schema changes")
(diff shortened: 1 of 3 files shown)


twthorn:
The mongodb test failure appears unrelated and probably flakey:
```
2024-11-07T20:43:49.9335356Z [ERROR]   MongoDbConnectorIT>AbstractMongoConnectorIT.afterEach:76->AbstractConnectorTest.stopConnector:152->AbstractConnectorTest.stopConnector:167 » IllegalState Cannot stop engine while tasks are starting, this may lead to leaked resource. Wait for the tasks to be fully started.
```

twthorn:
@jpechane new PR is ready for review

jpechane:
@twthorn Applied, thanks.

PR #5983, opened 2024-11-11
DBZ-8397 Make sure connector is fully running before stopping it

vjuranek:
https://issues.redhat.com/browse/DBZ-8397


jpechane:
@vjuranek Applied, thanks

PR #5984, opened 2024-11-11
DBZ-8399 Insert key with null when objectIdToTableId finds no matches

Naros:
https://issues.redhat.com/browse/DBZ-8399


Naros:

jpechane:
@Naros Applied, thanks

PR #5987, opened 2024-11-12
Update install.adoc

kavyaramaiah1991:


jpechane:
@kavyaramaiah1991 Applied via https://github.com/debezium/debezium/pull/5988, thanks for spotting!

PR #5993, opened 2024-11-15
DBZ-8409 Update Quarkus Outbox Extension to Quarkus 3.16.3

Naros:
https://issues.redhat.com/browse/DBZ-8409


jpechane:
@Naros Applied, thanks

PR #5995, opened 2024-11-15
DBZ-8414 Catch exception from stopping the connector

vjuranek:
Otherwise engine shutdown may get stuck as in such case the count down
of `shutDownLatch` would be skipped and engine would wait for `shutDownLatch` forever.

https://issues.redhat.com/browse/DBZ-8414

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEmbeddedEngine.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEmbeddedEngine.java
@@ -336,7 +336,7 @@ public final class AsyncEmbeddedEngine<R> implements DebeziumEngine<R>, AsyncEng
      */
     private void finishShutDown(Throwable exitError) {
         try {
-            shutDownLatch.await();
+            shutDownLatch.await(config.getLong(AsyncEngineConfig.TASK_MANAGEMENT_TIMEOUT_MS), TimeUnit.MILLISECONDS);
         }
         catch (InterruptedException e) {
             LOGGER.warn("Interrupted while waiting for shutdown to finish.");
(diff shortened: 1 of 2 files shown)


vjuranek:
DBZ server constantly fails with
```
 Error:  Failed to execute goal on project debezium-system-tests: Could not resolve dependencies for project io.debezium:debezium-system-tests:jar:3.0.3-SNAPSHOT
Error:  dependency: io.debezium:debezium-connector-db2:jar:3.0.3-SNAPSHOT (test)
Error:  	io.debezium:debezium-connector-db2:jar:3.0.3-SNAPSHOT was not found in https://s01.oss.sonatype.org/content/repositories/snapshots during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of oss has elapsed or updates are forced
Error:  	io.debezium:debezium-connector-db2:jar:3.0.3-SNAPSHOT was not found in https://packages.confluent.io/maven/ during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of confluent has elapsed or updates are forced
Error:  	io.debezium:debezium-connector-db2:jar:3.0.3-SNAPSHOT was not found in https://s01.oss.sonatype.org//content/groups/public/ during a previous attempt. This failure was cached in the local repository and resolution is not reattempted until the update interval of ossrh has elapsed or updates are forced
Error:  -> [Help 1]
Error:  
```
No idea what is the root cause, probably some issue in GH infrastructure or something?
I can download these bits locally without any issue, build of DBZ server locally passes as well (and of course I can see the bits in https://s01.oss.sonatype.org/content/repositories/snapshots/io/debezium/debezium-connector-db2/3.0.3-SNAPSHOT/)

jpechane:
@vjuranek Applied, thanks. The missed test is not affected by the change.

PR #5996, opened 2024-11-16
DBZ-8302: Treat oracle negative timestamp values as -infinity.

dongwook-chan:
We were hoping to create a PR on this issue but found that the repo is archived.
Below is the PR that we attempted to create.
 
We want to refer to config from ZonedTimestampType.
But it takes a lot of depth to reach config from this class.
So we would you to provide some advice on this issue pls.
https://github.com/dongwook-chan/debezium-connector-jdbc/blob/2e2aa669c2a4f50ba85196d5cf8f4115a5155c28/src/main/java/io/debezium/connector/jdbc/dialect/mysql/ZonedTimestampType.java#L37-L39

We tried to run integration test on jdbc sink connector but failed.
So if you have any instructions on this please provide.

### DIFF
--- a/debezium-connector-jdbc/pom.xml
+++ b/debezium-connector-jdbc/pom.xml
@@ -202,6 +202,7 @@
         <dependency>
             <groupId>com.oracle.database.jdbc</groupId>
             <artifactId>ojdbc11</artifactId>
+            <version>23.6.0.24.10</version>
         </dependency>
     </dependencies>
 
--- a/debezium-connector-jdbc/pom.xml
+++ b/debezium-connector-jdbc/pom.xml
@@ -202,7 +202,6 @@
         <dependency>
             <groupId>com.oracle.database.jdbc</groupId>
             <artifactId>ojdbc11</artifactId>
-            <version>23.6.0.24.10</version>
         </dependency>
     </dependencies>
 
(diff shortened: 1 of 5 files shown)


debezium-connector-jdbc/pom.xml
```diff
@@ -202,6 +202,7 @@
         <dependency>
             <groupId>com.oracle.database.jdbc</groupId>
             <artifactId>ojdbc11</artifactId>
+            <version>23.6.0.24.10</version>
```
jpechane, line 205: Why it is necessary to override the version here?

jpechane (changes requested):
@dongwook-chan Thanks for the PR. Could you please reformat the code so the tests are executed. lso could you please add a test case for the new feature? Thanks

Naros (commented):
Hi @dongwook-chan do you intend to add test cases?

debezium-connector-jdbc/pom.xml
dongwook-chan: Sorry, my bad! I removed the version tag.

dongwook-chan: @jpechane 
I also added config variable `ORACLE_INFINITY_CONVERTER_ENABLE` for the feature,
but have trouble accessing it from [convertInfinityTimestampValue()](https://github.com/dongwook-chan/debezium/blob/83b6c01558ad883633c443c4b42f248bd8ad53ec/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/ZonedTimestampType.java#L40-L54). Could you give some advice on this? And I would like some advice on running integration test too.

jpechane: I think this would require a different approach. The class `io.debezium.connector.jdbc.dialect.oracle.ZonedTimestampType` should provide two instances. The existing one and one with the requested behaviour. `io.debezium.connector.jdbc.dialect.oracle.OracleDatabaseDialect.registerTypes()` wil then register the one based on connector config.

Naros:

Hi @dongwook-chan, please sign off all commits with:

```
git commit -s
```

Naros:
Superseded with a slightly different design approach in #7878.

PR #5997, opened 2024-11-19
enable handling tombstones

msillence:


jpechane:
@vjuranek @msillence Please take a look. Feel free to approve/merge if/when you are happy with it.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/SinkRecordDescriptor.java
```diff
@@ -154,6 +157,9 @@ public Struct getKeyStruct(PrimaryKeyMode primaryKeyMode) {
     }
 
     public Struct getAfterStruct() {
+        if (isTombstone()) {
+            return ((Struct) record.key());
```
rk3rn3r, line 161: Doing this here could lead to potentially issues in the non-tombstone cases, don't you think?

msillence: I can see your concern this is generic code that is used in other places, though if this is a non-tombstone event this `if` condition will be skipped and the original path will be executed. If this is a tombstone then the there is no value struct as that's part of the isTombstone test so we would without this code return null.
I also checked and there's no after structure when we are dealing with a key. I think this is safe from my testing

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
```diff
@@ -68,6 +68,26 @@ public JdbcChangeEventSink(JdbcSinkConnectorConfig config, StatelessSession sess
         final DatabaseVersion version = this.dialect.getVersion();
         LOGGER.info("Database version {}.{}.{}", version.getMajor(), version.getMinor(), version.getMicro());
     }
+    
+    public boolean processDelete(JdbcSinkRecord record, final CollectionId collectionId, 
+    		final Map<CollectionId, Buffer> upsertBufferByTable, final Map<CollectionId, Buffer> deleteBufferByTable) {
+        if (!config.isDeleteEnabled()) {
```
mfvitale, line 74: I would put this out avoiding the return value since, as the method name suggests, it will process the delete.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/SinkRecordDescriptor.java
msillence: this is no longer required

msillence:
OK it turns out the maven formatter plugin was preventing me from building mvn 3.9.9 java 21
now I can build it I can confirm it all works and I've fixed the formatting.
The fix is much cleaner in trunk thanks to the changes to the sql delete path using the key fields already.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
```diff
@@ -95,8 +95,26 @@ public void execute(Collection<SinkRecord> records) {
             final TableId tableId = optionalTableId.get();
 
             if (sinkRecordDescriptor.isTombstone()) {
-                // Skip only Debezium Envelope tombstone not the one produced by ExtractNewRecordState SMT
-                LOGGER.debug("Skipping tombstone record {}", sinkRecordDescriptor);
+                LOGGER.debug("processing tombstone record {}", sinkRecordDescriptor);
+
+                if (!config.isDeleteEnabled()) {
```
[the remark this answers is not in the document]
rk3rn3r: thx for that cleanup and extracting to a method call!

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
```diff
@@ -88,14 +102,20 @@ public void execute(Collection<SinkRecord> records) {
...
             }
 
+            final CollectionId collectionId = optionalCollectionId.get();
+
             if (record.isTombstone()) {
-                // Skip only Debezium Envelope tombstone not the one produced by ExtractNewRecordState SMT
-                LOGGER.debug("Skipping tombstone record {}", record);
+                LOGGER.debug("processing tombstone record {}", record);
+
+                if (!config.isDeleteEnabled()) {
```
rk3rn3r, line 110: @jpechane @Naros I am currently hitting the question why deletes are disabled by default?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
```diff
@@ -88,14 +102,20 @@ public void execute(Collection<SinkRecord> records) {
                 continue;
             }
 
+            final CollectionId collectionId = optionalCollectionId.get();
+
             if (record.isTombstone()) {
```
rk3rn3r, line 107: @msillence Should we maybe move the below code for `if (record.isDelete()) { ... }` up here? `record.isDelete()` covers the case of tombstones and Debezium delete change event records. The code below can be deleted.

Mind that this new behavior will cause 2 deletes executed because `tombstones.on.delete` in source connectors and even with the deprecated `delete.handling​.mode`  option in ExtractNewRecordState SMT will cause 2 messages for a single delete event happening in the source datastore (one Debezium DELETE change event and one tombstone), depending on those configs. The implementation I am working on will de-duplicate those events so this is only temporary an issue. The reason for the previous code was that we avoided duplicate delete events for the same row/source event. But it excludes deletes for source connectors using ExtractNewRecordState SMT that drops the Debezium DELETE events and keeps tombstones (which might be semantically the best way to represent deletes in Kafka imho).

@jpechane As this code is moved to the AbstractChangeEventSink eventually with my current work, we should maybe evaluate if we close this PR as those lines will be dropped soon?

mfvitale:
Hi @msillence It is not clear to me how this is going to work. 

I have the feeling that with the `primary.key.mode` set to `record_value` a tombstone event from Debezium - both value and valueSchema to `null` - will lead to an error. 

Can you please  add some tests (in `AbstractJdbcSinkDeleteEnabledTest.java`) to cover the following cases?

1. `primary.key.mode=record_key` and a Debezium tombstone 
2.  `primary.key.mode=record_key` and a ExtractNewRecordState tombstone - only value null
3. `primary.key.mode=record_value` and a Debezium tombstone
4. `primary.key.mode=record_value` and a ExtractNewRecordState tombstone - only value null

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
rk3rn3r: @msillence After writing the above, maybe we can just drop those lines here for `if (record.isTombstone()) {...}` entirely and keep the  `if (record.isDelete()) { ... }` below as it handles tombstones and Debezium DELETE change event messages, yielding the previously mentioned duplicate deletes issued to the destination datastore. The idea was to skip those duplicates delete events in the default scenarion but excludes for certain ExtractNewRecordState SMT configs (which imho are even syntactically and semantically good and correct)!

Naros:

Hi @msillence, please sign off all commits with:

```
git commit -s
```

Naros:
I suspect we can close this PR, but before I do, @rk3rn3r @msillence @mfvitale @jpechane any opinion?

Naros:
Ok, I'll close. We can always reopen it if needed; same for the associated issue.

PR #5998, opened 2024-11-19
Implements sequence method for SourceInfo in Oracle connector

JuanmaBM:
This PR implemented the following feature request: https://issues.redhat.com/projects/DBZ/issues/DBZ-4489?filter=allopenissues

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/SourceInfo.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/SourceInfo.java
@@ -6,6 +6,7 @@
 package io.debezium.connector.oracle;
 
 import java.time.Instant;
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.Set;
@@ -187,5 +188,14 @@ public class SourceInfo extends BaseSourceInfo {
     @Override
     protected String database() {
         return (tableIds != null) ? tableIds.iterator().next().catalog() : null;
+    }
+
+    @Override
+    protected String sequence() {
+        return Arrays.asList(
+            this.scn,
+            this.sourceTime != null ? this.sourceTime.toEpochMilli() : null
+        )
+        .toString();
     }
 }
--- a/debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/SourceInfoTest.java
+++ b/debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/SourceInfoTest.java
@@ -7,6 +7,7 @@ package io.debezium.connector.oracle;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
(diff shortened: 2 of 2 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/SourceInfo.java
```diff
@@ -188,4 +189,13 @@ protected Instant timestamp() {
...
     }
+
+    @Override
+    protected String sequence() {
+        return Arrays.asList(
+            this.scn,
+            this.sourceTime != null ? this.sourceTime.toEpochMilli() : null
+        )
+        .toString();
+    }
```
Naros, line 200: Oracle's change time is always second-based precision, so I don't believe it will help identify the order of transactions if they're committed within proximity. 

Is there a reason why the transaction metadata topic cannot be used to resolve the order for post-processing?

Naros: Hi @JuanmaBM, in all honesty, you can't rely on SCN either because it's not unique. Oracle can assign up to 16k change vectors to the same SCN on Oracle 12 R2 and 96k on Oracle 19+. This means one or more transactions can share the same SCN.

The Oracle transaction identifier is also unsuitable for ordering since it is a composite value derived from 3 rollback segment attributes, which, when combined, are always unique.  The transaction identifier does not monotonically increase either.

One way to achieve this would be to write all table rows to the same topic, where this topic has a single partition. This can be a bottleneck in high-volume traffic scenarios, but you get the ordering guarantee automatically out of the box. In short, the Kafka topic will be chronologically ordered by commits.

Another approach would be to use KStreams to consume the transaction metadata topic, using the transaction commit order from that topic to enrich and write all table-level topic events to a new topic that has a single partition and would be chronologically ordered by commit. In short, you end up with the same type of topic as the above, except you don't take the performance bottleneck of the single partition while producing the data to Kafka; you take it during the post-processing.

The final option would be to consume the transaction metadata and table-level topics as-is and reorder the changes by transaction in Spark as part of the data analysis.

JuanmaBM: Hello @Naros 

So, if you agree, I think I will close this PR and add this information to the issue to document these approaches. This way, if anyone encounters the same problems, they can use or test them as a solution.

jpechane: @JuanmaBM Chris is on PTO this week. The documentation sounds like a good plan!

JuanmaBM: @jpechane Oh, I see. Great! I'll go ahead and add the information to the issue, then close the PR. Thanks!

PR #5999, opened 2024-11-19
DBZ-8429: fix install docs

blcksrx:
There is a typo error in installation docs that corrupted download link for `informix` and `jdbc` connector on the documentation online!


Naros:
Thanks @blcksrx applied to main (=3.0) and branch 2.7.

PR #6000, opened 2024-11-19
DBZ-8430 Use list of columns in incremental snapshot select stmt

wukachn:
https://issues.redhat.com/browse/DBZ-8430

https://debezium.zulipchat.com/#narrow/channel/302529-community-general/topic/Incremental.20Snapshots.20-.20Hidden.20Columns

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
@@ -81,15 +81,11 @@ public abstract class AbstractChunkQueryBuilder<T extends DataCollectionId>
     }
 
     protected String buildProjection(Table table) {
-        String projection = "*";
-        if (connectorConfig.isColumnsFiltered()) {
-            TableId tableId = table.id();
-            projection = table.columns().stream()
-                    .filter(column -> columnFilter.matches(tableId.catalog(), tableId.schema(), tableId.table(), column.name()))
-                    .map(column -> jdbcConnection.quotedColumnIdString(column.name()))
-                    .collect(Collectors.joining(", "));
-        }
-        return projection;
+        TableId tableId = table.id();
+        return table.columns().stream()
+                .filter(column -> !connectorConfig.isColumnsFiltered() || columnFilter.matches(tableId.catalog(), tableId.schema(), tableId.table(), column.name()))
+                .map(column -> jdbcConnection.quotedColumnIdString(column.name()))
+                .collect(Collectors.joining(", "));
     }
 
     protected void addLowerBound(IncrementalSnapshotContext<T> context, Table table, Object[] boundaryKey, StringBuilder sql) {
--- a/debezium-core/src/test/java/io/debezium/pipeline/source/snapshot/incremental/DefaultChunkQueryBuilderTest.java
+++ b/debezium-core/src/test/java/io/debezium/pipeline/source/snapshot/incremental/DefaultChunkQueryBuilderTest.java
@@ -95,11 +95,12 @@ public class DefaultChunkQueryBuilderTest {
                 .addColumn(val1)
                 .addColumn(val2)
                 .setPrimaryKeyNames("pk1").create();
-        assertThat(chunkQueryBuilder.buildChunkQuery(context, table, Optional.empty())).isEqualTo("SELECT * FROM \"s1\".\"table1\" ORDER BY \"pk1\" LIMIT 1024");
+        assertThat(chunkQueryBuilder.buildChunkQuery(context, table, Optional.empty())).isEqualTo(
+                "SELECT \"pk1\", \"val1\", \"val2\" FROM \"s1\".\"table1\" ORDER BY \"pk1\" LIMIT 1024");
         context.nextChunkPosition(new Object[]{ 1 });
         context.maximumKey(new Object[]{ 10 });
         assertThat(chunkQueryBuilder.buildChunkQuery(context, table, Optional.empty())).isEqualTo(
-                "SELECT * FROM \"s1\".\"table1\" WHERE (\"pk1\" > ?) AND NOT (\"pk1\" > ?) ORDER BY \"pk1\" LIMIT 1024");
+                "SELECT \"pk1\", \"val1\", \"val2\" FROM \"s1\".\"table1\" WHERE (\"pk1\" > ?) AND NOT (\"pk1\" > ?) ORDER BY \"pk1\" LIMIT 1024");
     }
 
     @Test
@@ -116,11 +117,11 @@ public class DefaultChunkQueryBuilderTest {
                 .addColumn(val2)
                 .setPrimaryKeyNames("pk1").create();
         assertThat(chunkQueryBuilder.buildChunkQuery(context, table, Optional.of("\"val1\"=foo")))
-                .isEqualTo("SELECT * FROM \"s1\".\"table1\" WHERE \"val1\"=foo ORDER BY \"pk1\" LIMIT 1024");
+                .isEqualTo("SELECT \"pk1\", \"val1\", \"val2\" FROM \"s1\".\"table1\" WHERE \"val1\"=foo ORDER BY \"pk1\" LIMIT 1024");
         context.nextChunkPosition(new Object[]{ 1 });
         context.maximumKey(new Object[]{ 10 });
         assertThat(chunkQueryBuilder.buildChunkQuery(context, table, Optional.of("\"val1\"=foo"))).isEqualTo(
-                "SELECT * FROM \"s1\".\"table1\" WHERE (\"pk1\" > ?) AND NOT (\"pk1\" > ?) AND \"val1\"=foo ORDER BY \"pk1\" LIMIT 1024");
+                "SELECT \"pk1\", \"val1\", \"val2\" FROM \"s1\".\"table1\" WHERE (\"pk1\" > ?) AND NOT (\"pk1\" > ?) AND \"val1\"=foo ORDER BY \"pk1\" LIMIT 1024");
     }
 
     @Test
@@ -140,11 +141,11 @@ public class DefaultChunkQueryBuilderTest {
                 .setPrimaryKeyNames("pk1", "pk2").create();
         context.addDataCollectionNamesToSnapshot("12345", List.of(table.id().toString()), List.of(), "pk2");
         assertThat(chunkQueryBuilder.buildChunkQuery(context, table, Optional.of("\"val1\"=foo")))
-                .isEqualTo("SELECT * FROM \"s1\".\"table1\" WHERE \"val1\"=foo ORDER BY \"pk2\" LIMIT 1024");
+                .isEqualTo("SELECT \"pk1\", \"pk2\", \"val1\", \"val2\" FROM \"s1\".\"table1\" WHERE \"val1\"=foo ORDER BY \"pk2\" LIMIT 1024");
         context.nextChunkPosition(new Object[]{ 1 });
         context.maximumKey(new Object[]{ 10 });
         assertThat(chunkQueryBuilder.buildChunkQuery(context, table, Optional.of("\"val1\"=foo"))).isEqualTo(
-                "SELECT * FROM \"s1\".\"table1\" WHERE (\"pk2\" > ?) AND NOT (\"pk2\" > ?) AND \"val1\"=foo ORDER BY \"pk2\" LIMIT 1024");
+                "SELECT \"pk1\", \"pk2\", \"val1\", \"val2\" FROM \"s1\".\"table1\" WHERE (\"pk2\" > ?) AND NOT (\"pk2\" > ?) AND \"val1\"=foo ORDER BY \"pk2\" LIMIT 1024");
     }
 
     @Test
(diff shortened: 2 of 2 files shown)


Naros:
Hi @wukachn, it seems the test failures are related. Could you take a look?

wukachn:
Ill take a look at the ITs at some point today

wukachn:
Looks like the initial change has made MySQL schema changes not supported at all during an incremental snapshot. If thats true then schema changes must have never been supported during a MySQL incremental snapshot if the user was using column filters. @Naros was the team aware of this nuance?

I haven't looked too deep into how to get it working again but I imagine it'd take some time for me to gain the confidence to make a meaningful change. Im happy to take suggestions from someone who has more knowledge of the codebase.

nathan-smit-1:
hey @wukachn, I've been tracking this issue as it relates to a problem we had in Prod for Mysql.  I think you may be right and that incremental snapshot has possibly just never worked with a column filter applied for some of the sources.  As far as I can tell there isn't a test in place for incremental snapshots with column filters applied.

I will ask about it in Zulip as well as I guess that would be unrelated to the specific issue of handling invisible columns but if I set column.include.list to include a subset of columns for my table and then run an incremental or blocking snapshot I get this warning:  {"timestamp":"2024-11-24T13:36:04.785293758+02:00","sequence":991,"loggerClassName":"org.slf4j.impl.Slf4jLogger","loggerName":"io.debezium.config.CommonConnectorConfig","level":"WARN","message":"The signal event 'Struct{}' should have 3 fields but has 0"

This is coming from the below code it seems in CommonConnectorConfig.  It would be cool if you could sanity check if you see the same behaviour in 3.0.1 for sql server.  I'm also not sure why that would cause the schema change to time out though which is what I saw when I ran tests using your code.

`    public Optional<String[]> parseSignallingMessage(Struct value, String fieldName) {
        final Struct event = value.getStruct(fieldName);
        if (event == null) {
            LOGGER.warn("Field {} part of signal '{}' is missing", fieldName, value);
            return Optional.empty();
        }
        List<org.apache.kafka.connect.data.Field> fields = event.schema().fields();
        if (fields.size() != 3) {
            LOGGER.warn("The signal event '{}' should have 3 fields but has {}", event, fields.size());
            return Optional.empty();
        }
        return Optional.of(new String[]{
                event.getString(fields.get(0).name()),
                event.getString(fields.get(1).name()),
                event.getString(fields.get(2).name())
        });
    }`

mfvitale:
> Looks like the initial change has made MySQL schema changes not supported at all during an incremental snapshot. If thats true then schema changes must have never been supported during a MySQL incremental snapshot if the user was using column filters. @Naros was the team aware of this nuance?

I tried to set the column filters in the `schemaChanges` test and it works with the old version. 

With you changes it works if you specify the column filters and not work if you not specify it. 

The scenario is

1. The table has initially two column: a,b
2. Some record inserted
3. Then the table will be modified with a third column: c
4. Some record will be inserted 
5. The column `c` will be removed
6. Some record inserted

So during the incremental snapshot can happen that when it is processing some chunk an the last table definition was with the column `c` but then the table has no more the `c` column. 

This before was not happening since the `*` always match the current table columns. 

@Naros @jpechane any idea?

mfvitale:
Maybe the only solution here is that when a table has hidden field, the `column.include.list` must be specified. WDYT?

nathan-smit-1:
@mfvitale I asked what I guess is a related question here:  https://debezium.zulipchat.com/#narrow/channel/348104-community-mysql-mariadb/topic/Expected.20behaviour.20of.20column.2Einclude.2Elist.20for.20multiple.20table

Basically just trying to understand the intended behaviour of the column filter as it appears to apply globally even to the debezium_signal table (i.e. in RelationalDatabaseSchema it excludes the signal table columns and you can't then read signals from it) and so if you specify include columns you need to make sure to include every column across all of your tables which is is not ideal

nathan-smit-1:
> What I think is wrong here is that we are using the select statement for chunk query also for read the table in `readSchema`.

Hey @mfvitale, I just wanted to follow up on what you were saying above as I've been tracking this issue.  Is there any downside to just having always having the select statement for readSchema be select * from table with conditions?  

Not sure what the least hacky way to do that would be though.  Maybe a new method so instead of calling buildChunkQuery in readSchema you call a new buildSchemaQuery  method or something like that so you can have something more similar to the old logic where you select * unless there is include/exclude column list?

Naros:
> So during the incremental snapshot can happen that when it is processing some chunk an the last table definition was with the column `c` but then the table has no more the `c` column.

There is a window between the schema change validation step and the execution of the query where I could imagine this error happening, and we should likely guard against this race condition. The question is whether or not we can efficiently guard against this :thinking: . 

> What I think is wrong here is that we are using the select statement for chunk query also for read the table in `readSchema`.

I believe we need both, its what we do for the initial snapshots I do believe, right?

Naros:
> Basically just trying to understand the intended behaviour of the column filter as it appears to apply globally even to the debezium_signal table (i.e. in RelationalDatabaseSchema it excludes the signal table columns and you can't then read signals from it) and so if you specify include columns you need to make sure to include every column across all of your tables which is is not ideal

In that case why not use `column.exclude.list` and remove the opposing data set @nathan-smit-1 ?

Naros:
> Not sure what the least hacky way to do that would be though. Maybe a new method so instead of calling buildChunkQuery in readSchema you call a new buildSchemaQuery method or something like that so you can have something more similar to the old logic where you select * unless there is include/exclude column list?

I believe the idea @nathan-smit-1 was to always apply a list of columns. For example in the initial snapshot, if we find the user supplies a `SELECT *` based override or there is no override, we turn that into a `SELECT <explicit list of columns>` query. It's my understanding that were attempting to mimic that exactly.

nathan-smit-1:
Although thinking about it now, the above would pass the tests but then likely not actually work for the invisible columns :(

So would need some other creative way, but I think conceptually having readSchema have its own logic could work.  Or maybe assume the schema changed based on the readSchema failure and have verifySchemaUnchanged handle things a little differently?  It's a tricky one.

nathan-smit-1:
> > Not sure what the least hacky way to do that would be though. Maybe a new method so instead of calling buildChunkQuery in readSchema you call a new buildSchemaQuery method or something like that so you can have something more similar to the old logic where you select * unless there is include/exclude column list?
> 
> I believe the idea @nathan-smit-1 was to always apply a list of columns. For example in the initial snapshot, if we find the user supplies a `SELECT *` based override or there is no override, we turn that into a `SELECT <explicit list of columns>` query. It's my understanding that were attempting to mimic that exactly.

@Naros Apologies, ended up deleting some responses for clarity.  As I understand the issue, when a column is dropped and we explicitly list out the columns readSchema fails because it does select a, b, c from table to generate the schema but [c] no longer exists.

I was just thinking now though that I recently added readSchemaForTable as part of [DBZ-4903](https://github.com/debezium/debezium/commit/a79aa0ba435aff46a143ca42728887b9ceffe662) and I'm wondering now if that can't be adapted to do the schema read instead?  I'm busy running some tests at the moment.

mfvitale:
> > What I think is wrong here is that we are using the select statement for chunk query also for read the table in `readSchema`.
> 
> I believe we need both, its what we do for the initial snapshots I do believe, right?

The main difference with the initial snapshot is that we lock tables while reading the schema.

nathan-smit-1:
@mfvitale @Naros   I don't wanna monopolise discussion to much on this as it's not my pull request, but FWIW, I did find that having a fallback like the below in the case of sql exception did pass the Mysql tests at least and does then allow snapshots on invisible columns.  The below is also kinda what I was referring to as a "hacky" solution though (i.e. checking the current schema in event of sql exception).  Definitely a problem that is much trickier than it initially seemed.

 ```
   private Table readSchema() {
        final String selectStatement = chunkQueryBuilder.buildChunkQuery(context, currentTable, 0, Optional.empty());
        LOGGER.debug("Reading schema for table '{}' using select statement: '{}'", currentTable.id(), selectStatement);

        try (PreparedStatement statement = chunkQueryBuilder.readTableChunkStatement(context, currentTable, selectStatement);
                ResultSet rs = statement.executeQuery()) {
            return getTable(rs);
        }
        catch (SQLException initialException) {
            LOGGER.warn("Initial schema read failed. Attempting alternative schema retrieval.", initialException);

            try {
                // Try reading schema using the alternative method
                Table alternativeTable = readSchemaForTable(currentTable.id());

                // Retry the original query with the alternative table
                final String retrySelectStatement = chunkQueryBuilder.buildChunkQuery(context, alternativeTable, 0, Optional.empty());
                try (PreparedStatement retryStatement = chunkQueryBuilder.readTableChunkStatement(context, alternativeTable, retrySelectStatement);
                        ResultSet retryRs = retryStatement.executeQuery()) {
                    return getTable(retryRs);
                }
            }
            catch (SQLException secondaryException) {
                // If both attempts fail, throw a comprehensive exception
                LOGGER.error("Failed to read schema for table '{}' after multiple attempts", currentTable.id(), secondaryException);
                throw new DebeziumException("Snapshotting of table " + currentTable.id() + " failed after multiple attempts", initialException);
            }
        }
    }
```

mfvitale:
> @mfvitale @Naros I don't wanna monopolise discussion to much on this as it's not my pull request, but FWIW, I did find that having a fallback like the below in the case of sql exception did pass the Mysql tests at least and does then allow snapshots on invisible columns. The below is also kinda what I was referring to as a "hacky" solution though (i.e. checking the current schema in event of sql exception). Definitely a problem that is much trickier than it initially seemed.
> 
> ```
>   private Table readSchema() {
>        final String selectStatement = chunkQueryBuilder.buildChunkQuery(context, currentTable, 0, Optional.empty());
>        LOGGER.debug("Reading schema for table '{}' using select statement: '{}'", currentTable.id(), selectStatement);
> 
>        try (PreparedStatement statement = chunkQueryBuilder.readTableChunkStatement(context, currentTable, selectStatement);
>                ResultSet rs = statement.executeQuery()) {
>            return getTable(rs);
>        }
>        catch (SQLException initialException) {
>            LOGGER.warn("Initial schema read failed. Attempting alternative schema retrieval.", initialException);
> 
>            try {
>                // Try reading schema using the alternative method
>                Table alternativeTable = readSchemaForTable(currentTable.id());
> 
>                // Retry the original query with the alternative table
>                final String retrySelectStatement = chunkQueryBuilder.buildChunkQuery(context, alternativeTable, 0, Optional.empty());
>                try (PreparedStatement retryStatement = chunkQueryBuilder.readTableChunkStatement(context, alternativeTable, retrySelectStatement);
>                        ResultSet retryRs = retryStatement.executeQuery()) {
>                    return getTable(retryRs);
>                }
>            }
>            catch (SQLException secondaryException) {
>                // If both attempts fail, throw a comprehensive exception
>                LOGGER.error("Failed to read schema for table '{}' after multiple attempts", currentTable.id(), secondaryException);
>                throw new DebeziumException("Snapshotting of table " + currentTable.id() + " failed after multiple attempts", initialException);
>            }
>        }
>    }
> ```

@nathan-smit-1 the problem with the hack is that you can effectively have a table schema without the hidden column and then read the data with the hidden column. 

I added `needs more thinking` as the solution will not be easy here. If you want to propose your way, feel free to work on it.

nathan-smit-1:
@mfvitale that makes sense.  Just to put another idea out there:  What if this was configurable such that you could choose to have Debezium list the column but accept the risk that you can't have schema changes during the snapshot?  So the equivalent of setting [snapshot.locking.mode](https://debezium.io/documentation/reference/3.0/connectors/mysql.html#mysql-property-snapshot-locking-mode) to none

mfvitale:
> @mfvitale that makes sense. Just to put another idea out there: What if this was configurable such that you could choose to have Debezium list the column but accept the risk that you can't have schema changes during the snapshot? So the equivalent of setting [snapshot.locking.mode](https://debezium.io/documentation/reference/3.0/connectors/mysql.html#mysql-property-snapshot-locking-mode) to none

@nathan-smit-1 I think this is not going to works since its what is happening in the tests. Schema must match otherwise you will get  `Data row is smaller than a column index, internal schema representation is probably out of sync with real database schema`

Naros:
My apologies for circling back to this relatively late @mfvitale and @nathan-smit-1. 

So I came across a similar situation with Oracle recently during streaming where any `GENERATED ALWAYS` columns are excluded and not part of the change record; however, the relational table metadata tracks such columns.

This makes me wonder whether it makes more sense to consider whether hidden/generated columns should be part of the change event payload, particularly given that they're often not easily obtainable or even recorded in some vendor transaction logs.

mfvitale:
> So I came across a similar situation with Oracle recently during streaming where any `GENERATED ALWAYS` columns are excluded and not part of the transaction log's change record; however, the relational table metadata tracks such columns.

Well this is interesting. It would be good if we can put down somewhere (DDD?) all this findings so we can also discuss better on the solution. 

> This makes me wonder whether it makes more sense to consider whether hidden/generated columns should be part of the change event payload, particularly given that they're often not easily obtainable or even recorded in some vendor transaction logs.

In case of the generated, that are not in the transaction logs, it sounds good but I have some doubts on the hidden one. Think in uses case where you just want to do same database replication.

> Perhaps once we introduce connector-specific relational model implementations, that might help here.

Also here, do you think is there any change to in someway exclude or optionally exclude those columns while we get the table metadata?

Naros:
> In case of the generated, that are not in the transaction logs, it sounds good but I have some doubts on the hidden one. Think in uses case where you just want to do same database replication.

Absolutely, I think this highlights the fact it could be useful to introduce a visibility attribute or mechanism for tracking column visibility in the relational model, since I suspect other databases are similar to Oracle where virtual/generated is tracked independently of column's visibility setting for generalized queries.

> do you think is there any change to in someway exclude or optionally exclude those columns while we get the table metadata?

We could, but column position details would also need to change with what I suspect may be a database-specific hook.  In Oracle for example, the driver by default provides generated columns first, but is this somehow influenced by NULL ordering, where some databases you can set NULLs to be first or last. 

Whatever we do, we'd need to take it connector by connector. For MariaDB, MySQL, and Oracle, combining a code change in `JdbcConnection` and updating the ANTLR grammar where needed to capture & track this detail would probably suffice. I can't speak for SQL Server as I don't know that area of the code well. For PostgreSQL, we'd face the same dilemma we do today with optionality, using an out-of-band query during relational messages to refresh the schema, which may or may not find the table if it's been dropped.

wukachn:
Closing due to inactivity.

PR #6001, opened 2024-11-20
DBZ-8434 Don't pass null records to consumer

vjuranek:
If user SMT for whatever reason return `null`, don't pass such records to consumer as it woudl result into NPE in async engine chain. This is in concordance with old `EmbeddedEngine` which skipped `null` records as well.

https://issues.redhat.com/browse/DBZ-8434


jpechane:
@vjuranek Applied, thanks! This is correct behaviour as returning `null` from the transformation means that the record should not be processed.

PR #6004, opened 2024-11-21
DBZ-8441 Make AbstractConnectorTest#createEngine method abstract

vjuranek:
With this change, all `EmbeddedEngine` stuff which is used in the tests should be localted only in `EmbeddedEngineTest` and `TestingEmbeddedEngine` classes, i.e. used only for running tests for `EmbeddedEngine` itself.

https://issues.redhat.com/browse/DBZ-8441


jpechane:
@vjuranek Applied, thanks

PR #6006, opened 2024-11-21
[docs] Add common IntelliJ run configs to git

ismailsimsek:
Saving common IntelliJ run configurations for developer convenience. These will be shared across the dev team.

- related question: https://stackoverflow.com/a/24642246/2679740
- related guide: https://www.jetbrains.com/guide/javascript/tips/share-run-configurations/

### DIFF
--- a/.run/build-quick.run.xml
+++ b/.run/build-quick.run.xml
@@ -1,0 +1,34 @@
+<component name="ProjectRunConfigurationManager">
+  <configuration default="false" name="build -Dquick" type="MavenRunConfiguration" factoryName="Maven">
+    <MavenSettings>
+      <option name="myGeneralSettings" />
+      <option name="myRunnerSettings" />
+      <option name="myRunnerParameters">
+        <MavenRunnerParameters>
+          <option name="cmdOptions" />
+          <option name="profiles">
+            <set />
+          </option>
+          <option name="goals">
+            <list>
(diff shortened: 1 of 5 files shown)


jcechace:
Are the name of the files mandatory? In general I'm not a fan of spaces in file names :). Otherwise LGTM!

mfvitale:
The only cons is that when we have this folder committed, if some one adds more custom run configuration it will end up to be committed maybe involuntary. 

Should we add a particular prefix to these OOTB configuration and add an entry in `.gitignore` to exclude personal ones?

mfvitale (approved):
LGTM

jcechace:
> > Are the name of the files mandatory? In general I'm not a fan of spaces in file names :). Otherwise LGTM!
> 
> @jcechace i believe file names are ignored by intelij, so different naming is OK. do you have any suggestion? it should be ok to rename files.
> 
> @mfvitale The user specifix run configs are stored in different location. so no chance that it can get mixed up. user should explicitly click`Store as project file` to commit it as `.run` file. and since it will go trough PR review, i think all the guardrails are in place. IMO additional setting is not needed.

Nope, literary anything a bit descriptive will do :).

jpechane:
@ismailsimsek Applied, thanks

rk3rn3r:
Ouch

PR #6010, opened 2024-11-22
DBZ-8450 Extra list-continuation (`+`) chars corrupt formatting of description lists

roldanbob:
[DBZ-8450](https://issues.redhat.com/browse/DBZ-8450)

This change removes extraneous list continuation characters (i.e., plus signs [`+`]) from the descriptions of the Oracle connector properties `log.mining.query.filter.mode` and `log.mining.strategy`.

The documentation for the two Oracle connector properties uses definition list formatting to present information about the available configuration options. In an AsciiDoc definition list, a pair of colon characters represent the delimiter between the term being defined and its description. In the previous version of the documentation, the definition list formatting is incorrectly combined with the use of list continuation characters, i.e., `+` signs. As a result, instead of serving as markup characters, the colon characters render literally in the published content.


jpechane:
@roldanbob Applied, thanks

PR #6014, opened 2024-11-26
DBZ-8461Fix randomly failing tests related to stop of async engine

vjuranek:
https://issues.redhat.com/browse/DBZ-8461


jpechane:
@vjuranek Applied, thanks

PR #6016, opened 2024-11-26
DBZ-8456 Aligns text/formatting of skipped ops props and truncate docs

roldanbob:
[DBZ-8456](https://issues.redhat.com/browse/DBZ-8456)

Applies consistent text and formatting to equivalent `skipped.operations` and `truncate` event docs across the connector different connectors.


documentation/modules/ROOT/pages/connectors/db2.adoc
```diff
@@ -2970,9 +2970,15 @@ In the resulting snapshot, the connector includes only the records for which `de
...
-|A comma-separated list of operation types that will be skipped during streaming.
-The operations include: `c` for inserts/create, `u` for updates, `d` for deletes, `t` for truncates, and `none` to not skip any operations.
-By default, truncate operations are skipped (not emitted by this connector).
+|A comma-separated list of the operation types that you want the connector to skip during streaming.
+You can configure the connector to skip the following types of operations:
+
+* `c` (insert/create)
+* `u` (update)
+* `d` (delete)
+* `t` (truncate)
```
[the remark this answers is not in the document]
jpechane: Please implement the latter 2.

jpechane (approved):
@roldanbob LGTM, thanks! Please feel free to merge yourself when you address the Db2 unsupported truncate.

documentation/modules/ROOT/pages/connectors/sqlserver.adoc
```diff
@@ -3156,9 +3156,15 @@ endif::community[]
...
-|A comma-separated list of operation types that will be skipped during streaming.
-The operations include: `c` for inserts/create, `u` for updates, `d` for deletes, `t` for truncates, and `none` to not skip any operations.
-By default, truncate operations are skipped (not emitted by this connector).
+|A comma-separated list of the operation types that you want the connector to skip during streaming.
+You can configure the connector to skip the following types of operations:
+
+* `c` (insert/create)
+* `u` (update)
+* `d` (delete)
+* `t` (truncate)
```
jpechane, line 3165: This is the same case as Db2.

documentation/modules/ROOT/pages/connectors/vitess.adoc
```diff
@@ -1712,9 +1712,15 @@ For example, if you set `max.queue.size=1000`, and `max.queue.size.in.bytes=5000
...
-By default, truncate operations are skipped (not emitted by this connector).
+|A comma-separated list of the operation types that you want the connector to skip during streaming.
+You can configure the connector to skip the following types of operations:
+
+* `c` (insert/create)
+* `u` (update)
+* `d` (delete)
+* `t` (truncate)
+
+Set the value to `none` if you do not want the connector to skip any operations.
```
jpechane, line 1723: @twthorn With the DDL parser in place, does Vitess emit truncate event or no?

documentation/modules/ROOT/pages/connectors/db2.adoc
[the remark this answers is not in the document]
roldanbob: @jpechane Is it accurate to state that Db2 does not support truncate change events? Or should the doc state that the Db2 connector does not support them?

documentation/modules/ROOT/pages/connectors/sqlserver.adoc
roldanbob: @jpechane Similar to the Db2 case, as I look at the SQL Server 2022 documentation, I find a reference to using the `TRUNCATE TABLE` command. Should the connector doc state that the Debezium SQL Server connector does not support `truncate` operations vs. SQL Server does not support `truncate` events?

roldanbob:
Merging without further change to the Vitess doc. If @twthorn can verify that the statements added for the Db2 and SQL Server connectors also apply to Vitess, I'll apply that separately later.

documentation/modules/ROOT/pages/connectors/vitess.adoc
twthorn: It emits the truncate events to the ddl topics, but we do not currently emit them within the data change topic (as I believe some other connectors do, eg mysql)

twthorn:
@roldanbob The only different statement I see is "Because the [connector] does not support `truncate` change events, setting the default `t` value has the same effect as setting the value to `none`."

We only emit truncate events to the schema history topic. If this applies to only data change topic, then this is accurate (ie no diff between t or none)

PR #6018, opened 2024-11-27
DBZ-8467 Emit traces even when no propagated trace context is found

VJean:
In the ActivateTracingSpan SMT, emitting traces when no propagated trace context is found is controlled by the `tracing.with.context.field.only` setting. However, the if/else condition was flawed, resulting in traces not being emitted event with this setting being set to `false`. This commit fixes the condition logic.

# Context

Debezium offers to create traces when processing records. The [documentation](https://debezium.io/documentation/reference/stable/integrations/tracing.html#_activatetracingspan_smt) states that when provided with a serialized upstream application tracing context (through a specific SQL table field), it will use this trace context as a parent trace context for its own spans. If it doesn't find any, it will create traces from scratch.

In reality, if no propagated trace context is found, Debezium skips emitting traces ([source](https://github.com/debezium/debezium/blob/v3.0.1.Final/debezium-core/src/main/java/io/debezium/transforms/tracing/ActivateTracingSpan.java#L119C1-L132C30))
```
if (propagatedSpanContext == null) {
    if (requireContextField) {
        return connectRecord;
    }
}
else {
    try {
        return TracingSpanUtil.traceRecord(connectRecord, envelope, source, propagatedSpanContext, operationName);
    }
    catch (NoClassDefFoundError e) {
        throw new DebeziumException("Failed to record tracing information, tracing libraries not available", e);
    }
}
return connectRecord;
```

The `requireContextField` parameter comes from a setting. It allows to specify whether we want to trace only records that provide a trace context.

So, [per the documentation](https://debezium.io/documentation/reference/stable/integrations/tracing.html#_activatetracingspan_smt), we have the following cases:

1. WHEN propagatedSpanContext is not null
   THEN create trace (and propagate span context as parent)
2. WHEN propagatedSpanContext is null AND requireContextField is true
   THEN skip trace creation
3. WHEN propagatedSpanContext is null AND requireContextField is false (default)
   THEN create trace

Case 3. is the one to be fixed here.

# Changes

Main diff in short:

* The SMT's `apply` method is modified to skip tracing when (and only when) `propagatedSpanContext == null && requireContextField`
* The `TracingSpanUtil.traceRecord` method is improved to always create a trace while still handling the case where its `propagatedSpanContext` parameter would be null.

### DIFF
--- a/debezium-bom/pom.xml
+++ b/debezium-bom/pom.xml
@@ -30,6 +30,7 @@
 
         <!-- Tracing -->
         <version.opentelemetry>1.23.0</version.opentelemetry>
+        <version.opentelemetry.testing>2.10.0</version.opentelemetry.testing>
 
         <!-- HTTP client -->
         <version.okhttp>4.8.1</version.okhttp>
@@ -611,6 +612,18 @@
               <groupId>com.squareup.okhttp3</groupId>
               <artifactId>logging-interceptor</artifactId>
               <version>${version.okhttp}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.opentelemetry.javaagent</groupId>
+                <artifactId>opentelemetry-testing-common</artifactId>
+                <version>${version.opentelemetry.testing}-alpha</version>
+                <scope>test</scope>
+            </dependency>
+            <dependency>
+                <groupId>io.opentelemetry.javaagent</groupId>
+                <artifactId>opentelemetry-agent-for-testing</artifactId>
+                <version>${version.opentelemetry.testing}-alpha</version>
+                <scope>test</scope>
             </dependency>
 
             <dependency>
--- a/debezium-core/pom.xml
+++ b/debezium-core/pom.xml
@@ -87,6 +87,16 @@
             <artifactId>groovy-json</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>io.opentelemetry.javaagent</groupId>
+            <artifactId>opentelemetry-testing-common</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>io.opentelemetry.javaagent</groupId>
+            <artifactId>opentelemetry-agent-for-testing</artifactId>
+            <scope>test</scope>
+        </dependency>
 
         <!-- Used for unit testing with Kafka -->
         <dependency>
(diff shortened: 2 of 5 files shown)


mfvitale:
@VJean Thanks for the PR. Can you also provide a tests for it?

VJean:
Sure! There is currently no tests for this class yet (which might have been part of the initial problem of not matching expectations). I'll work on adding one to ensure the expected behaviour is now correctly implemented.

debezium-bom/pom.xml
```diff
@@ -30,6 +30,7 @@
 
         <!-- Tracing -->
         <version.opentelemetry>1.23.0</version.opentelemetry>
+        <version.opentelemetry.testing>2.10.0</version.opentelemetry.testing>
```
VJean, line 33: Had to add the `2.10.0` opentelemetry testing javaagent as the `1.23.0` one would not support JDK 21 and would fail to initialize.

I did it without updating the opentelemetry dependency used in code to avoid introducing other changes.

debezium-core/pom.xml
```diff
@@ -146,6 +156,23 @@
...
+                            <goal>properties</goal>
+                        </goals>
+                    </execution>
+                </executions>
+            </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-surefire-plugin</artifactId>
+                <configuration>
+                    <argLine>-javaagent:${io.opentelemetry.javaagent:opentelemetry-agent-for-testing:jar}</argLine>
```
VJean, line 173: Tests have to be run with the opentelemetry java agent to actually test traces emission and TextMapPropagator implemetations behaviors. If not, the tests would go through the default API implementation which is noop, resulting in test failures.

VJean (commented):
Added a full TestSuite for the `ActivateTracingSpan` SMT class.

The following test cases test the condition flow that is mentioned in the PR description:
* `whenPropagationContextIsProvidedATraceIsCreated`
* `whenPropagationContextIsNotProvidedAndContextIsNotRequiredATraceIsCreated`
* `whenPropagationContextIsNotProvidedAndContextIsRequiredNoTraceIsCreated`

mfvitale:
@VJean Thanks for adding the tests! LGMT, let's wait the tests.

debezium-core/pom.xml
mfvitale: @jpechane any concern on this?

VJean:
@mfvitale Looks like all checks are green :)

mfvitale:
@VJean applied. Thanks a lot.

PR #6019, opened 2024-11-28
DBZ-8433 Adds a handling event processing failure method and refactorizes in dispatchers

JuanmaBM:
Fixes issue https://issues.redhat.com/projects/DBZ/issues/DBZ-7980?filter=allopenissues

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
@@ -206,32 +206,37 @@ public class EventDispatcher<P extends Partition, T extends DataCollectionId> im
     public void dispatchSnapshotEvent(P partition, T dataCollectionId, ChangeRecordEmitter<P> changeRecordEmitter,
                                       SnapshotReceiver<P> receiver)
             throws InterruptedException {
-        // TODO Handle Heartbeat
 
-        DataCollectionSchema dataCollectionSchema = schema.schemaFor(dataCollectionId);
+        try {
+            // TODO Handle Heartbeat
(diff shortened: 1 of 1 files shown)


debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
```diff
@@ -322,24 +327,33 @@ private boolean isASignalEventToProcess(T dataCollectionId, Operation operation)
...
-                            "Error while processing event at offset {}",
-                            changeRecordEmitter.getOffset().getOffset(), e);
-                    break;
-                case SKIP:
-                    LOGGER.debug(
-                            "Error while processing event at offset {}",
-                            changeRecordEmitter.getOffset().getOffset(), e);
-                    break;
-            }
+            handleEventProcessingFailure(e, changeRecordEmitter);
```
mfvitale, line 330: I'd pass just the offset as parameter instead of changeRecordEmitter. WDYT?

mfvitale:
Hi @JuanmaBM , thanks for the PR. It generally LGTM, I just leaved a minor comment.

debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
JuanmaBM: I was thinking that changeRecordEmitter might be useful if any additional actions or verifications are needed. But yes, you're right—let's stick to the KISS principle

mfvitale:
@JuanmaBM applied! Failures seems not related.  Thanks

JuanmaBM:
Great, thank you @mfvitale !!

PR #6021, opened 2024-11-28
DBZ-8465 Don't modify map in computeIfAbsent() function

vjuranek:
This is forbidden and `ConcurrentModificationException` can be eventually thrown.

https://issues.redhat.com/browse/DBZ-8465

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDatabaseSchema.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDatabaseSchema.java
@@ -44,6 +44,7 @@ public class OracleDatabaseSchema extends HistorizedRelationalDatabaseSchema {
 
     public static final String ATTRIBUTE_OBJECT_ID = "OBJECT_ID";
     public static final String ATTRIBUTE_DATA_OBJECT_ID = "DATA_OBJECT_ID";
+    private static final TableId NO_SUCH_TABLE = new TableId(null, null, "__NULL");
 
     private final OracleDdlParser ddlParser;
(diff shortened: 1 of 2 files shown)


vjuranek (commented):
I refactored the lambda into dedicated function. In general I don't like longer lambdas (like this one), however, in this case IMHO make some sense to keep it close to `if` clause which convert back the placeholder. Not sure which solution is better, leaving it to your judgement.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDatabaseSchema.java
```diff
@@ -149,32 +150,15 @@ public TableId getTableIdByObjectId(Long objectId, Long dataObjectId) {
...
-                        if (dataAttribute == null || !dataAttribute.asLong().equals(dataObjectId)) {
-                            // Did not match, continue
-                            continue;
-                        }
-                    }
-                    LOGGER.debug("Table lookup for object {} resolved to '{}'", tableObjectId, table.id());
-                    return table.id();
-                }
-            }
+        TableId cachedTableId = objectIdToTableId.putIfAbsent(objectId, tableObjectIdToTableId(objectId, dataObjectId));
```
jpechane, line 153: OK, `putIfAbsent` was not a good choice because right now I'd say we are back to square one, doing the expensive search in the `tableObjectIdToTableId` for something we might not need at all.
So in this regard the plain
```
var cachedTableId = objectIdToTableId.get(objectId);
if (cachedTableId == null) {
  cachedTableId = tableObjectIdToTableId(objectId, dataObjectId);
} 
```
would probably work best

jpechane (changes requested):

PR #6022, opened 2024-11-28
[docs] Correct the default value for log.mining.strategy

vjuranek:
ALso replace `redo_in_catalog` with correct `redo_log_catalog` option.


jpechane:
@vjuranek Applied, thanks

PR #6025, opened 2024-11-30
DBZ-XXXX Add AnonymiseField transformation. Transformation used to hash or encrypt given fields

ismailsimsek:
This transformation will hash given fields to do pseudonymisation

### DIFF
--- a/debezium-core/src/main/java/io/debezium/transforms/AESEncryption.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/AESEncryption.java
@@ -1,0 +1,101 @@
+package io.debezium.transforms;
+
+import java.security.NoSuchAlgorithmException;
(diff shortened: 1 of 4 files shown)


jpechane:
@ismailsimsek We do have https://debezium.io/documentation/reference/3.0/connectors/postgresql.html#postgresql-property-column-mask-hash
What is the motivation for this SMT? Is there something the existing feature cannot provide?

ismailsimsek:
@jpechane i will check it

rk3rn3r:
Furthermore let me add there is already Kryptonite available from our friend and former colleague @hpgrahsl: https://github.com/hpgrahsl/kryptonite-for-kafka
see https://developers.redhat.com/articles/2022/09/27/end-end-field-level-encryption-apache-kafka-connect

PR #6026, opened 2024-12-02
DBZ-8270 MySQL and MariaDB grammars with upstream versions 

ani-sha:
- [x] Format grammar using antlr-format-cli as upstream grammar
- [x] Compare the grammar and remove MariaDB-specific extensions that are not in the upstream MySQL grammar
- [x] Make sure all these extensions are properly handled in MariaDB grammar
- [x] MariaDB grammar is the same as the upstream grammar
- [x] MariaDB tests are the same as the upstream grammar
- [x] MySQL grammar contains only Percona extensions - this is the only difference from upstream grammar
- [x] MySQL grammar tests will be replaced with the upstream ones. 
- [x] Extra test statements in our grammar not covered in the  upstream grammar test will be moved to a separate file and kept for regression testing
- [x] Fix CI test failures; separate MariaDB ANTLR tests from `BinlogAntlrDdlParserTest`

closes https://issues.redhat.com/browse/DBZ-8270


jpechane:
@ani-sha Hi, can you intorduce a new class for MariaDB specific testing?

jpechane:
@ani-sha This is very nice progress! One question for you. Is/was the upstream grammar reformatting one-time event or is it done continuously in the same way we do for Java sources. If the latter we should probably introduce it to our build too.

ani-sha:
@jpechane The formatter tool is part of the script that runs for every PR. So I will add it to our CI checks as well.

ani-sha:
@jpechane the CI in ANTLR grammar has formatting checks but it just emits a warning; after reaching out to the team, they responded that they format and fix the grammar based on the warning later on if needed. 
I haven't seen any other formatting changes since the last major reformatting using the Antlr tool. So, should we still add this check to our CI or skip it now?

Naros:
> @Naros Not sure if you had the chance to go through this PR. I have a question for you. Since now MySQL and MariaDB grammars are completely separated. How do we address the MariaDB-specific unit tests? I see there is a common class `BinlogAntlrDdlParserTest`for both MySQL and MariaDB tests. For now, I have just removed and adjusted the MariaDB tests from `BinlogAntlrDdlParserTest`.

I like the separation but I do wonder if it would make sense to have a common base class still for situations where there is still overlap between the two grammars, even though the rules are different, so that we have the flexibility to add the test once rather than potentially twice where we know the test applies to both binlog connectors?  wdyt?

ani-sha:
@Naros The separation still exists. I added a new class `MariadbOnlyAntlrDdlParserTest` which only contains the mariadb-specific tests. `MariadbAntlrDdlParserTest` still extends `BinlogAntlrDdlParserTest` and tests the common grammar.

ani-sha:
@jpechane Fixed all failing tests, could give it a review? The one failing test seems unrelated.
```
MySqlConnectorConvertingFailureIT>BinlogConvertingFailureIT.shouldRecoverToSyncSchemaWhenFailedValueConvertByDdlWithSqlLogBinIsOff:104->AbstractConnectorTest.stopConnector:163->AbstractConnectorTest.stopConnector:178 » IllegalState Engine is already being shutting down.
```

jpechane (approved):
@ani-sha Great job! Please rebase the PR on the latest main and the PR can be merged. Thanks

jpechane:
@ani-sha Rebased on the latest main with infra tests fixed and force pushed. Can be merged after PR finish.

jpechane:
@ani-sha Applied, thanks a lot!

PR #6027, opened 2024-12-02
DBZ-8451 Support Mongo 8.0

ani-sha:
closes https://issues.redhat.com/browse/DBZ-8451


jpechane:
@ani-sha Thanks. I think we can replace 7 with 8 in this PR to keep time down and add it to internal Jenkins so we will test both 7 and 8 there.

jpechane:
@ani-sha Applied, thanks

PR #6029, opened 2024-12-04
[ci] Fix MariaDB profile names

vjuranek:


Naros (approved):
Wow, thanks for catching that @vjuranek, cut-n-paste failed apparently :grin:

PR #6031, opened 2024-12-06
DBZ-8492 Do not log and gather attributes for views

Naros:
https://issues.redhat.com/browse/DBZ-8492


jpechane:
@Naros Any idea why non oracle jobs were triggered here?

PR #6034, opened 2024-12-09
DBZ-8496 Make Connector schema snapshot field consist with SnapshotRecord Enum

fanyang:
Let Connector schema snapshot field allowed values reference SnapshotRecord Enum.


mfvitale (approved):

mfvitale:
Hi @fanyang Thanks for the PR. LGMT.

mfvitale:
@jpechane I don't think it should be back ported. WDYT?

jpechane:
@mfvitale No, it is not necessary. But we should mive schema definitions from this class into `io.debezium.schema.SchemaFactory` and start to version them.

PR #6036, opened 2024-12-10
DBZ-8500 using correct name for table.exists property in JdbcSchemaHistoryConfig

jcechace:
https://issues.redhat.com/browse/DBZ-8500


mfvitale (approved):

jpechane:
@jcechace Applied, thanks

PR #6037, opened 2024-12-10
DBZ-8499 Make source info versioned

jpechane:
https://issues.redhat.com/browse/DBZ-8499


rk3rn3r:
I was thinking if that has an impact on the schema generator and/or if it generally needs to be evaluated and maybe retired eventually?

PR #6039, opened 2024-12-10
Dbz 7094 ExtractNewRecordState value of optional null field which has default value

kmos:
## Context

A user reported an unexpected behavior when using the `ExtractNewRecordState` Single Message Transform (SMT). A table column defined with a default value in the source database is represented with that default value instead of `null` in the resulting event payload, even though the inserted record had `null` in that column.

This issue is not limited to PostgreSQL; it affects multiple RDBMS connectors (MySQL, PostgreSQL, etc.) when using the `ExtractNewRecordState` SMT. The root cause is how certain Kafka Connect transforms (`ExtractField`, `InsertField`) handle optional null fields and default values.

**Expected Behavior:**  
When a `null` value is inserted with a database-defined default value, Debezium’s change event should preserve that `null` in the final payload.

**Actual Behavior:**  
The final payload shows the column’s default value (e.g., `1.0`) instead of `null`, potentially causing downstream consumers to misinterpret the data.

## Changes

We add a configuration parameter `replace.null.with.default` that instruments `ExtractNewRecordState` and consequently the `InsertField` transformer in order to make the user decide the behavior of the extractor. This is possible thanks to the [KIP-1040](https://github.com/apache/kafka/pull/15756). In case the parameter is set to `true`, the behavior is to replace null value with the database-defined default. This behavior is the default one if the parameter is unset for backward compatibility.

closes https://issues.redhat.com/browse/DBZ-7094

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -651,3 +651,4 @@ Gaurav Miglani
 Ashish Binu
 Mohamed El Shaer
 Artyom Dubinin
+Giovanni Panice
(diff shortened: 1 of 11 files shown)


jpechane (approved):
@kmos LGTM! Thanks.
Could you please also provide a docs update for the SMT?

kmos:
@jpechane done! I have added some documentation for the new option `replace-null-with-default`. Do you think it is straight to the point or is it necessary something else?

jpechane:
@kmos Applied, thanks!

PR #6040, opened 2024-12-11
DBZ-8482 Fix MariaDB SSL configuration

vjuranek:
https://issues.redhat.com/browse/DBZ-8482


jpechane:
@vjuranek Applied, great job!

PR #6041, opened 2024-12-11
DBZ-8503 Use LAST_REDO_CHANGE#-1 from V$THREAD for range boundaries

Naros:
https://issues.redhat.com/browse/DBZ-8503


vjuranek (approved):

jpechane:
@Naros Applied, thanks

PR #6042, opened 2024-12-11
DBZ-8504 Support TEMPLATE_TABLE clause for Oracle data pump DDL

Naros:
https://issues.redhat.com/browse/DBZ-8504


jpechane:
@Naros Applied, thanks

PR #6044, opened 2024-12-11
DBZ-8491 Add metric to track partial rollback counters

Naros:
https://issues.redhat.com/browse/DBZ-8491


jpechane:
@Naros Applied, thanks

PR #6048, opened 2024-12-12
DBZ-3720 Error with debezium.sink.pulsar.client.serviceUrl and debezium-server

gustavolira:
Fixes https://issues.redhat.com/browse/DBZ-3720
Depends on https://github.com/debezium/debezium-server/pull/142

### DIFF
--- a/debezium-core/src/main/java/io/debezium/util/Strings.java
+++ b/debezium-core/src/main/java/io/debezium/util/Strings.java
@@ -25,6 +25,7 @@ import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 import java.util.regex.PatternSyntaxException;
+import java.util.stream.Collectors;
 
 import io.debezium.annotation.ThreadSafe;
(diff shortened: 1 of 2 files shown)


debezium-core/src/main/java/io/debezium/util/Strings.java
```diff
@@ -1253,4 +1254,24 @@ else if (c == ',') {
...
+     * Examples:
+     * - "example.string.input" becomes "ExampleStringInput"
+     * - "another_example_input" becomes "AnotherExampleInput"
+     *
+     * @param input the input string to be converted, containing words separated by '.' or '_'.
+     * @return the converted string in camel case format.
+     *
+     * @throws NullPointerException if the input string is null.
+     */
+    public static String convertToCamelCase(String input) {
```
jpechane, line 1272:
```suggestion
    public static String convertUnderscoredStringToCamelCase(String input) {
```

jpechane:
@gustavolira Applied, thanks

PR #6051, opened 2024-12-13
DBZ-8514 TimezoneConverter include.list should be respected if set (2.7.x)

martin-walsh:
Fix for what I would consider to be a bug in TimezoneConverter `include.list`.

If you specify entries e.g. 
```
transforms.convertTimezone.include.list=source:table_one:DATE_CREATED,table_one:LAST_UPDATED,table_two:CREATED_AT,table_two:UPDATED_AT
```

and an event from `table_three` is received, the TimezoneConverter will currently apply the conversion to all columns within. The fallback from not finding a table match in config is to convert all columns, instead of not converting any of them.

This leaves you with no way to opt in to select columns on select tables without adding workaround config e.g. `table_three:workaround_cfg`.

See https://issues.redhat.com/browse/DBZ-8514

### DIFF
--- a/debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
@@ -519,9 +519,6 @@ public class TimezoneConverter<R extends ConnectRecord<R>> implements Transforma
                 handleStructs(value, Type.ALL, matchName, fields);
             }
         }
-        else {
-            handleStructs(value, Type.ALL, table, Collections.emptySet());
-        }
     }
 
     private void handleExclude(Struct value, String table, String topic) {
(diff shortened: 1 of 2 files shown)


martin-walsh:
Tests are failing with errors similar to:
```Error:  io.debezium.connector.jdbc.e2e.JdbcSinkPipelineToMySqlIT -- Time elapsed: 0.419 s <<< ERROR!
java.util.concurrent.CompletionException: org.testcontainers.containers.ContainerFetchException: Can't get Docker image: RemoteDockerImage(imageName=debezium/example-postgres:latest, imagePullPolicy=DefaultPullPolicy(), ```

Do these just require a retry ?

martin-walsh:
See https://github.com/debezium/debezium/pull/6055 for PR to main.

PR #6053, opened 2024-12-13
DBZ-4590 Test the MS SQL Server Plugin with Transparent data encryption (TDE)

kmos:
## Context

Transparent Data Encryption (TDE) encrypts the entire database and transaction log at rest, ensuring unauthorized file access remains unreadable. It operates automatically and transparently, requiring no code changes, and interactions through JDBC remain unaffected. For this reason, the Debezium connector should work fine even with a database protected by TDE.

**Expected Behavior:**  
All the tests should works fine also with a ms SQL Server encrypted with TDE

## Changes

We add a new test container with an initial script that activate the encryption to database level. The test container is added in `TestInfrastructureHelper` in order to be used in `DebeziumSqlServerConnectorResourceIT`. The nature of TDE is to be trasparent to application level and my suggestion is to leave activated for all the tests. Add a new set of tests for only this feature can grow the pipeline execution time.

closes https://issues.redhat.com/browse/DBZ-4590


Naros (approved):
LGTM @kmos, great work adding TDE support checks for SQL Server.  Anything from your perspective @jpechane, before merging this into `main` ?

jpechane:
@kmos Applied, thanks!

PR #6054, opened 2024-12-14
DBZ-7051 Pass connector configuration to Column/Table naming strategies

gustavolira:
https://issues.redhat.com/browse/DBZ-7051

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
@@ -23,13 +23,16 @@ import io.debezium.config.Configuration;
 import io.debezium.config.EnumeratedValue;
 import io.debezium.config.Field;
 import io.debezium.config.Field.ValidationOutput;
+import io.debezium.connector.jdbc.naming.CollectionNamingStrategyFactory;
 import io.debezium.connector.jdbc.naming.ColumnNamingStrategy;
+import io.debezium.connector.jdbc.naming.ColumnNamingStrategyFactory;
 import io.debezium.connector.jdbc.naming.DefaultColumnNamingStrategy;
 import io.debezium.connector.jdbc.naming.TemporaryBackwardCompatibleCollectionNamingStrategyProxy;
 import io.debezium.sink.SinkConnectorConfig;
 import io.debezium.sink.filter.FieldFilterFactory;
 import io.debezium.sink.filter.FieldFilterFactory.FieldNameFilter;
 import io.debezium.sink.naming.CollectionNamingStrategy;
+import io.debezium.sink.naming.DefaultCollectionNamingStrategy;
 import io.debezium.util.Strings;
 
 /**
@@ -57,6 +60,13 @@ public class JdbcSinkConnectorConfig implements SinkConnectorConfig {
     public static final String SCHEMA_EVOLUTION = "schema.evolution";
     public static final String QUOTE_IDENTIFIERS = "quote.identifiers";
     public static final String COLUMN_NAMING_STRATEGY = "column.naming.strategy";
+    public static final String COLUMN_NAMING_STYLE = "column.naming.style";
+    public static final String COLUMN_NAMING_PREFIX = "column.naming.prefix";
+    public static final String COLLECTION_NAMING_PREFIX = "collection.naming.prefix";
+    public static final String COLLECTION_NAMING_SUFFIX = "collection.naming.suffix";
+    public static final String COLLECTION_NAMING_STYLE = "collection.naming.style";
+
+    public static final String COLUMN_NAMING_SUFFIX = "column.naming.suffix";
     public static final String POSTGRES_POSTGIS_SCHEMA = "dialect.postgres.postgis.schema";
     public static final String SQLSERVER_IDENTITY_INSERT = "dialect.sqlserver.identity.insert";
     public static final String USE_REDUCTION_BUFFER = "use.reduction.buffer";
@@ -188,6 +198,66 @@ public class JdbcSinkConnectorConfig implements SinkConnectorConfig {
             .withDefault(DefaultColumnNamingStrategy.class.getName())
             .withDescription("Name of the strategy class that implements the ColumnNamingStrategy interface.");
 
+    public static final Field COLUMN_NAMING_STYLE_FIELD = Field.create(COLUMN_NAMING_STYLE)
+            .withDisplayName("Column Naming Style")
+            .withType(Type.STRING)
+            .withWidth(ConfigDef.Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.LOW)
+            .withValidation(JdbcSinkConnectorConfig::validateColumnNamingStyle)
+            .withDefault("default")
+            .withDescription("The style of column naming: snake_case, camelCase, kebab-case, UPPERCASE, lowercase.");
+
+    public static final Field COLUMN_NAMING_PREFIX_FIELD = Field.create(COLUMN_NAMING_PREFIX)
+            .withDisplayName("Column Naming Prefix")
+            .withType(Type.STRING)
+            .withDefault("")
+            .withWidth(ConfigDef.Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.LOW)
+            .withDescription("Optional prefix to add to column names.");
+
+    public static final Field COLUMN_NAMING_SUFFIX_FIELD = Field.create(COLUMN_NAMING_SUFFIX)
+            .withDisplayName("Column Naming Suffix")
+            .withType(Type.STRING)
+            .withDefault("")
+            .withWidth(ConfigDef.Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.LOW)
+            .withDescription("Optional suffix to add to column names.");
+
+    public static final Field COLLECTION_NAMING_STRATEGY_FIELD = Field.create(COLLECTION_NAMING_STRATEGY)
+            .withDisplayName("Name of the strategy class that implements the CollectionNamingStrategy interface")
+            .withType(ConfigDef.Type.CLASS)
+            .withGroup(Field.createGroupEntry(Field.Group.CONNECTOR_ADVANCED, 2))
+            .withWidth(ConfigDef.Width.LONG)
+            .withImportance(ConfigDef.Importance.LOW)
+            .withDefault(DefaultCollectionNamingStrategy.class.getName())
+            .withDescription("Name of the strategy class that implements the CollectionNamingStrategy interface. " +
+                    "This field allows users to define a custom implementation for resolving collection names. " +
+                    "If not specified, the default strategy is used.")
+            .withDeprecatedAliases(DEPRECATED_TABLE_NAMING_STRATEGY)
+            .withValidation(JdbcSinkConnectorConfig::validateCollectionNamingStrategy);
+
+    public static final Field COLLECTION_NAMING_STYLE_FIELD = Field.create(COLLECTION_NAMING_STYLE)
+            .withDisplayName("Column Naming Style")
+            .withType(Type.STRING)
+            .withWidth(ConfigDef.Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.LOW)
+            .withDefault("default")
+            .withDescription("The style of table naming: snake_case, camel_case, upper_case, lower_case.");
+
+    public static final Field COLLECTION_NAMING_PREFIX_FIELD = Field.create(COLLECTION_NAMING_PREFIX)
+            .withDisplayName("Table Naming Prefix")
+            .withType(Type.STRING)
+            .withWidth(ConfigDef.Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.LOW)
+            .withDescription("Optional prefix to add to table names.");
+
+    public static final Field COLLECTION_NAMING_SUFFIX_FIELD = Field.create(COLLECTION_NAMING_SUFFIX)
+            .withDisplayName("Table Naming Suffix")
+            .withType(Type.STRING)
+            .withWidth(ConfigDef.Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.LOW)
+            .withDescription("Optional suffix to add to table names.");
+
     public static final Field POSTGRES_POSTGIS_SCHEMA_FIELD = Field.create(POSTGRES_POSTGIS_SCHEMA)
             .withDisplayName("Name of the schema where postgis extension is installed")
             .withType(Type.STRING)
@@ -253,7 +323,13 @@ public class JdbcSinkConnectorConfig implements SinkConnectorConfig {
                     SCHEMA_EVOLUTION_FIELD,
                     QUOTE_IDENTIFIERS_FIELD,
                     COLLECTION_NAMING_STRATEGY_FIELD,
+                    COLLECTION_NAMING_PREFIX_FIELD,
+                    COLLECTION_NAMING_STYLE_FIELD,
+                    COLLECTION_NAMING_SUFFIX_FIELD,
                     COLUMN_NAMING_STRATEGY_FIELD,
+                    COLUMN_NAMING_STYLE_FIELD,
+                    COLUMN_NAMING_PREFIX_FIELD,
+                    COLUMN_NAMING_SUFFIX_FIELD,
                     USE_TIME_ZONE_FIELD,
                     POSTGRES_POSTGIS_SCHEMA_FIELD,
                     SQLSERVER_IDENTITY_INSERT_FIELD,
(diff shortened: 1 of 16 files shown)


debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -9,6 +9,8 @@
 import java.util.Set;
 import java.util.function.Consumer;
 
+import io.debezium.connector.jdbc.naming.*;
```
Naros, line 12: We don't permit the use of `.*` imports. 

If you're using IDEA, for example, you can set `Settings > Editor > Code Style > Java > Imports > Class count to use import with '*'` with a value of `999` to avoid the IDE from doing this silly thing.

I'd also run `./mvnw clean install -DskipITs -DskipTests` so that imports get sorted before submitting the PR. We prefer to have packages grouped for legibility reasons.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -188,6 +194,79 @@ public class JdbcSinkConnectorConfig implements SinkConnectorConfig {
...
 
+    public static final Field COLUMN_NAMING_STYLE_FIELD = Field.create(COLUMN_NAMING_STYLE)
+            .withDisplayName("Column Naming Style")
+            .withType(Type.STRING)
+            .withWidth(ConfigDef.Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.LOW)
+            .withDefault("default")
+            .withDescription("The style of column naming: snake_case, camelCase, kebab-case, UPPERCASE, lowercase.");
+
+    public static final Field COLUMN_NAMING_PREFIX_FIELD = Field.create(COLUMN_NAMING_PREFIX)
```
Naros, line 205: What do you think about adding empty string defaults for the prefix/suffix? I believe this could simplify the checks in other places because these properties would either be a populated or empty string but never `null`.  wdyt?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -188,6 +194,79 @@ public class JdbcSinkConnectorConfig implements SinkConnectorConfig {
...
+
+                try {
+                    Class.forName(value);
+                } catch (ClassNotFoundException e) {
+                    problems.accept(field, value, "Invalid class specified for collection naming strategy.");
+                    return 1; // validation fail
+                }
+
+                return 0; // no problem
+            });
```
Naros, line 246: We've generally used a private static method reference here and placed the validation that way rather than using lambda expressions. I'd prefer we continue that trend for code consistency.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -419,12 +504,23 @@ public void validate() {
         if (!Strings.isNullOrEmpty(columnExcludeList) && !Strings.isNullOrEmpty(columnIncludeList)) {
             throw new ConnectException("Cannot define both column.exclude.list and column.include.list. Please specify only one.");
         }
+
+        validateColumnNamingProperties();
```
Naros, line 508: Could we use the `Field#withValidation` handler for this rather than calling this directly?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/naming/CollectionNamingStrategyFactory.java
```diff
@@ -0,0 +1,81 @@
...
+                return configureDefaultStrategy(props);
+            }
+
+            CollectionNamingStrategy strategy = config.getInstance(
+                    JdbcSinkConnectorConfig.COLLECTION_NAMING_STRATEGY_FIELD,
+                    CollectionNamingStrategy.class
+            );
+            strategy.configure(props);
+            return strategy;
+        } catch (Exception e) {
```
Naros, line 64:
```suggestion
        } 
        catch (Exception e) {
```
This will likely be caught by the formatter GitHub action task.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/naming/CustomCollectionNamingStrategy.java
```diff
@@ -0,0 +1,35 @@
...
+        this.prefix = properties.getOrDefault("collection.naming.prefix", "");
+        this.suffix = properties.getOrDefault("collection.naming.suffix", "");
+        String style = properties.getOrDefault("collection.naming.style", NamingStyle.DEFAULT.getValue());
+        this.namingStyle = NamingStyle.from(style);
+    }
+
+    @Override
+    public String resolveCollectionName(DebeziumSinkRecord record, String collectionFormat) {
+        String transformedName = NamingStyleUtils.applyNamingStyle(collectionFormat, namingStyle);
+        return (prefix != null ? prefix : "") + transformedName + (suffix != null ? suffix : "");
```
Naros, line 33:
```suggestion
        return Strings.defaultIfBlank(prefix, "") + transformedName + Strings.defaultIfBlank(suffix, "");
        return (prefix != null ? prefix : "") + transformedName + (suffix != null ? suffix : "");
```

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/util/NamingStrategyBase.java
```diff
@@ -0,0 +1,52 @@
...
+
+/**
+ * Base class for naming strategies that provides common functionality for applying prefixes, suffixes,
+ * and naming styles (e.g., snake_case, camelCase, etc.) to names.
+ * This class can be extended by naming strategies for columns, collections, or other entities that require
+ * customizable naming conventions.
+ *
+ * @author Gustavo Lira
+ */
+public abstract class NamingStrategyBase {
```
Naros, line 13: I'm not a fan of `Base`, though we have a few instances of its usage across the code base. I prefer using `Abstract` or `AbstractCommon` as a naming prefix. Still, I understand the desire to use suffixes to logically group classes in the IDE.  YMMV.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/util/NamingStrategyBase.java
```diff
@@ -0,0 +1,52 @@
...
+
+    /**
+     * Applies the configured naming style, prefix, and suffix to the given name.
+     *
+     * @param name the original name to be transformed
+     * @return the transformed name with applied prefix, suffix, and naming style
+     */
+    public String applyNaming(String name) {
+        if (name == null) {
+            throw new IllegalArgumentException("Name must not be null");
```
Naros, line 47: We should prefer to use `DebeziumException` or `ConnectException` here.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/util/NamingStyle.java
```diff
@@ -0,0 +1,60 @@
...
+     *
+     * @param value the string representation of the naming style
+     * @return the corresponding {@link NamingStyle}
+     * @throws IllegalArgumentException if the value does not match any naming style
+     */
+    public static NamingStyle from(String value) {
+        return Stream.of(values())
+                .filter(style -> style.value.equalsIgnoreCase(value))
+                .findFirst()
+                .orElseThrow(() -> new IllegalArgumentException(
```
Naros, line 35: Another use case of `DebeziumException` or `ConnectException`.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/util/NamingStyleUtils.java
```diff
@@ -0,0 +1,43 @@
...
+    /**
+     * Applies a naming style to a given name.
+     *
+     * @param name  the original name
+     * @param style the naming style as an enum
+     * @return the transformed name
+     */
+    public static String applyNamingStyle(String name, NamingStyle style) {
+        if (name == null || style == null) {
+            throw new IllegalArgumentException("Name and style must not be null");
```
Naros, line 31: `DebeziumException` or `ConnectException`

Naros (commented):
Hi @gustavolira, this is a great start. I like the idea you thought outside the box and added the suffix/prefix and case-handling features for columns.  I do think while we're at it, case-handling for table/collection names would be helpful and that would round out the complete feature for users who are working with Oracle/Db2 & PostgreSQL where their default cases are exact opposites.

Otherwise I've left a few inline comments if you could take a look.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -188,6 +194,79 @@ public class JdbcSinkConnectorConfig implements SinkConnectorConfig {
...
+                    Class.forName(value);
+                } catch (ClassNotFoundException e) {
+                    problems.accept(field, value, "Invalid class specified for collection naming strategy.");
+                    return 1; // validation fail
+                }
+
+                return 0; // no problem
+            });
+
+    public static final Field COLLECTION_NAMING_STYLE_FIELD = Field.create(COLLECTION_NAMING_STYLE)
```
rk3rn3r, line 248: Is a config option for a naming style really needed?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -188,6 +194,79 @@ public class JdbcSinkConnectorConfig implements SinkConnectorConfig {
...
+            .withDisplayName("Column Naming Style")
+            .withType(Type.STRING)
+            .withWidth(ConfigDef.Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.LOW)
+            .withDefault("default")
+            .withDescription("The style of table naming: snake_case, camel_case, upper_case, lower_case.");
+
+    public static final Field COLLECTION_NAMING_PREFIX_FIELD = Field.create(COLLECTION_NAMING_PREFIX)
```
rk3rn3r, line 256: Imo a prefix and suffix can already be added via the COLLECTION_NAME_FORMAT_FIELD (`collection.name.format`, prior named `table.name.format`). If this is really the case we could remove the suffix and prefix related code change. wdyt?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
rk3rn3r: See my other previous comment. On similar use-case we use `collection.name.format` to add an optional prefix and/or suffix. Should we keep this behavior instead of adding a new way by adding 2 options (prefix/suffix)? That would align better with how we do this and it allows combining prefix and suffix in a single option.

Naros: Hi @rk3rn3r, that's certainly another way to tackle it, though we'd need to introduce a placeholder for the name of where it should be injected as we did with `${topic}` and `${source.table}` for the collection name.  I'm not partial to either approach.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
gustavolira: While COLLECTION_NAME_FORMAT_FIELD could technically handle prefixes and suffixes, I believe keeping prefix and suffix as dedicated configuration options makes sense for simplicity and usability. These fields cover the most common customization scenarios (e.g., adding DEV_ or _ARCHIVE) without requiring users to define a custom format string. This separation also keeps basic configurations straightforward while leaving COLLECTION_NAME_FORMAT_FIELD for more advanced use cases. Removing prefix and suffix would risk complicating simple tasks for users. wdyt?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/util/NamingStyleUtils.java
```diff
@@ -0,0 +1,43 @@
...
+public class NamingStyleUtils {
+
+    /**
+     * Applies a naming style to a given name.
+     *
+     * @param name  the original name
+     * @param style the naming style as an enum
+     * @return the transformed name
+     */
+    public static String applyNamingStyle(String name, NamingStyle style) {
```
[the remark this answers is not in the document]
Naros: How would you propose the dialect know to use case sensitive vs insensitive naming? Respectfully, I think this is something where being explicitly set by the configuration works best, but I'm open to suggestions.

gustavolira:
Hi @Naros , I hope you're doing well! Just checking if there's anything else I need to update or address before the PR can be merged? Thanks!

Naros:
Hi @gustavolira it seems the test failure is related, can you take a look?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
rk3rn3r: You could use SMT/s to change topic and field names to whatever naming patterns you like to use?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/util/NamingStyleUtils.java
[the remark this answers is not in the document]
Naros: If we wanteed to push this up to a transform, then I would say we should provide the transform out of the box so that it would work both with flattened events but also Debezium's hierarchical events. That would introduce breaking changes, and I think it would be in our best interest to provide the solution for that.

Naros:
Hi @gustavolira, I understand where my colleague is coming from WRT transformations. While we've tried to be cheeky with case management with Db2, Oracle, and PostgreSQL, I do question if we could or should have tackled that differently from the connector's outset, but here we are.

That said, I do have a concern about your code. 

Suppose we have a PostgeSQL target database, and your connector configuration is set to use the custom naming strategy, setting the style as uppercase but identifiers are not quoted. I believe this will be problematic because the JDBC metadata returned by the JDBC driver will show the column names in lowercase, but the strategy is configured to use uppercase, and I suspect when schema evolution is enabled, we'll fail the column lookup and attempt to alter the table and re-add the columns.

So I would like to make the following proposals, as I think in hindsight this may be the best route forward.

First, lets split this PR into two pieces. The first piece is where the `ColumnNamingStrategy#configure` method is added and lets apply that in isolation as part of DBZ-7051.

Second, lets take the naming strategy work you've done and move it into one or more transformations in the JDBC sink connector's code base. I could imagine you could create one massive transformation or you could split the work into two where you have one that focuses on the collection name prefix/suffix/case and one that focuses on the field name prefix/suffix/case. I believe this will lay the ground work for us to consider removing the upper/lower casing magic in the JDBC sink entirely and have our community using your transformation for that work. This should also simplify the JDBC sink connector's code base and that will be advantageous with the work on the MongoDB sink alignment with the JDBC sink.

The only caveat with the column name prefix/suffix/case is you will need to support both flattened events, where you have a top level `Struct` that represents the value with a series of fields that are the columns, and unflattened events that are the Debezium complex events that have an envelope with a before and after `Struct` where you will need to apply the prefix/suffix/case to the fields of these nested before/after Structs.

Does that make sense @gustavolira and @rk3rn3r does that better align with you?

rk3rn3r:
@Naros almost. I don't feel good with passing/injecting the entire config as dependency.
Do we have an alternative? Like passing an immutable Java `record` as parameter instead of the entire config?
The record or whatever value object we are passing is then an explicit subset of the config options required for a global, shared class (it might eventually be shared in the `debezium-sink` module).

Naros:
> @Naros almost. I don't feel good with passing/injecting the entire config as dependency. Do we have an alternative? Like passing an immutable Java `record` as parameter instead of the entire config? The record or whatever value object we are passing is then an explicit subset of the config options required for a global, shared class (it might eventually be shared in the `debezium-sink` module).

We could certainly setup a configuration namespace, i.e. `column.naming.*`, where we would pull all the subset configs from that namespace and pass them, sure. I just think whatever we do needs to be generalized enough that it can be extendable without code changes for situations where a new strategy needs a specific configuration that the user can pass and the strategy use, but that has no influence on the remainder of the sink (think transforms or driver properties).

jpechane:
As the naming strtageies are quite deep in the core of Debezium anyway I think it does not make any harm to expose complete connector configuration. But this is an exception to the rule.

rk3rn3r:
@jpechane No that is inaccurate. You are mistaken source connector *NamingStrategies with those affected in this PR. Those here are only in the JDBC sink module and the `debezium-sink` module. Do you hold your previous statement under this condition?

jpechane:
@rk3rn3r Right, in this case it is not that clear cut as I originally thought. Also when I went back to Jira it epxlicitly states that prefix defined subset of config options should be passed.

Naros:
@gustavolira is this something you have the cycles for?

gustavolira:
@Naros  Yes, I have the cycles to make these changes. I'll start by splitting the PR as suggested, and will follow up with the necessary refactorings. I'll keep you posted on progress.

Naros:
Superseded by PR #6249

PR #6055, opened 2024-12-15
DBZ-8514 TimezoneConverter include.list should be respected if set

martin-walsh:
Fix for what I would consider to be a bug in TimezoneConverter `include.list`.

If you specify entries e.g. 
```
transforms.convertTimezone.include.list=source:table_one:DATE_CREATED,table_one:LAST_UPDATED,table_two:CREATED_AT,table_two:UPDATED_AT
```

and an event from `table_three` is received, the TimezoneConverter will currently apply the conversion to all columns within. The fallback from not finding a table match in config is to convert all columns, instead of not converting any of them.

This leaves you with no way to opt in to select columns on select tables without adding workaround config e.g. `table_three:workaround_cfg`.

See https://issues.redhat.com/browse/DBZ-8514

### DIFF
--- a/debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
@@ -519,9 +519,6 @@ public class TimezoneConverter<R extends ConnectRecord<R>> implements Transforma
                 handleStructs(value, Type.ALL, matchName, fields);
             }
         }
-        else {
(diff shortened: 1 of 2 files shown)


martin-walsh:
See https://github.com/debezium/debezium/pull/6051 for PR for 2.7.x line.

Naros (approved):
LGTM, let's check if CI passes.

jpechane:
@martin-walsh Applied, thanks!

PR #6058, opened 2024-12-17
DBZ-8291 DBZ-6709 Handle store.only.captured.tables.ddl regression

ani-sha:
closes https://issues.redhat.com/browse/DBZ-8291 and https://issues.redhat.com/browse/DBZ-6709

### DIFF
--- a/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogSnapshotSourceIT.java
+++ b/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogSnapshotSourceIT.java
@@ -57,7 +57,6 @@ import io.debezium.junit.logging.LogInterceptor;
 import io.debezium.relational.RelationalDatabaseConnectorConfig;
 import io.debezium.relational.RelationalDatabaseConnectorConfig.SnapshotTablesRowCountOrder;
 import io.debezium.relational.history.MemorySchemaHistory;
-import io.debezium.relational.history.SchemaHistory;
 
 /**
  * @author Randall Hauch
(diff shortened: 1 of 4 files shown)


debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogSnapshotSourceIT.java
```diff
@@ -155,7 +154,7 @@ private void snapshotOfSingleDatabase(boolean useGlobalLock, boolean storeOnlyCa
                     .with(BinlogConnectorConfig.USER, "cloud")
```
ani-sha, line 154: Modification in the test is required due to the conditional behavior of `useGlobalLock` here. Initially `STORE_ONLY_CAPTURED_TABLES_DDL` was set as true by default irrespective of the lock but now only this condition sets the config option.

jpechane:
@ani-sha Rebased and force-pushed

ani-sha:
Hi @mfvitale, I am uncertain about this failure. Do you believe it's connected to the change? Thanks.
```
2024-12-18T07:42:45.5123349Z [ERROR] NotificationsIT>AbstractNotificationsIT.notificationNotSentIfNoChannelIsConfigured:135->AbstractConnectorTest.assertNoRecordsToConsume:1029 expected:<[tru]e> but was:<[fals]e>
```

jpechane:
@ani-sha Hi, yes, it is related. You need to run whole test `NotificationsIT` locally, test method in isolation won't show it. It seems that the test expects no messages but probably schema changes are sent.

jpechane:
@ani-sha I tried to push a cumiit with the fix.

ani-sha:
Thanks @jpechane, it is working!

PR #6061, opened 2024-12-18
DBZ-8533 Always skip GoldenGate REPLICATION MARKER events

Naros:
https://issues.redhat.com/browse/DBZ-8533


Naros:
@jpechane the Oracle 21 error is unrelated, this seems good to merge for 3.0.5 if you're good w/it.

PR #6066, opened 2024-12-20
Add missing quick profile in test containers module

vjuranek:
https://issues.redhat.com/browse/DBZ-8545


jpechane:
@vjuranek Applied, thanks

PR #6068, opened 2024-12-24
Fix default value of `poll.interval.ms` in MongoDB connector docs

DominiQN:
The default value of `poll.interval.ms` is incorrectly documented as 1000ms in some parts of the documentation.

According to the source code and description, the correct default value is 500ms.

https://github.com/debezium/debezium/blob/6246808308200ae9877f25f3a04e378578def98c/debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java#L642-L650

https://github.com/debezium/debezium/blob/6246808308200ae9877f25f3a04e378578def98c/debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java#L572

image


mfvitale:
Hi @DominiQN thanks for spotting this. Can you please prefix your commit with `[docs]`?

DominiQN:
Hi @mfvitale, I’ve updated the commit message. Thank you!

mfvitale:
@DominiQN, applied. Thanks!

PR #6069, opened 2024-12-27
DBZ-8274 Azure Managed Identity support for Azure Blob Schema History

seanshi007:
https://issues.redhat.com/browse/DBZ-8274

### DIFF
--- a/debezium-bom/pom.xml
+++ b/debezium-bom/pom.xml
@@ -45,10 +45,13 @@
         <version.jedis>4.1.1</version.jedis>
 
         <!-- S3 -->
-        <version.s3>2.17.241</version.s3>
+        <version.s3>2.26.30</version.s3>
 
         <!-- Blob Storage -->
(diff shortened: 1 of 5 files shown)


Naros (commented):
Hi @seanshi007 overall this LGTM, just one housekeeping task remains.

Would you please update the documentation in the page:
`documentation/modules/ROOT/pages/operations/debezium-server.adoc`

You would need to include the new property(ies) & their respective descriptions, along with a small indicator that the `account.name` is used if and only if the `connectionstring` is omitted.

jpechane:
@seanshi007 Applied, thanks!

seanshi007:
@jpechane Sorry just noticed that S3SchemaHistoryIT failed after s3 sdk version bumped to 2.18+, created this issue https://issues.redhat.com/browse/DBZ-8569 which includes a detailed description why it fails. This is the PR https://github.com/debezium/debezium/pull/6089 to fix it, please help take a look when you get a chance. Thanks!

PR #6071, opened 2025-01-02
DBZ-8556 Upgrade logback to 1.2.15

jpechane:


jpechane:
Newer logback require slf4j 2.x that is incomaptible with 1.7.x provided by Kafka Connect. We need to align on the Connect version so upgrade is not straightforward.

PR #6074, opened 2025-01-03
DBZ-8270 Format DDL grammars with upstream version

ani-sha:
closes https://issues.redhat.com/browse/DBZ-8270


jpechane:
@ani-sha Our grammar is still significantly different from the upstream one. It seems they have addiitonal formatting instructions added in it. Could you please take a look and ideally make it look all the same - the only difference being our extensions?

ani-sha:
@jpechane I used the correct formatting options, and the grammar looks pretty much aligned (but there are a few differences, especially for MySqlParser, as I believe the ANTLR formatting was done long ago, and we are formatting it on our latest grammar). Please check if that's okay and let me know. Wdyt?

jpechane:
@ani-sha Could you please re-format the upstream grammar too and do the final comparison?

ani-sha:
@jpechane After formatting and comparing with the upstream, here are the stats 

- MySqlLexer ~ 15 additions; 10 deletions 
- MySqlParser ~ 72 additions; 29 deletions
- MariadbLexer ~ 4 additions; 0 deletions
- MariadbParser ~ 24 additions; 2 deletions

I just formatted oracle, but the grammar differences with the upstream are huge. I think the formatting is just fine.

debezium-ddl-parser/src/test/resources/mysql/examples/ddl_alter.sql
```diff
@@ -1,54 +1,49 @@
 #begin
 -- Alter Table
-alter table ship_class add column ship_spec varchar(150) first, add somecol int after start_build, algorithm=instant;
+alter table ship_class add column ship_spec varchar(150) first, add somecol int after start_build;
```
Naros, line 3: The `algorithm=instant` is valid for MySQL.

I understand the goal is to synchronize with upstream, but we are doing a disservice to ourselves if we take the upstream as-is without comparing the differences and those that are valid being retained downstream and sent upstream as a part of this effort. That's the only way we can avoid introducing any potential grammar regressions, no?

I haven't checked all the example SQL files, but we likely should.

Naros (commented):
Hi @ani-sha, I've left a few inline comments.

debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mariadb/generated/MariaDBParser.g4
```diff
@@ -1663,7 +1677,7 @@ privilege
     : ALL PRIVILEGES?
     | ALTER ROUTINE?
     | CREATE (TEMPORARY TABLES | ROUTINE | VIEW | USER | TABLESPACE | ROLE)?
-    | DELETE
+    | DELETE (HISTORY)? // HISTORY is MariaDB-specific
```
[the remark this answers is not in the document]
ani-sha: Yes, I had the same thought, but the upstream grammar has the comments; so I left them unchanged.

debezium-connector-binlog/src/test/resources/ddl/mysql-test-create.ddl
```diff
@@ -460,7 +460,7 @@ ALTER TABLE MBR_POL_USERS
 
 
 
-CREATE OR REPLACE VIEW  MBR_READ_ENTRIES (ENTRY_ID_P1,ENTRY_ID_P2,USER_NAME) AS 
+CREATE VIEW  MBR_READ_ENTRIES (ENTRY_ID_P1,ENTRY_ID_P2,USER_NAME) AS
```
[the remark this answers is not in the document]
ani-sha: Actually, the `createView` clause had the `orReplace is MariaDB-specific only` comment, and I didn't check with the MySQL 8 syntax. I will add it back and edit the tests, too. I will also make sure to check the other clauses if there are any missed cases.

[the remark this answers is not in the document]
jpechane: Ideally we should end up in the state where we would take the upstream grammar as drop-n replacement and apply minimum changes (IIRC for Percona there might be 2) and be done.

debezium-ddl-parser/src/test/resources/mysql/examples/ddl_alter.sql
ani-sha: The idea was to synchronize with upstream and add the remaining downstream-specific examples/grammars to a separate file. So, I have added them to a new file, `debezium_tests.sql`.

However, this would cause grammar duplication with subtle differences in the two files.

Naros: So ultimately I'll defer to @jpechane if he disagrees, but this `debezium_tests.sql` file would ideally only be the bits that are different between upstream/downstream that apply to things like Percona or unusual recordings of DDL that we say cannot confirm via the documented grammar. For example, think of the work we did for the Oracle data pump and data loader rules we added for it.

Ideally, if the documented grammar supports something, such as the `algorithm=instant` bit, that I would think belongs in the appropriate upstream SQL test files and contributed upstream, no?

ani-sha: Yes, @Naros, you are correct. After discussing this with Jiri yesterday, we clarified the final changes. I will contribute the differences to the upstream and add the specific downstream extensions to `debezium_tests.sql`. Additionally, I will separate the Oracle formatting into a separate PR and work on it after we have resolved the MariaDB and MySQL grammars.

ani-sha:
@jpechane The upstream PRs have been merged, and the grammar is completely synchronized. Please give a final set of reviews, and if you find anything I missed in the grammar/tests, please let me know, and I will fix it.

jpechane (approved):
@ani-sha Thanks, this will make the maintenance way easier. I rebase on the latest main. If JDBC tests passes then the PR is good to go.

jpechane:
@ani-sha And it is in, great job!

PR #6075, opened 2025-01-04
DBZ-8557: Implement keyset pagination in sqlserver connector

ramanenka:
Sqlserver connector requires unbounded memory to process big transactions. This happens due to the fact that `mssql-jdbc` driver buffers any not yet consumed result sets in memory in order to execute the next select statement. On the other hand the connector doesn't limit the number of rows it selects. This leads to almost all transaction rows being buffered in memory.

An obvious solution to that is to limit the number of rows selected. This patch does exactly that. It introduces a new connector setting called `streaming.fetch.size` (by analogy with `snapshot.fetch.size`) with the default value of `0` which means the old behavior is preserved.

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/Lsn.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/Lsn.java
@@ -22,6 +22,8 @@ public class Lsn implements Comparable<Lsn>, Nullable {
 
     public static final Lsn NULL = new Lsn(null);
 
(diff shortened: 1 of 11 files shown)


jpechane:
@ramanenka Rebease on the latest main and force-pushed

jpechane:
@ramanenka Great job, applied!

jpechane:
@ramanenka One more thing. There is a need for a Db2 PR to align the API changes.

ramanenka:
@jpechane Sure, I'll look into that tomorrow.

PR #6089, opened 2025-01-10
DBZ-8569 Fix S3SchemaHistoryIT

seanshi007:
https://issues.redhat.com/browse/DBZ-8569


jpechane:
@seanshi007 Applied, thanks!

PR #6092, opened 2025-01-13
DBZ-8564 Fix test failures using non-KRaft deployments

Naros:
https://issues.redhat.com/browse/DBZ-8564

Fixes issue with JDBC tests that use non-KRaft deployments.


Naros:
Test failures are unrelated.

PR #6093, opened 2025-01-13
[docs] Bind Kafka container to localhost in podman

vjuranek:
Kafka otherwise bind to some randomly assigned IP address and further step in the tutorial, which starts Kafka Connect, fails with connection timeout exception.


AlvarVG (approved):

jpechane:
@vjuranek Applied, thanks

rk3rn3r:
@vjuranek what's the issue with the random IP address? Don't we rely on hostnames for connecting?

AlvarVG:
Hey @rk3rn3r, I was running the tutorial and there was no way the connector connects to kafka unless i set that environment. This is the fastest way @vjuranek found to make it work using podman.

PR #6095, opened 2025-01-14
DBZ-8572 cover additional scenario with empty arrays

regular154:


regular154:
issue: https://issues.redhat.com/browse/DBZ-8572

jpechane:
@ani-sha FYI

ani-sha:
@regular154 Can you reformat the PR and push the changes again so the CI can run?

ani-sha:
@jpechane If the issue is urgent, it can be a temporary fix; otherwise, I will test it and add it to the restructuring PR while working on it.

jpechane:
@ani-sha IMHO we can do both as it is more or less done. So when finished it can go to 3.1 and be backported to 3.0. And you could incroporate it ito your PR which will go to 3.1 only.

jpechane:
@regular154 Test failures unrelated, applied. Thanks!
@ani-sha You can incorporate this case into your PR.

PR #6097, opened 2025-01-14
DBZ-6387 Document `isostring`, `microseconds`, `nanoseconds` Temporal Precision Modes

ismailsimsek:


ismailsimsek:
@Naros @jpechane its ready for review

jpechane:
@roldanbob Could you please take a look? Thanks

roldanbob:
> @roldanbob Could you please take a look? Thanks

Apologies. I was focused on other items this week and I just realized that I neglected this. I'll work on this first thing Monday.

documentation/modules/ROOT/pages/connectors/postgresql.adoc
```diff
@@ -1686,6 +1689,101 @@ Represents the number of milliseconds since the epoch, and does not include time
...
+|`INT32`
+|`io.debezium.time.Date` +
+ +
+Represents the number of days since the epoch.
+
+|`TIME([P])`
+|`INT64`
+|`io.debezium.time.NanoTime` +
+ +
+Represents the time value in nanoseconds and does not include timezone information. PostgreSQL allows precision P to be in the range 0-6 to store up to microsecond precision.
```
roldanbob, line 1776:
```suggestion
Represents the time value in nanoseconds and does not include timezone information. 
In PostgreSQL, the precision `p` parameter specifies the number of decimal places in the seconds part of a time value.
The precision can range from 0 (no fractional seconds) to 6 (microsecond precision).
```

documentation/modules/ROOT/pages/connectors/postgresql.adoc
```diff
@@ -1686,6 +1689,101 @@ Represents the number of milliseconds since the epoch, and does not include time
...
+
+|`TIMESTAMP([P])`
+|`INT64`
+|`io.debezium.time.NanoTimestamp` +
+ +
+Represents the number of nanoseconds past the epoch, and does not include timezone information. PostgreSQL allows precision P to be in the range 0-6 to store up to microsecond precision.
```
roldanbob, line 1782:
```suggestion
Represents the number of nanoseconds past the epoch, and does not include timezone information.
In PostgreSQL, the precision `p` parameter specifies the number of decimal places in the seconds part of a time value.
The precision can range from 0 (no fractional seconds) to 6 (microsecond precision).
```

roldanbob (changes requested):
@ismailsimsek I've added some minor wording and style suggestions. Let me know if you have any questions or comments. Thanks for the contribution!

documentation/modules/ROOT/pages/connectors/postgresql.adoc
```diff
@@ -1686,6 +1689,112 @@ Represents the number of milliseconds since the epoch, and does not include time
...
+ +
+Represents the number of microseconds past the epoch, and does not include timezone information. 
+In PostgreSQL, the precision `p` parameter specifies the number of decimal places in the seconds part of a time value.
+The precision can range from 0 (no fractional seconds) to 6 (microsecond precision).
+
+|===
+
+[[postgresql-time-precision-mode-nanoseconds]]
+.`time.precision.mode=nanoseconds`
+Set the `time.precision.mode` property to `nanoseconds`  to configure the connector to specify temporal values with nanosecond precision.
```
roldanbob, line 1763:
```suggestion
Set the `time.precision.mode` property to `nanoseconds` to configure the connector to specify temporal values with nanosecond precision.
```

roldanbob: Removes extra space

roldanbob (changes requested):
Except for one extra space on L1763, this LGTM.  Thanks for the quick turnaround. Sorry for holding this up by taking so long to do the initial review!

jpechane:
@ismailsimsek @roldanbob Squahed and applied, thanks!

Naros:
@jpechane these don't just apply to PostgreSQL right, so we need to adjust the documentation for all connectors?

ismailsimsek:
> @jpechane these don't just apply to PostgreSQL right, so we need to adjust the documentation for all connectors? @roldanbob FYI

i feel like this section of config should go to a global documentation page a connectors agnostic place. `Configuration/` section could be good place to add it.

roldanbob:
> i feel like this section of config should go to a global documentation page a connectors agnostic place. `Configuration/` section could be good place to add it.

I like that idea. If we were to implement that, _ideally we would really need to evaluate all of the connector config properties to identify which are supported globally_. Then we could _document them **all** in one place_, extracting the properties tables/lists from out of the individual connector files and placing those in a separate configuration file/chapter/"book". Currently, there is a fair amount of duplication (and inconsistency) among the various connector chapters. In the downstream docs, it might make sense to reorganize contents so that all of the properties information is presented in a separate guide, for example, _Configuring Debezium connectors_, or such-like. 

At one point I did some work to identify which properties were universal, and to try to establish some kind of hard line between which properties are required and which are "_advanced_", but I didn't have enough time or knowledge.
I think that I still have the spreadsheet that I created when I was reviewing the properties, although it's a bit out of date. It would be great if we could put our heads together to sort the properties to figure out which are common across all connectors, and then further differentiate them between required and advanced. Or just do away with that distinction if it really is too fuzzy to be meaningful.

If we started with a common location, rather than trying to share properties lists/tables via `include` directives, it would be much easier than trying to conditionalize the properties to show only those relevant to in a given connector context.   

Thoughts?

Naros:
@roldanbob I am certainly not opposed to finding a better way to represent the connector options to minimize the overhead caused by duplication as we do today, but I think we have to be careful about the approach.

For example, I think it makes sense to have things like the database, schema, and table include/exclude list properties next to one another as they describe how you define the connector's object capture set; however, not every connector uses these. For example, Oracle/PostgeSQL only use schema/table ones while MySQL/MariaDB uses database/table ones. 

And then there is a question of how are the values interpreted, even if the property is common. For example, for Oracle and PostgreSQL, you define the table include/exclude lists as `<schema-name>.<table-name>` whereas for MySQL we use the semantic description of `<database-name>.<table-name>`. 

Then we have common connector properties, but their values are interpreted differently, i.e. one may want the value to be `<database-name>.<schema-name>.<table-name>` while another wants a slightly different variant.

But I wholeheartedly agree that having properties grouped by section/relevance can help.  For example, I like the required and non-required separately we have today, but I do wonder if that separation is also in part attributing to users being unaware of non-required fields that can have some substantial impact on performance.

In the end, there won't be a solution that solves all the problems, and I think whatever makes the ingestion of the data easiest while keeping the maintenance of it easy for up/downstream, that's likely the best route.

roldanbob:
> @roldanbob To be honest that's something I've always dreamt of. We could you the source code oragnization - `CommonConnectorConfig` etc. to define the groups. Could the discrepancies mentioned by @Naros be handled via placeholder like we do with `data-collection` for incremental snapshot patials?

That's a good thought, but I think that it's only an option if we're bringing the shared content into different connector contexts, so that we can define unique attribute values per connector. But if we were to change to a model where we delivered the properties lists/tables outside of the connector chapters, in its own guide, we wouldn't be able to apply different values. At least not with the current in-flux tooling. I was wondering aloud to Chris the other day about whether the new tooling might support tabbed content, whereby the reader could choose specific versions of information based on their interest/focus, for example, show me what this looks like for SQL Server vs. MongoDB. We don't know yet, but we'll investigate it as a possibility.

PR #6002, opened 2024-11-20
DBZ-6387 Addding new isostring, microseconds, nanoseconds TemporalPrecisionModes

ismailsimsek:
Adding 3 new non **adaptive** temporal modes:
- `isostring`: converts to UTC ISO 8601
- `microseconds`: converts to int64, always with microseconds precision mode. not adaptive
- `nanoseconds`: converts to int64, always with nanoseconds precision mode. not adaptive

Implements [DBZ-6387](https://issues.redhat.com/browse/DBZ-6387)

remaining todos
- [x] update code comments
- [x] add it to postgresql array types as well

Requires https://github.com/debezium/debezium-connector-vitess/pull/219

PR #6098, opened 2025-01-14
DBZ-8564 Switch percona server to quay.io

vjuranek:


Naros:
The JDBC test failures are unrelated to this change.  I'll take a look at why these failed after I wrap up the vector database change PR @jpechane and @vjuranek .

jpechane:
@Naros It relates to changes intorduced in https://github.com/debezium/debezium/pull/6002

jpechane:
@vjuranek Applied, thanks

PR #6099, opened 2025-01-15
DBZ-6387 Disable testing with `isostring` temporal precision mode

Naros:
https://issues.redhat.com/browse/DBZ-6387

The JDBC sink tests were not designed to expect the temporal columns to be created as strings when `isostring` is set as the source's temporal precision mode. This PR disables that mode in the test suite for now, and a follow-up PR will update the tests to introduce this.


jpechane (approved):
@Naros LGTM, feel free to merge yourself.

Naros:
The one lone test failure is related, small test fix incoming for MICROSECOND precision w/SQL Server.

Naros:
The test failures are related to #6104, so I will proceed with the merge.

PR #6102, opened 2025-01-15
DBZ-8586 Implement Wasm accessors to all the Struct data types

andreaTP:
Opening in Draft as it's pending the update of the PDK.

cc. @jpechane


debezium-scripting/debezium-scripting/src/test/resources/wasm/go/go.mod
```diff
@@ -2,4 +2,7 @@ module example.com/debezium/filter
 
 go 1.22.7
 
+// TODO: remove me
+replace github.com/debezium/debezium-smt-go-pdk => ../debezium-smt-go-pdk/
```
andreaTP, line 6: Depends on: https://github.com/debezium/debezium-smt-go-pdk/pull/1

jpechane:
@andreaTP Could you please create a new Jira for this change and change the commit message to reflet the new Jira id?
I think we can now follow the standard process. Thanks!

debezium-scripting/debezium-scripting/src/main/java/io/debezium/transforms/scripting/wasm/ChicoryEngine.java
```diff
@@ -189,17 +196,97 @@ public int getString(int proxyObjectRef) {
     }
 
     @WasmExport
-    public int getInt(int proxyObjectRef) {
+    public int getBool(int proxyObjectRef) {
+        var proxyObject = objects.get(proxyObjectRef);
+        if (proxyObject.getClass().isAssignableFrom(Boolean.class)) {
```
jpechane, line 201: I wonder if we can use pattern matching here and at the other places instead of reflection.WDYT?

andreaTP: Changed to use `instanceof`, in this case the pattern matching is going to be always on a single element, so, an `if` is probably still the preferred construct for readability.
Not fussy on this detail though, if you prefer a different encoding just let me know :+1:

debezium-scripting/debezium-scripting/src/test/resources/wasm/go/go.mod
andreaTP: Now depends on https://github.com/debezium/debezium-smt-go-pdk/pull/2

andreaTP:
Updated everything and added array accessors, let me know if my mocked tests are correctly mimicking the runtime behavior of Kafka Connect not 100% sure about it.

debezium-scripting/debezium-scripting/src/test/resources/wasm/go/go.mod
andreaTP: updated

debezium-scripting/debezium-scripting/src/test/resources/wasm/go/router1.go
```diff
@@ -6,8 +6,8 @@ import (
 
 //export process
 func process(proxyPtr uint32) uint32 {
-	// 1
-	return debezium.SetInt(1)
+	// boolean
+	return debezium.SetBool(false)
```
andreaTP, line 10: Since we removed `SetInt` from the API it's hard for the user to pass an invalid value, the only mismatch can be across boolean and strings.

jpechane:
@andreaTP LGTM, applied thanks!
I've just added the `@Incubating` annotation se we are not bound by backward compatibility limits.
This fact should be mentioned in the docs too.

Now it is time for a blogpost! :-)

andreaTP:
Thanks @jpechane !

The annotation is an excellent idea :+1:

> Now it is time for a blogpost! :-)

I'll try to go through the docs and the blog sometime next week if it works for your timeline.

jpechane:
@andreaTP Of course, take your time!

PR #6103, opened 2025-01-16
DBZ-8583 Upgrade Outbox Extension to Quarkus 3.17.7

Naros:
https://issues.redhat.com/browse/DBZ-8583


jpechane:
@Naros Applied, thanks.

PR #6104, opened 2025-01-16
DBZ-8564 Fetch ryuk container from quay.io

vjuranek:


jpechane:
@vjuranek Cant we use `mirror.gcr.io/testcontainers/ryuk` ?

Naros:
@vjuranek can we also add this for the JDBC sink, if the one from the testing module doesn't influence it?

vjuranek:
@jpechane sure, I thought that quay.io is preferred place as you mentioned you have hit the limits also with gcr.io
@Naros yes, I'll add it and there will be more (possibly also outbox modules) - that's why it's a draft, as it's not complete yet. I'm running the testsuite on a machine without access to dockerhub now, as will update the PR as the testsuite fails.

rk3rn3r:
Do we need to customize other image sources too? https://java.testcontainers.org/features/configuration/#customizing-images

mfvitale:
We were discussing it on the zulip thread. I just tagged you.

rk3rn3r:
thx @mfvitale

PR #6107, opened 2025-01-18
DBZ-8549 Adds downstream JDBC connector deployment instructions

roldanbob:
[DBZ-8549](https://issues.redhat.com/browse/DBZ-8549)  Adds downstream deployment instructions to the JDBC connector doc.

Tested in a local Antora build. These changes have no effect on the community version of the documentation. 

This is ready for upstream review, but should not be merged yet, pending resolution of downstream build issues.

I would especially appreciate it if someone could review the CR examples. I modeled these examples after the source connector examples where I assumed that made sense, but I was unsure where to include certain entries, for example,
where do we define the `topics` array that lists the Kafka topics from which the connector consumes change events? 

I placed this in the connector YAML examples (L753 and L994), e.g., 

```
spec:
  class: io.debezium.connector.jdbc.JdbcSinkConnector  // <2>
  tasksMax: 1  // <3>
  topics: // <4>
    - orders
```

Is that correct?

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -756,14 +756,14 @@ metadata:
 spec:
   class: io.debezium.connector.jdbc.JdbcSinkConnector  // <2>
   tasksMax: 1  // <3>
-  topics: orders // <4>
-   config:  // <5>
+   config:  // <4>
     connector.class: io.debezium.connector.jdbc.JdbcSinkConnector
-    connection.url: jdbc://postgresql://<host>:<port>/<database-name> // <6>
-    connection.username: <database-user>  // <7>
-    connection.password: <database-pwd>  // <8>
+    connection.url: jdbc://postgresql://<host>:<port>/<database-name> // <5>
+    connection.username: <database-user>  // <6>
+    connection.password: <database-pwd>  // <7>
     insert.mode: upsert
     delete.enabled: true
+    topics: orders // <8>
     primary.key.mode: record_key
     schema.evolution: basic
     use.time.zone: UTC
@@ -785,24 +785,22 @@ spec:
 |The number of tasks that can operate concurrently.
 
 |4
-|Specifies a comma-separated list of Kafka topics that the connector reads.
-Events from each topic are streamed to tables with the same name in the sink database.
-
-|5
 |The connector’s configuration.
 
-|6
-|The JDBC address of the sink database.
+|5
+|The JDBC connection URL for the sink database.
+The URL specifies the port number and any other authentication properties that are required to connect to the database.
+For example, `jdbc:oracle:thin:@myhost.example.com:1521/myservice`
 
-|7
+|6
 |The name of the account that {prodname} uses to connect to the database.
 
-|8
+|7
 |The password that {prodname} uses to connect to the database user account.
 
-|9 ...
-|The connection URL specifies the port number and any other authentication properties that are required to connect to the database.
-For example, `jdbc:oracle:thin:@myhost.example.com:1521/myservice`
+|8
+|Specifies a comma-separated list of Kafka topics that the connector reads.
+Events from each topic are streamed to tables with the same name in the sink database.
 
 |===
 
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -2,8 +2,10 @@
 // Type: assembly
 [id="debezium-connector-for-jdbc"]
 = {prodname} connector for JDBC
-:context: JDBC
+:context: jdbc
 :mbean-name: {context}
+:connector-file: {context}
+:connector-name: JDBC
 ifdef::community[]
 :toc:
 :toc-placement: macro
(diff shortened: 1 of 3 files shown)


documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -544,30 +546,500 @@ endif::community[]
 [[jdbc-deployment]]
 == Deployment
 
+ifdef::community[]
 To deploy a {prodname} JDBC connector, you install the {prodname} JDBC connector archive, configure the connector, and start the connector by adding its configuration to Kafka Connect.
 
 .Prerequisites
 * link:https://zookeeper.apache.org/[Apache ZooKeeper], link:http://kafka.apache.org/[Apache Kafka], and link:{link-kafka-docs}.html#connect[Kafka Connect] are installed.
-* A destination database is installed and configured to accept JDBC connections.
+* Source and destination databases are installed and are configured to accept JDBC connections.
```
Naros, line 554: I don't see the need to reference a "Source" in this context, as the JDBC sink connector can be used with non-database sources.

Naros (approved):
Hi @roldanbob overall this LGTM, I just left one small nit-pick.

Naros:
As to your question, the `topics` shouldn't be an array as you implemented, but AFAICT its just a connector configuration that is a comma-separated list of values in the connector configuration section, just like you would provide this as part of the JSON, as shown [here](https://debezium.io/documentation/reference/3.0/connectors/jdbc.html#jdbc-connector-configuration).

roldanbob:
I addressed the review feedback, but I still need to resolve the downstream build issues.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -753,8 +756,7 @@ metadata:
 spec:
   class: io.debezium.connector.jdbc.JdbcSinkConnector  // <2>
   tasksMax: 1  // <3>
-  topics: // <4>
-    - orders
+  topics: orders // <4>
```
Naros, line 759: This needs to be moved into the `config` section where we define things like `connector.class` and `connection.url`.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -544,30 +546,503 @@ endif::community[]
...
+=====================================================================
+[source,yaml,subs="+attributes"]
+----
+apiVersion: kafka.strimzi.io/v1beta2
+kind: KafkaConnector
+metadata:
+  name: orders-topic-to-postgresql-via-jdbc-sink-connector // <1>
+  labels: strimzi.io/cluster: debezium-kafka-connect-cluster
+spec:
+  class: io.debezium.connector.jdbc.JdbcSinkConnector  // <2>
```
roldanbob, line 757:
```suggestion
  class: io.debezium.connector.jdbc.JdbcSinkConnector  // <2>
```
@Naros Should I remove this entry (L757) in the `spec` section?  It appears to be a duplicate of L761, but without the `connector-` prefix in the property name. This same entry appears downstream in the `spec` section of `KafkaConnector` examples for other connectors, for example,  in the MariaDB and MySQL docs

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -544,30 +546,503 @@ endif::community[]
...
+----
+apiVersion: {KafkaConnectApiVersion}
+kind: KafkaConnector
+metadata:
+  labels:
+  strimzi.io/cluster: my-connect-cluster // <1>
+  name: orders-topic-to-postgresql-via-jdbc-sink-connector // <2>
+...
+spec:
+  class: io.debezium.connector.jdbc.JdbcSinkConnector // <3>
```
roldanbob, line 992:
```suggestion
  class: io.debezium.connector.jdbc.JdbcSinkConnector // <3>
```
As with L757, I suspect that this `spec.class` entry should be a `config.connector.class` entry.

roldanbob (commented):
@Naros I posted a couple of questions related to `spec.class` entries in the examples.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
Naros: Unless something has changed, you only need L757.  I'm not aware of the `connector.class` property inside the `config` section at L761.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
Naros: Just to be sure, I just checked the `KafkaConnector` CRD yaml, and according to [this](https://github.com/strimzi/strimzi-kafka-operator/blob/main/install/cluster-operator/047-Crd-kafkaconnector.yaml), you need `spec.class` but you don't or would use `config.connector.class`.

roldanbob: > Just to be sure, I just checked the `KafkaConnector` CRD yaml, and according to [this](https://github.com/strimzi/strimzi-kafka-operator/blob/main/install/cluster-operator/047-Crd-kafkaconnector.yaml), you need `spec.class` but you don't or would use `config.connector.class`.

Thanks, @Naros. So, I think that part of the confusion surround this is that we list the following property in all of the source connector docs: `connector.class`, for example, this instance in the Db2 doc: https://github.com/debezium/debezium/blame/main/documentation/modules/ROOT/pages/connectors/db2.adoc#L2443
Also, in the JDBC doc the existing upstream example includes a reference to `connector.class`:  https://github.com/debezium/debezium/blame/main/documentation/modules/ROOT/pages/connectors/jdbc.adoc#L617

roldanbob: Is my understanding of what should be there correct? We only need `spec.class`, right? 

I was thrown off by what might be a typo when you said, 

> ... you need `spec.class` but you don't _or would use_ `config.connector.class`. [emphasis mine]

Is it that you would use `config.connector.class` if you don't specify `spec.class`?

Naros: So I double checked with Jakub from Strimzi and verified this also in the [docs](https://strimzi.io/docs/operators/latest/configuring.html#type-KafkaConnectorSpec-reference).

According to the docs, the `config` property is a `map` and it says:
> The Kafka Connector configuration. The following properties cannot be set: name, connector.class, tasks.max.

Perhaps in older Strimzi versions this wasn't the case, but at least based on 0.45, the documentation discourages and claims those shouldn't and cannot be set.

roldanbob: Ah, ok. So I'm guessing that where we specify `config.connect.class` in the community version of the instructions, that works, because in that context Strimzi is not part of the environment, yes?

roldanbob: To prevent confusion, for now, I can conditionalize the `connector.class` entries in the properties tables to apply to the community version only. WDTY?

roldanbob:
Closed in favor of #6121.

roldanbob:
Closed in favor of #6121

PR #6108, opened 2025-01-20
DBZ-8558 Correct schema.history.internal.store.only.captured.databases.ddl default value for MariaDB/MySQL

ani-sha:
https://issues.redhat.com/browse/DBZ-8558

### DIFF
--- a/debezium-core/src/main/java/io/debezium/relational/history/SchemaHistory.java
+++ b/debezium-core/src/main/java/io/debezium/relational/history/SchemaHistory.java
@@ -69,8 +69,8 @@ public interface SchemaHistory {
             .withWidth(Width.SHORT)
             .withImportance(Importance.LOW)
             .withDescription("Controls what DDL will Debezium store in database schema history. "
-                    + "By default (true) only DDL that manipulates a table from captured schema/database will be stored. "
-                    + "If set to false, then Debezium will store all incoming DDL statements.")
+                    + "By default (false) Debezium will store all incoming DDL statements. "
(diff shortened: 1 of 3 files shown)


jpechane:
@ani-sha @Naros I propose to step back and agree on the proper default for both MariaDB and MySQL - maybe have a short call. The issue seems to be a bit fuzzy now.

ani-sha:
@jpechane Just to clarify my further doubts I checked the details of DBZ-6709.

DBZ-6709 describes a regression in `STORE_ONLY_CAPTURED_TABLES_DDL`, but it's referring to `STORE_ONLY_CAPTURED_DATABASES_DDL`, which was misworded throughout the ticket. I worked on it in the previous PR, removing the regression. My mistake was overlooking the full property names.

So, these document changes clarify the correct default values. I have also reverted the test changes I made for `STORE_ONLY_CAPTURED_TABLES_DDL` in the other PR.

jpechane:
@Naros I wonder if we should just add this into release notes or maybe add a warning to MariaDB page as well?

Naros:
> @Naros I wonder if we should just add this into release notes or maybe add a warning to MariaDB page as well?

If I put my bureaucratic :tophat:  on, I'd say I'm inclined to avoid the documentation route mainly because we used to do the same around tracking what properties were deprecated, scheduled for removal, and what versions they were added in back when our documentation wasn't versioned, and since then we refrained from that. To me, this smells like a similar pattern that I wonder if we shouldn't just avoid to keep things streamlined.

So in my humble opinion, I'd err on the side of just the release notes.

jpechane:
@ani-sha Applied, thanks

PR #6110, opened 2025-01-21
DBZ-8564 Use better way how to specify dockerhub mirror

vjuranek:


debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/TestInfrastructureHelper.java
```diff
@@ -208,7 +208,7 @@ public static void setupDebeziumContainer(String connectorVersion, String restEx
         final String registry = debeziumContainerImageVersion.startsWith("1.2") ? "" : "quay.io/";
         final String debeziumVersion = debeziumContainerImageVersion.startsWith("1.2") ? "1.2.5.Final" : connectorVersion;
         String baseImageName = registry + "debezium/connect:nightly";
-        DEBEZIUM_CONTAINER = new DebeziumContainer(new ImageFromDockerfile("debezium/connect-rest-test:" + debeziumVersion)
+        DEBEZIUM_CONTAINER = new DebeziumContainer(new ImageFromDockerfile("quay.io/debezium/connect-rest-test:" + debeziumVersion)
```
rk3rn3r, line 211:
```suggestion
        DEBEZIUM_CONTAINER = new DebeziumContainer(new ImageFromDockerfile("debezium/connect-rest-test:" + debeziumVersion)
```

This is no actual image! This is just the identifier used when building this local container. I think this can safely be reverted!

vjuranek: No, as the commit message says, if there is no prefix of the image, TC tries (for whatever reason, no idea why as the image is created from Dockerfile) to pull the image from the docker registry mirror and of course fails.

rk3rn3r: Ah sure, it wants to base a build on a cache for different layers. Makes sense!
Then change for sure makes sense too!

rk3rn3r: I just learned the correct solution would have been using `localhost`. Will add this. FYI @vjuranek

vjuranek: TIL, thanks @rk3rn3r !

PR #6111, opened 2025-01-21
[docs] Add missing `include.schema.changes` configuration property

Naros:
@roldanbob just FYI - I copied the same from SQL Server, but this seems to be missing in Oracle docs.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/oracle.adoc
+++ b/documentation/modules/ROOT/pages/connectors/oracle.adoc
@@ -3791,7 +3791,9 @@ For example, if you set `max.queue.size=1000`, and `max.queue.size.in.bytes=5000
 
 |[[oracle-property-include-schema-changes]]<<oracle-property-include-schema-changes, `+include.schema.changes+`>>
 |`true`
-|Boolean value that specifies whether the connector should publish changes in the database schema to a Kafka topic with the same name as the database server ID. Each schema change is recorded with a key that contains the database name and a value that is a JSON structure that describes the schema update. This is independent of how the connector internally records database schema history. The default is `true`.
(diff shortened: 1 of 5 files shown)


documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -3789,6 +3789,10 @@ For example, if you set `max.queue.size=1000`, and `max.queue.size.in.bytes=5000
 |`500` (0.5 second)
 |Positive integer value that specifies the number of milliseconds the connector should wait during each iteration for new change events to appear.
 
+|[[oracle-property-include-schema-changes]]<<oracle-property-include-schema-changes, `+include.schema.changes+`>>
+|`true`
+|Boolean value that specifies whether the connector should publish changes in the database schema to a Kafka topic with the same name as the database server ID. Each schema change is recorded with a key that contains the database name and a value that is a JSON structure that describes the schema update. This is independent of how the connector internally records database schema history. The default is `true`.
```
roldanbob, line 3794:
```suggestion
|Boolean value that specifies whether the connector publishes changes in the database schema to a Kafka topic with the same name as the database server ID. 
The connector records each schema change with a key that contains the database name, and a value that is a JSON structure that describes the schema update. 
This mechanism for recording schema changes is independent of the connector's internal recording of changes to the database schema history. 
```
I know that you said this is copied directly from the SQL Server docs, but it's a good opportunity to make these updates. I can go into that file and copy these changes back to the source. WDYT?

roldanbob (changes requested):
@Naros Thanks for copying me on this. I'd like to apply a few changes:

- Separate sentences by line
- Tweak the wording a little. 
 
Obviously, these are not critical changes given that this text has been out there in the SQL Server doc, but since we're in there, maybe we can nudge things toward consistency.

roldanbob (approved):
LGTM

PR #6112, opened 2025-01-21
DBZ-8593 Fix ReducedRecord Buffer implementation in case of RECORD_VALUE mode

gaurav7261:
https://issues.redhat.com/browse/DBZ-8593

### DIFF
--- a/debezium-sink/src/main/java/io/debezium/bindings/kafka/KafkaDebeziumSinkRecord.java
+++ b/debezium-sink/src/main/java/io/debezium/bindings/kafka/KafkaDebeziumSinkRecord.java
@@ -6,7 +6,9 @@
 package io.debezium.bindings.kafka;
 
 import java.util.List;
+import java.util.Set;
 
+import org.apache.kafka.connect.data.Field;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.SchemaBuilder;
 import org.apache.kafka.connect.data.Struct;
@@ -122,10 +124,10 @@ public class KafkaDebeziumSinkRecord implements DebeziumSinkRecord {
     }
 
     @Override
-    public Struct getKeyStruct(SinkConnectorConfig.PrimaryKeyMode primaryKeyMode) {
+    public Struct getKeyStruct(SinkConnectorConfig.PrimaryKeyMode primaryKeyMode, Set<String> primaryKeyFields) {
         if (!keyFieldNames().isEmpty()) {
             switch (primaryKeyMode) {
-                case RECORD_KEY:
+                case RECORD_KEY -> {
                     final Schema keySchema = originalKafkaRecord.keySchema();
                     if (keySchema != null && Schema.Type.STRUCT.equals(keySchema.type())) {
                         return (Struct) originalKafkaRecord.key();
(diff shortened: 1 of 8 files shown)


gaurav7261:
@mfvitale  please review class SinkRecordFactory.java changes closely, as when returning SchemaBuilder class object, Reduced Record Buffer hashmap put hashcode method doesn't work as `SchemaBuilder` class don't have hashcode method, but ConnectSchema does, so changes are made so that `ConnectSchema` is return from it

Naros (approved):
LGTM from my PoV @gaurav7261

jpechane:
@gaurav7261 LGTM, thanks! I've one question. Is it possible to extend the assertion to verify the primary key definition?

gaurav7261:
sure @jpechane , adding that, thanks

debezium-sink/src/main/java/io/debezium/bindings/kafka/KafkaDebeziumSinkRecord.java
```diff
@@ -122,34 +124,51 @@ public Struct getPayload() {
...
+                        else {
+                            final SchemaBuilder keySchemaBuilder = SchemaBuilder.struct();
+                            primaryKeyFields.forEach(fieldName -> {
+                                Field field = afterPayload.schema().field(fieldName);
+                                keySchemaBuilder.field(field.name(), field.schema());
+                            });
+                            Struct keyStruct = new Struct(keySchemaBuilder.build());
+                            primaryKeyFields.forEach(fieldName -> {
+                                Field field = afterPayload.schema().field(fieldName);
+                                keyStruct.put(field.name(), afterPayload.get(field));
```
mfvitale, line 155: WDYT to just move this line in the other each loop, so that you can build the schema and the struct in just one iteration? In addition you can put this in a properly named method.

mfvitale:
@gaurav7261 LGTM, I just left a nitpik comment.

gaurav7261:
@jpechane are you talking about test case or some code flow, can you please help me at which file are you talking about

mfvitale:
> @jpechane are you talking about test case or some code flow, can you please help me at which file are you talking about

I suppose Jiri was talking about the `ReducedRecordBufferTest.java`. It will be good if you add to the config the list of primary key field and the assert that the records has the correct primary fields.

debezium-sink/src/main/java/io/debezium/bindings/kafka/KafkaDebeziumSinkRecord.java
gaurav7261: Hi @mfvitale , i have moved to a named method, but I'm not sure how we can do with single iteration, because as far as my understanding, I need to create schema with pkey list complete before building struct, or let me know if I'm missing something here

mfvitale: Yes, sorry you are right. Don't mind.

jpechane (approved):
@gaurav7261 That's what I was looking for, thanks!

@mfvitale If your comments are addressed then feel free to merge yourself,

mfvitale:
@gaurav7261 Applied, thanks!

gaurav7261:
Thanks @mfvitale , also @mfvitale  I want to know are we maintaining separate branch for 2.7, because classes name have been changed, I have changes for 2.7 as well, where can I raise that PR, can you please help in that flow as well

mfvitale:
@gaurav7261 Generally we maintain just one old release. In that case the 3.0 branch. I just backported the changes there. @jpechane Am I correct?

jpechane:
Yes, that's correct. You can prepare a PR against `2.7` branch but there are no plans for another 2.7 release so far

PR #6117, opened 2025-01-23
DBZ-8597 Fix hybrid column resolution for tables with generated columns

Naros:
https://issues.redhat.com/browse/DBZ-8597

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/parser/LogMinerColumnResolverDmlParser.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/parser/LogMinerColumnResolverDmlParser.java
@@ -5,8 +5,14 @@
  */
 package io.debezium.connector.oracle.logminer.parser;
 
+import java.util.HashMap;
+import java.util.Map;
+
+import io.debezium.DebeziumException;
 import io.debezium.connector.oracle.OracleConnectorConfig;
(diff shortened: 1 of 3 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/parser/LogMinerColumnResolverDmlParser.java
```diff
@@ -27,9 +28,38 @@ public LogMinerColumnResolverDmlParser(OracleConnectorConfig connectorConfig) {
     @Override
     protected int getColumnIndexByName(String columnName, Table table) {
         if (columnName.matches("COL \\d*")) {
+            final int columnIndex = Integer.parseInt(columnName.substring(4));
             // LogMiner always uses 1-based indices, 0-based indices are required.
-            return Integer.parseInt(columnName.substring(4)) - 1;
+            return (columnIndex + getGeneratedColumnCountSlicedByColumnIndex(table, columnIndex)) - 1;
```
jpechane, line 33: @Naros Is this something that can be cached per-table so it is not recomputed for every single record?

Naros: Reworked code and added the cache.

jpechane:
@Naros Applied, thanks!

PR #6118, opened 2025-01-23
DBZ-8598: Change schema-history producer configurations

rajdangwal:
Addressing https://issues.redhat.com/browse/DBZ-8598


jpechane:
@rajdangwal Could you please add a comment to each of the config option we explicitly set so the explanation is documented for future?

rajdangwal:
@jpechane I removed the `linger.ms` as it was being set to the default value (`0`).

I have added comments for the configs that made sense to me.
For some configs like `enable.auto.commit` I am still not sure why the value is changed from default. So, I left them.

jpechane:
@rajdangwal LGTM, applied!

PR #6121, opened 2025-01-24
Dbz 8549 provide streams deploy steps for jdbc connector

roldanbob:
[DBZ-8549](https://issues.redhat.com/browse/DBZ-8549) Adds downstream deployment instructions to the JDBC connector doc.

This PR supersedes #6107, which I closed. I've addressed any of the questions that were raised against that submission.

Tested in a local Antora and downstream builds. 

In the community version of the documentation, there are no visible changes, but I fixed link errors in the MongoDB and Oracle connector properties tables that resulted from mismatched anchors and links.  These mismatches prevented the downstream build from completing correctly.

This should be ready for reviewing and merging now.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -585,7 +585,8 @@ To deploy a {prodname} JDBC connector, you install the {prodname} JDBC connector
 
 .Prerequisites
 * link:https://zookeeper.apache.org/[Apache ZooKeeper], link:http://kafka.apache.org/[Apache Kafka], and link:{link-kafka-docs}.html#connect[Kafka Connect] are installed.
-* Source and destination databases are installed and are configured to accept JDBC connections.
+* You have a Kafka topic from which the connector can read change event records.
+* A destination database is installed and is configured to accept JDBC connections.
 
 .Procedure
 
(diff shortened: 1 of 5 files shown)


documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -700,6 +1165,7 @@ Information about the properties is organized as follows:
 * xref:jdbc-connector-properties-extendable[JDBC connector extendable properties]
 * xref:jdbc-connector-hibernate-passthrough-properties[JDBC connector `hibernate.*` passthrough properties]
 
+// Type: reference
 [[jdbc-connector-properties-generic]]
 === JDBC connector generic properties
```
Naros, line 1170: I wonder if this section should be renamed to something like "Kafka Connect consumer properties"?

Using the naming "JDBC connector generic properties" seems to indicate to me that these are specific to the JDBC sink when in reality, these aren't used by the JDBC sink at all but rather the Kafka Connect runtime.  In a similar way, the same holds true to the newly added `class` property, although for the downstream it's Strimzi specific, which passes it onto Kafka Connect.

roldanbob: @Naros TBH, I didn't take a further look at content beyond the deployment steps. 
I agree about changing the title of this section. WRT the `class` property, would you say that we should add that -- a least for the downstream docs -- to the properties listed in this section?

Naros (approved):

Naros:
LGTM @roldanbob, feel free to merge as you need.

PR #6124, opened 2025-01-27
[docs] Describes default Vitess `skipped.operations` behavior

roldanbob:
This change follows up on [DBZ-8456](https://issues.redhat.com/browse/DBZ-8456) to update the description of the `skipped.operations` property for the Vitess connector to align with other connectors.

Similar to Db2, MongoDB, and SQL Server, the Vitess connector does not emit truncate events to data change topics. Thus, setting the value of the `skipped-operations` property for the connector to the default `t` (truncate) value, is equivalent to configuring the connector not to skip any events. i.e, setting the value to `none`.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/vitess.adoc
+++ b/documentation/modules/ROOT/pages/connectors/vitess.adoc
@@ -1720,10 +1720,8 @@ You can configure the connector to skip the following types of operations:
 * `d` (delete)
 * `t` (truncate)
 
-Set the value to `none` if you do not want the connector to skip any operations.
-
-The {prodname} Vitess connector emits `truncate` events only to its internal schema history topic and not data change topics.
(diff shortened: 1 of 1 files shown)


roldanbob:
@twthorn FYI.

documentation/modules/ROOT/pages/connectors/vitess.adoc
```diff
@@ -1722,6 +1722,9 @@ You can configure the connector to skip the following types of operations:
 
 Set the value to `none` if you do not want the connector to skip any operations.
 
+The {prodname} Vitess connector emits `truncate` events only to its internal schema history topic and not data change topics.
```
twthorn, line 1725: Vitess connector actually does not have any internal schema history topic. It constructs schema entirely from the vstream events from vitess.

roldanbob: @twthorn I've removed the error. Let me know if this looks ok now. Thanks.

twthorn (approved):
lgtm, thanks for the update!

PR #6125, opened 2025-01-27
DBZ-8566 Adds notes re: deprecation of Containerfile deployment steps

roldanbob:
[DBZ-8566](https://issues.redhat.com/browse/DBZ-8566)  Adds note to the top-level Deployment topic in the product edition of the connector docs to notify users about future removal of the Containerfile version of the deployment instructions.


roldanbob:
@jpechane @Naros Does this look ok? It mirrors the statement that I inserted in the JDBC connector doc.

jpechane:
@roldanbob LGTM, applied. Thanks!

PR #6127, opened 2025-01-29
DBZ-8600 Support SEQUENCE_TABLE function

ani-sha:
https://issues.redhat.com/browse/DBZ-8600


ani-sha:
@Naros Could you review the addition of [SEQUENCE_TABLE](https://docs.percona.com/percona-server/8.0/sequence-table.html) function? I will add it to the upstream grammar after the review. Thanks.

Naros (commented):
Hi @ani-sha, one question. It would seem that in the Percona 8.0 update, there is a small note in the release notes that state:

> To maintain compatibility with existing third-party software, SEQUENCE_TABLE is no longer a reserved term and can be used as a regular identifier.

With that in mind, I think we should also add a test where `SEQUENCE_TABLE` without the function brackets is allowed as a column name, too.  wdyt?

Naros (approved):
Perfect, thanks @ani-sha :tada:

PR #6134, opened 2025-01-30
[docs] Update procedure documentation

HenkvanDyk:
Updating procedure docs to follow deprecation notices of parameters mentioned :

Maria + Mysql:

https://github.com/debezium/debezium/blob/cc04be65aa54dc7b4aa85d34f715d02e62296d0a/documentation/modules/ROOT/partials/modules/all-connectors/shared-mariadb-mysql.adoc?plain=1#L611-L612

https://github.com/debezium/debezium/blob/cc04be65aa54dc7b4aa85d34f715d02e62296d0a/documentation/modules/ROOT/partials/modules/all-connectors/shared-mariadb-mysql.adoc?plain=1#L621-L622

Same for Oracle, just can't find the source file 😅


Naros:
Great catch, thanks @HenkvanDyk, applied.

PR #6139, opened 2025-01-31
DBZ-8653 Emit LOB when available even when `lob.enabled` is off

Naros:
https://issues.redhat.com/browse/DBZ-8653


Naros:
Test failures are unrelated.

jpechane:
@Naros Applied, thanks

PR #6140, opened 2025-01-31
DBZ-8652 Add dedicated increment/decrement property

Naros:
https://issues.redhat.com/browse/DBZ-8652


jpechane:
@Naros Applied, thanks

PR #6148, opened 2025-02-05
DBZ-8665 Improve performance for handling constraint violations

Naros:
https://issues.redhat.com/browse/DBZ-8665

This change restores nearly the identical performance for constraint violations found in Debezium 2.7.4 before the processor and cache logic was refactored and unified.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
@@ -221,34 +221,72 @@ public abstract class AbstractLogMinerEventProcessor<T extends Transaction> impl
      * @param row the event row that contains the row identifier, must not be {@code null}
      */
     protected void removeEventWithRowId(LogMinerEventRow row) {
-        // locate the events based solely on XIDUSN and XIDSLT.
-        String basePrefix = getTransactionIdPrefix(row.getTransactionId());
-        List<String> eventKeysForBasePrefix = getTransactionKeysWithPrefix(basePrefix);
-
-        String transactionIdPrefix = row.getTransactionId() + "-";
-
-        // filter the existing list down to the events for the transaction
-        List<String> eventKeys = eventKeysForBasePrefix.stream()
-                .filter(k -> k.startsWith(transactionIdPrefix))
-                .toList();
-
-        if (eventKeys.isEmpty() && isTransactionIdWithNoSequence(row.getTransactionId())) {
+        final T transaction = getTransactionCache().get(row.getTransactionId());
+        if (transaction != null) {
+            if (removeTransactionEventWithRowId(transaction, row)) {
+                return;
+            }
+            Loggings.logWarningAndTraceRecord(LOGGER, row,
+                    "Cannot undo change on table '{}' since event with row-id {} was not found.",
+                    row.getTableId(), row.getRowId());
+        }
+        else if (isTransactionIdWithNoSequence(row.getTransactionId())) {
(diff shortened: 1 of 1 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
jeremy-l-ford, line 1: Is the method getTransactionKeysWithPrefix no longer used?  May be able to remove.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
```diff
@@ -221,36 +221,74 @@ protected Instant getLastProcessedScnChangeTime() {
...
             LOGGER.debug("Undo change refers to a transaction that has no explicit sequence, '{}'", row.getTransactionId());
-            LOGGER.debug("Checking all transactions with prefix '{}'", basePrefix);
-            eventKeys = eventKeysForBasePrefix;
-        }
+            LOGGER.debug("Checking all transactions with prefix '{}'", prefix);
+
+            final List<T> transactions = getTransactionCache().streamAndReturn(
+                    stream -> stream.filter(entry -> entry.getKey().startsWith(prefix))
+                            .map(LogMinerCache.Entry::getValue)
+                            .toList());
```
jeremy-l-ford, line 244: If you wanted to avoid streaming across the entire transaction cache and creating the list, you could do the following:

`
stream.filter(entry -> entry.getKey().startsWith(prefix))
                            .map(LogMinerCache.Entry::getValue)
                            .anyMatch(t -> removeTransactionEventWithRowId(t, row))

`

if false is reported, you could then log the cannot undo message.
Curious if this would gain any performance in your test.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
Naros: Thanks, I must have missed it when I looked for all the dead code. :+1:

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/AbstractLogMinerEventProcessor.java
Naros: Thanks, that certainly looks cleaner to me.  I will take a peek and see if it changes the performance.

Naros: I don't see a significant change in the performance, but it may have better performance with a cache that contains more than one active transaction like my tests.

Naros:
After testing the other cache implementations
- Ehcache is a 99.89% reduction in time, and is just about 1000x faster.
- Infinispan is a 98% reduction in time, and is about 50x faster.

So the change is clearly a success across all buffers.

Naros:
@jeremy-l-ford, so taking a slight detour to look at the performance of the non-heap implementations:

* With Ehcache, most time is spent processing a DML event, followed by checking whether the event cache has a key or the fetching of a transaction by key.
* With Infinispan, the most significant time spent is with `containsKey` and `size` calls, even on caches with zero entries.

Changing the processed transactions and schema changes cache to heap for Infinispan takes about ~15s off the entire processing with just one processing transaction for my test. I've reached out to the ISPN team to see if there is some additional tuning we should document or add in our tests.

PR #6154, opened 2025-02-10
DBZ-8651 Skip adding Apicurio bits into JDBC sink connector tests

vjuranek:
We don't run any JDBC sink test with Apicurio anyway at the moment, so there's no need to add these libs into JDBC sink pluing.

Related DBZ issues: DBZ-8675 and DBZ-8676.

https://issues.redhat.com/browse/DBZ-8651


AlvarVG (approved):

vjuranek:
Test failures are unrelated.

PR #6160, opened 2025-02-12
DBZ-8684 FIX SQL_ENUM_VALUES query

gaurav7261:
https://issues.redhat.com/browse/DBZ-8684

[FIX] adding order by enumsortorder will guarantee to have same array_agg output in case if a cdc topic is getting written from multiple sources and avoid new schema creation(avro)

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
@@ -66,7 +66,7 @@ public class TypeRegistry {
     private static final String CATEGORY_ARRAY = "A";
     private static final String CATEGORY_ENUM = "E";
 
-    private static final String SQL_ENUM_VALUES = "SELECT t.enumtypid as id, array_agg(t.enumlabel) as values "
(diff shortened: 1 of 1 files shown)


Naros:
Hi @gaurav7261 do you think it would be possible to create a test case that constructs the data types with enums in differing orders and then assert that the output in the event schema is as you expect?

gaurav7261:
I will try @Naros , but for a single postgres, it will always give in a way that is expected because there is only one db, we faced this issue in multi tenant environment where 2 source connectors are writing to same topic, do debezium have multi tenant test cases, if yes let me know, i can refer that

jpechane:
@gaurav7261 The issue is present with multiple databases only? Can't it be replicated with multiple schemas too?

gaurav7261:
Yes, multiple databases only, as a different number of records is needed in the pg_enum table to replicate the issue, so that the query returns the different order of enum in connect.parameters, wdyt? How can i replicate it in the test case @jpechane

Naros:
If it is only multiple databases, should we rethink the solution and modify the code instead of the query where we create the comma-separated list of values to be sorted? That would be easier to unit test while I don't believe that would yield any difference in the schema compatibility concerns as the current solution, right?

Naros:
Hi @gaurav7261, I think a string-based sort should be sufficient.

Naros:

Hi @gaurav7261, please sign off all commits with:

```
git commit -s
```

Naros:

PR #5912, opened 2024-10-03
DBZ-5920 Refactor MongoDB ExtractNewDocumentState SMT

ani-sha:
https://issues.redhat.com/browse/DBZ-5920

- [x] Refactor MongoDataConverter and ExtractNewDocumentState classes
- [x] Adjust all related unit tests and integration tests
- [x] Apply changes to outbox MongoEventRouter class
- [x] Fix outbox test cases
- [x] Incorporate changes from https://github.com/debezium/debezium/pull/6371

PR #6163, opened 2025-02-13
DBZ-8693 Fix omitted fields in nested JSON array

ruthbenjamin:
https://issues.redhat.com/browse/DBZ-8693

### DIFF
--- a/debezium-core/src/main/java/io/debezium/transforms/outbox/JsonSchemaData.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/outbox/JsonSchemaData.java
@@ -23,6 +23,7 @@ import org.apache.kafka.connect.transforms.util.SchemaUtil;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
+
 import io.debezium.schema.FieldNameSelector;
 import io.debezium.schema.FieldNameSelector.FieldNamer;
 import io.debezium.schema.SchemaNameAdjuster;
--- a/debezium-core/src/main/java/io/debezium/transforms/outbox/JsonSchemaData.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/outbox/JsonSchemaData.java
@@ -23,9 +23,6 @@ import org.apache.kafka.connect.transforms.util.SchemaUtil;
 
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.node.ArrayNode;
-import com.fasterxml.jackson.databind.node.JsonNodeType;
-import com.fasterxml.jackson.databind.node.NullNode;
-
 import io.debezium.schema.FieldNameSelector;
 import io.debezium.schema.FieldNameSelector.FieldNamer;
 import io.debezium.schema.SchemaNameAdjuster;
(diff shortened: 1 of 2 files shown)


debezium-core/src/main/java/io/debezium/transforms/outbox/JsonSchemaData.java
```diff
@@ -127,6 +126,11 @@ private Schema mergeSchema(Schema left, Schema right) {
             if (oldField == null) {
                 fields.put(field.name(), field);
             }
+             else if (oldField.schema().type() == Schema.Type.ARRAY && field.schema().type() == Schema.Type.ARRAY) {
+                 Schema mergedValueSchema = mergeSchema(oldField.schema().valueSchema(), field.schema().valueSchema());
+                 SchemaBuilder arrayBuilder = SchemaUtil.copySchemaBasics(field.schema(), SchemaBuilder.array(mergedValueSchema));
+                 fields.put(field.name(), new Field(field.name(), field.index(), arrayBuilder.build()));
+             }
```
Naros, line 133: Hi @ruthbenjamin, what about situations where the JSON objects in the array not only have missing fields but what if the field types differ, which I don't believe this code change handles:

```
{
  "example": [{
    "code": "300",
    "description": "a description"
  }, {
    "code": "200",
    "description": true
  }]
}
```
In this case, AFAICT the schema is generated solely based on the structure of the first element, which means that the "description" field in the second argument isn't a `boolean` as defined by the JSON.

In my humble opinion, I would expect if the above is the input JSON, I would expect an exception to be thrown given that the nested object has the same field with differing value types.

jpechane:
@ani-sha Could you please check if your unwrap SMT covers this scenario?

debezium-core/src/main/java/io/debezium/transforms/outbox/JsonSchemaData.java
ani-sha: @Naros, I have one question as I am working on Mongo Unwrap SMT: if the JSON is as such 
```
{
  "example": [{
    "code": "300",
    "description": "a description"
  }, {
    "code": "200",
    "description": null
  }]
}
```

Then, as per Mongo bson types, one description would be `BsonString`, and the other would be `BsonNull`, both are different types but would have the same schema type of `String`. In that case, it wouldn't throw an error, right? However, in the above case you mentioned, it would throw the error as the schema types are different too.

Naros: That's correct @ani-sha

vincentjames501:
In our case we have a Postgres JSONB column with roughly this shape:

```
{
  "id": "70c8646d-d45d-4972-88f1-545fa6363782",
  "type": "XXX",
  "payload": {
    "items": [
      {
        "id": "a0e88dda-4584-11f0-94fb-150dcd2bc7d8",
        "type": "YYY",
        "payload": {
          "overlap": "e406bfa71fa7",
          "a": -55
        }
      },
      {
        "id": "a0e88ddb-4584-11f0-94fb-534e6e918cf8",
        "type": "ZZZ",
        "payload": {
          "overlap": "e406bfa71fa7",
          "b": "blah"
        }
      }
    ]
  },
  "timestamp": "2025-06-09T22:53:58Z"
}
```

The schema of the items differs depending on its "type" key and when it gets produced to Kafka via Debezium It uses an invalid schema of some kind. For example Debezium produces this to Kafka:

```
{
  "payload": {
    "id": "70c8646d-d45d-4972-88f1-545fa6363782",
    "type": "XXX",
    "payload": {
      "items": [
        {
          "id": "a0e88dda-4584-11f0-94fb-150dcd2bc7d8",
          "type": "YYY",
          "payload": {
            "b": null,
            "overlap": "e406bfa71fa7"
          }
        },
        {
          "id": "a0e88ddb-4584-11f0-94fb-534e6e918cf8",
          "type": "ZZZ",
          "payload": {
            "b": "blah",
            "overlap": "e406bfa71fa7"
          }
        }
      ]
    },
    "timestamp": "2025-06-09T22:53:58Z"
  },
  "type": "AAA"
}
```

Note that it lost "a" and insert "b"=null but kept the overlapping value.

jpechane:
@vincentjames501 What Debezium version did you test it with?

vincentjames501:
@jpechane 3.1.2 . If there is a different version we should try please let me know. This issue definitely caught us off guard and was surprised when it didn't just treat json as an arbitrary blob!

jpechane:
@vincentjames501 Could you please try latest 3.2.0.Beta2? There was a significant rework in form of https://github.com/debezium/debezium/pull/5912

Naros:

Hi @ruthbenjamin, please sign off all commits with:

```
git commit -s
```

Naros:

PR #6169, opened 2025-02-19
Update format-maven-plugin to 2.26.0

Naros:
https://issues.redhat.com/browse/DBZ-8695


Naros:
@jpechane, I tested every plugin version between 2.20.0 and 2.25.0, and all of them worked except for 2.25.0.  I've opened an upstream bug because despite setting profiles specific to Java records, the plugin always changed the formatting for records.  :man_shrugging:

PR #6170, opened 2025-02-19
Mongo read events shoudl produce source info with ts_ms field set to current timestamp

jcechace:
https://issues.redhat.com/browse/DBZ-8708


jpechane:
@jcechace Applied, thanks

PR #6174, opened 2025-02-20
DBZ-8710 Allow specifying case-sensitive database and PDB names

Naros:
https://issues.redhat.com/browse/DBZ-8710

## Changes

The goal of this change is to be backward compatible by applying upper-case semantics to `database.dbname` and `database.pdb.name` when these two configuration properties are not specified in double-quotes. If these are defined in double quotes, the case will be treated as provided for case-sensitive use cases.

Example 1 - Values will be uppercased (case insensitive)
```
"database.dbname": "orclcdb",
"database.pdb.name": "pdb1"
```

Example 2 - Values left as-is (case sensitive)
```
"database.dbname": "\"orclcdb\"",
"database.pdb.name": "\"pdb1\"",
```


Naros:
The Oracle 21 XStream test failure is unrelated, its just the occasional timeout.

PR #6175, opened 2025-02-20
DBZ-8717 Prevent table-level read locks when minimal locking is enabled

kgalieva:
Closes https://issues.redhat.com/browse/DBZ-8717


jpechane:
@kgalieva What will happen in the situation when global lock was not available as it is not provided b cloud provider like - AWS. Will not then trigger this exception and fail the connector?

jpechane:
@kgalieva This seems like a better solution! Yet, maybe instead of a new config option we could introduce a new locking mode with defined semantics?

kgalieva:
Updated, ready for review

PR #6177, opened 2025-02-21
DBZ-8680 Postgres: Probe streaming connection during blocking snapshots

james-johnston-thumbtack:
If there is no activity on the streaming connection while the blocking snapshot is being taken, there is a possibility that the streaming connection might be disconnected.

Periodically run a probing query on the streaming connection during the blocking snapshot.

This is done by adding an overload of waitSnapshotCompletion which allows the user to provide a heartbeat callback function, which is regularly called while waiting.  This will allow other databases to easily add the functionality as well.

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
@@ -407,6 +407,8 @@ public class ChangeEventSourceCoordinator<P extends Partition, O extends OffsetC
 
     public class ChangeEventSourceContextImpl implements ChangeEventSourceContext {
 
+        private static final Duration PAUSE_BETWEEN_HEARTBEAT_CALLBACKS = Duration.ofSeconds(1);
+
         private final Lock lock = new ReentrantLock();
         private final Condition snapshotFinished = lock.newCondition();
         private final Condition streamingPaused = lock.newCondition();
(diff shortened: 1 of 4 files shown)


james-johnston-thumbtack:
NOTE: this commit was developed in a test-driven way:

1.  First the test was written and demonstrated to fail.
2. Then the implementation was written to fix the failing test (and without having to make further changes to the test).

debezium-core/src/main/java/io/debezium/pipeline/source/spi/ChangeEventSource.java
```diff
@@ -19,12 +21,44 @@ interface ChangeEventSourceContext {
...
         void streamingPaused();
 
+        /**
+         * Wait for the streamingPaused function to be called.
+         */
         void waitStreamingPaused() throws InterruptedException;
+
+        @FunctionalInterface
+        interface HeartbeatRunnable {
+            void run() throws SQLException;
```
mfvitale, line 61: Wondering if we can use `RuntimeException` here and the catch the `SQLException` on the implementation and myabe throw a `DebeziumException`. WDYT?

mfvitale:
Hi @james-johnston-thumbtack thanks for this enhancement! Overall it LGTM I just leaved a minor comment.

debezium-core/src/main/java/io/debezium/pipeline/source/spi/ChangeEventSource.java
james-johnston-thumbtack: Yeah, that looks better.  I double-checked and tested, and it looks like your proposal will not hurt the ability of the connector to restart if connection probing fails in the callback: it's still considered as a retriable exception.  See the latest commit version!

debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
```diff
@@ -435,13 +437,20 @@ public void resumeStreaming() {
 
         @Override
         public void waitSnapshotCompletion() throws InterruptedException {
+            waitSnapshotCompletion(null);
```
mfvitale, line 440: Just one last nitpick. Can you just use and empty function `() -> {}`?

jpechane (approved):
@james-johnston-thumbtack LGTM, thanks!

@mfvitale Feel free to merge yourself when you are satisified with the comments.

debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
james-johnston-thumbtack: Done

mfvitale:

PR #6178, opened 2025-02-21
DBZ-8721 Detect Default Value Changes in ExtractChangedRecordState

eisenschmidt:
Ensure that when a field is update from its default value to null or vice versa the change is detected by the SMT and added to the appropriate header.


jpechane:
@eisenschmidt Applied, thanks!

PR #6183, opened 2025-02-24
DBZ-8706 Add pipeline for package and distribute helm charts

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-8706

### DIFF
--- a/jenkins-jobs/pipelines/release/release-charts-pipeline.groovy
+++ b/jenkins-jobs/pipelines/release/release-charts-pipeline.groovy
@@ -20,7 +20,9 @@ GIT_CREDENTIALS_ID = 'debezium-github'
 QUAYIO_CREDENTIALS_ID = 'debezium-charts-quay'
 HOME_DIR = '/home/cloud-user'
 GPG_DIR = 'gpg'
+GITHUB_CLI_VERSION= '2.67.0'
 
+// Helm uses the semantic version format 3.1.0-final instead of the one used by Debezium 3.1.0.Final
 RELEASE_SEM_VERSION=RELEASE_VERSION.replaceAll(/\.(?=[^.]*$)/, '-').toLowerCase()
 
 MAVEN_CENTRAL = 'https://repo1.maven.org/maven2'
@@ -77,8 +79,8 @@ node('release-node') {
         }
 
         stage("Install GitHub CLI") {
-            sh 'curl -fLjsO https://github.com/cli/cli/releases/download/v2.67.0/gh_2.67.0_linux_amd64.tar.gz'
-            sh 'tar -xvzf gh_2.67.0_linux_amd64.tar.gz --one-top-level=gh-cli --strip-components=1'
+            sh "curl -fLjsO https://github.com/cli/cli/releases/download/v${GITHUB_CLI_VERSION}/gh_${GITHUB_CLI_VERSION}_linux_amd64.tar.gz"
+            sh "tar -xvzf gh_${GITHUB_CLI_VERSION}_linux_amd64.tar.gz --one-top-level=gh-cli --strip-components=1"
             sh 'sudo mv gh-cli/bin/gh /usr/local/bin'
             sh 'gh --version'
         }
--- a/jenkins-jobs/pipelines/release/release-charts-pipeline.groovy
+++ b/jenkins-jobs/pipelines/release/release-charts-pipeline.groovy
@@ -1,0 +1,206 @@
+import groovy.json.*
+import java.util.stream.*
+
+import com.cloudbees.groovy.cps.NonCPS
+
+
+if (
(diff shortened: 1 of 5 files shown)


jenkins-jobs/pipelines/release/release-charts-pipeline.groovy
```diff
@@ -0,0 +1,206 @@
...
+                      extensions                       : [[$class: 'RelativeTargetDirectory', relativeTargetDir: DEBEZIUM_OPERATOR_DIR]],
+                      submoduleCfg                     : [],
+                      userRemoteConfigs                : [[url: "https://$DEBEZIUM_OPERATOR_REPOSITORY", credentialsId: GIT_CREDENTIALS_ID]]
+            ]
+            )
+
+            checkout([$class                           : 'GitSCM',
+                      branches                         : [[name: "*/$DEBEZIUM_PLATFORM_BRANCH"]],
+                      doGenerateSubmoduleConfigurations: false,
+                      extensions                       : [[$class: 'RelativeTargetDirectory', relativeTargetDir: DEBEZIUM_PLATFORM_DIR]],
```
AlvarVG, line 82: My 2 cent: I tend to add the `depth: 1` so it makes quite faster all the checkout process. I don't know if you need all the git history. Add this to the extensions array in case you want to include it: 
    `[$class: 'CloneOption', noTags: false, depth: 1]`

AlvarVG (approved):

jenkins-jobs/pipelines/release/release-charts-pipeline.groovy
```diff
@@ -0,0 +1,206 @@
...
+
+RELEASE_SEM_VERSION=RELEASE_VERSION.replaceAll(/\.(?=[^.]*$)/, '-').toLowerCase()
+
+MAVEN_CENTRAL = 'https://repo1.maven.org/maven2'
+
+DEBEZIUM_OPERATOR_DIR='operator'
+DEBEZIUM_PLATFORM_DIR='platform'
+
+
+def modifyFile(filename, modClosure) {
```
jpechane, line 31: @mfvitale I wonder if we can create a Groovy file with library/utuilities shared by jobs and move it there.
Same for version checks etc.

mfvitale: I think we can. It will require some more time since then the library must be configured in Jenkins somewhere (or at least I know where the configuration is on the UI but not on our configuration files).

For `version checks` do you mean 

```groovy
 if (!(RELEASE_VERSION ==~ /\d+\.\d+.\d+\.(Final|(Alpha|Beta|CR)\d+)/)) {
                error "Release version '$RELEASE_VERSION' is not of the required format x.y.z.suffix"
            }
```

mfvitale: @AlvarVG Do you think it is possible to put the `vars` folder inside the `jenkins-job` one and then configure Jenkins  library to get the libs from there?

jpechane:
Converting to draft to avoid unintended merge before 3.0.8 release.

jenkins-jobs/pipelines/release/release-charts-pipeline.groovy
AlvarVG: Sorry for the late response. We should be able to do so, yes. I haven't check how we import it, but i dont think it is going to be a problem.

mfvitale: I was able to test it and works. Working on the changes.

mfvitale:
@jpechane I moved the `modifyFile` function and also the check about the release version format. I think that now we have in place this library there are lot of other stuff that can be refactored and put in the library. I that case a separate task is better.

jpechane (approved):

mfvitale:
@jpechane I think this can be merged now.

jenkins-jobs/pipelines/release/release-charts-pipeline.groovy
```diff
@@ -0,0 +1,191 @@
...
+    !OCI_ARTIFACT_REPO_URL
+) {
+    error 'Input parameters not provided'
+}
+
+GIT_CREDENTIALS_ID = 'debezium-github'
+QUAYIO_CREDENTIALS_ID = 'debezium-charts-quay'
+HOME_DIR = '/home/cloud-user'
+GPG_DIR = 'gpg'
+GITHUT_CLI_VERSION= '2.67.0'
```
jpechane, line 23:
```suggestion
GITHUB_CLI_VERSION= '2.67.0'
```
also the usage of this var

jpechane (approved):
@mfvitale Good job, please fix the type and feel free to merge yourself.

PR #6185, opened 2025-02-24
DBZ-8700 Only reset scale when column type is modified

Naros:
https://issues.redhat.com/browse/DBZ-8700


Naros:
The test failures are unrelated @jpechane.

jpechane:
@Naros Applied, thanks!

PR #6186, opened 2025-02-24
DBZ-8738 Fix stop notification id

mejiadaniel2021:
The correlation id is unset before sending aborted notification for incremental snapshot, resulting in randomly generated id

image

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
@@ -537,6 +537,7 @@ public abstract class AbstractIncrementalSnapshotChangeEventSource<P extends Par
             LOGGER.info("Removed current collection '{}' from incremental snapshot collection list.", stopCurrentTableId);
             tableScanCompleted(partition);
             stopped.add(stopCurrentTableId.identifier());
+            notificationService.incrementalSnapshotNotificationService().notifyAborted(context, partition, offsetContext, stopped);
             // If snapshot has no more collections, abort; otherwise advance to the next collection.
             if (!context.snapshotRunning()) {
                 LOGGER.info("Incremental snapshot has stopped.");
@@ -547,7 +548,9 @@ public abstract class AbstractIncrementalSnapshotChangeEventSource<P extends Par
                 nextDataCollection(partition, offsetContext);
             }
         }
-        notificationService.incrementalSnapshotNotificationService().notifyAborted(context, partition, offsetContext, stopped);
+        else {
+            notificationService.incrementalSnapshotNotificationService().notifyAborted(context, partition, offsetContext, stopped);
+        }
         if (!context.snapshotRunning()) {
             context.unsetCorrelationId();
         }


debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -547,7 +548,9 @@ private void checkAndProcessStopFlag(P partition, OffsetContext offsetContext) {
                 nextDataCollection(partition, offsetContext);
             }
         }
-        notificationService.incrementalSnapshotNotificationService().notifyAborted(context, partition, offsetContext, stopped);
+        else {
+            notificationService.incrementalSnapshotNotificationService().notifyAborted(context, partition, offsetContext, stopped);
+        }
```
rk3rn3r, line 553: Would it make sense to add a `continue;` statement in the right places and keep the previous code?

mfvitale:
@mejiadaniel2021 I have the filing that this is just an effect but not the cause. 

I think that the issue verifies when you stop the incremental snapshot for all the data collections. In this case all collection, expect the current one, will be removed with a call to `context.removeDataCollectionFromSnapshot(dataCollectionId)` and then added to the `stopped` collection. This to keep the current correction be the last to be removed bu effectively there is no call to `context.removeDataCollectionFromSnapshot(dataCollectionId)` it is just added to `stopped` list. This causes the code to gon in the `else` branch where there is the call to `nextDataCollection` but since all the other data collection has been already removed this cause the snapshot to be in a not running state and so calls the `unset`. 

So I think that adding the `context.removeDataCollectionFromSnapshot(dataCollectionId)` just before the `stopped.add(stopCurrentTableId.identifier());` should fix the issue. 

It will be good if you can create a test to reproduce it.

kgalieva:
I think the correlation ID gets unset when the current collection snapshot is stopped, not necessarily when all collections are stopped.

The change looks fine—it separates the logic for stopping the current collection from when other collections remain active.

Regarding this part:

> So I think that adding context.removeDataCollectionFromSnapshot(dataCollectionId) just before stopped.add(stopCurrentTableId.identifier()); should fix the issue.

The removal of the current data collection is already handled in nextDataCollection(). Did you mean skipping `nextDataCollection()` and adding a `resetChunk()` call instead?

Unrelated to this fix: The correlation ID does not map 1:1 with the original signal. It changes each time a connector receives a new execute-snapshot signal. If a new signal arrives while a snapshot is still in progress, subsequent notifications will carry the latest execute-snapshot ID. For example:

```
signal(123, sales)
(123, IN_PROGRESS, sales)
signal(456, products)
(456, IN_PROGRESS, sales)
(456, COMPLETED, sales)
(456, IN_PROGRESS, products)
(456, COMPLETED, products)
```

mfvitale:
> The removal of the current data collection is already handled in nextDataCollection(). Did you mean skipping nextDataCollection() and adding a resetChunk() call instead?

I mean that with the current code we always process the removing of the current data collection as last. But we don't remove it immediately with a call to `context.removeDataCollectionFromSnapshot(dataCollectionId)` but we just relay on the `nextDataCollection()`  that effectively calls the `context.nextDataCollection();` that poll from the Queue and cause `context.snapshotRunning()` return `false`. 

This cause the unset of the `correlationId` in

```java
 private void nextDataCollection(P partition, OffsetContext offsetContext) {
        context.nextDataCollection();
        if (!context.snapshotRunning()) {
            progressListener.snapshotCompleted(partition);
            notificationService.incrementalSnapshotNotificationService().notifyCompleted(context, partition, offsetContext);
            context.unsetCorrelationId();
        }
    }
```

Also the side effect of this is that a snapshot completed notification is emitted and the an abort one. 

So I think the code should be. 

```java
TableId stopCurrentTableId = null;
        for (String dataCollectionId : expandedDataCollectionIds) {
            final TableId collectionId = TableId.parse(dataCollectionId);
            if (currentTable != null && currentTable.id().equals(collectionId)) {
                stopCurrentTableId = currentTable.id();
            }
            else {
                if (context.removeDataCollectionFromSnapshot(dataCollectionId)) {
                    stopped.add(dataCollectionId);
                    LOGGER.info("Removed '{}' from incremental snapshot collection list.", collectionId);
                }
                else {
                    LOGGER.warn("Could not remove '{}', collection is not part of the incremental snapshot.", collectionId);
                }
            }
        }
        // If current is requested to stop, proceed with stopping it.
        if (stopCurrentTableId != null) {
             if (context.removeDataCollectionFromSnapshot(dataCollectionId)) {
                   LOGGER.info("Removed current collection '{}' from incremental snapshot collection list.", stopCurrentTableId);
                  tableScanCompleted(partition);
                  stopped.add(stopCurrentTableId.identifier());
                }
            
            // If snapshot has no more collections, abort; otherwise advance to the next collection.
            if (!context.snapshotRunning()) {
                LOGGER.info("Incremental snapshot has stopped.");
                progressListener.snapshotAborted(partition);
               notificationService.incrementalSnapshotNotificationService().notifyAborted(context, partition, offsetContext, stopped);
            }
            else {
                LOGGER.info("Advancing to next available collection in the incremental snapshot.");
                nextDataCollection(partition, offsetContext);
            }
        }
        notificationService.incrementalSnapshotNotificationService().notifyAborted(context, partition, offsetContext, stopped);
        if (!context.snapshotRunning()) {
            context.unsetCorrelationId();
        }
        LOGGER.info("Removed collections from incremental snapshot: '{}'", stopped); 
```

mfvitale:
> Unrelated to this fix: The correlation ID does not map 1:1 with the original signal.

@kgalieva this is a good point. Maybe it's worth to warn about this in the doc since having a correct mapping between signal and it's id required a completely different implementation. 

@jpechane WDYT?

jpechane:
@mfvitale Yes I agree that this is something that should be enioned in documentation.

Naros:

Hi @mejiadaniel2021, please sign off all commits with:

```
git commit -s
```

Naros:

PR #6188, opened 2025-02-25
DBZ-8287 Treat tombstones as delete events

Naros:
https://issues.redhat.com/browse/DBZ-8287

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
@@ -88,12 +88,6 @@ public class JdbcChangeEventSink implements ChangeEventSink {
                 continue;
             }
 
-            if (record.isTombstone()) {
-                // Skip only Debezium Envelope tombstone not the one produced by ExtractNewRecordState SMT
-                LOGGER.debug("Skipping tombstone record {}", record);
-                continue;
-            }
-
             final CollectionId collectionId = optionalCollectionId.get();
 
             if (record.isTruncate()) {
@@ -116,7 +110,7 @@ public class JdbcChangeEventSink implements ChangeEventSink {
                 }
             }
 
-            if (record.isDelete()) {
+            if (record.isDelete() || record.isTombstone()) {
                 if (!config.isDeleteEnabled()) {
                     LOGGER.debug("Deletes are not enabled, skipping delete for topic '{}'", record.topicName());
                     continue;
--- a/debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/AbstractJdbcSinkDeleteEnabledTest.java
+++ b/debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/AbstractJdbcSinkDeleteEnabledTest.java
@@ -162,9 +162,8 @@ public abstract class AbstractJdbcSinkDeleteEnabledTest extends AbstractJdbcSink
 
     @ParameterizedTest
     @ArgumentsSource(SinkRecordFactoryArgumentsProvider.class)
-    @FixFor("DBZ-6862")
-    public void testShouldSkipTombstoneRecord(SinkRecordFactory factory) {
-
+    @FixFor("DBZ-8287")
+    public void testTombstoneShouldDeleteRow(SinkRecordFactory factory) {
         final Map<String, String> properties = getDefaultSinkConfig();
         properties.put(JdbcSinkConnectorConfig.SCHEMA_EVOLUTION, SchemaEvolutionMode.BASIC.getValue());
         properties.put(JdbcSinkConnectorConfig.PRIMARY_KEY_MODE, PrimaryKeyMode.RECORD_KEY.getValue());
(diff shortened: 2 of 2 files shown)


jpechane:
@Naros Is there any relation to https://github.com/debezium/debezium/pull/5997 ?

jpechane:
@Naros One more question, what is expected behaviour if both delete and tombstone arrives? Are two `DELETE` statements executed?

Naros:
> @Naros Is there any relation to #5997 ?

It is, I just took a slightly simpler approach from what I can tell.

> @Naros One more question, what is expected behaviour if both delete and tombstone arrives? Are two `DELETE` statements executed?

The answer is yes and no :grin: 

The first question is whether does adding the new delete operation cause the buffer to exceed the batch size. If it does, the answer is always yes. If the operation does not overflow the buffer beyond the batch size, then it depends on whether the reduction buffer is enabled. If enabled, the answer is no, but if not enabled, then yes.

mfvitale:
> If they use the ExtractNewRecordState and pass tombstones along, they could enable this since there would never be a real "delete" otherwise. wdyt?

Why an user want to just maintain the tombstone and not the delete?

IMHO, I still think that JDBC sink should only manage real delete event and not tombstone that is a "special" message.

jpechane:
@mfvitale But int hat case there will be no deletes when `ExtractNewRecordState` is applied.

mfvitale:
> @mfvitale But int hat case there will be no deletes when `ExtractNewRecordState` is applied.

Got it. Maybe we can just simplify and, if you want to use the `ExtractNewRecordState` with the `JDBCSinkConnector` the only [delete.handling​.mode](https://debezium.io/documentation/reference/3.0/transformations/event-flattening.html#extract-new-record-state-delete-handling-mode) supported is `rewrite` and so we just check for the `__deleted` instead of the normal `op=d`.

Naros:
The problem with that @mfvitale is you're stuck on the idea that the source if Debezium, when it may not be.  In cases where the source could be anything else, the only real expectation for deletes IMO would be tombstones.

mfvitale:
> The problem with that @mfvitale is you're stuck on the idea that the source is Debezium, when it may not be. In cases where the source could be anything else, the only real expectation for deletes IMO would be tombstones.

Ok so no way for simplification. :)

Naros:
Hah, not necessarily, I just think we have to have a solution for that corner case of non-Debezium sources so that users can safely delete events that have some `record_key`. 

Though in hindsight, this reminds me that I didn't add a test where the `primary.key.mode` was `record_value` or `connect`, and at least in `record_value` scenarios, we should _skip_ the tombstone :thinking:

mfvitale:
> Hah, not necessarily, I just think we have to have a solution for that corner case of non-Debezium sources so that users can safely delete rows that have some `record_key`.
> 
> Though in hindsight, this reminds me that I didn't add a test where the `primary.key.mode` was `record_value` or `connect`, and at least in `record_value` scenarios, we should _skip_ the tombstone 🤔
> 
> I feel there is no "silver bullet" solution for some of these use cases :(

Yes, what I'm trying to say is that we are trying to solve all possible scenario resulting a very complicated code when maybe the real scenario are too few and should focus on that one.

IIRC the Debezium Sink was mainly created to support Debezium source users, so we should be more focus on that.

Naros:
> Though in hindsight, this reminds me that I didn't add a test where the primary.key.mode was record_value or connect, and at least in record_value scenarios, we should skip the tombstone 🤔

Actually we already address this when `delete.enabled` is `true`, you can only use `record_key`; so false alarm :tada:

jpechane:
OK, another proposal. Tombstone handling will be disabled by default. Could be enabled with a config option. When tombstone arrives and handling is disable then a warning is emitted - with its own logging category so it can be disabled?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
```diff
@@ -116,7 +110,7 @@ public void execute(Collection<SinkRecord> records) {
                 }
             }
 
-            if (record.isDelete()) {
+            if (record.isDelete() || record.isTombstone()) {
```
rk3rn3r, line 113: Don't you think `isDelete()` should return true when either the message is a Debezium envelope message with `op=d` or when it is a flattend / unwrapped message tombstone? Instead of adding this condition here do it inside `isDelete()`?

mfvitale:
> OK, another proposal. Tombstone handling will be disabled by default. Could be enabled with a config option. When tombstone arrives and handling is disable then a warning is emitted - with its own logging category so it can be disabled?

I fear that this will complicate instead of simplify. If we want to cover all the possible scenario there is nothing we can do I think.

rk3rn3r:
You should never disable Tombstones in Kafka because the log cleaner (log compaction) requires tombstones. That should be active by default and only really explicitly and with warnings be disabled!

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
Naros: Perhaps, but my only concern is what happens if we need a situation where we do something specific where we have `op=d` versus a Tombstone; then we've lost that differentiation.

Perhaps the best route in that case would be `isDeleteOrTruncate()` if we really want to hide that detail but be clear we're operating on both.

Naros:
> First, mind that a DELETE event (no matter what the message format is, aka Debezium envelope with `op=d`, Kafka tombstone or some other special tombstone format) only makes sense for keyed events. Those will always use a ReduceBuffer/DeduplicationBatch buffer implementation, right? 

Not based on the current implementation, the user has to opt into the reduction buffer, see `use.reduction.buffer`.

> Let me then move this towards the future implementation: No matter if the source of a DELETE event is a tombstone, a Debezium envelope `op=d` event or a some custom tombstone thingy, it will end up in a buffer/batch that is de-duplicated per key. So in that case you won't have duplicates, simply by the deduplication in the buffer. So only temporary we would maybe accept multi-deletes for a single event.

Are you suggesting that users must use deduplication without an opt-out strategy? 
If so, we should talk about this because I disagree. 

Before I joined RH, I worked on data replication across Oracle, SAP, and other ERP-like systems. There were specific laws and industry requirements about how the data had to be mirrored from one system to another for auditing, tracking, and reconciliation purposes. The common theme for those was that no deduplication was allowed.

So I strongly believe, while its okay to have the reduction buffer and perform deduplication, it should be something that the user opts into or out of, but they should have the option or else you effectively cut-off certain industries or sectors from being able to use the sink connector all together.

jpechane:
@Naros Applied. Let's merge it now so we can get it into 3.0.8. As was told the current downside is potential for two `DELETE`s to be exected sometimes. Let's log a follow-up Jira where we would define a final consensus.

jpechane:
@Naros This does not apply cleanly on 3.0. Could you please backport it? The Jira should be resloved after that. Thanks!

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
rk3rn3r: @Naros  but what is wrong with the above code? It's the current implementation of `isDelete()`. So there must be something wrong with that check?!

Naros:
> @Naros This does not apply cleanly on 3.0. Could you please backport it? The Jira should be resloved after that. Thanks!

Np, backported to 3.0 & closed Jira issue as requested :+1:

PR #6190, opened 2025-02-25
DBZ-8436 Add MariaDB `max_statement_time` user resource option

Naros:
https://issues.redhat.com/browse/DBZ-8436


jpechane:
@Naros Applied, thanks

PR #6191, opened 2025-02-25
DBZ-8159 Add section about Oracle mTLS connectivity

Naros:
https://issues.redhat.com/browse/DBZ-8159

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/oracle.adoc
+++ b/documentation/modules/ROOT/pages/connectors/oracle.adoc
@@ -3268,6 +3268,40 @@ For non-CDB installation, do *not* specify the `database.pdb.name` property.
 ----
 
 endif::community[]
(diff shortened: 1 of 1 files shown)


jpechane:
@Naros LGTM, thanks! If @roldanbob has any comments then they can be addressed in a separate PR.

roldanbob:
@Naros @jpechane Are we testing mTLS connectivity? Do we want this section to render in the downstream docs?

Naros:
> @Naros @jpechane Are we testing mTLS connectivity? Do we want this section to render in the downstream docs?

We don't specifically test this because we don't have Oracle Wallet configured in our installations, but we do have users who successfully use it in the community. It's worth including so it's documented for our downstream users, but I'll defer to @jpechane should he disagree.

PR #6194, opened 2025-02-26
DBZ-8595 Remove unnecesarry refreshes of schema

vjuranek:
Both `connectionCreated()` and `lockTablesForSchemaSnapshot()` are called only from `RelationalSnapshotChangeEventSource#doExecute()`. Before it we just determine snapshot tables, offset and do the locking of the captured tables. We don't need schema for these operations and it should be sufficient to refresh the schema only before taking the actual snapshot. At this point fo time tables are already lock, so we take schema refresh from consitent DB view.

https://issues.redhat.com/browse/DBZ-8595

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSnapshotChangeEventSource.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSnapshotChangeEventSource.java
@@ -106,7 +106,6 @@ public class PostgresSnapshotChangeEventSource extends RelationalSnapshotChangeE
         if (snapshotterService.getSnapshotter().shouldStreamEventsStartingFromSnapshot() && startingSlotInfo == null) {
             setSnapshotTransactionIsolationLevel(snapshotContext.onDemand);
         }
-        schema.refresh(jdbcConnection, false);
     }
 
     @Override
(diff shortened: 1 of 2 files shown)


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSnapshotChangeEventSource.java
```diff
@@ -106,7 +106,6 @@ protected void connectionCreated(RelationalSnapshotContext<PostgresPartition, Po
         if (snapshotterService.getSnapshotter().shouldStreamEventsStartingFromSnapshot() && startingSlotInfo == null) {
             setSnapshotTransactionIsolationLevel(snapshotContext.onDemand);
         }
-        schema.refresh(jdbcConnection, false);
```
animesh-hevo, line 109: @vjuranek In `PostgresStreamingChangeEventSource, `initSchema` is also called twice: once in the `init` method and once in the `execute` method. Can we remove the duplicate call there as well?

animeshsharma97, line 109: In `PostgresStreamingChangeEventSource` as well, `initSchema` is being called twice: once in `init` method and once in `execute` method. Are those duplicate calls as well?

vjuranek: Good point, I think the second call is redundant. I'm running the tests locally now, if they pass, I'll update the PR.

Naros (approved):
LGTM @vjuranek, applied!

PR #6203, opened 2025-02-28
DBZ-8749 Account for database system tz in snapshot source ts_ms

Naros:
https://issues.redhat.com/browse/DBZ-8749

Verified locally setting Oracle container TZ=CET and leaving the TZ unset (UTC), and in both scenarios, a quick test of the envelope `ts_ms` and the `source.ts_ms` were within milliseconds of one another, as expected.


Naros:
The Oracle XStream 19c failure about ORA-01466 is unrelated, bug logged a follow-up [DBZ-8750](https://issues.redhat.com/browse/DBZ-8750).

vjuranek:
Failure on XStream 21 is not related as well.

vjuranek:
Merged, thanks!

PR #6204, opened 2025-02-28
DBZ-8740 Add start/commit scn and time values to Oracle source info

Naros:
https://issues.redhat.com/browse/DBZ-8740


jpechane:
@Naros The test failures are unrelated. Applied, thanks.

PR #6205, opened 2025-02-28
DBZ-8744 Limit archive logs to current log incarnation only

Naros:
https://issues.redhat.com/browse/DBZ-8744

### Premise

In Oracle, when you refresh an instance from another, you do so by copying the other instance's database data files into the new instance along with all the recent archive logs, adding them to the catalog during the recovery process. This makes sure that when you refresh from the backup of the other instance, you have all changes. 

Once you've finished the refresh/recovery, the DBA most often elects to create a new log incarnation. You can think of this as closing off the log stream you restored from and starting a brand new log stream for the newly refreshed instance. When you do this, log sequences are reset.

If the DBA accidentally leaves the old archive logs from the prior incarnation in the catalog, this confuses Debezium and it will continuously report a log inconsistency. 

This fix addresses that by making sure when a new log incarnation is created, the archive logs that are fetched by the connector are only those that match the current incarnation, which avoids the need for the DBA to immediately remove the old archive logs from the catalog.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
@@ -5,7 +5,6 @@
  */
 package io.debezium.connector.oracle;
 
-import java.math.BigInteger;
 import java.sql.CallableStatement;
 import java.sql.Clob;
 import java.sql.DatabaseMetaData;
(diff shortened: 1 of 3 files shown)


jpechane (approved):
@Naros LGTM, thanks! Feel free to merge yourself. Just one final verification, given that the `SELECT` s now contain additional `JOIN` is not the perfromance impacted?

Naros:
Hi @jpechane, that's a great question, and the simple answer is no.

The `DATABASE_VIEW` refers to the database's dynamic performance view, which contains control file information about the entire database. This table has only one row.  In addition, these queries are executed ideally once per mining iteration unless the log state is inconsistent and triggers a consistency retry check.

jpechane:
@Naros Applied, thanks

jpechane:
@Naros It doies not apply cleanly to 3.1. Could you please backport it? Also do you also plan to backport it to 3.0?

jpechane:
@Naros Right now I don't forsee that. It is more a contingency - if the build would be done (vulerability etc.), is it important to have it in? If not, just skip it.

PR #6211, opened 2025-03-03
DBZ-8669 Add docs for CloudEvents OpenTelemetry integration

rkudryashov:


jpechane:
@rkudryashov Applied, thanks

PR #6212, opened 2025-03-04
DBZ-7810 Support empty passwords in jdbc connections

minleejae:
https://issues.redhat.com/browse/DBZ-7810


jpechane:
@minleejae Applied, thanks!

PR #6213, opened 2025-03-04
DBZ-8711 Reset retry count only if we successfully poll messages

ahmedbayraktar:


debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -373,11 +374,25 @@ protected void resetErrorHandlerRetriesIfNeeded(List<SourceRecord> records) {
         // error handler is passed into the new error handler, propagating the retry count. This method
         // allows resetting that counter when a successful poll iteration step contains new records so that when a
         // future failure is thrown, the maximum retry count can be utilized.
-        if (!records.isEmpty() && coordinator != null && coordinator.getErrorHandler().getRetries() > 0) {
+        if (containsMessages(records) && coordinator != null && coordinator.getErrorHandler().getRetries() > 0) {
             coordinator.getErrorHandler().resetRetries();
         }
     }
 
+    private boolean containsMessages(List<SourceRecord> records) {
```
jpechane, line 382: I beleive the name should be a bit more descriptive, like `containsChangeDataMessages`, otherwise it looks it it just checks empty/full state.
Also comment above should be updated to reflect new behaviour.

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -373,11 +374,25 @@ protected void resetErrorHandlerRetriesIfNeeded(List<SourceRecord> records) {
...
+        if (records == null || records.isEmpty()) {
+            return false;
+        }
+
+        for (SourceRecord record : records) {
+            if (record.valueSchema() != null && record.valueSchema().field(Envelope.FieldName.AFTER) != null) {
```
jpechane, line 388: I'd recommend to use `io.debezium.data.Envelope.isEnvelopeSchema` to check if it is change message or not.

jpechane (changes requested):
@ahmedbayraktar Thanks for the PR! I left two comments. Also would it be possible to add a test to `io.debezium.connector.common.BaseSourceTaskTest` to verify the new behaviour?

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
ahmedbayraktar: Changed method name as suggested.

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
ahmedbayraktar: Good point, using `isEnvelopeSchema` now.

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
ahmedbayraktar: Added unit test to verify this method.

jpechane:
@ahmedbayraktar Applied, thanks!

PR #6216, opened 2025-03-08
DBZ-8570 Reduce frequency table exists and column metadata queries

minleejae:
https://issues.redhat.com/browse/DBZ-8570

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/Buffer.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/Buffer.java
@@ -7,6 +7,8 @@ package io.debezium.connector.jdbc;
 
 import java.util.List;
 
+import io.debezium.connector.jdbc.relational.TableDescriptor;
+
 /**
  * An interface for implementing several kind of buffers
  *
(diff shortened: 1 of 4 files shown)


Naros (changes requested):
Overall I don't think this necessarily solves the broader problem, which also includes the call to `hasTable`, which fires for each event in a batch even with your changes.

What I had in mind was that we would look at somehow associating the `TableDescriptor` to the `RecordBuffer` objects. In this case, we would only need to associate the descriptor to the buffer once, when the buffer is created.

From there, we would simply compare the event schema to the buffer schema. If they differ, we know to flush the buffers & recreate, where we'd refetch the `TableDescriptor` again. 

This way if we received a batch of N inserts/updates for the same table with no changes in the table's structure, those would all be written in a transaction batch and we'd only fetch the table information once for that batch, rather than N times.

Does that make sense?

minleejae:
Hi, @Naros 
I agree with your feedback. I had overlooked the fact that hasTable is also a metadata query. As you suggested, I have modified the implementation to ensure that the table information is only re-fetched when a schema change is detected in the event. Could you review the changes? Thanks!

Naros (approved):
Great work @minleejae LGTM :tada: 

Will merge when the code freeze is lifted.

PR #6218, opened 2025-03-10
DBZ-8682 remove explicit tag for debezium platform in release pipeline

mfvitale:
Follow-up changes after Beta1 release.


jpechane:
@mfvitale Applied, thanks

PR #6220, opened 2025-03-10
Dbz 8687 update pg doc about replication slot failover

roldanbob:

Updates the PostgreSQL _Supported topologies_ topic to remove community-only conditional to expose information about using the PG connector in cluster environments that run PG 15 vs. 16 vs. 17. Adds a product-conditionalized note that specifies that using the connector with PG17 is a TP feature.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/postgresql.adoc
+++ b/documentation/modules/ROOT/pages/connectors/postgresql.adoc
@@ -2147,7 +2147,7 @@ The default `false` setting causes the custom converter to always return `null` 
 // ModuleID: setting-up-postgresql-to-run-a-debezium-connector
 // Title: Setting up PostgreSQL to run a {prodname} connector
 [[setting-up-postgresql]]
-== Setting up Postgres
+== Setting up PostgreSQL
 
 ifdef::community[]
 Before using the PostgreSQL connector to monitor the changes committed on a PostgreSQL server, decide which logical decoding plug-in you intend to use.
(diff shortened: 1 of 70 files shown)


documentation/modules/ROOT/pages/connectors/postgresql.adoc
```diff
@@ -2516,25 +2516,47 @@ host    replication     <youruser>  ::1/128                 trust   // <3>
...
+You cannot configure logical replication on replica servers in the cluster.
+Consequently, the {prodname} PostgreSQL connector can connect and communicate only with the primary server.
+If the primary server fails, the connector stops.
 
-When using PostgreSQL 17 or newer, you can set up logical replication slots on primary serves which are enabled for failover.
-The state of such a failover slot can automatically be propagated to one or more replica servers,
-allowing {prodname} to continue to ingest changes from a replica promoted to primary after a failure,
-without missing any events.
+Following a failure, after the cluster is repaired, you can once again promote the original primary server to `primary`, and then restart the connector.
+Alternately, you might promote a different PostgreSQL server to `primary`.
```
Naros, line 2532: I wonder if it would be helpful to consider including a procedure like we do for **Upgrading PostgreSQL** also to outline the expected steps when you fail over from a primary to a replica for PostgreSQL 15 or earlier. It's often one of the things we need to explain.

documentation/modules/ROOT/pages/connectors/postgresql.adoc
```diff
@@ -2516,25 +2516,47 @@ host    replication     <youruser>  ::1/128                 trust   // <3>
...
+Before you can promote a different server to primary, you must xref:setting-up-postgresql[properly configure] the server.
+Then, to enable the connector to consume from the new `primary` server, update properties in the connector configuration to match the details of the new server, for example, xref:postgresql-property-database-hostname[`database.hostname`], xref:postgresql-property-database-port[`database.port`], xref:postgresql-property-database-dbname[`database.dbname`], and xref:postgresql-property-slot-name[`slot.name`].
+After you update the configuration to point to the new primary, you can restart the connector.
 
-endif::community[]
+==== PostgreSQL 16 or later clusters
+
+When you deploy {prodname} with PostgreSQL 16 or later, you can set up logical replication slots on replica servers.
+In a PostgreSQL 16 cluster, {prodname} can capture change events from servers other than the primary server.
+However, be aware that {prodname} connections to replica servers generally experience higher latency than connections to a primary server.
```
Naros, line 2541: Perhaps we could pull up the tidbit from **Cluster failures** that emphasizes on PG16, you need to keep the replication slots on the replicas synchronized to support flawless failover post reconfiguration with Debeizum.

Naros (commented):
A few inline comments if you could take a look @roldanbob.
Let me know what you think.

documentation/modules/ROOT/pages/connectors/postgresql.adoc
roldanbob: Yes, again, I agree. Thanks. 
 ... Slowly putting the mind back into gear ...

roldanbob:
Closed in favor of #6225

PR #6221, opened 2025-03-11
DBZ-8771 fix docs and comment of minimum Java version requirement from 11 to 21

katsulau:
As mentioned in the title, I have updated the documentation to state that JDK 21 is required

https://issues.redhat.com/browse/DBZ-8771

### DIFF
--- a/CONTRIBUTE.md
+++ b/CONTRIBUTE.md
@@ -19,7 +19,7 @@ You can reach us here:
 The following software is required to work with the Debezium codebase and build it locally:
 
 * [Git 2.2.1](https://git-scm.com) or later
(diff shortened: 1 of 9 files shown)


Naros:
@roldanbob I don't know if it will affect downstream, but FYI on the new variable `min-java-connectors-version` in the antora.yml for the `install.adoc`. I believe that is entirely upstream, but in case not.

Naros (approved):

roldanbob:
> @roldanbob I don't know if it will affect downstream, but FYI on the new variable `min-java-connectors-version` in the antora.yml for the `install.adoc`. I believe that is entirely upstream, but in case not.

@Naros Thanks for the heads up, but to the best of my knowledge, the downstream docs don't pull in this content, so the change should not have any affect.

PR #6222, opened 2025-03-11
DBZ-8525: Centralise sensitive data logging using the Loggings Class

b-goyal:


Naros:
Hi @b-goyal I suspect many of these failures are related to tests that rely on `LogInterceptor` and the changing to the `Loggings` class changes which class now emits the log messages, causing failures.  Can you take a look?

Naros:
Hi @b-goyal, can you modify the second commit on this PR so that it is prefixed with DBZ-8525?

b-goyal:
@Naros , I have kept the last commit message as '[DBZ-8525: Centralise sensitive data logging using the Loggings Class](https://github.com/debezium/debezium/pull/6222/commits/190d4ba484110dec3fba2700847f437e546873d7)' - does this work?

b-goyal:
@Naros  I have updated the second commit message.

debezium-core/src/main/java/io/debezium/util/Loggings.java
```diff
@@ -72,4 +72,34 @@ public static void logErrorAndTraceRecord(Logger logger, Object record, String m
...
+     *
+     * @param logger the logger instance
+     * @param record the record the log entry is based upon
+     * @param message the trace message to be logged
+     * @param arguments the arguments passed to the trace message
+     */
+    public static void logTraceAndTraceRecord(Logger logger, Object record, String message,
+                                              Object... arguments) {
+        logger.trace(message, arguments);
+        LOGGER.trace("Source of trace is record '{}'", record);
```
Naros, line 103: I really don't think for use cases where we have an existing TRACE log operation that we should convert a single log entry into two log entries. That to me creates more noise than necessary in the logs.

Naros (commented):
Overall LGTM, I just have one nit-pick, see below.

jpechane:
@b-goyal Applied, thanks!

debezium-core/src/main/java/io/debezium/util/Loggings.java
b-goyal: Hey @Naros
This has been done to ensure we only log sensitive content using Loggings class. The problem as described in the [JIRA](https://issues.redhat.com/browse/DBZ-8525) is that we have a risk of exposing sensitive data when TRACE logging is enabled as multiple classes currently log sensitive content. With this change, we aim to limit the exposure of sensitive data from Loggings class only. If anyone wants to redact sensitive content, they would only have to care about this single class and not scan the entire core for sensitive logs.

rk3rn3r: @jpechane @Naros Some time back if I recall correctly that enabling TRACE log level was meant to expose sensitive data on purpose. How do I get this back? I didn't see any config added to enable it?

rk3rn3r:
@jpechane One general concern, this introduces a static dependency to many classes and it could make testing harder, especially looking at such an important non-functional part as logging. How do we approach this?

jpechane:
@rk3rn3r Hi, I understand your concern. From the purity PoV the best solution would be to create a descendant interface derived from `org.slf4j.Logger` and add new methods to it. Then new factory should be created that would create implementations that would underneath delegate to another logger.
I am fine with doing it as refactoring but current approach is from my PoV a pragmatic approach.
Can you please elaborate on the harder testing part?

mfvitale:
Maybe a better solution could be something like [this](https://www.baeldung.com/logback-mask-sensitive-data) or [this](https://github.com/augustd/owasp-security-logging/wiki)

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -302,8 +303,8 @@ protected void readChunk(P partition, OffsetContext offsetContext) throws Interr
                         continue;
                     }
                     if (LOGGER.isInfoEnabled()) {
-                        LOGGER.info("Incremental snapshot for table '{}' will end at position {}", currentTableId,
-                                context.maximumKey().orElse(new Object[0]));
+                        Loggings.logInfoAndTraceRecord(LOGGER, context.maximumKey().orElse(new Object[0]),
+                                "Incremental snapshot for table '{}' will end at position", currentTableId);
```
kgalieva, line 307: It looks like the end position is not getting logged after this change. 
Here is the log from one of the tests where I noticed it:
`2025-03-21 09:24:15,874 INFO   Postgres|test_server|0|streaming|postgres  Incremental snapshot for table 's1.part2' will end at position   [io.debezium.pipeline.source.snapshot.incremental.AbstractIncrementalSnapshotChangeEventSource]
`

debezium-core/src/main/java/io/debezium/util/Loggings.java
Naros: Hi @b-goyal I see your point. I've always approached this a bit differently, where I see it more broadly that you use the non-TRACE levels to log non-sensitive information and leave TRACE specifically for that purpose. This is why I questions the point behind two trace log entries from different loggers as being a bit over-engineered in PoV.  But if you're looking for a way to tune the exclusion of sensitive data by log class, does make sense.

@rk3rn3r you still have it that way. What @b-goyal has done is allow the environment to be configured with TRACE at the root or within the `io.debezium` package but they can then override `io.debezium.util.Loggings` to use INFO or DEBUG to prevent the output of sensitive data, while retaining everything else.

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
jpechane: @b-goyal I believe the wording of this message needs to be updated or maybe it will need to be altogether moved to a diffeent log level.
@kgalieva You are right, OTH (and I admit I never thougth about that) this really can be potential leak of data into logs.

b-goyal: Thanks @jpechane and @kgalieva , I will raise a followup PR with the fix.

b-goyal: A PR to improve this log line is raised here : https://github.com/debezium/debezium/pull/6247. Thanks!

b-goyal:
@Naros @jpechane - is it possible to backport this change to 3.0 ?

jpechane:
@b-goyal I am fine doing it if you prepare a PR for 3.0 branch

b-goyal:
> @b-goyal I am fine doing it if you prepare a PR for 3.0 branch

Thanks @jpechane , I will raise a PR for this.

twthorn:
This caused a severe performance degradation, put out a fix here https://github.com/debezium/debezium/pull/6298

PR #6223, opened 2025-03-12
DBZ-8776: Return delete.tombstone.handling.mode from ExtractNewRecordState config method

rajdangwal:
Addressing https://issues.redhat.com/browse/DBZ-8776

### DIFF
--- a/debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordStateConfigDefinition.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordStateConfigDefinition.java
@@ -121,7 +121,7 @@ public class ExtractNewRecordStateConfigDefinition {
     }
 
     public static final Field HANDLE_TOMBSTONE_DELETES = Field.create("delete.tombstone.handling.mode")
-            .withDisplayName("Handle delete records")
+            .withDisplayName("Handle delete and tombstone records")
             .withEnum(DeleteTombstoneHandling.class)
             .withWidth(ConfigDef.Width.MEDIUM)
             .withImportance(ConfigDef.Importance.MEDIUM)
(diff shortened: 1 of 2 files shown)


debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordStateConfigDefinition.java
```diff
@@ -121,7 +121,7 @@ public static DeleteTombstoneHandling parse(String value, String defaultValue) {
     }
 
     public static final Field HANDLE_TOMBSTONE_DELETES = Field.create("delete.tombstone.handling.mode")
-            .withDisplayName("Handle delete records")
+            .withDisplayName("Handle delete and tombstone records")
```
rk3rn3r, line 124: One could argue that a tombstone in Kafka is in fact a delete message. Thus `tombstone == delete`. And as we were using this in the UI as label for the field maybe we can leave the shorter name as it was before?

rajdangwal: Thanks for the review @rk3rn3r 
Although "tombstone" and "delete" are often used interchangeably, the connector can generate two distinct records: a tombstone record and a non-tombstone delete record with `__deleted: true`.

Since the new config `delete.tombstone.handling.mode` handles both types of records, I believe using a more descriptive name is appropriate.

In the UI, both `delete.tombstone.handling.mode` and `delete.handling.mode` currently have the same display name, which could cause confusion. Renaming the new config to something more specific would help clarify the difference. WDYT?

jpechane: I agree wit @rajdangwal on thise one. While tombstone is delete record in the plain Kafka context, in case of Debezium context these are two distinct records and it makes sense to differentiate between them. In fact tomstone is only technical one to enable compaction in Kafka, otherwise it would be completely unnecessary for Debezium.

jpechane:
@rajdangwal Applied, thanks!

debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordStateConfigDefinition.java
rajdangwal: @rk3rn3r are we referring to change the config name from `delete.tombstone.handling.mode` to something else?

jpechane:
@rajdangwal Hi, I backported the PR into 3.0 branch but there is not date sate for 3.0.9.Final

PR #6225, opened 2025-03-12
Dbz 8687 enhance pg docs w info about replication slot failover

roldanbob:

Updates the PostgreSQL _Supported topologies_ topic to remove community-only conditional to expose information about using the PG connector in cluster environments that run PG 15 vs. 16 vs. 17. Adds a product-conditionalized note that specifies that using the connector with PG17 is a TP feature.   

Supersedes #6220


jpechane:
@roldanbob Applied, thanks!

PR #6228, opened 2025-03-14
DBZ-8769 Change schema history table's history_data column type to TEXT

katvito:
When using debezium-storage-jdbc with MySQL, the creation of the default schema history table fails.
This PR changes the column type from VARCHAR to TEXT to resolve the issue.

Related Issue: https://issues.redhat.com/browse/DBZ-8769

### DIFF
--- a/debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/history/JdbcSchemaHistoryConfig.java
+++ b/debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/history/JdbcSchemaHistoryConfig.java
@@ -37,7 +37,7 @@ public class JdbcSchemaHistoryConfig extends JdbcCommonConfig {
     private static final String DEFAULT_TABLE_DDL = "CREATE TABLE %s" +
             "(" +
             "id VARCHAR(36) NOT NULL," +
-            "history_data VARCHAR(65000)," +
+            "history_data TEXT," +
             "history_data_seq INTEGER," +
             "record_insert_ts TIMESTAMP NOT NULL," +
             "record_insert_seq INTEGER NOT NULL" +


debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/history/JdbcSchemaHistoryConfig.java
```diff
@@ -37,7 +37,7 @@ public class JdbcSchemaHistoryConfig extends JdbcCommonConfig {
     private static final String DEFAULT_TABLE_DDL = "CREATE TABLE %s" +
             "(" +
             "id VARCHAR(36) NOT NULL," +
-            "history_data VARCHAR(65000)," +
+            "history_data TEXT," +
```
Naros, line 40: My only concern is whether there is even the potential corner case where a database supports VARCHAR but does not support TEXT. This may imply that users who had previously not specified a DDL override may need to.

I know this is a longer term goal, but I would much rather see a solution where data types are resolved by dialect so that we can avoid the DDL override and we create the table natively, i.e. using `TEXT` for some databases, `CLOB` or others where there is no `TEXT` type, etc. The same would apply to `VARCHAR` vs `VARCHAR2`, `INTEGER` vs `NUMERIC`, etc.

jpechane: Yes, that's exactly why `VARCHAR` was selected in the first place - to be a bit more compatible. And DDL override capability is something that is expected to be used.

katsulau: Thank you. Indeed, considering that this DDL is used across various RDBs, adopting a mechanism where the data type is appropriately resolved per database dialect seems to be the best approach.

In that case, rather than treating this as a patch fix, I believe it would be more appropriate to consider it as a new feature addition. Would it be okay if I create a ticket for this in JIRA?

I checked the existing tickets under the jdbc-connector component, but I couldn’t find anything that matches this issue, so I was thinking of creating a new one.

And also, I would like to confirm the reasoning behind using VARCHAR(65000).
Is it based on the fact that in MySQL, the maximum row size is **65,535 bytes**?

Additionally, I noticed the following behavior, which concerns me a little:

- In MySQL 5.7, [the default charset is latin1](https://dev.mysql.com/doc/refman/5.7/en/charset-mysql.html), so this DDL executes without any issues.
- From MySQL 8.0, [the default charset has changed to utf8mb4](https://dev.mysql.com/blog-archive/new-defaults-in-mysql-8-0/), which causes this DDL to always fail.

I understand that this is not a fundamental solution, but to ensure compatibility with as many databases as possible, would it be reasonable to reduce the size?

For example, considering that [SQL Server has a maximum VARCHAR size of 8000](https://learn.microsoft.com/en-us/sql/t-sql/data-types/char-and-varchar-transact-sql?view=sql-server-ver16#varchar---n--max--), if 8000 is sufficient for history_data, then using that size might be a viable option.

Naros: Hi @katsulau 

> In that case, rather than treating this as a patch fix, I believe it would be more appropriate to consider it as a new feature addition. Would it be okay if I create a ticket for this in JIRA?

I don't think we need a new issue; repurposing the existing one is perfectly acceptable.

> And also, I would like to confirm the reasoning behind using VARCHAR(65000).

Several databases, such as PostgreSQL and MariaDB, support this type and size column definition beyond MySQL.

> I understand that this is not a fundamental solution, but would it be reasonable to reduce the size to ensure compatibility with as many databases as possible?

For a dialect-based approach, resolving to TEXT or CLOB based on the dialect would be the most appropriate choice, as that eliminates the immediate concern of 65k being incompatible or too small.

> if 8000 is sufficient for history_data, then using that size might be a viable option.

My concern is that even 8k may be insufficient for wide tables.  

I think we should always resolve to the largest most effective character-based data type (i.e. TEXT, VARCHAR(MAX), CLOB, etc). Should a user want to optimize the DDL so that it uses a column type that is stored in the same data page row as the other data, they can supply a custom ddl and use VARCHAR(8000) and so on depending on their dialect.  wdyt?

katsulau: @Naros 
Thank you. Considering what you mentioned, I agree that covering each database type with a single DDL would be quite challenging, so I will proceed with implementing a mechanism in this branch to switch the DDL based on the database type.

Before starting the implementation, I’d like to share the approach.
I plan to define an interface like `DatabaseDialectProvider` and apply the Strategy Pattern for switching. The selection logic will be based on jdbcUrl, which will determine the appropriate implementation dynamically.

I’d appreciate it if you could let me know if anything seems off or if there are any concerns.

Naros: While using the JDBC URL might work in a general sense, I do wonder if using the `DatabaseMetaData` would be better, where you connect, get the metadata, and then create the Dialect based on the metadata. wdyt @katsulau ?

katsulau: @Naros 
Sounds great! Avoiding the need to parse the URL and using `DatabaseMetaData#getDatabaseProductName() `seems like a reliable way to determine the database type. I’ll proceed with that approach. Thank you!

Naros:

Hi @katsulau, please sign off all commits with:

```
git commit -s
```

Naros:

PR #6230, opened 2025-03-17
DBZ-8786 Improve Error Handling for Duplicate server_id / server_uuid in MySQL Connector

minleejae:
https://issues.redhat.com/browse/DBZ-8786


jpechane:
Hi, thanks for the PR. I wonder if you could use getErrorCode()/getSQLState() on SQL=Exception too to make the handler local independent. Please see `OracleErrorHandler` as an example.

minleejae:
Hi, @jpechane 
I’ve updated the logic to use getErrorCode() (1236).

References:
https://dev.mysql.com/doc/mysql-errors/8.0/en/server-error-reference.html
https://mariadb.com/kb/en/mariadb-error-codes/

jpechane:
@minleejae Applied, thanks!

PR #6231, opened 2025-03-17
DBZ-7689 Adding BYTES to VARBINARY type support for SQLServer in JDBC Sink connector

yuriyv:


jpechane:
@yuriyv Hi, thanks for PR. Please squash the commits and update both commit message and PR title to use correct identifier - [DBZ-8790](https://issues.redhat.com/browse/DBZ-8790)
Could you please also add a test? And it seems to me that in JavaDoc you refer by name not o yu but to an other person.

yuriyv:
should i close this PR?

PR #6232, opened 2025-03-17
DBZ-8790 Adding BYTES to VARBINARY type support for SQLServer in JDBC…

yuriyv:
… Sink connector


Naros:
hI @yuriyv if you remove the merge commit it will pass the commit message check.

PR #6234, opened 2025-03-18
Cherry pick metric endpoints commit into branch 3.0

AlvarVG:
Add commit 55d3938d394ebc83fab1d4ba6c92547a80ce1f3b into branch 3.0


vjuranek:
@AlvarVG thanks, but I don't think this is needed, we run the tests from the `main` branch, so I don't think we need to backport it.

PR #6237, opened 2025-03-19
DBZ-8780 fix Kafka Channel Signal to auto Commit

zakbenallal:


jpechane:
@mfvitale It seems that `ENABLE_AUTO_COMMIT_CONFIG` is configured twice in settings. Could you please double check the code? It seems to me that the PR is correct in the removal.

mfvitale:
@jpechane Yes, the PR correct. I was checking why our tests not caught the issue and this is because the `KafkaSignalChannel` does a `commit` on the `close()`, so during a gracefully shutdown (when we stop the connector in the tests) the offset are effectively committed. 

I asked @zakbenallal on Jira what was the step he did to reproduce the issue and seems (need a confirmation) that he is using Debezium on Kubernetes (I suppose with Debezium Server) so I would like to understand the details to see if it was a a gracefully or not shutdown to see if we miss something somewhere. 

In any case. I think that this can be merged.

zakbenallal:
Thank you for the merge! I confirm that this fixes our problem. Tested on MySQL and PostgreSQL. 🎉

With this fix, the signal message is now committed immediately, and the snapshot starts correctly. When the pod is restarted, Debezium resumes the snapshot properly.

**For your information: I use Debezium on a Kafka connector (via Strimzi CRD).**

mfvitale:
@zakbenallal Glad that this helped. 

>  the signal message is now committed immediately

Just a note on that. The commit happen based on `auto.commit.interval.ms` that has a default of  `5s`.

PR #6238, opened 2025-03-19
DBZ-8825 Make methods for adding fields into the record reuseable

vjuranek:
https://issues.redhat.com/browse/DBZ-8825


jpechane:
@vjuranek Applied, thanks.
Test failure is unrelated.

PR #6239, opened 2025-03-20
DBZ-8825 Make NewEntry record public

vjuranek:
The the API of methods for adding new fields into the record use `NewEntry`, it has to be public to make the methods actually usefull outside `core` package.

Sorry for this silly mistake and another PR:-(


vjuranek:
Converting to draft in case I find some other issue, to include all the fixes into one PR.

jpechane:
@vjuranek Given the scope of changes feel free to merge yourself when ready!

PR #6242, opened 2025-03-21
Dbz 8753 Align JDBC storage configuration naming

kmos:
## Context

JDBC storage configuration names are not aligned with the others storage module. This misalignment can create some issue on configurations. The Proposal convention is the following:

| current    | new |
| -------- | ------- |
| offset.storage.jdbc.*  | offset.storage.jdbc.connection.*    |
| offset.storage.jdbc.offset.table.* | offset.storage.jdbc.table.*     |
| schema.history.internal.jdbc.* | schema.history.internal.jdbc.connection.*|
| schema.history.internal.jdbc.schema.history.table.* | schema.history.internal.jdbc.table.*|

## Changes

The `JdbcSchemaHistoryConfig`, `JdbcOffsetBackingStoreConfig` fields are aligned with the new convention and the current names are deprecated for future removal.

closes https://issues.redhat.com/browse/DBZ-8573

### DIFF
--- a/debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/JdbcCommonConfig.java
+++ b/debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/JdbcCommonConfig.java
@@ -29,29 +29,29 @@ public class JdbcCommonConfig {
     public static final String CONFIGURATION_FIELD_PREFIX_STRING = "jdbc.";
     public static final String CONFIGURATION_FIELD_CONNECTION_GROUP = CONFIGURATION_FIELD_PREFIX_STRING + "connection.";
 
-    public static final Field PROP_JDBC_URL = Field.create(CONFIGURATION_FIELD_PREFIX_STRING + CONFIGURATION_FIELD_CONNECTION_GROUP + "url")
+    public static final Field PROP_JDBC_URL = Field.create(CONFIGURATION_FIELD_CONNECTION_GROUP + "url")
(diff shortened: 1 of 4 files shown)


debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/JdbcCommonConfig.java
```diff
@@ -27,28 +27,34 @@ public class JdbcCommonConfig {
     private static final Logger LOGGER = LoggerFactory.getLogger(JdbcCommonConfig.class);
 
     public static final String CONFIGURATION_FIELD_PREFIX_STRING = "jdbc.";
+    public static final String CONFIGURATION_FIELD_CONNECTION_GROUP = "connection.";
```
Naros, line 30:
```suggestion
    public static final String CONFIGURATION_FIELD_CONNECTION_GROUP = CONFIGURATION_FIELD_PREFIX_STRING + "connection.";
```
I wonder it this would just be a bit more legible because then we can minimize the concatenation in the Field defs below and just use `CONFIGURATION_FIELD_CONNECTION_GROUP`.  wdyt?

Naros (commented):
LGTM, just left one tiny nitpick for legibility.

jpechane:
This is going to be a breaking change. Given that we are late in the 3.1 release cycle we should merge this PR at the beginning og 3.2 cycle.

jpechane:
@kmos Applied, thanks! I am sorry I missed the backward compatibility behaviour.

PR #6243, opened 2025-03-21
Dbz 8832 Unexpected null value for Field Configuration deprecated aliases

kmos:
## Context

the method `withDefault()` for `Field` class used for configuration purpose create a new `Field` value without `deprecatedAliases`

**Expected Behavior:**  
the values for deprecated aliases should be mantained even if we have a default one

closes: https://issues.redhat.com/browse/DBZ-8832


jpechane:
@kmos Applied, thanks

PR #6247, opened 2025-03-24
DBZ-8525: Improves log message introduced in #6222

b-goyal:
This PR improves the log message introduced with the change done in #6222 
Current log line:
`Incremental snapshot for table 's1.part2' will end at position`
Update to log line
`Incremental snapshot for table 's1.part2' will end at maximum key`


jpechane:
@b-goyal LGTM, thanks

PR #6251, opened 2025-03-26
DBZ-8844 Don't skip transaction at snapshot boundary

Naros:
https://issues.redhat.com/browse/DBZ-8844


jpechane:
@Naros Applied, thanks.
Test failures unrelated,

PR #6254, opened 2025-03-26
DBZ-8702 Intorduce Debezium AI module with embeddings SMT

vjuranek:
https://issues.redhat.com/browse/DBZ-8702


debezium-ai/debezium-ai-embeddings-base/src/main/java/io/debezium/ai/embeddings/AbstractEmbeddingsSmt.java
```diff
@@ -0,0 +1,171 @@
...
+            .withDisplayName("Name of the field which would contain the embeddings of the input field.")
+            .withType(ConfigDef.Type.STRING)
+            .withWidth(ConfigDef.Width.SHORT)
+            .withImportance(ConfigDef.Importance.HIGH)
+            .required()
+            .withDescription(
+                    "Name of the field which which will be appended to the record and which would contain the embeddings of the content `filed.source` field. Supports also nested fields.");
+
+    private static final Field.Set ALL_FIELDS = Field.setOf(TEXT_FIELD, EMBEDDGINS_FIELD);
+    private static final Schema VECTOR_SCHEMA = SchemaBuilder.array(Schema.FLOAT32_SCHEMA).schema();
```
vjuranek, line 65: I wonder, if I should use `io.debezium.data.vector.FloatVector` (IMHO yes, will include it in final version).

debezium-ai/debezium-ai-embeddings-minilm/src/main/java/io/debezium/ai/embeddings/EmbeddingsMiniLmSmt.java
```diff
@@ -0,0 +1,22 @@
...
+import dev.langchain4j.model.embedding.onnx.allminilml6v2.AllMiniLmL6V2EmbeddingModel;
+
+/**
+ * Embeddings SMT which uses <a href="https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2">MiniLmL6V2 model</a> model for creating embeddings.
+ *
+ * @author vjuranek
+ */
+public class EmbeddingsMiniLmSmt<R extends ConnectRecord<R>> extends AbstractEmbeddingsSmt<R> {
+    public EmbeddingModel getModel() {
+        return new AllMiniLmL6V2EmbeddingModel();
```
mfvitale, line 20: I'm wondering if we should use the [HuggingFace](https://docs.langchain4j.dev/integrations/embedding-models/hugging-face) directly paired with a property to choose the model. As advanced case the users can use different models with a custom STM?

mfvitale: The best would be if we could permit choosing from [langchain4j suported models](https://docs.langchain4j.dev/category/embedding-models) directly with a config removing the need to create a custom STM.

vjuranek: I'm going to add it later on as a special module. The problematic thing is (not only for HUggingFace, but for most of the models), that you need an access token, so it's little bit hard to write tests (but more investigation in this regard is still WIP, so maybe there are some options).
Another one on the my TODO list is Ollama - here exists even [testconatiner](https://java.testcontainers.org/modules/ollama/) for it (possible issue here is size of the container and models).

mfvitale: > Another one on the my TODO list is Ollama - here exists even [testconatiner](https://java.testcontainers.org/modules/ollama/) for it (possible issue here is size of the container and models).

Yes, for the test part I was thinking to it or also `LocalAI` or the one you have used (`AllMiniLmL6V2EmbeddingModel`).

I dunno what are the performances of the `AllMiniLmL6V2EmbeddingModel` if they are not comparable with the best available model I think that we are just forcing users to create a custom SMT to have good results. This is my concern.

vjuranek: I don't expect this one to be used in the production, it meant it rather as a toy SMT, which works out of the box and can be used for PoC or for testing purposes (for us , but for users as well).

But in general yes, I expect users to eventually to write their SMTs, because there are too many models and service we wont be able to cover all of them anyway, so basically my idea is to make implementation of custom SMT as easy as possible (just provide model you want to use) and implement SMTs only for few most common models/services (huggingface and ollama, eventually chatgpt - possibly based on the community input)

debezium-ai/debezium-ai-embeddings-base/src/main/java/io/debezium/ai/embeddings/AbstractEmbeddingsSmt.java
```diff
@@ -0,0 +1,172 @@
...
+import dev.langchain4j.model.embedding.EmbeddingModel;
+
+/**
+ * Base class of the embeddings SMT, which provides common configuration and manipulations with the Connect record.
+ * Child classes has to implement at least {@link AbstractEmbeddingsSmt#getModel()} class, which provides specific
+ * model which will be used for creating the embeddings.
+ *
+ * @author vjuranek
+ */
+public abstract class AbstractEmbeddingsSmt<R extends ConnectRecord<R>> implements Transformation<R>, Versioned {
```
jpechane, line 44:
```suggestion
public abstract class AbstractEmbeddingsTransformation<R extends ConnectRecord<R>> implements Transformation<R>, Versioned {
```

debezium-ai/debezium-ai-embeddings-base/src/main/java/io/debezium/ai/embeddings/AbstractEmbeddingsSmt.java
```diff
@@ -0,0 +1,172 @@
...
+
+    private static final Field TEXT_FIELD = Field.create("field.source")
+            .withDisplayName("Name of the record field from which embeddings should be created.")
+            .withType(ConfigDef.Type.STRING)
+            .withWidth(ConfigDef.Width.SHORT)
+            .withImportance(ConfigDef.Importance.HIGH)
+            .required()
+            .withDescription("Name of the field of the record which content will be used as an input for embeddings. Supports also nested fields.");
+
+    private static final Field EMBEDDGINS_FIELD = Field.create("field.embeddings")
```
jpechane, line 56: Can we make it optional and replace the whole payload with embeddings hwn not provided?

debezium-ai/debezium-ai-embeddings-base/src/main/java/io/debezium/ai/embeddings/AbstractEmbeddingsSmt.java
```diff
@@ -0,0 +1,172 @@
...
+            .withDisplayName("Name of the field which would contain the embeddings of the input field.")
+            .withType(ConfigDef.Type.STRING)
+            .withWidth(ConfigDef.Width.SHORT)
+            .withImportance(ConfigDef.Importance.HIGH)
+            .required()
+            .withDescription(
+                    "Name of the field which which will be appended to the record and which would contain the embeddings of the content `filed.source` field. Supports also nested fields.");
+
+    private static final Field.Set ALL_FIELDS = Field.setOf(TEXT_FIELD, EMBEDDGINS_FIELD);
```
jpechane, line 65:
```suggestion
    public static final Field.Set ALL_FIELDS = Field.setOf(TEXT_FIELD, EMBEDDGINS_FIELD);
```

debezium-ai/debezium-ai-embeddings-base/src/main/java/io/debezium/ai/embeddings/AbstractEmbeddingsSmt.java
```diff
@@ -0,0 +1,172 @@
...
+    private static final Schema VECTOR_SCHEMA = FloatVector.schema();
```
jpechane, line 66:
```suggestion
    private static final Schema EMBEDDING_SCHEMA = FloatVector.schema();
```

debezium-ai/debezium-ai-embeddings-minilm/src/main/java/io/debezium/ai/embeddings/EmbeddingsMiniLmSmt.java
```diff
@@ -0,0 +1,22 @@
...
+
+import dev.langchain4j.model.embedding.EmbeddingModel;
+import dev.langchain4j.model.embedding.onnx.allminilml6v2.AllMiniLmL6V2EmbeddingModel;
+
+/**
+ * Embeddings SMT which uses <a href="https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2">MiniLmL6V2 model</a> model for creating embeddings.
+ *
+ * @author vjuranek
+ */
+public class EmbeddingsMiniLmSmt<R extends ConnectRecord<R>> extends AbstractEmbeddingsSmt<R> {
```
jpechane, line 18: Do we need to implement it as another PR? Can't we just have a model factory class that will receive configuration options from the transformation and provide model? Thus there will be only single transformation and it will just have a config options choosing the right factory. I guess this is in-line with @mfvitale comments.

debezium-ai/debezium-ai-embeddings-minilm/src/main/java/io/debezium/ai/embeddings/EmbeddingsMiniLmSmt.java
jpechane: @vjuranek I agree that we should provide just a subset of models, IMHO this set would make the most sense

- huggingface
- ollama
- chatgpt
- claude

We can let community to provide the others.

jpechane (commented):
@vjuranek Good start, thanks! I left couple of comments.

debezium-ai/debezium-ai-embeddings-minilm/src/main/java/io/debezium/ai/embeddings/EmbeddingsMiniLmSmt.java
vjuranek: This id definitely doable, but I think the main issue with this approach is that you would have to include all the dependencies into a single jar file, which may be eventually pretty large. I guess this is the reason why langchain4j (and other project like e.g. [Quarkus](https://github.com/quarkiverse/quarkus-langchain4j/blob/main/model-providers/pom.xml#L13-L30)) separates each model into separate module and build a dedicated artifact for each model.

debezium-ai/debezium-ai-embeddings-base/src/main/java/io/debezium/ai/embeddings/AbstractEmbeddingsSmt.java
vjuranek: sure

debezium-ai/debezium-ai-embeddings-minilm/src/main/java/io/debezium/ai/embeddings/EmbeddingsMiniLmSmt.java
mfvitale: I think that Jiri mean to have a separate maven module for each supported langchain4j model. So you will have something like this:

* debezium-ai-embeddings-allminiLmL6V2E
* debezium-ai-embeddings-huggingface
* debezium-ai-embeddings-ollama
* debezium-ai-embeddings-chatgpt
* debezium-ai-embeddings-claude

In that way each module as it dedicated model dependency and so the users don't need to use them all.

vjuranek: Well, this is exactly what I have now and what I mentioned at the very beginning (`I'm going to add it later on as a special module.`). So this is either misunderstanding and we are in agreement or I'm still missing what is your preferred way how to implement it.

mfvitale: Well my initial idea was to have all in a unique module but then you highlight the problem of dimensions and so I lead to what was your initial proposal. 

Another approach could be to have the `factory` approach proposed by Jiri and just add dependency as provided so the user then need to provide it. This approach could simplify the AI module structure. 

Honestly don't have strong opinion. The important thing is, for the users, to have an out-of-the-box way to use different models without going to customize.

jpechane: @vjuranek I think our only disagreemnt is if there should be multiple SMTs - each in a separate module. Or a single SMT that will just just use an embedding provided by module on a classpath.
I'd prefer the latter but it might make sense to sit on a call and discuss pros and cons of each of the options.

vjuranek: @jpechane @mfvitale done, please review if is this the direction you intended - there is only a single SMT `FieldToEmbedding` in `debezium-ai-embeddings` module, which relies on model factory which delivers concrete implementation of the model used for creating embeddings. Model factory is loaded via SPI and other modules contain implementations of this model factory.

jpechane (commented):
@vjuranek That's exactly what I was looking for! Thanks
There is only one thing missing - assembly descriptor similar to https://github.com/debezium/debezium/blob/main/debezium-assembly-descriptors/src/main/resources/assemblies/storage-distribution.xml

The built assamblies should not use `plugin` suffix but `embeddings` and should not contain cor, API etc. Debezium libraries, only what's needed to be copied into connector directory to get SMT running.

As we are almost there I'd really love to get it into 3.1.0.FInal as Incubating fetaure - it is omcpletely isolated so there is no risk that it breaks Debezium as such. And it will allow us to write a blogpost and example around it.

Documentation can be merged in a separate post-release PR and the announcement will just mention that the documentation is going to be published soon.

mfvitale:
@vjuranek great work!

vjuranek:
@jpechane done. I removed from the bits also `debezium-ai-embeddings` which contains SMT itself. So you have to download  `debezium-ai-embeddings-3.1.0-SNAPSHOT-embeddings.tar.gz` and e.g. `debezium-ai-embeddings-ollama-3.1.0-SNAPSHOT-embeddings.tar.gz`. If you want to have  `debezium-ai-embeddings` in all the bits so you would need only to download and deploy e.g. `debezium-ai-embeddings-ollama-3.1.0-SNAPSHOT-embeddings.tar.gz`, let me know.

jpechane:
@vjuranek Yes, I'd like to have embedding provider granularity. So the `debezium-ai-embeddings` should be included in ollama. So at the end there will be only two archives one for ollama and one for minilm.

vjuranek:
@jpechane done

jpechane:
@vjuranek And applied, huge thanks!
BTW it is funny if you compare size of ollama and minilm distrost and then try to fgure out what is mini about mini :-)

PR #6256, opened 2025-03-26
DBZ-8818 Removes obsolete admonition; adds note about idle thread exception

roldanbob:
[DBZ-8818](https://issues.redhat.com/browse/DBZ-8818)

Removes conditionalized incubating and TP admonitions.
Adds note describing factors that can contribute to idle thread exceptions and how to recover from the error by reverting the value of `snapshot.max.threads` and then retrying the snapshot.

This change should be backported to 3.0.


jpechane:
@roldanbob Applied, thanks

PR #6258, opened 2025-03-27
DBZ-8848 Update Azure Identity to 1.15.4

Naros:
https://issues.redhat.com/browse/DBZ-8848


jpechane:
@Naros Applied, thanks

PR #6264, opened 2025-03-27
DBZ-8823 Removes TP and incubating statements for LOB and hybrid mining

roldanbob:
[DBZ-8823](https://issues.redhat.com/browse/DBZ-8823)

The current downstream release exposes the use of *LOB character types and the hybrid mining strategy as GA features. This change removes admonitions that designated content for those features as Technology Preview or incubating, and also removes obsolete conditionals that prevented that content from rendering in the product edition of the documentation.

This change should be backported to 3.0.


Naros (approved):

Naros:
Applied, thanks @roldanbob !

PR #6267, opened 2025-03-28
Consolidated CI workflows

Naros:


Naros:
@jpechane Should the **IBMi** integration tests continue to be skipped?

jpechane:
@Naros Yes, they should, we need to have ITs on an infra with secure acess to a reliable/permanent iSeries instance with ideally more than one environment.

Naros:
> @Naros Yes, they should, we need to have ITs on an infra with secure acess to a reliable/permanent iSeries instance with ideally more than one environment.

Thanks @jpechane, that had completely slipped my mind.

PR #6269, opened 2025-03-28
DBZ-8755 Add binlog admonition

jpechane:
https://issues.redhat.com/browse/DBZ-8755


roldanbob (changes requested):
@jpechane Thanks for taking the first pass on this. I got confused for a sec, because I had this on my to-do list and I initially thought this was a dup!

I've made a few suggestions. LMKWYT. Thanks.

jpechane:
@roldanbob Thanks for the review

PR #6270, opened 2025-03-28
DBZ-8855 Add configured headers for read, insert, and delete events.

Naros:
https://issues.redhat.com/browse/DBZ-8855

@jpechane I'm on the fence on this change because, on one hand I like the idea that it simplifies and provides consistent logic when paired with other transforms; however at the same time, I can see the argument that when paired with other transforms, perhaps having a predicate chain to run one or the other transform makes sense.

Let me know what you think.


jpechane:
@Naros Applied, thanks. This is fine as esp. this SMT has multiple "fixes" built-in that were included before SMT predicates even existed.

PR #6280, opened 2025-03-31
DBZ-8788 Document mTLS using JKS with Oracle

Naros:
https://issues.redhat.com/browse/DBZ-8788

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/oracle.adoc
+++ b/documentation/modules/ROOT/pages/connectors/oracle.adoc
@@ -3299,21 +3299,24 @@ endif::community[]
 When connecting to Oracle using mutual TLS authentication (mTLS), this involves both the connector and the database service proving their identities.
 The {prodname} for Oracle connector relies on the Oracle JDBC driver's built-in functionality to support mTLS authentication.
 
-There are two ways to establish mTLS connections:
+You can use either of the following methods to establish mTLS connections between {prodname} and Oracle:
 
-* xref:#using-mtls-secure-connections-java-key-trust-stores[]
-* xref:#using-mtls-secure-connections-oracle-wallet[]
+* xref:#using-mtls-secure-connections-java-key-trust-stores[Using Java key and trust stores to configure mTLS connections]
+* xref:#using-mtls-secure-connections-oracle-wallet[Using Oracle Wallet to configure mTLS connections]
 
+// Type: procedure
+// ModuleID: oracle-connector-using-java-key-trust-stores-to-configure-mtls
+// Title: Using Java key and trust stores to configure the Oracle connector to use mTLS
 [[using-mtls-secure-connections-java-key-trust-stores]]
 ==== Java key/trust stores
 
 .Prerequisites
-* Verify connector can access key/trust store files directly.
-* Verify database TNS listener supports `tcps` secure connections.
+* Verify connector can access the configured key and trust store files.
+* Verify that the database TNS (Transparent Network Substrate) listener supports TCPS secure connections.
 
 .Procedure
-* Configure the {prodname} Oracle connector, as illustrated below in the example.
-
+* Configure the {prodname} Oracle connector, as illustrated in the following example.
++
 .Example: {prodname} Oracle connector TLS configuration
 [source,json,indent=0]
 ----
@@ -3327,13 +3330,18 @@ There are two ways to establish mTLS connections:
   "driver.javax.net.ssl.trustStoreType": "JKS"
 }
 ----
++
 
 [NOTE]
 ====
-It is important for TLS encryption to use `database.url` rather than `database.host`.
-The URL provides a TNS connection string which is what explicitly controls the secure connection requirement by setting the protocol to `tcps`.
+For {prodname} to use TLS encryption to communicate with Oracle, requires a TCPS connection with the server.
+To establish a TCPS connection, you must configure the connector to use the `database.url` property, and not the `database.host` property.
+Unlike the `database.host` property, the `database.url` property permits you to define an Oracle TNS (Transparent Network Substrate) connection string that explicitly requires use of the `TCPS` protocol.
 ====
 
+// Type: procedure
+// ModuleID: oracle-connector-using-oracle-wallet-to-configure-mtls
+// Title: Using Oracle wallet to configure the connector to use mTLS
 [[using-mtls-secure-connections-oracle-wallet]]
 ==== Oracle Wallet
 
--- a/documentation/modules/ROOT/pages/connectors/oracle.adoc
+++ b/documentation/modules/ROOT/pages/connectors/oracle.adoc
@@ -3299,6 +3299,44 @@ endif::community[]
 When connecting to Oracle using mutual TLS authentication (mTLS), this involves both the connector and the database service proving their identities.
 The {prodname} for Oracle connector relies on the Oracle JDBC driver's built-in functionality to support mTLS authentication.
 
+There are two ways to establish mTLS connections:
+
+* xref:#using-mtls-secure-connections-java-key-trust-stores[]
+* xref:#using-mtls-secure-connections-oracle-wallet[]
+
+[[using-mtls-secure-connections-java-key-trust-stores]]
+==== Java key/trust stores
+
+.Prerequisites
+* Verify connector can access key/trust store files directly.
+* Verify database TNS listener supports `tcps` secure connections.
+
+.Procedure
+* Configure the {prodname} Oracle connector, as illustrated below in the example.
+
+.Example: {prodname} Oracle connector TLS configuration
+[source,json,indent=0]
+----
+{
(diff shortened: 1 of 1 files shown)


documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -3299,6 +3299,44 @@ endif::community[]
...
+  "driver.javax.net.ssl.trustStore": "<path-to-jks-truststore>",
+  "driver.javax.net.ssl.trustStorePassword": "<truststore-password>",
+  "driver.javax.net.ssl.trustStoreType": "JKS"
+}
+----
+
+[NOTE]
+====
+It is important for TLS encryption to use `database.url` rather than `database.host`.
+The URL provides a TNS connection string which is what explicitly controls the secure connection requirement by setting the protocol to `tcps`.
```
roldanbob, line 3334:
```suggestion
For {prodname} to use TLS encryption to communicate with Oracle, requires a TCPS connection with the server.
To establish a TCPS connection, you must configure the connector to use the `database.url` property, and not the `database.host` property. 
Unlike the `database.host` property, the `database.url` property permits you to define an Oracle TNS (Transparent Network Substrate) connection string that explicitly requires use of the `TCPS` protocol. 
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -3299,6 +3299,44 @@ endif::community[]
 When connecting to Oracle using mutual TLS authentication (mTLS), this involves both the connector and the database service proving their identities.
 The {prodname} for Oracle connector relies on the Oracle JDBC driver's built-in functionality to support mTLS authentication.
 
+There are two ways to establish mTLS connections:
```
roldanbob, line 3302:
```suggestion
You can use either of the following methods to establish mTLS connections between {prodname} and Oracle:
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -3299,6 +3299,44 @@ endif::community[]
...
+
+* xref:#using-mtls-secure-connections-java-key-trust-stores[]
```
roldanbob, line 3304:
```suggestion
* xref:#using-mtls-secure-connections-java-key-trust-stores[Using Java key and trust stores to configure mTLS connections].
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -3299,6 +3299,44 @@ endif::community[]
...
+* xref:#using-mtls-secure-connections-oracle-wallet[]
```
roldanbob, line 3305:
```suggestion
* xref:#using-mtls-secure-connections-oracle-wallet[Using Oracle Wallet to configure mTLS connections].
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
roldanbob: In shared content (community/product), we must provide link text, because the text is not automatically resolved if the link is used downstream.  
In this particular case, the upstream link text will be used, rather than the topic title specified by the downstream Nebel annotation, but better that than no text, and we can live with the small discrepancy as long as the link works.

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -3299,6 +3299,44 @@ endif::community[]
...
+
```
roldanbob, line 3306:
```suggestion
// Type: procedure
// ModuleID: oracle-connector-using-java-key-trust-stores-to-configure-mtls
// Title: Using Java key and trust stores to configure the Oracle connector to use mTLS
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -3299,6 +3299,44 @@ endif::community[]
...
+  "driver.javax.net.ssl.trustStoreType": "JKS"
+}
+----
+
+[NOTE]
+====
+It is important for TLS encryption to use `database.url` rather than `database.host`.
+The URL provides a TNS connection string which is what explicitly controls the secure connection requirement by setting the protocol to `tcps`.
+====
+
```
roldanbob, line 3336:
```suggestion
// Type: procedure
// ModuleID: oracle-connector-using-oracle-wallet-to-configure-mtls
// Title: Using Oracle wallet to configure the connector to use mTLS
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -3299,6 +3299,44 @@ endif::community[]
...
+* xref:#using-mtls-secure-connections-java-key-trust-stores[]
+* xref:#using-mtls-secure-connections-oracle-wallet[]
+
+[[using-mtls-secure-connections-java-key-trust-stores]]
+==== Java key/trust stores
+
+.Prerequisites
+* Verify connector can access key/trust store files directly.
+* Verify database TNS listener supports `tcps` secure connections.
```
roldanbob, line 3312:
```suggestion
* Verify that the connector can access the configured key and trust store files.
* Verify that the database TNS (Transparent Network Substrate) listener supports TCPS secure connections.
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -3299,6 +3299,44 @@ endif::community[]
...
+
+.Procedure
+* Configure the {prodname} Oracle connector, as illustrated below in the example.
```
roldanbob, line 3315:
```suggestion
* Configure the {prodname} Oracle connector, as illustrated in the following example. +
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -3299,6 +3299,44 @@ endif::community[]
...
+
```
roldanbob, line 3316:
```suggestion
+
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -3299,6 +3299,44 @@ endif::community[]
...
+.Example: {prodname} Oracle connector TLS configuration
```
roldanbob, line 3317:
```suggestion
.Example: {prodname} Oracle connector TLS configuration
=====
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -3299,6 +3299,44 @@ endif::community[]
...
+  "database.url": "jdbc:oracle:thin@(DESCRIPTION=(ADDRESS=(PROTOCOL=tcps)(HOST=<host>)(PORT=<port>))(CONNECT_DATA(SERVICE_NAME=<service>)))",
+  "driver.javax.net.ssl.keyStore": "<path-to-jks-keystore>",
+  "driver.javax.net.ssl.keyStorePassword": "<keystore-password>",
+  "driver.javax.net.ssl.keyStoreType": "JKS",
+  "driver.javax.net.ssl.trustStore": "<path-to-jks-truststore>",
+  "driver.javax.net.ssl.trustStorePassword": "<truststore-password>",
+  "driver.javax.net.ssl.trustStoreType": "JKS"
+}
+----
+
```
roldanbob, line 3330:
```suggestion
=====
+
```
To use example titles, the example must be formatted as an example block.

roldanbob (changes requested):
@Naros Thanks for adding this. Reminds me that on a back burner I still have [DBZ-6168](https://issues.redhat.com/browse/DBZ-6168) for clarifying the connection string settings for XStream/TCPS. This looks to be very similar. I had written up something, but I never had a good feeling for whether it covered the required bases. Now that we're exposing the XStreams content, if you can provide verification of what needs to be done, maybe we can push that forward.

Naros:
Hi @roldanbob, so looking at that linked Jira, that one specifically was discussing Oracle Wallet for XStream.  The same would hold true whether the user is using LogMiner or XStream (as documented in the Wallet section).  What I've added is for mTLS using Java Keystores, which is a non-Wallet approach. This should also work for XStream, too.

We generally structured the documentation in a way to favor LogMiner but I think we might want to rethink that strategy long-term. It may pay to take one of two likely strategies; though there may be more.

* Split the documentation into General,  LogMiner, XStream, and OLR sections.  We have this mostly now except we've embedded the LogMiner content into the General content. As we want to have an emphasis on having both of these adapters as supported features, it may be wise to organize the content with equal weight.
* We tailor the content generally and then weave in each adapter's specifics into top-level generalized sections. 

I personally prefer the former, but I"d like your opinion.

roldanbob:
> I personally prefer the former, but I"d like your opinion.

I agree with the proposed approach. I don't think that I can address this for 3.0.8, but we can make it a priority for Debezium.next. For now, I think that you can merge this, yes? And post-3.0.8 we can work in the content for DBZ-6168 (waller or Java keystore).

PR #6281, opened 2025-04-01
DBZ-8867 Bump Chicory and use ByteArrayMemory

andreaTP:
`ByteArrayMemory` is a faster implementation of the Memory.

### DIFF
--- a/debezium-bom/pom.xml
+++ b/debezium-bom/pom.xml
@@ -38,7 +38,7 @@
         <version.groovy>4.0.17</version.groovy>
         <version.graalvm.js>23.0.2</version.graalvm.js>
         <!-- Wasm -->
-        <version.chicory>1.0.0</version.chicory>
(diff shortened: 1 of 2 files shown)


jpechane:
@andreaTP Hi, thanks for the PR. Could please create a Jira issue for this PR? Ideally if you could also link source about `ByteArrayMemory` being faster. This just to make sure that the descision/change is properly documented.

andreaTP:
Done @jpechane , for future reference, do you open JIRAs even for plain version bumps or only if they contain also code changes(even if minimal like in this case)?

Naros:
Hi @andreaTP we tend to prefer a Jira issue except under the following conditions:
- Making a typo fix in the documentation, then use `[docs]` prefix
- Making a GitHub CI change, then use `[ci]` prefix
- Performing a release task, then use `[release]` prefix

PR #6283, opened 2025-04-01
When the error 'Unable to register metrics as an old set with the same name ......' occurs

microhuang:
When the error 'Unable to register metrics as an old set with the same name' occurs, it can quickly locate the cause of the error


Naros:
Hi @microhuang can you please raise a Jira and fix the commit message?

microhuang:
> [DBZ-xxx JIRA issue key](https://github.com/debezium/debezium/blob/main/CONTRIBUTE.md#making-changes).

Sorry, I don't know where to get the 'issue key'.

mfvitale:
@microhuang You need to open an issue[here](https://issues.redhat.com/projects/DBZ/issues)

PR #6284, opened 2025-04-01
Log JMX MBean name when registration fails

microhuang:
https://issues.redhat.com/browse/DBZ-8862 

When the error 'Unable to register metrics as an old set with the same name ......' occurs, it can quickly locate the cause of the error.


Naros (approved):

jpechane:
@microhuang Applied, thanks

PR #6287, opened 2025-04-02
DBZ-8868 Make getHistoryRecordComparator() public

wukachn:
https://issues.redhat.com/browse/DBZ-8868

### DIFF
--- a/debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
+++ b/debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
@@ -320,7 +320,7 @@ public class MariaDbConnectorConfig extends BinlogConnectorConfig {
     }
 
     @Override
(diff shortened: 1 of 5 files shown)


Naros:
That would be great @wukachn.

wukachn:
https://github.com/debezium/debezium-connector-db2/pull/171
https://github.com/debezium/debezium-connector-informix/pull/74

@Naros How are versions tagged in the debezium repos? I see you've added the PR label for 3.2, will this change be added to `3.2.0-SNAPSHOT` on merge? then the above 2 PRs can get merged with their changes and a version bump of `3.2.0-SNAPSHOT`?

I see the informix connector is already on `3.2.0-SNAPSHOT`, whereas the db2 one is `3.1.0-SNAPSHOT`

jpechane:
@wukachn Db2 is also 3.2.0-SNAPSHOT so you probbably hit an interval when the versions were updated

jpechane:
@wukachn Applied, thanks

PR #6290, opened 2025-04-03
DBZ-8869 Fix several SQL relaxed quote corner cases

Naros:
https://issues.redhat.com/browse/DBZ-8869

### DIFF
--- a/.packit.yaml
+++ b/.packit.yaml
@@ -234,6 +234,7 @@ jobs:
     manual_trigger: true
     labels:
       - oracle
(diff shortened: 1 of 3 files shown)


Naros:
@jpechane I really have no idea why the workflow ran all the non-Oracle tests, clearly imo unrelated :)

jpechane:
@Naros Applied, thanks. I'd suppose that maybe the changed files actions were confused by the change in the root dir.

PR #6292, opened 2025-04-03
Dbz 8841 prep oracle xstream docs for use downstream

roldanbob:
[DBZ-8841](https://issues.redhat.com/browse/DBZ-8841)

Updates Oracle connector doc to expose XStream DP content downstream.

Tested in a local Antora and downstream builds.

This change requires a backport to 3.0.

Replaces #6265

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/oracle.adoc
+++ b/documentation/modules/ROOT/pages/connectors/oracle.adoc
@@ -937,9 +937,10 @@ If the {prodname} connector is configured to use the `online_catalog` mode, and 
 
 Following this procedure helps to ensure that Oracle LogMiner can safely reconstruct the SQL for all data changes.
 
+ifdef::community[]
 ==== Hybrid approach
 
-You can enable this strategy by setting the value of the xref:oracle-property-log-mining-strategy[`log.mining.strategy`] configuration property to `hybrid`.
+This is a new, experimental strategy that can be enabled by setting the strategy to `hybrid`.
 The goal of this strategy is to provide the reliability of the `redo_log_catalog` strategy with the performance and low overhead of the `online_catalog` strategy, without incurring the disadvantages of either strategy.
 
 The `hybrid` strategy works by primarily operating in the `online_catalog` mode, meaning that the {prodname} Oracle connector first delegates event reconstruction to Oracle LogMiner.
@@ -952,6 +953,7 @@ The connector reports a failure only if both Oracle LogMiner and the connector a
 You cannot use the `hybrid` mining strategy if the xref:oracle-property-lob-enabled[`lob.enabled`] property is set to `true`.
 If you require streaming `CLOB`, `BLOB`, or `XML` data, only the `online_catalog` or `redo_log_catalog` strategies can be used.
 ====
+endif::community[]
 
 [[oracle-logminer-query-modes]]
 === Query Modes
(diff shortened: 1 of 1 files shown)


documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -36,12 +36,28 @@ ifdef::product[]
 For information about the Oracle Database versions that are compatible with this connector, see the link:{LinkDebeziumSupportedConfigurations}[{NameDebeziumSupportedConfigurations}].
 endif::product[]
 
-{prodname} ingests change events from Oracle by using the native LogMiner database package
+{prodname} ingests change events from Oracle by using the native LogMiner database package,
```
Naros, line 39: Wasn't the `,` (comma) handling here correct in the old version? I would have expected for downstream it would have read "LogMiner database package or the ..." rather than "LogMiner database package, or the ...".  Am I mistaken here? :thinking:

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -1254,26 +1270,22 @@ In the case of the Oracle connector, the structure includes the following fields
 * The {prodname} version.
 * The connector name.
 * Whether the event is part of an ongoing snapshot or not.
-* The transaction id (not includes for snapshots).
-* The SCN of the change.
+* The transaction id (not included for snapshots).
+* The `commit_scn` for the change.
```
Naros, line 1274:
```suggestion
* The `scn` for the change and the `commit_scn` for enclosing transaction.
```
We should probably mention both, either as a single line or two lines. We may even want to include details about the new `start_scn`, `start_ts_ms`, and `commit_ts_ms` which represent the SCN when the transaction started, the time the transaction started, and the time the transaction was committed.  wdyt?

Naros (commented):
Just a few inline comments @roldanbob and then GTG imo.

documentation/modules/ROOT/pages/connectors/oracle.adoc
roldanbob: Sigh. Yeah, technically you're correct, and I was went back and forth about this, trying to balance moving the comma so that it gets inserted where it's not necessary, or having it render oddly with an extra leading space. It would seem that the original conditionalization would insert the comma immediately after the word `package`, but that's not how it renders in the built documentation. 
The other option would be offer two completely distinct sentences for the community and product versions.

documentation/modules/ROOT/pages/connectors/oracle.adoc
roldanbob: I defer to your knowledge on this. My change was actually naive -- I was trying to be more technically accurate without realizing that the two were not synonymous. I'm happy to add the additional details, but I'll have to look into the other options more ... unless you can provide a quick summary here?  :)

roldanbob:
Closing this one, too. Back to the drawing board. I'll have another version up soon.

documentation/modules/ROOT/pages/connectors/oracle.adoc
Naros: `scn`: the system change number associated with the change event
`start_scn`: the system change number for the change event's transaction start
`start_ts_ms`: the timestamp when the change event's transaction started
`commit_scn`: the system change number for the change event's transaction commit
`commit_ts_ms`: the timestamp when the change event's transaction was committed

PR #6294, opened 2025-04-04
DBZ-8877 MySQL/MariaDB connector enhancement

HarisBinSaif:
**Improve MySQL/MariaDB connector resilience during post-schema recovery reconnect**

While investigating an issue where the MySQL connector fails with:

_The client was disconnected by the server because of inactivity_

I found that even though the connector logs Reconnecting after finishing schema recovery, the actual connection object is reused and not revalidated. If the MySQL server silently drops the connection during a long schema recovery (due to wait_timeout), this leads to a failure when setAutoCommit(false) is called.

I contributed a patch that:

1. Validates the JDBC connection after schema recovery,
2. And reconnects if the existing connection is no longer valid

### DIFF
--- a/debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorTask.java
+++ b/debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorTask.java
@@ -142,8 +142,12 @@ public class MariaDbConnectorTask extends BinlogSourceTask<MariaDbPartition, Mar
         validateAndLoadSchemaHistory(connectorConfig, connection::validateLogPosition, previousOffsets, schema, snapshotter);
 
         LOGGER.info("Reconnecting after finishing schema recovery");
+        
         try {
-            if (!connection.isConnected()) {
+            try {
+                connection.execute("SELECT 1");
+            }
+            catch (SQLException e) {
                 LOGGER.warn("Connection was dropped during schema recovery. Reconnecting...");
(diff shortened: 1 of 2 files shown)


debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorTask.java
```diff
@@ -143,10 +143,15 @@ protected ChangeEventSourceCoordinator<MariaDbPartition, MariaDbOffsetContext> s
 
         LOGGER.info("Reconnecting after finishing schema recovery");
         try {
+            if (!connection.isConnected()) {
```
jpechane, line 146: I don't think this is going to work. If you look at the source code it just checks if the underlying connection was created and not explicitly closed. It does not check if the connection is healthy. For that you'd need to exectue a test statement to verify the connection is still alive and working.

HarisBinSaif: What do you prefer? connection.isValid(someTimeout) or connection.execute("SELECT 1").

jpechane: I'd say `connection.execute` - in this case we cn be really sure of end-to-end testing of the connection.

jpechane:
@HarisBinSaif Applied, thanks!

PR #6296, opened 2025-04-04
DBZ-8860 Improve transaction/event cache performance

Naros:
https://issues.redhat.com/browse/DBZ-8860 

### Highlights

- ~~Infinispan uses asynchronous write operations for writing transaction/event data. Should see 2-3x speed increase, particularly for very large transactions.~~ Doing async in a follow-up PR just for 3.2.
- Savepoint/Constraint Violation handling with partial rollbacks is substantially faster.
- General cache performance should be faster than 3.0.8.Final / 3.1.0.Final

### Premise

For all implementations, there is some added benefit we gain in performance by introducing a shared contract for the storage of transactions and events. All three cache implementations make use of some level of heap-based storage for quick access to hotspot areas, such as `contains`, `size`, `isEmpty`, and event identifiers by its transaction.

For non-heap implementations, there is a single minimalistic heap-backed cache. This cache stores a map of the transaction's identifier mapped to all active event identifiers. For small and short-lived transactions, this map adds minimal GC overhead because the event identifiers are pulled from the `IntegerCache` for values 0 to 127 thanks to how Integer auto-boxing works.

For the heap implementation, I opted for the dual-map approach for events rather than a single map that uses an inner LinkedHashMap due to the reverse traversal needed. When we move to Java 21 for runtime, this can be optimized even more by switching to a single map using a LinkedHashMap and then relying on SequencedMap for the reverse traversal which has minimal overhead by comparison to doing this in Java 17.


Naros:
@jeremy-l-ford as you've reviewed a few of these in the past, I'd love your feedback if you have a moment.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerTransactionCache.java
```diff
@@ -0,0 +1,187 @@
...
+
+    @Override
+    public void addTransaction(EhcacheTransaction transaction) {
+        transactionCache.put(transaction.getTransactionId(), transaction);
+        eventIdsByTransactionId.put(transaction.getTransactionId(), new TreeSet<>());
+    }
+
+    @Override
+    public void removeTransaction(EhcacheTransaction transaction) {
+        transactionCache.remove(transaction.getTransactionId());
```
jeremy-l-ford, line 54: Should eventIdsByTransactionId.remove(transaction.getTransactionId()) be called here?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerTransactionCache.java
```diff
@@ -0,0 +1,187 @@
...
+
+    @Override
+    public void eventKeys(Consumer<Stream<String>> consumer) {
+        try (var stream = StreamSupport.stream(eventCache.spliterator(), false)) {
+            consumer.accept(stream.map(Cache.Entry::getKey));
+        }
+    }
+
+    @Override
+    public Iterator<LogMinerEvent> eventsIterator(EhcacheTransaction transaction) {
```
jeremy-l-ford, line 94: Can we switch this to a consumer based implementation instead of returning the Iterator?  Due to the infinispan impl, we were trying to control the closing of the stream.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerTransactionCache.java
```diff
@@ -0,0 +1,187 @@
...
+    @Override
+    public LogMinerEvent getTransactionEvent(EhcacheTransaction transaction, int eventKey) {
+        return eventCache.get(transaction.getEventId(eventKey));
+    }
+
+    @Override
+    public EhcacheTransaction getAndRemoveTransaction(String transactionId) {
+        final EhcacheTransaction transaction = getTransaction(transactionId);
+        if (transaction != null) {
+            transactionCache.remove(transactionId);
```
jeremy-l-ford, line 110: eventIdsByTransactionId.remove

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerTransactionCache.java
Naros: So I can change this to do that, but doing so means more changes in the _processor_ class.  There we actually do the event removal and the transaction removal in two steps.  I'll take a look and see that changes fall out from doing this.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerTransactionCache.java
Naros: That should be doable. This is only used in 2 places: Commit and Abandonment.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerTransactionCache.java
Naros: So this is a really poor side effect of how we manage the transaction/events during the commit handler.  We eject the transaction, process the events, and then eject the events.  I think ideally I'd prefer we not eject anything from the cache until the end, and I think if we do that when `removeTransaction` and `getAndRemoveTransaction` and simply cleaer the transaction & events.  :+1:

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerTransactionCache.java
Naros: @jeremy-l-ford so for the non-heap implementations, the iterator returned has no underlying resources that needs management fwiw. If I change this, I'll need to likely do this using a `Predicate` because I need to break the loop in the commit handler.  Any issue from your PoV using this style?

jeremy-l-ford: @Naros - using a predicate should be fine.  There's a removeIf(Predicate<Entry<K, V>> predicate) method that's part of the LogMinerCache that lines up with this.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/InfinispanLogMinerTransactionCache.java
```diff
@@ -0,0 +1,247 @@
...
+            }
+        }));
+    }
+
+    private void waitForAllWrites() {
+        for (CompletableFuture<?> future : pendingWrites) {
+            try {
+                future.get();
+            }
+            catch (Exception e) {
```
jeremy-l-ford, line 237: If this is an InterruptedException, seems like maybe the loop should stop.  Also, would want to call Thread.currentThread().interrupt again.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/InfinispanLogMinerTransactionCache.java
```diff
@@ -0,0 +1,247 @@
...
+
+    @Override
+    public void close() throws Exception {
+        // Makes sure during shutdown that all pending writes are flushed
+        waitForAllWrites();
+    }
+
+    private void primeEventCountsByTransactionId() {
+        // Primes the heap-based cache if the Infinispan disk caches contained data on start-up
+        transactions(stream -> stream.forEach(transaction -> {
```
jeremy-l-ford, line 219: Can we avoid streaming over the transactions by simply streaming overall events and then looking up the transaction for each event?  Unsure of performance implications here, but I'm thinking it might be better.

jeremy-l-ford: @Naros - but I assume this is at startup, so may not be a major concern.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/InfinispanLogMinerTransactionCache.java
```diff
@@ -0,0 +1,247 @@
...
+    }
+
+    @Override
+    public void syncTransaction(InfinispanTransaction transaction) {
+        // todo:
+        // Perhaps we can look at pulling number of events out of Transaction and let that
+        // be managed in the cache's heap, in which case we can avoid this put.
+
+        // Necessary to synchronize state
+        putAsync(transactionCache, transaction.getTransactionId(), transaction);
```
jeremy-l-ford, line 208: How synced does this method need to be?  Should waitForAllWrites be called?

jeremy-l-ford:
Apologies in advance, but my comments be come in bursts.  I plan on looking at this some more later.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerTransactionCache.java
```diff
@@ -0,0 +1,187 @@
...
+        return eventIds.descendingSet().stream()
+                .map(i -> Map.entry(i, transaction.getEventId(i)))
+                .filter(entry -> {
+                    final LogMinerEvent event = eventCache.get(entry.getValue());
+                    return event != null && event.getRowId().equals(rowId);
+                })
+                .findFirst()
+                .map(entry -> {
+                    eventCache.remove(entry.getValue());
+                    eventIds.remove(entry.getKey());
```
jeremy-l-ford, line 144: We're removing entries from the TreeSet while streaming over it.  Should copy the set and iterate over that.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerTransactionCache.java
```diff
@@ -0,0 +1,187 @@
...
+
+    @Override
+    public void addTransactionEvent(EhcacheTransaction transaction, int eventKey, LogMinerEvent event) {
+        eventCache.put(transaction.getEventId(eventKey), event);
+        eventIdsByTransactionId.get(transaction.getTransactionId()).add(eventKey);
+    }
+
+    @Override
+    public void removeTransactionEvents(EhcacheTransaction transaction) {
+        eventIdsByTransactionId.get(transaction.getTransactionId())
```
jeremy-l-ford, line 123: Wondering if it would be better to removeAll the events from the cache at once.

```
Set<String> eventKeys = eventIdsByTransactionId.get(transaction.getTransactionId())
                .descendingSet()
                .stream()
                .map(transaction::getEventId)
                .collect(Collectors.toSet());
        eventCache.removeAll(eventKeys);
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/InfinispanLogMinerTransactionCache.java
```diff
@@ -0,0 +1,247 @@
...
+
+        eventIdsByTransactionId.remove(transaction.getTransactionId());
+    }
+
+    @Override
+    public boolean removeTransactionEventWithRowId(InfinispanTransaction transaction, String rowId) {
+        waitForAllWrites();
+
+        final TreeSet<Integer> eventIds = eventIdsByTransactionId.get(transaction.getTransactionId());
+        return eventIds.descendingSet().stream()
```
jeremy-l-ford, line 155: copy treeset before streaming.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/InfinispanLogMinerTransactionCache.java
Naros: This is performed at startup. As for performance, you're probably right. Worse case, there would be 1 transaction lookup for each event effectively evicting each transaction at some point, but likely have far more events associated with a single transaction in practice :+1:

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/InfinispanLogMinerTransactionCache.java
Naros: I was testing some research earlier today and I believe I found where Infinispan makes no guarantee about the processing order of asynchronous operations on the same key with different values. 

We could consider changing `pendingWrites` to be `Map<String, Future<?>>` and we'd need to maintain one for each internal cache, and check if there is any outstanding write for the cache-key, delegating to `waitForAllWrites` before the following asynchronous write.  I'm just not sure what the overhead of that would ultimately be and what potential errors this could introduce. 

Perhaps the safe and less risky play here would be to make DBZ-8860 synchronous for now, since we want to backport this back into 3.0. And then we can work on a future enhancement separately for asynchronous support.  Would you agree?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerTransactionCache.java
Naros: I understand the concern but given this is in `findFirst()` and the stream breaks after the removal, I'm not sure its an issue necessarily and I'd prefer to avoid a copy if it's not necessary.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerTransactionCache.java
Naros: In this case, we can probably drop the `descendingSet()` too as its superfluous.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/InfinispanLogMinerTransactionCache.java
Naros: If we think this is an issue, then I'd prefer using an alternative that avoids the copy. This method in particular is one of the critical reasons for this issue, so it needs to be as fast and efficient as possible. For large transactions, a copy of the set is `O(n)` and if I have a large transaction with hundreds of thousands or millions of partial rollbacks, that cost will eat away at any performance gains we're trying to achieve I believe.

Naros: I think for these specific cases, I'll fall back to a traditional loop to avoid the stream concurrency concerns.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerTransactionCache.java
Naros: So I see why we do it the way we currently do. The idea is we rely on the transaction cache after we remove the transaction to calculate the smallest scn based on the remaining transactions. I suppose we could leave the transaction in and simply have that transaction excluded from the computation as an alternative :+1:

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/InfinispanLogMinerTransactionCache.java
jeremy-l-ford: Less risky seems best.  Can continue to improve later.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerTransactionCache.java
jeremy-l-ford: Agreed about findFirst there.  Was thinking it was part of the stream when reading it.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/infinispan/InfinispanLogMinerTransactionCache.java
jeremy-l-ford: I think as implement it's fine since findFirst results in an Optional and stream is no longer an issue.  The issue may just be readability in that at my first glance, I mistook it to be removing things from the stream while it was being processed.

Naros:
@jeremy-l-ford I think like the async discussion with Infinispan, the tweaks to consolidate the remove logic really only make sense for 3.2+. I'd prefer to avoid any more risk than what this PR already brings.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerTransactionCache.java
```diff
@@ -0,0 +1,189 @@
...
+
+import io.debezium.connector.oracle.logminer.events.LogMinerEvent;
+import io.debezium.connector.oracle.logminer.processor.AbstractLogMinerTransactionCache;
+
+/**
+ * A concrete implementation of {@link AbstractLogMinerTransactionCache} for Ehcache.
+ *
+ * @author Chris Cranford
+ */
+public class EhcacheLogMinerTransactionCache extends AbstractLogMinerTransactionCache<EhcacheTransaction> {
```
jeremy-l-ford, line 29: Should the ehcache impl have its own primeEventCountsByTransactionId method?

jeremy-l-ford:
Other than the priming on ehcache, thinking everything is looking good.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/ehcache/EhcacheLogMinerEventProcessor.java
```diff
@@ -106,7 +106,22 @@ public LogMinerCache<String, String> getProcessedTransactionsCache() {
 
     @Override
     public void close() throws Exception {
+        transactionCache.close();
+
         if (cacheManager != null) {
```
jeremy-l-ford, line 111: Not sure if there should be a try/finally around any of this.

Naros: I can actually remove that call to `close` as it was used to synchronize Infinispan writes before shutdown, but since I reverted back to synchronous cache operations, I can remove this call.

vjuranek (commented):
Overall LGTM, concurrent put in ISPN cache was removed in later commits, so the only questions I have if we can use hash maps in in-memory cache or we should use concurrent hash maps instead.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/memory/MemoryLogMinerTransactionCache.java
```diff
@@ -0,0 +1,163 @@
...
+ * A concrete implementation of the {@link LogMinerTransactionCache} that stores transactions and events
+ * in the JVM heap for fast access and management.
+ *
+ * @author Chris Cranford
+ */
+public class MemoryLogMinerTransactionCache extends AbstractLogMinerTransactionCache<MemoryTransaction> {
+
+    private final Map<String, MemoryTransaction> transactionsByTransactionId = new HashMap<>();
+    private final Map<String, List<LogMinerEventEntry>> eventsByTransactionId = new HashMap<>();
+    private final Map<String, HashMap<Integer, LogMinerEvent>> eventsByEventIdByTransactionId = new HashMap<>();
```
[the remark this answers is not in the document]
Naros: Yes, that's a safe assumption as there is no multi-threaded access or handling during streaming.

danny-johnson-nisc:
I ran more rollback testing on this branch and saw great improvements
the test was rolling back a 100k row transaction and seeing how long it took to process it. I also confirmed after that the NumberOfPartialRollbackCount was at 100k
here are the results:

image

I noticed that the time between MEMORY and EHCACHE are much closer than before and also shorter. ehcache went from 30min ->1 min and MEMORY went from 5 min -> 1 min

one secondary observation was that the async engine seemed to take slightly longer and use more CPU during the processing. Is this expected?

Naros:
Those are great stats @danny-johnson-nisc, did you get a chance to look at ISPN as well?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/processor/memory/MemoryLogMinerTransactionCache.java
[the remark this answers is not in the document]
vjuranek: thanks, this is what I think but wanted to double check with you as you understand the code much better than me

vjuranek (approved):

vjuranek:
> one secondary observation was that the async engine seemed to take slightly longer and use more CPU during the processing. Is this expected? edit: confirmed the CPU increase was also on 3.0.8 for async, so not related to this PR

@danny-johnson-nisc that's interesting. In short: no, this is not expected. Depending on your test setup I can imagine test can run slightly longer on anync engine as it may wait for all threads to finish (assuming you measure from the engine start until it's stopped), but what is definitely not expected is higher CPU load. The connector is single threaded so the CPU load cannot be explained by parallel execution (and in such case one would expect that execution time would be shorter) and I don't see any reason or explanation why the engine itself should consume more CPU as it just starts the connector task and waits for the task to finish. Do you have any insight into it or can you please provide e.g. flame graphs for the normal and async execution? Or provide the source code for the tests so I can try reproduce it? Thanks!

Naros:
Hi @danny-johnson-nisc any chance the CPU spikes on 3.0.8.Final was related to what this PR is fixing? You could try comparing 3.0.7.Final and see if your test observed the same CPU increase?  If not, the CPU increase may be related to DBZ-8665.

Naros:
Rebased and simply added a new TMT matrix entry for ehcache.

jeremy-l-ford (approved):

danny-johnson-nisc:
@Naros
1) ISPN did show massive improvement
2) the 3.0.7 async test on MEMORY had even higher CPU than 3.2.0, so im thinking an improvement was made in 3.0.8 and it was about the same CPU usage from 3.0.8 to this branch. the times with (est.) are estimated, I waiting until 10k partial rollback and then multiplied the time by 10 so the tests didnt take all day

image

also disclaimer, the CPU usage numbers are approximate, im just using the jconsole CPU graph and giving a rough range. but I will say objectively and certainly that the async engine is definitely using more CPU than embedded engine

@vjuranek I will try and do some profiling today to see if I can find any more details for you on the async stuff

Naros:
@jpechane is this good from your PoV?

jpechane (commented):
@Naros LGTM, thanks!
Please feel free to merge yourself. Give that this is backported to 3.0 I'll leave to to you in case there are any conflicts.

jpechane (approved):

PR #6297, opened 2025-04-04
DBZ-8687 Edits `slot.failover` description, adds TP note.

roldanbob:

Followup to #6225 to add TP note to the description of the `slot.failover` property.

Please backport this change to 3.0.


roldanbob:
Not sure why this check is failing. Is it because I reused the issue #? 

> Hi @roldanbob, thanks for your contribution. Please prefix the commit message(s) with the [DBZ-xxx JIRA issue key](https://github.com/debezium/debezium/blob/main/CONTRIBUTE.md#making-changes).

Naros:
It's okay @roldanbob, sometimes when the GH API fails or the parser gets a bad response, it fails to recognize the data and it will fail the job when ti shouldn't. I reran the job and it passed :+1:

Naros (approved):

PR #6299, opened 2025-04-04
DBZ-8576: Incorrect NumberOfEventsFiltered metrics in streaming

adasari:
https://issues.redhat.com/browse/DBZ-8576

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/ReplicationMessage.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/ReplicationMessage.java
@@ -46,8 +46,7 @@ public interface ReplicationMessage {
         TRUNCATE,
         MESSAGE,
         BEGIN,
-        COMMIT,
-        NOOP
+        COMMIT
     }
 
     /**
@@ -186,6 +185,17 @@ public interface ReplicationMessage {
         return getOperation() == Operation.BEGIN || getOperation() == Operation.COMMIT;
     }
 
+    default boolean isFiltered() {
+        return false;
+    }
+
+    /**
+     * Returns true if the message is a skipped message.
+     */
+    default boolean isSkippedMessage() {
+        return false;
+    }
+
     /**
      * A special message type that is used to replace event filtered already at {@link MessageDecoder}.
      * Enables {@link PostgresStreamingChangeEventSource} to advance LSN forward even in case of such messages.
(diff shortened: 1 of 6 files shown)


adasari:
@vjuranek @jpechane I have retained NoopMessage but could't retain the NOOP operation as its corresponding mapping in global message is not available. Could you review the approach ?

debezium-core/src/main/java/io/debezium/pipeline/spi/ChangeRecordEmitter.java
```diff
@@ -40,6 +40,10 @@ public interface ChangeRecordEmitter<P extends Partition> {
      */
     Operation getOperation();
 
+    default boolean shouldIgnoreRecords() {
```
vjuranek, line 43: I'd suggest to use `shouldIgnoreRecord()` (singular `Record`) or just `ignoreRecord()`.
Also please add javadoc to this method.

vjuranek (commented):
One minor thing, but the approach LGTM.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/ReplicationMessage.java
```diff
@@ -186,6 +185,17 @@ default boolean isTransactionalMessage() {
...
     }
 
+    default boolean isFiltered() {
+        return false;
+    }
+
+    /**
+     * Returns true if the message is a skipped message.
+     */
+    default boolean isSkippedMessage() {
```
jpechane, line 195: What is the difference between these two messages?

adasari: My bad. `isFiltered` is a redundant and unused method. will remove it.

debezium-core/src/main/java/io/debezium/pipeline/spi/ChangeRecordEmitter.java
adasari: According to class doc https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/pipeline/spi/ChangeRecordEmitter.java#L15, it emits one or more records and mostly one record.
Let me know if i can go ahead update the method name to `ignoreRecord()`

vjuranek (approved):

jpechane:
@adasari Nice, applied!

PR #6302, opened 2025-04-05
Dbz 8840 doc update to sqlserver schema history mbean name

roldanbob:
[DBZ-8840](https://issues.redhat.com/browse/DBZ-8840)

Refactors the _Schema history metrics_ topic in the Db2, MariaDB, MySQL, Oracle, and SQL Server connector files  to correct the schema history MBean name for SQL Server to include the `task.id`. In the original version of the topics, a common MBean name was shared to all of the relevant connectors via an entry in a shared file  `ref-connector-monitoring-schema-history-metrics.adoc`.

To assign a unique MBean name for SQL Server, it was necessary to remove the single MBean name from the beginning of the shared file, and create two separate name entries in the fragment file {{/partials/frag-common-mbean-name.adoc}} that is already used to provide instances of other MBean names. The single MBean name is now replaced by one of the following: 

- A common MBean name for use in the Db2, MariaDB, MySQL, and Oracle connector Schema history topics
- A SQL Server-specific schema history MBean names 

Each of these entries in fragment file is implemented as a tagged region, which can be uniquely referenced via an `include` directive.

In the _Schema history metrics_ topic for each connector file, I added a second `include` directive to import just the MBean name for the connector (`common-schema-history` or `sqlserver-schema-history`) on the line that precedes the existing directive. The original `include` directive now provides just the metrics table. 

This approach should facilitate subsequent global changes to the naming format should that become necessary. 

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/db2.adoc
+++ b/documentation/modules/ROOT/pages/connectors/db2.adoc
@@ -3219,6 +3219,7 @@ include::{partialsdir}/modules/all-connectors/ref-connector-monitoring-streaming
 [[db2-schema-history-metrics]]
 === Schema history metrics
 
+include::{partialsdir}/modules/all-connectors/frag-common-mbean-name.adoc[leveloffset=+1,tags=common-schema-history]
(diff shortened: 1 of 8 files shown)


Naros (approved):
LGTM @roldanbob thanks!

PR #6303, opened 2025-04-06
DBZ-8882 Prefix Oracle infinispan profiles with `oracle-`

Naros:
https://issues.redhat.com/browse/DBZ-8882

Prefix the Infinispan-specific profiles to match the other buffer profiles. I've also added a new job on the testing farm to test the Oracle connector with embedded ISPN using Oracle 19.


jpechane:
@Naros Applied, thanks

PR #6305, opened 2025-04-07
DBZ-8885: Add an option to reset the lagBehindSource metric separately

ashoke-cube:
https://issues.redhat.com/browse/DBZ-8885

During idle periods at the source, when no records are received by the connector, the `lagBehindSource` metric retains its old value, incorrectly indicating that the connector is lagging behind. A mechanism is needed to reset the `lagBehindSource` metric in such cases.  

Currently, there is no way to reset this metric individually; only a group reset is available, which resets all metrics in `StreamingMeter` together. Additionally, `StreamingMeter` is a private variable in `DefaultStreamingChangeEventSourceMetrics`, making it inaccessible when extending this class. 

This PR provides a getter method for StreamingMeter object and provides a method to reset the `lagBehindSource` separately.

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/meters/StreamingMeter.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/meters/StreamingMeter.java
@@ -66,6 +66,11 @@ public class StreamingMeter implements StreamingMetricsMXBean {
         return lastTransactionId.get();
     }
 
+    @Override
+    public void resetLagBehindSource() {
+        lagBehindSource.set(null);
+    }
+
     public void onEvent(DataCollectionId source, OffsetContext offset, Object key, Struct value) {
         final Instant eventTimestamp = metadataProvider.getEventTimestamp(source, offset, key, value);
         if (eventTimestamp != null) {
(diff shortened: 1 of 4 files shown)


debezium-core/src/main/java/io/debezium/pipeline/meters/StreamingMeter.java
```diff
@@ -86,6 +86,10 @@ public void onEvent(DataCollectionId source, OffsetContext offset, Object key, S
         }
     }
 
+    public void resetLagBehindSource() {
```
jpechane, line 89: IMHO the method needs to be declared at `io.debezium.pipeline.metrics.ChangeEventSourceMetricsMXBean` to be visible in JMX or another one and implemented here or elswhere.

jpechane (changes requested):

debezium-core/src/main/java/io/debezium/pipeline/meters/StreamingMeter.java
jpechane: Right, that is probably better place. I wanted to place it sid-by-side with the global reset but this is better solution.

vjuranek (approved):

vjuranek (approved):

jpechane:
@ashoke-cube Thanks, it LGTM! Could you please remove the merge commit and squash the rest into one so the PR is kept clean for rebase?

jpechane:
@ashoke-cube Applied, thanks!

PR #6313, opened 2025-04-08
DBZ-8879: Fix performance regression in debezium-core

b-goyal:
This PR addresses the performance regression introduced as part of #6222.
JIRA: https://issues.redhat.com/browse/DBZ-8879
Zulip Chat: https://debezium.zulipchat.com/#narrow/channel/302529-community-general/topic/Performance.20Regression.20after.203.2E1.2E0.2EBeta1

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
@@ -6,6 +6,7 @@
 package io.debezium.pipeline;
 
 import static io.debezium.config.CommonConnectorConfig.WatermarkStrategy.INSERT_DELETE;
+import static io.debezium.util.Loggings.maybeRedactSensitiveData;
 
 import java.time.Instant;
 import java.util.Collection;
@@ -57,7 +58,6 @@ import io.debezium.schema.SchemaFactory;
 import io.debezium.schema.SchemaNameAdjuster;
 import io.debezium.spi.schema.DataCollectionId;
 import io.debezium.spi.topic.TopicNamingStrategy;
-import io.debezium.util.Loggings;
 
 /**
  * Central dispatcher for data change and schema change events. The former will be routed to the change event queue, the
@@ -227,8 +227,8 @@ public class EventDispatcher<P extends Partition, T extends DataCollectionId> im
                                          OffsetContext offset,
                                          ConnectHeaders headers)
                         throws InterruptedException {
-                    Loggings.logTraceAndTraceRecord(LOGGER, new Object[]{ "Key:", key, ", Value:", value }, "Received change record for {} operation with context {}",
-                            " ", operation, offset);
+                    LOGGER.trace("Received change record {} for {} operation on key {} with context {}", maybeRedactSensitiveData(value), operation,
+                            maybeRedactSensitiveData(key), offset);
 
                     eventListener.onEvent(partition, dataCollectionSchema.id(), offset, key, value, operation);
                     receiver.changeRecord(partition, dataCollectionSchema, operation, key, value, offset, headers);
@@ -289,8 +289,8 @@ public class EventDispatcher<P extends Partition, T extends DataCollectionId> im
                                              ConnectHeaders headers)
                             throws InterruptedException {
 
-                        Loggings.logTraceAndTraceRecord(LOGGER, new Object[]{ "Key:", key, ", Value:", value }, "Received change record for {} operation with context {}",
-                                " ", operation, offset);
+                        LOGGER.trace("Received change record {} for {} operation on key {} with "
+                                + "context {}", maybeRedactSensitiveData(value), operation, maybeRedactSensitiveData(key), offset);
 
                         if (isASignalEventToProcess(dataCollectionId, operation) && sourceSignalChannel != null) {
                             sourceSignalChannel.process(value);
@@ -510,8 +510,8 @@ public class EventDispatcher<P extends Partition, T extends DataCollectionId> im
 
             Objects.requireNonNull(value, "value must not be null");
 
-            Loggings.logTraceAndTraceRecord(LOGGER, new Object[]{ "Key:", key, ", Value:", value }, "Received change record for {} operation with context {}", " ",
-                    operation, offsetContext);
+            LOGGER.trace("Received change record {} for {} operation on key {} with context {}", maybeRedactSensitiveData(value), operation,
+                    maybeRedactSensitiveData(key), offsetContext);
 
             // Truncate events must have null key schema as they are sent to table topics without keys
             Schema keySchema = (key == null && operation == Operation.TRUNCATE) ? null
@@ -566,7 +566,7 @@ public class EventDispatcher<P extends Partition, T extends DataCollectionId> im
                 throws InterruptedException {
             Objects.requireNonNull(value, "value must not be null");
 
-            Loggings.logTraceAndTraceRecord(LOGGER, key, "Received change record for {} operation", operation);
+            LOGGER.trace("Received change record for {} operation on key {}", operation, maybeRedactSensitiveData(key));
 
             doPostProcessing(key, value);
 
@@ -648,7 +648,7 @@ public class EventDispatcher<P extends Partition, T extends DataCollectionId> im
                 throws InterruptedException {
             Objects.requireNonNull(value, "value must not be null");
 
-            Loggings.logTraceAndTraceRecord(LOGGER, key, "Received change record for {} operation", operation);
+            LOGGER.trace("Received change record for {} operation on key {}", operation, maybeRedactSensitiveData(key));
 
             Schema keySchema = dataCollectionSchema.keySchema();
             String topicName = topicNamingStrategy.dataChangeTopic((T) dataCollectionSchema.id());
--- a/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
@@ -227,8 +227,8 @@ public class EventDispatcher<P extends Partition, T extends DataCollectionId> im
                                          OffsetContext offset,
                                          ConnectHeaders headers)
                         throws InterruptedException {
-                    Loggings.logTraceAndTraceRecord(LOGGER, "Key: " + key + ", Value: " + value, "Received change record for {} operation with context {}", operation,
-                            offset);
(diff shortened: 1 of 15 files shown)


b-goyal:
@jpechane , I only found a single instance in `TableSchema` class. The other two introduced in original [PR](https://github.com/debezium/debezium/pull/6222/files#diff-da2269420d1d6278045f92b0bd57bf42b85969251c31c365c1634fa19640492f) in `HeaderToValue.java` are unavailable in latest `main`. Let me know if find any instance still uncovered in the PR. Thanks!

twthorn:
Some comments from [Zulip](https://debezium.zulipchat.com/#narrow/channel/302529-community-general/topic/Performance.20Regression.20after.203.2E1.2E0.2EBeta1/with/510974743) discussion including here
> One thing I'd like to add: Having the word "And" in a function name is a code smell. It indicates that Single Responsibility Principle is being broken. Several books (eg "Clean Code: A Handbook of Agile Software Craftsmanship") and articles discuss this (eg [this blog post](https://medium.com/@samanthaming/code-smell-no-and-in-function-name-c83d0fe5d46e#:~:text=in%20Function%20name-,Samantha%20Ming,is%20more%20flexible%20and%20modular.))
> From what I can tell, the regression was caused by trying to do two things in one method. Looking at the code:
```
    public static void logTraceAndTraceRecord(Logger logger, Object record, String message,
                                              Object... arguments) {
        logger.trace(message, arguments);
        LOGGER.trace("Source of trace is record '{}'", record);
    }
```
> Since we can only have one variable length argument Object... arguments we were eagerly concentrating a String message with its variable length args as a hacky solution to get around this. Each call to logger should have its own variable length args parameter. But this is impossible in a function that is making two log calls ie breaking SRP.

jpechane:
@twthorn As we are touching this code I am open to any solution. Is there an alternative aprocha woud envision? If yes, could you please add an example into the comment? I'd like to resolved the issue for good and when done we are going to do 3.1.1.Final right after.

twthorn:
The simplest way to fix SRP is to not be logging on behalf of another class, ie we change from
```
class Loggings {
    [...]
    public static void logTraceAndTraceRecord(Logger logger, Object record, String message,
                                              Object... arguments) {
        logger.trace(message, arguments);
        LOGGER.trace("Source of trace is record '{}'", record);
    }
   [...]
}
```
to
```
class Loggings {
    [...]
    public static void traceRecord(Object record) {
        LOGGER.trace("Source of trace is record '{}'", record);
    }
   [...]
}
```

So for callers, before it was
```
Loggings.logTraceAndTraceRecord(LOGGER, "Key: " + key + ", Value: " + value, "Received change record for {} operation with context {}", operation, offset);
```
Would now be two lines
```
LOGGER.trace("Received change record {} for {} operation on key {} with context {}", value, operation, key, offset);
Loggings.traceRecord(record)
```

This accomplishes the goal of all trace records with PII are logged by one class.

I understand the motivation of using "And" because it gives the appearance of DRY code. It assumes that whenever we log this message, we also need to log the record at trace level. This is tight coupling, which produces less duplication, but it reduces the modularity of the code. For example, what if we ever want to trace the record but we don't want to log a message? We need even more code to support that lack of modularity. Also it doesn't really produce completely DRY code because the Loggings class has many repetitive methods each for different log levels (eg logWarningAndTraceRecord, logDebugAndTraceRecord, etc.)

One other thing I notice, there are calls that are checking the log level before logging
```
        if (LOGGER.isTraceEnabled()) {
            Loggings.logTraceAndTraceRecord(LOGGER, columnData, "columnData from current stack", ", ");
        }
```
This is another bad practice. That is the first line of any log implementation to check what the log level is, we should not be repeating it here, that is not DRY. This would only be necessary when the next line is doing something inefficient like eager toString calls, which we should refactor to avoid.

If we *must* preserve this format of `Source of <level> is record`, we can refactor Loggings class to be a lot shorter:
```
public class Loggings {
    private static final Logger LOGGER = LoggerFactory.getLogger(Loggings.class);

    public static void traceRecord(String message, Level level, Object record) {
        LOGGER.trace("Source of {} is record '{}'", level, record);
    }
```

With multi-threaded log interleaving I am skeptical of how useful it is. If we can remove that level part it could just be
```
public class Loggings {
    private static final Logger LOGGER = LoggerFactory.getLogger(Loggings.class);

    public static void traceRecord(String message, Object record) {
        LOGGER.trace("Record '{}'", record);
    }
```
If we actually would like to make it extensible (ie we can support new message formats for tracing records) and modular we do the below and add an overload the method for all the callers that just want a standard/simple format. I would wager there are places in the code where the message with the trace record it may make sense to improve now that we have this flexibility (i.e., always saying "Source of `<level>` record" it may be unclear which `<level>` line we refer to, the operator needs to find that line and read that line to then interpret this record, due to log interleaving, instead a caller could say "Source of X problem is this record" with the record so all needed debug info is contained in one line).
```
public class Loggings {
    private static final Logger LOGGER = LoggerFactory.getLogger(Loggings.class);

    public static void traceRecord(String message, Object... arguments) {
        LOGGER.trace(message, arguments);
    }

    public static void traceRecord(Object record) {
        traceRecord("Record '{}'", record);
    }
```

debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
```diff
@@ -227,8 +227,8 @@ public void changeRecord(P partition,
                                          OffsetContext offset,
                                          ConnectHeaders headers)
                         throws InterruptedException {
-                    Loggings.logTraceAndTraceRecord(LOGGER, "Key: " + key + ", Value: " + value, "Received change record for {} operation with context {}", operation,
-                            offset);
+                    Loggings.logTraceAndTraceRecord(LOGGER, new Object[]{ "Key:", key, ", Value:", value }, "Received change record for {} operation with context {}",
```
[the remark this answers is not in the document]
Naros: But after reading Tom's comments below, I like the idea of pulling the dual logging up to the call site and then letting `Loggings` be called solely to log the sensitive parts with a common logger. Lets see if Jiri has any opinion one way or the other.

jpechane:
@twthorn To be honest I am not yet sold on idea of two calls. And your point with multithreading is definitely also another concern. I wonder then whether the solution could be using `MDC` and set the source record as par of context. Or the other option to have decoratior implementation of `Logger` class that would optionally add source record to messages.

twthorn:
For MDC, it seems this is used when not all information is available at the moment of the log call. In our case all info is available. I am not 100% clear on what you have in mind for decorator, can you share some code snippets?

Here's another idea:
```
public class PIIRedactor {

    private static final boolean PII_LOGGING_ENABLED;

    static {
        PII_LOGGING_ENABLED = <read from log4j.properties "pii.logging.enabled" so user can set in logging properties>
    }

    public static Object redactIfDisabled(Object piiValue) {
        return PII_LOGGING_ENABLED ? piiValue : "[REDACTED]";
    }
}
```

And in every call we do 
```
logger.info("Doing {} operation with other arg1 {} arg2 {} values for record {}",
    operation, arg1, arg2, redactIfDisabled(R));
```

Open to other ideas.

jpechane:
@twthorn I think we have a winner here. My intent was to have something like
```
public class PiiLogger implements Logger {
    private final Logger logger;
    private final piiEnabled;

    public PiiLogger(Logger logger) {
        this.logger = logger;
        piiEnabled = <some logic to determine if PII logging is enabled>;
    }

    // implement Logger methods using delegate logger

    public Object sensitiveValue(Object value) {
        return piiEnabled ? value : "<sensitive>";
    }
}
```

But it is probably overkill in comparison with the static utility class you proposed.

jpechane:
@b-goyal Hi, could you please update PR to be inline with https://github.com/debezium/debezium/pull/6313#issuecomment-2794515567 ? I'd like to do 3.1.1.Final release next week to get rid of that perf issue.
Thanks

jpechane (approved):
@b-goyal LGTM, thanks!

@twthorn Any additional concerns?

debezium-core/src/main/java/io/debezium/relational/TableSchema.java
```diff
@@ -126,7 +127,7 @@ public Envelope getEnvelopeSchema() {
      */
     public Struct keyFromColumnData(Object[] columnData) {
         if (LOGGER.isTraceEnabled()) {
```
twthorn, line 129: Any conditional checks of log level are redundant. You can use a lambda (e.g., directly or with [addArgument](https://slf4j.org/manual.html#fluent)) or you can wrap columnData in an encapsulating object to define a custom toString method. The latter is perhaps a refactor for another PR fyi @jpechane

debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -110,7 +110,7 @@ public AbstractIncrementalSnapshotChangeEventSource(RelationalDatabaseConnectorC
     @SuppressWarnings("unchecked")
     public void closeWindow(P partition, String id, OffsetContext offsetContext) throws InterruptedException {
         context = (IncrementalSnapshotContext<T>) offsetContext.getIncrementalSnapshotContext();
-        Loggings.logTraceAndTraceRecord(LOGGER, context.toString(), "Closing Window");
+        LOGGER.trace("Closing Window {}", maybeRedactSensitiveData(context.toString()));
```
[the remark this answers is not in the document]
b-goyal: Removed the toString call from here.

debezium-core/src/main/java/io/debezium/relational/TableSchema.java
b-goyal: The conditional checks and the other toString methods have existed before the change introduced in the PR. I would prefer to keep refactoring of those separate to avoid introducing any new regressions tbh.

Naros: I would say those changes don't necessarily need to be in a different PR, but they certainly could be in a separate commit from the "reverting" of the performance regression.

For this use case specifically, what about taking the most simple path introducing a method in the `Loggings` class that takes an `Object[]` that internally uses `Arrays.toString` on the argument if TRACE is enabled. It would continue to have negligible overhead and avoids the redundant check.

jpechane:
@b-goyal Applied, thanks!

rajdangwal:
@jpechane Just an FYI, this should be back-ported to 3.0 as well as we have DBZ-8525 in 3.0 too. Thanks.

jpechane:
@rajdangwal Thanks for the reminder. Could you please prepare 3.0 based PR?

PR #6317, opened 2025-04-09
DBZ-8895 Add Oracle virtual column `BY USER FOR STATISTICS` clause

Naros:
https://issues.redhat.com/browse/DBZ-8895

This is an undocumented Oracle clause, so I don't intend to push this upstream.


jpechane:
@Naros Applied, thanks

PR #6328, opened 2025-04-10
[docs] Clarifying default behavior of "delete.tombstone.handling.mode"

gunnarmorling:
The description says that `tombstone` is the default, but "Default" column says there is no default. Assuming the description is correct, I've updated the latter.

### DIFF
--- a/documentation/modules/ROOT/pages/transformations/event-flattening.adoc
+++ b/documentation/modules/ROOT/pages/transformations/event-flattening.adoc
@@ -281,7 +281,7 @@ In its place, use the xref:extract-new-record-state-delete-tombstone-handling-mo
 ====
 
 |[[extract-new-record-state-delete-tombstone-handling-mode]]xref:extract-new-record-state-delete-tombstone-handling-mode[`delete.tombstone.handling.mode`]
-|No default
+|`tombstone`
 |{prodname} generates a change event record for each `DELETE` operation. This setting determines how the event flattening SMT handles `DELETE` events from the stream.
 
 [NOTE]


mfvitale (commented):
Well, looking at the code seems that effectively there isn't a default value set. 
In that case if not set, it fallback to the deprecated `drop.tombstones` that as `true` as default. 

Honestly dunno if this was intention or not but maybe we should as add the default for `delete.tombstone.handling.mode` given that also in the `Field` description is mentioned that `tombstone` is the default.

gunnarmorling:
So IIUC, the default (when not specifying `delete.tombstone.handling.mode` at all) is a behavior, which actually is not represented by any of the values of that option? If so, this is confusing :) IMO, the sensible default would be to

* drop the delete events
* retain tombstone events

Which I think is what happens when setting the option to `tombstone` (docs aren't clearly unfortunately what happens to the delete event in that case), but it seems that this actually is not the default behavior right now?

gunnarmorling:
Ugh, seems there's an actual problem here, apart from docs. When setting `transforms.unwrap.delete.tombstone.handling.mode` to `tombstone`, it doesn't actually emit a tombstone record (i.e. a record without value), but a record whose value is a String `"null"`. This defeats the purpose of tombstone records, i.e. Kafka wouldn't compact that key, also it causes downstream serialization issues.

jpechane:
@gunnarmorling Are you sure? See io.debezium.transforms.ExtractNewRecordStateTest.testTruncateDroppedByDefault()
This is IMHO just icorrect/imprecise docs

jpechane:
@mfvitale This is mostly correct due to handling of legacy props. I propose to get rid of them now and just set the default value at `HANDLE_TOMBSTONE_DELETES` level. The current docs fix IMHO carries the meaning of the code.

gunnarmorling:
> Are you sure?

Yeah, kinda. When setting the option, I am receiving "null" strings in the output, which e.g. cause a JSON deserializer (in my case in Flink) to fail:

```
java.io.IOException: Failed to deserialize JSON 'null'.
```

This is not the case when I set `"transforms.unwrap.drop.tombstones" : "false"` instead. Then I'm receiving proper tombstones (i.e. `null` values). That said, I'm also not seeing in the code where that incorrect behavior could be coming from.

mfvitale:
> I propose to get rid of them

If you are referring to the deprecated options, I'm totally with you. This SMT has been re-worked a lot of times and having also the old config is not helping to have it working as expected. So first step should be to get rid of the old configurations and review the behavior of the current one. IIRC we have some work around "handle complex nested documents and arrays" so maybe it is better to clean first.

jpechane:
@mfvitale https://github.com/debezium/debezium/pull/6371

PR #6331, opened 2025-04-10
DBZ-8723 Update docs to mandate all column names of signal table

ani-sha:
https://issues.redhat.com/browse/DBZ-8723

### DIFF
--- a/documentation/modules/ROOT/pages/configuration/signalling.adoc
+++ b/documentation/modules/ROOT/pages/configuration/signalling.adoc
@@ -134,8 +134,8 @@ Each signal type requires a specific set of data.
 
 |===
 
-NOTE: The field names in a data collection are arbitrary.
-The preceding table provides suggested names.
+NOTE: The field names in a data collection are arbitrary, except for the first field, `id`, which is mandatory. This field name must be used to ensure the connector can correctly process the signal.
+The preceding table provides suggested names for the remaining fields, but you can use different names if you prefer.
 If you use a different naming convention, ensure that the values in each field are consistent with the expected content.
 
 // Type: procedure
--- a/documentation/modules/ROOT/pages/configuration/signalling.adoc
+++ b/documentation/modules/ROOT/pages/configuration/signalling.adoc
@@ -134,9 +134,9 @@ Each signal type requires a specific set of data.
 
 |===
 
-NOTE: The field names in a data collection are arbitrary, except for the first field, `id`, which is mandatory. This field name must be used to ensure the connector can correctly process the signal.
-The preceding table provides suggested names for the remaining fields, but you can use different names if you prefer.
-If you use a different naming convention, ensure that the values in each field are consistent with the expected content.
+NOTE: The signaling data collection must include columns with the names,  `id`, `type`, and `data`.
+Do not include quotation marks in the names.
+If you assign alternate names to these columns, the connector cannot process signals.
 
 // Type: procedure
 // Title: Creating a {prodname} signaling data collection


documentation/modules/ROOT/pages/configuration/signalling.adoc
```diff
@@ -134,8 +134,8 @@ Each signal type requires a specific set of data.
 
 |===
 
-NOTE: The field names in a data collection are arbitrary.
-The preceding table provides suggested names.
+NOTE: The field names in a data collection are arbitrary, except for the first field, `id`, which is mandatory. This field name must be used to ensure the connector can correctly process the signal.
```
roldanbob, line 137:
```suggestion
NOTE: The labels assigned to the `type` and `data` fields in a data collection are arbitrary.
However, the collection must include a field with the label `id`.
If this field is not present, the connector cannot process signals. 
```
Just to be sure that I correctly interpret what we're changing with this update, is it correct that as long as the structure of the collection provides for fields that contain `type` and `data` values, it's not required to use those actual labels? But, by contrast, the primary key label `id` must be used. 
   
I vaguely recall that when we first added this content, we were saying that as long as the tri-part structure is maintained (`id`/`type`/`data`), then the user can assign any labels to those fields. So the difference now is that we've since discovered that errors result if there is no field with the label `id`. Is that approximately correct?

Naros: To your latter question, yes. The issue is that when users use the delete watermarking strategy, the SQL used to clean up the signal table expects a column named `id`. If there is no `id` column in the table, the delete watermark strategy throws an error.

However, regarding the earlier question, I don't think the use of `labels` is the correct term in this context. We're speaking about a relational table where the first column must be defined as `id`. Unless there is a reason not to refer to this as a column, that's the most apt word choice. I would go one step beyond and clarify that the column name should also be given without quoting to avoid case sensitivity being an issue for someone who may mistakenly choose to quote column names.

roldanbob: > column name should also be given without quoting to avoid case sensitivity being an issue for someone who may mistakenly choose to quote column names.

So, do you mean that when setting up the table, you should never assign column names that are enclosed in quotation marks?

ani-sha: That's right @roldanbob. Thanks, @Naros for clarifying this.

ani-sha: @roldanbob I too think column/field names make more sense rather than labels.

Naros: > you should never assign column names that are enclosed in quotation marks?

Correct. If you do quote column names, the database engine treats the column as having a case-sensitive name and the `DeleteWindowCloser` does not quote the id column, so that could create an error that the id column does not exist in the table.

roldanbob: Thanks, both of you for the additional information. I revised my earlier comment.

@Naros My edit assumes that your comment about the prohibition on quotation marks in the column names applies only to the `id` column. Yes?

Naros: From a pure technical point of view, that's correct @roldanbob

james-johnston-thumbtack: My outsider two cents (from the person who was affected by this issue): would it make sense to use this opportunity to also pre-emptively mandate the names for `type` and `data` columns?  This way, in the future, if there ends up being a hard-coded dependency on those column names as well, it will be reasonable to do it since the documentation would have always said users need to use those names.

Otherwise, the same scenario that has happened with `id` column could also happen again with those other columns as well: Debezium initially lets users use whatever name they want, then Debezium makes a breaking change by hard-coding a column name rather than dynamically determine the correct column name to use.

jpechane: Well, that sounds reasonable. We can declare it in docs only and no change in the code is needed. We'd jus add it to breaking changes yet in fact nothing will change for the current installations.
I originally wanted to be really flexible but it seems that from the future development point of view it might be liability.

@mfvitale @Naros Any objections?

mfvitale: No objections from my side. This was exactly what I was thinking initially. I don't think in this case the flexibility give much advantages to the users.

Naros: No objection from my point of view either @jpechane

documentation/modules/ROOT/pages/configuration/signalling.adoc
```diff
@@ -134,8 +134,8 @@ Each signal type requires a specific set of data.
...
+The preceding table provides suggested names for the remaining fields, but you can use different names if you prefer.
 If you use a different naming convention, ensure that the values in each field are consistent with the expected content.
```
roldanbob, line 139: Remove this line.
```suggestion

```

documentation/modules/ROOT/pages/configuration/signalling.adoc
```diff
@@ -134,8 +134,8 @@ Each signal type requires a specific set of data.
 
 |===
 
-NOTE: The field names in a data collection are arbitrary.
-The preceding table provides suggested names.
+NOTE: The field names in a data collection are arbitrary, except for the first field, `id`, which is mandatory. This field name must be used to ensure the connector can correctly process the signal.
+The preceding table provides suggested names for the remaining fields, but you can use different names if you prefer.
```
roldanbob, line 138: Remove this line.
```suggestion

```

documentation/modules/ROOT/pages/configuration/signalling.adoc
roldanbob: Ok, based on the preceding discussion, I again amended my original comment, and added two further suggestions to remove L 138 and L139.

roldanbob (changes requested):
Taking into account the followup comments from @Naros , @james-johnston-thumbtack, @jpechane, and @mfvitale, I've revised my original review comment and suggested removal of the two lines that follow the table.

documentation/modules/ROOT/pages/configuration/signalling.adoc
ani-sha: Thanks all for your input, added the changes!

jpechane:
@ani-sha Applied, thanks!

PR #6339, opened 2025-04-11
[docs] move .github/README.md to avoid clash with main README.md

kmos:


Naros (commented):
Thanks for taking care of this @kmos :tada:

PR #6340, opened 2025-04-12
DBZ-8884 Populate source.user_name based on start events

Naros:
https://issues.redhat.com/browse/DBZ-8884

Hi @jchipmunk, this reverts the change [DBZ-7473](https://issues.redhat.com/browse/DBZ-7473) you introduced.  I know that the delayed transaction construction was important for your environment, but doing so creates an issue with both capturing the username in events and filtering by usernames.

Oracle only records the username attribute in the `START` event and its only reflected in DML events for that transaction thereafter if the DML event is captured in the same query as the `START`. If there are two mining iterations where on the first there is only the `START` and some non-captured tables and in the second is where we see the first DML for a captured table, the username attribute reference is lost and LogMiner always populates the field as `UNKNOWN`.

The easiest solution is to fall back to the behavior before the change, we were use the `START` event to create the actual transaction record in the buffer. I understand this makes the lower SCN watermark advance more slowly, but the only solution we have for users at the moment is to toggle `lob.enabled` on, which does precisely the same, even if they do not want to parse or handle LOB operations.

I'm open to other ideas if you have any. Thanks!

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
@@ -805,6 +805,12 @@ public class BufferedLogMinerStreamingChangeEventSource extends AbstractLogMiner
                     LOGGER.debug("Skipped transaction with excluded username {}", transaction.getUserName());
                     return true;
                 }
+                else if (!getConfig().getLogMiningUsernameIncludes().isEmpty()) {
+                    if (!getConfig().getLogMiningUsernameIncludes().contains(transaction.getUserName())) {
+                        LOGGER.debug("Skipped transaction with username {}", transaction.getUserName());
+                        return true;
+                    }
+                }
             }
 
             // Check whether transaction should be skipped by LogMiner CLIENT_ID field
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
@@ -794,40 +794,11 @@ public class BufferedLogMinerStreamingChangeEventSource extends AbstractLogMiner
      * mode.
      *
      * @param transaction the transaction, should not be {@code null}
-     * @return true if the transaction should be skipped and not disaptched, false otherwise
+     * @return true if the transaction should be skipped and not dispatched, false otherwise
      */
     private boolean isTransactionSkippedAtCommit(Transaction transaction) {
         // todo: can this be moved to earlier in the processing loop to avoid buffering?
-        if (transaction != null) {
-            // Check whether transaction should be skipped by LogMiner USERNAME field
-            if (!Strings.isNullOrBlank(transaction.getUserName())) {
-                if (getConfig().getLogMiningUsernameExcludes().contains(transaction.getUserName())) {
-                    LOGGER.debug("Skipped transaction with excluded username {}", transaction.getUserName());
-                    return true;
-                }
-                else if (!getConfig().getLogMiningUsernameIncludes().isEmpty()) {
-                    if (!getConfig().getLogMiningUsernameIncludes().contains(transaction.getUserName())) {
-                        LOGGER.debug("Skipped transaction with username {}", transaction.getUserName());
-                        return true;
-                    }
-                }
-            }
-
-            // Check whether transaction should be skipped by LogMiner CLIENT_ID field
-            if (!Strings.isNullOrBlank(transaction.getClientId())) {
-                if (getConfig().getLogMiningClientIdExcludes().contains(transaction.getClientId())) {
-                    LOGGER.debug("Skipped transaction with excluded client id {}", transaction.getClientId());
-                    return true;
-                }
-                else if (!getConfig().getLogMiningClientIdIncludes().isEmpty()) {
-                    if (!getConfig().getLogMiningClientIdIncludes().contains(transaction.getClientId())) {
-                        LOGGER.debug("Skipped transaction with client id {}", transaction.getClientId());
-                        return true;
-                    }
-                }
-            }
-        }
-        return false;
+        return transaction != null && (isUserNameSkipped(transaction.getUserName()) || isClientIdSkipped(transaction.getClientId()));
     }
 
     /**
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -2171,7 +2171,7 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             final String excludeList = config.getString(LOG_MINING_USERNAME_EXCLUDE_LIST);
 
             if (includeList != null && excludeList != null) {
-                problems.accept(TABLE_EXCLUDE_LIST, excludeList,
+                problems.accept(LOG_MINING_USERNAME_EXCLUDE_LIST, excludeList,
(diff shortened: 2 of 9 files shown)


Naros:
The only other alternative I can think of is if `lob.enabled` is disabled, we introduce a new `LogMinerCache<String,String>` cache that stores user names keyed by the transaction identifier. Then, when `lob.enabled` is disabled, we use this cache to resolve the username for all transaction events.

While this would retain the old behavior, I am equally concerned that we have these subtle corner cases where we do something very differently based on a given set of configuration options.  

But as I said, I'm open to ideas.

nathan-smit-1:
> The only other alternative I can think of is if `lob.enabled` is disabled, we introduce a new `LogMinerCache<String,String>` cache that stores user names keyed by the transaction identifier. Then, when `lob.enabled` is disabled, we use this cache to resolve the username at commit time for all the transaction, while also purging entries when transactions are abandoned or rolled back.
> 
> While this would retain the old behavior, I am equally concerned that we have these subtle corner cases where we do something very differently based on a given set of configuration options.
> 
> But as I said, I'm open to ideas.

BTW.  I think you intended to reference this ticket:  DBZ-7473

With the alternative proposal here just to make sure I understand:  You'd still have to revert so that you're capturing the "START" of the transaction from LOGMINER_CONTENTS.  You then cache that until commit?  I guess I'd wonder what would happen if the connector crashed.  You'd have to re-cache everything probably?

Regarding the problem that was initially solved:  What are the sorts of long-running transactions this was resolving i.e. when does deferring the transaction capture help?

Naros:
> With the alternative proposal here just to make sure I understand: You'd still have to revert so that you're capturing the "START" of the transaction from LOGMINER_CONTENTS. You then cache that until commit? I guess I'd wonder what would happen if the connector crashed. You'd have to re-cache everything probably?

That's a great counterpoint, @nathan-smit-1. It highlights that we need to revert and track the low watermark based on `START` events. Suppose we didn't do that and tried only to track the username in a separate cache, but delayed the creation of the transactions during a restart or crash. In that case, transactions risk continuing to have no username data and being included/excluded incorrectly. So I agree, reverting this is necessary.

> Regarding the problem that was initially solved: What are the sorts of long-running transactions this was resolving i.e. when does deferring the transaction capture help?
> 
> EDIT: I ask as it defers until DML occurs so I guess I'm just wondering what the subset of very long running transactions are that don't involve DML. 

Suppose we have a transaction that updates 100 million rows and only modifies non-captured tables. While this transaction occurs, other smaller transactions modify captured tables intermingled in the transaction logs by concurrent users.

In the current implementation, LogMiner would read all 100 million changes and the `START` / `COMMIT` operations, but only the `COMMIT` would mutate the offset state. As LogMiner returns data to the connector, the low watermark would advance as we read the interwoven transactions, setting the offset `scn` value based on those transactions for captured tables. If the connector is stopped and restarted midway through the 100 million changes, but we've at least seen one transaction for a captured table, the restart would only rewind to where that captured table change was observed.

The same behavior holds with this PR except how the offset `scn` is changed. Instead, when the connector is restarted, the rewind point will be where the 100 million change transaction started, even if we've captured transactions for tables that are captured midway through that large transaction.

But since early 2.x and when DBZ-7374 was added:

- The log collector system manages the log state and SCN availability reliably.
- Cache and processing performance have improved since 2.x

So I think it's worth taking the latest `main` paired with this PR and see if @jchipmunk can check their environment to see whether the issues of the past continue to be problematic today.

> I both don't want to break something that is working for others and also if this deferring is currently helping us we may end up with the same problem.

You use the non-deferred approach when you toggle on the `lob.enabled` feature. But if your connector has no interest in LOB operations, this will cause significantly higher network volume between the database & the connector.  This would play into the partition move use case. With `lob.enabled` toggled off, those operations would never enter the connector. 

The only alternative I can think of would be to drive this via configuration. Rather than `lob.enabled`, we would add yet another configuration option that would delay transaction creation, off by default. But then we need to caveat the configuration so that you can't rely on username include/exclude when delay is enabled, or automatically disable the delay configuration when you specify include/exclude lists for usernames.

But this again gets back to my original concern with all these subtle this/that toggles that continue to increase the connector's complexity, and questioning given where we are today, do we really need all of these?

nathan-smit-1:
Thanks for the detailed feedback!  Apologies, I didn't see you'd responded here before I replied on Zulip.

> 
> But since early 2.x and when DBZ-7374 was added:
> 
> * The log collector system manages the log state and SCN availability reliably.
> * Cache and processing performance have improved since 2.x
> 
> So I think it's worth taking the latest `main` paired with this PR and see if @jchipmunk can check their environment to see whether the issues of the past continue to be problematic today.
> 

I guess we'll know once testing has been done but I wonder if the symptom of the SCN being stuck wasn't necessarily connected to long-running transactions

Naros:
Hi @nathan-smit-1 that's my gut feeling as well.  It could have either been that or even GoldenGate's replication marker transactions which we also fixed, which looked exactly like an in-progress transaction that never ends.

nathan-smit-1:
hey @Naros I know this is in Draft still so was wondering if you were looking at this as a 3.2.x change potentially?  I'm not sure if there is any test case for the original issue?  I only have lob.enabled=true in one of our environments but we haven't had any issues with it so far in terms of getting stuck while mining.

I wanted to ask also if the issue highlighted here:  [#community-oracle > Unsupported operation in Oracle @ 💬](https://debezium.zulipchat.com/#narrow/channel/348250-community-oracle/topic/Unsupported.20operation.20in.20Oracle/near/506244433) where operations are listed as UNSUPPORTED unless lob.enabled=true would also be resolved by this?

Naros:
Hi @nathan-smit-1, my apologies for the late response as I have been a bit busy with the other unbuffered changes.

> was wondering if you were looking at this as a 3.2.x change potentially?

Yes, this will be added in 3.2.0.Alpha2. Should Andrey react after we merge it, we have plenty of time to rollback or adjust accordingly before 3.2.0.Final toward the end of June.

> I'm not sure if there is any test case for the original issue? I only have lob.enabled=true in one of our environments but we haven't had any issues with it so far in terms of getting stuck while mining.

So I prefer doing test-driven design (TDD), so I first added the tests, verified the code failed or did not do what it should, and then I made code changes to make the test pass. So I can confirm that I was able to see that without `lob.enabled`, the username fields were not always being populated as expected.

But I agree, I should probably include a test with `lob.enabled` toggled on for coverage. :+1: 

> where operations are listed as UNSUPPORTED unless lob.enabled=true would also be resolved by this?

No, that needs to be tackled separately. 

This change primarily makes sure that regardless of the configuration of `lob.enabled`, we always capture the username from the `START` events. Then later, when the transaction is filtered, we're always guaranteed that we'll have the user details so that the filtering works as expected.  It also makes sure that events emitted have the right username in the source information block, too.

Naros:
@jpechane, this should be good.  I added an option in case @jchipmunk wants to use the legacy behavior so that we can get this fix in for those who need this behavior.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -805,6 +805,12 @@ private boolean isTransactionSkippedAtCommit(Transaction transaction) {
                     LOGGER.debug("Skipped transaction with excluded username {}", transaction.getUserName());
                     return true;
                 }
+                else if (!getConfig().getLogMiningUsernameIncludes().isEmpty()) {
```
jpechane, line 808: Do we need here and below nested conditions? I beleive we can either comapre directly as empty string will not match current username or the conditons can be put in a single statement with and operator.

The same applies to the change below.

jpechane (commented):
@Naros LGTM, thanks! I left one comment related to code structure, other than that I think this is good to go

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
Naros: Hi @jpechane fixed, and I also chose to pull up the logic across both processors w/test adjustments.

jpechane:
@Naros That's better, applied, thanks!

PR #6342, opened 2025-04-13
DBZ-8925 Improve Hybrid strategy performance with the Oracle ObjectId cache

Naros:
https://issues.redhat.com/browse/DBZ-8925

The goal here is to optimize repeated lookups for the same `objectId` / `dataObjectId` tuple. The lookup cost is quite costly for large transactions, and by caching the prior lookup, we can avoid the cache call entirely.  For this unique case, we can gain around a 4x - 4.5x performance increase alone.


jpechane:
@Naros Interesting piece of critical path, thanks!

PR #6347, opened 2025-04-14
DBZ-8779 Use async engine factory as the default builder factory

vjuranek:
https://issues.redhat.com/browse/DBZ-8779


jpechane:
PostgreSQL tests are affected by the change

vjuranek:
Indeed, should be fixed now.

jpechane:
@vjuranek Applied, thanks

PR #6348, opened 2025-04-14
DBZ-8885: Backporting option to reset lagBehindSource metric separately

ashoke-cube:
Backporting changes made in PR https://github.com/debezium/debezium/pull/6305


jpechane:
@ashoke-cube Thanks for the PR. Our policy is the by default we backport bugfixes only. Given the isolated state of this change I am fine with backport but I'd backport to 3.1 not 3.0.
Is it OK for you? If yes could you then open the PR against 3.1 branch?

PR #6352, opened 2025-04-15
DBZ-8938 Remove unnecessary metadata query & map gets

Naros:
https://issues.redhat.com/browse/DBZ-8938


jpechane:
@Naros Applied, thanks!

PR #6361, opened 2025-04-23
DBZ-8945 Apicurio registry configuration should include instructions for confluent compatibility mode

kmos:
### Context

The documentation on Avro schema registry has two sections - Apicurio registry and Confluent registry.   However it doesn't have anything for Apicurio in "compatibility mode".  however the Apicurio mode is not very popular but It's possible to run Apicurio serialization in compatibilty mode with just a few settings

### Changes
We propose to add a section in the avro serialization documentation that explain how to use the __confluent compatibility mode__. 

closes: https://issues.redhat.com/browse/DBZ-8945

### DIFF
--- a/documentation/modules/ROOT/pages/configuration/avro.adoc
+++ b/documentation/modules/ROOT/pages/configuration/avro.adoc
@@ -80,6 +80,7 @@ This makes each record even smaller. For an I/O bound system like Kafka, this me
 
 * Avro _Serdes_ (serializers and deserializers) for Kafka producers and consumers.
 Kafka consumer applications that you write to consume change event records can use Avro Serdes to deserialize the change event records.
+* Apicurio's _Confluent compatibility mode_, enabled by setting `as-confluent: true`, allows Apicurio Registry to serialize Kafka messages using the same wire format as Confluent Schema Registry (including the magic byte and 4-byte schema ID). This enables full interoperability with Confluent clients and tools, making Apicurio a drop-in replacement without requiring changes to producers or consumers.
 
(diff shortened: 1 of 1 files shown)


jpechane:
@kmos LGTM, thanks. Still I'd appreciate if you could investigate a bit deeper and add a documentation for each of the settings necessary for compat mode explaining why it is needed. Also I am really not sure about `schemas.enable` setting. Is it really needed?

kmos:
> Also I am really not sure about `schemas.enable` setting. Is it really needed?

for what I know, yes. `schemas.enable` prevents from embedding the Avro schema inline in the Kafka message (which is not allowed with the Confluent schema registry model). If you leave it `true`:

- It may embed the full schema in the payload.
- This conflicts with the Confluent format (which expects a schema ID lookup via the registry).

>  Still I'd appreciate if you could investigate a bit deeper and add a documentation for each of the settings

Sure, I can explain all the settings. My concern is that for the previous configuration (The Apicurio mode) doesn't give any information about the need.

kmos:
IMHO there is no need to have both (Apicurio and Confluent Compatible mode). Maybe I can remove the previous one and leave the Confluent one with an explaination for each of them?

kmos:
> @kmos Well, thanks for volunteering in expansion of the additional part of the docs :-). But it is/will be different issue/PR.

hahahah ok :dancers:

kmos:
@jpechane  added details and opened another issue: https://issues.redhat.com/browse/DBZ-8980

jpechane:
@kmos Nice, applied!

PR #6362, opened 2025-04-23
DBZ-8948 Add polling start and stop engine callbacks

vjuranek:
https://issues.redhat.com/browse/DBZ-8948


jpechane:
@vjuranek Applied, thanks!

PR #6366, opened 2025-04-28
Dbz 8940 Processing error because of incomplete date part of DATETIME datatype in MariaDB

kmos:
## Context

MariaDB allows `DATETIME` values with incomplete date parts (e.g., `2015-00-12` is considered valid by MariaDB).
In Debezium MariaDB connector versions up to 3.0.7.Final, such values were silently converted to `null`.
Starting with 3.0.8.Final and 3.1.x, processing these incomplete DATETIME values causes the connector to fail with a `java.time.DateTimeException`.

## Changes

The current behavior causes the connector to fail when encountering invalid or incomplete DATETIME values. The root cause is a regression introduced by `maria-db-client` with version >`3.2.0`. We update the  client to `3.5.3`.

## References
[Supported values for datetime in MariaDB](https://mariadb.com/kb/en/datetime/#supported-values)
[Dates conatining zero day or month result into DateTimeException](https://jira.mariadb.org/browse/CONJ-1226)

closes https://issues.redhat.com/browse/DBZ-8940

### DIFF
--- a/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/AbstractBinlogConnectorIT.java
+++ b/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/AbstractBinlogConnectorIT.java
@@ -11,6 +11,7 @@ import org.apache.kafka.connect.source.SourceConnector;
 
 import io.debezium.connector.binlog.util.BinlogTestConnection;
 import io.debezium.embedded.async.AbstractAsyncEngineConnectorTest;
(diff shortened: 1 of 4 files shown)


jpechane:
I completely forgot about https://issues.redhat.com/browse/DBZ-8758?focusedId=26721265&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-26721265

jpechane:
@kmos Applied, thanks!

PR #6369, opened 2025-04-29
DBZ-8777 Delete rest code and test cases

AlvarVG:
Deleting rest code and test cases from the repository


debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/TestInfrastructureHelper.java
```diff
@@ -27,11 +27,9 @@
 import org.testcontainers.containers.startupcheck.MinimumDurationRunningStartupCheckStrategy;
```
rk3rn3r, line 27: ` debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/TestInfrastructureHelper.java` should be only used by REST tests and could then be entirely removed.

debezium-embedded/src/test/java/io/debezium/embedded/AbstractConnectorTest.java
```diff
@@ -63,7 +63,6 @@
 import org.awaitility.core.ConditionTimeoutException;
```
rk3rn3r, line 63: There should be more files in the `rest` package that were only used by RestExtension classes and the schema generator and can be removed too.
I think the schema generator module can be removed too. We can revisit the git history when we need to bring parts of it back for the Debezium Platform eventually. wdyt @jpechane @mfvitale ?

rk3rn3r:
I think this is on a good way! I did not called it "rabbit hole" without meaning it on Zulip ;).
Left some small comments inline.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/TestInfrastructureHelper.java
rk3rn3r: Never mind. Still in use for the sink connector test pipelines.

PR #6370, opened 2025-04-29
DBZ-8955 Avoid to download operator from maven during dry run

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-8955


jpechane:
@mfvitale Applied, thanks

PR #6372, opened 2025-04-30
DBZ-8995 Make StreamingChangeEventSource closeable

Naros:
https://issues.redhat.com/browse/DBZ-8995

### DIFF
--- a/debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
+++ b/debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
@@ -383,6 +383,11 @@ public class ChangeEventSourceCoordinator<P extends Partition, O extends OffsetC
                 blockingSnapshotExecutor.awaitTermination(SHUTDOWN_WAIT_TIMEOUT.toMillis(), TimeUnit.MILLISECONDS);
             }
 
+            if (streamingSource != null) {
+                // Close streaming source
+                streamingSource.close();
+            }
+
             Optional<SignalProcessor<P, O>> processor = getSignalProcessor(previousOffsets);
             if (processor.isPresent()) {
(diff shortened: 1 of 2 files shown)


debezium-core/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
```diff
@@ -311,15 +311,15 @@ protected CatchUpStreamingResult executeCatchUpStreaming(ChangeEventSourceContex
...
-        getSignalProcessor(previousOffsets).ifPresent(signalProcessor -> registerSignalActionsAndStartProcessor(signalProcessor,
-                eventDispatcher, this, connectorConfig));
-
         if (snapshotterService != null && !snapshotterService.getSnapshotter().shouldStream()) {
             LOGGER.info("Streaming is disabled for snapshot mode {}", snapshotterService.getSnapshotter().name());
             return;
         }
 
+        initStreamEvents(partition, offsetContext);
+        getSignalProcessor(previousOffsets).ifPresent(signalProcessor -> registerSignalActionsAndStartProcessor(signalProcessor,
```
mfvitale, line 320: If we move this here, it will not be possible to execute a `blocking` snapshot.

Naros: I don't understand @mfvitale, why would blocking snapshots be possible without streaming?

mfvitale: Yes, you're right but maybe in future blocking snapshot, since it is just an on-demand initial snapshot, could be possible to be trigger even if streaming is not running.  

Also, we are excluding the possibility to execute any signal.

vjuranek (approved):

PR #6375, opened 2025-04-30
DBZ-8979 JdbcSchemaHistory Fails to Handle Data Sharding When Recovering Records

tmq77:
[JdbcSchemaHistory Fails to Handle Data Sharding When Recovering Records](https://issues.redhat.com/browse/DBZ-8979)

https://issues.redhat.com/browse/DBZ-8979

### DIFF
--- a/debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/history/JdbcSchemaHistory.java
+++ b/debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/history/JdbcSchemaHistory.java
@@ -170,30 +170,26 @@ public final class JdbcSchemaHistory extends AbstractSchemaHistory {
                             StringBuilder historyDataBuilder = new StringBuilder();
                             boolean isNotFirst = false;
                             while (rs.next()) {
-                                // warning: No continuity check - nice to have
-                                int recordInsertSeq = rs.getInt("history_data_seq");
+
+                                int insertSeq = rs.getInt("history_data_seq");
                                 String historyDataSpc = rs.getString("history_data");
-                                // Reduce if else execution
+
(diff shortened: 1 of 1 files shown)


jpechane:
@tmq77 Thanks for the PR. Could you please build it locally first so the code is properly autoformatted?

jpechane (changes requested):
@tmq77 Thanks for the PR! This is good start but I beleive few more changes are needed

- We need to define the role of sequence field, it seems to me that on the store side it is total counter, so each record and each chunk has its own unique sequnece number, your code count with sequence to be reset with each record
- It is really important to add a test for this schema history that would test both store and recover for chunked records

ochedru:
I'm stepping in here because I've just encountered the same issue.
Column `record_insert_seq` is incremented at every insertion and a new `id` is also generated for each row. So we basically lack a column identifying a record split into multiple rows. I think we can fix the insertion to use `record_insert_seq` as a record identifier, and `history_data_seq` as a chunk identifier.

ochedru:
@tmq77 I was unable to push changes directly to your branch, so I pushed them here so you can review them and update your branch if you want to: https://github.com/ochedru/debezium/tree/fix/DBZ-8979-data-split. 
I kept the existing `debezium_database_history` schema for compatibility but we need only columns `record_insert_seq`, `history_data`, `history_data_seq`.

jpechane:
@ochedru Could you please create a new PR from your new branch so we can continue with the contribution process. We can the close this PR if your one will be finished.

ochedru:
@jpechane Thanks, I've just opened https://github.com/debezium/debezium/pull/6689.

Naros:

Hi @tmq77, please sign off all commits with:

```
git commit -s
```

Naros:

PR #6377, opened 2025-05-01
DBZ-8974 Include JSON source in deserialization exceptions

Naros:
https://issues.redhat.com/browse/DBZ-8974


Naros:
The test failures are all related to the REST tests that Alvar is fixing in #6369

PR #6378, opened 2025-05-02
DBZ-9001 Validate Oracle `signal.data.collection`

Naros:
https://issues.redhat.com/browse/DBZ-9001


jpechane:
@Naros Applied, thanks

PR #6379, opened 2025-05-03
DBZ-9002 Delay relational snapshot in-progress notification until export begins

Naros:
https://issues.redhat.com/browse/DBZ-9002


Naros:
Failures should be fixed in #6369

jpechane:
@Naros Applied, thanks

PR #6380, opened 2025-05-03
DBZ-9003 Prevent `idle in transaction` connection state with initial only snapshots

Naros:
https://issues.redhat.com/browse/DBZ-9003

@jpechane this is in reference to the Zulip conversation [#community-postgresql > Snapshot connectors cause idle in transaction @ 💬](https://debezium.zulipchat.com/#narrow/channel/348249-community-postgresql/topic/Snapshot.20connectors.20cause.20idle.20in.20transaction/near/515885512). I need to test this change against my recent expectations with the Oracle LogMiner unification, but I have little cause for concern.


Naros:
Failures should be fixed in #6369

jpechane:
@Naros Applied, thanks

PR #6381, opened 2025-05-04
DBZ-9004: Connection validation timeout for connectors

rajdangwal:
Addressing https://issues.redhat.com/browse/DBZ-9004


debezium-core/src/main/java/io/debezium/util/ConnectionValidationUtil.java
```diff
@@ -0,0 +1,39 @@
...
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Future;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class ConnectionValidationUtil {
```
jpechane, line 17: @rajdangwal Do we need a new class for it? IMHO this would fit well into `Threads` class.
In that case you could also refactor `io.debezium.jdbc.JdbcConnection.doClose()`

jpechane (commented):
@rajdangwal LGTM, thanks! I left one recommendation related to a bit more generalized approach.

debezium-core/src/main/java/io/debezium/util/Threads.java
```diff
@@ -310,4 +313,34 @@ public static ExecutorService newSingleThreadExecutor(Class<?> component, String
...
+     * Runs an operation with a timeout using a single-threaded executor.
+     *
+     * @param componentClass the class of the component using this method
+     * @param operation the operation to run
+     * @param timeoutMs the timeout in milliseconds
+     * @param componentName the name of the component
+     * @param operationName the name of the operation being executed with timeout
+     * @throws Exception if the operation fails or times out
+     */
+    public static void runWithTimeout(Class<?> componentClass, Runnable operation, long timeoutMs, String componentName, String operationName) throws Exception {
```
jpechane, line 327:
```suggestion
    public static void runWithTimeout(Class<?> componentClass, Runnable operation, Duration timeoutMs, String componentName, String operationName) throws Exception {
```

rajdangwal: Thanks @jpechane 
Addressed.

jpechane:
@rajdangwal Applied, thanks!

PR #6392, opened 2025-05-06
DBZ-8956 Create Maven Modules for Debezium Engine Quarkus Extension

kmos:
## Context
In traditional Debezium setups, developers must manually add two dependencies:

1. `debezium-engine` – the core library that powers the embedded change data capture (CDC) runtime.
2. A specific connector – e.g., `debezium-connector-postgres`, the actual __plugin__ that knows how to communicate with the underlying database.

This model works well for flexible setups but introduces friction when adapting to Quarkus, which aims for fast startup and native image compatibility via GraalVM. Having loosely coupled, runtime-loaded connectors with complex reflection/configuration mechanisms makes this fragile or incompatible with native compilation.

## Decision

We will provide a single Quarkus artifact that bundles both the engine and a specific connector into one dependency. For example:

```xml
<dependency>
    <groupId>io.quarkus</groupId>
    <artifactId>quarkus-debezium-postgres</artifactId>
    <version>${version.debezium}</version>
</dependency>
```

This will include:

- The Debezium engine
- The selected connector (e.g., Postgres)
- Sensible Quarkus-based defaults to eliminate boilerplate
- Automatic internal handling of the connector class setup (e.g., `connector.class=...`), hidden from the user

Furthermore, the features released in this PR are:
- [X] CI for Quarkus extensions
- [X] Skeleton for the Maven modules
- [X] `build steps` for native compilation
- [X] `build steps` for Quarkus Dev Services datasource instrumentation
- [X] Integration tests for native builds
- [X] `Debezium` bean abstraction

### Developer Experience
The proposed changes help developers create Debezium applications as follows:

1. Add the dependency in the `pom.xml`:
```xml
<dependency>
    <groupId>io.quarkus</groupId>
    <artifactId>quarkus-debezium-postgres</artifactId>
</dependency>
```

2. Add the configuration in `application.properties`:
```properties
# Debezium CDC configuration
quarkus.debezium.name=native
quarkus.debezium.topic.prefix=dbserver1
quarkus.debezium.table.include.list=inventory.products
quarkus.debezium.plugin.name=pgoutput
quarkus.debezium.snapshot.mode=never

# Transformation
quarkus.debezium.transforms.t0.add.fields=op,table
quarkus.debezium.transforms.t0.add.headers=db,table
quarkus.debezium.transforms.t0.negate=false
quarkus.debezium.transforms.t0.predicate=p2
quarkus.debezium.transforms.t0.type=io.debezium.transforms.ExtractNewRecordState
quarkus.debezium.transforms=t0
quarkus.debezium.predicates.p2.pattern=inventory.inventory.products
quarkus.debezium.predicates.p2.type=org.apache.kafka.connect.transforms.predicates.TopicNameMatches
quarkus.debezium.configuration.predicates=p2

# Agroal config
quarkus.datasource.db-kind=postgresql
quarkus.datasource.devservices.image-name=quay.io/debezium/postgres:15
```

3. Add a resource like:
```java
@Path("/api/debezium")
public class DebeziumEndpoint {

    @Inject
    private Debezium debezium;

    @GET
    @Path("manifest")
    public DebeziumManifest getManifest() {
        return debezium.manifest();
    }
}
```

4. Launch the application in dev mode using `mvn clean quarkus:dev`  
   This will use [Dev Services](https://quarkus.io/guides/dev-services).

### Design Decisions

- The proposal introduces an interface `Debezium` as an abstraction that exposes the engine's snapshot status.
- The Maven module structure shifts the focus from __Debezium__ to __Quarkus__ (with packages like `io.quarkus.xxx`) to ensure compatibility with Quarkus (potentially for future migration to the Quarkus repository) and support multiple extensions.

### Dev Services

The Postgres Dev Service provided by the Quarkus Postgres Extension uses an image that is incompatible with Debezium (WAL not active). It's necessary to override it in the properties:

```properties
quarkus.datasource.devservices.image-name=quay.io/debezium/postgres:15
```

In future, to improve developer experience, we should override the datasource image.

### Poor Native Build

Currently, the native build has some limitations:
- `META-INF/services` is injected using a wildcard
- Reflection classes are included based on empirical assumptions

In both cases, we should provide a more reliable way to configure reflection for Quarkus.

### Space for a Core Extension Module

Although the proposal currently places all code inside a single extension (YAGNI), in the future — when additional connectors are added — there is room for a core module that contains shared logic (especially for native builds).

## References
closes: https://issues.redhat.com/browse/DBZ-8956
link: https://github.com/debezium/debezium-design-documents/pull/16


Naros:
@kmos what do you think about if we introduced a `quarkus-debezium-engine` artifact that has deployment, integration-test, and runtime sub-modules?  I believe that would be a great place for us to place common behavior that isn't specific to a given connector implementation.

quarkus-debezium-parent/quarkus-debezium-postgres-parent/deployment/src/main/java/io/quarkus/debezium/postgres/deployment/QuarkusDatasource.java
```diff
@@ -0,0 +1,23 @@
...
+
+import java.net.URI;
+import java.util.Map;
+
+public class QuarkusDatasource {
+    public static Map<String, String> generateDebeziumConfiguration(Map<String, String> properties) {
+        URI uri = URI.create(properties.get("quarkus.datasource.jdbc.url").substring(5));
+
+        return Map.of(
+                "quarkus.debezium.configuration.database.hostname", uri.getHost(),
```
mfvitale, line 17: I'm wondering if we should just use `quarkus.debezium.*` instead of `quarkus.debezium.configuration.*`.

quarkus-debezium-parent/quarkus-debezium-postgres-parent/runtime/src/main/java/io/quarkus/debezium/configuration/DebeziumEngineConfiguration.java
```diff
@@ -0,0 +1,24 @@
...
+/**
+ * Debezium configuration.
+ */
+@ConfigMapping(prefix = "quarkus.debezium")
+@ConfigRoot(phase = ConfigPhase.BUILD_AND_RUN_TIME_FIXED)
+public interface DebeziumEngineConfiguration {
+    /**
+     * Configuration properties
+     */
+    Map<String, String> configuration();
```
mfvitale, line 23: I'm wondering if we should just use quarkus.debezium.* instead of quarkus.debezium.configuration.*.

quarkus-debezium-parent/quarkus-debezium-postgres-parent/runtime/src/main/java/io/quarkus/debezium/engine/DebeziumRunner.java
```diff
@@ -0,0 +1,40 @@
...
+
+    public void shutdown() throws RuntimeException {
+        LOGGER.info("Shutting down Debezium Engine...");
+        try {
+            engine.close();
+        }
+        catch (IOException e) {
+            throw new RuntimeException("Impossible to shutdown Debezium Engine ", e);
+        }
+        executorService.shutdown();
```
mfvitale, line 38: Should this be in a `finally`?

quarkus-debezium-parent/quarkus-debezium-postgres-parent/runtime/src/main/java/io/quarkus/debezium/configuration/DebeziumEngineConfiguration.java
kmos: I think it's feasible

quarkus-debezium-parent/quarkus-debezium-postgres-parent/runtime/src/main/java/io/quarkus/debezium/engine/DebeziumRunner.java
kmos: yes definitely!

quarkus-debezium-parent/quarkus-debezium-postgres-parent/runtime/src/main/java/io/quarkus/debezium/engine/Debezium.java
```diff
@@ -0,0 +1,36 @@
...
+import java.util.concurrent.ExecutorService;
+
+import io.debezium.engine.DebeziumEngine.Signaler;
+
+/**
+ * The Debezium engine abstraction in the Quarkus CDI
+ * <p>
+ * The Engine is submitted to an {@link Executor} or {@link ExecutorService} for execution by a single thread
+ */
+public interface Debezium {
```
mfvitale, line 19: Should we mark this as @Incubating?

quarkus-debezium-parent/quarkus-debezium-postgres-parent/runtime/src/main/java/io/quarkus/debezium/engine/ManifestHandler.java
```diff
@@ -0,0 +1,17 @@
...
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package io.quarkus.debezium.engine;
+
+import io.debezium.engine.DebeziumEngine;
+
+public interface ManifestHandler {
```
mfvitale, line 11: Should we mark this as @Incubating?

quarkus-debezium-parent/quarkus-debezium-postgres-parent/runtime/src/main/java/io/quarkus/debezium/engine/SourceRecordDebezium.java
```diff
@@ -0,0 +1,72 @@
...
+
+import io.debezium.config.Configuration;
+import io.debezium.embedded.Connect;
+import io.debezium.engine.DebeziumEngine;
+import io.debezium.engine.DebeziumEngine.Signaler;
+import io.debezium.engine.format.ChangeEventFormat;
+import io.quarkus.debezium.configuration.DebeziumEngineConfiguration;
+
+@ApplicationScoped
+public class SourceRecordDebezium extends RunnableDebezium {
```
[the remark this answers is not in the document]
kmos: Well we should mark with `@Incubating` `Debezium` interface and `DebeziumManifest`

quarkus-debezium-parent/quarkus-debezium-postgres-parent/runtime/src/main/java/io/quarkus/debezium/engine/Debezium.java
kmos: exactly

quarkus-debezium-parent/quarkus-debezium-postgres-parent/runtime/src/main/java/io/quarkus/debezium/engine/SourceRecordDebezium.java
```diff
@@ -0,0 +1,72 @@
...
+
+        this.engine = DebeziumEngine.create(ChangeEventFormat.of(Connect.class))
+                .using(Configuration.empty()
+                        .withSystemProperties(Function.identity())
+                        .edit()
+                        .with(Configuration.from(debeziumEngineConfiguration.configuration()))
+                        .build().asProperties())
+                .using(this.manifestHandler.connectorCallback())
+                .using(this.manifestHandler.completionCallback())
+                .notifying(event -> LOGGER.info("**EXPERIMENTAL** {}", event.record().value().toString()))
```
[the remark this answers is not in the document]
kmos: We can do whatever we want. The point is that if we want something useful not only for us (internal feedback) but also for users (external feedback), then we should aim higher. 

I mean, if we want to include it in the next alpha, we could make an extra effort to provide a basic listener and deliver value to users — and also get feedback for ourselves.

My approach is always to keep short to facilitate the review but if we want to push a bit, I am in. WDYT @jpechane

quarkus-debezium-parent/quarkus-debezium-postgres-parent/runtime/src/main/java/io/quarkus/debezium/configuration/DebeziumEngineConfiguration.java
Naros: I absolutely agree @mfvitale, I really would prefer to avoid unnecessary config option name bloat.

quarkus-debezium-parent/quarkus-debezium-postgres-parent/runtime/src/main/java/io/quarkus/debezium/engine/SourceRecordDebezium.java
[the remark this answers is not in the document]
jpechane: Yes, why not? Just make sure to use @Incubating annotation as I really expect non-trivila changes in the future so it is important that users play with it but they understand that at this stage everything is open to breaking changes.

kmos:
> @kmos what do you think about if we introduced a `quarkus-debezium-engine` artifact that has deployment, integration-test, and runtime sub-modules? I believe that would be a great place for us to place common behavior that isn't specific to a given connector implementation.

Hi @Naros ,

please review the [last commit](https://github.com/debezium/debezium/tree/c2266c49ba86c5f979aea6b97de084247775192d): _DBZ-8956 add `quarkus-debezium-engine` and `quarkus-debezium-engine-spi`_.

It contains two new _quarkus module structure_ in particular:

- `quarkus-debezium-engine-spi` defines interfaces and `BuildItem` to produce correctly a _Debezium Engine_
- `quarkus-debezium-engine` defines common (and concrete) classes based on the `spi`

Naros:
I like the split, great job.  I have a few questions, but I'll ask those directly.

PR #6395, opened 2025-05-07
DBZ-9016 Avoid to process streaming event when incremental snapshot is not running

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-9016


vjuranek (approved):

vjuranek:
Merged, thanks!

PR #6396, opened 2025-05-07
DBZ-9018: fix issues using latest jdbc driver

jonomorris:
https://issues.redhat.com/browse/DBZ-9018 

Apache Camel-Quarkus debezium-postgres update & delete tests are failing after upgrading the Postgres JDBC driver from 42.7.4 to 42.7.5 [1].

In Version 42.7.5 `getTables` includes the database name [2] and `TableIds` created using 'catalogName/databaseName', 'schemaName', and 'table' from the result-set do not equal the `TableIds` created from messages read from the replication stream that feature only 'schemaName' and 'table'.

Fixed unit test problem in `PostgresConnectorIT`. In Version 42.7.0 the LSN of the of the keepAlive message is used to advance the flush LSN [3].  This causes the test to fail if a 'status-update' packet is sent during the test.  We can set the 'status-update' interval so that it is not sent during test execution.

Also Added the `SkipTestDependingOnDecoderPluginNameRule` to `CloundEventsConverterIT` in order to correctly skip a test that should not be run when the decoderPlungin name is not 'pgoutput'

1. https://github.com/apache/camel-quarkus/issues/6911
2. https://github.com/pgjdbc/pgjdbc/blob/REL42.7.5/pgjdbc/src/main/java/org/postgresql/jdbc/PgDatabaseMetaData.java#L1434
3. https://github.com/pgjdbc/pgjdbc/pull/2941/files#diff-5d411513d3e0fcbc7d3eefaaaf24e5a845a9655d277f6ab1e54f7f9de66d7df9R241

### DIFF
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/CloudEventsConverterIT.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/CloudEventsConverterIT.java
@@ -18,13 +18,16 @@ import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.source.SourceRecord;
 import org.apache.kafka.connect.transforms.HeaderFrom;
 import org.junit.Before;
+import org.junit.Rule;
 import org.junit.Test;
+import org.junit.rules.TestRule;
 
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.ObjectNode;
 
 import io.debezium.config.Configuration;
 import io.debezium.connector.postgresql.PostgresConnectorConfig.SnapshotMode;
+import io.debezium.connector.postgresql.junit.SkipTestDependingOnDecoderPluginNameRule;
 import io.debezium.connector.postgresql.junit.SkipWhenDecoderPluginNameIsNot;
 import io.debezium.connector.postgresql.transforms.DecodeLogicalDecodingMessageContent;
 import io.debezium.converters.AbstractCloudEventsConverterTest;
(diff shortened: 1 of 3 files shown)


Naros:
Hi @jonomorris, thanks.  

So my concern is with the change to `PostgresConnectorIT` regarding DBZ-5118. While this fixes the test failure, the upgrade to the 42.7.1+ driver (which is currently blocked in #5314) introduces a change in the flush behavior that was not present in 42.7.0 or earlier versions.  So while extending the update interval to a value greater than the test's run test makes the test pass, it does not address the need for Debezium to only control the LSN flush.

I genuinely would prefer that we advocate for https://github.com/pgjdbc/pgjdbc/pull/3403 to be added and a new PG driver cut. That would provide us a hook where the driver's automatic LSN flushing can be disabled and preserves the behavior of DBZ-5118 in the `PostgresConnectorIT` test.

jonomorris:
@Naros, thanks for that, it makes sense.  I've reverted the change to `PostgresConnectorIT`, commit `c10b3aa`.

@davecramer has merged https://github.com/pgjdbc/pgjdbc/pull/3403, hopefully that will be added to a new PG driver cut.

Another issue I found with 42.7.5 was the removal of 'application name' [1].  Without it Debezium has problems identifying rows in the `pg_stat_activity` table [2].   The 'application name' was added back last week [3]. Hopefully that will be also added to the new PG driver cut.

1. https://github.com/pgjdbc/pgjdbc/pull/3476/files#diff-dba283e2ce6b2f6a8a889e9dac54cce198602b32dea901c429eb845e02b2d7b0L400-L402

2. https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java#L389

3. https://github.com/pgjdbc/pgjdbc/pull/3509/files

Naros:
Hi @jonomorris, thanks for helping identify & champion these pgjdbc items. I've reached out to the pgjdbc maintainers and it seems they have a few changes they're looking to include, but we should expect 42.7.6 "soon" (tm) :grin:

jpechane:
@jonomorris I suppose this can be closed now, can't it?

jonomorris:
@jpechane yeah this can be closed with the changes applied in https://github.com/debezium/debezium/pull/6472

PR #6398, opened 2025-05-12
DBZ-8777 Fix failing MongoDB SinkTest

AlvarVG:


rk3rn3r:
This is not correct and breaking CI. Can you explain why you did this change? It looks like the classpath was not correct when an error happened. Could this be? @vjuranek  @AlvarVG ?

PR #6403, opened 2025-05-13
Dbz 8777 Restore RequireAssemblyProfile and isAssemblyProfileActive

AlvarVG:


rk3rn3r:
Isn't this only required to bring back to MongoDB (sink) connector and for JDBC connector?

AlvarVG:
@rk3rn3r No. It is only for MongoDB. But I have added the isAssemblyProfileActive for all the poms i have deleted it from, just in case.

PR #6404, opened 2025-05-13
DBZ-9022 Added support for a new enum in delete tombstone handling mode

vaibhav-yb:
The code to create tombstones in [EventDispatcher.StreamingChangeRecordReceived#changeRecord](https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java#L545) is something like:

```java
queue.enqueue(record);

if (shouldEmitTombstones) {
  queue.enqueue(tombstoneRecord);
}
```

Speaking in terms of Postgres semantics (would be similar for other databases), let's say that the delete record is published to Kafka, so our `lastCompletelyProcessedLsn` would take the value as specified by the delete record (assume that it's 100) and after publishing the delete record, before the connector can publish a tombstone record, it crashes or restarts. Upon restart, we would start from this `lastCompletelyProcessedLsn` and will not receive this delete record again and hence no tombstone would be published.

This will be an eventual data loss.

## Changes

This PR introduces a workaround which will help all the downstream connectors who rely on tombstone records like the JDBCSinkConnector by Confluent.

This introduces another value `rewrite-deletes` to the `delete.tombstone.handling.mode` which:
1. Converts all the delete records to tombstone records - ensuring we process the deletes for the target table.
2. Drops all the tombstone records published by the source connectors.

Additional change in the PR is to modify the documentation to add the newly added value for the configuration property `delete.tombstone.handling.mode`.

This closes https://issues.redhat.com/browse/DBZ-9022

### DIFF
--- a/debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordStateConfigDefinition.java
+++ b/debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordStateConfigDefinition.java
@@ -20,7 +20,8 @@ public class ExtractNewRecordStateConfigDefinition {
         DROP("drop"),
         TOMBSTONE("tombstone"),
         REWRITE("rewrite"),
-        REWRITE_WITH_TOMBSTONE("rewrite-with-tombstone");
+        REWRITE_WITH_TOMBSTONE("rewrite-with-tombstone"),
+        REWRITE_DELETES("rewrite-deletes");
 
         private final String value;
 
@@ -78,7 +79,8 @@ public class ExtractNewRecordStateConfigDefinition {
                     + "drop - Remove the delete event and tombstone from the stream."
                     + "tombstone (default) - For each delete event, leave only a tombstone in the stream."
                     + "rewrite - Remove tombstone from the record, and add a `__deleted` field with the value `true`."
-                    + "rewrite-with-tombstone - Retain tombstone in record and add a `__deleted` field with the value `true`.");
+                    + "rewrite-with-tombstone - Retain tombstone in record and add a `__deleted` field with the value `true`."
+                    + "rewrite-deletes - Convert delete events to tombstone events and drop tombstone events.");
 
     public static final Field ROUTE_BY_FIELD = Field.create("route.by.field")
             .withDisplayName("Route by field name")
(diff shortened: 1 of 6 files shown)


debezium-core/src/main/java/io/debezium/transforms/extractnewstate/AbstractExtractRecordStrategy.java
vaibhav-yb, line 1: I'll need to add `tombstoneDelegate.close()` to the `close()` method.

Naros (commented):
Hi @vaibhav-yb this LGTM, just two comments.

1. The call in `AbstractExtractRecordStrategy#close()` you previously mentioned.
2. Can you add the new enum option to `documentation/modules/ROOT/pages/transformations/event-flattening.adoc`.

debezium-core/src/main/java/io/debezium/transforms/extractnewstate/DefaultDeleteHandlingStrategy.java
```diff
@@ -57,6 +58,9 @@ public R handleDeleteRecord(R record) {
                 // a record only with null value that by JDBC connector is treated as a flattened delete.
                 // Any change to this behavior can have impact on JDBC connector.
                 return afterDelegate.apply(record);
+            case REWRITE_DELETES:
+                LOGGER.trace("Delete message {} requested to be converted to tombstone", maybeRedactSensitiveData(record.key()));
+                return tombstoneDelegate.apply(record);
```
jpechane, line 63: @vaibhav-yb There is no need for the delegate, you can just inline `record.newRecord` here.

debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordStateConfigDefinition.java
```diff
@@ -20,7 +20,8 @@ public enum DeleteTombstoneHandling implements EnumeratedValue {
         DROP("drop"),
         TOMBSTONE("tombstone"),
         REWRITE("rewrite"),
-        REWRITE_WITH_TOMBSTONE("rewrite-with-tombstone");
+        REWRITE_WITH_TOMBSTONE("rewrite-with-tombstone"),
+        REWRITE_DELETES("rewrite-deletes");
```
jpechane, line 24: @Naros I wonder if the option name is not a bit misleading. Should not `delete-to-tombstone` or `delete-as-tombstone` be better option?

jpechane (commented):
@vaibhav-yb LGTM, thanks! I am still a bit unsure about the final option name. I asked @Naros for feedback and when this is decided we are good to go.

vaibhav-yb:
@jpechane Just curious about the GitHub actions that were run with this PR, I see that one Spanner related action has failed. But the failure doesn't look related to the PR - is it flaky?

jpechane:
@vaibhav-yb Yes, it is flaky test, no worries.

debezium-core/src/main/java/io/debezium/transforms/ExtractNewRecordStateConfigDefinition.java
Naros: @jpechane / @vaibhav-yb I agree and I believe I'd prefer `delete-to-tombstone`, but ymmv.

jpechane:
@vaibhav-yb Applied, thanks!

PR #6405, opened 2025-05-13
DBZ-8982 Leave last event w/i single PK in JDBC

ArtDu:
With a reduction buffer, we can leave only the last event within all buffers (delete and upsert buffers) since we can reduce events not only within a single buffer but across both of them.

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
@@ -119,7 +119,12 @@ public class JdbcChangeEventSink implements ChangeEventSink {
                 if (upsertBufferToFlush != null && !upsertBufferToFlush.isEmpty()) {
                     // When a delete event arrives, update buffer must be flushed to avoid losing the delete
                     // for the same record after its update.
-                    flushBufferWithRetries(collectionId, upsertBufferToFlush);
+                    if (config.isUseReductionBuffer()) {
+                        upsertBufferToFlush.remove(record);
+                    }
+                    else {
+                        flushBufferWithRetries(collectionId, upsertBufferToFlush);
+                    }
                 }
 
                 flushBufferRecordsWithRetries(collectionId, getRecordsToFlush(deleteBufferByTable, collectionId, record));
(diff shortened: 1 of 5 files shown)


debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
```diff
@@ -119,7 +119,12 @@ public void execute(Collection<SinkRecord> records) {
...
-                    flushBufferWithRetries(collectionId, upsertBufferToFlush);
+                    if (config.isUseReductionBuffer()) {
+                        upsertBufferToFlush.remove(record);
+                    }
+                    else {
+                        flushBufferWithRetries(collectionId, upsertBufferToFlush);
+                    }
                 }
 
                 flushBufferRecordsWithRetries(collectionId, getRecordsToFlush(deleteBufferByTable, collectionId, record));
```
rk3rn3r, line 130: I was wondering if flushing of the delete buffer can be removed here and moved entirely to the end of this method now (with the `flushBuffers()` code there)? Is there still a case where flushing of the delete buffer is required here?

ArtDu: It doesn't flush until the buffer is full. 
This line just needs to be checked to see whether the buffer is full and flushed if it is.
But this code is not obvious at first look.

rk3rn3r: Agreed. The method name could also be more intuitive/descriptive. Thx!

rk3rn3r (approved):
LGTM. Thx a lot @ArtDu

ArtDu:
@rk3rn3r 

I also need to have this feature in 2.7 version.
Should I create the same PR for 2.7 branch, or what is the process?

jpechane:
@ArtDu Hi, yes, you can prepare a PR against 2.7 branch and we'll merge. But please bear in mind that we don't plan to do another 2.7 release so you'll need to built it from source.

ArtDu:
> @ArtDu Hi, yes, you can prepare a PR against 2.7 branch and we'll merge. But please bear in mind that we don't plan to do another 2.7 release so you'll need to built it from source.

I just saw that you have patch releases on 2.7 (2.7.3,  2.7.4). 
Are you not planning to release version 2.7.5 at all, or are you releasing a patch only to fix bugs? 
If so I will have 2.7.5 with my patch only when serious bug needed to be fixed in 2.7, am I right?

Naros:
Hi @ArtDu there are no plans to release a 2.7.x build; however, you are free to make a local build of the 2.7 branch if you need to remain on the 2.7 branch behavior but with the patch.

PR #6411, opened 2025-05-13
[docs] Fix several Antora doc build errors/warnings

Naros:
Hi @roldanbob, can you check if these fixes create any issues for the downstream builds? These were generating several build warnings and errors.


roldanbob:
These changes all appear to be benign wrt the downstream docs. Thanks for checking!

roldanbob (approved):
LGTM

PR #6414, opened 2025-05-14
[ci] Add drop all dbs

Naros:
Just curious if this has any performance benefits on the test suite.  if it does, I'll polish up the commits & raise a jira.

### DIFF
--- a/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogBinaryModeIT.java
+++ b/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogBinaryModeIT.java
@@ -54,6 +54,8 @@ public abstract class BinlogBinaryModeIT<C extends SourceConnector> extends Abst
         finally {
             Files.delete(SCHEMA_HISTORY_PATH);
         }
+
+        dropAllDatabases();
     }
 
     @Test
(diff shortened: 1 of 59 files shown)


vjuranek:
Good catch and it seems it does have impact - IIRC MySQL/MariaDB testsuites run between 30-40 min, now around 20 min!

jpechane:
Could be there an `@After` method in `io.debezium.connector.binlog.AbstractBinlogConnectorIT<C>` dropping the databases?

Naros:
> Good catch and it seems it does have impact - IIRC MySQL/MariaDB testsuites run between 30-40 min, now around 20 min!

Ya but I think the 30-40 minutes you're thinking of was back when we had much higher logging verbosity, and Jiri and I heavily cut that back and that dropped MySQL down to around 22-23m for non-gtid and 34m for gtids. For MariaDB the prior numbers were 29m-31m, and so this change really don't have the impact I had hoped :(

> Could be there an `@After` method in `io.debezium.connector.binlog.AbstractBinlogConnectorIT<C>` dropping the databases?

We could, but given it did not make as much of a dent on the runtime of the tests like I had hoped, I'm not sure if its worth adding. Given that we always spin up a unique database per test class, we generally don't need to be concerned with taint database state from other databases that prior tests leave behind.  wdyt?

jpechane:
@Naros IMHO if you feel it does not make much difference then I'd either say we don't need to be concerned about it or I'd aim for the parent class `@After` method so it is centralized and not spread around in tests.

Naros:
Let's table it for now. If we need to resurrect the PR later, we can certainly do so.

PR #6417, opened 2025-05-14
DBZ-8992 Add Hugging Face inference model provider

vjuranek:
https://issues.redhat.com/browse/DBZ-8992


vjuranek:
Keeping as draft until I'm able to test it. Currently there is either bug in HF langchain4j client or an issue with HF infra (more likely, as the requests fails also using `curl`, but still looking on it).

vjuranek:
This code in this PR works (tested with my build of langchain4j), what is now broken is langchain4j - needs https://github.com/langchain4j/langchain4j/pull/3020 and bump to fixed langchain4j release.

debezium-bom/pom.xml
```diff
@@ -575,17 +576,22 @@
             <dependency>
                 <groupId>dev.langchain4j</groupId>
                 <artifactId>langchain4j-embeddings-all-minilm-l6-v2</artifactId>
-                <version>${version.langchain4j}</version>
+                <version>${version.langchain4j.unstable}</version>
```
jpechane, line 579: @vjuranek Do we need to have it marked as `unstable`? We don't use such approach elswhere.

jpechane (approved):
@vjuranek LGTM, thanks! I left one question, feel free to merge yourself when addressed/answered.

debezium-bom/pom.xml
vjuranek: I'll take a look if it can be workarounded somehow (e.g. importing bom file), but this seems to the approach langchain4j decided to go - until recently all artifacts have the same version, e.g. `1.0.0-beta2`. recently, they did first stable release, but only part of the artifacts (e.g. `core`) are considered stable. These artifacts have version wihout `-beta` suffix, e.g. `1.0.1`, while artifacts which are not considered stable has this suffix, e.g. `1.0.1-beta6`. I don't know if they plan different release cadence for stable/unstable artifacts, or what is the advantage here, but they use two different versions for the artifacts in the same release.

vjuranek: well, switching to bom file seems to fix the issue

PR #6420, opened 2025-05-15
DBZ-9043 Add apicurio workflow

AlvarVG:


.github/workflows/debezium-workflow-pr.yml
```diff
@@ -495,6 +495,14 @@ jobs:
   ## Please define callable workflows here in alphabetical item name order.
   ########################################################################################
 
+  apicurio_checks:
+    name: "Apicurio checks"
+    needs: [ check_style, file_changes ]
+    if: ${{ needs.file_changes.outputs.documentation-only-changed == 'false' }}
```
Naros, line 501: What if we introduce a new step in the `file_changes` job that checks for all the common, monogdb, mysql, postgres, testing, or the ddl-parser for mysql and have a new output that we use here instead:

```
      - name: Get modified files (Apicurio Tests)
        id: changed-files-apicurio
        uses: tj-actions/changed-files@v46.0.5
        with:
          files: |
            support/checkstyle/**
            support/revapi/**
            debezium-api/**
            debezium-assembly-descriptors/**
            debezium-core/**
            debezium-embedded/**
            debezium-revapi/**
            debezium-ide-configs/**
            debezium-parent/pom.xml
            debezium-bom/pom.xml
            debezium-storage/** 
            debezium-scripting/**          
            pom.xml
            .mvn/**
            .github/actions/**
            .github/workflows/debezium-workflow-pr.yml
            debezium-sink/**
            debezium-connector-mongodb/**
            debezium-connector-mysql/**
            debezium-connector-binlog/**
            debezium-connector-postgres/**
            debezium-testing/**
            debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mysql/**
            debezium-ddl-parser/src/main/java/io/debezium/antlr/**
            debezium-ddl-parser/src/test/resources/mysql/examples/**
```
and then this as an output, i.e.:
```
      apicurio-only-changed: ${{ steps.changed-files-apicurio.outputs.only_changed }}                                                            
```
and lastly change this condition as:
```
if: ${{ needs.file_changes.outputs.apicurio-only-changed == 'true' }}      
```

This makes sure we only run these 3 jobs if they're needed to avoid saturating GHA.

Naros (commented):
One inline comment @AlvarVG, otherwise LGTM.

PR #6419, opened 2025-05-15
DBZ-9027 Fix MariaDB column type mapping for UUID

Naros:
https://issues.redhat.com/browse/DBZ-9027

PR #6422, opened 2025-05-15
DBZ-9027 Support UUID columns in debezium-connector-mariadb

coekie:
https://issues.redhat.com/browse/DBZ-9027

This makes UUIDs columns in mariadb work.

I just noticed before submitting this PR that by pure coincidence work was being done on this already today in #6419 .
My changes here are bit more complete: they also handle the value conversion.

I put the tests in the mariadb module, instead of in the common binlog module with annotations on it to ignore them for mysql. I'm not sure what the preferred way is.
Note that this PR is putting the first file in `debezium-connector-mariadb/src/test/resources/ddl/`.

In the test I added a consumeSkippingSchemaChanges helper method because the number of records to skip is different between testing streaming vs snapshots; and I didn't want to hardcode imo fragile numbers of how many records to skip when.


Naros (approved):

Naros:
Applied @coekie, and congratulations on your first contribution :tada: !

PR #6426, opened 2025-05-16
DBZ 8961 Debezium Engine Quarkus Extension: Introduce Debezium `Capturing` Listener

kmos:
## Context

Debezium emits change events (INSERT, UPDATE, DELETE) in a generic form via the `{{RecordChangeEvent<K, V>}}` interface. Quarkus developers has the ability to capture change events from specific tables using annotations — similar to how message consumers are declared in Quarkus Kafka or REST endpoints in JAX-RS.

## Decision

Introduce CDI-compatible annotations to declaratively register event consumers for specific tables:

```java
import io.debezium.engine.ChangeEvent;


@ApplicationScoped  
class Orders {
  
    @Capturing("order")  
    public void capture(RecordChangeEvent<SourceRecord> event) {  
        /// some logic to apply
    } 
} 
```

in this PR we introduce:
- [X] `Capturing` annotation
- [X] native build for `Invoker` generated classes

## References
closes: https://issues.redhat.com/browse/DBZ-8961
link: https://github.com/debezium/debezium-design-documents/pull/16


quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DefaultQualifiedTableNameResolver.java
```diff
@@ -0,0 +1,25 @@
...
+
+import io.debezium.engine.RecordChangeEvent;
+
+class DefaultQualifiedTableNameResolver implements FullyQualifiedTableNameResolver {
+
+    @Override
+    public String resolve(RecordChangeEvent<SourceRecord> event) {
+        SourceRecord record = event.record();
+        Struct payload = (Struct) record.value();
+        String table = ((Struct) payload.get("source")).getString("table");
```
jpechane, line 20: No need for typecasting, this and the rest of the code can work like
```suggestion
        String table = payload.getStruct("source").getString("table");
```

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DefaultQualifiedTableNameResolver.java
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+package io.quarkus.debezium.engine;
+
+import org.apache.kafka.connect.data.Struct;
+import org.apache.kafka.connect.source.SourceRecord;
+
+import io.debezium.engine.RecordChangeEvent;
+
+class DefaultQualifiedTableNameResolver implements FullyQualifiedTableNameResolver {
```
jpechane, line 14: What if there is an SMT in the stack that removes source info block?

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DefaultQualifiedTableNameResolver.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @Override
+    public String resolve(RecordChangeEvent<SourceRecord> event) {
```
jpechane, line 17: In general this needs to be rethhought, please see `io.debezium.relational.RelationalDatabaseConnectorConfig.tableIdMapper` - it is a bit more per-database specific.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DefaultQualifiedTableNameResolver.java
Naros: I also suggest we avoid trying to tie ourselves to relational constructs, as that will likely bite us when we consider this from the standpoint of MongoDB and potentially future non-relational connectors. Is there a specific reason why we can't target the topic name rather than the table name here?

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DefaultQualifiedTableNameResolver.java
Naros: See my other comment @kmos, can we avoid the `TableId` notion all together & rather focus on topic name?

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DefaultQualifiedTableNameResolver.java
kmos: > I also suggest we avoid trying to tie ourselves to relational constructs, as that will likely bite us when we consider this from the standpoint of MongoDB and potentially future non-relational connectors. Is there a specific reason why we can't target the topic name rather than the table name here?
> 
No, any specific reason. I based it on [this](https://debezium.io/documentation/reference/3.1/configuration/signalling.html#debezium-signaling-enabling-source-signaling-channel) which cover also no-sql sources (to drive the include list).

> I realize if you're trying to use the annotation to drive the include list, this somewhat breaks that paradigm, but at the same time, I'd prefer to walk first. Maybe having just something that simply dictates this handler is for X topic?

I don't have a strong opinion. At this point I suggest to remove all this routing logic and leave only the listener and go back to the [ddd](https://github.com/debezium/debezium-design-documents/pull/16) (example):

```java
import io.debezium.engine.InsertEvent;

@ApplicationScoped  
class CaptureHandler {  
  
    @Capturing
    public void capture(RecordChangeEvent<SourceRecord> event) {
        captureService.capture();
    }
}
```

In this way we have a first MVP for the next alpha and keep focusing on the native build.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DefaultQualifiedTableNameResolver.java
mfvitale: @Naros Focusing on `topic` name can be a solution since effectively the records here are already processed by transforms. Wondering if there can be uses cases where the users could want to filter based on the real source ant not the `topic` when this are not 1:1.

jpechane: IMHO we need both. I'd say let's use topic by default but allow user to feref to a message field or header tho override the filter. That way it would be possible to use source info block and not make it hardcoded from relational database.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DefaultQualifiedTableNameResolver.java
kmos: I have removed the filtering logic (nothing to cry about :smile: ). Actually, the `CapturingInvokerRegistry<T>` return the first `CapturingInvoker<T>` defined by the user.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DefaultQualifiedTableNameResolver.java
kmos: If I think to other libraries like *spring-kafka* and in particular the `@KafkaListener`, it's possible to filter/configure by:

- topic
- ~consumer group~
- ~partition~
- `RecordFilterStrategy`


I think that we can evaluate to use concept like `topic` and `RecordFilterStrategy` which is a `predicate` of `ConsumerRecord<K,V> -> boolean` and use a similar approach. What I am wondering is there any generic *identifier* that we can use for the `source` that isn't strictly related to the Kafka domain?

jpechane (approved):
@kmos Good job, thanks!
Please double check that the tests were executed as part of this PR.
If/when yes and pass then feel free to merge yourself.

PR #6427, opened 2025-05-17
DBZ-9052 Add validation for KafkaSignalChannel

Naros:
https://issues.redhat.com/browse/DBZ-9052


jpechane:
@Naros Applied, thanks

PR #6429, opened 2025-05-20
DBZ-7996 Part 1 moving to a shared sink connector framework

rk3rn3r:
relates to https://issues.redhat.com/browse/DBZ-7996


Naros:
@rk3rn3r can you do me a favor and please change the package `value_bind` to `binding`?  We really shouldn't have package names using underscores.

rk3rn3r:
@Naros Sure. Just one question on that: what is the reason behind not having underscore names? In general they are valid and it is used. Any reference on the downsides?

I was wondering, it seems parts of our code are Java 21 already. I thought it is active for all newer things. It seems it isn't. Can I make JDBC sink Java 21 or do I have to keep 17? 
How do we usually do this? From what I found 21 is standard and 17 should be active only for the older components (connectors etc)?

Naros:
Hi @rk3rn3r, regarding underscores, we don't use this style in the code base. For example, in the core module under `io.debezium.transforms`, you'll see we have `extractnewstate`, not `extract_new_state`. We should be consistent.

Regarding your question about Java versions other than Debezium Server, Operator, and Platform, all the remaining parts of the code base must target Java 17. The JDBC sink is part of the downstream builds, and we cannot change the target version for the sink. It must remain Java 17 as the target bytecode format.

jpechane:
@rk3rn3r Connectors are Java 17 - as Kafka Connect still runs with Java 17. Non-connectors are Java 21.
WRT underscore names - it is a kind of home rule estabslihed long time ago. And TBH I personally consider package names ugly. I understand that this is personal preference but I'd refrain from underscores to keep the codebase consistent.

rk3rn3r:
I feel you are personally biased and I think you should evaluate your opinion, because there are cases it can make sense when resulting names can be easier to understand.
But sure, I see consistency as a good reason.
But search your feelings on that! ;) :alien: (there is no :yoda: emoji unfortunately)

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/ReducedRecordBuffer.java
```diff
@@ -105,7 +105,8 @@ public void remove(JdbcSinkRecord record) {
             return;
         }
 
-        Struct keyStruct = record.getKeyStruct(connectorConfig.getPrimaryKeyMode(), connectorConfig.getPrimaryKeyFields());
+        // Struct keyStruct = record.getFilteredKey(connectorConfig.getPrimaryKeyMode(), connectorConfig.getPrimaryKeyFields(), connectorConfig.getFieldFilter());
```
Naros, line 108:
```suggestion
```

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/GeneralDatabaseDialect.java
```diff
@@ -712,13 +713,13 @@ protected String columnQueryBindingFromField(String fieldName, TableDescriptor t
         final ColumnDescriptor column = table.getColumnByName(columnName);
 
         final Object value;
-        if (record.getNonKeyFieldNames().contains(fieldName)) {
+        if (record.getPayload().schema().fields().stream().map(Field::name).toList().contains(fieldName)) {
```
Naros, line 716: This is called for every column when we construct an insert statement. I am concerned with the potential GC and overhead that constantly creating an intermediate list to apply a contains check may incur rather than the contains on a precomputed list we used previously.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/sqlserver/BytesType.java
```diff
@@ -30,20 +29,17 @@ class BytesType extends AbstractBytesType {
...
         final int columnSize = Integer.parseInt(getSourceColumnSize(schema).orElse("0"));
         if (columnSize > 0) {
-            return dialect.getTypeName(Types.VARBINARY, Size.length(columnSize));
+            return getDialect().getJdbcTypeName(Types.VARBINARY, Size.length(columnSize));
         }
-        else if (key) {
-            return dialect.getTypeName(Types.VARBINARY, Size.length(dialect.getMaxVarbinaryLength()));
-        }
-        return dialect.getTypeName(Types.VARBINARY);
+        return getDialect().getJdbcTypeName(Types.VARBINARY, Size.length(getDialect().getMaxVarbinaryLength()));
```
Naros, line 42: This is a behavior change. For non-key columns, the column will now be constructed as `VARBINARY(8000)` rather than `VARBINARY(MAX)`. This changes the potential of storing 2GB of bytes to only 8000 bytes.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/type/AbstractType.java
```diff
@@ -54,9 +54,9 @@ public String getDefaultValueBinding(DatabaseDialect dialect, Schema schema, Obj
             case FLOAT64:
                 return Double.toString((double) value);
             case STRING:
-                return "'" + ((String) value).replaceAll("(?<!')'(?!')", "''") + "'";
+                return "'" + value + "'";
```
Naros, line 57:
```suggestion
                return "'" + ((String) value).replaceAll("(?<!')'(?!')", "''") + "'";
```
This needs to use the old behavior, otherwise default values that contain `'` characters will be mishandled. This was a recent change that may have gotten lost in your fork.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/type/connect/ConnectBytesType.java
```diff
@@ -31,20 +31,20 @@ public String[] getRegistrationKeys() {
...
         final int columnSize = Integer.parseInt(getSourceColumnSize(schema).orElse("0"));
+        DatabaseDialect dialect = getDialect();
         if (columnSize > 0) {
-            return dialect.getTypeName(Types.VARBINARY, Size.length(columnSize));
+            return dialect.getJdbcTypeName(Types.VARBINARY, Size.length(columnSize));
         }
-        else if (key) {
-            return dialect.getTypeName(Types.VARBINARY, Size.length(dialect.getMaxVarbinaryLength()));
+        else {
+            return dialect.getJdbcTypeName(Types.VARBINARY, Size.length(dialect.getMaxVarbinaryLength()));
```
Naros, line 46: This will present the same issue as the SQL Server dialect with `VARBINARY` as 8000 vs MAX.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/type/connect/ConnectStringType.java
```diff
@@ -59,33 +60,25 @@ public String getTypeName(DatabaseDialect dialect, Schema schema, boolean key) {
...
         }
         else {
             final int jdbcType = hasNationalizedCharacterSet(schema) ? Types.NVARCHAR : Types.VARCHAR;
-            // The column propagation details either don't exist or did not map to a logical STRING type
-            // In this case, we apply the size defaults (no size for non-keys and max-key for keys).
-            if (key) {
-                return dialect.getTypeName(jdbcType, Size.length(getMaxSizeInKey(dialect, jdbcType)));
-            }
-            return dialect.getTypeName(jdbcType);
+            return dialect.getJdbcTypeName(jdbcType, Size.length(getMaxSizeInKey(dialect, jdbcType)));
```
Naros, line 69: Another case where non-key columns will be scoped to smaller sizes than previously.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/type/connect/ConnectStringType.java
```diff
@@ -59,33 +60,25 @@ public String getTypeName(DatabaseDialect dialect, Schema schema, boolean key) {
...
         int columnSize = Integer.parseInt(getSourceColumnSize(schema).orElse("0"));
-        if (key) {
-            final int maxSizeInKey = getMaxSizeInKey(dialect, jdbcType);
-            if (columnSize > 0) {
-                columnSize = Math.min(columnSize, maxSizeInKey);
-            }
-            else {
-                columnSize = maxSizeInKey;
-            }
+        final int maxSizeInKey = getMaxSizeInKey(dialect, jdbcType);
```
Naros, line 75: The max size in key should only be applied to key columns, not non-key columns.

debezium-sink/src/main/java/io/debezium/bindings/kafka/DebeziumSinkRecord.java
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.bindings.kafka;
+
+import org.apache.kafka.connect.data.Struct;
+
+public interface DebeziumSinkRecord extends io.debezium.sink.DebeziumSinkRecord {
```
Naros, line 10: Could we add some documentation here, given it's an interface?

debezium-sink/src/main/java/io/debezium/bindings/kafka/KafkaDebeziumSinkRecord.java
```diff
@@ -95,91 +244,181 @@ public boolean isTombstone() {
...
+        else {
+            if (isDelete()) {
+                return rawPayload.getStruct(Envelope.FieldName.BEFORE);
+            }
+            return rawPayload.getStruct(Envelope.FieldName.AFTER);
+        }
+
+        // if (!configuredPrimaryKeyFields.isEmpty()) {
+        // recordFields = recordFields.filter(field -> configuredPrimaryKeyFields.contains(field.name()));
+        // }
```
Naros, line 421:
```suggestion
```

Naros (changes requested):
Hi @rk3rn3r, a lot of good stuff in here with the contract cleanup you've done, especially around the type system :tada: .  

I've left a few comments and identified where I saw concerns with the type system issues, keys vs. non-keys, and a few others.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/RecordWriter.java
```diff
@@ -118,39 +121,39 @@ private void bindValues(JdbcSinkRecord record, QueryBinder queryBinder) {
...
-        return bindFieldValuesToQuery(record, query, index, record.getPayload(), record.getNonKeyFieldNames());
+        return bindFieldValuesToQuery(record, query, index, record.getPayload(), record.nonKeyFieldNames());
     }
 
-    private int bindFieldValuesToQuery(JdbcSinkRecord record, QueryBinder query, int index, Struct source, List<String> fields) {
-        for (String fieldName : fields) {
-            final JdbcSinkRecord.FieldDescriptor field = record.allFields().get(fieldName);
+    private int bindFieldValuesToQuery(JdbcSinkRecord record, QueryBinder query, int index, Struct source, Set<String> fields) {
+        if (source.schema().type().isPrimitive()) {
+            List<ValueBindDescriptor> boundValues = dialect.bindValue(record.jdbcFields().get(fields.iterator().next()), index, source);
```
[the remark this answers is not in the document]
rk3rn3r: Do you mean `fields.forEach(..)`?

I think the main reason was that prior I changed `keyFieldNames()` and `nonKeyFieldNames()` to `LinkedHashMap` the fields were out of order. That way the ordering was broken and tests check for the correct expected ordering of field / field names. I think  since I fixed ordering there I could give this a try again! :+1:

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/ReducedRecordBuffer.java
rk3rn3r: just a leftover. already remove.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/type/AbstractType.java
rk3rn3r: Yep. Merged from the other dev branch incorrectly.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/type/connect/ConnectBytesType.java
rk3rn3r: wdym?
It's the same code as before (with the new method name that collide with the same name used in here from another class)

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/type/connect/ConnectStringType.java
rk3rn3r: It's the same logic as before, only head and tail inverted, no?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/sqlserver/BytesType.java
rk3rn3r: Yes, that's why I asked you on Zulip y'day about the key field handling. I re-introduced key handling this morning.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/type/connect/ConnectStringType.java
rk3rn3r: key handling re-introduced. see above.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/type/connect/ConnectStringType.java
rk3rn3r: same

debezium-sink/src/main/java/io/debezium/bindings/kafka/DebeziumSinkRecord.java
rk3rn3r: Done

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/type/connect/ConnectBytesType.java
Naros: It is now after adding the `isKey` argument.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/type/connect/ConnectStringType.java
Naros: With the new `isKey` argument, it is now :+1:

Naros:
Just responded @rk3rn3r, there is still an issue with primitive key bindings. Please see my comment.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/RecordWriter.java
```diff
@@ -118,39 +121,38 @@ private void bindValues(JdbcSinkRecord record, QueryBinder queryBinder) {
...
-        return bindFieldValuesToQuery(record, query, index, record.getPayload(), record.getNonKeyFieldNames());
+        return bindFieldValuesToQuery(record, query, index, record.getPayload(), record.nonKeyFieldNames());
     }
 
-    private int bindFieldValuesToQuery(JdbcSinkRecord record, QueryBinder query, int index, Struct source, List<String> fields) {
-        for (String fieldName : fields) {
-            final JdbcSinkRecord.FieldDescriptor field = record.allFields().get(fieldName);
+    private int bindFieldValuesToQuery(JdbcSinkRecord record, QueryBinder query, int index, Struct source, Set<String> fieldNames) {
+        if (source.schema().type().isPrimitive()) {
+            List<ValueBindDescriptor> boundValues = dialect.bindValue(record.jdbcFields().get(fieldNames.iterator().next()), index, source);
```
rk3rn3r, line 137: here is `dialect.bindValue(..)` for the primitive

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/RecordWriter.java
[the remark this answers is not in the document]
rk3rn3r: I don't know if I get it. I call `here is `dialect.bindValue(..)` for the primitive in the if-body on line 137 and also `dialect.bindValue(..)` for the Struct fields on line 151 (see below).

I don't know what you mean is missing. all tests are green (tests, ITs, e2e-mysql).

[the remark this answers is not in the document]
rk3rn3r: Gotya now after [#dev > test specifics in JDBC sink tests @ 💬](https://debezium.zulipchat.com/#narrow/channel/302533-dev/topic/test.20specifics.20in.20JDBC.20sink.20tests/near/519586959) and ff

rk3rn3r:
still running on i9-14900K :face_exhaling: :clock1130:

rk3rn3r:
I ran out of network addresses/ports when finally Oracle tests started :facepalm:

rk3rn3r:
`test.tags=it-oracle` were green, `-Dsource.connectors=mysql,postgres -Dtest.tags=e2e-oracle` was green too

rk3rn3r:
@Naros @jpechane @mfvitale should be ready for final review

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/transforms/ConvertCloudEventToSaveableForm.java
```diff
@@ -1,266 +0,0 @@
-/*
```
rk3rn3r, line 1: @rkudryashov You might want to have a look at the removal of the ConvertCloudEventToSaveableForm SMT.

The result of the JDBC sink is now aligned with the conventional output of (sink) connectors. Basically it extracts the real payload (flattened or Debezium Envelope) from the `data` field of the CloudEvents message. Similar to handling the Debezium Envelope.
If you need a different structure of your events (that SMT felt like it solves your specific problem and that should not be in the core repo) please use other SMT combinations like ReplaceField for renaming etc or a custom SMT, or stream processing, to re-structure events like you need them.

Still I appreciate your feedback!

debezium-sink/src/main/java/io/debezium/bindings/kafka/KafkaDebeziumSinkRecord.java
```diff
@@ -5,30 +5,63 @@
...
 import io.debezium.data.Envelope;
-import io.debezium.sink.DebeziumSinkRecord;
 import io.debezium.sink.SinkConnectorConfig;
+import io.debezium.sink.field.FieldDescriptor;
+import io.debezium.sink.filter.FieldFilterFactory;
 import io.debezium.util.Strings;
 
 @Immutable
-public class KafkaDebeziumSinkRecord implements DebeziumSinkRecord {
+public class KafkaDebeziumSinkRecord extends Values implements DebeziumSinkRecord {
```
Naros, line 31: I would prefer that we drop `Values` here in the class and instead create an inner class that extends `Values`, moving the `inferSchemaFromMap` static method to the inner class, making it a private concern. That avoids exposing all the public static methods that `Values` exposes, which is unnecessary.

debezium-sink/src/main/java/io/debezium/bindings/kafka/KafkaDebeziumSinkRecord.java
```diff
@@ -47,8 +80,17 @@ public long offset() {
...
+    public Struct kafkaCoordinates() {
+        return kafkaCoordinates;
+    }
+
+    @Override
+    public Struct kafkaHeader() {
+        return kafkaHeader;
+    }
+
+    public Map<String, FieldDescriptor> allFields() {
```
Naros, line 92:
```suggestion
    @Override
    public Map<String, FieldDescriptor> allFields() {
```

debezium-sink/src/main/java/io/debezium/bindings/kafka/KafkaDebeziumSinkRecord.java
```diff
@@ -95,91 +244,177 @@ public boolean isTombstone() {
...
+            }
+            case RECORD_HEADER -> {
+                if (originalKafkaRecord.headers().isEmpty()) {
+                    throw new ConnectException("Configured primary key mode 'record_header' cannot have empty message headers");
+                }
+                return getRecordHeaders();
+            }
+            case KAFKA -> {
+                return kafkaCoordinates();
+            }
```
Naros, line 350: What do you think about adding a default handler in the switch that raises an exception so that if we add any new primary key modes in the future, we don't mistakenly cause NullPointerExceptions?

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/DebeziumContainer.java
```diff
@@ -90,7 +91,9 @@ private static String lazilyRetrieveAndCacheLatestStable() {
     }
 
     public static DebeziumContainer nightly() {
-        return new DebeziumContainer(String.format("%s:%s", DEBEZIUM_CONTAINER, DEBEZIUM_NIGHTLY_TAG));
+        return new DebeziumContainer(String.format("%s:%s", DEBEZIUM_CONTAINER, DEBEZIUM_NIGHTLY_TAG))
+                .withImagePullPolicy(PullPolicy.ageBased(Duration.ofDays(1)));
```
Naros, line 95: This change is unrelated to this specific work; I'd suggest that if this is needed, we do this in a separate PR.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/GeneralDatabaseDialect.java
```diff
@@ -260,12 +258,15 @@ public String getCreateTableStatement(JdbcSinkRecord record, CollectionId collec
         builder.append(getQualifiedTableName(collectionId));
         builder.append(" (");
 
+        Map<String, FieldDescriptor> allFields = record.allFields();
```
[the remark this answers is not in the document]
rk3rn3r: I think you mistake this a bit with jdbcFields which should be the relevant member for write operations, but I  see someone else could do that kind of mistake in the future. Basically `allFields` should be a registry for all seen fields for the record. I addressed this with a new commit.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/type/Type.java
```diff
@@ -22,7 +18,8 @@
  *
  * @author Chris Cranford
  */
-public interface Type {
+public interface Type extends io.debezium.sink.type.Type {
```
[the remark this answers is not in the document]
rk3rn3r: `jdbc.type.JdbcType` 

;D 

as you command

debezium-sink/src/main/java/io/debezium/bindings/kafka/KafkaDebeziumSinkRecord.java
rk3rn3r: Good point.

debezium-sink/src/main/java/io/debezium/bindings/kafka/DebeziumSinkRecord.java
```diff
@@ -0,0 +1,28 @@
...
+ */
+package io.debezium.bindings.kafka;
+
+import org.apache.kafka.connect.data.Struct;
+
+/**
+ * A Kafka-specific representation of a {@link io.debezium.sink.DebeziumSinkRecord} that offers access to Kafka message headers and coordinates (offset, partition, offset).
+ * @author rk3rn3r
+ */
+public interface DebeziumSinkRecord extends io.debezium.sink.DebeziumSinkRecord {
```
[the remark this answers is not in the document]
rk3rn3r: But here I would say it is intended, because this is inside the Kafka binding and basically the SAME interface with some Kafka specific extensions. But I'll keep the comment unresolved and let's follow-up after my PTO in 3 weeks.

debezium-sink/src/main/java/io/debezium/bindings/kafka/KafkaDebeziumSinkRecord.java
rk3rn3r: Done and also other `@Override` annotations added where they were missing.

debezium-sink/src/main/java/io/debezium/bindings/kafka/KafkaDebeziumSinkRecord.java
rk3rn3r: Done

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/DebeziumContainer.java
rk3rn3r: No, it is/was not. It was related to test errors when this is not added. A previously fetched old nightly image (in my case from 3.0.0-SNAPSHOT) broke my builds and caused trouble.
Please accept it as part of this PR because of my PTO I can only come back to this in 3 weeks. I am off for the next 2 weeks. Thank you!

rk3rn3r:
Thx @Naros!
I applied the requested changes directly.

It would be really nice if you could merge it until Monday noon latest. I am on PTO now for 2 weeks and I don't like coming back in 3 weeks to another updated codebase that is hard to rebase. I can have a final look on Friday evening, Saturday, Sunday morning or Monday till afternoon (all my time).

debezium-sink/src/main/java/io/debezium/bindings/kafka/DebeziumSinkRecord.java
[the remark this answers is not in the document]
Naros: Sounds good to me.

PR #6430, opened 2025-05-20
DBZ-9020: OpenLineage integration

mfvitale:
Adds support for OpenLineage

The lineage graph will be:

image

Todos:
- [ ] Send running event with input dataset info for MongoDB
- [ ] Add an SMT to provide the send information about the output dataset
- [ ] Move `openlineage.integration.*` properties to CommonConnectorConfig
- [ ] Address minor code todos

closes: https://issues.redhat.com/browse/DBZ-9020


debezium-core/pom.xml
```diff
@@ -56,6 +56,13 @@
             <scope>provided</scope>
         </dependency>
 
+        <!-- OpenLineage Integration -->
+        <dependency>
+            <groupId>io.openlineage</groupId>
+            <artifactId>openlineage-java</artifactId>
+            <scope>provided</scope>
```
mfvitale, line 63: I preferred to put this in `provided` scope so that by default we will not include the `openlineage-java` dependency and also its transitive deps.

debezium-core/src/main/java/io/debezium/openlineage/dataset/DefaultOutputDatasetNamespaceResolver.java
```diff
@@ -0,0 +1,27 @@
...
+
+import java.util.function.Function;
+
+import io.debezium.config.Configuration;
+
+public class DefaultOutputDatasetNamespaceResolver implements OutputDatasetNamespaceResolver {
+
+    private static final String LIST_SEPARATOR = ",";
+    private static final String KAFKA_NAMESPACE_FORMAT = "kafka://%s";
+    private static final String SCHEMA_HISTORY_INTERNAL_KAFKA_BOOTSTRAP_SERVERS_PROPERTY = "schema.history.internal.kafka.bootstrap.servers";
```
mfvitale, line 16: @jpechane I don't recall if this is in some way required for KC connectors. Can you confirm?

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -238,6 +240,10 @@ public final void start(Map<String, String> props) {
         try {
             setTaskState(State.INITIAL);
             config = Configuration.from(props);
+
+            DebeziumOpenLineageEmitter.init(config, connectorName());
```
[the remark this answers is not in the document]
jpechane: Also, what is expected behaviour in case of multi-task connectors?
Will not it flip-flop between the states depedning on the state of each task?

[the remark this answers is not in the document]
kmos: The OpenLineage integration is the motivation behind #6458 : the need for a registry that can be used in `BaseSourceTask`. IMHO, I like the idea of using `ConnectorTaskLifecycleListener` along with a registry.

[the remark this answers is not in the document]
mfvitale: > Or an altenrative approach might be re-using of https://github.com/debezium/debezium/pull/6458
If there is a way how to use the registry to implement that kind of listener then I think we should keep it as is right now and then modify it based on the registry. WDYT?

Yes, having the `ConnectorTaskLifecycleListener` and the different implementations managed with the registry will be the best option. 

I created a Jira to track it https://issues.redhat.com/browse/DBZ-9109

> Also, what is expected behaviour in case of multi-task connectors?

Well, I forgot about this. Maybe we can, in case of multiple tasks, emitting a multiple job just adding the "-#tasknumber" as suffix. WDYT?

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -1097,6 +1097,59 @@ public static SnapshotQueryMode parse(String value, String defaultValue) {
             .withDescription("The maximum time in milliseconds to wait for connection validation to complete. Defaults to 60 seconds.")
             .withValidation(Field::isPositiveLong);
 
+    public static final Field OPENLINEAGE_INTEGRATION_ENABLED = Field.create("openlineage.integration.enabled")
```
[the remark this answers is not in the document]
mfvitale: Well touches the topic to have configuration in a single place or spread it. IIRC there was some discussion also with @Naros on this.

@jpechane Do you mean just to move there config in a separate class and the reference them in the common config?

debezium-core/src/main/java/io/debezium/openlineage/dataset/DefaultDatasetNamespaceResolverFactory.java
```diff
@@ -0,0 +1,25 @@
...
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.openlineage.dataset;
+
+public class DefaultDatasetNamespaceResolverFactory implements DatasetNamespaceResolverFactory {
+
+    private static final String MONGODB = "mongodb";
+    private static final String POSTGRESQL = "postgresql";
+
+    public InputDatasetNamespaceResolver createInput(String connectorName) {
```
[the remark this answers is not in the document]
mfvitale: Honestly I don't see the advantage here to use the `ServiceLoader` since I don't expect any further implementations here and also there will not be any custom implementation from outside. 

Maybe we can revisit with the registry. WDYT?

debezium-core/src/main/java/io/debezium/openlineage/dataset/DefaultOutputDatasetNamespaceResolver.java
mfvitale: So for the KC deployment this is the only option, right?

debezium-core/src/main/java/io/debezium/relational/RelationalDatabaseSchema.java
```diff
@@ -121,9 +127,19 @@ protected void buildAndRegisterSchema(Table table) {
         if (tableFilter.isIncluded(table.id())) {
             TableSchema schema = schemaBuilder.create(topicNamingStrategy, table, columnFilter, columnMappers, customKeysMapper);
             schemasByTableId.put(table.id(), schema);
+            DebeziumOpenLineageEmitter.emit(BaseSourceTask.State.RUNNING, List.of(extractDatasetMetadata(table)));
```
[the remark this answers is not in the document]
mfvitale: As for the task it will be addressed in a separate PR

debezium-core/src/main/java/io/debezium/transforms/openlineage/OpenLineage.java
```diff
@@ -0,0 +1,109 @@
...
+        return new ConfigDef();
+    }
+
+    @Override
+    public void configure(Map<String, ?> props) {
+    }
+
+    @Override
+    public R apply(R record) {
+
```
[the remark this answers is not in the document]
mfvitale: Well, them could be good to emit the "running" state but for sure not the relative output dataset. 
Do you recall if we already have some utility for filtering out these events?

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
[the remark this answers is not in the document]
jpechane: @mfvitale Sounds good!

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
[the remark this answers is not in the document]
jpechane: @mfvitale Yes, it from my PoV necessary step for potential undbundling of OLNG into a separate package, in a similar way as is done for storages.

debezium-core/src/main/java/io/debezium/openlineage/dataset/DefaultDatasetNamespaceResolverFactory.java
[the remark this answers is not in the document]
jpechane: Yes, registry sounds like the best solution now.

debezium-core/src/main/java/io/debezium/transforms/openlineage/OpenLineage.java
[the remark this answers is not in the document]
jpechane: @mfvitale There is `io.debezium.transforms.SmtManager<R>`. It probably could be split into a specific utility class not associated exclusively with SMTs and it could be extended to contain other message type checks.

debezium-core/src/main/java/io/debezium/openlineage/dataset/DefaultOutputDatasetNamespaceResolver.java
jpechane: Usually yes

mfvitale:
@jpechane I added the support for multi task connectors. As discussed the solution uses record headers.

mfvitale:
@jpechane Failure of PostgreSQL 17 seems to be unrelated.

debezium-core/src/main/java/io/debezium/openlineage/ConnectorContext.java
```diff
@@ -0,0 +1,109 @@
...
+     * Creates a ConnectorContext instance from a configuration object and connector type name.
+     *
+     * <p>This factory method extracts the topic prefix and task ID from the provided configuration.
+     * If no task ID is found in the configuration, it defaults to "0".</p>
+     *
+     * @param configuration the configuration object containing connector settings
+     * @param connectorTypeName the name of the connector type
+     * @return a new ConnectorContext instance populated with configuration values
+     */
+    public static ConnectorContext from(Configuration configuration, String connectorTypeName) {
```
jpechane, line 52: Whey can't you pass `CommonConnectorConfig` as both parameters come from it?

debezium-core/src/main/java/io/debezium/connector/common/CdcSourceTaskContext.java
```diff
@@ -42,7 +43,8 @@ public CdcSourceTaskContext(CommonConnectorConfig config,
                                 Map<String, String> customMetricTags,
                                 Supplier<Collection<? extends DataCollectionId>> collectionsSupplier) {
         this.connectorType = config.getContextName();
-        this.connectorName = config.getLogicalName();
+        this.connectorLogicalName = config.getLogicalName();
```
jpechane, line 46: This is now quite confusing. Connector name reflects the connector type itself and connector logical name reflects - well the topic prefix aka logical name of a connector instance. I believe we need to differentiate between those. Maybe use something like `getConnectorInstanceLogicalName()`?

debezium-core/src/main/java/io/debezium/connector/common/DebeziumHeaderProducer.java
```diff
@@ -0,0 +1,35 @@
...
+import static io.debezium.connector.common.DebeziumHeaders.DEBEZIUM_CONNECTOR_LOGICAL_NAME_HEADER;
+import static io.debezium.connector.common.DebeziumHeaders.DEBEZIUM_CONNECTOR_NAME_HEADER;
+import static io.debezium.connector.common.DebeziumHeaders.DEBEZIUM_TASK_ID_HEADER;
+
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.header.ConnectHeaders;
+
+import io.debezium.service.Service;
+
+public class DebeziumHeaderProducer implements Service {
```
jpechane, line 27: Please add JavaDoc describing the headers.

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/transforms/ExtractNewDocumentStateTestIT.java
```diff
@@ -1238,7 +1238,7 @@ public void shouldPropagatePreviousRecordHeaders() throws InterruptedException {
         // Perform transformation
         final SourceRecord transformed = transformation.apply(record);
 
-        assertThat(transformed.headers()).hasSize(1);
+        assertThat(transformed.headers()).hasSize(4);
```
jpechane, line 1241: I think for future refernce in this and other places the count of headers should not be hardcoded but conturcted dynamically - there should be a method to which for example 1 is added and it returns 1 + Debezium standard headers.

debezium-core/src/main/java/io/debezium/openlineage/ConnectorContext.java
```diff
@@ -0,0 +1,109 @@
...
+ * @author Mario Fiore Vitale
+ */
+public record ConnectorContext(String connectorLogicalName, String connectorName, String taskId, Configuration config) {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ConnectorContext.class);
+
+    private static final String KEY_FORMAT = "%s:%s";
+
+    /** Configuration key for the topic prefix setting */
+    public static final String TOPIC_PREFIX = "topic.prefix";
```
jpechane: This should be initialize for CommonConnectorConfig config field not hardcoded.

jpechane (approved):
@mfvitale We are almost there, thanks! I left only few minor comments. When addressed feel free to merge yourself.

debezium-core/src/main/java/io/debezium/openlineage/ConnectorContext.java
mfvitale: Well this is used in the `DebeziumOpenLineageEmitter.init` that is used in the `BaseSourceTask.start(Map<String, String> props)`. We effectively create the `CommonConnectorConfig` in the concrete source tasks.

jpechane: Makes sense

debezium-core/src/main/java/io/debezium/connector/common/CdcSourceTaskContext.java
mfvitale: What we had before

`connectorName` -> `topic.prefix` ->  `my.prefix`
`connectorType` -> `Module.contextName()` -> `Postgres`

what we have now

`connectorName` -> `Module.name()` -> `postgresql`
`connectorType` -> `Module.contextName()` -> `Postgres`
`connectorLogicalName` -> `topic.prefix` ->  `my.prefix`

I think that what create confusion is that `connectorName` should effectively be the Kafka Connect `name` value . WDYT?

jpechane: So how about channging `connectorName` to `pluginName` or `connectorPluginName`?

mfvitale: I'm fine with `connectorPluginName`. Do we want to change this only in the `CdcSourceTaskContext` or also in `CommonConnectorConfig`?

and well there is a long list of places where `connectorName` is used...

So I think we should change it also in `CommonConnectorConfig` but I would like to do it a different PR. WDYT?

mfvitale: and well there is a long list o places where `connectorName` is used...

jpechane: Only here, there is no need to do codebase wide change right now.

dolfinus:
Hi. Can I create a pull request with some changes to this implementation? I see that all PRs here are created with some Jira issue bound, but I don't have access to create Jira tasks.

PR #6432, opened 2025-05-20
DBZ-8562 - Move schema history recovery out of task start

akanimesh7:
## Problem
https://issues.redhat.com/browse/DBZ-8562

In most production scenarios, there is an operator which tracks the state of the task and re-starts it when it is failed. Also it might do it at certain frequency (let's say a minute). Schema history recovery takes place inside the task's `start` method. Once this `start` metod is complete, the task is transitioned to RUNNING state. Hence, if the schema history recovery is not complete (let's say it takes 10 minutes), then after a minute operator would re-start the task and hence, it would create another task thread. The previous task thread would still be running for 9 more minutes.

This can lead to thread leak, high CPU usage and other issues in production environments. The reason why this happens is connect runtime's assumption/contract for task's `start` method is that it would be used to initialize certain objects needed and would finish early. 

## Solution

To solve the above issue, we are moving the schema history recovery out of task's `start` method. Rather it is placed before snapshotting or streaming starts in the ChangeEventSourceCoordinator. By doing this, the start method finishes early, hence task moves to RUNNING state quickly and is not re-started unnecessarily by state tracking mechanisms like an opeator.

### DIFF
--- a/debezium-core/src/main/java/io/debezium/relational/history/SchemaHistory.java
+++ b/debezium-core/src/main/java/io/debezium/relational/history/SchemaHistory.java
@@ -164,7 +164,7 @@ public interface SchemaHistory {
      * @deprecated Use {@link #recover(Offsets, Tables, DdlParser)} instead.
      */
     @Deprecated
-    default void recover(Map<String, ?> source, Map<String, ?> position, Tables schema, DdlParser ddlParser) {
+    default void recover(Map<String, ?> source, Map<String, ?> position, Tables schema, DdlParser ddlParser) throws InterruptedException {
         recover(Collections.singletonMap(source, position), schema, ddlParser);
     }
 
@@ -181,7 +181,7 @@ public interface SchemaHistory {
      *            may not be null
      * @param ddlParser the DDL parser that can be used to apply DDL statements to the given {@code schema}; may not be null
      */
-    default void recover(Offsets<?, ?> offsets, Tables schema, DdlParser ddlParser) {
+    default void recover(Offsets<?, ?> offsets, Tables schema, DdlParser ddlParser) throws InterruptedException {
         Map<Map<String, ?>, Map<String, ?>> offsetMap = new HashMap<>();
         for (Entry<? extends Partition, ? extends OffsetContext> entry : offsets) {
             if (entry.getValue() != null) {
@@ -196,7 +196,7 @@ public interface SchemaHistory {
      * @deprecated Use {@link #recover(Offsets, Tables, DdlParser)} instead.
      */
     @Deprecated
-    void recover(Map<Map<String, ?>, Map<String, ?>> offsets, Tables schema, DdlParser ddlParser);
+    void recover(Map<Map<String, ?>, Map<String, ?>> offsets, Tables schema, DdlParser ddlParser) throws InterruptedException;
 
     /**
      * Stop recording history and release any resources acquired since {@link #configure(Configuration, HistoryRecordComparator, SchemaHistoryListener, boolean)}.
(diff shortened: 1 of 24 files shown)


debezium-storage/debezium-storage-kafka/src/main/java/io/debezium/storage/kafka/history/KafkaSchemaHistory.java
```diff
@@ -316,13 +322,16 @@ protected void recoverRecords(Consumer<HistoryRecord> records) {
                             "The database schema history couldn't be recovered. Consider to increase the value for " + RECOVERY_POLL_INTERVAL_MS.name());
                 }
 
+                checkForInterruption();
```
akanimesh7, line 325: The schema history recovery is moved out of task start and is now happening in the change-event-source-coordinator thread. If recovery time is high and task's stop method is called, it would send an interrupt to the change-event-source-coordinator thread(`Executor.shutdownNow`). It is important to catch this interrupt early and terminate the recovery process so that this thread is terminated. `checkForInterruption` has been put at certain places in the code where we are calling external code (KafkaConsumer). Reason being, if the thread is interrupted and is not handled properly (we can't be sure about interrupt handling in external code), the interrupt could be missed/cleared and thread would keep on running forever.

debezium-core/src/main/java/io/debezium/relational/history/AbstractFileBasedSchemaHistory.java
```diff
@@ -124,7 +124,19 @@ protected void storeRecord(HistoryRecord record) throws SchemaHistoryException {
 
     @Override
     protected void recoverRecords(Consumer<HistoryRecord> records) {
-        lock.write(() -> getRecords().forEach(records));
+        lock.write(() -> {
+            for (HistoryRecord record : getRecords()) {
+                records.accept(record);
+                try {
+                    LOGGER.info("Sleeping for {} ms to emulate CPU-intensive environment", config.getLong(SCHEMA_HISTORY_RECOVERY_DELAY_MS));
```
jpechane, line 131: This class is used for non-test components too so it must not have something like it hard-coded. For testing a new test only schema history implementation should be provided.

debezium-core/src/main/java/io/debezium/relational/history/SchemaHistory.java
```diff
@@ -53,6 +53,15 @@ public interface SchemaHistory {
                     + "which it cannot parse. If skipping is enabled then Debezium can miss metadata changes.")
             .withDefault(false);
 
+    // Not intended to be used in production code. This is only for testing.
+    Field SCHEMA_HISTORY_RECOVERY_DELAY_MS = Field.createInternal(CONFIGURATION_FIELD_PREFIX_STRING + "recovery.delay.ms")
```
jpechane, line 57: I'd like to have not this intorduce it here at all.

jpechane (changes requested):
@akanimesh7 Thanks for the PR. It looks good but I am a bit unsure about having the test code being place in `AbstractFileBasedSchemaHistory` that is used as base class for production components.
I recommend to remove the introduced change and provide test only implementation of the `SchemaHistory` that would emulate the timeouts needed for testing.

akanimesh7:
https://github.com/debezium/debezium-connector-informix/pull/83
and 
https://github.com/debezium/debezium-connector-db2/pull/182
are PRs for informix and db2 respectively. 

@jpechane Do we need to first merge this PR for the other two to progress?

jpechane:
@akanimesh7 LGTM! Could you please just rebase on the latest main so I can do the clean merge?
I'll take care of Informix/Db2 failurues (testing).
Thanks

akanimesh7:
@jpechane resolved conflicts

jpechane:
@akanimesh7 Applied, thanks!

PR #6436, opened 2025-05-21
DBZ-9056 Remove embeddgins prefix from embeddgins SMT config

vjuranek:
https://issues.redhat.com/browse/DBZ-9056


jpechane:
@vjuranek Should we for ollama do deprecation instead of hard removal?

vjuranek:
> @vjuranek Should we for ollama do deprecation instead of hard removal?

good point, agree we should

vjuranek:
And yes, I should set CI to run also for `debezium-ai` moule.

jpechane:
@vjuranek Please rabse on the latest main. There should be no need for workaround now. Also the CI changes were merged too.

vjuranek:
> There should be no need for workaround now. 

unfortunately this is not the case - DBZ-9060 fixed only the issue that `deprecatedAliases()` doesn't have to be the last one config call on the field, but the issue with validator is still there (I haven't even file a bug for it yet - will do once I'm done with the documentation PR) and the main issue is in SMT config validation, which is done based on the error messages and not on based on the validator return value - here I'm not completely sure yet if it's a bug or not as it's little bit awkward how to propagate validator return value (`deprecatedAliases()` adds warning to the messages that value is deprecated, which is later on considered as an error and validation fails).

PR #6444, opened 2025-05-22
DBZ-9000 Improve connector-level transaction filtering

### LINKED ISSUE #1360: Unify LogMiner transaction filtering semantics [DBZ-9000]
Migrated from [DBZ-9000](https://issues.redhat.com/browse/DBZ-9000)

The Oracle LogMiner unbuffered implementation processes transactions without interweaving them, which means the implementation can use a single boolean flag to track whether a transaction is skipped. This flag is set at the {{START}} event and reset when the {{COMMIT}} event is observed. 

However, the buffered implementation receives transactions interwoven, which means that a single boolean flag won't be sufficient and that we'll need to track skipped transactions by transaction ID in a HashSet or equivalent container that provides fast lookup. It may even be helpful to wrap this with a String equality check for situations where we get multiple events back-to-back for the same transaction.

This would allow us to avoid buffering transactions where the connector does not enable database-level filtering. Still, we want to avoid unnecessarily using the heap to store a transaction that will ultimately be discarded at commit time.

Naros:
Fixes debezium/dbz#1360 

https://issues.redhat.com/browse/DBZ-9000

When using LogMiner in buffered mode paired with connector-level filtering, the connector would previously perform the filter at commit time to unnecessarily buffer the entire transaction to discard it later.

This change moves that filtering logic to the start of the transaction so it can be discarded immediately if the username or client id filters apply.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
@@ -34,6 +34,7 @@ import io.debezium.connector.oracle.logminer.LogMinerChangeRecordEmitter;
 import io.debezium.connector.oracle.logminer.LogMinerStreamingChangeEventSourceMetrics;
 import io.debezium.connector.oracle.logminer.SqlUtils;
 import io.debezium.connector.oracle.logminer.TransactionCommitConsumer;
+import io.debezium.connector.oracle.logminer.buffered.ProcessedTransaction.ProcessType;
 import io.debezium.connector.oracle.logminer.buffered.ehcache.EhcacheCacheProvider;
 import io.debezium.connector.oracle.logminer.buffered.ehcache.EhcacheTransactionFactory;
 import io.debezium.connector.oracle.logminer.buffered.infinispan.EmbeddedInfinispanCacheProvider;
@@ -253,7 +254,7 @@ public class BufferedLogMinerStreamingChangeEventSource extends AbstractLogMiner
     }
 
     @VisibleForTesting
-    protected LogMinerCache<String, String> getProcessedTransactionsCache() {
+    protected LogMinerCache<String, ProcessedTransaction> getProcessedTransactionsCache() {
         return cacheProvider.getProcessedTransactionsCache();
     }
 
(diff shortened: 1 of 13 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -674,6 +685,16 @@ private Scn calculateNewStartScn(Scn startScn, Scn endScn, Scn maxCommittedScn)
...
     }
 
+    /**
+     * Marks the specific transaction as skipped.
+     *
+     * @param transactionId the transaction identifier, should not be {@code null}
+     * @param eventScn the transaction's starting system change number, should not be {@code null}
+     */
+    protected void markTransactionSkipped(String transactionId, Scn eventScn) {
+        getProcessedTransactionsCache().put(transactionId, eventScn.toString());
```
jpechane: @Naros I wonder if we just want to put the transaction into processed cache. Would not it be better for potential future debugging also record the reason for putting into the cache -either finalized or skipped?

jpechane: @Naros Any opinion?

Naros: That's a good question, :thinking: 

My only reservation would be that it requires the introduction of another marshaller to serialize an object as the value in the Map rather than a native String, which might add some overhead, tho it might be negligible. So, we either do that or use a separate Map for this _skipped_ use case and let the code differentiate the two, thereby avoiding the marshaller requirement.

Do you have any opinion? :wink:

jpechane: I'd say it depends :-). But for starter I'd probaly just use the cache so the code can be kept simple and only if we see some performance impact it might make sense to do separate paths.

Naros: Ok, replaced cache value with an object. Talk about boilerplate ;-)

vjuranek (approved):
LGTM

Naros:
Rebased on main.

Naros:
I am going to move this back to draft because this requires a bit more thinking. 

While the filter logic works, the logging has the potential to be overly excessive by comparison to the commit time filtering where the number of captured events is known and omits logging transactions with 0 captured events. This is not possible when we first consume the `START` event.

Naros:

```
git commit -s
```

Naros:
Closing as no longer planned.

PR #6452, opened 2025-05-26
DBZ-9066 Warn user about deprecated option via standard logger

vjuranek:
Other Debezium parts, e.g. `SmtManager` use validation error as a signal for invalida configuration and thus using deprecated option results into the failure.

https://issues.redhat.com/browse/DBZ-9066


Naros (commented):
The code changes LGTM, but it seems the test failures are related @vjuranek

vjuranek:
Fixed, now the failures are unrelated.

PR #6458, opened 2025-05-27
Move `ServiceRegistry` and `BeanRegistry` in `BaseSourceTask` 

kmos:
This PR is a draft of moving `beanRegistry` and `serviceRegistry` from `CommonConnectorConfig` to `BaseSourceTask`. It illustrates the overall changes required in the codebase.

@Naros fill free to add comments

### DIFF
--- a/debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorTask.java
+++ b/debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorTask.java
@@ -126,8 +126,7 @@ public class MariaDbConnectorTask extends BinlogSourceTask<MariaDbPartition, Mar
         getBeanRegistry().add(StandardBeanNames.VALUE_CONVERTER, valueConverters);
         getBeanRegistry().add(StandardBeanNames.OFFSETS, previousOffsets);
 
-        // Service providers
-        registerServiceProviders();
+        getServiceRegistry().registerServiceProvider(new MariaDbCharsetRegistryServiceProvider());
 
         final SnapshotterService snapshotterService = getServiceRegistry().tryGetService(SnapshotterService.class);
         final Snapshotter snapshotter = snapshotterService.getSnapshotter();
@@ -301,9 +300,4 @@ public class MariaDbConnectorTask extends BinlogSourceTask<MariaDbPartition, Mar
         return new MariaDbFieldReader(connectorConfig, new MariaDbCharsetRegistry());
     }
 
-    @Override
-    protected void registerServiceProviders() {
-        getServiceRegistry().registerServiceProvider(new MariaDbCharsetRegistryServiceProvider());
-        super.registerServiceProviders();
-    }
 }
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorTask.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorTask.java
@@ -118,9 +118,6 @@ public final class MongoDbConnectorTask extends BaseSourceTask<MongoDbPartition,
             getBeanRegistry().add(StandardBeanNames.DATABASE_SCHEMA, schema);
             getBeanRegistry().add(StandardBeanNames.OFFSETS, previousOffsets);
 
-            // Service providers
-            registerServiceProviders();
-
             final SnapshotterService snapshotterService = getServiceRegistry().tryGetService(SnapshotterService.class);
 
             final EventDispatcher<MongoDbPartition, CollectionId> dispatcher = new EventDispatcher<>(
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorTask.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorTask.java
@@ -116,8 +116,7 @@ public class MySqlConnectorTask extends BinlogSourceTask<MySqlPartition, MySqlOf
         getBeanRegistry().add(StandardBeanNames.VALUE_CONVERTER, valueConverters);
         getBeanRegistry().add(StandardBeanNames.OFFSETS, previousOffsets);
 
-        // Service providers
-        registerServiceProviders();
+        getServiceRegistry().registerServiceProvider(new MySqlCharsetRegistryServiceProvider());
 
         final SnapshotterService snapshotterService = getServiceRegistry().tryGetService(SnapshotterService.class);
         final Snapshotter snapshotter = snapshotterService.getSnapshotter();
@@ -285,10 +284,4 @@ public class MySqlConnectorTask extends BinlogSourceTask<MySqlPartition, MySqlOf
         return MySqlConnectorConfig.ALL_FIELDS;
     }
 
-    @Override
-    protected void registerServiceProviders() {
-        super.registerServiceProviders();
-
-        getServiceRegistry().registerServiceProvider(new MySqlCharsetRegistryServiceProvider());
-    }
 }
--- a/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
+++ b/debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
@@ -245,6 +245,10 @@ public abstract class BaseSourceTask<P extends Partition, O extends OffsetContex
             config = Configuration.from(props);
             this.beanRegistry = new DefaultBeanRegistry();
             this.serviceRegistry = new DefaultServiceRegistry(config, beanRegistry);
+            serviceRegistry.registerServiceProvider(new PostProcessorRegistryServiceProvider());
+            serviceRegistry.registerServiceProvider(new SnapshotLockProvider());
+            serviceRegistry.registerServiceProvider(new SnapshotQueryProvider());
+            serviceRegistry.registerServiceProvider(new SnapshotterServiceProvider());
 
             retriableRestartWait = config.getDuration(CommonConnectorConfig.RETRIABLE_RESTART_WAIT, ChronoUnit.MILLIS);
             // need to reset the delay or you only get one delayed restart
@@ -588,13 +592,6 @@ public abstract class BaseSourceTask<P extends Partition, O extends OffsetContex
 
     public List<NotificationChannel> getNotificationChannels() {
         return notificationChannels;
-    }
-
-    protected void registerServiceProviders() {
-        serviceRegistry.registerServiceProvider(new PostProcessorRegistryServiceProvider());
-        serviceRegistry.registerServiceProvider(new SnapshotLockProvider());
-        serviceRegistry.registerServiceProvider(new SnapshotQueryProvider());
-        serviceRegistry.registerServiceProvider(new SnapshotterServiceProvider());
     }
 
     public BeanRegistry getBeanRegistry() {
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
@@ -95,9 +95,6 @@ public class OracleConnectorTask extends BaseSourceTask<OraclePartition, OracleO
         getBeanRegistry().add(StandardBeanNames.VALUE_CONVERTER, valueConverters);
         getBeanRegistry().add(StandardBeanNames.OFFSETS, previousOffsets);
 
-        // Service providers
-        registerServiceProviders();
-
         final SnapshotterService snapshotterService = getServiceRegistry().tryGetService(SnapshotterService.class);
 
         validateRedoLogConfiguration(connectorConfig, snapshotterService);
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorTask.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorTask.java
@@ -130,9 +130,6 @@ public class PostgresConnectorTask extends BaseSourceTask<PostgresPartition, Pos
         getBeanRegistry().add(StandardBeanNames.VALUE_CONVERTER, valueConverter);
         getBeanRegistry().add(StandardBeanNames.OFFSETS, previousOffsets);
 
-        // Service providers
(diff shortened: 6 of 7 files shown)


debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
```diff
@@ -291,7 +290,6 @@ public MariaDbConnectorConfig(Configuration config) {
         this.snapshotLockingMode = SnapshotLockingMode.parse(config.getString(SNAPSHOT_LOCKING_MODE));
         this.snapshotLockingStrategy = new MariaDbSnapshotLockingStrategy(snapshotLockingMode);
 
-        getServiceRegistry().registerServiceProvider(new MariaDbCharsetRegistryServiceProvider());
```
Naros, line 294: So I am a bit concerned about this specific change.

The point behind the `ServiceProvider` was to define a way that you effectively _register_ a service that may or may not need to be created depending on the environment. But when the service is created, you would only ever have one copy in the service registry, and that same copy would be handed out to all interested parties.

The concrete problem with throwing this away is with `MariaDbCharsetRegistry`.

In a variety of places now, the code just instantiates a new copy of this registry class, but the impact of that is for each instance, you are reparsing the JSON file off disk, populating the same maps with that data once for each instance rather than reusing a single instance that the service registry is guaranteed to provide you.

So I think we should think about retaining the singular charset registry instance behavior.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorTask.java
```diff
@@ -227,28 +236,29 @@ protected ChangeEventSourceCoordinator<MariaDbPartition, MariaDbOffsetContext> s
...
-                        queue,
-                        snapshotterService),
+                mariaDbChangeEventSourceFactory,
                 new MariaDbChangeEventSourceMetricsFactory(streamingMetrics),
                 dispatcher,
                 schema,
                 signalProcessor,
                 notificationService,
-                snapshotterService);
+                snapshotterService, getBeanRegistry(), getServiceRegistry());
```
Naros, line 261:
```suggestion
                snapshotterService, 
                getBeanRegistry(), 
                getServiceRegistry());
```

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbSnapshotChangeEventSource.java
```diff
@@ -44,9 +45,9 @@ public MariaDbSnapshotChangeEventSource(MariaDbConnectorConfig connectorConfig,
                                             BlockingConsumer<Function<SourceRecord, SourceRecord>> lastEventProcessor,
                                             Runnable preSnapshotAction,
                                             NotificationService<MariaDbPartition, MariaDbOffsetContext> notificationService,
-                                            SnapshotterService snapshotterService) {
+                                            SnapshotterService snapshotterService, BeanRegistry beanRegistry) {
```
Naros, line 48: I wonder if we shouldn't pass the `ServiceRegistry` as a future proof need and then fetch the `BeanRegistry` out of it. This way as we add new services to the registry, the signature would remain unchanged. wdyt? If you agree, I'd think of applying this in all places where it makes sense.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbChangeEventSourceFactory.java
```diff
@@ -64,7 +68,7 @@ public SnapshotChangeEventSource<MongoDbPartition, MongoDbOffsetContext> getSnap
                 snapshotProgressListener,
                 errorHandler,
                 notificationService,
-                snapshotterService);
+                snapshotterService, this.beanRegistry);
```
Naros, line 71:
```suggestion
                snapshotterService, 
                this.beanRegistry);
```
I know we aren't uniform about this behavior, but as a side note, I'd prefer if we have blocks of code that use the style here where arguments are per line, adding new arguments should adhere to that same style. I think I may have caught a few others, but I could have missed a few.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/MongoDbConnectorTask.java
```diff
@@ -132,7 +129,7 @@ public ChangeEventSourceCoordinator<MongoDbPartition, MongoDbOffsetContext> star
                     DataChangeEvent::new,
                     metadataProvider,
                     schemaNameAdjuster,
-                    signalProcessor);
+                    signalProcessor, getServiceRegistry());
```
Naros, line 132: Another use case of the style.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
```diff
@@ -346,8 +345,6 @@ public MySqlConnectorConfig(Configuration config) {
         final String gtidSetExcludes = config.getString(GTID_SOURCE_EXCLUDES);
         this.gtidSourceFilter = gtidSetIncludes != null ? Predicates.includesUuids(gtidSetIncludes)
                 : (gtidSetExcludes != null ? Predicates.excludesUuids(gtidSetExcludes) : null);
-
-        getServiceRegistry().registerServiceProvider(new MySqlCharsetRegistryServiceProvider());
```
Naros, line 350: I suspect if we reverse course on the `MariaDbCharsetRegistry` behavior, it would make sense to do that here as well; although unlikely MariaDB, the MySQL registry doesn't parse anything and acts as a delegate to the driver.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresChangeEventSourceCoordinator.java
```diff
@@ -47,9 +49,10 @@ public PostgresChangeEventSourceCoordinator(Offsets<PostgresPartition, PostgresO
                                                 EventDispatcher<PostgresPartition, ?> eventDispatcher, DatabaseSchema<?> schema,
                                                 SnapshotterService snapshotterService, SlotState slotInfo,
                                                 SignalProcessor<PostgresPartition, PostgresOffsetContext> signalProcessor,
-                                                NotificationService<PostgresPartition, PostgresOffsetContext> notificationService) {
+                                                NotificationService<PostgresPartition, PostgresOffsetContext> notificationService, BeanRegistry beanRegistry,
```
Naros, line 52: I must have missed this in earlier changes in the PR, but I noticed here we pass both the BeanRegistry and the ServiceRegistry. Would you agree its satisfactory to pass the ServiceRegistry and only then get the bean registry where needed to keep the signature shorter and "more clean" than it currently is? :grin:

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -238,6 +243,13 @@ public final void start(Map<String, String> props) {
         try {
             setTaskState(State.INITIAL);
             config = Configuration.from(props);
+            this.beanRegistry = new DefaultBeanRegistry();
+            this.serviceRegistry = new DefaultServiceRegistry(config, beanRegistry);
+            serviceRegistry.registerServiceProvider(new PostProcessorRegistryServiceProvider());
+            serviceRegistry.registerServiceProvider(new SnapshotLockProvider());
+            serviceRegistry.registerServiceProvider(new SnapshotQueryProvider());
+            serviceRegistry.registerServiceProvider(new SnapshotterServiceProvider());
```
Naros, line 251: I certainly like this approach better than my original choice in the `CommonConnectorConfig` class. 

One suggestion here may be to add a protected method where you pass in the `ServiceRegistry` as an argument, and that would be where you move the `registerServiceProvider` calls. Then in the future if a specific connector has the need to register some custom service providers, the hook already exists as an overridable point. wdyt?

Naros (commented):
Hi @kmos, great work so far and I like the changes. 

I've left a few comments. I did not catch all the style concerns that I brought up, so I'll leave that as an exercise for you to find the others. 

I did mention a comment about favoring the passing of `ServiceRegistry` rather than `BeanRegistry`. I think there are arguments from both sides of what makes the most sense from both a testability/mock perspective versus providing only what's needed. But I also would like to see us really go back over our classes that have massive ctor lists and clean those up with the help of the service registry, hence my preference there, but YMMV.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
kmos: Yeah, I completely agree. This was one of the shortcuts I took to highlight the drawbacks of this approach.

In some classes like `BinlogFieldReader`, it’s not possible to inject the service registry from an instance created in `BaseSourceTask`. As a workaround, I created new instances wherever the registry was needed. 

If we decide to proceed with this refactoring/approach and aim to maintain a single instance of `charsetRegistry`, one option is to expose the `ServiceRegistry` as a static reference—making it accessible even in components like `BinlogFieldReader`.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorTask.java
```diff
@@ -96,21 +105,20 @@ public ChangeEventSourceCoordinator<MySqlPartition, MySqlOffsetContext> start(Co
...
+        getBeanRegistry().add(StandardBeanNames.CONFIGURATION, config);
+        getBeanRegistry().add(StandardBeanNames.CONNECTOR_CONFIG, connectorConfig);
+        getBeanRegistry().add(StandardBeanNames.DATABASE_SCHEMA, schema);
+        getBeanRegistry().add(StandardBeanNames.JDBC_CONNECTION, beanRegistryJdbcConnection);
+        getBeanRegistry().add(StandardBeanNames.VALUE_CONVERTER, valueConverters);
+        getBeanRegistry().add(StandardBeanNames.OFFSETS, previousOffsets);
 
-        // Service providers
-        registerServiceProviders(connectorConfig.getServiceRegistry());
+        getServiceRegistry().registerServiceProvider(new MySqlCharsetRegistryServiceProvider());
```
kmos, line 119: here I moved `getServiceRegistry().registerServiceProvider(new MySqlCharsetRegistryServiceProvider());`

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
kmos: Apologies, I didn’t fully understand your comment, could you please elaborate?

In this case, I removed the load call because I assumed it was intended to eagerly load the `MySqlCharsetRegistryServiceProvider`, which I’ve now moved inside the `ConnectorTask` (please see the comment in [MySqlConnectorTask.java](https://github.com/debezium/debezium/pull/6458/files#diff-2fb991f953c491e9f94e023c0f9c1cf0d910594df1d09fcb457a76e97b1409fd)).

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
kmos: Totally Agree, I was thinking the same!

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresChangeEventSourceCoordinator.java
kmos: Regarding that, we can continue the discussion in the other comment.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbSnapshotChangeEventSource.java
Naros: It appears that the `BeanRegistry` now has everything purely by coincidence, but moving forward, that may not always hold as we consider adding more services. The `BeanRegistry` was primarily a short-term solution to enable a connector to register named objects in a map that other parts of the code could look up and retrieve, which were not yet ported to be `Service` or had not yet been assigned a `ServiceProvider`. Honestly, I'm not a fan of (2) as I would prefer for `BeanRegistry` to be removed.

As for (3), I am not opposed to limiting the number of objects passed to only those that are necessary. That can often make things easier to mock and test. However, I would like us to reevaluate whether some of our constructors and methods should be repurposed to use parameter objects. I believe it would be helpful to consider whether parameter objects are a better design pattern for compatibility reasons in certain instances. 

So, yes, I think it's reasonable to question how viable using `ServiceRegistry` alone is in hindsight; perhaps a more contextual object approach would be better in the long term?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
Naros: Sorry that my comment wasn't clear. I was conveying that if we chose to revert back so that there is just one `BinlogCharsetRegistry` instance created , it would make sense to align the MariaDB/MySQL behavior despite the fact that using different instances does not have IO concerns like MariaDB's implementation does.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbSnapshotChangeEventSource.java
mfvitale: I agree with @Naros and also seems that we all agree that the best solution will be the (3) but with the support of a `ServiceRegistry` that will do auto discovery starting from a root Service (maybe our Coordinator ?).

But until we reach the point where all classes are a `Service` I think we cannot remove the `BeanRegistry`.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
Naros: I wouldn't expose the service registry as a static reference, that will likely bite us in some environments where static resources are not expected like that

PR #6459, opened 2025-05-28
DBZ-9082 Add option to specify custom keystore and truststore for Redis sink

nange10v:
This change introduces additional configuration options for specifying custom keystore/truststore for the jedis client in redis sink. This allows the client to use different keys and certificates than the default ones passed to the JVM.

This change updates the Jedis client to version 6.0.0. and takes advantage of the newly introduced SSL options to set the keystore/truststore. The options are applied only if redis.ssl.enabled is true.

Configuration properties for each new SSL options:
`redis.ssl.truststore.path`: Specifies the path to the truststore file used for SSL.
`redis.ssl.truststore.password`: Specifies the password for the truststore file.
`redis.ssl.truststore.type`: Specifies the type of the truststore (e.g., JKS, PKCS12) (default: JKS).
`redis.ssl.keystore.path`: Specifies the path to the keystore file used for SSL.
`redis.ssl.keystore.password`: Specifies the password for the keystore file.
`redis.ssl.keystore.type`: Specifies the type of the keystore (e.g., JKS, PKCS12) (default: JKS).

This update introduces a factory method that creates a RedisConnection directly from a shared configuration object (RedisCommonConfig). Previously, both Debezium and Debezium Server needed to duplicate any changes made to Redis connection parameters, leading to tight coupling and maintenance overhead. By centralizing connection creation through a common config, any updates to the Redis connection settings now only need to be made in Debezium.

### DIFF
--- a/debezium-bom/pom.xml
+++ b/debezium-bom/pom.xml
@@ -43,7 +43,7 @@
         <!-- Storages -->
 
         <!-- Redis -->
-        <version.jedis>4.1.1</version.jedis>
+        <version.jedis>6.0.0</version.jedis>
 
         <!-- S3 -->
         <version.s3>2.26.30</version.s3>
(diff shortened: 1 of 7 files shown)


jpechane:
@nange10v Applied, thanks!

PR #6460, opened 2025-05-28
DBZ-9075 move packages from `io.quarkus.*` to `io.debezium.*`

kmos:


quarkus-debezium-parent/pom.xml
```diff
@@ -40,22 +40,22 @@
                 <scope>import</scope>
             </dependency>
             <dependency>
-                <groupId>io.quarkus</groupId>
+                <groupId>io.debezium.quarkus</groupId>
```
jpechane, line 43: I know we don't use it elsewhere but would `${project.groupId}` work? If yes it might be useful to use it for potential future move to Quarkus project.

kmos: it seems that properties in parent definition are not allowed like:

```xml
    <parent>
        <groupId>${project.groupId}</groupId>
        <artifactId>quarkus-debezium-engine-spi-parent</artifactId>
        <version>3.2.0-SNAPSHOT</version>
    </parent>
```

jpechane: @kmos Yes, that makes sense, but it should work in dependencies, should not it?

kmos: Yeah in dependencies for sure

PR #6461, opened 2025-05-28
DBZ-9085 Exclude TRUNCATE and REPLACE statements from schema history

akanimesh7:


jpechane:
@akanimesh7 Don't forget that `TRUNCATE` is processed and emitted as truncate event.

jpechane:
@akanimesh7 Applied, thanks!

PR #6463, opened 2025-05-28
DBZ-9086 Fix Postgres reselectColumns for serial primary key

jasonspeck:
https://issues.redhat.com/browse/DBZ-9086

`::serial` cast caused errors when primary key was serial type


jpechane:
@jasonspeck Applied, thanks!

PR #6464, opened 2025-05-29
DBZ-9091: Escape special characters in SQL Server connector

morozov:


morozov:
This ended up being a much larger change than I anticipated, and it still contains a workaround for the prepared statement cache in a couple of tests. Before I proceed, @jpechane please let me know how you would like to approach reviewing/accepting these changes. I broke them down into commits, each of which is a dependency for the following one and could be reviewed/merged separately.

debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -82,7 +82,29 @@ public class JdbcConnection implements AutoCloseable {
...
+     *
+     * <ol>
+     *     <li>{@code %} and {@code _} are wildcard characters defined by the ANSI SQL standard.</li>
+     *     <li>{@code [} is used in SQL Server's extended pattern syntax to define character ranges or sets.</li>
+     *     <li>The escape character itself (denoted by {@code ESCAPE_CHAR}) must also be escaped.</li>
+     * </ol>
+     *
+     * @see <a href="https://learn.microsoft.com/en-us/sql/t-sql/language-elements/like-transact-sql#pattern">SQL Server LIKE pattern syntax</a>
+     */
+    private static final Set<Character> SPECIAL_CHARS = Set.of('%', '_', '[', ESCAPE_CHAR);
```
morozov, line 106: Technically, the SQL Server specific character doesn't belong to the generic `JdbcConnection` class. But this is so nice an cozy to have everything defined in a single `private static final` field. We can take it apart, but then we'd have to make it non-static and protected and initialize/override in constructors.

jpechane:
@morozov I don't mind accepting the PR as is. Thanks a lot for the fix!

debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
Naros: Hi @morozov so this specific commit has broken Oracle tests. In Oracle, its perfectly valid to use `_` characters in table names and they do not require any sort of escaping. Are we sure that `_` requires escaping in this context or is it something unique to just SQL Server?

AFAIK we don't currently have a universal test that checks table names across the database gambit with these special characters to make sure that this does not break other implementations with such use cases like we do for Oracle, but this was missed since Oracle's tests are manually triggered. :cry: 

So we're trying to understand should this specific behavior be scoped to SQL Server or is it universal for other databases and Oracle is just the odd one out here.

morozov: > Are we sure that `_` requires escaping in this context or is it something unique to just SQL Server?

I believe so. My assumption was and is that `DatabaseMetadata` uses LIKE to evaluate the pattern, and `_` is a special character in all SQL implementations including Oracle:
```sql
SELECT CASE WHEN 'X' LIKE '_' THEN 'YES' ELSE 'NO' END
FROM DUAL;
-- YES
```

What I didn't account for is:
1. As you found out, the Oracle's JDBC driver uses a different escape character.
2. It analyzes and pre-processes the pattern before using it in SQL. If the pattern doesn't contain the expected escape character, it uses it as a literal (this is I believe what caused the failure).

If we correctly specify the expected escape character and escape `_` using it, it will make the driver use it as a pattern, not as a literal.

> So we're trying to understand should this specific behavior be scoped to SQL Server or is it universal for other databases and Oracle is just the odd one out here.

Are you asking about the behavior of Debezium escaping special characters in database object names? I believe it should be universally applied to all databases assuming that all respective JDBC drivers implement `DatabaseMetadata` correctly and interpret the patterns. Debezium needs to prevent that by escaping special characters.

I can do some exploratory work and see if there are any similar quirks in other drivers.

PR #6470, opened 2025-06-02
DBZ-9098 Reduce calls to getEndOffset during kafka schema history rec…

akanimesh7:


akanimesh7:
@jpechane 
Please check

jpechane:
@akanimesh7 Applied, thanks!

PR #6472, opened 2025-06-02
DBZ-7533 Upgrade PostgreSQL driver to 42.7.6

jonomorris:
Updated Postgres JDBC driver to 42.7.6, and added config to turn off the automatic LSN flush now that we are able to do so [1].    

Included a fix in order to match `TabledIds` created from the database schema with those created using messages from the replication stream following a change in 42.7.5 to include the database name in `getTables` [2].

Updated 'quarkus-platform' to 3.23.1 as this version uses driver 42.7.6 [3].

1. https://github.com/pgjdbc/pgjdbc/pull/3403
2. https://github.com/pgjdbc/pgjdbc/pull/3390
3. https://github.com/quarkusio/quarkus/blob/3.23.2/bom/application/pom.xml#L122

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
@@ -1276,7 +1276,6 @@ public class PostgresConnectorConfig extends RelationalDatabaseConnectorConfig {
                     QUERY_TIMEOUT_MS,
                     PLUGIN_NAME,
                     SLOT_NAME,
-                    // AUTOMATIC_FLUSH,
                     PUBLICATION_NAME,
                     PUBLICATION_AUTOCREATE_MODE,
                     REPLICA_IDENTITY_AUTOSET_VALUES,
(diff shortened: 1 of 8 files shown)


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
```diff
@@ -683,6 +683,14 @@ public static SnapshotLockingMode parse(String value, String defaultValue) {
             .withImportance(Importance.LOW)
             .withDescription("The timeout in seconds for the creation of the replication slot.");
 
+    public static final Field AUTOMATIC_FLUSH = Field.create("automatic.flush")
```
jpechane: I don't think we want this option configurable as it IMHO always breaks Debezium's semnatics.
@Naros WDYT?

Naros: Agreed, this should be a Debezium-opinionated concept @jpechane @jonomorris

jonomorris: Updated so that the `PGReplicationStream` will always have automatic flush disabled, and removed the option.

jpechane:
@jonomorris Applied, thanks!

PR #6473, opened 2025-06-02
DBZ-9100 Bump Chicory to 1.4.0

andreaTP:
Both, compiler and annotations are now released as stable.

I took the chance to align the naming, no-one should be directly referencing those from code as far as I understand.


jpechane:
@andreaTP Applied, thanks!

PR #6477, opened 2025-06-03
DBZ-8956 fix names for Quarkus Extension poms

kmos:


jpechane (changes requested):
@kmos Thanks!. Could you please just modify the names a bit to reflect the content a bit more or make it consistent?
In one case it is `Quarkus Debezium` set of modules and in other it is `Debezium Quarkus`.

When addressed feel free to merge the PR yourself.

PR #6479, opened 2025-06-03
DBZ-9017 Skip null values in FieldNameTransformation

vjuranek:
https://issues.redhat.com/browse/DBZ-9017

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/transforms/FieldNameTransformation.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/transforms/FieldNameTransformation.java
@@ -186,7 +186,11 @@ public class FieldNameTransformation<R extends ConnectRecord<R>> implements Tran
                 }
 
                 Struct struct = new Struct(schema.build());
-                newValues.forEach(struct::put);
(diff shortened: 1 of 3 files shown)


jpechane (commented):
@vjuranek The code looks correct. Yet I wonder if the real root cause is fixed. IIUC the source message has non-optional field and we tried to put null value into non-optional field of the struct. But if the field was non-optional how is it possible that is is null?

vjuranek:
It's `null` when parent struct is optional and is not included in the record, which is the case of e.g. create event and `before` field. `before` field is optional (and missing) in create record, but child elements in `before` struct schema are not optional.

Current implementation of the transformation iterates over all stucts/fields in the schema and place a new value for each of them, which can be empty struct for missing optional field:
```
Struct value = new Struct(schema.build());
        if (originalValue != null) {
            originalSchema.fields().forEach(field -> value.put(transformFieldName(field.name()), originalValue.get(field)));
        }

        return new TransformedSchemaValue(value.schema(), value);
```
so instead of `null` (in our case value of `before` struct) we get empty struct as a transformed value.

jpechane:
@vjuranek Thanks for the explanation! Applied.

PR #6480, opened 2025-06-03
DBZ-8319 - Debezium readonly connection

lgazire:
DBZ-8319


lgazire:
@jpechane I think now everything is fine.

jpechane:
The test failures are related. There is a NPE in initialization.

jpechane:
@lgazire Could you please remove the merge commit and use rebase? We prefer lienar history. Thanks for understanding.

lgazire:
@jpechane while i fix the pull commited can you help me with the error in the Testing Farm Result? It seems not related to this change.

Naros:
Hi @lgazire are you sure its not related? It would seem that the connection used by the reselect post processor cannot find the table it needs to reselect, which would indicate that perhaps that connection is pointing to the CDB rather than the PDB? I haven't looked at the code changes recently, but that would be my first pass check.

PR #6481, opened 2025-06-03
DBZ-9092 Column name encrypted is not supported by MySqlParser

kmos:
## Context

The following query is valid in Mysql but invalid for mariaDB:

```sql
CREATE TABLE IF NOT EXISTS table_with_encrypted_column_name(uuid VARCHAR(40) NOT NULL PRIMARY KEY, encrypted VARCHAR(40) DEFAULT NULL)
```

## changes
Removed the keyword `encrypted` in mysql grammar


jpechane:
@kmos Applied, thanks! This is quite interesting s really `ENCRYPTED` does not seem to be part of MySQL grammar. I wonder how it got in.

kmos:
Well, it seems something that is available in mariaDB and not for mysql. Maybe a copy&paste

PR #6482, opened 2025-06-03
extend SQL parser to accept GRANT/REVOKE FLASHBACK

moench-tegeder:
The old code recognized "FLASHBACK ARCHIVE" as an object privilege, but "FLASHBACK" ("Issue a SQL Flashback Query on the table") without the ARCHIVE part was ommitted. Flashback Queries ("AS OF") are avaible for a very long time in Oracle Database - and in fact, the Debezium Oracle Connector uses Flashback Queries when creating snapshots.


debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/oracle/generated/PlSqlParser.g4
```diff
@@ -5210,7 +5210,7 @@ object_privilege
     | DEBUG
     | DELETE
     | EXECUTE
-    | FLASHBACK ARCHIVE
+    | FLASHBACK ARCHIVE?
```
Naros:
```suggestion
    | FLASHBACK
    | FLASHBACK ARCHIVE
```

debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/oracle/generated/PlSqlParser.g4
```diff
@@ -5210,7 +5210,7 @@ object_privilege
...
     | INDEX
     | INHERIT PRIVILEGES
```
Naros, line 7532:
```suggestion
    | INHERIT PRIVILEGES
    | INHERIT REMOTE PRIVILEGES    
```

For this we need to add `REMOTE` into the Lexer as follows:
```
REMOTE:                       'REMOTE';
```
and then add it to the `non_reserved_keywords_in_12c` rule, this way we cover all object privileges that exist to avoid any future issues with the same rule.

Naros (changes requested):
Hi @moench-tegeder this is very close, I left a few inline comments.

I'd also suggest updating the `ddl_revoke.sql` file in the tests and adding these lines at the end:
```
REVOKE SELECT,FLASHBACK,INHERIT REMOTE PRIVILEGES ON hr.employees FROM hr;
REVOKE FLASHBACK ARCHIVE ON hr.employees FROM hr;
```

This way we have a specific test for the rule as well.

debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/oracle/generated/PlSqlParser.g4
moench-tegeder: fixed in 8958aa9bcb726882a460909e029fb3091bde087a

debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/oracle/generated/PlSqlParser.g4
moench-tegeder: also fixed in 8958aa9bcb726882a460909e029fb3091bde087a

jpechane:
@moench-tegeder The Oracle errors are directly related to grammar. Could you please check it out? Thanks

moench-tegeder:
> @moench-tegeder The Oracle errors are directly related to grammar. Could you please check it out? Thanks

I'm not sure why your version of Oracle Database reports this as syntax error... A very recent Oracle (23) shows that we would need an actual flashback archive, and that we need additional privileges to create one. So I decided to not test GRANT/REVOKE FLASHBACK ARCHIVE.

jpechane:
@moench-tegeder IIRC the error was reported already during the parser test, so either there was an issue in grammar or the test case itself.

moench-tegeder:
> @moench-tegeder IIRC the error was reported already during the parser test, so either there was an issue in grammar or the test case itself.

Even better: I believe I uncovered another problem.
In https://github.com/debezium/debezium/blob/cd75ed1c01d2245c93554aad25b09b19938b8dce/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/oracle/generated/PlSqlParser.g4#L1283
that grant_statement has to have a closing ';', but
https://github.com/debezium/debezium/blob/cd75ed1c01d2245c93554aad25b09b19938b8dce/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/oracle/generated/PlSqlParser.g4#L1292
does not force a closing ';' on the statement.
The closing semicolon would be already matched in https://github.com/debezium/debezium/blob/cd75ed1c01d2245c93554aad25b09b19938b8dce/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/oracle/generated/PlSqlParser.g4#L36 by SEMICOLON - in fact, this is relevant when the input has more than one command, as the inner SEMICOLONs are not optional, only the last one. And funnily enough, the ddl_grant tests only fail when there is more than one command in ddl_grant.sql. (Remember that there is no GRANT-coverage in the main branch - adding some in this PR uncovered this).
Looking further, the end-';' are all over the place in PlSqlParser.g4 - I guess I should clean that up, too. This PR morphes into a "clean up the parser" job...

jpechane:
Well, in this case it probably makes sense to pause the PR and pull the grammar and exmaples from upstream grammar. It contains also examples for GRANT commands.
When this gets passing the relevant chnages in the grammar should be made.
It probably does not make sense to try to find issues that are probably no longer present in the upstream grammmar.

Naros:
The alignment PR with ANTLR upstream is https://github.com/debezium/debezium/pull/6778. Once that is merged, we can review if there are any remaining changes needed here.

Naros:
Included changes from this PR in https://github.com/antlr/grammars-v4/pull/4640 in commit
[[PlSql] Improve object permissions](https://github.com/antlr/grammars-v4/pull/4640/commits/e4b9503f8fcba60f0ed84aa6919f9a782fc2076f)

Naros (approved):

PR #6485, opened 2025-06-04
DBZ-9011 Don't use deprecated txid_current()

vjuranek:
https://issues.redhat.com/browse/DBZ-9011


jpechane:
@vjuranek Nice, applied!

PR #6486, opened 2025-06-04
DBZ-9117: Create schemaNamePattern from schemaName for schema introspection

morozov:
This is a continuation of the effort started in https://github.com/debezium/debezium/pull/6464.

Here's the change summary:

### Use consistent parameter and variable naming in `JdbcConnection`

The parameter and variable names in some `JdbcConnection` methods are inconsistent:
1. There is `tableName` but `databaseCatalog`. The latter has been renamed to `catalogName` for consistency.
2. The collision between the `tableName` parameter and the table name introspected via the JDBC is addressed by naming the variable representing the introspected value as `metaTableName`, but `catalogName` is not prefixed with "metadata" because it doesn't conflict with `databaseCatalog`.

Before making the actual code changes, I'd like to unify the naming. This way, the changes themselves are easier to review and validate.

### Rename `schemaNamePattern` parameter to `schemaName` in `JdbcConnection` methods

Even though all modified methods accept the `schemaNamePattern` parameter, all callers pass a schema name there. This is the root cause of the bug. Since there are no cases when a caller actually needs to pass a pattern, we change the name and the semantics of this parameter. Now it's `schemaName`.

#### No changes in `JdbcConnection#readTableNames()`

Note that this method still accepts `schemaNamePattern` and `tableNamePattern`. The reason is that all callers of this method pass `null` for both of the parameters, so whether these arguments are interpreted as names or patterns doesn't matter. In order to prevent similar escaping issues from happening in the future, it might be a good idea to just remove these parameters at all.

### Create `schemaNamePattern` from `schemaName`

This is the actual bug fix. The same was done in https://github.com/debezium/debezium/pull/6464 for table names.

`SpecialCharsInNamesIT` has been modified to use schema names with special characters (from both the SQL and LIKE pattern syntaxes) to cover the change.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
@@ -615,7 +615,7 @@ public class OracleConnection extends JdbcConnection {
 
     @Override
     protected Map<TableId, List<Column>> getColumnsDetails(String catalogName,
-                                                           String schemaNamePattern,
+                                                           String schemaName,
                                                            String tableName,
                                                            Tables.TableFilter tableFilter,
                                                            ColumnNameFilter columnFilter,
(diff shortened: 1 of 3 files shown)


Naros:
@jpechane ideally we should get a consensus on my recent PR #6487 as this will potentially introduce further regressions without it since this PR uses the method I fixed.

jpechane:
I agree, let's get the @Naros PR in first and then we can continue with this one

morozov:
I rebased my changes in top of the ones from https://github.com/debezium/debezium/pull/6487, so they are ready for review.

All build failures (including the SQL Server ones) seem irrelevant. All the failures I've examined (about a half) are caused by `assertConnectorIsRunning()` failing.

jpechane:
@morozov Applied, thanks!

PR #6487, opened 2025-06-05
DBZ-9091 Use DatabaseMetadata for escape character resolution

Naros:
https://issues.redhat.com/browse/DBZ-9091

@morozov can you provide some feedback here? 

This allows the Oracle tests to pass because Oracle's driver uses the `/` character to identify escape requirements in the metadata lookups and not `\`. Given the metadata can provide this detail, I added that. My only issue here is the use of `String.valueOf(c)` but given this shouldn't be a hotspot in the code, it is likely okay. wdyt?

### DIFF
--- a/debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
+++ b/debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
@@ -1269,23 +1269,28 @@ public class JdbcConnection implements AutoCloseable {
      * Given a database object name, creates a pattern, escaping any special characters such that the pattern matches
      * the name itself.
      */
-    protected String createPatternFromName(String name, String databaseSearchEscapeCharacter) {
+    protected String createPatternFromName(String name, String searchEscapeCharacter) {
         if (name == null) {
             return null;
+        }
+
+        if (Strings.isNullOrBlank(searchEscapeCharacter) || searchEscapeCharacter.length() > 1) {
+            throw new DebeziumException("The driver's search escape character should be a length of 1");
         }
 
         final int length = name.length();
         StringBuilder pattern = null;
 
+        final char escapeCharacter = searchEscapeCharacter.charAt(0);
         for (int i = 0; i < length; i++) {
             char c = name.charAt(i);
 
-            if (likeWildcardCharacters.contains(c) || databaseSearchEscapeCharacter.equals(String.valueOf(c))) {
+            if (likeWildcardCharacters.contains(c) || escapeCharacter == c) {
                 if (pattern == null) {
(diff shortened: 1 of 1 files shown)


jpechane (approved):
@Naros LGTM, thanks!
Vitess failure seems related, Informix depends on https://github.com/debezium/debezium-connector-informix/pull/84

When solved feel free to merge yourself!

debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -1275,7 +1269,7 @@ protected String resolveCatalogName(String catalogName) {
      * Given a database object name, creates a pattern, escaping any special characters such that the pattern matches
      * the name itself.
      */
-    protected String createPatternFromName(String name) {
+    protected String createPatternFromName(String name, String databaseSearchEscapeCharacter) {
```
morozov: I'd rename `databaseSearchEscapeCharacter` to `searchEscapeCharacter`:
1. It aligns with the `DatabaseMetadata` method name.
2. The actual value is defined not so by the database as by the driver (technically any SQL query can use its own escape character, and the driver uses one of them).

debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -1286,12 +1280,12 @@ protected String createPatternFromName(String name) {
         for (int i = 0; i < length; i++) {
             char c = name.charAt(i);
 
-            if (likeWildcardCharacters.contains(c) || LIKE_ESCAPE_CHARACTER.equals(c)) {
+            if (likeWildcardCharacters.contains(c) || databaseSearchEscapeCharacter.equals(String.valueOf(c))) {
```
morozov: Alternatively, before the `for` loop, we can assert that the `databaseSearchEscapeCharacter` string consists of one character and extract the character. Then use the character for comparison.

If some driver uses more than one character for escaping (which I'm not sure is allowed by SQL), the condition `databaseSearchEscapeCharacter.equals(String.valueOf(c))` would have to be reworked as it would never evaluate to `true`.

morozov (commented):
@Naros thanks a lot for addressing the issue. The approach looks right (my assumption of the escape character being always the same wasn't). Please see a couple of inline comments.

jpechane:
@vjuranek Would you be able to finishe the job? If yes, could you please modify commit messages to refer to https://issues.redhat.com/browse/DBZ-9131 ?

vjuranek (approved):
Test failures are unrelated

vjuranek:
Merged, thanks!

PR #6488, opened 2025-06-05
[DBZ-9118] Fix warn log if snapshot mode when needed is required but not set

akanimesh7:


akanimesh7:
Some tests failed. Checking them.

jpechane:
I triggered re-executin of failed tests, The POstgres failures might be related to changes in tests.

akanimesh7:
One of the failing tests -- `shouldRegularlyFlushLsnWithTxMonitoring` requires https://github.com/debezium/debezium/pull/6494 to be merged.

Other failing tests for postgres from ReplicationConnectionIT are failing in other unrelated PR also. This one -- https://github.com/debezium/debezium/pull/6497 . Hence seems unrelated.

`AbstractNotificationsIT.sinkNotificationWillCorrectlySaveOffsetAfterSnapshot` --> For test cases where restart of connector will happen, it's necessary to preserve the older replication slot. Hence, added `slot.drop.on.stop` value as false for the first config instance. There could be better solution rather than putting this in the Abstract class, it could be put in the NotificationsIT in the postgres module, but that would need change in the API. Put it in the abstract class for now, considering the effort vs gain tradeoff.

`PostgresConnectorIT.shouldStopConnectorOnSlotRecreation` --> Connector now fails early if slot is recreated. Updated error log message for that.

jpechane:
@akanimesh7 Makes sense, applied, thanks

jpechane:
@akanimesh7 I just hit https://productionresultssa19.blob.core.windows.net/actions-results/0d39e667-e085-4fef-84fc-015423544acc/workflow-job-run-c6c6859f-d60d-5c7e-85f9-5797cc493788/logs/job/job-logs.txt?rsct=text%2Fplain&se=2025-06-13T05%3A25%3A23Z&sig=ponmzdznfChWEeYMorI8cHA55h2%2FA5sd9kVh4E9ks0c%3D&ske=2025-06-13T16%3A14%3A58Z&skoid=ca7593d4-ee42-46cd-af88-8b886a2f84eb&sks=b&skt=2025-06-13T04%3A14%3A58Z&sktid=398a6654-997b-47e9-b12b-9515b896b4de&skv=2025-05-05&sp=r&spr=https&sr=b&st=2025-06-13T05%3A15%3A18Z&sv=2025-05-05

I am not sure but there may be some edge cases that would cause this change to break. So thinking about that while I am fine doing this by default could you please extend the behaviour with `io.debezium.config.CommonConnectorConfig.EVENT_PROCESSING_FAILURE_HANDLING_MODE` support?
So fail (default) will be the current one. warn will be the oroiginal behaviour and ignore will jus silently skip over it.

With thois option if there is an issue for existing users they can easily continue operations while reporting a bug.

akanimesh7:
Makes sense @jpechane. I'll do that in a new PR.

akanimesh7:
@jpechane The link in [this](https://github.com/debezium/debezium/pull/6488#issuecomment-2969114566) one is not working

jpechane:
@akanimesh7 The run was probably lost. Generally the issue was in Postgres' blocking snapshot test, esp. `aFailedBlockingSnapshotShouldNotCauseInitialSnapshotOnRestart`

PR #6489, opened 2025-06-05
DBZ-9120 Add keepAlive to Schema Snapshot Table Lock Connection

jtanza:
Addresses https://issues.redhat.com/browse/DBZ-9120

Adds a keepAlive thread on the connection obtaining global locks during the binlog schema snapshot phase, periodically sending heartbeats (via a `SELECT 1`) to MySQL in order to keep the connection active and avoid premature socket disconnects before `UNLOCK TABLES` can be called.


debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogSnapshotChangeEventSource.java
```diff
@@ -669,4 +678,32 @@ protected void aborted(SnapshotContext<P, O> snapshotContext) throws Interrupted
 
         super.aborted(snapshotContext);
     }
+
+    private void startLockHeartbeat() {
+        if (lockKeepAliveExecutor == null || lockKeepAliveExecutor.isShutdown()) {
+            LOGGER.info("Starting lock heartbeat");
+            lockKeepAliveExecutor = Executors.newSingleThreadScheduledExecutor(r -> {
```
jpechane: Please use `io.debezium.util.Threads` class

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogSnapshotChangeEventSource.java
```diff
@@ -69,6 +72,7 @@ public abstract class BinlogSnapshotChangeEventSource<P extends BinlogPartition,
 
     private static final Logger LOGGER = LoggerFactory.getLogger(BinlogSnapshotChangeEventSource.class);
     private static final Logger ROW_ESTIMATE_LOGGER = LoggerFactory.getLogger(BinlogSnapshotChangeEventSource.class.getName() + ".RowEstimate");
+    private static final Integer LOCK_HEARTBEAT_INTERVAL = 30;
```
jpechane:
```suggestion
    private static final Duration LOCK_HEARTBEAT_INTERVAL = Duration.ofSeconds(30);
```

jpechane (changes requested):
@jtanza Hi, thanks for PR. I left few suggestions for changes. But I am more concerned with one issue - `io.debezium.jdbc.JdbcConnection` is explicitly marked as `@NotThreadSafe`. Also can we be sure that underlying connection for both MySQL and MariaDB can be used by multiple threads?

jtanza:
Hey @jpechane thanks for the review. Yeah I completely missed the `NotThreadSafe` annotation. I'll refactor things a bit to synchronize access to the `BinlogConnectorConnection` with a mutex. Unless there's an issue w/ that i'll plan to get those changes in soon.

jpechane:
@jtanza If there is no other way then yes. But one question though. At which point exactly is the inactvity and timeout thrown? Can't we integrate the "heartbeat" code with the main loop is it is executed in the same thread?

jtanza:
@jpechane the exception is being thrown when `globalUnlock()` is being executed. So from my understanding: we execute the lock statement (without commiting), call our `SHOW CREATE TABLE` statements in our snapshot connectionPool, and finally release our lock (and in our case, fail with an exception). Given the fact that the main `connection` thread just obtains the lock and does nothing else, it fails w/ the timeout when calling unlock. I dont think moving the heartbeat to the main thread addresses that?

jpechane:
@jtanza Yes, in this case you are right. Then it leads me to a question - if the main thread is passive, do we need than do the mutex as we could declare that ownership is passed to the heartbeat thread and main thread can use it again only after the heartbeat thread is stopped?

jtanza:
@jpechane  I think that makes sense, yes. I am assuming you mean using a semaphore over the mutex for signaling between the lock and heartbeat thread?

jpechane:
@jtanza I guess the best way forward would be for you to send the mutex proposal and we can discuss it over. Right now it is a bit complicated to get on the same page.

ahmedbayraktar:
@jpechane Addressed your comments and added connection mutex to synchronize access to the non-thread-safe `BinlogConnectorConnection`

ahmedbayraktar:
@jpechane Please let me know if you have any thoughts on the latest changes for this PR.

jpechane:
@jtanza Applied, thanks!

PR #6490, opened 2025-06-06
DBZ-9097 Support KRaft for Kafka Cluster

AlvarVG:
With this PR we can run Kafka in KRaft mode without adding/changing any of the actual parameters. 
I have also added the param test.force.kraft in case we want to force the use of KRaft. 

And then, if the kafka version is equals or greater than 4.0.0, KRaft is used. And if the build is not productized and strimzi version is equals or greater than 0.46.0, we are also using KRaft (in this version Zookeeper is deprecated and no longer available).

### DIFF
--- a/debezium-testing/debezium-testing-system/src/main/java/io/debezium/testing/system/tools/kafka/builders/FabricKafkaBuilder.java
+++ b/debezium-testing/debezium-testing-system/src/main/java/io/debezium/testing/system/tools/kafka/builders/FabricKafkaBuilder.java
@@ -13,11 +13,15 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
 import io.debezium.testing.system.tools.fabric8.FabricBuilderWrapper;
 import io.fabric8.kubernetes.api.model.Secret;
 import io.strimzi.api.kafka.model.common.template.PodTemplate;
 import io.strimzi.api.kafka.model.common.template.PodTemplateBuilder;
 import io.strimzi.api.kafka.model.kafka.EphemeralStorage;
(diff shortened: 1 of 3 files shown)


debezium-testing/debezium-testing-system/src/main/java/io/debezium/testing/system/tools/kafka/builders/FabricKafkaBuilder.java
```diff
@@ -29,14 +35,20 @@
...
+import io.strimzi.api.kafka.model.nodepool.ProcessRoles;
 import io.strimzi.api.kafka.model.zookeeper.ZookeeperClusterSpec;
 import io.strimzi.api.kafka.model.zookeeper.ZookeeperClusterSpecBuilder;
 
 /**
  * This class simplifies building of kafka by providing default configuration for whole kafka or parts of its definition
  */
 public final class FabricKafkaBuilder extends FabricBuilderWrapper<FabricKafkaBuilder, KafkaBuilder, Kafka> {
+    private static final Logger LOGGER = LoggerFactory.getLogger(FabricKafkaBuilder.class);
+    private static Boolean USE_KRAFT;
```
vjuranek: Convention in Java is to use upper case names for constant. This is not a constant and is eventually changed later on. Using upper case name can cause the confusion or false assumption in later code edits, so I'd suggest to use lower case letters, or, even better, remove this variable completely.

AlvarVG: Thanks for the catch @vjuranek!! I am going to update the name, i dont want to remove the variable because i dont want to evaluate the method every time (only the first time).

vjuranek (approved):
LGTM, feel free to merge yourself once the release is done.

PR #6491, opened 2025-06-06
[docs] Fix page alias that cannot reference itself

mfvitale:
@vjuranek fyi I found that the doc was broken.


vjuranek:
thanks @mfvitale

PR #6493, opened 2025-06-07
DBZ-8510 Add Merge Into strategy in JDBC postgres sink connector

bpaquet:
Allow to handle complex scenario like partitionning: from a single id to (id, xx) with partition on xx.

closes https://issues.redhat.com/browse/DBZ-8510


Naros:
Hi @bpaquet, would it make sense that for PG15+, rather than introduce a special merge strategy, we instead use the `MERGE INTO` rather than the `INSERT ... ON CONFLICT UPDATE`, like as follows in the `getUpsertStatement`:

```java
if (getDatabaseVersion().isSameOrAfter(15)) {
  /* return the MERGE INTO */
}
else {
  /* return INSERT ... ON CONFLICT UPDATE */
}
```
Given that PG12 is already in EoL, PG13 is one month away, and PG14 is just over a year, I think this aligns more in supporting the future with standardized syntax. We can also drop the unnecessary CTE as that's not required for PG15+ any longer as well.

Naros:
Superseded by #6761.

PR #6494, opened 2025-06-07
[DBZ-9125] Fix test shouldRegularlyFlushLsnWithTxMonitoring

akanimesh7:


akanimesh7:
Failures are unrelated. Only affected test with this change is `shouldRegularlyFlushLsnWithTxMonitoring` and that is passing.

akanimesh7:
@jpechane to review

jpechane:
@vjuranek Could you please take a look at this PR? It intorduces a mirror operation you did in the past. But if I look and think about the code. IIRC the method in quesiton is shared by multiple connectors and for connectors like SQL Server, Db2 or MySQL the semicolo is part of standard transaction id. Does not have it the potential to break the test in some circumstances?

vjuranek:
@jpechane if doesn't do any harm as this is actually used only in Postgres tests. But in general (if someone start to use is in test for other connector), it starts to be an issue. We should probably add a helper method and override it in Postgres tests.

vjuranek:
Added a commit with proposed change to make it more general and usable for other connectors as well.

akanimesh7:
@vjuranek thanks for the addition !
`shouldRegularlyFlushLsnWithTxMonitoring` is passing. Failures seem to be unrelated.

jpechane:
@akanimesh7 Applied, thanks

PR #6496, opened 2025-06-09
DBZ-9110 Create debezium-openlineage dedicated module

mfvitale:


mfvitale:
Oracle and PostgreSQL17 failures seems not to be related.

debezium-api/src/main/java/io/debezium/config/ConfigurationDefinition.java
```diff
@@ -0,0 +1,16 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.config;
+
+public interface ConfigurationDefinition {
```
rk3rn3r: Doesn't this diverge too much (in naming) from how we do this in the connector modules?
Furthermore this classname could confuse users with Kafka's ConfDef class.

debezium-openlineage/debezium-openlineage-api/src/main/java/io/debezium/openlineage/DebeziumOpenLineageEmitter.java
```diff
@@ -56,7 +53,7 @@ public class DebeziumOpenLineageEmitter {
      * @param configuration The Debezium connector configuration containing OpenLineage settings
      * @param connectorTypeName The name of the connector used for event attribution
      */
-    public static void init(Configuration configuration, String connectorTypeName) {
+    public static void init(Map<String, String> configuration, String connectorTypeName) {
```
rk3rn3r, line 56: Why did this change from `Configuration` to `Map` type? If it is required, why is this `<String, String>` that could convert some values to something unexpected. Would `<String, ?>` make more sense and handle types more accurate, if we need to use `Map`?

debezium-openlineage/debezium-openlineage-core/src/main/java/io/debezium/openlineage/DebeziumOpenLineageConfiguration.java
```diff
@@ -34,22 +36,31 @@ public record Config(String path) {
...
                         owners));
     }
+
+    public static <T> List<T> getList(Map<String, String> configuration, String key, String separator, Function<String, T> converter) {
+        var value = configuration.get(key);
+        return value == null ? List.of()
+                : Arrays.stream(value.split(Pattern.quote(separator)))
+                        .map(String::trim)
+                        .map(converter)
+                        .collect(Collectors.toList());
```
rk3rn3r: Wouldn't it be better to return an immutable list?

debezium-api/src/main/java/io/debezium/config/ConfigurationDefinition.java
```diff
@@ -0,0 +1,16 @@
...
+
+    String TOPIC_PREFIX_PROPERTY_NAME = "topic.prefix";
+    String DATABASE_CONFIG_PREFIX = "database.";
```
rk3rn3r: `CommonConnectorConfig.DATABASE_CONFIG_PREFIX` should point here then to avoid duplication and instead of breaking the API?

jpechane:
@mfvitale In general this LGTM, but I really wonder if we are really exposing classes that are part of *API* in Debezium API module. Maybe it really makes sense to introduce one more module that will be shared by Debezium modules and move the classes there.

debezium-api/src/main/java/io/debezium/config/ConfigurationDefinition.java
mfvitale: Maybe `ConfigurationNames`?

debezium-openlineage/debezium-openlineage-api/src/main/java/io/debezium/openlineage/DebeziumOpenLineageEmitter.java
mfvitale: This method is called in `BaseSourceTask.start()` method that implements the `public abstract void start(Map<String, String> props);` from `org.apache.kafka.connect.source.sourceSourceTask`.

debezium-api/src/main/java/io/debezium/config/ConfigurationDefinition.java
mfvitale: I'm sure I changed any reference to it. Can you explain what you intend?

mfvitale:
> @mfvitale In general this LGTM, but I really wonder if we are really exposing classes that are part of _API_ in Debezium API module. Maybe it really makes sense to introduce one more module that will be shared by Debezium modules and move the classes there.

Yeah, you're are right. These classes are a little bit "forced" there. Maybe it's time for `debezium-common` module?

debezium-api/src/main/java/io/debezium/config/ConfigurationDefinition.java
jpechane: @mfvitale I believe @rk3rn3r wants you to do something like CommonConnectorConfig.DATABASE_CONFIG_PREFIX = ConfigurationDefinition.DATABASE_CONFIG_PREFIX

debezium-api/src/main/java/io/debezium/config/ConfigurationDefinition.java
jpechane: @mfvitale Sounds good

jpechane:
> > @mfvitale In general this LGTM, but I really wonder if we are really exposing classes that are part of _API_ in Debezium API module. Maybe it really makes sense to introduce one more module that will be shared by Debezium modules and move the classes there.
> 
> Yeah, you're are right. These classes are a little bit "forced" there. Maybe it's time for `debezium-common` module?

Yes, I think that's the best option.

mfvitale:
@jpechane I have added the `debezium-common` module. I need to fix the other connectors too but before I would like your though on the changes.

mfvitale:
Other connectors aligned.

jpechane (approved):
@mfvitale LGTM, thanks!
Please speak with @vjuranek about the name of the `ConnectorState` class. Kafka Connect differs between connector state and task state so the name can be IMHO a bit misleading.
When clarified feel free to merge this and releavnt PRs yourself.

vjuranek:
> Please speak with @vjuranek about the name of the `ConnectorState` class. Kafka Connect differs between connector state and task state so the name can be IMHO a bit misleading.

I'd call it `DebeziumTaskState` or something similar. It's task specific and moreover it differs from Kafka task state (or to be pre precise, AFAICT Kafka tasks have status instead of state) and is specific only to Debezium.

kmos (approved):

PR #1090, opened 2019-10-18
DBZ-1428 Restore TX isolation level after table read lock

jpechane:
https://issues.jboss.org/browse/DBZ-1428

PR #6498, opened 2025-06-09
DBZ-9006 Events inserted during snapshot are being duplicated

kmos:
## Context

The `binlog` snapshot process may produce duplicated events when other sessions perform `WRITE`s concurrently. This occurs because `InnoDB` is not instructed to create a consistent read view before releasing the `GLOBAL READ LOCK`, allowing duplicates between snapshot and streaming events.

The actual process as SQL:

```sql
-- 1. Disable autocommit to keep the session in manual commit mode
SET autocommit = 0;

-- 2. Prevent new writes (flush dirty pages, freeze binlog)
FLUSH TABLES WITH READ LOCK;

-- 3. Capture binlog position for snapshot
SHOW MASTER STATUS;

-- 4. set session transaction isolation (still under global read lock)
SET SESSION TRANSACTION ISOLATION LEVEL REPEATABLE READ;

-- 5. release the global read lock — snapshot is safe
UNLOCK TABLES;

-- 6. Do SELECTs 
SELECT * FROM customers;
SELECT * FROM orders;

-- 7. Finish the transaction (ends the snapshot)
COMMIT;
```

## Changes

We issue `START TRANSACTION WITH CONSISTENT SNAPSHOT` *before* releasing the global read lock to ensure that InnoDB captures a consistent read view. This guarantees that all subsequent `SELECT` queries operate on a stable snapshot, even after writes are re-enabled. Additionally, all intermediate `COMMIT`s have been removed to preserve snapshot isolation throughout the backup process.

the process after the changes:

```sql
-- 1. Disable autocommit to keep the session in manual commit mode
SET autocommit = 0;

-- 2. Prevent new writes (flush dirty pages, freeze binlog)
FLUSH TABLES WITH READ LOCK;

-- 3. Capture binlog position for snapshot
SHOW MASTER STATUS;

-- 4. set session transaction isolation (still under global read lock)
SET SESSION TRANSACTION ISOLATION LEVEL REPEATABLE READ;

-- 4.1 set consistent snapshot before unlock
START TRANSACTION WITH CONSISTENT SNAPSHOT;

-- 5. release the global read lock — snapshot is safe
UNLOCK TABLES;

-- 6. Do SELECTs 
SELECT * FROM customers;
SELECT * FROM orders;

-- 7. Finish the transaction (ends the snapshot)
COMMIT;
```

## Considerations

The current process treats the `snapshotLock` as something associated only with the schema (`lockTablesForSchemaSnapshot`). However, the `global read lock` also plays a role in defining the binlog offset. We suggest refactoring the lock strategy to account for both schema and offset consistency.

## Side Notes

The issue was introduced by #1090 due to incompatibility with RocksDB engine in MariaDB. For this reason we introduced a new snapshot mode `single_transaction` that is compatible with `innoDB` engine (only for mariaDB). The default snapshotLock for mariaDB is still `minimal`.

closes https://issues.redhat.com/browse/DBZ-9006


debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
```diff
@@ -72,6 +72,12 @@ public enum SnapshotLockingMode implements EnumeratedValue {
          * set to schema_only or schema_only_recovery.
          */
         NONE("none"),
+
+        /**
+         * the connector holds the global read lock and set a transaction `SET TRANSACTION ISOLATION LEVEL REPEATABLE READ` to ensure
+         * that the snapshot process will have a view from the mvcc. This mode is invalid for rocksDB
```
jpechane: RocksDB?

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
```diff
@@ -72,6 +72,12 @@ public enum SnapshotLockingMode implements EnumeratedValue {
...
+         */
+        SINGLE_TRANSACTION("single_transaction"),
```
jpechane: Is the name really correct? If this setting is not used then the snapshot is executed in multiple transactions?

jpechane: Also do we really need a new mode? Should not this be done by default?

jpechane (commented):
@kmos LGTM, thanks! I left few comments hofelly leading to addition simplification of the code.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
kmos: Yeah to be precise the engine is myRocks that adds the rocksDB database to mariaDB.


Reference: [https://mariadb.com/kb/en/myrocks/](https://mariadb.com/kb/en/myrocks/)

kmos: The error in the comment is that `single-transaction` apply  the following: `START TRANSACTION WITH CONSISTENT SNAPSHOT`

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
kmos: > Is the name really correct? If this setting is not used then the snapshot is executed in multiple transactions?

The name is used for context reason. It's the same name used in `mysqldump` for the same [option](https://dev.mysql.com/doc/refman/8.4/en/mysqldump.html#option_mysqldump_single-transaction) but we can use another name if it's misleading.



> Also do we really need a new mode? Should not this be done by default?

In a first draft, I thought the same, but then I found the [issue](https://issues.redhat.com/browse/DBZ-1428) that introduced this regression which led me to realize that using myrocks storage engine is not so uncommon as myISAM in mySQL.

kmos: We decided after an offline discussion to:
- to make default `START TRANSACTION WITH CONSISTENT SNAPSHOT` for mariaDB
- add a common snapshot mode called like *at-least-once* semantics in order to underline that can be duplicated events

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogConnectorConfig.java
```diff
@@ -262,6 +262,10 @@ public interface SnapshotLockingStrategy {
         default boolean preventsTableLocks() {
             return false;
         }
+
+        default boolean isSingleTransaction() {
```
jpechane:
```suggestion
        default boolean useConsistentSnapshotTransaction() {
```

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogSnapshotChangeEventSource.java
```diff
@@ -191,6 +191,15 @@ protected void lockTablesForSchemaSnapshot(ChangeEventSourceContext sourceContex
                 // FLUSH TABLES resets TX and isolation level
                 connection.executeWithoutCommitting("SET TRANSACTION ISOLATION LEVEL REPEATABLE READ");
             }
+
+            if (connectorConfig.getSnapshotLockingStrategy().isSingleTransaction()) {
+                try {
+                    connection.executeWithoutCommitting("START TRANSACTION WITH CONSISTENT SNAPSHOT");
+                }
+                catch (SQLException e) {
+                    LOGGER.warn("It's possible to receive duplicated events between snapshot and streaming. It can be caused by an unsupported engine", e);
```
jpechane: Emit `ERROR` and rethrow the Exception.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/snapshot/lock/AtLeastOnceSnapshotLock.java
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package io.debezium.connector.mariadb.snapshot.lock;
+
+import java.util.Map;
+
+import io.debezium.connector.mariadb.MariaDbConnectorConfig;
+import io.debezium.snapshot.spi.SnapshotLock;
+
+public class AtLeastOnceSnapshotLock extends DefaultSnapshotLock implements SnapshotLock {
```
jpechane: I think we agreed on putting this mode into binlog module and make it available for both MariaDB and MySQL in case the user uses arbitrary storage module for which we cannt be sure that TX with consistent snapshot is working.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
```diff
@@ -72,6 +72,12 @@ public enum SnapshotLockingMode implements EnumeratedValue {
          * set to schema_only or schema_only_recovery.
          */
         NONE("none"),
+
+        /**
+         * the connector holds the global read lock but does not set a transaction `START TRANSACTION WITH CONSISTENT SNAPSHOT` to ensure
+         * that the snapshot process will have a view from the mvcc. This mode is valid for myRocks engine that doesn't provide mvcc support
+         */
+        AT_LEAST_ONCE("at_least_once"),
```
jpechane: Do we need new snapshot mode? Should not this be done via `snapshot.locking.mode`? I can envision name like `minimal_at_least_once`.

jpechane (commented):
@kmos I think we are almost there. I proposed some changes in naming and error handling and I believe it should be configured at lock level not snaphost level.

rk3rn3r:
Is there an impact on the datastore performance? Is this locking the affected tables more than it did before? What happens on busy database servers, if this adds a longer table lock as before?

debezium-connector-binlog/src/test/resources/ddl/innodb_mvcc.sql
```diff
@@ -0,0 +1,106 @@
+CREATE TABLE test_table (
+					id INT AUTO_INCREMENT PRIMARY KEY,
+					name VARCHAR(255),
+					email VARCHAR(255),
+					last_updated TIMESTAMP);
+
+INSERT INTO test_table (name,email,last_updated) VALUES ('John Doe1','johndoe@gmail.com',NOW());
```
rk3rn3r, line 7: IT nerds start counting with 0, no? ;))
btw, what are you doing with my email address here?

(don't take this comment serious)

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
```diff
@@ -89,13 +95,17 @@ public String getValue() {
...
         public boolean usesMinimalLocking() {
-            return value.equals(MINIMAL.value);
+            return value.equals(MINIMAL.value) || value.equals(AT_LEAST_ONCE.value);
         }
 
         public boolean usesLocking() {
             return !value.equals(NONE.value);
         }
 
+        public boolean useSingleTransaction() {
```
rk3rn3r: Is there an option  for 1 transaction per table? If my previous comment about table locks vs performance is valid and the transaction locks all tables until  the entire snapshot finishes with `COMMIT`, then 1 transaction per table might help on busy database servers that have high write load. wdyt?

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
kmos: I am bit confused :eyes: . If I understand correctly it's already a `snapshot.locking.mode` this one. Now I have moved it on `BinlogConfiguration` to be in common between mariaDB and mysql. We need another mode to specify if it's necessary to apply the query `START TRANSACTION WITH CONSISTENT SNAPSHOT`

debezium-connector-binlog/src/test/resources/ddl/innodb_mvcc.sql
kmos: There's a John Doe inside each of us, and it's screaming

kmos:
> Is there an impact on the datastore performance? Is this locking the affected tables more than it did before? What happens on busy database servers, if this adds a longer table lock as before?

@rk3rn3r good question. The answer is *yes or not*. It depends in what do you mean (or care) in performance degradation. The quick answer is *yes* because nothing is for free, but let me explain :writing_hand: (and sorry for the wall of text):

Actually `START TRANSACTION WITH CONSISTENT SNAPSHOT;` is part of the [MVCC mechanism](https://dl.acm.org/doi/abs/10.1145/319996.319998) that provides **consistent reads** without blocking **writes**. The idea is that each **write** produces a new copy (or *version*) of that data. Since **writes** don't overwrite each other, the **reads** can read any version without any **lock**.

so your question:

> if this adds a longer table lock as before?

if you mean longer **write lock**, the answer is **no**. The performance are perfectly the same as **writes** produces new versions.

> Is there an impact on the datastore performance? Is this locking the affected tables more than it did before? What happens on busy database servers [...]

it's clear that such mechanism in a system under heavy (**write**) load, forces the engine to keep more data (in this case the **undo log**) to reconstruct past versions of rows for the snapshot. which means:
- Increase disk I/O
- longer purge operations (clean-up of old undo entries)
- transactions can be longer (for the undo log volume)

So **yes** in the context that the operations in the database can increase I/O due to **undo log** dimension but (as I said before) **without locking any table**.

**so what really do** `START TRANSACTION WITH CONSISTENT SNAPSHOT;` **??**
For what I have read/search, it takes a consistent **read view** taking into account active transactions. Every row in InnoDB has two hidden fields:
- `trx_id`: The transaction that last modified the row
- `roll_pointer`: Points to undo data (old version)

During the snapshot, innoDB compares `trx_id` of each row against the **read view**. If the row was created or modified by a transaction that was still active when the snapshot began, the data are not **visibile** .


The snapshot (**opinion alert**) should guarantee the consistency of the data. As nothing is free, we decide to have some degradation on database performance. If the user needs that the database keeps performances during the snapshot phase, it can disable this mode but we should keep it consistent as default. 

so WDYT? is it reasonable? 

**side note**: I didn't investigate yet for postgreSQL but (based on what I know) there is no need to apply that query because it's already addressed when you `FLUSH TABLES WITH READ LOCK;` and disable autocommit.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
jpechane: @kmos My bad, I looked at the diff only and somehow mixed snapshot mode with snapshot locking mode.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
kmos: > Is there an option for 1 transaction per table? If my previous comment about table locks vs performance is valid and the transaction locks all tables until the entire snapshot finishes with `COMMIT`, then 1 transaction per table might help on busy database servers that have high write load. wdyt?


I think that the name is misleading but I have used this one to refer to the option used by `mysqldump` ([reference](https://dev.mysql.com/doc/refman/8.4/en/mysqldump.html#option_mysqldump_single-transaction)) which is the same. referring to the docs:

>It dumps the consistent state of the database at the time when [START TRANSACTION](https://dev.mysql.com/doc/refman/8.4/en/commit.html) was issued without blocking any applications.

it's not intended as a transaction that makes any `write lock` (or `read lock`). 

> then 1 transaction per table might help on busy database servers that have high write load.

At the beginning to this journey I thought the same but let me explain:

at point *5* of the process:

```sql
-- 5. release the global read lock — snapshot is safe
UNLOCK TABLES;
```
When we reach the point 5, we have a perfect view of the system state (rows versions + binlog position) and we can release the global lock. And we start reading the table without `commit` in order to keep using the version defined at the beginning. It's impossible to `commit` or do this process table by table because:

- you will lose the point in time of your snapshot which is something global and not local
- the `binlog` can go in another point in time which lead to inconsistent snapshot

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorConfig.java
kmos: don't worry! let me know if it's reasonable now moved in `binlog` module :rocket:

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/snapshot/lock/MinimalAtLeastOnceSnapshotLock.java
```diff
@@ -0,0 +1,31 @@
...
+
+import static io.debezium.connector.binlog.BinlogConnectorConfig.SnapshotLockingMode.MINIMAL_AT_LEAST_ONCE;
+
+import java.time.Duration;
+import java.util.Map;
+import java.util.Optional;
+
+import io.debezium.snapshot.spi.SnapshotLock;
+
+public class MinimalAtLeastOnceSnapshotLock implements SnapshotLock {
```
Naros, line 17: So if we look at all the `SnapshotLock` modes the only ones that are not annotated with `@ConnectorSpecific` are those in `debezium-core`. @mfvitale can you remind me, do you think we'll need that here?

Naros (commented):
LGTM, I've simply left one comment I'd like feedback from @mfvitale on.

I recall we had a class loader issue with ServiceLoader usages in the past when we first introduced `SnapshotLock` impls, but I don't remember the specifics.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/snapshot/lock/MinimalAtLeastOnceSnapshotLock.java
mfvitale: The `@ConnectorSpecific` is used when there is the same snapshot mode in the core and in a specific connector. 

In this case `MinimalAtLeastOnceSnapshotLock` is just for binlog connector so it's fine.

jpechane:
@kmos Applied, thanks

PR #6499, opened 2025-06-10
DBZ-9126 Add TF images for Oracle 19.3.0 SE

vjuranek:
https://issues.redhat.com/browse/DBZ-9126


vjuranek:
To make it working we need to add SE composes into Testing farm. Filed [TFT-3413](https://issues.redhat.com/browse/TFT-3413) for it.

PR #6500, opened 2025-06-12
DBZ-9141: Avoid messing up fields while using fields.include.list

bpaquet:
Before the PR, the data.get(field) is doing a lookup by id. But the id of field come from filteredData, so schema can be different, especially if we use `fields.include.list`.


bpaquet:
I did not include test because the fix is really small, and existing tests seems to be enough for non regression.

We can reconsider this if needed.

jpechane:
@bpaquet Good catch, thanks!

PR #6503, opened 2025-06-13
DBZ-xxyy Remove redundancy in intializing DataTypeResolver

akanimesh7:


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/MySqlAntlrDdlParser.java
```diff
@@ -50,6 +50,7 @@ public class MySqlAntlrDdlParser extends AntlrDdlParser<MySqlLexer, MySqlParser>
     private final ConcurrentMap<String, String> charsetNameForDatabase = new ConcurrentHashMap<>();
     private final TableFilter tableFilter;
     private final BinlogCharsetRegistry charsetRegistry;
+    private static final DataTypeResolver dataTypeResolver = initializeDataTypeResolver();
```
Naros: Are you sure you intended the variable to be static here? I'd prefer we avoid that much like you have done in the MariaDB and Oracle implementations.

Naros (changes requested):

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/MySqlAntlrDdlParser.java
akanimesh7: I intended them to be static only. This resolver should be bound to the type of parser right and not the specifc instantiated object ? 

If we don't keep it static, two objects of MySqlAntlrDdlParser.java will both have their own objects of resolver. However, it isn't needed and increases memory footprint ?

On a parallel note, as of today all the type specific parsers are instantiated only once.  Hence it would not matter. But would still love to hear your thoughts.

akanimesh7:
@Naros replied to comment

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/MySqlAntlrDdlParser.java
jpechane: @akanimesh7 I'd prefer to have it as non-static attribut esp. as you mentioned the parsers are instatiated only once.

akanimesh7:
@jpechane 
Accomodated the suggestion to not use static.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/MySqlAntlrDdlParser.java
akanimesh7: Just curious on this one. Does using static in this case lead to some overhead?

jpechane: @akanimesh7 No, in this case it is more matter of style/opinion on the code. I don't see any perf impact.

jpechane:
@akanimesh7 Applied, thanks

PR #6504, opened 2025-06-13
DBZ-9049 - Support for Informix 15

nrkljo:


jpechane:
Rebased on the latest main. Will be merged if Ifx (15) tests passes

PR #6506, opened 2025-06-13
DBZ-9161 Postgres: log errors from keepalive thread for replication

james-johnston-thumbtack:
A background keepalive thread is used to periodically send status updates on the replication stream. If this status update fails, we should immediately log the failure.

The old code catches and rethrows the exception.  This is a problem because:

1.  The exception is not logged at all.  The ExecutorService would rethrow the exception when the "get" function is called on the Future.  However, we never check for results of this future at all by calling "get", so... nothing gets logged.

2.  Even if we did call "get" on the Future in the stopKeepAlive function, the exception would not be immediately logged.  Having the log message be outputted at the correct time the exception actually happened is better for end-user debugging.

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
@@ -827,7 +827,10 @@ public class PostgresReplicationConnection extends JdbcConnection implements Rep
                                 metronome.pause();
                             }
                             catch (Exception exp) {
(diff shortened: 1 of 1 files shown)


matthewmcnew (approved):

jpechane:
@james-johnston-thumbtack LGTM, thanks. But could you please create a separate Jira bug for it and associate the PR and commit with it?

james-johnston-thumbtack:
> @james-johnston-thumbtack LGTM, thanks. But could you please create a separate Jira bug for it and associate the PR and commit with it?

@jpechane sure, see https://issues.redhat.com/browse/DBZ-9161

jpechane:
@james-johnston-thumbtack Applied, thanks

PR #6507, opened 2025-06-16
DBZ-9146 Remove nominalTime facet from OpenLineage events

dolfinus:
According to OpenLineage docs:
https://openlineage.io/docs/spec/facets/run-facets/nominal_time
> The nominal usually means the time the job run was expected to run (like a scheduled time)

But facet contains current time, which is already passed to eventTime field.


jpechane:
@dolfinus Applied, thanks

PR #6511, opened 2025-06-17
DBZ-9153 Typo in the example code of the Custom Converter

Living-Hell:
This commit fixes the closing bracket typo in the example code of the Custom Converter 'IsbnConverter' due to an extra closing bracket on line 15, which was causing the code to throw an error.

image


Naros (approved):
Great catch @Living-Hell :tada:

Living-Hell:
> Great catch @Living-Hell 🎉

@Naros Thank you! : )

PR #6514, opened 2025-06-18
DBZ-9150: Decouple LSN Flush from Offset Committer Thread and Introduce Timeout Handling

rajdangwal:
Addressing https://issues.redhat.com/browse/DBZ-9150

- Moved the offset flush logic as follows:
  - The `commit` method now only sets a flag `shouldPerformCommit = true`.
  - The actual flush is triggered during the next poll invocation, which checks the flag, performs the LSN flush if set, and then resets the flag.
- Introduced timeout handling for the LSN flush operation:
  - A default timeout of `30 seconds` is configured for the LSN flush to complete.
  - If the timeout is reached, the connector's response depends on a configurable behavior: By default, the connector fails, since the inability to flush the LSN is considered a critical issue that requires immediate attention. Alternatively, connector cab be configured to ignore the timeouts and continue processing.


rajdangwal:
Can we please re-run the build job?
The failure looks irrelevant.

jpechane:
@rajdangwal Re-executed. We really need to fix testsuite first before we'll move forward with merging of more PRs.

rajdangwal:
@jpechane I have fixed the Postgres related tests. They were not related to my changes though.

Also, the failures in other connectors are not related to my changes. The Spanner tests seems flaky as it was passed in previous iteration. 

For SqlServer one I cannot debug locally as I have a `ARM64` machine and the docker image is not available for it.
From the added logs, it seems the tests are failing because of the log position validity check:
```
INFO   ||||  Oldest SCN in logs is '0000002A00000B60003A'   [io.debezium.connector.sqlserver.SqlServerConnection]
INFO   ||||  Stored LSN is 'NULL'   [io.debezium.connector.sqlserver.SqlServerConnection]
WARN   ||||  Last recorded offset is no longer available on the server.   [io.debezium.connector.common.BaseSourceTask]
ERROR  SchemaHistoryTopicIT|||engine|  1 task(s) out of 1 failed to start.   [io.debezium.embedded.async.AsyncEmbeddedEngine]
```

jpechane:
@rajdangwal Just a recommendations - by coincidence there is https://debezium.zulipchat.com/#narrow/channel/302533-dev/topic/Running.20Integration.20tests.20on.20Apple.20Silicon
It might be useful for you in future too.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresStreamingChangeEventSource.java
```diff
@@ -434,33 +438,77 @@ private void maybeWarnAboutGrowingWalBacklog(boolean dispatched) {
...
+    /**
+     * Handles the scenario when an LSN flush timeout occurs.
+     *
+     * @param lsn the LSN that failed to flush
+     */
+    private void handleTimeout(Lsn lsn) {
+        String action = connectorConfig.lsnFlushTimeoutAction().getValue();
+        long timeoutMillis = connectorConfig.lsnFlushTimeout().toMillis();
+
+        if ("fail".equals(action)) {
```
jpechane: Why string based condition when we have enum?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
```diff
@@ -993,6 +993,88 @@ public static AutoCreateMode parse(String value, String defaultValue) {
...
+            .withImportance(Importance.MEDIUM)
+            .withDescription("Maximum time in milliseconds to wait for LSN flush operation to complete. "
+                    + "If the flush operation does not complete within this timeout, the action specified by "
+                    + "lsn.flush.timeout.action will be taken. Defaults to 30 seconds.")
+            .withValidation(Field::isPositiveInteger);
+
+    /**
+     * The set of predefined LSN flush timeout action options
+     */
+    public enum LsnFlushTimeoutAction implements EnumeratedValue {
```
jpechane, line 1011: Maybe we should use the same options we do elsewhere so fail/warn/ignore - where ignore logs message on DEBUG level.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresStreamingChangeEventSource.java
```diff
@@ -434,33 +438,77 @@ private void maybeWarnAboutGrowingWalBacklog(boolean dispatched) {
...
+            try {
+                future.get(connectorConfig.lsnFlushTimeout().toMillis(), TimeUnit.MILLISECONDS);
             }
-            else {
-                LOGGER.debug("Streaming has already stopped, ignoring commit callback...");
+            catch (TimeoutException e) {
+                // Handle the timeout according to configuration
+                handleTimeout(lsn);
+            }
+            catch (InterruptedException e) {
```
jpechane, line 482: What will happen during regular connector shutdown? Will the LSN be properly flueshed then or not?

rajdangwal: Thanks for pointing that out.
During connector shutdown, there is a possibility that the latest LSN might not get flushed. This is because the `commit()` method, which is invoked during task shutdown, now only sets a flag, and since `poll()` won't be called afterward, the actual flush doesn't happen.

Do you anticipate any major concerns with this approach? From my side, it seems acceptable, but I would appreciate your thoughts.
The offsets will still be updated and persisted on the connector side, so on restart, the connector will resume from the correct position. The only minor drawback is that the server might retain older WALs a bit longer, until the connector flushes the next LSN after restarting.

jpechane: Well, there is another issue. What if the connector is stopped, removed or not restarted for quite some time. In this case the WAL will be kept for much long than a bit longer. From my PoV we effectivelly do not clean-up resources and if possible I'd like to avoid that.

rajdangwal: You're right that if the connector is stopped and the replication slot isn’t dropped, the WAL files will remain indefinitely. But that's already the case today, regardless of this change.

We do periodically flush the LSNs (typically every minute, based on default `offset.flush.interval.ms`), so in practice, a connector stop would only delay the latest flush. The amount of additional WAL retained due to that should be minimal in normal conditions.

Unfortunately, I haven’t found any hook other than commit that reliably runs after flushing uncommitted records and before the connector stops. So, this seems to be the most suitable option.
That said, if you have any other suggestions in mind, I would be happy to explore and try implementing them.
Thanks.

jpechane:
Rebase don the latest main and added one commit with if converted to switch

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresStreamingChangeEventSource.java
jpechane: @rajdangwal Can we call `performCommit` from stop method after committing offsets?

rajdangwal: Hi @jpechane 
It looks like I had overlooked the fact that the task's `stop` method is guaranteed to run after the `commit`.
So, we will have the latest offsets available when flushing in `stop`.

I've now added the `performCommit` call inside the `stop` method. Thanks for pointing that out!

PR #6515, opened 2025-06-18
Connector/cockroachdb

viragtripathi:


jpechane:
@viragtripathi Thanks for the PR! One question related to the orgnization. I suppose we can expect you'll take more active lead in the connector. I'd prefer if the connector then would reside in a separate project (see Informix, IBMi, Db2 and others). In that case I could you give you committers rights too immediately. We'll then integrate the repo into our release train etc. WDYT?

viragtripathi:
> @viragtripathi Thanks for the PR! One question related to the orgnization. I suppose we can expect you'll take more active lead in the connector. I'd prefer if the connector then would reside in a separate project (see Informix, IBMi, Db2 and others). In that case I could you give you committers rights too immediately. We'll then integrate the repo into our release train etc. WDYT?

Hi @jpechane — thank you for the quick feedback!

You're right that I do intend to take a more active lead on this connector going forward, and I'd be happy to help maintain and iterate on it long-term.

That said, I’m curious about your suggestion to host this in a separate project. Since CockroachDB is wire-compatible with PostgreSQL and leverages much of the same ecosystem (JDBC, wire protocol, data types), my initial assumption was that it fits within the main Debezium repo alongside connectors like Postgres, MySQL, and MongoDB — especially since the core Debezium APIs are being used here (e.g., OffsetContext, EventDispatcher, etc.).

The primary difference lies in the CDC mechanism: CockroachDB uses a push-based changefeed (enriched envelope) rather than logical decoding. But otherwise, the connector structure mirrors the Debezium style and is modeled on the Postgres implementation.

So my questions are:

Is the push-based changefeed mechanism the reason you're leaning toward an external project?

Would you envision this as a separate repo like debezium-connector-cockroachdb, to be integrated into the release train like Informix or Db2?

Are there specific trade-offs (e.g., testing, dependency management, or CI) that make an external repo better in this case?

Or is this mainly a maintainership boundary to help contributors manage connector evolution more independently?

Happy to proceed either way — just want to make sure I understand the long-term expectations and that the connector’s placement aligns with the broader Debezium architecture.

Thanks again! 🙏

— Virag

jpechane:
@viragtripathi Hi, the placement in a separate project (under Debezium org) is just only about organizational issues, not about technical ones - so it i exactly mainly a maintainership boundary to help contributors manage connector evolution more independently. In case of Db2 a lot of code/functionality is shared with SQL Server connector with no issues.

The connector would be included in release train as soon as the initial PR is reviewd by the team and merged. There is not probation period or something like it. We'll just mark the connector as incubating which we did in the past and we do with some other part of Debezium now too. The reason is not about quality but a kind of message to the user that they might be backward compatinbility changes. This is usually useful as it allows us more freedom. We usually don't break the compatibility but at the beginning it might be necessary. Just to exaplin by that term I mean for eaxmaple changes into message schema etc.

jpechane:
@viragtripathi If you don't mind I went ahead aand created https://github.com/debezium/debezium-connector-cockroachdb
You were also invited into group `cockroachdb-conmmitters` that has write access to this repo.

I also create a new channel on Zulip - https://debezium.zulipchat.com/#narrow/channel/510960-community-cockroachdb

viragtripathi:
> @viragtripathi If you don't mind I went ahead aand created https://github.com/debezium/debezium-connector-cockroachdb You were also invited into group `cockroachdb-conmmitters` that has write access to this repo.
> 
> I also create a new channel on Zulip - https://debezium.zulipchat.com/#narrow/channel/510960-community-cockroachdb

Thanks so much, @jpechane - I truly appreciate you setting up the repo and granting access!

This all sounds perfect: the separate project structure, the incubating status to allow for iteration, and the Zulip channel for collaboration. I’ll start migrating the existing connector work into debezium-connector-cockroachdb and will follow the established patterns from other external connectors like Db2 and Informix.

Looking forward to working closely with the Debezium team and community to get the CockroachDB connector production-ready. Thanks again for the trust and support!

viragtripathi:
https://github.com/debezium/debezium-connector-cockroachdb

PR #6516, opened 2025-06-19
DBZ-9154 Make Executor Shutdown Timeout configurable

nrkljo:


jpechane:
@nrkljo I pushed one commit changing default valune handling.

PR #6517, opened 2025-06-19
DBZ-8959 Lifecycle CDI event support

Naros:
https://issues.redhat.com/browse/DBZ-8959

### DIFF
--- a/quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/events/AbstractDebeziumLifecycleEvent.java
+++ b/quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/events/AbstractDebeziumLifecycleEvent.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.quarkus.debezium.engine.events;
+
+import io.debezium.runtime.Debezium;
+
+/**
+ * Abstract base class for all Debezium lifecycle events.
+ *
+ * @author Chris Cranford
+ */
+public abstract class AbstractDebeziumLifecycleEvent {
+
+    private final Debezium engine;
+
+    public AbstractDebeziumLifecycleEvent(Debezium engine) {
+        this.engine = engine;
+    }
+
+    public Debezium getEngine() {
+        return engine;
+    }
+}
--- a/quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/EngineProcessor.java
+++ b/quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/EngineProcessor.java
@@ -64,6 +64,7 @@ import io.quarkus.debezium.deployment.items.DebeziumConnectorBuildItem;
 import io.quarkus.debezium.deployment.items.DebeziumGeneratedInvokerBuildItem;
 import io.quarkus.debezium.deployment.items.DebeziumMediatorBuildItem;
 import io.quarkus.debezium.engine.DebeziumRecorder;
+import io.quarkus.debezium.engine.DefaultStateHandler;
 import io.quarkus.debezium.engine.capture.CapturingInvoker;
 import io.quarkus.debezium.engine.capture.CapturingSourceRecordInvokerRegistryProducer;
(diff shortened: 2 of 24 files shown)


quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/events/AbstractDebeziumLifecycleEvent.java
```diff
@@ -0,0 +1,26 @@
...
+package io.quarkus.debezium.engine.events;
+
+import io.debezium.runtime.Debezium;
+
+/**
+ * Abstract base class for all Debezium lifecycle events.
+ *
+ * @author Chris Cranford
+ */
+public abstract class AbstractDebeziumLifecycleEvent {
```
Naros: In hindsight, I suspect we want to move these events to the SPI module.  wdyt?

Naros:
@jpechane I think we had discussed this before but from your PoV, is it acceptable for our first pass to be 1 Quarkus App = 1 Debezium Engine instance? We can certainly look at adding `@Qualifier` support to have multi-engine capabilities within a single app, but I do think that's a decision we need to make sooner rather than latter, given how much it does impact how we handle item registrations, proxy/class generation, etc.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/events/AbstractDebeziumLifecycleEvent.java
kmos: Yeah for sure. I was thinking that maybe we can have these events as `sealed class` something like: 

```java
public sealed abstract class DebeziumLifecycleEvent permits ConnectorStartedEvent, DebeziumCompletionEvent {
    private final Debezium engine;

    public DebeziumLifecycleEvent(Debezium engine) {
        this.engine = engine;
    }

    public Debezium getEngine() {
        return engine;
    }
}

public final class ConnectorStartedEvent extends AbstractDebeziumLifecycleEvent {
    public ConnectorStartedEvent(Debezium engine) {
        super(engine);
    }
}

public final class DebeziumCompletionEvent extends AbstractDebeziumLifecycleEvent {
    public DebeziumCompletionEvent(Debezium engine) {
        super(engine);
    }
}
```

jpechane: I agree with the separate SPI module but I am not sold on sealed class. What danger do you see in keeping it open?

jpechane:
@Naros I am in for one app == one engine relation at the start. Let's get this out soon so we can get community feedback.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/events/AbstractDebeziumLifecycleEvent.java
kmos: I don't have any strong opinion, we can have two main things with sealed class:

1. enabling exhaustiveness analysis
2. better performance (the hierarchy is resolved statically)

let me show you a possible DE for a common use case with events + pattern matching with a sealed class and java `21`:

```java
@ApplicationScoped
public class DebeziumEventHandler {
    public void handleEvent(@Observes DebeziumLifecycleEvent event) {
        switch (event) { // exhaustive
            case ConnectorStartedEvent connectorStartedEvent -> {
            }
            case DebeziumCompletionEvent debeziumCompletionEvent -> {
            }
        }
    }
}
```

with abstract:

```java
@ApplicationScoped
public class DebeziumEventHandler {
    public void handleEvent(@Observes AbstractDebeziumLifecycleEvent event) {
        switch (event) { // need default
            default -> throw new RuntimeException("");
        }
    }
}
```

but for me are fine both also because you can reach the same with multiple method that `@observes` different events

kmos (approved):

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/events/AbstractDebeziumLifecycleEvent.java
jpechane: Good usecase! Yes in that context it make sense.

Naros: I have no particular opinion either way, we can add them as `sealed` and change it later as needed.

PR #6518, opened 2025-06-20
DBZ-9162 Fix truncate present in history topic if it is a skipped operation

akanimesh7:
Fixes a case missed in https://issues.redhat.com/browse/DBZ-9085


jpechane:
@akanimesh7 Could we have a test for this case so it is not missed in future?

akanimesh7:
@jpechane I can try.

akanimesh7:
@jpechane Added a test for the same

jpechane:
@akanimesh7 Applied, thanks

PR #6520, opened 2025-06-21
DBZ-9168: Escape special characters in MySQL connector

morozov:
The implementation of `BinlogSnapshotChangeEventSource#quote(String)` and `BinlogSnapshotChangeEventSource#quote(TableId)` is incorrect as it doesn't quote the backtick character properly. The source should use the corresponding `JdbcConnection` methods instead.


jpechane:
@morozov Applied, thanks

PR #6521, opened 2025-06-22
DBZ-9169: Use JdbcConnection#quoteIdentifier(), remove JdbcConnection#quotedColumnIdString()

morozov:
This PR removes the deprecated `JdbcConnection#quotedColumnIdString()` in favor of `JdbcConnection#quoteIdentifier()`.

It doesn't include any additional tests based on the following:
1. https://github.com/debezium/debezium/pull/6464 and https://github.com/debezium/debezium/pull/6520 demonstrate that `JdbcConnection#quotedColumnIdString()` correctly quotes special characters on SQL Server, MySQL and MariaDB.
2. The implementation of `PostgresConnection#quotedColumnIdString()` is identical to `JdbcConnection#quotedColumnIdString()`, so it doesn't require additional testing.

This change also affects the Oracle connector. The thing is that Oracle doesn't allow double quotes in object names (see the [documentation](https://docs.oracle.com/en/database/oracle/oracle-database/23/sqlrf/Database-Object-Names-and-Qualifiers.html)):
> Quoted identifiers can contain any characters and punctuations marks as well as spaces. However, neither quoted nor nonquoted identifiers can contain double quotation marks or the null character (`\0`).

So, technically, neither the old nor the new implementations are correct for Oracle. However, the Oracle connector already uses `TableId#toDoubleQuotedString()`, which doubles the double quote:
https://github.com/debezium/debezium/blob/3a53db5dbb74b7371943355d79ec3b96f47ab6cc/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java#L260-L262

Therefore, using `JdbcConnection#quoteIdentifier()` in the Oracle connector at least makes it consistent with the rest of the queries generated by the connector.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/BaseChangeRecordEmitter.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/BaseChangeRecordEmitter.java
@@ -147,7 +147,7 @@ public abstract class BaseChangeRecordEmitter<T> extends RelationalChangeRecordE
     private String getReselectQuery(List<Column> reselectColumns, Table table, OracleConnection connection) {
         final TableId id = new TableId(null, table.id().schema(), table.id().table());
         final StringBuilder query = new StringBuilder("SELECT ")
-                .append(reselectColumns.stream().map(c -> connection.quotedColumnIdString(c.name())).collect(Collectors.joining(", ")))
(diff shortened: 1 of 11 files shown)


morozov:
Please let me know if any if the test failures look relevant. The ones I've examined are caused by timeouts. The Oracle failures all happened in different tests (i.e. couldn't be caused by a flaw in the logic).

jpechane:
@morozov Applied, thanks!

PR #6523, opened 2025-06-23
DBZ-9170 Align configuration for post processors to transforms, predicates

kmos:
## Context

Actually the configuration for post processing is the following:

```txt
  "post.processors": "reselector", 
  "reselector.type": "io.debezium.processors.reselect.ReselectColumnsPostProcessor", 
  "reselector.reselect.columns.include.list": "<schema>.<table>:<column>,<schema>.<table>:<column>", 
  "reselector.reselect.unavailable.values": "true", 
  "reselector.reselect.null.values": "true" 
  "reselector.reselect.use.event.key": "false" 
  "reselector.reselect.error.handling.mode": "WARN"
```

which differs from the canonical convention used for `transformers`, `predicates` like:

```txt
transforms=t0
transforms.t0.add.fields=op,table
transforms.t0.add.headers=db,table
transforms.t0.negate=false
```

## Changes
We support both aligning the new configuration to `transformers` and `predicates`.

```txt
  "post.processors": "reselector", 
  "post.processors.reselector.type": "io.debezium.processors.reselect.ReselectColumnsPostProcessor", 
  "post.processors.reselector.reselect.columns.include.list": "<schema>.<table>:<column>,<schema>.<table>:<column>", 
  "post.processors.reselector.reselect.unavailable.values": "true", 
  "post.processors.reselector.reselect.null.values": "true" 
  "post.processors.reselector.reselect.use.event.key": "false" 
  "post.processors.reselector.reselect.error.handling.mode": "WARN"
```

PR #6526, opened 2025-06-23
DBZ 8965 Debezium Engine Quarkus Extension: introduce PostProcessor handler

kmos:
## Context

Post processors perform lightweight, per-message mutations, similar to the modifications that are performed by single message transformations (SMTs). Quarkus developers should be able to define `post-processor` in the *normal* way and using a Quarkus way.

## Decision
In the context of Quarkus, we introduce an annotation `PostProcessing` in order to create `post-processor` inside the Quarkus CDI. like in this example:

```java
    @ApplicationScoped
    static class PostProcessingHandler {

        @PostProcessing()
        public void postProcessing(Object key, Struct struct) {
             // do your stuff
        }
    }
```

in this PR we introduce:

- [X]  `PostProcessing` annotation
- [X]  native build for PostProcessing generated classes

## References

#depends https://github.com/debezium/debezium/pull/6523
#closes https://issues.redhat.com/browse/DBZ-8965
#link: https://github.com/debezium/debezium-design-documents/pull/16

## Side notes

We have introduced a `factory` to retrieve CDI beans (`post-processors`) into the debezium core. Due to time and complexity of `BeanRegistry` and `ServiceRegistry`, we have introduced it in `debezium-core/src/main/java/io/debezium/processors/PostProcessorRegistryServiceProvider.java` . [We should evaluate a better approach](https://issues.redhat.com/browse/DBZ-8958) taking into account the lesson learned from #6458

### DIFF
--- a/debezium-core/src/main/java/io/debezium/processors/PostProcessorRegistryServiceProvider.java
+++ b/debezium-core/src/main/java/io/debezium/processors/PostProcessorRegistryServiceProvider.java
@@ -5,8 +5,10 @@
  */
 package io.debezium.processors;
 
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
+import java.util.ServiceLoader;
 import java.util.stream.Collectors;
 
 import org.slf4j.Logger;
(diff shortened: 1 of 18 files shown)


debezium-core/src/main/java/io/debezium/processors/PostProcessorRegistryServiceProvider.java
```diff
@@ -31,17 +33,32 @@ public class PostProcessorRegistryServiceProvider implements ServiceProvider<Pos
 
     private static final String POST_PROCESSOR_MISS_CONFIGURATION_ERROR_MESSAGE = "Post processor '%s' is missing '%s.type' and/or '%s.<option>' configurations";
     private final String TYPE_SUFFIX = ".type";
+    private final ServiceLoader<PostProcessorFactory> postProcessorFactory = ServiceLoader.load(PostProcessorFactory.class);
```
Naros, line 36: I would prefer a slightly different architectural & strategic change here.

Rather than using `ServiceLoader` directly & side-step the `ServiceRegistry`, could we instead create a dedicated dependency injection service & matching provider, of which there would be a Quarkus/Arc implementation, and this would lay the framework for future DI frameworks like Spring.

Then here we ask the `ServiceRegistry` for this DI service, it would return us the Quarkus/Arc implementation behind that common DI service interface, and we can then ask to get a list of all beans with a given interface.

This means we pay the cost for this hook once and we can use it for PostProcessors or anything in the the future that we want to hook in.  What do you think @kmos ?

kmos: This is exactly one of the reason behind #6458   :rocket: for the refactoring of `baseSourceTask` and `ServiceRegistry/BeanRegistry`, as well as for this [issue](https://issues.redhat.com/browse/DBZ-8958). Indeed, for now, I took this shortcut both for the `PostProcessor`, `Heartbeat` and `Converter` (using `ServiceLoader` inside the class).

My only point is whether to implement this now or wait until we tackle this [issue](https://issues.redhat.com/browse/DBZ-8958), so we can incorporate the lesson learned from #6458. I have left that activity has the last one of this release to address all the injections with a single solution and have an entire review dedicated on it. WDYT?

Naros: That's not a bad idea, and it gives us the chance to take a birds eye view of the picture to make sure that we design the Quarkus/Arc/DI stuff well that fits all our needs and leaves room for alternatives.

PR #6527, opened 2025-06-24
DBZ-9173: Documentation on running integration tests on Apple Silicon

morozov:


README.md
```diff
@@ -98,6 +98,39 @@ In order to run testcontainers against [colima](https://github.com/abiosoft/coli
...
     export DOCKER_HOST="unix://${HOME}/.colima/default/docker.sock"
 
+#### Docker Desktop on Apple Silicon
+
+When running on Apple Silicon, the Docker Maven Plugin needs to be configured to use the `linux/amd64` platform. Additionally, it should be pointed to the Docker socket created by Docker Desktop using the `docker.host` system property.
+
+For example:
+
+```shell
+mvn docker:start \
```
rk3rn3r, line 108:
```suggestion
./mvnw docker:start \
```

README.md
```diff
@@ -98,6 +98,39 @@ In order to run testcontainers against [colima](https://github.com/abiosoft/coli
...
+        <docker.platform>linux/amd64</docker.platform>
+      </properties>
+    </profile>
+  </profiles>
+
+  <activeProfiles>
+    <activeProfile>default</activeProfile>
+  </activeProfiles>
+</settings>
+```
```
rk3rn3r, line 133: Should this last paragraph be reworded that configure this in `settings.xml` is the preferred option and not just recommended for simplicity?

README.md
morozov: There are a few code snippets after this one that use `mvn` directly. Let’s keep this consistent for now. If it’s important to recommend the wrapper, this could be done separately.

README.md
morozov: I don’t think so. It should be up to the developer to decide what they prefer. Their choice will have no effect on the end result of their contribution.

jpechane: I agree, this depends on developer

README.md
jpechane: @rk3rn3r Good point, but I gree that w'll keep `mvn` for consistency in this PR and a separate one that will use `mvnw` in both readme and contribiting guide would be better.

jpechane:
@morozov Applied, thanks

PR #6529, opened 2025-06-24
Add EOS support to main source connectors

vjuranek:
https://issues.redhat.com/browse/DBZ-9177

I intentionally do it in separate commits so we can easily revert it if there is any issue reported for given connector.


vjuranek:
@Naros yes, this is all we need, EOS framework itself is implemented by Kafka Connect and other stuff needed, like offset tracking etc. is already implemented by the connectors, thus we just have to declare we support it.
The only thing I need to add is to mention it somewhere in the documentation, which I'll do in separate PR.

PR #6536, opened 2025-06-26
DBZ-9025 Update snapshot links

jpechane:


jpechane:
@Naros I create a script that would update antora docs to reflect latest snapshots. Could you please take a look? I'd recommend to add it as GHA periodically executed every day at midnight UTC. WDYT?

Naros:
Hi @jpechane, the script works well. Should we make it executable? 

With the YML update, I think we can tackle this one of two ways:

1. We have GHA push a commit after modifying `antora.yml` and trigger a remote docs workflow build. This may require us to pass a PAT so it can safely push to `main`.
2. We modify antora.yml on website builds. This is where it gets a bit trickier because Antora handles the checkout automatically, so we would need to check the Rakefile to checkout the source first manually and then have the build use that rather than relying on checking out directly from GH.

There is also a third option, and I suspect this may be the cleanest, and that is we use this script as is, but we modify the filename so that it changes the `playbook.yml` file rather than `antora.yml` and we run this after the website is checked out and before we run the website builder.

This would require a small change to `install.adoc` to refer to the attributes from new names in the `playbook.yml` for now rather than `antora.yml` until this is fixed my Sonatype and we avoid the need to make any commits to the main repository? 

If you like this latter approach, I'll throw together the fixes and incorporate your script.

Naros:
Superseded by https://github.com/debezium/debezium/pull/6549

PR #6537, opened 2025-06-27
DBZ-9186 Avoid storing irrelevant DDL statements in history topic

akanimesh7:


akanimesh7:
Failing tests look unrelated to me.

Naros:
@akanimesh7, Gunnar also suggested skipping things around permissions, such as grants/revokes, too.

Naros:
Hi @akanimesh7 can you please take a look at the test failures? The test `BinlogTableMaintenanceStatementsIT` uses ANALYZE, OPTIMIZE, and REPAIR.

akanimesh7:
@Naros Yes. I see that. Fixed the test and added `REPAIR` also to the list of DDLs to not capture.

akanimesh7:
@Naros Tests are passing now. Plase do review.

The tests which are failing for SQL Server are passing on my machine. Hence those look to be flaky. I will raise another PR to fix the flakiness in them.

akanimesh7:
@Naros @jpechane Bump on this

Naros:
It does not look like we tested the Oracle test suite with the change, just kicked that off while I check the SQL Server failures, as some look suspicious.

Naros:
Hi @akanimesh7 so all the SQL Server errors are related to DBZ-9118, which I believe you worked on and somehow these were missed. I've logged DBZ-9217 for this. My concern here is that even setting `WHEN_NEEDED` in the SQL Server test as suggested by the error, then we get past this exception but then we don't get the expected output. It could be a faulty test, but would you have some cycles to take a look at that?

akanimesh7:
Hey @Naros Sure i'll check DBZ-9217. As mentioned in [this comment](https://github.com/debezium/debezium/pull/6537#issuecomment-3049138539), those tests are passing on my local. In limited time i spent, i had some ideas to fix them. Will create a PR with DBZ-9217.

akanimesh7:
@Naros 
I have made some changes in the sql server tests and i am able to get them passed now.
However, some of the other tests are failing, which look unrelated to me. 
LMK if you think otherwise.

vjuranek (approved):

Naros (approved):
Thanks @akanimesh7 LGTM as well.

PR #6538, opened 2025-06-27
DBZ-9158: postgres doc update

Kalla-Shubham:


documentation/modules/ROOT/pages/connectors/postgresql.adoc
```diff
@@ -4020,6 +4020,12 @@ For more information, see xref:postgresql-toasted-values[toasted values].
 |Determines whether the connector generates events with transaction boundaries and enriches change event envelopes with transaction metadata. Specify `true` if you want the connector to do this.
 For more information, see xref:postgresql-transaction-metadata[Transaction metadata].
 
+|[[postgresql-property-publish-via-partition-root]]<<postgresql-property-publish-via-partition-root, `+publish.via.partition.root+`>>
+|`false`
+|A boolean that determines whether the connector should auto-create publication for partitioned tables.
```
jpechane: @Kalla-Shubham I am not sure this text is accurate. The publication will be created, it just defines how the publication will be configured for partirioned tables.

Kalla-Shubham: Ack.
Right auto-create is a separate config so that mention here would not be correct.
let me know if you meant this only?

jpechane: Yes, that's the only issue I have with the PR.

Kalla-Shubham: @jpechane plz have a look now

documentation/modules/ROOT/pages/connectors/postgresql.adoc
```diff
@@ -4020,6 +4020,12 @@ For more information, see xref:postgresql-toasted-values[toasted values].
 |Determines whether the connector generates events with transaction boundaries and enriches change event envelopes with transaction metadata. Specify `true` if you want the connector to do this.
 For more information, see xref:postgresql-transaction-metadata[Transaction metadata].
 
+|[[postgresql-property-publish-via-partition-root]]<<postgresql-property-publish-via-partition-root, `+publish.via.partition.root+`>>
+|`false`
+|Determines whether streaming changes for partitioned tables are captured at the root partition level (`true`) or at the individual partition level (`false`).
```
Naros: I wonder for the layperson if it would be advantageous here to include some examples or be less technical. The goal here is to indicate the following:

* When set to `true`, partitioned table changes are emitted using the base table name.
* When set to `false`, partitioned table changes are emitted using the partition table name.

The people who configure the connector or manage it may not be database experts, and so trying to make this less technical, I think, is a win for all.

@roldanbob any suggestions from your side?

documentation/modules/ROOT/pages/connectors/postgresql.adoc
```diff
@@ -4020,6 +4020,12 @@ For more information, see xref:postgresql-toasted-values[toasted values].
...
+When `true`, publications will be created with `publish_via_partition_root = true` parameter for partitioned tables.
+When `false`, publications will be created without this parameter and it will not capture root level table changes.
```
roldanbob: UPDATE: _I've revised my comment to reflect the latest feedback._

If I understand this correctly, I think that the following revision might work:

```suggestion
Specifies how the connector captures and emits events for changes that it captures from partitioned tables.
This setting applies only if the xref:postgresql-publication-autocreate-mode[`publication.autocreate.mode`] property is set to `all_tables` or `filtered`, and {prodname} creates the publication for the captured tables.
 
Set one of the following options:
 
`true`::
The connector emits change events for all partitions to a topic with the name of the base table.
 +
When the connector creates a publication, it submits a `CREATE PUBLICATION` statement in which the `publish_via_partition_root` parameter is set to `true`.
As a result, the publication ignores the partition in which changes originate, and only records the name of the name of the source table. 

`false`::
The connector emits changes from each source partition to a topic that reflects the name of the partition.
 +
When the connector creates the publication, the `CREATE PUBLICATION` statement omits the `publish_via_partition_root` parameter so that the publication always uses the name of the source partition to publish change events.
```

roldanbob (changes requested):
I took a stab at breaking this down. Let me know if I missed the mark.

documentation/modules/ROOT/pages/connectors/postgresql.adoc
Naros: My only concern is the wording in line 4037 with `does not capture root level table changes`, and I am concerned how that may be perceived. Changes are captured regardless, it's just a matter of whether or not the change is represented as having come from the base table or the table partition.

roldanbob: Ah, ok, I think I get it. 
The original wording states that "it will not capture root level table changes," so I was reflecting that. 
But IIUC, what you're saying is that these root level changes _are_ captured, but the connector always emits changes to tables that are named for the source partition. 
In that case, the second sentence should be amended to say s.t. like the connector only considers the name of the source table and disregards the originating partition
And similarly, in the description for the `true` option, the second sentence likewise needs to be changed so that it states the opposite, i.e,  the connector only considers that name of the partition and disregards the name of the base table. Is that right?

rajdangwal: How does the following sound to you?

This configuration is applicable only when the connector is responsible for creating the publication in the source database. It determines how change events from partitioned tables are captured and emitted.

When set to `true`, the connector includes the `publish_via_partition_root = true` parameter in the publication creation statement, instructing the source database to publish change events using the root table name. As a result, changes from all partitions are emitted under the root table in the change stream.

When set to `false` (the default), the `publish_via_partition_root` parameter is omitted, and change events are published using the individual partition names, reflecting the exact partition where each change occurred.

roldanbob: @rajdangwal Another question to clarify exactly when the setting applies. Your latest comment states that it applies only if the connector is responsible for creating the publication. Does that imply that `publication.autocreate.mode` is set to either `all_tables` or `filtered`, and, further, that the connector actually creates the publication, rather than defaulting to using an existing publication?

Naros: Hi @roldanbob, you're correct if Debezium does not create the publication, this option has no real effect. This is a setting that gets applied only when Debezium manages the publication.

Kalla-Shubham: @roldanbob are we good on what @rajdangwal suggested?

roldanbob: @Kalla-Shubham, apologies for the delay; we had a long holiday weekend here. 
I've revised my suggested revision based on the latest from @Naros and @rajdangwal. After those changes are incorporated, this can be merged.

roldanbob (changes requested):
Please see my updated review comment, based on the latest feedback from @rajdangwal and @Naros. 

Contact me if you have further questions or if you believe that my revision does not accurately reflect the original intent.

documentation/modules/ROOT/pages/connectors/postgresql.adoc
Kalla-Shubham: thanks @roldanbob

Kalla-Shubham:
@roldanbob thanks But i am not able to merge, it show commit message format check failure even when I had done 
DBZ-9158: pr suggest

PR #6542, opened 2025-06-30
DBZ-9192 Fix PostgreSQL OpenLinage input dataset name

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-9192


jpechane:
@mfvitale Applied, thanks

PR #6543, opened 2025-06-30
DBZ-8610 Add SnapshotSkipped metric

nange10v:
https://issues.redhat.com/browse/DBZ-8610 

Add SnapshotSkipped metric as an additive metric which doesn't reset the other snapshot metrics. 

**Build CI errors**
This commit is related to a [PR](https://github.com/debezium/debezium-connector-spanner/pull/115) which implements the changes to the SnapshotProgressListener in `debezium-connector-spanner`. Due to a circular dependency (the spanner connector depends on debezium-core interfaces and the debezium CI depends on the spanner-connector CI test) the failing spanner connector build step should probably be manually ignored. Once this change is in main the corresponding PR in the spanner connector can be merged which will resolve the issue.

**Alternative approach to adding the snapshotSkipped metric**
The currently proposed implementation will preserve backward compatibility in external code depending on the snapshot state metrics. Alternatively SnapshotSkipped can be treated as another independent state of the snapshot. Note that this might cause issues in external code depending on the snapshot state always being in one of the current states -SnapshotCompleted, SnapshotAborted, SnapshotStarted, SnapshotPaused, SnapshotResumed.

Example alternative implementation:
```java
    public void snapshotSkipped() {
        this.snapshotCompleted.set(false);
        this.snapshotAborted.set(false);
        this.snapshotRunning.set(false);
        this.snapshotPaused.set(false);
        this.snapshotSkipped.set(true);
        this.stopTime.set(clock.currentTimeInMillis());
    }
```

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/metrics/SqlServerSnapshotPartitionMetrics.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/metrics/SqlServerSnapshotPartitionMetrics.java
@@ -56,6 +56,11 @@ class SqlServerSnapshotPartitionMetrics extends AbstractSqlServerPartitionMetric
     }
 
     @Override
+    public boolean getSnapshotSkipped() {
+        return snapshotMeter.getSnapshotSkipped();
+    }
+
+    @Override
     public long getSnapshotDurationInSeconds() {
         return snapshotMeter.getSnapshotDurationInSeconds();
     }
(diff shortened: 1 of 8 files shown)


Naros:
I've started the Oracle tests. I don't suspect they'll be failures related to this change, but I'd rather check to be sure.

Naros (approved):
Overall change LGTM, just pending CI results.

Naros:
The Oracle test failures are unrelated.

jpechane:
@nange10v Applied, thanks!

PR #6544, opened 2025-06-30
DBZ-8964 Debezium Engine Quarkus Extension: expose Debezium Notification Events

kmos:
## Context

Debezium connectors can emit notification events related to internal state changes, such as:

- Snapshot completion
- Other Events

Quarkus developers need a declarative mechanism to observe these events.

## Decision

Introduce a series of events that are observable using `Observes` like:

```java
    @ApplicationScoped
    static class SnapshotNotificationsHandler {

        public void observe(@Observes SnapshotEvent event) {
            // here
        }

    }
```

## References

#link: https://github.com/debezium/debezium-design-documents/pull/16
#closes: https://issues.redhat.com/browse/DBZ-8964

### DIFF
--- a/quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/notification/NotificationHandler.java
+++ b/quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/notification/NotificationHandler.java
@@ -6,8 +6,23 @@
 
 package io.quarkus.debezium.notification;
 
+/**
+ *
+ * {@link NotificationHandler} is used to manage Debezium Notification
+ *
+ */
 public interface NotificationHandler {
+
+    /**
+     * Define the handler availability {@param aggregateType}
+     * @param aggregateType
+     * @return
+     */
     boolean isAvailable(String aggregateType);
 
+    /**
+     * handle the Debezium Notification
+     * @param notification
+     */
     void handle(io.debezium.pipeline.notification.Notification notification);
 }
--- a/quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/notification/NotificationHandler.java
+++ b/quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/notification/NotificationHandler.java
@@ -1,0 +1,13 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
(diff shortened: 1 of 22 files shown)


quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/notification/SnapshotHandler.java
```diff
@@ -0,0 +1,120 @@
...
+        return kinds.contains(aggregateType);
+    }
+
+    @Override
+    public void handle(io.debezium.pipeline.notification.Notification notification) {
+        Kind.from(notification.getAggregateType())
+                .ifPresent(kind -> fire(kind, notification));
+    }
+
+    private void fire(Kind kind, Notification notification) {
```
Naros: I wonder if could introduce a Map of Kind to a Consumer, i.e.
```
private final Map<String, Consumer<NotificationContext>> handlers = new HashMap<>();
```
and then register these in the constructor or in a `@PostConstruct` method.  wdyt?

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/notification/NotificationHandler.java
```diff
@@ -0,0 +1,13 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package io.quarkus.debezium.notification;
+
+public interface NotificationHandler {
```
Naros, line 14: Lets ad some 📖

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/notification/SnapshotHandler.java
kmos: Totally agree...I'll try some investigations to create the map using Quarkus `<Instance>` that gives a list of beans...something like:

```java
public class SnapshotHandler implements NotificationHandler {
 private final Map<String, Consumer<NotificationContext>> handlers = new HashMap<>();

    @Inject
    public SnapshotHandler(Instance<Event<Notification>> notifications) {
        this.handlers = notifications.stream()
                .collect(Collectors.toMap(Notification::key, item -> item));
    }

// [...]
```

jpechane (approved):
@kmos Nice job, thanks! I'd just recommend you to think if we want to have all the notifications in the single package or maube there could be snapshot subpackage that would contain everything related to snapshot notifications and then future subpackages if needed.
Feel free to merge the PR yourself.

PR #6547, opened 2025-07-03
[docs] Mention Embeddings SMT Java requirements

vjuranek:


jpechane:
@vjuranek Applied, thanks

PR #6548, opened 2025-07-03
DBZ-9201 Fix OpenLineageIT tests

AlvarVG:
Test are failing because we are waiting for port 5432 (postgres) but we are actually working with port (55432) and that makes the test fail. Checking it dynamically solves the problem.


Naros:
Looks as tho the test failures could still be related @AlvarVG ?

jpechane:
There are additional OpenLineage tests fixes in https://github.com/debezium/debezium/pull/6496

Maybe it is the root cause for these failures? @mfvitale Could you please check it and merge the PR if this is the case?

AlvarVG:
> Looks as tho the test failures could still be related @AlvarVG ?

Definitely. I let them complete before merging. Working on it again.

AlvarVG:
> There are additional OpenLineage tests fixes in #6496
> 
> Maybe it is the root cause for these failures? @mfvitale Could you please check it and merge the PR if this is the case?

I dont think they are related. That PR seems to be a huge refactor on the OpenLineage integration itself, but i think we are still going to have the same issues once merged.

AlvarVG:
The failures are not related to this PR it self. I an merging it. cc\ @Naros @jpechane @mfvitale

PR #6551, opened 2025-07-03
DBZ-9199 Gracefully skip unexpected LogMiner entries for temporary tables

Naros:
https://issues.redhat.com/browse/DBZ-9199

@jpechane as I indicated on the Jira, I am unable to reproduce this locally, so there must be some unique setup in the reporter's system that allowed this to happen. LogMiner generally should never include these, but I've added a sanity check here to avoid them triggering a parser failure.


jpechane:
@Naros Applied, thanks

PR #6552, opened 2025-07-04
Dbz 3829 custom binlog position

kebab-mai-haddi:
## Description
This PR implements support for configuring custom binlog file and position to start CDC from, addressing DBZ-3829.

## Problem
Currently, Debezium MySQL connector cannot start streaming from a custom binlog file/position. When users try to set a custom offset:
- The snapshot task assumes a previous snapshot was done and skips schema creation
- This causes "schema isn't known" errors

## Solution
- Added configuration parameters `binlog.start.filename` and `binlog.start.position`
- Modified offset context to handle custom start positions
- Updated snapshot logic to properly initialize schema when custom offset is provided
- Added validation for binlog file existence and position validity

## Testing
- Added unit tests for configuration validation
- Added integration tests for custom binlog position startup
- Manually tested with MySQL 8.0

## Documentation
- [ ] Updated configuration documentation
- [ ] Added example in connector documentation

Fixes https://issues.redhat.com/browse/DBZ-3829
Fixes https://github.com/debezium/dbz/issues/1156


debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogConnectorConfig.java
```diff
@@ -898,6 +956,110 @@ private static int validateTimePrecisionMode(Configuration config, Field field,
...
+     */
+    private static int validateBinlogStartPosition(Configuration config, Field field, ValidationOutput problems) {
+        final String binlogFilename = config.getString(BINLOG_START_FILENAME.name());
+        final Long binlogPosition = config.getLong(BINLOG_START_POSITION.name());
+        final String gtidSet = config.getString(BINLOG_START_GTID_SET.name());
+
+        int errorCount = 0;
+
+        // Check mutual exclusivity with GTID set
+        if (gtidSet != null && !gtidSet.trim().isEmpty()) {
```
vjuranek: you can leverage here and in other places Debezium's `Strings.isNullOrEmpty(str)`

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogConnectorConfig.java
```diff
@@ -898,6 +956,110 @@ private static int validateTimePrecisionMode(Configuration config, Field field,
...
+            }
+            if (binlogPosition != null) {
+                problems.accept(BINLOG_START_POSITION, binlogPosition.toString(),
+                        "binlog.start.position cannot be used together with binlog.start.gtid.set");
+                errorCount++;
+            }
+        }
+
+        // If one of filename/position is specified, both must be specified
+        if ((binlogFilename != null && !binlogFilename.trim().isEmpty()) && binlogPosition == null) {
```
vjuranek: this and other `if` branch can be in the `else` branch if checking if the `gtid` is not `null` and thus checked only in that case

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogConnectorConfig.java
```diff
@@ -898,6 +956,110 @@ private static int validateTimePrecisionMode(Configuration config, Field field,
...
+            }
+        }
+
+        return errorCount;
+    }
+
+    /**
+     * Validate GTID set format and mutual exclusivity with binlog file/position.
+     */
+    private static int validateBinlogStartGtidSet(Configuration config, Field field, ValidationOutput problems) {
```
vjuranek: IMHO this can be merged with `validateBinlogStartPosition()` into one validation function.

debezium-connector-mysql/src/test/resources/test-mysql-config.properties
```diff
@@ -0,0 +1,27 @@
+# Test configuration for MySQL Debezium connector with custom binlog position
```
vjuranek: Is this file used anywhere?

vjuranek (commented):
Thanks for the PR @kebab-mai-haddi ! Overall LGTM, just a few comments regarding code structure.

Naros (changes requested):
I don't see anywhere in your original description of the issue, nor in the code, where you intend to handle what happens on subsequent restarts of the connector if those properties are left in the connector configuration. This was one of the main factors for why we chose not to implement this behavior this way.

You also said (in code) that providing the position must trigger a repopulation of the schema history topic, but that the capture of the schema will be based on the current schema for those tables, not the table's structure at the binlog position provided. This has the potential to create problems in a volatile system by emitting events if these structures differ, much like you would encounter when using `recovery` snapshot mode when it's not applicable.

Changing the position should not force a recapture of schema. It should leave the schema history topic untouched, just like if you were manually manipulating the offsets.

But we need to consider the former use case for rebalances, restarts, etc.

I would genuinely think a more robust approach would be using signals to send the position/filename/gtid set and then have the signal processing stop the connector, adjust the offsets, and restart the connector.

Naros:
Hi @kebab-mai-haddi do you have any input on my inquiry above?

kebab-mai-haddi:
@Naros  Thank you for the thorough review and valuable feedback! I've completely reimplemented this feature using the signal-based approach as you suggested.

kebab-mai-haddi:
update: trying to run tests. currently failing badly at being able to run integration tests. access denied. need to dive deeper into why i can connect with mysql in docker but tests fail, with the same creds.

Naros:
When you run the MySQL tests, its essential you use the right profiles.  I generally recommend setting the `mysql-ci` profile as enabled if you're running in the IDE, and then build the docker container and run it like so:
```
cd debezium-connector-mysql
mvn docker:build -Pmysql-ci
mvn docker:run -Pmysql-ci
```
When building and running the containers, make sure that you use the same profile that you intend to use in the IDE. if you want to run all tests via the CLI across all configurations, it's a similar process:

```
./mvnw clean install -pl debezium-connector-mysql -am -Passembly
```

The assembly profile will run all the various CI flavors, MySQL as a single server, the primary/replica, SSL and Percona all back to back with a single toggle. Can you give that a shot and let us know if that works @kebab-mai-haddi ?

Naros:
Hi @kebab-mai-haddi could you take a look at the test failures?

Naros:
> I've completely reimplemented this feature using the signal-based approach as you suggested.

@kebab-mai-haddi we should remove all the necessary changes that don't relate to the signal approach, such as the connector configuration options, etc correct?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/signal/MySqlSignalActionProvider.java
```diff
@@ -0,0 +1,44 @@
...
+import io.debezium.pipeline.signal.actions.SignalActionProvider;
+import io.debezium.pipeline.spi.Partition;
+import io.debezium.spi.schema.DataCollectionId;
+
+/**
+ * Provider for MySQL-specific signal actions.
+ *
+ * @author Debezium Authors
+ */
+public class MySqlSignalActionProvider implements SignalActionProvider {
```
Naros, line 25: Could we introduce a specialization  for MariaDB. Since both of them are nearly identical, except for their GTID format, this should be relatively straightforward.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/signal/SetBinlogPositionSignal.java
```diff
@@ -0,0 +1,162 @@
...
+                LOGGER.info("Setting binlog position to GTID set: {}", gtidSet);
+                offsetContext.setCompletedGtidSet(gtidSet);
+            }
+            else {
+                LOGGER.info("Setting binlog position to file: {}, position: {}", binlogFilename, binlogPosition);
+                offsetContext.setBinlogStartPoint(binlogFilename, binlogPosition);
+            }
+
+            // Force a new offset commit to persist the change
+            eventDispatcher.alwaysDispatchHeartbeatEvent(signalPayload.partition, offsetContext);
```
Naros, line 93: What about if the user has not configured `heartbeat.interval.ms`? In that case, this would never be persisted until the following change that is read and emitted.

I'm okay with defining that the heartbeats must be turned on for this to work initially, but it should be clearly documented and highlighted.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlOffsetContext.java
```diff
@@ -34,7 +34,26 @@ public static MySqlOffsetContext initial(MySqlConnectorConfig config) {
...
+        }
+        else if (customBinlogFilename != null && customBinlogPosition != null) {
+            // Use custom binlog position
+            offset.setBinlogStartPoint(customBinlogFilename, customBinlogPosition);
+        }
+        else {
+            // Start from the beginning of the binlog
+            offset.setBinlogStartPoint("", 0L);
+        }
+
```
Naros: If we are only going to use signals for this, which is what I would recommend, this can be removed.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorConfig.java
```diff
@@ -409,6 +409,28 @@ public boolean isSslModeEnabled() {
...
+        return position != null ? Long.valueOf(position) : null;
+    }
+
+    /**
+     * @return the custom GTID set to start reading from
+     */
+    public String getBinlogStartGtidSet() {
+        return getConfig().getString(io.debezium.connector.binlog.BinlogConnectorConfig.BINLOG_START_GTID_SET);
+    }
+
```
Naros: If we are only going to use signals for this, which is what I would recommend, this can be removed.

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/util/UniqueDatabase.java
```diff
@@ -233,7 +233,10 @@ public Configuration.Builder defaultJdbcConfigBuilder() {
                 .with(BinlogConnectorConfig.PORT, System.getProperty("database.port", "3306"))
                 .with(BinlogConnectorConfig.USER, "snapper")
                 .with(BinlogConnectorConfig.PASSWORD, "snapperpass")
-                .with("driver.allowPublicKeyRetrieval", "true");
+                .with("driver.allowPublicKeyRetrieval", "true")
+                // Read driver.connectionTimeZone from system property to support local development in different timezones
+                // Defaults to UTC to avoid ambiguous timezone errors (e.g., EDT) when running tests locally
+                .with("driver.connectionTimeZone", System.getProperty("driver.connectionTimeZone", "UTC"));
```
Naros: This shouldn't be needed.

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/util/BinlogTestConnection.java
```diff
@@ -162,7 +163,8 @@ protected static JdbcConfiguration.Builder getReplicaJdbcConfig(String databaseN
                         .merge(Configuration.fromSystemProperties(DRIVER_CONFIG_PREFIX)
                                 .merge(Configuration.fromSystemProperties(DATABASE_CONFIG_PREFIX))))
                 .withDatabase(databaseName)
-                .with("characterEncoding", "utf8");
+                .with("characterEncoding", "utf8")
+                .with("connectionTimeZone", "UTC");
```
Naros: Ths shouldn't be needed.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogSnapshotChangeEventSource.java
```diff
@@ -294,6 +294,32 @@ protected abstract void setOffsetContextBinlogPositionAndGtidDetailsForSnapshot(
...
+
+        // Check for custom binlog file/position
+        if (customBinlogFilename != null && customBinlogPosition != null) {
+            // If both custom filename and position are specified, this indicates a custom start position
+            // We should perform schema snapshot but skip data snapshot
+            return true;
+        }
+
+        return false;
+    }
```
Naros: If we are only going to use signals for this, which is what I would recommend, this can be removed.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogConnectorConfig.java
```diff
@@ -450,6 +450,40 @@ default boolean useConsistentSnapshotTransaction() {
...
+            .withDisplayName("Initial GTID set")
+            .withType(ConfigDef.Type.STRING)
+            .withWidth(ConfigDef.Width.LONG)
+            .withImportance(ConfigDef.Importance.MEDIUM)
+            .withValidation(BinlogConnectorConfig::validateBinlogStartGtidSet)
+            .withGroup(Field.createGroupEntry(Field.Group.CONNECTOR_ADVANCED, 6))
+            .withDescription("The GTID set to start reading from when using GTID-enabled MySQL. "
+                    + "When this property is set, the connector will start reading from this GTID set instead of "
+                    + "binlog file/position. Cannot be used together with binlog.start.filename and binlog.start.position. "
+                    + "Format example: 'server-uuid:1-100,other-server-uuid:1-50'");
```
Naros: If we are only going to use signals for this, which is what I would recommend, this can be removed along with the other changes in the config class.

debezium-connector-mysql/src/test/docker/init/001-setup.sql
```diff
@@ -1,3 +1,6 @@
+-- Set timezone to avoid JDBC connection issues
```
Naros: Changes here are superfluous, no?

debezium-connector-mysql/src/test/docker/server/my.cnf
```diff
@@ -39,6 +39,8 @@ symbolic-links=0
 # Server ID is required, but this will vary on production systems
 server-id                   = 112233
 log_bin                     = mysql-bin
+# Set default timezone to avoid JDBC connection issues
+default_time_zone           = '+00:00'
```
Naros: Changes here are superfluous, we should never assume the database's time zone and we should instead make sure that the code works regardless of database or host time zones.

jpechane:
I really like the idea and the direction in which this is going. But for additional safety I'd recommend to stop the connector automatically instead of realying on the restart only.
My idea is to have one more field in the signal that would be two state - `stop` and `continue` with stop as the default.
When `stop` is set the connector is automatically stopped after the offsets are modified.
Later on we can add third state `restart` that would automatically restart the connector after the changes are applied.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/signal/MySqlSignalActionProvider.java
kebab-mai-haddi: I'd like to address the MariaDB specialization in a follow-up PR since the MySQL implementation is complete and tested. Would that be acceptable?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/signal/SetBinlogPositionSignal.java
kebab-mai-haddi: Done, is that okay?

kebab-mai-haddi:
## Integration Test Status Update

### Current Situation
The integration test  is experiencing an issue where INSERT events during streaming are not being captured as change events.

### What Works
- Signal implementation (unit tests pass)
- Action field implementation (stop/continue/restart)
- Connector initialization and startup
- Snapshot completion
- Streaming phase startup
- Database connection and binlog reader connection

### The Problem
When INSERT statements are executed during the streaming phase:
```java
connection.execute("INSERT INTO test_table VALUES (1, 'value1')");
connection.execute("INSERT INTO test_table VALUES (2, 'value2')");
connection.commit();
```

The binlog reader successfully reads 12 events, but when consuming records:
- **Expected**: 2 INSERT change events on topic `binlog_signal_test.signal_test_xxx.test_table`
- **Actual**: Only heartbeat messages on topic `__debezium-heartbeat.binlog_signal_test`

The INSERT data changes are not appearing in the consumed records despite:
- Explicit transaction commits
- waitForStreamingRunning() to ensure streaming is active
- waitForAvailableRecords() before consumption
- Binlog reader confirming it read the events

### Investigation Conducted
This appears to be a test pattern issue rather than a signal implementation issue. Similar tests like `SignalsIT` and `IncrementalSnapshotIT` extend binlog parent classes that have different consumption patterns. Currently investigating the correct approach for testing streaming INSERT consumption in this test context.

### Next Steps
1. Compare consumption patterns with working streaming tests
2. Investigate if table filtering or event dispatching is preventing INSERT events from reaching the consumer
3. Review whether additional configuration is needed for NO_DATA snapshot mode with streaming inserts

kebab-mai-haddi:
@Naros @jpechane Test update:

**All 4 MySQL CI profiles pass** for MySqlBinlogPositionSignalIT:
- mysql-ci: 2/2
- mysql-ci-gtids: 2/2  
- mysql-ci-percona: 2/2
- mysql-ci-ssl: 2/2

Assembly profile ran 235 test classes with 0 failures before hitting OOM on our test server (3.8GB RAM).

Reviewer feedback addressed:
- No connector config changes in PR (signal-only approach)
- stop/continue/restart action field implemented

Naros:
Hi @kebab-mai-haddi, thanks for the update. I have a few functional questions.

In `SetBinlogPositionSignal`, we have 3 different types of actions (`stop`, `restart`, and `continue`). I agree with the stop and restart, but I am heavily torn on this idea of `continue`.

In the documentation, you mention that `continue` is designed to continue processing where the connector is presently reading, but then will resume from that position on the next restart. However, isn't the position overwritten when you process the next change event from the binlog?

Furthermore, the test coverage for `continue` makes an assumption that the connector is stopped after the signal. This does not align with the wording in the documentation, which gives the impression the signal is handled, the connector can continue to process events, and only after some indefinite future point you restart and the connector resumes from the position specified in the signal. But if I understand the code, this breaks due to my above observation.

Unless I am missing something in the code, I really believe the only reasonable actions here are either `stop` or `restart`, which avoids any sort of state mismanagement with offsets. wdyt?

kebab-mai-haddi:
Hi @Naros,

Apologies for the confusion with the `continue` action. I initially implemented it as suggested, but realized I missed the point - `continue` would be misleading because even after updating the offset in memory, the connector keeps reading from its current binlog position. The new offset only takes effect after a restart, so `continue` would give users a false impression that something changed when it didn't.

I've removed `continue` and kept only `stop` (default) with `restart` reserved for future use. This way users get predictable behavior: signal updates the offset, connector stops, restart applies the new position.

vjuranek (changes requested):
hi @kebab-mai-haddi , it seems that mysql module doesn't compile now. Could you please take a look and fix it?

kebab-mai-haddi:
hey @vjuranek , my branch wasnt rebased on top of master.  can you re-trigger CI?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/signal/SetBinlogPositionSignal.java
```diff
@@ -0,0 +1,206 @@
...
+            // Stop the connector as requested by the action field (default: stop)
+            // This ensures the new offset takes effect on the next restart
+            // Schedule the stop in a separate thread to avoid lifecycle issues when
+            // calling stop from within the signal handler context
+            if (action == Action.STOP) {
+                LOGGER.info("Stopping connector to apply new binlog position. Restart the connector for changes to take effect.");
+                Thread stopThread = new Thread(() -> {
+                    try {
+                        // Small delay to allow signal processing to complete
+                        Thread.sleep(100);
```
vjuranek: I wonder how we can be sure that by this time heartbeat is already processed.

Naros:

Hi @kebab-mai-haddi, please sign off all commits with:

```
git commit -s
```

Naros:
Hi @kebab-mai-haddi if you rebase on `main`, you can have it sign-off each commit to avoid the bot spam
```
git rebase --signoff main
```
Hope that helps. :rocket:

Naros:
Hi @kebab-mai-haddi, it appears the email address in your last commit differs from the ones in previous commits, and it does not match the email address associated with the commit's author email. Can you please adjust it for db9e15c?

kebab-mai-haddi:
Hi @Naros , re-requested your review. And this time, CI checks passed :) 

Thanks again and apologies for that sign-off issue I had.

jpechane:
@kebab-mai-haddi Thanks! Have you had a chance to look at https://github.com/debezium/debezium/pull/6552#issuecomment-3471698125 ? Do you think it is feasible? Is it something that can be implemented in this or a follow-up PR?

kebab-mai-haddi:
@jpechane Yes, it's feasible. I'll implement the `stop` behavior (connector automatically stops after offset update) in a follow-up PR. The `continue` action was removed since it was confusing (the new offset only takes effect after restart anyway), and `restart` will be reserved for future use.

kebab-mai-haddi:
@Naros Thanks for the approval\! I've created a separate PR for the documentation in `mysql.adoc`. Here's the follow-up PR for reference: https://github.com/debezium/debezium/pull/7061. Can this PR now be merged? Thanks\!

Naros:
Hi @kebab-mai-haddi I think we can merge this, but one question. Do you think you could submit the follow-up PR before the end of the 3.5 development cycle in mid/late March?

kebab-mai-haddi:
> Hi @kebab-mai-haddi I think we can merge this, but one question. Do you think you could submit the follow-up PR before the end of the 3.5 development cycle in mid/late March?

Yes @Naros , 100%. I assure you I would submit the two follow up PRs, after the merge of the [documentation PR](https://github.com/debezium/debezium/pull/7061) that you have. And will get started right away.

Naros:
None from my side @jpechane, so once the `.md` content is moved to the docs, we're good :tada:

kebab-mai-haddi:
@Naros @jpechane I've removed the inline docs file as requested -- the documentation has been moved to PR #7061. All review feedback has been addressed. Could you please merge this when you get a chance? Thanks!

kebab-mai-haddi:
## CI Failure Analysis

Two CI jobs are failing on this PR. Both are **pre-existing failures on the `main` branch** and are unrelated to the changes in this PR.

### 1. MySQL / MySQL 9.1 - mysql-ci-gtids

The Docker container fails to start during initialization when running `001-setup.sql` (line 21):

```
ERROR 1064 (42000) at line 21: You have an error in your SQL syntax; check the manual
that corresponds to your MySQL server version for the right syntax to use near
'MASTER -- MySQL 8.x' at line 1
```

MySQL 9.1 removed the `RESET MASTER` command (replaced by `RESET BINARY LOGS AND GTIDS`). The init script `001-setup.sql` has both commands on consecutive lines, and MySQL 9.1 rejects the old syntax before reaching the new one. No tests execute -- the container itself cannot start.

**This PR does not modify `001-setup.sql`.**

### 2. Storage Module

All 7 test errors are in `JedisClusterClientIT` and `RedisOffsetBackingStoreIT`:

```
Could not find a valid Docker environment
```

The CI runner cannot find a Docker environment to spin up Redis containers. This is a CI infrastructure issue.

**This PR does not modify any files in the storage module.**

### Evidence from `main` branch

The same two jobs fail on every recent push to `main`:

| Run | `MySQL 9.1 - mysql-ci-gtids` | `Storage Module` |
|-----|-----|-----|
| [22004330142](https://github.com/debezium/debezium/actions/runs/22004330142) | FAIL | FAIL |
| [21989395361](https://github.com/debezium/debezium/actions/runs/21989395361) | FAIL | FAIL |
| [21972427795](https://github.com/debezium/debezium/actions/runs/21972427795) | FAIL | FAIL |

### All MySQL-relevant jobs pass

Every CI job relevant to the MySQL connector changes in this PR passes:

- MySQL 8.0 - mysql-ci
- MySQL 8.4 - mysql-ci
- MySQL 9.1 - mysql-ci
- MySQL 8.0 - mysql-ci-percona
- MySQL 8.4 - mysql-ci-percona
- MySQL 9.1 - mysql-ci-percona
- MySQL 9.1 - mysql-ci-ssl
- MySQL - Apicurio (9.1)
- Checkstyle and Formatting
- Commit message

jpechane:
@kebab-mai-haddi Applied, thanks!

PR #6557, opened 2025-07-09
DBZ-8319 Create readonly connection to oracle database

lgazire:


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/DualOracleConnectionFactory.java
```diff
@@ -0,0 +1,50 @@
...
+import io.debezium.jdbc.JdbcConnection;
+import io.debezium.jdbc.MainConnectionProvidingConnectionFactory;
+
+/**
+ * Factory to manage between read-only and regular connection for the purpose of using only what is necessary for
+ * start the connector while the read-only will be used to log mining.
+ * @param <T>
+ * @author Lucas Gazire
+ */
+public class DualOracleConnectionFactory<T extends JdbcConnection> implements MainConnectionProvidingConnectionFactory<T> {
```
Naros:
```suggestion
public class OracleConnectionFactory<T extends JdbcConnection> implements MainConnectionProvidingConnectionFactory<T> {
```
I think the `Dual` prefix is a bit superfluous, we can just call it `OracleConnectionFactory`.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/LogMinerSessionContext.java
```diff
@@ -108,7 +109,7 @@ public void removeAllLogFilesFromSession() throws SQLException {
      * @throws SQLException if a database exception occurred starting the mining session
      * @throws RetriableLogMinerException if the operation should be retried
      */
-    public void startSession(Scn startScn, Scn endScn, boolean committedDataOnly) throws SQLException {
+    public void startSession(Scn startScn, Scn endScn, boolean committedDataOnly, String dictionaryFilePath) throws SQLException {
```
Naros: Could we pass the `dictionaryFilePath` as a constructor argument rather than as a method argument? I realize that I did not do this for `committedDataOnly`, but in hindsight, I should have since these are static values that do not change over the connector's lifecycle.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/DualOracleConnectionFactory.java
```diff
@@ -0,0 +1,50 @@
...
+
+    public DualOracleConnectionFactory(ConnectionFactory<T> mainConnectionFactory, ConnectionFactory<T> readOnlyConnectionFactory, OracleConnectorConfig config) {
+        this.mainConnectionFactory = mainConnectionFactory;
+        this.readOnlyConnectionFactory = readOnlyConnectionFactory;
+        this.mainConnection = mainConnectionFactory.newConnection();
+        this.readOnlyConnection = this.readOnlyConnectionFactory.newConnection();
+        this.isLogMiningReadonly = config.isLogMiningReadOnly();
+    }
+
+    public T getConnection() {
```
Naros: Unless my IDE indexes are broken, I don't see this method nor the `newReadonlyConnection` being used from what I could see.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/ReadOnlyOracleConnection.java
```diff
@@ -0,0 +1,30 @@
...
+import java.sql.Connection;
+import java.sql.SQLException;
+
+import io.debezium.jdbc.JdbcConfiguration;
+
+/**
+ * Read-only connection class, extends the oracle connection, for read-only databases.
+ * @author Lucas Gazire
+ */
+public class ReadOnlyOracleConnection extends OracleConnection {
```
Naros: Could we use `ConnectionFactoryDecorator` or a custom `ConnectionFactory` instead to apply the read-only semantic to the connection?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/DualOracleConnectionFactory.java
lgazire: This in fact is not being used, i was thinking in using it in the bean registry but i don’t know how it works so i let it with the standard connection, i will remove it from the class.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/ReadOnlyOracleConnection.java
lgazire: Not sure how to do this, i set the semantic because it seemed the right approach  to avoid unnecessary commands (insert, update …) Can you give more insight?

lgazire:
@Naros thanks for the feedback!

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/ReadOnlyOracleConnection.java
Naros: @lgazire  so what I might recommend is inside the `OracleConnectionFactory`, introduce a new private class similar to what we do inside the `JdbcConnection` with `ConnectionFactoryDecorator`, something like `ReadOnlyConnectionDecorator`.

Then wrap the `readOnlyConnectionFactory` constructor argument in `OracleConnectionFactory` by this decorator, mirroring the decorator wrapping the factory in the `JdbcConnection` constructor. The decorator would be responsible for setting the connection to read-only.

That should eliminate the need for the `ReadOnlyOracleConnection` implementation I think. 

I haven't tested this theory, but I would prefer to wrap this inside the factories somehow rather than the standalone class.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
```diff
@@ -85,7 +87,7 @@ public ChangeEventSourceCoordinator<OraclePartition, OracleOffsetContext> start(
 
         // The bean registry JDBC connection should always be pinned to the PDB
         // when the connector is configured to use a pluggable database
-        beanRegistryJdbcConnection = connectionFactory.newConnection();
+        beanRegistryJdbcConnection = dualConnectionFactory.newConnection();
```
Naros: Could we use the original `connectionFactory` naming, do you think?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/AbstractLogMinerStreamingAdapter.java
```diff
@@ -87,8 +87,7 @@ public OracleOffsetContext determineSnapshotOffset(RelationalSnapshotContext<Ora
         final Optional<Scn> currentScn;
         if (isPendingTransactionSkip(connectorConfig)) {
             currentScn = getCurrentScn(latestTableDdlScn, connection);
-        }
-        else {
+        } else {
```
Naros: Some minor formatting changes have been made that will likely break code style checks. Can you please do me a favor and just run `./mvnw clean install -DskipITs -DskipTests` locally and let that reformat the code.  Thanks.

Naros (changes requested):
Just a few more comments with the latest round of commits.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/ReadOnlyOracleConnection.java
lgazire: ok, i understand your explanation but there’s two ConnectionFactory in the solution, what should be used to extend my decorator? Also in the OracleConnectionFactory i’m using T extends JdbcConnection the connect override is returning a java.sql Connection should i change this as well to use java’s Connection?

Naros:
Hi @lgazire I rebased the PR and I also added some small adjustments. Could you take a look and see if you agree with the changes?  Thanks!

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/ReadOnlyOracleConnection.java
Naros: I added some changes to reflect what I was after, let me know if you agree with them.

lgazire:
Hi @Naros, it's way better than what I was doing, awesome.

vjuranek (approved):

vjuranek:
Test failures are unrelated.

vjuranek:
Merged, thanks @lgazire and @Naros !

PR #6558, opened 2025-07-10
DBZ-9212 Validation for Log Position for SqlServer can fail

kmos:
## Context

in some cases the validation for the LSN in SQLserver can fail due to nullability check.

## Side notes
It reduce flaky tests

## reference
#closes https://issues.redhat.com/browse/DBZ-9212


debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
```diff
@@ -765,7 +765,7 @@ public boolean validateLogPosition(Partition partition, OffsetContext offset, Co
 
             LOGGER.info("Oldest SCN in logs is '{}'", oldestScn);
             LOGGER.info("Stored LSN is '{}'", storedLsn);
-            return storedLsn == null || Lsn.valueOf(oldestScn).compareTo(storedLsn) < 0;
+            return storedLsn.equals(Lsn.NULL) || Lsn.valueOf(oldestScn).compareTo(storedLsn) < 0;
```
vjuranek: to avoid possible NPE, I'd use 
```suggestion
            return Lsn.NULL.equals(storedLsn) || Lsn.valueOf(oldestScn).compareTo(storedLsn) < 0;
```
but shouldn't we still consider the case when `storedLsn` is `null` in evaluating validity of the log position?

kmos: Yeah, I totally agree! I was thinking the same but I wasn't so sure to be so _defensive_ (and I was a bit tired of the failing tests :eyes: ). If you think it's ok, I can evaluate also it

mfvitale: Given a quick look seems that `storedLsn` should never be `null` but maybe it's better to also check for nullability

Naros:
@kmos, as I added this on the Jira, is this in any way related to DBZ-9217? I just want to be sure.

kmos:
> @kmos, as I added this on the Jira, is this in any way related to DBZ-9217? I just want to be sure.


Well, yes, I did find a failure there, but the root cause isn't related. I believe it could fail in many other cases as well.

vjuranek (approved):

kmos:
> @kmos, as I added this on the Jira, is this in any way related to DBZ-9217? I just want to be sure.

I have to take a look the error is different but can be related.

kmos:
> > @kmos, as I added this on the Jira, is this in any way related to DBZ-9217? I just want to be sure.
> 
> I have to take a look the error is different but can be related.

@Naros confirmed

vjuranek:
Merge, thanks @kmos !

akanimesh7:
@vjuranek @kmos @Naros 
The lsn from offsets should never be null ideally right ? As identified in the fix done in https://github.com/debezium/debezium/pull/6537 for some of the tests, i think there is some flakiness which we might miss catching due to the introduction of this change ?

jpechane:
@kmos @akanimesh7 
Well, this is good point. I looked in the code and it seems to me that this should not happen on the happy path. Also check is not `Lsn.NULL.equals`, but should be `storedLsn.isAavailable`.

PR #6560, opened 2025-07-14
DBZ-8582 Support MariaDB 11.7+ vector data type in JDBCSinkConnector

pranavt84:
https://issues.redhat.com/browse/DBZ-8582

- [ ]  Support MariaDB 11.7+ vector data type in JDBCSinkConnector

### DIFF
--- a/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogBlockingSnapshotIT.java
+++ b/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogBlockingSnapshotIT.java
@@ -12,7 +12,6 @@ import java.sql.SQLException;
 import java.util.List;
 
 import org.apache.kafka.connect.source.SourceConnector;
-import org.jetbrains.annotations.NotNull;
 import org.junit.After;
 import org.junit.Before;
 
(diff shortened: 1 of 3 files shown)


debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/maria/DoubleVectorType.java
```diff
@@ -0,0 +1,34 @@
...
+ * @author Pranav Tiwari
+ */
+public class DoubleVectorType extends AbstractDoubleVectorType {
+
+    public static DoubleVectorType INSTANCE = new DoubleVectorType();
+
+    @Override
+    public String getTypeName(Schema schema, boolean isKey) {
+        final Optional<String> size = getSourceColumnSize(schema);
+        return size.map(s -> String.format("vector(%s)", s)).orElse("vector(16383)");
```
Naros, line 30: I would add a comment in this and in the other class about the reasoning for using 16383 and 2048.

Naros (commented):
Just left one comment inline.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/maria/DoubleVectorType.java
pranavt84: Added a comment to explain the reasoning in both files.

Naros:
@pranavt84 I assume there is the other side of this to add support in the MariaDB source connector, which is why the PR is still in draft state, right?

pranavt84:
> @pranavt84 I assume there is the other side of this to add support in the MariaDB source connector, which is why the PR is still in draft state, right?

@Naros 
Yes, that's correct.
However, this part can be merged independently since there's no direct dependency on the source connector changes. If you feel so, I’ll move it from draft to ready for review.

I've started working on the source connector side, but it may take some time as I'm still going through the codebase to understand what changes are needed and how best to implement them.

Naros (approved):
Just waiting on CI, but overall LGTM @pranavt84.

Could you also consider opening a PR in the upstream https://github.com/antlr/grammars-v4/tree/master/sql/mariadb with the changes you've made to the `.g4` and `.sql` files.

Naros:
Thanks @pranavt84, applied!

PR #6561, opened 2025-07-15
DBZ-9172 Use CONSTRAINT keyword in DROP statement

ani-sha:
https://issues.redhat.com/browse/DBZ-9172


Naros:
Applied, thanks @ani-sha !

PR #6563, opened 2025-07-16
DBZ-8960 Debezium Engine Quarkus Extension: Support Heartbeat Event Listener

kmos:
## Context

Heartbeat events in Debezium are synthetic change events emitted at fixed intervals when no other data changes are detected. They serve multiple operational purposes:
 * Confirm that the connector is still connected to the source database

 * Detect and recover from liveness failures

## Decision

Quarkus users should be able to listen specifically for heartbeat events in a **Quarkus** way like:

```java
import io.debezium.engine.ChangeEvent;


@ApplicationScoped  
class HeartbeatListener {

    public void heartbeat(@Observes DebeziumHeartbeat event) {  
        /// some logic to apply 
    }  
}
```

## References

#link https://github.com/debezium/debezium-design-documents/pull/16
#closes https://issues.redhat.com/browse/DBZ-8960

### DIFF
--- a/quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/heartbeat/QuarkusHeartbeatEmitter.java
+++ b/quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/heartbeat/QuarkusHeartbeatEmitter.java
@@ -21,17 +21,17 @@ import io.debezium.runtime.events.DebeziumHeartbeat;
 public class QuarkusHeartbeatEmitter implements Heartbeat {
 
     private final Debezium debezium;
-    private final Event<DebeziumHeartbeat> heartBeat;
+    private final Event<DebeziumHeartbeat> heartbeat;
 
     @Inject
-    public QuarkusHeartbeatEmitter(Debezium debezium, Event<DebeziumHeartbeat> heartBeat) {
+    public QuarkusHeartbeatEmitter(Debezium debezium, Event<DebeziumHeartbeat> heartbeat) {
         this.debezium = debezium;
-        this.heartBeat = heartBeat;
+        this.heartbeat = heartbeat;
     }
 
     @Override
     public void emit(Map<String, ?> partition, OffsetContext offset) {
-        heartBeat.fire(new DebeziumHeartbeat(
+        heartbeat.fire(new DebeziumHeartbeat(
                 debezium.connector(),
                 debezium.status(),
                 partition,


kmos:
side note: I think that the `heartbeat` is a feature really important in the context of microservices/SOA. Maybe in the case of the Quarkus Extension should be enabled by default? WDYT @mfvitale @Naros ?

(adding a default `interval`)

Naros:
I understand your reasonings, but I'm also hesitate to consider having something have different defaults in one runtime versus another. That will only create confusion for users as well as for us to both maintain and document. We should probably strive to keep these as consistent across all runtimes as possible @kmos 

For now at the very least :)

debezium-core/src/main/java/io/debezium/heartbeat/HeartbeatFactory.java
```diff
@@ -50,6 +58,11 @@ public HeartbeatFactory(CommonConnectorConfig connectorConfig, TopicNamingStrate
         this.schemaNameAdjuster = schemaNameAdjuster;
         this.connectionProvider = connectionProvider;
         this.errorHandler = errorHandler;
+        this.heartbeatFactories = new ArrayList<>(externalHeartbeatFactory
+                .stream()
+                .map(ServiceLoader.Provider::get)
+                .toList());
+        this.heartbeatFactories.add(new DatabaseHeartbeatFactory());
```
Naros: Should we prioritize our implementation (the connector level one) before external ones?

Naros (commented):
LGTM, I just have one comment inline.

debezium-core/src/main/java/io/debezium/heartbeat/HeartbeatFactory.java
kmos: yeah, I agree

kmos:
> I understand your reasonings, but I'm also hesitate to consider having something have different defaults in one runtime versus another. That will only create confusion for users as well as for us to both maintain and document. We should probably strive to keep these as consistent across all runtimes as possible @kmos
> 
> For now at the very least :)

I think you are right, let's keep consistent :rocket:

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/heartbeat/QuarkusHeartbeatEmitter.java
```diff
@@ -0,0 +1,45 @@
...
+import io.debezium.heartbeat.Heartbeat;
+import io.debezium.pipeline.spi.OffsetContext;
+import io.debezium.runtime.Debezium;
+import io.debezium.runtime.events.DebeziumHeartbeat;
+
+@ApplicationScoped
+public class QuarkusHeartbeatEmitter implements Heartbeat {
+
+    private final Debezium debezium;
+    private final Event<DebeziumHeartbeat> heartBeat;
```
jpechane:
```suggestion
    private final Event<DebeziumHeartbeat> heartbeat;
```

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/heartbeat/QuarkusHeartbeatEmitter.java
```diff
@@ -0,0 +1,45 @@
...
+
+    @Inject
+    public QuarkusHeartbeatEmitter(Debezium debezium, Event<DebeziumHeartbeat> heartBeat) {
```
jpechane:
```suggestion
    public QuarkusHeartbeatEmitter(Debezium debezium, Event<DebeziumHeartbeat> heartbeat) {
```

jpechane (approved):
@kmos Nice work, thanks! I left two super tiny comments. When addressed feel free to merge yourself.

jpechane:
@kmos Applied, thanks!

PR #6566, opened 2025-07-16
DBZ-9229 Break processing when log switch occurs

Naros:
https://issues.redhat.com/browse/DBZ-9229

- [ ] Restrict the query based on read SCN within online redo log SCN boundaries
- [ ] Add test case

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/AbstractLogMinerStreamingChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/AbstractLogMinerStreamingChangeEventSource.java
@@ -78,6 +78,8 @@ import io.debezium.util.Loggings;
 import io.debezium.util.Metronome;
 import io.debezium.util.Stopwatch;
 import io.debezium.util.Strings;
+import io.debezium.util.Threads;
+import io.debezium.util.Threads.Timer;
 
 /**
  * An abstract implementation of the {@link StreamingChangeEventSource} for Oracle LogMiner, that is the basis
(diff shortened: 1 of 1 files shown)


nathan-smit-1:
hey @Naros ,  is there a discussion somewhere on Zulip around the issue that this resolves?  Just want to get a bit more context.  I will check out the change though and see if we can test it somewhere

Naros:
Hi @nathan-smit-1, unfortunately, there is not. This PR was born out of an internal case where an Oracle system was over-saturated, and before DBZ-9241. 

The use case was that when the connector fell behind, the mining range would no longer be based on the LogMiner batch size max settings, but would rather jump the upper boundary to the current write position of the database, `current_scn`. In this case, that would cause a wide window of changes to be read in one pass. 

It was observed that the metrics `OffsetScn` value would not change for extended periods when this happened, and raised questions about whether a commit was being missed. After reviewing the LogMiner extracts, we observed that in some cases, the transaction commit occurred within seconds of the start of the transaction, and in other cases, several minutes later.

The only consistent thing I found was that it happened only during a large pass across many logs. So it made me question whether LogMiner may mistakenly not report the online redo log has rolled when streaming changes, and rather than throw an exception, it omits changes.

So the question stands whether we should be checking periodically for log switches while processing batches.

gong:
> Hi @nathan-smit-1, unfortunately, there is not. This PR was born out of an internal case where an Oracle system was over-saturated, and before DBZ-9241.
> 
> The use case was that when the connector fell behind, the mining range would no longer be based on the LogMiner batch size max settings, but would rather jump the upper boundary to the current write position of the database, `current_scn`. In this case, that would cause a wide window of changes to be read in one pass.
> 
> It was observed that the metrics `OffsetScn` value would not change for extended periods when this happened, and raised questions about whether a commit was being missed. After reviewing the LogMiner extracts, we observed that in some cases, the transaction commit occurred within seconds of the start of the transaction, and in other cases, several minutes later.
> 
> The only consistent thing I found was that it happened only during a large pass across many logs. So it made me question whether LogMiner may mistakenly not report the online redo log has rolled when streaming changes, and rather than throw an exception, it omits changes.
> 
> So the question stands whether we should be checking periodically for log switches while processing batches.

@Naros Hi, I met data miss when endscn is `current_scn` and endscn-startscn is 5m.

Naros:
Hi @gong is that using this PR or are you using a specific version of Debezium?

gong:
> Hi @gong is that using this PR or are you using a specific version of Debezium?

@Naros I use 1.9.7 version. I search DBZ-9241 and found this PR. I guess miss data whether scn gap is too large cause problem.

Naros:
Hi @gong, the change you're needing is DBZ-8747, which requires Debezium 3.2.4.Final, 3.3.1.Final, or 3.4+.  I would suggest updating from 1.9.7.Final if possible, there have been significant improvements since the 1.x days.

Naros:

```
git commit -s
```

PR #6568, opened 2025-07-17
DBZ-9166 Toggle for legacy numeric decimal handling behavior

Naros:
https://issues.redhat.com/browse/DBZ-9166

In DBZ-7882, we introduced a breaking behavior that enforces `decimal.handling.mode` when set to `string` or `double` to be applied to all numeric columns, similar to other connectors. However, there have been repeated reports that this change was too disruptive.

I first considered introducing this as a `CustomConverter`, but upon reviewing the required code changes, I found that we would also need a way to determine the database's nationalized character set, which the `OracleValueConverters` class already provides. And as I looked deeper, the easiest and cleanest solution would be to introduce a legacy toggle in the connector rather than trying to make this a standalone converter component.

So this introduces `legacy.decimal.handling.strategy`, which defaults to `false`. When set to `true`, this restores the old Debezium 2.7 behavior from before DBZ-7882 was introduced. 

- [ ] Needs documentation


vjuranek (approved):

vjuranek:
Merged, thanks @Naros !

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -5755,3 +5755,7 @@ ORA-00600: internal error code, arguments: [kdli_iread], ...
 ```
 +
 This exception can be avoided by upgrading the Oracle db to 19.25 and applying patch 32209850.
+
+*Debezium 2.7 introduced breaking changes with `decimal.handling.mode`, is there a way to use the legacy behavior?*::
+Yes, the legacy behavior can be enabled by setting the connector configuration property `legacy.decimal.handling.strategy` to a value of `true`.
```
roldanbob, line 5760: @Naros Are we intentionally omitting `legacy.decimal.handling.strategy` from the configuration properties table?

Naros: Hi @roldanbob, yes, we generally do not document _internal_ properties and recommend those on an as-needed basis.

roldanbob: > Hi @roldanbob, yes, we generally do not document _internal_ properties and recommend those on an as-needed basis.

Ah, sure. I didn't realize this is internal. Don't we usually prefix those with `internal.` ?

jpechane: Well, this one is probably not internal as such but ore preferred not to be used.

PR #6569, opened 2025-07-17
[ci] Improve/Guard octocat from corner case failures

Naros:


Naros:
Ya I was going to add this to the `DBZ-xxxx` prefix check too, which are the two that most often fail due to these reasons.  I've just been ping ponging around on various contexts the last few days.  I'll add a commit for this shortly, if you can review it on Monday as a final cursory check.

.github/workflows/sanity-check.yml
```diff
@@ -22,31 +22,43 @@ jobs:
           pull_request_number: ${{ github.event.pull_request.number }}
           github_repository: ${{ github.repository }}
         run: |
-          COMMIT_MSGS=$(curl --silent -X "GET" https://api.github.com/repos/$github_repository/pulls/$pull_request_number/commits | jq '.[] | {message: .commit.message}' | jq '.message' | cut -c 2-)
+          echo "PREFIX_COMMITS=false" >> $GITHUB_OUTPUT
+          
+          RESPONSE=$(curl --select -X "GET" https://api.github.com/repos/$github_repository/pulls/$pull_request_number/commits)
```
vjuranek:
```suggestion
          RESPONSE=$(curl --silent -X "GET" https://api.github.com/repos/$github_repository/pulls/$pull_request_number/commits)
```

ani-sha (approved):

PR #6572, opened 2025-07-18
Draft: Optimize the Postgres LSN flushing logic and remove lsnSeen cache

rajdangwal:


jpechane:
Debezium now supports only PostgreSQL versions that support `pg_replication_origin_advance`. I propose to either

* Get rid of `WalPositionLocator` completely or
* keep `slot.seek.to.known.offset.on.start` internal but flip default to `true`. In this case `WalPositionLocator` will be untouched but could be enabled as fallback. The locator logic will be removed in future version.

WDYT?

jpechane:
@rajdangwal @pranavt84 There'll be impact on https://github.com/debezium/debezium/pull/6662 too.
We need to agree on the sequence in whcih we are going to apply the PRs to avoid wasting effort in repeated rebasing of work.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
```diff
@@ -71,76 +65,66 @@ public WalPositionLocator() {
...
+        // Case 2: Found the last processed LSN earlier - now looking for the next LSN to resume from
+        if (foundLastProcessedLsn) {
+            /*
+             * Even after discovering the lastEventStoredLsn, it is possible that the current LSN is still
+             * equals to lastEventStoredLsn, because of the following scenario:
+             *
+             *   LSN | operation
+             *   ---------------
+             *   123 | COMMIT
+             *   123 | BEGIN
```
pranavt84, line 89: I have never seen two separate transactions (i.e. TX1 and TX2) having WAL records (e.g. COMMIT of TX1 and BEGIN of TX2) with the same LSN.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
```diff
@@ -39,13 +36,10 @@ public class WalPositionLocator {
     private final Lsn lastCommitStoredLsn;
     private final Lsn lastEventStoredLsn;
     private final Operation lastProcessedMessageType;
-    private Lsn txStartLsn = null;
-    private Lsn lsnAfterLastEventStoredLsn = null;
     private Lsn firstLsnReceived = null;
     private boolean passMessages = true;
     private Lsn startStreamingLsn = null;
-    private boolean storeLsnAfterLastEventStoredLsn = false;
-    private Set<Lsn> lsnSeen = new HashSet<>(1_000);
```
pranavt84, line 48: Are we allowing duplicate processing of messages?

Although we find an LSN from which to start, and use this LSN while polling during actual processing, we will always retrieve the entire committed batch to which this LSN belongs.
This will result in duplicate processing of some messages.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
rajdangwal: @pranavt84 It was reported and fixed in this PR: https://github.com/debezium/debezium/pull/4357

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
rajdangwal: Thanks for looking into the changes.
Can you please explain which change seems to be allowing duplicate processing of messages?
Removal of lsnSeen cache?

rajdangwal: If yes, I don’t think that would be the cause. The `lsnSeen` cache is only used to store the LSNs encountered during the searching phase and to compare them with the LSNs observed during streaming before reaching the resume LSN. It does not impact the handling of messages that were already processed prior to a connector restart.

I have explained this [here](https://debezium.zulipchat.com/#narrow/channel/348249-community-postgresql/topic/LsnSeen.20cache.20causing.20OOM/near/529435564) (see point 4) describing why the `lsnSeen` cache is not essential for validating LSNs and was most likely introduced due to an incorrect assumption about LSN ordering.

pranavt84: I went through the chat, and I agree that lastProcessedLSN should be flushed each time rather than lastCommitLSN. This was the issue I also faced.

I want to highlight a few observations I have seen while testing. 

1- lastProcessedLSN is used when connecting to PostgreSQL to stream changes. And it is being done twice, first to search for the correct LSN, and once the correct LSN found start again from the lastProcessedLSN, and this time actual message processing will be done, and lsnSeenCache will be used to skip or process messages.

2- If lastProcessedLSN is at commit 123, we will not get the LSNs of this transaction. We will get some other transaction which committed after 123.

For example, a new LSN can be: 120 (BEGIN) -> 120 (INSERT) -> 121 (INSERT) -> 124 (COMMIT). When we get LSN 124, we start streaming from 120 by setting variable startLSN = 120, which belongs to this new transaction and will not be skipped.

3- In your case, where lsn_proc = 275 and commit_lsn = 300 (of T3), we will start streaming from lsn_proc 275 and will get all transactions of T1 (100 ....... 275 .....400). All the LSNs till 275 will be stored in lsnSeen, and when we encounter LSN 276 of T1, we will switch the mode, and the method skipMessage(lsn) will be used to skip the messages until 275.

(So if somehow we are not starting streaming from lsn_proc = 275, we need to fix that.)

Flushing of LSN is not immediate. Even if we are able to flush LSN 275 and restart the connector, we will get all the LSNs of this transaction, and since lastLsnSeen has been removed, we will process the messages again.

You can refer the unit test class I just added for this entire WAL - https://github.com/debezium/debezium/pull/6727

rajdangwal: @pranavt84 Thanks for taking time to explain the scenarios.
However, I still don’t agree with the following statement:
> and lsnSeenCache will be used to skip or process messages.

The LSNs stored in the lsnSeen map are only used to validate that the LSNs observed during the streaming phase are exactly within the set of LSNs seen during the searching phase ([ref](https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java#L164-L169)). It is not involved in the message skipping logic. The scenario in 2 and 3 above would still happen even if the `lsnSeen` map is not there.

The `skipMessage` method only skips messages until it encounters the one with LSN as `startStreamingLsn`. Once that is found, it stops skipping any further messages.

gunnarmorling:
Hey all, agreed that this "cache" shouldn't be needed at all.

In fact, I think that the logic could still be simplified some more even. The ordering guarantee Postgres gives is that the tuple `{Commit LSN, LSN}` is monotonically increasing. So all that's needed is, when resuming after a restart, to skip all those events whose value for that tuple is smaller than what has been stored in the connector offsets as the latest value. No need to look at the type of the events or anything like that.

The reason we even get duplicates after restarting (instead of only everything after the commit flush LSN of the replication slot) is that Debezium unfortunately commits mid-TX LSNs by default (as it doesn't publish COMMIT events to Kafka, unless the TX metadata feature is enabled), in which case Postgres will continue from the beginning of that current in-flight transaction.

Wrote about the details recently here: https://www.morling.dev/blog/postgres-replication-slots-confirmed-flush-lsn-vs-restart-lsn/.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
pranavt84: Gotcha.
Yes, it is not being used directly to skip messages, but it ensures we start processing only after the lastProcessedLSN, otherwise throws the exception.
Earlier, it made sense because the flushing logic was not correct, but after your change, this should not happen.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
```diff
@@ -71,76 +65,66 @@ public WalPositionLocator() {
...
         if (firstLsnReceived == null) {
             firstLsnReceived = currentLsn;
             LOGGER.info("First LSN '{}' received", firstLsnReceived);
         }
-        if (storeLsnAfterLastEventStoredLsn) {
-            // Event that immediately follows the last event seen
-            // We can resume streaming from it
+
+        // Case 1: No previous state exists - start from the beginning
+        if (lastEventStoredLsn == null) {
```
gunnarmorling, line 74: I think all this logic here should be based on the tuple `{Commit LSN, LSN}`. This is what Postgres guarantees to be monotonically increasing. So the logic can be as easy as this:

```
record WalPosition (Lsn commitLsn, Lsn Lsn) {

  boolean smallerThan(WalPosition other) {
    if(commitLsn < other.commitLsn) {
      return true;
    }
    else if(commitLsn > other.commitLsn) {
      return false;
    }
    else {
      // same commit LSN, compare LSN...
    }
  }
}

...
if (new WalPosition.from(incomingEvent).smallerThan(WalPosition.from(committedOffsets)) {
    skip();
}
...
```

Optionally, `pg_replication_slot_advance()` could be used on newer PG versions to skip to the committed offset position right away.

rajdangwal: Thanks for the suggestion, @gunnarmorling.

Even today, the connector uses both the commitLSN and changeLSN to establish a total ordering of events. However, it cannot determine the WAL resume position in a single pass since change events contain only their own LSNs, not the corresponding commit LSNs.

The approach you are suggesting seems to assume that each change event carries its associated commit LSN at processing time. Is that correct? I couldn’t find a way to retrieve the commit LSN for each event during its processing. Please let me know if I have misunderstood anything.

gunnarmorling: Heyy @rajdangwal, sorry for the late response, missed your comment.

> However, it cannot determine the WAL resume position in a single pass since change events contain only their own LSNs, not the corresponding commit LSNs.

Are you referring to the events coming in from Postgres, or the events emitted by Debezium? The latter do contain both commit LSN and event LSN under `source.sequence`, so there is a way to establish the commit LSN when producing the Debezium change events. Specifically, the transaction`s `BEGIN` event contains the commit LSN, so it is known when handling the events coming out of a transaction.

rajdangwal: Thanks for the response, Gunnar.

I was referring to events emitted by PostgreSQL. But yes, as you mentioned, the BEGIN event carries the transaction's [final(commit) LSN](https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java#L253) but it is different from the end LSN that is tracked by the connector.
The COMMIT event exposes both the [commit LSN and the end LSN](https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java#L271-L272). Our connector has been persisting the end LSN as the commit watermark. But the one present in the BEGIN event corresponds to the commit LSN.

See the following example:
```
LSN        | data
-----------+-------------------------------------------------------------
0/166C7C0  | BEGIN 766   (BEGIN shows final LSN = 0/166D920 for this txn)
0/166C828  | INSERT
0/166D8F0  | INSERT
0/166D920  | (commit LSN per BEGIN's "final LSN")
0/166DA20  | COMMIT 766  (end LSN > final LSN)
```

On `source.sequence`: it includes both the event LSN and a commit LSN, but that commit LSN corresponds to the last fully processed transaction, not the transaction that the current event belongs to.

To implement your suggestion, we would have to track the transaction's "final LSN" (as announced in BEGIN) and store it in offsets alongside the event LSN.
One question: can we rely on always receiving a BEGIN before any row-change messages for a transaction? In other words, does the slot guarantee that replay starts at the transaction boundary rather than mid-stream?

gunnarmorling: > The COMMIT event exposes both the [commit LSN and the end LSN](https://github.com/debezium/debezium/blob/main/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/PgOutputMessageDecoder.java#L271-L272).

I see; I wasn't actually aware that there is a difference between the last LSN of the TX and the LSN of the COMMIT event (and that the BEGIN event contains the former rather than the latter). 

> Our connector has been persisting the end LSN as the commit watermark. But the one present in the BEGIN event corresponds to the commit LSN.

Gotcha. In the light of the above, I think that's the right thing to do.

> On source.sequence: it includes both the event LSN and a commit LSN, but that commit LSN corresponds to the last fully processed transaction, not the transaction that the current event belongs to.

That sounds worrisome. I think this should be the end LSN of the current TX. Worth logging a bug report and fixing it (can happen separately though).

> To implement your suggestion, we would have to track the transaction's "final LSN" (as announced in BEGIN) and store it in offsets alongside the event LSN.

SGTM.

> One question: can we rely on always receiving a BEGIN before any row-change messages for a transaction? In other words, does the slot guarantee that replay starts at the transaction boundary rather than mid-stream?

Yes, that's the case as per the testing I did for [this post](https://www.morling.dev/blog/postgres-replication-slots-confirmed-flush-lsn-vs-restart-lsn/) (see section "Mid-Transaction Recovery"). Unfortunately, the connector resumes always from mid transaction by default, i.e. when not enabling the TX metadata topic. The reason being that it doesn't emit an event for the COMMIT message by default, hence no offset can be committed for that, and the connector needs to replay the entire transaction.

jpechane: > That sounds worrisome. I think this should be the end LSN of the current TX. Worth logging a bug report and fixing it (can happen separately though).

@gunnarmorling That's by design. If we'd emit LSN of current TX it would mean that we'd have to accumulate the complete transaction in memory till we get the commit and it's LSN and then emit all changes at once. In fact the `sequence` is only for total ordering so using previous commit LSN serves the same purpose well.

Naros:

Hi @rajdangwal, please sign off all commits with:

```
git commit -s
```

Naros:

PR #6575, opened 2025-07-18
DBZ-9235 Log offset map rather than offset class

### LINKED ISSUE #1376: Reduce duplicity of offset logging at startup [DBZ-9235]
Migrated from [DBZ-9235](https://issues.redhat.com/browse/DBZ-9235)

h3. Problem
The below log line --

{code:java}
[2025-07-18 15:33:55,157] [task-thread-testTrigger-55-0] INFO [testTrigger-55|task-0] Found previous offset BinlogOffsetContext{sourceInfoSchema=Schema{io.debezium.connector.mysql.Source:STRUCT}, sourceInfo=BinlogSourceInfo{currentGtid='null', currentBinlogFilename='mysql-binn.000011', currentBinlogPosition=5520, currentRowNumber=0, serverId=0, sourceTime=null, threadId=-1, currentQuery='null', tableIds=[], databaseName='null'}, snapshotCompleted=false, transactionContext=TransactionContext [currentTransactionId=null, perTableEventCount={}, totalEventCount=0], restartGtidSet='8fede472-41b1-11f0-80d3-9ff8011a7d28:1-39', currentGtidSet='8fede472-41b1-11f0-80d3-9ff8011a7d28:1-39', restartBinlogFilename='mysql-binn.000011', restartBinlogPosition=5520, restartRowsToSkip=1, restartEventsToSkip=2, currentEventLengthInBytes=0, inTransaction=false, transactionId='null', incrementalSnapshotContext=IncrementalSnapshotContext [windowOpened=false, chunkEndPosition=null, dataCollectionsToSnapshot=[], lastEventKeySent=null, maximumKey=null]} (io.debezium.connector.mysql.MySqlConnectorTask:150)
{code}

Even though the server did have GTIDs enabled, the `currentGtid` value in the above log is always printed null. This creates confusion for the users.

h3. Solution
While logging the offset, there is no need to include the `sourceInfo`. Semantically `sourceInfo` is only relevant as a field in the change events created alongside fields `before` and `after`. `SourceInfo` describes the real-time position and context of the connector while processing events (current binlog file, position, GTID, query, threadid etc.), whereas `Offset` represents the last durable checkpoint from which the connector will resume after a restart.

Looking at it this way, it looks like `sourceInfo` should not be a part of `toString` for binlog related offset class.

akanimesh7:
Fixes debezium/dbz#1376

### DIFF
--- a/debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorTask.java
+++ b/debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/MariaDbConnectorTask.java
@@ -177,7 +177,7 @@ public class MariaDbConnectorTask extends BinlogSourceTask<MariaDbPartition, Mar
             LOGGER.info("No previous offset found");
         }
         else {
-            LOGGER.info("Found previous offset {}", previousOffset);
+            LOGGER.info("Found previous offset {}", previousOffset.getOffset());
         }
 
         // Set up the task record queue ...
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorTask.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorTask.java
@@ -169,7 +169,7 @@ public class MySqlConnectorTask extends BinlogSourceTask<MySqlPartition, MySqlOf
             LOGGER.info("No previous offset found");
         }
         else {
-            LOGGER.info("Found previous offset {}", previousOffset);
+            LOGGER.info("Found previous offset {}", previousOffset.getOffset());
         }
 
         // Set up the task record queue ...
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
@@ -119,7 +119,7 @@ public class OracleConnectorTask extends BaseSourceTask<OraclePartition, OracleO
             LOGGER.info("No previous offset found");
         }
         else {
-            LOGGER.info("Found previous offset {}", previousOffset);
+            LOGGER.info("Found previous offset {}", previousOffset.getOffset());
         }
 
         Clock clock = Clock.system();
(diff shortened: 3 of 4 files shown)


Naros (commented):
Hi @akanimesh7 I'm not sure I would necessarily agree with this change as is. For example, there are some attributes, like `currentQuery` that would only be visible when the class is logged by including the `SourceInfo`.

I'd rather like to see the logged data cleaned up to minimize what gets logged so we don't have the same value listed multiple times called different things, rather than loosing visibility to something that's only part of the `SourceInfo` object.

akanimesh7:
Ideally `SourceInfo` should contain the state of the source at the time of processing a particular event. That's the reason it contains fields like -- 
1. `currentGtid` which is the txn id of the specific txn that is being dealt with.
2. `serverId` -- the original server from where event originated
3. `threadID` -- related to client session. Not relevant on a task re-start, since it's bound to change.

However, Looking at the specific log line `Found previous offset`. This log is intended to provide an idea of the offset from where the task is starting up. There is no importance/need/relevance of `sourceInfo` here currently.  Adding more --

I think that's why there is a subtle difference in the fields that offset info has. For example -- `restartGtidSet` -- this is the set of gtids that have been processed already on the contrary to `currentGtid` in the souce info.

Consider a setup in which a mysql connector is connecting to a replica which is getting events from three sources -- u1, u2 and u3. Let's say at any given point in time the event being processed currently is `u2:123`. So at this point, source info will have the `currentGtid` as `u2:123`. Whereas the offset will have `restartGtidSet` as `u1:101,u2:122.u3:189`. To re-start task from correct point what's relevant is `u1:101,u2:122.u3:189` and not `u2:123`. Also this is the reason why even though fields like `currentBinlogFilename` and `currentBinlogPosition` are populated in the source info from the offset, but `currentGtid` is still kept null, since it's not possible to determine in this case.

Naros:
Then would it make since to:
* Use `OffsetContext#getOffset` so you have a Map that gets logged instead to align with the offset state that is in the offset topic rather than `#toString()`.
* Also align the behavior across all connectors, not just make MariaDB and MySQL different from the others

akanimesh7:
@Naros That makes sense. Updated the PR

jpechane (approved):
@akanimesh7 LGTM, thanks. It seems we need a change in spanner and vitess connector too. Then this is good to go.

Naros:

Hi @akanimesh7, please sign off all commits with:

```
git commit -s
```

Naros:
Superseded by #7877.

PR #6577, opened 2025-07-18
DBZ-9237 Do not immediately fail JDBC sink if create table fails

Naros:
https://issues.redhat.com/browse/DBZ-9237

So the issue here is that when more than one task executes a `CREATE TABLE`, the non-first task will throw a runtime exception, `ConstraintViolationException` in case of PostgreSQL. These runtime exceptions were not being caught previously, and immediately forced the connector to fail. 

This PR is designed to instead fallthrough to the `ALTER TABLE` if the `CREATE TABLE` failed, as a more resilient way to deal with multiple tasks that may attempt to create the table concurrently, allowing the non-first tasks to safely fall through and process the insert, update, or upsert without failing.


vjuranek (commented):
The changes LGTM, but test failures seem to be related.

jpechane:
@Naros Do we really want to take broad approach? Wouldn't it be safer to just add the another exception to the list instead of handling everything?

Naros:
I think that's reasonable @jpechane, I pulled back to using SQLException and JDBCException. 
Do we agree that's reasonable?

jpechane:
@Naros Applied, thanks/

PR #6582, opened 2025-07-21
DBZ-9241 Cap each large SCN jump based on max batch size

Naros:
https://issues.redhat.com/browse/DBZ-9241


Naros:
@nathan-smit-1 this might be an interesting PR for you

Naros:
Based on the feedback I've received from internal testing, this change has reduced latency during high bursts from 70-80 minutes to 20-25 minutes. I should caveat that this test is creating a significant amount of batch data and is using an upper bounds `log.mining.batch.size.max` of 1M (to give an idea on their volume).  

That's nearly a 69% savings :tada:

vjuranek:
Merged, thanks @Naros ! And really nice improvement!

PR #6589, opened 2025-07-23
DBZ-9247 Added undefined_table error handling for PostgresConnectorTask heartbeat

ifig:
Added undefined_table error handling for PostgresConnectorTask heartbeat
Error code `42P01` throws a new `DebeziumException`


jpechane:
@ifig Applied, thanks!

PR #6590, opened 2025-07-23
DBZ-9272 Add option to use CTE-based LogMiner queries

Naros:
https://issues.redhat.com/browse/DBZ-9272 

While the `log.mining.query.filter.mode` provides ways to push query criteria to the Oracle database to filter the change stream so that we get the least amount of events in the connector, this is not always adequate, especially in situations where there may be far more activity for non-captured tables. In these situations, this can lead to the connector receiving a large number of transaction commit or rollback marker events that have no changes in the transaction.

This PR introduces a new internal, experimental feature that enables the use of a CTE-based query. The CTE query is designed to apply the same filter as the main query, but limits the scope to only non-transaction marker changes (inserts, updates, deletes, LOB operations, DDL, and so on). The outer query then uses the list of transaction IDs generated by the CTE query to limit the final result set provided to Debezium only to contain relevant transactions.

There are a few caveats:

1. The use of `internal.log.mining.use.cte.query` requires `log.mining.query.filter.mode` to be set.
2. Due to the CTE being cached in the PGA, large `log.mining.batch.size.*` settings may lead to PGA_AGGREGATE_LIMIT errors. If so, the `log.mining.batch.size.*` settings should be reduced to a smaller chunk pass.
3. The CTE uses an Oracle hint to enforce join orders, and if the database chooses to ignore the hint, the events could  be provided out of order, which could cause events to be missed. This will require some testing across variety of environments.
4. The CTE query and the outer query will force LogMiner to read the logs twice. This should generally be a non-issue if your logs are not sized too large and your batch sizes are reasonable. But this will may cause slightly higher IO on the database. In effect, we're trying to take advantage of the database to optimize, sort, and return the data set as efficiently as possible before it goes over the network wire.

@nathan-smit-1 @jeremy-l-ford - This may be an interesting thing to play with :tada:

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -737,6 +737,15 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             .withDefault(false)
             .withDescription("Uses the legacy decimal handling behavior before DBZ-7882");
 
+    public static final Field LOG_MINING_USE_CTE_QUERY = Field.createInternal("log.mining.use.cte.query")
+            .withDisplayName("Use CTE-based query")
+            .withType(Type.BOOLEAN)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withDefault(false)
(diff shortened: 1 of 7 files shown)


vjuranek:
Test failures are unrelated

vjuranek:
Merged, thanks!

PR #6593, opened 2025-07-24
DBZ-9248 Add extended.headers.enabled configuration to enable/disable extended headers

mfvitale:
With OpneLineage integration we added  [Extended Context Headers](https://github.com/debezium/debezium/blob/d4779a879bcffdd36b632d237b02dc5a8c331eb0/debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java#L559) to every emitted records to support [OpenLineage SMT.](https://github.com/debezium/debezium/blob/main/debezium-core/src/main/java/io/debezium/transforms/openlineage/OpenLineage.java)

In scenario where OpenLineage integration is not of interest this could cause waste of space. 

We this change, we added a configuration to eventually disable the emission of those headers, there remain enabled by default.

closes: https://issues.redhat.com/browse/DBZ-9248


Naros (approved):

Naros:
Thanks @mfvitale, applied!

PR #6599, opened 2025-07-28
DBZ-9288

nrkljo:
Uppdate Informix JDBC Driver to 4.50.12


jpechane:
@nrkljo Applied, thanks

PR #6603, opened 2025-07-28
DBZ-8962 Debezium Engine Quarkus Extension: custom deserializer

kmos:
## Context

By default, Debezium emits change events as structures like `ChangeEvent<V, T>`. While flexible, this model forces developers to manually deserialize payloads during the *capturing phase* in the quarkus extension.

## Decision

We propose mapping events to POJOs or Java records using a Jackson deserializer. The user can receive directly mapped events:

```java
@ApplicationScoped  
class ProductHandler {
  
    @Capturing(destination = "dbserver1.public.products")
    public void capture(CapturingEvent<Product> event) {
      // your logic
    }
} 
```

or

```java
@ApplicationScoped  
class ProductHandler {
  
    @Capturing(destination = "dbserver1.public.products")
    public void capture(Product event) {
      // your logic
    }
} 
```

the events are filtered by `destination` which is the `topic`.

To setup the deserializer, the user should implements a `Deserializer<T>`:

```java
public class ProductDeserializer extends ObjectMapperDeserializer<Product> {
    public ProductDeserializer() {
        super(Product.class);
    }
}
```

map it in the configuration with the associated `destination`:

```properties
## deserializer
quarkus.debezium.capturing.product.destination=dbserver1.public.products
quarkus.debezium.capturing.product.deserializer=io.quarkus.sample.app.deserializer.ProductDeserializer
```

## Side notes

- introduced the `CapturingEvent<T>` interface
- introduced the `Deserializer<T>` that can be used to add new deserializer like `jsonb`

## References

#closes https://issues.redhat.com/browse/DBZ-8962
#link https://github.com/debezium/debezium-design-documents/pull/16

### DIFF
--- a/quarkus-debezium-parent/quarkus-debezium-postgres-parent/runtime/pom.xml
+++ b/quarkus-debezium-parent/quarkus-debezium-postgres-parent/runtime/pom.xml
@@ -12,6 +12,11 @@
 
     <name>Quarkus Debezium :: Extension :: Postgres :: Runtime</name>
 
+    <properties>
+        <!-- The below override should be removed when kafka is upgraded to 4.0.0 and quarkus to 3.25.0 -->
+        <kafka.version.override>4.0.0</kafka.version.override>
+    </properties>
+
     <dependencies>
         <dependency>
             <groupId>${project.groupId}</groupId>
@@ -58,6 +63,7 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>connect-runtime</artifactId>
+            <version>${kafka.version.override}</version>
             <exclusions>
                 <exclusion>
                     <groupId>com.fasterxml.jackson.core</groupId>
@@ -120,6 +126,7 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>connect-json</artifactId>
+            <version>${kafka.version.override}</version>
             <exclusions>
                 <exclusion>
                     <groupId>com.fasterxml</groupId>
@@ -157,6 +164,13 @@
         <dependency>
             <groupId>org.apache.kafka</groupId>
             <artifactId>connect-transforms</artifactId>
+            <version>${kafka.version.override}</version>
+            <exclusions>
+                <exclusion>
+                    <groupId>org.apache.kafka</groupId>
+                    <artifactId>kafka-clients</artifactId>
+                </exclusion>
+            </exclusions>
         </dependency>
 
         <dependency>


quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/EngineProcessor.java
```diff
@@ -202,6 +229,33 @@ void registerClassesThatAreLoadedThroughReflection(
...
+                                .reason(getClass().getName())
+                                .constructors(true)
+                                .methods(true)
+                                .fields(true)
+                                .build()));
+
+        debeziumEngineConfiguration
+                .capturing()
+                .values()
+                .forEach(capturing -> {
```
mfvitale: Maybe here a method could improve the readability.

quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/EngineProcessor.java
```diff
@@ -202,6 +229,33 @@ void registerClassesThatAreLoadedThroughReflection(
                         .reason(getClass().getName())
                         .build()));
 
+        indexBuildItem
+                .getIndex()
+                .getAllKnownSubclasses(DotName.createSimple(ObjectMapperDeserializer.class))
+                .stream()
+                .flatMap(clazz -> clazz.superClassType().asParameterizedType().arguments().stream())
+                .forEach(clazz -> reflectiveClasses.produce(
```
mfvitale, line 231: Maybe here a method could improve the readability.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/SourceRecordDeserializer.java
```diff
@@ -0,0 +1,74 @@
...
+    @Override
+    public CapturingEvent<T> deserialize(CapturingEvent<SourceRecord> event) {
+        byte[] data = converter.fromConnectData(
+                event.record().topic(),
+                event.record().valueSchema(),
+                event.record().value());
+
+        return switch (event) {
+            case Create<SourceRecord> record -> new Create<>(
+                    deserializer.deserialize(data, "after"),
```
mfvitale: What if here the event has been processed with `ExtractNewRecordState` SMT?

mfvitale:
Great work @kmos . I left few minor comments. 

> @ApplicationScoped  
> class ProductHandler {
>   
>     @Capturing(destination = "dbserver1.public.products")
>     public void capture(Product event) {
>       // your logic
>     }
> } 
> 
> ## deserializer
> quarkus.debezium.capturing.product.destination=dbserver1.public.products
> quarkus.debezium.capturing.product.deserializer=io.quarkus.sample.app.deserializer.ProductDeserializer

Is really necessary to declare the destination both on the `@Capture` annotation and in the properties?
Also I could see another annotation for the deserializer.

quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/InvokerGenerator.java
```diff
@@ -23,14 +26,25 @@
...
 import io.quarkus.runtime.util.HashUtil;
 
-public class InvokerGenerator {
+/**
+ *
+ * this generator is substituted by {@link CapturingInvokerGenerator} which use {@link io.debezium.runtime.CapturingEvent}
+ *
+ */
+@Deprecated
+public class InvokerGenerator implements CapturingInvokerGenerator {
```
Naros: I don't think we need to worry about using `@Deprecated` at this stage. I suggest removing the implementation altogether.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/OperationMapper.java
```diff
@@ -0,0 +1,69 @@
...
+import io.debezium.runtime.CapturingEvent.Update;
+
+public class OperationMapper {
+
+    public static final String NOT_AVAILABLE = "NOT_AVAILABLE";
+
+    public static CapturingEvent<SourceRecord> from(ChangeEvent<SourceRecord, SourceRecord> record) {
+        Struct payload = (Struct) record.value().value();
+
+        if (payload.schema().field("op") == null) {
```
Naros: We should guard against the schema being null for tombstones.

Naros (commented):
There are a few interfaces I'd prefer to see some documentation on just for future knowledge and understanding by those who aren't so familiar with the code.  Otherwise, just a few inline comments, too.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/OperationMapper.java
jpechane: Also for value above

quarkus-debezium-parent/quarkus-debezium-engine-spi-parent/runtime/src/main/java/io/debezium/runtime/Capturing.java
```diff
@@ -17,4 +17,7 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Incubating()
 public @interface Capturing {
+    String ALL = "*";
+
+    String destination() default ALL;
```
jpechane, line 22: What if there are multiple `@Capturing` instrances with the same destination or multiple catch-all instances?

quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/CapturingObjectInvokerGenerator.java
```diff
@@ -0,0 +1,123 @@
...
+                capture.invokeVirtualMethod(methodDescriptor, delegate, event);
+                capture.returnVoid();
+            }
+
+            try (MethodCreator destination = invoker.getMethodCreator("destination", String.class)) {
+                Optional.ofNullable(methodInfo
+                        .annotation(DebeziumDotNames.CAPTURING)
+                        .value("destination"))
+                        .map(AnnotationValue::asString)
+                        .ifPresentOrElse(s -> destination.returnValue(destination.load(s)),
```
jpechane: What if `destination` is an empty string?

quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/InvokerGenerator.java
jpechane: +1
You can either remove or re-use existing implmentation

quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/InvokerGenerator.java
```diff
@@ -23,14 +26,25 @@
...
 
     private final ClassOutput output;
 
     public InvokerGenerator(ClassOutput classOutput) {
         this.output = classOutput;
     }
 
+    @Override
+    public boolean isCompatible(Type type) {
+        return ParameterizedType.create(RecordChangeEvent.class, Type.create(SourceRecord.class)).equals(type);
```
jpechane: How frequently is this called? If for each event then `ParameterizedType.create(RecordChangeEvent.class, Type.create(SourceRecord.class))` could be pre-created and re-used.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/OperationMapper.java
```diff
@@ -0,0 +1,69 @@
...
+            return new Message<>(
+                    record.value(),
+                    record.destination(),
+                    NOT_AVAILABLE,
+                    record.headers());
+        }
+
+        return switch (Operation.forCode(payload.getString("op"))) {
+            case READ -> new Read<>(record.value(),
```
jpechane: Please use ate same argumenta fromatting in all branches.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/OperationMapper.java
```diff
@@ -0,0 +1,69 @@
...
+            case CREATE -> new Create<>(record.value(),
+                    record.destination(),
+                    NOT_AVAILABLE,
+                    record.headers());
+            case UPDATE -> new Update<>(
+                    record.value(),
+                    record.destination(),
+                    NOT_AVAILABLE,
+                    record.headers());
+
```
jpechane: Remove or add empy line after each branch

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/SourceRecordDeserializer.java
```diff
@@ -0,0 +1,74 @@
...
+
+public class SourceRecordDeserializer<T> implements CapturingEventDeserializer<T, SourceRecord> {
+
+    private final Deserializer<T> deserializer;
+    private final Converter converter;
+
+    public SourceRecordDeserializer(Deserializer<T> deserializer, Converter converter) {
+        this.deserializer = deserializer;
+        this.converter = converter;
+        this.converter.configure(Map.of("schemas.enable", true), false);
```
jpechane:
```suggestion     

this.converter.configure(Map.of(org.apache.kafka.connect.json.JsonConverterConfig.SCHEMAS_ENABLE_CONFIG, true), false);
```

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/SourceRecordDeserializer.java
jpechane:
```suggestion
                    deserializer.deserialize(data, Enevelope.FieldName.AFTER),
```

Same for others

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/OperationMapper.java
```diff
@@ -0,0 +1,69 @@
...
+        if (payload.schema().field("op") == null) {
+            return new Message<>(
+                    record.value(),
+                    record.destination(),
+                    NOT_AVAILABLE,
+                    record.headers());
+        }
+
+        return switch (Operation.forCode(payload.getString("op"))) {
```
jpechane:
```suggestion
        return switch (Operation.forCode(payload.getString(Envelope.FieldName.OPERATION))) {
```

quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/InvokerGenerator.java
kmos: This method is called at [build time](https://quarkus.io/blog/quarkus-metaprogramming/) to create the classes that invoke the methods annotated with `@Capturing`. It should be called as many times as there are annotations in your code. Once the classes are created, the method is not called anymore.

kmos:
> There are a few interfaces I'd prefer to see some documentation on just for future knowledge and understanding by those who aren't so familiar with the code. Otherwise, just a few inline comments, too.

@Naros Yeah, I think I rushed too much in this pull request :eyes:

quarkus-debezium-parent/quarkus-debezium-engine-spi-parent/runtime/src/main/java/io/debezium/runtime/Capturing.java
kmos: Interesting question. Actually if there are multiple `@Capturing` with catch-all destination, it should be called the first in the list. To be more precise, if there are also some `capturing` with destination and a catch-all, the catch-all is called only for the destinations that aren't mapped with a method.

WDYT? What should be the default behavior? I honestly didn't think too much about it :eyes:

jpechane: IMHO this is then non-deterministic behaviour. My proposal is that if there is more than one catch all `@Capturing` or more than one for the same destination there should be an error on application start. Later on we can relax this behaviour if we find other ways how to discriminate between the methods to be called or maybe we allow all of them to be called - but that would open another can of worms.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/SourceRecordDeserializer.java
kmos: Well, if I understand correctly you are talking about the case of an user that wants to use an `ExtractNewRecordState` SMT in conjuction with a `Deserializer`. My suggestion for the user should be to override also the method `deserialize.data(bytes, String)` to match the case. Instead, if we want to be super smooth, we should understand if `ExtractNewRecordState` is applied and then change the *strategy* to deserialize the data. But my question is: what are the SMT that we want to support for this?

WDYT? @mfvitale @Naros @jpechane

quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/InvokerGenerator.java
kmos: ok!

kmos:
> Why do we need to go thourg JSON serialization and desrialization to get domain class? Can't we skip it and go directly from `SourceRecord` to a Java object?

I agree. Since we weren't sure whether this feature would be used in many cases, we went with the quickest solution.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/SourceRecordDeserializer.java
jpechane: I think for start we could just do it simple

1. Is it Envelope? If yes extract after part
2. It is not enevlope? Then work directly with the whole message

IIRC we do it more or less the same way on JDBC Sink

quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/CapturingObjectInvokerGenerator.java
kmos: well actually is acceptable, but we can be more robust :eyes:

jpechane:
@kmos Good job, applied, thanks!

PR #6604, opened 2025-07-28
DBZ-9292 Unchecked exception from OffsetStorageWriter.doFlush() in As…

silan0176:
Unchecked exception from OffsetStorageWriter.doFlush() in AsyncEmbeddedEngine leaves semaphore in OffsetStorageWriter unreleased and probably causes engine to fail


silan0176:
invalid username

PR #6605, opened 2025-07-28
DBZ-9293 Fix reselect using VariableScaleDecimal column type

Naros:
https://issues.redhat.com/browse/DBZ-9293

### DIFF
--- a/debezium-core/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
+++ b/debezium-core/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
@@ -276,8 +276,11 @@ public class ReselectColumnsPostProcessor implements PostProcessor, BeanRegistry
 
     private Object resolveKeyFieldValue(Struct key, org.apache.kafka.connect.data.Field field) {
         if (field.schema() != null && VariableScaleDecimal.LOGICAL_NAME.equals(field.schema().name())) {
-            SpecialValueDecimal decimal = VariableScaleDecimal.toLogical(key.getStruct(field.name()));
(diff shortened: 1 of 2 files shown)


Naros:
@jpechane test failures are unrelated.

debezium-core/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
```diff
@@ -272,6 +274,14 @@ public void injectBeanRegistry(BeanRegistry beanRegistry) {
         this.schema = beanRegistry.lookupByName(StandardBeanNames.DATABASE_SCHEMA, RelationalDatabaseSchema.class);
     }
 
+    private Object resolveKeyFieldValue(Struct key, org.apache.kafka.connect.data.Field field) {
+        if (field.schema() != null && VariableScaleDecimal.LOGICAL_NAME.equals(field.schema().name())) {
+            SpecialValueDecimal decimal = VariableScaleDecimal.toLogical(key.getStruct(field.name()));
+            return decimal.getWrappedValue();
```
jpechane: Can the field value be null? If yes then this line is susceptible to NPE.

jpechane (commented):
@Naros LGTM, thanks. I left one comment wrt NPE. When addressed feel fre to merge yourself.

debezium-core/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
Naros: I suppose it's possible for some databases that allow it for primary key columns, so I'll guard it to be safe.

PR #6606, opened 2025-07-28
DBZ-9132 Fix parsing of UNISTR values with nested `||` sequence

Naros:
https://issues.redhat.com/browse/DBZ-9132


jpechane:
@Naros Applied, thanks

PR #6607, opened 2025-07-28
DBZ-8870 Allow ORA-00310 to trigger a retry on mine start

Naros:
https://issues.redhat.com/browse/DBZ-8870


jpechane:
@Naros Applied, thanks

PR #6608, opened 2025-07-28
DBZ-8747 Decouple mining range from fetch range

Naros:
https://issues.redhat.com/browse/DBZ-8747

## Details

When the mining range and fetch range are identical, this can lead to a variety of issues, including but not limited to seeing unsupported operations due to transactions starting outside the mine window.

This change aims to decouple the mining and fetch ranges, where the mining range will always start at the eldest transaction in the cache while the fetch window will always increase based on batch windows.

This change also simplifies the mining window logic, as the same behavior is used regardless of LOB settings. This means that connectors using LOB will now read logs slightly faster, as the read operation no longer needs to rollback to the start of the transaction, but rather to where the last step left off; thus improving performance.

---

- [x] Test against stadnalone Oracle with SQL Loader use case
- [x] Test against multi-node Oracle RAC with SQL Loader use case

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/AbstractLogMinerQueryBuilder.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/AbstractLogMinerQueryBuilder.java
@@ -111,14 +111,14 @@ public abstract class AbstractLogMinerQueryBuilder implements LogMinerQueryBuild
             //
             // Oracle always stores usernames as upper-case.
             // This predicate build applies upper-case to the provided value lists from the configuration.
-            return applyToNonTransactionMarkers(IncludeExcludeInClause.builder()
+            return IncludeExcludeInClause.builder()
                     .withField("USERNAME")
                     .withFilterMode(LogMiningQueryFilterMode.IN)
                     .withDefaultIncludeValues(Collections.singletonList(UNKNOWN_USERNAME))
                     .withIncludeValues(connectorConfig.getLogMiningUsernameIncludes())
                     .withExcludeValues(connectorConfig.getLogMiningUsernameExcludes())
                     .caseInsensitive()
-                    .build());
+                    .build();
         }
         return EMPTY;
     }
@@ -132,22 +132,15 @@ public abstract class AbstractLogMinerQueryBuilder implements LogMinerQueryBuild
         if (!LogMiningQueryFilterMode.NONE.equals(connectorConfig.getLogMiningQueryFilterMode())) {
             // Only filter client ids when using IN and REGEX modes
             // Client id filters always use an IN-clause predicate
-            return applyToNonTransactionMarkers(IncludeExcludeInClause.builder()
+            return IncludeExcludeInClause.builder()
                     .withField("CLIENT_ID")
                     .withFilterMode(LogMiningQueryFilterMode.IN)
                     .withIncludeValues(connectorConfig.getLogMiningClientIdIncludes())
                     .withExcludeValues(connectorConfig.getLogMiningClientIdExcludes())
                     .caseInsensitive()
-                    .build());
+                    .build();
         }
         return EMPTY;
-    }
-
-    private String applyToNonTransactionMarkers(String fragment) {
-        if (!Strings.isNullOrBlank(fragment)) {
-            return "(CASE WHEN OPERATION_CODE IN (6,7,36) THEN 1 ELSE CASE WHEN " + fragment + " THEN 1 ELSE 0 END END = 1)";
-        }
-        return fragment;
     }
 
     /**
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerQueryBuilder.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerQueryBuilder.java
@@ -11,6 +11,7 @@ import java.util.List;
 
 import io.debezium.connector.oracle.OracleConnectorConfig;
 import io.debezium.connector.oracle.logminer.AbstractLogMinerQueryBuilder;
+import io.debezium.util.Strings;
 
 /**
  * A builder that is responsible for producing the query to be executed against LogMiner when operating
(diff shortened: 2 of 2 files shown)


Naros:
@nathan-smit-1 do you happen to have a UAT environment you could throw this PR into and see if you notice any differences from the current 3.2.0.Final in terms of performance without `lob.enabled` set?

Naros:
As mentioned on Zulip, I've prepared a few pre-built options for testing this PR
Build Artifact - [here](https://drive.google.com/file/d/1ov37ZWGSdgTANOJpUhaTEbILFaPUwmDq/view?usp=sharing) 
Container image - quay.io/debezium/connect:DBZ-8747

Naros:
Just an update @nathan-smit-1, but Oracle RAC is still broken, even with this fix. Looking into it.

danny-johnson-nisc:
I am seeing odd behavior when testing this compared to 3.2. I built debezium from the DBZ-8747 branch directly yesterday so it is up to date. 
When I switch to 3.3 my connector starts spamming diffs to kafka after the snapshot, even when no database activity is happening. There is also a lag spike in the metrics that is climbing at a consistent rate (the little dip back down was me restarting the program):
image
image

here is an example log: `21179 records sent during previous 00:01:20.119, last recorded offset of {server=REDACTED} partition is {commit_scn=692425956:1:050000009c820700, txId=050000009c820700, snapshot_scn=692355392, scn=692425954`

when I switch back to 3.2 it stops churning after the snapshot. I tested and re-confirmed 3 times so it was not a fluke.
I also confirmed our database is standalone and not RAC, as `SELECT value FROM v$parameter WHERE name = 'cluster_database';` returns FALSE


edit:
this was being caused by DBZ-9340

Naros:
As discussed, the latest commit about the username/client id filtering should address situations where large transactions are split across mining steps and this makes sure the decoupling behavior should work for filtering the data events as intended.

danny-johnson-nisc:
The client_id filtering is now working for large transactions after commit a3ffa33 😄

Naros:
Test failures are unrelated.

vjuranek (approved):
LGTM

vjuranek:
Merged, thanks @Naros !

gong:
Will it cause session start scn don't be changed and mining gap is too large when database have long running transaction? @Naros

Naros:
The longer you have long-running transactions, the more logs that are added to the mining session, but unless a redo entry in the data fetch start/end range requires going back in time, older logs should be skipped unless needed @gong

gong:
> The longer you have long-running transactions, the more logs that are added to the mining session, but unless a redo entry in the data fetch start/end range requires going back in time, older logs should be skipped unless needed @gong

Hi, @Naros . This PR could cause mining gap is too large for long-running transactions, this will effect database performance? Event if fetch range is small.

gong:
Hi, @Naros. I met a missing data situation. oracle is RAC. I miss data scn is belong to thread1 and `lastProcessedScn` of current mining is updated from thread2.  It take place online redo log switch to archive log. Next mining start scn  is pushed forward, it cause miss data from thread1. Maybe we need compare latest scn between with multi thread and get minimal value.

Naros:
Hi @gong 

> > The longer you have long-running transactions, the more logs that are added to the mining session, but unless a redo entry in the data fetch start/end range requires going back in time, older logs should be skipped unless needed @gong
> 
> Hi, @Naros . This PR could cause mining gap is too large for long-running transactions, this will effect database performance? Event if fetch range is small.

I understand the concern, but I'm afraid that Debezium is at the mercy of LogMiner. There are simply situations where logs are mandatory this way to avoid things like _unsupported operations_, _partial redo entries_, and various other oddities, all because we were not adhering to LogMiner's quirks.

The only option I see is if users don't like the fact LogMiner works this way, we could allow you to use the old legacy behavior at your own risk, but my concern is that doing that could lead to more questions about data loss because someone wants faster handling at the sake of reliability. :shrug:   For me, it's a no-win situation, unfortunately.

> I met a missing data situation. oracle is RAC. I miss data scn is belong to thread1 and `lastProcessedScn` of current mining is updated from thread2. It take place online redo log switch to archive log. Next mining start scn is pushed forward, it cause miss data from thread1. Maybe we need compare latest scn between with multi thread and get minimal value.

I don't see how this is possible. Before we ever fetch the data, we look at `V$THREAD` across all redo threads, and we cap the upper boundary to be the MIN across all threads, so in your use case, we should never read data from thread 2 that has any SCN that exceeds the last flushed SCN of redo thread 1.

If you have evidence of this, please raise an issue https://github.com/debezium/dbz and please zip and attach the connector logs on trace level that illustrate the data being missed.

gong:
> Hi @gong
> 
> > > The longer you have long-running transactions, the more logs that are added to the mining session, but unless a redo entry in the data fetch start/end range requires going back in time, older logs should be skipped unless needed @gong
> > 
> > 
> > Hi, @Naros . This PR could cause mining gap is too large for long-running transactions, this will effect database performance? Event if fetch range is small.
> 
> I understand the concern, but I'm afraid that Debezium is at the mercy of LogMiner. There are simply situations where logs are mandatory this way to avoid things like _unsupported operations_, _partial redo entries_, and various other oddities, all because we were not adhering to LogMiner's quirks.
> 
> The only option I see is if users don't like the fact LogMiner works this way, we could allow you to use the old legacy behavior at your own risk, but my concern is that doing that could lead to more questions about data loss because someone wants faster handling at the sake of reliability. 🤷 For me, it's a no-win situation, unfortunately.
> 
> > I met a missing data situation. oracle is RAC. I miss data scn is belong to thread1 and `lastProcessedScn` of current mining is updated from thread2. It take place online redo log switch to archive log. Next mining start scn is pushed forward, it cause miss data from thread1. Maybe we need compare latest scn between with multi thread and get minimal value.
> 
> I don't see how this is possible. Before we ever fetch the data, we look at `V$THREAD` across all redo threads, and we cap the upper boundary to be the MIN across all threads, so in your use case, we should never read data from thread 2 that has any SCN that exceeds the last flushed SCN of redo thread 1.
> 
> If you have evidence of this, please raise an issue https://github.com/debezium/dbz and please zip and attach the connector logs on trace level that illustrate the data being missed.

Hi, @Naros Thanks for your replay, I can't reproduce _unsupported operation when transaction across mining. About RAC scn problem, I don't have some full log information, I can provide some mining file info.

Naros:
That would be great @gong, just raise a new issue in the [issue tracker](https://github.com/debezium/dbz) and attach the details and we'll take a look.

PR #6615, opened 2025-07-30
DBZ-8991 Add `NumberOfEventsInBuffer` JMX metric

Naros:
https://issues.redhat.com/browse/DBZ-8991


jpechane:
@Naros Applied, thanks

PR #6616, opened 2025-07-30
DBZ-9296 Preload Mockito agent to avoid Java agent dynamic loading restriction in further JDK

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-9296

[JEP-451](https://openjdk.org/jeps/451) added a warning to prepare to Disallow the Dynamic Loading of Agents.

This changes instruct `maven-surfire-plugin` and `maven-failsafe-plugin` to pre-loads the mockito agent. 

Effectively the WARN disappears.

### DIFF
--- a/debezium-bom/pom.xml
+++ b/debezium-bom/pom.xml
@@ -70,7 +70,6 @@
         <version.fest>1.4</version.fest>
         <version.assertj>3.27.3</version.assertj>
         <version.jmh>1.21</version.jmh>
-        <version.mockito>5.17.0</version.mockito>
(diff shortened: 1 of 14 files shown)


jpechane:
It seems the failures are related

mfvitale:
Yes, working on it. If this works, it requires PR on other repos that uses mockito. So let's see if this at least passes.

mfvitale:
@jpechane this should be ready now.

vjuranek:
Merged, thanks!

PR #6619, opened 2025-07-30
DBZ-9305 Fix pgoutput decoder metadata query with wildcard characters

Naros:
https://issues.redhat.com/browse/DBZ-9305


Chitrae92 (approved):
Changes looks good to me 👍

Naros:
The test failures were related to schema object taint from the added test.  I've made sure the test cleans up after itself, so we should ideally be green now on CI.  Thanks for the catch @Chitrae92 !

jpechane:
@Naros Test failuer unrelated, applied, thanks!

PR #6620, opened 2025-07-31
DBZ-9306 Pause mining in archive log only mode when logs mined

Naros:
https://issues.redhat.com/browse/DBZ-9306

The prior logic for mining archive logs in archive log only mode did not account for all use cases that the connector could be in when an archive log was fully mined.

This change uses heuristics to determine the appropriate point when the connector should enter a wait state for a new archive log to be created.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/AbstractLogMinerStreamingChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/AbstractLogMinerStreamingChangeEventSource.java
@@ -818,7 +818,7 @@ public abstract class AbstractLogMinerStreamingChangeEventSource
                 return true;
             }
         }
-        else if (!getMetrics().getBatchMetrics().hasProcessedAnyTransactions()) {
+        else if (isNoDataProcessedInBatchAndAtEndOfArchiveLogs()) {
             if (endScn.compareTo(getMaximumArchiveLogsScn()) == 0) {
                 // Prior iteration mined up to the last entry in the archive logs and no data was returned.
                 return isArchiveLogOnlyModeAndScnIsNotAvailable(endScn.add(Scn.ONE));


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/AbstractLogMinerStreamingChangeEventSource.java
```diff
@@ -801,6 +801,17 @@ protected void handleXmlEndEvent(LogMinerEventRow event) throws InterruptedExcep
...
+    /**
+     * Waits for the range if it's not completely available in the archive logs yet.
+     *
+     * @param startScn the range starting SCN number
+     * @param endScn the range ending SCN number
+     * @return {@code true} if the connector loop should break, {@code false} otherwise
+     * @throws SQLException if a database exception occurs
+     * @throws InterruptedException if the thread is interrupted
+     */
+    protected abstract boolean waitForRangeAvailabilityInArchiveLogs(Scn startScn, Scn endScn) throws SQLException, InterruptedException;
```
vjuranek: Is it needed to make this method abstract? The implementations are identical.

Naros: They're close but the `else if` condition is slightly different between the implementations. I suppose I could abstract that aspect instead and pull the vast majority of the logic up into the abstract class.

Naros: @vjuranek can you take a look at the latest commit, does that look better to you?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/AbstractLogMinerStreamingChangeEventSource.java
```diff
@@ -801,6 +801,40 @@ protected void handleXmlEndEvent(LogMinerEventRow event) throws InterruptedExcep
...
+     */
+    protected boolean waitForRangeAvailabilityInArchiveLogs(Scn startScn, Scn endScn) throws SQLException, InterruptedException {
+        if (endScn.isNull()) {
+            // There was no prior iteration yet, sanity check to verify starting SCN
+            if (isArchiveLogOnlyModeAndScnIsNotAvailable(startScn)) {
+                LOGGER.error("Could not find the start SCN {} in the archive logs, stopping connector.", startScn);
+                return true;
+            }
+        }
+        else if (!getMetrics().getBatchMetrics().hasProcessedAnyTransactions()) {
```
vjuranek:
```suggestion
        else if (isNoDataProcessedInBatchAndAtEndOfArchiveLogs()) {
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/AbstractLogMinerStreamingChangeEventSource.java
vjuranek: Indeed, sorry, I didn't noticed the difference:-( Anyway, this looks to me better now:-)

vjuranek:
Let's wait for CI, but LGTM now

Naros:
The XStream test failures are unrelated.

PR #6621, opened 2025-08-01
DBZ-9310 Publication timeouts

Kalla-Shubham:


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
```diff
@@ -176,7 +178,16 @@ protected void initPublication() {
                                     LOGGER.info("Creating Publication with statement '{}'", createPublicationStmt);
                                     // Publication doesn't exist, create it.
                                     if (!isOnlyRead) {
-                                        stmt.execute(createPublicationStmt);
+                                        try {
+                                            stmt.setQueryTimeout(toIntExact(connectorConfig.createSlotCommandTimeout()));
+                                            stmt.execute(createPublicationStmt);
+                                        } catch (SQLException ex) {
+                                            if ("57014".equals(ex.getSQLState())) {
```
jpechane: @Kalla-Shubham Could you please extract the state code as a consant?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
```diff
@@ -176,7 +178,16 @@ protected void initPublication() {
                                     LOGGER.info("Creating Publication with statement '{}'", createPublicationStmt);
                                     // Publication doesn't exist, create it.
                                     if (!isOnlyRead) {
-                                        stmt.execute(createPublicationStmt);
+                                        try {
+                                            stmt.setQueryTimeout(toIntExact(connectorConfig.createSlotCommandTimeout()));
+                                            stmt.execute(createPublicationStmt);
```
mfvitale, line 185: If I'm not wrong the effective method called here is https://github.com/debezium/debezium/blob/31ddc2c720003636c7ccd72413f2b6f083d5bd79/debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java#L429 that seems to not use the like https://github.com/debezium/debezium/blob/31ddc2c720003636c7ccd72413f2b6f083d5bd79/debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java#L450

Maybe we can add the query timeout also to this method so that you can then appropriate configure it via https://github.com/debezium/debezium/blob/31ddc2c720003636c7ccd72413f2b6f083d5bd79/debezium-core/src/main/java/io/debezium/relational/RelationalDatabaseConnectorConfig.java#L248?

Kalla-Shubham: I don't think this execute method is called.

Kalla-Shubham: its from the java.sql library

Naros: You're right @Kalla-Shubham, but it does raise the question of whether or not we should move away from using the raw `java.sql` API here and instead use the abstraction on `JdbcConnection`, for example, methods like `query`, `queryAndMap`, or `execute` as Mario points out.

jpechane: Given that we already use direct JDBC API in the code I'd prefer to keep it as is and handle rewrite from JDBC to JDBC connection as a separate PR.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
Kalla-Shubham: I have added them as constants

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
Kalla-Shubham: Hi 
I agree we should do the rewrite in separate PR.

I have a couple of questions here :
- Should we have retries here what are your thoughts I don't think retries would be much of a help in case of VACUUM process going on it would anyway fail even after retries and the intention is to not keep waiting on the query.
- Should we keep the timeout config as https://github.com/debezium/debezium/blob/31ddc2c720003636c7ccd72413f2b6f083d5bd79/debezium-core/src/main/java/io/debezium/relational/RelationalDatabaseConnectorConfig.java#L248
or create separate one for Publication query

LMK your thoughts @jpechane

Kalla-Shubham: BUMP
@jpechane 
I have tested it timeout looks good just want the final go ahead for which config to use.

Kalla-Shubham:
@jpechane I am not getting the option to merge

jpechane:
@Kalla-Shubham Applied, thanks. Sorry the comment was not meant that you could merge, it was just the final assessment.

PR #6622, opened 2025-08-01
DBZ-8471: Support for TSVECTOR data types for postgres Sink connector.

pranavt84:
https://issues.redhat.com/browse/DBZ-8471

### DIFF
--- a/debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/e2e/AbstractJdbcSinkPipelineIT.java
+++ b/debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/e2e/AbstractJdbcSinkPipelineIT.java
@@ -2927,8 +2927,12 @@ public abstract class AbstractJdbcSinkPipelineIT extends AbstractJdbcSinkIT {
                 List.of("to_tsvector('english', 'This is a test for direct tsvector insert')"),
                 List.of("'direct':6 'insert':8 'test':4 'tsvector':7"),
                 (record) -> {
-                    String actualDataType = resolveExpectedDataType(sink);
-                    assertColumn(sink, record, "data", actualDataType);
+                    if (sink.getType().is(SinkType.POSTGRES)) {
+                        assertColumn(sink, record, "data", "tsvector");
+                    }
+                    else {
+                        assertColumn(sink, record, "data", getTextType(false));
+                    }
                 },
                 ResultSet::getString);
     }
@@ -2942,29 +2946,14 @@ public abstract class AbstractJdbcSinkPipelineIT extends AbstractJdbcSinkIT {
                 List.of("'full:3 postgre:1 search:5 support:2 text:4'"),
                 List.of("'full':3 'postgre':1 'search':5 'support':2 'text':4"),
                 (record) -> {
-                    String actualDataType = resolveExpectedDataType(sink);
-                    assertColumn(sink, record, "data", actualDataType);
+                    if (sink.getType().is(SinkType.POSTGRES)) {
+                        assertColumn(sink, record, "data", "tsvector");
+                    }
+                    else {
+                        assertColumn(sink, record, "data", getTextType(false));
+                    }
                 },
                 ResultSet::getString);
-    }
-
-    private String resolveExpectedDataType(Sink sink) {
-        SinkType sinkType = sink.getType();
-
-        switch (sinkType) {
-            case DB2:
-                return "CLOB";
-            case SQLSERVER:
-                return "varchar";
-            case MYSQL:
-                return "longtext";
-            case POSTGRES:
-                return "tsvector";
-            case ORACLE:
-                return "VARCHAR2";
-            default:
-                return "text";
-        }
     }
 
     private static List<ZonedDateTime> getExpectedZonedDateTimes(Sink sink) {


debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresDatabaseDialect.java
```diff
@@ -196,6 +196,8 @@ protected void registerTypes() {
         registerType(SparseDoubleVectorType.INSTANCE);
         registerType(FloatVectorType.INSTANCE);
         registerType(DoubleVectorType.INSTANCE);
+
+        registerType(TsvectorType.INSTANCE);
```
Naros:
```suggestion
        registerType(TsvectorType.INSTANCE);
```

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/TsvectorType.java
```diff
@@ -0,0 +1,53 @@
...
+    public static TsvectorType INSTANCE = new TsvectorType();
+
+    @Override
+    public String getTypeName(Schema schema, boolean isKey) {
+        return "tsvector";
+    }
+
+    @Override
+    public String[] getRegistrationKeys() {
+        return new String[]{ "io.debezium.data.Tsvector" };
```
Naros:
```suggestion
        return new String[]{ io.debezium.data.TsVector.LOGICAL_NAME };
```
Since the `TsVector` type was added to `debezium-core`, we can use the logical name directly.

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/e2e/JdbcSinkPipelineToPostgresIT.java
```diff
@@ -244,4 +244,28 @@ public void testHalfVectorDataType(Source source, Sink sink) throws Exception {
                 ResultSet::getString);
     }
 
+    @TestTemplate
```
Naros, line 247: Could we add another test in `AbstractJdbcSinkPipelineIT` where again it's scoped to the POSTGRES source, but we verify that when we send this to non-Postgres systems, the field is constructed as a text-based column? Since you have the Postgres sink ones here, you could annotate this new method to skip sink POSTGRES type.

Naros (changes requested):
I've left a few inline comments, and it seems the test failures are related. Could you take a look?  Thanks.

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/e2e/JdbcSinkPipelineToPostgresIT.java
pranavt84: After adding the test case, I found that different sink connectors resolve to different data types for tsvector of PostgreSQL. So, I updated the test case based on the default values returned in each case, which allowed all tests to pass successfully.

jpechane: @pranavt84 _nteresting. Do you please have this information somewhere? What database map tsvector to what? I am a bit surprised that this differes as I'd expect this to be consistent.

pranavt84: @jpechane 
I traced the type resolution process. For the tsvector type, our code resolves it to ConnectStringType for all databases except PostgreSQL, where a specific type is used.

The ConnectStringType maps to CLOB, and the actual SQL type that CLOB resolves to depends on the database dialect.

Each dialect (e.g., DB2Dialect, MySQLDialect, etc.) defines how CLOB maps to the actual SQL type via registerColumnType(...).

However, I couldn’t find any centralized documentation that lists how CLOB (or other types) maps across different dialects — these mappings are only found in the source code of each respective dialect class.

jpechane: @pranavt84 Make sense, thanks! In that case could you please extend the JavaDoc of `TsvectorType` describing this fact and add a test for one of the other databases, for example MySQL?

I believe we'll be good to go then.

pranavt84: @jpechane 
The test case is already there. Let me know if you are referring to something else.
https://github.com/debezium/debezium/pull/6622/files#:~:text=testTsvectorDataType

Added this mapping information to the doc of TsvectorType class.

Naros: Thanks @pranavt84, what you describe is expected. The logical type is only recognized as a `tsveector` for PG, and for all others, it would be seen as its native schema type of of `string`.

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/e2e/AbstractJdbcSinkPipelineIT.java
```diff
@@ -2918,6 +2918,55 @@ public void testVectorDataType(Source source, Sink sink) throws Exception {
...
+    @ForSource(value = { SourceType.POSTGRES }, reason = "The tsvector data type only applies to PostgreSQL")
+    public void testTsvectorDataType(Source source, Sink sink) throws Exception {
+        assertDataTypeNonKeyOnly(source,
+                sink,
+                "tsvector",
+                List.of("to_tsvector('english', 'This is a test for direct tsvector insert')"),
+                List.of("'direct':6 'insert':8 'test':4 'tsvector':7"),
+                (record) -> {
+                    String actualDataType = resolveExpectedDataType(sink);
+                    assertColumn(sink, record, "data", actualDataType);
```
Naros:
```suggestion
                    assertColumn(sink, record, "data", getTextType(false));
```

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/e2e/AbstractJdbcSinkPipelineIT.java
```diff
@@ -2918,6 +2918,55 @@ public void testVectorDataType(Source source, Sink sink) throws Exception {
...
+    @ForSource(value = SourceType.POSTGRES, reason = "The tsvector data type only applies to PostgreSQL")
+    public void testTsvectorDataTypeWithStaticValue(Source source, Sink sink) throws Exception {
+        assertDataTypeNonKeyOnly(source,
+                sink,
+                "tsvector",
+                List.of("'full:3 postgre:1 search:5 support:2 text:4'"),
+                List.of("'full':3 'postgre':1 'search':5 'support':2 'text':4"),
+                (record) -> {
+                    String actualDataType = resolveExpectedDataType(sink);
+                    assertColumn(sink, record, "data", actualDataType);
```
Naros:
```suggestion
                    assertColumn(sink, record, "data", getTextType(false));
```

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/e2e/AbstractJdbcSinkPipelineIT.java
```diff
@@ -2918,6 +2918,55 @@ public void testVectorDataType(Source source, Sink sink) throws Exception {
...
+            case MYSQL:
+                return "longtext";
+            case POSTGRES:
+                return "tsvector";
+            case ORACLE:
+                return "VARCHAR2";
+            default:
+                return "text";
+        }
+    }
```
Naros:
```suggestion
```

Naros (commented):
LGTM, just left one inline request, we should be able to use `getTextType` for the non-PG implementations so we can reuse existing logic without creating a new impl.

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/e2e/AbstractJdbcSinkPipelineIT.java
pranavt84: incorporated changes

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/e2e/AbstractJdbcSinkPipelineIT.java
pranavt84: incorporated changes

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/e2e/AbstractJdbcSinkPipelineIT.java
pranavt84: incorporated changes

Naros (approved):

Naros:
Applied @pranavt84, thanks!

PR #6623, opened 2025-08-03
DBZ-9322 Log all V$LOGMNR_LOGS rows on failures

Naros:
https://issues.redhat.com/browse/DBZ-9322


jpechane:
@Naros Applied, thanks

PR #6626, opened 2025-08-05
DBZ-9327 Add retries to the github calls for checking the commit messages

AlvarVG:


.github/workflows/sanity-check.yml
```diff
@@ -23,17 +23,41 @@ jobs:
           github_repository: ${{ github.repository }}
         run: |
           echo "PREFIX_COMMITS=false" >> $GITHUB_OUTPUT
-          
-          RESPONSE=$(curl --silent -X "GET" https://api.github.com/repos/$github_repository/pulls/$pull_request_number/commits)
-          
-          MESSAGE=$(echo "$RESPONSE" | jq -r '.[].commit.message // empty')
+
```
jpechane, line 26: @AlvarVG This is very nice functionality! Would it be possible to extract this into functiona and call the function with the command to be executed? The reason is that we could then copy and re-use the function elsewhere when needed?

AlvarVG: Done!

Naros (commented):
This looks great, could you also update the octocat workflow with this, too?

AlvarVG:
> This looks great, could you also update the octocat workflow with this, too?

Done!

jpechane:
@AlvarVG Applied, thanks!

PR #6629, opened 2025-08-06
DBZ-9330 Add timeout to build image jobs

AlvarVG:


jpechane:
@AlvarVG Applied, thanks!

PR #6636, opened 2025-08-11
DBZ-9329 - Memory Leak Fix for Debezium MySQL Connector

### LINKED ISSUE #1383: MySQL Connector Memory Leak and Thread Accumulation During Snapshot Interruption [DBZ-9329]
Migrated from [DBZ-9329](https://issues.redhat.com/browse/DBZ-9329)

h1. Bug report

MySQL Connector, version v3.0.8.Final and later versions [Issue still exists in latest version because it has not been addressed.]

Any MySQL connector configuration that performs initial snapshots, particularly when:
 * Large tables are being snapshotted

 * Network connectivity to Kafka broker is unstable

 * Consumer applications may disconnect during snapshot operations

 * High volume snapshot operations that may trigger queue buffering

MySQL 8.0+ (any version)
Deployment mode: Any (on-premises, cloud providers like AWS RDS, GCP Cloud SQL, Azure Database for MySQL)

When a MySQL connector task is stopped or fails during a snapshot operation:
 # The task should shut down gracefully within a reasonable time (< 30 seconds)
 # All coordinator threads should be properly cleaned up
 # Memory should be released without accumulation or leaks
 # Buffered snapshot records should be handled gracefully without causing indefinite blocking
 # Subsequent connector restarts should work normally without memory pressure

h2. What behavior do you see?
 # Memory Accumulation: Coordinator threads get stuck indefinitely when trying to flush buffered snapshot records
 # Thread Leaks: Snapshot cleanup threads remain blocked and never terminate, holding memory references
 # Indefinite Blocking: The modifyAndFlushLastRecord method blocks forever when the consumer queue is full and no consumer is available
 # Resource Exhaustion: Over time, multiple failed snapshot attempts lead to accumulated memory usage and thread pool exhaustion
 # Delayed Shutdown: Task shutdown can take extremely long or never complete when coordinator threads are stuck

Specific symptoms:
 * Connector tasks that never finish shutting down

 * Memory usage that continues to grow after task failures

 * Thread dumps showing blocked threads in ChangeEventQueue.doEnqueue() and MySqlChangeEventSourceFactory.modifyAndFlushLastRecord()

 * OutOfMemoryError in environments with repeated snapshot failures

Yes, this issue exists in Debezium v3.0.8.Final and affects all subsequent versions that don't include the proposed fix. The issue is particularly prevalent in production environments with:
 * Intermittent network connectivity

 * Large snapshot operations

 * High-availability setups where consumer applications may restart frequently

Typical log patterns showing the issue:

 
{code:java}
[Coordinator thread] DEBUG - Attempting to flush buffered record during snapshot cleanup
[Coordinator thread] WARN - Flush attempt timed out, coordinator thread may be stuck
[Task shutdown] INFO - Stopping MySQL connector task
[Task shutdown] WARN - Task shutdown taking longer than expected, coordinator threads may be blocked {code}
 

 
 # Setup: Use the standard Debezium MySQL tutorial setup with a database containing large tables
 # Configure: Set up MySQL connector with initial snapshot mode
 # Trigger: Start the connector to begin initial snapshot of large tables
 # Simulate failure: During snapshot operation, either:
 ## Stop the Kafka Connect worker abruptly
 ## Disconnect the consumer application
 ## Fill up the Kafka topic to cause backpressure
 ## Introduce network latency/disconnection to Kafka broker
 #  Observe: Monitor memory usage and thread counts during and after the failure
 # Verify: Attempt to restart the connector and observe memory accumulation over multiple failure cycles

h3. Root Cause Analysis:
 # Simple Blocking Flush: The original modifyAndFlushLastRecord method had a simple implementation:
 
 
{code:java}
   private void modifyAndFlushLastRecord(Function<SourceRecord, SourceRecord> modify) throws InterruptedException {
       queue.flushBuffer(dataChange -> new DataChangeEvent(modify.apply(dataChange.getRecord())));
       queue.disableBuffering();
   } {code}
 

 # No Queue Shutdown: The ChangeEventQueue had no mechanism to signal shutdown to waiting threads.
 # No Coordinated Shutdown: The MySqlConnectorTask.doStop() method did not coordinate with the queue to unblock waiting threads.

h1. Feature request or enhancement (Bug Fix)

This enhancement addresses critical production reliability requirements:
 # High Availability Operations: Ensures MySQL connectors can handle broker outages and consumer failures gracefully without memory leaks
 # Resource Management: Prevents memory accumulation that can lead to OutOfMemoryError in production environments
 # Operational Stability: Enables reliable connector lifecycle management with predictable shutdown behavior
 # Scalability: Supports large-scale deployments where snapshot operations are frequent and failures are expected
 # Production Resilience: Improves connector robustness in unstable network environments or high-load scenarios

 # Queue Shutdown Mechanism: Add a shutdown() method to ChangeEventQueue that signals waiting threads to unblock gracefully
 # Enhanced Snapshot Cleanup: Improve modifyAndFlushLastRecord() in MySqlChangeEventSourceFactory to handle interruptions and use the queue shutdown signal
 # Coordinated Task Shutdown: Modify MySqlConnectorTask.doStop() to trigger queue shutdown early in the shutdown process

 
h2. Testing of the proposed implementation

The proposed implementation above was tested locally, under following conditions:

{*}Database in picture{*}: A 200 row, 10MB each row data. *[DELIBERATE SELECTION OF TABLE SIZE TO REPRO THE ISSUE]*

{*}Cluster{*}: Standard

{*}Task Failing Cause{*}: Due to standard cluster {{max.request.size}} is {{{}8388698{}}}, so task was failing with
{code:java}
WorkerSourceTask{id=lcc-devckv61v2-0} Task threw an uncaught and unrecoverable exception. Task is being killed and will not recover until manually restartedorg.apache.kafka.connect.errors.ConnectException: Unrecoverable exception from producer send callback	at org.apache.kafka.connect.runtime.WorkerSourceTask.maybeThrowProducerSendException(WorkerSourceTask.java:343)	at org.apache.kafka.connect.runtime.WorkerSourceTask.prepareToSendRecord(WorkerSourceTask.java:135)	at org.apache.kafka.connect.runtime.AbstractWorkerSourceTask.sendRecords(AbstractWorkerSourceTask.java:445)	at org.apache.kafka.connect.runtime.AbstractWorkerSourceTask.execute(AbstractWorkerSourceTask.java:391)	at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:272)	at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:328)	at org.apache.kafka.connect.runtime.AbstractWorkerSourceTask.run(AbstractWorkerSourceTask.java:88)	at org.apache.kafka.connect.runtime.isolation.Plugins.lambda$withClassLoader$7(Plugins.java:341)	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:572)	at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:317)	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1144)	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:642)	at java.base/java.lang.Thread.run(Thread.java:1583)Caused by: org.apache.kafka.common.errors.RecordTooLargeException: The message is 13981496 bytes when serialized which is larger than 8388698, which is the value of the max.request.size configuration. {code}
 

 

*Timeline of events*

!image-2025-08-05-22-53-39-338.png|width=482,height=179!

00:20 - 0:28 → Connector configured running on existing v3.0.8.Final image

00:28 → Connector deleted

00:34 → Connect pod restarted with new image (one with above changes on top of 3.0.8)

00:35 → New Connector configured with same config

 

*Observation*

On older image, at each task failure, older memory was not getting free, as we can see a step pattern in first half of the above image.

On the newer image (our fix), at each task failure, older memory was getting free, as we observe a near flatline.

Restarts in both instances of connectors were more or less same as observed from logs.

yashisrivastava-confluent:
Fixes debezium/dbz#1383

Fix memory leak in binlog-based connectors during snapshot interruptions

This PR resolves coordinator thread memory leaks that occur when snapshot operations are interrupted, causing threads to remain blocked indefinitely in ChangeEventQueue operations.

**Key Changes:**
- Added thread-safe shutdown mechanism to ChangeEventQueue with graceful interruption
- Created BinlogSourceTask and BinlogChangeEventSourceFactory base classes for code reuse
- Updated MySQL and MariaDB connectors to use centralized shutdown logic
- Added comprehensive test coverage (10 test scenarios) for shutdown edge cases

**Impact:** Prevents resource exhaustion in production deployments by ensuring coordinator threads can exit cleanly when connectors are stopped during snapshot operations.


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorTask.java
```diff
@@ -274,6 +274,12 @@ protected Optional<ErrorHandler> getErrorHandler() {
 
     @Override
     protected void doStop() {
+        // Shut down the queue early to prevent memory leaks from blocked threads
+        if (queue != null) {
+            LOGGER.info("Shutting down queue to prevent thread leaks");
+            queue.shutdown();
+        }
```
Naros: Would this also be necessary for MariaDB? If so, it may be useful to create a `doStop()` implementation in the `BinlogSourceTask` class and place this logic there, having both MySQL and MariaDB call their respective super type methods.  What do you think?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlChangeEventSourceFactory.java
```diff
@@ -85,8 +89,35 @@ private void preSnapshot() {
     }
 
     private void modifyAndFlushLastRecord(Function<SourceRecord, SourceRecord> modify) throws InterruptedException {
```
Naros: This likely applies to MariaDB. I might suggest we create a `BinlogChangeEventSourceFactory` abstract base class that both MySQL and MariaDB extend from, and put this method in the abstract base class.

Naros (changes requested):
I've left a few questions around refactoring and applicability.

yashisrivastava-confluent:
@Naros I have made the fixes recommended by you, can I get a re-review on the same?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogSourceTask.java
```diff
@@ -94,4 +96,16 @@ public void onError(SQLException exception) throws RuntimeException {
...
+
+    @Override
+    protected void doStop() {
+        // Get the queue field from the subclass to shut it down early
+        ChangeEventQueue<DataChangeEvent> queue = getChangeEventQueue();
+        if (queue != null) {
+            LOGGER.info("Shutting down queue to prevent thread leaks");
+            queue.shutdown();
+        }
+    }
```
Naros: I understand that MariaDB and MySQL are the only ones that toggle buffering in the `ChangeEventQueue`, but I wonder if we shouldn't think about moving this up one level to the `BaseSourceTask`?  @jpechane do you have any specific preference?

jpechane:
Could you please remove merge commit and do just rebase to keep linear history? Thanks

Naros (commented):
Overall, this LGTM. 

The only question I have, and I am on the fence about it, is whether or not we should be setting up every connector to return its _queue_ in the `getChangeEventQueue`. Do you have any preference @jpechane ?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogChangeEventSourceFactory.java
```diff
@@ -0,0 +1,73 @@
...
+            // Queue was shut down or thread was interrupted - this is expected during task shutdown
+            LOGGER.info("Buffered record flush interrupted during snapshot cleanup, likely due to task shutdown");
+            throw e;
+        }
+        finally {
+            // Always disable buffering to prevent memory leaks
+            try {
+                queue.disableBuffering();
+            }
+            catch (AssertionError e) {
```
jpechane: I definitely don't like this part of the code. If it is acceptable to have something when buffering is disabled in the buffer then assertion should not fail. Maybe and extra method needs to be added to reflect that fact?

jpechane:
> Overall, this LGTM.
> 
> The only question I have, and I am on the fence about it, is whether or not we should be setting up every connector to return its _queue_ in the `getChangeEventQueue`. Do you have any preference @jpechane ?

Do we have any connector that does not use queue? I don;t think so. If it is the case then I believe the current implementation is OK and there should be a follow-up PR refactoring that would pull queue up into `BaseSourceTask`.

debezium-core/src/main/java/io/debezium/connector/base/ChangeEventQueue.java
```diff
@@ -241,15 +244,19 @@ protected void doEnqueue(T record) throws InterruptedException {
...
                 // signal poll() to drain queue
                 this.isFull.signalAll();
                 // queue size or queue sizeInBytes threshold reached, so wait a bit
                 this.isNotFull.await(pollInterval.toMillis(), TimeUnit.MILLISECONDS);
             }
 
-            queue.enqueue(record);
-            // If we pass a positiveLong max.queue.size.in.bytes to enable handling queue size in bytes feature
+            // If not running, throw InterruptedException to allow graceful exit
+            if (!running) {
```
jpechane, line 279: Is `volatile` enough here? What if `running` flag is change between this check and the next `enqueue` call?

debezium-core/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -507,6 +516,18 @@ private void stop(boolean restart) {
...
 
+    /**
+     * Override this method to provide access to the connector's ChangeEventQueue.
+     * If the connector uses a ChangeEventQueue, this should return it to enable
+     * automatic queue shutdown during connector stop to prevent thread leaks.
+     * If the connector doesn't use a ChangeEventQueue, return null.
+     *
+     * @return the ChangeEventQueue instance or null if not applicable
+     */
+    protected ChangeEventQueue<DataChangeEvent> getChangeEventQueue() {
```
jpechane: I'd prefer `Optional` here if possible.

jpechane (commented):
@yashisrivastava-confluent Thanks for the PR. It definitely goes into right direction. Could you please take a look at my comments? Thanks

yashisrivastava-confluent:
@jpechane I have made requested changes, please re-review

debezium-core/src/main/java/io/debezium/connector/base/ChangeEventQueue.java
jpechane: I think this comment was not addressed.

support/checkstyle/src/main/java/io/debezium/checkstyle/ExcludeTestPackages.java
```diff
@@ -11,9 +11,9 @@
 import com.puppycrawl.tools.checkstyle.api.Filter;
 
 /**
- * Allows disabling some rules for the test suite source. Any violation will be suppressed if it's generated by a source file
- * having {@code /src/test/java} included in its path and if the violation message contains the keyword
- * "[not required for tests]".
+ * Allows disabling some rules for the test suite source. Any violation will be suppressed if it's generated by a source
```
jpechane, line 14: Please don't include change sin `support` module.

debezium-core/src/main/java/io/debezium/connector/base/ChangeEventQueue.java
```diff
@@ -212,20 +213,44 @@ record = bufferedEvent.getAndSet(record);
...
+     *
+     * @return true if buffering was disabled successfully, false if buffer was not empty
+     */
+    public boolean disableBufferingSafely() {
+        if (bufferedEvent.get() != null) {
+            LOGGER.warn("Disabling buffering with non-empty buffer - this may indicate incomplete flush during shutdown");
+            // Clear the buffer to prevent memory leaks
+            bufferedEvent.set(null);
+        }
+        buffering = false;
```
jpechane, line 250: I think this should go to the top of the method. Otherwise you can celan the buffer but another thread can fill it again before buffering is disabled.

Naros:

Hi @yashisrivastava-confluent, please sign off all commits with:

```
git commit -s
```

Naros:
Hi @yashisrivastava-confluent are you in a position to rebase this PR, add commit sign-offs and address Jiri's remaining open questions?

PR #6637, opened 2025-08-11
DBZ-9342 Use mockito only in testcontainers module

vjuranek:
https://issues.redhat.com/browse/DBZ-9342


vjuranek:
AFAICT `debezium-testing/debezium-testing-system` module doesn't use Mocito at all, so no need to set it there.

AlvarVG:
Merging the PR as the failures does not seem to be related. Thank you very much @vjuranek

PR #6640, opened 2025-08-11
DBZ-9345 Use equality check for improved performance

Naros:
https://issues.redhat.com/browse/DBZ-9345

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/UnchangedToastedReplicationMessageColumn.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/UnchangedToastedReplicationMessageColumn.java
@@ -56,7 +56,11 @@ public class UnchangedToastedReplicationMessageColumn extends AbstractReplicatio
     }
 
     public static boolean isUnchangedToastedValue(Object value) {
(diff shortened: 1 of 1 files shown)


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/UnchangedToastedReplicationMessageColumn.java
```diff
@@ -56,7 +56,11 @@ public boolean isToastedColumn() {
     }
 
     public static boolean isUnchangedToastedValue(Object value) {
-        return UNCHANGED_TOAST_VALUES.contains(value);
+        // Using Set#contains triggers comparison by the value's hash. For some types, like a Java String, the
```
jpechane, line 55: @Naros What about turning `UNCHANGED_TOAST_VALUES` into `Object[]` array and just use raw iteration for best performance?

Naros: @jpechane done.

jpechane:
@Naros Applied, thanks

PR #6641, opened 2025-08-11
DBZ-9300 Add reflections dependency into Oracle module

vjuranek:


Naros (approved):
@vjuranek I'd say as long as the Oracle module compiles, no need to run the tests if you've verified this locally.  I can then rebase main on the other PR where you saw this.

PR #6645, opened 2025-08-12
DBZ-9351 Add section about adapter modes

Naros:
https://issues.redhat.com/browse/DBZ-9351

@roldanbob, as discussed, could you review this? It made sense to me to put this at the top and then for the XStream and OpenLogReplicator bits, guide users to those sections as needed. In the long term, there may be a better navigational flow we can consider, but that's secondary imo.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/oracle.adoc
+++ b/documentation/modules/ROOT/pages/connectors/oracle.adoc
@@ -72,6 +72,31 @@ Information and procedures for using a {prodname} Oracle connector are organized
   * xref:debezium-oracle-connector-frequently-asked-questions[]
 endif::product[]
 
+// Type: assembly
+// ModuleID: how-debezium-oracle-connectors-work
+// Title: How {prodname} Oracle connectors work
+[[how-the-oracle-connector-works]]
+== How the Oracle connector works
+
+To optimally configure and run a {prodname} Oracle connector, it is helpful to understand how the connector works.
+
+ifdef::product[]
+For more information, see the following topics:
+
+* xref:descriptions-of-debezium-oracle-connector-adapters[]
+* xref:how-debezium-oracle-connectors-perform-database-snapshots[]
+* xref:debezium-oracle-ad-hoc-snapshots[]
+* xref:debezium-oracle-incremental-snapshots[]
+* xref:default-names-of-kafka-topics-that-receive-debezium-oracle-change-event-records[]
+* xref:how-debezium-oracle-connectors-expose-database-schema-changes[]
+* xref:debezium-oracle-connector-generated-events-that-represent-transaction-boundaries[]
+* xref:how-the-debezium-oracle-connector-uses-event-buffering[]
+
+endif::product[]
+
+// Type: concept
+// ModuleID: descriptions-of-debezium-oracle-connector-adapters
+// Title: Descriptions of {prodname} Oracle connector adapters
 [[oracle-adapter-modes]]
 == Adapter Modes
 
@@ -143,27 +168,6 @@ The {prodname} Oracle connector can also use Oracle XStream, a commercial compon
 +
 Enable Oracle XStream by setting the xref:oracle-property-database-connection-adapter[`database.connection.adapter`] property to `xstream`.
 For more details, see the xref:oracle-xstreams-support[XStream] section.
-w
-// Type: assembly
-// ModuleID: how-debezium-oracle-connectors-work
-// Title: How {prodname} Oracle connectors work
-[[how-the-oracle-connector-works]]
-== How the Oracle connector works
-
-To optimally configure and run a {prodname} Oracle connector, it is helpful to understand how the connector performs snapshots, streams change events, determines Kafka topic names, uses metadata, and implements event buffering.
-
-ifdef::product[]
-For more information, see the following topics:
-
-* xref:how-debezium-oracle-connectors-perform-database-snapshots[]
-* xref:debezium-oracle-ad-hoc-snapshots[]
-* xref:debezium-oracle-incremental-snapshots[]
-* xref:default-names-of-kafka-topics-that-receive-debezium-oracle-change-event-records[]
-* xref:how-debezium-oracle-connectors-expose-database-schema-changes[]
-* xref:debezium-oracle-connector-generated-events-that-represent-transaction-boundaries[]
-* xref:how-the-debezium-oracle-connector-uses-event-buffering[]
-
-endif::product[]
 
 // Type: concept
 // Title: How {prodname} Oracle connectors perform database snapshots


documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,63 @@ Information and procedures for using a {prodname} Oracle connector are organized
   * xref:debezium-oracle-connector-frequently-asked-questions[]
 endif::product[]
 
+[[oracle-adapter-modes]]
+== Adapter Modes
+
+The {prodname} Oracle connector supports multiple _adapters_ for capturing changes from the Oracle database’s transaction logs.
```
```suggestion
The {prodname} Oracle connector supports multiple adapters for capturing changes from the Oracle database transaction logs.
```
Per Oracle TM guidelines, _Oracle trademarks are adjectives and should not be used as nouns, or in the possessive or plural form. For example, “Oracle database’s benefits...” not “Oracle’s benefits...”_

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,63 @@ Information and procedures for using a {prodname} Oracle connector are organized
...
+You can configure the adapter by setting the xref:oracle-property-database-connection-adapter[`database.connection.adapter`] configuration property.
```
```suggestion
You can configure the connector to use a specific adapter by setting the xref:oracle-property-database-connection-adapter[`database.connection.adapter`] configuration property.
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,63 @@ Information and procedures for using a {prodname} Oracle connector are organized
...
+
+=== LogMiner
```
```suggestion
The connector supports the following adapters:
 
LogMiner::
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,63 @@ Information and procedures for using a {prodname} Oracle connector are organized
...
+
+By default, the {prodname} Oracle connector uses the native Oracle LogMiner API to read and stream changes from database transaction logs.
+LogMiner can operate in two distinct modes, both supported by {prodname}.
```
```suggestion
You can configure {prodname} to use LogMiner in either of the following distinct modes:
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,63 @@ Information and procedures for using a {prodname} Oracle connector are organized
...
+
+==== Uncommitted changes mode
```
```suggestion
Uncommitted changes mode:::
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,63 @@ Information and procedures for using a {prodname} Oracle connector are organized
...
+Because Oracle interleaves operations from multiple transactions, the connector must buffer events until a corresponding commit or rollback is detected.
+
+As this is the connector’s default mode, ensure that the connector runtime is allocated sufficient memory to handle your largest transactions and highest concurrency levels across all captured tables.
+
+This mode minimizes load on the source database by delegating most processing to the connector.
+Oracle LogMiner is responsible only for reading redo and archive logs from disk and streaming the raw changes.
+
+Enable this mode by setting the xref:oracle-property-database-connection-adapter[`database.connection.adapter`] property to `logminer` (default).
+
+==== Committed changes mode
```
```suggestion
LogMiner committed changes mode:::
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
roldanbob: Per new DITA migration requirements, nested headings (level 3 or higher) are not permitted. Guidelines recommend converting these to description lists. 
TBH, it's still unclear how the combination of the migration process and Nebel tooling will process these -- we have a lot of them sprinkled across the docs. My goal for now is to avoid introducing more instances.

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,63 @@ Information and procedures for using a {prodname} Oracle connector are organized
...
+
+In uncommitted changes mode, the {prodname} Oracle connector receives a continuous stream of change events for both committed and in-flight uncommitted transactions.
+Because Oracle interleaves operations from multiple transactions, the connector must buffer events until a corresponding commit or rollback is detected.
+
+As this is the connector’s default mode, ensure that the connector runtime is allocated sufficient memory to handle your largest transactions and highest concurrency levels across all captured tables.
+
+This mode minimizes load on the source database by delegating most processing to the connector.
+Oracle LogMiner is responsible only for reading redo and archive logs from disk and streaming the raw changes.
+
+Enable this mode by setting the xref:oracle-property-database-connection-adapter[`database.connection.adapter`] property to `logminer` (default).
```
```suggestion
In uncommitted changes mode, the {prodname} Oracle connector receives a continuous stream of change events for both committed and in-flight uncommitted transactions.
Because Oracle interleaves operations from multiple transactions, the connector must buffer events until a corresponding commit or rollback is detected.
+
Because this the default mode for the connector, it's important to allocate enough memory to the connector to handle the largest transactions and highest concurrency levels across all captured tables.
+
Uncommitted changes mode minimizes the load on the source database by delegating most processing to the connector.
Oracle LogMiner is responsible only for reading redo and archive logs from disk and streaming the raw changes.
+
Enable this mode by setting the xref:oracle-property-database-connection-adapter[`database.connection.adapter`] property to `logminer` (default).
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,63 @@ Information and procedures for using a {prodname} Oracle connector are organized
...
+
+This mode is recommended only if you can guarantee that all transactions fit within the database’s PGA memory space.
+Otherwise, use the uncommitted changes mode.
+
+Enable this mode by setting the xref:oracle-property-database-connection-adapter[`database.connection.adapter`] property to `logminer_unbuffered`.
+
+[NOTE]
+====
+This feature is currently *incubating* and might change in future releases.
+====
```
```suggestion
In committed changes mode, the {prodname} Oracle connector instructs Oracle LogMiner to stream only committed changes.
Because the connector receives only committed events, it can immediately forward them to destination topics without maintaining an internal transaction buffer.
+
This mode requires significantly less connector memory, but shifts a much greater processing and memory burden to the source database.
For large transactions, this can easily exceed the database’s available PGA memory, potentially leading to `PGA_AGGREGATE_LIMIT` errors.
+
Configure this mode only if all transactions in your environment are smaller than the PGA memory that is configured for the database.
Otherwise, use the uncommitted changes mode.
+
Enable this mode by setting the xref:oracle-property-database-connection-adapter[`database.connection.adapter`] property to `logminer_unbuffered`.
ifdef::community[]
+
[NOTE]
====
This feature is currently *incubating* and might change in future releases.
====
endif::community[]
```
For productization, we should not use the `incubating` terminology or refer to potential future changes.
If the feature is in any way experimental, we should mark it as DP or TP.
If we want to commit to marking this feature as DP or TP, I'll add a product conditional and apply the appropriate conditional.

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,63 @@ Information and procedures for using a {prodname} Oracle connector are organized
...
+
+ifdef::community[]
+=== OpenLogReplicator
```
```suggestion
OpenLogReplicator:::
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,63 @@ Information and procedures for using a {prodname} Oracle connector are organized
...
+
+The {prodname} Oracle connector can be configured to use OpenLogReplicator, an open-source application that reads Oracle changes directly from redo and archive logs with minimal impact on the database.
```
```suggestion
You can configure the {prodname} Oracle connector to use OpenLogReplicator, an open-source application that reads Oracle changes directly from the redo and archive logs with minimal impact on the database.
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,63 @@ Information and procedures for using a {prodname} Oracle connector are organized
...
+
+Enable OpenLogReplicator by setting the xref:oracle-property-database-connection-adapter[`database.connection.adapter`] property to `olr`.
+For more information, see the xref:oracle-openlogreplicator-support[] section.
+endif::community[]
```
```suggestion
+
Enable OpenLogReplicator by setting the xref:oracle-property-database-connection-adapter[`database.connection.adapter`] property to `olr`.
For more information, see the xref:oracle-openlogreplicator-support[] section.
endif::community[]
```
I'm going to hold off on suggesting any change for the in-line link, because we use them so frequently that if we really need to eliminate them, that's going to be part of a much broader effort. But we should be conscious of how and when we use them, and consider alternate constructions. The recommended workaround is to add all such links to an _Additional resources_ section. Stay tuned.

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,63 @@ Information and procedures for using a {prodname} Oracle connector are organized
...
+
+=== Oracle XStream
```
```suggestion
Oracle XStream:::
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,63 @@ Information and procedures for using a {prodname} Oracle connector are organized
...
+
+The {prodname} Oracle connector can also use Oracle XStream, a commercial component of Oracle GoldenGate.
+
+Enable Oracle XStream by setting the xref:oracle-property-database-connection-adapter[`database.connection.adapter`] property to `xstream`.
+For more details, see the xref:oracle-xstreams-support[] section.
```
```suggestion
+
Enable Oracle XStream by setting the xref:oracle-property-database-connection-adapter[`database.connection.adapter`] property to `xstream`.
For more details, see the xref:oracle-xstreams-support[XStream] section.
```
To ensure that link text renders in the product docs, which are associated with the alternate, ModuleID anchor, it's necessary to hard code link text here, even if it's going to differ slightly between the community and product versions (via the Nebel _Title_ annotation). In sections that are conditionalized for product use only, we can safely omit the implied link text.

documentation/modules/ROOT/pages/connectors/oracle.adoc
Naros: I chose to use `Committed changes mode:::` here to be consistent with the `Uncommitted changes mode:::` and not to introduce unnecessary noise since this is under the `LogMiner` section already. Let me know if you disagree.

documentation/modules/ROOT/pages/connectors/oracle.adoc
Naros: The unbuffered (committed changes mode) is TP. Once I have your suggestions applied, if you'd like to review and suggest further changes, let me know.

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,67 @@ Information and procedures for using a {prodname} Oracle connector are organized
...
++
+Enable this mode by setting the xref:oracle-property-database-connection-adapter[`database.connection.adapter`] property to `logminer_unbuffered`.
+ifdef::community[]
++
+[NOTE]
+====
+This feature is currently *incubating* and might change in future releases.
+====
+endif::community[]
+
```
roldanbob, line 155:
```suggestion
ifdef::product[]
[IMPORTANT]
====
The Committed changes mode setting is a Technology Preview feature.
Technology Preview features are not supported with Red Hat production service level agreements (SLAs) and might not be functionally complete.
Red Hat does not recommend using them in production.
These features provide early access to upcoming product features, enabling customers to test functionality and provide feedback during the development process.
For more information about the support scope of Red Hat Technology Preview features, see link:https://access.redhat.com/support/offerings/techpreview/[Technology Preview Features Support Scope].
====
endif::product[]
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -72,6 +72,67 @@ Information and procedures for using a {prodname} Oracle connector are organized
   * xref:debezium-oracle-connector-frequently-asked-questions[]
 endif::product[]
 
+[[oracle-adapter-modes]]
```
roldanbob, line 100: Should we insert this section at the end of the Overview (which is where it now sits), or as the first topic within the How the connector works assembly? 
I'm slightly partial to the latter option, if only because, in the product version, we could add an entry for it in the preceding list of topics. 
```suggestion
// Type: concept
// ModuleID: descriptions-of-debezium-oracle-connector-adapters
// Title: Descriptions of {prodname} Oracle connector adapters
[[oracle-adapter-modes]]
```

roldanbob (changes requested):
Added the TP admonition (L119) and Nebel annotations on LL75-77

After we finalize placement, I think this is good to go.

documentation/modules/ROOT/pages/connectors/oracle.adoc
Naros: I'm okay with that, but the reason I didn't initially was how the initial text in that section was written. We've added several subsections, and I think that initial text might need some adjustments.

We should also reconsider a long-overdue overhaul to the subsections in this part of the documentation, likely across all connectors. In the Oracle docs, for example, the initial flow was reasonable with snapshotting, schema changes, etc; however, as we've added new sections, the flow is a bit disjointed.

Would you agree that introducing one additional layer here might be worth it, grouping multiple snapshot topics under "Snapshotting", and we could do similar things with a few others. We could introduce a top-level section about "Mining Query Tuning" with subsections that discuss various configuration knobs we have, how they can be used, etc. This would go a long way to provide quick references to users.

The idea is to tailor "How it works" as an in-depth guide stepping across high-level topics that users should know about and understand even before considering deployment. If we bring up pertinent parts around set up, deployment, or even data types, we can always jump link to those sections like we do for configuration properties.

But as I said, this is a long-term goal; certainly not something in this PR. For now, I think if we can just tweak the first text in the "How it works" a tad, that should suffice for now.

Thoughts?

roldanbob: What you're saying makes sense. Let's talk about an overhaul -- it's certainly needed. I agree that's out of scope for this PR. Re: tweaking the first text, do you have something in mind?

roldanbob: I'm thinking something like:

> You can configure the connector to optimize how it performs snapshots, ingests and streams change events, determines Kafka topic names, uses metadata, and implements event buffering.

Naros: Nothing specific, I feel like we touch on some, but not all, the sections in this area. Should we consider making it more generic rather than more inclusive of the subsections? wdyt?

> To optimally configure and run a Debezium Oracle connector, it is helpful to understand how the connector performs snapshots, streams change events, determines Kafka topic names, uses metadata, and implements event buffering.

roldanbob: So, maybe just leave it at ...

> ... it is helpful to understand how the connector works. 

??

Naros:
@roldanbob I believe I captured all the changes as suggested, if you could give it one last review.  Thanks!

roldanbob (approved):
LGTM, @Naros!

PR #6650, opened 2025-08-13
DBZ-9214 Remove Confluent dependencies

AlvarVG:


debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/DebeziumContainer.java
```diff
@@ -55,7 +54,7 @@ public class DebeziumContainer extends GenericContainer<DebeziumContainer> {
 
     private static final int KAFKA_CONNECT_PORT = 8083;
     private static final Integer DEFAULT_JMX_PORT = 13333;
-    private static final Duration DEBEZIUM_CONTAINER_STARTUP_TIMEOUT = Duration.ofSeconds(waitTimeForRecords() * 30);
+    private static final Duration DEBEZIUM_CONTAINER_STARTUP_TIMEOUT = Duration.ofMinutes(waitTimeForRecords() * 2L);
```
vjuranek, line 57: `records.waittime` is in secods, so using it as minutes may results into unwanted consequences (very long delay if `records.waittime` is set to some higher value and things goes wrong), so I'd keep it as seconds.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/DebeziumContainer.java
```diff
@@ -198,6 +197,7 @@ public void registerConnector(String name, ConnectorConfiguration configuration)
         // To avoid a 409 error code meanwhile connector is being configured.
         // This is just a guard, probably in most use cases you won't need that as preparation time of the test might be enough to configure connector.
         Awaitility.await()
+                .pollInterval(Duration.ofSeconds(5))
```
Naros, line 200: Is there any harm in using a 1 second interval rather than 5?

debezium-testing/debezium-testing-testcontainers/src/test/java/io/debezium/testing/testcontainers/DebeziumContainerIT.java
```diff
@@ -70,19 +67,27 @@ public static void startContainers() {
...
+        final String connectorName = "my-connector-1";
+        final String connectorState = "RUNNING";
+        debeziumContainer.registerConnector(connectorName, getConfiguration(1));
 
         // task initialization happens asynchronously, so we might have to retry until the task is RUNNING
         Awaitility.await()
-                .pollInterval(Duration.ofMillis(250))
-                .atMost(Duration.ofSeconds(30))
+                .pollInterval(Duration.ofSeconds(5))
+                .atMost(Duration.ofSeconds(60))
```
Naros, line 77: Is there a difference in container performance that justified doubling the wait time? We shouldn't be spinning these up on each `@Test`, so the cost here should be negligible, but I'm just curious.

PR #6652, opened 2025-08-13
Fix typo in example configuration

Naros:


Naros:
Applied, thanks @Marci0817 !

PR #6656, opened 2025-08-14
DBZ-9356 Fix comparing ByteBuffer with byte array

spsheni:
The PostProcessor, ReselectColumnsPostProcessor, bug that otherwise occurs when the source table contains a BINARY type column is fixed by the modifications below. With these modifications, HeapByteBuffer and byte array objects will be accurately compared.

### DIFF
--- a/debezium-core/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
+++ b/debezium-core/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
@@ -42,6 +42,7 @@ import io.debezium.relational.Table;
 import io.debezium.relational.TableId;
 import io.debezium.relational.ValueConverter;
 import io.debezium.relational.ValueConverterProvider;
+import io.debezium.util.ByteBuffers;
 import io.debezium.util.Strings;
 
 /**
@@ -318,7 +319,13 @@ public class ReselectColumnsPostProcessor implements PostProcessor, BeanRegistry
     private boolean isUnavailableValueHolder(Schema schema, Object value) {
         switch (schema.type()) {
             case BYTES:
-                return unavailableValuePlaceholderBytes.equals(value);
+                if (value instanceof byte[] valueArray) {
+                    return ByteBuffers.equals(unavailableValuePlaceholderBytes, valueArray);
+                }
+                else if (value instanceof ByteBuffer valueBuffer) {
+                    return unavailableValuePlaceholderBytes.equals(valueBuffer);
+                }
+                return false;
             case MAP:
                 return unavailableValuePlaceholderMap.equals(value);
             case STRING:
--- a/debezium-core/src/main/java/io/debezium/util/ByteBuffers.java
+++ b/debezium-core/src/main/java/io/debezium/util/ByteBuffers.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.util;
+
+import java.nio.ByteBuffer;
+import java.util.Arrays;
+
+import io.debezium.annotation.ThreadSafe;
+
+/**
+ * ByteBuffers-related utility methods.
+ *
+ * @author Shyama Praveena S
+ */
+@ThreadSafe
+public final class ByteBuffers {
+
+    /**
+     * Compares ByteBuffer with byte array.
+     *
+     * @param buffer the buffer to compare
+     * @param array the byte array to compare
+     * @return true if the content of buffer matches the array, false otherwise
+     */
+    public static boolean equals(ByteBuffer buffer, byte[] array) {
+        // Null checks
+        if (buffer == null && array == null) {
+            return true;
+        }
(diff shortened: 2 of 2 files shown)


spsheni:
I invite you for the review @jcechace . Please do the needful.

debezium-core/src/main/java/io/debezium/util/ByteBuffers.java
```diff
@@ -0,0 +1,66 @@
...
+     * @param buffer the buffer to compare
+     * @param array the byte array to compare
+     * @return true if the content of buffer matches the array, false otherwise
+     */
+    public static boolean equals(ByteBuffer buffer, byte[] array) {
+        // Null checks
+        if (buffer == null && array == null)
+            return true;
+        if (buffer == null || array == null)
+            return false;
```
Naros:
```suggestion
        if (buffer == null || array == null {
            return buffer == array;
        }
```

debezium-core/src/main/java/io/debezium/util/ByteBuffers.java
```diff
@@ -0,0 +1,66 @@
...
+
+        // Length check
+        if (buffer.remaining() != array.length)
+            return false;
```
Naros:
```suggestion
        if (buffer.remaining() != array.length) {
            return false;
        }
```
We should use `{` and `}` even for oneliners to be consistent with the rest of the code-base.

debezium-core/src/main/java/io/debezium/util/ByteBuffers.java
```diff
@@ -0,0 +1,66 @@
...
+            for (int i = 0; i < array.length; i++) {
+                if (buffer.get() != array[i]) {
+                    return false;
+                }
+            }
+            return true;
+        }
+        finally {
+            buffer.position(originalPosition);
+        }
```
Naros:
```suggestion
        final int originalPosition = buffer.position();
        for (int i = 0; i < array.length; i++) {
            if (buffer.get(originalPosition + i) != array[i]) {
                return false;
            }
        }
        return true;
```
This avoids the mutation of position within the buffer.

debezium-core/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
```diff
@@ -318,7 +319,15 @@ private boolean isUnavailableValueHolder(org.apache.kafka.connect.data.Field fie
     private boolean isUnavailableValueHolder(Schema schema, Object value) {
         switch (schema.type()) {
             case BYTES:
-                return unavailableValuePlaceholderBytes.equals(value);
+                if (value instanceof byte[]) {
+                    byte[] valueArray = (byte[]) value;
+                    return ByteBuffers.equals(unavailableValuePlaceholderBytes, valueArray);
+                }
```
Naros:
```suggestion
                if (value instanceof byte[] valueArray) {
                    return ByteBuffers.equals(unavailableValuePlaceholderBytes, valueArray);
                }
```
Let's favor pattern matching since we're Java17+ compliant.

debezium-core/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
```diff
@@ -318,7 +319,15 @@ private boolean isUnavailableValueHolder(org.apache.kafka.connect.data.Field fie
...
+                else if (value instanceof ByteBuffer) {
+                    ByteBuffer valueBuffer = (ByteBuffer) value;
+                    return unavailableValuePlaceholderBytes.equals(valueBuffer);
+                }
```
Naros:
```suggestion
                else if (value instanceof ByteBuffer valueBuffer) {
                    return unavailableValuePlaceholderBytes.equals(valueBuffer);
                }
```
Let's favor pattern matching since we're Java17+ compliant.

debezium-core/src/main/java/io/debezium/util/ByteBuffers.java
spsheni: we can't compare `buffer == array`,

15:32:57,510 [ERROR] /Users/shyama/personal/project/debezium/debezium-core/src/main/java/io/debezium/util/ByteBuffers.java:[31,27] incomparable types: java.nio.ByteBuffer and byte[]

spsheni: so will keep the above code only, with added code formatting.

spsheni:
@Naros , Thank you for the review. I have updated PR with the changes.

Naros (approved):
This looks much better. Now let's wait for CI.

vjuranek:
Merged, thanks @spsheni !

spsheni:
@Naros , @vjuranek ,
Can you please confirm above fix is marked to 3.2.2.Final version?
Asking because, above pull request (PR) is raised to main branch only. Do I need to create separate PR for taking the fix to 3.2.2 ? Please let know if any action needed from my side.

jpechane:
@spsheni It is selectivelly applied to 3.2 branch, please see https://github.com/debezium/debezium/commits/3.2/

spsheni:
@jpechane , cool. Thank you.

PR #6663, opened 2025-08-18
DBZ-9364 SQL Server Heartbeats in Empty LSN Ranges

mwbayley:
SQL Server connector now emits heartbeats while catching up to the current LSN through empty LSN ranges.

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerStreamingChangeEventSource.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerStreamingChangeEventSource.java
@@ -245,6 +245,7 @@ public class SqlServerStreamingChangeEventSource implements StreamingChangeEvent
                         changeTables[i].next();
                     }
 
+                    boolean anyData = false;
                     for (;;) {
                         SqlServerChangeTablePointer tableWithSmallestLsn = null;
                         for (SqlServerChangeTablePointer changeTable : changeTables) {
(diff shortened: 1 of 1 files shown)


debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerStreamingChangeEventSource.java
```diff
@@ -346,6 +348,10 @@ else if (!checkAgent) {
                         tableWithSmallestLsn.next();
                     }
                     streamingExecutionContext.setLastProcessedPosition(TxLogPosition.valueOf(toLsn));
+                    if (!anyData) {
+                        offsetContext.setChangePosition(TxLogPosition.valueOf(toLsn), 0);
+                        dispatcher.dispatchHeartbeatEvent(partition, offsetContext);
+                    }
                     // Terminate the transaction otherwise CDC could not be disabled for tables
                     dataConnection.rollback();
```
Naros: On the slightest chance that rollback triggers a `SQLException`, would it make more sense to move the heartbeat and offset manipulation post rollback?

mwbayley: I'm new to the codebase and not sure of the consequences of the rollback, so I'll defer to you here. I assumed it was essentially just cleanup.

mwbayley: [rollback before heartbeat](https://github.com/debezium/debezium/pull/6663/commits/594cf55455568a6350074131ccf90563198b3bdd)

Naros: No problem. It's rolling back the current database transaction.  I suppose there is a tiny window here where we execute queries to get the data, including the calculation of `toLsn`. We don't emit any changes, but when rolling back the database connection, we encounter a network break and a `SQLException` is thrown. In this case, it's generally better that we let the exception throw, the connector terminates, and the offsets are not manipulated. Only when there is a successful pass over the complete iteration should we emit a heartbeat event on `noData`.

Naros:
Hi @mwbayley can you prefix your last commit with `DBZ-9364` as well?

mwbayley:
> Hi @mwbayley can you prefix your last commit with `DBZ-9364` as well?

Sure thing, I assumed it would be squashed initially.

vjuranek:
Merged, thanks @mwbayley !

PR #6668, opened 2025-08-21
DBZ-9374 Fix OpenTelemetry testing agent path

AlvarVG:


pom.xml
```diff
@@ -180,7 +180,7 @@
         <version.opentelemetry.testing>2.10.0</version.opentelemetry.testing>
```
jpechane:
```suggestion
        <version.opentelemetry.testing>2.10.0-alpha</version.opentelemetry.testing>
```

jpechane (changes requested):
@AlvarVG Thanks, I belive there is still one more change necessay.

pom.xml
AlvarVG: I have done it, and when Intellij reloads the maven project complains about the version 2.10.0-alpha-alpha not being available. I guess it gets concatenated automatically .

jpechane:
`debezium-bom/pom.xml` contains hardcoded `-alpha` suffix too.

jpechane:
BOM pdated and pushed

jpechane:
Applied, test failures unrelated

AlvarVG:
Thank you for taking over @jpechane

PR #6669, opened 2025-08-21
DBZ-9340 Debezium constantly performs heartbeat.action.query instead of honoring heartbeat.interval.ms

kmos:
closes: https://issues.redhat.com/browse/DBZ-9340

### DIFF
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/RecordsStreamProducerIT.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/RecordsStreamProducerIT.java
@@ -18,12 +18,14 @@ import static junit.framework.TestCase.assertTrue;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.entry;
 import static org.assertj.core.api.Assertions.fail;
+import static org.hamcrest.Matchers.equalTo;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertNotNull;
 
 import java.math.BigDecimal;
 import java.nio.ByteBuffer;
 import java.nio.charset.StandardCharsets;
+import java.time.Duration;
 import java.time.Instant;
 import java.time.LocalTime;
 import java.time.OffsetDateTime;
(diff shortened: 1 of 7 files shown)


debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/RecordsStreamProducerIT.java
```diff
@@ -3845,6 +3853,48 @@ record = consumer.remove();
         VerifyRecord.isValidTombstone(record, PK_FIELD, 1);
     }
 
+    @Test
+    @FixFor("DBZ-9340")
+    public void shouldSendHeartbeatBasedOnDefinedInterval() throws InterruptedException {
+        Logger logger = (Logger) LoggerFactory.getLogger(TestHeartbeatFactory.class);
```
mfvitale: You can use `io.debezium.junit.logging.LogInterceptor`

debezium-core/src/main/java/io/debezium/pipeline/EventDispatcher.java
```diff
@@ -453,6 +453,7 @@ public void alwaysDispatchHeartbeatEvent(P partition, OffsetContext offset) thro
     public void dispatchHeartbeatEvent(P partition, OffsetContext offset) throws InterruptedException {
         if (heartbeat instanceof ScheduledHeartbeat scheduledHeartbeat) {
```
jpechane, line 454: OK, this seems to be a bit of issue wrt the code robustness and maintenance. Is there a way how we could modify the contract/interface in the way we can drop `instanceof`?

kmos: Yes, we should align all the external connectors (Spanner, Vitess, etc.) to use the factory. I initially thought this could be addressed in a separate PR, but we can apply it in this one.

jpechane: @kmos I am fine doiing it in a separate PR. Let's get the correct behaviour first and refactoring as follow-up.

kmos: Ok, I can prepare a PR for the migration (using the factory and moving the constructors for `EventDispatcher` to `ScheduledHeartbeat`). The removal of this fragility should be straightforward.

jpechane:
Test failures seems unrelated but I rather re-execute the failed jobs.

jpechane:
@kmos Postgres test failures are related, the test doesn't fail on current main.
Tested also locally.

jpechane:
@kmos Applied, thanks

PR #6670, opened 2025-08-21
DBZ-9366 Adjust relaxed quote insert column value list parsing

Naros:
https://issues.redhat.com/browse/DBZ-9366

## Changes

When relaxed quote parsing was enabled and string contains escaped quotes during the inserts and updates, the logic for parsing the column value would reduce the `''` sequence a single `'` for updates, but would leave the `''` untouched for inserts. This caused inserts to publish a value that was not identical to the same value present for updates.


jpechane:
@Naros Applied, thanks

PR #6672, opened 2025-08-22
DBZ-9367 Add details to MariaDB connector binlog compression lack of support

jschnab:
Changes: Document that the MariaDB connector does not support binary log compression.
Closes: https://issues.redhat.com/browse/DBZ-9367

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/mariadb.adoc
+++ b/documentation/modules/ROOT/pages/connectors/mariadb.adoc
@@ -396,9 +396,9 @@ A new binary log format was introduced in MariaDB 11.4.
 When you use {prodname} with MariaDB 11.4 or later, you must set the MariaDB server variable https://mariadb.com/kb/en/replication-and-binary-log-system-variables/#binlog_legacy_event_pos[`binlog_legacy_event_pos`] to `1` (ON) to ensure that the connector can consume events in the new format.
 If you leave this variable at its default setting (`0`, or OFF), after a connector restart, {prodname} might not be able to find the resume point.
 
-The {prodname} {connector-name} connector currently does not support
-link:https://mariadb.com/docs/server/server-management/server-monitoring-logs/binary-log/compressing-events-to-reduce-size-of-the-binary-log[binary
-log compression]. Make sure to set the MariaDB server configuration option `log_bin_compress` to `0` to ensure the connector can consume all events.
+The {prodname} {connector-name} connector currently does not support link:https://mariadb.com/docs/server/server-management/server-monitoring-logs/binary-log/compressing-events-to-reduce-size-of-the-binary-log[binary
+log compression]. 
+To enable the connector to consume all events, you must set the MariaDB server configuration option `log_bin_compress` to `0`.
 ====
 
 // Type: procedure


documentation/modules/ROOT/pages/connectors/mariadb.adoc
```diff
@@ -324,13 +324,81 @@ include::{partialsdir}/modules/all-connectors/shared-mariadb-mysql.adoc[leveloff
...
 
 [IMPORTANT]
 ====
 A new binary log format was introduced in MariaDB 11.4.
 When you use {prodname} with MariaDB 11.4 or later, you must set the MariaDB server variable https://mariadb.com/kb/en/replication-and-binary-log-system-variables/#binlog_legacy_event_pos[`binlog_legacy_event_pos`] to `1` (ON) to ensure that the connector can consume events in the new format.
 If you leave this variable at its default setting (`0`, or OFF), after a connector restart, {prodname} might not be able to find the resume point.
+
+The {prodname} {connector-name} connector currently does not support
+link:https://mariadb.com/docs/server/server-management/server-monitoring-logs/binary-log/compressing-events-to-reduce-size-of-the-binary-log[binary
+log compression]. Make sure to set the MariaDB server configuration option `log_bin_compress` to `0` to ensure the connector can consume all events.
```
roldanbob:
```suggestion
The {prodname} {connector-name} connector currently does not support link:https://mariadb.com/docs/server/server-management/server-monitoring-logs/binary-log/compressing-events-to-reduce-size-of-the-binary-log[binary
log compression]. 
To enable the connector to consume all events, you must set the MariaDB server configuration option `log_bin_compress` to `0`.
```

roldanbob (changes requested):
The proposed change departs from the current model, which was implemented so that when a change affects both MariaDB and MySQL, it would be possible to update the content in one location.
Departing from the current model not only means that potential future changes to content in this section must be authored in two places, but it also makes it difficult to identify and trace the lineage of the changes.

If possible, I'd like to adhere to the current model. Ideally, when a piece of content applies to only one of the connectors, we could use a conditional that controls the context in which the content renders. However, in an earlier attempt, I discovered that although content conditionalized  with the `MYSQL` and `MARIADB` conditionals rendered as intended in the community version of the docs, this was not the case in the downstream docs.

However, I think that we can work around that by breaking the content in tagged region of the current shared file into separate regions and then interpolating a new region anywhere that MariaDB-specific changes are required. New `include` directives in the parent `mariadb.adoc` file can be used to incorporate the delta.  TBH, full disclosure: I'm debating whether that's worth the trouble.

Based on the following changes that I see in your PR, we would need 3 additional tagged regions in the shared file, and corresponding `include` directives in `mariadb.adoc`:

- Added the log_bin_compress property (L350)
- Added a log_bin_compress entry to the table of binlog config properties (L386-3890). This might also require converting the table of properties to a description list as in `ref-mariadb-mysql-*-connector-cfg-props.adoc`, because I'm not sure whether tables behave properly when you use an  `include` directive to insert a row.
- Added a paragraph to the admonition that specifies that the connector does not support binlog compression (L398-401)

PR #6678, opened 2025-08-26
DBZ-9392 Fix CLOB/BLOB column handing to support large data

shybovycha:
Fix the logic that handles BLOB/CLOB values to support large binary objects (2GB+).

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
@@ -92,7 +92,6 @@ public class OracleValueConverters extends JdbcValueConverters {
 
     private static final Pattern TO_TIMESTAMP_TZ = Pattern.compile("TO_TIMESTAMP_TZ\\('(.*)'\\)", Pattern.CASE_INSENSITIVE);
     private static final BigDecimal MICROSECONDS_PER_SECOND = new BigDecimal(1_000_000);
-    private static final long BLOB_SIZE_2GB = 2000000000L;
 
     private final OracleConnection connection;
     private final boolean legacyDecimalModeStrategy;
@@ -288,7 +287,7 @@ public class OracleValueConverters extends JdbcValueConverters {
 
             try {
                 // use buffered read for large CLOB values
-                if (clob.length() >= BLOB_SIZE_2GB) {
+                if (clob.length() >= Integer.MAX_VALUE) {
                     try (
                             BufferedReader reader = new BufferedReader(clob.getCharacterStream());
                             StringWriter writer = new StringWriter()) {
@@ -346,7 +345,7 @@ public class OracleValueConverters extends JdbcValueConverters {
             else if (data instanceof Blob) {
                 Blob blob = (Blob) data;
 
-                if (blob.length() >= BLOB_SIZE_2GB) {
+                if (blob.length() >= Integer.MAX_VALUE) {
                     // use buffered read to support large BLOB values
                     try (
                             BufferedInputStream inputStream = new BufferedInputStream(blob.getBinaryStream());
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
@@ -92,7 +92,7 @@ public class OracleValueConverters extends JdbcValueConverters {
 
     private static final Pattern TO_TIMESTAMP_TZ = Pattern.compile("TO_TIMESTAMP_TZ\\('(.*)'\\)", Pattern.CASE_INSENSITIVE);
     private static final BigDecimal MICROSECONDS_PER_SECOND = new BigDecimal(1_000_000);
-    private static final long BLOB_SIZE_2GB = 2L * 1024 * 1024 * 1024;
+    private static final long BLOB_SIZE_2GB = 2000000000L;
 
     private final OracleConnection connection;
     private final boolean legacyDecimalModeStrategy;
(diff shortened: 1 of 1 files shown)


Naros:
Hi @shybovycha have you profiled the code difference when the vast majority of the database has CLOB or BLOB data sets that are several hundred kilobytes or simply less than 2GB versus greater than 2GB?  My concern is that the driver's call to things like `getBytes` or `getSubString` are often optimized to minimize the number of round trips to the database, and the streaming path could introduce additional network calls. 

Is there any advantage here in checking the length? If it exceeds 2GB, you can use your approach; otherwise, you can use the original style approach.

shybovycha:
@Naros good point! will do the profiling tomorrow. i think you might be right about the branching approach - sounds reasonable, at least.

shybovycha:
@Naros you were right, streaming does indeed introduce a *significant* latency for smaller values - tested on 1MB values and it was approx. 3 times slower with streaming. 2KB values were fine - I presume because the block size is around or larger than 2KB, so they fit in one packet of data from the start. Added the branching.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
```diff
@@ -278,15 +284,28 @@ protected Object convertString(Column column, Field fieldDefn, Object data) {
...
             try {
-                Clob clob = (Clob) data;
-                // Note that java.sql.Clob specifies that the first character starts at 1
-                // and that length must be greater-than or equal to 0. So for an empty
-                // clob field, a call to getSubString(1, 0) is perfectly valid.
-                return clob.getSubString(1, (int) clob.length());
+                // use buffered read for large CLOB values
+                if (clob.length() >= BLOB_SIZE_2GB) {
+                    try (
+                            BufferedReader reader = new BufferedReader(clob.getCharacterStream());
```
jpechane, line 292: We have `io.debezium.util.IoUtil` class. Maybe that's a place where the strea/writer related code could be placed and called from both places of this PR.

jpechane:
@Naros Give there is a branch now, do you think the PR needs a new test to test 2GB case?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
shybovycha: the code is quite different between CLOB and BLOB cases - one uses `StringWriter` and the other uses `ByteArrayOutputStream`. moving these to the utility class is possible, but i don't see too much value in doing so, tbh, for the sake of three lines of code with no real abstraction benefits - since the writer either has to be passed as a parameter, diminishing the value of the change from sheer lines of code perspective, or it would be two implementation-specific methods, diminishing the value of abstraction:

```java
public static byte[] copyBytes(InputStream in) throws IOException {
    try (
        BufferedInputStream inputStream = new BufferedInputStream(in);
        ByteArrayOutputStream writer = new ByteArrayOutputStream()
    ) {
        inputStream.transferTo(writer);
        return writer.toByteArray();
    }
}

public static String copyString(Reader in) throws IOException {
    try (
        BufferedReader reader = new BufferedReader(in);
        StringWriter writer = new StringWriter()
    ) {
        reader.transferTo(writer);
        return writer.toString();
    }
}
```

and then

```java
if (clob.length() >= BLOB_SIZE_2GB) {
    try {
        data = IoUtil.copyString(clob.getCharacterStream());
    } catch (IOException e) {
        throw new DebeziumException("Couldn't read binary data for column " + column.name(), e);
    }
}
```

or

```java
if (blob.length() >= BLOB_SIZE_2GB) {
    try {
        data = IoUtil.readBytes(blob.getBinaryStream());
    }
    catch (SQLException | IOException e) {
        throw new DebeziumException("Couldn't read binary data for column " + column.name(), e);
    }
}
```

moreover, the other branch is still going to remain implementation-specific

```java
data = blob.getBytes(1, Long.valueOf(blob.length()).intValue());
```

this just creates unnecessary inconsistency in the code, in my opinion

shybovycha:
@jpechane on my end i can't think of a good way to make a test for 2GB+ dataset - storing a pre-computed blob in a repo is a really bad idea, but creating one row with 2GB CLOB data in Oracle (using Oracle sprocs) on RDS instance takes approx. 4 hrs. other than mocking and unit-testing, i can't think of a good way to test this new branch. do you have any ideas?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
jpechane: Good points!

jpechane:
> @jpechane on my end i can't think of a good way to make a test for 2GB+ dataset - storing a pre-computed blob in a repo is a really bad idea, but creating one row with 2GB CLOB data in Oracle (using Oracle sprocs) on RDS instance takes approx. 4 hrs. other than mocking and unit-testing, i can't think of a good way to test this new branch. do you have any ideas?

@shybovycha Have you had a chance to test it manually? If yes then under these conditions it is good enough.

shybovycha:
@jpechane tested on `1GB`, `1MB` and `2KB` CLOBs and `2KB` BLOBs - the approach works perfectly fine.

the `2097152000` (`2.09GB` - decimal part is the key here) bytes case failed though, but worked after reducing the `BLOB_SIZE_2GB` constant from `2L * 1024 * 1024 * 1024` (which i assumed was `2GB`) to `2000000000L` fixed the issue. the difference in translation - `GiB` vs `GB` 😞

i guess i will need to re-run the packit tests

Naros:
Applied, thanks @shybovycha :tada:

PR #6679, opened 2025-08-26
DBZ-9393 Relax check on `version` for OpenLineage emitted events

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-9393


jpechane:
@mfvitale Applied, thanks

PR #6681, opened 2025-08-26
DBZ-9375 Configure TCCL in threads where some classloading happens

vjuranek:
https://issues.redhat.com/browse/DBZ-9375

### DIFF
--- a/debezium-core/src/main/java/io/debezium/config/Instantiator.java
+++ b/debezium-core/src/main/java/io/debezium/config/Instantiator.java
@@ -25,12 +25,21 @@ public class Instantiator {
     }
 
     /**
-     * Instantiates the specified class either using the no-args constructor.
+     * Instantiates the specified class using the no-args constructor.
      *
      * @return The newly created instance or {@code null} if no class name was given
      */
     public static <T> T getInstance(String className) {
         return getInstanceWithProvidedConstructorType(className, null, null);
+    }
+
+    /**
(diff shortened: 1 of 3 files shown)


debezium-core/src/main/java/io/debezium/config/Instantiator.java
```diff
@@ -15,7 +15,15 @@
  */
 public class Instantiator {
 
+    private static ClassLoader classLoader;
```
Naros: Would this not present some concerns with it being static in Kafka Connect with different class loaders being applied to different plug-in paths?

jpechane: I agree, having a classloader in static attribute sound like a recipe for future issues. Also the access is unsychronized and I can easily imagin mulutithread approach.
If we are going to sotre custome classloader we'd IMHO need to move from static `Instantiator` to a class instance.

vjuranek: Agree that this is not the best way how to use user-specified class loader. I switch to another approach when class loader is specified as a parameter of the method.
Btw: it's actually not needed to fix this issue to use `Instantiator` with Spring class loader, but IMHO if we allow user to specify class loader, we should use it, thus I added it here.

jpechane:
@vjuranek Test failures unrelated, applied, thanks

PR #6684, opened 2025-08-27
[docs] Fix snippets dir location

jpechane:


roldanbob:
In this case, the path was rendered incorrect, because the content was moved directly into mariadb.adoc from the shared mariadb/mysql file. Reqarding quoting of the `snippetsdir` attribute value, while that's harmless, quoting is only required if the value starts with an attribute reference, or if the path contains spaces or other characters that require quoting in YAML.

jpechane:
@roldanbob The quoting change was merely cosmetic. Other path definitions are also quoted so I added it for consistency.

PR #6687, opened 2025-08-27
DBZ-9328 MongoEventRouter ignores tracing.* configurations

jpechane:


jpechane:
This can be merged when MongoDB CI passes.

PR #6688, opened 2025-08-27
DBZ-9398 Use only one batch to avoid RejectedExecutionException during shutdown

vjuranek:
https://issues.redhat.com/browse/DBZ-9398


jpechane:
@vjuranek Applied, thanks

PR #6689, opened 2025-08-27
DBZ-8979 JdbcSchemaHistory Fails to Handle Data Sharding When Recover…

ochedru:
https://issues.redhat.com/browse/DBZ-8979


debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/history/JdbcSchemaHistory.java
```diff
@@ -167,16 +169,29 @@ protected synchronized void recoverRecords(Consumer<HistoryRecord> records) {
                         try (
                                 Statement stmt = conn.createStatement();
                                 ResultSet rs = stmt.executeQuery(config.getTableSelect())) {
+                            StringBuilder sb = new StringBuilder();
+                            int currentRecordSeq = Integer.MAX_VALUE;
                             while (rs.next()) {
+                                int recordSeq = rs.getInt("record_insert_seq");
```
jpechane: The column names should be configurable (ithis is oversight in "history_data") too.

ochedru:
Thank you for your feedback. Actually, I've been testing further and I confirm `record_insert_seq` cannot be used to sort rows when recovering records because `record_insert_seq values` are reused when the server is restarted (this is an atomic counter).
I will update this PR accordingly.

jpechane:
Applied, thanks!

ochedru:
@jpechane In the case of large schemas (like mine), it can be beneficial in terms of I/O to compress json before storing it as a base64-encoded string into the database. With gzip, I get a compression ratio of approximately 1 to 4 in my case: from around 800 MB to 200 MB if I sum the length of history_data on all rows. Of course, this would be disabled by default. Do you think this is worth creating a new feature ticket? Thanks.

PR #6701, opened 2025-08-29
DBZ-9413 Fix offset changes for incremental/blocking snapshots

Naros:
https://issues.redhat.com/browse/DBZ-9413

@mfvitale this was the fix and the reason we were seeing odd behaviors with offsets during blocking snapshots :sob: I am curious to see how tests fair with this change, along with incremental snapshots :crossed_fingers:

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleOffsetContext.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleOffsetContext.java
@@ -14,7 +14,6 @@ import java.util.stream.Collectors;
 
 import org.apache.kafka.connect.data.Schema;
 
-import io.debezium.connector.AbstractSourceInfo;
 import io.debezium.connector.SnapshotRecord;
 import io.debezium.connector.SnapshotType;
 import io.debezium.pipeline.CommonOffsetContext;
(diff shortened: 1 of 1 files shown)


Naros:
At least one test failure is related, will look at it next week.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleOffsetContext.java
```diff
@@ -177,58 +177,48 @@ public static Builder create() {
...
+
+            final String encodedPendingTransactions = getEncodedSnapshotPendingTransactions();
+            if (!Strings.isNullOrEmpty(encodedPendingTransactions)) {
+                result.put(SNAPSHOT_PENDING_TRANSACTIONS_KEY, encodedPendingTransactions);
             }
-            offset.put(SNAPSHOT_SCN_KEY, snapshotScn != null ? snapshotScn.isNull() ? null : snapshotScn.toString() : null);
+        }
 
-            return offset;
+        if (sourceInfo.getLcrPosition() != null) {
```
mfvitale, line 192: Would't it better to explicitly check if the adapter is `XStream` instead to implicitly check on nullability?

mfvitale (approved):
LGTM. I just left a minor comment.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleOffsetContext.java
Naros: I could, but I actually wonder if it wouldn't be better to check nullability for each field and append it to the resulting Map if it's available, regardless of the adapter in play. In the long-term, I'd prefer to try and isolate as much of this sort of logic in an object-oriented way by having different offset context implementations per adapter, but that's likely a 4.0 task. For now, since I want to tackle how I manage this differently with a OO approach, I'll leave this as-is for now to avoid introducing any behavior differences.

vjuranek: Makes sense @Naros

vjuranek (approved):

vjuranek:
Test failures doesn't seem to be related.

vjuranek:
Merged, thanks @Naros !

PR #6704, opened 2025-08-29
DBZ-9415 Upgrade grpc version to align with netty upgrade

twthorn:
We upgraded netty. We need to upgrade version.grpc to a version that is compatible with those newer netty libs.


twthorn:
Closing this PR, we need to complete [DBZ-9426](https://issues.redhat.com/browse/DBZ-9426) before we do this

PR #6708, opened 2025-08-30
DBZ-9416 Avoid creating source Struct with partial data

Naros:
https://issues.redhat.com/browse/DBZ-9416


jpechane:
@Naros Test failures undrelated, applied, thanks!

PR #6709, opened 2025-08-31
DBZ-9395: Selectively call ALTER PUBLICATION for filtered publication

rajdangwal:
Addressing https://issues.redhat.com/browse/DBZ-9395

Also, replaced client-side `Statement#setQueryTimeout(...)` with server-side `SET statement_timeout` at session scope to prevent orphaned queries when the connector stops or shuts down ungracefully. `setQueryTimeout` only cancels on the client, if the client disappears while the DB is waiting/blocked, the server query can keep running, leaking resources. `statement_timeout` is enforced by PostgreSQL, so the server cancels the statement even if the client is gone.

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
@@ -158,20 +158,20 @@ public class PostgresReplicationConnection extends JdbcConnection implements Rep
      * Captures the current statement_timeout value from the database.
      *
      * @param stmt the Statement to use for querying
-     * @return the current statement_timeout value as a String, or null if unable to capture
+     * @return the current statement_timeout value as an Optional<String>, empty if unable to capture
      */
-    private String captureCurrentStatementTimeout(Statement stmt) {
+    private Optional<String> captureCurrentStatementTimeout(Statement stmt) {
         try (ResultSet timeoutRs = stmt.executeQuery("SHOW statement_timeout;")) {
             if (timeoutRs.next()) {
                 String timeout = timeoutRs.getString(1);
                 LOGGER.debug("Captured original statement_timeout: {}", timeout);
-                return timeout;
+                return Optional.of(timeout);
             }
         }
         catch (SQLException ex) {
             LOGGER.warn("Failed to capture current statement_timeout", ex);
         }
-        return null;
+        return Optional.empty();
     }
 
     /**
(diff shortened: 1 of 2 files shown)


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
```diff
@@ -154,6 +154,66 @@ private ServerInfo.ReplicationSlot getSlotInfo() throws SQLException, Interrupte
         }
     }
 
+    /**
+     * Captures the current statement_timeout value from the database.
+     *
+     * @param stmt the Statement to use for querying
+     * @return the current statement_timeout value as a String, or null if unable to capture
+     */
+    private String captureCurrentStatementTimeout(Statement stmt) {
```
jpechane: I'd prefer `Optional` here

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
```diff
@@ -154,6 +154,66 @@ private ServerInfo.ReplicationSlot getSlotInfo() throws SQLException, Interrupte
...
+     * Executes a statement with a temporary statement_timeout, automatically restoring the original timeout afterwards.
+     *
+     * @param stmt the Statement to use for execution
+     * @param statementToExecute the SQL statement to execute
+     * @throws SQLException if the execution fails
+     */
+    private void executeWithTimeout(Statement stmt, String statementToExecute) throws SQLException {
+        String originalTimeout = captureCurrentStatementTimeout(stmt);
+        try {
+            String lineSeparator = System.lineSeparator();
```
jpechane: Is this temporary variable needed?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
```diff
@@ -291,6 +352,82 @@ private boolean isReadOnlyDb() throws SQLException {
         return isReadOnly.get();
     }
 
+    /**
+     * Gets the list of tables currently configured in the publication.
+     *
+     * @param stmt the statement to use for the query
+     * @return Set of TableId objects representing tables in the publication, or null if unable to query
+     */
+    private Set<TableId> getCurrentPublicationTables(Statement stmt) {
```
jpechane: Please use `Optional`

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
```diff
@@ -291,6 +352,82 @@ private boolean isReadOnlyDb() throws SQLException {
...
+        catch (Exception e) {
+            throw new SQLException("Failed to determine captured tables", e);
+        }
+
+        if (desiredTables.isEmpty()) {
+            LOGGER.warn("No table filters found for filtered publication {}", publicationName);
+            return false;
+        }
+
+        boolean updateRequired = !currentPublicationTables.equals(desiredTables);
```
jpechane: We use multiple returns in this method so it can me modified like
```suggestion
        boolean updateRequired = !currentPublicationTables.equals(desiredTables);
        if (currentPublicationTables.equals(desiredTables)) {
            LOGGER.info("Publication '{}' is already up to date with desired tables", publicationName);
            return false;
        }
        
```

jpechane (commented):
@rajdangwal Overalll, LGTM thanks! I left couple of really pedantic comments

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresReplicationConnection.java
rajdangwal: Removed.

rajdangwal:
Thanks @jpechane 
Addressed the comments.

jpechane:
@rajdangwal Applied, thanks!

PR #6711, opened 2025-09-02
DBZ-9422 Fix duplicate context headers during incremental snapshot

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-9422


Naros:
Applied, thanks @mfvitale

PR #6718, opened 2025-09-05
DBZ-9427: Introduce guardrail to limit number of tables captured

rajdangwal:
Addressing: https://issues.redhat.com/browse/DBZ-9427

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorTask.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorTask.java
@@ -6,9 +6,11 @@
 package io.debezium.connector.sqlserver;
 
 import java.sql.SQLException;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
+import java.util.Set;
 import java.util.stream.Collectors;
 
 import org.apache.kafka.connect.source.SourceRecord;
@@ -114,6 +116,9 @@ public class SqlServerConnectorTask extends BaseSourceTask<SqlServerPartition, S
         connectorConfig.getBeanRegistry().add(StandardBeanNames.CDC_SOURCE_TASK_CONTEXT, taskContext);
 
         final SnapshotterService snapshotterService = connectorConfig.getServiceRegistry().tryGetService(SnapshotterService.class);
+
+        // Validate guardrail limits for captured tables to prevent loading excessive table schemas into memory
+        validateGuardrailLimits(connectorConfig, dataConnection);
 
         validateSchemaHistory(connectorConfig, dataConnection::validateLogPosition, offsets, schema,
                 snapshotterService.getSnapshotter());
@@ -236,6 +241,30 @@ public class SqlServerConnectorTask extends BaseSourceTask<SqlServerPartition, S
     @Override
     protected Iterable<Field> getAllConfigurationFields() {
         return SqlServerConnectorConfig.ALL_FIELDS;
+    }
+
+    private void validateGuardrailLimits(SqlServerConnectorConfig connectorConfig, SqlServerConnection connection) {
+        try {
+            // Get all table IDs that match the connector's filters
+            Set<TableId> allTableIds = new HashSet<>();
+            for (String databaseName : connectorConfig.getDatabaseNames()) {
+                allTableIds.addAll(connection.readTableNames(databaseName, null, null, new String[]{ "TABLE" }));
+            }
+
+            Set<TableId> capturedTables = allTableIds.stream()
+                    .filter(tableId -> connectorConfig.getTableFilters().dataCollectionFilter().isIncluded(tableId))
+                    .collect(java.util.stream.Collectors.toSet());
+
+            List<String> tableNames = capturedTables.stream()
+                    .map(TableId::toString)
+                    .collect(java.util.stream.Collectors.toList());
+
+            connectorConfig.validateGuardrailLimits(capturedTables.size(), tableNames);
+
+        }
+        catch (SQLException e) {
+            throw new DebeziumException("Failed to validate guardrail limits", e);
+        }
     }
 
 }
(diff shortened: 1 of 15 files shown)


Naros:
Hi @rajdangwal please make sure that you're also adding a comment `/packit test --labels oracle` on this PR to make sure that the Oracle Integration Tests are being executed so any changes needed there are also fixed.

rajdangwal:
@Naros the currently failing tests are irrelevant to the change, FYI.
Thanks.

vjuranek (approved):
LGTM

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnectorTask.java
```diff
@@ -238,4 +242,25 @@ protected Iterable<Field> getAllConfigurationFields() {
...
+            // Get all table IDs using the connection's method, similar to PostgresConnectorTask
+            Set<TableId> allTableIds = connection.getAllTableIds(connectorConfig.getDatabaseNames());
+
+            Set<TableId> capturedTables = allTableIds.stream()
+                    .filter(tableId -> connectorConfig.getTableFilters().dataCollectionFilter().isIncluded(tableId))
+                    .collect(java.util.stream.Collectors.toSet());
+
+            List<String> tableNames = capturedTables.stream()
+                    .map(TableId::toString)
+                    .collect(java.util.stream.Collectors.toList());
```
Naros: What do you think about collapsing these two streams into a single stream call, i.e.
```suggestion
            Set<String> capturedTables = allTableIds.stream()
                    .filter(tableId -> connectorConfig.getTableFilters().dataCollectionFilter().isIncluded(tableId))
                    .map(TableId::toString)
                    .collect(java.util.stream.Collectors.toSet());
```
I think this can be applied to several implementations in the PR. Unless I missed something, it did not seem clear to me why we needed a `Set` and `List` for the validate method.

rajdangwal: Thanks @Naros 
Yes, the suggestion makes sense. I have addressed it.

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -1571,6 +1659,58 @@ public boolean skipMessagesWithoutChange() {
...
+
+            // Log table list at debug level for troubleshooting
+            if (tableNames != null && LOGGER.isDebugEnabled()) {
+                String tableList = tableNames.stream()
+                        .sorted()
+                        .collect(java.util.stream.Collectors.joining(", "));
+                LOGGER.debug("Tables/Collections configured for capture: {}", tableList);
+            }
+
+            GuardrailLimitAction action = getGuardrailLimitAction();
```
jpechane: There is no need for a separate variable.

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -1288,6 +1350,28 @@ public static SnapshotQueryMode parse(String value, String defaultValue) {
...
+            .withGroup(Field.createGroupEntry(Field.Group.ADVANCED, 33))
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.MEDIUM)
+            .withDescription("The maximum number of tables that can be captured by the connector. " +
+                    "When this limit is exceeded, the action specified by 'guardrail.limit.action' will be taken. " +
+                    "Set to 0 to disable this guardrail.")
+            .withDefault(0)
+            .withValidation(Field::isNonNegativeInteger);
+
+    public static final Field GUARDRAIL_LIMIT_ACTION = Field.create("guardrail.limit.action")
```
jpechane:
```suggestion
    public static final Field GUARDRAIL_TABLES_LIMIT_ACTION = Field.create("guardrail.tables.limit.action")
```

Let's start buikding a namespace tree

jpechane (commented):
@rajdangwal Thanks. I left two minor comments.
I have one bigger question though. It seems to me that `validateGuardrailLimits` method is very similar in each connector. Doesn't make sense to move the validation logic to guradrail validation logic into connection classes completely with possibility of additional code re-use/refactoring?

rajdangwal:
> It seems to me that `validateGuardrailLimits` method is very similar in each connector. Doesn't make sense to move the validation logic to guradrail validation logic into connection classes completely with possibility of additional code re-use/refactoring?

Thanks for the review @jpechane 
Yes, the methods are similar but not exactly the same. IIUC, you are suggesting to move the guardrail validation logic to connector specific connection classes, right?
Even if we move the logic to the connection classes, there would still be duplication, IMO.
We would probably save some lines for Binlog based connectors (MySql and MariaDb) but not much. 
Please let me know if you think otherwise and I would move the logic to the connection classes.

jpechane:
> > It seems to me that `validateGuardrailLimits` method is very similar in each connector. Doesn't make sense to move the validation logic to guradrail validation logic into connection classes completely with possibility of additional code re-use/refactoring?
> 
> Thanks for the review @jpechane Yes, the methods are similar but not exactly the same. IIUC, you are suggesting to move the guardrail validation logic to connector specific connection classes, right? Even if we move the logic to the connection classes, there would still be duplication, IMO. We would probably save some lines for Binlog based connectors (MySql and MariaDb) but not much. Please let me know if you think otherwise and I would move the logic to the connection classes.

OK, let's keep it as is. Thanks

jpechane:
@rajdangwal MongoDB test failure seems to be related. The failing test passes locally on `main` but fails with the PR.

rajdangwal:
> @rajdangwal MongoDB test failure seems to be related. The failing test passes locally on `main` but fails with the PR.

My bad @jpechane 
I missed that failing case. I have fixed it now.

jpechane:
Test failures unrelated
@rajdangwal Applied, thanks!
Could you please also send a docs PR?

som-design:
ฉันเสียใจ ที่ไม่สามารถ ปรับเปลี่ยนแก้ไขได้ เนื่องจาก ไม่สามารถ เข้าไป
แก้ไขได้

ในวันที่ ศ. 12 ก.ย. 2025 13:45 Jiri Pechanec ***@***.***>
เขียนว่า:

PR #6721, opened 2025-09-05
DBZ-9402-adds-section-to-discuss-unsupported-oracle-datatypes

roldanbob:
[DBZ-9402](https://issues.redhat.com/browse/DBZ-9402)

Adds information to Oracle _Data Types_ section about data types that the connector does not support.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/oracle.adoc
+++ b/documentation/modules/ROOT/pages/connectors/oracle.adoc
@@ -1927,8 +1927,8 @@ The connector cannot process columns that include the following types of data:
 * UROWID (Universal ROWID)
 * VECTOR
 
-If the connector attempts to capture an event that includes a column with one of these unsupported data types, it logs a warning, and skips the column with the unsupported type.
-The connector continues to process other columns and tables, but data from the unsupported column is lost.
+If the connector attempts to capture an event that includes a column with one of these unsupported data types, it skips the event and then continues to process subsequent events.
+The connector continues to process other columns and tables that are configured for capture, but data from the unsupported column is lost.
 
 [id="oracle-character-types"]
 === Character types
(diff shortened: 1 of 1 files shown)


documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -1906,6 +1909,27 @@ Support for further data types is planned for subsequent releases.
...
+* JavaScript-based stored procedures
+* LONG
+* LONG RAW
+* ROWID (not supported on XStream)
+* Table value constructor (TVCs)
+* UROWID (Universal ROWID)
+* VECTOR
+
+If the connector attempts to capture an event that includes a column with one of these unsupported data types, it logs a warning, and skips the column with the unsupported type.
+The connector continues to process other columns and tables, but data from the unsupported column is lost.
```
Naros:
```suggestion
If the connector attempts to capture an event that includes a column with one of these unsupported data types, it skips the event and continues to process subsequent events.
```
Unforunately most of these generate an `UNSUPPORTED` event type in LogMiner and when this happens, it returns absolutely nothing about the event, not even the other columns :disappointed:

Naros (commented):
I left just one comment inline.

documentation/modules/ROOT/pages/connectors/oracle.adoc
roldanbob: Oh ....  Ok, so it's more a case of:

> If the connector attempts to capture an event that includes a column with one of these unsupported data types, it skips the event.

But we don't throw an exception, right? So processing for other events continues?

Naros: Correct, we do not throw an exception @roldanbob .

Naros:
Did you @roldanbob? It looks the same as it did previously, did you fail to push the commit?

roldanbob:
Oh, geez. Why was I thinking that this was my review of your PR. Doh! 
I'll make the change and push it.

Naros (approved):

Naros:
Applied, thanks @roldanbob.

PR #6730, opened 2025-09-12
DBZ-9446: handle wrapped reselect SQLException

kinolaev:
Fix for https://issues.redhat.com/browse/DBZ-9446

### DIFF
--- a/debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
+++ b/debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
@@ -1545,6 +1545,15 @@ public class JdbcConnection implements AutoCloseable {
         return Set.of('%', '_');
     }
 
+    public StatementPreparer statementValuesSetter(List<Object> values) {
+        return statement -> {
+            int index = 1;
+            for (final Object parameter : values) {
+                statement.setObject(index++, parameter);
+            }
+        };
+    }
+
     public <T> ResultSetMapper<T> singleResultMapper(ResultSetExtractor<T> extractor, String error) throws SQLException {
         return (rs) -> {
             if (rs.next()) {
@@ -1554,6 +1563,25 @@ public class JdbcConnection implements AutoCloseable {
                 }
             }
             throw new IllegalStateException(error);
+        };
+    }
+
+    public <T> ResultSetMapper<T> singleResultMapper(ResultSetExtractor<T> extractor, T defaultValue, String msg) throws SQLException {
+        return singleResultMapper(extractor, defaultValue, msg, msg);
+    }
+
+    public <T> ResultSetMapper<T> singleResultMapper(ResultSetExtractor<T> extractor, T defaultValue, String notFoundMsg, String multipleResultMsg)
+            throws SQLException {
+        return (rs) -> {
+            if (!rs.next()) {
+                LOGGER.warn(notFoundMsg);
+                return defaultValue;
+            }
+            final T ret = extractor.apply(rs);
+            if (rs.next()) {
+                LOGGER.warn(multipleResultMsg);
+            }
+            return ret;
         };
     }
 
@@ -1722,21 +1750,12 @@ public class JdbcConnection implements AutoCloseable {
 
     protected boolean reselectColumns(String query, TableId tableId, List<String> columns, List<Object> bindValues, ResultSetConsumer resultConsumer)
             throws SQLException {
-        final PreparedStatement statement = createPreparedStatement(query);
-        int index = 1;
-        for (final Object parameter : bindValues) {
-            statement.setObject(index++, parameter);
-        }
-        try (ResultSet resultSet = statement.executeQuery()) {
-            if (!resultSet.next()) {
-                LOGGER.warn("No data found for re-selection on table {}.", tableId);
-                return false;
-            }
+        return prepareQueryAndMap(query, statementValuesSetter(bindValues), singleResultMapper(resultSet -> {
             resultConsumer.accept(resultSet);
-            if (resultSet.next()) {
-                LOGGER.warn("Re-selection detected multiple rows for the same key in table {}, using first.", tableId);
-            }
-        }
-        return true;
+            return true;
+        },
+                false,
+                "No data found for re-selection on table %s.".formatted(tableId),
+                "Re-selection detected multiple rows for the same key in table %s, using first.".formatted(tableId)));
     }
 }


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
```diff
@@ -692,9 +692,17 @@ public Map<String, Object> reselectColumns(Table table, List<String> columns, Li
...
+        SQLException sqlException = null;
+        if (exception instanceof SQLException) {
+            sqlException = (SQLException) exception;
+        }
+        else if (exception.getCause() instanceof SQLException) {
+            sqlException = (SQLException) exception.getCause();
+        }
+        return sqlException != null &&
+                ORACLE_RESELECT_ERROR_CODE_FALLBACK.contains(sqlException.getErrorCode()) ||
+                ORACLE_RESELECT_ERROR_PREFIX_FALLBACK.stream().anyMatch(sqlException.getMessage()::startsWith);
```
Naros: I would suggest using a `Throwable` here, and walk up the cause call stack to avoid any potential cases where a DebeziumException gets wrapped by another DebeziumException but where the original cause is a SQLException. 

By walking the exception stack, we avoid these corner cases falling through the cracks.  wdyt @kinolaev ?

kinolaev: Thanks @Naros , totally agree, I'll update the PR soon

kinolaev: Done in 854dbf36d3ae780fbe0b793295637a7444f58ea2

Naros:
Hi @kinolaev so looking a bit more closely at the issue, I'm not convinced this addresses the problem. I do believe the change is good nonetheless, but it doesn't address the call to `getConvertedValue` in the reselect post processor class, which is where I believe the Jira issue's error stems from. wdyt?

kinolaev:
Yes, you are right. I faced the issue today even with the PR applied(
Seems like we should read clobs just after reselection from an scn in the OracleConnector. But if this operation is not cached it'll double the load on the database, that is not acceptable for us. Do you now whether clob.getSubString is cached or not?
Or maybe we should return string value of clobs from reselectColums method?
I will explore both possibilities further this week

Naros:
The `ReselectPostProcessor` currently performs two iterations of the selectables, where the first is the returned HashMap while the latter is where we apply the conversion. By pushing the value converter down into the `reselectColumns` call, we actually could reduce this to 1 iteration, but we'd also need to pass in the `Struct` for the `after`.

The question is, are we pushing too much into the connection class where we're mixing concerns?

I definitely believe that query construction and perhaps binding should be handled by the `JdbcConnection` or concrete connector derivatives, giving us the ability to deal with those query semantics. Still, a lot of the result set handling is particular to the reselect post-processor's needs.

What do you think about the `ReselectPostProcessor` providing a `ParameterResutlSetConsumer` implementation? 

In this case we could pull up all the iteration that revolves around the ResultSet into the post-processor, which would prevent us from needing to pass the value converter nor the after `Struct` into the connection class method, and allows us to reduce the iteration to a single pass while keeping non-query build/bind semantics out of the `JdbcConnection` class?

In the long term, I believe this will certainly align better with some future work where we'd ideally like to see other state mediums like RocksDB be the source for the reselect values rather than the relational database, where you gain much better cache locality & performance, but that's outside the scope of this change.

@jpechane care to share any thoughts?

jpechane:
> @jpechane care to share any thoughts?

Sounds good to me! Also the change is not invasive in the sense that it would affect a public API so it is safe to go forward with it.

kinolaev:
As was suggested earlier I moved conversions into ParameterResultSetConsumer https://github.com/debezium/debezium/pull/6730/commits/af12b46a2490a142cb29319e037b1093a1283f43

debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -1710,29 +1710,30 @@ public String getQualifiedTableName(TableId tableId) {
...
-        return reselectColumns(query, table.id(), columns, keyValues);
+        return reselectColumns(query, table.id(), columns, keyValues, resultConsumer);
     }
 
-    protected Map<String, Object> reselectColumns(String query, TableId tableId, List<String> columns, List<Object> bindValues) throws SQLException {
-        final Map<String, Object> results = new HashMap<>();
+    protected boolean reselectColumns(String query, TableId tableId, List<String> columns, List<Object> bindValues, ParameterResultSetConsumer resultConsumer)
+            throws SQLException {
+        final boolean[] flags = { false };
         prepareQuery(query, bindValues, (params, rs) -> {
```
Naros: Could we use `prepareQueryAndMap` and have the lambda return true/false to the caller? We could then get rid of the `boolean[]` flags variable.

kinolaev: Done in ec5dde6916c2894b75ab8ec4ebd308da9121df15

kinolaev: I think, it looks cleaner without prepareQueryAndMap:
```java
    protected boolean reselectColumns(String query, TableId tableId, List<String> columns, List<Object> bindValues, ResultSetConsumer resultConsumer)
            throws SQLException {
        final PreparedStatement statement = createPreparedStatement(query);
        int index = 1;
        for (final Object parameter : bindValues) {
            statement.setObject(index++, parameter);
        }
        try (ResultSet resultSet = statement.executeQuery()) {
            if (!resultSet.next()) {
                LOGGER.warn("No data found for re-selection on table {}.", tableId);
                return false;
            }
            resultConsumer.accept(resultSet);
            if (resultSet.next()) {
                LOGGER.warn("Re-selection detected multiple rows for the same key in table {}, using first.", tableId);
            }
        }
        return true;
    }
```
wdyt @Naros ?

Naros: That seems reasonable to me @kinolaev

jpechane: There are also `io.debezium.jdbc.JdbcConnection.singleResultMapper(ResultSetExtractor<T>, String)` and `io.debezium.jdbc.JdbcConnection.querySingleValue(Connection, String, StatementPreparer, ResultSetExtractor<T>)`. Maybe it would be good to provide alternatives that emit a warning instead of exception and we'll get best of the both worlds.

kinolaev: Done without helper methods in 1a37cc646945ac72903d68a19e1a748b808b71b2.
@jpechane, I looked at the methods you suggested but sorry I didn't get what version of them could make reselectColumns cleaner. It looks like it's sort of special case where we don't need a mapper or an extractor, but a consumer plus an indication whether it was called or not.
For now I think it's better to keep it as is. But if you give me a hint I'll look at it again)

jpechane: Well, the methods I pointed do have exactly the same structure as your plain JDBC code. So I propose to introduce generic methods that are the same as the ones I pointed with single difference, instead of throwing an error they'll log warning - exatcly as you do and use them here. The general methods can be useful in the codebase elsewhere.

kinolaev: @jpechane , please look at fee7cbbe85c397560714d6160243922f1543e2a4, I hope it's close to how you see it. Additionally I've introduced `statementValuesSetter` method that accept a list of statement values and returns `StatementPreparer`. Let me know if you want me to remove it.

debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -1710,29 +1742,24 @@ public String getQualifiedTableName(TableId tableId) {
...
-        });
-        return results;
+    protected boolean reselectColumns(String query, TableId tableId, List<String> columns, List<Object> bindValues, ResultSetConsumer resultConsumer)
+            throws SQLException {
+        return prepareQueryAndMap(query, statementValuesSetter(bindValues), singleResultMapper(resultSet -> {
+            resultConsumer.accept(resultSet);
+            return true;
+        },
+                false,
+                String.format("No data found for re-selection on table {}.", tableId),
```
jpechane:
```suggestion
                "No data found for re-selection on table %s.".formatted(tableId),
```

debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -1710,29 +1742,24 @@ public String getQualifiedTableName(TableId tableId) {
...
+                String.format("Re-selection detected multiple rows for the same key in table {}, using first.", tableId)));
```
jpechane:
```suggestion
                "Re-selection detected multiple rows for the same key in table %s, using first.".formatted(tableId)));
```

debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
kinolaev: I've amended  the last commit, see 635321563a24bafa71f3dae85118e38258e147b2

debezium-core/src/main/java/io/debezium/jdbc/JdbcConnection.java
kinolaev: I've amended  the last commit, see 635321563a24bafa71f3dae85118e38258e147b2

kinolaev:
I guess, I have to remove `<T> ResultSetMapper<T> singleResultMapper(ResultSetExtractor<T> extractor, T defaultValue)` variant because it conflicts with the original `<T> ResultSetMapper<T> singleResultMapper(ResultSetExtractor<T> extractor, String error)` when T is String.

upd: removed in a8e3cfb0e001b81c89b84e847838bb778bb66081

kinolaev:
Could you please help me understand if the failed tasks are related to the PR?

Naros:
> Could you please help me understand if the failed tasks are related to the PR?

Hi @kinolaev the test failures are unrelated.

jpechane:
@kinolaev Appllied, thanks!

PR #6731, opened 2025-09-12
DBZ-9160 Multi-engine Support

kmos:
## Context

Currently, the extension supports only one engine at a time in the same Quarkus Application (JVM). We want to enable the use of more than one engine within the same Quarkus application. The **Debezium Multi-Engine** support for Quarkus makes it possible to capture events from different sources or to use multiple engines from the same source.

## Decision

To address multi-engine support in Quarkus, we have introduced a series of new concepts:

- **Capture Groups**: a capture group is a set of working Debezium Servers that attempt to capture log events using the same configuration for a particular source (for example, in Postgres using the same slot). In a Quarkus application, only one engine can be assigned to a capture group. A capture group is defined by an ID.
- **Debezium Context**: contextual information assigned to a particular engine.
- **CaptureGroup Qualifier**: in cases where a user is *observing* events coming from the engine (such as *notifications*, *heartbeats*, *lifecycle events*), it is possible to filter these events by engine using the `CaptureGroup` qualifier, as shown below:

```java
@ApplicationScoped
public class SnapshotEventObserver {

    private final List<SnapshotEvent> defaultSnapshotEvents = new ArrayList<>();
    private final List<SnapshotEvent> alternativeSnapshotEvents = new ArrayList<>();

    public void defaultSnapshot(@Observes @CaptureGroup("default") SnapshotEvent snapshot) {
        defaultSnapshotEvents.add(snapshot);
    }

    public void alternativeSnapshot(@Observes @CaptureGroup("alternative") SnapshotEvent snapshot) {
        alternativeSnapshotEvents.add(snapshot);
    }

}
```

In the example above, default and alternative are two groups assigned to two different engines that are capturing from two different sources.

The features supported in the multi-engine implementation are:
- [x] Capturing Listener
- [x] Custom Deserializer
- [x] Lifecycle events
- [x] Heartbeat events
- [x] Custom Data Type Converter
- [x] Notification events
- [x] Post Processors

### Different configuration convention

Currently, in a single-engine configuration, we support the following convention:

```properties
# Debezium CDC configuration
quarkus.debezium.name=native
quarkus.debezium.topic.prefix=dbserver1
quarkus.debezium.plugin.name=pgoutput
quarkus.debezium.snapshot.mode=initial

## deserializer
quarkus.debezium.capturing.default.destination=dbserver1.public.products
quarkus.debezium.capturing.default.deserializer=io.quarkus.sample.app.deserializer.ProductDeserializer

quarkus.datasource.devservices.enabled=false
quarkus.datasource.db-kind=postgresql
quarkus.datasource.username=native
quarkus.datasource.password=native
quarkus.datasource.jdbc.url=jdbc:postgresql://localhost:5432/native
```

This convention is not applicable to multiple engines, so it is necessary to add different configurations inside the `capturing` prefix, as shown below:

```properties
# Debezium Default CDC configuration
quarkus.debezium.name=native
quarkus.debezium.topic.prefix=dbserver1
quarkus.debezium.plugin.name=pgoutput
quarkus.debezium.snapshot.mode=initial

# Debezium CDC configuration for alternative

quarkus.debezium.capturing.alternative.group-id=alternative
quarkus.debezium.capturing.alternative.offset.storage=org.apache.kafka.connect.storage.MemoryOffsetBackingStore
quarkus.debezium.capturing.alternative.name=alternative
quarkus.debezium.capturing.alternative.topic.prefix=dbserver2
quarkus.debezium.capturing.alternative.plugin.name=pgoutput
quarkus.debezium.capturing.alternative.snapshot.mode=initial


# Default Quarkus datasource configuration

quarkus.datasource.devservices.enabled=false
quarkus.datasource.db-kind=postgresql
quarkus.datasource.username=native
quarkus.datasource.password=native
quarkus.datasource.jdbc.url=jdbc:postgresql://localhost:5432/native

# Alternative Quarkus datasource configuration

quarkus.datasource.alternative.db-kind=postgresql
quarkus.datasource.alternative.username=alternative
quarkus.datasource.alternative.password=alternative
quarkus.datasource.alternative.jdbc.url=jdbc:postgresql://localhost:5433/alternative
quarkus.debezium.capturing.alternative.heartbeat.interval.ms=5

quarkus.debezium.capturing.alternative.orders.destination=dbserver2.public.orders
quarkus.debezium.capturing.alternative.orders.deserializer=io.quarkus.sample.app.deserializer.OrderDeserializer
```

### Engine Unique Constraint

In multi-engine support, we do not validate whether two engines use the same configuration because equality rules for two or more engines have not yet been defined (for example, in Postgres, two engines must never use the same slot). We only check if the ID assigned to the capture group is the same.

## What's next

The related documentation for the multi engine support


Naros:
Hi @kmos, I still need to review the code, but I wanted to understand if there were any specific reasons why you chose to go with method-level `@CaptureGroup` annotations rather than class-level?

Using your `SnapshotEventListener` example, I am curious if this is a clearer approach in terms of SoC:
```
public interface SnapshotEventListener {
  void snapshot(SnapshotEvent event);
}

@ApplicationScoped
@CaptureGroup("default")
public DefaultSnapshotEventListener implements SnapshotEventListener {
  @Override
  public void snapshot(@Observes SnapshotEvent event) {
  }
}

@ApplicationScoped
@CaptureGroup("alternative")
public AlternativeSnapshotEventListener implements SnapshotEventListener {
  @Override
  public void snapshot(@Observes SnapshotEvent event) {
  }
}
```

Perhaps it's a bit of a preference, but I find this is a bit clearer where you aren't directly able to mix handlers across engine instances, and it forces you to think about a tertiary object that is injected into both. I get the impression that by using `@CaptureGroup` at the method-level, it potentially encourages god-class patterns.

What do you think?

I am not saying we need to necessarily change it, unless we all agree that this makes sense. But I just wanted to share my perspective and to understand yours.

kmos:
> Hi @kmos, I still need to review the code, but I wanted to understand if there were any specific reasons why you chose to go with method-level `@CaptureGroup` annotations rather than class-level?
> 
> Using your `SnapshotEventListener` example, I am curious if this is a clearer approach in terms of SoC:
> 
> ```
> public interface SnapshotEventListener {
>   void snapshot(SnapshotEvent event);
> }
> 
> @ApplicationScoped
> @CaptureGroup("default")
> public DefaultSnapshotEventListener implements SnapshotEventListener {
>   @Override
>   public void snapshot(@Observes SnapshotEvent event) {
>   }
> }
> 
> @ApplicationScoped
> @CaptureGroup("alternative")
> public AlternativeSnapshotEventListener implements SnapshotEventListener {
>   @Override
>   public void snapshot(@Observes SnapshotEvent event) {
>   }
> }
> ```
> 
> Perhaps it's a bit of a preference, but I find this is a bit clearer where you aren't directly able to mix handlers across engine instances, and it forces you to think about a tertiary object that is injected into both. I get the impression that by using `@CaptureGroup` at the method-level, it potentially encourages god-class patterns.
> 
> What do you think?
> 
> I am not saying we need to necessarily change it, unless we all agree that this makes sense. But I just wanted to share my perspective and to understand yours.

tbh, I agree 100 %. I really don't like to chain annotations in the method parameters. I have proposed this solution due to the jakarta event `observes` that can be used in that way (parameter). To make an example, an user could have multiple observations (`@Observes`) in the same class but with different capture group. Again, I agree 100 % and I found it really ugly but at the same time if we do it by class, we reduce the power of `@Observes` (which isn't necessarily a bad thing).

In the meantime I'll check if it works also by class.

Naros:
Hi @kmos how do you arrive at that reducing the effectiveness of `@Observes`, as I don't follow?

mfvitale:
> Capture Groups: a capture group is a set of working Debezium Servers that attempt to capture log events using the same configuration for a particular source (for example, in Postgres using the same slot). In a Quarkus application, only one engine can be assigned to a capture group. A capture group is defined by an ID.

I don't get the point here. You are talking about `group` that usually mean a set of "things" sharing something in common but then you say that 

> In a Quarkus application, only one engine can be assigned to a capture group. A capture group is defined by an ID.

Is this `limitation` temporary or not?

kmos:
> > Capture Groups: a capture group is a set of working Debezium Servers that attempt to capture log events using the same configuration for a particular source (for example, in Postgres using the same slot). In a Quarkus application, only one engine can be assigned to a capture group. A capture group is defined by an ID.
> 
> I don't get the point here. You are talking about `group` that usually mean a set of "things" sharing something in common but then you say that
> 
> > In a Quarkus application, only one engine can be assigned to a capture group. A capture group is defined by an ID.
> 
> Is this `limitation` temporary or not?



> Hi @kmos how do you arrive at that reducing the effectiveness of `@Observes`, as I don't follow?

Just saying with a class based annotation you can't have multiple observations in a single class, but with a parameter one you can do both. Anyway I understand that force a developer to be more SRP is the right DX to support but I am not sure if it's feasible with jakarta events (`event.select(typeLiteral).fire`). I'll give a look!

quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/CapturingObjectInvokerGenerator.java
```diff
@@ -113,6 +114,20 @@ public GeneratedClassMetaData generate(MethodInfo methodInfo, BeanInfo beanInfo)
                                 });
             }
 
+            try (MethodCreator group = invoker.getMethodCreator("group", String.class)) {
```
Naros: This same logic also exists in `CapturingEventGenerator`, should we have this in a helper class or some super type to minimize code dupe?

kmos:
> Hi @kmos how do you arrive at that reducing the effectiveness of `@Observes`, as I don't follow?

Just saying with a class based annotation you can't have multiple observations in a single class, but with a parameter one you can do both. Anyway I understand that force a developer to be more SRP is the right DX to support but I am not sure if it's feasible with jakarta events (`event.select(typeLiteral).fire`). I'll give a look!

quarkus-debezium-parent/quarkus-debezium-engine-spi-parent/runtime/src/main/java/io/debezium/runtime/CaptureGroup.java
```diff
@@ -0,0 +1,13 @@
...
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+
+package io.debezium.runtime;
+
+import io.debezium.common.annotation.Incubating;
+
+@Incubating
+public record CaptureGroup(String id) {
```
jpechane: I am not sure about the naming. Yo mention that captrue group represent set of Debezium Servers and there can  be only one engine per-group in Quarkus app. Here I'd argue that you are leaking concepts from upper level to lower level.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/configuration/DebeziumConfigurationEngineParser.java
```diff
@@ -0,0 +1,59 @@
...
+                .values()
+                .stream()
+                .filter(configuration -> configuration.groupId().isPresent())
+                .collect(Collectors.toMap(configuration -> configuration.groupId().get(), DebeziumEngineConfiguration.Capturing::configurations));
+
+        /*
+         * SmallRye Config in Quarkus is not able to map configuration on multiple fields if are used raw types
+         * removing the keys with capturing
+         */
+        globalConfiguration.defaultConfiguration().keySet().removeIf(key -> key.startsWith("capturing"));
```
jpechane: Make this constant.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DebeziumThreadHandler.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    /**
+     *
+     * @return the {@link DebeziumContext} associated to the running engine
+     */
+    public static DebeziumContext context() {
+        if (context.get() == null) {
+            LOGGER.warn("Debezium context not initialized, using testing context");
+            return () -> new CaptureGroup("testing");
```
jpechane: Why testing?

quarkus-debezium-parent/quarkus-debezium-engine-spi-parent/runtime/src/main/java/io/debezium/runtime/events/Engine.java
```diff
@@ -0,0 +1,38 @@
...
+import java.lang.annotation.Target;
+
+import jakarta.enterprise.util.AnnotationLiteral;
+import jakarta.inject.Qualifier;
+
+@Qualifier
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.FIELD, ElementType.PARAMETER, ElementType.METHOD })
+public @interface CaptureGroup {
+    String value();
```
jpechane, line 21: Should not be default value specified?

jpechane (commented):
I wonder if there should be `default` group. Maybe the group should not be `String` but  `Optional<String>`. If the "default" config.handler is in place then it would be empty, otherwise a group id will be provided.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DebeziumThreadHandler.java
kmos: The context is created through the dev/prod flow of the extension, thanks to the `getThreadFactory` method (line 40). When the context() method is used in production code, it should always be initialized. However, if the context() method is called without starting the Quarkus extension (such as during integration tests) it will not be initialized. This scenario is primarily relevant for testing purposes.

quarkus-debezium-parent/quarkus-debezium-engine-spi-parent/runtime/src/main/java/io/debezium/runtime/events/Engine.java
kmos: if we specify a default value means something like:

```java
    public void defaultSnapshot(@Observes @CaptureGroup SnapshotEvent snapshot) {
        defaultSnapshotEvents.add(snapshot);
    }
```

which in my opinion hide that you want to see the default group. I mean, even if it's a bit ugly I prefer to be explicit.

quarkus-debezium-parent/quarkus-debezium-engine-spi-parent/runtime/src/main/java/io/debezium/runtime/CaptureGroup.java
kmos: I get your point. We can use at this point something like `engineId` which is at extension level. At debezium server level, we will work on it in the DDD.

WDYT ? @jpechane @Naros @mfvitale

mfvitale: `engineId` sounds good for me

kmos:
> > Capture Groups: a capture group is a set of working Debezium Servers that attempt to capture log events using the same configuration for a particular source (for example, in Postgres using the same slot). In a Quarkus application, only one engine can be assigned to a capture group. A capture group is defined by an ID.
> 
> I don't get the point here. You are talking about `group` that usually mean a set of "things" sharing something in common but then you say that

To address the confusion at this point, it's better to change the name.

> 
> > In a Quarkus application, only one engine can be assigned to a capture group. A capture group is defined by an ID.
> 
> Is this `limitation` temporary or not?

Forget about capture group and think about it as an ID. This ID should be unique because there is a 1:1 association with the agroal datasource which is the same (please take a look to the configuration).

kmos:
> I wonder if there should be `default` group. Maybe the group should not be `String` but `Optional<String>`. If the "default" config.handler is in place then it would be empty, otherwise a group id will be provided.

I would like to have an association 1:1 with the `agroal` datasource which is used. The default datasource under the hood is always named (in our case `<default>`).

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DebeziumThreadHandler.java
jpechane: Can we record this as a comment?

quarkus-debezium-parent/quarkus-debezium-engine-spi-parent/runtime/src/main/java/io/debezium/runtime/events/Engine.java
jpechane: So does it mean that I can either use `@Observes SnapshotEvent snapshot` or `@Observes @CaptureGroup("default") SnapshotEvent snapshot` as equivalent? If yes then it is OK.

kmos: > So does it mean that I can either use `@Observes SnapshotEvent snapshot` or `@Observes @CaptureGroup("default") SnapshotEvent snapshot` as equivalent? If yes then it is OK.

well not :sweat_smile: . If you are in a single engine instance you can use `@Observes SnapshotEvent snapshot`. If you are in a multi-engine configuration you have to specify the engine that you want to observe otherwise you will have all the events (default, alternative and so on). Maybe we can create a fixed annotation (like `DefaultCaptureGroup` or `DefaultEngine` if we change the name) that is always defaulted to "default" which is the first `agroal` datasource. Use `@CaptureGroup` without anything that specify the engine sounds a bit weird to me :eyes: . WDYT? Sound it reasonable? As the annotation is created at extension level and fixed, is feasible...

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DebeziumThreadHandler.java
kmos: sure!

quarkus-debezium-parent/quarkus-debezium-engine-spi-parent/runtime/src/main/java/io/debezium/runtime/events/Engine.java
jpechane: `@DefaultEngine` sounds OK.

quarkus-debezium-parent/quarkus-debezium-engine-spi-parent/runtime/src/main/java/io/debezium/runtime/CaptureGroup.java
kmos: ok, I'll change all the references to `CaptureGroup` to `Engine` and `EngineId`

kmos:
Hi All,

I think that I have addressed mostly all the comments. Please give a further review about annotations and overall DX. Let me know if I have missed something. 

If we are in a *semi-stable* status, I can proceed adding some documentation related to the multi-engine in the relative section. cc: @jpechane @Naros @mfvitale

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DebeziumThreadHandler.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(DebeziumThreadHandler.class.getName());
+
+    /**
+     * Generate a {@link ThreadFactory} and initialize the {@link DebeziumContext} for a give {@link Debezium} engine
+     * @param debezium engine that use generated {@link ThreadFactory}
+     * @return the {@link ThreadFactory} customized for Quarkus
+     */
+    static ThreadFactory getThreadFactory(Debezium debezium) {
+        return runnable -> {
+            int num = manifests
```
mfvitale: I think that here there could be potential issue with concurrent threads. Since you read the value and the update the corresponding key element but the two operation are not atomic. 

```java
manifests.computeIfAbsent(manifest, k -> new AtomicInteger(0))
                        .incrementAndGet();
```

This should be enough since is the key is not present it will be initialized with AtomicInteger(0) and then incremented atomically. If the value is present it is then returned and incremented.

kmos: If it's ok for you, I'll remove all the concurrent data structure in favor of `syncronized`.

quarkus-debezium-parent/quarkus-debezium-engine-parent/runtime/src/main/java/io/quarkus/debezium/engine/DebeziumThreadHandler.java
```diff
@@ -0,0 +1,60 @@
...
+    private static final InheritableThreadLocal<DebeziumContext> context = new InheritableThreadLocal<>();
+    private static final Map<EngineManifest, Integer> manifests = new HashMap<>();
+    private static final Logger LOGGER = LoggerFactory.getLogger(DebeziumThreadHandler.class.getName());
+
+    /**
+     * Generate a {@link ThreadFactory} and initialize the {@link DebeziumContext} for a give {@link Debezium} engine
+     * @param debezium engine that use generated {@link ThreadFactory}
+     * @return the {@link ThreadFactory} customized for Quarkus
+     */
+    static ThreadFactory getThreadFactory(Debezium debezium) {
```
jpechane, line 33: Please use `io.debezium.util.Threads`

kmos: I have done some tests and I cannot use `io.debezium.util.Threads.threadFactory()` out of the box. In `DebeziumThreadHandler` the class is setting the `context` which is a `InheritableThreadLocal` variable. To do that, the callback available in `io.debezium.util.Threads.threadFactory()` (in my test) cannot set the `Context`. We should change the `io.debezium.util.Threads.threadFactory()` in something like:

```java

    public static ThreadFactory threadFactory(Class<?> component, String componentId, String name, boolean indexed, boolean daemon,
                                              Consumer<Thread> callback, Runnable runnable) {
        if (LOGGER.isInfoEnabled()) {
            LOGGER.info("Requested thread factory for component {}, id = {} named = {}", component.getSimpleName(), componentId, name);
        }

        return new ThreadFactory() {
            private final AtomicInteger index = new AtomicInteger(0);

            @Override
            public Thread newThread(Runnable r) {
                StringBuilder threadName = new StringBuilder(DEBEZIUM_THREAD_NAME_PREFIX)
                        .append(component.getSimpleName().toLowerCase())
                        .append('-')
                        .append(componentId)
                        .append('-')
                        .append(name);
                if (indexed) {
                    threadName.append('-').append(index.getAndIncrement());
                }
                LOGGER.info("Creating thread {}", threadName);
                final Thread t = new Thread(() -> {
                    if (runnable != null) {
                        runnable.run();
                    }
                    r.run();
                }, threadName.toString());
                t.setDaemon(daemon);
                if (callback != null) {
                    callback.accept(t);
                }
                return t;
            }
        };
    }
```

To be precise, add the possibility to run some code beforehand:

```java
//[...]
() -> {
                    if (runnable != null) {
                        runnable.run();
                    }
                    r.run();
                }
//[...]
```

and in the `DebeziumThreadHandler` we can have something like:

```java
    static ThreadFactory getThreadFactory(Debezium debezium) {
        return Threads.threadFactory(debezium.connector().getClass(),
                "quarkus-extension",
                debezium.manifest().id(),
                true,
                false, null, () -> context.set(debezium::manifest));
    }
```

WDYT? @jpechane Let me know so I can add these changes to `io.debezium.util.Threads`

jpechane: @kmos Sounds reasonable. Just of course not replace the existing method but add a new one and existing modify to call the new one.
Also please rename `runnable` argument to samethig more descriptive and you can also use `Optional<Runnable>` as the type.

jpechane:
@kmos Applied, thanks!

kmos:
> @kmos Applied, thanks!

I'll open a jira ticket for the related docs!

PR #6733, opened 2025-09-14
DBZ-9452 Correctly skip LogMiner options for file strategy

lgazire:
https://issues.redhat.com/browse/DBZ-9452


Naros:
Applied thanks @lgazire!

PR #6741, opened 2025-09-16
DBZ-9462 Update Docker Rhel system test fixtures and images

AlvarVG:


jpechane:
@AlvarVG Applied, thanks!

PR #6742, opened 2025-09-16
DBZ-9473 Remove unnecessary dependencies from Debezium Engine

kmos:
## Context

Debezium Engine contains many dependencies taken from Kafka libraries. These dependencies are necessary to make the engine able to work with the same API and data that are used in the kafka runtime. Some of them are unnecessary for the correct work of the engine.

## Decision

Remove the unnecessary dependencies coming from `org.apache.kafka:connect-runtime` which are mainly related to `jetty` and `jersey`.


debezium-connector-binlog/pom.xml
```diff
@@ -149,6 +149,11 @@
             <artifactId>kafka_${version.kafka.scala}</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
```
Naros: @jpechane do you see any harm if these test dependencies were pushed up to the parent pom?

debezium-connector-oracle/pom.xml
```diff
@@ -196,6 +196,12 @@
             <artifactId>kafka_${version.kafka.scala}</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>jakarta.validation</groupId>
+            <artifactId>jakarta.validation-api</artifactId>
+            <version>3.0.2</version>
```
Naros: If this dependency remains in the Oracle POM (see my earlier comment to Jiri), let's move the version at least to a property that's maintained either in the Oracle POM or the parent pom.

debezium-connector-binlog/pom.xml
jpechane: Yes, it can be moved to parent pom.

debezium-embedded/pom.xml
```diff
@@ -35,6 +35,58 @@
                     <groupId>log4j</groupId>
                     <artifactId>log4j</artifactId>
                 </exclusion>
+                <exclusion>
```
jpechane, line 58: Please make sure you also build and make basic test of Debezium Server distribution so we are sure it does not affect it.

debezium-connector-binlog/pom.xml
kmos: done even if I don't feel comfortable to move these dependencies in the parent and having them imported in all our modules

debezium-embedded/pom.xml
kmos: these dependencies are already remove from the `quarkus-debezium-engine` which is used inside the last version of Debezium Server (3.7)

kmos:
I revamp a bit the topic based on offline discussion @jpechane @Naros

PR #6744, opened 2025-09-17
DBZ-9469:  Support OpenLineage on Debezium sinks

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-9469

image

TODO: 

- [x] Emit OpenLineage event for output dataset
- [x] Add OpenLIneage for MongoDB sink


jpechane:
@mfvitale Could you please rebase the PR?

debezium-common/src/main/java/io/debezium/config/ConfigurationNames.java
```diff
@@ -14,4 +14,5 @@ public interface ConfigurationNames {
     String MONGODB_CONNECTION_STRING_PROPERTY_NAME = "mongodb.connection.string";
     String TASK_ID_PROPERTY_NAME = "task.id";
     String EXTENDED_HEADERS_ENABLED = "extended.headers.enabled";
+    String CONNECTOR_NAME_PROPERTY = "name";
```
jpechane, line 17:
```suggestion
    String CONNECTOR_NAME_PROPERTY = "connector.name";
```
Would not this be a bit more descriptive?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
```diff
@@ -57,14 +67,19 @@ public class JdbcChangeEventSink implements ChangeEventSink {
     private final RecordWriter recordWriter;
     private final int flushMaxRetries;
     private final Duration flushRetryDelay;
+    private final ConnectorContext connectorContext;
 
-    public JdbcChangeEventSink(JdbcSinkConnectorConfig config, StatelessSession session, DatabaseDialect dialect, RecordWriter recordWriter) {
+    private final ExecutorService executor = Executors.newFixedThreadPool(1, ThreadUtils.createThreadFactory(this.getClass().getSimpleName() + "-%d", false));
```
jpechane: Should not be the `executor` closed at the end?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorTask.java
```diff
@@ -56,6 +72,7 @@ private enum State {
 
     private final AtomicReference<State> state = new AtomicReference<>(State.STOPPED);
     private final ReentrantLock stateLock = new ReentrantLock();
+    private final ExecutorService executor = Executors.newFixedThreadPool(1, ThreadUtils.createThreadFactory(this.getClass().getSimpleName() + "-%d", false));
```
jpechane: Should be closed at the end
Also what is the eadvantage of not using `io.debezium.util.Threads`?

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/oracle/OpenLineageJdbcSinkTestIT.java
```diff
@@ -0,0 +1,25 @@
...
+
+import io.debezium.connector.jdbc.integration.AbstractOpenLineageJdbcSinkTest;
+import io.debezium.connector.jdbc.junit.jupiter.OracleSinkDatabaseContextProvider;
+import io.debezium.connector.jdbc.junit.jupiter.Sink;
+
+@Tag("all")
+@Tag("it")
+@Tag("it-oracle")
+@ExtendWith(OracleSinkDatabaseContextProvider.class)
+public class OpenLineageJdbcSinkTestIT extends AbstractOpenLineageJdbcSinkTest {
```
jpechane, line 19: What is the risk of running test with only single database?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/sink/MongoDbChangeEventSink.java
```diff
@@ -36,14 +48,17 @@ final class MongoDbChangeEventSink implements ChangeEventSink, AutoCloseable {
     private final MongoDbSinkConnectorConfig sinkConfig;
     private final MongoClient mongoClient;
     private final ErrorReporter errorReporter;
+    private final ConnectorContext connectorContext;
+    private final ExecutorService executor = Executors.newFixedThreadPool(1, ThreadUtils.createThreadFactory(this.getClass().getSimpleName() + "-%d", false));
```
jpechane: Same as for other instances

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/sink/MongoDbSinkConnectorTask.java
```diff
@@ -23,14 +33,22 @@
...
+import io.debezium.openlineage.dataset.DatasetMetadata;
 import io.debezium.sink.DebeziumSinkRecord;
 
 public class MongoDbSinkConnectorTask extends SinkTask {
     static final Logger LOGGER = LoggerFactory.getLogger(MongoDbSinkConnectorTask.class);
     private static final String CONNECTOR_TYPE = "sink";
     private MongoDbChangeEventSink mongoSink;
+    private ConnectorContext connectorContext;
+    private DatasetDataExtractor datasetDataExtractor;
+    private final ExecutorService executor = Executors.newFixedThreadPool(1, ThreadUtils.createThreadFactory(this.getClass().getSimpleName() + "-%d", false));
```
jpechane: Same as for the other instances

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -1302,6 +1302,32 @@ public static SnapshotQueryMode parse(String value, String defaultValue) {
...
+     * <strong>Format:</strong> Host and port (e.g., {@code host:port})<br>
+     * <strong>Default:</strong> None<br>
+     * <strong>Importance:</strong> Low</p>
+     *
+     * <p><strong>Example:</strong>
+     * <pre>{@code
+     * openlineage.integration.dataset.kafka.bootstrap.server=localhost:9092
+     * }</pre>
+     */
+    public static Field OPEN_LINEAGE_INTEGRATION_DATASET_KAFKA_BOOTSTRAP_SERVER = Field.create(OpenLineageConfig.OPEN_LINEAGE_INTEGRATION_DATASET_KAFKA_BOOTSTRAP_SERVER)
```
jpechane, line 1323: As there is no default is not this effectivelly mandatory even for non-OpenLineage deployments?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/sink/MongoDbSinkConnectorTask.java
jpechane: I wonder if there should be `Executors` spread around the code or if there should be a central service through which the OpenLineage tasks are submitted.

debezium-common/src/main/java/io/debezium/config/ConfigurationNames.java
mfvitale: No, this is exactly the `name` property from connector configuration. With the source we decided to use the `topic.prefix` since this is for us a logical name but for sinks we do no have this property.

mfvitale:
> I wonder if we can introduce one more namespace at the connector level that would contain the broker configuration. If present then OL integration, Kafka internal schema history and Debezium kafka sink should be able to use it. If not the per compoment config will be checked for.

We could but to me seems a bit more confusing to have this property explicit since is more for internal stuff and more a Kafka Connect things (except for Kafka Sink).

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/oracle/OpenLineageJdbcSinkTestIT.java
mfvitale: Not so much risk it just helps checking the output dataset structure.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/sink/MongoDbSinkConnectorTask.java
mfvitale: Well, effectively the emit is already async, here is more for the places where we need to build input/dataset from records. Let me check if I can improve in someway.

debezium-core/src/main/java/io/debezium/config/CommonConnectorConfig.java
mfvitale: It is not marked as `required`.

mfvitale:
@jpechane I reworked a bit the executor stuff. Let me know if it is better now.

jpechane:
@mfvitale OK, the more I think about the less happy I am about it. My main gripe with it is the issue of executors/threading. We are starting the threads and using async submission regardless of OpnLineage being used.
On the source side we do just easy sync calls that are relatively cheaply handled by `NoOpLineageEmitter` so there is very limited impact. I feel that in this case we are very heavy on the caller side when most deployments will not use provided functionality.

mfvitale:
Ok, so do we want to remove the async call completely and then if there will be performance issues for those using lineage we can improve it. WDYT?

jpechane:
Yes, I think it is going to be a better approach now.

jpechane:
@mfvitale Applied, thanks!

PR #6746, opened 2025-09-18
DBZ-9479 Remove redundant backtick when performing unquoted on field …

leoloel:
…names

https://issues.redhat.com/browse/DBZ-9479


jpechane:
@leoloel Applied, thanks!

PR #6747, opened 2025-09-18
DBZ-9480 Update to Mockito 5.19.0

Naros:
https://issues.redhat.com/browse/DBZ-9480


jpechane:
@Naros Applied, thanks

PR #6748, opened 2025-09-18
DBZ-9481 Update Hibernate to 7.1.0.Final for JDBC sink

Naros:
https://issues.redhat.com/browse/DBZ-9481


jpechane:
Test failures are related

Naros:
There are also Oracle test failures, looking at those locally.

jpechane:
@Naros Applied, thanks!

PR #6749, opened 2025-09-18
DBZ-9041 Support archive destination name precedence

Naros:
https://issues.redhat.com/browse/DBZ-9041

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
@@ -428,9 +428,9 @@ public class OracleConnection extends JdbcConnection {
     }
 
     public boolean validateLogPosition(Partition partition, OffsetContext offset, CommonConnectorConfig config) {
-
-        final Duration archiveLogRetention = ((OracleConnectorConfig) config).getArchiveLogRetention();
-        final String archiveDestinationName = ((OracleConnectorConfig) config).getArchiveLogDestinationName();
+        final OracleConnectorConfig connectorConfig = (OracleConnectorConfig) config;
+        final Duration archiveLogRetention = connectorConfig.getArchiveLogRetention();
+        final String archiveDestinationName = getArchiveLogDestinationByPrecedence(connectorConfig.getArchiveLogDestinationNames());
         final Scn storedOffset = ((OracleConnectorConfig) config).getAdapter().getOffsetScn((OracleOffsetContext) offset);
 
         try {
@@ -868,6 +868,25 @@ public class OracleConnection extends JdbcConnection {
     public String getDatabaseParameterValue(String parameterName) throws SQLException {
         final String query = "SELECT VALUE FROM V$PARAMETER WHERE UPPER(NAME) = UPPER(?)";
         return prepareQueryAndMap(query, ps -> ps.setString(1, parameterName), rs -> rs.next() ? rs.getString(1) : null);
+    }
+
+    public String getArchiveLogDestinationByPrecedence(Set<String> destinationNames) {
+        try {
+            if (destinationNames != null && !destinationNames.isEmpty()) {
+                for (String destinationName : destinationNames) {
+                    if (isArchiveLogDestinationValid(destinationName)) {
+                        LOGGER.debug("Using archive destination '{}'.", destinationName);
+                        return destinationName;
+                    }
+                }
+                return destinationNames.stream().findFirst().orElse(null);
+            }
+            // Fallback to using a random one that's available in query
+            return null;
+        }
+        catch (SQLException e) {
+            throw new DebeziumException("Error while checking validity of archive destination configuration", e);
+        }
     }
 
     private static Scn readScnColumnAsScn(ResultSet rs, String columnName) throws SQLException {
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -895,7 +895,7 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
     private final boolean lobEnabled;
     private final Set<String> logMiningUsernameIncludes;
     private final Set<String> logMiningUsernameExcludes;
-    private final String archiveLogDestinationName;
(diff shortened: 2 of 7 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -976,7 +976,7 @@ public OracleConnectorConfig(Configuration config) {
         this.archiveLogOnlyMode = config.getBoolean(LOG_MINING_ARCHIVE_LOG_ONLY_MODE);
         this.logMiningUsernameIncludes = Strings.setOfTrimmed(config.getString(LOG_MINING_USERNAME_INCLUDE_LIST), String::new);
         this.logMiningUsernameExcludes = Strings.setOfTrimmed(config.getString(LOG_MINING_USERNAME_EXCLUDE_LIST), String::new);
-        this.archiveLogDestinationName = config.getString(ARCHIVE_DESTINATION_NAME);
+        this.archiveLogDestinationNames = Strings.setOfTrimmed(config.getString(ARCHIVE_DESTINATION_NAME), String::new);
```
jpechane: The order of destination will be lost, is it OK?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
```diff
@@ -205,20 +205,18 @@ protected String connectorName() {
         return Module.name();
     }
 
-    private void checkArchiveLogDestination(OracleConnection connection, String destinationName) {
+    private void checkArchiveLogDestination(OracleConnection connection, Set<String> destinationNames) {
         try {
-
-            if (!Strings.isNullOrBlank(destinationName)) {
+            if (destinationNames != null && !destinationNames.isEmpty()) {
```
jpechane: How about a new utility method in `io.debezium.util.Collect` class?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
```diff
@@ -870,6 +870,25 @@ public String getDatabaseParameterValue(String parameterName) throws SQLExceptio
         return prepareQueryAndMap(query, ps -> ps.setString(1, parameterName), rs -> rs.next() ? rs.getString(1) : null);
     }
 
+    public String getArchiveLogDestinationByPrecedence(Set<String> destinationNames) {
+        try {
+            if (destinationNames != null && !destinationNames.isEmpty()) {
```
jpechane: See below

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
```diff
@@ -205,20 +205,18 @@ protected String connectorName() {
...
+                final String destinationName = connection.getArchiveLogDestinationByPrecedence(destinationNames);
                 if (!connection.isArchiveLogDestinationValid(destinationName)) {
```
jpechane: Is not this check superfluous now? Or if it is in place then it should throw an error?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
```diff
@@ -870,6 +870,25 @@ public String getDatabaseParameterValue(String parameterName) throws SQLExceptio
         return prepareQueryAndMap(query, ps -> ps.setString(1, parameterName), rs -> rs.next() ? rs.getString(1) : null);
     }
 
+    public String getArchiveLogDestinationByPrecedence(Set<String> destinationNames) {
```
jpechane: The name implies some ordering but was not it lost above?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
```diff
@@ -870,6 +870,25 @@ public String getDatabaseParameterValue(String parameterName) throws SQLExceptio
...
+        try {
+            if (destinationNames != null && !destinationNames.isEmpty()) {
+                for (String destinationName : destinationNames) {
+                    if (isArchiveLogDestinationValid(destinationName)) {
+                        LOGGER.debug("Using archive destination '{}'.", destinationName);
+                        return destinationName;
+                    }
+                }
+                return destinationNames.stream().findFirst().orElse(null);
```
jpechane: When this is this branch valid?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
Naros: Doh, thanks for the catch :pray:

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
Naros: What I am attempting to do is to avoid a behavior change. 

Before this change, if the archive destination name was not valid, we logged a warning and let the connector proceed, which would eventually lead to an error in later checks. So, here I take the first name, since not all of them are valid, and use it. The `orElse` part was just there as an alternative to using `.get()` .

So in short, this is the equivalent of setting `archive.destination.name` with a single name that's not valid.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
Naros: Given that the connection method can return an invalid name, unfortunately, it is not. 

Overall, I'm not super happy with the logic split between the warnings in the task and the bulk of the logic in the connection class. I started with a single method in the Connection class that included the warnings. I was concerned that calling the Connection class method multiple times would lead to repeated warnings when the process was executed, which I didn't like.

I'm open to other ideas, if you have any.

Naros:
@jpechane So I chose to move all the logic into `ArchiveDestinationNameResolver` and exposed this on the config class. This provided a solid way to do some Mockito unit tests quite easily and keeps the logic centralized. Let me know what you think of this approach.

Naros:
Test failures are unrelated.

jpechane:
@Naros Good job, applied!

PR #6754, opened 2025-09-22
DBZ-7772 Retry errors in JdbcChangeEventSink.hasTable

coekie:
Retries on errors when flushing were added in DBZ-7291. But when a database connection breaks, and the next received record is the first record for that table, then no retry was attempted.

Concretely, it would then fail at:
JdbcChangeEventSink.getRecordsToFlush calling
JdbcChangeEventSink.getOrCreateBuffer through computeIfAbsent calling checkAndApplyTableChangesIfNeeded calling hasTable.

This applies the same retry mechanism from flushBufferWithRetries (DBZ-7291) to hasTable.

There are more SQL calls in checkAndApplyTableChangesIfNeeded that could fail, in createTable and alterTableIfNeeded. This change does not add retries to those because seems more tricky/dangerous; and right after successfully executing another query (in hasTable) transient connection errors are far less likely.

For context: I'm submitting this improvement because in our setup the jdbc connector is crashing with this error daily.

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
@@ -16,6 +16,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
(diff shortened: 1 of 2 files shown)


Naros:
Applied, thanks @coekie

PR #6756, opened 2025-09-22
Add StrimziKafkaCluster to the tests and replace current impl

see-quick:
This PR updates tests to use the `StrimziKafkaCluster` [1].  Many features could be handy, such as multi-node setup, OAuth and in the next release, dedicated roles for nodes. 

[1] - https://github.com/strimzi/test-container


gunnarmorling:
Hey @see-quick! Debezium's `KafkaCluster` runs solely in-process, which makes it very fast and thus is great for testing. Would this change cause tests to run in a container which previously didn't require Docker/podman? If so, do we know what's the impact of test execution time? Thanks for clarifying!

see-quick:
> Hey @see-quick! Debezium's `KafkaCluster` runs solely in-process, which makes it very fast and thus is great for testing. Would this change cause tests to run in a container which previously didn't require Docker/podman? If so, do we know what's the impact of test execution time? Thanks for clarifying!

Hi, generally when you start a cluster with `1-node`, it usually takes `2 seconds` to start. The overall design of those tests is that you always deploy the cluster before the whole test execution, so it should not be a huge deal. (Number of suites affected * (2 seconds + cleaning phase)  => result. 

> Would this change cause tests to run in a container which previously didn't require Docker/podman?

Correct.

gunnarmorling:
Gotcha, thanks. So what's the benefit of this change then? I'm concerned we're slowing down tests (and they take long time to run already) without a clear upside.

jpechane:
@gunnarmorling Well, the reason is easy. Rewrite from Kafka 3.x support to Kafka 4.x support was not exactly straightforward and we dropped some capabilities in the process to make sure we do it relatively fast.
This PR removes the dependency on Kafka internal APIs and while there be a slight slowdown (negligible as it is not used in many tests) it is better from the mainitenanace PoV.
In fact I was thinking about rewriting the class for TC myself so in this regard we just re-use what is already ready made.

gunnarmorling:
@jpechane, ah yes, that's the context I was looking for. Wasn't aware of the fact that `KafkaCluster` isn't trivial to migrate across versions. Makes sense then, thanks!

see-quick:
btw, what is the easiest way to run those integration tests? Do I need to run some databases manually and then run tests, or can those tests also create databases? Thanks :)

jpechane:
@see-quick For the newcomer the easiest is probably `mvnw install -DskipTests -Dit.test=... -Passembly`

see-quick:
> @see-quick For the newcomer the easiest is probably `mvnw install -DskipTests -Dit.test=... -Passembly`

I have managed to run those tests, but I can see I did need to explicitly set my `DOCKER_HOST` and `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE` to point at the `podman` socket. Those tests are not compatible with podman? 

Also, I can see that you are using some docker plugin, so I assume I need to somehow also override the default settings to use podman (hope that somehow works) 🤞.

see-quick:
> > @see-quick For the newcomer the easiest is probably `mvnw install -DskipTests -Dit.test=... -Passembly`
> 
> I have managed to run those tests, but I can see I did need to explicitly set my `DOCKER_HOST` and `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE` to point at the `podman` socket. Those tests are not compatible with podman?
> 
> Also, I can see that you are using some docker plugin, so I assume I need to somehow also override the default settings to use podman (hope that somehow works) 🤞.

Okay, it seems that when using with `podman`, one has to pay attention to set `DOCKER_HOST` to a direct link and not a symlink (e.g., `unix:///var/.........podman/podman-machine-default-api.sock` Now it works.

see-quick:
> > > @see-quick For the newcomer the easiest is probably `mvnw install -DskipTests -Dit.test=... -Passembly`
> > 
> > 
> > I have managed to run those tests, but I can see I did need to explicitly set my `DOCKER_HOST` and `TESTCONTAINERS_DOCKER_SOCKET_OVERRIDE` to point at the `podman` socket. Those tests are not compatible with podman?
> > Also, I can see that you are using some docker plugin, so I assume I need to somehow also override the default settings to use podman (hope that somehow works) 🤞.
> 
> Okay, it seems that when using with `podman`, one has to pay attention to set `DOCKER_HOST` to a direct link and not a symlink (e.g., `unix:///var/.........podman/podman-machine-default-api.sock` Now it works.

Now I am stucked at 
```
[ERROR] DOCKER> Error occurred during container startup, shutting down...
[ERROR] DOCKER> I/O Error [Unable to create container for [debezium/mysql-server-test-database] : {"cause":"image not known","message":"no such image: localhost/debezium/mysql-server-test-database:latest: image not known","response":404} (Not Found: 404)]
```
I have succesfully done with `DOCKER> [debezium/mysql-server-test-database:latest]: Built image sha256:53bde` but eventually error with 404.  I don't know if that's a problem with podman or what 🤷 .

jpechane:
So, if you list the images you have available locally, what tags contain string `mysql-server-test-database`?

see-quick:
> So, if you list the images you have available locally, what tags contain string `mysql-server-test-database`?

```
localhost/debezium/mysql-server-test-database   latest   53bdec3bbd83  19 hours ago   612 MB
```
i.e., `latest`

see-quick:
> > So, if you list the images you have available locally, what tags contain string `mysql-server-test-database`?
> 
> ```
> localhost/debezium/mysql-server-test-database   latest   53bdec3bbd83  19 hours ago   612 MB
> ```
> 
> i.e., `latest`

I have found the issue. That plugin builds images by default for `x64` 🙄. Do you also test on ARM (I can see that in each  `pom.xml` you do not specify arches (e.g.,
```
<buildx>
      <platforms>
                <platform>linux/amd64</platform>
                <platform>linux/arm64</platform>
                              ...
      </platforms>
</buildx>
```

UPDATE: I need to make a hack to directly use `<from>${mysql.server.image.source}:${version.mysql.server}-aarch64</from>` aarch64 base image :) as `podman buildx` is not compatible with `docker buildx`.

see-quick:
The build is passing (ARM). I have also modified and removed some classes, and PR should be ready for review :) 

> [!NOTE]
I didn't fully migrate to just StrimziKafkaCluster, as there were places where you used just DebeziumKafkaContainer (I have replaced it with our StrimziKafkaContainer). But for the future, we should migrate to use `StrimziKafkaCluster`.

jpechane:
It seems to me that  MySQL, MariaDB and JDBC failures are related, could you please take a look?

jpechane:
@see-quick Thank you very much, we are almost there!

* Please use `DBZ-9512` prefix instead of `[test]` in all commits for all PRs
* Informix and Db2 failures are fixed in theappropriate repositories so thee are not of concern
* MySQL failure is a transient one
* SQL Server failure is cause by this PR. I tried tests locally and they work for `main` but fails with this PR, also failure is

```
java.lang.ClassCastException: class org.apache.logging.slf4j.Log4jLogger cannot be cast to class ch.qos.logback.classic.Logger (org.apache.logging.slf4j.Log4jLogger and ch.qos.logback.classic.Logger are in unnamed module of loader 'app')
	at io.debezium.junit.logging.LogInterceptor.<init>(LogInterceptor.java:52)
	... 33 more
```

Which implies there can be a (transitive?) dependency clash on the classpath.

see-quick:
> @see-quick Thank you very much, we are almost there!
> 
> * Please use `DBZ-9512` prefix instead of `[test]` in all commits for all PRs
> * Informix and Db2 failures are fixed in theappropriate repositories so thee are not of concern
> * MySQL failure is a transient one
> * SQL Server failure is cause by this PR. I tried tests locally and they work for `main` but fails with this PR, also failure is
> 
> ```
> java.lang.ClassCastException: class org.apache.logging.slf4j.Log4jLogger cannot be cast to class ch.qos.logback.classic.Logger (org.apache.logging.slf4j.Log4jLogger and ch.qos.logback.classic.Logger are in unnamed module of loader 'app')
> 	at io.debezium.junit.logging.LogInterceptor.<init>(LogInterceptor.java:52)
> 	... 33 more
> ```
> 
> Which implies there can be a (transitive?) dependency clash on the classpath.

Fixed. MongoDB jobs have some issues with pulling PostgreSQL that are not related to this PR.

jpechane:
@see-quick Applied, thanks a lot!

PR #6757, opened 2025-09-22
DBZ-9493 Declare missing transform/sink in META/INF

mfvitale:
closes: https://issues.redhat.com/browse/DBZ-9493


Naros:
Applied, thanks @mfvitale

PR #6732, opened 2025-09-13
DBZ-9427: Documentation for guardrail configs

rajdangwal:
Ref: https://github.com/debezium/debezium/pull/6718

PR #6760, opened 2025-09-23
DBZ-9427: Renamed config to use collections instead of tables

rajdangwal:
Ref: https://github.com/debezium/debezium/pull/6732#discussion_r2366827245


jpechane:
@rajdangwal Applied, thanks!

PR #6761, opened 2025-09-24
DBZ-8510 Use MERGE syntax for PostgreSQL 15+ targets

### LINKED ISSUE #1024: Add a Merge Into insert mode to handle PK changes in jdbc sink [DBZ-8510]
Migrated from [DBZ-8510](https://issues.redhat.com/browse/DBZ-8510)

Even with the upsert insert mode with Postgres, when a PK changes, we end with two lines in the database.

Instead of using INSERT INTO ON CONFLICT, we can use MERGE INTO which is fully supported in recent PG Versions.

Requirement to use MERGE INTO
 * Recent PG version
 * Old the fields used in the key must be present in the record value to extract the old key from the before field
 * Not compatible with flat records

Work in progress in this branch: [https://github.com/debezium/debezium-connector-jdbc/compare/main...bpaquet:debezium-connector-jdbc:merge_into]

Naros:
Fixes debezium/dbz#1024

https://issues.redhat.com/browse/DBZ-8510

Supersedes #6493


jpechane (changes requested):
Test failure is related, it seems the SQL is not syntatically correct

Naros:
Thanks @jpechane I'll take a look.  I had ran a few local tests, but maybe there is a corner case I missed.

Naros:
Hi @jpechane I've reached out to Dave because the behavior of MERGE in PG is vastly different from other dialects, and it requires that we express all values using cast clauses, which will severely bloat the SQL statement, which we don't have to do when using the current behavior. Until I hear back, moving this to draft.

bpaquet:
Hello, please be aware that `merge into` is slower and `insert into on conflict`. That's why I created a different strategy.

fuyar:
Could we specify the join columns ? That's really the added value here compared to `INSERT ON CONFLICT`.

Also to backup Bertrand's point, IIRC you can't do batched inserts using MERGE INTO (unless you create a CTE where you put all the values but that's not out of the box).

Naros:
> Hello, please be aware that `merge into` is slower and `insert into on conflict`. That's why I created a different strategy.

I was unaware of the performance difference. Do you have any documentation or references that illustrate the extent of the variance we're discussing? 

If it's too great, then perhaps rather than a strategy, we consider a dialect-specific toggle, as we've already done for `dialect.postgres.postgis.schema` and `dialect.sqlserver.identity.insert`, that would allow you to specify which statement should be used.

Naros:
> Could we specify the join columns ? That's really the added value here compared to `INSERT ON CONFLICT`.

This isn't clear to me. Could you elaborate?

> Also to backup Bertrand's point, IIRC you can't do batched inserts using MERGE INTO (unless you create a CTE where you put all the values but that's not out of the box).

If you're referring to the fact that the driver executes each MERGE statement as a separate statement and doesn't inline values like it does with the `INSERT` variant, I've at least read the same. 

However, this presents a unique opportunity to reach out to the JDBC driver development team and see if we can help nudge some work in this space to improve performance, bringing it closer in line with the INSERT variant.

One of the most significant issues currently is how the JDBC driver parses the MERGE statement, as it does not handle inference in the same efficient manner as other dialects. I'd ideally like to see that improved as well. The team is investigating this matter, and I'll also inquire about the performance.

Naros:

```
git commit -s
```

PR #6765, opened 2025-09-24
DBZ-9500 No need to snapshot when the table has no columns.

leoloel:


jpechane (approved):
LGTM, this can be merged when a test is provide for one of the connectors.

jpechane:
@leoloel Applied, thanks!

PR #6776, opened 2025-09-29
DBZ-9508 Added support for SDO_GEOMETRY columns in jdbc-connector.

piatekmarcin:
Added support for SDO_GEOMETRY using a similar approach as in GeometryType of PostgreSQL dialect.
Ingestion of WKB achieved using the Oracle function SDO_UTIL.FROM_WKBGEOMETRY(TO_BLOB( ? ), ? ).


debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/oracle/GeometryType.java
```diff
@@ -0,0 +1,43 @@
...
+import io.debezium.data.geometry.Geometry;
+import io.debezium.sink.SinkConnectorConfig;
+import io.debezium.sink.column.ColumnDescriptor;
+
+public class GeometryType extends AbstractGeoType {
+
+    public static final JdbcType INSTANCE = new GeometryType();
+
+    static final String GEO_FROM_WKB_FUNCTION = "SDO_UTIL.FROM_WKBGEOMETRY(TO_BLOB(?), ?)";
+    private static final String TYPE_NAME = "MDSYS.SDO_GEOMETRY";
```
Naros, line 18: We should consider introducing a `dialect` specific configuration property that defines the tablespace where the spatial components are installed. By default Oracle will place them in `MDSYS`; however, the DBA can move these objects to a unique, separate tablespace if they want.  

See the `postgis` property for PostgreSQL as an example.

We should also include the dialect configuration property in the jdbc sink docs.

piatekmarcin: Hi, Oracle Spatial, unlike PostGIS, cannot be installed in different schema than MDSYS. Technically it's possible but such a configuration would be broken. It is possible to move physical location of MDSYS schema to dedicated tablespace but it will be still same schema. Tablespace is not used in DDL or DML references to Spatial types. Only schema is used and it will be always MDSYS.

piatekmarcin:
Hi, see my comment to the inline suggestion. I will add some tests tomorrow.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/oracle/GeometryType.java
Naros: Ah, I misunderstood the documentation; thanks for the clarification.

Naros:
@piatekmarcin I have fixed the test case so it passes, but I am not really happy about the fact that with certain versions the SRID is populated and in other versions it isn't due to the bindable values for `FROM_WKBGEOMETRY`.

I wonder if there is a better way in which we can construct the `oracle.sql.STRUCT` data type directly with the values and bind the `STRUCT` object instead? That might allow us to guarantee that the SRID is always set regardless of which database version is used.  wdyt?

If we stick with the current flavor for now we need to document the version variances around SRID, that it requires Oracle 19.26 or later of Oracle 19 or Oracle 23+ for the SRID to be stored.

We should also consider adding a geometry end-to-end integration test at least between something like PostgreSQL as the source and Oracle as the target just to verify that the mappings work as expected across vendors.

piatekmarcin:
Hi @Naros, thanks for picking this up further. I've searched docs and apparently `SDO_GEOMETRY` type has constructor, which accepts WKB and SRID, so we can use constructor instead of `FROM_WKBGEOMETRY`.
I've checked and it works on 19.3.0, and is also documented even in release 18 and 12.2

Below test from 19.3.0:

```
SQL> select SDO_GEOMETRY(to_blob('0101000000000000000000F03F0000000000000040'), 4326) from dual;

SDO_GEOMETRY(TO_BLOB('0101000000000000000000F03F0000000000000040'),4326)(SDO_GTY
--------------------------------------------------------------------------------
SDO_GEOMETRY(2001, 4326, SDO_POINT_TYPE(1, 2, NULL), NULL, NULL)

```

Naros:
@piatekmarcin can you take a look at the latest changes to see if they work for you?

piatekmarcin:
Hi @Naros, thanks, it works perfectly.

Naros:
Ok, then we should try and add a few E2E tests that use Postgis as source and Oracle as target for geometry to make sure that cross vendor works and then we're good to go.  Do you think you can add that @piatekmarcin ?

piatekmarcin:
I can add it @Naros

piatekmarcin:
Hi @Naros, I've added e2e tests and some handling of `SDO_GEOMETRY` for tests, however I could not execute them due to some issues with test containers (connect:nightly). Could you crosscheck if it is working?

Naros:
@piatekmarcin did you take a look at `db2gse.ST_Geometry` for Db2 or `geometry::STGeomFromWKB` for SQL Server by any chance? If not, I can look into it.

piatekmarcin:
@Naros, no I haven't looked into it, but based on documentation, it seems like it should be doable in similar way as for Oracle and PostGIS.

Naros:
> @Naros, no I haven't looked into it, but based on documentation, it seems like it should be doable in similar way as for Oracle and PostGIS.

Ok, I have almost a full working solution across all dialects, standby for my commits.

piatekmarcin:
@Naros, I just noticed two more issues in values used for tests and posted fixes

Naros:
Hi @jpechane and @piatekmarcin, I've stumbled on a small nuance with EWKB/WKB when crossing vendor boundaries.

If we assume a basic `POINT` geometry type, e.g.
```
'SRID=4326;POINT (8 19)'::postgis.geography
```

Different database vendors have different rules when it comes to the coordinate axis order. For example, the following chart describes the rules when using SRID=4326 (WGS 84 (GPS coordinates, most common for global lat/lon data)):

|Database/Engine|Format|
|---|---|
|Postgres|Uses `latitude longitude` or `y x` format as it respects EPSG for SRID=4326|
|SQL Server|Uses `latitude longitude` or `y x` format as it respects EPSG for SRID=4326|
|Oracle|Uses `longitude latitude` or `x y` format as it ignores EPSG|
|MySQL|Uses `longitude latitude` or `x y` format as it ignores EPSG|

When you replicate and insert the EWKB/WKB value where the source/target are the same or respect the same rules, then when you read the value from the target, the coordinates will align with the values as inserted on the source. But when you replicate across vendors where this is not the case, the coordinates are inverted. This is the case when you replicate changes from PostGIS to Oracle, as Oracle treats the coordinates as `(x, y)` even if PostGIS serialized the EWKB as `(y, x)`.

Unfortunately, the EWKB/WKB format does not specify the axis order. 

While we could try and bake this into the JDBC sink or on the source, it may be easier to consider introducing a custom transform that can be placed on either side of the pipeline:

```json
"transforms": "flip",
"transforms.flip.type": "io.debezium.transforms.FlipGeometryCoordinates",
"transforms.flip.srids": "4326,3857,4269", // this is optional but these would be the defaults
```

The transform would look for `io.debezium.data.Geometry` fields in the before/after or in the flattened event, and based on the configured `srids` (which would default to `4326,3857,4269` but can be overwritten for flexibility), if the field's srid matches, the coordinates in the points and coordinates lists would be flipped.

This would mean:

- The JDBC sink writes the EWKB/WKB data as it's provided with no assumptions about order.
- Replicating EWKB/WKB within the same vendor or vendors with the same EPSG rules, the transform can be omitted.

If you both agree with this approach, I'll add this to this PR tomorrow. Otherwise, if you have any other suggestions, I'm all ears on what you think might work better.

One additional question wrt dependencies, @jpechane is there any issue that you're aware by using EPL-2.0, BSD-3 licensed software? I'm specifically referring to [JTS](https://github.com/locationtech/jts).

jpechane:
@Naros Unbelievable :-(.
I like your approach to the solution, the SMT is definitely way to go. WRT EPL and BSD, IIRC we already have such in our deps so there is no change. One question though, do we really need anothere dependency? Can't we just operate directly at the bytes level to do the switch?

piatekmarcin:
@Naros, good find. SMT is perfectly fine for me. Moreover, Oracle doesn’t support EWKB, so in case EWKB would appear in Connect, then SRID needs to be extracted and additional transformation to WKB is required.

Naros:
> @Naros Unbelievable :-(. I like your approach to the solution, the SMT is definitely way to go. WRT EPL and BSD, IIRC we already have such in our deps so there is no change. One question though, do we really need anothere dependency? Can't we just operate directly at the bytes level to do the switch?

That should be doable @jpechane.

> Moreover, Oracle doesn’t support EWKB, so in case EWKB would appear in Connect, then SRID needs to be extracted and additional transformation to WKB is required.

I've currently solved this by forcing the conversion at bind time from EWKB to WKB if the dialect only uses WKB and the value is provided as an EWKB. Aside from the coordinate flipping that's needed, I'm close. :crossed_fingers:

Naros:
@maver1ck I've incorporated the new Spatial API from DBZ-9556. 

The tests will fail here until we merge https://github.com/debezium/debezium/pull/6827 (DBZ-9556) and rebase this after the `quay.io/debezium/connect:nightly` image is built with the new SwapGeometryCoordinates SMT. But I've tested locally with a local baking of the connect image.

Naros:
:warning: Waiting on Nighty connect image to be built before running tests.

PR #6784, opened 2025-09-30
[doc] Fix URL in link to release notes from Extended strings topic

roldanbob:
A note in the Oracle connector Extended Strings topic linked to the Tech Preview release note for the feature. In the 3.2.3 release, the feature is promoted to GA. This change updates the link to reflect that change.


roldanbob:
Backported to 3.2 Because the target content reflects a point in time, this change is not strictly needed in `main`, but it's benign. I'm not backporting the change to 3.3.

PR #6791, opened 2025-10-01
DBZ-9521 Flush commit queue on non-LOB tables

Naros:
https://issues.redhat.com/browse/DBZ-9521

### Changes

When LOB enabled tables are observed, their LOB data is stitched together over the course of the transaction; however, when a table without an LOB column is seen, such events are immediately dispatched. But this dispatch would lead to event sequences being out of order.

This change is designed that if an immediate dispatch is necessary that any pending changes in the queue are flushed by event order first and only then will the non-LOB table be emitted, preserving event order.

This fixes a regression introduced by https://issues.redhat.com/browse/DBZ-8060


jpechane:
Part of the test failures can be fixed by https://github.com/debezium/debezium/pull/6798
I recommend to rebase the PR when merged but I wonder if `OracleConnectorIT.shouldPopulateCommitScnAndTimestampInSourceInfoBlock` is related to this.

Naros:
Applied a fix for the one problem.  Will rerun the tests & merge if all looks good. Thanks @jpechane

vjuranek (approved):

vjuranek:
Test failures not related.

vjuranek:
Merged, thanks @Naros !

PR #6796, opened 2025-10-03
DBZ-9517 Update to Ehcache 3.11.1

Naros:
https://issues.redhat.com/browse/DBZ-9517


jpechane:
It seems the test failures are caused by https://github.com/debezium/debezium/pull/6756

PR #6805, opened 2025-10-06
DBZ-9532 resolves metric name clash with multiple connectors

msillence:


Naros:
Hi @msillence, I'm afraid it's expected behavior that metrics should clash if you deploy two connectors with the exact same `topic.prefix`. This property is designed to be unique across all connector deployments.

msillence:
> Hi @msillence, I'm afraid it's expected behavior that metrics should clash if you deploy two connectors with the exact same `topic.prefix`. This property is designed to be unique across all connector deployments.

It seems a shame when we have a known unique name for the connector that we chose to rely on a field that does not have to be unique. I believe the topic.prefix is used as named as the prefix for the topics and it's not obvious why this needs to be unique or should be forced to be so.

The fix is reasonably straightforward only the "name" is a confusing term there being 
1/ a "connectorName" which is the implementing class 
and 
2/ the "name" of the connector in the json we use to create it

Strangely I can't find anything exposing the "name" though we do have a constant to extract it.

jpechane:
@msillence `connector.name` from Kafka Connect is (was?) never accessible form connectors in the past. As a workaround there was a `database.server.name` config option later renamed to `topic.prefix` that represents logical connector name and was always intended to be unique.

msillence:
> @msillence `connector.name` from Kafka Connect is (was?) never accessible form connectors in the past. As a workaround there was a `database.server.name` config option later renamed to `topic.prefix` that represents logical connector name and was always intended to be unique.

Sadly it isn't unique in our configurations and unlike the connector.name there's no uniqueness validation so we didn't know this was the source of the issue. We've ignored the metrics errors until recently where we needed to fix it for alerting purposes.

I guess if you don't want this fix then you can close it, we can use `custom.metric.tags` instead. It seems a shame as there is a unique name that could easily be used now.

jpechane:
@msillence How about introducing some special value(s) for custom tag mapping so it will look like for example
`cname=${param:topic.prefix}`
This would add a tag named `cname` with value taken from connector parameter named `topic.prefix`.

mfvitale:
> @msillence `connector.name` from Kafka Connect is (was?) never accessible form connectors in the past. As a workaround there was a `database.server.name` config option later renamed to `topic.prefix` that represents logical connector name and was always intended to be unique.

@jpechane Seems that the `name` configuration is accessible in Kafka Connect. I used it [here](https://github.com/debezium/debezium/blob/767a41f71858182c3f506bb9b4b63aaed92d10c0/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorTask.java#L111)

KC reference https://github.com/apache/kafka/blob/7527a8bac088ac551a34c7ef2447588f70d8ec7e/connect/runtime/src/main/java/org/apache/kafka/connect/runtime/rest/resources/ConnectorsResource.java#L390

jpechane:
@msillence Given the information provided by @mfvitale, wdyt about the advanced tag mapping? In your case the connectr config would then look like `cname=${param:name}`.

Naros:

Hi @msillence, please sign off all commits with:

```
git commit -s
```

PR #6813, opened 2025-10-08
SchemaUtil does not escape JSON strings correctly

hamnis:
We fix this by using jackson ObjectMapper.
This will correcly escape the JSON string.

Add tests that tests parseability and escaping.


vjuranek (approved):
LGTM. Please modify commit message and add corresponding Jira issue ID at the beginning of the commit message.

hamnis:
This has been done

vjuranek:
Merged, thanks @hamnis !

PR #6814, opened 2025-10-08
Refactor CdcSourceTaskContext

mfvitale:
In anticipation of a refactoring for the OpenLinage to remove `runId` sharing through static field, the idea is to put it in the `CdcSourceContext` so that it can then be passed via record headers and used in the SMT. 

The problem is that currently the sub-classes of `CdcSourceContext` have a lot of other information like the `schema`. This is not only a design flaw but also create a chicken egg problem problem since the `runId` needs to be used in the `schema` but then the `schema` is required by  `CdcSourceContext`

This refactor is a preliminary step to remove the design flaws and prepare the `CdcSourceContext` to be used as a pure context.


mfvitale:
@jpechane , @Naros , @kmos feedback are really appreciated.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/XStreamStreamingChangeEventSourceMetrics.java
```diff
@@ -18,7 +18,8 @@
 public class XStreamStreamingChangeEventSourceMetrics extends AbstractOracleStreamingChangeEventSourceMetrics {
     public XStreamStreamingChangeEventSourceMetrics(CdcSourceTaskContext taskContext,
                                                     ChangeEventQueueMetrics changeEventQueueMetrics,
-                                                    EventMetadataProvider metadataProvider) {
-        super(taskContext, changeEventQueueMetrics, metadataProvider);
+                                                    EventMetadataProvider metadataProvider,
+                                                    java.util.function.Supplier<java.util.Collection<? extends io.debezium.spi.schema.DataCollectionId>> dataCollectionsSupplier) {
```
Naros: In some metrics classes we imported `Supplier`, `Collection`, and `DataCollectionId`, but in others we did not. Could we be consistent? I didn't look at all the usages across the metrics classes but I suspect there are likely no reason we can't import the classes for readability?

mfvitale: Yes, sure. This draft PR is mainly to validate the idea and if its ok I'll adjust the nitpiks.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/metrics/BinlogStreamingChangeEventSourceMetrics.java
```diff
@@ -44,11 +43,12 @@ public class BinlogStreamingChangeEventSourceMetrics<T extends BinlogDatabaseSch
 
     public BinlogStreamingChangeEventSourceMetrics(BinlogTaskContext<T> taskContext,
                                                    ChangeEventQueueMetrics changeEventQueueMetrics,
-                                                   EventMetadataProvider eventMetadataProvider) {
-        super(taskContext, changeEventQueueMetrics, eventMetadataProvider);
-        this.client = taskContext.getBinaryLogClient();
+                                                   EventMetadataProvider eventMetadataProvider,
+                                                   java.util.function.Supplier<java.util.Collection<? extends io.debezium.spi.schema.DataCollectionId>> dataCollectionsSupplier,
```
jpechane: Would not in this case be better to define a specific functional interface?

mfvitale: Good idea.

mfvitale:
Changes completed. I have updated the other connectors to reflect the core changes but since the branch name is not `DBZ` the maven cross check is not running.

PR #6817, opened 2025-10-09
DBZ-9376 Remove PostgreSQL 12, baseline is now 14

Naros:
https://issues.redhat.com/browse/DBZ-9376

Sibling PR in container repository - https://github.com/debezium/container-images/pull/470


Naros:
Thanks @jpechane, but I think in 3.4 the images are all based on PG18, but I'll give one more pass.

jpechane:
The cleanup needs to be done in https://github.com/debezium/container-images/tree/main/postgres

jpechane:
@Naros Applied, thanks!

PR #6818, opened 2025-10-09
DBZ-9385 Document escape requirements for Quarkus expansion

Naros:
https://issues.redhat.com/browse/DBZ-9385

### DIFF
--- a/documentation/modules/ROOT/pages/operations/debezium-server.adoc
+++ b/documentation/modules/ROOT/pages/operations/debezium-server.adoc
@@ -64,8 +64,10 @@ This means that the application can be configured from disparate sources like co
 [NOTE]
 ====
 Configuration properties that contain the `$` character are normally interpreted by Quarkus for property expansion.
-If a property value contains a `++${...}++` syntax and should not be expanded, the `$` character must be escaped using `++$$++` so that the value is written as `++$${name}++`.
-This syntax guarantees that the property value is passed to {prodname} as `++${name}++` without expansion.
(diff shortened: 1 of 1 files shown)


jpechane:
@Naros `2.5` label?

Naros:
> @Naros `2.5` label?

I must have not paid attention, my intent was 3.4.

documentation/modules/ROOT/pages/operations/debezium-server.adoc
```diff
@@ -61,6 +61,13 @@ explained here: xref:{link-oracle-connector}#obtaining-oracle-jdbc-driver-and-xs
 {prodname} Server uses https://github.com/eclipse/microprofile-config[MicroProfile Configuration] for configuration.
 This means that the application can be configured from disparate sources like configuration files, environment variables, system properties etc.
 
+[NOTE]
+====
+Configuration properties that contain the `$` character are normally interpreted by Quarkus for property expansion.
+If a property value contains a `++${...}++` syntax and should not be expanded, the `$` character must be escaped using `++$$++` so that the value is written as `++$${name}++`.
+This syntax guarantees that the property value is passed to {prodname} as `++${name}++` without expansion.
```
roldanbob:
```suggestion
If a property value includes a `$` character as a prefix, such as `++${my.property}++`,  the Microprofile configuration parser determines that the value is a  https://quarkus.io/guides/config-reference#property-expressions[property expression] and attempts to expand it.

To prevent the parser from attempting to expand values that use this syntax, escape the `$` prefix with a second `$` character; that is, provide the value as `++$${my.property}++`.
The value of the property is then passed to {prodname} as `++${my.property}++`.
```

Naros: Looks good to me, applied the edits. :+1:

PR #6820, opened 2025-10-10
DBZ-9465 Adding pluggable off-heap  providers for memory buffer type

nathan-smit-1:
Relates to discussion here:  [discussion](https://debezium.zulipchat.com/#narrow/channel/348250-community-oracle/topic/Hitting.20heap.20limits.20during.20large.20transaction).  Creating as draft PR for feedback.  There is still work to be done but I've already made a lot of decisions which may be questionable and so wanted to send this through.

High Level:  When there are large transactions in Oracle, Debezium can OOM as the entire transaction must be stored in heap.  In order to avoid this, allow a configurable threshold at which we spill from heap to an off-heap provider.  This allows most transactions to use heap and only move to slower off-heap providers when necessary.

Current principles
1. Spilled data should be ephemeral extension of heap and not survive restarts
2. Spilled transactions are split between heap and off-heap (i.e. don't try to push the existing transaction data to disk, just spill from the next event on that meets threshold)

Current Implementation:
I've added Chronicle Queue and Rocksdb as new off-heap providers.  I then added the existing Ehcache buffer type as a new spill provider which requires some small code changes.  I've created CompositeTransactionCache which is a sort of facade to coordinate things which is one of the main things I'm unsure about as it created some weirdness managing the MemoryTransaction.

The advantage is I can make specific decisions about spill and heap.  I was also able add a configurable index to hopefully help the off-heap providers when it comes to partial rollbacks.  I created three new profiles for testing that set the threshold to 0 to always spill.  All tests should be passing but I added exclusions for the infinispanITs as I don't think those make sense here.

Currently the spill is just based on number of events seen and I don't have any new metrics yet but I've tried to cater for potential of other strategies being used.

Chronicle Queue Specific concerns:
1.   As discussed the Chronicle Queue team only publish EA jars to MVN and so currently I have selected 5.27ea8 but this isn't ideal
2. For Java 17+ additional parameters have to be passed as outlined here:  https://chronicle.software/chronicle-support-java-17/  I currently set these in the profile and in my local DBZ server testing I add them in the run.sh file
3. For chronicle specifically, I made each transaction its own individual folder and manage each separately.  I'm not sure if this is the right decision but it helps keep things clean and in my view it isn't the intention to have thousands of transactions spilling concurrently to disk.

Rocksdb specific concerns:
1. There are some issues with the fat jar on the latest version and so I selected 10.2.1 but probably not a huge issue:  https://github.com/facebook/rocksdb/issues/13893#issuecomment-3240464232
2. When I used defaults I was failing on some of the lob tests (timing out) and so had to add additional configurable options but I'm not a rocksdb expert and don't know the best way to tune it.

I haven't done a lot of testing yet.  I need a more extensive test for partial rollbacks.  Currently I have tested all three providers with the following test:  Set threshold to 50 000, Insert 10 000 000 rows in batches of 100 000.  Update all rows.  Delete all rows.  None of the solutions ran out of heap (currently set to 8gb).  I would like to do more benchmarks especially against just the memory buffer type to see if I've added overhead somewhere.

### DIFF
--- a/debezium-connector-oracle/pom.xml
+++ b/debezium-connector-oracle/pom.xml
@@ -112,6 +112,20 @@
             <version>${version.jaxb.implementation}</version>
         </dependency>
 
+        <!-- Chronicle Queue Dependencies -->
+        <dependency>
+            <groupId>net.openhft</groupId>
+            <artifactId>chronicle-queue</artifactId>
+            <version>${version.chronicle.queue}</version>
+        </dependency>
+
+        <!-- RocksDB Dependencies -->
+        <dependency>
+            <groupId>org.rocksdb</groupId>
+            <artifactId>rocksdbjni</artifactId>
+            <version>${version.rocksdbjni}</version>
+        </dependency>
+
         <!-- Introduce ehcache dependencies on Maven Central -->
         <dependency>
             <groupId>com.sun.istack</groupId>
@@ -227,6 +241,9 @@
         -->
         <version.istack>3.0.7</version.istack>
         <version.fastinfoset>1.2.15</version.fastinfoset>
+        
+        <version.chronicle.queue>2026.1</version.chronicle.queue>
+        <version.rocksdbjni>10.2.1</version.rocksdbjni>
 
         <protobuf.output.directory>${project.basedir}/generated-sources</protobuf.output.directory>
 
@@ -308,7 +325,17 @@
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-surefire-plugin</artifactId>
                 <configuration>
-                    <argLine>-Djavax.xml.parsers.SAXParserFactory=com.sun.org.apache.xerces.internal.jaxp.SAXParserFactoryImpl</argLine>
+                    <argLine>
+                        -Djavax.xml.parsers.SAXParserFactory=com.sun.org.apache.xerces.internal.jaxp.SAXParserFactoryImpl
+                        --add-opens java.base/sun.nio.ch=ALL-UNNAMED
+                        --add-opens java.base/java.nio=ALL-UNNAMED
+                        --add-opens java.base/sun.security.action=ALL-UNNAMED
+                        --add-opens java.base/java.lang.reflect=ALL-UNNAMED
+                        --add-opens java.base/java.lang=ALL-UNNAMED
+                        --add-opens java.base/java.util=ALL-UNNAMED
+                        --add-opens java.base/jdk.internal.misc=ALL-UNNAMED
+                        --add-opens java.base/jdk.internal.ref=ALL-UNNAMED
+                    </argLine>
                     <systemPropertyVariables combine.children="append">
                         <database.connection.adapter>${adapter.name}</database.connection.adapter>
                         <log.mining.buffer.type>${log.mining.buffer.type.name}</log.mining.buffer.type>
@@ -336,11 +363,21 @@
                 <configuration>
                     <skipTests>${skipITs}</skipTests>
                     <enableAssertions>true</enableAssertions>
-                    <forkCount>0</forkCount>
-                    <argLine>-Djava.library.path=${instantclient.dir} -Ddebezium.embedded.shutdown.pause.before.interrupt.ms=20000</argLine>
+                    <argLine>
+                        -Djava.library.path=${instantclient.dir} -Ddebezium.embedded.shutdown.pause.before.interrupt.ms=20000
+                        --add-opens java.base/sun.nio.ch=ALL-UNNAMED
+                        --add-opens java.base/java.nio=ALL-UNNAMED
+                        --add-opens java.base/sun.security.action=ALL-UNNAMED
+                        --add-opens java.base/java.lang.reflect=ALL-UNNAMED
+                        --add-opens java.base/java.lang=ALL-UNNAMED
+                        --add-opens java.base/java.util=ALL-UNNAMED
+                        --add-opens java.base/jdk.internal.misc=ALL-UNNAMED
+                        --add-opens java.base/jdk.internal.ref=ALL-UNNAMED
+                    </argLine>
                     <systemPropertyVariables>
                         <database.connection.adapter>${adapter.name}</database.connection.adapter>
                         <log.mining.buffer.type>${log.mining.buffer.type.name}</log.mining.buffer.type>
+                        <log.mining.buffer.spill.provider>${log.mining.buffer.spill.provider}</log.mining.buffer.spill.provider>
                         <internal.log.mining.read.only>${log.mining.read.only.mode}</internal.log.mining.read.only>
                         <javax.xml.parsers.SAXParserFactory>com.sun.org.apache.xerces.internal.jaxp.SAXParserFactoryImpl</javax.xml.parsers.SAXParserFactory>
                         <version.oracle.server>${version.oracle.server}</version.oracle.server>
@@ -805,6 +842,178 @@
                 </plugins>
             </build>
         </profile>
+
+        <!-- Memory spillover profiles (chronicle/rocksdb/ehcache) -->
+        <profile>
+            <id>oracle-memory-chronicle-spillover</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <properties>
+                <log.mining.buffer.spill.provider>chronicle</log.mining.buffer.spill.provider>
+            </properties>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <excludes>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceIT.java</exclude>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceTest.java</exclude>
+                            </excludes>
+                        </configuration>
+                    </plugin>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-failsafe-plugin</artifactId>
+                        <configuration>
+                            <excludes>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceIT.java</exclude>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceTest.java</exclude>
+                            </excludes>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
+        <profile>
+            <id>oracle-memory-rocksdb-spillover</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <properties>
+                <log.mining.buffer.spill.provider>rocksdb</log.mining.buffer.spill.provider>
+            </properties>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <excludes>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceIT.java</exclude>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceTest.java</exclude>
+                            </excludes>
+                        </configuration>
+                    </plugin>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-failsafe-plugin</artifactId>
+                        <configuration>
+                            <excludes>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceIT.java</exclude>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceTest.java</exclude>
+                            </excludes>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
+        <profile>
+            <id>oracle-memory-ehcache-spillover</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <properties>
+                <log.mining.buffer.spill.provider>ehcache</log.mining.buffer.spill.provider>
+            </properties>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <excludes>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceIT.java</exclude>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceTest.java</exclude>
+                            </excludes>
+                        </configuration>
+                    </plugin>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-failsafe-plugin</artifactId>
+                        <configuration>
+                            <excludes>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceIT.java</exclude>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceTest.java</exclude>
+                            </excludes>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
+        <profile>
+            <id>oracle-memory-infinispan-embedded-spillover</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <properties>
+                <log.mining.buffer.spill.provider>infinispan_embedded</log.mining.buffer.spill.provider>
+            </properties>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <excludes>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceIT.java</exclude>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceTest.java</exclude>
+                            </excludes>
+                        </configuration>
+                    </plugin>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-failsafe-plugin</artifactId>
+                        <configuration>
+                            <excludes>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceIT.java</exclude>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceTest.java</exclude>
+                            </excludes>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
+        <profile>
+            <id>oracle-memory-infinispan-remote-spillover</id>
+            <activation>
+                <activeByDefault>false</activeByDefault>
+            </activation>
+            <properties>
+                <log.mining.buffer.spill.provider>infinispan_remote</log.mining.buffer.spill.provider>
+            </properties>
+            <build>
+                <plugins>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-surefire-plugin</artifactId>
+                        <configuration>
+                            <excludes>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceIT.java</exclude>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceTest.java</exclude>
+                            </excludes>
+                        </configuration>
+                    </plugin>
+                    <plugin>
+                        <groupId>org.apache.maven.plugins</groupId>
+                        <artifactId>maven-failsafe-plugin</artifactId>
+                        <configuration>
+                            <excludes>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceIT.java</exclude>
+                                <exclude>**/EmbeddedInfinispanStreamingChangeEventSourceTest.java</exclude>
+                            </excludes>
+                        </configuration>
+                    </plugin>
+                </plugins>
+            </build>
+        </profile>
+
         <!-- This profile should be used for testing connector with EhCache only -->
         <profile>
             <id>oracle-ehcache</id>
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -77,6 +77,12 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
 
     protected final static long DEFAULT_RESUME_POSITION_INTERVAL = 10_000L;
 
+    // Buffer spill configuration constants
+    protected static final int SPILL_DISABLED_THRESHOLD = -1;
+    protected static final int DEFAULT_SPILL_IN_MEMORY_INDEX_THRESHOLD = 10000;
+    protected static final String DEFAULT_CHRONICLE_ROLL_CYCLE = "FAST_DAILY";
+    protected static final int DEFAULT_ROCKSDB_BATCH_FLUSH_THRESHOLD = 10000;
+
     public static final Field PORT = RelationalDatabaseConnectorConfig.PORT
             .withDefault(DEFAULT_PORT);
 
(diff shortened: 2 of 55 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -207,6 +207,20 @@ private LogWriterFlushStrategy resolveFlushStrategy() {
 
     @SuppressWarnings("unchecked")
     private <T extends Transaction> CacheProvider<T> createCacheProvider(OracleConnectorConfig connectorConfig) {
+        if (connectorConfig.getLogMiningBufferType() == OracleConnectorConfig.LogMiningBufferType.MEMORY && connectorConfig.isLogMiningBufferSpillEnabled()) {
+            OracleConnectorConfig.LogMiningSpillProvider provider = connectorConfig.getLogMiningBufferSpillProvider();
```
Naros: Could we drive the instance creation from the enum, like we do with the adapter?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -217,6 +231,24 @@ private <T extends Transaction> CacheProvider<T> createCacheProvider(OracleConne
 
     @SuppressWarnings("unchecked")
     private <T extends Transaction> TransactionFactory<T> createTransactionFactory(OracleConnectorConfig connectorConfig) {
+        if (connectorConfig.getLogMiningBufferType() == OracleConnectorConfig.LogMiningBufferType.MEMORY && connectorConfig.isLogMiningBufferSpillEnabled()) {
+            OracleConnectorConfig.LogMiningSpillProvider provider = connectorConfig.getLogMiningBufferSpillProvider();
```
Naros: Could we drive the factory from the enum rather than using a switch? See adapter.

debezium-connector-oracle/pom.xml
```diff
@@ -347,11 +374,21 @@
...
+                    <argLine>
+                        -Djava.library.path=${instantclient.dir} -Ddebezium.embedded.shutdown.pause.before.interrupt.ms=20000
+                        --add-opens java.base/sun.nio.ch=ALL-UNNAMED
+                        --add-opens java.base/java.nio=ALL-UNNAMED
+                        --add-opens java.base/sun.security.action=ALL-UNNAMED
+                        --add-opens java.base/java.lang.reflect=ALL-UNNAMED
+                        --add-opens java.base/java.lang=ALL-UNNAMED
+                        --add-opens java.base/java.util=ALL-UNNAMED
+                        --add-opens java.base/jdk.internal.misc=ALL-UNNAMED
+                        --add-opens java.base/jdk.internal.ref=ALL-UNNAMED
```
Naros, line 375: I'm curious what ChronicleQueue intends to do on newer JVMs where the sun packages are deprecated and schedued for removal 🤔

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -978,7 +1137,20 @@ public OracleConnectorConfig(Configuration config) {
         this.logMiningUsernameExcludes = Strings.setOfTrimmed(config.getString(LOG_MINING_USERNAME_EXCLUDE_LIST), String::new);
         this.logMiningBufferType = LogMiningBufferType.parse(config.getString(LOG_MINING_BUFFER_TYPE));
         this.logMiningBufferTransactionEventsThreshold = config.getLong(LOG_MINING_BUFFER_TRANSACTION_EVENTS_THRESHOLD);
-        this.logMiningBufferDropOnStop = config.getBoolean(LOG_MINING_BUFFER_DROP_ON_STOP);
+        this.logMiningBufferSpillInMemoryEventsThreshold = config.getInteger(LOG_MINING_BUFFER_SPILL_IN_MEMORY_EVENTS_THRESHOLD);
```
Naros, line 1170: @jpechane, the Oracle connector configuration property list is getting quite large at this point. I wonder if we shouldn't consider splitting the class in a future PR? 

I'm not really sure how we'd go about it in a way that doesn't create any more friction, but I do want to consider at a minimum exposing properties only if they're common, otherwise adapter-specific ones should be made available through a Record or adapter-specific class wrapper.

Thoughts/Suggestions?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -431,9 +431,140 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
...
+            .withDisplayName("The spill strategy to use for determining when to spill events")
+            .withEnum(LogMiningSpillStrategy.class, LogMiningSpillStrategy.EVENT_COUNT)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withDescription("The strategy implementation to use for determining when transactions should begin spilling events to disk.");
+
+    public static final Field LOG_MINING_BUFFER_SPILL_IN_MEMORY_EVENTS_THRESHOLD = Field.create("log.mining.buffer.spill.in.memory.events.threshold")
+            .withDisplayName("The number of events to keep in memory before spilling to disk")
+            .withType(Type.INT)
+            .withDefault(-1)
```
Naros: Lets move defaults to constants and if there are cases where a property could one of several values, lets consider implementing an enumeration for it, e.g. 'chronicle.roll.cycle`.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/SpillStrategy.java
```diff
@@ -0,0 +1,33 @@
...
+     * Determines if a transaction should begin spilling events to the spillover provider.
+     *
+     * @param txId the transaction ID
+     * @param latest the latest event being added to the transaction
+     * @param view a view of the transaction's current state
+     * @return true if the transaction should begin spilling, false otherwise
+     */
+    boolean shouldSpill(String txId, LogMinerEvent latest, TransactionView view);
+
+    default void onTransactionCreated(String txId) {
```
Naros, line 35: Let's be sure to document, even default methods on interfaces.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/MemoryTransactionCacheAdapter.java
```diff
@@ -0,0 +1,145 @@
...
+import io.debezium.connector.oracle.logminer.buffered.memory.MemoryTransaction;
+import io.debezium.connector.oracle.logminer.events.LogMinerEvent;
+
+/**
+ * An adapter that maps provider transactions to internal MemoryTransaction instances
+ */
+public class MemoryTransactionCacheAdapter<T extends Transaction> {
+
+    private final MemoryLogMinerTransactionCache inner;
+    private final Map<String, MemoryTransaction> txMap = new ConcurrentHashMap<>();
```
Naros: Lets find a better name for `inner` and change `txMap` to perhaps just `transactions` or `transactionsById`.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/CompositeTransactionCache.java
```diff
@@ -0,0 +1,533 @@
...
+        void clear() {
+            rowIndex.clear();
+            insertionOrder.clear();
+        }
+    }
+
+    // Core components
+    private final LogMinerTransactionCache<T> spilloverCache;
+    private final SpillStrategy spillStrategy;
+    private final io.debezium.connector.oracle.logminer.buffered.MemoryTransactionCacheAdapter<T> memoryEventStore;
```
Naros: Prefer imports for readability.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/CompositeTransactionCache.java
```diff
@@ -0,0 +1,533 @@
...
+     */
+    public CompositeTransactionCache(LogMinerTransactionCache<T> spilloverCache,
+                                     SpillStrategy spillStrategy,
+                                     int indexThreshold) {
+        this.spilloverCache = spilloverCache;
+        this.spillStrategy = spillStrategy;
+        this.indexThreshold = indexThreshold;
+
+        // Create memory event store internally (adapter maps provider transactions to memory)
+        this.memoryEventStore = new io.debezium.connector.oracle.logminer.buffered.MemoryTransactionCacheAdapter<>();
```
Naros: Prefer imports.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/CompositeTransactionCache.java
```diff
@@ -0,0 +1,533 @@
...
+    }
+
+    /**
+     * Lightweight index structure used only for spilled transactions. Keeps rowId -> eventKey lists
+     * and minimal index bookkeeping. Created lazily when a transaction starts spilling.
+     * Implements FIFO eviction using LinkedHashMap to maintain a sliding window of the most recent indexed events.
+     */
+    private static class SpilloverRowIdIndex {
+        private final ConcurrentHashMap<String, ConcurrentLinkedDeque<Integer>> rowIndex = new ConcurrentHashMap<>();
+        private final LinkedHashMap<IndexKey, Integer> insertionOrder;
```
jeremy-l-ford, line 50: Since most all other maps in this class are Concurrent, does this map need to be synchronized?

nathan-smit-1: hey @jeremy-l-ford, thanks for the feedback.  This is a good question.  My understanding is that Debezium is largely (always?) single-threaded and so the use of concurrent structures in general is largely defensive and I'm then making a concession here with the LinkedHashMap as I'm using it because I want to leverage the removeEldestEntry() callback which creates this inconsistency.

If I look at the existing ehcache implementation for example the structures it uses for managing its cache are not concurrent:   
private final Map<String, TreeSet<Integer>> eventIdsByTransactionId = new HashMap<>();  

So perhaps it makes sense to avoid the defensive concurrent maps?

nathan-smit-1:
@Naros was just messing with some things here which is why this is building again.  Removed WAL for rocksdb and then based on what Jeremy was saying removed all the concurrent structures

Naros:
It's in draft, so when you think its in a solid state for review, just mark it ready. If there's something specific you need guidance on, feel free to ping, @nathan-smit-1.

nathan-smit-1:
@Naros Okay I marked it as ready for review.  I don't think there is anything more I wanted to do on.  Mainly just needs feedback on all the many decisions I made while working on it

nathan-smit-1:
@Naros sorry one other thing here:  I don't think we had any resolution on the Chronicle Queue licensing so probably also need a call on whether using EA versions is okay

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -1633,6 +1808,66 @@ private static LogMiningBufferType parseOrDefault(String value, String defaultVa
 
     }
 
+    public enum LogMiningSpillProvider implements EnumeratedValue {
```
Naros, line 1817: Do you think an Infinispan implementation would equally be possible?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/CompositeTransactionCache.java
```diff
@@ -0,0 +1,535 @@
...
+
+    @Override
+    public void transactions(Consumer<Stream<T>> consumer) {
+        consumer.accept(transactionRegistry.values().stream());
+    }
+
+    @Override
+    public void eventKeys(Consumer<Stream<String>> consumer) {
+        memoryEventStore
+                .eventKeys(memoryStream -> spilloverCache.eventKeys(spilloverStream -> consumer.accept(java.util.stream.Stream.concat(memoryStream, spilloverStream))));
```
Naros:
```suggestion
                .eventKeys(memoryStream -> spilloverCache.eventKeys(spilloverStream -> consumer.accept(Stream.concat(memoryStream, spilloverStream))));
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/CompositeTransactionCache.java
```diff
@@ -0,0 +1,535 @@
...
+            if (isSpilling(transactionId)) {
+                // Already spilling - add to spillover
+                spilloverCache.addTransactionEvent(transaction, eventKey, event);
+                updateSpilloverRowIdIndex(transactionId, eventKey, event);
+            }
+            else {
+                // Check if we should start spilling
+                TransactionView view = new TransactionViewImpl(transactionId);
+                if (spillStrategy != null && spillStrategy.shouldSpill(transactionId, event, view)) {
+                    LOGGER.debug("Transaction {} is transitioning to spillover state.", transactionId);
```
Naros: Do we need to announce we "are transitioning" and then "we have transitioned" between this point and the following function call? I'm just curious if one log entry would be sufficient?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/MemoryTransactionCacheAdapter.java
```diff
@@ -0,0 +1,146 @@
...
+import io.debezium.connector.oracle.logminer.buffered.memory.MemoryTransaction;
+import io.debezium.connector.oracle.logminer.events.LogMinerEvent;
+
+/**
+ * An adapter that maps provider transactions to internal MemoryTransaction instances.
+ * Single-threaded since LogMiner processing is single-threaded.
+ */
+public class MemoryTransactionCacheAdapter<T extends Transaction> {
+
+    private final MemoryLogMinerTransactionCache inner;
```
Naros:
```suggestion
    private final MemoryLogMinerTransactionCache delegate;
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/MemoryTransactionCacheAdapter.java
```diff
@@ -0,0 +1,146 @@
...
+    private final Map<String, MemoryTransaction> txMap = new HashMap<>();
```
Naros:
```suggestion
    private final Map<String, MemoryTransaction> transactionsById = new HashMap<>();
```
Just to keep with the naming in the other caches.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/chronicle/ChronicleCacheProvider.java
```diff
@@ -0,0 +1,152 @@
...
+    private final MemoryBasedLogMinerCache<String, String> schemaChangesCache;
+
+    public ChronicleCacheProvider(OracleConnectorConfig connectorConfig) {
+
+        if (!connectorConfig.isLogMiningBufferSpillEnabled()) {
+            throw new DebeziumException("Chronicle Queue spill must be enabled to use ChronicleCacheProvider");
+        }
+
+        String spillPath = connectorConfig.getLogMiningBufferSpillPath();
+        if (spillPath == null || spillPath.trim().isEmpty()) {
```
Naros:
```suggestion
        if (Strings.isNullOrEmpty(spillPath)) {
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/chronicle/ChronicleCacheProvider.java
```diff
@@ -0,0 +1,152 @@
...
+            throw new DebeziumException("Chronicle Queue spill path must be configured when spill is enabled");
+        }
+
+        Path spillDirectory = Paths.get(spillPath);
+        long spillThreshold = connectorConfig.getLogMiningBufferSpillThreshold();
+        String rollCycle = connectorConfig.getChronicleRollCycle();
+
+        if (rollCycle == null || rollCycle.trim().isEmpty()) {
```
Naros:
```suggestion
        if (Strings.isNullOrEmpty(rollCycle)) {
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/chronicle/ChronicleCacheProvider.java
```diff
@@ -0,0 +1,152 @@
...
+            throw new DebeziumException("Chronicle roll cycle must be configured when spill is enabled");
+        }
+        try {
+            net.openhft.chronicle.queue.RollCycles.valueOf(rollCycle);
```
Naros: Lets import RollCycles.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/chronicle/ChronicleEventSerializer.java
```diff
@@ -0,0 +1,379 @@
...
+
+        protected void writeObject(ValueOut valueOut, Object value) {
+            if (value == null) {
+                valueOut.int8(VALUE_TYPE_NULL);
+            }
+            else if (value == OracleValueConverters.UNAVAILABLE_VALUE) {
+                valueOut.int8(VALUE_TYPE_UNAVAILABLE);
+                writeString(valueOut, UNAVAILABLE_VALUE_SENTINEL);
+            }
+            else if (value instanceof String) {
```
Naros: Let's prefer instanceof pattern matching here for these `else if` cases.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/chronicle/ChronicleEventSerializer.java
```diff
@@ -0,0 +1,379 @@
...
+                case VALUE_TYPE_LONG:
+                    return valueIn.int64();
+                case VALUE_TYPE_DOUBLE:
+                    return valueIn.float64();
+                case VALUE_TYPE_FLOAT:
+                    return valueIn.float32();
+                case VALUE_TYPE_BOOLEAN:
+                    return valueIn.bool();
+                case VALUE_TYPE_BIG_DECIMAL:
+                    return new java.math.BigDecimal(readString(valueIn));
```
Naros: Import BigDecimal.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/chronicle/ChronicleEventSerializer.java
```diff
@@ -0,0 +1,379 @@
...
+
+        @Override
+        public void writeEvent(ValueOut valueOut, LogMinerEvent event) {
+            writeCommonFields(valueOut, event);
+        }
+
+        @Override
+        public LogMinerEvent readEvent(ValueIn valueIn) {
+            byte eventTypeOrdinal = valueIn.int8();
+            EventType eventType = EventType.values()[eventTypeOrdinal];
```
Naros: Let's guard the ordinal against the EventType length to avoid IndexOutOfBoundsExceptions with a more appropriate error message.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/rocksdb/RocksDbLogMinerTransactionCache.java
```diff
@@ -0,0 +1,596 @@
...
+                String name = new String(handle.getName(), StandardCharsets.UTF_8);
+                if ("events".equals(name)) {
+                    eventsHandle = handle;
+                }
+                else if ("rowIdIndex".equals(name)) {
+                    indexHandle = handle;
+                }
+                else if ("transactions".equals(name)) {
+                    txHandle = handle;
+                }
```
Naros: Could we construct 3 static member variables that represent the `byte[]` for each of the strings `events`, `rowIdIndex`, and `transactions` allowing us to just compare the `byte[]` during the loop without allocating an extra `String`?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/rocksdb/RocksDbLogMinerTransactionCache.java
```diff
@@ -0,0 +1,596 @@
...
+
+            if (indexHandle == null) {
+                ColumnFamilyDescriptor indexDescriptor = new ColumnFamilyDescriptor("rowIdIndex".getBytes(StandardCharsets.UTF_8), new ColumnFamilyOptions());
+                indexHandle = db.createColumnFamily(indexDescriptor);
+            }
+
+            if (txHandle == null) {
+                ColumnFamilyDescriptor txDescriptor = new ColumnFamilyDescriptor("transactions".getBytes(StandardCharsets.UTF_8), new ColumnFamilyOptions());
+                txHandle = db.createColumnFamily(txDescriptor);
+            }
```
Naros, line 134: We could again use the 3 static member `byte[]` members for the descriptor's first argument.

Naros (commented):
Hi @nathan-smit-1 here's a quick pass. I still need to play more with this more. 

I would assume since we're handling the spill on the "processing" side, if we were to split the database read loop from the processing loop, then with this PR we'd likely just plug in some buffer much like we have between Debezium and Kafka Connect where we simply push the `LogMinerEventRow` onto this buffer and let the processing loop dequeue and work with the spillage.  Would you agree?

nathan-smit-1:
> Hi @nathan-smit-1 here's a quick pass. I still need to play more with this more.
> 
> I would assume since we're handling the spill on the "processing" side, if we were to split the database read loop from the processing loop, then with this PR we'd likely just plug in some buffer much like we have between Debezium and Kafka Connect where we simply push the `LogMinerEventRow` onto this buffer and let the processing loop dequeue and work with the spillage. Would you agree?

Is this something you're planning currently?  Would the processing loop in this scenario become multi-threaded?

I think what you're suggesting here makes sense though, yes.  In general would probably need to see how this PR should be impacted by e.g. DBZ-9636 as well.

Will try go through your comments today.

Naros:
Hi @nathan-smit-1 

> Is this something you're planning currently? Would the processing loop in this scenario become multi-threaded?

It's something I'd like to look at in 2026, yes. I am still trying to decide if it makes sense to just split the commit handler to a separate thread or if it makes sense to split more from the LogMiner data read pass.

nathan-smit-1:
@Naros regarding your question on Chronicle Queue:  On this version support page they mention having support for Java 25 in version that released 2025/10 (this happened after I did my pull request) https://github.com/OpenHFT/OpenHFT/blob/b69b7dd26721c189d9705bb9a0edc05613440823/docs/Version-Support.adoc#L50

I'm not sure how that maps to the EA releases but that's probably part of broader discussion about using EA jars. 5.27ea11 released in November

Edit:  I could potentially upgrade to 5.27ea11 and remove those jvm arguments and see if it works?

nathan-smit-1:
@Naros I've made some changes.  Side note:  When you review is it easier for you if I do a second commit and squash later?  Currently I always just squash everything into the same commit but I don't know if that makes reviewing more difficult.  Just some specific points:

- I added CompositeTransactionCache for Infinispan Embedded and Infinispan Remote which should allow them to be used as spillover providers
- My main concern on having so many providers is just my own bandwidth to keep track of and test them all.  I ran the full set of tests for Infinispan Embedded which passed though
- The way I've done the implementation is that there are new profiles for every spillover provider that run all the tests with a threshold of zero for the spill (I just excluded the infinispanITs).  I dunno how much overhead this adds though as it means running all of the tests 5 more times currently?

Would be nice to get some more eyes on how well this runs in the wild as well

Naros:

Hi @nathan-smit-1, please sign off all commits with:

```
git commit -s
```

nathan-smit-1:
@Naros pushed this one again with the sign off on the commit.  Let me know if I need to do anything else on these older PRs

nathan-smit:
@Naros just FYI I just tried  5.27ea11 for Chronicle Queue on this one as that was on my todo list but the JVM arguments are still required on that version

Naros:
Hi @nathan-smit I wonder if we should consider bumping to the stable 2026 release. I just noticed they changed their release process to be more formal and no more EA releases [here](https://github.com/OpenHFT/OpenHFT/blob/ea/docs/Version-Support.adoc#release-and-development-model). 

This doesn't change the JVM arguments issue, that's still _ongoing_

nathan-smit-1:
@Naros okay great!  That's still good news that we can move away from the EA builds.  I bumped to 2026.1 and reran all the tests under the chronicle spillover profile and all still seemed good.  Pushed that change through as well

Naros:
As an update, I did talk with them and they are actively working on removing the requirement for the JVM opts. @jpechane from your PoV, should we wait until Chronicle removes that as a requirement (see point 2 in the PR description)?

jpechane:
@Naros This is not a blocker, we need it for Cassandra connectors by default so I believe we might require them for this senario too.

Naros:
In that case, I'll circle back to this @nathan-smit-1 in the early 3.6 cycle. Thanks @jpechane.

nathan-smit-1:
> In that case, I'll circle back to this @nathan-smit-1 in the early 3.6 cycle. Thanks @jpechane.

Sounds good.  There are some minor code adjustments I want to do on this still.  Just clean-up based on feedback I've gotten on some other PRs.

I can then hopefully also do docs for this before you review again?  Not sure where docs for this should sit though as I noticed that the existing Ehcache, Infinispan buffer options don't appear to have documentation currently

PR #6821, opened 2025-10-10
Handle TOASTed Geography columns

prburgu:


jpechane:
@prburgu Could you please handle GEO arrays too?

PR #6823, opened 2025-10-12
DBZ-9545 Introduce mining/fetch SCN boundary metrics

Naros:
https://issues.redhat.com/browse/DBZ-9545


jpechane:
Test failures unrelated, applied, thanks!

PR #6824, opened 2025-10-12
DBZ-9546 Move to bcpix-jdk18on 1.82

Naros:
https://issues.redhat.com/browse/DBZ-9546

@AlvarVG FYI


jpechane:
@Naros Applied, thanks

PR #6825, opened 2025-10-14
DBZ-9549 Couple clearance of DdlChanges with parseDdl

akanimesh7:
- Users of DdlParser are storing/handling the reset of DdlChanges manually. It was done correctly in MySqlDatabaseSchema and OracleSchemaChangeEventEmitter, but was missed in AbstractDatabaseSchemaHistory. This resulted in hogging up all the events that are parsed while performing schema history recovery. This fills up the memory and is not useful for functioning of the connector.
- We are coupling the DdlChanges reset part with the DdlParser itself. Users of DdlParser should just bother about calling `parse` and expect the DdlChanges to have the result of parsing the most recent DDL.

### DIFF
--- a/debezium-ddl-parser/src/main/java/io/debezium/antlr/AntlrDdlParser.java
+++ b/debezium-ddl-parser/src/main/java/io/debezium/antlr/AntlrDdlParser.java
@@ -8,7 +8,6 @@ package io.debezium.antlr;
 
 import java.util.Collection;
 
-import io.debezium.relational.ddl.DdlChanges;
 import org.antlr.v4.runtime.CharStream;
 import org.antlr.v4.runtime.CharStreams;
 import org.antlr.v4.runtime.CodePointCharStream;
@@ -24,6 +23,7 @@ import org.antlr.v4.runtime.tree.ParseTreeWalker;
 import io.debezium.relational.TableId;
 import io.debezium.relational.Tables;
 import io.debezium.relational.ddl.AbstractDdlParser;
+import io.debezium.relational.ddl.DdlChanges;
 import io.debezium.text.MultipleParsingExceptions;
 import io.debezium.text.ParsingException;
 
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogDatabaseSchema.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogDatabaseSchema.java
@@ -305,9 +305,10 @@ public abstract class BinlogDatabaseSchema<P extends BinlogPartition, O extends 
             return schemaChangeEvents;
         }
 
+        DdlChanges ddlChanges = new DdlChanges();
         try {
             this.ddlParser.setCurrentSchema(databaseName);
-            this.ddlParser.parse(ddlStatements, tables());
+            ddlChanges = this.ddlParser.parse(ddlStatements, tables());
         }
         catch (ParsingException | MultipleParsingExceptions e) {
             if (skipUnparseableDdlStatements()) {
@@ -317,8 +318,6 @@ public abstract class BinlogDatabaseSchema<P extends BinlogPartition, O extends 
                 throw e;
             }
         }
-
-        DdlChanges ddlChanges = this.ddlParser.getAndResetDdlChanges();
 
         // No need to send schema events or store DDL if no table has changed
         if (!storeOnlyCapturedTables() || isGlobalSetVariableStatement(ddlStatements, databaseName) || ddlChanges.anyMatch(filters)) {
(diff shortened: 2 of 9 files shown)


jpechane:
Test failures are related

akanimesh7:
@Naros can you check this ? The tests which were failing were authored by you. It would be good to have your input.

debezium-ddl-parser/src/main/java/io/debezium/antlr/AntlrDdlParser.java
```diff
@@ -61,6 +61,9 @@ public AntlrDdlParser(boolean throwErrorsFromTreeWalk, boolean includeViews, boo
     public void parse(String ddlContent, Tables databaseTables) {
         this.databaseTables = databaseTables;
 
+        // DDL events from previous parse operation are no more relevant.
+        ddlChanges.reset();
```
Naros: What if instead `DdlChanges` implemented `AutoCloseable` and then where we use it in the Oracle and Binlog classes we use a `try-with-resources` block, letting `close()` handle the `reset`?

In that case we remove this `reset` call in the `AntlrDdlParser`.

This also gives the flexibility for situations where we'd like to let the `DdlChanges` class not reset by not using the `try-with-resources` block.

@akanimesh7 how does that sound?

akanimesh7: The problem is we want to reset "just before the next parse call" not "after every parse call". How it works is once parse is completed, `ddlChanges` is expected to contain relevant events. The code accessing these events is not coupled with the code populating these events. 

> This also gives the flexibility for situations where we'd like to let the DdlChanges class not reset by not using the try-with-resources block.

Isn't it same as not calling reset ?

@Naros does that sound reasonable ?

Naros: So the current code as it stands both in `main` and your suggestion simply call `reset` on the `DdlChanges` right before we are about to parse the next DDL event. This means that the prior DDL event data is kept in memory unnecessarily. 

By applying a `try-with-resources` approach where `close` clears the `DdlChanges` buffer, we avoid at runtime the parser holding onto potentially long DDL strings. We have some users who have tables with hundreds to thousands of columns, and a `CREATE TABLE` statement could potentially be quite large.

The suggestion is purely an efficiency optimization so that at runtime we cleanup earlier rather than waste valuable heap retaining something we no longer truly need.

So yes, it's the same as callling `reset` like we do now, but it's done after we've processed the changes in the lower block of the schema/emitter code rather than earlier in the method using clearer semantics imo and removes the call from inside the parser.

akanimesh7: @Naros 
I tried implementing this and few issues that i can see
1. The `Autocloseable` interface was originally intended to be used for objects which are holding file or socket handles etc. (Heavy resources). One notable thing is -  the `close` method throws an Exception, which makes sense if the `close` involves relinquishing some heavy resources like file or socket, etc. But in our case, the Exception would never occur since we are just clearing an in-memory array. But still, since it's a part of API contract, it has to be caught at all places where `ddlChanges` would be used. This looks like a code smell to me.
2. For places where the `ddlChanges` is useful, it makes sense to surround in `try-with-resources`.  For example -- `BinlogDatabaseSchema` and `OracleSchemaChangeEventEmitter`. But for usages such as the one in `AbstractSchemaHistory`, we are just interested in making the `parse` call. What we need here is the in-memory schema to be adjusted according to the DDL parsing. We don't need the `ddlChanges` here. So if we were to remove the `reset` call from inside the parser, it would not solve the current problem of memory accumulation during schema history recovery. Additionally, if we just add a try with resources for DdlChanges array here and do nothing inside the block, since we just want to clear the `ddlChanges` array, that also looks like a code smell. 

> waste valuable heap retaining something we no longer truly need.

Yes this makes sense. The ideal way to solve this should have been an opt-in where the caller of `parse` could signify whether it needs the `ddlChanges` array or not. If it needs, it will use the array with try-with-resource and hence the memory will get cleared right after the usage. Otherwise, the ddlChanges would not be populated if it is not needed ? I am not sure how big of a change this is and whether it will be worth investing into it.

Naros: I see your point with (2), and that makes perfect sense. 

I do like your latter idea. I'd say for now, one of the less disruptive changes would be to have parse return the DdlChanges to the caller. Then at the end of the `parse` method, we copy the member instance to a local variable, create a new instance as the member variable and return the local instance. 

It's one extra allocation per parse call, but we avoid the `AutoCloseable` smell, and we completely avoid any heap issues with retained SQL for both the `AbstractSchemaHistory` class and the residual SQL after each call to `parse`.  

What do you think @akanimesh7 ?

Naros: @akanimesh7 would you mind rebasing the PR on `main` ?

akanimesh7: @Naros Yes i agree. The elegant solution would be to simply return DdlChanges on every parse call. Have implemented that.

Also have rebased from `main`

akanimesh7:
@Naros PTAL

jpechane:
@akanimesh7 Applied, thanks!

PR #6826, opened 2025-10-14
DBZ-9551 Add Oracle Autonomous Database (ADB) support

martinzuern:
See https://issues.redhat.com/browse/DBZ-9551

Oracle Autonomous Database restricts access to certain views and log file handling.

This PR adds a flag to respect the subtle differences between ADB and normal PDB scenarios.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -93,15 +93,6 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             .withDescription("Name of the pluggable database when working with a multi-tenant set-up. "
                     + "The CDB name must be given via " + DATABASE_NAME.name() + " in this case.");
 
-    public static final Field DATABASE_AUTONOMOUS_MODE = Field.create(ConfigurationNames.DATABASE_CONFIG_PREFIX + "autonomous.mode")
-            .withDisplayName("Autonomous Database Mode")
-            .withType(Type.BOOLEAN)
-            .withWidth(Width.SHORT)
-            .withImportance(Importance.LOW)
-            .withDefault(false)
-            .withGroup(Field.createGroupEntry(Field.Group.CONNECTION_ADVANCED, 0))
-            .withDescription("Enable when connecting to Oracle Autonomous Database (ADB).");
-
     public static final Field XSTREAM_SERVER_NAME = Field.create(ConfigurationNames.DATABASE_CONFIG_PREFIX + "out.server.name")
             .withDisplayName("XStream out server name")
             .withType(Type.STRING)
@@ -779,7 +770,6 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
                     DATABASE_NAME,
                     QUERY_TIMEOUT_MS,
                     PDB_NAME,
-                    DATABASE_AUTONOMOUS_MODE,
                     XSTREAM_SERVER_NAME,
                     SNAPSHOT_MODE,
                     CONNECTOR_ADAPTER,
@@ -939,7 +929,6 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
     private final Integer openLogReplicatorPort;
 
     private final Duration resumePositionUpdateInterval;
-    private final boolean databaseAutonomousMode;
 
     public OracleConnectorConfig(Configuration config) {
         super(
@@ -962,8 +951,6 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
         this.lobEnabled = config.getBoolean(LOB_ENABLED);
         this.objectIdToTableIdCacheSize = config.getInteger(OBJECT_ID_CACHE_SIZE);
         this.legacyDecimalHandlingStrategy = config.getBoolean(LEGACY_DECIMAL_HANDLING_STRATEGY);
-
-        this.databaseAutonomousMode = config.getBoolean(DATABASE_AUTONOMOUS_MODE);
 
         this.streamingAdapter = this.connectorAdapter.getInstance(this);
         if (this.streamingAdapter == null) {
@@ -1036,10 +1023,6 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
 
     public String getPdbName() {
         return pdbName;
-    }
-
-    public boolean isAutonomousDatabaseMode() {
-        return databaseAutonomousMode;
     }
 
     public String getCatalogName() {
@@ -1872,8 +1855,7 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
      * @return true if the connector is to mine archive logs only, false to mine all logs.
      */
     public boolean isArchiveLogOnlyMode() {
-        // ADB mode requires archive-log-only since online redo logs cannot be used
-        return archiveLogOnlyMode || isAutonomousDatabaseMode();
+        return archiveLogOnlyMode;
     }
 
     /**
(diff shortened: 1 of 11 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
```diff
@@ -39,14 +39,19 @@ public class SqlUtils {
     private static final String ARCHIVED_LOG_VIEW = "V$ARCHIVED_LOG";
     private static final String ARCHIVE_DEST_STATUS_VIEW = "V$ARCHIVE_DEST_STATUS";
     private static final String ALL_LOG_GROUPS = "ALL_LOG_GROUPS";
+    private static final String DBA_SUPPLEMENTAL_LOGGING = "DBA_SUPPLEMENTAL_LOGGING";
```
Naros, line 42: This should be documented as a table used by Oracle for Autonomous in `oracle.adoc.`

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -1855,7 +1872,8 @@ public Duration getLogMiningTransactionRetention() {
      * @return true if the connector is to mine archive logs only, false to mine all logs.
      */
     public boolean isArchiveLogOnlyMode() {
-        return archiveLogOnlyMode;
+        // ADB mode requires archive-log-only since online redo logs cannot be used
```
Naros: This criteria should be documented in `oracle.adoc`.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -93,6 +93,15 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             .withDescription("Name of the pluggable database when working with a multi-tenant set-up. "
                     + "The CDB name must be given via " + DATABASE_NAME.name() + " in this case.");
 
+    public static final Field DATABASE_AUTONOMOUS_MODE = Field.create(ConfigurationNames.DATABASE_CONFIG_PREFIX + "autonomous.mode")
```
Naros: This configuration should be documented in `oracle.adoc`.

Naros (changes requested):
I've left a few comments. 

In addition to those, in the past Oracle Autonomous did not allow starting LogMiner from within the CDB and required that you do so from within the context of the PDB. Is that no longer the case, as I did not see any logic changes that forced streaming to stay within the PDB rather than switch to CDB.

Naros:
Also @martinzuern would it be possible to open a separate PR with the same Jira number in https://github.com/debezium/container-images that adds an [ADB image setup](https://www.oracle.com/autonomous-database/free-trial/#free-container-image) similar to what you see there for the SE and EE images? We can then use that image to add a testing farm setup to verify autonomous in the PR workflow.

martinzuern:
@Naros thanks for your prompt feedback! :)

> Oracle Autonomous did not allow starting LogMiner from within the CDB and required that you do so from within the context of the PDB.

Yes, that is still the case. Therefore we must not set the pdb name and add the flag for autonomous mode. Then, I was able to successfully extract changes from the ADB. Also there is undocumented behavior (that was verified with Oracle) that you can only use the GGADMIN user to access the logs.
So configuring it withoud cdb and then using the GGADMIN user mostly does the trick. In addition, you cannot add log files to the logminer session as ADB does that automagically for you.

I'll add the documentation in oracle.adoc and take a look into https://github.com/debezium/container-images.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/SqlUtils.java
martinzuern: ✅ Fixed.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
martinzuern: ✅ Fixed.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
martinzuern: ✅ Fixed.

martinzuern:
@Naros added PR for container-images https://github.com/debezium/container-images/pull/472

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
Naros: @martinzuern could we use a query to resolve autonmous mode rather than a configuration property?

In Hibernate for example, we use
```sql
SELECT sys_context('USERENV', 'CLOUD_SERVICE') FROM DUAL;
```
If that query returns `OLTP`, `DWCS` or `JDCS`, then we're connected to an autonomous environment.

Naros:
@martinzuern aside from my comment above, I've pushed a few commits. Once we settle on how to define if we are in autoonmous mode, I think we're good. I need to test the ADB container, but I'll do that tomorrow and will push one additional commit to cover adding the testing farm bits for ADB.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
martinzuern: Great catch! Implemented.

martinzuern:
@Naros Added discovery of autonomous mode and tested it. Instead of the config flag, I added some basic validation for configuration that is needed for ADB – let me know if that works for you :)

Naros:
Thanks for the update @martinzuern. So I've been playing around with your PR, and I think we need to discuss internally if we're comfortable applying these changes without running our test suite.  Right now most of the the suite is designed around having access to the redo logs, and the enforcement of archive log only mode introduces some complexities we'll need to think on.  Let me talk with the team and get back to you.

martinzuern:
@Naros thanks for driving this forward. I believe support for Autonomous Database is likely to become more and more important for Debezium’s Oracle users, as Oracle is actively promoting it. Let me know if there’s something I can do to help bring this over the finish line.

Naros:
So if I understand the documentation correctly, in Autonomous, both Oracle XStream and GoldenGate have access to mine online redo logs whereas using the pubic native LogMiner API is limited to archive logs only. 

I wonder from a pure CDC point of view, would a shift in focus toward XStream in this use case make more sense to verify if it works as expected and what may be extra steps needed for ADB + XStream @martinzuern ?

I'm still certainly open to a LogMiner implementation, even if it relies on the archive log only mode feature, but I can imagine that most users prefer or have a business need for near real time changes.

martinzuern:
@Naros Unfortunately XStream isn’t available on Autonomous Database (see [List of Unavailable Oracle Database Features](https://docs.oracle.com/en/cloud/paas/autonomous-database/serverless/adbsb/unavailable-oracle-database-features.html)), so I didn’t try that path. GoldenGate should work, but it comes with an extra license requirement.

martinzuern:
Hi @Naros, just a friendly ping — is there anything I can do to help move this across the finish line? 🙂

martinzuern:
Hi @Naros  We’ve been running this branch for the past few weeks and wanted to share an update.

The connector has been stable under load, and we haven’t observed any regressions or data inconsistencies. I also fixed an additional issue related to log handling that appeared during extended testing.

In our environment, a log switch typically occurs about once per hour. That results in roughly hourly change propagation—while not real-time, it’s still sufficient for many practical use cases with Oracle ADB.

Overall, this implementation provides a solid way to capture changes from ADB within its constraints. From our perspective, it’s ready to be merged once any remaining review points are addressed. Happy to help with the follow-ups needed to make that happen. :)

Naros:
Hi @martinzuern, thanks for the update. I've discussed this internally, and for now, the best course of action is to keep this PR on pause. The main issue for us is the inability to run the test suite against ADB. If there is any work you can do on this front to make it work, that would help expedite the merge process.

Naros:

Hi @martinzuern, please sign off all commits with:

```
git commit -s
```

Naros:
Superseded in https://github.com/debezium/debezium/pull/7942, attributions retained.

PR #6828, opened 2025-10-15
DBZ-9558 Fix JDK23+ edge case for timestamps in Postgres

shybovycha:
See [this Zullip thread](https://debezium.zulipchat.com/#narrow/channel/348249-community-postgresql/topic/JDK23.20timestamp.20exceptions/with/544111326) and the [related Jira issue](https://issues.redhat.com/browse/DBZ-9558) for more details.

Fix edge case with `timestamp` values in Postgres on JDK23+.

### DIFF
--- a/debezium-core/src/main/java/io/debezium/time/Timestamp.java
+++ b/debezium-core/src/main/java/io/debezium/time/Timestamp.java
@@ -75,18 +75,14 @@ public class Timestamp {
         }
 
         // Fix rare JDK issue where one of the components of ChronoLocalDateTime is null
-        long dateTimeInstant;
         try {
-            dateTimeInstant = dateTime.toInstant(ZoneOffset.UTC).toEpochMilli();
+            return dateTime.toInstant(ZoneOffset.UTC).toEpochMilli();
         }
         catch (NullPointerException e) {
             // Fallback for NPE from ChronoLocalDateTime#toLocalDate, see DBZ-9558
             var ignored = dateTime.toString();
-            dateTimeInstant = dateTime.toInstant(ZoneOffset.UTC).toEpochMilli();
+            return dateTime.toInstant(ZoneOffset.UTC).toEpochMilli();
         }
-        return dateTimeInstant;
-
-        return dateTime.toInstant(ZoneOffset.UTC).toEpochMilli();
     }
 
     private Timestamp() {


debezium-core/src/main/java/io/debezium/time/Timestamp.java
```diff
@@ -74,6 +74,9 @@ public static long toEpochMillis(Object value, TemporalAdjuster adjuster) {
             dateTime = dateTime.with(adjuster);
         }
 
+        // Fix rare JDK issue where one of the components of ChronoLocalDateTime is null
+        String ignored = dateTime.toString();
```
vjuranek: As it happens only from time to time and only in some JDKs (which may suggest a bug in JDK), it's probably not needed to add performance penalty for every single call. I'd call it only when NPE is thrown:

```suggestion
        long dateTimeInstant;
        try {
            dateTimeInstant = dateTime.toInstant(ZoneOffset.UTC).toEpochMilli();
        }
        catch (NullPointerException e) {
            // Fallback for NPE from ChronoLocalDateTime#toLocalDate, see DBZ-9558
            var ignored = dateTime.toString();
            dateTimeInstant = dateTime.toInstant(ZoneOffset.UTC).toEpochMilli();
        }
        return dateTimeInstant;
```

jpechane: @vjuranek Good point!

shybovycha: i would just remove the unnecessary intermediate variable and return right away

Naros:
I find this fix extremely odd. The exception in the Jira indicated that the `date` member variable of the `LocalDateTime` is `null`, but if that were the case, then I would also expect calling `toString()` on the `LocalDateTime` to also fail with the same exception since that implementation is merely:

```java
    public String toString() {
        return date.toString() + 'T' + time.toString();
    }
```

shybovycha:
@Naros indeed it is weird. figured it by trying. i assume this _might_ be some internal JDK23+ optimizations like lazy initialization or something, but could not see anything in the JDK sources. maybe some runtime optimizations? for what it's worth, this solves the problem 🤷

Naros:
Hi @shybovycha so I did some digging and I found [JDK-8322854](https://bugs.openjdk.org/browse/JDK-8322854). This looks very similar to your use case.

Can you document which specific versions of JDK 23, 24, and 25 you tested with?  It would stand to reason this may likely mean there is another bug in the JDK given that the above bug was fixed in JDK23b10.

shybovycha:
@Naros i tested with Temurin on Github Actions - `23.0.2+7`, `24.0.2+12` and `25.0.0+36`. i see that issue was resolved over a year ago, so i'd assume the fix would have been released quite a while ago at this point?

Naros:
> @Naros i tested with Temurin on Github Actions - `23.0.2+7`, `24.0.2+12` and `25.0.0+36`. i see that issue was resolved over a year ago, so i'd assume the fix would have been released quite a while ago at this point?

I would concur, unless there is a corner case that was missed :thinking:

jpechane:
@shybovycha  I think this is good to go in the current form. Thanks
Still it would be great if you manage to reproduce the issue with plain Java (somehow) and report it to OpenJDK team.

PR #6832, opened 2025-10-16
DBZ-9565 Use StrimziKafkaClsuter and remove KafkaCluster

AlvarVG:
As talked with @Naros we should not use the custom made calss KafkaCluster. Instead, we should use StrimziKafkaCluster class. In this PR, i have identified that KafkaCluster is only being used in the KafkaClusterTests class and almost everywhere else we use StrimziKafkaCluster. 

Thats is why i have deleted both files (KafkaCluster and KafkaClusterTest) and removed quite a few non used methods from KafkaClusterUtils.


jpechane:
@AlvarVG Applied, thanks

PR #6833, opened 2025-10-16
DBZ-9552 Add drop-transaction signal to Oracle connector

nathan-smit-1:
Hi there,

This relates to this discussion:  [#community-oracle > OldestScn not advancing in 3.2.1 @ 💬](https://debezium.zulipchat.com/#narrow/channel/348250-community-oracle/topic/OldestScn.20not.20advancing.20in.203.2E2.2E1/near/536311724)

The goal here is to allow dropping a transaction using a signal.  Signal exists for Oracle only at the moment.  The case we ran into is where devs with write access would create "stub" transactions when using SQL Developer.  These then get tracked as long-running transactions (i.e. transactions with a START but no further operations).  We can ask our DBAs to kill these transactions but it would be helpful for us to be able to drop them in a targeted way without needing DBA intervention.

Regarding the test:  I'm not certain of a good mechanism to grab the transaction during the IT, so there is a negative test for a txid I know doesn't exist just to test that the signal processing all works and then a unit test to test the drop functionality.  I have also tested in my own local environment and seems to be working fine.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/oracle.adoc
+++ b/documentation/modules/ROOT/pages/connectors/oracle.adoc
@@ -510,29 +510,57 @@ endif::community[]
 
 include::{partialsdir}/modules/all-connectors/con-connector-blocking-snapshot.adoc[leveloffset=+3]
 
+// Type: procedure
+// ModuleID: debezium-oracle-using-the-drop-transaction-signal-to-abandon-stale-transactions
+// Title: Using the drop-transaction signal to abandon stale transactions
 [id="oracle-drop-transaction-signal"]
 === Dropping a transaction
 
-The {prodname} Oracle connector includes a _drop-transaction_ signal that you can use to manually drop (abandon) a transaction from the LogMiner buffer at run time.
-This signal is useful for resolving situations where a long-running transaction is preventing advancement of the low watermark without generating any change events.
+You can use {prodname} signaling to send a `drop-transaction` signal to the connector at run time to remove a transaction from the LogMiner buffer.
+Removing a transaction can be helpful when a long-running transaction negatively affects connector performance.
 
-A drop-transaction signal works by removing the specified transaction from the connector's internal buffer and cleaning up any associated resources.
+A long-running transaction prevents the connector from processing subsequent change events.
+Further, the stalled transaction can result in excessive memory buffering, affect log mining continuity, and result in SCN (system change number) drift, a condition in which the connector’s current SCN falls behind the range that is available in the redo logs.
+
+During normal processing, {prodname} periodically flushes offsets to the Kafka Connect offset topic.
+If a transaction runs longer than the defined offset flush interval, the connector is unable to commit an offset.
+If the connector restarts after the last known offset is removed from the redo logs, it cannot find the resume point, which results in an error.
+
+By submitting a `drop-transaction` signal, you can avoid errors by removing a problem transaction from the connector's internal LogMiner buffer before it results in negative consequences.
+
+.Prerequisites
+* The connector is configured to use one of the available link:{debezium-signaling}#sending-signals-to-a-debezium-connector[{prodname} signaling channels].
+* You have access to the connector log file and can identify the ID of the transaction that you want to drop.
 
 [IMPORTANT]
 ====
-When a transaction is dropped, any uncommitted changes in that transaction are abandoned and no change events are emitted for them.
-Use this signal only when you are certain that the transaction should not be processed.
+After you drop a transaction, the connector abandons uncommitted changes for that transaction and does not emit change events for them.
+Use the `drop-transaction` signal only when you are certain that the transaction should not be processed.
 ====
 
-.Configure drop-transaction signal
-
-You can set the following properties in the `data` component of a signal:
-
-For example, to drop a transaction with ID `1a.2b.001f`:
+.Procedure
+* Send a signal that includes the standard `id`, `type`, and `data` properties of a {prodname} signal to your preferred signaling channel.
+In the  `data` component of the signal, specify the ID of the transaction that you want to remove.
++
+For example, to drop a transaction with ID `1a.2b.001f`, send the signal based on your chosen signal channel:
++
+.Example SQL using the `source` signal channel
 [source,sql,indent=0,subs="+attributes"]
 ----
 INSERT INTO debezium.debezium_signal VALUES('drop-tx-1', 'drop-transaction', '{"transaction-id": "1a.2b.001f"}');
 ----
++
+.Example message payload using the `kafka` signal channel
+[source,json,indent=0,subs="+attributes"]
+----
+{
+  "type": "drop-transaction",
+  "data": {
+    "transaction-id": "1a.2b.001f"
+  }
+}
+----
+
 
 // Type: concept
 // ModuleID: default-names-of-kafka-topics-that-receive-debezium-oracle-change-event-records


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -1097,6 +1097,57 @@ private void logActiveTransactions() {
...
+    /**
+     * Abandons a single transaction identified by its transaction id.
+     * This method is public so it can be invoked by external signal actions.
+     *
+     * @param transactionId the transaction id to abandon, must be in lowercase hex format
+     * @return true if the transaction was found and abandoned, false otherwise
+     * @throws InterruptedException if interrupted while performing cleanup
+     */
+    public boolean abandonTransactionById(String transactionId) throws InterruptedException {
+        if (transactionId == null || transactionId.trim().isEmpty()) {
```
Naros:
```suggestion
        if (Strings.isNullOrEmpty(transactionId)) {
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -1097,6 +1097,57 @@ private void logActiveTransactions() {
...
+            LOGGER.warn("Abandon transaction requested with null/empty transaction id");
+            return false;
+        }
+        final String txId = transactionId.trim();
```
Naros: Since the signal handler trims the value, let's avoid a second call to `trim` and omit this.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -1097,6 +1097,57 @@ private void logActiveTransactions() {
...
+        getMetrics().setActiveTransactionCount(getTransactionCache().getTransactionCount());
+        getMetrics().setBufferedEventCount(getTransactionCache().getTransactionEvents());
+
+        // Update oldest scn metric after manual abandonment
+        getTransactionCache().getEldestTransactionScnDetailsInCache().ifPresentOrElse(
+                scnDetails -> getMetrics().setOldestScnDetails(scnDetails.scn(), scnDetails.changeTime()),
+                () -> getMetrics().setOldestScnDetails(Scn.NULL, null));
+
+        // notify dispatcher to keep offsets moving
+        getEventDispatcher().dispatchHeartbeatEvent(getPartition(), getOffsetContext());
```
Naros: Do you think it might be useful to check `heartbeatsEnabled()` on the event dispatcher and conditionalize this block only if they're enabled? Then for the path where we identify heartbeats are not, we log a message that the offsets will be updated on the next committed transaction?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -1097,6 +1097,57 @@ private void logActiveTransactions() {
...
+        }
+
+        LOGGER.info("Manually abandoning transaction {} as requested", txId);
+        try {
+            cleanupAfterTransactionRemovedFromCache(transaction, true);
+        }
+        catch (Exception e) {
+            LOGGER.error("Failed to cleanup after abandoning transaction {}", txId, e);
+            getMetrics().incrementErrorCount();
+            throw e instanceof InterruptedException ? (InterruptedException) e : new InterruptedException(e.getMessage());
```
Naros:
```suggestion
            throw e;
```
Can we just let the exception bubble up as is?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/spi/DropTransactionAction.java
```diff
@@ -0,0 +1,100 @@
...
+        LOGGER.info("Processing {} signal: {}", NAME, signalPayload.id);
+
+        if (!(signalPayload.partition instanceof OraclePartition)) {
+            throw new DebeziumException(
+                    String.format("Signal '%s' with id '%s' is only supported by Oracle connector, but was sent to connector handling partition type: %s",
+                            NAME, signalPayload.id, signalPayload.partition.getClass().getSimpleName()));
+        }
+
+        final String transactionId = signalPayload.data.getString(FIELD_TRANSACTION_ID);
+        if (transactionId == null || transactionId.trim().isEmpty()) {
```
Naros:
```suggestion
        if (Strings.isNullOrEmpty(transactionId)) {
```

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/SignalsIT.java
```diff
@@ -149,4 +152,63 @@ public void signalSchemaChange() throws Exception {
...
+        final String fakeTransactionId = "fake.transaction.id.12345";
+        try (OracleConnection signalConnection = TestHelper.testConnection()) {
+            signalConnection.execute(String.format(
+                    "INSERT INTO debezium.debezium_signal VALUES('drop-tx-1', 'drop-transaction', '{\"transaction-id\": \"%s\"}')",
+                    fakeTransactionId));
+        }
+
+        // Wait for signal to be processed (signal processor runs every 5s by default)
+        // We need to wait longer than the signal processing interval to ensure the signal is processed
+        waitForAvailableRecords(6000, TimeUnit.MILLISECONDS);
```
Naros: Could we use the `LoggingInterceptor` instead and wait for a specific log entry rather than hardcoding a time that may vary on different CI load scenarios?

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/SignalsIT.java
```diff
@@ -149,4 +152,63 @@ public void signalSchemaChange() throws Exception {
...
+        connection.execute("INSERT INTO debezium.customer VALUES (2, 'After-Signal', 88.88, TO_DATE('2024-01-02', 'yyyy-mm-dd'))");
+        connection.execute("COMMIT");
+
+        // Wait for both transactions to be processed
+        waitForAvailableRecords(3000, TimeUnit.MILLISECONDS);
+
+        // Should receive 2 records (the first uncommitted transaction + the After-Signal transaction)
+        SourceRecords records = consumeRecordsByTopic(3);
+        List<SourceRecord> customerRecords = records.recordsForTopic("server1.DEBEZIUM.CUSTOMER");
+        assertThat(customerRecords).hasSize(2);
```
Naros, line 220: Let's add some content validation here that we got the right transaction events.

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -510,6 +510,31 @@ endif::community[]
 
 include::{partialsdir}/modules/all-connectors/con-connector-blocking-snapshot.adoc[leveloffset=+3]
 
+// Type: concept
```
Naros:
```suggestion
```
We'll leave this out and let @roldanbob look at how this metadata needs to be created for downstream documentation in a future step.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/spi/DropTransactionAction.java
```diff
@@ -0,0 +1,100 @@
...
+     * The streamingSource field is protected in ChangeEventSourceCoordinator, so reflection is required
+     * for connector-specific signal actions to access connector-specific streaming sources.
+     *
+     * @return the BufferedLogMinerStreamingChangeEventSource, or null if not available or wrong type
+     */
+    private BufferedLogMinerStreamingChangeEventSource getStreamingSource() {
+        final Optional<? extends io.debezium.pipeline.source.spi.StreamingChangeEventSource<?, ?>> opt = changeEventSourceCoordinator.getStreamingSource();
+        if (opt.isEmpty()) {
+            return null;
+        }
```
Naros:
```suggestion
        final Optional<?> source = changeEventSourceCoordinator.getStreamingSource();
        if (source.isEmpty()) {
            return null;
        }
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/spi/DropTransactionAction.java
```diff
@@ -0,0 +1,100 @@
...
+
+        final Object streamingSource = opt.get();
+        if (!(streamingSource instanceof BufferedLogMinerStreamingChangeEventSource)) {
+            throw new DebeziumException(
+                    String.format("Signal '%s' is only supported by BufferedLogMinerStreamingChangeEventSource, but found %s",
+                            NAME, streamingSource.getClass().getSimpleName()));
+        }
+
+        return (BufferedLogMinerStreamingChangeEventSource) streamingSource;
```
Naros:
```suggestion
        return source.filter(BufferedLogMinerStreamingChangeEventSource.class::isInstance)
                .map(BufferedLogMinerStreamingChangeEventSource.class::cast)
                .orElseThrow(() -> new DebeziumException(
                        String.format("Signal '%s' is only supported by BufferedLogMinerStreamingChangeEventSource, but found %s",
                                NAME, source.get().getClass().getSimpleName())));
```
Just modern Java API sugar.

Naros (changes requested):
This is looking good so far, I've left a few suggestions/comments. I'll take a look at your comments about the test and reach back out on that separately.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
nathan-smit-1: Will do

nathan-smit-1: Should be fixed in latest commit

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
nathan-smit-1: Thanks, I moved this back to the signal and then missed this one when removing the trims.  Should be fixed.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
nathan-smit-1: Yeah, makes sense.  Have added

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
nathan-smit-1: Done

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/spi/DropTransactionAction.java
nathan-smit-1: Done

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/SignalsIT.java
nathan-smit-1: Sure switched to the LoggingInterceptor here

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/SignalsIT.java
nathan-smit-1: So I've added some validation here but this is the test that I'm not really happy with currently as I don't have any clean way to grab a buffered transaction id.  I now just have this test to see that the signal and transactions all work but can't actually test the drop.

documentation/modules/ROOT/pages/connectors/oracle.adoc
nathan-smit-1: Removed

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/spi/DropTransactionAction.java
nathan-smit-1: Done

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/spi/DropTransactionAction.java
nathan-smit-1: Done

nathan-smit-1:
@Naros Changes made and pushed.  Think I got to everything

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/SignalsIT.java
Naros: I think what you have is fine. Given that transactions are emitted in commit order and you committed the transaction after the drop but then verify the received events before and after are received without the dropped transaction in the middle, I think that's more than sufficient.

Naros:
Thanks @nathan-smit-1, I pushed one small commit to remove a redundant variable. Starting CI.

Naros (approved):

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -510,6 +510,30 @@ endif::community[]
...
+=== Dropping a transaction
+
+The {prodname} Oracle connector includes a _drop-transaction_ signal that you can use to manually drop (abandon) a transaction from the LogMiner buffer at run time.
+This signal is useful for resolving situations where a long-running transaction is preventing advancement of the low watermark without generating any change events.
+
+A drop-transaction signal works by removing the specified transaction from the connector's internal buffer and cleaning up any associated resources.
+
+[IMPORTANT]
+====
+When a transaction is dropped, any uncommitted changes in that transaction are abandoned and no change events are emitted for them.
```
```suggestion
After you drop a transaction, the connector abandons uncommitted changes for that transaction and does not emit change events for them.
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -510,6 +510,30 @@ endif::community[]
 
 include::{partialsdir}/modules/all-connectors/con-connector-blocking-snapshot.adoc[leveloffset=+3]
 
+[id="oracle-drop-transaction-signal"]
+=== Dropping a transaction
+
+The {prodname} Oracle connector includes a _drop-transaction_ signal that you can use to manually drop (abandon) a transaction from the LogMiner buffer at run time.
+This signal is useful for resolving situations where a long-running transaction is preventing advancement of the low watermark without generating any change events.
+
+A drop-transaction signal works by removing the specified transaction from the connector's internal buffer and cleaning up any associated resources.
```
```suggestion
You can use {prodname} signaling to send a `drop-transaction` signal to the connector at run time to remove a transaction from the LogMiner buffer.
Removing a transaction can be helpful when a long-running transaction negatively affects connector performance.

A long-running transaction prevents the connector from processing subsequent change events.
Further, the stalled transaction can result in excessive memory buffering, affect log mining continuity, and result in SCN (system change number) drift, a condition in which the connector’s current SCN falls behind the range that is available in the redo logs.

During normal processing, {prodname} periodically flushes offsets to the Kafka Connect offset topic.
If a transaction runs longer than the defined offset flush interval, the connector is unable to commit an offset.
If the connector restarts after the last known offset is removed from the redo logs, it cannot find the resume point, which results in an error.

By submitting a `drop-transaction` signal, you can avoid errors by removing a problem transaction from the connector's internal LogMiner buffer before it results in negative consequences.
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -510,6 +510,30 @@ endif::community[]
...
+
+[IMPORTANT]
+====
+When a transaction is dropped, any uncommitted changes in that transaction are abandoned and no change events are emitted for them.
+Use this signal only when you are certain that the transaction should not be processed.
```
```suggestion
Use the `drop-transaction` signal only when you are certain that the transaction should not be processed.
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -510,6 +510,30 @@ endif::community[]
 
 include::{partialsdir}/modules/all-connectors/con-connector-blocking-snapshot.adoc[leveloffset=+3]
 
+[id="oracle-drop-transaction-signal"]
```
roldanbob, line 516:
```suggestion
// Type: procedure
// ModuleID: debezium-oracle-using-the-drop-transaction-signal-to-abandon-stale-transactions
// Title: Using the drop-transaction signal to abandon stale transactions
[id="oracle-drop-transaction-signal"]
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -510,6 +510,30 @@ endif::community[]
...
+
+A drop-transaction signal works by removing the specified transaction from the connector's internal buffer and cleaning up any associated resources.
+
+[IMPORTANT]
+====
+When a transaction is dropped, any uncommitted changes in that transaction are abandoned and no change events are emitted for them.
+Use this signal only when you are certain that the transaction should not be processed.
+====
+
+.Configure drop-transaction signal
```
```suggestion
.Procedure

File: documentation/modules/ROOT/pages/connectors/oracle.adoc [hunk 16]
```diff
@@ -510,6 +510,30 @@ endif::community[]
...
+This signal is useful for resolving situations where a long-running transaction is preventing advancement of the low watermark without generating any change events.
+
+A drop-transaction signal works by removing the specified transaction from the connector's internal buffer and cleaning up any associated resources.
+
+[IMPORTANT]
+====
+When a transaction is dropped, any uncommitted changes in that transaction are abandoned and no change events are emitted for them.
+Use this signal only when you are certain that the transaction should not be processed.
+====
+
```
```suggestion

The `drop-transaction` signal requires the ID of the transaction you want to drop.
You can find the ID in the connector log file.
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -510,6 +510,30 @@ endif::community[]
...
 include::{partialsdir}/modules/all-connectors/con-connector-blocking-snapshot.adoc[leveloffset=+3]
 
+[id="oracle-drop-transaction-signal"]
+=== Dropping a transaction
+
+The {prodname} Oracle connector includes a _drop-transaction_ signal that you can use to manually drop (abandon) a transaction from the LogMiner buffer at run time.
+This signal is useful for resolving situations where a long-running transaction is preventing advancement of the low watermark without generating any change events.
+
+A drop-transaction signal works by removing the specified transaction from the connector's internal buffer and cleaning up any associated resources.
+
```
roldanbob, line 534:
```suggestion
.Prerequisites

* The connector is configured to use one of the available link:{debezium-signaling}#sending-signals-to-a-debezium-connector[{prodname} signaling channels].
* You have access to the connector log file and can identify the ID of the transaction that you want to drop.
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -510,6 +510,30 @@ endif::community[]
...
+.Configure drop-transaction signal
+
+You can set the following properties in the `data` component of a signal:
+
+For example, to drop a transaction with ID `1a.2b.001f`:
+[source,sql,indent=0,subs="+attributes"]
+----
+INSERT INTO debezium.debezium_signal VALUES('drop-tx-1', 'drop-transaction', '{"transaction-id": "1a.2b.001f"}');
+----
```
```suggestion
* Send a signal that includes the standard `id`, `type`, and `data` properties of a {prodname} signal to your preferred signaling channel.  
In the  `data` component of the signal, specify the ID of the transaction that you want to remove. 
+
For example, to drop a transaction with ID `1a.2b.001f`, send the following signal:
+
[source,sql,indent=0,subs="+attributes"]
----
INSERT INTO debezium.debezium_signal VALUES('drop-tx-1', 'drop-transaction', '{"transaction-id": "1a.2b.001f"}');
----
```

nathan-smit-1:
@Naros was there anything else that you needed me to do here?

Naros:
@nathan-smit-1 just needed to fix one test issue and we had to get past the 3.4 Alpha1 release.  Going to do one more CI pass and if it passes without any issues, I'll merge.

roldanbob (changes requested):
@nathan-smit-1 @Naros 
I tried to submit this review a couple of times, but for some reason it never posted ....

I've added a few suggestions. I'm not sure whether the additional content that I added to the **Dropping a transaction** topic needs more work. Appreciate any input. 

Also, I'm wondering in the procedure section it would be best to include multiple examples, to provide guidance for using the different signaling channels?

Naros:
@roldanbob I'm not opposed to adding other examples. But I would say this is another place where it begins to raise the question about whether we shouldn't find a better way to present multiple examples. 

For example, we had discussed tabbed code blocks in the past. I'm not sure how viable that is or if it makes sense to conditionalize that in the community docs and use another way for downstream. 

Another option would be we find a way to represent the signal data agnostically at these places in the documentation, linking back to a common Signaling section that covers the semantics of how to use that agnostic data per signal channel type. For example the agnostic data might be something like:

|Type|Data|
|:---|:---|
|`drop-transaction`|`{"transaction-id": "1a.2b.001f"}`|

And then in the Signaling section we illustrate 

```
For `source` signal channels, use a database insert into the signal table:
[source,sql]
----
INSERT into <signal-table> VALUES ('<a unique id>', <type>, <data>);
----
```

Naros:
@roldanbob can you take a look at the docs and let me know if that LGTY.

Naros:
Applied @nathan-smit-1, thanks!

PR #6834, opened 2025-10-17
DBZ-9568 Debezium Quarkus & Outbox Extensions to Quarkus 3.28.4

Naros:
https://issues.redhat.com/browse/DBZ-9568

### DIFF
--- a/quarkus-debezium-parent/pom.xml
+++ b/quarkus-debezium-parent/pom.xml
@@ -24,6 +24,7 @@
         <maven.compiler.testTarget>${debezium.java.specific.target}</maven.compiler.testTarget>
         <maven.compiler.testRelease>${debezium.java.specific.target}</maven.compiler.testRelease>
 
+        <quarkus.platform.version>${quarkus.version.extension}</quarkus.platform.version>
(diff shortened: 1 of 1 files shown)


Naros:
Talked with Jiri and we concluded that we'd align both extensions on the same version, FYI @kmos.

quarkus-debezium-parent/pom.xml
```diff
@@ -24,7 +24,6 @@
         <maven.compiler.testTarget>${debezium.java.specific.target}</maven.compiler.testTarget>
         <maven.compiler.testRelease>${debezium.java.specific.target}</maven.compiler.testRelease>
 
-        <quarkus.platform.version>3.26.3</quarkus.platform.version>
```
jpechane, line 27: @Naros 
Given that Quarkus Extension still goes thorugh intensive development I'd prefer to use
`<quarkus.platform.version>${quarkus.version.extension}</quarkus.platform.version>` and remove the ability to override the version only when all connectors are properly handled in the extension.
WDYT?

Naros: That makes sense, fixed.

PR #6836, opened 2025-10-17
DBZ-9564 Fix handling of compressed transactions in MySQL

akanimesh7:


debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
```diff
@@ -177,7 +177,14 @@ public void execute(ChangeEventSourceContext context, P partition, O offsetConte
         eventHandlers.put(EventType.ROTATE, (event) -> handleRotateLogsEvent(effectiveOffsetContext, event));
         eventHandlers.put(EventType.TABLE_MAP, (event) -> handleUpdateTableMetadata(partition, effectiveOffsetContext, event));
         eventHandlers.put(EventType.QUERY, (event) -> handleQueryEvent(partition, effectiveOffsetContext, event));
-        eventHandlers.put(EventType.TRANSACTION_PAYLOAD, (event) -> handleTransactionPayload(partition, effectiveOffsetContext, event));
+        eventHandlers.put(EventType.TRANSACTION_PAYLOAD, (event) -> {
+            TransactionPayloadEventData transactionPayloadEventData = event.getData();
```
vjuranek: Is there any reason to inline `handleTransactionPayload()` method here?

akanimesh7: Yes @vjuranek 
The `handleTransactionPayload` method in it's current state, just calls the corresponding event handlers on all the events that are obtained after decompressing the compressed payload.

However, `handleEvent` method does a few more things in addition to calling appropriate event handler. Such as -- check if streaming is paused , emit heartbeat, update offsets, incrementing skip counter, etc. What this leads to is, if a compressed txn contains a lot of events, then the task would keep on running event handlers and would fail to
1. register that streaming was required to be paused
2. increment skip counter appropriately
3. produce correct offsets for a compressed transaction

Hence this refactoring is required. For confidence i have added a IT which checks two things in the process of reading an uncompressed transaction --
1. the offset constructed has correct value for events to be skipped
2. the events to be skipped is correctly honored (which is not happening currently)

jpechane: @akanimesh7 LGTM. Still just one question. Would not it be possible to refactor `handleTransactionPayload` and call it from here instead of inlining the code?

akanimesh7: @jpechane It makes sense ! 
Addressed the concern

akanimesh7:
@jpechane @vjuranek PTAL

jpechane:
@akanimesh7 Applied, thanks

PR #6838, opened 2025-10-20
DRAFT DBZ-9323: Add built-in support for acquiring MSAL access tokens 

IssamElNass:
Added Azure Core and MSAL4J packages to the Debezium BOM. Since the BOM is used in the SQL Server connector, these dependencies should be included by default.

**Reason**
This change resolves issues when using Managed Identity, which previously failed due to a missing MSAL4J package. It should fix the issue with ActiveDirectoryAccessToken authentication as well.

### DIFF
--- a/debezium-bom/pom.xml
+++ b/debezium-bom/pom.xml
@@ -54,12 +54,6 @@
         <!-- Azure Identity -->
         <version.azure.identity>1.15.4</version.azure.identity>
 
-        <!-- Azure Core -->
-        <version.azure.core>1.57.0</version.azure.core>
-
-        <!-- MSAL4J-->
-        <version.msal4j>1.23.1</version.msal4j>
-
         <!-- K8s Config Map Storage -->
         <version.kubernetes.client>6.13.4</version.kubernetes.client>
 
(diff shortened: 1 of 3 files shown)


debezium-bom/pom.xml
```diff
@@ -54,6 +54,12 @@
         <!-- Azure Identity -->
         <version.azure.identity>1.15.4</version.azure.identity>
 
+        <!-- Azure Core -->
+        <version.azure.core>1.57.0</version.azure.core>
+
+        <!-- MSAL4J-->
+        <version.msal4j>1.23.1</version.msal4j>
```
Naros: The `com.azure:azure-storage-blob` already brings in azure-core 1.54.0 and azure-identity 1.15.4 brings in msal4j 1.19.1. We should ideally try and make sure that we are aligning all Azure dependencies so that we don't run into potential conflicts across different component versions being incompatible.

IssamElNass: You’re right, I removed it from the pom.xml. Since the issue is about MS SQL Server, I added the Azure SDK BOM to the SQL Server connector as it includes all the dependencies needed for managed identity authentication to work. I’ll make sure to test it properly.

jpechane:
This is BOM only so I don't think the solve the issue.
Also I wonder if we include Azure SDK how much larger the archive would become.

IssamElNass:
@jpechane My initial idea was to include Azure Identity which should have msal4j included. But I think that the sql server connector is missing more than just that. There are other dependencies that are required such as Azure Core, ...

PR #6845, opened 2025-10-21
DBZ-9384 Debezium Extensions for Quarkus: support for Sql Server

kmos:
## Scope

In this PR we propose the Sql Server Debezium Extensions for Quarkus  `quarkus-debezium-sqlserver`. In addition to the extension, we propose also a `agroal` extension useful to parse and manage datasources based on it. Actually the `agroal` module is applied only for `sqlserver` but in a future PR can be applied to `postgres` and other datasources like `mariadb` `mysql`.

Below a list of features proposed in the PR:

- [X] sql server extension
- [X] native build for the sql server extension
- [X] `multi-engine` support for sql server
- [X] agroal extension for Debezium on Quarkus

## How to use

To explain how to use the Quarkus extension, below a minimum Jbang script and `application.properties` necessary to make CDC from a sql server source:

```java
package sqlserver;
//usr/bin/env jbang "$0" "$@" ; exit $?
//DEPS io.quarkus.platform:quarkus-bom:3.26.3@pom
//DEPS io.quarkus:quarkus-arc
//DEPS io.debezium.quarkus:quarkus-debezium-sqlserver:3.4.0-SNAPSHOT
//FILES application.properties
//JAVAC_OPTIONS -parameters
//JAVA_OPTIONS -Djava.util.logging.manager=org.jboss.logmanager.LogManager


import org.apache.kafka.connect.source.SourceRecord;
import org.slf4j.Logger;
import org.slf4j.LoggerFactory;

import io.debezium.engine.RecordChangeEvent;
import io.debezium.runtime.Capturing;
import io.quarkus.runtime.Quarkus;

@ApplicationScoped
public class sqlserver {

    private final Logger logger = LoggerFactory.getLogger(sqlserver.class);

    public static void main(String[] args) {
        Quarkus.run();
    }

    @Capturing()
    public void capture(RecordChangeEvent<SourceRecord> event) throws Exception {
        logger.info("capturing orders for destination {} with data {}");
    }
}
```

and the relative `application.properties`:

```properties
quarkus.debezium.schema.history.internal=io.debezium.relational.history.MemorySchemaHistory
quarkus.debezium.name=test
quarkus.debezium.topic.prefix=topic
quarkus.debezium.topic.naming.strategy=io.debezium.schema.DefaultTopicNamingStrategy
quarkus.debezium.database.names=inventory
quarkus.debezium.database.encrypt=false
quarkus.debezium.database.trustServerCertificate=false
quarkus.debezium.snapshot.mode=initial
quarkus.debezium.capturing.orders.destination=topic.inventory.orders
quarkus.debezium.include.schema.changes=false
quarkus.debezium.heartbeat.interval.ms=5

quarkus.datasource.db-kind=mssql
quarkus.datasource.username=sa
quarkus.datasource.password=Password!

quarkus.datasource.jdbc.url=jdbc:sqlserver://localhost:1434;encrypt=false
quarkus.datasource.jdbc.max-size=16
```

as you can see from the `application.properties` we use the same configuration property used for agroal datasources in [quarkus](https://quarkus.io/guides/datasource)

## References

#closes https://issues.redhat.com/browse/DBZ-9384

### DIFF
--- a/quarkus-debezium-parent/quarkus-debezium-agroal-parent/runtime/src/main/java/io/quarkus/debezium/agroal/configuration/AgroalDatasourceRecorder.java
+++ b/quarkus-debezium-parent/quarkus-debezium-agroal-parent/runtime/src/main/java/io/quarkus/debezium/agroal/configuration/AgroalDatasourceRecorder.java
@@ -41,8 +41,7 @@ public class AgroalDatasourceRecorder {
         String password = ConfigProvider.getConfig().getConfigValue(PREFIX + DOT + name + PASSWORD).getValue();
 
         if (jdbcUrl == null) {
-            LOGGER.warn("JDBC URL is null");
-            return null;
+            throw new NullPointerException("JDBC url is null, please specify database url");
         }
 
         return createConfiguration(name, jdbcUrl, username, password, false, dbKind);
--- a/quarkus-debezium-parent/quarkus-debezium-agroal-parent/runtime/src/main/java/io/quarkus/debezium/agroal/configuration/DatasourceParser.java
+++ b/quarkus-debezium-parent/quarkus-debezium-agroal-parent/runtime/src/main/java/io/quarkus/debezium/agroal/configuration/DatasourceParser.java
@@ -17,7 +17,7 @@ import io.debezium.jdbc.JdbcConfiguration;
 
 public class DatasourceParser {
 
-    private static final Logger logger = LoggerFactory.getLogger(DatasourceParser.class);
+    private static final Logger LOGGER = LoggerFactory.getLogger(DatasourceParser.class);
     public static final String REGEX = "jdbc:[a-z]+://(?<hostname>[^:/;?]+)(:(?<port>\\d+))?([/;](?<dbname>[^?;]+))?";
     private static final Pattern pattern = Pattern.compile(REGEX);
     private final String value;
@@ -30,7 +30,7 @@ public class DatasourceParser {
         Matcher matcher = pattern.matcher(value);
 
         if (matcher.find()) {
-            logger.trace("Found datasource definition for {}", value);
+            LOGGER.trace("Found datasource definition for {}", value);
 
             String host = matcher.group(JdbcConfiguration.HOSTNAME.name());
             String port = matcher.group(JdbcConfiguration.PORT.name());
@@ -39,7 +39,7 @@ public class DatasourceParser {
             return Optional.of(new JdbcDatasource(host, port, database));
         }
 
-        logger.warn("Unable to parse datasource: {}", value);
+        LOGGER.warn("Unable to parse datasource: {}", value);
         return Optional.empty();
     }
 
--- a/quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/EngineProcessor.java
+++ b/quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/EngineProcessor.java
@@ -6,6 +6,9 @@
 
 package io.quarkus.debezium.deployment.engine;
 
+import static io.debezium.config.CommonConnectorConfig.TOPIC_NAMING_STRATEGY;
+import static io.debezium.embedded.EmbeddedEngineConfig.OFFSET_STORAGE;
+import static io.debezium.relational.HistorizedRelationalDatabaseConnectorConfig.SCHEMA_HISTORY;
 import static io.quarkus.debezium.deployment.engine.ClassesInConfigurationHandler.POST_PROCESSOR;
 import static io.quarkus.debezium.deployment.engine.ClassesInConfigurationHandler.PREDICATE;
 import static io.quarkus.debezium.deployment.engine.ClassesInConfigurationHandler.TRANSFORM;
@@ -110,9 +113,9 @@ import io.quarkus.deployment.recording.RecorderContext;
 public class EngineProcessor {
 
     public static final List<String> PROPERTIES_WITH_CLASSES = List.of(
-            "schema.history.internal",
-            "topic.naming.strategy",
-            "offset.storage");
+            SCHEMA_HISTORY.name(),
+            TOPIC_NAMING_STRATEGY.name(),
+            OFFSET_STORAGE.name());
 
     @BuildStep
     void features(BuildProducer<FeatureBuildItem> producer, List<DebeziumExtensionNameBuildItem> debeziumExtensionNameBuildItems) {
--- a/quarkus-debezium-parent/quarkus-debezium-sqlserver-parent/integration-tests/pom.xml
+++ b/quarkus-debezium-parent/quarkus-debezium-sqlserver-parent/integration-tests/pom.xml
@@ -14,6 +14,9 @@
     <artifactId>quarkus-debezium-sqlserver-integration-tests</artifactId>
 
     <name>Quarkus Debezium :: Extension :: Sql Server :: Integration Tests</name>
+    <properties>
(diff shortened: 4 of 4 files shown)


jpechane (approved):
@kmos Great job! I've no comments so far.

@Naros Any comment from you?

quarkus-debezium-parent/quarkus-debezium-agroal-parent/runtime/src/main/java/io/quarkus/debezium/agroal/configuration/AgroalDatasourceRecorder.java
```diff
@@ -0,0 +1,69 @@
...
+        LOGGER.trace("Extracting datasource configuration for {}", name);
+
+        String jdbcUrl = ConfigProvider.getConfig().getConfigValue(PREFIX + DOT + name + JDBC_URL).getValue();
+        String username = ConfigProvider.getConfig().getConfigValue(PREFIX + DOT + name + USERNAME).getValue();
+        String password = ConfigProvider.getConfig().getConfigValue(PREFIX + DOT + name + PASSWORD).getValue();
+
+        if (jdbcUrl == null) {
+            LOGGER.warn("JDBC URL is null");
+            return null;
+        }
```
Naros, line 45: Should we have a hard exception here in this case? 

Inside the `AgroalEngineProcessor` class, we pass the supplier to the `SyntheticBeanBuildItem` and when null this leads to a basic `NullPointerException`. While the warn is good, I believe a hard exception makes more sense clarifying that the JDBC url wasn't provided.  wdyt?

quarkus-debezium-parent/quarkus-debezium-agroal-parent/runtime/src/main/java/io/quarkus/debezium/agroal/configuration/DatasourceParser.java
```diff
@@ -0,0 +1,49 @@
...
+import java.util.regex.Pattern;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.debezium.jdbc.JdbcConfiguration;
+
+public class DatasourceParser {
+
+    private static final Logger logger = LoggerFactory.getLogger(DatasourceParser.class);
```
Naros:
```suggestion
    private static final Logger LOGGER = LoggerFactory.getLogger(DatasourceParser.class);
```
For consistency, and adjust usages in class.

quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/EngineProcessor.java
```diff
@@ -109,6 +109,11 @@
 
 public class EngineProcessor {
 
+    public static final List<String> PROPERTIES_WITH_CLASSES = List.of(
+            "schema.history.internal",
+            "topic.naming.strategy",
+            "offset.storage");
```
Naros: Is there a reason why we opted to inline the names here rather than reference the exposed fields or property names on classes/interfaces in core module?

quarkus-debezium-parent/quarkus-debezium-sqlserver-parent/deployment/src/test/java/io/quarkus/debezium/sqlserver/deployment/SqlServerResource.java
```diff
@@ -0,0 +1,47 @@
...
+import java.time.temporal.ChronoUnit;
+
+import org.testcontainers.containers.MSSQLServerContainer;
+import org.testcontainers.containers.startupcheck.MinimumDurationRunningStartupCheckStrategy;
+import org.testcontainers.containers.wait.strategy.LogMessageWaitStrategy;
+import org.testcontainers.utility.DockerImageName;
+
+public class SqlServerResource {
+
+    private static final MSSQLServerContainer<?> SQL_SERVER_CONTAINER = new MSSQLServerContainer<>(DockerImageName.parse("mcr.microsoft.com/mssql/server:2019-latest"))
```
Naros, line 21: Not something specifically for this PR, but I believe we are starting to reference container image names in multiple places (JDBC sink, TestInfrastructure, and now in the Quarkus extension resources). 

I wonder in the long-term, would it make sense to create some centralized constants class for things like this that we can modify as needed, change once and it updates all modules? @jpechane any thoughts?

quarkus-debezium-parent/quarkus-debezium-sqlserver-parent/integration-tests/pom.xml
```diff
@@ -0,0 +1,228 @@
...
+                        <id>stop-sqlserver</id>
+                        <phase>post-integration-test</phase>
+                        <goals><goal>stop</goal></goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <images>
+                        <image>
+                            <!-- A Docker image using the SQL Server -->
+                            <name>mcr.microsoft.com/mssql/server:2022-latest</name>
```
Naros: Can we pull the image name up into a property? Also, our deployment uses 2019, but here we use 2022. Should we be consistent, or is the difference intended?

quarkus-debezium-parent/quarkus-debezium-sqlserver-parent/integration-tests/pom.xml
```diff
@@ -0,0 +1,228 @@
...
+                                </wait>
+                            </run>
+                            <external>
+                                <type>properties</type>
+                                <mode>override</mode>
+                            </external>
+                        </image>
+                        <image>
+                            <!-- A Docker image using the SQL Server -->
+                            <name>mcr.microsoft.com/mssql/server:2022-latest</name>
```
Naros: Same for image name.

Naros (commented):
Looks good @kmos, I've left a few inline comments.

quarkus-debezium-parent/quarkus-debezium-agroal-parent/runtime/src/main/java/io/quarkus/debezium/agroal/configuration/AgroalDatasourceRecorder.java
kmos: Totally agree, I was also in doubt. It's mostly not recoverable state

quarkus-debezium-parent/quarkus-debezium-engine-parent/deployment/src/main/java/io/quarkus/debezium/deployment/engine/EngineProcessor.java
kmos: do you mean use like `SCHEMA_HISTORY.name()`? nothing. Good point, I can use them

quarkus-debezium-parent/quarkus-debezium-sqlserver-parent/deployment/src/test/java/io/quarkus/debezium/sqlserver/deployment/SqlServerResource.java
kmos: I agree also because we should give to the users information related to the containers used by the devServices

quarkus-debezium-parent/quarkus-debezium-sqlserver-parent/integration-tests/pom.xml
kmos: Well, I can use the 2019. I used the 2022 because it's newer

kmos: As mentioned in the above, we should have a central point which is also documented about our container images.

Naros:
Applied thanks @kmos !

PR #6846, opened 2025-10-21
DBZ-806 Support Infinity and Nan floating point constants

Naros:
https://issues.redhat.com/browse/DBZ-806

This introduces a new `binary.decimal.handling.mode` configuration option that defaults to `numeric`, keeping with existing behavior; however, the option can be set to `string` to enable the serialization of infinity or not-a-number non-numeric constants.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
@@ -18,6 +18,7 @@ import java.nio.charset.StandardCharsets;
 import java.sql.Blob;
 import java.sql.Clob;
 import java.sql.SQLException;
+import java.sql.Timestamp;
 import java.sql.Types;
 import java.time.Instant;
 import java.time.ZoneId;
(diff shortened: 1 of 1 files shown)


Naros:
Test failures appear related, looking into them :thinking:

jpechane:
@Naros I'd argue that the new config option is completely unnecessary.
The infity at the FLOAT level can be handled by mapping the string value `Inf` to the correct one so `Double.parseDouble('Infinity')` will work.
If you want to add support for different handling modes then in that case you can just use the default `binary.handling.mode` as the infity will be handled properly with fix above.

Naros:
> The infity at the FLOAT level can be handled by mapping the string value `Inf` to the correct one so `Double.parseDouble('Infinity')` will work.

This is definitely more elegant. I'll adjust.

> I'd argue that the new config option is completely unnecessary. If you want to add support for different handling modes then in that case you can just use the default `binary.handling.mode` as the infity will be handled properly with fix above.

I would have opted to use `decimal.handling.mode` here if I had known that it would not affect how numeric values with a length or length minus scale greater than 19 were handled. It's quite common for users to use `INT` or `INTEGER` as type which is just an alias for `NUMBER(38, 0)`, which has a length greater than 19. So by default, the connector will rely on what users have set for `decimal.handling.mode`, which is often either `precise` or  `double` for ease of use.

Similarly, using `binary.handling.mode` leads to a similar issue with `BLOB` and `RAW` column types if the user were to change the serialization mode just to satisfy these binary float/double columns.

I just want to point out the use of "binary" here is really a poor choice on Oracle's part. It simply refers to how the value is stored in the database storage layer, but otherwise its represented as a 32-bit `Float` or a 64-bit `Double`.

So in the end, I landed on a special option to be the least disruptive, but I'm open to suggestions. Who knows, maybe the right answer here is to simply use `SpecialDecimalValue` and let `decimal.handling.mode` control it and accept the potential impact on those other numeric column types :man_shrugging: ?

jpechane:
@Naros Sorry I of course meant `decimal.handling.mode`.

jpechane:
@Naros But if you look at `DOUBLE PRECISION` for PostgreSQL or `REAL` or `DOUBLE`/`FLOAT` for Oracle these are not affected by `decimal.handling.mode`. So I'd sayjust add infinity support and we are done.

Naros:

```
git commit -s
```

PR #6847, opened 2025-10-22
DBZ-9380 Debezium Quarkus Extension for MySQL

Naros:
This bulds on atop of the existing framework in PR #6845


quarkus-debezium-parent/quarkus-debezium-mysql-parent/integration-tests/src/test/java/io/quarkus/sample/app/Scripts.java
```diff
@@ -0,0 +1,119 @@
...
+            //
+            // CREATE TABLE users(id INT IDENTITY(1,1) PRIMARY KEY, name VARCHAR(255) NOT NULL)
+            // CREATE TABLE products(id INT NOT NULL PRIMARY KEY, name VARCHAR(255) NOT NULL, description VARCHAR(255) NOT NULL)
+            //
+            // INSERT INTO users (name) VALUES ('alvar'), ('anisha'), ('chris'), ('indra'), ('jiri'), ('giovanni'), ('mario'), ('rené'), ('Vojtěch')
+            // INSERT INTO products (id, name, description) VALUES (1, 't-shirt','red hat t-shirt'), (2,'sweatshirt','blue ibm sweatshirt')
+            //
+            // EXEC sys.sp_cdc_enable_table @source_schema = 'dbo', @source_name = 'products', @role_name = NULL, @supports_net_changes = 0
+            // EXEC sys.sp_cdc_enable_table @source_schema = 'dbo', @source_name = 'users', @role_name = NULL, @supports_net_changes = 0
+            // """;
```
Naros: I missed these, will fix.

jpechane:
@kmos See

quarkus-debezium-parent/quarkus-debezium-mysql-parent/integration-tests/src/test/java/io/quarkus/sample/app/Scripts.java
kmos: I think that here can be simplified. Sql Server doesn't permits to use dir like `docker-entrypoint-initdb.d` to apply sql scripts. If mysql permits to apply in `docker-entrypoint-initdb.d` sql script with enough priviliges for CDC, I think it's simplier to use that approach.

quarkus-debezium-parent/quarkus-debezium-mysql-parent/integration-tests/pom.xml
```diff
@@ -0,0 +1,195 @@
...
+                    <execution>
+                        <id>stop-mysql</id>
+                        <phase>post-integration-test</phase>
+                        <goals><goal>stop</goal></goals>
+                    </execution>
+                </executions>
+                <configuration>
+                    <images>
+                        <image>
+                            <!-- A Docker image using the SQL Server -->
```
kmos: too much copy&paste ;)

quarkus-debezium-parent/quarkus-debezium-mysql-parent/runtime/src/main/java/io/quarkus/debezium/engine/MySqlEngineProducer.java
```diff
@@ -0,0 +1,90 @@
...
+        return new DebeziumConnectorRegistry() {
+            private final Map<String, Debezium> engines = multiEngineConfigurations
+                    .stream()
+                    .map(engine -> {
+                        EngineManifest engineManifest = new EngineManifest(engine.engineId());
+
+                        Map<String, String> debeziumConfiguration = engine.configuration();
+
+                        // remove unnecessary configuration
+                        debeziumConfiguration.remove(DATABASE_CONFIG_PREFIX + JdbcConfiguration.DATABASE.name());
```
kmos, line 64: Just to confirm: `dbname` is not available in mysql right?

quarkus-debezium-parent/quarkus-debezium-mysql-parent/deployment/src/main/java/io/quarkus/deployment/MySqlEngineProcessor.java
```diff
@@ -0,0 +1,71 @@
...
+                ExtendedSnapshotLock.class,
+                MinimalPerconaNoTableLocksSnapshotLock.class,
+                MinimalPerconaSnapshotLock.class,
+                MinimalSnapshotLock.class,
+                NoneSnapshotLock.class,
+                SelectAllSnapshotQuery.class)
+                .reason(getClass().getName())
+                .build());
+    }
+
```
kmos: What’s currently missing is only the `devservice` for MySQL. For SQL Server, there's nothing specific needed to make the database CDC-ready, but I don't think that's the case for MySQL.

It would be nice to have an empty (without any  example database data) CDC-ready database/schema (maybe the default one), but I am not sure if we already have that. Can we still use `example-mysql` n any case? I have some doubts, especially if a user expects a devservice with example data. When I started developing the extensions, I used the example one for PostgreSQL, but I’ve since changed my mind (😬 ). WDYT?

quarkus-debezium-parent/quarkus-debezium-mysql-parent/runtime/src/main/java/io/quarkus/debezium/engine/MySqlEngineProducer.java
Naros: It is.

quarkus-debezium-parent/quarkus-debezium-mysql-parent/deployment/src/main/java/io/quarkus/deployment/MySqlEngineProcessor.java
Naros: That's probably going to be tough given that a database may have its own data types that are not universal.  For example, many databases use `VARCHAR`, but Oracle because it wants to be a special snowflake uses `VARCHAR2`.

kmos (approved):

Naros:
Thanks for the review @kmos.

PR #6848, opened 2025-10-22
DBZ-9214 Remove Confluent images/references from testing suite

AlvarVG:
Removing Confluent images from testing suite.

https://issues.redhat.com/browse/DBZ-9214


AlvarVG:
Hey @Naros, as we talked yesterday. I am removing the usage of SchemaRegistry we used to have with confluent images/references. (It was not being used)

Naros (approved):
LGTM, feel free to merge after 3.4.0.Alpha1 is out.

jpechane:
@AlvarVG @AlvarVG I am reverting the change as this is used in Debezium Server tests. We need to decide if we will keep Schema Registry compat testing with Debezium Server and if yes then the classes should be moved there.

jpechane:
@AlvarVG We discussed this with @Naros and the best course forward is moving the classes from core repo into server repo and keep the tests.

AlvarVG:
@jpechane Sounds totally reasonable! Thanks for the feedback!

PR #6662, opened 2025-08-18
DBZ-9309- Postgres cdc support START_REPLICATION option streaming is on.

pranavt84:

PR #6850, opened 2025-10-23
DBZ-9566 Update `slot.stream.params` documentation to clarify use

roldanbob:
[DBZ-9566](https://issues.redhat.com/browse/DBZ-9566)

Tested in a local Antora build

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/postgresql.adoc
+++ b/documentation/modules/ROOT/pages/connectors/postgresql.adoc
@@ -4003,7 +4003,7 @@ endif::community[]
 
 `pgoutput` plugin::
 For connectors that use the `pgoutput` plug-in, add the `origin` parameter to the `slot.stream.params` property to specify whether a subscription requests the publisher to send all changes, or only local changes.
-
++
 For example to request the publisher to send all changes, that is, changes that originate locally on the publisher, and changes that were replicated to the publisher from another source, set the following value:
 
  origin=any


documentation/modules/ROOT/pages/connectors/postgresql.adoc
```diff
@@ -3985,7 +3985,40 @@ Set a delay value that is higher than the value of the {link-kafka-docs}/#connec
...
+For example,
++
+ publicationnames=dbzpublication;messages=true;binary=true;two_phase=true;origin=any
+
+
+[NOTE]
+====
+The parameters that are available for use with the `pgoutput` plug-in depend on the `proto_version`, which defines the version of the logical streaming replication protocol, and which is specified when you create the replication slot.
+If you configure {prodname} to create replication slots automatically, the value of `proto_version` is hardcoded to `1`.
+This version of the protocol does not support streaming of large in-progress transactions, or the use of two-phase commits.
```
jpechane: @roldanbob  These two changes chose as an exmaple could not be used in Debezium OOTB. So either different examples should be use or maybe the paragraph proposing the change of protocol version should be removed even if it was mentioned on the original Jira.

For reference PR https://github.com/debezium/debezium/pull/6662 provides support for larg TX.

roldanbob: Hey, @jpechane I'm a little bit unclear on how to address your comment. Which two changes are you referring to -- do you mean both the `decoderbufs` and `pgoutput` examples?  
I can certainly remove the reference to the `proto_version` and to the ability to process large in-progress transactions.  
Can you provide examples of some of the specific parameters that are valid for use with each of the decoder plug-ins, and in the case of `pgoutput` would be valid regardless of the the `proto_version` setting? Could a user specify the `proto_version` as one of the parameters for this property?

jpechane: As was mentioned in Jira, there is only `debug-mode` option available for protobuf. This option just prints logical messages to the database console/log without passing them to Debezium.
For pgoutput I'd recommend to use for example `origin`.

roldanbob: Ah, ok, thanks. 
Yes, I saw the reference to `debug-mode` for protobuf, but I assumed that other options might be available, and was looking to point to some sort of comprehensive list. Similarly, for decoderbufs, I was hoping to list a range of options. 
I've addressed your feedback. Please take another look and let me know if this looks good. Thanks.

jpechane:
@roldanbob Applied, thanks

PR #6859, opened 2025-10-28
DBZ-9623: Allow table named CYCLE in MySQL

ochedru:
https://issues.redhat.com/browse/DBZ-9623


jpechane:
@ochedru Applied, thanks

PR #6862, opened 2025-10-30
DBZ-9634 Unparseable DDLs for Mariadb

pricelessjunk:
Solution for the unparseable DDLs in MariaDb

- Rename tables
- alter table ... convert to charset

### DIFF
--- a/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mariadb/generated/MariaDBParser.g4
+++ b/debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mariadb/generated/MariaDBParser.g4
@@ -758,8 +758,8 @@ alterSpecification
     | ENABLE KEYS                                                         # alterByEnableKeys
     | RENAME renameFormat = (TO | AS)? (uid | fullId)                     # alterByRename
     | ORDER BY uidList                                                    # alterByOrder
-    | CONVERT TO CHARACTER SET charsetName (COLLATE collationName)?       # alterByConvertCharset
-    | DEFAULT? CHARACTER SET '=' charsetName (COLLATE '=' collationName)? # alterByDefaultCharset
+    | CONVERT TO charSet charsetName (COLLATE collationName)?             # alterByConvertCharacterset
+    | DEFAULT? charSet '=' charsetName (COLLATE '=' collationName)?       # alterByDefaultCharset
     | DISCARD TABLESPACE                                                  # alterByDiscardTablespace
     | IMPORT TABLESPACE                                                   # alterByImportTablespace
     | FORCE                                                               # alterByForce
@@ -856,7 +856,7 @@ dropSequence // sequence is MariaDB-specific only
 //    Other DDL statements
 
 renameTable
-    : RENAME TABLE ifExists? renameTableClause (',' renameTableClause)*
+    : RENAME (TABLE | TABLES) ifExists? renameTableClause (',' renameTableClause)*
     ;
 
 renameTableClause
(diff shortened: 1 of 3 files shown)


debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mariadb/generated/MariaDBParser.g4
```diff
@@ -758,7 +758,8 @@ alterSpecification
     | ENABLE KEYS                                                         # alterByEnableKeys
     | RENAME renameFormat = (TO | AS)? (uid | fullId)                     # alterByRename
     | ORDER BY uidList                                                    # alterByOrder
-    | CONVERT TO CHARACTER SET charsetName (COLLATE collationName)?       # alterByConvertCharset
+    | CONVERT TO CHARACTER SET charsetName (COLLATE collationName)?       # alterByConvertCharacterset
```
jpechane: Maybe this coud be changed to

```suggestion
    | CONVERT TO (CHARACTER SET) | CHARSET charsetName (COLLATE collationName)?       # alterByConvertCharacterset
```
Also `alterByDefaultCharset` might need a similar update

debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mariadb/generated/MariaDBParser.g4
```diff
@@ -856,7 +857,7 @@ dropSequence // sequence is MariaDB-specific only
 //    Other DDL statements
 
 renameTable
-    : RENAME TABLE ifExists? renameTableClause (',' renameTableClause)*
+    : RENAME (TABLE|TABLES) ifExists? renameTableClause (',' renameTableClause)*
```
jpechane:
```suggestion
    : RENAME (TABLE | TABLES) ifExists? renameTableClause (',' renameTableClause)*
```

pricelessjunk: Done

debezium-ddl-parser/src/main/antlr4/io/debezium/ddl/parser/mariadb/generated/MariaDBParser.g4
pricelessjunk: Took the solution. I just adjusted the brackets a little.

jpechane: @pricelessjunk I am not sure the parentheses are correct now, but let's wait for tests.
Also could oyu please afdress `alterByDefaultCharset` rule below that I believe sufferes from the same defect.

pricelessjunk: Hey ho.

I tested the brackets locally. Should be fine.

Regarding the default charset, it seems to be okay. I added in another 2 lines in the test sql.

The lines for default charset
```sql
MariaDB [mysql]> ALTER TABLE EMP DEFAULT CHARACTER SET = utf8mb4 COLLATE = utf8mb4_unicode_ci;
Query OK, 0 rows affected (0.014 sec)
Records: 0  Duplicates: 0  Warnings: 0

MariaDB [mysql]> ALTER TABLE EMP DEFAULT CHARSET = utf8mb4 COLLATE = utf8mb4_unicode_ci;
Query OK, 0 rows affected (0.008 sec)
Records: 0  Duplicates: 0  Warnings: 0
```

pricelessjunk: Hey ho,
Builds should be done. The default one was fine. 😃

jpechane:
@pricelessjunk Well, I still have one more proposal for the change. While reviewing I noticed there is `charSet` rule that defines the different alternatives. Can you please take a look? Maybe that would be the best option to re-use it?

pricelessjunk:
> @pricelessjunk Well, I still have one more proposal for the change. While reviewing I noticed there is `charSet` rule that defines the different alternatives. Can you please take a look? Maybe that would be the best option to re-use it?

Hey @jpechane 
Good catch. Made the changes. PTAL.

jpechane:
@pricelessjunk Applied, thanks!

PR #6864, opened 2025-10-31
DBZ-9638 Remove superfluous metrics updates

Naros:
https://issues.redhat.com/browse/DBZ-9638


jpechane:
@Naros Applied, thanks

PR #6868, opened 2025-11-03
DBZ-9640 Update to use only `StrimziKafkaCluster`

see-quick:
This PR updates the codebase to use `StrimziKafkaCluster` instead of `StrimziKafkaContainer` because of future/planned deprecations/removals from the public API.


vjuranek (commented):
Test failures are related, please take a look @see-quick

see-quick:
> Test failures are related, please take a look @see-quick

I have fixed it. Now tests are passing, now I would just prefix all commits with the JIRA number, and we should be fine.

jpechane:
@see-quick Applied, thanks!

PR #6871, opened 2025-11-06
DBZ-9636 Reduce Oracle heap cache footprint

Naros:
https://issues.redhat.com/browse/DBZ-9636

- [x] Remove the caching of the `LogMinerDmlEntry` in the cache. This introduces duplicity with the EventType, object owner and name, which is unnecessary.
- [ ] Look at whether the collection footprint in `MemoryLogMinerTransactionCache` can be reduced.
- [ ] Look at deduplicating new/old values
- [ ] Optionally introduce value compression


Naros:
So far tests failures are unrelated to LogMiner changes.

PR #6873, opened 2025-11-07
DBZ-9504 Always return NoOpLineageEmitter when open lineage integration not enabled

seanshi007:
https://issues.redhat.com/browse/DBZ-9504

Despite that debezium would expect topic.prefix to be unique for each connector, there are at least two issues with the introduction of open lineage integration. 

1. It's a bit invasive that it touches all connectors main execution path and despite there is the default option to not enable open lineage integration, it still makes a lot of debezium users feels this new feature (by creating emitters and get emitters and throw exceptions if can't get the emitter). And as discussed in [#community-db2 > DebeziumOpenLineageEmitter not initialized for connector @ 💬](https://debezium.zulipchat.com/#narrow/channel/348103-community-db2/topic/DebeziumOpenLineageEmitter.20not.20initialized.20for.20connector/near/546845894), I think we should make this feature not so invasive by properly honor the enabled flag. 
2. When exception is thrown, because the ConnectorContext is a record and contains the whole connector config, the exception will print out all the sensitive information such as DB connection string, password, etc. 

There are multiple ways to address this problem:
1. Similar to JMX metrics, where the metric name will consist of the tags as well which gives debezium users a lot of flexibility to work around of similar issue. (As a side note, I generally feel we shouldn't rely on topic.prefix as the only unique thing for connector, if we need such a unique property, we should introduce one rather than rely on topic.prefix which also has impact on how the event looks like in the payload). So, if we include open lineage tags as part of emitter key, the uniqueness can also be resolved. 
2. This PR to always use NoOpLineageEmitter when open lineage integration not enabled

This could be considered just as a starting point, as we would like to get this resolved ASAP. Open to further discussions as this PR is just one way to resolve these issues.

### DIFF
--- a/debezium-openlineage/debezium-openlineage-api/src/main/java/io/debezium/openlineage/DebeziumOpenLineageEmitter.java
+++ b/debezium-openlineage/debezium-openlineage-api/src/main/java/io/debezium/openlineage/DebeziumOpenLineageEmitter.java
@@ -5,6 +5,8 @@
  */
 package io.debezium.openlineage;
 
+import static io.debezium.openlineage.OpenLineageConfig.OPEN_LINEAGE_INTEGRATION_ENABLED;
+
 import java.util.List;
 import java.util.Map;
 import java.util.ServiceLoader;
@@ -43,6 +45,8 @@ public class DebeziumOpenLineageEmitter {
     // Thread-safe map to store emitters per connector
     private static final ConcurrentHashMap<String, LineageEmitter> emitters = new ConcurrentHashMap<>();
 
+    private static final NoOpLineageEmitter noOpLineageEmitter = new NoOpLineageEmitter();
+
     /**
      * Initializes the lineage emitter with the given configuration for a specific connector.
      * <p>
@@ -71,6 +75,9 @@ public class DebeziumOpenLineageEmitter {
      * @param connectorContext The connector context
      */
     public static void init(ConnectorContext connectorContext) {
+        if (isOpenLineageDisabled(connectorContext)) {
+            return;
+        }
 
         LOGGER.debug("Calling init for connector with context {}", connectorContext);
 
@@ -170,11 +177,19 @@ public class DebeziumOpenLineageEmitter {
     }
 
     private static LineageEmitter getEmitter(ConnectorContext connectorContext) {
+        if (isOpenLineageDisabled(connectorContext)) {
+            return noOpLineageEmitter;
+        }
+
         LineageEmitter emitter = emitters.get(connectorContext.toEmitterKey());
         LOGGER.debug("Available emitters {}", emitters);
         if (emitter == null) {
             throw new IllegalStateException("DebeziumOpenLineageEmitter not initialized for connector " + connectorContext + ". Call init() first.");
         }
         return emitter;
+    }
+
+    private static boolean isOpenLineageDisabled(ConnectorContext connectorContext) {
+        return !Boolean.parseBoolean(connectorContext.config().get(OPEN_LINEAGE_INTEGRATION_ENABLED));
     }
 }


mfvitale:
> The change itself LGTM, but AFAICT (I'm not very familiar with OpenLIneage ingeration in Debezium), when OpenLineage is not configured, Debezium already uses no-op emitter. So has to be some other issue which this PR doesn't fix, it's just a workaround for it and I'm afraid the error will pop up somewhere else.

You're right. The problem is that sharing the same `topic.prefix` is not an option for OpenLineage since it is used to uniquely identify the task and create the relative emitter. So if multiple connector shares it, i f a connector is stopped/restart the cleanup code will remove the emitter from the map and this cause the error. 

One possibility here is to maybe use the `<connector name>.<topic.prefix>.<taskId>` as identifier. But I need to think more about it.

mfvitale:
Looking to it better this PR should work as a quick workaround meanwhile we decide the final solution.

seanshi007:
Thanks both, as I have shared in the community channel. We have thousands of customers potentially impacted by this. So the goal is to try to get this resolved ASAP. 

There are essentially 2 problems:
1. Whether open lineage uniqueness should depends on topic.prefix
2. Despite topic.prefix is unique (in our dedicated kafka connect cluster case), we still occasionally see this exception, and that exception will leak all sensitive information in the log

DBZ-9504 solves these 2 problems by honor open lineage enabled config, i.e. as long as people not interested in open lineage, they should have zero code path that could run into problems, which is the approach taken by this pull request. 

Other approaches could be using a different unique identifier like @mfvitale mentioned. But like I pointed out in problem 2, we still occasionally see secrets being leaked due to this exception being thrown despite uniqueness is guaranteed (although we didn't spend enough time to further look into why uniqueness still leads to this exception).  

I would personally prefer to merge this pull request as a quick workaround to unblock all potentially impacted debezium users. And the community could take more time to think about these problems and design a better solution.

mfvitale:
> we still occasionally see secrets being leaked due to this exception being thrown despite uniqueness is guaranteed (although we didn't spend enough time to further look into why uniqueness still leads to this exception).

Just out of curiosity are you on Kafka Connect 4.1?

seanshi007:
No, all of our customers are on Kafka Connect 3.6.

seanshi007:
> Looking to it better this PR should work as a quick workaround meanwhile we decide the final solution.

@vjuranek @mfvitale Shall we merge this PR and maybe build a debezium 3.3.2?

jpechane:
Have you tried to build and test it locally? Does it really fixes the issue?

seanshi007:
@jpechane Yes, I've just built it locally and ran a PG connector locally, everything works as expected. I believe it really fixes the 2 problems outlined. This PR essentially always uses the NoOpLineageEmitter and works fine regardless of what connector configuration is. I believe that's what each new feature should be.

debezium-openlineage/debezium-openlineage-api/src/main/java/io/debezium/openlineage/DebeziumOpenLineageEmitter.java
```diff
@@ -71,6 +75,9 @@ public static void init(Map<String, String> configuration, String connectorTypeN
      * @param connectorContext The connector context
      */
     public static void init(ConnectorContext connectorContext) {
+        if (!Boolean.parseBoolean(connectorContext.config().get(OPEN_LINEAGE_INTEGRATION_ENABLED))) {
+            return;
+        }
```
Naros: Could we at least move the parse portion into a `isOpenLineageEnabled` function?

seanshi007: Sure, moved to a `isOpenLineageDisabled` method.

vjuranek:
I'm going to merge it as a quick workaround and let @mfvitale adjust it as he wants one he is back.

vjuranek:
Merged, thanks @seanshi007 !

PR #6875, opened 2025-11-09
DBZ-9657 fix to handle space in collection.include.list

sumitjha4321:
# Fix: Trim whitespace from collection.include.list to prevent incorrect regex patterns in MongoDB change streams

## Problem

When `collection.include.list` contains whitespace after commas (either from multiline YAML processing via Strimzi or explicit single-line format with spaces), the MongoDB connector generates a change stream aggregation pipeline with an incorrect regex pattern that includes spaces. This causes the regex pattern to fail to match collection names correctly, resulting in change stream events not being captured for collections listed after the first one.

### Root Cause

The issue occurs in two scenarios:

1. **Multiline YAML format** (when deployed via Strimzi):
   ```yaml
   collection.include.list: >-
     db1.col1,
     db1.col2
   ```
   When Strimzi processes the multiline YAML value, it converts newlines to comma followed by space, resulting in `"db1.col1, db2.col2"` being passed to Debezium.

2. **Single-line format with spaces after commas**:
   ```yaml
   collection.include.list: db1.col1, db1.col2
   ```

When `ChangeStreamPipelineFactory` processes this comma-separated list and converts it to a regex pattern, it generates:
- **Broken regex:** `"db1.col1| db1.col2"` (space after pipe)
- **Expected regex:** `"db1.col1|db1.col2"` (no space)

The space in the regex pattern causes the second collection to never match, as the regex expects a literal space character before the collection name.

### Evidence

Connector log showing the broken regex pattern:
```
[2025-11-09 13:05:07,342] INFO [realtime-stream-test|task-0] Effective change stream pipeline: 
...
{"namespace": {"$regularExpression": {"pattern": "db1.col1| db1.col2", "options": "i"}}}
...
```

**Key issue:** Notice the space after the pipe character (`|`) in the pattern.

## Solution

This PR fixes the issue by trimming whitespace from collection names when parsing `collection.include.list` configuration values before converting them to regex patterns. This ensures that:

1. Whitespace is removed from collection names during parsing
2. Regex patterns are generated without spaces
3. Both single-line and multiline YAML configurations work correctly
4. The fix is defensive and handles various whitespace scenarios

### Changes

- Modified `ChangeStreamPipelineFactory` (or relevant parsing method) to trim whitespace from collection names when processing `collection.include.list`
- Added/updated tests to verify whitespace handling in comma-separated collection lists

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ChangeStreamPipelineFactory.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ChangeStreamPipelineFactory.java
@@ -6,7 +6,6 @@
 package io.debezium.connector.mongodb;
 
 import static java.util.function.Predicate.not;
-import static java.util.stream.Collectors.joining;
 import static java.util.stream.Collectors.toList;
 
 import java.util.ArrayList;
@@ -27,6 +26,7 @@ import com.mongodb.client.model.changestream.OperationType;
 import io.debezium.DebeziumException;
 import io.debezium.connector.mongodb.Filters.FilterConfig;
 import io.debezium.data.Envelope;
+import io.debezium.util.Strings;
 
 /**
  * A factory to produce a MongoDB change stream pipeline expression.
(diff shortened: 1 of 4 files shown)


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/ChangeStreamPipelineFactory.java
```diff
@@ -347,4 +348,23 @@ private static List<Bson> splitNamespaceList(String input) {
...
+
+    /**
+     * Converts a comma-separated list of values to a regex pattern by splitting on commas,
+     * trimming whitespace from each part, and joining with pipe (|) characters.
+     * This handles cases where multiline YAML configurations result in spaces after commas.
+     *
+     * @param input the comma-separated input string
+     * @return the regex pattern with values joined by pipes, with no spaces
+     */
+    private static String commaSeparatedToRegex(String input) {
```
Naros: Could we somehow leverage `io.debezium.util.Strings` here?

sumitjha4321: @Naros Made changes as you suggested. could you please review ?

jpechane:
Test failures unrelated
@sumitjha4321  Applied, thanks!

PR #6876, opened 2025-11-09
DBZ-9660 Add LogMiner session hash/sort area size tuning options

Naros:
https://issues.redhat.com/browse/DBZ-9660

This introduces two undocumented, internal tuning knobs for adjusting the Oracle JDBC session's `HASH_AREA_SIZE` and the `SORT_AREA_SIZE` values that can directly improve the performance of LogMiner.


jpechane:
@Naros Applied, thanks

PR #6877, opened 2025-11-09
DBZ-9661 LogMiner to always exclude tables prefixed `MLOG$`

Naros:
https://issues.redhat.com/browse/DBZ-9661

When refreshing a materialized view, Oracle uses `MLOG$%` (materialized view log) tables to track the changes that will eventually be placed into the materialized view. In an effort to optimize the LogMiner data fetch, when the query filter mode is not set or an exclude list is specified, the query will now always exclude such log tables by default.


jpechane:
@Naros Applied, thanks

PR #6878, opened 2025-11-10
DBZ-9662 Remove left join in LogMiner log query

Naros:
https://issues.redhat.com/browse/DBZ-9662

The left join in this query does not always perform well in some environments that have more archive log records. It's actually unnecessary, and a code-based deduplication pass can be used in its place as a more efficient alternative.


jpechane:
@Naros Applied, thanks

PR #6879, opened 2025-11-10
DBZ-9372 Timezone conversion support for Epoch time fields in source block

chaitradalawai:
https://issues.redhat.com/browse/DBZ-9372

Updated the **TimezoneConverter** SMT to support timezone conversion for **epoch-based fields** (ts_ms, ts_us, ts_ns) present under the **Debezium source block**.
Added logic to identify and convert **epoch-type fields (INT64)** in the source struct when included in the **include.list**.

Added unit test to verify:
- Correct epoch conversion for ts_ms, ts_us, ts_ns fields.
- Handling of multiple fields in a single include list.
- Validation that conversion occurs only for configured fields.

### DIFF
--- a/documentation/modules/ROOT/pages/transformations/timezone-converter.adoc
+++ b/documentation/modules/ROOT/pages/transformations/timezone-converter.adoc
@@ -52,11 +52,6 @@ The `include.list` and `exclude.list` configuration options are mutually exclusi
 
 The SMT also allows conversion of event metadata fields in the source information block, such as `ts_ms` to the target timezone. In order to convert the metadata fields, you must include the `source` prefix in the `fieldname` of the `include.list` or `exclude.list` configuration option.
 
-[NOTE]
-====
-If the schema for timestamp fields in the source information block, like `ts_ms`, is currently set to `INT64`, which is not a timestamp type, future releases aim to support the conversion of such fields by introducing compatibility for a timestamp schema.
-====
-
 // Type: concept
 // Title: Example: Basic {prodname} timezone converter SMT configuration
 // ModuleID: example-basic-debezium-timezone-converter-smt-configuration


debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -361,12 +361,14 @@ private void handleValueForFields(Struct value, Type type, Set<String> fields) {
         for (org.apache.kafka.connect.data.Field field : schema.fields()) {
             String schemaName = field.schema().name();
 
-            if (schemaName == null) {
+            boolean isEpochType = field.schema().type() == Schema.Type.INT64;
```
Naros: I'm afraid this is a bit too relaxed, as it allows for the conversion of any INT64 field.

The converter should only be permitted to convert the following:
- Logical types that are within the supported logical name list
- Source information fields named `ts_ms`, `ts_us`, and `ts_ns`, which happen to be INT64.

Using your test case you added in the PR, introduce a new `INT64_SCHEMA` field in the `source` schema, .e.g.
```java
        Schema sourceSchema = SchemaBuilder.struct()
                .name("source")
                .field("ts_ms", Schema.INT64_SCHEMA)
                .field("ts_us", Schema.INT64_SCHEMA)
                .field("ts_ns", Schema.INT64_SCHEMA)
                .field("table", Schema.STRING_SCHEMA)
                .field("random", Schema.INT64_SCHEMA)
                .build();

        Struct source = new Struct(sourceSchema)
                .put("ts_ms", 1762652621071L)
                .put("ts_ns", 1762652621071088000L)
                .put("ts_us", 1762652621071088L)
                .put("random", 125L)
                .put("table", "customers");
````

Now, include the `customers:source.random` in the inclusion list, and the field should not be converted; however, it is and its value is set to 19800125.

File: debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java [hunk 2]
```diff
@@ -387,10 +389,42 @@ private void handleValueForFields(Struct value, Type type, Set<String> fields) {
...
+        else if (fieldName.endsWith("ts_ns")) {
+            scaleFactor = 1_000_000;
+        }
+        else {
+            scaleFactor = 1;
+        }
+
+        long epochMillis = epochValue / scaleFactor;
+
+        ZoneOffset offset = zoneId.getRules().getOffset(Instant.ofEpochMilli(epochMillis));
```
Naros: Take a look at `Conversions` class method's `toInstantFromMicros` and `toInstantFromMillis`. Perhaps we introduce a new method there that handles nanos?

debezium-core/src/main/java/io/debezium/transforms/TimezoneConverter.java
```diff
@@ -387,10 +389,42 @@ private void handleValueForFields(Struct value, Type type, Set<String> fields) {
...
         String fieldName = field.name();
         Schema schema = field.schema();
-        Object newValue = getTimestampWithTimezone(schema.name(), struct.get(fieldName));
+        Object fieldValue = struct.get(fieldName);
+        Object newValue = fieldValue;
+        if (fieldValue != null) {
+            if (schema.name() != null) {
+                newValue = getTimestampWithTimezone(schema.name(), struct.get(fieldName));
+            }
+            else if (schema.name() == null && schema.type() == Schema.Type.INT64) {
```
Naros: This aligns with my comment above about being too relaxed and should be restricted to the specific source information block field names and only within the source info block.

Naros (approved):
Thanks @chaitradalawai, this LGTM.  Great work and congrats on your first contribution to Debezium :tada:.

Let's see how CI runs, but I suspect it should be all good.

jpechane:
Test failures unrelated.

@chaitradalawai Nice job, thanks!

PR #6882, opened 2025-11-10
DBZ-9668 Clarify Debezium Server documentation for signaling via source table channel

jfelipends:
Include notes in the signaling documentation to explicitly describe the required configuration for Debezium Server to work with the source signaling channel.

Fixes https://github.com/debezium/dbz/issues/1433

### DIFF
--- a/documentation/modules/ROOT/pages/configuration/signalling.adoc
+++ b/documentation/modules/ROOT/pages/configuration/signalling.adoc
@@ -79,7 +79,7 @@ For example,
 +
 For more information about setting the `signal.data.collection` property, see the table of configuration properties for your connector.
 
-NOTE: If you use the `debezium.source.table.include.list` property, you do not need to include the signaling data collection in it. However, if you use the `debezium.source.column.include.list` property, you must include the signaling table columns (`id`, `type`, and `data`).
+NOTE: If you use the `table.include.list` property, you do not need to include the signaling data collection in it. However, if you use the `column.include.list` property, you must include the signaling table columns (`id`, `type`, and `data`).
 
 .Additional resources
 * xref:format-for-specifying-fully-qualified-names-for-data-collections[Formats for specifying fully qualified names for data collections].


documentation/modules/ROOT/pages/configuration/signalling.adoc
```diff
@@ -79,6 +79,8 @@ For example,
 +
 For more information about setting the `signal.data.collection` property, see the table of configuration properties for your connector.
 
+NOTE: If you use the `debezium.source.table.include.list` property, make sure to include the signaling data collection in it. Also, if you use the `debezium.source.column.include.list` property, include the columns of the signaling table (`id`, `type`, and `data`) as well.
```
Naros: Hi @jfelipends, users should not have to add the signal data collection to the table include list since 2.0, since DBZ-3293 handles adding this automatically for you.

jfelipends: Hi @Naros, I tried this in Debezium Server 3.2, and the only way I got it to work was by adding it to the include list. Maybe it’s a bug in Debezium Server though?

Naros: Which specific connector were you using @jfelipends ?

jfelipends: I'm using Debezium Server with MySQL source and Redis Sink.

jpechane: @jfelipends Could we please go back to the Jira issue and try to find the root cause?
Could you please share the Server config you used before you added the additional the table to the list and the columns to the list?

I wonder if the root cause is that you have colmun.include.list added and while Debezium adds the table automativcally to the list of captured tables it should add columns automatically to the list of capture columns.

jfelipends: @jpechane in the Jira Issue, I did mentioned the warning that I was getting when I was trying to perform the incremental snapshot via source (`The signal event 'Struct{}' should have 3 fields but has 0`). At the time I added only these properties to the application.properties:
```
debezium.source.signal.enabled.channels=source
debezium.source.signal.data.collection=debezium_signal
```
When I added:
```
debezium.source.table.include.list=..., debezium_signal
```
The problem still happened.

And the incremental snapshot only worked when I added:
```
debezium.source.column.include.list=..., debezium_signal.id, debezium_signal.type, debezium_signal.data
```
And gave INSERT privilege to Debezium MySQL user for the `debezium_signal` table.

jpechane: But what happens if you only add columns into `column.include.list` but not table into `table.include.list`?

jfelipends: @jpechane Just tested here, and in this scenario it works.

jfelipends: In this case it's considered a bug in Debezium Server or we can just add the `column.include.list` part to documentation? @Naros @jpechane

jpechane: My proposal is to add it to the documentation as a limitation and create a follw-up Jira to remove that limitation.

jfelipends: @jpechane So, I will keep the `NOTE` part regarding to the `column.include.list` in this PR. And then create the follow-up on Jira, ok?

Naros:

Hi @jfelipends, please sign off all commits with:

```
git commit -s
```

documentation/modules/ROOT/pages/configuration/signalling.adoc
jfelipends: Hi @jpechane just did the changes.

documentation/modules/ROOT/pages/configuration/signalling.adoc
```diff
@@ -79,6 +79,8 @@ For example,
 +
 For more information about setting the `signal.data.collection` property, see the table of configuration properties for your connector.
 
+NOTE: If you use the `debezium.source.table.include.list` property, make sure to include the signaling data collection in it. Also, if you use the `debezium.
```
Naros: This note seems to contradict
* https://github.com/debezium/debezium/pull/6882#discussion_r2524337568
* https://github.com/debezium/debezium/pull/6882#discussion_r2527673722

I thought we were just keeping a note about `column.include.list` usage, given that adding it without adding it to the `table.include.list` worked, or am I missing something?

jfelipends: Sure. I was testing the commit signing thing and forgot to push the correct note.

documentation/modules/ROOT/pages/configuration/signalling.adoc
```diff
@@ -79,6 +79,8 @@ For example,
 +
 For more information about setting the `signal.data.collection` property, see the table of configuration properties for your connector.
 
+NOTE: If you use the `debezium.source.table.include.list` property, you do not need to include the signaling data collection in it. However, if you use the `debezium.source.column.include.list` property, you must include the signaling table columns (`id`, `type`, and `data`).
```
vjuranek: `debezium.` prefix is used only for DBZ server, while this documentation is included also in connector documentation and uses Kafka Connect config (i.e. without `debezium.` prefix). We should keep it unified. I'd recommend to remove `debezium.` prefix from the config options mentioned here.

jfelipends: @vjuranek I saw that in most of the documentation the `.source` part is also omitted. Like [here](https://debezium.io/documentation/reference/3.5/connectors/mysql.html#mysql-property-table-include-list). Do you think it's ok to it keep only  `table.include.list` and `column.include.list`?

jpechane:
@jfelipends Could you please remove the merge commit? We prefer linear history. Thanks

documentation/modules/ROOT/pages/configuration/signalling.adoc
vjuranek: @jfelipends yes, I'm sorry, I meant `debezium.source.` prefix

jpechane:
@jfelipends Applied, thanks!

PR #6885, opened 2025-11-13
DBZ-9679 Introduce a passthrough collection naming strategy

Naros:
https://issues.redhat.com/browse/DBZ-9679


jpechane:
@Naros Applied, thanks

PR #6886, opened 2025-11-13
Added record_value support for DELETE and tombstone events

apache88:
- debezium has only support record_key for DELETE event, however record_value should also be supported
- added record_value support partially, more time is needed to make it work
- added unit test works well, added intergration test failing

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/RecordWriter.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/RecordWriter.java
@@ -133,6 +133,12 @@ public class RecordWriter {
 
     private int bindFieldValuesToQuery(JdbcSinkRecord record, QueryBinder query, int index, Struct source, Set<String> fieldNames) {
         for (String fieldName : fieldNames) {
+
+            // TODO: Remove this workaround after fixing the root cause in the schema generation
+            if (fieldName.equals("nick_name_")) {
+                fieldName = "nick_name$";
+            }
+
             final JdbcFieldDescriptor field = record.jdbcFields().get(fieldName);
 
             Object value;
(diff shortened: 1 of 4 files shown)


debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/RecordWriter.java
```diff
@@ -133,6 +133,12 @@ private int bindNonKeyValuesToQuery(JdbcSinkRecord record, QueryBinder query, in
 
     private int bindFieldValuesToQuery(JdbcSinkRecord record, QueryBinder query, int index, Struct source, Set<String> fieldNames) {
         for (String fieldName : fieldNames) {
+
+            // TODO: Remove this workaround after fixing the root cause in the schema generation
```
Naros, line 137: Can you explain what's wrong in the schema generation?

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfigTest.java
```diff
@@ -66,7 +66,7 @@ public void testNonDefaultInsertModeProperty() {
     public void testNonDefaultDeleteEnabledPropertyWithPrimaryKeyModeNotRecordKey() {
         final Map<String, String> properties = new HashMap<>();
         properties.put(JdbcSinkConnectorConfig.DELETE_ENABLED, "true");
-        properties.put(JdbcSinkConnectorConfig.PRIMARY_KEY_MODE, "record_value");
+        properties.put(JdbcSinkConnectorConfig.PRIMARY_KEY_MODE, "kafka");
```
Naros, line 69: Is there a specific reason why an old test needs to be changed like this?

Naros:
Superseded by #7180.

PR #6894, opened 2025-11-16
DBZ-9682 Added a nullity check when Subtracting UUIDSets

hagar-yasser1:
When subtracting null from a UUIDSet, the UUIDSet should be returned and a NullPointerException shouldn't be thrown. This takes place when subtracting 2 GTIDSets and a new UUIDSet is added to the left Operand due to restart or maintanence. 
Link to the Issue on JIRA: https://issues.redhat.com/browse/DBZ-9682


jpechane:
@hagar-yasser1 Applied, thanks

PR #6896, opened 2025-11-17
DBZ-9681-Debezium Oracle connector fails to parse Oracle database version

poonam-meghnani:
## External Links
[DBZ-9681](https://issues.redhat.com/browse/DBZ-9681)

## Description
The change fixes the bug in validating Oracle database version in Oracle Connector. The existing query for validating Oracle connection used V$VERSION view that had version in full banner string and each version has a different format for this banner string, for which there are regex patterns to match and extract each segment of database version.

However, the newer versions have below banner that does not match any of the existing regex:
```
SQL> select banner_full from V$VERSION;

BANNER_FULL
--------------------------------------------------------------------------------
Oracle AI Database 26ai Free Release 23.26.0.0.0 - Develop, Learn, and Run for F
ree
Version 23.26.0.0.0

```
The change queries from V$INSTANCE instead that does not need to parse the banner like above but VERSION string which adheres by the 5 segment version format -
example - 23.26.0.0.0

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
@@ -159,7 +159,7 @@ public class OracleConnection extends JdbcConnection {
             try {
                 // Oracle 18.1 introduced BANNER_FULL as the new column rather than BANNER
                 // This column uses a different format than the legacy BANNER.
-                versionStr = queryAndMap("SELECT BANNER_FULL FROM V$VERSION WHERE BANNER_FULL LIKE 'Oracle Database%'", (rs) -> {
+                versionStr = queryAndMap("SELECT VERSION_FULL FROM V$INSTANCE", (rs) -> {
                     if (rs.next()) {
                         return rs.getString(1);
                     }
@@ -168,8 +168,8 @@ public class OracleConnection extends JdbcConnection {
             }
             catch (SQLException e) {
                 // exception ignored
-                if (e.getMessage().contains("ORA-00904: \"BANNER_FULL\"")) {
-                    LOGGER.debug("BANNER_FULL column not in V$VERSION, using BANNER column as fallback");
+                if (e.getMessage().contains("ORA-00904: \"VERSION_FULL\"")) {
+                    LOGGER.debug("VERSION_FULL column not in V$INSTANCE, using VERSION column as fallback");
                     versionStr = null;
                 }
                 else {
@@ -180,7 +180,7 @@ public class OracleConnection extends JdbcConnection {
             // For databases prior to 18.1, a SQLException will be thrown due to BANNER_FULL not being a column and
             // this will cause versionStr to remain null, use fallback column BANNER for versions prior to 18.1.
             if (versionStr == null) {
-                versionStr = queryAndMap("SELECT BANNER FROM V$VERSION WHERE BANNER LIKE 'Oracle Database%'", (rs) -> {
+                versionStr = queryAndMap("SELECT VERSION FROM V$INSTANCE", (rs) -> {
                     if (rs.next()) {
                         return rs.getString(1);
                     }
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDatabaseVersion.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDatabaseVersion.java
@@ -8,8 +8,6 @@ package io.debezium.connector.oracle;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import io.debezium.annotation.VisibleForTesting;
-
 /**
  * Represents the Oracle database version.
  *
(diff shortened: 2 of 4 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
```diff
@@ -159,7 +159,7 @@ private OracleDatabaseVersion resolveOracleDatabaseVersion() {
             try {
                 // Oracle 18.1 introduced BANNER_FULL as the new column rather than BANNER
                 // This column uses a different format than the legacy BANNER.
-                versionStr = queryAndMap("SELECT BANNER_FULL FROM V$VERSION WHERE BANNER_FULL LIKE 'Oracle Database%'", (rs) -> {
+                versionStr = queryAndMap("SELECT VERSION_FULL FROM V$INSTANCE", (rs) -> {
```
poonam-meghnani, line 162: Pretty sure, there is only 1 row in V$INSTANCE. However, can add some logic to pick the first if needed at all.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDatabaseVersion.java
```diff
@@ -8,36 +8,29 @@
...
  */
 public class OracleDatabaseVersion {
     private final static Pattern VERSION_PATTERN = Pattern
-            .compile("(?:.*)(?:Release )([0-9]+)\\.([0-9]+)\\.([0-9]+)\\.([0-9]+)\\.([0-9]+)(?:.*)");
-    private final static Pattern VERSION_18_1_PATTERN = Pattern
-            .compile("^Oracle Database.*(?:\\r\\n|\\r|\\n)^(?:Version )([0-9]+)\\.([0-9]+)\\.([0-9]+)\\.([0-9]+)\\.([0-9]+)", Pattern.MULTILINE);
-
-    @VisibleForTesting
-    public static final OracleDatabaseVersion VERSION_19_3 = new OracleDatabaseVersion(19, 3, 0, 0, 0, "");
+            .compile("^([0-9]+)\\.([0-9]+)\\.([0-9]+)\\.([0-9]+)\\.([0-9]+)$");
```
poonam-meghnani, line 18: The version from V$INSTANCE does not have any text but 5-segment version format. and hence removed other regex's

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleDatabaseVersionTest.java
```diff
@@ -19,63 +17,14 @@
...
-        assertOracleVersion(version, 21, 3, 0, 0, 0, banner);
+    public void shouldParseOracleVersion() {
+        String version = "23.4.0.23.10";
+        OracleDatabaseVersion oracleDatabaseVersion = OracleDatabaseVersion.parse(version);
+        assertThat(oracleDatabaseVersion.getMajor()).isEqualTo(23);
+        assertThat(oracleDatabaseVersion.getMaintenance()).isEqualTo(4);
+        assertThat(oracleDatabaseVersion.getAppServer()).isEqualTo(0);
+        assertThat(oracleDatabaseVersion.getComponent()).isEqualTo(23);
+        assertThat(oracleDatabaseVersion.getPlatform()).isEqualTo(10);
+        assertThat(oracleDatabaseVersion.getVersion()).isEqualTo(version);
```
poonam-meghnani, line 28: This test covers pretty much all version cases so replaced the others. Please raise a concern if any of the older ones might cause issues.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnection.java
Naros: You're correct, there is only ever one row.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleDatabaseVersion.java
Naros: Thanks, this should make reading the version metadata much easier.

Naros:
@poonam-meghnani I just force pushed to fix the commit message, just FYI.

poonam-meghnani:
> /packit test --labels oracle

@Naros, 
Most of the tests fail on starting test container -
```
9:09:57,892 [ERROR] io.debezium.testing.testcontainers.ApicurioRegistryTestIT -- Time elapsed: 1.335 s <<< ERROR!
java.lang.RuntimeException: Failed to start Kafka containers
	at io.strimzi.test.container.StrimziKafkaCluster.start(StrimziKafkaCluster.java:299)
```
These do not seem to be related, do you think the change could have caused it or needs rebase?

poonam-meghnani:
> > /packit test --labels oracle
> 
> @Naros, Most of the tests fail on starting test container -
> 
> ```
> 9:09:57,892 [ERROR] io.debezium.testing.testcontainers.ApicurioRegistryTestIT -- Time elapsed: 1.335 s <<< ERROR!
> java.lang.RuntimeException: Failed to start Kafka containers
> 	at io.strimzi.test.container.StrimziKafkaCluster.start(StrimziKafkaCluster.java:299)
> ```
> 
> These do not seem to be related, do you think the change could have caused it or needs rebase?

@Naros 
My bad, I didnt see this error in the failure logs - 
`	UnixSocketClientProviderStrategy: failed with exception BadRequestException (Status 400: {"message":"client version 1.32 is too old. Minimum supported API version is 1.44, please upgrade your client to a newer version"}`

Looks like the test-containers version and docker client version are incompatible? 
I am looking forward to merging this change asap as its blocks us on using newer Oracle versions, but I am happy to wait if it is expected to be resolved before next release?
Thanks for the support!

jpechane:
@Naros Do you think we need to upgrade TF image?

Naros:
Hi @jpechane, it's actually an issue with testcontainers. I noticed it today after updating my local system to Docker 29; the classes failed to communicate appropriately with the Docker subsystem. Once I updated Testcontainers to 2.0.2, the issue was fixed. We were on a pretty old version of Testcontainers :cry: .

Naros (approved):
This LGTM. We just need to merge the TC update so that addresses the issue you've seen with regard to the Docker compatibility issue in the TF and we're good.

Naros:
Actually there is one small issue @poonam-meghnani. The call to `#getOracleVersion` needs to be predicated to run only when the connection is in the container root because `V$INSTANCE` is not accessible from within the context of a PDB. That's a unique difference from `V$VERSION` which was accessible from both contexts. I am working on a small commit to address that.

If  you're curious what I mean, you can check out `OracleConnectorIT#shouldGetOracleDatabaseVersion` as an example use case.

Naros:
For now @poonam-meghnani I've gone ahead with a slightly different implementation in https://github.com/debezium/debezium/pull/6911, which I added you as the author so you get credit. I'd rather have some time to think about the `V$INSTANCE` accessibility within the CDB, and adjusting the regex to be less restrictive should satisfy the immediate need without jeopardizing backward compatibility.

Can you check if that works for you?

Naros (changes requested):
As indicated earlier, this needs some more thought about the fact that we should ideally consider auto-switching between PDB/CDB for version details when using the `V$INSTANCE` performance view; however, due to how some users may configure the connector, there isn't a good way to resolve this when users set `database.pdbname` and `database.pdb.name` to the PDB.

Naros:

Hi @poonam-meghnani, please sign off all commits with:

```
git commit -s
```

Naros:
Superseded by #7190

PR #6976, opened 2025-12-16
debezium/dbz#1417 Add support for quantiles

vjuranek:
Add support for metrics statistics like averages and quantiles.

debezium/dbz#1417

### DIFF
--- a/debezium-bom/pom.xml
+++ b/debezium-bom/pom.xml
@@ -600,6 +600,13 @@
                 <version>${version.docling}</version>
             </dependency>
 
+            <!-- Statistics dependencies -->
+            <dependency>
+                <groupId>com.datadoghq</groupId>
+                <artifactId>sketches-java</artifactId>
+                <version>${version.datadog.sketches}</version>
+            </dependency>
+
             <!-- Testing utilities -->
             <dependency>
                 <groupId>org.testcontainers</groupId>
--- a/debezium-connector-common/src/main/java/io/debezium/metrics/stats/SingleValueMeasurement.java
+++ b/debezium-connector-common/src/main/java/io/debezium/metrics/stats/SingleValueMeasurement.java
@@ -17,11 +17,11 @@ public interface SingleValueMeasurement<T extends SingleValueEvent<V>, V> extend
 
     void reset();
 
-    Long getLastValue();
+    V getLastValue();
 
-    Long getMinValue();
+    V getMinValue();
 
-    Long getMaxValue();
+    V getMaxValue();
 
-    Long getAverageValue();
+    V getAverageValue();
 }
--- a/debezium-connector-common/src/main/java/io/debezium/metrics/stats/SingleValueMeasurement.java
+++ b/debezium-connector-common/src/main/java/io/debezium/metrics/stats/SingleValueMeasurement.java
@@ -1,0 +1,26 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.metrics.stats;
+
+import io.debezium.metrics.event.SingleValueEvent;
+
+/**
+ * // TODO: Document this
+ * @author vjuranek
+ * @since 4.0
+ */
+public interface SingleValueMeasurement<T extends SingleValueEvent<V>, V> extends Measurement<T> {
+
+    void reset();
+
+    Long getLastValue();
+
+    Long getMinValue();
+
+    Long getMaxValue();
+
+    Long getAverageValue();
+}
--- a/debezium-connector-common/src/main/java/io/debezium/metrics/stats/SingleValueMeasurement.java
+++ b/debezium-connector-common/src/main/java/io/debezium/metrics/stats/SingleValueMeasurement.java
@@ -8,9 +8,10 @@ package io.debezium.metrics.stats;
 import io.debezium.metrics.event.SingleValueEvent;
 
 /**
- * // TODO: Document this
+ * A measurement which measures only a single variable.
+ * Supports basic statistics over some time period, like minimal, maximal and average values.
+ *
  * @author vjuranek
- * @since 4.0
  */
 public interface SingleValueMeasurement<T extends SingleValueEvent<V>, V> extends Measurement<T> {
 
--- a/debezium-connector-common/src/main/java/io/debezium/metrics/stats/SingleValueMeasurement.java
+++ b/debezium-connector-common/src/main/java/io/debezium/metrics/stats/SingleValueMeasurement.java
@@ -23,5 +23,5 @@ public interface SingleValueMeasurement<T extends SingleValueEvent<V>, V> extend
 
     V getMaxValue();
 
-    V getAverageValue();
+    Double getAverageValue();
 }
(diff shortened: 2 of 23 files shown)


Naros:

Hi @vjuranek, please sign off all commits with:

```
git commit -s
```

debezium-connector-common/src/main/java/io/debezium/metrics/stats/SingleValueMeasurement.java
```diff
@@ -0,0 +1,27 @@
...
+
+import io.debezium.metrics.event.SingleValueEvent;
+
+/**
+ * A measurement which measures only a single variable.
+ * Supports basic statistics over some time period, like minimal, maximal and average values.
+ *
+ * @author vjuranek
+ */
+public interface SingleValueMeasurement<T extends SingleValueEvent<V>, V> extends Measurement<T> {
```
Naros, line 16: In the Oracle connector, we have some metrics that not only measure the last, min, and max values but also track the total value. Do we think those are meaningful to include here?

mfvitale:
@vjuranek just a question. How this work fits with the introduction of `micrometer` that we have planned for Q2?

vjuranek:
> @vjuranek just a question. How this work fits with the introduction of `micrometer` that we have planned for Q2?

@mfvitale well, TBH I don't know much about Micrometer and what are the exact plan how to use it in Debezium. This work just extends Debezium metrics. My understanding (based on the Debezium roadmap `Micrometer monitoring using OpenTelemetry`) is that Mictometer will replace OTEL tracing, providing general observability capabilities and OTEL will be only one of the possible implementations. So if this understanding is correct, there is no overlap between these two.

debezium-embedded/src/test/java/io/debezium/pipeline/AbstractMetricsTest.java
```diff
@@ -275,6 +311,56 @@ protected void assertStreamingMetrics(boolean checkAdvancedMetrics, long expecte
...
+    protected void assertStreamingStatistics(long expectedEvents) throws Exception {
+        final MBeanServer mBeanServer = ManagementFactory.getPlatformMBeanServer();
+
+        waitForStreamingRunning(connector(), server(), getStreamingNamespace(), task());
+
+        // Insert new records and wait for them to become available
+        executeInsertStatements();
+
+        consumeRecordsByTopic((int) expectedEvents);
+        Thread.sleep(Duration.ofSeconds(2).toMillis());
```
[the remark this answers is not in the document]
vjuranek: This seems to be fine -  we just need to wait little bit to be sure that metrics is update. Moreover, similar approach is used across the whole test.

mfvitale:
> > @vjuranek just a question. How this work fits with the introduction of `micrometer` that we have planned for Q2?
> 
> @mfvitale well, TBH I don't know much about Micrometer and what are the exact plan how to use it in Debezium. This work just extends Debezium metrics. My understanding (based on the Debezium roadmap `Micrometer monitoring using OpenTelemetry`) is that Mictometer will replace OTEL tracing, providing general observability capabilities and OTEL will be only one of the possible implementations. So if this understanding is correct, there is no overlap between these two.

I'm not a super expert of Micrometer but I'm wondering if we could also use it for the metrics too. Micrometer should offers already the concept of quantiles.
So in that case we can avoid to use another dependency to do the calculations. WDYT?

jpechane:
@rk3rn3r Could you please take a look at this as IIRC you'll be involved in Micrometer in future?

debezium-connector-common/src/main/java/io/debezium/metrics/stats/SingleValueMeasurement.java
vjuranek: Sure, added.

vjuranek:
> I'm not a super expert of Micrometer but I'm wondering if we could also use it for the metrics too. Micrometer should offers already the concept of quantiles.
> So in that case we can avoid to use another dependency to do the calculations. WDYT?

Yes, I was looking on Mircometer yesterday little bit (and will continue today) and it indeed provides support for histograms and quantiles. However, my understanding is that it's just a facade, so you won't avoid additional dependency on e.g. Datadog lib anyway, but it gives you the freedom to switch the underlying data processing library.

vjuranek:
I did some investigation over the weekend and also consult with the Claude. It really depends what out intentions with Micrometer are. If it's limited only to tracing, there is of course no overlap. If we are going to replace all current metrics (JMX) with Micrometer, it's questionable if this PR should be merged. Claude concluded (and I more or less agree with it), that if switch to Micrometer will be done any time soon (in Q2 as planned), this PR shouldn't be merged. If it takes loner time (as it will be bigger effort replace all metrics with Micrometer), it makes sense to merge this PR, eventually add into the PR another abstraction layer, which would make the switch to Micrometer straightforward.

jpechane:
@rk3rn3r Could you please comment above on the timing?
@vjuranek I don't think we want to remove JMX metrics altogether, as it is universal tool in Java world, so maybe it should still stay?

mfvitale:
> I don't think we want to remove JMX metrics altogether, as it is universal tool in Java world, so maybe it should still stay?

@jpechane, @vjuranek Well I think that we should use `Micrometer` as primary instrumentation library and then export the metrics via micrometer `registry` (look to [Composite Registries](https://docs.micrometer.io/micrometer/reference/concepts/registry.html#_composite_registries)) to the different technology (jmx, prometheus, otel)

Think to it as `SLF4J` and its implementations (Logback/Log4j2). 

In that way we use a more powerful instrumentation library that has concepts like `Gauges`, `Counters`, `Timers` and so on and the let the registry adapt them to the destination.

vjuranek:
```quote
Well I think that we should use Micrometer as primary instrumentation library and then export the metrics via micrometer registry
```
@mfvitale,  in general I agree and if I was starting a new project, I'd use Micrometer as the primary monitoring/tracing library in such case. On the other hand, if you have quite large project with lots of existing JMX metrics (which doesn't require any dependency)  and switching to Micrometer would require significant effort, I'd be quite cautious. AFAICT Micrometer is not under any foundation, so if they decide to change the license and forbid it for commercial use or do any other disrupting change, we would have to invest once again big effort to switch to another solution or move JMX metics back.

mfvitale:
> ```
> Well I think that we should use Micrometer as primary instrumentation library and then export the metrics via micrometer registry
> ```
> 
>  AFAICT Micrometer is not under any foundation, so if they decide to change the license and forbid it for commercial use or do any other disrupting change, we would have to invest once again big effort to switch to another solution or move JMX metics back.

Well this is a good point.

rk3rn3r:
Sorry, I talked to Jiri about it last week and wrote my reply and just realized now, that I didn't submit. 

I agreed with Jiri. Ours and my personal take was we should merge the PR. It's possible to start from here and it will move things forward.
If we replace / move to metrics with Micrometer eventually we are still able to decide in H2-26 when we get a feeling for Micrometer.
My personal feeling is that it's easier to decide when you started using it.

debezium-bom/pom.xml
```diff
@@ -605,6 +605,13 @@
                 <version>${version.docling}</version>
             </dependency>
 
+            <!-- Statistics dependencies -->
+            <dependency>
```
jpechane, line 604: @vjuranek Are there any transitive dependencies pulled with it?

jpechane:
@vjuranek Can you please add dew screenshots from a JMX console?

debezium-bom/pom.xml
vjuranek: No transitive dependency (optional dependency on protobuf, which we already include)

vjuranek:
@jpechane sorry for the delay, here are the screenshots:

Screenshot_20260504_135403
Screenshot_20260504_135416
Screenshot_20260504_135427
Screenshot_20260504_135439
Screenshot_20260504_135448
Screenshot_20260504_135456

jpechane (approved):
@vjuranek LGTM, thanks! Feel free to merge and continue with the changes we discussed during yesterday call.

PR #6990, opened 2026-01-02
debezium/dbz#1490 Fix JDBC Sink Connector to cast string values to PostgreSQL enum types

### LINKED ISSUE #1490: JDBC Sink Connector fails to insert string values into PostgreSQL enum columns
## Bug report

 When using the JDBC Sink Connector to write records to a PostgreSQL table containing an enum column from string kafka connect field, the connector fails with:

```ERROR: column "source" is of type my_enum_type but expression is of type character varying
Hint: You will need to rewrite or cast the expression.```

---

No specific configuation

---

PG 15

---

Casting the string to the enum type.

---

**What behavior do you see?**

SQL error :)

---

Tested on 3.2.2.Final and on main.

---

N/A

---

See attached PR.

### PR DESCRIPTION
**Pull request by bpaquet**
Fixes https://github.com/debezium/dbz/issues/1490

When inserting string values into PostgreSQL enum columns, the connector now detects user-defined enum types and generates appropriate cast expressions.

PostgreSQL enums are reported by JDBC as VARCHAR with a custom type name. The fix detects these by checking if the type name is not a standard PostgreSQL string type (varchar, text, char, etc.) and applies: cast(? as "enum_type_name")

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresDatabaseDialect.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresDatabaseDialect.java
@@ -9,10 +9,15 @@ import static io.debezium.connector.jdbc.type.debezium.DebeziumZonedTimestampTyp
 import static io.debezium.connector.jdbc.type.debezium.DebeziumZonedTimestampType.POSITIVE_INFINITY;
 
 import java.sql.Connection;
+import java.sql.ResultSet;
 import java.sql.SQLException;
+import java.sql.Statement;
+import java.sql.Types;
 import java.time.format.DateTimeFormatter;
 import java.time.temporal.TemporalAccessor;
+import java.util.HashSet;
 import java.util.Optional;
+import java.util.Set;
 
 import org.apache.kafka.connect.data.Schema;
 import org.hibernate.SessionFactory;
@@ -54,6 +59,8 @@ public class PostgresDatabaseDialect extends GeneralDatabaseDialect {
         }
     }
 
+    private final Set<String> enumTypeNames = new HashSet<>();
+
     private PostgresDatabaseDialect(JdbcSinkConnectorConfig config, SessionFactory sessionFactory) {
         super(config, sessionFactory);
     }
@@ -78,7 +85,20 @@ public class PostgresDatabaseDialect extends GeneralDatabaseDialect {
             // This means that the table will be stored as lower-case
             collectionId = collectionId.toLowerCase();
         }
+        cacheEnumTypes(connection);
         return super.readTable(connection, collectionId);
+    }
+
+    private void cacheEnumTypes(Connection connection) throws SQLException {
+        if (!enumTypeNames.isEmpty()) {
+            return;
+        }
+        try (Statement stmt = connection.createStatement();
+                ResultSet rs = stmt.executeQuery("SELECT typname FROM pg_type WHERE typtype = 'e'")) {
+            while (rs.next()) {
+                enumTypeNames.add(rs.getString(1).toLowerCase());
+            }
+        }
     }
 
     @Override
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresDatabaseDialect.java [hunk 1]
```diff
@@ -54,6 +59,8 @@ public DatabaseDialect instantiate(JdbcSinkConnectorConfig config, SessionFactor
         }
     }
 
+    private final Set<String> enumTypeNames = new HashSet<>();
```
bpaquet, line 62: This cache here is not really cool. I can consider moving it to ColumnDescriptor to add a bool isEnum, but I have to create an extension point somewhere to activate it and fill it only for PG, and/or break the immutability of ColumnDescriptor to do a backfill after table reading. We can also consider a postprocessing after the `readTables` call in the same class. These solutions seems too heavy to me compared to this cache.

What do you think?

GeckoSplinter (commented):
I have read quickly but the solution with the cache tradeoff seems OK to me

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresDatabaseDialect.java
bpaquet: Another option is to implement a known type list. If the `typeName  == "varchar"`, we know this is not an enum ...
I still prefer the cache solution above

Naros: Hi @bpaquet I don't think placing this logic in the Dialect is the correct location for this. I realize there are some things already there, but in general, any casting should be handled by the type subsystem in `io.debezium.connector.jdbc.type`.

As a simple pass, could this be solved by using a transform to make sure the field is encoded as a specia logical type, i.e. `io.debezium.data.Enum` and then letting the `EnumType` handle the casting? And is the enum cache even necessary in this case?

But looking at this from a long-term view, I am wondering if it makes sense to consolidate ColumnDescriptor and FieldDescriptor contracts here, or at the very least move the JdbcType from FieldDescriptor to ColumnDescriptor. The idea I have here is that doing this when we fetch the JDBC metadata to create the TableDescriptor provides several key benefits:

1. We can deduce complex type mappings using the field schema and the column JDBC metadata.
2. We could introduce composite type mappings quite easily, data cast to A then cast to B.
3. Type resolution ends up being cached along side TableDescriptor during batches for performance.

What are your thoughts?

Naros:

Hi @bpaquet, please sign off all commits with:

```
git commit -s
```

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresDatabaseDialect.java
```diff
@@ -78,9 +85,22 @@ public TableDescriptor readTable(Connection connection, CollectionId collectionI
...
     }
 
+    private void cacheEnumTypes(Connection connection) throws SQLException {
+        if (!enumTypeNames.isEmpty()) {
+            return;
+        }
+        try (Statement stmt = connection.createStatement();
+                ResultSet rs = stmt.executeQuery("SELECT typname FROM pg_type WHERE typtype = 'e'")) {
+            while (rs.next()) {
+                enumTypeNames.add(rs.getString(1).toLowerCase());
```
Ritabrata1080, line 99: Check if resultset object can be closed to avoid any memory leaks.

Naros: The `ResultSet` is part of a `try-with-resources` block, it will be closed automatically when the scope exits.

Naros:

PR #7023, opened 2026-01-19
[WIP] [JUST FOR DISCUSSION] DBZ-7996 pt2

rk3rn3r:


debezium-sink/src/main/java/io/debezium/sink/AbstractChangeEventSink.java
```diff
@@ -0,0 +1,108 @@
...
+
+    protected AbstractChangeEventSink(SinkConnectorConfig config) {
+        this.config = config;
+        if (SinkConnectorConfig.PrimaryKeyMode.NONE.equals(config.getPrimaryKeyMode())) {
+            buffer = new KeylessBuffer(config);
+        }
+        else {
+            // might also require: !record.keyFieldNames().isEmpty()
+            buffer = new DeduplicatedBuffer(config);
+        }
```
Naros, line 47: There should be a third option here to write the changes from the topic as they exist. From an audit point of view, there may be a need for 1:1 mapping from source to target systems, and forcing the deduplication buffer eliminates that option for users. The dedup buffer should be optional, or at least togglable for users who need the 1:1.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
```diff
@@ -81,6 +74,14 @@ public JdbcChangeEventSink(JdbcSinkConnectorConfig config, StatelessSession sess
     }
 
     public void execute(Collection<SinkRecord> records) {
+        if (config.isSharedChangeEventSinkEnabled()) {
+            var batch = put(records);
+            if (batch != null && !batch.isEmpty()) {
+                recordWriter.write(batch);
+            }
+            return;
+        }
```
Naros, line 83: If this is optional, would it be better to have two ChangeEventSink implementations: one that retains the old behavior and one that uses the sharded change event sink feature?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/RecordWriter.java
```diff
@@ -43,13 +59,159 @@ public RecordWriter(SharedSessionContract session, QueryBinderResolver queryBind
         this.dialect = dialect;
     }
 
-    public void write(List<JdbcSinkRecord> records, String sqlStatement) {
+    private String getSqlStatement(TableDescriptor table, JdbcSinkRecord record) {
```
Naros, line 62: The recent PostgreSQL unnest PR may be of interest here @rk3rn3r. It will likely be merged early next week, and I believe that shifts this behavior here slightly.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
rk3rn3r: It's only there for a migration phase and for benchmarking and testing purposes. Do you think it still makes sense to split implementations?

debezium-sink/src/main/java/io/debezium/sink/AbstractChangeEventSink.java
rk3rn3r: So you mean there should be a way to set PrimaryKeyMode another value than NONE but still using KeylessBuffer? That would make every message processed but the same result in the downstream system (because the datastore is keyed). Would that make sense?

Naros: Yes, although in that context I'd argue that `KeylessBuffer` is a poor name, and something like `PassthroughBuffer`, `EventBuffer`, or simply `Buffer` makes far more sense. But I digress.

At a past employer, we had two ERP systems connected via a data pipeline. For every operation in Oracle ERP, there had to be a corresponding audit trail in SAP. We were not allowed to take N changes on the Oracle side and consolidate them into a single SAP change, even though they were for the same PK. There was some leniency for corner cases where we had to resend the data, and some audit entries in SAP were duplicated. However, not having at least 1 audit entry in SAP for the same audit entry in Oracle was not allowed.

So, in these highly regulated environments where strict auditing is mandatory, we should ensure we can accommodate it. Deduplication could be the default, but falling back to a 1:1 batch approach, as we currently do, is important.

One final note: I did not notice it previously, but it is inside `AbstractBuffer#truncate`. We are making an assumption that it's fine to discard events without writing them prior to the truncate. We should rethink that, again for the reason I outlined above.

rk3rn3r: @Naros Because I am about to finish the work I am coming back to this comment. I see the requirement, but we might be able to agree it is not smart (omitting a better fitting word ;) ). Especially considering that modern JDBC drivers tend to eliminate duplicate physical writes to the same key, right? So I would argue in 2026 this might not be easy to implement and that yields the question if such a specific use-case should be covered by Debezium?

Arguments against it is that this would break efficiency a lot. For a very rare use-case we would introduce much higher code / flow complexity.

It is not simply about using the KeylessBuffer, but a different DuplicatesBuffer with a different behavior:
- always need to lookup if an added record's key is already in the buffer, if so return the buffer (before batch size is reached) for flushing - breaking the benefits of strict batch size and commit intervals
- same applies to delete and truncate events (as you pointed out)
- modern JDBC drivers de-duplicate writes to the same physical row - duplicate writes might not hit the datastore at all
- probably more, but I am not aware of rn

Are you okay with I am only logging an issue for that and do it later after a more detailed discussion?

Naros: > Especially considering that modern JDBC drivers tend to eliminate duplicate physical writes to the same key, right? 

No, they don't. At the JDBC layer, the main optimization is JDBC batching, which minimizes the database network round-trip by sending SQL statements and data in a single network operation. Behavior such as statement reordering, eliminating dual writes to the same keys, etc., is rarely a driver concern and, more often than not, a database engine concern. 

This means to have the same expected behavior, you either need to accept that, depending on the variant of a given database, these advanced features may not be available, or you design code at a higher level to account for this at the application level.

But as we have seen with `upsert` across multiple vendors, while we send 1 batch to the engine, it applies those to the database in a multi-pass fashion by 

1. Evaluate all the keys, setting whether the row exists or does not 
2. Iterate all data + the computed data from (1) and perform the insert or update

This is why, when you do an `upsert` with a batch that has 2 events for the same key and the row does not yet exist, the database engine attempts 2 inserts, causing a fault on the second due to a constraint violation.

This is precisely why the reduction buffer is super important. But it's a decision point that users have to account for in their setup. Do they allow writes to be dropped for efficiency, or does their environment mandate 1:1 writes across systems? If they require 1:1, either `batch.size=1` or a JDBC sink flush after each row toggle becomes critical for proper operation.

rk3rn3r: I can't find it anymore, but I think I saw it last year. But I agree it could've been only one driver and maybe specific for the underlying datastore. 👍 

You are more or less saying what I said before. It's possible, but not nice.

I might give it a chance to do it now, but if it would delay the PR even more because it needs tests etc, would you be okay if I'll file this extra buffer and primary key mode as a new issue? @Naros

Naros: I think we can do it incrementally, but we need to be strategic about it. 

My suggestion would be that we either introduce the loss in Alpha1, and we have it fixed by BetaN or CRn in the same cycle, especially if the release is slated for downstream GA. It's not good for customers to be in a position where we want them to upgrade but can't because of an intentional decision we made that has no real customer-facing justification for the loss in the first place.

@jpechane any thoughts?

jpechane: Yes, I agree. The feature can go out in Alpha butwe must be absolutely committed to bring it back no later than in CR. If we are not 100 % we can achieve this thn incremental approach can't be used.

PR #7098, opened 2026-02-25
DBZ-9816 Update strimzi-test-container

see-quick:
Hi folks... I have just checked that you still using 0.112.0 so I bumped to recently released version (which should also have 4.2.0).

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -104,8 +104,8 @@
         <version.build-helper.plugin>3.6.1</version.build-helper.plugin>
 
         <!-- Kafka and it's dependencies MUST reflect what the Kafka version uses -->
-        <version.kafka>4.1.1</version.kafka>
-        <version.strimzi-test-container>0.112.0</version.strimzi-test-container>
+        <version.kafka>4.2.0</version.kafka>
+        <version.strimzi-test-container>0.115.0</version.strimzi-test-container>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -104,8 +104,8 @@
         <version.build-helper.plugin>3.6.1</version.build-helper.plugin>
 
         <!-- Kafka and it's dependencies MUST reflect what the Kafka version uses -->
-        <version.kafka>4.1.1</version.kafka>
-        <version.strimzi-test-container>0.112.0</version.strimzi-test-container>
+        <version.kafka>4.2.0</version.kafka>
```
Naros: Does this update require the use of Kafka 4.2?

see-quick: It is not required, but it may be better to align both versions?

Naros: In that case, it may be best to align with #7105, cc @jpechane

jpechane: @Naros Thanks for pointing it out.
@see-quick Yes, please limit this PR to update of the Strimzi TC only. Upgrade to Kafka 4.2 is handled separately as it will require more changes.

see-quick: Sure, I have just updates the version of stc.

jpechane:
Applied via https://github.com/debezium/debezium/pull/7120
@see-quick Closing, thanks!

PR #7105, opened 2026-02-26
debezium/dbz#1651 Upgrade Kafka to 4.2.0

### LINKED ISSUE #1651: Upgrade to Kafka 4.2.0

jpechane:
Fixes https://github.com/debezium/dbz/issues/1651

Requires https://github.com/debezium/debezium-platform/pull/369

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEmbeddedEngine.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEmbeddedEngine.java
@@ -41,6 +41,7 @@ import org.apache.kafka.connect.json.JsonConverterConfig;
 import org.apache.kafka.connect.runtime.AbstractHerder;
 import org.apache.kafka.connect.runtime.ConnectorConfig;
 import org.apache.kafka.connect.runtime.WorkerConfig;
+import org.apache.kafka.connect.runtime.rest.entities.ConfigInfo;
 import org.apache.kafka.connect.runtime.rest.entities.ConfigInfos;
 import org.apache.kafka.connect.source.SourceConnector;
 import org.apache.kafka.connect.source.SourceRecord;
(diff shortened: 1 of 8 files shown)


jamesnetherton:
Any idea when we'll see this upgrade in a Debezium release?

@Naros FYI In Quarkus 3.35.0, kafka-clients will be upgraded to 4.2.0. So I assume it'll cause issues for the Quarkus Debezium bits? The tests seem to be disabled in the Quarkus Platform, so I guess any problems wont be caught.

Naros:
Hi @jamesnetherton, I would wager this quarter's Debezium 3.6 release.

jpechane:
@vjuranek Could you please take a look? `AsyncEmbeddedEngineTest` and otheres required adding this property. I want to double check if the issue is the property missing or the test can/should be updated and the freshly added property can be removed.

vjuranek:
> I want to double check if the issue is the property missing or the test can/should be updated and the freshly added property can be removed.

@jpechane unfortunately it has to stay (exactly as you added it), as prior Kafka 4.2.0 it had the default value, but since 4.2.0 the default was removed and user is now required to provide some value in the configuration. See [KIP-1161](https://cwiki.apache.org/confluence/display/KAFKA/KIP-1161%3A+Unifying+LIST-Type+Configuration+Validation+and+Default+Values) for more details.

debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEmbeddedEngine.java
```diff
@@ -797,7 +799,8 @@ private Map<String, String> validateAndGetConnectorConfig(final SourceConnector
         final ConfigInfos configInfos = AbstractHerder.generateResult(connectorClassName, Collections.emptyMap(), validatedConnectorConfig.configValues(),
                 connector.config().groups());
         if (configInfos.errorCount() > 0) {
-            final String errors = configInfos.values().stream()
+            @SuppressWarnings("unchecked")
+            final String errors = ((List<ConfigInfo>) Reflections.invokeMethodWithFallbackName(configInfos, "configs", "values", List.class)).stream()
```
vjuranek, line 805: I'd add TODO note to clean this up once we require Kafka 4.2 or higher. Or a comment that this is needed to handle Kafka < 4.2.0 and Kafka >= 4.2.0 at the same time, so that one immediately know why the reflection is needed here without digging into the git history.

vjuranek (approved):
Small note about commenting unusual access pattern, otherwise LGTM.

jpechane:
@vjuranek Comment added and merged

PR #7106, opened 2026-02-26
debezium/dbz#1557 Keep OTel span scope active during format conversion

### LINKED ISSUE #1557: OpenTelemetry span context not active during format conversion (schema registry HTTP calls not traced)
## Summary

I've been playing around with the entire setup since we fairly recently implemented Otel support for Apicurio Registry. When using Debezium Server with OpenTelemetry tracing and a schema registry (e.g., Apicurio Registry), HTTP calls made by format converters to register/fetch schemas are **not included in the distributed trace**, even though the `ActivateTracingSpan` SMT is properly configured. 

I have it working in an example, but I was wondering if it's worth it to add some code to the Debezium project, or maybe document the setup/workaround.

## Root Cause

The "issue" is in `TracingSpanUtil.traceRecord()` - the span scope is closed before the method returns, which means downstream components (like Avro format converters) don't have an active trace context when they make HTTP calls.

```java
// TracingSpanUtil.java (simplified)
public static <R extends ConnectRecord<R>> R traceRecord(R connectRecord, ...) {
    Span txLogSpan = ...;
    
    try (Scope ignored = txLogSpan.makeCurrent()) {  // ← Scope opens
        // ... span attributes set
        // ... context injected into HEADERS
        
    }  // ← Scope CLOSES HERE
    finally {
        txLogSpan.end();
    }
    
    return connectRecord;  // ← Returns with NO active thread-local span context
}
```

The trace context is propagated via **Kafka headers** (for downstream Kafka consumers), but **not** via thread-local context (for same-process operations like format conversion and schema registry HTTP calls).

## "Expected" vs Actual Behavior

| Component | Expected | Actual |
|-----------|----------|--------|
| Debezium CDC span | ✅ Traced | ✅ Traced |
| Kafka message headers | ✅ Context injected | ✅ Context injected |
| Format converter HTTP calls | ✅ Traced as child spans | ❌ **Separate trace or not traced** |
| Downstream Kafka consumer | ✅ Can extract context from headers | ✅ Works |

## Reproduction Steps

1. Set up Debezium Server with:
   - OpenTelemetry Java Agent
   - Avro format with Apicurio Registry (or Confluent Schema Registry)
   - `ActivateTracingSpan` SMT configured
   - Jaeger or similar trace collector

2. Create a database change that triggers CDC

3. Observe in Jaeger:
   - Debezium spans appear correctly
   - Schema registration POST to registry appears in a **separate trace** (not linked to the CDC trace)

## Environment

- Debezium Server 3.0
- OpenTelemetry Java Agent 2.10.0
- Apicurio Registry (but issue applies to any schema registry)
- PostgreSQL source connector

### Option 1: Keep scope open during format conversion (breaking change)

I'm guessing this is not an option, since this seems to be a design trade-off right now but extending the span scope to cover the entire SMT pipeline including format conversion. This would require architectural changes to how the SMT interacts with converters.

### Option 2: Configuration option for extended scope

Add a configuration property like `tracing.keep.scope.active=true` that keeps the span context active after `traceRecord()` returns.

### Option 3: Document the limitation and workaround

Document that users who need schema registry calls traced should either:
- Use a custom SMT that keeps the scope open
- Configure the OTEL agent to create parent context from headers

## Workaround

I have been playing with implementing a custom SMT that keeps the span scope active:

```java
public class OtelActivateTracingSpan<R extends ConnectRecord<R>> implements Transformation<R> {
    private Span currentSpan;
    
    @Override
    public R apply(R record) {
        // Close previous span
        if (currentSpan != null) {
            currentSpan.end();
        }
        
        SpanContext parentContext = extractTraceContext(record.value());
        
        if (parentContext != null && parentContext.isValid()) {
            Context parentOtelContext = Context.current().with(Span.wrap(parentContext));
            
            currentSpan = tracer.spanBuilder(operationName)
                    .setParent(parentOtelContext)
                    .setSpanKind(SpanKind.CONSUMER)
                    .startSpan();
            
            // Keep scope OPEN - don't use try-with-resources
            currentSpan.makeCurrent();
        }
        
        return record;  // Returns WITH active span context
    }
    
    @Override
    public void close() {
        if (currentSpan != null) {
            currentSpan.end();
        }
    }
}
```

This allows the OTEL Java Agent to automatically instrument HTTP calls made by the format converter as child spans of the CDC trace.

## Related

- Example demonstrating the issue and workaround: https://github.com/Apicurio/apicurio-registry/tree/main/examples/debezium-otel-tracing

carlesarnal:
Closes https://github.com/debezium/dbz/issues/1557

## Description

When using Debezium with OpenTelemetry tracing and a schema registry (Apicurio, Confluent), HTTP calls made by format converters to register/fetch schemas are **not included in the distributed trace**. The root cause is in `TracingSpanUtil.traceRecord()` — the span scope is closed via try-with-resources before the method returns, so downstream components have no active thread-local trace context.

This PR adds an optional `tracing.keep.scope.active` configuration property (default `false`) to the `ActivateTracingSpan` SMT. When enabled, the OTel span scope remains active after `apply()` returns, allowing the OTel Java Agent to automatically instrument downstream HTTP calls (e.g. schema registry requests) as child spans of the CDC trace.

### Changes

- **`TracingSpanUtil`**: Added `TracingResult<R>` inner class and `traceRecordKeepingScope()` method that creates the same tracing structure as `traceRecord()` but does not close the scope or end the span, returning them for the caller to manage
- **`ActivateTracingSpan`**: Added `tracing.keep.scope.active` boolean config field. When enabled, `apply()` delegates to `traceRecordKeepingScope()` and stores the active span/scope. The previous scope is automatically cleaned up on the next `apply()` call or on `close()`
- **`EventRouterConfigDefinition`** / **`MongoEventRouterConfigDefinition`**: Added the new field to the "Tracing" config group
- **`ActivateTracingSpanTest`**: Added two new tests verifying that the span context remains active after `apply()` and that the previous scope is properly cleaned up on subsequent calls

### Default behavior is unchanged

When `tracing.keep.scope.active` is not set (or set to `false`), the existing `traceRecord()` code path is used with no behavioral change.

## Test plan
- [x] Two new unit tests added to `ActivateTracingSpanTest`
- [x] All 5 tracing tests pass (3 existing + 2 new)
- [x] Code formatting verified (`mvn process-sources` — 0 files changed)

### DIFF
--- a/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/outbox/MongoEventRouterConfigDefinition.java
+++ b/debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/outbox/MongoEventRouterConfigDefinition.java
@@ -124,7 +124,8 @@ public class MongoEventRouterConfigDefinition {
         Field.group(
                 config,
                 "Tracing",
-                ActivateTracingSpan.TRACING_SPAN_CONTEXT_FIELD, ActivateTracingSpan.TRACING_OPERATION_NAME, ActivateTracingSpan.TRACING_CONTEXT_FIELD_REQUIRED);
+                ActivateTracingSpan.TRACING_SPAN_CONTEXT_FIELD, ActivateTracingSpan.TRACING_OPERATION_NAME, ActivateTracingSpan.TRACING_CONTEXT_FIELD_REQUIRED,
+                ActivateTracingSpan.TRACING_KEEP_SCOPE_ACTIVE);
         return config;
     }
 }
--- a/debezium-schema-generator/pom.xml
+++ b/debezium-schema-generator/pom.xml
@@ -28,6 +28,10 @@
       <version>1.1.0</version>
     </dependency>
     <dependency>
+      <groupId>io.opentelemetry</groupId>
+      <artifactId>opentelemetry-api</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.apache.maven</groupId>
       <artifactId>maven-plugin-api</artifactId>
       <version>${version.maven}</version>
--- a/debezium-transforms/src/main/java/io/debezium/transforms/outbox/EventRouterConfigDefinition.java
+++ b/debezium-transforms/src/main/java/io/debezium/transforms/outbox/EventRouterConfigDefinition.java
@@ -327,7 +327,8 @@ public class EventRouterConfigDefinition {
         Field.group(
                 config,
                 "Tracing",
-                ActivateTracingSpan.TRACING_SPAN_CONTEXT_FIELD, ActivateTracingSpan.TRACING_OPERATION_NAME, ActivateTracingSpan.TRACING_CONTEXT_FIELD_REQUIRED);
+                ActivateTracingSpan.TRACING_SPAN_CONTEXT_FIELD, ActivateTracingSpan.TRACING_OPERATION_NAME, ActivateTracingSpan.TRACING_CONTEXT_FIELD_REQUIRED,
+                ActivateTracingSpan.TRACING_KEEP_SCOPE_ACTIVE);
         return config;
     }
 
--- a/debezium-transforms/src/main/java/io/debezium/transforms/tracing/ActivateTracingSpan.java
+++ b/debezium-transforms/src/main/java/io/debezium/transforms/tracing/ActivateTracingSpan.java
@@ -24,6 +24,8 @@ import io.debezium.pipeline.EventDispatcher;
 import io.debezium.transforms.Module;
 import io.debezium.transforms.SmtManager;
 import io.opentelemetry.api.GlobalOpenTelemetry;
+import io.opentelemetry.api.trace.Span;
+import io.opentelemetry.context.Scope;
 
 /**
  * This SMT enables integration with a tracing system.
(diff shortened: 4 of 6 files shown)


carlesarnal:
@Naros is there interest in this? If so, I can fix the conflicts and resume, if not, we can close and move on.

Naros:
Hi @carlesarnal, thanks for reaching out.

The only place where I have some doubts is in the Quarkus Extensions, Server, and Embedded Engine, all of which use the same underlying async engine for event processing, as we have several different concurrency modes available for transformations & converter serialization. I don't believe thread-local propagation works in these modes the same way it does within Kafka Connect's single-task = single-thread model.

@vjuranek @jpechane would you concur that could be a limitation?

vjuranek:
> I don't believe thread-local propagation works in these modes the same way it does within Kafka Connect's single-task = single-thread model.
> 
> @vjuranek @jpechane would you concur that could be a limitation?

Yes, this is correct. But if it's well documented, I'd defer to user decision if they want to turn it on or not, so basically I don't have any principal issue with this approach.

If merged, I'd suggest to create an issue for further possible improvements in this regard and assign to me.
I'll be reviewing OTel support in Debezium (https://github.com/debezium/dbz/issues/1907) and as part of it I can investigate if we have better possibilities how to handle that or not (if yes, we can improve, it not, better to have this fix which works in most common case, synchronous processing, than nothing).

PR #7140, opened 2026-03-06
debezium/dbz#197 Add support for MongoDB ByLogicalTableRouter

### LINKED ISSUE #197: ByLogicalTableRouter should support MongoDB [DBZ-1075]
Migrated from [DBZ-1075](https://issues.redhat.com/browse/DBZ-1075)

{{io.debezium.transforms.ByLogicalTableRouter.updateEnvelopeSchema(Schema, String)}} requires {{before}} field to be present in the envelope. This is not the case for the MongoDB connector.

kartikangiras:
### Description: 

Fix NPE for MongoDB envelopes by falling back from before to after when updating or renaming envelope schemas and values along with it added a unit test to cover MongoDB envelope.

Fixes https://github.com/debezium/dbz/issues/197
https://issues.redhat.com/browse/DBZ-1075


debezium-transforms/src/main/java/io/debezium/transforms/ByLogicalTableRouter.java
```diff
@@ -371,16 +371,25 @@ private Schema updateEnvelopeSchema(Schema oldEnvelopeSchema, String newTopicNam
             return newEnvelopeSchema;
         }
 
-        final Schema oldValueSchema = oldEnvelopeSchema.field(Envelope.FieldName.BEFORE).schema();
-        final SchemaBuilder valueBuilder = copySchemaExcludingName(oldValueSchema, SchemaBuilder.struct());
-        valueBuilder.name(schemaNameAdjuster.adjust(newTopicName + ".Value"));
-        final Schema newValueSchema = valueBuilder.build();
+        final org.apache.kafka.connect.data.Field oldValueField = oldEnvelopeSchema.field(Envelope.FieldName.BEFORE) != null
+                ? oldEnvelopeSchema.field(Envelope.FieldName.BEFORE)
+                : oldEnvelopeSchema.field(Envelope.FieldName.AFTER);
```
Naros: This approach results in two `O(1)` calls, rather than a single call as before, when the schema includes the `before` field. In transformations, we should strive to be as efficient as possible because they can often become a bottleneck during bursts of events in the pipeline.

I might suggest a new helper function like:
```java
Field getSchemaFieldWithFallback(Schema schema, String fieldName, String fallbackFieldName) {
  final Field field = schema.field(fieldName);
  if (field != null) {
    return field;
  }
  return schema.field(fallbackFieldName);
}
```

This can be used in the other place where it's now two `O(1)` calls for the `before` field, too.

Naros (changes requested):

debezium-transforms/src/main/java/io/debezium/transforms/ByLogicalTableRouter.java
kartikangiras: I have added the proposed helper function to avoid two 0(1)calls.

Naros (approved):

kartikangiras:
@Naros Are there any required changes for this or is it good to go for the merge. Thanks.

Naros:
Hi @kartikangiras we should hold off merging this until we fix CI.  The CI pipeline did not fire all jobs.

@mfvitale I'm not sure if you have fixed this in another PR, but we should add any changes in the new modules we've recently moved code into as part of the "core" or "common" changes list so that connectors are being built.  It does not appear in this PR that the transform change queued up any connector jobs :sob:

kartikangiras:
@Naros alright that's totally fine.

jpechane:
@mfvitale Have you seen @Naros comment above?

mfvitale:
> @mfvitale Have you seen @Naros comment above?

Yes, I have addressed it in [24d89b77dce37ac66738488ef6c876946eaa997d](https://github.com/debezium/debezium/pull/7084/changes/24d89b77dce37ac66738488ef6c876946eaa997d)

rk3rn3r:
@Naros @mfvitale Should the code maybe moved to a new `ByLogicalCollectionRouter` class and let the old class extend the new one for compatibility and mark the old one as deprecated? We usually renamed Table to Collection when we moved support from relational tables to generic collections. wdyt?

Naros:
With the MongoDB connector, do we actually emit events where changes to the same collection, distributed across different shards, have different topic names? Looking back at the code and examples, I always thought MongoDB emitted events in the format `<prefix>.<db>.<collection>`, regardless of how the collection is distributed. 

If so, I do wonder whether changing `ByLogicalTableRouter` is really the right approach. If the goal is simply to reroute topics from A to B, use a more general TopicRegexRouter SMT or Content-based Routing, not the `ByLogicalTableRouter`.

This SMT is designed explicitly for situations where the connector emits changes for a single table/collection using different names, e.g., `public.my_table_parta`, `public.my_table_partb`, `public.my_table_partc`. In PostgreSQL, each of these is a table partition that all refer to a single root table `public.my_table`. Users can optionally configure PG to do this at the publication level, but this SMT allows it at the event level when the publication isn't set to do it.

So IMHO, if MongoDB does not actually handle sharded collections the same way a relational database like PostgreSQL handles partitions, then I don't believe the SMT should be changed to support MongoDB, because now we're effectively applying two distinct semantics.

jpechane:
@Naros This is a tough one. While I agree that the SMT is originally aimed at merging of partitioned tables into a single one I'd also argue that in community it is used as regular router too.

Also let's suppose that you are doing your own sharding in MongoDB - that would cover that scenario too,

Naros:
I see your point @jpechane. 

Then, in terms of @rk3rn3r's point, I suppose we could consider a rename, but I must admit that the term "collection" in a relational database has a very different meaning. To me, my native Oracle hat immediately thinks in terms of VARRAY, Associative Array, or Nested Table setups, as these are called Collections in PL/SQL. It's not a term that relational database experts would immediately see as "table".

For that reason, I'd probably recommend that if we did a rename, we find a more neutral name, perhaps favoring something like `Topic` instead, given that it is predominantly what the SMT is manipulating anyway. Furthermore, this ties into Jiri's point about it being used as regular router, too.

wdyt?

jpechane:
@Naros `ToLogicalTopicRouter` ?

kartikangiras:
@Naros @jpechane Could you please summarise if any changes are required for this code change Thanks for the help.

Naros:
@kartikangiras, see the above discussion on the renaming.  

1. We should move all the logic into a new class called `ToLogicalTopicRouter`, then have `ByLogicalTableRouter` extend it, marking `ByLogicalTableRouter` as deprecated.
2. In the `ByLogicalTableRouter` class, inside the `configure` method, before calling the super method, log a warning indicating that the old class is deprecated and that the connector configuration should be adjusted to use the new name.

Please make sure to do these refactoring changes in a separate commit.

There should be a follow-up PR that updates the documentation to use the new name and notes that the former name has been deprecated and will be removed in a future build.

Does that make sense?

kartikangiras:
@naros @jpechane Could you PTAL I have made the refactoring changes in a separate commit. As soon as this one gets approved & merged, I will start working upon the follow-up documentation PR.

debezium-connect-plugins/src/main/java/io/debezium/transforms/ToLogicalTopicRouter.java
```diff
@@ -0,0 +1,479 @@
...
+
+    private static final Field TOPIC_REPLACEMENT = Field.create("topic.replacement")
+            .withDisplayName("Topic replacement")
+            .withType(ConfigDef.Type.STRING)
+            .withWidth(ConfigDef.Width.LONG)
+            .withImportance(ConfigDef.Importance.LOW)
+            .required()
+            .withDescription("The replacement string used in conjunction with " + TOPIC_REGEX.name() +
+                    ". This will be used to create the new topic name.");
+    private static final Field KEY_ENFORCE_UNIQUENESS = Field.create("key.enforce.uniqueness")
```
Naros, line 77: I realize this wasn't in the original code, but since we're creating a new class, can we make sure each `private static final Field` declaration has a space between them? It's just easier to read and is consistent with the other config blocks in the codebase.

Naros (commented):
Overall, LGTM @kartikangiras, just one tiny formatting point.  Otherwise lets wait for @jpechane to give his input next week after the holiday weekend.

debezium-connect-plugins/src/main/java/io/debezium/transforms/ToLogicalTopicRouter.java
kartikangiras: I have fixed the formatting issue.

Naros:
Rebased the PR on the latest `main` to get a better sense of failures.

Naros:
Test failures are unrelated. @jpechane any other input on this?

jpechane:
LGTM, still I'd like to have a follow-up commit that would clean-up the naming in the rename SMT like using `source` naming instread of `table` etc. `logical.table.cache.size` config option can be renamed to `logical.destination.cache.size` etc.
@mfvitale One question for you, with this deprecation approach should the old SMT be removed from metadata provider or not?

mfvitale:
>  @mfvitale One question for you, with this deprecation approach should the old SMT be removed from metadata provider or not?

Well I think that we should't remove it util we remove it from code. The missing part is that we should mark the deprecation also in the descriptors so that the UI will be able to show a warning so that the user can then change to the new one.

kartikangiras:
@naros @jpechane I have made a seperate commit to ensure the cleanup and naming changes.

debezium-connect-plugins/src/main/java/io/debezium/transforms/ToLogicalTopicRouter.java
```diff
@@ -0,0 +1,483 @@
...
+            .withDisplayName("Schema Name Adjustment")
+            .withEnum(SchemaNameAdjustmentMode.class, SchemaNameAdjustmentMode.NONE)
+            .withWidth(ConfigDef.Width.MEDIUM)
+            .withImportance(ConfigDef.Importance.LOW)
+            .withDescription(
+                    "Specify how the message key schema names derived from the resulting topic name should be adjusted for compatibility with the message converter used by the connector, including:"
+                            + "'avro' replaces the characters that cannot be used in the Avro type name with underscore (default)"
+                            + "'none' does not apply any adjustment");
+
+    private static final Field LOGICAL_DESTINATION_CACHE_SIZE = Field.create("logical.destination.cache.size")
```
Naros, line 131: Let's use `topic` rather than `destination` to align with the SMT name. In addition, this should still honor the old `logical.table.cache.size` property as a deprecated alias for backward compatibility.

debezium-connect-plugins/src/main/java/io/debezium/transforms/ToLogicalTopicRouter.java
```diff
@@ -0,0 +1,483 @@
...
+                    "guaranteed to be unique across different sources. We need some identifier added to the key that distinguishes " +
+                    "the different physical tables.");
+
+    private static final Field KEY_FIELD_NAME = Field.create("key.field.name")
+            .withDisplayName("Key field name")
+            .withType(ConfigDef.Type.STRING)
+            .withWidth(ConfigDef.Width.LONG)
+            .withImportance(ConfigDef.Importance.LOW)
+            // Default is prefixed with "__dbz__" to minimize the likelihood of a conflict with an existing key field name.
+            .withDefault("__dbz__sourceIdentifier")
```
Naros, line 106: Changing the default value here introduces a backward-incompatible configuration change for anyone currently using the old SMT. We do generally prefer to avoid these unless they're absolutely needed.  

@jpechane, do you have a specific preference here? Mark it as backward-incompatible so we note the change in breaking changes, or leave the default so this does not impact any existing deployments?

jpechane: Ideally when old SMT name is used then old default value should be in place. When new SMT is used then new default value should be used.
In the meantime this issue should be marked as backward incompatible and the text should inform both about deprecation and future change of the default identified value.

Naros (approved):
LGTM, just waiting on CI.

Naros:
Applied, thanks @kartikangiras

PR #7180, opened 2026-03-13
debezium/dbz#1421 Allow using record value and record header with row deletion

### LINKED ISSUE #1421: Allow using record_value and record_header with row deletion [DBZ-9583]
Migrated from [DBZ-9583](https://issues.redhat.com/browse/DBZ-9583)

The JDBC sink connector currently only supports deletion if the connector configuration has set the {{primary.key.mode}} to {{record_key}}.  

There's not an immediately obvious reason why we shouldn't be able to support deletion with {{record_value}} (with optional {{primary.key.fields}} set) or {{record_header}}.

pugarte7:
As described on the issue, the JDBC sink connector currently only supports deletion if the connector configuration has set the {{primary.key.mode}} to {{record_key}}.

We thought that a better approach would be to let the user delete the connector configuration when the {{primary.key.mode}} has the {{record_value}}, {{record_header}} or {{record_key}} defined. 

In this PR we have changed the JDBC Sink Connector Configuration logic, so It allows the new logic. We have also adapted the corresponding test logic and the documentation regarding the connector's new logic.

I have ensured that the new logic passes the tests, and ran the linting and formatting before submitting this PR.

Fixes debezium/dbz#1421

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -1770,7 +1770,7 @@ Use with caution where strong data consistency is required.
 |[[jdbc-property-delete-enabled]]<<jdbc-property-delete-enabled, `+delete.enabled+`>>
 |`false`
 |Specifies whether the connector processes `DELETE` or _tombstone_ events and removes the corresponding row from the database.
-Use of this option requires that you set the xref:jdbc-property-primary-key-mode[`primary.key.mode`] to `record.key`, `record.value` or `record.header`.
+Use of this option requires that you set the xref:jdbc-property-primary-key-mode[`primary.key.mode`] to `record_key`, `record_value` or `record_header`.
 
 |[[jdbc-property-truncate-enabled]]<<jdbc-property-truncate-enabled, `+truncate.enabled+`>>
 |`false`
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -1770,7 +1770,7 @@ Use with caution where strong data consistency is required.
 |[[jdbc-property-delete-enabled]]<<jdbc-property-delete-enabled, `+delete.enabled+`>>
 |`false`
 |Specifies whether the connector processes `DELETE` or _tombstone_ events and removes the corresponding row from the database.
-Use of this option requires that you set the xref:jdbc-property-primary-key-mode[`primary.key.mode`] to `record.key`.
+Use of this option requires that you set the xref:jdbc-property-primary-key-mode[`primary.key.mode`] to `record.key`, `record.value` or `record.header`.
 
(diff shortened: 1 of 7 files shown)


documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -1557,7 +1557,7 @@ Use with caution where strong data consistency is required.
 |[[jdbc-property-delete-enabled]]<<jdbc-property-delete-enabled, `+delete.enabled+`>>
 |`false`
 |Specifies whether the connector processes `DELETE` or _tombstone_ events and removes the corresponding row from the database.
-Use of this option requires that you set the xref:jdbc-property-primary-key-mode[`primary.key.mode`] to `record.key`.
+Use of this option requires that you set the xref:jdbc-property-primary-key-mode[`primary.key.mode`] to `record.key`, `record.value` or `record.header`.
```
Naros:
```suggestion
Use of this option requires that you set the xref:jdbc-property-primary-key-mode[`primary.key.mode`] to `record_key`, `record_value` or `record_header`.
```

Naros (commented):
Hi @pugarte7, were there no changes required to handle the new modes in the underlying primary-key-handling runtime code?

jpechane:
@pugarte7 Hi, I'd expect integration tests to verify the new modes are working.

pugarte7:
> Hi @pugarte7, were there no changes required to handle the new modes in the underlying primary-key-handling runtime code?

Yes, they are indeed needed, but I am a little bit unsure where to make those changes. Could you give me some hints about what parts are responsible for that?

Naros:
I'd suggest taking a look at `KafkaDebeziumSinkRecord#getFilteredKey`. There is a switch there for the various modes, so it may already be fully taken care of, but it's worth double checking. I'd suggest adding the integration tests requested by Jiri first, and then verify whether the code works as expected. If it doesn't, then you have a starting point. 

If you have any questions, just ping us @pugarte7.

battmush:
This isn't my PR, but this is functionality that would be of great use to my team.  I would be willing to complete the integration test work if this is otherwise stalled.

pugarte7:
Hi guys! Apologies for the delay, I started a new job and have been completely out. I just uploaded the integration tests, so the PR should be good to merge!

Naros (commented):
No problem @pugarte7, could you do an interactive rebase and change the issue key from `DBZ-9583` to `debezium/dbz#1421` now that we've moved to GH issues. 

It would also be useful to rebase on `main` as this branch is 900+ commits behind and there has been a lot of recent changes on the JDBC sink connector code base.

pugarte7:
Hi @Naros! I just rebased everything and looks good. Thank you!

vjuranek (approved):
LGTM

pugarte7:
Hi @Naros! There is anything I can on my side to have this PR merged?

Naros (approved):

Naros:
Applied, thanks @pugarte7 :tada:

PR #7183, opened 2026-03-13
debezium/dbz#258 feat: add S3LargeMessagePostProcessor for large column values

### LINKED ISSUE #258: Explore SMT for externalizing large column values [DBZ-1541]
Migrated from [DBZ-1541](https://issues.redhat.com/browse/DBZ-1541)

This came up in a Twitter [discussion|https://twitter.com/gunnarmorling/status/1181687283071803392]: it could be useful to have an SMT which externalizes large BLOB/CLOB column values and propagates the external reference in data change events. The motivation is to avoid large messages in Apache Kafka.

One particular implementation could be based on Amazon S3: when creating a change event, the values of any configured large columns would be written to the S3 object storage, and in change events the corresponding field value would describe bucket name and object id. The object id should be based on the offset of the change event (and the column name, and optionally a before/after flag), so that the same id would be used when processing the same offset a second time after a connector restart.

Consumers might either resolve the reference, retrieve the referenced object and persist that in a sink datastore. More common usage would probably simply persist the object reference, pushing the object retrieval to readers of that sink datastore.

Eventually, multiple object stores may be supported, but S3 will be a good candidate for an initial PoC.

kartikangiras:
### Description:

Added new Debezium PostProcessor in the debezium-storage-s3 module that externalizes large BLOB/CLOB field values to Amazon S3 and replaces them in change events with a structured reference along with new unit cases.

fixes https://github.com/debezium/dbz/issues/258


debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,391 @@
...
+        thresholdBytes = config.getInteger(THRESHOLD_BYTES_CONFIG, DEFAULT_THRESHOLD_BYTES);
+        if (thresholdBytes <= 0) {
+            throw new DebeziumException("Configuration '" + THRESHOLD_BYTES_CONFIG + "' must be a positive integer.");
+        }
+
+        LOGGER.info("S3LargeMessagePostProcessor configured: bucket='{}', region='{}', thresholdBytes={}",
+                bucket, regionName, thresholdBytes);
+
+        if (s3ClientOverride != null) {
+            s3Client = s3ClientOverride;
```
kmos: add a return in the if and remove the else

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,391 @@
...
+        }
+        else {
+            final Region region = Region.of(regionName);
+
+            final AwsCredentialsProvider credentialsProvider;
+            final String accessKeyId = config.getString(ACCESS_KEY_ID_CONFIG);
+            final String secretAccessKey = config.getString(SECRET_ACCESS_KEY_CONFIG);
+            if (accessKeyId != null && secretAccessKey != null) {
```
kmos: extract a method that returns the `credtianlsProvider` use a quick return and remove the redundant else. Inline in 126

debezium-core/src/main/java/io/debezium/processors/externalize/ExternalizeColumnsPredicateBuilder.java
```diff
@@ -0,0 +1,55 @@
...
+ */
+package io.debezium.processors.externalize;
+
+import java.util.Arrays;
+import java.util.function.Predicate;
+import java.util.stream.Collectors;
+
+import io.debezium.function.Predicates;
+
+public class ExternalizeColumnsPredicateBuilder {
```
jpechane: What is purpose of this class?

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,391 @@
...
+            .version(1)
+            .doc("Reference to a large field value stored in Amazon S3")
+            .field("bucket", Schema.STRING_SCHEMA)
+            .field("objectId", Schema.STRING_SCHEMA)
+            .build();
+
+    private static final java.lang.reflect.Field STRUCT_SCHEMA_FIELD;
+    private static final java.lang.reflect.Field STRUCT_VALUES_FIELD;
+
+    static {
```
jpechane: Why is the reflection needed?

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,391 @@
...
+            return false;
+        }
+        switch (schema.type()) {
+            case STRING:
+                return ((String) value).getBytes(StandardCharsets.UTF_8).length >= thresholdBytes;
+            case BYTES:
+                if (value instanceof byte[]) {
+                    return ((byte[]) value).length >= thresholdBytes;
+                }
+                else if (value instanceof ByteBuffer) {
```
kmos: else if is redundant: use only if. extract all the case statement for bytes in a method

debezium-storage/debezium-storage-s3/src/test/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessorTest.java
```diff
@@ -0,0 +1,299 @@
...
+    @Test
+    void testConfigureValidConfig() {
+        processor.configure(baseConfig());
+    }
+
+    @Test
+    void testConfigureMissingBucket() {
+        Map<String, Object> config = baseConfig();
+        config.remove(S3LargeMessagePostProcessor.BUCKET_NAME_CONFIG);
+        assertThrows(DebeziumException.class, () -> processor.configure(config));
```
kmos, line 79: can you give a space between assertion and setup? in this way is easy to identify the assertions. It's valid for all the tests

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,391 @@
...
+    private volatile S3Client s3Client;
+
+    S3Client s3ClientOverride = null;
+
+    @Override
+    public void configure(Map<String, ?> properties) {
+        final Configuration config = Configuration.from(properties);
+
+        bucket = config.getString(BUCKET_NAME_CONFIG);
+        if (bucket == null || bucket.isBlank()) {
```
Naros:
```suggestion
        if (Strings.isNullorBlank(bucket)) {
```

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,391 @@
...
+            throw new DebeziumException("Configuration '" + BUCKET_NAME_CONFIG + "' is required for " + getClass().getSimpleName());
+        }
+
+        final String regionName = config.getString(REGION_NAME_CONFIG);
+        if (regionName == null || regionName.isBlank()) {
```
Naros:
```suggestion
        if (Strings.isNullOrBlank(regionName)) {
```

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
Naros:
```suggestion
            if (!Strings.isNullorBlank(accessKeyId) && !Strings.isNullOrBlank(secretAccessKey)) {
```
If you intend for null and empty values to skip the if-block, I'd suggest using the above.

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,391 @@
...
+                LOGGER.info("Using DefaultCredentialsProvider for S3 authentication.");
+                credentialsProvider = DefaultCredentialsProvider.create();
+            }
+
+            S3ClientBuilder builder = S3Client.builder()
+                    .region(region)
+                    .credentialsProvider(credentialsProvider);
+
+            final String endpointStr = config.getString(ENDPOINT_CONFIG);
+            if (endpointStr != null && !endpointStr.isBlank()) {
```
Naros:
```suggestion
            if (!Strings.isNullOrBlank(endpointStr)) {
```

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,391 @@
...
+        }
+    }
+
+    private boolean meetsThreshold(Schema schema, Object value) {
+        if (value == null) {
+            return false;
+        }
+        switch (schema.type()) {
+            case STRING:
+                return ((String) value).getBytes(StandardCharsets.UTF_8).length >= thresholdBytes;
```
Naros: Let's prefer the newer arrow-based switch statement.
```suggestion
        return switch (schema.type()) {
            case STRING -> ((String) value).getBytes(StandardCharsets.UTF_8).length >= thresholdBytes;
```

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
Naros: Let's also favor using the modern `instanceof` pattern match syntax, e.g.:
```java
if (value instanceof byte[] byteData) {
    return byteData.length >= thresholdBytes;
}
```

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,391 @@
...
+                else if (value instanceof ByteBuffer) {
+                    return ((ByteBuffer) value).remaining() >= thresholdBytes;
+                }
+                return false;
+            default:
+                return false;
+        }
+    }
+
+    private byte[] toBytes(Schema schema, Object value) {
```
Naros, line 279: Another case to use `return switch` and instanceof pattern matching.

Naros (changes requested):

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
kartikangiras: done

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
kartikangiras: I have added a method for this.

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
kartikangiras: Added a new method for bytes switch case and used the new instanceof pattern

debezium-storage/debezium-storage-s3/src/test/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessorTest.java
kartikangiras: Done

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
kartikangiras: done

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
kartikangiras: done

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
kartikangiras: Yeah, i have updated the switch statement to use the new arrow based statement

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
kartikangiras: done

kartikangiras:
@kmos @Naros I have resolved all the changes requested on the code change.

debezium-storage/debezium-storage-s3/src/test/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessorTest.java
```diff
@@ -0,0 +1,315 @@
...
+        Struct resultAfter = envelope.getStruct(Envelope.FieldName.AFTER);
+        Object replacedValue = resultAfter.get("content");
+        assertInstanceOf(Struct.class, replacedValue);
+
+        Struct ref = (Struct) replacedValue;
+        assertEquals(BUCKET, ref.getString("bucket"));
+
+        String objectId = ref.getString("objectId");
+        assertNotNull(objectId);
+        assert objectId.contains("content");
```
kmos: Why here `assert`? Can you use an assertions from junit5 ? My suggestion is to uniform the test with only one test framework. An alternative is to use `assertj` with junit5 as test runner

kartikangiras: Resolved, now the entire test file uses junit5 as the only testing framework with all the assertions from it.

debezium-core/src/main/java/io/debezium/processors/externalize/ExternalizeColumnsPredicateBuilder.java
kartikangiras: I added this new class for future usecase when processor that would use it to filter which columns to externalise. but I have removed this as there is no immediate usecase now.

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
kartikangiras: this Reflection is used to directly modify internal fields of Kafka Connect's Struct class.

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,393 @@
...
+
+    private static final String REFERENCE_SCHEMA_NAME = "io.debezium.storage.s3.processor.LargeMessageReference";
+
+    public static final Schema REFERENCE_SCHEMA = SchemaBuilder.struct()
+            .name(REFERENCE_SCHEMA_NAME)
+            .version(1)
+            .doc("Reference to a large field value stored in Amazon S3")
+            .field("bucket", Schema.STRING_SCHEMA)
+            .field("objectId", Schema.STRING_SCHEMA)
+            .build();
```
Naros: Should this perhaps go into `SchemaFactory`, @jpechane wdyt?

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,393 @@
...
+        catch (NoSuchFieldException e) {
+            throw new ExceptionInInitializerError("Cannot access Kafka Connect Struct internal fields: " + e.getMessage());
+        }
+    }
+
+    private String bucket;
+    private int thresholdBytes;
+    private volatile S3Client s3Client;
+
+    S3Client s3ClientOverride = null;
```
Naros: I really don't care for the package private member variable for test mocking.  Could we instead create a method that creates the client and then in the test suite we mock the method to return our test-constructed object instead?

Naros (commented):
Hi @kartikangiras, this is coming along well. I left a few more comments.

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
kartikangiras: I have created a method for this.

kmos:
@kartikangiras Can you rebase the branch ?

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
Naros: @jpechane, it appears it's used to bypass the immutability of the `Struct` since this `PostProcessor` changes the layout of the field to a nested `Struct`.

Naros (changes requested):
@kartikangiras, so this introduces a small problem in that events when they do not meet the threshold, will emit fields as either `STRING` or `BYTES` depending on the field type. However, once the threshold is met, we change the event layout. 

From a schema registry perspective, this change is most likely an incompatible shift in how events are represented for a given subject, which creates all sorts of issues.

I propose that, rather than changing the event's schema, we find a way to represent the bucket and object ID in the field's existing form.

For example, if the field type is `STRING`, find a way to combine the two values with a delimiter. Consumers would check whether the field is prefixed with the bucket name + delimiter and decode it accordingly. If it is `BYTES`, you create the string representation and serialize it as `bytes`.

This avoids the schema registry issue and provides a way to supply the downstream consumers with the S3 locator data.  wdyt?

kartikangiras:
@naros I came up with an implementation in which there will be encoding of the S3 reference inline within the original field value instead of changing the event schema structure along with type preserved Encoding. Wdyt about this.

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,281 @@
...
+            return bytes;
+        }
+        throw new DebeziumException("Unsupported field type for S3 offloading: BYTES");
+    }
+
+    String buildObjectKey(Struct source, String fieldName, String qualifier) {
+        final StringBuilder sb = new StringBuilder();
+
+        if (source != null) {
+            appendSourceField(sb, source, "db");
```
jpechane: I don't think thid is good solution. `source` is specific for each connector. Should not you use fields from `source` info block schema?

@Naros WDYT?

jpechane (commented):
@kartikangiras LGTM, I believe we are almost there.  I left one comment.

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
Naros: I agree, this should at the very least should iterate the the `source` information block fields.

But I do have some reservations about the validity of the object key. @kartikangiras, if the goal here is to have a 1:1 relationship between an event and its S3 reference for each column, we need to check whether the information in the `source` information block is effectively unique for each event from a given source.

As an example, in older Oracle connector versions, the `source` block wasn't unique when a user performed two back-to-back updates to the same logical row, changing the large object value of the same column. We have fixed this in recent versions, but I wonder what other connectors likely have the same issue.

Just a brief skim across our connectors in the main repo, I think we're safe. Each of them, I believe, has a distinct set of values in their `source` information block to avoid ambiguity, but I can't say the same holds true for our non-core connectors. So we may want to document this as a requirement so users of the post processor are aware.

My last concern is with the default "unknown-source". I would argue that if you cannot resolve the object key to a valid value, then, for the "unknown-source" use case, the value passes as is. Multiple events that resolve to "unknown-source" thrash each other's values, leading to potential data loss at an event level, right?

kartikangiras:
Hi @jpechane,
I have refactored the `buildObjectKey` to iterate over the source fields instead of the previous hardcoded fields. Also in the case of no fields it returns null instead of unknown resource so the edge case is also covered now.

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,301 @@
...
+     * @return a unique S3 object key, or {@code null} if a unique key cannot be resolved
+     */
+    String buildObjectKey(Struct source, String fieldName, String qualifier) {
+        final StringBuilder sb = new StringBuilder();
+
+        if (source != null) {
+            for (org.apache.kafka.connect.data.Field field : source.schema().fields()) {
+                final Object val = source.get(field);
+                if (val != null) {
+                    sb.append('/').append(field.name()).append('=').append(val);
```
Naros: This creates the object key with a leading `/`, which can cause issues with S3 console navigation and some SDK listing operations. Could we omit the `/` as the first character, or add a prefix before it, to avoid the problem?

In addition, if the field's value contains a `/` or `=` character, this creates a malformed key that breaks the parsing logic, so we should likely encode the value to prevent that.

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,301 @@
...
+        }
+    }
+
+    private void processRecordStruct(Struct recordStruct, Struct source, String qualifier) {
+        final Schema originalSchema = recordStruct.schema();
+        final List<org.apache.kafka.connect.data.Field> fields = originalSchema.fields();
+        final List<String> fieldsToReplace = new ArrayList<>();
+        for (org.apache.kafka.connect.data.Field field : fields) {
+            final Object fieldValue = recordStruct.get(field);
+            if (meetsThreshold(field.schema(), fieldValue)) {
```
Naros: This automatically offloads all string/bytes fields that meet the threshold. 

I wonder if we should consider a follow-up where we default to all fields, but a user could configure the post processor with a `columns.[include|exclude].list` property, like the `ReselectPostProcessor`, to be as granular as needed?

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,301 @@
...
+        return sb.toString();
+    }
+
+    private void uploadToS3(String objectKey, byte[] bytes, Schema.Type originalType) {
+        final PutObjectRequest request = PutObjectRequest.builder()
+                .bucket(bucket)
+                .key(objectKey)
+                .contentType(originalType == Schema.Type.STRING ? "text/plain; charset=utf-8" : "application/octet-stream")
+                .build();
+        s3Client.putObject(request, RequestBody.fromBytes(bytes));
```
Naros: We should consider adding an `ErrorHandlingMode` here that provides various strategies for handling S3 upload failures. At a minimum, the failure should be wrapped in a DebeziumException.

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,301 @@
...
+    private String encodeInlineReference(String objectKey) {
+        return INLINE_REFERENCE_PREFIX + bucket + INLINE_REFERENCE_DELIMITER + objectKey;
+    }
+
+    private boolean meetsThreshold(Schema schema, Object value) {
+        if (value == null) {
+            return false;
+        }
+        return switch (schema.type()) {
+            case STRING -> ((String) value).getBytes(StandardCharsets.UTF_8).length >= thresholdBytes;
```
Naros: For every `STRING` field, this allocates a byte array just to check the length. This can create unnecessary GC back-pressure in memory-constrained environments.

I'd suggest a small short-circuit to minimize the allocations except where necessary:
```suggestion
            case STRING -> {
                final String s = (String) value;
                yield s.length() >= thresholdBytes || s.getBytes(StandardCharsets.UTF_8).length >= thresholdBytes;
            }
```

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,301 @@
...
+    public static final String ENDPOINT_CONFIG = "large.message.s3.endpoint";
+
+    public static final int DEFAULT_THRESHOLD_BYTES = 100 * 1024;
+
+    private static final String INLINE_REFERENCE_PREFIX = "__debezium:s3:ref:v1:";
+    private static final String INLINE_REFERENCE_DELIMITER = "|";
+
+    private String bucket;
+    private int thresholdBytes;
+    private volatile S3Client s3Client;
```
Naros:
```suggestion
    private S3Client s3Client;
```
I don't believe this needs to be volatile.

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
```diff
@@ -0,0 +1,301 @@
...
+
+    @Override
+    public void close() {
+        if (s3Client != null) {
+            LOGGER.debug("Closing S3 client.");
+            s3Client.close();
+        }
+    }
+
+    private void processRecordStruct(Struct recordStruct, Struct source, String qualifier) {
```
Naros, line 204: This method iterates the fields twice. This could be collapsed into a single pass over the fields to improve performance.

Naros (changes requested):
Hi @kartikangiras I've left a few comments. In addition to the follow, we should also add documentation for the post processor by updating `post-processors/index.adoc` and adding a dedicated documentation page for this implementation.

debezium-storage/debezium-storage-s3/src/main/java/io/debezium/storage/s3/processor/S3LargeMessagePostProcessor.java
kartikangiras: This is covered now. Processor exposes Include List and Exclude List config options. When neither is provided, all threshold meeting string/bytes fields are offloaded by default and include/exclude lists then narrow that down, similar to ReselectPostProcessor.

documentation/modules/ROOT/pages/post-processors/s3-large-message.adoc
```diff
@@ -0,0 +1,165 @@
...
+ifdef::community[]
+:toc:
+:toc-placement: macro
+:linkattrs:
+:icons: font
+:source-highlighter: highlight.js
+
+toc::[]
+
+== Overview
```
Naros, line 17: Unlike the `ReselectColumnsPostProcessor` that ships with Debezium's common framework, this one explicitly requires users to include the S3 storage module, `debezium-storage-s3`. Could you add a callout to this prerequisite so that users who want to use this know an explicit additional dependency is required?

Naros (commented):
Hi @kartikangiras this LGTM, just one final update to docs and it's G2G.

documentation/modules/ROOT/pages/post-processors/s3-large-message.adoc
kartikangiras: Sure.

Naros (approved):

Naros:
Applied, thanks @kartikangiras :tada:

PR #7194, opened 2026-03-17
debezium/dbz#1663 Optionally exclude client_id, username, commit_timestamp, start_timestamp tracking by configuration

### LINKED ISSUE #1663: Allow non-critical metadata columns to be removed from Logminer query

Relates broadly to discussion [here](https://debezium.zulipchat.com/#narrow/channel/348250-community-oracle/topic/Oracle.20Heap.20Footprint)

As part of the heap footprint changes it was made possible to drop one of the metadata used in the oracle connector by setting log.mining.buffer.track.rs_id=false.  Based on the feedback I think it would be useful to add similar configurations for other metadata columns that are non-critical.  In particular, I have been looking at these:

- CLIENT_ID
- USERNAME
- COMMIT_TIMESTAMP
- START_TIMESTAMP 

The first three in particular I've found to consistently improve performance by a small amount (about 3%) when removed (even when there is no output for the query) and there is likely no real downside to having users able to toggle these on/off.

nathan-smit-1:
Fixes debezium/dbz#1663

## Description
Relates to this conversation [here](https://debezium.zulipchat.com/#narrow/channel/348250-community-oracle/topic/ROWID.20based.20snapshot.20for.20oracle)

The goal here is to allow a user to optionally exclude certain metadata columns.  This means excluding it from the source_info block as well as the V$LOGMNR_CONTENTS query.  The goal of the PR is to prevent the need for a hardcoded list by using a new ResultSetValueResolver.  For every row fetched, it simply loops through this pre-built array to map the ResultSet values into the LogMinerEventRow

I've currently focused on implementing this for metadata columns that appear to be the most impactful from my own testing.  I've found removing all of the optional columns leads to ~3% performance improvement.  I've also done testing vs current main branch and I don't believe I've added any overhead to the existing process.

Note:  This has also been added for RS_ID and essentially replaces the optional removal process that was implemented there.  Let me know if that's going to be a problem...

## PR Checklist

- [ x] I have read the [contribution guidelines](https://github.com/debezium/debezium/blob/main/CONTRIBUTING.md) and the [governance document](https://github.com/debezium/governance/blob/main/GOVERNANCE.md) on PR expectations.
- [ x] Minimal changes to code not directly related to your change (e.g. no unnecessary formatting changes or refactoring to existing code)
- [ x] One feature/change per PR unless tightly coupled
- [ x] Do a rebase on upstream `main`

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -306,7 +306,56 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             .withValidation(Field::isRequired)
             .withDescription("This controls whether the 'RS_ID' column values are tracked. " +
                     "When set to true (the default), the 'RS_ID' values are buffered and provided in events when available. " +
-                    "When set to false, the 'RS_ID' values are not buffered and can reduce the memory footprint.");
+                    "When set to false, the 'RS_ID' column is excluded from the LogMiner query and its values are not buffered, " +
+                    "reducing both the memory footprint and query bandwidth.");
+
+    public static final Field LOG_MINING_BUFFER_TRACK_CLIENT_ID = Field.create("log.mining.buffer.track.client_id")
+            .withDisplayName("Toggle whether the 'client_id' value is tracked and buffered")
+            .withType(Type.BOOLEAN)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withDefault(true)
+            .withValidation(Field::isRequired)
+            .withDescription("This controls whether the 'CLIENT_ID' column values are tracked. " +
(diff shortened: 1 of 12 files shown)


Naros (approved):
Great work @nathan-smit-1. I spot-tested it locally, and everything appeared to pass. 

I also made a small change to push all optional columns to the end of the query to make the ordinal math a bit easier, rather than having them spread out over the column list. This allows the use of the minimum number of lambdas.

Let's see how CI goes, but LGTM.

nathan-smit-1:
@Naros I see there were errors here but I tdon't think related to my change.  I imagine though I need to rebase as I see there are conflicts?

Naros:
Hi @nathan-smit-1 if you could rebase, that'd help. The `BlockingSnapshotIT` IIRC were fixed in another PR.

nathan-smit-1:
@Naros Okay, hopefully I did that right.  Did the rebase and locally ran the tests and everything passed.  Left your commit in place rather than squash it for now

Naros (approved):

Naros:
Thanks @nathan-smit-1, applied :tada:

PR #7201, opened 2026-03-18
debezium/dbz#1563 Clear incremental snapshot state

### LINKED ISSUE #1563: State leaks from aborted incremental snapshot to the next
## Bug report

debezium-connector-mysql 3.4.0

---

Not relevant

---

Running local mysql in docker/podman; mysql:8.0.26

---

After I start an incremental snapshot of all tables (with datacollection `.*`), then abort that incremental snapshot - I expect that I can start another incremental snapshot on a single table.

---

**What behavior do you see?**

After "immediately" aborting the snapshot of "all tables" and starting a new incremental snapshot of a single table, parts of the state from the first snapshot leaks into the next snapshot execution.

What I've seen is that io.debezium.pipeline.source.snapshot.incremental.AbstractIncrementalSnapshotChangeEventSource#addDataCollectionNamesToSnapshot is called with a signalPayload where `signalPayload.offsetContext.incrementalSnapshotContext` contains e.g. `chunkEndPosition`, `lastEventKeySent`, `maximumKey` etc from the first table in the previous snapshot execution. As a result the new snapshot execution resumes the snapshot from the maximumKey from another snapshot of _another table_, which is obviously not correct.

Also I observed that `total_rows_scanned` in the TABLE_SCAN_COMPLETED notification includes rows from the previous aborted snapshot.

---

Yes, I have seen it in 3.1.x and the same behaviour is in 3.4.0 which is the latest version.

---

No, but I can provide DEBUG/TRACE logs on request.

---

Idk, I have built my own rest api to trigger snapshots, but it is just a thin wrapper around the JmxSignalChannel from debezium. I saw some mentioning of a feature to send signals from debezium-ui in recent versions, but I have never used debezium-ui so not sure if this issue can be reproduced from there.

tommyk-gears:
Fixes debezium/dbz#1563

## Description
Clear incremental snapshot state when a snapshot is aborted, so that it does not "leak" into a future incremental snapshot.

## PR Checklist

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
+++ b/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
@@ -660,6 +660,8 @@ public abstract class AbstractIncrementalSnapshotChangeEventSource<P extends Par
                     final List<String> ids = databaseSchema
                             .tableIds()
                             .stream()
+                            .map(databaseSchema::tableFor) // Get the id from the Table object to get original tableId case
+                            .map(Table::id)
                             .map(TableId::identifier)
                             .filter(t -> Pattern.compile(x).matcher(t).matches())
                             .collect(Collectors.toList());
--- a/debezium-connector-common/src/test/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSourceTest.java
+++ b/debezium-connector-common/src/test/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSourceTest.java
@@ -5,6 +5,8 @@
  */
 package io.debezium.pipeline.source.snapshot.incremental;
 
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
 import static org.mockito.Mockito.RETURNS_DEEP_STUBS;
 import static org.mockito.Mockito.doReturn;
 import static org.mockito.Mockito.mock;
@@ -15,6 +17,7 @@ import static org.mockito.Mockito.when;
 import java.sql.SQLException;
 import java.sql.SQLNonTransientConnectionException;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 
 import org.junit.jupiter.api.BeforeEach;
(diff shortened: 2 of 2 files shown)


mfvitale:
@tommyk-gears thanks for the PR, would you be able to provide an integration test?

tommyk-gears:
I have tried (hard) to reproduce the issue in an integration test. I've finally come to the conclusion that my first fix is incorrect (even though it solves the issue at hand). The _real_ problem as I understand it now is caused by the underlying fact that I'm running MySql with `lower_case_table_names=2` which gives case insensitive table names, but table names are still stored in mixed-case in mysql.

Now, one may wonder how does this lead to "state leak" in incremental snapshots?
It boils down to this line: https://github.com/debezium/debezium/blob/5178bdb585d25a7c415d35c136f0bc4cd192046f/debezium-core/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java#L589

Where the `collectionId` has the table-id in all lowercase, but the `currentTable` has the table-id in mixed case. Hence, the equals-check returns false and `stopCurrentTableId` is never set, which in turn "disables" the whole logics around stopping of ongoing snapshot a few lines further down.

Anyway - now that I understand the root cause better, I still unfortunately cannot produce an integration test that verifies this - mainly because the docker setup has case sensitive file system, and MySql refuses to use `lower_case_table_names=2` on a case sensitive file system.

I'm leaning towards changing `AbstractIncrementalSnapshotChangeEventSource#expandAndDedupeDataCollectionIds` so that it returns table ids in their original case rather than the current all-lowercase. At the same time I am slightly concerned that `TableId#equals` is unaware of case-sensitivity. Gut feeling is that it _might_ make sense to make the `TableId` object itself aware of the case sensitivity setting, but maybe you have already considered that option and decided it is a bad idea?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros, line 670: So the problem here is the following:
|Scenario | PR as written |
|---|---|
|Start `.*`, Stop `.*` | works |
|Start with explicit `game.game_integration`, Stop `.*`|Stop misses the stored lowercase id|
|Start with lowercase regex `game.\.game_.*`|no longer matches, falls through as bogus|

I would recommend a slightly different strategy:
```suggestion
                .flatMap(x -> {
                	final List<String> ids = databaseSchema
                			.tableIds()
                			.stream()
                			.map(TableId::identifier)
                			.filter(t -> Pattern.compile(x).matcher(t).matches())
                			.toList();
                	return ids.isEmpty() ? Straem.of(x) : ids.stream();
                })
                .map(this::toDeclaredIdentifier)
                .distinct()
                .toList();
```
Then for `toDeclaredIdentifier`:
```java
private String toDeclaredIdentifier(String dataCollectionId) {
	final Table table = databaseSchema.tableFor(TableId.parse(dataCollectionId));
	return table != null ? table.id().identifier() : dataCollectionId;
}
```
This fixes the last two broken scenarios.

Naros: Now ids that were stored by any earlier offsets may not be in declared form, so this helps deal with that by also matching via schema:
```suggestion
final List<String> matching = context.getDataCollections().stream()
	.map(dataCollection -> dataCollection.getId().identifier())
	.filter(stored -> toDeclaredIdentifier(stored).equals(dataCollectionId))
	.toList();
	
bool removed = false;	
for (String stored : matching) {
	removed |= context.removeDataCollectionFromSnapshot(stored);
}

if (removed) {
	stopped.add(dataCollectionId);
	LOGGER.info("Removed '{}' from incremental snapshot collection list.", collectionId);
}
else {
	LOGGER.warn("Could not remove '{}', collection is not part of the incremental snapshot.", collectionId);
}
```

Naros (commented):
Hi @tommyk-gears, I believe we're close. Could you see my comment below? Perhaps we need to expand the test suite a bit?

PR #7210, opened 2026-03-20
debezium/dbz#1726 Order by primary key when ROWID pseudocolumn is used as surrogate key

### LINKED ISSUE #1726: When using ROWID pseudocolumn as surrogate-key sort by primary key of table

We have found that when using the rowid as the surrogate-key during an incremental snapshot, the ORDER BY ROWID operation becomes very expensive and leads to significantly degraded performance.

I would like to instead decouple this such that ROWID is used in the projection and the PK of the table is used in the order by.

Relates to this discussion [here]([#community-oracle > Unable to transfer a large table @ 💬](https://debezium.zulipchat.com/#narrow/channel/348250-community-oracle/topic/Unable.20to.20transfer.20a.20large.20table/near/579705339))

---

nathan-smit-1:
Fixes debezium/dbz#1726

## Description
The goal here is to allow decoupling so that when ROWID is used in the projection we can still use the PK for the order by as ordering by the ROWID creates significant overhead

## PR Checklist

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
+++ b/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
@@ -52,25 +52,33 @@ public abstract class AbstractChunkQueryBuilder<T extends DataCollectionId>
 
     @Override
     public String buildChunkQuery(IncrementalSnapshotContext<T> context, Table table, int limit, Optional<String> additionalCondition) {
+        final Optional<Object[]> upperBound = getChunkUpperBound(context);
+        final List<Column> queryColumns = getQueryColumns(context, table);
         String condition = null;
         // Add condition when this is not the first query
         if (context.isNonInitialChunk()) {
-            final Object[] maximumKey = context.maximumKey().get();
             final Object[] chunkEndPosition = context.chunkEndPosititon();
             final StringBuilder sql = new StringBuilder();
             // Window boundaries
             addLowerBound(context, table, chunkEndPosition, sql);
             // Table boundaries
-            sql.append(" AND ");
-            addUpperBound(context, table, maximumKey, sql);
+            if (upperBound.isPresent()) {
+                sql.append(" AND ");
+                addUpperBound(context, table, upperBound.get(), sql);
+            }
             condition = sql.toString();
         }
-        final List<Column> queryColumns = getQueryColumns(context, table);
+        else if (upperBound.isPresent()) {
+            final StringBuilder sql = new StringBuilder();
+            addUpperBound(context, table, upperBound.get(), sql);
+            condition = sql.toString();
+        }
+        final List<Column> orderByColumns = getOrderByColumns(context, table);
         if (jdbcConnection.nullsSortLast().isEmpty() && queryColumns.stream().anyMatch(Column::isOptional)) {
             // You need to override nullsSortLast on JdbcConnection for your connector if you want to be able to chunk based on nullable columns.
             throw new UnsupportedOperationException("The sort order of NULL values in the incremental snapshot key is unknown.");
         }
-        final String orderBy = queryColumns.stream()
+        final String orderBy = orderByColumns.stream()
                 .map(c -> jdbcConnection.quoteIdentifier(c.name()))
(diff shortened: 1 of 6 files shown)


Naros:
Test failures are unrelated.

jpechane:
@nathan-smit-1 Can you please rebase on the latest main? Thanks

ezazwar1:
What if there is no primary key for the table in that case what will happen, I believe that there should be a property in which we could specify what column to order by

nathan-smit-1:
> @nathan-smit-1 Can you please rebase on the latest main? Thanks

Have done this now

nathan-smit-1:
> What if there is no primary key for the table in that case what will happen, I believe that there should be a property in which we could specify what column to order by

So this is probably clearer if you look at the original ROWID PR here:  https://github.com/debezium/debezium/pull/6930

It is still a requirement that in order to do an incremental snapshot your table needs to have a primary key.  This is primarily because in the dedupe step the PK is still required.  It's probably possible to create a snapshot type using the physical row identifier that doesn't require a PK at all but that was outside of the scope of what I was looking to implement.

ezazwar1:
> probably possible to create a snapshot type using the physical row identifier that doesn't require a PK at all but that was outside of the scope of what I was looking to implem

For using ROWID to take a snapshot for tables without a primary key, i did search and didnt find any good options available

1) I cannot take the snapshot as a initial table in that case the row_id is null.
2) Only way to do the snapshot was using a signal table and specifying the surrogate key as ROWID but that has the problem in which it does an order by ROWID which is very slow.

If you can point me out how to do it that will be great help

Naros:
If your table does not have a PK and there is no suitable column to serve as a surrogate or where you can define a custom primary key using `message.key.columns`, then I'm afraid that, for incremental snapshots, the `ROWID` pseudo-column is your only option as the surrogate. 

Your other option is to rely on initial or blocking snapshots for the keyless table.

ezazwar1:
> If your table does not have a PK and there is no suitable column to serve as a surrogate or where you can define a custom primary key using `message.key.columns`, then I'm afraid that, for incremental snapshots, the `ROWID` pseudo-column is your only option as the surrogate.
> 
> Your other option is to rely on initial or blocking snapshots for the keyless table.

What i am trying to do is to use Debezium to do replication from oracle to MSSQL, the problem is that if i do not have ROWID in the keyless tables it won't work for the deletes and updates as it cannot govern which record to update and delete

I tried to write the custom query snapshot.select.statement.overrides by specifying ROWID in that query itself but it didn't work as the schema for ROWID was unknown to the Debezium, Another try i did was to select a column which had null values always and i did ROWID AS column_name and designated this column with message.key.columns as kind of the primary key, The initial sync was fast and it worked but then the problem was with any update in the table because the logminner is not aware about snapshot.select.statement.overrides and it failed because now it finds the null values in the column that is part of message.key.columns

Naros:
Hi @ezazwar1, using `ROWID` for anything other than a surrogate key for snapshots will cause a myriad of issues. These are simply not sufficiently stable for any sort of replication, and certain database operations can invalidate a `ROWID` for a given row, even when the row is not deleted.

The best alternative is to use `message.key.columns` at the source if there are 1 or more logical columns (columns that are not virtually generated or hidden) that can be used to represent the key. If that's not possible, the alternative at the sink for deletions is to enforce the PK to use all columns by setting `primary.key.mode` to `record_value` when using the Debezium JDBC sink.

Naros:
@nathan-smit-1 rebased on latest main to get the BlockingSnapshotIT changes in for a cleaner test pass.

nathan-smit-1:
> @nathan-smit-1 rebased on latest main to get the BlockingSnapshotIT changes in for a cleaner test pass.

Thanks!  The DBZ team is making changes faster than I can rebase!

jpechane:
@nathan-smit-1 I am afraid we need one more rebase

nathan-smit-1:
@jpechane   Done.  Rebase didn't seem to break anything

jpechane (approved):
@nathan-smit-1 LGTM, thanks! I triggered Oracle CI just once again. Can be merged if CI passes.

Naros:
Test failures are unrelated.

Naros (approved):

Naros:
Applied, thanks @nathan-smit-1  :tada:

PR #7211, opened 2026-03-20
debezium/dbz#1395 Schema Inconsistency Fixing

### LINKED ISSUE #1395: MariaDB Connector Fails with Schema Inconsistency on Tables Using Generated Columns in Unique Constraints [DBZ-9412]
Migrated from [DBZ-9412](https://issues.redhat.com/browse/DBZ-9412)

h1. Bug report

The Debezium MariaDB source connector encounters a schema inconsistency error when a table includes a *generated column* that is part of a {*}unique key constraint{*}. In such cases, MariaDB automatically creates a *hidden index column* to support the constraint, which is not reflected in the table schema retrieved by Debezium. However, during CDC processing, Debezium receives {{before}} and {{after}} row data that includes this hidden/generated column, causing a mismatch between the internal schema and the binlog row data.
 * *Connector:* io.debezium.connector.mariadb.MariaDbConnector

 * *Version:* *3.0.7.Final* 

{code:java}
{
"connector.class": "io.debezium.connector.mariadb.MariaDbConnector",
"snapshot.locking.mode": "none",
"errors.log.include.messages": "false",
"transforms": "unwrap",
"transforms.unwrap.drop.tombstones": "true",
"transforms.unwrap.type": "io.debezium.transforms.ExtractNewRecordState",
"errors.log.enable": "true",
"key.converter": "io.confluent.connect.avro.AvroConverter",
"transforms.unwrap.field.name.adjustment.mode": "avro",
"errors.tolerance": "none",
"max.batch.size": "2048",
"snapshot.mode": "no_data",
"connect.timeout.ms": "10000",
"transforms.unwrap.delete.handling.mode": "rewrite",
"max.queue.size": "8192",
"incremental.snapshot.chunk.size": "32768",
"transforms.unwrap.schema.name.adjustment.mode": "avro",
"schema.history.internal.store.only.captured.databases.ddl": "true",
"schema.history.internal.store.only.captured.tables.ddl": "true",
"binary.handling.mode": "bytes",
"value.converter.ignore.default.for.nullables": "true",
"read.only": "false",
"converters": "JdbcSinkDataTypesConverter",
"JdbcSinkDataTypesConverter.type": "io.debezium.connector.binlog.converters.JdbcSinkDataTypesConverter",
"JdbcSinkDataTypesConverter.selector.real": "",
"JdbcSinkDataTypesConverter.selector.boolean": "",
"transforms.unwrap.add.headers": "op",
"database.port": "3306",
"database.ssl.mode": "disable",
"database.connectionTimeZone": "SERVER",
"database.hostname": "mariadb",
"transforms.unwrap.add.fields": "source.ts_ms,ts_ms",
"database.useGtid": "none",
"table.include.list": "mydatabase.tbl_settings",
"key.converter.schema.registry.url": "...",
…
} {code}
 * {*}Database{*}: MariaDB (10.6.19).

 * {*}Deployment{*}: cloud & docker.

Debezium should either ignore the hidden generated columns or properly account for them in its internal schema representation to prevent schema mismatch errors.
h2. What behavior do you see?

Debezium MariaDB connector fails with a schema mismatch when a generated column is used in a unique constraint—MariaDB creates a hidden column (e.g., {{{}DB_ROW_HASH_*{}}}) for the supporting index, which is excluded from the snapshot but appears in binlog row data, causing inconsistency.

Yes
{code:java}
Task threw an uncaught and unrecoverable exception. Task is being killed and will not recover until manually restarted (org.apache.kafka.connect.runtime.WorkerTask)
	at io.debezium.pipeline.ErrorHandler.setProducerThrowable(ErrorHandler.java:67)
	at io.debezium.connector.binlog.BinlogStreamingChangeEventSource.handleEvent(BinlogStreamingChangeEventSource.java:591)
	at io.debezium.connector.binlog.BinlogStreamingChangeEventSource.lambda$execute$17(BinlogStreamingChangeEventSource.java:209)
	at com.github.shyiko.mysql.binlog.BinaryLogClient.notifyEventListeners(BinaryLogClient.java:1281)
	at com.github.shyiko.mysql.binlog.BinaryLogClient.listenForEventPackets(BinaryLogClient.java:1103)
	at com.github.shyiko.mysql.binlog.BinaryLogClient.connect(BinaryLogClient.java:657)
	at com.github.shyiko.mysql.binlog.BinaryLogClient$7.run(BinaryLogClient.java:959)
	at java.base/java.lang.Thread.run(Thread.java:840)
Caused by: io.debezium.DebeziumException: Error processing binlog event
	... 7 more
Caused by: io.debezium.DebeziumException: Error processing row in tbl_settings, internal schema size 17, but row size 18 , restart connector with schema recovery mode.
	at io.debezium.connector.binlog.BinlogStreamingChangeEventSource.validateChangeEventWithTable(BinlogStreamingChangeEventSource.java:1065)
	at io.debezium.connector.binlog.BinlogStreamingChangeEventSource.lambda$handleInsert$28(BinlogStreamingChangeEventSource.java:837)
	at io.debezium.connector.binlog.BinlogStreamingChangeEventSource.handleChange(BinlogStreamingChangeEventSource.java:1101)
	at io.debezium.connector.binlog.BinlogStreamingChangeEventSource.handleInsert(BinlogStreamingChangeEventSource.java:832)
	at io.debezium.connector.binlog.BinlogStreamingChangeEventSource.lambda$execute$7(BinlogStreamingChangeEventSource.java:183)
	at io.debezium.connector.binlog.BinlogStreamingChangeEventSource.handleEvent(BinlogStreamingChangeEventSource.java:571)
	... 6 more {code}
 * Create a MariaDB table with a *generated column* used in a *unique key constraint*

 * Insert or update data in the table

 * Observe Debezium connector logs and failure behavior

We introduced a configurable workaround to optionally ignore schema inconsistency errors caused by hidden columns like {{{}DB_ROW_HASH_*{}}}. A new flag {{ignore.schema.row.inconsistency}} was added, along with support for a list of specific tables to ignore ({{{}ignore.schema.row.inconsistency.tables{}}}). When enabled, the connector skips schema validation and logs a warning if the number of columns in binlog data doesn't match the internal schema, allowing CDC to continue without failure.

KMohnishM:
Fixes debezium/dbz#1395<the Debezium Issue Number>

## Description

This PR addresses DBZ-1395 where the MariaDB connector fails due to a schema mismatch when generated columns are used in unique constraints.

MariaDB introduces hidden columns (e.g. DB_ROW_HASH_*) that appear in binlog events but are not exposed via JDBC metadata, leading to a mismatch between the internal schema and row data.

To handle this safely, the connector now queries INFORMATION_SCHEMA to detect such hidden/generated columns. These are tracked internally but excluded from the emitted schema, and validation accounts for them without shifting any visible columns.

In non-fail modes, rows are skipped rather than attempting unsafe alignment, avoiding any risk of silent data misplacement.

Added integration test (MariaDbHiddenColumnsIT) to cover this scenario.

## PR Checklist

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -779,6 +779,7 @@ M. Gökhan Akgül
 Maithem
 Mark Allanson
 Mark Banierink
+Mohnish Kodukulla
 Rahul Khanna
 Razvan Laurus
 Ronak Jain
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -546,6 +546,7 @@ Mohamed El Shaer
 Mohamed Pudukulathan
 Mohammad Akhil
 Mohammad Yousuf Minhaj Zia
+Mohnish Kodukulla
 Moira Tagle
 Mostafa Ghadimi
 Moustapha Mahfoud
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
@@ -8,6 +8,7 @@ package io.debezium.connector.binlog;
 import static io.debezium.util.Strings.isNullOrEmpty;
 
 import java.io.IOException;
+import java.io.Serializable;
 import java.security.GeneralSecurityException;
 import java.security.KeyStore;
 import java.security.KeyStoreException;
@@ -1225,24 +1226,42 @@ public abstract class BinlogStreamingChangeEventSource<P extends BinlogPartition
         informAboutUnknownTableIfRequired(partition, offsetContext, event, tableId, null);
     }
 
-    private void validateChangeEventWithTable(Table table, Object[] before, Object[] after) {
+    private boolean validateChangeEventWithTable(Table table, Serializable[] before, Serializable[] after) {
         if (table != null) {
             int columnSize = table.columns().size();
-            String message = "Error processing {} of row in {} because it's different column size with internal schema size {}, but {} size {}, " +
-                    "restart connector with schema recovery mode.";
+
+            String message = "Error processing {} of row in {} because it's different column size with " +
+                    "internal schema size {}, but {} size {}, restart connector with schema recovery mode.";
             if (before != null && columnSize != before.length) {
-                LOGGER.error(message, "before", table.id().table(), columnSize, "before", before.length);
-                throw new DebeziumException(
-                        "Error processing row in " + table.id().table() + ", internal schema size " + columnSize + ", but row size " + before.length + " , " +
-                                "restart connector with schema recovery mode.");
+                if (inconsistentSchemaHandlingMode == EventProcessingFailureHandlingMode.FAIL) {
+                    LOGGER.error(message, "before", table.id().table(), columnSize, "before", before.length);
+                    throw new DebeziumException(
+                            "Error processing row in " + table.id().table() + ", internal schema size " + columnSize
+                                    + ", but row size " + before.length
+                                    + ", restart connector with schema recovery mode.");
+                }
+
+                LOGGER.warn(message, "before", table.id().table(), columnSize, "before", before.length);
+                LOGGER.warn("Skipping before row for table {} due to schema/row length mismatch "
+                        + "to avoid potential column misalignment.", table.id().table());
+                return false;
             }
             if (after != null && columnSize != after.length) {
-                LOGGER.error(message, "after", table.id().table(), columnSize, "after", after.length);
-                throw new DebeziumException(
-                        "Error processing row in " + table.id().table() + ", internal schema size " + columnSize + ", but row size " + after.length + " , " +
-                                "restart connector with schema recovery mode.");
+                if (inconsistentSchemaHandlingMode == EventProcessingFailureHandlingMode.FAIL) {
+                    LOGGER.error(message, "after", table.id().table(), columnSize, "after", after.length);
+                    throw new DebeziumException(
+                            "Error processing row in " + table.id().table() + ", internal schema size " + columnSize
+                                    + ", but row size " + after.length
+                                    + ", restart connector with schema recovery mode.");
+                }
+
+                LOGGER.warn(message, "after", table.id().table(), columnSize, "after", after.length);
+                LOGGER.warn("Skipping after row for table {} due to schema/row length mismatch "
+                        + "to avoid potential column misalignment.", table.id().table());
+                return false;
             }
         }
+        return true;
     }
 
     private <T extends EventData, U> void handleChange(P partition,
@@ -1275,7 +1294,10 @@ public abstract class BinlogStreamingChangeEventSource<P extends BinlogPartition
             if (startingRowNumber < numRows) {
                 for (int rowIndex = startingRowNumber; rowIndex != numRows; ++rowIndex) {
                     U row = rows.get(rowIndex);
-                    changeEventValidator.validate(tableId, row);
+                    // Skip this row if validation fails (e.g., schema/row column count mismatch)
+                    if (!changeEventValidator.validate(tableId, row)) {
+                        continue;
+                    }
                     offsetContext.setRowNumber(rowIndex, numRows);
                     offsetContext.event(tableId, eventTimestamp);
                     changeEmitter.emit(tableId, row);
@@ -1536,7 +1558,7 @@ public abstract class BinlogStreamingChangeEventSource<P extends BinlogPartition
 
     @FunctionalInterface
     private interface ChangeEventValidator<U> {
-        void validate(TableId tableId, U row);
+        boolean validate(TableId tableId, U row);
     }
 
     /**
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
@@ -1329,24 +1329,42 @@ public abstract class BinlogStreamingChangeEventSource<P extends BinlogPartition
         return aligned;
     }
 
-    private void validateChangeEventWithTable(Table table, Object[] before, Object[] after) {
+    private boolean validateChangeEventWithTable(Table table, Serializable[] before, Serializable[] after) {
         if (table != null) {
             int columnSize = table.columns().size();
-            String message = "Error processing {} of row in {} because it's different column size with internal schema size {}, but {} size {}, " +
-                    "restart connector with schema recovery mode.";
+
+            String message = "Error processing {} of row in {} because it's different column size with " +
+                    "internal schema size {}, but {} size {}, restart connector with schema recovery mode.";
             if (before != null && columnSize != before.length) {
-                LOGGER.error(message, "before", table.id().table(), columnSize, "before", before.length);
-                throw new DebeziumException(
-                        "Error processing row in " + table.id().table() + ", internal schema size " + columnSize + ", but row size " + before.length + " , " +
-                                "restart connector with schema recovery mode.");
+                if (inconsistentSchemaHandlingMode == EventProcessingFailureHandlingMode.FAIL) {
+                    LOGGER.error(message, "before", table.id().table(), columnSize, "before", before.length);
+                    throw new DebeziumException(
+                            "Error processing row in " + table.id().table() + ", internal schema size " + columnSize
+                                    + ", but row size " + before.length
+                                    + ", restart connector with schema recovery mode.");
+                }
+
+                LOGGER.warn(message, "before", table.id().table(), columnSize, "before", before.length);
+                LOGGER.warn("Skipping before row for table {} due to schema/row length mismatch "
+                        + "to avoid potential column misalignment.", table.id().table());
+                return false;
             }
             if (after != null && columnSize != after.length) {
-                LOGGER.error(message, "after", table.id().table(), columnSize, "after", after.length);
-                throw new DebeziumException(
-                        "Error processing row in " + table.id().table() + ", internal schema size " + columnSize + ", but row size " + after.length + " , " +
-                                "restart connector with schema recovery mode.");
+                if (inconsistentSchemaHandlingMode == EventProcessingFailureHandlingMode.FAIL) {
+                    LOGGER.error(message, "after", table.id().table(), columnSize, "after", after.length);
+                    throw new DebeziumException(
+                            "Error processing row in " + table.id().table() + ", internal schema size " + columnSize
+                                    + ", but row size " + after.length
+                                    + ", restart connector with schema recovery mode.");
+                }
+
+                LOGGER.warn(message, "after", table.id().table(), columnSize, "after", after.length);
+                LOGGER.warn("Skipping after row for table {} due to schema/row length mismatch "
+                        + "to avoid potential column misalignment.", table.id().table());
+                return false;
             }
         }
+        return true;
     }
 
     private <T extends EventData, U> void handleChange(P partition,
@@ -1379,7 +1397,10 @@ public abstract class BinlogStreamingChangeEventSource<P extends BinlogPartition
             if (startingRowNumber < numRows) {
                 for (int rowIndex = startingRowNumber; rowIndex != numRows; ++rowIndex) {
                     U row = rows.get(rowIndex);
-                    changeEventValidator.validate(tableId, row);
+                    // Skip this row if validation fails (e.g., schema/row column count mismatch)
+                    if (!changeEventValidator.validate(tableId, row)) {
+                        continue;
+                    }
                     offsetContext.setRowNumber(rowIndex, numRows);
                     offsetContext.event(tableId, eventTimestamp);
                     changeEmitter.emit(tableId, row);
@@ -1650,7 +1671,7 @@ public abstract class BinlogStreamingChangeEventSource<P extends BinlogPartition
 
     @FunctionalInterface
     private interface ChangeEventValidator<U> {
-        void validate(TableId tableId, U row);
+        boolean validate(TableId tableId, U row);
     }
 
     /**
--- a/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogSchemaValidateIT.java
+++ b/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogSchemaValidateIT.java
@@ -11,15 +11,18 @@ import static org.junit.jupiter.api.Assertions.fail;
 
 import java.nio.file.Path;
 import java.sql.SQLException;
+import java.time.Duration;
 import java.util.List;
 import java.util.concurrent.atomic.AtomicReference;
 
 import org.apache.kafka.connect.source.SourceConnector;
 import org.apache.kafka.connect.source.SourceRecord;
+import org.awaitility.Awaitility;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
+import io.debezium.config.CommonConnectorConfig;
 import io.debezium.config.Configuration;
 import io.debezium.connector.binlog.BinlogConnectorConfig.SnapshotMode;
 import io.debezium.connector.binlog.util.BinlogTestConnection;
@@ -28,6 +31,7 @@ import io.debezium.connector.binlog.util.UniqueDatabase;
 import io.debezium.doc.FixFor;
 import io.debezium.jdbc.JdbcConnection;
 import io.debezium.junit.SkipWhenDatabaseVersion;
+import io.debezium.junit.logging.LogInterceptor;
 
 /**
  * @author Inki Hwang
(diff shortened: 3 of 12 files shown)


KMohnishM:
@Naros @jpechane 
Adding a bit more context on the approach I took here.

The main issue seems to come from MariaDB introducing hidden columns (like DB_ROW_HASH_*) for generated columns in unique constraints. These show up in binlog row events but are not visible through JDBC metadata, which leads to the row vs schema length mismatch during streaming.
Instead of going with a mismatch-tolerance or ignore-based approach, I tried to handle this at the schema level for MariaDB:

- I am querying INFORMATION_SCHEMA.COLUMNS directly to detect hidden/generated columns that JDBC omits.
- These columns are tracked internally (counted), but explicitly excluded from Debezium’s relational schema so they don’t appear in emitted records.
- During binlog validation, I account for these hidden columns when comparing expected vs actual row width, without attempting any positional remapping.

The goal was to avoid the unsafe case you mentioned (e.g. [a, b, hidden, c] vs [a, b, c]) where shifting could silently corrupt data. So there is no column realignment happening  only validation that tolerates the extra hidden columns in a controlled way.

For now I have kept this scoped to MariaDB/binlog connectors, without changing any shared/core behavior across other connectors.

A couple of things I am looking for feedback are on:
- Whether this schema-level handling (vs config-based workaround) is the right direction
- If querying INFORMATION_SCHEMA is acceptable here, or if there’s a preferred metadata source/pattern in Debezium
- Any edge cases I might be missing around hidden column ordering or visibility

Happy to adjust direction based on your feedback.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogConnectorConnection.java
```diff
@@ -643,4 +643,12 @@ public void appendToDdlStatement(String dbName, StringBuilder ddl) {
         }
     }
 
+    /**
+     * Gets the number of hidden columns mathematically excluded from the table's schema.
+     * Default is 0, overridden by databases like MariaDB that hide invisible sequence hashes.
+     */
+    public int getHiddenColumnCount(TableId tableId) {
```
jpechane: I'd rather call it system column than hidden column.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/jdbc/MariaDbConnection.java
```diff
@@ -103,4 +128,78 @@ public GtidSet filterGtidSet(Predicate<String> gtidSourceFilter, String offsetGt
...
+                            }
+
+                            ColumnEditor editor = Column.editor()
+                                    .name(columnName)
+                                    .type(rs.getString(2))
+                                    .length(rs.getInt(3) == 0 ? -1 : rs.getInt(3))
+                                    .scale(rs.getInt(5) == 0 ? null : rs.getInt(5))
+                                    .optional("YES".equalsIgnoreCase(rs.getString(6)))
+                                    .position(rs.getInt(7))
+                                    .generated(true);
```
jpechane: I think we need an additional flag for this scenario. Generated means that the values for the column are generated not that the column as such is generated by database.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
```diff
@@ -1088,24 +1089,40 @@ private void informAboutUnknownTableIfRequired(P partition, O offsetContext, Eve
         informAboutUnknownTableIfRequired(partition, offsetContext, event, tableId, null);
     }
 
-    private void validateChangeEventWithTable(Table table, Object[] before, Object[] after) {
+    private boolean validateChangeEventWithTable(Table table, Serializable[] before, Serializable[] after) {
         if (table != null) {
             int columnSize = table.columns().size();
+            int hiddenColumnCount = connection.getHiddenColumnCount(table.id());
```
jpechane: I'd prefer to use naming invisible to alig with offical database naming.

jpechane (changes requested):
@KMohnishM Thanks, this is great start. I belive it just needs a bit of polish and we are done.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/jdbc/MariaDbConnection.java
```diff
@@ -18,20 +22,37 @@
...
  */
 public class MariaDbConnection extends BinlogConnectorConnection {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(MariaDbConnection.class);
 
+    /**
+     * Tracks the number of mathematically invisible DB_ROW_HASH columns per table so Debezium can explicitly exclude
+     * them from the payload while safely forgiving the array length mismatch natively.
+     */
+    private final java.util.Map<TableId, Integer> hiddenColumnCounts = new java.util.concurrent.ConcurrentHashMap<>();
```
Naros: No need for FQCN, the classes are imported.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/jdbc/MariaDbConnection.java
```diff
@@ -103,4 +128,78 @@ public GtidSet filterGtidSet(Predicate<String> gtidSourceFilter, String offsetGt
...
+                                    .generated(true);
+
+                            columns.add(editor.create());
+                        }
+                    }
+                    return null;
+                });
+
+                // Ensure columns are sorted by physical ORDINAL_POSITION naturally
+                Collections.sort(columns);
```
Naros, line 182: Is it possible to apply the ordinal sort in the query, and avoid the Java-level sort here?

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/jdbc/MariaDbConnection.java
Naros: Would it make sense to avoid the ambiguity between those generated by the database and those that are not to use the term `virtual column` in this context? It's what other vendors, like Oracle, use to make this distinction clearer. wdyt @jpechane ?

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/jdbc/MariaDbConnection.java
KMohnishM: @Naros 
At this point the columns list already contains the base columns from JDBC metadata (super.getColumnsDetails(...)). The INFORMATION_SCHEMA query only retrieves the missing generated/invisible columns.

Applying ORDER BY ORDINAL_POSITION there would only sort those additional columns among themselves, but they would still be appended to the existing list rather than placed at their correct positions relative to the JDBC-discovered columns.

Because of that, I am merging both sources and doing a final sort on the combined collection to ensure the full column list reflects the correct ordinal order.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/jdbc/MariaDbConnection.java
jpechane: @Naros `synthetic`?

Naros: That works for me, keeps it vendor neutral :wink:

KMohnishM:
@Naros @jpechane Thanks for the feedback!

I have updated the naming to use “synthetic” to keep it vendor-neutral and avoid ambiguity with generated/virtual columns.

Also incorporated the other suggested changes. Please let me know if anything else needs adjustment

KMohnishM:
@jpechane @Naros I am not sure why this merge conflicts appeared as I am not seeing any in my local repo , Also most of the changes in those files where additions , Could you Please look into it once and suggest me the next steps .

jpechane:
@KMohnishM Have you tried to rebase it on the latest main?

KMohnishM:
@jpechane Yeah I have rebased onto the latest main and i think changes can be cleanly merged , on local i am not getting any merge conflicts as such. 
Could you please review once all the changes?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
```diff
@@ -1088,24 +1089,47 @@ private void informAboutUnknownTableIfRequired(P partition, O offsetContext, Eve
         informAboutUnknownTableIfRequired(partition, offsetContext, event, tableId, null);
     }
 
-    private void validateChangeEventWithTable(Table table, Object[] before, Object[] after) {
+    private boolean validateChangeEventWithTable(Table table, Serializable[] before, Serializable[] after) {
         if (table != null) {
             int columnSize = table.columns().size();
+            int invisibleColumnCount = connection.getSyntheticColumnCount(table.id());
```
jpechane: Please make sure the new naming is used everywhere in the code and messages

KMohnishM: @jpechane 

I've updated the naming to use `synthetic` consistently across the codebase, including method names, variables, and related messages.

Also ensured there are no leftover references to the previous naming.

Could you please take another look and let me know if anything still needs adjustment?

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/jdbc/MariaDbConnection.java
```diff
@@ -18,20 +23,37 @@
...
  * A concrete implementation of {@link BinlogConnectorConnection} for MariaDB.
- *
  * @author Chris Cranford
  */
 public class MariaDbConnection extends BinlogConnectorConnection {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(MariaDbConnection.class);
 
+    // No longer need manual tracking of synthetic column counts as they are part of the relational model as synthetic columns
+
```
Naros: This can be removed, yes?

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/jdbc/MariaDbConnection.java
```diff
@@ -18,20 +23,37 @@
...
+    /**
+     * Gets the number of synthetic columns explicitly excluded for the given table.
+     * Now returns 0 as these columns are included in the relational model but marked as synthetic columns.
+     *
+     * @param tableId the table identifier; may not be null
+     * @return the number of synthetic columns excluded
+     */
+    public int getSyntheticColumnCount(TableId tableId) {
+        return 0;
+    }
```
Naros: As the relational model tracks this, this can be removed, yes, and the one in Binlog Connection?

COPYRIGHT.txt
```diff
@@ -748,6 +742,7 @@ M. Gökhan Akgül
 Maithem
 Mark Allanson
 Mark Banierink
+  Mohnish Kodukulla
```
Naros:
```suggestion
Mohnish Kodukulla
```
This way the release script name matching works and there are no leading spaces.

Naros (commented):
Hi @KMohnishM I left a few inline comments.

KMohnishM:
@Naros Thanks for the feedback. I have addressed the inline comments. Mention if any more changes are required

KMohnishM:
Hi @jpechane @Naros,

Could you please take another look when you get a chance? Happy to make any final adjustments.

Thanks!

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/jdbc/MariaDbConnection.java
```diff
@@ -18,10 +23,14 @@
...
+import io.debezium.relational.TableId;
+import io.debezium.relational.Tables.ColumnNameFilter;
+import io.debezium.relational.Tables.TableFilter;
 
 /**
  * A concrete implementation of {@link BinlogConnectorConnection} for MariaDB.
- *
  * @author Chris Cranford
  */
 public class MariaDbConnection extends BinlogConnectorConnection {
```
Naros, line 37: There are a few random tabs/line changes that can probably be avoided here to keep the diff clean.

Naros (commented):
Hi @KMohnishM just one tiny nit-pick with the diff, others LGTM.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/jdbc/MariaDbConnection.java
KMohnishM: @Naros Thanks for pointing that out. I cleaned up the unnecessary whitespace/tab-only changes in MariaDbConnection.java and kept the diff limited to the actual changes. Could you please take another look when you get a chance?

Naros:
Hi @KMohnishM can you take a look at the formatting for `BinlogSchemaValidateIT`?

KMohnishM:
@Naros I have fixed it . Could You take a look at it once ? Thanks!

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
```diff
@@ -1111,24 +1114,43 @@ private void informAboutUnknownTableIfRequired(P partition, O offsetContext, Eve
...
         if (table != null) {
             int columnSize = table.columns().size();
-            String message = "Error processing {} of row in {} because it's different column size with internal schema size {}, but {} size {}, " +
-                    "restart connector with schema recovery mode.";
-            if (before != null && columnSize != before.length) {
-                LOGGER.error(message, "before", table.id().table(), columnSize, "before", before.length);
-                throw new DebeziumException(
-                        "Error processing row in " + table.id().table() + ", internal schema size " + columnSize + ", but row size " + before.length + " , " +
-                                "restart connector with schema recovery mode.");
+            int expectedSize = columnSize;
```
Naros: The use of `expectedSize` here is redundant, let's use `columnSize` throughout.

debezium-connector-common/src/main/java/io/debezium/relational/TableSchemaBuilder.java
```diff
@@ -174,6 +174,7 @@ public TableSchema create(TopicNamingStrategy topicNamingStrategy, Table table,
 
         table.columns()
                 .stream()
+                .filter(column -> !column.isSynthetic())
```
Naros, line 180: @jpechane, if we are filtering synthetic columns from the non-key, should keys get the same treatment? Generated/Synthetic/Virtual columns in some databases cannot participate in primary keys, while in others, they can. But typically, where they cannot participate in the physical primary key, if the table is keyless, they can be present in unique constraints, which we use as a fallback for table keys.

Any opinion one way or the other?

KMohnishM:
Thanks @Naros, I’ll update `BinlogStreamingChangeEventSource.java` to remove the redundant `expectedSize` variable and use `columnSize` throughout.

For the `TableSchemaBuilder.java` key-column filtering, my current understanding is that synthetic columns should probably be excluded from emitted keys as well, since they are database-internal implementation details rather than user-visible columns. But I’ll wait for @jpechane’s opinion and align the change with the direction you both prefer.

KMohnishM:
Hi @Naros @jpechane,
 
 I have updated the branch with the following changes:
 
 1. **Code Cleanup**: Removed the redundant `expectedSize` variable in `BinlogStreamingChangeEventSource` and updated it to use `columnSize` directly throughout the method.
 2. **Key Schema Filtering**: Updated `TableSchemaBuilder` to also filter out synthetic columns from the generated `keyColumns` schema. This ensures database-internal hidden columns (like MariaDB's fallback unique constraint `DB_ROW_HASH_*`) are excluded from both key and value schemas consistently.
 
 Please let me know if you have any feedback or if anything else is needed!

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogSchemaValidateIT.java
```diff
@@ -388,6 +392,90 @@ public void shouldRecoverToSyncSchemaWhenAddColumnToEndWithSqlLogBinIsOffAndColu
         assertTombstone(tombstoneEvent);
     }
 
+    @Test
+    @FixFor("DBZ-1395")
```
Naros: Let's please use `@FixFor("dbz#1395")` or `@FixFor("debezium/dbz#1395")`, as the style you've chosen here uses the older Jira-based issue key format. This is here and a few other places in your PR.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/jdbc/MariaDbConnection.java
```diff
@@ -103,4 +113,86 @@ public GtidSet filterGtidSet(Predicate<String> gtidSourceFilter, String offsetGt
...
+
+                prepareQueryAndMap(sql, ps -> {
+                    ps.setString(1, tableId.catalog());
+                    ps.setString(2, tableId.table());
+                }, rs -> {
+                    while (rs.next()) {
+                        String columnName = rs.getString(1);
+                        // Only process it if the JDBC driver missed it completely
+                        boolean alreadyGenerated = columns.stream().anyMatch(c -> c.name().equals(columnName));
+                        if (!alreadyGenerated) {
```
Naros: Just to avoid confusion with generated columns, could we instead just do:
```suggestion
                        if (!columns.stream().anyMatch(c -> c.name().equals(columnName))) {
```
or name the variable something like `columnIsPresent`?

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/jdbc/MariaDbConnection.java
```diff
@@ -103,4 +113,86 @@ public GtidSet filterGtidSet(Predicate<String> gtidSourceFilter, String offsetGt
...
+                            boolean isSynthetic = false;
+                            if (columnName.startsWith("DB_ROW_HASH")) {
+                                LOGGER.debug("Including DB_ROW_HASH as synthetic column for {}: {}",
+                                        tableId, columnName);
+                                isSynthetic = true;
+                            }
+
+                            LOGGER.debug("Injecting synthetic/generated column intelligently for {}: {}",
+                                    tableId, columnName);
```
Naros:
```suggestion
                            LOGGER.debug("Injecting synthetic/generated column for {}: {}",
                                    tableId, columnName);
```

Naros (commented):
Just a few inline comments @KMohnishM and I think this will be good for merge.

KMohnishM:
@Naros 
 I have addressed all the review comments:
 1. Updated the `@FixFor` annotations and test comments to use the modern GitHub issue key format (`dbz#1395`).
 2. Inlined the stream check in `MariaDbConnection` to eliminate the `alreadyGenerated` variable.
 3. Simplified the debug log message as suggested.
 4. Rebased the branch onto the latest `main`.

 Please let me know if anything else is needed!

KMohnishM:
@Naros Please let me know if you’d like me to rebase or make any other changes.

Naros:
Hi @KMohnishM yes, it appears there are conflicts, so a rebase would be welcomed.

KMohnishM:
@jpechane Could you PTAL when you get time , Thanks !

PR #7212, opened 2026-03-21
debezium/dbz#1729 Add "AI Agents OSS Helper" support for Debezium

### LINKED ISSUE #1729: Add "AI Agents OSS Helper" support for Debezium
Since https://github.com/orpiske/ai-agents-oss-helper/issues/5 was merged AI Agents OSS Helper reads metadata from the `.oss-ai-helper-rules` directory.

Provide the rule files for AI Agents OSS Helper in `.oss-ai-helper-rules` directory.

rk3rn3r:
Add "AI Agents OSS Helper" support for Debezium

Provide the rule files for AI Agents OSS Helper in `.oss-ai-helper-rules/` directory.

closes debezium/dbz#1729

### DIFF
--- a/.oss-ai-helper-rules/project-guidelines.md
+++ b/.oss-ai-helper-rules/project-guidelines.md
@@ -1,0 +1,26 @@
+# Project Guidelines
+
+This rule file contains branching, commit, PR, and task-finding conventions for the project. Commands read this file to determine how to name branches, format commits, and search for tasks.
+
+Debezium does not provide any way to create a quick fix branch and quick-fix commits.
+There are exceptions for docs, CI fixes and releases. 
+
+- **Fix-issue branch:** `debezium/dbz#<ISSUE_NUMBER>-<short-slug>`
+- **Quick-fix branch:** NOT ALLOWED!
+- **SonarCloud branch:** _(not configured)_
+- **Commit format (fix-issue):** `debezium/dbz#<ISSUE_NUMBER> <brief description>`
+- **Commit format (quick-fix):** NOT ALLOWED!
+- **Commit format (docs):** `[docs] <brief description>`
+- **Commit format (release):** `[release] <brief description>`
+- **CI-fix branch:** `ci/debezium/dbz#<ISSUE_NUMBER>-<short-slug>`
+- **Commit format (ci-fix):** `[ci] <brief description>`
+- **PR creation:** always
+- **Find-task source:** GitHub labels
+- **Find-task beginner label:** `good first issue`
+- **Find-task experienced label:** _(none)_
+- **Find-task intermediate:** _(none)_
+- **Find-task repo:** `debezium/dbz` (issues are tracked in a separate repo)
+- **Scope-too-large redirect:** `/oss-create-issue`
+
+## Version
+1.0.0
--- a/.oss-ai-helper-rules/project-info.md
+++ b/.oss-ai-helper-rules/project-info.md
@@ -1,0 +1,43 @@
+# Project Information
+
+This rule file contains project-specific metadata used by OSS Helper commands. Commands detect the current project by matching `git remote get-url origin` against the remote pattern below.
+
+- **Remote pattern:** `debezium/debezium`
+- **GitHub repo:** `debezium/debezium`
+- **Issue tracker:** GitHub
+- **Issue tracker URL:** `https://github.com/debezium/dbz/issues`
+- **Issue ID format:** numeric
+- **SonarCloud component key:** _(none)_
+- **Documentation URL:** `https://debezium.io/documentation/`
+- **Related repositories:**
+  - `debezium/container-images` - Docker images for Debezium
+  - `debezium/debezium-charts` - Helm charts for Debezium
+  - `debezium/debezium-connector-cassandra` - CDC connector for Apache Cassandra
+  - `debezium/debezium-connector-cockroachdb` - CDC connector for CockroachDB
+  - `debezium/debezium-connector-db2` - CDC connector for Db2 LUW and z/OS
+  - `debezium/debezium-connector-ibmi` - CDC connector for IBM i (AS/400)
+  - `debezium/debezium-connector-informix` - CDC connector for IBM Informix
+  - `debezium/debezium-connector-ingres` - CDC connector for Ingres
+  - `debezium/debezium-connector-spanner` - CDC connector for Google Spanner
+  - `debezium/debezium-connector-vitess` - CDC connector for Vitess
+  - `debezium/debezium-descriptors-registry` - JSON configuration schemas registry for connectors, SMTs, and sinks
+  - `debezium/debezium-design-documents` - Design documents
+  - `debezium/debezium-examples` - Examples (configuration, Docker Compose files, etc.)
+  - `debezium/debezium.github.io` - Debezium Website
+  - `debezium/debezium-operator` - Kubernetes/OpenShift operator for Debezium Server
+  - `debezium/debezium-operator-manifests` - Operator bundle manifests for released versions
+  - `debezium/debezium-platform` - Opinionated data-centric view of Debezium components
+  - `debezium/debezium-quarkus` - Quarkus extensions for Debezium
+  - `debezium/debezium-server` - Standalone runtime for Debezium connectors
+  - `debezium/debezium-smt-go-pdk` - Go PDK for writing SMT transformations
+  - `debezium/jbang-catalog` - JBang scripts catalog
+  - `debezium/microservices-lab` - Lab for deploying streaming applications on Strimzi
+  - `debezium/mysql-binlog-connector-java` - Java library to process MySQL Binary Logs
+  - `debezium/onetimeserver` - MySQL instances for test environments
+  - `debezium/oracle-vagrant-box` - Vagrant box for Oracle DB Docker image (testing)
+  - `debezium/performance` - Performance testing
+  - `debezium/postgres-decoderbufs` - PostgreSQL logical decoder output plugin (Protocol Buffers)
+- **Create-issue supported:** yes
+
+## Version
+1.0.0
--- a/.oss-ai-helper-rules/project-standards.md
+++ b/.oss-ai-helper-rules/project-standards.md
@@ -1,0 +1,24 @@
+# Project Standards
+
+This rule file contains build tools, commands, and code style constraints for the project. Commands read this file to determine how to build, test, and format code.
+
+- **Build tool:** Maven
+- **Build command:** `./mvnw clean install`
+- **Test command:** `./mvnw clean install`
+- **Format command:** `./mvnw process-sources`
+- **Module-specific build:** yes (always run `./mvnw` in the module directory where changes occurred)
+- **Parallelized Maven:** no (resource intensive, do NOT parallelize Maven jobs)
+- **Code style restrictions:** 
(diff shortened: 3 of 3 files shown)


.oss-ai-helper-rules/project-guidelines.md
```diff
@@ -0,0 +1,26 @@
...
+There are exceptions for docs, CI fixes and releases. 
+
+- **Fix-issue branch:** `debezium/dbz#<ISSUE_NUMBER>-<short-slug>`
+- **Quick-fix branch:** NOT ALLOWED!
+- **SonarCloud branch:** _(not configured)_
+- **Commit format (fix-issue):** `debezium/dbz#<ISSUE_NUMBER> <brief description>`
+- **Commit format (quick-fix):** NOT ALLOWED!
+- **Commit format (docs):** `[docs] <brief description>`
+- **Commit format (release):** `[release] <brief description>`
+- **CI-fix branch:** `ci/debezium/dbz#<ISSUE_NUMBER>-<short-slug>`
```
rk3rn3r, line 15: Any preference about CI branch naming?

.oss-ai-helper-rules/project-standards.md
```diff
@@ -0,0 +1,22 @@
...
+- **Parallelized Maven:** no (resource intensive, do NOT parallelize Maven jobs)
+- **Code style restrictions:**
+  - Do NOT use Lombok (unless already present in the file)
+  - Do NOT change public API signatures without justification
+  - Do NOT add new dependencies without justification
+  - Maintain backwards compatibility for public APIs
+  - Eclipse-based formatter rules applied automatically during build
+  - Code style violations will fail CI
+  - Use `./mvnw process-sources` to auto-format code
+  - Commits MUST be signed off (`git commit -s`) and all code changes MUST be reviewed by a human user, understood by that human user and MUST explicitly be confirmed by that human user before committing
```
Naros, line 20: Agents should never sign off on commits.

.oss-ai-helper-rules/project-info.md
```diff
@@ -0,0 +1,43 @@
...
+  - `debezium/debezium-quarkus` - Quarkus extensions for Debezium
+  - `debezium/debezium-server` - Standalone runtime for Debezium connectors
+  - `debezium/debezium-smt-go-pdk` - Go PDK for writing SMT transformations
+  - `debezium/jbang-catalog` - JBang scripts catalog
+  - `debezium/microservices-lab` - Lab for deploying streaming applications on Strimzi
+  - `debezium/mysql-binlog-connector-java` - Java library to process MySQL Binary Logs
+  - `debezium/onetimeserver` - MySQL instances for test environments
+  - `debezium/oracle-vagrant-box` - Vagrant box for Oracle DB Docker image (testing)
+  - `debezium/performance` - Performance testing
+  - `debezium/postgres-decoderbufs` - PostgreSQL logical decoder output plugin (Protocol Buffers)
```
rk3rn3r, line 39: Is this list of projects okay? The relation should be declared to make AI better understand related code and examples. eg. probably `oracle-vagrant-box` and others could be removed? wdyt?

.oss-ai-helper-rules/project-standards.md
```diff
@@ -0,0 +1,22 @@
...
+
```
rk3rn3r, line 22: Please review those restrictions above.

.oss-ai-helper-rules/project-guidelines.md
Naros: I don't see why it needs to be any different from `Fix-issue branch`.

.oss-ai-helper-rules/project-standards.md
```diff
@@ -0,0 +1,22 @@
...
+This rule file contains build tools, commands, and code style constraints for the project. Commands read this file to determine how to build, test, and format code.
+
+- **Build tool:** Maven
+- **Build command:** `./mvnw clean install`
+- **Test command:** `./mvnw clean install`
+- **Format command:** `./mvnw process-sources`
+- **Module-specific build:** yes (always run `./mvnw` in the module directory where changes occurred)
+- **Parallelized Maven:** no (resource intensive, do NOT parallelize Maven jobs)
+- **Code style restrictions:**
+  - Do NOT use Lombok (unless already present in the file)
```
rk3rn3r: Any other library that AI and devs should avoid? `Guava` maybe?

.oss-ai-helper-rules/project-guidelines.md
rk3rn3r: Easier for AI (and human) to understand the context of the work in that branch. Probably obvious for humans that understand the context from human memory etc, harder for an AI machine definitely.

Probably we could even add that the agent makes another consistency check of affected files and runs related tests after the confirmation and only commits when all tests are green (entire test suite). So managing the AI commit might be better than trying to not allow commits that might be forced by the user somehow anyways.

.oss-ai-helper-rules/project-standards.md
rk3rn3r: Like I commented in the AI contribution guidelines PR, when the user instructs to commit, the agent will do. So I would personally say let's specify a process for that scenario and enforce human review. Wdyt?

Naros: I understand, and it's fine if the agent commits; it's just that the agent cannot add the sign-off.

rk3rn3r: Why not? Mine does it regularly. I get a quick popup to provide my personal passphrase and it gets signed. The agent simply executes the `git` shell call. Or wdym?

Naros: You are looking at the process purely through the lens of your use of AI, but I am looking at it from a much broader, large-scale perspective. 

There are numerous GitHub organizations that get code submitted by AI bots regularly, and some of those PRs are worthwhile, great changes in a vacuum where accountability doesn't matter. 

But when you factor in accountability, the problem is that PRs have to undergo an extremely painful process of authenticity and verification by reviewers. There's no other liable entity in the chain that can accept liability if a disputable piece of code makes it into the codebase. The DCO itself is there to help protect the process and share the liability and blame should this ever happen.

The guardrail here is to inform the AI that it isn't human, has no legal personhood, is incapable of entering into any legally binding contract, and therefore cannot bear any liability. As a consequence, the AI cannot verify the authenticity or provenance of the submitted changes because of how it's been trained and is therefore incapable of signing commits.

This does not mean AI cannot contribute; it means a human must always remain the accountable party in the chain, and the sign-off is where that accountability is formally recorded. So in your personal use case, you're formally accepting liability.

rk3rn3r: I promise you, I am not. Here I am partially looking at it from the angle of what is going to happen in reality and that'll be users having a `CLAUDE.local.md` that has highest priority that basically ignores a simple restriction like that.

I see your point from the legal perspective, but this is a technical ruleset and I think it's too easy to escape when only restricting commits in a system prompt.

Should we maybe split it into 2 rules (although I am not sure that'll be that effective)? One to restrict commits, one to provide requirements for finished work that is required before submitting / committing changes?

Naros: If you think that would help, I'm all for it. There's no magic bullet here to address this, which is part of what I dislike about the whole landscape, but here we are. I simply want to make sure we are protecting ourselves as much as we possibly can while being as open as possible.

.oss-ai-helper-rules/project-guidelines.md
```diff
@@ -0,0 +1,26 @@
...
+
+Debezium does not provide any way to create a quick fix branch and quick-fix commits.
+There are exceptions for docs, CI fixes and releases. 
+
+- **Fix-issue branch:** `debezium/dbz#<ISSUE_NUMBER>-<short-slug>`
+- **Quick-fix branch:** NOT ALLOWED!
+- **SonarCloud branch:** _(not configured)_
+- **Commit format (fix-issue):** `debezium/dbz#<ISSUE_NUMBER> <brief description>`
+- **Commit format (quick-fix):** NOT ALLOWED!
+- **Commit format (docs):** `[docs] <brief description>`
```
jpechane, line 13: This should be limited to only small docs updates likes typos etc. not larger ones.

.oss-ai-helper-rules/project-guidelines.md
```diff
@@ -0,0 +1,26 @@
+# Project Guidelines
+
+This rule file contains branching, commit, PR, and task-finding conventions for the project. Commands read this file to determine how to name branches, format commits, and search for tasks.
+
+Debezium does not provide any way to create a quick fix branch and quick-fix commits.
+There are exceptions for docs, CI fixes and releases. 
+
+- **Fix-issue branch:** `debezium/dbz#<ISSUE_NUMBER>-<short-slug>`
```
jpechane, line 8: Somewhere it should be stressed that if change go across multiple repositories then the branch names should be the same to have CI working properly.

rk3rn3r:
We should keep this PR until we decide how to proceed? (AA)

.oss-ai-helper-rules/project-guidelines.md
Naros: I genuinely don't think the AI gives $0.02 about the branch name. :) 

But one reason to consider a prefix is more for humans, particularly in the IDE, where if you had the branches named `ci-dbz#<issuekey>` or `ci/dbz#<issuekey>`, you would at least get all your CI-related branches grouped together. So there, it's more of an organizational benefit for humans.

But I rarely find adding extra slug-like information to branch names useful, but that's a matter of preference. It's not a hill I'll die on defending, so I'll defer to the team for a quorum. 

Worst case, I can always add a `CLAUDE.local.md` file or enforce a project memory that records my preferences to override these changes and avoid changing how I've worked.

.oss-ai-helper-rules/project-info.md
Naros: I genuinely wonder if these are needed. 

I've already made fixes across 2 or more repositories with OSS and didn't need to define these. The AI was smart enough to look at the Debezium org and resolve the right repositories given the context, but YMMV. My concern is whether this could become stale by accident, undermining its purpose.

Naros (commented):
One question Rene, should we consider adding to these files (either guidelines or in the standards file) a small section that says effective:

```
## Required reading before contributing:

- **Contribution guidelines:** see `CONTRIBUTING.md` in `debezium/debezium`
- **AI policy:** see `AI_POLICY.md` in `debezium/debezium`
```

Otherwise agents are more likely not to read either of the files in my experience using the OSS harness for the last several months.

rk3rn3r:
@Naros it depends on if they are used remotely or locally, right? But I agree, the specification makes sense.

Naros:
> it depends on if they are used remotely or locally, right?

I suspect it depends more on the AI agent in question. I primarily use just one agent, and so far every OSS harness fix it's made or recommended has never followed our guidelines in either of those files. 

So ya, I think being explicit is just the better strategy.

.oss-ai-helper-rules/project-standards.md
Naros: I'd rather not get in the business of specifically calling out dependencies, as it can be viewed negatively for many reasons. Instead, perhaps we should say that the AI agent should not add any new dependencies without a clear reason, and it must ask & justify the reason to the human before adding them. WDYT?

PR #7233, opened 2026-03-25
debezium/dbz#1715 Created dedicated SMT to convert MongoDB events into relational-style structs

### LINKED ISSUE #1715: Create dedicated SMT to convert MongoDB events into relational-style structs
## Description:

Reference: Issue #594 

Currently, unlike relational connectors that emit before and after states as nested `Struct` objects, MongoDB events emit them as simple JSON strings. To enable a shared behavior pipeline for both relational and non-relational events, we need to extract the before/after struct conversion logic into its own dedicated Single Message Transform (SMT).

Requirements for the new SMT:

- Relational Event Formatting: The SMT should focus solely on parsing the MongoDB sequence and converting it into a structured format where before and after are nested Struct objects. This will allow MongoDB events to seamlessly feed into the existing `ExtractChangedRecordState` SMT.

- Schema Mapping: The new SMT must allow users to define the final target schema. This will make it possible to normalize per-collection documents into a pre-defined schema (when configured) and effectively solve the missing field values problem.

div-dev123:
Fixes debezium/dbz#1715

## Description

MongoDB CDC events currently emit the `before` and `after` states as stringified JSON. This makes it difficult to use relational SMTs like `ExtractChangedRecordState` or the Outbox Event Router, which expect strongly-typed `Struct` values with a consistent schema.

This PR introduces a new SMT, `MongoToRelationalConverter`, which converts MongoDB envelopes into a relational-compatible format. Instead of flattening the document (like `ExtractNewDocumentState`), this transformation focuses only on converting the envelope payload from JSON into properly typed Kafka Connect `Struct`s, while preserving the standard Debezium envelope (`source`, `op`, `ts_ms`, `transaction`).

The goal is to allow MongoDB events to flow through the same downstream pipelines as relational connectors without additional custom handling.

### Changes

- **Pure envelope conversion**  
  Implements `Transformation<R>` directly and avoids extending existing flattening SMTs. This keeps the behavior focused and predictable.

- **Schema mapping support (`schema.mapping`)**  
  Allows users to define a fixed schema per topic using a JSON configuration. These mappings are parsed and compiled into Kafka Connect `Schema`s during initialization.

- **Dynamic schema inference**  
  If no mapping is provided, the SMT infers a unified schema by combining fields from both `before` and `after` documents.

- **Handling missing fields (`add.missing.fields`)**  
  Since relational SMTs expect identical schemas for `before` and `after`, the transformer ensures consistency by injecting `null` for any missing fields in either document.

### Testing

**Current coverage**

- Added `MongoToRelationalConverterTest.java`
- Verified heartbeat messages pass through unchanged (`shouldPassHeartbeatMessages`)
- Tested JSON → `Struct` conversion (`shouldConvertBsonStringsToStructs`), including:
  - preservation of envelope metadata (`op`, `ts_ms`, etc.)
  - correct handling of missing fields via null injection
- Confirmed that the existing MongoDB connector test suite runs without regressions

**Follow-ups**

- Add tests for `schema.mapping` to validate static schema parsing and application
- Add tests for `add.missing.fields=false` to ensure null injection can be disabled
- Cover edge cases:
  - empty `before` (insert events)
  - empty `after` (delete events)

## PR Checklist


debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/MongoToRelationalConverter.java
```diff
@@ -0,0 +1,335 @@
...
+
+/**
+ * Converts MongoDB CDC events to relational-style format where 'before' and 'after'
+ * fields are nested Struct objects instead of JSON strings. This enables MongoDB events
+ * to be processed by relational SMTs like ExtractChangedRecordState.
+ *
+ * @param <R> the subtype of {@link ConnectRecord} on which this transformation will operate
+ * @author Divyansh Agrawal
+ */
+public class MongoToRelationalConverter<R extends ConnectRecord<R>> implements Transformation<R>, Versioned {
```
mfvitale: Please implement also `io.debezium.metadata.ConfigDescriptor` interface.

div-dev123: Done @mfvitale thanks!

div-dev123:
@Naros @jpechane Could you take a look at this? Also I see that these CI failures are occurring on other PRs too so i am not sure they are related to my changes

rk3rn3r:
The MongoDB CI failures are fixed in Commit#2 of PR #7174 , should be merged tomorrow.

div-dev123:
> The MongoDB CI failures are fixed in Commit#2 of PR #7174 , should be merged tomorrow.

All right, thank you!

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/transforms/MongoToRelationalConverterTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+import io.debezium.connector.AbstractSourceInfo;
+import io.debezium.data.Envelope;
+
+/**
+ * Unit test for {@link MongoToRelationalConverter}.
+ *
+ * @author Divyansh Agrawal
+ */
+public class MongoToRelationalConverterTest {
```
Naros: There should be tests that cover other use cases for operations (`c`, `r`, and `d`) and different `capture.mode` values, etc.

Naros (changes requested):

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/transforms/MongoToRelationalConverterTest.java
div-dev123: Yeah the test coverage is a bit incomplete, will add the remaining too
Thanks!

div-dev123: I have added some of them, there are still 2-3 tests like `op=r`  and case where both after and before are populated on update left, will do it soon!

div-dev123: Also I can see that `ExtractNewDocumentState` has IT test, should those be added for this SMT too and if yes should it be a part of this PR or a new one?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/MongoToRelationalConverter.java
```diff
@@ -0,0 +1,337 @@
...
+
+/**
+ * Converts MongoDB CDC events to relational-style format where 'before' and 'after'
+ * fields are nested Struct objects instead of JSON strings. This enables MongoDB events
+ * to be processed by relational SMTs like ExtractChangedRecordState.
+ *
+ * @param <R> the subtype of {@link ConnectRecord} on which this transformation will operate
+ * @author Divyansh Agrawal
+ */
+public class MongoToRelationalConverter<R extends ConnectRecord<R>> implements Transformation<R>, Versioned, ConfigDescriptor {
```
jpechane: `Converter` has very specific meaning in Kafka Connect, I'd recommend to use a different term.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/MongoToRelationalConverter.java
```diff
@@ -0,0 +1,337 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(MongoToRelationalConverter.class);
+    private static final ObjectMapper MAPPER = new ObjectMapper();
+
+    // Configuration field for schema mapping
+    private static final Field SCHEMA_MAPPING = Field.create("schema.mapping")
```
jpechane: I'd prefer per collection option mapping like `schema.mapping.db.collection=` instead of everything in a single JSON.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/MongoToRelationalConverter.java
```diff
@@ -0,0 +1,337 @@
...
+                    JsonNode fieldsNode = collectionEntry.getValue();
+
+                    SchemaBuilder builder = SchemaBuilder.struct().optional().name(collectionName + ".envelope");
+
+                    // Translate the user's string type definitions into actual Connect Schema types
+                    fieldsNode.fields().forEachRemaining(fieldEntry -> {
+                        String fieldName = fieldEntry.getKey();
+                        String fieldType = fieldEntry.getValue().asText().toLowerCase();
+
+                        switch (fieldType) {
```
jpechane: Can we add logical name support like `io.debezium.time.Date`?

Also precise decimal numbers needs to be supported.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/MongoToRelationalConverter.java
div-dev123: would `MongoToRelationalStructMapper` be a fine alternative?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/MongoToRelationalConverter.java
div-dev123: Should we also extend it to other names like `io.debezium.time.Timestamp` , `io.debezium.time.MicroTimestamp` etc..?

jpechane: Yes, ideally it should support full set of debezium datatypes.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/MongoToRelationalConverter.java
div-dev123: @jpechane  Also Should i do that once all the changes are finalised? I think the commit history might be cleaner that way and changes might be more easily reviewable

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/MongoToRelationalConverter.java
div-dev123: done @jpechane

debezium-connector-mongodb/src/test/java/io/debezium/connector/mongodb/transforms/MongoToRelationalConverterTest.java
div-dev123: @Naros , I have added the asked tests , also I was thinking should i add tests to check if it converts nested documents properly `shouldConvertNestedDocuments` and also to check it converts arrays inside documents properly , `shouldConvertArrays`

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/MongoToRelationalConverter.java
div-dev123: I have skipped the `io.debezium.data.Bits` because its schema builder requires a mandatory parameter, `length` and it doesn't fit  into the `field:type` configuration string. We would need to change the parsing logic to handle something like `io.debezium.data.Bits(16)`. This adds a bit of complexity to the `parseFieldDefinitions` logic like handling parentheses, extracting the integer, etc.. I skipped it as its relatively rare but can add it if needed

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/MongoToRelationalConverter.java
Naros: Hi @div-dev123, what about just `MongoToRelationalMapper`? 

Also, I wouldn't be so concerned about the commit history on the PR. Making small changes like this in a single commit and in a PR with multiple commits is actually good practice during reviews like this, rather than squashing and rebasing.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/MongoToRelationalMapper.java
```diff
@@ -0,0 +1,398 @@
...
+        this.smtManager = new SmtManager<>(config);
+
+        addMissingFields = config.getBoolean(ADD_MISSING_FIELDS);
+
+        // Per-collection schema mapping strategy:
+        // Users define a strict target schema for each collection/topic using individual properties.
+        // Example: schema.mapping.myserver.mydb.orders=_id:string,total:float64,created:io.debezium.time.Date
+        for (Map.Entry<String, ?> entry : configs.entrySet()) {
+            if (entry.getKey().startsWith(SCHEMA_MAPPING_PREFIX)) {
+                String collectionName = entry.getKey().substring(SCHEMA_MAPPING_PREFIX.length());
```
Naros, line 96: We should probably guard that `collectionName` is not empty in this context, in case someone mistakenly adds the prefix without any collection name.

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/MongoToRelationalMapper.java
```diff
@@ -0,0 +1,398 @@
...
+
+        // Relational records expect 'before' and 'after' to share the same schema
+        Schema unifiedPayloadSchema = getOrInferPayloadSchema(record, beforeDoc, afterDoc);
+
+        // Convert the BSON documents into Kafka Connect Structs
+        Struct beforeStruct = convertToStruct(beforeDoc, unifiedPayloadSchema);
+        Struct afterStruct = convertToStruct(afterDoc, unifiedPayloadSchema);
+
+        // Create a new envelope schema that uses nested Structs instead of JSON strings
+        Schema envelopeSchema = buildEnvelopeSchema(unifiedPayloadSchema, value.schema());
```
Naros, line 257: I don't think we need to do this in this PR, but we should add a follow-up here to consider whether it makes sense for the envelope schema to be driven by a pluggable factory. There was a recent PR to make Debezium's default envelope customizable, and if we do that, it makes sense to apply that same pattern here.

In the meantime, it would be nice if there were a way to build off the existing SchemaFactory here, but I'm not sure whether we want to go that far. The other option would be in the SchemaFactory document: we should also align this mapper if the envelope schema ever changes.  @jpechane thoughts?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/MongoToRelationalConverter.java
div-dev123: All right @naros thanks!

div-dev123: changed the name thanks

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/transforms/MongoToRelationalMapper.java
div-dev123: done @Naros

Naros (approved):

minleejae:
Hi @Naros, it would be great to see this merged — it was approved back in April but has been idle since. What's the usual way forward in a case like this, where an approved PR just needs a refresh but the author seems inactive?

Naros:
Hi @minleejae in this case, start a new PR based on their work (so they get the initial attribution), and just note that the new PR supersedes this one because the prior one went idle/stale.

Naros:
Superseded by https://github.com/debezium/debezium/pull/8008.

PR #7272, opened 2026-04-02
debezium/dbz#1605 Auto-enable heartbeat when lsn.flush.mode=connector_and_driver

### LINKED ISSUE #1605: flush_lsn may not advance with lsn.flush.mode=connector_and_driver when only unmonitored tables are updated
## Bug report

When `lsn.flush.mode` is set to `connector_and_driver` and `provide.transaction.metadata` is set to `false` (default), I observed a case where flush_lsn in pg_stat_replication and confirmed_flush_lsn in pg_replication_slots do not advance even when there is ongoing WAL-generating activity.

PostgreSQL connector 3.4.1

---

```json
{
  "name": "test-connector",
  "config": {
    "database.hostname": "localhost",
    "database.port": "15432",
    "database.user": "atorik",
    "database.password": "xxx",
    "topic.prefix": "example",
    "slot.name": "example_slot",
    "plugin.name": "pgoutput",
    "publication.name": "p1",
    "lsn.flush.mode":"connector_and_driver"
  }
}
```

Note that when `provide.transaction.metadata`: `true` was added to the configuration, this problem does NOT occur.

---

on-premises

---

Even if WAL activity occurs on tables that are not part of the publication (i.e., not captured by CDC), both:

- flush_lsn in pg_stat_replication, and
- confirmed_flush_lsn in pg_replication_slots

should continue to advance.

---

**What behavior do you see?**

Even when WAL activity occurs on tables that are not part of the publication, both:

- flush_lsn in pg_stat_replication, and
- confirmed_flush_lsn in pg_replication_slots

remain unchanged and do not advance.

---

Yes.

---

Yes. I believe the same logs can be obtained by following the reproduction steps below. Please let me know if you need the actual log files.

---

The issue can be reproduced by:

- Updating a table that is part of the publication, and then
- Generating WAL activity on tables that are not part of the publication.

If only non-publication tables are updated (without first updating a publication table), the issue does NOT occur.

In the steps below:

- `t1` is included in the publication.
- Non-publication WAL activity is generated via pgbench (e.g., updates to pgbench_accounts).

Create `t1` table and the publication:

```
=# create table t1(i int primary key, t text);
=# create publication p1 for table t1;
```

Initialize pgbench tables:

```
$ pgbench postgres -h localhost -U postgres -i
```

Connector configuration:

```
{
  "name": "test-connector",
  "config": {
    "database.port": "5432",
    "database.user": "postgres",
    "topic.prefix": "example",
    "slot.name": "example_slot",
    "plugin.name": "pgoutput",
    "publication.name": "p1",
    "lsn.flush.mode": "connector_and_driver"
  }
}
```

Register the connector:

```
$ curl -i -X POST \
  -H "Accept:application/json" \
  -H "Content-Type:application/json" \
  http://localhost:8083/connectors/ \
  -d @docker-non-proceed-lsn.json
```

Update a table included in the publication:

```
=# insert into t1 values (1, 'aaa');
```

At this point, flush_lsn and confirmed_flush_lsn are 0/37DABB0:

```
=# select now(), sent_lsn, write_lsn, flush_lsn, replay_lsn from pg_stat_replication;

              now              | sent_lsn  | write_lsn | flush_lsn | replay_lsn
-------------------------------+-----------+-----------+-----------+------------
 2026-02-12 07:25:30.583864+00 | 0/37DACF8 | 0/37DACF8 | 0/37DABB0 | 0/37DABB0


=# select now(), confirmed_flush_lsn from pg_replication_slots;

              now              | confirmed_flush_lsn
-------------------------------+---------------------
 2026-02-12 07:25:23.319627+00 | 0/37DABB0
```

Generate activity on non-publication tables:

```
$ pgbench -d postgres -T 30 -U postgres
```

At this point, both flush_lsn and confirmed_flush_lsn remain unchanged:

```
              now              | confirmed_flush_lsn
-------------------------------+---------------------
 2026-02-12 07:29:58.925503+00 | 0/37DABB0
(1 row)

              now              | confirmed_flush_lsn
-------------------------------+---------------------
 2026-02-12 07:30:45.896185+00 | 0/37DABB0
 ```

---
**Cause Analysis**

The direct cause appears to be the condition for updating `lastFlushedLSN` in the keepalive message handling of the JDBC driver:

```java
  private boolean processKeepAliveMessage(ByteBuffer buffer) {

    // if the client has confirmed flush of last XLogData msg and KeepAlive shows ServerLSN is still
    // advancing, we can safely advance FlushLSN to ServerLSN
    if (automaticFlush && explicitlyFlushedLSN.asLong() >= startOfLastMessageLSN.asLong()
        && lastServerLSN.asLong() > explicitlyFlushedLSN.asLong()
        && lastServerLSN.asLong() > lastFlushedLSN.asLong()) {
      lastFlushedLSN = lastServerLSN;
```

When the issue occurs, debugging shows something like:

```
explicitlyFlushedLSN: 97663152 (5D238B0)
startOfLastMessageLSN: 97663328 (5D23960)
```

The condition `explicitlyFlushedLSN.asLong() >= startOfLastMessageLSN.asLong()` is not satisfied.

This might be wrong, but from my investigation:

- The JDBC driver updates `startOfLastMessageLSN` even for transaction messages such as COMMIT
(see [processXLogData()](https://github.com/atorik/pgjdbc/blob/c07721af972eee4f10873b1a23b3811336454436/pgjdbc/src/main/java/org/postgresql/core/v3/replication/V3PGReplicationStream.java#L267))
- On the other hand, `explicitlyFlushedLSN` is derived from commitOffset(), which uses
PostgresOffsetContext.LAST_COMPLETELY_PROCESSED_LSN_KEY.

When `provide.transaction.metadata` = `false`, transaction metadata events are not emitted, and therefore the LSN used to update `explicitlyFlushedLSN` does not advance in certain scenarios.

---
**Possible approaches?**

One idea I considered was modifying commitOffset() to use effectiveOffset.lastCompletelyProcessedLsn() instead of PostgresOffsetContext.LAST_COMPLETELY_PROCESSED_LSN_KEY.

However, this would acknowledge LSNs to PostgreSQL that may not yet have been fully transferred to Kafka, potentially causing data loss. Therefore, I believe this approach is unsafe.

At the moment, the only workaround I can think of is:

When `lsn.flush.mode` = `connector_and_driver`, require `provide.transaction.metadata` = `true`, and fail fast otherwise. If this direction is acceptable, I am willing to prepare a PR.

What do you think?

atorik:
Fixes debezium/dbz#1605

## Description
As discussed in #1605, this PR enables automatic heartbeat when conditions are met.

When `lsn.flush.mode` is set to `connector_and_driver` and `heartbeat.interval.ms` is left at its default value (0),
the connector may never get a chance to propagate LSNs to Kafka if only unmonitored tables generate WAL activity.

In this situation, `confirmed_flush_lsn` does not advance and replication slots retain WAL indefinitely.

To provide periodic opportunities for LSN propagation without requiring users to be aware of this subtle interaction, the connector now internally sets `heartbeat.interval.ms` to 10 minuntes when:

  - lsn.flush.mode = connector_and_driver
  - heartbeat.interval.ms = 0
  - provide.transaction.metadata = false

A warning is logged to make this behavior explicit.

## PR Checklist

### DIFF
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresConnectorIT.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresConnectorIT.java
@@ -2977,6 +2977,16 @@ public class PostgresConnectorIT extends AbstractAsyncEngineConnectorTest {
     }
 
     @Test
+    @FixFor("DBZ-1605")
+    void shouldApplyLsnFlushModeHeartbeatFallbackWhenNoOpportunityForFlush() {
+        PostgresConnectorConfig config = new PostgresConnectorConfig(TestHelper.defaultConfig()
+                .with(PostgresConnectorConfig.LSN_FLUSH_MODE, PostgresConnectorConfig.LsnFlushMode.CONNECTOR_AND_DRIVER.getValue())
+                .build());
+
+        assertThat(config.getHeartbeatInterval()).isGreaterThan(java.time.Duration.ZERO);
+    }
+
+    @Test
     @FixFor("DBZ-1292")
     @SkipWhenKafkaVersion(check = EqualityCheck.EQUAL, value = KafkaVersion.KAFKA_1XX, description = "Not compatible with Kafka 1.x")
     public void shouldOutputRecordsInCloudEventsFormat() throws Exception {
--- a/debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
+++ b/debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
@@ -1525,7 +1525,7 @@ public abstract class CommonConnectorConfig {
     private final Duration pollInterval;
     protected final String logicalName;
     private final String heartbeatTopicsPrefix;
-    private final Duration heartbeatInterval;
+    private Duration heartbeatInterval;
     private final Duration snapshotDelay;
     private final Duration streamingDelay;
     private final Duration retriableRestartWait;
@@ -1725,6 +1725,10 @@ public abstract class CommonConnectorConfig {
 
     public Duration getHeartbeatInterval() {
         return heartbeatInterval;
+    }
+
+    protected void setHeartbeatInterval(int heartbeatIntervalMs) {
+        this.heartbeatInterval = Duration.ofMillis(heartbeatIntervalMs);
     }
 
     public Duration getSnapshotDelay() {
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
@@ -33,6 +33,7 @@ import io.debezium.connector.postgresql.connection.PostgresConnection;
 import io.debezium.connector.postgresql.connection.ReplicationConnection;
 import io.debezium.connector.postgresql.connection.pgoutput.PgOutputMessageDecoder;
 import io.debezium.connector.postgresql.connection.pgproto.PgProtoMessageDecoder;
+import io.debezium.heartbeat.Heartbeat;
 import io.debezium.jdbc.JdbcConfiguration;
 import io.debezium.relational.ColumnFilterMode;
 import io.debezium.relational.RelationalDatabaseConnectorConfig;
@@ -1410,6 +1411,8 @@ public class PostgresConnectorConfig extends RelationalDatabaseConnectorConfig {
         this.publishViaPartitionRoot = config.getBoolean(PUBLISH_VIA_PARTITION_ROOT);
         this.lsnFlushTimeoutAction = LsnFlushTimeoutAction.parse(config.getString(LSN_FLUSH_TIMEOUT_ACTION));
         this.offsetSlotMismatchStrategy = resolveOffsetSlotMismatchStrategy(config);
+
+        applyLsnFlushModeHeartbeatFallback(config);
     }
 
     protected String hostname() {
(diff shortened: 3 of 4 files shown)


debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresConnectorIT.java
```diff
@@ -2976,6 +2976,16 @@ public void shouldFlushLsnOfUnmonitoredActivityInConnectorAndDriverMode() throws
...
 
+    @Test
+    @FixFor("DBZ-1605")
+    void shouldApplyLsnFlushModeHeartbeatFallbackWhenNoOpportunityForFlush() {
+        PostgresConnectorConfig config = new PostgresConnectorConfig(TestHelper.defaultConfig()
+                .with(PostgresConnectorConfig.LSN_FLUSH_MODE, PostgresConnectorConfig.LsnFlushMode.CONNECTOR_AND_DRIVER.getValue())
+                .build());
+
+        assertThat(config.getHeartbeatInterval()).isGreaterThan(java.time.Duration.ZERO);
+    }
```
Naros, line 2987: While this is a great unit test, it's insufficient to verify that a heartbeat event is emitted in this use case and that the WAL shrinks as expected. If I am not mistaken, for the WAL-to-shrink use case you mentioned with a non-captured table, `heartbeat.action.query` would be necessary, right?

atorik: Thank you for the quick review.

Hmm, my understanding is that `heartbeat.action.query` is not required here.

When `lsn.flush.mode` is set to `connector_and_driver`, the connector-side LSN is advanced by PostgreSQL KeepAlive messages. Because of this, the LSN recognized by the connector progresses even without `heartbeat.action.query`.

In fact, after applying this PR, I followed the steps described in the issue and confirmed that both `pg_stat_replication.flush_lsn` and `pg_replication_slots.confirmed_flush_lsn` advance as expected.

atorik:
Added documentation for this behavior.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresConnectorIT.java
Naros: What confused me was this in your original description:
> In this situation, confirmed_flush_lsn does not advance and replication slots retain WAL indefinitely.

But you're right, in this context events like TX begin/commit will trigger the rolling forward of the LSN data in the offsets, and so the periodic heartbeats will adderss updating the Kafka offset metadata. :heavy_check_mark:

Naros:
Hi @atorik can you rebase this PR on `main` to fix the conflicts?  Thanks.

atorik:
Thanks for pointing this out!
Fixed the conflict.

jpechane:
@atorik LGTM but I re-triggered the failed jobs as the PostgreSQL/pgouput failure can be related.

atorik:
> LGTM but I re-triggered the failed jobs as the PostgreSQL/pgouput failure can be related.

As far as I can tell from the CI results, the failures are limited to the Db2 and Spanner jobs.
I don’t see evidence that the PostgreSQL/pgoutput failure is related to those.
Is that understanding correct? > @jpechane

jpechane:
Yes, the repeated build has passed so it was probably just on-off issue.

jpechane (approved):
@atorik LGTM, thanks!
The PR can be merged after Alpha1 release is finished.

vjuranek (approved):

vjuranek:
Merged, thanks @atorik !

PR #7286, opened 2026-04-03
debezium/dbz#453 Derive field group order implicitly from ConfigDefinition ordering

### LINKED ISSUE #453: Order Field definitions in Config classes according to the Field.Group and the position in that group [DBZ-3447]
Migrated from [DBZ-3447](https://issues.redhat.com/browse/DBZ-3447)

Order Field definitions in Config classes according to the Field.Group and the position in that group

Binayak490-cyber:
## Summary

Fixes debezium/dbz#453

This PR repurposes the original issue (which asked to reorder `Field` declarations to match explicit `positionInGroup` numbers) into a cleaner solution: remove manual position numbers entirely and derive `groupOrder` implicitly from the order fields appear in `ConfigDefinition`.

### Changes

- **`DebeziumDescriptorSchemaCreator`**: Added `buildFieldGroupOrderMap()` which iterates `ConfigDefinition.all()` and assigns sequential position per `Field.Group` based on list order — following the same implicit pattern used in `Field#group()`. No more reading explicit `positionInGroup` numbers from field declarations.

- **All connector config classes** (MySQL, PostgreSQL, MongoDB, Oracle, SQL Server, MariaDB, Binlog, JDBC, Sink, Kafka/RocketMQ storage): 
  Removed explicit position numbers from all `createGroupEntry(Field.Group.XXX, N)` calls — replaced with `createGroupEntry(Field.Group.XXX)`.

- **`AbstractFieldTest`**: Removed `shouldNotAllowDuplicateGroupAssignment` test which is now a no-op since no explicit positions exist to conflict.

### Note

The concrete `FieldTest` subclasses (Postgres, Oracle, SQL Server, MongoDB, MySQL, MariaDB) now run 0 tests as they only set up fields for the removed test. These can be cleaned up in a follow-up.


Binayak490-cyber:
Hi @Naros, @mfvitale, @jpechane - opened a PR for this - https://github.com/debezium/dbz/issues/453

Removed all explicit positionInGroup numbers from connector configs and updated DebeziumDescriptorSchemaCreator to derive groupOrder implicitly from ConfigDefinition.all() per Field.Group, following the same pattern as Field#group(). All schema generator tests pass. Happy to get feedback!

debezium-schema-generator/src/main/java/io/debezium/schemagenerator/schema/debezium/DebeziumDescriptorSchemaCreator.java
```diff
@@ -75,14 +81,33 @@ public ComponentDescriptor buildDescriptor() {
                 buildGroups(usedGroups));
     }
 
-    private Property buildProperty(Field field) {
+    /**
+     * Builds a map of field name to its implicit order within its {@link Field.Group},
+     * based on the order fields appear in the {@link ConfigDefinition}.
+     * This avoids the need for explicit position numbers on each field declaration.
+     */
+    private Map<String, Integer> buildFieldGroupOrderMap(ConfigDefinition configDefinition) {
```
mfvitale: We are effectively doing this calculation in two places. 

We already do it now in https://github.com/debezium/debezium/blob/7610a1553ebe7d2e0c9af9aee82e292c89114ff9/debezium-config/src/main/java/io/debezium/config/Field.java#L464. 

Could you evaluate it we could do this just one time in the [io.debezium.config.ConfigDefinitionEditor#create](https://github.com/debezium/debezium/blob/7610a1553ebe7d2e0c9af9aee82e292c89114ff9/debezium-config/src/main/java/io/debezium/config/ConfigDefinitionEditor.java#L72)?

So effectively we reprocess the fields adding the calculated group order?

mfvitale:
@Binayak490-cyber We are in the good path. I left a comment on a possible optimization. 
Could you also restore the abstract test so that we can effectively see if we are not creating any duplicates?

Naros (commented):
Hi @Binayak490-cyber, to add to what Mario has requested, we should also check across all the other non-main repo connectors like cassandra, db2, ibmi, informix, cockroachdb, and ingres; opening PRs in those to update the `Field.createGroupEntry`.

Binayak490-cyber:
> Hi @Binayak490-cyber, to add to what Mario has requested, we should also check across all the other non-main repo connectors like cassandra, db2, ibmi, informix, cockroachdb, and ingres; opening PRs in those to update the `Field.createGroupEntry`.

Yes @Naros, I know that other repos needs to be updated with the same changes, its a follow-up task, not part of this PR which i will go through for other non-main repos also which I mentioned too in the PR description.

debezium-schema-generator/src/main/java/io/debezium/schemagenerator/schema/debezium/DebeziumDescriptorSchemaCreator.java
Binayak490-cyber: Sure @mfvitale! I'll move the group order calculation into ConfigDefinitionEditor#create so it's computed once at definition time and available directly from the ConfigDefinition, rather than being recalculated in DebeziumDescriptorSchemaCreator. Will update the PR shortly.

Binayak490-cyber: @mfvitale, addressed both points:

1. Moved buildFieldGroupOrderMap() into ConfigDefinition constructor — computed once, exposed via fieldGroupOrderMap() getter. DebeziumDescriptorSchemaCreator now delegates to it.
2. Restored AbstractFieldTest with updated shouldNotHaveDuplicateGroupOrders() using counter-based implicit ordering.

Point 3 (external repos) of @Naros will follow as separate PRs once this is merged and will be done accordingly.

Binayak490-cyber:
@mfvitale, @Naros, updated the PR according to the latest review...anything else needs to be changed or updated...?

debezium-config/src/main/java/io/debezium/config/ConfigDefinition.java
```diff
@@ -37,6 +41,33 @@ public class ConfigDefinition {
...
         this.events = Collections.unmodifiableList(events);
+        this.fieldGroupOrderMap = Collections.unmodifiableMap(buildFieldGroupOrderMap());
+    }
+
+    /**
+     * Builds a map of field name to its implicit order within its {@link Field.Group},
+     * based on the order fields appear in this {@link ConfigDefinition}.
+     * This avoids the need for explicit position numbers on each field declaration.
+     */
+    private Map<String, Integer> buildFieldGroupOrderMap() {
```
mfvitale: @Binayak490-cyber Any reason to not do it in the `io.debezium.config.ConfigDefinitionEditor#type` method (and others)?

I was thinking to something like 

```java
public ConfigDefinitionEditor type(Field... fields) {
        type.addAll(IntStream.range(0, fields.length)
                .mapToObj(i -> fields[i].withGroup(Field.createGroupEntry(fields[i].group().getGroup(), i)))
                .toList());
        return this;
    }
```

so you effectively do it by group and it simplify the logic and then we can reuse the `io.debezium.config.Field.GroupEntry#getPositionInGroup`. 

WDYT?

Binayak490-cyber: @mfvitale, that's a cleaner approach! My initial thinking was to keep the ordering logic centralized in ConfigDefinition so that any consumer (not just DebeziumDescriptorSchemaCreator) could reuse the map without recomputing it. But I see your point - assigning the position directly into the Field at insertion time in ConfigDefinitionEditor is simpler, avoids the extra map, and reuses the existing getPositionInGroup() infrastructure. If so then I'll go ahead with your approach. So according to this approach want to confirm that should the position be 0-based (as in your snippet with i) or 1-based (like i+1)?

Binayak490-cyber: Any update to this @mfvitale, @jpechane...?

mfvitale: @Binayak490-cyber it should be 0-based

Binayak490-cyber: @mfvitale, is my reasoning correct as I have given it according to my perspective, so wanted to ask that is it so or I will revert the changes and make it according to what you said...? Need your guidance as many other non-main repos needs to be changed and addressed since it is a big refactor and clean-up.

Binayak490-cyber:
@mfvitale, implemented your approach - position is now assigned directly in `ConfigDefinitionEditor#type()` (and `connector()`, `history()`, `events()`) at insertion time using 0-based index `i`. `DebeziumDescriptorSchemaCreator` goes back to using `field.group().getPositionInGroup()` directly. `ConfigDefinition` is back to its clean state with no extra map. Can you look at it now...?

debezium-schema-generator/src/main/java/io/debezium/schemagenerator/schema/debezium/DebeziumDescriptorSchemaCreator.java
```diff
@@ -56,10 +57,12 @@ public ComponentDescriptor buildDescriptor() {
                 componentMetadata.getComponentDescriptor().getDisplayName(),
                 null);
 
+        ConfigDefinition configDefinition = componentMetadata.getConfigDefinition();
+
         List<Property> properties = new ArrayList<>();
         Set<String> usedGroups = new LinkedHashSet<>();
-        StreamSupport.stream(componentMetadata.getConfigDefinition().all().spliterator(), false)
-                .map(this::buildProperty)
+        StreamSupport.stream(configDefinition.all().spliterator(), false)
```
mfvitale: You can rollback these changes now.

Binayak490-cyber: @mfvitale, rolled back the changes in DebeziumDescriptorSchemaCreator - it's now back to its original form. Thanks for the guidance!

mfvitale:
@Naros Do you recall if the current orders as a real meaning (maybe done with the old UI) or is just something casual?

I'm asking because with this refactoring we will not keep the current orders, but the order will be guided by how the configurations appears in the `ConfigDefinition`.

Naros:
@mfvitale, the old "OG" configuration field ordinal positions did have some meaning for the old UI. When we initially set those years ago, we did so with the legacy UI in mind.  Now, as new configuration properties were added, if they were assigned an ordinal, I at least shifted things around to keep like properties grouped together, but I don't think we genuinely followed that principle universally.

With this PR's work, I would genuinely expect that if there is any order that must be maintained or adjusted, we'd just apply that to the `CONFIG_DEFINITION` as needed. It might be useful to run a Platform before/after across the connector configuration forms to identify any gaps or misalignments introduced by this PR. wdyt?

mfvitale:
> With this PR's work, I would genuinely expect that if there is any order that must be maintained or adjusted, we'd just apply that to the `CONFIG_DEFINITION` as needed. It might be useful to run a Platform before/after across the connector configuration forms to identify any gaps or misalignments introduced by this PR. wdyt?

Well I think that the order and maybe also the group association should be revisited. Therefore I just opened an [issue](https://github.com/debezium/dbz/issues/1821) for the `TOPIC_PREFIX`. 

So I think that we have two option: go with the current order and then fix things accordingly or try to keep the old order re arranging in the  `CONFIG_DEFINITION`. 

The choice should take in consideration that we should in any case do some organization and reorder work, so if keeping the current order is not a pain we should aim of it in this PR otherwise let's keep it as is and then we can do the re-ordering. @Naros WDYT?

@Binayak490-cyber How difficult you think will be to keep the current configuration order?

Binayak490-cyber:
@mfvitale, @Naros, Keeping the current order should be I think **very straightforward**.

After checking the codebase, no field in the main repo uses an explicit `positionInGroup` value - every `createGroupEntry` call uses the no-argument overload which defaults to `9999`. This means the current ordering is already entirely implicit, driven purely by the sequence fields appear in each `ConfigDefinitionEditor` call (`type()`, `connector()`, `history()`, `events()`).

So @mfvitale's proposed approach of assigning `i` at insertion time in `ConfigDefinitionEditor` will naturally preserve the existing order - it just makes it explicit. No reordering work would be needed in this PR.

Is that so and should I go ahead with the change @mfvitale, @Naros...?

mfvitale:
> So @mfvitale's proposed approach of assigning `i` at insertion time in `ConfigDefinitionEditor` will naturally preserve the existing order - it just makes it explicit.

@Binayak490-cyber Yes, this is the current order, implicit from the order in the ConfigDefinitionEditor, but this order **could** be different from the the old order expressed within the `Field` definition. 

This is why I asked

> How difficult you think will be to keep the current configuration order?

Binayak490-cyber:
> > So @mfvitale's proposed approach of assigning `i` at insertion time in `ConfigDefinitionEditor` will naturally preserve the existing order - it just makes it explicit.
> 
> @Binayak490-cyber Yes, this is the current order, implicit from the order in the ConfigDefinitionEditor, but this order **could** be different from the the old order expressed within the `Field` definition.
> 
> This is why I asked
> 
> > How difficult you think will be to keep the current configuration order?

@mfvitale, After looking at the actual `CONFIG_DEFINITION` chains (e.g. `BinlogConnectorConfig`, `RelationalDatabaseConnectorConfig`), fields from multiple groups are mixed within a single `connector()` or `events()` call. Assigning `i` at insertion time means two fields from the same group that are non-consecutive in the list will get non-consecutive `positionInGroup` values - but that's exactly what the current implicit ordering already produces.

So keeping the current order is I think **not difficult at all** - the `ConfigDefinitionEditor` approach with `i` naturally preserves it. No manual reordering of any `CONFIG_DEFINITION` is needed.

Naros:
> The choice should take in consideration that we should in any case do some organization and reorder work, so if keeping the current order is not a pain we should aim of it in this PR otherwise let's keep it as is and then we can do the re-ordering. @Naros WDYT?

@mfvitale, my opinion is to keep this PR as is, knowing we need to revisit the field order/groups anyway. Then we can do all that massaging and shuffling as needed in the follow-up, since those changes are very similar.

mfvitale (approved):

mfvitale:
@Binayak490-cyber I took another pass and I found that the current solution is not going to work (sorry was my suggestion). 

The problem is that we define the group on each field and the we add them in the `ConfigDefinition` in some groups that are not related. 

For example, if we take from `io.debezium.relational.RelationalDatabaseConnectorConfig`

```java
protected static final ConfigDefinition CONFIG_DEFINITION = CommonConnectorConfig.CONFIG_DEFINITION.edit()
            .type(
                    CommonConnectorConfig.TOPIC_PREFIX)
            .connector(
                    DECIMAL_HANDLING_MODE,
                    TIME_PRECISION_MODE,
                    SNAPSHOT_LOCK_TIMEOUT_MS)
            .events(
                    COLUMN_INCLUDE_LIST,
                    COLUMN_EXCLUDE_LIST,
                    TABLE_INCLUDE_LIST,
                    TABLE_EXCLUDE_LIST,
                    TABLE_IGNORE_BUILTIN,
                    SCHEMA_INCLUDE_LIST,
                    SCHEMA_EXCLUDE_LIST,
                    MSG_KEY_COLUMNS,
                    SNAPSHOT_SELECT_STATEMENT_OVERRIDES_BY_TABLE,
                    MASK_COLUMN_WITH_HASH,
                    MASK_COLUMN,
                    TRUNCATE_COLUMN,
                    INCLUDE_SCHEMA_CHANGES,
                    INCLUDE_SCHEMA_COMMENTS,
                    PROPAGATE_COLUMN_SOURCE_TYPE,
                    PROPAGATE_DATATYPE_SOURCE_TYPE,
                    SNAPSHOT_FULL_COLUMN_SCAN_FORCE,
                    SNAPSHOT_TABLES_ORDER_BY_ROW_COUNT,
                    DatabaseHeartbeatImpl.HEARTBEAT_ACTION_QUERY)
            .create();
```

we have this `events` grouping and if you take for example the `TABLE_INCLUDE_LIST` property it has the `FILTERS` group but event worst the `MSG_KEY_COLUMNS` is in `CONNECTOR_ADVANCED` group!

@Naros So we should decide where we want to define the group, in the field or in the `ConfigDefinition`. 

@Binayak490-cyber For now I think that the quick solution is to go back to your old implementation where you get the groups looking to all fields.

mfvitale (changes requested):
See my comment

Binayak490-cyber:
> @Binayak490-cyber I took another pass and I found that the current solution is not going to work (sorry was my suggestion).
> 
> The problem is that we define the group on each field and the we add them in the `ConfigDefinition` in some groups that are not related.
> 
> For example, if we take from `io.debezium.relational.RelationalDatabaseConnectorConfig`
> 
> ```java
> protected static final ConfigDefinition CONFIG_DEFINITION = CommonConnectorConfig.CONFIG_DEFINITION.edit()
>             .type(
>                     CommonConnectorConfig.TOPIC_PREFIX)
>             .connector(
>                     DECIMAL_HANDLING_MODE,
>                     TIME_PRECISION_MODE,
>                     SNAPSHOT_LOCK_TIMEOUT_MS)
>             .events(
>                     COLUMN_INCLUDE_LIST,
>                     COLUMN_EXCLUDE_LIST,
>                     TABLE_INCLUDE_LIST,
>                     TABLE_EXCLUDE_LIST,
>                     TABLE_IGNORE_BUILTIN,
>                     SCHEMA_INCLUDE_LIST,
>                     SCHEMA_EXCLUDE_LIST,
>                     MSG_KEY_COLUMNS,
>                     SNAPSHOT_SELECT_STATEMENT_OVERRIDES_BY_TABLE,
>                     MASK_COLUMN_WITH_HASH,
>                     MASK_COLUMN,
>                     TRUNCATE_COLUMN,
>                     INCLUDE_SCHEMA_CHANGES,
>                     INCLUDE_SCHEMA_COMMENTS,
>                     PROPAGATE_COLUMN_SOURCE_TYPE,
>                     PROPAGATE_DATATYPE_SOURCE_TYPE,
>                     SNAPSHOT_FULL_COLUMN_SCAN_FORCE,
>                     SNAPSHOT_TABLES_ORDER_BY_ROW_COUNT,
>                     DatabaseHeartbeatImpl.HEARTBEAT_ACTION_QUERY)
>             .create();
> ```
> 
> we have this `events` grouping and if you take for example the `TABLE_INCLUDE_LIST` property it has the `FILTERS` group but event worst the `MSG_KEY_COLUMNS` is in `CONNECTOR_ADVANCED` group!
> 
> @Naros So we should decide where we want to define the group, in the field or in the `ConfigDefinition`.
> 
> @Binayak490-cyber For now I think that the quick solution is to go back to your old implementation where you get the groups looking to all fields.

@mfvitale, Done - reverted `ConfigDefinitionEditor` back to simple field addition, moved `buildFieldGroupOrderMap()` into `ConfigDefinition` (0-based ordering), and `DebeziumDescriptorSchemaCreator` now delegates to `configDefinition.fieldGroupOrderMap()`. All tests passing.

Also, no worries at all! This is actually exactly what I had originally thought and implemented - the map computation iterating over all fields and counting per `Field.Group`. The reason I had it that way from the start was precisely because the `ConfigDefinitionEditor` buckets (`type`, `connector`, `history`, `events`) mix fields from multiple `Field.Group` values, so position assignment there wouldn't be meaningful per group. We've now come back to that approach but with the logic correctly placed in `ConfigDefinition` rather than in `DebeziumDescriptorSchemaCreator`. Can you please take a look at it now @mfvitale, @Naros!

Naros:
> So we should decide where we want to define the group, in the field or in the ConfigDefinition.

@mfvitale I think there are pros/cons to either approach. 

At the `Field` level, we keep all the information about a field together, but this makes it much harder to reason order/grouping across a wide array of fields.  At the `ConfigDefinition` level, we have a lot less noise, making it easier to reason about order/grouping.

The only thing that gives me pause with moving to `ConfigDefinition` is the remaining UI-centric declarations at the `Field` level, e.g., width and importance. I think these should ideally remain together somehow. Would you concur?

mfvitale:
> > So we should decide where we want to define the group, in the field or in the ConfigDefinition.
> 
> @mfvitale I think there are pros/cons to either approach.
> 
> At the `Field` level, we keep all the information about a field together, but this makes it much harder to reason order/grouping across a wide array of fields. At the `ConfigDefinition` level, we have a lot less noise, making it easier to reason about order/grouping.

I agree that with just `Field` it will be difficult to figure out the ordering and the grouping.

 
> The only thing that gives me pause with moving to `ConfigDefinition` is the remaining UI-centric declarations at the `Field` level, e.g., width and importance. I think these should ideally remain together somehow. Would you concur?

`ConfigDefinition` seems to a be a better place where grouping things together since we also use in the connector config for example with ```java public static Field.Set ALL_FIELDS = Field.setOf(CONFIG_DEFINITION.all());``` and this is also the one used by the component descriptor. 

I could see that in long term we can remove the group associated with the property, then in the `ConfigDefinition` we just need to add the `Field` to the desired group and the order in which the it appears will be the final group order.

Binayak490-cyber:
@mfvitale, @Naros, wanted to know that this PR has been open for about 3 weeks now. The current implementation looks stable and tests are passing. Let me know if anything needs to be changed in this PR. The long-term direction you're discussing (removing group from Field entirely) seems like a natural follow-up rather than a blocker for this PR which I think is also made by @mfvitale that is this https://github.com/debezium/dbz/issues/1821.

Also, the six non-main repo PRs (cassandra, db2, ibmi, informix, cockroachdb, ingres) are waiting on this to merge first before I can open those. Would be great to get a decision on whether the current state is good enough to merge or changes are required in this PR so we can unblock that work.

mfvitale:
@Binayak490-cyber Sorry for taking this too long but with the current implementation, if we want to change a field order it is more difficult compared with the past. 

I added a `needs-more-thinking` label since I need to think better what we can do at this point. I'll take a look to it next week. Thanks for the patience.

Naros:
@Binayak490-cyber, one thing to note, the sibling repos, if you want to work on code changes that align with this PR's current state, that is doable, and they'll build as long as the branches in the other repos are named exactly the same as the branch name you used for this PR, `fix/dbz-453-implicit-group-order`. 

The only concern I would have with proceeding, and you would be the best judge of the effort, is how much time it would take to tweak those PRs based on our recent discussion, or how much throwaway that might be. The choice is ultimately up to you, but I wouldn't want to you to do a lot of work to later effectively start over.

Binayak490-cyber:
> @Binayak490-cyber Sorry for taking this too long but with the current implementation, if we want to change a field order it is more difficult compared with the past.
> 
> I added a `needs-more-thinking` label since I need to think better what we can do at this point. I'll take a look to it next week. Thanks for the patience.

@mfvitale, No worries, take your time and I wanted that clarity only that it may take time - thanks for the transparency on the label.

Binayak490-cyber:
> @Binayak490-cyber, one thing to note, the sibling repos, if you want to work on code changes that align with this PR's current state, that is doable, and they'll build as long as the branches in the other repos are named exactly the same as the branch name you used for this PR, `fix/dbz-453-implicit-group-order`.
> 
> The only concern I would have with proceeding, and you would be the best judge of the effort, is how much time it would take to tweak those PRs based on our recent discussion, or how much throwaway that might be. The choice is ultimately up to you, but I wouldn't want to you to do a lot of work to later effectively start over.

@Naros, Thanks for the heads-up on the sibling repos. I think as @mfvitale is still thinking through the approach, I'll hold off on those for now to avoid potentially throwaway work which you exactly told right and therefore I was not going up for with the sibling repos as it may take a full rework if design changes. I'll wait for the direction to settle on this PR first before proceeding with the others.

mfvitale:
@Naros 

I could image something like 

```java
ConfigDefinition CONFIG_DEFINITION = ConfigDefinition.editor()
      .name("Postgres")
      .group(Field.Group.CONNECTION,
          HOSTNAME,      // auto-assigned position 0
          PORT,          // auto-assigned position 1
          DATABASE)      // auto-assigned position 2
      .group(Field.Group.CONNECTION_ADVANCED_REPLICATION,
          SLOT_NAME,     // auto-assigned position 0
          PUBLICATION)   // auto-assigned position 1
      .group(Field.Group.CONNECTOR_SNAPSHOT,
          SNAPSHOT_MODE) // auto-assigned position 0
      .create();
```

So this will permit to easily describe the groups and the fields that affect to them. Order is then auto calculated based on position in the list. In addition, `ConfigDefinition` will be more generic in this way to also be used to describe not only connector configuration. 

This will also solve the fact that currently we just exposes to KC the wrong groups and not effectively the Field ones that are the correct. 

@jpechane WDYT?

jpechane:
@mfvitale 

> @Naros
> 
> I could image something like
> 
> ```java
> ConfigDefinition CONFIG_DEFINITION = ConfigDefinition.editor()
>       .name("Postgres")
>       .group(Field.Group.CONNECTION,
>           HOSTNAME,      // auto-assigned position 0
>           PORT,          // auto-assigned position 1
>           DATABASE)      // auto-assigned position 2
>       .group(Field.Group.CONNECTION_ADVANCED_REPLICATION,
>           SLOT_NAME,     // auto-assigned position 0
>           PUBLICATION)   // auto-assigned position 1
>       .group(Field.Group.CONNECTOR_SNAPSHOT,
>           SNAPSHOT_MODE) // auto-assigned position 0
>       .create();
> ```
> 
> So this will permit to easily describe the groups and the fields that affect to them. Order is then auto calculated based on position in the list. In addition, `ConfigDefinition` will be more generic in this way to also be used to describe not only connector configuration.
> 
> This will also solve the fact that currently we just exposes to KC the wrong groups and not effectively the Field ones that are the correct.
> 
> @jpechane WDYT?

Do youplan to introduce inhertiance into that proposal so we don't nneed to redefine everyhing at each connector level?

mfvitale:
> @mfvitale
> 
> > @Naros
> > I could image something like
> > ```java
> > ConfigDefinition CONFIG_DEFINITION = ConfigDefinition.editor()
> >       .name("Postgres")
> >       .group(Field.Group.CONNECTION,
> >           HOSTNAME,      // auto-assigned position 0
> >           PORT,          // auto-assigned position 1
> >           DATABASE)      // auto-assigned position 2
> >       .group(Field.Group.CONNECTION_ADVANCED_REPLICATION,
> >           SLOT_NAME,     // auto-assigned position 0
> >           PUBLICATION)   // auto-assigned position 1
> >       .group(Field.Group.CONNECTOR_SNAPSHOT,
> >           SNAPSHOT_MODE) // auto-assigned position 0
> >       .create();
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > So this will permit to easily describe the groups and the fields that affect to them. Order is then auto calculated based on position in the list. In addition, `ConfigDefinition` will be more generic in this way to also be used to describe not only connector configuration.
> > This will also solve the fact that currently we just exposes to KC the wrong groups and not effectively the Field ones that are the correct.
> > @jpechane WDYT?
> 
> Do youplan to introduce inhertiance into that proposal so we don't nneed to redefine everyhing at each connector level?

Yes, it should support the same inheritance mechanism that we already have and also `exclude` and `include` if children should remove some field,

Naros:
@mfvitale That proposal LGTM, :rocket:

mfvitale:
For the inheritance we can thing of something like 

```java

PostgresConnectorConfig.CONFIG_DEFINITION =
      RelationalDatabaseConnectorConfig.CONFIG_DEFINITION.edit()
          .name("Postgres")

          // Simple append (most common)
          .group(Field.Group.CONNECTION,
              PLUGIN_NAME,
              SLOT_NAME)

          // Insert after specific field (when order matters)
          .groupAfter(Field.Group.CONNECTION, HOSTNAME,
              DATABASE_NAME)

         // Inserts IMPORTANT_FIELD before MAX_BATCH_SIZE
          .groupBefore(Field.Group.CONNECTOR, MAX_BATCH_SIZE, IMPORTANT_FIELD)


          // Insert at beginning (high priority)
          .groupFirst(Field.Group.CONNECTOR,
              CRITICAL_SETTING)

          .create();
```

@Binayak490-cyber seems that we have a plan. Do you still have some time to look to this?

Binayak490-cyber:
> For the inheritance we can thing of something like
> 
> ```java
> PostgresConnectorConfig.CONFIG_DEFINITION =
>       RelationalDatabaseConnectorConfig.CONFIG_DEFINITION.edit()
>           .name("Postgres")
> 
>           // Simple append (most common)
>           .group(Field.Group.CONNECTION,
>               PLUGIN_NAME,
>               SLOT_NAME)
> 
>           // Insert after specific field (when order matters)
>           .groupAfter(Field.Group.CONNECTION, HOSTNAME,
>               DATABASE_NAME)
> 
>          // Inserts IMPORTANT_FIELD before MAX_BATCH_SIZE
>           .groupBefore(Field.Group.CONNECTOR, MAX_BATCH_SIZE, IMPORTANT_FIELD)
> 
> 
>           // Insert at beginning (high priority)
>           .groupFirst(Field.Group.CONNECTOR,
>               CRITICAL_SETTING)
> 
>           .create();
> ```
> 
> @Binayak490-cyber seems that we have a plan. Do you still have some time to look to this?

Yes, sure @mfvitale! What needs to be done and what is finalised...?

mfvitale:
> > For the inheritance we can thing of something like
> > ```java
> > PostgresConnectorConfig.CONFIG_DEFINITION =
> >       RelationalDatabaseConnectorConfig.CONFIG_DEFINITION.edit()
> >           .name("Postgres")
> > 
> >           // Simple append (most common)
> >           .group(Field.Group.CONNECTION,
> >               PLUGIN_NAME,
> >               SLOT_NAME)
> > 
> >           // Insert after specific field (when order matters)
> >           .groupAfter(Field.Group.CONNECTION, HOSTNAME,
> >               DATABASE_NAME)
> > 
> >          // Inserts IMPORTANT_FIELD before MAX_BATCH_SIZE
> >           .groupBefore(Field.Group.CONNECTOR, MAX_BATCH_SIZE, IMPORTANT_FIELD)
> > 
> > 
> >           // Insert at beginning (high priority)
> >           .groupFirst(Field.Group.CONNECTOR,
> >               CRITICAL_SETTING)
> > 
> >           .create();
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @Binayak490-cyber seems that we have a plan. Do you still have some time to look to this?
> 
> Yes, sure @mfvitale! What needs to be done and what is finalised...?

Effectively we should aim for the user experience I showed in the code snippet. 

We should remove the static grouping in the `ConfigDefinition`  

```java
 private final List<Field> type;
    private final List<Field> connector;
    private final List<Field> history;
    private final List<Field> events;
```

and replace them with a `private Map<Field.Group, List<Field>> fieldsByGroup = new LinkedHashMap`` that you populate with the method I showed in the code snippet, finally on the `create` you will then calculate the position in the group similar how you already did. You should keep attention on the fact that inheritance as mentioned in previous comments.

Binayak490-cyber:
@mfvitale, @Naros, As there were some conflicts over the files changed, I rebased on latest main and squashed into a single clean commit. Here's a summary of what has been done in this PR till now:

**Core changes:**
- `buildFieldGroupOrderMap()` in `ConfigDefinition` - computes 0-based implicit ordering per `Field.Group` based on the order fields appear in the definition
- Respects explicit `positionInGroup` values when set (not the default 9999), falls back to implicit counter-based ordering otherwise
- `DebeziumDescriptorSchemaCreator` now delegates to `configDefinition.fieldGroupOrderMap()` instead of computing order inline
- `AbstractFieldTest` restored with `shouldNotHaveDuplicateGroupOrders()` using counter-based validation

**Missing `.withGroup()` entries added across connector configs:**
- `BinlogConnectorConfig`, `CommonConnectorConfig`, `RelationalDatabaseConnectorConfig`
- `MySqlConnectorConfig`, `MariaDbConnectorConfig`, `PostgresConnectorConfig`
- `SqlServerConnectorConfig`, `OracleConnectorConfig`, `MongoDbConnectorConfig`
- `JdbcSinkConnectorConfig`, `SinkConnectorConfig`, `KafkaSchemaHistory`, `RocketMqSchemaHistory`

**Pending - awaiting for your review after the recent big changes is reviewed:**
The larger redesign @mfvitale proposed - replacing the 4 fixed buckets (`type`, `connector`, `history`, `events`) in `ConfigDefinitionEditor` with a `Map<Field.Group, List<Field>>` and introducing new API methods like `group()`, `groupAfter()`, `groupBefore()`, `groupFirst()` - has **not** been implemented yet. This was left out since a big change has been done in the last squashed commit and needs to be reviewed then I can proceed with the remaining changes. Can you please take a look at it now...!

All tests passing.

Naros:
Hi @Binayak490-cyber I'm not sure this is a clean rebase, the `never` snapshot mode was added back in this PR and it was officially dropped out of `main`.

mfvitale:
@Binayak490-cyber could you take a look?

Binayak490-cyber:
> @Binayak490-cyber could you take a look?

Hi @mfvitale, @Naros, please check it out now, I have made the changes.

mfvitale:
I think you can proceed with the final implementation as agreed.

Binayak490-cyber:
> I think you can proceed with the final implementation as agreed.

I think @mfvitale, I have gone through the final agreed implementation only.

mfvitale:
> For the inheritance we can thing of something like
> 
> ```java
> PostgresConnectorConfig.CONFIG_DEFINITION =
>       RelationalDatabaseConnectorConfig.CONFIG_DEFINITION.edit()
>           .name("Postgres")
> 
>           // Simple append (most common)
>           .group(Field.Group.CONNECTION,
>               PLUGIN_NAME,
>               SLOT_NAME)
> 
>           // Insert after specific field (when order matters)
>           .groupAfter(Field.Group.CONNECTION, HOSTNAME,
>               DATABASE_NAME)
> 
>          // Inserts IMPORTANT_FIELD before MAX_BATCH_SIZE
>           .groupBefore(Field.Group.CONNECTOR, MAX_BATCH_SIZE, IMPORTANT_FIELD)
> 
> 
>           // Insert at beginning (high priority)
>           .groupFirst(Field.Group.CONNECTOR,
>               CRITICAL_SETTING)
> 
>           .create();
> ```
> 
> @Binayak490-cyber seems that we have a plan. Do you still have some time to look to this?

I'm expecting to see this. Am I'm missing something?

> Pending - awaiting for your review after the recent big changes is reviewed:
The larger redesign @mfvitale proposed - replacing the 4 fixed buckets (type, connector, history, events) in ConfigDefinitionEditor with a Map<Field.Group, List<Field>> and introducing new API methods like group(), groupAfter(), groupBefore(), groupFirst() - has not been implemented yet. This was left out since a big change has been done in the last squashed commit and needs to be reviewed then I can proceed with the remaining changes. 

Given what you mentioned earlier I thougth you first did an initial part and then the second part will follow. Isn't it?

Binayak490-cyber:
> > For the inheritance we can thing of something like
> > ```java
> > PostgresConnectorConfig.CONFIG_DEFINITION =
> >       RelationalDatabaseConnectorConfig.CONFIG_DEFINITION.edit()
> >           .name("Postgres")
> > 
> >           // Simple append (most common)
> >           .group(Field.Group.CONNECTION,
> >               PLUGIN_NAME,
> >               SLOT_NAME)
> > 
> >           // Insert after specific field (when order matters)
> >           .groupAfter(Field.Group.CONNECTION, HOSTNAME,
> >               DATABASE_NAME)
> > 
> >          // Inserts IMPORTANT_FIELD before MAX_BATCH_SIZE
> >           .groupBefore(Field.Group.CONNECTOR, MAX_BATCH_SIZE, IMPORTANT_FIELD)
> > 
> > 
> >           // Insert at beginning (high priority)
> >           .groupFirst(Field.Group.CONNECTOR,
> >               CRITICAL_SETTING)
> > 
> >           .create();
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @Binayak490-cyber seems that we have a plan. Do you still have some time to look to this?
> 
> I'm expecting to see this. Am I'm missing something?
> 
> > Pending - awaiting for your review after the recent big changes is reviewed:
> > The larger redesign @mfvitale proposed - replacing the 4 fixed buckets (type, connector, history, events) in ConfigDefinitionEditor with a Map<Field.Group, List> and introducing new API methods like group(), groupAfter(), groupBefore(), groupFirst() - has not been implemented yet. This was left out since a big change has been done in the last squashed commit and needs to be reviewed then I can proceed with the remaining changes.
> 
> Given what you mentioned earlier I thougth you first did an initial part and then the second part will follow. Isn't it?

Ah..yes @mfvitale 👍. I forgot this only. Sorry for this.

Naros:
Hi @Binayak490-cyber it looks like you may need to rebase and/or check `@Overrides` annotations. The error in the formatting stage I've often see where you've applied an `@Overrides` annotation but the annotation shouldn't exist.

Binayak490-cyber:
> Hi @Binayak490-cyber it looks like you may need to rebase and/or check `@Overrides` annotations. The error in the formatting stage I've often see where you've applied an `@Overrides` annotation but the annotation shouldn't exist.

@Naros, I am trying to do it and pushed every and many changes for it, but still it is breaking.

Naros:
Hi @Binayak490-cyber, I took a look and spotted the issue. 

So the first thing I noticed is that the error happens during the `testCompile` phase of `debezium-connector-common`. This is important because it highlights specifically where the problem is.

The first thing I did was modify the pom for `debezium-connector-common` and I adjusted the compiler plugin as follows:
```xml
            <plugin>
                <groupId>org.apache.maven.plugins</groupId>
                <artifactId>maven-compiler-plugin</artifactId>
                <configuration>
                    <compilerArgs>
                        <arg>-XDuseUnsharedTable=false</arg>
                    </compilerArgs>
                </configuration>
            </plugin>
```
This `-XDuseUnsharedTable=false` setting is a gem because by default, Maven sets this to `true`. When this is set to false, the compiler can often provide better compiler failure results for these corner cases because it doesn't use specific optimizations that can often obscure hard to understand compiler failures.

With this set to `false`, I now see:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:testCompile (default-testCompile) on project debezium-connector-common: Compilation failure
[ERROR] /home/ccranfor/github/debezium/debezium/debezium-connector-common/src/test/java/io/debezium/config/ConfigurationTest.java:[536,48] cannot find symbol
[ERROR]   symbol:   method stream()
[ERROR]   location: interface java.lang.Iterable<io.debezium.config.Field>
```

For what it's worth, I also noticed this opening the java file in my IDE using Java 21, which highlighted that the method `stream()` isn't applicable for an `Iterable`.

This worked before because `connector()` returned a `List`, but you cannot use the same API on `Iterable` objects.  You need to rely on using `StreamSupport.stream(theIterable.spliterator(), false)` to get the same outcome.

Hope that helps solve the riddle for you.

Binayak490-cyber:
> > Pending - awaiting for your review after the recent big changes is reviewed:
> > The larger redesign @mfvitale proposed - replacing the 4 fixed buckets (type, connector, history, events) in ConfigDefinitionEditor with a Map<Field.Group, List> and introducing new API methods like group(), groupAfter(), groupBefore(), groupFirst() - has not been implemented yet. This was left out since a big change has been done in the last squashed commit and needs to be reviewed then I can proceed with the remaining changes.
> 
> Given what you mentioned earlier I thougth you first did an initial part and then the second part will follow. Isn't it?

Hi @mfvitale, @Naros, @jpechane, I have implemented the remaining second part, can you please go now for reviewing this PR.

debezium-config/src/main/java/io/debezium/config/ConfigDefinition.java
```diff
@@ -25,18 +28,85 @@
...
+    private final Map<String, Integer> fieldGroupOrderMap;
 
+    ConfigDefinition(String connectorName, Map<Field.Group, List<Field>> fieldsByGroup) {
+        this.connectorName = connectorName;
+        this.fieldsByGroup = Collections.unmodifiableMap(fieldsByGroup);
+        this.fieldGroupOrderMap = Collections.unmodifiableMap(buildFieldGroupOrderMap());
+    }
+
+    // Legacy constructor for backward compatibility
+    @Deprecated
```
mfvitale: Since this is something not used externally, I think we can just drop it and adapt the code to use the new.

debezium-config/src/main/java/io/debezium/config/ConfigDefinition.java
```diff
@@ -25,18 +28,85 @@
 public class ConfigDefinition {
 
     private final String connectorName;
-    private final List<Field> type;
-    private final List<Field> connector;
-    private final List<Field> history;
-    private final List<Field> events;
+    private final Map<Field.Group, List<Field>> fieldsByGroup;
+    private final Map<String, Integer> fieldGroupOrderMap;
```
mfvitale: Is this really required? 

Could we instead put the position inside the field with the `withGroup(Field.createGroupEntry(Field.Group.CONNECTOR, 1))`? You can calculate the final positions per group inside the `io.debezium.config.ConfigDefinitionEditor#create` method?

debezium-config/src/main/java/io/debezium/config/ConfigDefinition.java
```diff
@@ -55,53 +125,54 @@ public ConfigDefinitionEditor edit() {
...
+        }
         return config;
     }
 
     public String connectorName() {
         return connectorName;
     }
 
+    // Legacy methods for backward compatibility - deprecated
+    @Deprecated
```
mfvitale: I'll try to remove all this deprecated.

debezium-config/src/main/java/io/debezium/config/ConfigDefinitionEditor.java
```diff
@@ -17,77 +19,221 @@
...
+     *
+     * @param group the field group
+     * @param fields the fields to add
+     * @return this editor for method chaining
+     */
+    public ConfigDefinitionEditor group(Field.Group group, Field... fields) {
+        List<Field> groupFields = fieldsByGroup.computeIfAbsent(group, k -> new ArrayList<>());
+        for (Field field : fields) {
+            removeFromOtherGroups(group, field.name());
+            int existingIdx = indexOfByName(groupFields, field.name());
```
mfvitale: Could you explain why this?

debezium-config/src/main/java/io/debezium/config/ConfigDefinitionEditor.java
```diff
@@ -17,77 +19,221 @@
...
+    public ConfigDefinitionEditor groupAfter(Field.Group group, Field anchor, Field... fields) {
+        List<Field> groupFields = fieldsByGroup.computeIfAbsent(group, k -> new ArrayList<>());
+        int anchorIndex = groupFields.indexOf(anchor);
+        if (anchorIndex == -1) {
+            // Anchor not found, append to end
+            return group(group, fields);
+        }
+        int insertIndex = anchorIndex + 1;
+        int insertOffset = 0;
+        for (int i = 0; i < fields.length; i++) {
```
mfvitale: Here the logic seems very similar to the `group` method. Could we maybe have another `group` method that takes as input also the position so that we can reuse the main logic?

debezium-config/src/main/java/io/debezium/config/ConfigDefinitionEditor.java
```diff
@@ -17,77 +19,221 @@
...
+    /**
+     * Adds fields to the specified group, inserting them before the specified anchor field.
+     * If a field with the same name already exists in the group, it is replaced in-place.
+     *
+     * @param group the field group
+     * @param anchor the field before which to insert
+     * @param fields the fields to insert
+     * @return this editor for method chaining
+     */
+    public ConfigDefinitionEditor groupBefore(Field.Group group, Field anchor, Field... fields) {
```
mfvitale, line 83: Same as above

debezium-config/src/main/java/io/debezium/config/ConfigDefinitionEditor.java
```diff
@@ -17,77 +19,221 @@
...
+                groupFields.add(field);
+            }
+        }
+    }
+
+    // Legacy methods for backward compatibility during transition.
+    // Fields with an explicit .withGroup() annotation keep their annotated group via group();
+    // unannotated fields fall back to the default group so they are not silently dropped.
+    @Deprecated
+    public ConfigDefinitionEditor type(Field... fields) {
```
mfvitale: Let's try to remove this deprecates

mfvitale:
@Binayak490-cyber We are in the right direction. I left some comments. Plus could you please add some unite test to test the `groupAfter`, `groupBefore` and `groupFirst` and also a inheritance scenario?

debezium-config/src/main/java/io/debezium/config/ConfigDefinition.java
Binayak490-cyber: The `fieldGroupOrderMap` is indeed redundant - the position is already implicit in the ordering of fields within `fieldsByGroup`. Then I'll compute the positions in `ConfigDefinitionEditor#create()` using `field.withGroup(Field.createGroupEntry(group, position))` and embed them directly into each field's `GroupEntry`. Then I'll update `DebeziumDescriptorSchemaCreator` to read `field.group().getPositionInGroup()` instead of the map, and remove `fieldGroupOrderMap` from `ConfigDefinition` entirely.

debezium-config/src/main/java/io/debezium/config/ConfigDefinitionEditor.java
Binayak490-cyber: Here, these two calls work together to support child connectors overriding parent field definitions safely.

`removeFromOtherGroups` ensures that if a field with the same name already exists in a different group (added by a parent `ConfigDefinition` via `.edit()`), it is moved to the new group rather than existing in two groups simultaneously - which would cause Kafka's `ConfigDef` to throw a "field defined twice" error which I faced in CI checks failures while pushing the changes in the PR.

`indexOfByName` then checks whether a field with the same name already exists in this group. If it does, the field is replaced in-place (preserving its position in the list) rather than appended as a duplicate. This allows child connectors to override a parent's field definition - for example, to attach a custom validator - while keeping the same ordering.

Binayak490-cyber:
> @Binayak490-cyber We are in the right direction. I left some comments. Plus could you please add some unite test to test the `groupAfter`, `groupBefore` and `groupFirst` and also a inheritance scenario?

@mfvitale, However, I've added the unit tests for `groupAfter`, `groupBefore`, `groupFirst` and the inheritance scenarios to `FieldTest.java` since it already uses `ConfigDefinition.editor()` extensively. But I think them to be given a dedicated `ConfigDefinitionEditorTest.java` if you feel that's a better fit, since these tests are specifically about editor behavior rather than field behavior. Let me know your preference!

debezium-config/src/main/java/io/debezium/config/ConfigDefinitionEditor.java
```diff
@@ -17,77 +19,169 @@
...
                 .collect(java.util.stream.Collectors.toList());
     }
-}
\ No newline at end of file
+
+    private List<Field> assignPositions(Field.Group group, List<Field> fields) {
+        List<Field> result = new ArrayList<>(fields.size());
+        int implicitPosition = 0;
+        for (Field field : fields) {
+            if (field.group() != null && field.group().getPositionInGroup() != 9999) {
```
mfvitale: I think this should be not permitted. Users should set group on the  `ConfigDefinitionEditor.`

debezium-config/src/main/java/io/debezium/config/ConfigDefinitionEditor.java
```diff
@@ -17,77 +19,169 @@
...
+                result.add(field);
+            }
+            else {
+                Field.Group effectiveGroup = field.group() != null ? field.group().getGroup() : group;
```
mfvitale: I think this could lead to an issue. 

Imagine that you set a the group `A` on a field but then you put that field, using the `ConfigDefinitionEditor.group()` in group B, then the position in the group will be based on the items on the group `B`. 

I think we should avoid completely to let users set the group on at field level. 

Then in the `DebeziumSchemaDescriptorCreator` I'll use the `ConfigDefinition.fieldsByGroup()`. WDYT?

mfvitale:
> But I think them to be given a dedicated ConfigDefinitionEditorTest.java

Totally agree!

debezium-config/src/main/java/io/debezium/metadata/ComponentMetadata.java
```diff
@@ -39,7 +39,7 @@ public interface ComponentMetadata {
     default ConfigDefinition getConfigDefinition() {
         return ConfigDefinition.editor()
                 .name(getClass().getName())
-                .type(getComponentFields().asArray())
+                .group(Field.Group.CONNECTOR, getComponentFields().asArray())
```
mfvitale: Could we create a `GENERIC` group and us it?

Binayak490-cyber:
@mfvitale, I have gone through your latest reviews and updated the PR, can you please now go for another round of reviewing this PR.

debezium-config/src/main/java/io/debezium/config/ConfigDefinitionEditor.java
```diff
@@ -17,77 +19,168 @@
...
+
+    private List<Field> assignPositions(Field.Group group, List<Field> fields) {
+        List<Field> result = new ArrayList<>(fields.size());
+        int implicitPosition = 0;
+        for (Field field : fields) {
+            if (field.group() != null && field.group().getPositionInGroup() != 9999) {
+                result.add(field);
+            }
+            else {
+                result.add(field.withGroup(Field.createGroupEntry(group, implicitPosition)));
```
mfvitale: I think you should remove the if/else entirely and just use this since we do not want to get group and position from the field.

debezium-config/src/main/java/io/debezium/metadata/ComponentMetadata.java
```diff
@@ -37,10 +43,16 @@ public interface ComponentMetadata {
...
-        return ConfigDefinition.editor()
-                .name(getClass().getName())
-                .type(getComponentFields().asArray())
-                .create();
+        Map<Field.Group, List<Field>> byGroup = new LinkedHashMap<>();
+        for (Field field : getComponentFields()) {
+            Field.Group group = field.group() != null ? field.group().getGroup() : Field.Group.GENERIC;
+            byGroup.computeIfAbsent(group, k -> new ArrayList<>()).add(field);
+        }
+        ConfigDefinitionEditor editor = ConfigDefinition.editor().name(getClass().getName());
```
mfvitale: Instead of `getClass().getName()` you can now use `getComponentDescriptor().getClassName()`.

debezium-config/src/main/java/io/debezium/metadata/ComponentMetadata.java
```diff
@@ -37,10 +43,16 @@ public interface ComponentMetadata {
      * In future only this method should be used and the {@code getConnectorFields()} removed
      */
     default ConfigDefinition getConfigDefinition() {
-        return ConfigDefinition.editor()
-                .name(getClass().getName())
-                .type(getComponentFields().asArray())
-                .create();
+        Map<Field.Group, List<Field>> byGroup = new LinkedHashMap<>();
```
mfvitale, line 49: I'll add a comment that this code and the `default` should be removed in future so that this method must be implemented.

debezium-config/src/main/java/io/debezium/metadata/ComponentMetadata.java
```diff
@@ -37,10 +43,16 @@ public interface ComponentMetadata {
...
+            byGroup.computeIfAbsent(group, k -> new ArrayList<>()).add(field);
+        }
+        ConfigDefinitionEditor editor = ConfigDefinition.editor().name(getClass().getName());
+        for (Map.Entry<Field.Group, List<Field>> entry : byGroup.entrySet()) {
+            editor.group(entry.getKey(), entry.getValue().toArray(new Field[0]));
+        }
+        return editor.create();
     }
 
     Field.Set getComponentFields();
```
mfvitale, line 62: I'll also add `@Deprecated` here

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogConnectorConfig.java
```diff
@@ -299,7 +299,7 @@ default boolean useConsistentSnapshotTransaction() {
             .withType(ConfigDef.Type.STRING)
             .withWidth(ConfigDef.Width.LONG)
             .withImportance(ConfigDef.Importance.LOW)
-            .withGroup(Field.createGroupEntry(Field.Group.CONNECTION_ADVANCED, 4))
+            .withGroup(Field.createGroupEntry(Field.Group.CONNECTION_ADVANCED))
```
mfvitale, line 302: At this point this should no longer used since the group must be defined with the `ConfigDefinitorEditor`

debezium-schema-generator/src/main/java/io/debezium/schemagenerator/schema/debezium/DebeziumDescriptorSchemaCreator.java
```diff
@@ -51,20 +51,25 @@ public DebeziumDescriptorSchemaCreator(ComponentMetadata componentMetadata, Fiel
...
-        StreamSupport.stream(componentMetadata.getConfigDefinition().all().spliterator(), false)
-                .map(this::buildProperty)
-                .filter(Objects::nonNull)
-                .forEach(property -> {
+        ConfigDefinition configDefinition = componentMetadata.getConfigDefinition();
+        for (Map.Entry<Field.Group, List<Field>> entry : configDefinition.fieldsByGroup().entrySet()) {
+            String groupName = formatGroupName(entry.getKey());
+            List<Field> groupFields = entry.getValue();
+            for (int i = 0; i < groupFields.size(); i++) {
+                Property property = buildProperty(groupFields.get(i), groupName, i);
```
mfvitale: At this point, since you in `ComponentMetadata.getConfigDefinition()` reconstructed the map with the info that are in `Field`, that you added in `ConfigDefinitionEditor.create()` you could just use that. So this part could remain untouched. Isn't it?

mfvitale:
> @mfvitale, I have gone through your latest reviews and updated the PR, can you please now go for another round of reviewing this PR.

@Binayak490-cyber done! I left some comments.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogConnectorConfig.java
Binayak490-cyber: @mfvitale, Understood but However, removing `.withGroup(Field.createGroupEntry(...))` from field definitions is a much larger change than it might appear - there are 414 such calls across the entire codebase. Also, the default `ComponentMetadata.getConfigDefinition()` implementation currently reads `field.group()` to determine the group assignment, so removing `.withGroup()` from fields would break that too. Should all 414 be removed as part of this PR or do you mean to remove from a particular part of the call or file ?

debezium-schema-generator/src/main/java/io/debezium/schemagenerator/schema/debezium/DebeziumDescriptorSchemaCreator.java
Binayak490-cyber: @mfvitale, you're right that after fixing `ComponentMetadata.getConfigDefinition()` and `assignPositions()` but the old approach of reading `field.group()` in `DebeziumDescriptorSchemaCreator` would also work correctly, since `assignPositions()` always overwrites the field's group with the editor's group and position. However, we kept the `fieldsByGroup()` approach because it is more explicit - it reads the group directly from the editor's authoritative map rather than relying on the side effect of `assignPositions()` updating `field.group()`. Is that so which you wanted or needs some changes ?

Binayak490-cyber:
> > @mfvitale, I have gone through your latest reviews and updated the PR, can you please now go for another round of reviewing this PR.
> 
> @Binayak490-cyber done! I left some comments.

@mfvitale, I have addressed your some comments and some I need more clearance and direction before proceeding...!

debezium-config/src/main/java/io/debezium/metadata/ComponentMetadata.java
```diff
@@ -35,13 +41,23 @@ public interface ComponentMetadata {
...
-                .type(getComponentFields().asArray())
-                .create();
+        Map<Field.Group, List<Field>> byGroup = new LinkedHashMap<>();
+        for (Field field : getComponentFields()) {
+            Field.Group group = field.group() != null ? field.group().getGroup() : Field.Group.GENERIC;
+            byGroup.computeIfAbsent(group, k -> new ArrayList<>()).add(field);
+        }
+        ConfigDefinitionEditor editor = ConfigDefinition.editor().name(getComponentDescriptor().getClassName());
+        for (Map.Entry<Field.Group, List<Field>> entry : byGroup.entrySet()) {
+            editor.group(entry.getKey(), entry.getValue().toArray(new Field[0]));
```
mfvitale, line 56: Just one point here. There are some classes, like for example transformations, that currently do no use the `ConfigDefinitionEditor` (they will do after this refactoring) so here we should compute the group position too.

debezium-schema-generator/src/main/java/io/debezium/schemagenerator/schema/debezium/DebeziumDescriptorSchemaCreator.java
mfvitale: @Binayak490-cyber You can use the `fieldsByGroup()` but the I do not expect to use the `i` to get the position since the correct position should already be in the `field.group().getPositionInGroup()`. Isn't it?

mfvitale: So I think that at least the `buildProperty` could remain unchanged.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogConnectorConfig.java
mfvitale: Let's keep them for now.

Binayak490-cyber:
@mfvitale, updated what you pointed out, can you go for another look at it.

debezium-config/src/main/java/io/debezium/metadata/ComponentMetadata.java
```diff
@@ -53,7 +54,9 @@ default ConfigDefinition getConfigDefinition() {
         }
         ConfigDefinitionEditor editor = ConfigDefinition.editor().name(getComponentDescriptor().getClassName());
         for (Map.Entry<Field.Group, List<Field>> entry : byGroup.entrySet()) {
-            editor.group(entry.getKey(), entry.getValue().toArray(new Field[0]));
+            List<Field> fields = entry.getValue();
+            fields.sort(Comparator.comparingInt(f -> f.group() != null ? f.group().getPositionInGroup() : Integer.MAX_VALUE));
```
mfvitale: I do not understand this. We should calculate the implicit order based on the list position.

mfvitale:
@Binayak490-cyber Great work so far. I left a minor comment and then you could also start working on align the non core repositories (DB2, informix, etc. )

debezium-config/src/main/java/io/debezium/metadata/ComponentMetadata.java
Binayak490-cyber: Yes, correct. Done.

debezium-config/src/main/java/io/debezium/config/Field.java
```diff
@@ -272,7 +272,8 @@ public enum Group {
         CONNECTOR,
         ADVANCED_HEARTBEAT,
         CONNECTOR_ADVANCED,
-        ADVANCED
+        ADVANCED,
+        GENERIC
```
Naros, line 282: @mfvitale does this exist solely for non-connector configurations that may not have specified groups, e.g. transformations, schema history, offset stores, etc?

mfvitale: Yes, this is the idea for now. Then the idea is to use `ConfigDefinition/ConfigDefinitioEditor` on all configurable components. WDYT?

debezium-config/src/main/java/io/debezium/config/ConfigDefinition.java
```diff
@@ -25,18 +26,11 @@
...
-    private final List<Field> type;
-    private final List<Field> connector;
-    private final List<Field> history;
-    private final List<Field> events;
+    private final Map<Field.Group, List<Field>> fieldsByGroup;
 
-    ConfigDefinition(String connectorName, List<Field> type, List<Field> connector, List<Field> history,
-                     List<Field> events) {
+    ConfigDefinition(String connectorName, Map<Field.Group, List<Field>> fieldsByGroup) {
         this.connectorName = connectorName;
```
mfvitale: Could we rename `connectorName` to `componentName`?

debezium-config/src/main/java/io/debezium/config/Field.java
Naros: That's what I thought. 

The only concern I see is that, since `GENERIC` is in the public Field API, anyone creating a new connector field could assign that group to it, and that's perfectly legal. If `GENERIC` is eventually going away, then I would at least add a comment that it should not be used and that it will be removed in the future.  WDYT?

mfvitale: @Binayak490-cyber Could you also add a comment on the `GENERIC` group?

Binayak490-cyber:
> Could we rename `connectorName` to `componentName`?

@mfvitale, @Naros, added a comment on the `GENERIC` group and also renamed `connectorName` to `componentName`. Can you check it out once.

mfvitale (approved):
@Binayak490-cyber LGTM! Great work. 

Before proceeding with the merge also the non core connectors should be aligned. So waiting for the relative PRs.

mfvitale:
@Binayak490-cyber Any update on the progress of this?

Binayak490-cyber:
Hi @mfvitale, @Naros, @kmos - I wanted to give an update on the non-core connector follow-up PRs for this work.

I went through all of the affected connectors (DB2, Informix, IBM i, Vitess, Ingres, CockroachDB, Spanner, and Cassandra) and opened follow-up PRs removing explicit `positionInGroup` values and migrating to the new `.group()` API, following the same pattern as this PR:

* DB2: debezium/debezium-connector-db2#233
* Informix: debezium/debezium-connector-informix#175
* IBM i: debezium/debezium-connector-ibmi#100
* Vitess: debezium/debezium-connector-vitess#291
* Ingres: debezium/debezium-connector-ingres#18
* CockroachDB: debezium/debezium-connector-cockroachdb#51
* Spanner: debezium/debezium-connector-spanner#164

I think so that there is one exception that is **Cassandra**. `CassandraConnectorConfig` never adopted the `ConfigDefinition`/`Field.Group` system (it doesn't use `.type()`, `.connector()`, `.events()`, `createGroupEntry`, or `configDef()`), so there's nothing to migrate there. I believe it's out of scope for this follow-up.

All seven PRs currently fail CI as expected because they depend on the new `.group()` API introduced by this PR. Once this PR is merged, those follow-ups should build against the updated core.

Please let me know if there's anything else outstanding or if this clears the path for all other non-core repos related to this PR.

mfvitale:
@Binayak490-cyber Thanks for the update. I re-run the postgres tests. Let's see if they passes. I then plan to merge this and re trigger CI on non core connectors.

mfvitale:
@Binayak490-cyber Could you take a look to the postgres failure?

Binayak490-cyber:
> @Binayak490-cyber Could you take a look to the postgres failure?

@mfvitale, done. Can you check it out once ?

mfvitale:
> > @Binayak490-cyber Could you take a look to the postgres failure?
> 
> @mfvitale, done. Can you check it out once ?

@Binayak490-cyber What was the problem?

Binayak490-cyber:
> > > @Binayak490-cyber Could you take a look to the postgres failure?
> > 
> > 
> > @mfvitale, done. Can you check it out once ?
> 
> @Binayak490-cyber What was the problem?

@mfvitale, the issue was that during a previous rebase conflict resolution, our commit accidentally removed the `applyLsnFlushModeHeartbeatFallback` implementation that was added by dbz#1605. Specifically:

- `CommonConnectorConfig.java`: `heartbeatInterval` was reverted back to `private final` and the `setHeartbeatInterval()` setter was dropped
- `PostgresConnectorConfig.java`: the `import io.debezium.heartbeat.Heartbeat`, the `applyLsnFlushModeHeartbeatFallback(config)` constructor call, and the method itself were all removed

The test `shouldApplyLsnFlushModeHeartbeatFallbackWhenNoOpportunityForFlush` was correctly catching this regression. We've restored the missing code in the latest push.

mfvitale:
@Binayak490-cyber Could you please do the last rebase and we are good to go?

Binayak490-cyber:
> @Binayak490-cyber Could you please do the last rebase and we are good to go?

@mfvitale, Done.

mfvitale:
@Binayak490-cyber Applied, thanks!

PR #7302, opened 2026-04-07
debezium/dbz#1713 Use log count/size in place of batch/sleep

### LINKED ISSUE #1713: Add configurable way to scale the Oracle LogMiner batch size window.
### Motivation

In Debezium 1.x, 2.x, and earlier versions of Debezium 3.x, the connector used a mining window algorithm that, when it had fallen significantly behind, would use the current database write position and perform a single large pass over the archive and redo logs to catch up. However, this proved problematic in environments that were handling massive data loads or had a high volume of activity, as the connector could never catch up. 

As a solution in Debezium 3.2, we adjusted the mining logic so that the batch window scales linearly rather than with a single large jump. This algorithm works by always extending the batch window size `log.mining.batch.size.max`. For customers with a high volume of activity, this approach often worked better and helped maintain sustained ingestion of changes, but there were a few issues.

## Implementation

The idea is to introduce a new `log.mining.batch.size.window.scale` option that gives the user control over exactly how the connector scales the mining window when it falls behind and reaches `log.mining.batch.size.max`. The window scale can be set to one of three values

**`linear`**
This will be the default mode, and is the current behavior in Debezium 3.4 and 3.5. It works by checking if the batch size needs to be extended when we continue to remain a `log.mining.batch.size`, and takes the previous value and adds `log.mining.batch.size.max` on each step, growing the mining window's maximum range by a linear scale.

**`current_write`**
This is the old behavior from Debezium 1.x, 2.x, and early 3.x days. It works by determining if the connector needs to make a large leap to catch up, potentially enqueuing dozens of archive logs at once. For systems with small bursts of activity, this allows them to use the legacy behavior, which may be faster on smaller systems than `linear`.

**`exponential`**
This is a new mode that works almost identically to `linear`; however, the increment occurs exponentially. This provides a stable middle ground between the two extremes, which aligns with the `linear` approach during low-activity periods and trends toward `current_write` during high-activity periods.

Naros:
Fixes debezium/dbz#1713

## Description
One of the most common pieces of feedback about the Oracle connector is that it's very difficult to determine which settings to use for `log.mining.batch.size.*`. Given that an SCN is not unique, and batch jobs or high activity throughput can cause multiple changes to be assigned the same SCN, values you might use during normal activity time are not sufficient for high-volume periods. 

So what is needed is a consistent way to read a consistent block of data, regardless of how many unique SCN changes are within that block. The most logical representation of that is a log's size in bytes.

At the forefront of this PR, all logic related to the following properties is gone:
* `log.mining.batch.size.*`
* `log.mining.sleep.time.*`
* `log.mining.scn.gap.detection.*`

These options attempted to help the old algorithm provide a sliding scale based on whether the connector was close to real time or falling behind. This is what made it difficult to tune.

For `redo_log_catalog`, this means each mining step pauses for exactly 1s between polls, and the connector will use the `UnboundedLogFileSessionSelector`, reading all available log files from the current read position to present. This is the same behavior as existed before, as this is needed due to the flushing of the data dictionary.

_This new feature may be patched in for redo_log_catalog users, but that's currently out of scope_.

For other mining strategies (`hybrid` and `online_catalog`), a new `CappedLogFileSessionSelector` implementation is used, based on the new configuration property `log.mining.log.count.min`. This setting specifies the minimum number of logs per redo thread that should be mined in a given mining step, defaulting to `2`. If the user overrides this option and sets the count to `0`, the connector will fall back to `UnboundedLogFileSessionSelector` behavior.

With `hybrid` or `online_catalog` and given a redo log file size of 1GB on an Oracle RAC cluster with 2 nodes, the connector will attempt to read at least 4GB of data per mining step (2 * 1GB * number of nodes). This makes it much easier to identify the IO cost per mining step with consistent values.

In addition, this `CappedLogFileSessionSelector` has built-in long-running transaction support. When a subsequent mining step resolves exactly to the same log files as the previous mining step, the minimum number of logs will be incremented by 1, effectively adding 1 more log per thread to the mine log list. This helps keep the IO load contained during long-running transactions while also preventing the connector from stalling due to capped logs.

In addition to the removal of metrics for batch/sleep, a new metric, `CurrentMinedLogCount`, was added to track the current number of logs in the active or prior mining session. This complements the existing min/max tracking, providing a solid histogram-like representation of the upper/lower bounds and the current value for trend analysis.

## PR Checklist

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/CappedLogFileSessionSelector.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/CappedLogFileSessionSelector.java
@@ -1,0 +1,143 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.oracle.logminer;
+
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.stream.Collectors;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.debezium.DebeziumException;
+import io.debezium.connector.oracle.RedoThreadState.RedoThread;
+import io.debezium.connector.oracle.Scn;
+import io.debezium.connector.oracle.logminer.LogFileCollector.LogFilesResult;
+
+/**
+ * A LogMiner log file selector that caps the returned logs based on user configuration, while capping the
+ * mining session window upper boundary to the minimum upper system change number across all threads.
+ *
+ * @author Chris Cranford
+ */
+public class CappedLogFileSessionSelector implements LogFileSessionSelector {
+
+    private final Logger LOGGER = LoggerFactory.getLogger(CappedLogFileSessionSelector.class);
+
+    private final int minimumLogsPerRedoThread;
+    private final long redoLogSizeInBytes;
+
+    private int logsPerRedoThread;
+    private Map<Integer, List<LogFile>> previousCappedLogsByThread;
+
+    public CappedLogFileSessionSelector(int minimumLogsPerRedoThread, long redoLogSizeInBytes) {
+        this.minimumLogsPerRedoThread = minimumLogsPerRedoThread;
+        this.redoLogSizeInBytes = redoLogSizeInBytes;
+        this.logsPerRedoThread = minimumLogsPerRedoThread;
+    }
+
+    @Override
+    public SessionLogSelection selectLogsForSession(LogFilesResult logFilesResult, Scn upperBoundary) {
+        Scn effectiveUpperBoundary = upperBoundary;
+
+        // Groups all collected logs by redo thread, sorted in ascending order by sequence.
+        // The ordering is important for this algorithm when inspecting what is the first/last logs per thread.
+        final Map<Integer, List<LogFile>> logsByThread = logFilesResult.logFiles().stream()
+                .sorted(Comparator.comparing(LogFile::getSequence))
+                .collect(Collectors.groupingBy(LogFile::getThread));
+
+        Map<Integer, List<LogFile>> cappedLogsByThread = getThreadLogsCappedBySize(logsByThread, (long) logsPerRedoThread * redoLogSizeInBytes);
+
+        if (previousCappedLogsByThread != null) {
+            if (cappedLogsByThread.equals(previousCappedLogsByThread)) {
+                // Same log set as last iteration: the lower watermark did not advance, so a long-running
+                // transaction may extend beyond the current cap. Grow by one log to find the end.
+                logsPerRedoThread++;
+                LOGGER.debug("Capped log set unchanged, growing log count per redo thread to {}.", logsPerRedoThread);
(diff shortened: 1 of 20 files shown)


Naros:
Hi @nathan-smit-1, this is my first pass at the new logic for capping the LogMiner IO based on log file sizes. If you have some spare time to give it a test driven and share your thoughts, I'd appreciate it. Overall, I believe it's mostly feature complete, although I need to fully test it in corner case deployments like archive-log-only mode, etc.

nathan-smit-1:
> Hi @nathan-smit-1, this is my first pass at the new logic for capping the LogMiner IO based on log file sizes. If you have some spare time to give it a test driven and share your thoughts, I'd appreciate it. Overall, I believe it's mostly feature complete, although I need to fully test it in corner case deployments like archive-log-only mode, etc.

I'm going to test this a bit today.  What are good metrics to track as I see batch size is no longer there?  I just briefly scanned but looks like there aren't any new metrics.

BTW, extractJarPath in KafkaConnectDiscoveryService.java is causing builds to fail on windows it looks like.  Seems like a recent change.  I'll maybe log a ticket for it.  I think it's just related to the path

Naros:
Hi @nathan-smit-1, yes please raise a ticket, if there isn't one already.

As for metrics, I would say the best ones for now are the lag-based metrics. During loads, do you observe that these values remain more consistent, or at the very least, when they spike, they drop much faster?

The only really valuable metrics I can foresee with the changes would be to incorporate something like:
* Current minimum log count used for the mining step
* Total bytes to read for a mining step

But these metrics break down somewhat when we reach the online redo log range, particularly the latter, since the total number of bytes for online redo logs is technically unknown until it rolls to the archive. We could use the current redo log file size, but it would be a bit inaccurate IO if the log just rolled and the current one isn't yet full.

If you have any ideas, I'm certainly open.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/CappedLogFileSessionSelector.java
```diff
@@ -0,0 +1,143 @@
...
+                logsPerRedoThread++;
+                LOGGER.debug("Capped log set unchanged, growing log count per redo thread to {}.", logsPerRedoThread);
+                cappedLogsByThread = getThreadLogsCappedBySize(logsByThread, (long) logsPerRedoThread * redoLogSizeInBytes);
+            }
+            else if (logsPerRedoThread > minimumLogsPerRedoThread) {
+                // Log set changed: the watermark advanced, so reset the count back to the configured minimum.
+                logsPerRedoThread = minimumLogsPerRedoThread;
+                LOGGER.debug("Capped log set changed, resetting log count per redo thread to {}.", logsPerRedoThread);
+                cappedLogsByThread = getThreadLogsCappedBySize(logsByThread, (long) logsPerRedoThread * redoLogSizeInBytes);
+            }
```
jpechane, line 70: There is no need for `else` branch?

jpechane (approved):
@Naros Nice work, thanks! I left one small comment to double check, if OK feel free to merge yourself.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/CappedLogFileSessionSelector.java
Naros: There is not. The `else` block is an implicit "log set changed AND logsPerRedoThread == min", and in this case, the `cappedLogsByThread` computation before the if handled that case already.  The `if` and `else if` handle a recompute pass when the log set is the same OR log set changed and we're not at the min.

PR #7347, opened 2026-04-14
debezium/dbz#1813 Reduce memory usage using Interner

### LINKED ISSUE #1813: Schema history: optimize memory usage

`We can greatly improve memory usage when the database schema contains identical tables. This is a typical situation for e.g. multi-tenant databases where each tenant has its own schema with the same structure as the other tenants.`

---

`By "interning" (i.e. similarly to String.intern() API) schema history objects (basically tables, columns, attributes), we keep only one copy for each of them, thus reducing drastically memory usage.`

ochedru:
## Summary
- Introduce a generic `Interner` utility to deduplicate frequently repeated strings and objects in memory
- Apply interning to `TableId`, `TableImpl`, `ColumnImpl`, and `AttributeImpl` to reduce heap usage
- Add a configurable `schema.history.internal.memory.optimization` option to enable the `Interner` (default is `off`), either for the connector only (`on`) or shared with other connectors (`shared`).

Closes https://github.com/debezium/dbz/issues/1813

## Test plan
- [x] `InternerTest` — verifies interning behavior (identity, concurrency, null handling)
- [x] `TableImplInterningTest` — verifies interning applied correctly to relational model objects

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/relational/history/InternerMXBean.java
+++ b/debezium-connector-common/src/main/java/io/debezium/relational/history/InternerMXBean.java
@@ -1,0 +1,34 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.relational.history;
+
+/**
+ * JMX metrics for the schema object interner.
+ * Exposed under {@code debezium.<connector>:type=connector-metrics,context=interner,server=<logical-name>}.
+ *
+ * <p>Hit and miss counts cover non-string objects only (columns, attributes, lists).
+ * String values are delegated to the JVM string pool and are not counted here.</p>
+ */
+public interface InternerMXBean {
+
+    /** Number of times {@code intern()} returned an already-cached canonical instance. */
+    long getHitCount();
+
+    /** Number of times {@code intern()} stored a new instance in the pool. */
+    long getMissCount();
+
+    /**
+     * Current number of entries held in the pool.
+     * Includes entries whose referents may have been garbage-collected but not yet expunged.
+     */
+    int getPoolSize();
+
+    /**
+     * Hit ratio as a percentage ({@code hitCount / (hitCount + missCount) * 100}),
+     * or {@code 0} if no calls have been recorded yet.
+     */
+    double getHitRatioPercent();
+}
--- a/debezium-connector-common/src/main/java/io/debezium/relational/history/SchemaHistory.java
+++ b/debezium-connector-common/src/main/java/io/debezium/relational/history/SchemaHistory.java
@@ -131,6 +131,16 @@ public interface SchemaHistory {
             .withDescription("The unique identifier of the Debezium connector")
             .withNoValidation();
 
+    Field MEMORY_OPTIMIZATION = Field.create(CONFIGURATION_FIELD_PREFIX_STRING + "memory.optimization")
+            .withDisplayName("Enable memory optimization for schema history")
+            .withType(Type.BOOLEAN)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withDescription("Controls whether Debezium deduplicates identical schema objects (tables, columns, attributes) "
+                    + "in memory using an interner. When enabled, this can significantly reduce heap usage for connectors "
+                    + "that track many tables with similar structures. When disabled (the default), no deduplication is performed.")
+            .withDefault(false);
+
     // Temporary preference for DDL over logical schema due to DBZ-32
     Field INTERNAL_PREFER_DDL = Field.create(CONFIGURATION_FIELD_PREFIX_STRING + "prefer.ddl")
             .withDisplayName("Prefer DDL for schema recovery")
--- a/debezium-connector-common/src/main/java/io/debezium/relational/history/SchemaHistory.java
+++ b/debezium-connector-common/src/main/java/io/debezium/relational/history/SchemaHistory.java
@@ -131,6 +131,14 @@ public interface SchemaHistory {
             .withDescription("The unique identifier of the Debezium connector")
             .withNoValidation();
 
+    Field MEMORY_OPTIMIZATION = Field.create(CONFIGURATION_FIELD_PREFIX_STRING + "memory.optimization")
+            .withDisplayName("Memory optimization mode for schema history")
+            .withEnum(MemoryOptimizationMode.class, MemoryOptimizationMode.OFF)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withDescription("Controls how Debezium deduplicates identical schema objects (tables, columns, attributes) "
+                    + "in memory using an interner.");
+
     // Temporary preference for DDL over logical schema due to DBZ-32
     Field INTERNAL_PREFER_DDL = Field.create(CONFIGURATION_FIELD_PREFIX_STRING + "prefer.ddl")
             .withDisplayName("Prefer DDL for schema recovery")
--- a/debezium-util/src/main/java/io/debezium/util/Interner.java
+++ b/debezium-util/src/main/java/io/debezium/util/Interner.java
@@ -1,0 +1,144 @@
+/*
+ * Copyright Debezium Authors.
+ *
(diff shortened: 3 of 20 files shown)


jpechane:
@ochedru Hi,how does this work in case a schema of a table/column is updated?

ochedru:
> @ochedru Hi,how does this work in case a schema of a table/column is updated?

Hi @jpechane, all the interned objects are immutable (final fields of immutable types): `String`, `Attribute`, `ColumnImpl` and `TableImpl` as well as unmodifiable collections of these types.

When a schema is updated, new objects are created and the Interner takes care of discarding unreferenced entries.

ochedru:
Hi @jpechane, is there any chance to get this merged? I've been running it in production successfully for one month.

debezium-util/src/main/java/io/debezium/util/Interner.java
```diff
@@ -0,0 +1,144 @@
...
+ * when thousands of database schemas share identical table structures.</p>
+ *
+ * <p>Warning: do not use with mutable objects. The interning contract assumes that
+ * objects do not change after being interned.</p>
+ */
+public final class Interner {
+
+    private static final ConcurrentMap<WeakEntry, WeakEntry> MAP = new ConcurrentHashMap<>();
+    private static final ReferenceQueue<Object> QUEUE = new ReferenceQueue<>();
+    private static boolean enabled = false;
```
Naros: If I am using two connectors in Kafka Connect, where I would want to apply interning to only one of the two connectors, with this being a static state, I would be unable to do that. It's either all or nothing, which I'm not sure is the best choice.

debezium-util/src/main/java/io/debezium/util/Interner.java
```diff
@@ -0,0 +1,144 @@
...
+    private Interner() {
+    }
+
+    /**
+     * Enables or disables the interner.
+     * When disabled, {@link #intern(Object)} returns the input value without deduplication.
+     *
+     * @param enabled {@code true} to enable interning, {@code false} to bypass it
+     */
+    public static void setEnabled(boolean enabled) {
```
Naros: When enabled, should there be metrics that track hits vs. new entries to help validate whether the intern optimization is worth enabling for a given connector deployment?

debezium-util/src/main/java/io/debezium/util/Interner.java
ochedru: Thanks for your feedback. I'll fix this.

debezium-util/src/main/java/io/debezium/util/Interner.java
ochedru: Not sure I can easily track the number of references, but I'll have a look. One can verify the optimization is working by monitoring the heap usage.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogSourceInfo.java
```diff
@@ -153,7 +154,7 @@ public String getQuery() {
      */
     public void setBinlogPosition(String binlogFilename, long positionOfFirstEvent) {
         if (binlogFilename != null) {
-            this.currentBinlogFilename = binlogFilename;
+            this.currentBinlogFilename = Interner.intern(binlogFilename);
```
Naros: If we're going to do this for MySQL/MariaDB source info blocks, should this not be applied universally? In my genuinely opinion, this seems a bit outside the PR scope, but ymmv.

ochedru: You're right, this is probably overkill.

ochedru: I reverted this and added some doc about the new JMX metrics.

ochedru:
Hi @Naros, please let me know if there is anything else I should do. Thanks!

debezium-connector-common/src/main/java/io/debezium/relational/history/InternerMXBean.java
```diff
@@ -0,0 +1,34 @@
...
+
+    /**
+     * Current number of entries held in the pool.
+     * Includes entries whose referents may have been garbage-collected but not yet expunged.
+     */
+    int getPoolSize();
+
+    /**
+     * Hit ratio as a percentage ({@code hitCount / (hitCount + missCount) * 100}),
+     * or {@code -1} if no calls have been recorded yet.
```
Naros:
```suggestion
     * or {@code 0} if no calls have been recorded yet.
```
Either the Javadoc is wrong or the implementation is.

debezium-connector-common/src/main/java/io/debezium/relational/history/SchemaHistory.java
```diff
@@ -131,6 +131,19 @@ public interface SchemaHistory {
             .withDescription("The unique identifier of the Debezium connector")
             .withNoValidation();
 
+    Field MEMORY_OPTIMIZATION = Field.create(CONFIGURATION_FIELD_PREFIX_STRING + "memory.optimization")
```
Naros, line 134: Should we consider this to be an `EnumeratedValue` property rather than string-based?

Naros (commented):
I only see a few remaining items, shown below. @jpechane any feedback?

ochedru:
Hi @Naros, thanks for your feedback! I've taken into account your comments. (I had to force push because I forgot to sign-off the commits.)

documentation/modules/ROOT/partials/modules/all-connectors/ref-connector-monitoring-schema-history-metrics.adoc
```diff
@@ -38,3 +38,32 @@ The following table lists the schema history metrics that are available.
...
+
+|[[connectors-interner-metric-misscount_{context}]]<<connectors-interner-metric-misscount_{context}, `MissCount`>>
+|`long`
+|Number of times `intern()` stored a new instance in the pool.
+
+|[[connectors-interner-metric-poolsize_{context}]]<<connectors-interner-metric-poolsize_{context}, `PoolSize`>>
+|`int`
+|Current number of entries held in the pool. Includes entries whose referents may have been garbage-collected but not yet expunged.
+
+|[[connectors-interner-metric-hitratioppercent_{context}]]<<connectors-interner-metric-hitratioppercent_{context}, `HitRatioPercent`>>
```
Naros:
```suggestion
|[[connectors-interner-metric-hitratiopercent_{context}]]<<connectors-interner-metric-hitratiopercent_{context}, `HitRatioPercent`>>
```

Naros (commented):
Hi @ochedru, I only spotted one remaining item in the docs. @jpechane where do you stand on this one?

documentation/modules/ROOT/partials/modules/all-connectors/ref-connector-monitoring-schema-history-metrics.adoc
ochedru: Indeed that'a typo. Thanks for the suggested change! Applied.

jpechane:
@ochedru I think this is something thayt can be merged in 3.7.
I still have one big concern and that's maintainability of the code. In future code changes it is pretty easy to forget to use interner etc. I still wonder if there is a way how to spot or automate that.

ochedru:
Hi @jpechane, thank you for your feedback! I'll check if I can add a test verifying all schema objects have their fields properly interned.

ochedru:
> @ochedru I think this is something thayt can be merged in 3.7. I still have one big concern and that's maintainability of the code. In future code changes it is pretty easy to forget to use interner etc. I still wonder if there is a way how to spot or automate that.

Hi @jpechane, I've added `SchemaObjectInterningReflectionTest` to make sure all fields in schema objects are interned when required.

jpechane:
@ochedru We are on 3.7 now. Could you please do the final rebase?

jpechane:
@ochedru Could you please check all commits and make sure they are signed off and the message is properly fortmatted?

ochedru:
> @ochedru Could you please check all commits and make sure they are signed off and the message is properly fortmatted?

Hi @jpechane
I've squashed all my commits with a proper commit message, and force pushed the branch.

Naros (approved):

Naros:
Thanks @ochedru, applied! :tada:

PR #7350, opened 2026-04-15
debezium/dbz#1819 ChangeEventSourceCoordinator correctly initialize streaming state

### LINKED ISSUE #1819: Triggering a blocking snapshot can occur before the streaming loop is ready
After a recent change to `ChangeEventSourceCoordinator` in debezium/dbz#1778, the `ChangeEventSourceCoordinator` does not initialize the `streaming` member variable when streaming actually begins. This previously used to be triggered when `doBlockingSnapshot` was called, but was removed.

As a result, when `doBlockingSnapshot` is triggered by a signal, it starts the blocking snapshot without waiting for the streaming loop to indicate it is in a state for the blocking snapshot to execute. This is because the `streaming` member variable is never initialized to `true`, and when the context waits for streaming to pause, the condition wait is skipped as a result.

For Oracle, this yields a ResultSet exception as the Snapshot and Streaming reuse the same JDBC connection, and the triggering on the blocking snapshot invalidates the ResultSet that the streaming loop is still using.

Naros:
Fixes debezium/dbz#1819

## Description
The first commit in this PR addresses a runtime regression introduced by debezium/dbz#1778 where the initial streaming state was not being initialized in the `ChangeEventSourceCoordinator`, leading to unintended mishandling of the trade off from streaming to the blocking snapshot for some connectors.

The second commit in this PR addresses a test regression introduced by the same change: a new test case uses the `LogInterceptor` to capture events for a specific package. Unfortunately, Loggers are not hierarchical, so if a specific connector module creates a dedicated logger for a class that emits the message the test is interested in, the package logger never sees it, and the test stalls. This commit introduces a specialization of `LogInterceptor` to listen for logged messages from a package, and registers the interceptor for the package, its classes, and any loggers defined in subpackages or subpackage classes, recursively.

## PR Checklist

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
+++ b/debezium-connector-common/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java
@@ -523,6 +523,7 @@ public class ChangeEventSourceCoordinator<P extends Partition, O extends OffsetC
             streamingMetrics.connected(status);
             LOGGER.info("Connected metrics set to '{}'", status);
         }
+        this.streaming = status;
     }
 
     protected class CatchUpStreamingResult {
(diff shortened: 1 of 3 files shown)


Naros:
I've verified the Oracle tests locally, so I don't believe there is a need to execute the full 6hr pass, so I'll trigger just the LogMiner ones for sanity sake.

jpechane (approved):
@Naros LGTM, thanks. Feel free to merge yourself.

PR #7362, opened 2026-04-16
debezium/dbz#1829 Add parallel incremental snapshot for relational connectors

### LINKED ISSUE #1829: Add parallel incremental snapshot with configurable batch flush, SPI completion signal, and snapshot transforms
### Description

Add support for parallel incremental snapshots with table-level workers, configurable batch flush size, a new SPI method to signal table completion, and snapshot transforms applied directly in the worker pipeline.

### Motivation

The current incremental snapshot implementation has several limitations:

1. **Single-threaded**: Tables are processed one at a time, leaving CPU and I/O bandwidth underutilized
2. **Fixed buffer size**: The internal batch buffer is hardcoded, causing OOM on tables with wide rows or constrained environments
3. **No table completion signal**: Sink connectors receive chunks via `onTableSnapshotCompleted()` but have no way to know when a table is finished — forcing sinks to create a new writer per chunk, causing massive overhead and producing many small files
4. **No transforms on snapshot path**: Snapshot events bypass the engine's SMT chain — they are delivered as raw `Object[]` rows to completion handlers. Sink connectors must implement their own column casting/type conversion, duplicating logic that already exists in the CDC path via SMTs like `ExtractNewRecordState`

### Proposed Changes

1. **New SPI**: `SnapshotTableCompletionHandler` with `onTableSnapshotCompleted()` (per chunk), `onTableSnapshotFinished()` (per table, default no-op), and `shouldHandle()` filter
2. **Configurable `incremental.snapshot.batch.flush.size`**: Default 20,000 rows per flush (previously hardcoded)
3. **Parallel snapshot infrastructure**: `ParallelIncrementalSnapshotCoordinator`, `TableSnapshotWorker`, `IncrementalSnapshotRetryPolicy`
4. **Work-queue pattern**: Replaces batch-and-wait scheduling with a shared `ConcurrentLinkedQueue` — workers pick the next table immediately upon completing the current one, eliminating idle time
5. **PostgreSQL read-only support**: Parallel-safe JDBC connections via `createSnapshotConnection()` override
6. **Snapshot transforms in worker pipeline**: `RecordTransformer` SPI and `SnapshotRecordBuilder` convert JDBC rows to `SourceRecord` objects and apply the registered SMT chain inside `TableSnapshotWorker`, before delivering to handlers. This aligns snapshot and CDC paths — handlers receive pre-transformed `List<SourceRecord>` ready for direct writing

### Backward Compatibility

- Default `threads=1` preserves original single-threaded behavior
- SPI methods have default no-op implementations
- If no `RecordTransformer` is registered, records pass through unchanged
- No breaking API changes

MrIvv:
## Fixes
[debezium/dbz#1829](https://github.com/debezium/dbz/issues/1829)

## Summary
This PR adds **table-level parallelism** to incremental snapshots. Multiple data collections in a signal are scanned concurrently within a single DBLog watermark window, bounded by `snapshot.max.threads`. Per-chunk deduplication semantics are preserved.

Supporting changes: a retry policy with exponential backoff for transient JDBC failures during chunk reads, and a work-queue scheduler that keeps workers busy across the whole signal.

## Motivation
Today's incremental snapshot scans tables strictly sequentially. On any reasonable hardware (multi-core, bonded NICs, fast storage) the bottleneck is one slow table at a time, not the total work — leaving CPU and JDBC bandwidth idle while the connector waits.

This PR removes that bottleneck without changing the semantics of the snapshot itself.

### 1. Parallel table scan during incremental snapshot
`AbstractIncrementalSnapshotChangeEventSource` runs each round (one DBLog watermark window) across `snapshot.max.threads` worker threads, each holding its own `TableSnapshotContext` and JDBC connection. The window is opened once for the whole round and closed once at the end — workers do not contend on the watermark, they only contend on the per-table chunk position.

`SignalBasedIncrementalSnapshotChangeEventSource` and `PostgresReadOnlyIncrementalSnapshotChangeEventSource` are wired to use the parallel coordinator when `snapshot.max.threads > 1` and degrade to the existing sequential path when `= 1` (default).

### 2. Work-queue scheduling (no batch-and-wait idle)
When the signal carries N tables and the pool has K workers, all N tables go into a `ConcurrentLinkedQueue`. Each worker picks the next table immediately upon completing its current one — no `awaitCompletion()` between sub-batches.

| Aspect | Sequential / batch-and-wait | This PR |
|---|---|---|
| Worker idle between tables | Yes, worker waits for slowest sibling | No, picks next from shared queue |
| JDBC connection | Acquired per table | Held by worker across all its tables |
| `awaitCompletion()` calls | One per sub-batch | One total at end of round |
| Tasks submitted | N per sub-batch | Exactly `workerCount` total |

### 3. New helper: `ParallelIncrementalSnapshotCoordinator`
Owns the JDBC connection pool, the persistent worker executor, and the per-table window buffers. Pool lifecycle is on-demand:

- Opens lazily on the first signal arrival, with `snapshot.max.threads` JDBC connections.
- Reused across back-to-back signals (no per-signal pool churn).
- Released after a configurable grace period (`incremental.snapshot.pool.release.delay.ms`, default `60_000`) once no signal is in flight. A `0` value releases immediately.
- Race-free release: an `AtomicLong` generation counter aborts a scheduled close if a new signal arrives mid-grace.

Pooled connections are validated on borrow via `JdbcConnection.isValid()` and lazily reallocated on failure. `CopyOnWriteArrayList` tracks live connections so concurrent eviction stays safe. The worker executor is a single `Threads.newFixedThreadPool` for the whole snapshot lifetime; it is shut down only when the coordinator itself terminates.

Connection threading is end-to-end: `ChunkQueryBuilder.readTableChunkStatement` now takes the worker's `JdbcConnection` as a parameter, so workers do not contend on a shared field or perform connection-swap dances mid-chunk.

The coordinator is annotated `@NotThreadSafe` for state changes and uses a `ReentrantLock` + `AtomicReference` pattern (consistent with `BaseSourceTask`) for the public lifecycle methods (`ensurePoolOpen`, `scheduleReleaseIfNotPending`, `shutdown`).

### 4. Retry policy with exponential backoff: `RetryExecutor`
New utility in `debezium-util`. Generic over `Callable`/`Runnable`, accepts any retryable predicate, exponential backoff with jitter, configurable cap. Used by the parallel chunk round to recover from transient JDBC failures (lock timeout, deadlock victim, broken connection in the pool) without aborting the whole snapshot.

## Configuration
| Property | Type | Default | Description |
|---|---|---|---|
| `snapshot.max.threads` | INT | 1 | Pre-existing. `1` = sequential (no behavior change). `> 1` = parallel incremental + chunked snapshot |
| `incremental.snapshot.pool.release.delay.ms` | LONG | 60000 | Grace period before the snapshot JDBC pool is released after the last signal completes. `0` releases immediately. |
| `incremental.snapshot.retry.max.attempts` | INT | 5 | Max retries on transient failure during a chunk read |
| `incremental.snapshot.retry.initial.delay.ms` | LONG | 1000 | Initial backoff delay |
| `incremental.snapshot.retry.max.delay.ms` | LONG | 60000 | Max backoff delay cap |
| `incremental.snapshot.retry.backoff.multiplier` | DOUBLE | 2.0 | Exponential growth factor |

All five new properties carry `DEFAULT_*` `public static final` constants. A new `Field::isPositiveDouble` validator is added for the backoff multiplier; the other properties use existing validators.

## Testing
* Runtime-validated on production-configured PostgreSQL 17 deployments with the Iceberg sink (Lakekeeper REST catalog on GCS), `snapshot.max.threads=2`, chunk size 20000. The parallel coordinator initializes the pool on first signal arrival, reuses it across back-to-back signal bursts, and releases it cleanly once the queue drains past the grace period. Workers stay continuously busy via the shared queue — verified zero idle time between sub-batches in the live logs. No leaked connections or scheduled-release races observed across multiple recovery cycles.

## Backward compatibility
- `snapshot.max.threads` defaults to `1` → no behavior change unless explicitly opted in
- Notification format unchanged (same `aggregateType="Incremental Snapshot"` + same per-type schema)
- New retry properties are additive, all have safe defaults

## AI usage disclosure
I used Claude (Anthropic) as a code assistant and for troubleshooting during this work. I reviewed every change, ran the test suite locally, and validated the runtime behaviour. I take full responsibility for the code submitted.

## PR Checklist


Naros:
Hi @MrIvv, there appears to be a lot of overlap with the new Chunked Initial Snapshot behavior, just refactored to its own classes.  Can you see if we can reuse that somehow?

Also please prefix all your commits as debezium/dbz#1829

MrIvv:
Hi @Naros, you're right. I'm already working on the refactoring and I'll update this pr soon

Naros:
@MrIvv before you get too far along, it appears your PR is reintroducing debezium-core, which it should not. You may want to rebase on the latest main so that the classes that have been recently reorganized into various new modules are not showing as new files in the PR.

Naros:
Hi @MrIvv, it appears there is one commit in the PR that is missing the proper prefix.

debezium-api/pom.xml
```diff
@@ -37,8 +37,6 @@
             <artifactId>assertj-core</artifactId>
             <scope>test</scope>
         </dependency>
-
-        <!-- Used for unit testing with Kafka -->
```
Naros, line 41: This removal seems unrelated to the change at hand.

debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -766,6 +766,50 @@ public static SnapshotQueryMode parse(String value, String defaultValue) {
...
 
+    public static final Field INCREMENTAL_SNAPSHOT_BATCH_FLUSH_SIZE = Field.create("incremental.snapshot.batch.flush.size")
+            .withDisplayName("Incremental snapshot batch flush size")
+            .withType(Type.INT)
+            .withWidth(Width.MEDIUM)
+            .withImportance(Importance.MEDIUM)
+            .withDescription("Number of rows accumulated in memory before flushing to the sink during incremental snapshot. "
+                    + "Lower values reduce memory usage but increase flush overhead. "
+                    + "Default 20000 keeps memory under 300MB per worker thread.")
+            .withDefault(20_000)
```
Naros: Move default values to `public static final` fields in the configuration class. This applies here and with the other added configuration properties. These values also probably deserve some sort of field validation, making sure they're positive at a minimum.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/IncrementalSnapshotRetryPolicy.java
```diff
@@ -0,0 +1,306 @@
...
+ *
+ * Object result = retryPolicy.executeWithRetry(
+ *     () -> performDatabaseOperation(),
+ *     "database operation"
+ * );
+ * </pre>
+ *
+ * @author Ivan Senyk
+ */
+public class IncrementalSnapshotRetryPolicy {
```
Naros: Let's rename this class to RetryExecutor or equivalent, and move it to `debezium-utils`. This way we can actually reuse it in Debezium Server, which already has a very similar class there, to minimize code dupe.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/RecordTransformer.java
```diff
@@ -0,0 +1,23 @@
...
+ * snapshot events receive the same transformations as CDC events.
+ *
+ * <p>If no transformer is registered, the worker passes records through unchanged.
+ *
+ * @author Ivan Senyk
+ */
+@FunctionalInterface
+public interface RecordTransformer {
+    SourceRecord transform(SourceRecord record);
+}
```
Naros: It seems very unusual that we'd do this for incremental snapshots only, so can you explain why this is necessary and why wouldn't you just use the existing standard mutation mechanisms of `CustomConverter`, `PostProcessor`, or Kafka `Transformation`?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/SnapshotRecordBuilder.java
```diff
@@ -0,0 +1,126 @@
...
+ *
+ * <p>Produces records matching the output format of ExtractNewRecordState (unwrap SMT):
+ * flat struct with column fields + metadata fields (__op, __table, __source_ts_ms, etc.).
+ *
+ * <p>This class lives in debezium-connector-common so that TableSnapshotWorker can use it
+ * directly without depending on debezium-embedded.
+ *
+ * @author Ivan Senyk
+ */
+public class SnapshotRecordBuilder {
```
Naros: Why are we doing this here rather than not using the existing TableSchema/EventDispatcher logic that takes the raw `Object[]` values and creates events?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/SnapshotTableCompletionHandler.java
```diff
@@ -0,0 +1,47 @@
...
+ * <p>Implementations receive pre-transformed SourceRecords — the configured SMT chain
+ * has already been applied by the snapshot worker. This matches the CDC path where
+ * the engine applies transforms before delivering records to the consumer.
+ *
+ * <p>Implementations can register via {@link java.util.ServiceLoader} to receive
+ * notifications when an incremental snapshot worker completes processing a table.
+ *
+ * @author Ivan Senyk
+ */
+public interface SnapshotTableCompletionHandler {
```
Naros: This bypasses things like `PostProcessor`, metrics collection, and more, which are handled by the `EventDispatcher`. So why is this SPI needed rather than using the existing Debezium common connector framework pipeline process?

It seems to me that you want to bypass much of the existing downstream pipeline, why?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/TableSnapshotWorker.java
```diff
@@ -0,0 +1,417 @@
...
+    private final OffsetContext offsetContext;
+    private final List<SnapshotTableCompletionHandler> completionHandlers;
+
+    private static volatile RecordTransformer RECORD_TRANSFORMER = null;
+
+    public static void registerRecordTransformer(RecordTransformer transformer) {
+        LoggerFactory.getLogger(TableSnapshotWorker.class)
+                .info("Registered RecordTransformer for snapshot events");
+        RECORD_TRANSFORMER = transformer;
+    }
```
Naros: None of this is ever used.

debezium-embedded/src/main/java/io/debezium/embedded/DebeziumEventFactory.java
```diff
@@ -0,0 +1,189 @@
...
+
+/**
+ * Factory to create Debezium change events from JDBC ResultSet rows.
+ *
+ * This allows us to reuse the tested EventConverter code from memiiso/debezium-server-iceberg
+ * for handling snapshot data, ensuring consistent null handling and type conversion.
+ *
+ * @author Ivan Senyk
+ */
+public class DebeziumEventFactory {
```
Naros: This frankly looks like you're trying to bypass the `Transformation` process and instead bake the raw `SourceRecord` into the desired downstream consumer format you need. I would argue this is not the most appropriate way.

A more appropriate approach would be to introduce a separate feature that fully customizes the event envelope, ensuring it satisfies all event pipeline paths (initial/blocking snapshots, streamed events, and incremental snapshots). 

But until such a time that a feature like that is available, this really should be logic that belongs solely inside a `Transformation` imo.

Naros (changes requested):
Hi @MrIvv, while I appreciate the work you've put into this PR, I believe this is a bit away from being mergeable. Can you please take some time to review my comments and provide some feedback/reasoning for my questions? Thanks.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/SnapshotTableCompletionHandler.java
MrIvv: I need to bypass EventDispatcher because the multi-threaded workers deliver records directly to the consumer handler, outside the standard ChangeEventQueue → SourceTask.poll() path. I'll integrate PostProcessor and SnapshotProgressListener metrics into the worker so the SPI path is functionally equivalent

debezium-api/pom.xml
MrIvv: Done, the comment is back.

debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
MrIvv: Done. Extracted `DEFAULT_*` constants for all 5 properties (`BATCH_FLUSH_SIZE` + 4 retry) and added `Field::isPositiveInteger` / `Field::isPositiveLong` validation. Also lowered the default batch flush size from 20_000 to 5_000 — wide-row tables (VARCHAR/TEXT/JSONB) saturate worker memory at higher values, users with narrow rows can raise it via config.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/IncrementalSnapshotRetryPolicy.java
MrIvv: Done. Now `io.debezium.util.RetryExecutor` in `debezium-util`, generalized (no `SQLException`-specific handling, accepts any `Callable`/`Runnable`).

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/RecordTransformer.java
MrIvv: You're right — this was a workaround for a sink-specific concern that doesn't belong in the snapshot pipeline. Removed entirely in the latest push.

Sink-side column casting / type conversion now goes through the standard `Transformation` / `CustomConverter` / `PostProcessor` mechanisms; the worker just produces standard Debezium envelope `SourceRecord`s via `SnapshotChangeRecordEmitter` and lets the existing pipeline handle the rest.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/TableSnapshotWorker.java
MrIvv: Gone, removed with `RecordTransformer`.

debezium-embedded/src/main/java/io/debezium/embedded/DebeziumEventFactory.java
MrIvv: Removed entirely. The JDBC-row → flat-struct logic is gone (snapshot events flow through `SnapshotChangeRecordEmitter` and the standard `Transformation` chain), and the only thing left after the refactor — a single `wrapSourceRecord(SourceRecord)` factory — was redundant: `ConverterBuilder.toFormat()` configured with `Connect` format already does the same raw `SourceRecord → EmbeddedEngineChangeEvent` wrap.

The consumer side (memiiso/debezium-server-iceberg#693) now builds the wrapper once via `ConverterBuilder` and reuses it: no new API surface added to `debezium-embedded`.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/SnapshotRecordBuilder.java
MrIvv: Refactored. The worker now uses `SnapshotChangeRecordEmitter` + `EventDispatcher.emitReadRecord()` — same path as the chunked initial snapshot. My earlier note about `EventDispatcher` being single-threaded was wrong: with each worker holding its own `TableSnapshotContext` the dispatcher works fine here. `SnapshotRecordBuilder` is now down to 66 lines, just wraps the envelope `Struct` produced by the emitter into a `SourceRecord` for delivery to the SPI handler (see reply on the SPI itself for why this last step still exists).

debezium-connector-common/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -1765,6 +1766,9 @@ public Object getColumnValue(ResultSet rs, int columnIndex, Column column, Table
      * Sets value on {@link PreparedStatement} and set explicit SQL type for it if necessary
      */
     public void setQueryColumnValue(PreparedStatement statement, Column column, int pos, Object value) throws SQLException {
+        if (value instanceof SpecialValueDecimal) {
+            value = ((SpecialValueDecimal) value).getDecimalValue().orElse(null);
```
Naros:
```suggestion
        if (value instanceof SpecialValueDecimal specialValueDecimal) {
            value = specialValueDecimal.getDecimalValue().orElse(null);
```

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/SnapshotTableCompletionHandler.java
MrIvv: Done on the metrics/post-processor side: `PostProcessorRegistry` and `SnapshotProgressListener` are now wired into `TableSnapshotWorker`, so the SPI path is functionally equivalent to CDC (transforms, post-processors, metrics).

The SPI itself stays for the `onTableSnapshotFinished(tableName)` callback — a per-table boundary signal that nothing in the existing pipeline exposes. `EventDispatcher` operates per-record with no concept of table boundaries; `SnapshotProgressListener` has `dataCollectionSnapshotCompleted` but it's observer-only (no return, no failure propagation, used for metrics/JMX).

The use case is sinks that keep a writer open across all chunks of a table and commit once at the end. Iceberg streaming flush is the concrete one — see [memiiso/debezium-server-iceberg#693](https://github.com/memiiso/debezium-server-iceberg/pull/693). Numbers measured on a production workload (PostgreSQL 16, 116 tables, ~128M rows, largest 30M):

| Metric | Per-chunk writer | Per-table streaming writer |
|---|---|---|
| Iceberg writers / table | ~1,500 | 1 (with periodic file splits) |
| Iceberg commits / table | ~1,500 | ~6-10 (one per ~512MB Parquet file) |
| Throughput | ~14K rows/min | ~80-120K rows/min |
| Peak memory / worker | ~1.5 GB | ~200-300 MB |

Same pattern fits any sink with batched-write semantics — JDBC sinks coalescing inserts into a single COPY, S3/Parquet writers, columnar formats with expensive open/close.

Alternatives I considered:

- sniff completion from offset state in the consumer → fragile under retry/rebalancing, leaks snapshot internals into every sink
- add the callback directly on `EventDispatcher` → wider blast radius (affects all event paths) for a signal only batched-write sinks need
- extend `SnapshotProgressListener` with delivery semantics → breaks the observer contract retroactively for everyone using it for metrics

A small, focused, opt-in SPI registered via `ServiceLoader` felt like the smallest change. If you see a cleaner fit inside the existing pipeline I'm open.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/ParallelIncrementalSnapshotCoordinator.java
```diff
@@ -0,0 +1,164 @@
...
+ * the chunked initial snapshot infrastructure ({@code ThreadedSnapshotExecutor}
+ * and {@code ConcurrentLinkedQueue<JdbcConnection>} pool).
+ *
+ * <p>Each table gets its own isolated window buffer to prevent thread conflicts.
+ *
+ * @author Ivan Senyk
+ * @param <P> the type of partition
+ * @param <T> the type of data collection identifier
+ */
+public class ParallelIncrementalSnapshotCoordinator<P extends Partition, T extends DataCollectionId> {
```
Naros, line 48: Take a look at `RelationalSnapshotChangeEventSource`, specifically `ThreadedSnapshotExecutor` and `PooledWork`. It seems here you're mixing concerns, and while having a coordinator that faciliates a higher-abstraction may make sense, perhaps reusing the lower-level components here reduces code duplication?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/SnapshotTableCompletionHandler.java
Naros: > Done on the metrics/post-processor side: PostProcessorRegistry and SnapshotProgressListener are now wired into TableSnapshotWorker, so the SPI path is functionally equivalent to CDC (transforms, post-processors, metrics).

This is not entirely correct.

Today, we generate events (streaming, incremental, blocking, initial) and those traverse through the EventDispatcher, get queued up into the `ChangeEventQueue`, are popped off by the runtime, passed through the transformation chain, and handled to the `ChangeEventConsumer` on the sink side, where consumer implementations like Iceberg ingest the changes.

The completion handler effectively bypasses all those bits, AFAICT, based on the PR. The events are either sent to the completion handlers (some opaque black box) our code doesn't use, or placed into the window buffer, which is never consumed. This doesn't seem to produce any functional component when enabled upstream.

I would recommend focusing on the core feature, which is adding parallel incremental snapshots, rather than layering all these extra SPIs and features that are completely unrelated to the task. In all fairness, the SPI signal and transforms you speak of are really separate PRs that can stand on their own.

Naros (changes requested):

MrIvv:
Hi @Naros, force-pushed after rebase on latest main. The branch is now 5 commits.

Summary of what changed in response to the review:

- **Pipeline alignment** (`SnapshotRecordBuilder`, SPI integration, `DebeziumEventFactory`): snapshot events flow through `SnapshotChangeRecordEmitter` + `EventDispatcher.emitReadRecord()` and have `PostProcessorRegistry` + `SnapshotProgressListener` applied. `RecordTransformer` and the flat-struct logic in `DebeziumEventFactory` are gone.
- **Code reuse** (`RetryExecutor`): moved to `debezium-util`, generalized for reuse from Debezium Server.
- **Config** (`CommonConnectorConfig`): defaults extracted to `public static final` constants with `Field::isPositive*` validation. Default batch flush size lowered from 20_000 to 5_000 for safety on wide-row tables.
- **Cleanup** (`pom.xml` revert, dead code removal): unrelated `pom.xml` change reverted, dead code removed.

The remaining `SnapshotTableCompletionHandler` SPI exists only for the per-table completion signal needed by sinks with writer-per-table semantics — concrete consumer at memiiso/debezium-server-iceberg#693, details in the inline replies.

DCO, email, and commit prefix issues from the previous push are also fixed.

Naros:
> The remaining SnapshotTableCompletionHandler SPI exists only for the per-table completion signal needed by sinks with writer-per-table semantics — concrete consumer at https://github.com/memiiso/debezium-server-iceberg/pull/693, details in the inline replies.

But I would argue this is accomplishable today without this SPI.

What I don't understand is why the Iceberg ' ChangeEventConsumer ' couldn't subscribe to one of the pre-existing notification channels about initial, blocking, and incremental snapshots, and alternate between the current event-batch handler logic and a persistent writer mode by table?

For example, Debezium emits "Incremental Snapshot started for Table ABC". The Iceberg sink receives it and adds the table "ABC" to its list using a persisted writer. As batches of events come in for table ABC, you use the persisted writer for them; they are buffered in memory or on disk, depending on your use case, and acknowledge receipt of the events, just like any other. The only difference is they're not yet written to the target system. Once Debezium emits "Incremental Snapshot finished for Table ABC", you flush the writer to disk, close it, and transition Table ABC back to the normal batch mode it operates in today.

The only question is whether or not there is currently a place in a sink for it to register for such events, e.g. is Debezium available for it to register at a given good point. Obviously, this could be done lazily in the `handleBatch` callback, but ideally, doing this earlier would be much better. If there isn't a good spot for this "sync-point", I'd say that might be worth a discussion and potentially a feature PR for DS.

This keeps what I believe are sink-related aspects where they belong, on the sink and not the producer. What you are doing right now is introducing the ability to wire something in, potentially in a non-DS environment, that won't work and makes no sense. The connector code itself should be runtime-agnostic in this regard, which is why I believe moving this work to the sink and using notifications to manage the writes makes far more architectural sense.

What do you think?

Naros:
>  I'd say that might be worth a discussion and potentially a feature PR for DS.

I talked with a colleague, and he reminded me about Debezium Server's lifecycle management.  There is already `ConnectorStartedEvent` and `ConnectorStoppedEvent` that the consumer could observe and handle this setup quite nicely. Even a [custom notification channel](https://debezium.io/documentation/reference/stable/configuration/notification.html#debezium-notification-custom-channel) that acts as a bridge between the producer & the consumer fits well here.

kmos:
> > I'd say that might be worth a discussion and potentially a feature PR for DS.
> 
> I talked with a colleague, and he reminded me about Debezium Server's lifecycle management. There is already `ConnectorStartedEvent` and `ConnectorStoppedEvent` that the consumer could observe and handle this setup quite nicely. Even a [custom notification channel](https://debezium.io/documentation/reference/stable/configuration/notification.html#debezium-notification-custom-channel) that acts as a bridge between the producer & the consumer fits well here.

Hi @MrIvv ,  thank you for your contribution. I went through your [issue](https://github.com/debezium/dbz/issues/1829) and pull request and based on your motivation:

> **No table completion signal**: Sink connectors receive chunks via onTableSnapshotCompleted() but have no way to know when a table is finished — forcing sinks that keep a writer open across chunks (e.g. Iceberg streaming flush) to either create a new writer per chunk (massive overhead, many small files) or sniff completion from offset state (fragile)

as mentioned by @Naros there is the [custom notification channel](https://debezium.io/documentation/reference/stable/configuration/notification.html#debezium-notification-custom-channel) that can help your motivation: you can observe the event `TABLE_SCAN_COMPLETED` and try to address your need.

if you can give a try and check if it fits your motivation, it may help to reduce the complexity of the contribution. What do you think?

MrIvv:
Hello @kmos , your suggestion is amazing! It’s a great point that simplifies things a lot. I’ll update the code today to include these parts as well

MrIvv:
Hi @Naros and @kmos , thanks for the patient review and for pointing me toward the right pattern.
I've taken your suggestion all the way through and force-pushed the result. Two things changed:
**1. SPI removed in favor of `TABLE_SCAN_COMPLETED` notifications**
**2. `ThreadedSnapshotExecutor` and `PooledWork` reused**
I've updated the PR description to reflect the reduced scope.
Let me know if there's anything else i can improve

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -643,6 +969,17 @@ private List<String> expandAndDedupeDataCollectionIds(List<String> dataCollectio
...
+    private boolean createDataEventsForTable(P partition, JdbcConnection connection) throws SQLException {
+        JdbcConnection originalConnection = this.jdbcConnection;
+        try {
+            this.jdbcConnection = connection;
+            return createDataEventsForTable(partition);
+        }
+        finally {
+            this.jdbcConnection = originalConnection;
+        }
+    }
```
Naros: Are we sure this is thread-safe and that at no point two threads won't fight for what connection is the member variable connection here? I think it's much safer to pass the desired connection through the call where it's needed rather than relying on this hacky swap approach.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -800,6 +1146,60 @@ protected void postIncrementalSnapshotCompleted() {
         // no-op
     }
 
+    public void shutdown() {
+        if (parallelCoordinator != null) {
+            LOGGER.info("Shutting down parallel incremental snapshot coordinator");
+            parallelCoordinator.shutdown();
+        }
+    }
```
Naros, line 1206: I don't see where this is ever called.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/ParallelIncrementalSnapshotCoordinator.java
```diff
@@ -0,0 +1,226 @@
...
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ParallelIncrementalSnapshotCoordinator.class);
+
+    private final int threadCount;
+    private final ConcurrentHashMap<T, Map<Struct, Object[]>> windowBuffers;
+    private final Queue<JdbcConnection> connectionPool;
+    private final List<JdbcConnection> allConnections;
+    private final Map<T, TableSnapshotWorker<P, T>> activeWorkers;
+    private final Queue<DataCollection<T>> pendingTables;
+    private final List<T> completedTables;
```
Naros, line 67: This `ArrayList` is written to by multiple threads and isn't guarded.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -800,6 +1146,60 @@ protected void postIncrementalSnapshotCompleted() {
...
+        if (remainingTables < 2) {
+            LOGGER.trace("[{}] Only {} table(s) remaining, using sequential read",
+                    Thread.currentThread().getName(), remainingTables);
+            return false;
+        }
+
+        LOGGER.debug("[{}] Multiple tables available ({} remaining), using parallel read",
+                Thread.currentThread().getName(), remainingTables);
+        return true;
+    }
```
Naros: This method is also never used.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -86,6 +93,9 @@ public abstract class AbstractIncrementalSnapshotChangeEventSource<P extends Par
     protected ChunkQueryBuilder<T> chunkQueryBuilder;
     protected final Map<Struct, Object[]> window = new LinkedHashMap<>();
     protected final NotificationService<P, ? extends OffsetContext> notificationService;
+    protected ParallelIncrementalSnapshotCoordinator<P, T> parallelCoordinator;
+    protected final RetryExecutor retryPolicy;
+    protected final PostProcessorRegistry postProcessorRegistry;
```
Naros: The `PostProcessorRegistry` is set but never used, so it can be removed.

debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -755,6 +755,43 @@ public static SnapshotQueryMode parse(String value, String defaultValue) {
...
+            .withDisplayName("Incremental snapshot retry max delay (ms)")
+            .withType(Type.LONG)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withDescription("Upper bound on the exponential backoff delay between retries.")
+            .withDefault(DEFAULT_INCREMENTAL_SNAPSHOT_RETRY_MAX_DELAY_MS);
+
+    public static final Field INCREMENTAL_SNAPSHOT_RETRY_BACKOFF_MULTIPLIER = Field.create("incremental.snapshot.retry.backoff.multiplier")
+            .withDisplayName("Incremental snapshot retry backoff multiplier")
+            .withType(Type.STRING)
```
Naros:
```suggestion
            .withType(Type.DOUBLE)
```
Also, add a `withValidation` that this is a valid positive numeric value, and change the default to a double value.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -361,13 +549,138 @@ protected void readChunk(P partition, OffsetContext offsetContext) throws Interr
...
+        }
+
+        final List<T> completedThisRound = Collections.synchronizedList(new ArrayList<>());
+        final Map<T, TableSnapshotWorker<P, T>> activeWorkers = parallelCoordinator.getActiveWorkers();
+        final int roundSize = Math.min(parallelCoordinator.getThreadCount(), activeWorkers.size());
+
+        DebeziumException primaryFailure = null;
+        try {
+            LOGGER.info("Creating parallel chunk round worker pool with {} worker thread(s)", roundSize);
+            ExecutorService executor = Executors.newFixedThreadPool(roundSize);
```
Naros: We are creating a new thread pool for each incremental snapshot window, which is wasteful and inefficient, particularly for large snapshots. Should we think about creating this when the snapshot is started with the desired size and cleaning this up when the snapshot ends, or when the connector is stopped?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -254,10 +387,56 @@ protected void readChunk(P partition, OffsetContext offsetContext) throws Interr
...
+                    effectiveConnection = createSnapshotConnection();
+                    createdOnDemandConnection = true;
+                }
+                catch (UnsupportedOperationException e) {
+                    LOGGER.trace("createSnapshotConnection not supported, using default connection");
+                }
+                catch (Exception e) {
+                    LOGGER.debug("Could not create snapshot connection, using default: {}", e.getMessage());
+                }
+            }
```
Naros: Can we avoid relying on an exception here to use the default connection? I assume this can be easily controlled by a flag or an abstract method that dictates whether the incremental snapshot operates in concurrent or single-threaded mode without the need for exception handling.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -800,6 +1146,60 @@ protected void postIncrementalSnapshotCompleted() {
...
+     * context after restart. Default implementation reflectively calls a
+     * {@code validateRestoredContext()} method if present and falls back to
+     * checking that at least one table remains to snapshot.
+     */
+    protected boolean validateRestoredContext(IncrementalSnapshotContext<T> context) {
+        try {
+            java.lang.reflect.Method method = context.getClass().getMethod("validateRestoredContext");
+            Boolean result = (Boolean) method.invoke(context);
+            return result != null ? result : true;
+        }
```
Naros: What purpose does this logic serve now? This seems unnecessary is it not?

debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -755,6 +755,43 @@ public static SnapshotQueryMode parse(String value, String defaultValue) {
...
+            .withDescription("Initial backoff delay before retrying a failed chunk read.")
+            .withDefault(DEFAULT_INCREMENTAL_SNAPSHOT_RETRY_INITIAL_DELAY_MS);
+
+    public static final Field INCREMENTAL_SNAPSHOT_RETRY_MAX_DELAY_MS = Field.create("incremental.snapshot.retry.max.delay.ms")
+            .withDisplayName("Incremental snapshot retry max delay (ms)")
+            .withType(Type.LONG)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withDescription("Upper bound on the exponential backoff delay between retries.")
+            .withDefault(DEFAULT_INCREMENTAL_SNAPSHOT_RETRY_MAX_DELAY_MS);
```
Naros: If these should be always positive, they should have validation applied.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -167,14 +216,62 @@ protected String getSignalTableName(String dataCollectionId) {
...
+            return parallelCoordinator.getWindowBuffer(dataCollectionId);
+        }
+        return window;
+    }
+
+    /**
+     * Returns true if any per-table window buffer (parallel mode) or the
+     * legacy single window (sequential mode) currently holds events. Required
+     * because in parallel mode the legacy {@code window} field is always empty
+     * — chunk reads write into the coordinator's per-table buffers — so a
```
Naros: The use of emdashes is very AI-centric, let's avoid these. In fact, our AGENTS.md specifically mention to avoid these types of characters, among others. I'd suggest if you're using AI that it adhere or our AI Policy and Agent rules.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -167,14 +216,62 @@ protected String getSignalTableName(String dataCollectionId) {
...
-        LOGGER.debug("Sending {} events from window buffer", window.size());
+        if (parallelCoordinator != null) {
+            sendParallelWindowEvents(partition, offsetContext);
+            return;
+        }
+
+        final T dataCollectionId = context.currentDataCollectionId().getId();
+        final Map<Struct, Object[]> currentWindow = getWindowForDataCollection(dataCollectionId);
+
+        LOGGER.debug("[{}] Sending {} events from window buffer for table {}",
```
Naros: There are a number of use cases where `Thread.currentThread().getName()` is used even on the single-threaded code path, which only adds noise and I believe is less useful than the actual parallel thread paths. Can we clean these up?

Naros (changes requested):

debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
MrIvv: I will add

debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
MrIvv: Hi Naros, regarding the validation, I checked this file: https://github.com/debezium/debezium/blob/main/debezium-config/src/main/java/io/debezium/config/Field.java#L1642. I noticed there isn't a check for isPositiveDouble, so I will go ahead and add it

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
MrIvv: Understood, I will clean them up

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
MrIvv: I can easily change this from per-window to coordinator setup. However, a per-signal lifecycle allocation is certainly the best option. Should we do this in the current PR?

MrIvv (commented):
Hello Naros, I've made a few changes to improve the thread safety of this code, mainly focusing on the lifecycle of the connections

MrIvv:
Hello @Naros, I reviewed all CI jobs and found some regressions in my code.
I've added a fix in this new commit for easier review, I think all critical issues have now been covered

Naros:
Thanks for the update @MrIvv, I've started CI. I'll review it once CI finishes.

Naros:
Hi @MrIvv can you just check that the failure in MariaDB isn't related.

MrIvv:
Hi @Naros,
force-pushed one extra fix into the review-iteration commit:
TableSnapshotWorker.isChunkPositionComplete now uses Arrays.equals instead of a per-key compareTo loop.

On the MariaDB CI failure (BinlogReadOnlyIncrementalSnapshotIT.filteredEvents) I do not believe it is caused by this PR. The stack is entirely in MariaDB-specific code that this PR does not touch:

  NPE: this.highWatermark is null
    at MariaDbReadOnlyIncrementalSnapshotContext.serverStreamSetChanged:121
    at MariaDbReadOnlyIncrementalSnapshotContext.hasServerIdentifierChanged:63
    at BinlogReadOnlyIncrementalSnapshotChangeEventSource.emitWindowClose:123

The exception is swallowed by the catch in AbstractIncrementalSnapshotChangeEventSource.addDataCollectionNamesToSnapshot ("Error while executing incremental snapshot ... skipping and continuing streaming"), so the snapshot is silently abandoned and the consumer
  times out with 0 records. 
Only 1 of 4 MariaDB matrix variants failed, the others passed on the same code, consistent with a timing-sensitive race on highWatermark initialization rather than a regression from this PR.

Happy to dig deeper if useful

MrIvv:
Hi @Naros, I noticed the CI/CD jobs failed, but I'm not sure if it's related to my PR. Have you had a chance to look into it?

kmos:
@MrIvv please rebase with main branch

jpechane:
@MrIvv Could you please fix the formatting?

Naros:
Hi @MrIvv, could you give this one more rebase, please. Thanks.

MrIvv:
Hi @Naros, done

MrIvv:
Rebased on current main. 
The rebase shrank the change: the boundary-condition building and null-safe triangular binding that this PR originally carried in CascadingOrBoundaryConditions are now provided by main (generateBoundaryParams/bindBoundaryParams), so the PR adopts those helpers and only passes the per-worker connection through readTableChunkStatement. 
Also added a third commit that closes an observability gap: the parallel path now emits the same per-table IN_PROGRESS notifications the sequential path emits at every window close

debezium-config/src/main/java/io/debezium/config/Field.java
```diff
@@ -1661,6 +1661,22 @@ public static int isDouble(Configuration config, Field field, ValidationOutput p
...
+        if (value == null) {
+            return 0;
+        }
+        try {
+            if (Double.parseDouble(value) > 0) {
+                return 0;
+            }
+        }
+        catch (NumberFormatException e) {
+        }
```
Naros, line 1675:
```suggestion
        catch (Throwable e) {
        }
```

debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -793,6 +793,60 @@ public static SnapshotQueryMode parse(String value, String defaultValue) {
             .withDefault(1024)
             .withValidation(Field::isNonNegativeInteger);
 
+    public static final int DEFAULT_INCREMENTAL_SNAPSHOT_RETRY_MAX_ATTEMPTS = 5;
+    public static final long DEFAULT_INCREMENTAL_SNAPSHOT_RETRY_INITIAL_DELAY_MS = 1000L;
+    public static final long DEFAULT_INCREMENTAL_SNAPSHOT_RETRY_MAX_DELAY_MS = 60_000L;
+    public static final double DEFAULT_INCREMENTAL_SNAPSHOT_RETRY_BACKOFF_MULTIPLIER = 2.0;
```
Naros, line 799: These and other added `private static final` constants should be placed adjacent to the other values like `DEFAULT_MAX_QUEUE_SIZE`, rather than randomly inline. This would also apply to the new field `DEFAULT_INCREMENTAL_SNAPSHOT_POOL_RELEASE_DELAY_MS` added further below.

debezium-connector-common/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -1794,6 +1795,9 @@ public Object getColumnValue(ResultSet rs, int columnIndex, Column column, Table
      * Sets value on {@link PreparedStatement} and set explicit SQL type for it if necessary
      */
     public void setQueryColumnValue(PreparedStatement statement, Column column, int pos, Object value) throws SQLException {
+        if (value instanceof SpecialValueDecimal specialValueDecimal) {
+            value = specialValueDecimal.getDecimalValue().orElse(null);
+        }
```
Naros, line 1800: This specific change seems unrelated to parallel incremental snapshots. Is this fixing something unrelated that you discovered? If so, this should ideally be done in a separate issue & PR.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -185,6 +276,13 @@ protected void sendEvent(P partition, EventDispatcher<P, T> dispatcher, OffsetCo
                 dispatcher.getIncrementalSnapshotChangeEventReceiver(dataListener));
     }
 
+    private void sendEventForDataCollection(P partition, OffsetContext offsetContext, T dataCollectionId, Object[] row) throws InterruptedException {
```
Naros, line 279: This method never calls `context.sendEvent(keyFromRow(row))`. Shouldn't it?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -254,10 +389,59 @@ protected void readChunk(P partition, OffsetContext offsetContext) throws Interr
...
+            readParallelRound(partition, offsetContext);
+            return;
+        }
+
+        JdbcConnection effectiveConnection = null;
+        boolean acquiredPoolConnection = false;
+        boolean createdOnDemandConnection = false;
+
         try {
+            if (parallelCoordinator != null) {
```
Naros, line 419: There are a lot of places in the code where there's now a notion of "here we do something because it's parallel, and for non-parallel we don't," or the implementation between the two now differs. It raises the question of whether this should be behind a strategy pattern used by the event source based on whether parallel or non-parallel is used.

documentation/modules/ROOT/pages/connectors/sqlserver.adoc
```diff
@@ -3427,6 +3427,33 @@ Increasing the chunk size provides greater efficiency, because the snapshot runs
 However, larger chunk sizes also require more memory to buffer the snapshot data.
 Adjust the chunk size to a value that provides the best performance in your environment.
 
+|[[sqlserver-property-incremental-snapshot-pool-release-delay-ms]]<<sqlserver-property-incremental-snapshot-pool-release-delay-ms, `+incremental.snapshot.pool.release.delay.ms+`>>
+|`60000` (1 minute)
+|Milliseconds to wait after the last active or pending table completes before closing the parallel snapshot connection pool.
+A new signal arriving within this window cancels the pending release, so back-to-back signal bursts reuse the same pool instead of paying open/close overhead on every signal.
+Set to `0` to release the pool as soon as no work remains.
+The setting only has an effect when `snapshot.max.threads` is greater than 1, since the parallel pool is only allocated in that mode.
```
Naros, line 3435: All of these parallel incremental snapshot properties are documented for all connectors, but the only way they can be used is if the connector implements a read-only incremental snapshot feature. Therefore, I'm not sure it makes sense to document these for connectors that don't support this.

This would mean we'd expect an implementation for PostgreSQL, MariaDB, and MySQL, yet the only implementation the PR includes is PostgreSQL. Do you intend to supply the others?

debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -1632,6 +1691,11 @@ protected CommonConnectorConfig(Configuration config, int defaultSnapshotFetchSi
         this.queryFetchSize = config.getInteger(QUERY_FETCH_SIZE);
         this.incrementalSnapshotChunkSize = config.getInteger(INCREMENTAL_SNAPSHOT_CHUNK_SIZE);
         this.incrementalSnapshotAllowSchemaChanges = config.getBoolean(INCREMENTAL_SNAPSHOT_ALLOW_SCHEMA_CHANGES);
+        this.incrementalSnapshotRetryMaxAttempts = config.getInteger(INCREMENTAL_SNAPSHOT_RETRY_MAX_ATTEMPTS);
+        this.incrementalSnapshotRetryInitialDelayMs = config.getLong(INCREMENTAL_SNAPSHOT_RETRY_INITIAL_DELAY_MS);
+        this.incrementalSnapshotRetryMaxDelayMs = config.getLong(INCREMENTAL_SNAPSHOT_RETRY_MAX_DELAY_MS);
+        this.incrementalSnapshotRetryBackoffMultiplier = Double.parseDouble(config.getString(INCREMENTAL_SNAPSHOT_RETRY_BACKOFF_MULTIPLIER));
```
Naros, line 1697:
```suggestion
        this.incrementalSnapshotRetryBackoffMultiplier = config.getDouble(INCREMENTAL_SNAPSHOT_RETRY_BACKOFF_MULTIPLIER);
```

debezium-config/src/main/java/io/debezium/config/Field.java
```diff
@@ -1661,6 +1661,22 @@ public static int isDouble(Configuration config, Field field, ValidationOutput p
         return 0;
     }
 
+    public static int isPositiveDouble(Configuration config, Field field, ValidationOutput problems) {
```
Naros, line 1664: Should we check `Double.isFinite()` here and reject `Infinity`, which would otherwise validate as a positive double?

debezium-util/src/main/java/io/debezium/util/RetryExecutor.java
```diff
@@ -0,0 +1,181 @@
...
+import java.util.concurrent.ThreadLocalRandom;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+/**
+ * Retry executor with exponential backoff and jitter. {@code maxAttempts = -1}
+ * disables the attempt cap and retries indefinitely.
+ */
+public class RetryExecutor {
```
Naros, line 19: Now that we have `RetryingCallable` and `RetryingRunnable`, I wonder whether it would make sense to refine them and see whether they replace `RetryExecutor` at this point. There seems to be a significant amount of overlap and duplication that can now be eliminated.

Naros (changes requested):
Hi @MrIvv, besides the obvious here that this needs unit and integration tests, I've added a few questions and points that still need some improvements.

debezium-connector-common/src/main/java/io/debezium/jdbc/JdbcConnection.java
MrIvv: yes, i already done one fast pr for this.

debezium-util/src/main/java/io/debezium/util/RetryExecutor.java
MrIvv: Okay, I can certainly work on this refactoring. Maybe we can do it in a separate PR? The idea would be to first add the necessary integrations to the current retry classes, and then refactor this part

Naros:
Hi @MrIvv could we get a rebase on the PR?

MrIvv:
Hi @Naros, i've rebased the branch, but I'll hold off on further changes here until the other PR (consolidating the retries) is closed. 
Is this the correct approach?

PR #7289, opened 2026-04-03
debezium/dbz#1760 Disable auto-commit: MySQL/MariaDB-ChunkedSnapshotIT

Naros:
Fixes debezium/dbz#1760

## Description
Disable auto-commit for MySQL/MariaDB ChunkedSnapshot integration tests.

## PR Checklist

- [x] I have read the [contribution guidelines](https://github.com/debezium/debezium/blob/main/CONTRIBUTING.md) and the [governance document](https://github.com/debezium/governance/blob/main/GOVERNANCE.md) on PR expectations.
- [x] Minimal changes to code not directly related to your change (e.g. no unnecessary formatting changes or refactoring to existing code)
- [x] One feature/change per PR unless tightly coupled
- [x] Do a rebase on upstream `main`

PR #7368, opened 2026-04-18
debezium/dbz#1832 Remove MySQL `never` snapshot mode

### LINKED ISSUE #1832: Remove the `never` option from `snapshot.mode` for MySQL connectors
The MySQL connector has long supported a `never` snapshot mode, which replays all available binlog events starting from the oldest entry rather than taking an initial snapshot. During the Debezium 3.0 release, we opted to keep this mode temporarily with the understanding that it would be deprecated and eventually removed.

This issue covers four pieces of work:

1. **Remove `never` as a valid snapshot mode.** The connector should fail to start with a clear configuration error if `never` is specified.

2. **Add an internal, undocumented escape hatch** that re-enables the legacy `never` behavior. This gives us a way to unblock users on a case-by-case basis without maintaining it as a public API.

3. **Migrate existing tests off `never`.** Any tests currently relying on the `never` mode should be updated to use one of the remaining valid snapshot modes.

4. **Add a single regression test for the escape hatch.** This test should validate that the legacy behavior still works when the internal option is enabled. It should be disabled by default and run manually as needed, to avoid interference with the rest of the test suite.

Naros:
Fixes debezium/dbz#1832

## Description
Removes `never` snapshot mode; replaced by config-based

* This removes the `never` snapshot mode for MySQL
* Users are expected to use the configuration-based snapshot to mimic `never`
* Adjusts most tests to use `initial`
* Introduces a legacy test for supporting configuration-based mimiced `never` behavior

## PR Checklist

Required for #7289.


Naros:
The MariaDB test failures are related, looking :eyes:

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogConnectorConfig.java
```diff
@@ -488,15 +488,28 @@ default boolean useConsistentSnapshotTransaction() {
...
-                    + "'never': The connector does not run a snapshot. Upon first startup, the connector immediately begins reading from the beginning of the binlog. "
-                    + "The 'never' mode should be used with care, and only when the binlog is known to contain all history.");
+                    + "'initial_only': The connector performs a snapshot as it does for the 'initial' option, but after the connector completes the snapshot, it stops, and does not stream changes from the binlog.");
+
+    /**
+     * Internal-only property that permits the use of {@code snapshot.mode=never}.
+     * This mode is not supported for production use; it exists solely to allow targeted
+     * testing of the historical binlog-from-beginning behaviour.
+     */
+    public static final Field SNAPSHOT_MODE_NEVER_ALLOWED = Field.createInternal("snapshot.mode.never.allowed")
```
mfvitale, line 505: Since the `never` snapshot mode was marked for removal in 3.0, should we just remove it and then if someone want to still use this kind of mode can just implement a custom snapshot mode or use the `io.debezium.snapshot.mode.ConfigurationBasedSnapshotter`?

So we can just document it and avoiding to add another configuration?

vjuranek: If marked for removal in 3.0, IMHO we should keep it in 3.X and remove in 4.0 (or later). No strong opinion about this, this is just my understanding about backward compatibility guarantees in Debezium (and lots of other project - mark something deprecated/for removal in X.Y and remove in next major (X+1).Z)

vjuranek (approved):
Overall LGTM

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogConnectorConfig.java
Naros: @mfvitale I think the only gotcha with either of those approaches is the `instanceof` check in `BinlogStreamingChangeEventSource#execute`:
```java
        if (!(snapshotterService.getSnapshotter() instanceof NeverSnapshotter)) {
            schema.assureNonEmptySchema();
        }
```
There wouldn't be a way to skip this check otherwise, I don't think.

mfvitale: > @mfvitale I think the only gotcha with either of those approaches is the `instanceof` check in `BinlogStreamingChangeEventSource#execute`:
> 
> ```java
>         if (!(snapshotterService.getSnapshotter() instanceof NeverSnapshotter)) {
>             schema.assureNonEmptySchema();
>         }
> ```
> 
> There wouldn't be a way to skip this check otherwise, I don't think.

Hum why this should be specific to binlog and snapshot mode `never`? We call `assureNonEmptySchema` also in https://github.com/debezium/debezium/blob/e89689a3b93022c216034ecc77135499a30c9a74/debezium-connector-common/src/main/java/io/debezium/pipeline/ChangeEventSourceCoordinator.java#L322

Maybe it can be checked in any case also when snapshot is skipped? Seems that this message is more a warning to be in the start of the streaming phase. Isn't it?

Naros: It is, and I suppose we could just emit it in that case.  :thinking:

mfvitale: > If marked for removal in 3.0, IMHO we should keep it in 3.X and remove in 4.0 (or later). No strong opinion about this, this is just my understanding about backward compatibility guarantees in Debezium (and lots of other project - mark something deprecated/for removal in X.Y and remove in next major (X+1).Z)

I agree with the process but the mark for removal here I think was intended to drop in 3.0 since this comment was added into 2.6.

mfvitale: > It is, and I suppose we could just emit it in that case. 🤔

@Naros So in that case maybe we can just drop the `never` and use a configuration based for tests or document the possible alternative?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogConnectorConfig.java
```diff
@@ -965,4 +978,19 @@ public long getBinlogNetWriteTimeout() {
...
         return config.getLong(BINLOG_NET_READ_TIMEOUT);
     }
+
+    /**
+     * Validates that {@link SnapshotMode#NEVER} is not used unless the internal opt-in property
+     * ({@link #SNAPSHOT_MODE_NEVER_ALLOWED}) is explicitly set to {@code true}.
+     */
+    protected static int validateSnapshotModeNever(Configuration config, Field field, ValidationOutput problems) {
+        if (SnapshotMode.NEVER == SnapshotMode.parse(config.getString(field))) {
+            if (!config.getBoolean(SNAPSHOT_MODE_NEVER_ALLOWED.name(), false)) {
```
jpechane: @Naros One question, this valu is problematic only if the snapshot needs to be taken. When the snapshot is already done we don't need to fail.
How about emitting a warning in case we are going directly to streaming and throw an error when we need to take the snapshot?
The location of the valdiation will be probably elsewhere.
And we would do hard validation in 4.x here, WDYT?

jpechane (approved):

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogConnectorConfig.java
jpechane: The `NEVER` can definitely be dropped.
I added https://github.com/debezium/debezium/pull/7368#discussion_r3116970132 as a proposal to further change this into a soft landing.
@mfvitale If there is a configuration based option then it would be even better than the  new property.

mfvitale: @Naros @jpechane 

```
snapshot.mode=configuration_based
snapshot.mode.configuration.based.snapshot.data=false
snapshot.mode.configuration.based.snapshot.schema=false
snapshot.mode.configuration.based.start.stream=true
snapshot.mode.configuration.based.snapshot.on.schema.error=false
snapshot.mode.configuration.based.snapshot.on.data.error=false
```

this configuration should enable the same behavior of  `never`.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogConnectorConfig.java
```diff
@@ -173,11 +173,6 @@ public enum SnapshotMode implements EnumeratedValue {
          * otherwise some events during the gap may be processed with an incorrect schema and corrupted.
          */
         RECOVERY("recovery"),
-        /**
-         * Never perform a snapshot and only read the binlog. This assumes the binlog contains all the history of those
-         * databases and tables that will be captured.
-         */
-        NEVER("never"),
```
mfvitale, line 180: Should we also remove `io.debezium.snapshot.mode.NeverSnapshotter`?

Naros: Yes I am working on that now, just trying to get to a clean pass on tests.

Naros: Yes I am working on that now, just trying to get to a clean pass on tests.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
```diff
@@ -174,9 +173,8 @@ public BinlogStreamingChangeEventSource(BinlogConnectorConfig connectorConfig,
 
     @Override
     public void execute(ChangeEventSourceContext context, P partition, O offsetContext) throws InterruptedException {
-        if (!(snapshotterService.getSnapshotter() instanceof NeverSnapshotter)) {
-            schema.assureNonEmptySchema();
-        }
+        schema.assureNonEmptySchema();
+
```
Naros, line 177: I thought about checking if the snapshotter was config-based and then seeing if the condition ("snapshot_data=false, snapshot_schema=false") was true and skip the logged warning.

However, that would require passing in dummy values here since we don't have all the context, so I opted to just log the warning, considering its unlikely to be an issue for 80/20 or even 90/10 of users.  wdyt?

jpechane (approved):
@Naros Nice job, feel free to merge yourself.

PR #7374, opened 2026-04-20
debezium/dbz#579 Handle LOB_WRITE/LOB_TRIM/LOB_ERASE events in XStream adapter

### LINKED ISSUE #579: Unify Xstream behavior with LogMiner when using DBMS_LOB methods [DBZ-4741]
Migrated from [DBZ-4741](https://issues.redhat.com/browse/DBZ-4741)

In DBZ-4366, LogMiner adapter was adjusted to emit multiple {{DBSM_LOB}} method calls to methods such as {{APPEND}} as a single event rather than individual events.  This allows the consumer to receive the data for LOB package methods as a single event.

Unfortunately, Xstream emits these as separate events.  We've documented that this is a nuanced behavior of Xstream currently but would like to look into how to unify the behavior and avoid this nuance if possible.

pyaroslav:
Fixes debezium/dbz#579

## Description

Oracle XStream emits `LOB_WRITE`, `LOB_TRIM`, and `LOB_ERASE` command-type LCRs for `DBMS_LOB.WRITE / WRITEAPPEND / TRIM / ERASE` operations. The XStream adapter previously handled none of them:

- `XStreamChangeRecordEmitter.getOperation()` — `default:` branch threw `IllegalArgumentException`
- `LcrEventHandler.processRowLCR()` — `LOB_ERASE` logged a warning; `LOB_WRITE` / `LOB_TRIM` fell through unhandled

In production this surfaced as `ORA-26824: user-defined XStream callback error`, a crashed streaming task, and a stalled outbound server on any schema that uses the `DBMS_LOB` package (very common in enterprise Oracle apps with CLOB/BLOB columns).

This PR aligns the XStream adapter with the LogMiner adapter (cf. DBZ-4366) by treating these events as regular DML with a LOB re-read:

### `XStreamChangeRecordEmitter`
- `LOB_WRITE` / `LOB_TRIM` / `LOB_ERASE` now map to `Operation.UPDATE` in `getOperation()`.
- New `operationOverride` field lets the caller force `Operation.CREATE` when the LOB delivery immediately follows an `INSERT` for the same `ROWID`+transaction (XStream splits large-LOB inserts into an `INSERT` LCR followed by one or more `LOB_WRITE` chunks — without the override these would be mis-reported as `UPDATE`).
- The old constructor is preserved; a new one takes the override (`null` preserves prior behavior).

### `LcrEventHandler`
- New optional `OracleConnection` parameter (overloaded constructor keeps the original signature working with `jdbcConnection=null`).
- Tracks `INSERT` `ROWID`s per transaction, cleared on commit, so a following `LOB_WRITE` / `LOB_TRIM` is emitted as `CREATE`.
- Routes `LOB_WRITE` / `LOB_TRIM` / `LOB_ERASE` with chunk data through the existing chunk-assembly path. Without chunk data they are re-selected and dispatched directly.
- After chunk assembly for LOB events, re-selects the current LOB value(s) from the source database by PK so downstream consumers receive the complete value, not the partial delta carried by the `LOB_WRITE` chunk. Graceful fallbacks: if the table has no PK, or if the re-select fails, the partial chunk value is used and a warning is logged. This mirrors how `ReselectColumnsPostProcessor` is already used from LogMiner.

### `XstreamStreamingChangeEventSource`
- Passes the existing `jdbcConnection` into `LcrEventHandler` so the re-read above can run on the original CDC connection without opening another session.

### Backward compatibility

- Callers using the non-`jdbcConnection` `LcrEventHandler` constructor keep the prior behavior (partial chunks, no re-read).
- The `getOperation()` change is additive: these event types previously threw, so no caller could have been depending on a specific exception type from them.

### Testing

Validated against Oracle 19c (19.28) with Debezium 3.5.0 + XStream outbound against a ~1,693-table Curam schema with frequent CLOB/BLOB activity driven by `DBMS_LOB.WRITE` / `WRITEAPPEND` / `TRIM`, for several days at sustained DML. After this patch: no `ORA-26824`, no streaming-task crashes on LOB events, and LOB column values in Kafka match the source after each event (including partial updates).

## PR Checklist

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
@@ -327,11 +327,25 @@ class LcrEventHandler implements XStreamLCRCallbackHandler {
     }
 
     /**
+     * Pins the streaming JDBC connection's session to the configured PDB, when the
+     * connector is configured against a CDB+PDB topology. Called once after construction
+     * and before any LCRs are processed, while the connection is already active from the
+     * upstream XStream attach. The reconnect branch in {@link #getConnection()} re-pins
+     * the session if the underlying connection has to be reopened later (e.g. after an
+     * Oracle restart).
+     */
+    void init() throws SQLException {
+        if (connectorConfig.isUsingPluggableDatabase()) {
+            jdbcConnection.setSessionToPdb(connectorConfig.getPdbName());
+        }
+    }
+
+    /**
      * Returns the streaming JDBC connection ready for an out-of-bands query against the
      * captured database. Reconnects the underlying connection if it is currently
-     * disconnected (e.g. after an Oracle restart) and pins the session to the configured
-     * PDB on (re)connection. Reused across out-of-bands callbacks (DDL fetch, LOB
-     * reselect) so we don't open and tear down a JDBC connection per LCR.
+     * disconnected (e.g. after an Oracle restart) and re-pins the session to the
+     * configured PDB on reconnection. Reused across out-of-bands callbacks (DDL fetch,
+     * LOB reselect) so we don't open and tear down a JDBC connection per LCR.
      *
      * <p>The connection is dedicated to the XStream change-event source — the snapshot
      * phase has its own connection — so swapping its session to a PDB once is safe.
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
@@ -63,7 +63,6 @@ class LcrEventHandler implements XStreamLCRCallbackHandler {
     private final boolean tablenameCaseInsensitive;
     private final XstreamStreamingChangeEventSource eventSource;
     private final XStreamStreamingChangeEventSourceMetrics streamingMetrics;
-    private final OracleConnection jdbcConnection;
     private final Map<String, ChunkColumnValues> columnChunks;
     private RowLCR currentRow;
 
@@ -71,8 +70,7 @@ class LcrEventHandler implements XStreamLCRCallbackHandler {
                     EventDispatcher<OraclePartition, TableId> dispatcher, Clock clock,
                     OracleDatabaseSchema schema, OraclePartition partition, OracleOffsetContext offsetContext,
                     boolean tablenameCaseInsensitive, XstreamStreamingChangeEventSource eventSource,
-                    XStreamStreamingChangeEventSourceMetrics streamingMetrics,
-                    OracleConnection jdbcConnection) {
+                    XStreamStreamingChangeEventSourceMetrics streamingMetrics) {
         this.connectorConfig = connectorConfig;
         this.errorHandler = errorHandler;
         this.dispatcher = dispatcher;
@@ -83,7 +81,6 @@ class LcrEventHandler implements XStreamLCRCallbackHandler {
         this.tablenameCaseInsensitive = tablenameCaseInsensitive;
         this.eventSource = eventSource;
         this.streamingMetrics = streamingMetrics;
-        this.jdbcConnection = jdbcConnection;
         this.columnChunks = new LinkedHashMap<>();
     }
 
@@ -456,7 +453,7 @@ class LcrEventHandler implements XStreamLCRCallbackHandler {
      * fallback logic with the {@code ReselectColumnsPostProcessor} path.
      */
     private void reselectLobValues(RowLCR row, Table table, Map<String, Object> chunkValues) {
-        if (chunkValues.isEmpty() || jdbcConnection == null) {
+        if (chunkValues.isEmpty()) {
             return;
         }
 
@@ -501,12 +498,20 @@ class LcrEventHandler implements XStreamLCRCallbackHandler {
 
         final List<String> lobColumnNames = new ArrayList<>(chunkValues.keySet());
 
-        try {
+        // Use a separate, short-lived connection for the reselect — pinned to the PDB the
+        // table lives in. The streaming JDBC connection cannot be reused: it is bound to
+        // the CDB root in CDB+PDB deployments and would not see the user table. This
+        // mirrors getTableMetadataDdl above and BaseChangeRecordEmitter's reselect path.
+        final String pdbName = connectorConfig.getPdbName();
+        try (OracleConnection connection = new OracleConnection(connectorConfig, false)) {
+            if (!Strings.isNullOrBlank(pdbName)) {
+                connection.setSessionToPdb(pdbName);
+            }
             // reselectColumns advances the ResultSet cursor internally before invoking
             // the consumer, so we read column values directly without calling rs.next().
             // A return of `false` means no matching row was found — typically the row
             // was deleted between the LCR emission and our reselect.
-            final boolean found = jdbcConnection.reselectColumns(table, lobColumnNames, pkColumns, pkValues, null, rs -> {
+            final boolean found = connection.reselectColumns(table, lobColumnNames, pkColumns, pkValues, null, rs -> {
                 for (String colName : lobColumnNames) {
                     final Column column = table.columnWithName(colName);
                     if (column == null) {
@@ -533,10 +538,14 @@ class LcrEventHandler implements XStreamLCRCallbackHandler {
             }
         }
         catch (SQLException e) {
-            LOGGER.warn("Failed to reselect LOB values for table {} (tx={}, scn={}, rowId={}): {}. "
-                    + "Emitting partial chunk data.",
-                    table.id(), row.getTransactionId(), offsetContext.getScn(),
-                    row.getAttribute("ROW_ID"), e.getMessage());
+            // Match BaseChangeRecordEmitter.emitUpdateAsPrimaryKeyChangeRecord — surface
+            // the failure rather than emit partial chunk data with UNAVAILABLE_VALUE
+            // placeholders. A silent fallback masks real configuration problems
+            // (wrong PDB, missing privileges, table not yet visible) and produces
+            // wire records that fail downstream assertions in non-obvious ways.
+            throw new DebeziumException("Failed to reselect LOB values for table " + table.id()
+                    + " (tx=" + row.getTransactionId() + ", scn=" + offsetContext.getScn()
+                    + ", rowId=" + row.getAttribute("ROW_ID") + ")", e);
         }
     }
 
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
@@ -5,23 +5,21 @@
  */
 package io.debezium.connector.oracle.xstream;
 
-import java.sql.Connection;
-import java.sql.PreparedStatement;
-import java.sql.ResultSet;
 import java.sql.SQLException;
 import java.time.Instant;
+import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.stream.Collectors;
+import java.util.Set;
 
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import io.debezium.DebeziumException;
 import io.debezium.connector.oracle.OracleConnection;
-import io.debezium.data.Envelope.Operation;
 import io.debezium.connector.oracle.OracleConnection.NonRelationalTableException;
 import io.debezium.connector.oracle.OracleConnectorConfig;
 import io.debezium.connector.oracle.OracleDatabaseSchema;
@@ -30,6 +28,7 @@ import io.debezium.connector.oracle.OraclePartition;
 import io.debezium.connector.oracle.OracleSchemaChangeEventEmitter;
 import io.debezium.connector.oracle.OracleValueConverters;
 import io.debezium.connector.oracle.xstream.XstreamStreamingChangeEventSource.PositionAndScn;
+import io.debezium.data.Envelope.Operation;
 import io.debezium.pipeline.ErrorHandler;
 import io.debezium.pipeline.EventDispatcher;
 import io.debezium.relational.Column;
@@ -39,6 +38,7 @@ import io.debezium.util.Clock;
 import io.debezium.util.Strings;
 
 import oracle.streams.ChunkColumnValue;
+import oracle.streams.ColumnValue;
 import oracle.streams.DDLLCR;
 import oracle.streams.DefaultRowLCR;
 import oracle.streams.LCR;
@@ -68,20 +68,18 @@ class LcrEventHandler implements XStreamLCRCallbackHandler {
     private final XStreamStreamingChangeEventSourceMetrics streamingMetrics;
     private final OracleConnection jdbcConnection;
     private final Map<String, ChunkColumnValues> columnChunks;
-    // Track INSERT operations per ROW_ID+txId so that LOB_WRITE/LOB_TRIM following an INSERT
-    // can be correctly mapped to Operation.CREATE instead of Operation.UPDATE (DBZ-4741).
-    // Key: "ROW_ID|transactionId", cleared on COMMIT.
-    private final Map<String, Boolean> pendingInsertRows;
+    // DBZ-4741: track INSERTs in the *current* transaction only so that
+    // LOB_WRITE / LOB_TRIM / LOB_ERASE events that belong to that same
+    // transaction can be mapped to Operation.CREATE (the LOB data is part
+    // of the INSERT from the user's point of view, not a later UPDATE).
+    // We deliberately scope by the active transaction id instead of a
+    // global Map<rowId+txId,..> so the state is self-evicting: as soon as
+    // we observe an LCR from a different transaction, the set is cleared,
+    // even if the COMMIT LCR for the previous transaction was not
+    // received (e.g. if XStream delivered them out of order).
+    private String currentTxId;
+    private final Set<String> currentTxnInsertedRowIds = new HashSet<>();
     private RowLCR currentRow;
-
-    LcrEventHandler(OracleConnectorConfig connectorConfig, ErrorHandler errorHandler,
-                    EventDispatcher<OraclePartition, TableId> dispatcher, Clock clock,
-                    OracleDatabaseSchema schema, OraclePartition partition, OracleOffsetContext offsetContext,
-                    boolean tablenameCaseInsensitive, XstreamStreamingChangeEventSource eventSource,
-                    XStreamStreamingChangeEventSourceMetrics streamingMetrics) {
-        this(connectorConfig, errorHandler, dispatcher, clock, schema, partition, offsetContext,
-                tablenameCaseInsensitive, eventSource, streamingMetrics, null);
-    }
 
     LcrEventHandler(OracleConnectorConfig connectorConfig, ErrorHandler errorHandler,
                     EventDispatcher<OraclePartition, TableId> dispatcher, Clock clock,
@@ -101,7 +99,6 @@ class LcrEventHandler implements XStreamLCRCallbackHandler {
         this.streamingMetrics = streamingMetrics;
         this.jdbcConnection = jdbcConnection;
         this.columnChunks = new LinkedHashMap<>();
-        this.pendingInsertRows = new HashMap<>();
     }
 
     @Override
@@ -156,56 +153,17 @@ class LcrEventHandler implements XStreamLCRCallbackHandler {
     }
 
     private void processRowLCR(RowLCR row) throws InterruptedException {
-        // Track INSERT operations so LOB_WRITE/LOB_TRIM following an INSERT for the same row
-        // can be correctly mapped to Operation.CREATE instead of Operation.UPDATE.
-        // For large LOBs, XStream delivers: INSERT LCR, then LOB_WRITE LCR(s) with the data.
+        // Observe transaction boundaries and track per-txn INSERTs so LOB ops
+        // that share a txId with a tracked INSERT (i.e. the LOB data for the
+        // INSERT, delivered as separate chunks by XStream) emit Operation.CREATE
+        // rather than Operation.UPDATE.
+        resetTxScopeIfNewTransaction(row);
         if (RowLCR.INSERT.equals(row.getCommandType())) {
             final Object rowId = row.getAttribute("ROW_ID");
             if (rowId != null) {
-                final String key = rowId.toString() + "|" + row.getTransactionId();
-                pendingInsertRows.put(key, Boolean.TRUE);
+                currentTxnInsertedRowIds.add(rowId.toString());
                 LOGGER.trace("Tracking INSERT for row {} in tx {}", rowId, row.getTransactionId());
             }
-        }
-
-        // Handle LOB WRITE, LOB TRIM, and LOB ERASE as DML operations (DBZ-4741).
-        // These are emitted by DBMS_LOB.WRITE/WRITEAPPEND/TRIM/ERASE calls.
-        // LOB_WRITE with chunks: cache row, process chunks, then re-read full LOB value.
-        // LOB_TRIM/LOB_ERASE without chunks: re-read LOB value directly and dispatch.
-        if (row.getCommandType().equals(RowLCR.LOB_WRITE)
-                || row.getCommandType().equals(RowLCR.LOB_TRIM)
-                || row.getCommandType().equals(RowLCR.LOB_ERASE)) {
-            if (row.hasChunkData()) {
-                LOGGER.debug("Processing {} for table '{}' as UPDATE with chunk data.",
-                        row.getCommandType(), row.getObjectName());
-                currentRow = row;
-            }
-            else {
-                // LOB_TRIM (and rarely LOB_WRITE) without chunk data: dispatch immediately.
-                // Re-read the LOB value directly from the database to get the trimmed value.
-                LOGGER.debug("Processing {} without chunk data for table '{}', re-reading LOB and dispatching.",
-                        row.getCommandType(), row.getObjectName());
-                if (jdbcConnection != null) {
-                    // Build chunkValues with LOB columns for the re-read
-                    TableId tableId = getTableId(row);
-                    Table table = schema.tableFor(tableId);
-                    if (table != null) {
-                        Map<String, Object> lobValues = new HashMap<>();
-                        for (Column column : schema.getLobColumnsForTable(table.id())) {
-                            lobValues.put(column.name(), OracleValueConverters.UNAVAILABLE_VALUE);
-                        }
-                        reselectLobValues(row, lobValues);
-                        dispatchDataChangeEvent(row, lobValues);
-                    }
-                    else {
-                        dispatchDataChangeEvent(row, null);
-                    }
-                }
-                else {
-                    dispatchDataChangeEvent(row, null);
-                }
-            }
-            return;
         }
 
         if (row.hasChunkData()) {
@@ -224,9 +182,12 @@ class LcrEventHandler implements XStreamLCRCallbackHandler {
         LOGGER.debug("Processing DML event {}", lcr);
 
         if (RowLCR.COMMIT.equals(lcr.getCommandType())) {
-            // Clear pending INSERT tracking on COMMIT — LOB operations in subsequent
-            // transactions should not be mapped to CREATE.
-            pendingInsertRows.clear();
+            // Clear per-txn INSERT tracking on COMMIT so it doesn't carry over into
+            // subsequent transactions. resetTxScopeIfNewTransaction() (invoked at
+            // the top of processRowLCR and below) is the belt-and-braces case for
+            // when the COMMIT is missed or arrives out of order.
+            currentTxnInsertedRowIds.clear();
+            currentTxId = null;
             final Instant commitTimestamp = lcr.getSourceTime().timestampValue().toInstant();
             dispatcher.dispatchTransactionCommittedEvent(partition, offsetContext, commitTimestamp);
             return;
@@ -304,25 +265,34 @@ class LcrEventHandler implements XStreamLCRCallbackHandler {
             }
         }
 
+        // For LOB_WRITE/LOB_TRIM/LOB_ERASE, the chunk stream (if any) only carries the partial
+        // delta written by DBMS_LOB, not the full post-operation value. Reselect against the
+        // current row so downstream consumers see a consistent full LOB state. Gated on these
+        // specific command types so regular DML is not slowed down — callers who want
+        // per-DML reselection continue to use ReselectColumnsPostProcessor.
+        if (RowLCR.LOB_WRITE.equals(lcr.getCommandType())
+                || RowLCR.LOB_TRIM.equals(lcr.getCommandType())
+                || RowLCR.LOB_ERASE.equals(lcr.getCommandType())) {
+            reselectLobValues(lcr, table, chunkValues);
+        }
+
         final Object rowIdObject = lcr.getAttribute("ROW_ID");
         if (rowIdObject != null) {
             offsetContext.setRowId(rowIdObject.toString());
         }
 
         // Determine operation override for LOB events following an INSERT.
-        // If LOB_WRITE/LOB_TRIM/LOB_ERASE follows an INSERT for the same row+tx,
-        // the operation should be CREATE (not UPDATE) since the LOB data is part of the INSERT.
+        // If LOB_WRITE/LOB_TRIM/LOB_ERASE shares a transaction with a tracked INSERT
+        // of the same row, the operation is logically part of the INSERT — emit CREATE.
         Operation operationOverride = null;
         if (RowLCR.LOB_WRITE.equals(lcr.getCommandType())
                 || RowLCR.LOB_TRIM.equals(lcr.getCommandType())
                 || RowLCR.LOB_ERASE.equals(lcr.getCommandType())) {
-            if (rowIdObject != null) {
-                final String key = rowIdObject.toString() + "|" + lcr.getTransactionId();
-                if (pendingInsertRows.containsKey(key)) {
-                    operationOverride = Operation.CREATE;
-                    LOGGER.debug("LOB operation {} for row {} follows INSERT, mapping to CREATE",
-                            lcr.getCommandType(), rowIdObject);
-                }
+            resetTxScopeIfNewTransaction(lcr);
+            if (rowIdObject != null && currentTxnInsertedRowIds.contains(rowIdObject.toString())) {
+                operationOverride = Operation.CREATE;
+                LOGGER.debug("LOB op {} for row {} follows tracked INSERT in same tx — mapping to CREATE",
+                        lcr.getCommandType(), rowIdObject);
             }
         }
         dispatcher.dispatchDataChangeEvent(
@@ -507,16 +477,6 @@ class LcrEventHandler implements XStreamLCRCallbackHandler {
 
             columnChunks.clear();
 
-            // DBZ-4741: For LOB_WRITE/LOB_TRIM/LOB_ERASE, the chunk data contains only the partial
-            // delta written by DBMS_LOB.WRITE/WRITEAPPEND/TRIM/ERASE, not the full LOB value.
-            // Re-read the full LOB value from the source database directly.
-            if (jdbcConnection != null && currentRow != null
-                    && (RowLCR.LOB_WRITE.equals(currentRow.getCommandType())
-                        || RowLCR.LOB_TRIM.equals(currentRow.getCommandType())
-                        || RowLCR.LOB_ERASE.equals(currentRow.getCommandType()))) {
-                reselectLobValues(currentRow, resolvedChunkValues);
-            }
-
             dispatchDataChangeEvent(currentRow, resolvedChunkValues);
         }
         catch (InterruptedException e) {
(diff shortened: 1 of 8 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/XStreamChangeRecordEmitter.java
```diff
@@ -56,10 +71,6 @@ private static Object[] getColumnValues(Table table, ColumnValue[] columnValues,
         Object[] values = new Object[table.columns().size()];
         if (columnValues != null) {
             for (ColumnValue columnValue : columnValues) {
-                // Skip Oracle ROW_ARCHIVAL column
-                if ("ORA_ARCHIVE_STATE".equals(columnValue.getColumnName())) {
-                    continue;
-                }
```
Naros, line 62: Is there a specific reason why the archival column exclusion was removed?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
```diff
@@ -133,9 +156,55 @@ else if (lcr instanceof DDLLCR) {
...
+                    else {
+                        dispatchDataChangeEvent(row, null);
+                    }
+                }
+                else {
+                    dispatchDataChangeEvent(row, null);
+                }
+            }
             return;
         }
```
Naros: Is there a reason this couldn't be moved to `dispatchDataChangeEvent` after table lookup? This would have several benefits from what I can see:

1. The guard for `currentRow = row` when chunkData is `true` can be dropped, and the existing if/else that is at the end of this method should suffice.
2. The number of schema lookups for `TableId` to `Table` gets reduced to `1`. Furthermore, this also guards against prematurely reselecting table columns when the `Table` is tracked, but it's not captured when `schema.history.internal.store.only.captured.tables.ddl` is set to `false`. This allows the single lookup of `Table` to be passed into `reselectLobValues` as an argument, too.
3. The call to `reselectLobValues` inside `resolveAndDispatchCurrentChunkedRow` could be dropped since the `reselectLobValues` is now inline inside `dispatchDataChangeEvent`.
4. Keeps all LOB_WRITE/LOB_TRIM, and LOB_ERASE logic in one place.

Do you think that is feasible?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
```diff
@@ -426,4 +525,105 @@ private void resolveAndDispatchCurrentChunkedRow() {
...
+                                    if (strVal != null) {
+                                        chunkValues.put(colName, strVal);
+                                        LOGGER.debug("Reselected CLOB/XMLTYPE column {} ({} chars)", colName, strVal.length());
+                                    }
+                                }
+                            }
+                        }
+                    }
+                    else {
+                        LOGGER.warn("Reselect for {} returned no rows (row may have been deleted).", tableId);
```
Naros: Should we consider providing more context here: transactionId, scn, rowId? I'm just thinking that if a user sees this message, they can work with the DBA in that context to query LogMiner directly. Otherwise, it's quite opaque which row in tableId had this issue.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
```diff
@@ -60,14 +66,29 @@ class LcrEventHandler implements XStreamLCRCallbackHandler {
...
     private final XStreamStreamingChangeEventSourceMetrics streamingMetrics;
+    private final OracleConnection jdbcConnection;
     private final Map<String, ChunkColumnValues> columnChunks;
+    // Track INSERT operations per ROW_ID+txId so that LOB_WRITE/LOB_TRIM following an INSERT
+    // can be correctly mapped to Operation.CREATE instead of Operation.UPDATE (DBZ-4741).
+    // Key: "ROW_ID|transactionId", cleared on COMMIT.
+    private final Map<String, Boolean> pendingInsertRows;
     private RowLCR currentRow;
 
     LcrEventHandler(OracleConnectorConfig connectorConfig, ErrorHandler errorHandler,
```
Naros, line 69: This old constructor is no longer used; there's no need to keep the stub.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
```diff
@@ -426,4 +525,105 @@ private void resolveAndDispatchCurrentChunkedRow() {
...
+            }
+        }
+
+        if (pkValues.size() < pkColumns.size()) {
+            LOGGER.warn("Cannot reselect LOB values for table {}: incomplete primary key in LCR.", tableId);
+            return;
+        }
+
+        // Build SELECT for each LOB column
+        final String lobColumns = String.join(", ", chunkValues.keySet());
```
Naros: I would prefer if the reselection logic itself were kept in `OracleConnection` alongside the existing code. Perhaps the existing API can be adjusted to accommodate both call paths, or at the very least have a single internal implementation that two public methods can call into with context-specific arguments to control what the reselection query does.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/XStreamChangeRecordEmitter.java
```diff
@@ -27,23 +27,38 @@
 public class XStreamChangeRecordEmitter extends BaseChangeRecordEmitter<ColumnValue> {
 
     private final RowLCR lcr;
+    private final Operation operationOverride;
 
     public XStreamChangeRecordEmitter(OracleConnectorConfig connectorConfig, Partition partition, OffsetContext offset, RowLCR lcr,
```
Naros, line 31: This ctor is now dead code, it can be removed.

Naros (changes requested):
Hi @pyaroslav, great work so far. I've left some recommendations, if you could take a look. Also, these changes will require integration tests, if you could add those that'd be great. Thanks.

pyaroslav:
Thanks for the thorough review, @Naros. All six inline points are addressed — individual replies threaded under each comment. I've also added six integration tests (three each for CLOB and BLOB) covering `WRITEAPPEND`, `TRIM`, and `ERASE`.

**What's in this push:**

- `XStreamChangeRecordEmitter`: restored the `ORA_ARCHIVE_STATE` skip; removed the unused 9-arg constructor.
- `LcrEventHandler`: moved the LOB reselect into `dispatchDataChangeEvent` after the single `Table` lookup; delegated the reselection itself to `OracleConnection.reselectColumns` so this path shares the SQL builder, identifier quoting, and flashback-fallback with the existing `ReselectColumnsPostProcessor` path; removed the unused 10-arg constructor; log lines now carry `txId`, `scn`, and `rowId`.
- Replaced the map-based INSERT tracking with a per-transaction `Set` that self-evicts when a new transaction is observed (belt-and-braces for out-of-order or missed COMMIT LCRs).
- Integration tests: narrowed the existing `shouldNotStreamAnyChangesWhenLobEraseIsDetected` to LogMiner-only because XStream now emits an UPDATE for `LOB_ERASE` as of this PR; added XStream-only ITs for WRITEAPPEND/TRIM/ERASE on both BLOB and CLOB.

**One open issue I want to flag proactively — this matters for whether the new ITs pass on CI:**

I deployed the PR code into a live pipeline against Oracle 19c + XStream and ran the exact scenarios the new ITs assert. **Every LOB value is byte-exact** — `WRITEAPPEND`, `TRIM`, and `ERASE` all produce the expected reselected values. But the standalone-transaction LOB operations emit **`op=c` instead of `op=u`** in the Kafka record, even though debug logs show the `operationOverride` branch correctly resolving to `null` and the switch case for `LOB_WRITE`/`LOB_TRIM`/`LOB_ERASE` returns `Operation.UPDATE`. I couldn't root-cause the gap before pushing. If your test environment reproduces this, the `VerifyRecord.isValidUpdate(...)` assertion in the new ITs will fail. If yours is green, that's useful information (may be specific to a particular XStream config on my side). Either way, I'd appreciate your eyes — it isn't a data-integrity issue (values are correct), but it does affect downstream consumers that branch on `op`, so I'd consider it a blocker for merge.

<details>
<summary><b>Verification evidence (live pipeline, Oracle 19c XStream, 8 KB seed, exact IT scenarios)</b></summary>

| Scenario | INSERT event | LOB-op event | Desc UPDATE event |
|---|---|---|---|
| BLOB WRITEAPPEND (8 000 + 4) | `op=c`, BLOB=8 000 ✓ | `op=c` ❌ (expected `u`), BLOB=**8 004** ✓ | `op=u`, unavailable ✓ |
| BLOB TRIM (8 000 → 5 000)   | `op=c`, BLOB=8 000 ✓ | `op=c` ❌ (expected `u`), BLOB=**5 000** ✓ | `op=u`, unavailable ✓ |
| BLOB ERASE (100 @ 1)        | `op=c`, BLOB=8 000 ✓ | `op=c` ❌ (expected `u`), BLOB=8 000 ✓ (zero-filled 0..99) | `op=u`, unavailable ✓ |
| CLOB WRITEAPPEND (8 000 + 4)| `op=c`, CLOB=8 000 ✓ | `op=c` ❌ (expected `u`), CLOB=**8 004** ✓ | `op=u`, unavailable ✓ |
| CLOB TRIM (8 000 → 5 000)   | `op=c`, CLOB=8 000 ✓ | `op=c` ❌ (expected `u`), CLOB=**5 000** ✓ | `op=u`, unavailable ✓ |
| CLOB ERASE (100 @ 1)        | `op=c`, CLOB=8 000 ✓ | `op=c` ❌ (expected `u`), CLOB=8 000 ✓ (space-filled 0..99) | `op=u`, unavailable ✓ |

Byte-level content probes: BLOB ERASE → first 100 bytes `0x00`, offset 100+ `0x43`; CLOB ERASE → first 100 chars space, offset 100+ `'C'`; BLOB WRITEAPPEND trailing 4 bytes `0x41 0x42 0x43 0x44`.

</details>

**What I verified locally:** `./mvnw -pl debezium-connector-oracle -am install -DskipTests -Passembly,oracle-xstream` builds cleanly on Temurin 21; all 290 existing unit tests pass; the six new ITs compile (structural check — I don't have access to the licensed Oracle EE test image locally, CI is the authoritative check).

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/XStreamChangeRecordEmitter.java
pyaroslav: Unintentional removal in my original commit — restored.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
pyaroslav: Done. The reselect now runs inside `dispatchDataChangeEvent` after the single `Table` lookup, which is threaded through as an argument. The special-case branch in `processRowLCR` is gone — the uniform dispatch flow handles LOB events the same way as regular DML. The explicit call from `resolveAndDispatchCurrentChunkedRow` is also removed. Good catch on the `schema.history.internal.store.only.captured.tables.ddl=false` case — the new code benefits because the `Table` lookup happens exactly once per event.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
pyaroslav: Reselect warnings now include `transactionId`, `scn`, and `rowId` so a DBA can correlate against LogMiner without extra instrumentation.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
pyaroslav: Removed. `XstreamStreamingChangeEventSource` already calls the 11-arg ctor.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
pyaroslav: Delegated to the existing `OracleConnection.reselectColumns(Table, columns, keyColumns, keyValues, source, consumer)`. The LOB path supplies a `ResultSetConsumer` that reads each LOB column via `rs.getBytes` (BLOB) or `rs.getString` (CLOB/XMLTYPE) and updates the `chunkValues` map. No parallel SQL builder — the XStream path and `ReselectColumnsPostProcessor` now share the same internal implementation, including the flashback-fallback and identifier quoting.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/XStreamChangeRecordEmitter.java
pyaroslav: Removed. Only the 10-arg ctor with `operationOverride` is reached now.

pyaroslav:
Pushed follow-up commit `f8c8583` that replaces `operationOverride` + per-tx INSERT tracking with a smaller fix. End-to-end validation on Oracle 19c XStream surfaced a primary-key bug the earlier commit didn't catch.

**What the earlier commit missed.** For `LOB_WRITE` / `LOB_TRIM` / `LOB_ERASE` LCRs XStream delivers no PK on the old side (only the modified LOB column shows up in `newValues`). The prior commit merged `oldValues → newValues`, so `newKey` had the PK but `oldKey` didn't. `RelationalChangeRecordEmitter.emitUpdateRecord` then saw `oldKey != newKey`, routed through `emitUpdateAsPrimaryKeyChangeRecord`, and emitted a `DELETE + CREATE` — so the wire format showed `op='c'` on every standalone `DBMS_LOB.*`, not `op='u'`.

**What the follow-up does.** Builds both the `old` and `new` column arrays from `mergeColumnValues(oldValues, newValues, chunkValues)` for LOB ops; overlays `oldChunkValues` (UNAVAILABLE_VALUE markers) onto the old side so `before.<lobCol>` is reported as unavailable instead of inheriting the post-op value. Removes the per-tx INSERT tracking and `operationOverride` plumbing entirely — with old/new populated correctly, `getOperation()` produces the right `Operation` from the LCR command type alone. Constructor is back to 9 args.

**Semantic change.** A same-tx `INSERT` + `LOB_WRITE` now emits two records (`op=c` for the INSERT, `op=u` for the LOB reselect) instead of a merged `op=c`. This matches how `ReselectColumnsPostProcessor` already behaves on the LogMiner path. Happy to reconsider if I'm missing a LogMiner case.

**Validated** against Oracle 19c + XStream:

| Scenario | Observed |
|---|---|
| `INSERT(id=N, BLOB[8000])` | `op=c key.ID=N blob=8000` ✅ |
| `DBMS_LOB.WRITEAPPEND(+4)` | `op=u key.ID=N blob=8004` ✅ |
| `DBMS_LOB.TRIM(→5000)` | `op=u key.ID=N blob=5000` ✅ |
| `DBMS_LOB.ERASE(100 @ 1)` | `op=u key.ID=N blob=8000` ✅ |
| Same three ops on `CLOB` | all `op=u key.ID=N` with reselected post-op value ✅ |

PR description updated to match the simpler final design.

cc @Naros — whenever you have a moment, a fresh `/packit test --labels oracle` against `f8c8583` would be great; the previous XStream-variant failures should be gone with this commit.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/XStreamChangeRecordEmitter.java
pyaroslav: Follow-up commit `f8c8583` removes `operationOverride` entirely — the old/new column arrays are now merged symmetrically for LOB ops so `getOperation()` returns the right `Operation` from the LCR command type alone. The 9-arg ctor is the only one that exists now. See the [top-level PR comment](https://github.com/debezium/debezium/pull/7374#issuecomment-4316525477) for the full rationale.

Naros:
@pyaroslav can you do me a favor and rebase the PR on `main`? Thanks.

pyaroslav:
Rebased onto current `main` (`223610c4f`) — the conflicts were just `COPYRIGHT.txt` and `jenkins-jobs/scripts/config/Aliases.txt`, both of which were "both branches added a contributor at the bottom" patterns. Resolved by keeping both entries.

Also picked up two small items from this round of review while I was in there:

- **`LcrEventHandler.reselectLobValues`** — drop the `if (blobVal != null)` / `if (strVal != null)` guards in the reselect consumer. Previously, a multi-LOB row whose `LOB_WRITE`/`LOB_TRIM`/`LOB_ERASE` only modified one column would emit `UNAVAILABLE_VALUE` for the other LOB columns when those columns were genuinely `NULL` on the row, instead of `null`. Always overwriting `chunkValues` with the reselected value (including `null`) makes `before`/`after` accurately represent NULL LOB columns.
- **`OracleBlobDataTypesIT`** — fix a stray `// WRITEAPPEND 4 KB` comment that described the appended payload as kilobytes when the test actually appends 4 bytes (`0x41 0x42 0x43 0x44`). Comment-only.

I deliberately did *not* take Copilot's suggestion to switch the reselect to `AS OF SCN` flashback semantics. The current `source=null` matches `ReselectColumnsPostProcessor` (which @Naros explicitly asked us to share an implementation with), and adding `AS OF SCN` reintroduces ORA-1555 risk on long-running operations. The other two Copilot comments referenced PR-description text that I rewrote on Apr 24 — they're against the older description.

End-to-end re-validated against Oracle 19c + XStream after the rebase + null-fix:

| Scenario | Observed |
|---|---|
| `INSERT(id=N, BLOB[8000])`           | `op=c key.ID=N blob=8000` ✅ |
| `DBMS_LOB.WRITEAPPEND(BLOB,+4)`      | `op=u key.ID=N blob=8004` ✅ |
| `DBMS_LOB.TRIM(BLOB,→5000)`          | `op=u key.ID=N blob=5000` ✅ |
| `DBMS_LOB.ERASE(BLOB, 100 @ 1)`      | `op=u key.ID=N blob=8000` ✅ |
| Same three ops on `CLOB`             | all `op=u key.ID=N` with reselected value ✅ |

Branch is at `ce30a0381` and is now mergeable. cc @Naros — happy to retrigger packit when convenient.

Naros:
Hi @pyaroslav, I am seeing errors in the `OracleClobDataTypeIT` and `OracleBlobDataTypeIT` tests, which show that what the test expects (which is the full reselection) isn't happening in the LCR handler:

* OracleClobDataTypeIT
  * shouldStreamUpdateWithReselectedValueForXStreamLobErase
  * shouldStreamUpdateWithReselectedValueForXStreamLobTrim
  * shouldStreamUpdateWithReselectedValueForXStreamLobWriteAppend

* OracleBlobDataTypeIT
  * shouldStreamUpdateWithReselectedValueForXStreamLobErase
  * shouldStreamUpdateWithReselectedValueForXStreamLobTrim
  * shouldStreamUpdateWithReselectedValueForXStreamLobWriteAppend

Can you please double-check these tests?

pyaroslav:
@Naros — found the CDB+PDB-mode root cause for the six XStream IT failures.

**Diagnosis.** `LcrEventHandler.reselectLobValues` was reusing the connector's long-lived `jdbcConnection`. In CDB+PDB the streaming connection is bound to the CDB root container, so a SELECT against the user table (which lives in the PDB) raises `ORA-00942: table or view does not exist`. My old code caught the `SQLException`, logged at `WARN`, and silently fell through — `chunkValues` retained the pre-seeded `UNAVAILABLE_VALUE` markers. The wire record's `after.VAL_BLOB` came out as the unavailable placeholder instead of the post-write payload, which is exactly the assertion-failure pattern in `shouldStreamUpdateWithReselectedValueForXStreamLob*`. Our non-CDB production XStream didn't catch this — `connectorConfig.getPdbName()` is empty there, the table is reachable on the streaming connection directly, and the reselect happened to work.

**Fix in commit `a8e6f0c`.** Match the pattern already used elsewhere in the same file (`getTableMetadataDdl`) and in `BaseChangeRecordEmitter.emitUpdateAsPrimaryKeyChangeRecord` (the LogMiner reselect):

```java
final String pdbName = connectorConfig.getPdbName();
try (OracleConnection connection = new OracleConnection(connectorConfig, false)) {
    if (!Strings.isNullOrBlank(pdbName)) {
        connection.setSessionToPdb(pdbName);
    }
    final boolean found = connection.reselectColumns(...);
}
catch (SQLException e) {
    throw new DebeziumException("Failed to reselect LOB values for table " + table.id() + " ...", e);
}
```

Three things changed:

1. **Open a fresh `OracleConnection` per LOB op and switch its session to the PDB** if one is configured. This is the only difference that matters for your CI environment — non-CDB callers (like our pipeline) skip the PDB switch and behave as before.
2. **Promote the `SQLException` path from WARN+fall-through to a thrown `DebeziumException`.** Matches the existing `BaseChangeRecordEmitter` reselect path and surfaces real misconfig instead of producing wire records with unavailable placeholders that fail downstream assertions in non-obvious ways. This is what made the original failure mode confusing — silent fallback hid the actual cause.
3. **Drop the `jdbcConnection` field and constructor parameter from `LcrEventHandler`** (now 10-arg). It had no other consumer; aligns with your earlier review preference to remove dead-code stubs. Single call site in `XstreamStreamingChangeEventSource` updated.

**Validation.** Live e2e on Oracle 19c XStream (non-CDB) on fresh IDs 1030–1035 — six-scenario LOB battery (BLOB/CLOB × WRITEAPPEND/TRIM/ERASE):

| Scenario | Observed |
|---|---|
| `INSERT(id=N, BLOB[8000])` | `op=c key.ID=N blob=8000` ✅ |
| `DBMS_LOB.WRITEAPPEND(BLOB,+4)` | `op=u key.ID=N blob=8004` ✅ |
| `DBMS_LOB.TRIM(BLOB,→5000)` | `op=u key.ID=N blob=5000` ✅ |
| `DBMS_LOB.ERASE(BLOB, 100 @ 1)` | `op=u key.ID=N blob=8000` ✅ |
| Same three ops on `CLOB` | all `op=u key.ID=N` with reselected value ✅ |

CDB+PDB I can't reproduce locally (our test env is non-CDB), so the proof is in the pattern alignment with the precedents you pointed at — happy to run another `/packit test --labels oracle` cycle whenever you're ready, branch is at `a8e6f0c`.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
```diff
@@ -499,12 +496,20 @@ private void reselectLobValues(RowLCR row, Table table, Map<String, Object> chun
 
         final List<String> lobColumnNames = new ArrayList<>(chunkValues.keySet());
 
-        try {
+        // Use a separate, short-lived connection for the reselect — pinned to the PDB the
+        // table lives in. The streaming JDBC connection cannot be reused: it is bound to
+        // the CDB root in CDB+PDB deployments and would not see the user table. This
+        // mirrors getTableMetadataDdl above and BaseChangeRecordEmitter's reselect path.
+        final String pdbName = connectorConfig.getPdbName();
+        try (OracleConnection connection = new OracleConnection(connectorConfig, false)) {
```
Naros: While a new connection works here, I am curious about the performance impact. You could reuse the provided `JdbcConnection` you had here previously, but you'd just need to account for disconnection and for PDBs, which you could solve this way:
```java
    /**
     * Get the underlying connection, ready for use.
     *
     * @return the JDBC connection, connected, and inside the pluggable database, if a PDB is used
     * @throws SQLException if there was an error connecting or switch contexts
     */
    private OracleConnection getConnection() throws SQLException {
        if (!jdbcConnection.isConnected()) {
            jdbcConnection.connect();

            if (connectorConfig.isUsingPluggableDatabase()) {
                jdbcConnection.setSessionToPdb(connectorConfig.getPdbName());
            }
        }
        return jdbcConnection;
    }
```
In this example, I added a helper method to the configuration class, e.g
```java
    public boolean isUsingPluggableDatabase() {
        return !Strings.isNullOrBlank(pdbName);
    }
```
Given we use this in a number of places, I think it makes sense for this helper to exist here, and we can cleanly streamline the if/set block as you can see.

The log entry about using a separate JDBC connection in `getTableMetadataDdl`, I believe, can also use this strategy, since XStream now uses its own dedicated connection rather than the one used by the Snapshot.

I think doing it this way will help in the future when we add downstream mining support for XStream.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
```diff
@@ -499,12 +496,20 @@ private void reselectLobValues(RowLCR row, Table table, Map<String, Object> chun
...
+            }
             // reselectColumns advances the ResultSet cursor internally before invoking
             // the consumer, so we read column values directly without calling rs.next().
             // A return of `false` means no matching row was found — typically the row
             // was deleted between the LCR emission and our reselect.
-            final boolean found = jdbcConnection.reselectColumns(table, lobColumnNames, pkColumns, pkValues, null, rs -> {
+            final boolean found = connection.reselectColumns(table, lobColumnNames, pkColumns, pkValues, null, rs -> {
                 for (String colName : lobColumnNames) {
                     final Column column = table.columnWithName(colName);
                     if (column == null) {
```
Naros, line 541: There should never be a point in time when the event we receive has a column that isn't in the relational model; ideally, this should throw an exception rather than be silently dropped. What situation have you observed where this happens?

If we keep this logic, at the very least a warning should be raised.

Naros (changes requested):
Hi @pyaroslav I've left two inline comments, can you take a look?  I think once we address these issues, we should be in a good shape.

pyaroslav:
@Naros — addressed both inline comments in commit `5c75794`.

**Comment 1 (line 504, perf / lazy-reconnect).** Adopted your pattern. `LcrEventHandler` regains the `jdbcConnection` field and a small `getConnection()` helper that lazy-reconnects + pins to the PDB on (re)connection:

```java
private OracleConnection getConnection() throws SQLException {
    if (!jdbcConnection.isConnected()) {
        jdbcConnection.connect();
        if (connectorConfig.isUsingPluggableDatabase()) {
            jdbcConnection.setSessionToPdb(connectorConfig.getPdbName());
        }
    }
    return jdbcConnection;
}
```

Added the `OracleConnectorConfig.isUsingPluggableDatabase()` helper as you sketched — pulls the `Strings.isNullOrBlank(pdbName)` check up so other call sites can share it. Per your hint, also routed `getTableMetadataDdl` through the same helper now that the streaming path has its own dedicated connection — single source of connection lifecycle for both out-of-bands callbacks.

**Comment 2 (line 515, column null skip).** You're right — that branch is unreachable by construction (`lobColumnNames` is derived from `schema.getLobColumnsForTable(table.id())`, the same `Table` we use for the lookup). Replaced the silent `continue` with:

```java
throw new DebeziumException("Schema cache for table " + table.id()
        + " does not include LOB column '" + colName
        + "' that was registered for reselection (tx=" + row.getTransactionId()
        + ", scn=" + offsetContext.getScn() + ").");
```

Carries `tableId`, `txId` and `scn` so a future invariant violation is debuggable rather than just producing partial-but-passing wire data.

**Validation.** Live e2e on Oracle 19c XStream (non-CDB) on fresh IDs 1040–1045 — same six-scenario LOB battery as v9, identical results:

| Scenario | Observed |
|---|---|
| `INSERT(id=N, BLOB[8000])` | `op=c key.ID=N blob=8000` ✅ |
| `DBMS_LOB.WRITEAPPEND(BLOB,+4)` | `op=u key.ID=N blob=8004` ✅ |
| `DBMS_LOB.TRIM(BLOB,→5000)` | `op=u key.ID=N blob=5000` ✅ |
| `DBMS_LOB.ERASE(BLOB, 100 @ 1)` | `op=u key.ID=N blob=8000` ✅ |
| Same three ops on `CLOB` | all `op=u key.ID=N` with reselected value ✅ |

Branch is at `5c75794`. The `Debezium Platform - conductor` failure on this commit is the same upstream-main failure pattern (also red on `main` HEAD itself; PRs #7390/#7387/#7383/#7382 all merged with it red), not introduced by this PR. Happy for a fresh `/packit test --labels oracle` against `5c75794` whenever convenient.

Side-finding (filed separately, not for this PR): the `XStreamChangeRecordEmitter.overlayColumnValues` / `getColumnValues` paths still NPE on schema-cache gaps when an LCR carries an unknown column (a different code path from this reselect — that one iterates LCR-supplied columns, not our internally-derived list). Filed as **dbz#1867** with the same defensive-pattern reference (dbz#1774's `ORA_ARCHIVE_STATE` skip). Also added a 2026-04-28 production-impact comment to **dbz#801 / DBZ-6373** (CTAS DDL parsing) since CTAS-recreated tables are the most common real-world trigger for that NPE.

Naros (approved):

Naros:
@pyaroslav would you mind raising a follow-up to align the LogMiner behavior with XStream?  In the meantime, lets see how CI fairs :tada:

pyaroslav:
> **Update (after Naros's review):** This v11 commit was superseded by `49a3cfe` — see the [review reply on the `getConnection` thread](https://github.com/debezium/debezium/pull/7374#discussion_r3185163414) for the v12 design (PDB pinning via `LcrEventHandler.init()`, no boolean field). The root-cause diagnosis below still applies; only the implementation has changed.

---

@Naros — diagnosed the CI red on the v10 commit and pushed a fix as `0734700`. Happy for a fresh `/packit test --labels oracle` run whenever convenient.

**Root cause.** In v10's `getConnection()`, the PDB switch was gated behind `!isConnected()`:

```java
if (!jdbcConnection.isConnected()) {
    jdbcConnection.connect();
    if (connectorConfig.isUsingPluggableDatabase()) {
        jdbcConnection.setSessionToPdb(connectorConfig.getPdbName());
    }
}
```

That's correct after a reconnect, but wrong on first use: `XstreamStreamingChangeEventSource` opens the JDBC connection before this handler runs (to attach the Outbound Server), so by the time `getConnection()` is first called the connection is already alive — `isConnected()` returns true, the branch is skipped, and the session is left pointing at the CDB root. DDL fetch and LOB reselect then run in the wrong container, which is what the testing-farm CI surfaced as `ORA-00942: table or view does not exist` on the LOB reselect path in CDB+PDB mode.

**Fix (commit `0734700`).** Track PDB-set state explicitly with a per-connection flag so the switch fires on first use *and* again after any reconnect, independent of `isConnected()`:

```java
private boolean pdbSetOnSession = false;

private OracleConnection getConnection() throws SQLException {
    if (!jdbcConnection.isConnected()) {
        jdbcConnection.connect();
        pdbSetOnSession = false;
    }
    if (!pdbSetOnSession && connectorConfig.isUsingPluggableDatabase()) {
        jdbcConnection.setSessionToPdb(connectorConfig.getPdbName());
        pdbSetOnSession = true;
    }
    return jdbcConnection;
}
```

Reset on disconnect so the next `connect()` re-pins the new physical session. Single field, no behaviour change for non-CDB.

**Validation.**

- *Non-CDB regression check (production-style 19c XStream attach to source primary).* Six-scenario LOB battery on fresh IDs 1050–1055; identical results to v9/v10:

  | Scenario | Observed |
  |---|---|
  | `INSERT(id=N, BLOB[8000])` → `WRITEAPPEND(+4)` → `UPDATE(desc)` | `c blob=8000` → `u blob=8004` → `u (desc)` ✅ |
  | `INSERT(id=N, BLOB[8000])` → `TRIM(→5000)` → `UPDATE(desc)` | `c blob=8000` → `u blob=5000` → `u (desc)` ✅ |
  | `INSERT(id=N, BLOB[8000])` → `ERASE(100@1)` → `UPDATE(desc)` | `c blob=8000` → `u blob=8000` → `u (desc)` ✅ |
  | Same three on `CLOB` | identical ✅ |

  All key.ID values non-zero; all op codes (`c`/`u`/`u`); payload lengths match expectations. So the v10 → v11 change doesn't regress non-CDB behaviour.

- *CDB+PDB path.* Our deployment is non-CDB so I can't drive `XstreamLobIT` locally; the definitive verification is the testing-farm matrix that was red on v10 — happy for `/packit test --labels oracle` against `0734700`.

Branch: `dbz-4741-xstream-lob-write-trim` @ `0734700`. The `Debezium Platform - conductor` failure is the same upstream-main failure that was red on `5c75794` and on `main` HEAD itself; not introduced by this commit.

pyaroslav:
> @pyaroslav would you mind raising a follow-up to align the LogMiner behavior with XStream?

@Naros — ack, will file the tracking issue + PR once this one merges. Plan is to address dbz#1867 (null-guard on the XStream emitter's column iteration; same code area as this PR) first, then the LogMiner alignment as a separate PR. Don't want to interleave them while 7374 is still in review.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
```diff
@@ -341,9 +358,11 @@ private String getTableMetadataDdl(TableId tableId) throws NonRelationalTableExc
...
             jdbcConnection.connect();
-            if (connectorConfig.isUsingPluggableDatabase()) {
-                jdbcConnection.setSessionToPdb(connectorConfig.getPdbName());
-            }
+            pdbSetOnSession = false;
+        }
+        if (!pdbSetOnSession && connectorConfig.isUsingPluggableDatabase()) {
+            jdbcConnection.setSessionToPdb(connectorConfig.getPdbName());
+            pdbSetOnSession = true;
         }
```
Naros: We could probably drop the boolean member variable by doing two things:

1. Inside `init`, we check if we are using a PDB, and if so, call `setSessionToPdb` directly. At this point, the connection should be active, so there shouldn't be any concern with "isConnected" there and it sets up the connection for streaming at the outset. If you wanted to use `getConnection()` regardless, this still yields at most two `setSessionToPdb` calls if the connection had broken, but in practice, only one.
2. Inside the `getConnection`, we can move the shift to the PDB inside the `!isConnected` block and initialize it on reconnect. 

wdyt?

pyaroslav: Good call — thanks. Adopted approach (1): added `LcrEventHandler.init()` that calls `setSessionToPdb` when a PDB is configured, invoked from `XstreamStreamingChangeEventSource.execute()` once the streaming connection is attached. `getConnection()` reverts to the simpler form — its `!isConnected()` branch handles reconnect re-pinning. Boolean field is gone.

Force-pushed as `49a3cfe` (replaces `0734700`). Non-CDB regression check on fresh IDs 1060-1065 — same six-scenario LOB battery, identical results to v11. CDB+PDB verification still relies on the testing-farm matrix.

| Scenario | Observed |
|---|---|
| `INSERT(BLOB[8000])` → `WRITEAPPEND(+4)` → `UPDATE(desc)` | `c blob=8000` → `u blob=8004` → `u (desc)` ✅ |
| `INSERT(BLOB[8000])` → `TRIM(→5000)` → `UPDATE(desc)` | `c blob=8000` → `u blob=5000` → `u (desc)` ✅ |
| `INSERT(BLOB[8000])` → `ERASE(100@1)` → `UPDATE(desc)` | `c blob=8000` → `u blob=8000` → `u (desc)` ✅ |
| Same three on `CLOB` | identical ✅ |

pyaroslav:
@Naros — heads-up: my response on the boolean-flag thread is now hidden behind GitHub's "outdated" filter (the force-push for `49a3cfe` rewrote line 366, so the conversation got auto-collapsed). Reposting the gist here for visibility:

Adopted approach (1): added `LcrEventHandler.init()` that calls `setSessionToPdb` when a PDB is configured, invoked from `XstreamStreamingChangeEventSource.execute()` once the streaming connection is attached. `getConnection()` reverts to the simpler form — its `!isConnected()` branch handles reconnect re-pinning. No boolean field.

Non-CDB regression check on IDs 1060–1065 (same six-scenario LOB battery as v11) is clean. CDB+PDB verification still relies on the testing-farm matrix.

| Scenario | Observed |
|---|---|
| `INSERT(BLOB[8000])` → `WRITEAPPEND(+4)` → `UPDATE(desc)` | `c blob=8000` → `u blob=8004` → `u (desc)` ✅ |
| `INSERT(BLOB[8000])` → `TRIM(→5000)` → `UPDATE(desc)` | `c blob=8000` → `u blob=5000` → `u (desc)` ✅ |
| `INSERT(BLOB[8000])` → `ERASE(100@1)` → `UPDATE(desc)` | `c blob=8000` → `u blob=8000` → `u (desc)` ✅ |
| Same three on `CLOB` | identical ✅ |

Original threaded reply: https://github.com/debezium/debezium/pull/7374#discussion_r3185163414

pyaroslav:
@Naros — quick observation on the testing-farm results for `49a3cfe`: the six tests originally flagged on v10 (`OracleClobDataTypeIT` and `OracleBlobDataTypeIT` `shouldStreamUpdateWithReselectedValueForXStreamLob*`) are now passing across all three XStream-mode variants — the v12 init-time PDB switch landed clean.

The remaining red on the XStream variants is from unrelated tests (`OracleConnectorIT.shouldCreateSnapshotSchemaOnlyRecovery`, `OracleDefaultValueIT` JMX `InstanceNotFoundException`, `IncrementalSnapshotCaseSensitiveIT.stopCurrentIncrementalSnapshot…`). None touch the LCR/LOB code path; they look infrastructure-level (snapshot recovery, MBean registration timing, signal-channel edge case). Happy to dig into any specific one if it ends up looking PR-related, but on first read they're noise.

Naros:
Thanks @pyaroslav, just one final question, and is there any plan to adjust the behavior so that LM and XStream both handle these LOB operations in the same way? Right now there is a difference in their behavior, and they ideally should be handled in precisely the same way.

Is this something you plan to do as a follow-up? If not, could you at least raise a GH issue as a follow-up so we don't lose sight of the fact there is a behavior discrepancy we need to close.

pyaroslav:
@Naros — both: I do plan to do this as a follow-up PR (was already in my queue, ack'd in the [outdated-collapsed thread reply](https://github.com/debezium/debezium/pull/7374#discussion_r3185163414) from the boolean-flag review), and I just filed [dbz#1900](https://github.com/debezium/dbz/issues/1900) so the discrepancy is tracked durably.

Sequence: [dbz#1867](https://github.com/debezium/dbz/issues/1867) PR first (null-guard, same code area as this PR — would conflict in review otherwise), then the LogMiner alignment as the second follow-up. Will reference both #7374 and the new issue from the alignment PR when it goes up.

pyaroslav:
@Naros — just checking in when you get a moment. PR is at `49a3cfe` (v12), the six XStream LOB tests pass on testing-farm, and the LogMiner alignment is tracked as [dbz#1900](https://github.com/debezium/dbz/issues/1900). Anything else needed from me before merge, or are we waiting on something specific?

Naros:
Hi @pyaroslav could you rebase the PR please.

pyaroslav:
@Naros rebased onto current `main` (force-pushed `938d336`). Conflicts were only the additive `COPYRIGHT.txt` / `Aliases.txt` contributor entries — resolved by keeping both. The 6 code commits applied cleanly; I compiled + test-compiled `debezium-connector-oracle` against current main to confirm no API drift. Ready for a fresh `/packit test --labels oracle` whenever convenient.

Naros:
Hi @pyaroslav, could I get a quick rebase on `main`?

pyaroslav:
> Hi @pyaroslav, could I get a quick rebase on `main`?

Hi @Naros 
Rebased on current main. The only change beyond the rebase is one small adaptation to the OracleConnectionFactory refactor, kept as a separate commit for easy review. Is this targeted for 3.7?

Naros:
Hi @pyaroslav, could I get one more rebase? I'm sorry for these 2 files being problematic; hopefully, once we get a PR merged that maintains these in alphabetical order, merge conflicts will become smaller for them.

> Is this targeted for 3.7?

That's my plan. :crossed_fingers:

pyaroslav:
Hi @Naros,  rebased on main (e43e32ad9). Thanks for targeting 3.7.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
```diff
@@ -309,20 +318,50 @@ private TableId getTableId(LCR lcr) {
...
+    }
+
+    /**
+     * Returns the streaming JDBC connection ready for an out-of-bands query against the
+     * captured database. Reconnects the underlying connection if it is currently
+     * disconnected (e.g. after an Oracle restart) and re-pins the session to the
+     * configured PDB on reconnection. Reused across out-of-bands callbacks (DDL fetch,
+     * LOB reselect) so we don't open and tear down a JDBC connection per LCR.
+     *
+     * <p>The connection is dedicated to the XStream change-event source — the snapshot
```
Naros: This claim does not hold for the standard connection factory, and it is the root of a real problem.

`StandardOracleConnectionFactory.mainConnection()` returns a single shared connection: `OracleConnectorTask`, `OracleSnapshotChangeEventSource`, and now this handler all hold the same object. Everywhere else in the module that connection rests in `CDB$ROOT` and callers switch to the PDB and reset around each PDB-scoped use (`validateGuardrailLimits`, `OracleSnapshotChangeEventSource.prepare()`/`close()`, `checkDatabaseAndTableState`, the incremental snapshot source).

`init()` pins it to the PDB and never resets. The concrete failure is a blocking snapshot: `XstreamStreamingChangeEventSource` pauses, `OracleSnapshotChangeEventSource` runs on this same connection, and its `close()` calls `resetSessionToCdb()`. When streaming resumes, `getConnection()` sees the connection still open, skips the re-pin, and the next DDL fetch or LOB reselect runs in `CDB$ROOT`. With the fail-fast choice in `reselectLobValues` that is an ORA-00942 wrapped in a `DebeziumException`, and the task dies.

I know the reuse was my suggestion, so apologies for the detour. I think we can keep the shared connection and fix this in the streaming source rather than here:

1. Move `init()` into `XstreamStreamingChangeEventSource` under a more descriptive name, something like `pinConnectionToPdb()`, and call it where `eventHandler.init()` is called today.
2. Call it again right after `context.waitSnapshotCompletion()` returns in the receive loop, so the session is back in the PDB before the next LCR is processed. A short comment there explaining that the blocking snapshot resets the shared connection to the CDB would protect the next person who touches this.

With that in place the handler no longer needs `init()`, and the re-pin branch in `getConnection()` can go too, since `isConnected()` only reports false after an explicit close. I checked the other users of the main connection during streaming: the task only touches it at start and stop, and incremental snapshots, heartbeats, and the bean registry hold their own connections, so the blocking snapshot is the only thing that changes the container underneath us today.

See also the note on the constructor call in `XstreamStreamingChangeEventSource` about which factory the connection should come from; the two go together.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/XstreamStreamingChangeEventSource.java
```diff
@@ -112,10 +112,12 @@ public void execute(ChangeEventSourceContext context, OraclePartition partition,
         this.effectiveOffset = offsetContext;
 
         LcrEventHandler eventHandler = new LcrEventHandler(connectorConfig, errorHandler, dispatcher, clock, schema,
-                partition, offsetContext, isTableCaseInsensitive(), this, streamingMetrics);
+                partition, offsetContext, isTableCaseInsensitive(), this, streamingMetrics,
+                connectionFactory.streamingConnectionFactory().mainConnection());
```
Naros: This needs to come from the primary side rather than the streaming factory.

`capture.mode=downstream` (and `physical_standby`) is only supported for XStream, and in those modes `connectionFactory.streamingConnectionFactory().mainConnection()` is the read-only secondary connection to the mining or standby instance (`DualOracleConnectionFactory`). Before this PR the DDL fetch in `LcrEventHandler` opened a connection from the primary connector config, so it always hit the source database. With this change both the DDL fetch and the LOB reselect run against the mining instance, where the captured tables do not exist, so new-table DDL lookups fail and every LOB event throws.

`connectionFactory.snapshotConnectionFactory().mainConnection()` resolves to the primary in both factory implementations, so passing that here covers standard, downstream, and standby modes with the same shared connection. Combined with re-pinning the session after `context.waitSnapshotCompletion()` (see the comment on `getConnection()` in `LcrEventHandler`), that also handles the blocking snapshot case.
```suggestion
                connectionFactory.snapshotConnectionFactory().mainConnection());
```

I would suggest a comment on why we pass the snapshot connection for clarity.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
```diff
@@ -428,4 +468,117 @@ private void resolveAndDispatchCurrentChunkedRow() {
...
+        }
+
+        final List<String> lobColumnNames = new ArrayList<>(chunkValues.keySet());
+
+        try {
+            // reselectColumns advances the ResultSet cursor internally before invoking
+            // the consumer, so we read column values directly without calling rs.next().
+            // A return of `false` means no matching row was found — typically the row
+            // was deleted between the LCR emission and our reselect.
+            final boolean found = getConnection().reselectColumns(table, lobColumnNames, pkColumns, pkValues, null, rs -> {
```
Naros: Small follow-up to the earlier Copilot thread, which was closed without a change here. Passing `null` as the source means the reselect always reads the row's current state, so a later committed change to the same row can leak into this event's `after` value.

By the time we get here `processLCR` has already called `offsetContext.setEventCommitScn(...)`, and `OracleSourceInfoStructMaker` writes that into `commit_scn`. Passing `offsetContext.getSourceInfo()` instead of `null` turns on the existing `AS OF SCN` query in `OracleConnection.reselectColumns`, with its built-in fallback to the plain query on ORA-01555 and ORA-01466. That would give the same consistency the LogMiner reselect path gets for free.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
```diff
@@ -428,4 +468,117 @@ private void resolveAndDispatchCurrentChunkedRow() {
...
+                                + " does not include LOB column '" + colName
+                                + "' that was registered for reselection (tx=" + row.getTransactionId()
+                                + ", scn=" + offsetContext.getScn() + ").");
+                    }
+                    // Always overwrite — including with null — so a column whose row truly
+                    // holds NULL is reported as null, not as the UNAVAILABLE_VALUE placeholder
+                    // that was pre-seeded in chunkValues. (Without this, a multi-LOB row whose
+                    // LOB op only touched one column would emit the unavailable marker for the
+                    // others.)
+                    if (column.jdbcType() == java.sql.Types.BLOB) {
```
Naros: Question rather than a blocker. `chunkValues` was pre-seeded from `schema.getLobColumnsForTable(...)`, and that set also includes `SQLXML`, extended `VARCHAR2`/`NVARCHAR2`, and `JSON` columns, not only CLOB/NCLOB/BLOB. So this branch reads everything that is not a BLOB with `rs.getString`.

Have you checked that `getString` returns what the value converters expect for XMLTYPE and JSON columns? `ReselectColumnsPostProcessor` uses `rs.getObject` and lets the converters handle the type, and `OracleConnection.getColumnValue(rs, index, column, table)` already encodes the Oracle-specific rules. Reusing one of those here would keep the two reselect paths behaving the same. Also a nit: `java.sql.Types.BLOB` can be an import.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleBlobDataTypesIT.java
```diff
@@ -899,7 +898,7 @@ record = records.recordsForTopic(topicName("BLOB_TEST")).get(2);
 
     @Test
     @FixFor({ "DBZ-2948", "DBZ-5773" })
-    @SkipWhenAdapterNameIs(value = SkipWhenAdapterNameIs.AdapterName.OLR, reason = "OpenLogReplicator does not differentiate between LOB operations")
+    @SkipWhenAdapterNameIsNot(value = SkipWhenAdapterNameIsNot.AdapterName.ANY_LOGMINER, reason = "DBZ-4741 enables LOB_ERASE propagation for XStream; OLR handled by its own test")
```
Naros: The annotation change is right (both XStream and OLR should skip this now), but the reason text says OLR is "handled by its own test" and I could not find an OLR test for LOB_ERASE. Suggest something like: `"LOB_ERASE is propagated as an UPDATE by XStream (debezium/dbz#579); OpenLogReplicator does not differentiate between LOB operations"`.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleBlobDataTypesIT.java
```diff
@@ -1006,6 +1005,165 @@ record = records.recordsForTopic(topicName("BLOB_TEST")).get(0);
         assertNoRecordsToConsume();
     }
 
+    @Test
+    @FixFor("DBZ-4741")
```
Naros: Convention nit for all six new tests and the skip reasons: the issue was migrated to GitHub, and the commits and PR title already use `debezium/dbz#579`. Newer tests in this module use that form as well.
```suggestion
    @FixFor("debezium/dbz#579")
```

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleClobDataTypeIT.java
```diff
@@ -1072,7 +1071,7 @@ record = records.recordsForTopic(topicName("CLOB_TEST")).get(2);
 
     @Test
     @FixFor({ "DBZ-2948", "DBZ-5773" })
-    @SkipWhenAdapterNameIs(value = SkipWhenAdapterNameIs.AdapterName.OLR, reason = "OpenLogReplicator does not differentiate between LOB operations")
+    @SkipWhenAdapterNameIsNot(value = SkipWhenAdapterNameIsNot.AdapterName.ANY_LOGMINER, reason = "DBZ-4741 enables LOB_ERASE propagation for XStream; OLR handled by its own test")
```
Naros: Same as the BLOB test: the reason text refers to an OLR test that does not exist. Suggest the same wording as over there.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleClobDataTypeIT.java
```diff
@@ -1178,6 +1177,161 @@ record = records.recordsForTopic(topicName("CLOB_TEST")).get(0);
         assertNoRecordsToConsume();
     }
 
+    @Test
+    @FixFor("DBZ-4741")
```
Naros: Same convention nit as in `OracleBlobDataTypesIT`, for the three CLOB tests.
```suggestion
    @FixFor("debezium/dbz#579")
```

Naros (changes requested):
Hi @pyaroslav after the rework with the connection factory, I believe there are some issues with the changes and the pinned connection. If you could take a look below. Appreciate it.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/XstreamStreamingChangeEventSource.java
pyaroslav: Adopted your suggestion — the handler now gets `connectionFactory.snapshotConnectionFactory().mainConnection()` with a comment explaining why the snapshot factory is the right source in downstream/standby modes. Thanks for catching that; my rebase adaptation picked the wrong factory.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
pyaroslav: Implemented exactly as you outlined: `pinConnectionToPdb()` now lives in `XstreamStreamingChangeEventSource`, called before the receive loop and again right after `context.waitSnapshotCompletion()` returns, with a comment explaining the blocking-snapshot reset. `init()` is gone and `getConnection()` no longer re-pins. No apologies needed — the blocking-snapshot interaction was subtle, glad you caught it before merge.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
pyaroslav: Done — `reselectColumns` now receives `offsetContext.getSourceInfo()`, and the comment documents the commit_scn-keyed `AS OF SCN` read plus the ORA-01555/ORA-01466 fallback.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/xstream/LcrEventHandler.java
pyaroslav: Good point. Non-BLOB values now go through `OracleConnection.getColumnValue(rs, i, column, table)` so JSON/XMLTYPE/temporal handling matches the snapshot reads and `ReselectColumnsPostProcessor`. I kept BLOB on `rs.getBytes` deliberately: these values are converted at dispatch time, after the ResultSet is closed, and I didn't want a `Blob` locator's session-state dependency in that window — happy to unify fully if you'd rather. Also added the `java.sql.Types` import.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleBlobDataTypesIT.java
pyaroslav: Updated with your suggested wording.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleBlobDataTypesIT.java
pyaroslav: Done — switched all the new tests to `@FixFor("debezium/dbz#579")` and used the same form in the skip reasons.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleClobDataTypeIT.java
pyaroslav: Fixed here too, same wording as the BLOB test.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleClobDataTypeIT.java
pyaroslav: Done for the three CLOB tests as well.

Naros:
Hi @pyaroslav I made one specific change here about the LOB handling, and that is, if the reselect fails, the code will fall back to the unavailable value placeholder rather than passing the partial value, which would be interpreted as the "full value" rather than "partial", which I believe is systematically wrong. WDYT?

Naros (approved):
Just waits on CI.

Naros:
Applied, thanks @pyaroslav :tada:

pyaroslav:
Agreed — the placeholder is the right semantics. Emitting a partial `LOB_WRITE` chunk as if it were the full column value would be silently wrong downstream; `UNAVAILABLE_VALUE` tells consumers exactly what happened. Thanks for that change, for finishing the PDB re-pinning through `getOutOfBandsConnection()`, and for the merge. 🎉

PR #7377, opened 2026-04-23
debezium/dbz#188 Introduce Debezium offset storage interfaces

### LINKED ISSUE #188: Offset storage contract should be independent of Kafka Connect [DBZ-1971]
Migrated from [DBZ-1971](https://issues.redhat.com/browse/DBZ-1971)

Embedded engine and Debezium Server currently re-use the {{OffsetBackingStore}} contract from Kafka Connect. That can be confusing when implementing a custom store. We therefore should provide a Debezium-specific contract instead. This also applies to the configuration properties used e.g. to configure the file based store with the embedded engine.

One specific requirement we should address is to avoid the usage of FQNs in the configuration. Instead, each implementation should expose a short name such as "file", "kafka", "dynamodb" etc. This would be retrieved from a store-specific factory which is discovered via the service loader. In the configuration then those short names woul be used by default, and the engine looks up the right implementation via the service loader. FQNs might still be supported as a fallback, should there be multiple implementations exposing the same short name.

vjuranek:
Fixes debezium/dbz#188

## Description

This commit introduces Debezium storage interfaces for offset storage and outlines a way how to remove Kafka dependency from Debezium embedded module.

It creates two new modules:

* `debezium-storage-api` introduces new Debezium storage APIs. Currently it mimics Kafka APIs, but they are marked as incubating and will eventually evolve in subsequent commits or in the future.

* `debezium-kafka-adapters` contains adapters to Kafka classes. These adapters implement interfaces from `debezium-storage-api` and delegates the calls to the undelying Kafka instance. There are also adapters which implement Kafka APIs, but in the future Kafka APIs should be replaced by Debezium equivalents. The idea here is to move most of the Kafka stuff into dedicated module. It can be in the end coverted to an implementation of Debezium interfaces which uses Kafka as the Debezium runtime, but to be only one of several other possible implementations of Debezium runtime.

These modules are used in `debezium-embedded` module to start reducing Kafka classes usage and making them optional. Main change here is creation of offset store. Offset stores are loaded via SPI, using the loader name, eventually fully qualified class name. Legacy loading of the Kafka offset stores is still preseved as a backup. These Kafka offset stored are located in `debezium-storage-kafka` modules and are wrapped into adapter classes which implement Debezium `OffsetStoreProvider` API.

## PR Checklist


rk3rn3r:
Awesome! I will have a look.

debezium-storage/debezium-storage-api/src/main/java/io/debezium/spi/storage/OffsetStoreProvider.java
```diff
@@ -0,0 +1,64 @@
...
+     * such as creating converters or setting up connection pools.
+     * <p>
+     * Note: The caller is responsible for calling {@link OffsetStore#configure(Configuration)}
+     * and {@link OffsetStore#start()} on the returned instance.
+     *
+     * @param config the configuration
+     * @param connectorConfig the connector-specific configuration
+     * @return a configured offset backing store instance, never null
+     */
+    OffsetStore create(Configuration config, Map<String, String> connectorConfig);
```
mfvitale: Why both configuration? What is the difference?

debezium-kafka-adapters/src/main/java/io/debezium/storage/kafka/KafkaStorageAdapter.java
```diff
@@ -0,0 +1,41 @@
...
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.storage.kafka;
+
+/**
+ * Set of interfaces which mark implementations as Kafka-based and allows access to underlying Kafka instances.
+ *
+ * @author Debezium Authors
+ */
+public interface KafkaStorageAdapter {
```
mfvitale: Is `KafkaConnectStorageAdapter` more appropriate here?

mfvitale: Same for the other adapters

debezium-storage/debezium-storage-api/src/main/java/io/debezium/spi/storage/OffsetStoreProvider.java
```diff
@@ -0,0 +1,64 @@
...
+     * Returns the fully qualified class name of the {@link OffsetStore} implementation
+     * this provider creates.
+     * <p>
+     * This is used for backward compatibility, allowing configuration by class name to still work.
+     * <p>
+     * Default implementation returns null, indicating no class name mapping.
+     *
+     * @return the offset store class name, or null if not applicable
+     */
+    default String getOffsetStoreClassName() {
```
mfvitale: Maybe we could return an `Optional<String>`?

mfvitale:
@vjuranek The direction sound good to me. I left some comments.

debezium-kafka-adapters/src/main/java/io/debezium/source/kafka/KafkaSourceTaskContextAdapter.java
```diff
@@ -0,0 +1,68 @@
...
+import org.apache.kafka.common.metrics.PluginMetrics;
+import org.apache.kafka.connect.source.SourceTaskContext;
+import org.apache.kafka.connect.storage.OffsetStorageReader;
+
+/**
+ * Adapter that wraps Kafka Connect's {@link SourceTaskContext} to adapt it for Debezium's internal usage.
+ *
+ * @author Debezium Authors
+ */
+public class KafkaSourceTaskContextAdapter implements SourceTaskContext {
```
Naros: Should this extend the Kaka contract? It seems like this adapter isn't decoupling, right? The same holds true for the other class in this package, so I assume these are still WIP?

debezium-kafka-adapters/src/main/java/io/debezium/storage/kafka/KafkaStorageAdapter.java
vjuranek: I wanted to keep the names as short as possible, but yes, just `Kafka` prefix can be misleading. Will rename it.

debezium-storage/debezium-storage-api/src/main/java/io/debezium/spi/storage/OffsetStoreProvider.java
vjuranek: It's for historical reasons, but this is a good point, it shouldn't be used in brand-new interfaces. I'll do additional refactoring so that only one config is used.

debezium-storage/debezium-storage-api/src/main/java/io/debezium/spi/storage/OffsetStoreProvider.java
vjuranek: Good point, thanks.

debezium-kafka-adapters/src/main/java/io/debezium/source/kafka/KafkaSourceTaskContextAdapter.java
vjuranek: Ops, thanks for pointing this out. This is leftover from the step-by-step refactroring. In the code is actually used `getDelegate()`:
```
// TODO: remove switching to Kafka
                task.initialize(
                        new KafkaSourceTaskContextAdapter(taskContext.config(),
                                ((KafkaStorageAdapter.OffsetStorageReader) taskContext.offsetStorageReader()).getDelegate()).getDelegate()); // Initialize Kafka Connect source task
```
Will fix it in the next iteration.

vjuranek:
Now, all the `org.apache.kafka.connect.storage.*` imports are removed from Debezium embedded module and either replaced by Debezium alternative or moved into `debezium-kafka-adapters` or `debezium-storage-kafka` modules and no Kafka storage API is direcly used in the embedded module.

jpechane:
@vjuranek Do we want to emulate Connect API and use `Future` in the API? What is the adavnatage of having it there? Should we prescibe the asnc model?

vjuranek:
> @vjuranek Do we want to emulate Connect API and use `Future` in the API? What is the adavnatage of having it there? Should we prescibe the asnc model?

The main motivation was to have very similar API as Kafka Connect, at least in the initial step. However, we use it the synchronous manner and don't use callbacks (at least for now), so there is not strong reason why to provide async API. TBH, I don't have any strong opinion on it. I'd slightly prefer close similarity with Kafka API as it would make very easy to use anything from Kafka Connect in Debezium, but this is rather minor reason.

vjuranek:
Added migration of offset stores in `debezium-storage` module to new `OffsetStore` interface.

vjuranek:
Related PRs which should fix failing jobs:
* https://github.com/debezium/debezium-server/pull/286
* https://github.com/debezium/debezium-connector-vitess/pull/283

debezium-storage/debezium-storage-kafka/src/main/java/io/debezium/storage/kafka/offset/KafkaFileOffsetProvider.java
```diff
@@ -0,0 +1,77 @@
...
+        LOGGER.debug("Creating Kafka FileOffsetBackingStore");
+
+        // Create JSON converter for offset serialization
+        final Converter converter = createConverter();
+
+        // Create Kafka Connect FileOffsetBackingStore and wrap it
+        FileOffsetBackingStore kafkaStore = new FileOffsetBackingStore(converter);
+        if (config != null && !config.isEmpty()) {
+            WorkerConfig workerConfig = new EmbeddedWorkerConfig(config.asMap());
+            kafkaStore.configure(workerConfig);
```
Naros: Does this not cause a double configure when the adapter is configured below? I believe that violates the `OffsetStoreProvider` contract, if I'm not mistaken.

debezium-storage/debezium-storage-common/src/main/java/io/debezium/spi/storage/DefaultOffsetStorageReader.java
```diff
@@ -0,0 +1,107 @@
...
+                if (!serializedToOriginal.containsKey(rawEntry.getKey())) {
+                    LOGGER.error("Unexpected key {} in response from backing store", rawEntry.getKey());
+                    continue;
+                }
+                Map<String, T> origKey = serializedToOriginal.get(rawEntry.getKey());
+                if (rawEntry.getValue() == null) {
+                    continue;
+                }
+
+                byte[] valueBytes = rawEntry.getValue().array();
```
Naros: Is there any possibility that the ByteBuffer here is not heap-backed, where calling `array()` could lead to an exception because it's not supported, e.g., direct or read-only buffers?

Naros (commented):
Hi @vjuranek, great work. I've left a few inline questions.

debezium-storage/debezium-storage-kafka/src/main/java/io/debezium/storage/kafka/offset/KafkaFileOffsetProvider.java
mfvitale: I agree with @Naros. Should the `configure` be called on the adapter?

debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -662,6 +662,15 @@ public static SnapshotQueryMode parse(String value, String defaultValue) {
                     "since it is used as a prefix for all Kafka topic names that receive events emitted by this connector. " +
                     "Only alphanumeric characters, hyphens, dots and underscores must be accepted.");
 
+    public static final Field TASKS_MAX = Field.create("tasks.max")
```
mfvitale, line 670: Since this is a default Kafka Connect property, I wonder if we should put a in a separate place, also because this is effectively a configuration that we do not expose.

debezium-kafka-adapters/src/main/java/io/debezium/converter/kafka/KafkaConnectConverterAdapter.java
```diff
@@ -0,0 +1,52 @@
...
+
+    public KafkaConnectConverterAdapter(org.apache.kafka.connect.storage.Converter delegate) {
+        this.delegate = delegate;
+    }
+
+    public KafkaConnectConverterAdapter(Configuration converterConfig, String key, boolean isKey) {
+        final org.apache.kafka.connect.storage.Converter kafkaConverter = converterConfig.getInstance(key,
+                org.apache.kafka.connect.storage.Converter.class);
+        kafkaConverter.configure(converterConfig.asMap(), isKey);
+        this.delegate = kafkaConverter;
```
mfvitale, line 35: Same problem with potential double call to `configure`?

debezium-kafka-adapters/src/main/java/io/debezium/source/kafka/KafkaConnectSourceConnectorContextAdapter.java
```diff
@@ -0,0 +1,67 @@
...
+            }
+        };
+    }
+
+    /**
+     * Returns the underlying Kafka SourceConnectorContext.
+     *
+     * @return the wrapped SourceConnectorContext
+     */
+    public SourceConnectorContext getDelegate() {
```
mfvitale, line 64: Why we expose the delegate externally? Should't we just use the adapter and avoid to direct use the delegate?

mfvitale:
@vjuranek Great work. I left some comments

debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
Naros: Is this property even passed into Debezium from Kafka Connect or is it sanitized?

mfvitale: It's passed together with `name`. Initially it was not.

vjuranek: Yes, we use it, even in embedded engine ([here](https://github.com/debezium/debezium/blob/main/debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEmbeddedEngine.java#L165)). So this is not anything Kafka specific and e.g. for SQL server it can be something else then 1.

debezium-kafka-adapters/src/main/java/io/debezium/source/kafka/KafkaConnectSourceConnectorContextAdapter.java
vjuranek: We still depend a lot on Kafka stuff and at some point we have to switch to Kafka, so this (or at some other place) we have to expose the delegate. In this specific case we use it when initializing Kafka connector (class `org.apache.kafka.connect.connector.Connector` which requires `org.apache.kafka.connect.connector.ConnectorContext`). So in the future, when we replace Kafka Connect `Connector` class with our own `DebeziumConnector` class, it won't be needed, but unfortunately this is not the case yet.

debezium-storage/debezium-storage-kafka/src/main/java/io/debezium/storage/kafka/offset/KafkaFileOffsetProvider.java
vjuranek: Thanks for spotting this, you are right. Fixed.

debezium-kafka-adapters/src/main/java/io/debezium/converter/kafka/KafkaConnectConverterAdapter.java
vjuranek: This is actually not the case - it's just a shortcut for calling 
```
KafkaConnectConverterAdapter converter = new KafkaConnectConverterAdapter();
converter.configure(Configuration.from(...))
```
and this is never used in the code as KafkaConnectConverterAdapter is the only implementation and we always call the constructor with config (contrary to `OffsetStore`, which was really called twice on some places and violates explicit `OffsetStoreProvider` contract). 

It can be subject of future evolution of `Converter` interface if we lean towards splitting it into constructor and mandatory call of `configure()` method.

debezium-storage/debezium-storage-common/src/main/java/io/debezium/spi/storage/DefaultOffsetStorageReader.java
vjuranek: Great spot, indeed this could be an issue, depending on the implementation. Fixed.

vjuranek:
DBZ server failure (Redis sink) is caused by this change.

debezium-kafka-adapters/src/main/java/io/debezium/storage/kafka/KafkaConnectOffsetStorageWriterAdapter.java
```diff
@@ -0,0 +1,69 @@
...
+        delegate.offset(partition, offset);
+    }
+
+    @Override
+    public boolean beginFlush(long timeout, TimeUnit timeUnit) {
+        try {
+            return delegate.beginFlush(timeout, timeUnit);
+        }
+        catch (InterruptedException | TimeoutException e) {
+            return false;
```
Naros, line 53: Should we reset the interrupt flag here if an `InterruptedException` is thrown?

vjuranek: Ops, sure. Thanks for spotting this!

vjuranek:
Relates PRs:
* debezium server: https://github.com/debezium/debezium-server/pull/286
* vitess: https://github.com/debezium/debezium-connector-vitess/pull/283

debezium-bom/pom.xml
```diff
@@ -828,6 +828,11 @@
                 <artifactId>debezium-sink</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
```
Naros, line 841: Should this include `debezium-storage-api` and `debezium-kafka-adapters`?

debezium-storage/debezium-storage-api/src/main/java/io/debezium/spi/storage/OffsetStore.java
```diff
@@ -0,0 +1,112 @@
...
+     * Set the values for the specified key-value pairs.
+     * <p>
+     * The operation is asynchronous and returns immediately with a Future.
+     * The callback will be invoked when the operation completes.
+     *
+     * @param values the key-value pairs to store
+     * @param callback optional callback to invoke on completion (may be null)
+     * @return a Future that completes when the values are persisted
+     */
+    Future<Void> set(Map<ByteBuffer, ByteBuffer> values, OffsetStore.Callback<Void> callback);
```
Naros, line 74: This is more a long-term question: should this and `get` use something that is of a higher-level type (e.g., `Map<String, Object>`) for cases where offset store implementors currently need to handle serialization themselves?

Naros (commented):
Hi @vjuranek I found just a few things, see below.

debezium-bom/pom.xml
vjuranek: Definitely, I forgot on it. Fixed now.

debezium-storage/debezium-storage-api/src/main/java/io/debezium/spi/storage/OffsetStore.java
vjuranek: Yes, this is a good point, most of the implementation need to do the conversion to `ByteBuffer`. Would you like me to change it as part of this PR?

Naros: Since it's `@Incubating` and there isn't an immediate need, I do think leaving it as is works, and we can continue to evaluate if we need to adjust it as we get feedback. WDYT?

vjuranek: Yes, agree, we can adjust any time later as we need.

jpechane (approved):

PR #7378, opened 2026-04-23
[ci] Adjust the operator helm chart dir after QOSDK update

mfvitale:


jpechane:
@mfvitale Stable 3.4 and 3.5 are still on older QOSDK. Will the PR work for them too or not?

mfvitale:
> @mfvitale Stable 3.4 and 3.5 are still on older QOSDK. Will the PR work for them too or not?

Good catch. I fixed it and tests again with 3.6 release.

jpechane:
@mfvitale LGTM, thanks!!

PR #7381, opened 2026-04-23
 Add shutdown strategies in Debezium Engine debezium/dbz#1745

### LINKED ISSUE #1745: Add shutdown strategies in Debezium Engine
## Context

The Debezium Engine does not currently provide a way to shut down gracefully. In addition, there is no built‑in mechanism to stop the engine based on specific conditions, such as:

- when a snapshot has completed
- after processing a fixed number of events
- when the event queue has been fully drained

## Changes

We propose introducing a new new `shutdownStrategy` API to the `EmbeddedEngine`:

```java
EmbeddedEngine.create()
                .shutdownStrategy(event -> {
                    return false;
                })
                .using(config)
                .using(this.getClass().getClassLoader())
                .using(Clock.SYSTEM)
                .notifying(this::sendRecord)
                .build();
```

kmos:
Signed-off-by: kmos <kmos@commonhaus.dev><!-- Make sure all your commits are signed before submitting your pull request -->
<!-- Run `git commit -s` to sign off your commits to satisfy the DCO check -->
<!-- Ensure your commit messages start with your GitHub issue, e.g., debezium/dbz#<issue_number> -->
closes debezium/dbz#1745

## Description
This branch adds automatic shutdown strategies to the Debezium embedded engine, allowing it to gracefully stop based on configurable conditions during record processing.                                                      
                                                                                                                                                                                                                                 
 ### Key Changes                                                                                                                                                                                                                    
                  
  1. New Public API (`DebeziumEngine.java`)                                                                                                                                                                                        
  	- `ShutdownStrategy<T>` - Functional interface (extends `Predicate<T>`) for defining shutdown conditions
  	- `Watcher` interface - Observability interface to check if engine is actively consuming records
  	- New `shutdown()` method in `Builder<R>` - Entry point for configuring shutdown                                                                                                                               
  	- `ShutdownBuilder<R>` - Fluent builder API with:
    	- `before()` / `after()` - Position shutdown evaluation before or after consumer processing                                                                                                                                      
    	- `consuming()` - Specify consumption-based conditions                                                                                                                                                                         
    	- `records(int n)` - Simple count-based shutdown (e.g., stop after N records)                                                                                                                                                  
    	- `custom(ShutdownStrategy)` - Custom predicate-based shutdown logic                                                                                                                                                                                                                                                                                       
 
## PR Checklist
<!-- Please review the following checklist and mark items with an 'x' before submitting your pull request. -->


jpechane:
@kmos Good start! This is definitely working well for shutdowns based on records only.
But I beleive the sthudown stratgey be also configurable from connector configuration, so either full connector config or some partial namespace should be available po it.
Also one of the usecase was  shutdown based on connector state - i.e. shutdown affter snapshot is completed. Right now this can be achieved with notifications enabled but IMHO we need something a bit better.

kmos:
> @kmos Good start! This is definitely working well for shutdowns based on records only. But I beleive the sthudown stratgey be also configurable from connector configuration, so either full connector config or some partial namespace should be available po it. 

What do you mean with connector configuration? like a property to set: `connector.shutdown=SNAPSHOT_COMPLETED` 

> Also one of the usecase was shutdown based on connector state - i.e. shutdown affter snapshot is completed. Right now this can be achieved with notifications enabled but IMHO we need something a bit better.

On this point, I completely agree. I can address it in another PR.

jpechane:
> 
> What do you mean with connector configuration? like a property to set: `connector.shutdown=SNAPSHOT_COMPLETED`
> 

Yes and in the (custom) condition you can have somethig like var endMessage = config.get("connector.shutdown");
Or in case of the count it could be either configured either via code or connector config proeprty.

jpechane:
@kmos I wonder if we should introduce something like `ShutdownContext` that will contain the current record and connector config and can be extended in future with more functions so exisiting ShutDownStrategy implementations will be compatible whene there are changes done?

kmos:
> @kmos I wonder if we should introduce something like `ShutdownContext` that will contain the current record and connector config and can be extended in future with more functions so exisiting ShutDownStrategy implementations will be compatible whene there are changes done?

Well we can discuss about the content of `ShutdownContext` and for now could only contain the record. Something like `ShutdownContext<T>` where `T` is the `DebeziumEngine<T>` ? 

We should also think about the name :thinking: which it sounds meh

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
```diff
@@ -247,6 +486,70 @@ interface Builder<R> {
...
+         * <p><strong>Note:</strong> Shutdown configuration is optional. If not specified,
+         * the engine will run indefinitely until explicitly stopped via {@link #close()}
+         * or thread interruption.
+         *
+         * @return a shutdown builder for configuring automatic shutdown conditions; never null
+         * @see ShutdownBuilder
+         * @see ShutdownStrategy
+         * @since 3.6.0
+         */
+        default ShutdownBuilder<R> shutdown() {
```
vsantonastaso: I’m wondering if it makes sense for the default method to return null. Would it be better to make it abstract, or at least throw an exception instead? I think that might also improve readability. WDYT?

debezium-embedded/src/main/java/io/debezium/embedded/async/ShutdownConsumer.java
```diff
@@ -0,0 +1,45 @@
...
+                            ShutdownHandler<R> after,
+                            Consumer<R> consumer) {
+        this.before = before;
+        this.after = after;
+        this.consumer = consumer;
+    }
+
+    @Override
+    public void accept(R record) {
+        before.evaluate(record);
```
vsantonastaso, line 50: I have a doubt. The `before.evaluate(record)` may trigger the shutdown right? If so the `consumer.accept(record);` will be executed as well and the record consumed. Do we want it? Because you commented "*       If the condition is met, the record triggering shutdown is not processed.</li>" in the `debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java`

debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEmbeddedEngine.java
```diff
@@ -591,27 +596,53 @@ private String selectRecordProcessor() {
...
         throw new IllegalStateException("Unable to create RecordProcessor instance, this should never happen.");
     }
 
+    private Runnable shutdownWorkflow() {
+        return () -> {
+            try {
+                close();
+            }
+            catch (IOException e) {
+                throw new RuntimeException(e);
```
mfvitale: Should this be a `DebeziumException`?

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
mfvitale: Maybe an `Optional`?

vsantonastaso: why not abstract?

debezium-embedded/src/main/java/io/debezium/embedded/async/ShutdownConsumer.java
Naros: I agree. It's as if the `ShutdownStrategy` provided by the user needs to be wrapped by an internal implementation that flushes the offsets for the _after_ but doesn't for the _before_ use case.

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
mfvitale: It could be abstract to. I haven't a strong opinion on that.

kmos: My preference is to leave it not implemented. The problem is that it breaks the `ravapi` and for this reason I put a default implementation. How can I add a breaking change in the `DebeziumEngine` API ?

debezium-embedded/src/main/java/io/debezium/embedded/async/ShutdownConsumer.java
kmos: I left it like this to start a discussion: If we want this behavior, the `watcher` must be added before consumption, which requires evaluating the status for every message and introduces additional overhead and I am not sure about it. Do you have any suggestion ?

debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEmbeddedEngine.java
kmos: right!

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
mfvitale: > The problem is that it breaks the ravapi

Well in this case the `default` approach is the right one.

kmos: if I have to keep the method implemented, I prefer an exception

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
```diff
@@ -199,6 +200,56 @@ default boolean supportsTombstoneEvents() {
...
+     * the records being processed. This is a functional interface extending {@link Predicate}
+     * that evaluates each record to decide if shutdown conditions have been met.
+     *
+     * <p>Shutdown strategies can be configured to evaluate records either before or after
+     * they are passed to the consumer, using the {@link ShutdownBuilder} API:
+     *
+     * <pre>{@code
+     * // Shutdown after consuming exactly 100 records
+     * engine = DebeziumEngine.create(Connect.class)
+     *     .shutdown()
```
vjuranek: IIRC it was already mentioned somewhere, but I cannot find it ... I'd prefer something more self-container, like 
```
.withShutdownStrategy(
    ShutdownStrategyBuilder.create(ShutdownStrategy.AFTER)
        .consumed()
        .records(100)
        .build()
)
```
instead of mixing shurdown config options with high-level engine options.

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
```diff
@@ -224,6 +275,194 @@ interface Signaler {
...
+         * Provides access to engine-level state monitoring.
+         *
+         * @return the engine watcher; never null
+         */
+        EngineWatcher engine();
+
+        /**
+         * Monitors the consumption state of the engine.
+         */
+        interface EngineWatcher {
```
vjuranek: De we need dedicated implementation for it? To be able to work, engine would have to expose its state publicly and we have to pass engine instance into the watcher. Won't be more easy just to query the engine directly about its state? We can eventually add some watcher service later on, but I'd try to minimize changes into the public engine API as we have to keep and support public API for a long time.

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
kmos: Okay, but if we want to proceed with this approach, we would lose the event type in the `custom` API. For example:

```java
DebeziumEngine.create(Connect.class)
          .shutdown()
          .before()
          .consumed()
          .custom(record -> record.value().toString().contains("STOP"))
          .notifying(consumer)
          .build();
```

The record type is the `R` defined in `DebeziumEngine<R>`. In your case:

```java
.withShutdownStrategy(
    ShutdownStrategyBuilder.create(ShutdownStrategy.AFTER)
        .consumed()
        .custom(record -> record.value().toString().contains("STOP"))
        .build()
```

record type should be `Object`.

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
kmos: Yeah, I think we can do something better here. The interface is defined in `DebeziumEngine`, but a working implementation isn’t available outside the engine internals (see the Javadocs and the instantiation logic) and this service is not intended to be used outside. In fact, it’s used as an internal service to query the engine, because I’m not sure that exposing the engine’s state is the right way to proceed. We can remove the interface and use the service and later on think about it.

debezium-embedded/src/main/java/io/debezium/embedded/async/ShutdownConsumer.java
vsantonastaso: In my opinion, providing the ability to customize the shutdown behavior both before and after consumption would be very useful. So I think we should add a watcher guard before each consumption. Regarding the overhead, I’m not entirely sure how critical it might be, but at the moment I can’t identify any valid alternative approach.

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
vjuranek: Why the `ShutdownStrategy` cannot be parametrized in the similar way you have it now?

debezium-api/src/main/java/io/debezium/engine/DebeziumEngine.java
```diff
@@ -247,6 +328,21 @@ interface Builder<R> {
...
+         * <p>The policy is evaluated for every record: the {@link Shutdown#before() before} strategy
+         * is tested before the record reaches the consumer, and the {@link Shutdown#after() after}
+         * strategy is tested after the consumer has processed it.
+         *
+         *
+         * @param shutdown the shutdown policy; may be null to disable record-driven shutdown
+         */
+        default Builder<R> shutdown(Shutdown<R> shutdown) {
+            return null;
+        }
```
Naros, line 344: Shold we return `UnsupportedOperationException` or `DebeziumException` here rather than returning `null`?

debezium-embedded/src/main/java/io/debezium/embedded/async/DefaultShutdownHandler.java
```diff
@@ -0,0 +1,97 @@
...
+    public void evaluate(R record) {
+        if (shutdownStrategy.test(createContext(record))) {
+            try {
+                if (record instanceof EmbeddedEngineChangeEvent<?, ?, ?>) {
+                    committer.markProcessed(((EmbeddedEngineChangeEvent<?, ?, ?>) record).sourceRecord());
+                    committer.markBatchFinished();
+                }
+            }
+            catch (InterruptedException e) {
+                throw new RuntimeException(e);
```
Naros, line 55: Should this restore the interrupt flag?

debezium-embedded/src/main/java/io/debezium/embedded/async/DefaultShutdownHandler.java
```diff
@@ -0,0 +1,97 @@
...
+    }
+
+    @Override
+    public void evaluate(R record) {
+        if (shutdownStrategy.test(createContext(record))) {
+            try {
+                if (record instanceof EmbeddedEngineChangeEvent<?, ?, ?>) {
+                    committer.markProcessed(((EmbeddedEngineChangeEvent<?, ?, ?>) record).sourceRecord());
+                    committer.markBatchFinished();
+                }
```
Naros: Did we decide if this should mark the event processed when used as a `before` rather than `after` handler?

debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEmbeddedEngine.java
```diff
@@ -195,6 +200,7 @@ private AsyncEmbeddedEngine(Properties config,
         Map<String, String> internalConverterConfig = Collections.singletonMap(JsonConverterConfig.SCHEMAS_ENABLE_CONFIG, "false");
         offsetKeyConverter.configure(internalConverterConfig, true);
         offsetValueConverter.configure(internalConverterConfig, false);
+        this.watcher = () -> () -> State.POLLING_TASKS.compareTo(state.get()) == 0;
```
Naros:
```suggestion
        this.watcher = () -> () -> State.POLLING_TASKS.equals(state.get());
```

debezium-embedded/src/main/java/io/debezium/embedded/async/DefaultShutdownHandler.java
kmos: I have added a `watcher` before the `consumer` inside the `ShutdownChangeConsumer` which should address the use case. Or I missed something?

debezium-embedded/src/main/java/io/debezium/embedded/async/DefaultShutdownHandler.java
```diff
@@ -0,0 +1,98 @@
...
+        this.shutdown = shutdown;
+        this.committer = committer;
+        this.configuration = configuration;
+    }
+
+    @Override
+    public void evaluate(R record) {
+        if (shutdownStrategy.test(createContext(record))) {
+            try {
+                if (record instanceof EmbeddedEngineChangeEvent<?, ?, ?>) {
```
jpechane: Is not it possible to use pattern matchin instanceof?

Naros (approved):

mfvitale (approved):

debezium-embedded/src/main/java/io/debezium/embedded/async/DebeziumShutdown.java
```diff
@@ -0,0 +1,163 @@
...
+            return new DebeziumShutdown<>(this);
+        }
+
+        /**
+         * Secondary builder returned by {@link Builder#before()} and {@link Builder#after()} that
+         * selects the concrete shutdown strategy.
+         *
+         * @param <R> the record type produced by the engine
+         */
+        public interface ConsumingBuilder<R> {
```
vjuranek: I'm not good in naming, but I'd use something like `ShutdownStrategyBuilder` instead of `ConsumingBuilder`. `ConsumingBuilder` sounds little bit confusing to me.

vjuranek (commented):
Overall LGTM, just a two questions about (better?) naming and one regarding functionality.

debezium-embedded/src/main/java/io/debezium/embedded/async/DebeziumShutdown.java
vjuranek: I'm not good in naming, but I'd use something like `ShutdownStrategyBuilder` instead of `ConsumingBuilder`. `ConsumingBuilder` sounds little bit confusing to me.

debezium-embedded/src/main/java/io/debezium/embedded/async/ShutdownChangeConsumer.java
```diff
@@ -0,0 +1,62 @@
...
+                                  Watcher watcher) {
+        this.before = before;
+        this.after = after;
+        this.consumer = consumer;
+        this.watcher = watcher;
+    }
+
+    @Override
+    public void handleBatch(List<R> records, DebeziumEngine.RecordCommitter<R> committer) throws InterruptedException {
+        records.forEach(before::evaluate);
```
vjuranek, line 55: I'm not sure I follow here - if I have a batch of 10 records and shutdown strategy to shut down after 3 records, 3 records will trigger engine shutdown and the first two records won't be processed by the consumer as engine already shut down?

debezium-embedded/src/main/java/io/debezium/embedded/async/Watcher.java
```diff
@@ -0,0 +1,44 @@
...
+
+        /**
+         * Checks whether the engine is currently in the consuming/polling state,
+         * actively reading and processing records from the source.
+         *
+         * @return {@code true} if the engine is actively consuming records from the
+         *         source database; {@code false} if the engine is in any other state
+         *         (initializing, starting tasks, stopping, stopped, etc.)
+         */
+        boolean isConsuming();
```
vjuranek: Maybe `isPolling()` woiuld be more consistent with rest of the engine code, as we use "polling" for the state when engine is actually running and consuming records.

debezium-embedded/src/main/java/io/debezium/embedded/async/DebeziumShutdown.java
Naros: Haha, why does this give me pictures of the Cookie Monster in my head :thinking:

debezium-embedded/src/main/java/io/debezium/embedded/async/ShutdownChangeConsumer.java
kmos: Yeah, they are treated as a unique unit in the case of batch processing. Do you think this is wrong?

vjuranek: Not necessary wrong, but it may be confusing.

Naros (approved):

debezium-embedded/src/main/java/io/debezium/embedded/async/ShutdownChangeConsumer.java
kmos: Yeah I agree about the confusion. Based on the code, we should refactor a lot of pieces to address in the batch processing the shutdown for single event. Do you have any suggestion?

jpechane (approved):

PR #7382, opened 2026-04-23
debezium/dbz#1833 Fix PostgreSQL timestamp infinity overflow in nanos…

### LINKED ISSUE #1833: PostgreSQL timestamp infinity overflows in time.precision.mode=nanoseconds and becomes negative NanoTimestamp
## Bug report

```
[{"class":"io.debezium.connector.postgresql.PostgresConnector","type":"source","version":"3.3.1.Final"},{"class":"org.apache.kafka.connect.mirror.MirrorCheckpointConnector","type":"source","version":"3.6.1"},{"class":"org.apache.kafka.connect.mirror.MirrorHeartbeatConnector","type":"source","version":"3.6.1"},{"class":"org.apache.kafka.connect.mirror.MirrorSourceConnector","type":"source","version":"3.6.1"}]
```

---

```
    plugin.name: pgoutput
    publication.autocreate.mode: "disabled"
    snapshot.mode: initial

    table.include.list: <table list, separated by comma>
    topic.delimiter: "-"

    key.converter: org.apache.kafka.connect.json.JsonConverter
    value.converter: org.apache.kafka.connect.json.JsonConverter
    key.converter.schemas.enable: "false"
    value.converter.schemas.enable: "false"

    schema.history.internal.kafka.bootstrap.servers: <bootstrap servers>

    transforms: route,unwrap
    transforms.unwrap.type: io.debezium.transforms.ExtractNewRecordState
    transforms.route.type: io.debezium.transforms.ByLogicalTableRouter
    transforms.route.topic.regex: "<topic regex>"
    transforms.route.key.enforce.uniqueness: "false"
    topic.prefix: <prefix>
    database.dbname: <dbname>
    slot.name: <replication slot name>
    publication.name: <publication name>
    transforms.route.topic.replacement: "<topic replacement>"

    topic.creation.enable: "false"
    topic.creation.default.replication.factor: 3
    topic.creation.default.partitions: 10

    decimal.handling.mode: "string"
    time.precision.mode: "nanoseconds"
```

---

Deployment mode: AWS RDS PotgreSQL cluster
PotgreSQL: 14.17

---

One of the following would be preferable:

- Special handling for infinity / -infinity in time.precision.mode=nanoseconds so that overflow cannot occur
- represent infinity / -infinity in a way that preserves semantics for consumers

We can't switch to `time.precision.mode= isostring`, or `adaptive`.

---

**What behavior do you see?**

When using the Debezium PostgreSQL connector with: `"time.precision.mode": "nanoseconds"`

and a PostgreSQL TIMESTAMP column contains the special value infinity, the value emitted to Kafka becomes: `-29575808000`

This looks like an int64 overflow during conversion to nanoseconds. Why this looks incorrect

Debezium documents PostgreSQL TIMESTAMP infinity as the numeric sentinel: `9223372036825200000`

That value is already very close to Long.MAX_VALUE: `Long.MAX_VALUE = 9223372036854775807`

If time.precision.mode=nanoseconds converts the documented infinity representation by multiplying it by 1000, the result is: `9223372036825200000 * 1000 = 9223372036825200000000` which does not fit into signed int64.

The overflowed signed 64-bit result is exactly: `-29575808000`

So it appears that timestamp infinity is first converted to Debezium’s documented numeric sentinel and then multiplied by 1000 for NanoTimestamp, causing overflow and producing a finite negative value in Kafka.

Why this is a problem

The emitted value: -29575808000 is no longer recognizable as infinity.

From a consumer perspective, this is problematic because:

- PostgreSQL contains a special non-finite value (infinity)
- Debezium emits a regular finite negative timestamp
- downstream consumers cannot distinguish this from a real timestamp value
- the semantic meaning of the original database value is lost

---

We can't upgrade debezium version because of Kafka version

---

No ERROR or WARN logs detected, connectors working fine and transfers data successfully, the main problem is the exact value.

---

Example table:

```
CREATE TABLE test_infinity (
    id INT PRIMARY KEY,
    ts TIMESTAMP
);
```

`INSERT` INTO test_infinity (id, ts) VALUES (1, `'infinity');`

Example connector configuration:
```
{
  "name": "pg-test",
  "config": {
    "plugin.name": "pgoutput",
    "database.port": "5432",
    "database.user": "postgres",
    "topic.prefix": "test",
    "table.include.list": "public.test_infinity",
    "time.precision.mode": "nanoseconds"
  }
}
```
Observed record payload contains: `"ts": -29575808000`

jpechane:
Fixes debezium/dbz#1833

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java
@@ -1186,6 +1186,33 @@ public class PostgresValueConverter extends JdbcValueConverters {
     }
 
     @Override
+    protected Object convertTimestampToEpochNanos(Column column, Field fieldDefn, Object data) {
+        if (data == null) {
+            return null;
+        }
+
+        if (data instanceof Instant instant) {
+            if (POSITIVE_INFINITY_INSTANT.equals(instant)) {
+                return Long.MAX_VALUE;
+            }
+            else if (NEGATIVE_INFINITY_INSTANT.equals(instant)) {
+                return Long.MIN_VALUE;
+            }
+        }
+
+        if (data instanceof LocalDateTime localDateTime) {
+            if (POSITIVE_INFINITY_LOCAL_DATE_TIME.equals(localDateTime)) {
+                return Long.MAX_VALUE;
+            }
+            else if (NEGATIVE_INFINITY_LOCAL_DATE_TIME.equals(localDateTime)) {
+                return Long.MIN_VALUE;
+            }
+        }
+
+        return super.convertTimestampToEpochNanos(column, fieldDefn, data);
+    }
+
+    @Override
     protected int getTimePrecision(Column column) {
         return column.scale().orElse(-1);
     }
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresTemporalPrecisionHandlingIT.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresTemporalPrecisionHandlingIT.java
@@ -344,6 +344,56 @@ public class PostgresTemporalPrecisionHandlingIT extends AbstractAsyncEngineConn
     }
 
     @Test
+    void shouldConvertInfinityTimestampsToLongMinMax() throws Exception {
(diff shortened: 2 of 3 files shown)


jpechane:
Test failure unrelated

vjuranek (approved):
LGTM

vjuranek:
Merged, thanks @jpechane !

PR #2326, opened 2021-04-23
DBZ-3458 No need to exclude test scoped dependency from assembly

gunnarmorling:
https://issues.redhat.com/browse/DBZ-3458

PR #7388, opened 2026-04-27
debezium/dbz#1848 Preserve nulls in ByLogicalTableRouter

### LINKED ISSUE #1848: ByLogicalTableRouter materializes schema defaults for nullable fields, clobbering real NULL values
## Bug report

MySQL connector, `3.0.7.Final`. The bug is also present on `main` (code inspection — the two call sites are unchanged).

---

Relevant portion. Any `ByLogicalTableRouter` config is affected; the Avro converter flag is included only to illustrate why the bug matters (see "behavior you see" below):

```json
{
  "transforms": "table_router",
  "transforms.table_router.type": "io.debezium.transforms.ByLogicalTableRouter",
  "transforms.table_router.topic.regex": "(.+)\\.(.+)\\.(.+)",
  "transforms.table_router.topic.replacement": "$1.$3",
  "transforms.table_router.key.enforce.uniqueness": "false",
  "value.converter.ignore.default.for.nullables": "true"
}
```

The source table has a column of the form `flag tinyint(1) NULL DEFAULT 0` (any nullable column with a non-null schema default reproduces it — `INT NULL DEFAULT 0`, `VARCHAR NULL DEFAULT ''`, etc.).

---

MySQL 8, self-hosted on GCP. The bug is not MySQL-specific — it's in the SMT. Any source connector using `ByLogicalTableRouter` whose records contain nullable columns with non-null defaults is affected.

---

A real `NULL` in the source survives the SMT. When `value.converter.ignore.default.for.nullables=true` is set on the Confluent AvroConverter (added in CP 7.3 via [schema-registry#2314](https://github.com/confluentinc/schema-registry/issues/2314) / [PR#2326](https://github.com/confluentinc/schema-registry/pull/2326)), the published Avro record carries the `null` branch of the field's union.

---

**What behavior do you see?**

`ByLogicalTableRouter.updateValue` (and `.updateKey`) rebuilds the `Struct` by iterating fields and calling `oldX.get(field)`. `Struct.get(Field)` substitutes the schema default whenever the slot is null, so the new `Struct` stores the default (e.g. `0`) in place of the real `NULL`. Downstream consumers see `0`.

This also neutralizes the Confluent AvroConverter's `ignore.default.for.nullables` fix: by the time the converter runs, the `Struct` no longer contains a `null` for it to preserve.

Offending code on `main` (`debezium-connect-plugins/src/main/java/io/debezium/transforms/ByLogicalTableRouter.java`):

- `updateValue`, L412-L418 — `newValue.put(field.name(), oldValue.get(field));`
- `updateKey`, L341-L367 — `newKey.put(field.name(), oldKey.get(field));`

Same code on tag `v3.0.7.Final` at the pre-rename path `debezium-core/src/main/java/io/debezium/transforms/ByLogicalTableRouter.java`.

**Proposed fix:** use `getWithoutDefault(field.name())` at both call sites. Happy to open a PR with a regression test in `ByLogicalTableRouterTest`.

---

Confirmed via code inspection on `main` — the two call sites are unchanged. The bug is reproducible without a live database via a `Struct` fixture:

```java
Schema flagSchema = SchemaBuilder.int16().optional().defaultValue((short) 0).build();
Schema valueSchema = SchemaBuilder.struct().name("shard_1.example.Value")
    .field("id", Schema.INT64_SCHEMA)
    .field("flag", flagSchema)
    .build();
Schema sourceSchema = SchemaBuilder.struct().name("source")
    .field("table", Schema.STRING_SCHEMA)
    .build();

Envelope envelope = Envelope.defineSchema()
    .withName("shard_1.example.Envelope")
    .withRecord(valueSchema)
    .withSource(sourceSchema)
    .build();

Struct after = new Struct(valueSchema).put("id", 1L).put("flag", null);
Struct source = new Struct(sourceSchema).put("table", "example");
Struct envelopeValue = envelope.update(after, after, source, Instant.now());

SourceRecord in = new SourceRecord(null, null, "shard_1.example.example", null,
        null, null, envelope.schema(), envelopeValue);

ByLogicalTableRouter<SourceRecord> smt = new ByLogicalTableRouter<>();
smt.configure(Map.of(
    "topic.regex", "(.+)\\.(.+)\\.(.+)",
    "topic.replacement", "$1.$3",
    "key.enforce.uniqueness", "false"));

SourceRecord out = smt.apply(in);
Struct outAfter = ((Struct) out.value()).getStruct("after");

outAfter.getWithoutDefault("flag");
// Expected: null
// Actual:   (short) 0
```

---

Not relevant here — the bug is deterministic inside the SMT's transform path, not a lifecycle issue. The unit-level reproducer above exercises the same code path without needing a connector. Happy to provide DEBUG logs from a live deployment if it would help triage.

---

The `Struct`-fixture reproducer above is simpler and doesn't need the tutorial stack. To reproduce via the tutorial:

1. Start the standard tutorial MySQL stack, but replace `JsonConverter` with `AvroConverter` + a Confluent Schema Registry container on both key and value.
2. Add a `ByLogicalTableRouter` SMT to the MySQL source connector config:
   ```json
   "transforms": "logical_table",
   "transforms.logical_table.type": "io.debezium.transforms.ByLogicalTableRouter",
   "transforms.logical_table.topic.regex": "(.+)\\.inventory\\.(.+)",
   "transforms.logical_table.topic.replacement": "$1.inventory.$2",
   "transforms.logical_table.key.enforce.uniqueness": "false"
   ```
3. Add a nullable-with-default column to one of the sample tables:
   ```sql
   ALTER TABLE inventory.customers ADD COLUMN flag TINYINT(1) NULL DEFAULT 0;
   ```
4. `UPDATE inventory.customers SET flag = NULL WHERE id = 1001;`
5. Consume the resulting topic with an Avro-aware consumer.
   - Observed: `flag` is `0` in the published record.
   - Expected: `flag` is `null`.

willhorn:
Closes debezium/dbz#1848.

## Summary

`ByLogicalTableRouter` rebuilds the key and value `Struct` for every record so it can apply a new logical schema name. Inside that rebuild, both `updateKey` and `updateValue` read source field values via `oldStruct.get(field)`, which substitutes the schema default whenever the slot is null. For any nullable column that carries a non-null default (e.g. MySQL `TINYINT(1) NULL DEFAULT 0`, `INT NULL DEFAULT 0`), real `NULL` values are silently replaced by the default before the record reaches the value converter. Downstream consumers see the default in place of the original null.

This also makes Confluent's `value.converter.ignore.default.for.nullables=true` AvroConverter fix (CP 7.3, schema-registry [#2326](https://github.com/confluentinc/schema-registry/pull/2326)) effectively a no-op for any pipeline that includes `ByLogicalTableRouter`, because by the time the converter runs the `Struct` no longer holds a `null` for it to preserve.

## Change

Switch both `updateKey` and `updateValue` to read field values via `Struct.getWithoutDefault(String)`, which returns the raw stored slot exactly as it was set. Real `NULL` values in columns with non-null defaults now survive the SMT.

## How was it tested?

* New regression test `ByLogicalTableRouterTest#shouldPreserveNullForNullableFieldWithNonNullDefault` constructs a record with a nullable `INT16` field carrying a non-null default (`(short) 0`) and a real `null` value in both the key `Struct` and the envelope's `before`/`after` `Struct`s. It applies the SMT and asserts that all three field positions still report `null` via `Struct.getWithoutDefault` after the rebuild. The test fails on `main` and passes with the fix.
* `./mvnw -pl debezium-connect-plugins -am test` is green for the full module test suite (11 tests, including the new one).
* The full build (`./mvnw clean install -DskipITs`) also runs `Debezium Kafka Connect Plugins` cleanly. An unrelated container-startup failure in `debezium-connector-mysql`'s `KafkaSchemaHistoryTest` was observed locally; it does not touch any code path changed in this PR.
* I observed the bug end-to-end in a real Debezium 3.0.7.Final + MySQL pipeline (a record produced by an `UPDATE ... SET col = NULL` against a nullable-with-default column) and verified that backporting this same fix into a local fork of `ByLogicalTableRouter` resolves the issue and lets the resulting Avro records carry the `null` branch of the field's union.

## AI assistance disclosure

Per `AI_USAGE_POLICY.md`, I used Claude (Anthropic) assistively while investigating this bug, reading the relevant Debezium and Kafka Connect source, drafting the fix, and drafting the regression test. I reviewed every line, ran the test suite locally to verify the change, observed the bug in our own pipeline beforehand, and verified that the same fix resolves it after deploying a local fork of the SMT to our dev environment. I am responsible for the contents of this PR.

## PR Checklist

### DIFF
--- a/debezium-connect-plugins/src/main/java/io/debezium/transforms/ByLogicalTableRouter.java
+++ b/debezium-connect-plugins/src/main/java/io/debezium/transforms/ByLogicalTableRouter.java
@@ -342,7 +342,7 @@ public class ByLogicalTableRouter<R extends ConnectRecord<R>> implements Transfo
     private Struct updateKey(Schema newKeySchema, Struct oldKey, String oldTopic) {
         final Struct newKey = new Struct(newKeySchema);
         for (org.apache.kafka.connect.data.Field field : oldKey.schema().fields()) {
-            newKey.put(field.name(), oldKey.get(field));
+            newKey.put(field.name(), oldKey.getWithoutDefault(field.name()));
         }
 
         String physicalTableIdentifier = oldTopic;
@@ -413,7 +413,7 @@ public class ByLogicalTableRouter<R extends ConnectRecord<R>> implements Transfo
     private Struct updateValue(Schema newValueSchema, Struct oldValue) {
         final Struct newValue = new Struct(newValueSchema);
         for (org.apache.kafka.connect.data.Field field : oldValue.schema().fields()) {
-            newValue.put(field.name(), oldValue.get(field));
+            newValue.put(field.name(), oldValue.getWithoutDefault(field.name()));
         }
         return newValue;
     }
--- a/debezium-connect-plugins/src/test/java/io/debezium/transforms/ByLogicalTableRouterTest.java
+++ b/debezium-connect-plugins/src/test/java/io/debezium/transforms/ByLogicalTableRouterTest.java
@@ -8,6 +8,7 @@ package io.debezium.transforms;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 
+import java.time.Instant;
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
@@ -19,6 +20,7 @@ import org.apache.kafka.connect.errors.ConnectException;
 import org.apache.kafka.connect.source.SourceRecord;
 import org.junit.jupiter.api.Test;
 
+import io.debezium.data.Envelope;
 import io.debezium.doc.FixFor;
 import io.debezium.relational.history.HistoryRecord.Fields;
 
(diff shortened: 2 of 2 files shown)


Naros (changes requested):
There is one additional use case that is missed in `updateEnvelope`.  Can you also address it there as well, please @willhorn ?

willhorn:
> There is one additional use case that is missed in `updateEnvelope`. Can you also address it there as well, please @willhorn ?

@Naros I'm looking at [this line](https://github.com/willhorn/debezium/blob/2ad15abccbf60ee9accd05c7f30fa77a1cb95b76/debezium-connect-plugins/src/main/java/io/debezium/transforms/ByLogicalTableRouter.java#L402). I'm a bit reluctant to change that. The other changes are specific to fields within the `before` and `after` structs -- i.e. they correspond to database fields and address the bug I encountered. If I understand correctly, the line in `updateEnvelope` affects non-database fields within the change log schema. I'm less confident about how `null` values should be treated for these, and they don't address the bug. However, if you're confident the change should be made there as well, I'll add a test make the change.

Naros (approved):

PR #7393, opened 2026-04-29
debezium/dbz#1561 Detect network errors in MySqlConnection to prevent misleading SHOW MASTER STATUS fallback

### LINKED ISSUE #1561: Missleading error message when loss of database connectivity happens with mysql >= 8.4.X
## Bug report

Debezium-connector-mysql version 3.3.2.Final.

Bundle was downloaded from https://debezium.io/releases/3.3/

This bundle has been baked in a kafka-3.6.1 docker image that is then deployed in a kafka-connect cluster via the strimzi operator.

---

  config:
    database.user: [REDACTED]
    database.password: [REDACTED]
    database.server.id: 11
    schema.history.internal.kafka.bootstrap.servers: ilot-cluster-kafka-bootstrap:9092
    schema.history.internal.kafka.topic: schemahistory.my-connector
    database.port:  [REDACTED]
    database.hostname:  [REDACTED]
    database.include.list: my_db
    table.include.list: my_db.outbox
    topic.prefix: my.prefix
    transforms: outbox
    transforms.outbox.type: io.debezium.transforms.outbox.EventRouter
    transforms.outbox.route.by.field: topic
    transforms.outbox.table.field.event.id: kafka_key
    transforms.outbox.table.field.event.key: kafka_key
    transforms.outbox.table.field.event.payload.id: kafka_key
    transforms.outbox.table.field.event.payload: payload
    transforms.outbox.table.field.event.timestamp : dt
    transforms.outbox.route.topic.replacement: my.prefix.${routedByValue}
    transforms.outbox.table.expand.json.payload: false
    transforms.outbox.table.fields.additional.placement: tp:header:__TypeId__
    key.converter: org.apache.kafka.connect.storage.StringConverter
    key.converter.schemas.enable: false
    value.converter: org.apache.kafka.connect.storage.StringConverter
    value.converter.schemas.enable: false
    event.deserialization.failure.handling.mode: ignore
    database.history.skip.unparseable.ddl: true
    schema.history.internal.skip.unparseable.ddl: true
    inconsistent.schema.handling.mode: skip
    database.history.store.only.monitored.tables.ddl: true
    use.nongraceful.disconnect: true
    snapshot.mode: no_data

---

On-premises InnoDB mysql cluster, version 8.4.5 running on cluster mode with 2 replicaSets.
The connectors connect to the innodb clusters through multiple mysql-routers that
have been deployed in kubernetes as a replicaSet. 

---

**What is the source of the issue?**

The source of this seems to be the constructor of the MysqlConnection class (https://github.com/debezium/debezium/blob/v3.3.2.Final/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java)

The logic implemented tries to execute the  Mysql.Version>=8.4.0 expected binlog command, `SHOW BINARY LOG STATUS` , and if that errors out it falls back to the Mysql.Version<8.4.0 version, `SHOW MASTER STATUS`.

So the MysqlConnection instance created will have the variable binaryLogStatusStatement  set to 'SHOW MASTER STATUS'.
From this point any additional GTID related function execution will always fail with the error message mentioned above, like the functions isGtidModeEnabled() or knownGtidSet(), even if the network connectivity issue was established again, since this sql command will never be successfully executed on a mysql >=8.4.0.

---

**What behavior do you see?**

Upon facing issues related to the database connectivity from the cluster due to new members joining the cluster and
adding a huge load due to replication of a high volume, the routers lost connectivity for some time. Based on the logs
we have captured, my estimation is 1 minute.

The connector then reported that the issue that happened was due to the sql command `SHOW MASTER STATUS`
failing to execute successfully.

Image

---

The connector should remain in a failed state but the error message should be accurate. 
An error message such as:
- Driver did not receive packets for a long time
- Database connection lost
- Cannot establish connection to the database
or basically any other of the related and common network/database errors that should have been emitted.

This means that the constructor should have failed instantiation and should have thrown the correct error
message instead instead of setting the wrong binaryLogStatusStatement value.

---

This has only been tested on 3.1.1.Final and 3.3.2.Final.

---

N/A

---

I did not have the chance to try and reproduce the error in a test environment but this has happened multiple days in the past few days

aryadharmadhikari:
## Description
- Fixes debezium/dbz#1561

## Problem
When the MySQL connector loses database connectivity during the constructor's version detection, it silently falls back to using `SHOW MASTER STATUS` on MySQL 8.4+ servers. This causes all subsequent operations like `knownGtidSet()` to fail with a misleading error about `SHOW MASTER STATUS` not being supported, rather than reporting the actual network connectivity issue.

## Solution
Added network error detection using JDBC SQLState codes:
- Network errors (SQLState 08xxx) now throw `DebeziumException` with a clear message
- Non-network errors continue to trigger version detection and legitimate fallback
- MySQL version is logged when falling back to aid debugging

## Changes
- Made `isNetworkError()` a static method that checks for connection-related SQLState codes
- Updated constructor to call `isNetworkError()` before attempting fallback
- Added unit tests for `isNetworkError()` covering all edge cases

## Testing
- [x] Unit tests verify SQLState detection logic
- [x] Full MySQL connector test suite passes
- Manual verification pending: constructor behavior with actual network failures

## Future Enhancement
A version compatibility check could be added to detect cases where `SHOW BINARY LOG STATUS` fails on MySQL 8.4+ for reasons other than network errors (e.g., permissions). I kept this out of the current PR to keep the fix minimal and focused on the reported issue, but happy to add it or track it separately if it feels worthwhile.

## Flowchart
Screenshot 2026-04-24 at 11 36 35 PM

## PR Checklist

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java
@@ -46,11 +46,23 @@ public class MySqlConnection extends BinlogConnectorConnection {
                 String version = queryAndMap("SELECT VERSION()", rs -> {
                     return rs.next() ? rs.getString(1) : "unknown";
                 });
-                LOGGER.info("MySQL version {} detected, falling back to '{}'",
-                        version, MASTER_STATUS_STATEMENT);
+
+                String[] parts = version.split("\\.");
+                int major = Integer.parseInt(parts[0]); // 8
+                int minor = Integer.parseInt(parts[1]); // 4
+
+                if (major > 8 || (major == 8 && minor >= 4)) {
+                    throw new DebeziumException("MySQL version " + version +
+                            " should support SHOW BINARY LOG STATUS but it failed. Check database permissions or connectivity.", e);
+                }
+
+                LOGGER.info("MySQL version {} detected, falling back to '{}'", version, MASTER_STATUS_STATEMENT);
             }
             catch (SQLException ve) {
                 LOGGER.warn("Could not determine MySQL version during fallback detection: {}", ve.getMessage());
+            }
+            catch (NumberFormatException nfe) {
+                LOGGER.warn("Could not parse MySQL version string during fallback detection: {}", nfe.getMessage());
             }
 
             binaryLogStatusStatement = MASTER_STATUS_STATEMENT;
@@ -60,9 +72,17 @@ public class MySqlConnection extends BinlogConnectorConnection {
         binaryLogStatusStatement = BINARY_LOG_STATUS_STATEMENT;
     }
 
-    public boolean isNetworkError(SQLException e) {
+    public static boolean isNetworkError(SQLException e) {
         String sqlState = e.getSQLState();
-        return sqlState != null && sqlState.startsWith("08");
+        if (sqlState == null) {
+            return false;
+        }
+
+        // Network-related connection failures (not config/auth errors)
+        return sqlState.equals("08001") // Cannot connect to server
+                || sqlState.equals("08003") // Connection does not exist
+                || sqlState.equals("08006") // Connection failure
+                || sqlState.equals("08S01"); // Communication link failure
     }
 
     public String binaryLogStatusStatement() {
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java
@@ -38,12 +38,31 @@ public class MySqlConnection extends BinlogConnectorConnection {
             });
         }
         catch (SQLException e) {
-            LOGGER.info("Using '{}' to get binary log status", MASTER_STATUS_STATEMENT);
+            if (isNetworkError(e)) {
+                throw new DebeziumException("Failed to establish connection with the database: ", e);
+            }
+
+            try {
+                String version = queryAndMap("SELECT VERSION()", rs -> {
+                    return rs.next() ? rs.getString(1) : "unknown";
+                });
+                LOGGER.info("MySQL version {} detected, falling back to '{}'",
+                        version, MASTER_STATUS_STATEMENT);
+            }
+            catch (SQLException ve) {
+                LOGGER.warn("Could not determine MySQL version during fallback detection: {}", ve.getMessage());
+            }
+
             binaryLogStatusStatement = MASTER_STATUS_STATEMENT;
             return;
         }
         LOGGER.info("Using '{}' to get binary log status", BINARY_LOG_STATUS_STATEMENT);
         binaryLogStatusStatement = BINARY_LOG_STATUS_STATEMENT;
+    }
+
+    public boolean isNetworkError(SQLException e) {
+        String sqlState = e.getSQLState();
+        return sqlState != null && sqlState.startsWith("08");
     }
 
     public String binaryLogStatusStatement() {
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java
@@ -33,9 +33,28 @@ public class MySqlConnection extends BinlogConnectorConnection {
     public MySqlConnection(MySqlConnectionConfiguration connectionConfig, BinlogFieldReader fieldReader) {
         super(connectionConfig, fieldReader);
 
+        this.binaryLogStatusStatement = resolveBinaryLogStatusStatement();
+    }
+
+    public static boolean isNetworkError(SQLException e) {
+        String sqlState = e.getSQLState();
+        if (sqlState == null) {
+            return false;
+        }
+
+        return sqlState.equals("08001") // Cannot connect to server
+                || sqlState.equals("08003") // Connection does not exist
+                || sqlState.equals("08004") // Connection rejected by server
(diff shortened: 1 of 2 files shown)


debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java
```diff
@@ -38,14 +38,53 @@ public MySqlConnection(MySqlConnectionConfiguration connectionConfig, BinlogFiel
...
+                LOGGER.info("MySQL version {} detected, falling back to '{}'", version, MASTER_STATUS_STATEMENT);
+            }
+            catch (SQLException ve) {
+                LOGGER.warn("Could not determine MySQL version during fallback detection: {}", ve.getMessage());
+            }
+            catch (NumberFormatException nfe) {
+                LOGGER.warn("Could not parse MySQL version string during fallback detection: {}", nfe.getMessage());
+            }
+
             binaryLogStatusStatement = MASTER_STATUS_STATEMENT;
```
Naros: At this point, could we move this logic into its own private method, given its breadth?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java
```diff
@@ -38,14 +38,53 @@ public MySqlConnection(MySqlConnectionConfiguration connectionConfig, BinlogFiel
...
+        String sqlState = e.getSQLState();
+        if (sqlState == null) {
+            return false;
+        }
+
+        // Network-related connection failures (not config/auth errors)
+        return sqlState.equals("08001") // Cannot connect to server
+                || sqlState.equals("08003") // Connection does not exist
+                || sqlState.equals("08006") // Connection failure
+                || sqlState.equals("08S01"); // Communication link failure
```
Naros: My understanding is that MySQL classifies all `08nnn` state codes as communication-related. In that case, do we think it would be better, and future-proof, to check that the code is simply prefixed by `08`?

The reason I ask is that `08004` is too many connections, too.

aryadharmadhikari: @Naros at first I tried simply the 08 prefix check but when I ran the integration check the build kept failing, only for me to realize that it catches TLS configuration errors as "network errors" when they are really configuration errors. So I felt that fix was too broad so I narrowed it down to these specific 08 codes.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java
aryadharmadhikari: @Naros sure I will move it to a new private helper method.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java
aryadharmadhikari:
``` java 
public static boolean isNetworkError(SQLException e) {
    String sqlState = e.getSQLState();
    if (sqlState == null) {
        return false;
    }
    return sqlState.equals("08001") // Cannot connect to server
                || sqlState.equals("08003") // Connection does not exist
                ||sqlState.equals("08004");
                || sqlState.equals("08006") // Connection failure
                || sqlState.equals("08S01"); // Communication link failure
}
```
I suppose this might work, having some issues with the Integration tests, will resolve them meanwhile.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/MySqlConnection.java
```diff
@@ -42,28 +42,7 @@ public MySqlConnection(MySqlConnectionConfiguration connectionConfig, BinlogFiel
...
-
-                LOGGER.info("MySQL version {} detected, falling back to '{}'", version, MASTER_STATUS_STATEMENT);
-            }
-            catch (SQLException ve) {
-                LOGGER.warn("Could not determine MySQL version during fallback detection: {}", ve.getMessage());
-            }
-            catch (NumberFormatException nfe) {
-                LOGGER.warn("Could not parse MySQL version string during fallback detection: {}", nfe.getMessage());
-            }
+            validateVersionForFallback(e);
```
Naros: I think GitHub didn't highlight everything as I had hoped. Sorry for that. But my intent was to have the entire try, including where we call `query(BINARY_LOG_STATUS_STATEMENT, rs -> { });`, be placed into a method that simply returns the value to assign to `binaryLogStatusStatement`. This way, all the logic for resolving this, including its fallback, is encapsulated in a single method.

But if you want to keep just the fallback encapsulated, that's fine. It's just a preference.

aryadharmadhikari: @Naros, I agree, this approach is much cleaner. I have refactored the logic to encapsulate the entire logic within the resolveBinaryLogStatusStatement method. This keeps the constructor to a simple assignment. I have amended the commit and force-pushed the update:)

vjuranek (commented):
LGTM, just please  check and fix (`mvn process-sources`) formatting to avoid formatting check failures.

aryadharmadhikari:
@vjuranek thanks for the heads up, I have amended the commit fixing the formatting.
I was holding off on the push because I was troubleshooting the Integration Test failures. While my unit tests for are passing, the ITs which I managed to build successfully 5 days ago, now failed. Was trying to fix but it failed persistently:(

Naros (approved):

Naros:
Thanks @aryadharmadhikari, great work!

PR #7399, opened 2026-04-30
ORA-26913 RAC instance pinning issue for XStream adapter

### LINKED ISSUE #1880: Document ORA-26913 RAC instance pinning issue for XStream adapter
The current XStream documentation does not mention that ORA-26913 can occur in Oracle RAC environments when the JDBC connection is load-balanced to a different RAC instance than the one where the XStream outbound server is running.

When database.hostname and database.port are used, the Oracle listener can route the connection to any available RAC instance. If it lands on the wrong instance, XStream attach fails with:

oracle.streams.StreamsException: ORA-26913: must connect to instance X where XStream Outbound server is running

Additionally, the documentation does not mention that a PDB service name must not be used, as the XStream outbound server is defined at CDB level and connecting via a PDB service results in ORA-26701: Streams process does not exist.

The fix is to use database.url with a full TNS descriptor specifying an instance-pinned SERVICE_NAME and ENABLE=broken for TCP keepalive.

Add a FAQ entry to documentation/modules/ROOT/pages/connectors/oracle.adoc documenting the root cause and fix.

aqhil07:
Fixes debezium/dbz#1880

Add a FAQ entry to the Oracle Xstream connector documentation explaining ORA-26913 
which occurs in Oracle RAC environments when the XStream outbound server is 
pinned to a specific RAC instance but the JDBC connection is load-balanced 
to a different instance.

Documents the fix: use database.url with a full TNS descriptor specifying 
an instance-pinned SERVICE_NAME and ENABLE=broken for TCP keepalive.

Also documents that a PDB service name must not be used as the XStream 
outbound server is defined at CDB level.

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/oracle.adoc
+++ b/documentation/modules/ROOT/pages/connectors/oracle.adoc
@@ -6242,3 +6242,23 @@ yum install -y libaio
 ----
 apt-get install -y libaio1
 ----
+
+*What causes `ORA-26913` when using XStream with Oracle RAC?*::
+This error only occurs in Oracle Real Application Clusters (RAC) environments and does not affect standalone Oracle installations.
++
+In a RAC environment, the XStream outbound server is pinned to a specific RAC instance.
+When the JDBC connection is load-balanced by the Oracle listener to a different instance, XStream cannot attach and fails with:
+
+----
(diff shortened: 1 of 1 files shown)


aqhil07:
Thanks @Naros  for the updates on the commits.

Naros:
Thanks @aqhil07, applied! :tada:

PR #7411, opened 2026-05-06
debezium/dbz#1185 Sink connector framework pt.2.2 - add shared sink connector framework for JDBC sink connector

rk3rn3r:
relates to debezium/dbz#1185

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
@@ -14,10 +14,15 @@ import java.sql.SQLException;
 import java.sql.Statement;
 import java.time.Duration;
 import java.time.temporal.ChronoUnit;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Callable;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
 
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.errors.ConnectException;
@@ -34,6 +39,8 @@ import io.debezium.connector.jdbc.dialect.DatabaseDialect;
 import io.debezium.connector.jdbc.field.JdbcFieldDescriptor;
 import io.debezium.connector.jdbc.relational.TableDescriptor;
 import io.debezium.metadata.CollectionId;
+import io.debezium.sink.batch.Batch;
+import io.debezium.sink.batch.BatchRecord;
 import io.debezium.sink.field.FieldDescriptor;
 import io.debezium.sink.spi.SinkProgressListener;
 import io.debezium.sink.valuebinding.ValueBindDescriptor;
@@ -283,6 +290,76 @@ public class DefaultRecordWriter implements RecordWriter {
             }
         }
         throw new ConnectException("Exceeded max retries " + flushMaxRetries + " times, failed to " + description, lastException);
+    }
+
+    public void write(Batch records) {
+        // resolve tables outside of doWork (DDL needs session access that conflicts with active doWork connection)
+        Map<TableDescriptor, List<JdbcSinkRecord>> insertsByTable = new LinkedHashMap<>();
+        Map<TableDescriptor, List<JdbcSinkRecord>> deletesByTable = new LinkedHashMap<>();
+
+        for (BatchRecord batchRecord : records) {
+            var record = (JdbcKafkaSinkRecord) batchRecord.record();
+            var collectionId = batchRecord.collectionId();
+
+            final TableDescriptor table;
+            try {
+                table = checkAndApplyTableChangesIfNeeded(collectionId, record);
+            }
+            catch (SQLException | JDBCException e) {
+                throw new ConnectException("Error while checking and applying table changes for collection '" + collectionId + "'", e);
+            }
+
+            if (record.isTruncate()) {
+                try {
+                    writeTruncate(collectionId);
+                }
+                catch (SQLException | JDBCException e) {
+                    throw new ConnectException("Failed to truncate collection '" + collectionId + "'", e);
+                }
+            }
+            else if (record.isDelete()) {
+                deletesByTable.computeIfAbsent(table, k -> new ArrayList<>()).add(record);
+            }
+            else {
+                insertsByTable.computeIfAbsent(table, k -> new ArrayList<>()).add(record);
+            }
+        }
+
+        if (insertsByTable.isEmpty() && deletesByTable.isEmpty()) {
+            return;
+        }
+
+        // write records inside doWork/transaction with retry support
+        String tableNames = Stream.concat(
+                insertsByTable.keySet().stream(),
+                deletesByTable.keySet().stream())
+                .map(t -> t.getId().name())
+                .distinct()
+                .collect(Collectors.joining(", "));
+        executeWithRetries("flush records for table(s) '" + tableNames + "'",
+                () -> {
+                    Stopwatch writeStopwatch = Stopwatch.reusable();
+                    writeStopwatch.start();
+                    final Transaction transaction = session.beginTransaction();
+                    try {
+                        session.doWork(conn -> {
+                            for (var entry : insertsByTable.entrySet()) {
+                                performWrite(conn, getSqlStatementInfo(entry.getKey(), entry.getValue()), entry.getValue());
+                            }
+                            for (var entry : deletesByTable.entrySet()) {
+                                performWrite(conn, getSqlStatementInfo(entry.getKey(), entry.getValue()), entry.getValue());
+                            }
+                        });
+                        transaction.commit();
+                    }
+                    catch (Exception e) {
+                        transaction.rollback();
+                        throw e;
+                    }
+                    writeStopwatch.stop();
+                    LOGGER.trace("[PERF] Total write execution time {}", writeStopwatch.durations());
+                    return null;
+                });
     }
 
     @Override
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
@@ -94,6 +94,10 @@ public class DefaultRecordWriter implements RecordWriter {
         return dialect;
     }
 
+    protected SinkProgressListener progressListener() {
+        return progressListener;
+    }
+
     /**
      * Bind key field values to the query for a single record.
      */
@@ -157,7 +161,7 @@ public class DefaultRecordWriter implements RecordWriter {
             transaction.rollback();
             throw e;
         }
-        progressListener.tableCreated();
+        progressListener().tableCreated();
 
         return readTable(collectionId);
     }
@@ -212,7 +216,7 @@ public class DefaultRecordWriter implements RecordWriter {
             transaction.rollback();
             throw e;
         }
-        progressListener.tableAltered();
+        progressListener().tableAltered();
 
         return readTable(collectionId);
     }
@@ -372,6 +376,7 @@ public class DefaultRecordWriter implements RecordWriter {
         try {
             getSession().doWork(processBatch(statementInfo, records));
             transaction.commit();
+            processMetrics(records, statementInfo);
         }
         catch (Exception e) {
             transaction.rollback();
@@ -379,6 +384,19 @@ public class DefaultRecordWriter implements RecordWriter {
         }
         writeStopwatch.stop();
         LOGGER.trace("[PERF] Total write execution time {}", writeStopwatch.durations());
+    }
+
+    protected void processMetrics(List<JdbcSinkRecord> records, SqlStatementInfo statementInfo) {
+        if (statementInfo.isDelete()) {
+            progressListener().deleted(records.size());
+        }
+        else {
+            switch (getConfig().getInsertMode()) {
+                case INSERT -> progressListener().inserted(records.size());
+                case UPDATE -> progressListener().updated(records.size());
+                case UPSERT -> progressListener().upserted(records.size());
+            }
+        }
     }
 
     public void writeTruncate(CollectionId collectionId) throws SQLException {
@@ -390,6 +408,7 @@ public class DefaultRecordWriter implements RecordWriter {
 
             query.executeUpdate();
             transaction.commit();
+            progressListener().truncated();
         }
         catch (Exception e) {
             transaction.rollback();
@@ -408,7 +427,6 @@ public class DefaultRecordWriter implements RecordWriter {
             Stopwatch allbindStopwatch = Stopwatch.reusable();
             allbindStopwatch.start();
             for (JdbcSinkRecord record : records) {
-
                 Stopwatch singlebindStopwatch = Stopwatch.reusable();
                 singlebindStopwatch.start();
                 bindValues(record, queryBinder);
@@ -472,15 +490,15 @@ public class DefaultRecordWriter implements RecordWriter {
         JdbcSinkRecord firstRecord = records.get(0);
 
         if (!firstRecord.isDelete()) {
-            switch (config.getInsertMode()) {
+            switch (getConfig().getInsertMode()) {
                 case INSERT:
                     // Try batch insert first (e.g., UNNEST for PostgreSQL)
                     Optional<String> batchInsert = dialect.getBatchInsertStatement(table, records);
                     if (batchInsert.isPresent()) {
                         LOGGER.debug("Using batch INSERT statement with {} records", records.size());
-                        return new SqlStatementInfo(batchInsert.get(), true);
+                        return new SqlStatementInfo(batchInsert.get(), true, false);
                     }
-                    return new SqlStatementInfo(dialect.getInsertStatement(table, firstRecord), false);
+                    return new SqlStatementInfo(dialect.getInsertStatement(table, firstRecord), false, false);
 
                 case UPSERT:
                     if (firstRecord.keyFieldNames().isEmpty()) {
@@ -490,18 +508,18 @@ public class DefaultRecordWriter implements RecordWriter {
                     Optional<String> batchUpsert = dialect.getBatchUpsertStatement(table, records);
                     if (batchUpsert.isPresent()) {
                         LOGGER.debug("Using batch UPSERT statement with {} records", records.size());
-                        return new SqlStatementInfo(batchUpsert.get(), true);
+                        return new SqlStatementInfo(batchUpsert.get(), true, false);
                     }
-                    return new SqlStatementInfo(dialect.getUpsertStatement(table, firstRecord), false);
+                    return new SqlStatementInfo(dialect.getUpsertStatement(table, firstRecord), false, false);
 
                 case UPDATE:
                     // UPDATE doesn't have batch optimization yet
-                    return new SqlStatementInfo(dialect.getUpdateStatement(table, firstRecord), false);
+                    return new SqlStatementInfo(dialect.getUpdateStatement(table, firstRecord), false, false);
             }
         }
         else {
             // DELETE doesn't have batch optimization yet
-            return new SqlStatementInfo(dialect.getDeleteStatement(table, firstRecord), false);
+            return new SqlStatementInfo(dialect.getDeleteStatement(table, firstRecord), false, false);
         }
         throw new DataException(String.format("Unable to get SQL statement for %s", firstRecord));
     }
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
@@ -22,7 +22,6 @@ import java.util.Optional;
 import java.util.Set;
 import java.util.concurrent.Callable;
 import java.util.stream.Collectors;
-import java.util.stream.Stream;
 
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.errors.ConnectException;
@@ -296,16 +295,18 @@ public class DefaultRecordWriter implements RecordWriter {
         throw new ConnectException("Exceeded max retries " + flushMaxRetries + " times, failed to " + description, lastException);
     }
 
-    public void write(Batch records) {
-        // resolve tables outside of doWork (DDL needs session access that conflicts with active doWork connection)
-        Map<TableDescriptor, List<JdbcSinkRecord>> insertsByTable = new LinkedHashMap<>();
-        Map<TableDescriptor, List<JdbcSinkRecord>> deletesByTable = new LinkedHashMap<>();
+    protected record ResolvedTableRecords(TableDescriptor table, List<JdbcSinkRecord> inserts, List<JdbcSinkRecord> deletes) {
+        boolean isEmpty() {
+            return inserts.isEmpty() && deletes.isEmpty();
+        }
+    }
 
-        JdbcSinkConnectorConfig.InsertMode insertMode = getConfig().getInsertMode();
-        SinkProgressListener progressedListener = progressListener();
-        List<Callable<?>> processMetrics = new ArrayList<>();
+    protected Map<CollectionId, ResolvedTableRecords> resolveBatchByTable(Batch batch) {
+        Map<CollectionId, TableDescriptor> tables = new LinkedHashMap<>();
+        Map<CollectionId, List<JdbcSinkRecord>> inserts = new LinkedHashMap<>();
+        Map<CollectionId, List<JdbcSinkRecord>> deletes = new LinkedHashMap<>();
 
-        for (BatchRecord batchRecord : records) {
+        for (BatchRecord batchRecord : batch) {
             var record = (JdbcKafkaSinkRecord) batchRecord.record();
             var collectionId = batchRecord.collectionId();
 
@@ -316,6 +317,7 @@ public class DefaultRecordWriter implements RecordWriter {
             catch (SQLException | JDBCException e) {
                 throw new ConnectException("Error while checking and applying table changes for collection '" + collectionId + "'", e);
             }
+            tables.put(collectionId, table);
 
             if (record.isTruncate()) {
                 try {
@@ -325,43 +327,36 @@ public class DefaultRecordWriter implements RecordWriter {
                     throw new ConnectException("Failed to truncate collection '" + collectionId + "'", e);
                 }
             }
+            else if (record.isDelete()) {
+                deletes.computeIfAbsent(collectionId, k -> new ArrayList<>()).add(record);
+            }
             else {
-                if (record.isDelete()) {
-                    deletesByTable.computeIfAbsent(table, k -> new ArrayList<>()).add(record);
-                    processMetrics.add(() -> {
-                        progressedListener.deleted(1);
-                        return null;
-                    });
-                }
-                else {
-                    insertsByTable.computeIfAbsent(table, k -> new ArrayList<>()).add(record);
-                    switch (insertMode) {
-                        case INSERT -> processMetrics.add(() -> {
-                            progressedListener.inserted(1);
-                            return null;
-                        });
-                        case UPDATE -> processMetrics.add(() -> {
-                            progressedListener.updated(1);
-                            return null;
-                        });
-                        case UPSERT -> processMetrics.add(() -> {
-                            progressedListener.upserted(1);
-                            return null;
-                        });
-                    }
-                }
+                inserts.computeIfAbsent(collectionId, k -> new ArrayList<>()).add(record);
             }
         }
 
-        if (insertsByTable.isEmpty() && deletesByTable.isEmpty()) {
+        Map<CollectionId, ResolvedTableRecords> result = new LinkedHashMap<>();
+        for (var entry : tables.entrySet()) {
+            CollectionId id = entry.getKey();
+            result.put(id, new ResolvedTableRecords(
+                    entry.getValue(),
+                    inserts.getOrDefault(id, List.of()),
+                    deletes.getOrDefault(id, List.of())));
+        }
+        return result;
+    }
+
+    public void write(Batch records) {
+        Map<CollectionId, ResolvedTableRecords> resolved = resolveBatchByTable(records);
+
+        if (resolved.values().stream().allMatch(ResolvedTableRecords::isEmpty)) {
             return;
         }
 
-        // write records inside doWork/transaction with retry support
-        String tableNames = Stream.concat(
-                insertsByTable.keySet().stream(),
-                deletesByTable.keySet().stream())
-                .map(t -> t.getId().name())
+        SinkProgressListener progressedListener = progressListener();
+
+        String tableNames = resolved.values().stream()
+                .map(r -> r.table().getId().name())
                 .distinct()
                 .collect(Collectors.joining(", "));
         executeWithRetries("flush records for table(s) '" + tableNames + "'",
@@ -371,17 +366,22 @@ public class DefaultRecordWriter implements RecordWriter {
                     final Transaction transaction = session.beginTransaction();
                     try {
                         session.doWork(conn -> {
-                            for (var entry : insertsByTable.entrySet()) {
-                                performWrite(conn, getSqlStatementInfo(entry.getKey(), entry.getValue()), entry.getValue());
-                            }
-                            for (var entry : deletesByTable.entrySet()) {
-                                performWrite(conn, getSqlStatementInfo(entry.getKey(), entry.getValue()), entry.getValue());
+                            for (var entry : resolved.values()) {
+                                if (!entry.inserts().isEmpty()) {
+                                    performTableWrite(conn, entry.table(), entry.inserts());
+                                }
+                                if (!entry.deletes().isEmpty()) {
+                                    performTableWrite(conn, entry.table(), entry.deletes());
+                                }
                             }
                         });
                         transaction.commit();
-                        if (!processMetrics.isEmpty()) {
-                            for (Callable<?> processMetric : processMetrics) {
-                                processMetric.call();
+                        for (var entry : resolved.values()) {
+                            if (!entry.inserts().isEmpty()) {
+                                progressedListener.written(entry.inserts().size());
+                            }
+                            if (!entry.deletes().isEmpty()) {
+                                progressedListener.deleted(entry.deletes().size());
                             }
                         }
                     }
@@ -420,11 +420,7 @@ public class DefaultRecordWriter implements RecordWriter {
             progressListener().deleted(records.size());
         }
         else {
-            switch (getConfig().getInsertMode()) {
-                case INSERT -> progressListener().inserted(records.size());
-                case UPDATE -> progressListener().updated(records.size());
-                case UPSERT -> progressListener().upserted(records.size());
-            }
+            progressListener().written(records.size());
         }
     }
 
@@ -445,11 +441,20 @@ public class DefaultRecordWriter implements RecordWriter {
         }
     }
 
+    /**
+     * Write records for a single table to the connection.
+     * Resolves the SQL statement and delegates to {@link #performWrite}.
+     * Subclasses can override to change the write strategy (e.g., UNNEST).
+     */
+    protected void performTableWrite(Connection conn, TableDescriptor table, List<JdbcSinkRecord> records) throws SQLException {
+        performWrite(conn, getSqlStatementInfo(table, records), records);
+    }
+
     private Work processBatch(SqlStatementInfo statementInfo, List<JdbcSinkRecord> records) {
         return conn -> performWrite(conn, statementInfo, records);
     }
 
-    private void performWrite(Connection conn, SqlStatementInfo statementInfo, List<JdbcSinkRecord> records) throws SQLException {
+    protected void performWrite(Connection conn, SqlStatementInfo statementInfo, List<JdbcSinkRecord> records) throws SQLException {
         try (PreparedStatement prepareStatement = conn.prepareStatement(statementInfo.statement())) {
             QueryBinder queryBinder = getQueryBinderResolver().resolve(prepareStatement);
             Stopwatch allbindStopwatch = Stopwatch.reusable();
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
@@ -301,6 +301,10 @@ public class DefaultRecordWriter implements RecordWriter {
         Map<TableDescriptor, List<JdbcSinkRecord>> insertsByTable = new LinkedHashMap<>();
         Map<TableDescriptor, List<JdbcSinkRecord>> deletesByTable = new LinkedHashMap<>();
 
+        JdbcSinkConnectorConfig.InsertMode insertMode = getConfig().getInsertMode();
+        SinkProgressListener progressedListener = progressListener();
+        List<Callable<?>> processMetrics = new ArrayList<>();
+
         for (BatchRecord batchRecord : records) {
             var record = (JdbcKafkaSinkRecord) batchRecord.record();
             var collectionId = batchRecord.collectionId();
@@ -321,11 +325,31 @@ public class DefaultRecordWriter implements RecordWriter {
                     throw new ConnectException("Failed to truncate collection '" + collectionId + "'", e);
                 }
             }
-            else if (record.isDelete()) {
-                deletesByTable.computeIfAbsent(table, k -> new ArrayList<>()).add(record);
-            }
             else {
-                insertsByTable.computeIfAbsent(table, k -> new ArrayList<>()).add(record);
+                if (record.isDelete()) {
+                    deletesByTable.computeIfAbsent(table, k -> new ArrayList<>()).add(record);
+                    processMetrics.add(() -> {
+                        progressedListener.deleted(1);
+                        return null;
+                    });
+                }
+                else {
+                    insertsByTable.computeIfAbsent(table, k -> new ArrayList<>()).add(record);
+                    switch (insertMode) {
+                        case INSERT -> processMetrics.add(() -> {
+                            progressedListener.inserted(1);
+                            return null;
+                        });
+                        case UPDATE -> processMetrics.add(() -> {
+                            progressedListener.updated(1);
+                            return null;
+                        });
+                        case UPSERT -> processMetrics.add(() -> {
+                            progressedListener.upserted(1);
+                            return null;
+                        });
+                    }
+                }
             }
         }
 
@@ -355,6 +379,11 @@ public class DefaultRecordWriter implements RecordWriter {
                             }
                         });
                         transaction.commit();
+                        if (!processMetrics.isEmpty()) {
+                            for (Callable<?> processMetric : processMetrics) {
+                                processMetric.call();
+                            }
+                        }
                     }
                     catch (Exception e) {
                         transaction.rollback();
@@ -422,7 +451,6 @@ public class DefaultRecordWriter implements RecordWriter {
 
     private void performWrite(Connection conn, SqlStatementInfo statementInfo, List<JdbcSinkRecord> records) throws SQLException {
         try (PreparedStatement prepareStatement = conn.prepareStatement(statementInfo.statement())) {
-
             QueryBinder queryBinder = getQueryBinderResolver().resolve(prepareStatement);
             Stopwatch allbindStopwatch = Stopwatch.reusable();
             allbindStopwatch.start();
@@ -519,7 +547,7 @@ public class DefaultRecordWriter implements RecordWriter {
         }
         else {
             // DELETE doesn't have batch optimization yet
-            return new SqlStatementInfo(dialect.getDeleteStatement(table, firstRecord), false, false);
+            return new SqlStatementInfo(dialect.getDeleteStatement(table, firstRecord), false, true);
         }
         throw new DataException(String.format("Unable to get SQL statement for %s", firstRecord));
     }
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
@@ -57,6 +57,9 @@ public class JdbcChangeEventSink extends AbstractChangeEventSink implements Chan
     private final ConnectorContext connectorContext;
     private final SinkProgressListener progressListener;
 
+    final Map<CollectionId, Buffer> upsertBufferByTable = new LinkedHashMap<>();
+    final Map<CollectionId, Buffer> deleteBufferByTable = new LinkedHashMap<>();
+
     public JdbcChangeEventSink(JdbcSinkConnectorConfig config, StatelessSession session, DatabaseDialect dialect, RecordWriter recordWriter,
                                ConnectorContext connectorContext, SinkProgressListener progressListener) {
         super(config);
@@ -83,9 +86,6 @@ public class JdbcChangeEventSink extends AbstractChangeEventSink implements Chan
             }
             return;
         }
-
-        final Map<CollectionId, Buffer> upsertBufferByTable = new LinkedHashMap<>();
-        final Map<CollectionId, Buffer> deleteBufferByTable = new LinkedHashMap<>();
 
         for (SinkRecord kafkaSinkRecord : records) {
             JdbcSinkRecord record = new JdbcKafkaSinkRecord(kafkaSinkRecord, config);
@@ -119,7 +119,6 @@ public class JdbcChangeEventSink extends AbstractChangeEventSink implements Chan
                 try {
                     final TableDescriptor table = recordWriter.checkAndApplyTableChangesIfNeeded(collectionId, record);
                     recordWriter.writeTruncate(table.getId());
-                    progressListener.truncated();
                     continue;
                 }
                 catch (SQLException | JDBCException e) {
@@ -147,7 +146,6 @@ public class JdbcChangeEventSink extends AbstractChangeEventSink implements Chan
                 }
 
                 flushBufferRecordsWithRetries(collectionId, getRecordsToFlush(deleteBufferByTable, collectionId, record));
-                progressListener.deleted();
             }
             else {
                 final Buffer deleteBufferToFlush = deleteBufferByTable.get(collectionId);
@@ -163,20 +161,11 @@ public class JdbcChangeEventSink extends AbstractChangeEventSink implements Chan
                 }
 
                 flushBufferRecordsWithRetries(collectionId, getRecordsToFlush(upsertBufferByTable, collectionId, record));
-                recordWriteOperation();
             }
         }
 
         flushBuffers(upsertBufferByTable);
         flushBuffers(deleteBufferByTable);
-    }
-
-    private void recordWriteOperation() {
-        switch (config.getInsertMode()) {
-            case INSERT -> progressListener.inserted();
-            case UPDATE -> progressListener.updated();
-            case UPSERT -> progressListener.upserted();
-        }
     }
 
     private BufferFlushRecords getRecordsToFlush(Map<CollectionId, Buffer> bufferMap, CollectionId collectionId, JdbcSinkRecord record) {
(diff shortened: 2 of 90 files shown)


debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
```diff
@@ -280,6 +285,70 @@ public <T> T executeWithRetries(String description, Callable<T> callable) {
...
+                    final Transaction transaction = session.beginTransaction();
+                    try {
+                        session.doWork(conn -> {
+                            for (var entry : insertsByTable.entrySet()) {
+                                performWrite(conn, getSqlStatementInfo(entry.getKey(), entry.getValue()), entry.getValue());
+                            }
+                            for (var entry : deletesByTable.entrySet()) {
+                                performWrite(conn, getSqlStatementInfo(entry.getKey(), entry.getValue()), entry.getValue());
+                            }
+                        });
```
Naros, line 377: So I am concerned that the notion of inserts/deletes per table here leaks the buffer abstraction. 

This predominantly works right now because, for keyed tables, deduplication is used, which means the buffer already ensures that there is never an insert/update, or delete for the same key. And for keyless tables with no configured primary key mode, deletes aren't possible, so you only have a stream of inserts.

But as soon as you want to implement any other type of Buffer abstraction, this logic here breaks down because you prioritize inserts/updates over deletes, which may not provide the right order of operation for the event key from the topic.

Given that the buffer's job is already to deduplicate or maintain the order of events, what's to say that the `Buffer#poll` method couldn't return a `List<BufferWork>` and the goal of the `RecordWriter` in this method is simply to iterate the `BufferWork` and execute that operation? This opens the door for things like:

* Insert records A-D into TAB1
* Delete record B from TAB1
* Insert records X-Z into TAB2
* Update records Y in TAB2
* Delete record A in TAB1
* Delete record Z in TAB2

In essence, what the current logic does today, where you fill a delete/insert-update buffer and then flush it based on what you see for the same table, you get that here by treating the buffer as the source of providing the work units based on what you enqueue.

The `RecordWriter` simply becomes a contract where you supply a `List<BufferWork>` and it fetches the table descriptors, generates the query bindings, and executes the work units. How the work units should be organized is a concern of the `Buffer`.  wdyt?

rk3rn3r: @Naros I was thinking about it and trying out things this week. You are right, similar to the previous implementation a direct flush is also required when a delete arrives. In case of `SimpleBuffer` the poll must return when a delete arrives and again when a non-delete arrives. Correct?

I was not sure if I understood you suggestion entirely but I think I will give it a try (already started on it). When I understood your approach correct this is super smart and we won't need to make intermediate flushes because the original order is maintained? I don't like to move UnitOfWork aka BufferWork into that layer, but maybe it is a very smart idea.

Naros: > In case of SimpleBuffer the poll must return when a delete arrives and again when a non-delete arrives. Correct?

Exactly.

> When I understood your approach correct this is super smart and we won't need to make intermediate flushes because the original order is maintained? I don't like to move UnitOfWork aka BufferWork into that layer, but maybe it is a very smart idea.

I haven't looked closely, but maybe there is a way to split the logic so it doesn't. Let me know what you find. And feel free in the future, if I make a suggestion and it isn't clear, by all means ping me directly and we can chat. Text isn't always the best way to convey stuff.

rk3rn3r:
I am currently trying to figure out why many tests are currently failing here, while locally the same unchanged branch runs without errors on my machine.

edit: I think I found the issue.
edit2: I think someone else did too (Apicurio)

debezium-sink/src/main/java/io/debezium/sink/batch/Batch.java
```diff
@@ -0,0 +1,18 @@
...
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Collections;
+
+public class Batch extends ArrayList<BatchRecord> {
+
+    public Batch(Collection<? extends BatchRecord> collection) {
+        super(Collections.unmodifiableList(new ArrayList<>(collection)));
+    }
```
Naros, line 15: I'm not sure this yields the intended immutable result you're looking for. Given that the unmodified list is not an `ArrayList` class type, the immutable semantics are discarded, are they not?

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/sink/MongoDbSinkConnectorConfig.java
```diff
@@ -204,9 +204,18 @@ public String useTimeZone() {
         return "UTC";
     }
 
+    @Override
+    public KeyedMessageBatchMode getKeyedMessageBatchMode() {
+        return null;
```
Naros, line 209: Would it be safer here to return some default rather than `null`?

debezium-sink/src/main/java/io/debezium/sink/batch/AbstractBuffer.java
```diff
@@ -0,0 +1,80 @@
...
+    }
+
+    protected record BatchKey(CollectionId collectionId, Object key) {
+    }
+
+    @Override
+    public int size() {
+        synchronized (records) {
+            return records.size();
+        }
```
Naros: This seems unusual: synchronizing only `size` but not the others if your goal is thread safety. Should we either make all methods thread-safe or none?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -492,12 +497,12 @@ public String getCollectionNameFormat() {
 
     @Override
     public PrimaryKeyMode getPrimaryKeyMode() {
-        return primaryKeyMode;
+        return this.primaryKeyMode;
```
Naros: There seem to be some superfluous `this.` prefixes added here that are irrelevant. Ideally the PR should not be making style changes like this to keep the delta minimal.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -410,11 +411,13 @@ public String getValue() {
     private final long flushRetryDelayMs;
     private final int batchSize;
     private final boolean useReductionBuffer;
+    private final KeyedMessageBatchMode keyedMessageBatchMode;
     private final boolean connectionRestartOnErrors;
     private final String cloudEventsSchemaNamePattern;
     private final PrimaryKeyMode primaryKeyMode;
     private final Set<String> primaryKeyFields;
     private final FieldNameFilter fieldsFilter;
+    private final boolean isSharedChangeEventSinkEnabled;
```
Naros, line 421:
```suggestion
    private final boolean sharedChangeEventSinkEnabled;
```
Generally, I expect to see the `is` or `has` prefixes on accessors, not member variables.

debezium-testing/debezium-testing-testcontainers/src/test/java/io/debezium/testing/testcontainers/ApicurioRegistryTestIT.java
```diff
@@ -234,13 +234,15 @@ public void shouldNotErrorWithBadHeader() {
...
                 .with("value.converter.avro.apicurio.registry.artifact.group-id", "dummy")
-                .with("value.converter.avro.apicurio.registry.request.ssl.truststore.location", "TroubleMaker");
+                .with("value.converter.avro.apicurio.registry.request.ssl.truststore.location", "TroubleMaker")
+                .with("value.converter.avro.apicurio.kafkasql.security.ssl.truststore.location", "TroubleMaker");
 
-        debeziumContainer.registerConnector("my-connector-test-apicurio-header", config);
-        debeziumContainer.ensureConnectorTaskState("my-connector-test-apicurio-header", 0, Connector.State.FAILED);
+        String connectorName = "my-connector-test-apicurio-header";
+        debeziumContainer.registerConnector(connectorName, config);
+        debeziumContainer.ensureConnectorTaskState(connectorName, 0, Connector.State.RUNNING);
```
Naros: Are all these changes needed for things you're changing in this PR or are these changes needed for the recent failures with this test class? If it's the latter, these should be applied in a separate PR and not here as they're unrelated to the JDBC sink work.

debezium-sink/src/test/java/io/debezium/sink/AbstractChangeEventSinkTest.java
```diff
@@ -0,0 +1,357 @@
...
+            DebeziumSinkRecord record = invocation.getArgument(0);
+            return record.topicName().replace(".", "_");
+        });
+        when(config.getCollectionNamingStrategy()).thenReturn(namingStrategy);
+
+        sink = new TestChangeEventSink(config);
+    }
+
+    @Test
+    void putShouldBufferRecords() {
```
Naros, line 39: Could we add the `FixFor` annotations to all the newly added tests.

debezium-sink/src/main/java/io/debezium/util/BufferExtractor.java
```diff
@@ -0,0 +1,31 @@
...
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+public class LinkedHashMapExtractor {
+
+    public static <K, V> List<V> extractFirstEntries(LinkedHashMap<K, V> map, int count) {
```
Naros, line 19: Would you consider using `drainFirstEntries`? 

The term extract implies that you are slicing and taking a subset of the original map, but in fact this class mutates the provided map, so using `drain` or `remove` might be clearer. wdyt?

debezium-sink/src/main/java/io/debezium/sink/SinkConnectorConfig.java
```diff
@@ -186,8 +233,20 @@ public String getValue() {
...
+            .withType(ConfigDef.Type.BOOLEAN)
+            .withGroup(Field.createGroupEntry(Field.Group.ADVANCED, 0))
+            .withWidth(ConfigDef.Width.SHORT)
+            .withImportance(ConfigDef.Importance.MEDIUM)
+            .withDefault(false)
+            .withDescription("Enables usage of the new shared sink connector framework logic.");
+
     String getCollectionNameFormat();
 
+    KeyedMessageBatchMode getKeyedMessageBatchMode();
```
Naros, line 256: Let's make sure that interface contracts are documented.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
Naros: So one solution to address this in the current code with the least refactor would be to use writer groups that preserve the right order regardless of strategy, e.g.
```
  public void write(Batch records) {
      record WriteGroup(TableDescriptor table, List<JdbcSinkRecord> records) {}
      List<WriteGroup> writeGroups = new ArrayList<>();

      for (BatchRecord batchRecord : records) {
          var record = (JdbcKafkaSinkRecord) batchRecord.record();
          var collectionId = batchRecord.collectionId();

          final TableDescriptor table;
          try {
              table = checkAndApplyTableChangesIfNeeded(collectionId, record);
          }
          catch (SQLException | JDBCException e) {
              throw new ConnectException("Error while checking and applying table changes for collection '" + collectionId + "'", e);
          }

          if (record.isTruncate()) {
              try { writeTruncate(collectionId); }
              catch (SQLException | JDBCException e) {
                  throw new ConnectException("Failed to truncate collection '" + collectionId + "'", e);
              }
              continue;
          }
  
          // Coalesce with the last group if same table + same operation type
          if (!writeGroups.isEmpty()) {
              WriteGroup last = writeGroups.get(writeGroups.size() - 1);
              boolean lastIsDelete = last.records.get(0).isDelete();
              if (last.table.equals(table) && lastIsDelete == record.isDelete()) {
                  last.records.add(record);
                  continue;
              }
          }
          writeGroups.add(new WriteGroup(table, new ArrayList<>(List.of(record))));
      }

      if (writeGroups.isEmpty()) {
          return;
      }

      executeWithRetries("flush records", () -> {
          Stopwatch writeStopwatch = Stopwatch.reusable();
          writeStopwatch.start();
          final Transaction transaction = session.beginTransaction();
          try {
              session.doWork(conn -> {
                  for (WriteGroup group : writeGroups) {
                      performWrite(conn, getSqlStatementInfo(group.table, group.records), group.records);
                  }
              });
              transaction.commit();
          }
          catch (Exception e) {
              transaction.rollback();
              throw e;
          }
          writeStopwatch.stop();
          LOGGER.trace("[PERF] Total write execution time {}", writeStopwatch.durations());
          return null;
      });
  }
```
This unblocks the correctness and should at least give you time to investigate how to move this logic into the proper `Buffer` class, as I discussed above.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
rk3rn3r: Is it okay to say I feel this is very personal opinion? :grin: 
E.g. `useReductionBuffer` above is okay?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
rk3rn3r: Agreed. I think that was my little coding buddy. Dunno why.

debezium-sink/src/main/java/io/debezium/sink/batch/Batch.java
rk3rn3r: You are absolutely right. Thanks for the pointer. The copied list will be returned which is then mutable again, I suppose. I think just using the copy is guarding enough, will remove the `unmodifiableList()` call. Or wdyt?

edit: Code change DONE

debezium-connector-mongodb/src/main/java/io/debezium/connector/mongodb/sink/MongoDbSinkConnectorConfig.java
rk3rn3r: This is just dummy and will lead to the NOP mode, it is not used at all. It is just required because of the introduction of changes in the common shared sink connector framework interface. It will be replaced with the next PR introducing the sink framework to the MongoDB connector.

debezium-sink/src/main/java/io/debezium/sink/batch/AbstractBuffer.java
rk3rn3r: Hmmm. This code is more than 6 months old. I don't even recall why I added it back then. :grinning:  I think the buffers are initialized on a per connector task level and should then not be accessed concurrently at all. Do you agree?

edit: synchronizing removed, code change DONE

debezium-testing/debezium-testing-testcontainers/src/test/java/io/debezium/testing/testcontainers/ApicurioRegistryTestIT.java
rk3rn3r: This test was failing on CI. This is why I touched the code. It is required to make tests pass. How should I proceed?

debezium-sink/src/test/java/io/debezium/sink/AbstractChangeEventSinkTest.java
rk3rn3r: Also for the initial tests? I mean git blame game shows the related issue already. Should this be added as a checkstyle rule?

debezium-sink/src/main/java/io/debezium/util/BufferExtractor.java
rk3rn3r: No, it extracts (`pop`s ) the first `count` entries from the list and returns them. Do you think drain is a proper name then?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
rk3rn3r: @Naros Are you sure this is really needed? The new `KeyedPassthrough` splits `Batch`-es before every duplicate. Based on this assumption it is irrelevant if deletes or upserts are done first. There is always the potential that something fails when there are CASCADE-ing constraints in the downstream datastore, right? Or is there anything that WriterGroups solve in this regard or am I overseeing anything else?

rk3rn3r: Furthermore on those `CASCADE` constraints downstream: Did you ever had a thought on this one? I assume always when we do batching and not force-flush on every event type change from upsert to delete or vice-versa we are going to have issues, right? Any idea how this can be tackled?

debezium-sink/src/main/java/io/debezium/sink/SinkConnectorConfig.java
rk3rn3r: DONE

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
Naros: Hah, that one, I would prefer to see `reductionBufferEnabled`, but :man_shrugging: :grin:

debezium-sink/src/test/java/io/debezium/sink/AbstractChangeEventSinkTest.java
Naros: The checkstyle rule idea sounds like a nice addition tbh, this does come up from time to time. Would love to know what others on the team think about it.

debezium-sink/src/main/java/io/debezium/util/BufferExtractor.java
Naros: As you and I discussed the other day, a lot of this is open to interpretation, I guess :grin:. 

I just think using a term that is more emblematic in Java makes more sense, e.g. `remove` or `drain`. If we want to get into the weeds, even the use of Entries seems unusual because I would otherwise expect a `List<Map.Entry<K,V>>` return, but if you want to keep it as-is, up to you.

debezium-sink/src/main/java/io/debezium/sink/batch/KeyedPassthroughBuffer.java
```diff
@@ -0,0 +1,68 @@
...
+    public void truncate(CollectionId collectionId, DebeziumSinkRecord record) {
+        completedBatches.replaceAll(batch -> {
+            List<BatchRecord> filtered = batch.stream()
+                    .filter(br -> !br.collectionId().equals(collectionId))
+                    .toList();
+            return new Batch(filtered);
+        });
+        completedBatches.removeIf(Batch::isEmpty);
+        super.truncate(collectionId, record);
+    }
```
Naros: For this use case where we want to preserve all events for audit purposes with no deduplication, then shouldn't we not flush the in-memory buffers?

The way I read this is if table X has 10 events in the buffer and a truncate arrives, then we discard the 10 events and only truncate what is present in the target table. I would expect here we'd simply append the truncate to the buffer here and not clear any of the existing batches, right?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
Naros: I went back and reviewed `KeyedPassthrough` again, and I think it solves the issue. I did notice one issue with truncates and left a comment on it; could you take a look?

With regard to cascading, this is an inherent problem in any batch scenario. I think in this case, rather than trying to solve that now, it should just be documented, with the expectation that the batch size be set to 1 to force a flush on each operation. I'm not aware of this coming up, so I would say this is quite rare, and the existing solution solves 90-95% of use cases.

So then I would say that the insert/delete bucketing here is moot and would serve no purpose, if we agree with the cascading requiring batch size of 1. wdyt?

debezium-sink/src/main/java/io/debezium/sink/batch/Buffer.java
```diff
@@ -0,0 +1,44 @@
...
+
+    /**
+     * Add a {@link DebeziumSinkRecord} to the internal buffer.
+     * @param record the Sink record descriptor
+     */
+    void enqueue(CollectionId collectionId, DebeziumSinkRecord record);
+
+    int size();
+
+    void truncate(CollectionId collectionId, DebeziumSinkRecord record);
```
Naros, line 41: Javadocs

debezium-sink/src/main/java/io/debezium/sink/batch/Batch.java
Naros: @rk3rn3r if you wanted to maintain immutability for a `Batch` but give it `List` collection semantics with your unique class name, could you not simply extend `AbstractList<BatchRecord>` and used composition?
```java
public class Batch extends AbstractList<BatchRecord> {
  private final List<BatchRecord> records;

  public Batch(Collection<? extends BatchRecord> collection) {
    this.records = List.copyOf(collection); // creates an immutable wrapper
  }
  /* implement get/size to access records */
  /* if you don't implement, it throws UnsupportedOperationException by default, which you want for set */
  /* things like iterator, stream, contains are all for free */
}
```
WDYT?

debezium-sink/src/test/java/io/debezium/sink/AbstractChangeEventSinkTest.java
mfvitale: > The checkstyle rule idea sounds like a nice addition tbh, this does come up from time to time. Would love to know what others on the team think about it.


Dunno. How to differentiate when a test is part of a new feature and not a fix then?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
rk3rn3r: @Naros Nice! You are right, thx. The outcome of this discussion is awesome. I reflected so many things and learned from that.

> I did notice one issue with truncates and left a comment on it; could you take a look?

For sure!

> rather than trying to solve that now, it should just be documented, with the expectation that the batch size be set to 1 to force a flush on each operation.

Sure! I agree!

> So then I would say that the insert/delete bucketing here is moot and would serve no purpose

I think it is still required because of we are sending batches of upserts or batches of deletes downstream and I can only have a single (prepared) statement per table and operation type? But you are the Hibernate expert, am I wrong? Can I send batches of mixed upserts/deletes?

debezium-sink/src/main/java/io/debezium/sink/batch/Batch.java
rk3rn3r: I think I discussed it once with my AI (not Vojtech's heheheeee) and we agreed that there is lots of access over the ArrayList semantics and it is not worth the effort. But we considered this option.

debezium-sink/src/test/java/io/debezium/sink/AbstractChangeEventSinkTest.java
rk3rn3r: @mfvitale I think Chris wants to have it always, also for the initial tests?

debezium-sink/src/main/java/io/debezium/util/BufferExtractor.java
rk3rn3r: Every computer science student learns this as `pop` operation (if we look for technically highly correct naming _giggles_), but I don't like to call it like that. :popcorn:

I agree the method is biased, not really a generic reusable `LinkedHashMapExtractor`. Maybe the classname should be `BufferExtractor` which makes it clear why the key `<K>` is kept internal. wdyt?

debezium-sink/src/main/java/io/debezium/sink/batch/KeyedPassthroughBuffer.java
rk3rn3r: I think you are right. Let me double-check. I am not sure what exactly should happen here. either complex:

1) flush all records in completedBatches **and** in the buffer
2) call `enqueue(collectionId, record);` to enqueue the truncate event (as first event for this collection id)

That way all pending records get flushed immediately

or simply:

1) Simply enqueue the record with `enqueue(collectionId, record)`

plus use a different RecordWriter implementation that flushes pending `deletesByTable` and `instertsByTable` before executing the `writeTruncate()`.

Wdyt @Naros ?

debezium-sink/src/main/java/io/debezium/sink/batch/Buffer.java
rk3rn3r: I am going over all the interfaces rn and adding missing docs!

debezium-sink/src/main/java/io/debezium/util/BufferExtractor.java
Naros: Up to your discretion :wink:

rk3rn3r: It feels like you lean more towards the more generic framework like version and discard the key from the `Entry`?

debezium-sink/src/main/java/io/debezium/sink/batch/KeyedPassthroughBuffer.java
Naros: I don't think you need anything special for this use case; most of your logic is already there.

In `RecordWriter`, you already handle transactional boundaries and flush semantics per batch that gets returned from the buffers, so we can use that to our advantage.

In `DeduplicatingBuffer`, you could reuse the same `completedBatches` concept you use for the passthrough. When you observe a truncate, the truncate method would:

1. Remove pending entries in `records` for the truncated table.
2. Move all remaining `records` entries into a `completedBatches` batch.
3. Enqueue the truncate as its own `completedBatch` batch.
4. Continue on with enqueuing events into the now empty `records`.

In the poll, you would return the events in `completedBatches` order, plus whatever remains in `records`, just like you already do for the `PassthroughBuffer`. 

For `PassthroughBuffer`, it would be handled exactly the same way except omitting (1). 

So during the `RecordWriter` loop, you'd flush everything pre-truncate, execute the truncate separately, and then execute whatever comes after the truncate, keeping that clean separation.

WDYT, would that work?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
Naros: Yes, those buckets are still required for using JDBC batching with a stateless session; I simply got ahead of myself. If this were a stateful session, that'd be entirely different :grin: So, ya those buckets are needed.

debezium-sink/src/main/java/io/debezium/util/BufferExtractor.java
Naros: I do, but it's just a matter of taste.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
rk3rn3r: Then this code section is okay. Marked as DONE.
Keeping this discussion open until a docs section on the `CASCADE` constraints was added.

debezium-sink/src/main/java/io/debezium/util/BufferExtractor.java
rk3rn3r: Will rename it to `BufferExtractor` in order to make clear it is not a generic framework-like operation that users would expect from the generic classname.

debezium-sink/src/main/java/io/debezium/sink/batch/KeyedPassthroughBuffer.java
rk3rn3r: I figured that I don't need 4 implementations. I changed the code to `PassthroughBuffer` and `DeduplicatingBuffer` and both have a parameter for keyed operation mode. This should work for all use-cases now, except the CASCADE effects that we discussed above.

Truncate events are always the first event in a Batch. That way everything is flushed before the truncate event.

debezium-sink/src/test/java/io/debezium/sink/AbstractChangeEventSinkTest.java
rk3rn3r: DONE in all _new_ tests and ITs.
Moving the discussion to the team call.

rk3rn3r:
@Naros @jpechane I think the PR is ready for the final review. I am on PTO today and will check back tomorrow.

rk3rn3r:
@Naros re-integrated changes from last week.
Can I have a final review and if nothing new comes up do you think we can merge?

rk3rn3r:
@Naros I finally finished re-integration and tried to get rid of all the CI test failures (except Informix connector, DBZ platform, hopefully) in seperate new commits. I hope tests will soon be green.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
```diff
@@ -285,6 +296,105 @@ public <T> T executeWithRetries(String description, Callable<T> callable) {
         throw new ConnectException("Exceeded max retries " + flushMaxRetries + " times, failed to " + description, lastException);
     }
 
+    public void write(Batch records) {
```
Naros: If someone enables the `enable.sces` toggle and then uses Unnest, the connector will fail because of the binding strategy being used.

So either the Unnest approach should override `write(Batch)`, even in the short term if it's going to eventually be folded, or we need to account for this somehow.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
```diff
@@ -55,8 +57,12 @@ public class JdbcChangeEventSink implements ChangeEventSink {
     private final ConnectorContext connectorContext;
     private final SinkProgressListener progressListener;
 
+    final Map<CollectionId, Buffer> upsertBufferByTable = new LinkedHashMap<>();
+    final Map<CollectionId, Buffer> deleteBufferByTable = new LinkedHashMap<>();
```
Naros, line 61: In the previous code, these buffers were intentionally re-created on each call to `execute` because if any code path were to ever throw a `RetriableException`, the Kafka Connect `WorkerSinkTask` does not destroy the `SinkTask`, but instead pauses and then recalls `SinkTask#put` again with the same batch. Because these are now instance-based, you end up with duplicates.

I'm okay if we want to minimize the allocation by keeping these instance-based, but we should wrap the code inside execute with a `try/finally` that clears these on each `execute` call.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
Naros: This hasn't been fixed @rk3rn3r, could you take a look.

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/mysql/JdbcSinkRetryIT.java
```diff
@@ -107,12 +105,9 @@ public void testRetryToFlushBufferWhenRetriableExceptionOccurred(SinkRecordFacto
...
         catch (Exception e) {
-            assertThat(e.getCause().getMessage()).matches(
-                    "Exceeded max retries [0-9]* times, failed to flush records for table '" + tableName + "'");
-            // PessimisticLockException exception is retriable in mysql dialect.
-            assertThat(e.getCause().getCause()).isInstanceOf(PessimisticLockException.class);
-            assertThat(e.getCause().getCause().getCause().getMessage()).matches(
-                    "Lock wait timeout exceeded; try restarting transaction");
+            assertExceptionCauseMessage(e,
+                    "Exceeded max retries [0-9]* times, failed to flush records for table.*'" + tableName + "'");
+            assertExceptionCauseMessage(e, "Lock wait timeout exceeded; try restarting transaction");
```
Naros: I like the cleanup, but this drops checking that it's a `PessimisticLockException`. Intended?

debezium-sink/src/main/java/io/debezium/sink/batch/AbstractBuffer.java
```diff
@@ -0,0 +1,66 @@
...
+        if (batchSize <= records.size()) {
+            return Collections.singletonList(new Batch(BufferExtractor.extractFirstEntries(records, batchSize)));
+        }
+        return List.of();
+    }
+
+    @Override
+    public List<Batch> forcePoll() {
+        if (!records.isEmpty()) {
+            return Collections.singletonList(new Batch(BufferExtractor.extractFirstEntries(records, records.size())));
```
Naros, line 61: This intentionally only returns 1 batch per poll. Would it be more efficient to have this generate multiple batches until all of the `records` have been exhausted now that we return `List<Batch>`?

debezium-sink/src/main/java/io/debezium/sink/batch/Buffer.java
Naros: This is still an open item.

debezium-sink/src/main/java/io/debezium/sink/spi/ChangeEventSink.java
```diff
@@ -25,8 +27,9 @@ public interface ChangeEventSink extends AutoCloseable {
      *
      * @param records the sink records, never {@code null}
      */
-    void execute(Collection<SinkRecord> records);
+    List<Batch> put(Collection<SinkRecord> records);
 
     CollectionId getCollectionId(String collectionName);
 
+    List<Batch> forcePoll();
```
Naros, line 45: There are also methods added here with no javadocs.

debezium-sink/src/main/java/io/debezium/util/BufferExtractor.java
```diff
@@ -0,0 +1,34 @@
...
+package io.debezium.util;
+
+import java.util.ArrayList;
+import java.util.Iterator;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+/**
+ * Extractor helper that fetches the first entries from the @see LinkedHashMap inside a @see io.debezium.sink.batch.Buffer .x
```
Naros:
```suggestion
 * Extractor helper that fetches the first entries from the @see LinkedHashMap inside a @see io.debezium.sink.batch.Buffer
```

debezium-sink/src/main/java/io/debezium/sink/AbstractChangeEventSink.java
```diff
@@ -0,0 +1,134 @@
...
+    public CollectionId getCollectionIdFromRecord(DebeziumSinkRecord record) {
+        String tableName = this.config.getCollectionNamingStrategy().resolveCollectionName(record, config.getCollectionNameFormat());
+        if (tableName == null) {
+            return null;
+        }
+        return getCollectionId(tableName);
+    }
+
+    @Override
+    public List<Batch> put(Collection<SinkRecord> records) {
```
Naros, line 65: We aren't accumulating the filtered metric here that was added when SCES is enabled. We should consider adding this here to maintain parity.

Naros (commented):
Hi @rk3rn3r this is getting close, I've left a few inline comments.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
rk3rn3r: Are you sure? I am convinced I moved it from there to local some time ago and it was only because it was not used outside. I agree this was a side-effect. :thinking:  :grin: 

Are you sure in the previous version on normal shutdown the remaining messages (that were already `put` by KC) are properly flushed? I think KC commits those offsets on shutdown?
I aligned this behavior now with the new code to ensure when offsets get flushed buffers are flushed as well.

Then wrapping with `try/finally` is probably the best way.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
rk3rn3r: Currently it will use the non-unnest path, no? But I agree this should be added. Sorry, it was added in between and I missed that case.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
rk3rn3r: It was. Maybe auto re-integration brought it back. Maybe a `commit --fixup`. 🤦

rk3rn3r: DONE. (again)

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/mysql/JdbcSinkRetryIT.java
rk3rn3r: I think it was a fix for a CI issue and I thought there is no other exception. But I'll bring it back.

I see for old behavior:
```
org.apache.kafka.connect.errors.ConnectException: JDBC sink connector failure
org.apache.kafka.connect.errors.ConnectException: Exceeded max retries 1 times, failed to flush records for table '...'
org.hibernate.exception.LockTimeoutException: Error executing work [Lock wait timeout exceeded; try restarting transaction] [n/a]
java.sql.BatchUpdateException: Lock wait timeout exceeded; try restarting transaction
com.mysql.cj.jdbc.exceptions.MySQLTransactionRollbackException: Lock wait timeout exceeded; try restarting transaction
```

and sink framework:
```
org.apache.kafka.connect.errors.ConnectException: Exceeded max retries 1 times, failed to flush records for table(s) '...'
org.hibernate.exception.LockTimeoutException: Error executing work [Lock wait timeout exceeded; try restarting transaction] [n/a]
java.sql.BatchUpdateException: Lock wait timeout exceeded; try restarting transaction
com.mysql.cj.jdbc.exceptions.MySQLTransactionRollbackException: Lock wait timeout exceeded; try restarting transaction
```

Which one should we use? And because the trace is different we need to change the logic a bit @Naros

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/TestInfrastructureHelper.java
```diff
@@ -59,16 +58,19 @@ public enum DATABASE {
...
-            DockerImageName.parse("quay.io/debezium/kafka:" + DEBEZIUM_CONTAINER_IMAGE_VERSION_LATEST).asCompatibleSubstituteFor("kafka"))
-            .withImagePullPolicy(PullPolicy.ageBased(Duration.ofHours(8)))
-            .withNetworkAliases(KAFKA_HOSTNAME)
-            .withNetwork(NETWORK)
-            .withEnv("HOST_NAME", KAFKA_HOSTNAME)
-            .withEnv("CLUSTER_ID", "5Yr1SIgYQz-b-dgRabWx4g")
-            .withEnv("NODE_ID", "1")
-            .withEnv("NODE_ROLE", "combined")
-            .withEnv("KAFKA_CONTROLLER_QUORUM_BOOTSTRAP_SERVERS", KAFKA_HOSTNAME + ":9093");
+    public static StrimziKafkaCluster createKafkaCluster() {
```
AlvarVG, line 61: Hey @rk3rn3r I am totally agree with this changes so we avoid extra complexity we have on our side. 

But i have the feeling that if we merge this PR we are going to loose backwards compatibility for older versions either kafka/debezium. What do you think?

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/mysql/JdbcSinkRetryIT.java
rk3rn3r: I have added a helper method to match Exception type and message.

```
            assertExceptionCausedBy(e, ConnectException.class,
                    "Exceeded max retries [0-9]* times, failed to flush records for table.*'" + tableName + "'");
            assertExceptionCausedBy(e, BatchUpdateException.class, "Lock wait timeout exceeded; try restarting transaction");
```

debezium-sink/src/main/java/io/debezium/sink/batch/AbstractBuffer.java
rk3rn3r: The buffer should always have `batch.size - 1` records max. So it'll always return a single batch.
But what comes to my mind: can a connector-close that calls forceFlush() happen while a parallel normal processing (with flush() at the end) is in action? Can then the connector-close force-flush happen before the normal flush?

debezium-sink/src/main/java/io/debezium/sink/AbstractChangeEventSink.java
rk3rn3r: The metrics were quite misleading putting them before the buffers, probable transformation and deduplication. All metrics were moved to the RecordWriter level in the legacy and the new framework code.

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/testhelper/TestInfrastructureHelper.java
rk3rn3r: @AlvarVG Yes and no. Thx for raising it. The main repo already switched to StrimziKafkaCluster. It was just not easy to do correctly for the testcontainers repo and this was just a leftover. Check out #6756 and #6868

Furthermore see the commit messages. There were mixed implementations now (StrimziKafkaCluster in a source connector interfering with the sink connector for example that was using the old Kafka container/s - which here are finally updated)

debezium-sink/src/main/java/io/debezium/util/BufferExtractor.java
rk3rn3r: ooopsy 😁

rk3rn3r: Applied.

rk3rn3r:
Applied and pushed fixes for the smaller issues.

Still working on `write(Batch)` for UnnestRecordWriter 

We need to clarify if "add `try/finally` to the legacy put()/execute() behavior to reset buffers" is really needed

(Updated. FYI @Naros)

Naros:
Sounds good @rk3rn3r, I'll wait until those 2 are in and do a follow-up review, if that's okay.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
rk3rn3r: I looked into this and I think we are fine, no?
Btw it looks on the first glance as if we even loose events when the buffer is not on the instance level? Could this be?
We catch the Exception in `io.debezium.connector.jdbc.JdbcSinkConnectorTask#put()` and mark the records as not processed. Which is fine. So the next run will flush the events and mark the records as processed in the offsets (or mark not processed and retry). When the buffer gets reset (like now on every execute call) we mark the events as not processed but we can't/don't retry flushing them on the next run?

This part is rather complex, so maybe I am wrong on this? @Naros

Naros: Today, when a batch of events (A, B, and C) arrives, and one of them fails to write, the exception is thrown; all events (A, B, and C) are marked as unprocessed, and the consumer offsets remain untouched.

If the sink throws a non-RetriableException, the sink is destroyed. This is where the instance-scoped collections are a non-issue because the `JdbcChangeEventSink` instance is re-created.

If the sink throws a RetriableException, the `JdbcChangeEventSink` is not destroyed; instead, the worker briefly pauses and resends the same batch of events (A, B, and C). So there's no concern here about events not being flushed; this is already handled correctly before this PR.

Where this becomes a problem is when a RetriableException is thrown with these changes: when events A, B, and C are sent again, they are already enqueued in the collection buffers, so we reprocess them and create duplicates, which can break the connector's idempotent replay.

All this needs inside `JdbcChangeEventSink#execute` is:
```java
    public void execute(Collection<SinkRecord> records) {
        if (config.isSharedChangeEventSinkEnabled()) {
            for (Batch batch : put(records)) {
                writeBatch(batch);
            }
            return;
        }
        try {
            for (SinkRecord kafkaSinkRecord : records) {
				// existing logic
            }
        }
        finally {
            upsertBufferByTable.clear();
			deleteBufferByTable.clear();
        }
    }
```
This is effectively the same behavior the connector uses now, except that we're clearing the buffers at the end of the method rather than at the top.

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/AbstractBaseJdbcSinkTest.java
```diff
@@ -74,6 +74,17 @@ protected void assertExceptionCauseMessage(Throwable t, String messagePattern) {
         assertThat(cause).as("No exception in cause chain matches pattern: " + messagePattern).isNotNull();
     }
 
+    protected void assertExceptionCausedBy(Throwable t, Class<? extends Throwable> causedBy, String messagePattern) {
```
Naros, line 77: Could we opt to move to using the standard AssertJ methods here?
```java
assertThat(e).cause().isInstanceOf(TheClass.class).hasMessageMatching(thePattern);
```
IIRC, it does exactly what you're trying to accomplish here, and I would argue even the sibling method above could be replaced with a consolidated one too.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
rk3rn3r: > If the sink throws a non-RetriableException, the sink is destroyed. This is where the instance-scoped collections are a non-issue because the `JdbcChangeEventSink` instance is re-created.

According to https://github.com/debezium/debezium/blob/main/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorTask.java#L157-L186 every exception is catched `catch (Throwable throwable) {` and converted to a `ConnectException` which is never retriable? What do I overlook?

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/AbstractBaseJdbcSinkTest.java
rk3rn3r: The different implementations generate a different stacktraces. That's why `.cause()` is not the right exception and we need to iterate and find the exception in the stacktrace that matches message and type.
There is no AssertJ method for this iteration, is it?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
Naros: Of course, but what happens when someone changes that, our test coverage doesn't cover this case we've identified, and we cut a release, and people get dupes? 

The code, as written before, was designed to guard against that. The current design in this PR breaks that guard. I'd rather be defensive and avoid introducing behavior changes in the new sink framework on the legacy path without a strong justification. 

As I said, minimizing the collection allocation, if that was a concern, and making it instance-scoped is fine, but we should keep the collection clearing behavior as it was before for the legacy code paths.

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/AbstractBaseJdbcSinkTest.java
Naros: You're right, my apologies. I had thought there was one, but apparently not :sob: 
You'd think it would make sense for that LOL.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
rk3rn3r: > Of course, but what happens when someone changes that, our test coverage doesn't cover this case we've identified, and we cut a release, and people get dupes?

Our new tests do. For example the new SinkProgressListener tests count the number of processed events!

> The code, as written before, was designed to guard against that. The current design in this PR breaks that guard.

That's not precisely enough? The code behaves exactly the same (as long as all exceptions are mapped into a ConnectException - what is currently the case) and there are tests that ensure the event count is correct (SinkProgressListener).

What happens in KC is that the connector is stopped / failed ( task::stop() is called for resource cleanup) and if it gets restarted/recreated it will start from the uncommitted offsets and with a new JdbcSinkConnectorTask and a new JdbcChangeEvent instance.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/metrics/JdbcSinkConnectorMetrics.java
```diff
@@ -35,15 +34,12 @@ public class JdbcSinkConnectorMetrics implements JdbcSinkConnectorMetricsMXBean,
...
-    private final AtomicLong totalNumberOfInsertEventsSeen = new AtomicLong();
-    private final AtomicLong totalNumberOfUpdateEventsSeen = new AtomicLong();
-    private final AtomicLong totalNumberOfUpsertEventsSeen = new AtomicLong();
-    private final AtomicLong totalNumberOfDeleteEventsSeen = new AtomicLong();
-    private final AtomicLong totalNumberOfTruncateEventsSeen = new AtomicLong();
-    private final AtomicLong numberOfFilteredEvents = new AtomicLong();
+    private final AtomicLong totalNumberOfWrites = new AtomicLong();
+    private final AtomicLong totalNumberOfDeletes = new AtomicLong();
+    private final AtomicLong totalNumberOfTruncates = new AtomicLong();
+    private final AtomicLong totalNumberOfFilteredEvents = new AtomicLong();
```
rk3rn3r, line 40: @minleejae we like to cleanup the sink metrics a bit. We like to combine insert/upsert/update into `TotalNumberOfWrites`.  A connector config has a single `insert.mode`  which maps to one of those types. In conclusion 2 of those 3 values will be 0.
Furthermore the info for "connected" was identical to the sink task run status reported by the Kafka Connect REST API.

Do you have important use-cases for keeping those metrics?

Naros:
@rk3rn3r we have about ~6 PRs that are blocked by this work. Any ETA from your side?

rk3rn3r:
@Naros Do you have a list, maybe some are not that conflicting and I can merge those?
Edit: never mind, I just saw you linked them.

Probably when CI is finished we might be ready to merge. Did you take a look on the added commit with the UnnestRecordWriter support for `write(Batch)` and the metrics?

Naros (commented):
@rk3rn3r the removal of `Connected` from the JMX metrics breaks code in Debezium Server `main`, FWIW. Are you planning to address that concurrently so we don't end up with breaking builds?

Naros:
> Did you take a look on the added commit with the UnnestRecordWriter support for `write(Batch)` and the metrics?

For the unnest record writer, I did, and I think the solution is good. I left one comment above about the metrics.

rk3rn3r:
> For the unnest record writer, I did, and I think the solution is good. I left one comment above about the metrics.

@Naros Great! I fixed that the PR workflow is using the PR's branch name in the DS repo where the fix is provided, else falls back to the base ref / PR target branch name.

Hopefully the CI build will be green now. :crossed_fingers:

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/metrics/JdbcSinkConnectorMetrics.java
minleejae: Hi @rk3rn3r 

Thanks for checking. I don’t have an important use case for keeping the three write metrics separate. Since insert.mode is fixed for a task, `TotalNumberOfWrites` seems sufficient. I also agree with removing Connected, as its current implementation reflects the task lifecycle rather than actual JDBC connectivity. If we need database connection health later, it would be better to expose a separate metric with that specific semantics.

rk3rn3r:
@Naros Tests are green. The failing Mongo DB OpenLineage tests seem to be unrelated (I still re-triggered because those seem to be flaky before and sometime pass CI).

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/metrics/JdbcSinkConnectorMetrics.java
rk3rn3r: Thx a lot for coming back so quickly @minleejae ! Highly appreciated. :pray: 
FYI @Naros

Naros (approved):

Naros:
Thanks @rk3rn3r, applied! :tada:

PR #7415, opened 2026-05-08
DBZ-1912 Allow customizing the zero-date fallback value (per type) for MySQL/MariaDB

minleejae:
Closes [debezium/dbz#1912](https://github.com/debezium/dbz/issues/1912).

## Summary

Adds four optional connector properties to the MySQL and MariaDB connectors that let users replace the hardcoded `1970-01-01` zero-date fallback with a sentinel of their choice, with per-type overrides:

| Property | Default | Purpose |
|---|---|---|
| `zero.date.fallback.value` | `1970-01-01` | General default |
| `zero.date.fallback.value.date` | (unset → inherit) | DATE override |
| `zero.date.fallback.value.datetime` | (unset → inherit) | DATETIME override |
| `zero.date.fallback.value.timestamp` | (unset → inherit) | TIMESTAMP override |

Cascading lookup: type-specific override → general default → `LocalDate.EPOCH`. Validator: ISO-8601 `yyyy-MM-dd`, range `0001-01-01..9999-12-31`. Both DDL parsing (schema default) and row-event conversion (payload value) read from the same cascading source.

The three sentinels are bundled into an immutable `io.debezium.jdbc.ZeroDateFallback` parameter object; the existing `JdbcValueConverters` constructor used by `postgres`/`oracle`/`sqlserver` delegates with `ZeroDateFallback.defaultEpoch()`, so other connectors are unaffected.

## Commits

1. `DBZ-1912 Allow customizing the zero-date fallback value (per type) for MySQL/MariaDB` — config + cascading getter + 12 fallback site swaps + `ZeroDateFallback` parameter object + tests + docs.
2. `DBZ-1912 Detect zero-date in streaming TIMESTAMP deserialization` — adds the `epochSecond == 0` guard in `RowDeserializers.deserializeTimestamp(V2)` so streaming `TIMESTAMP` zero-dates flow through the same fallback infrastructure as `DATE`/`DATETIME`. Without this guard the per-type `*.timestamp` override is silently inert during streaming. MySQL `TIMESTAMP`'s legal range starts at `1970-01-01 00:00:01` UTC, so `epochSecond == 0` unambiguously signals the zero-date sentinel.

## Backward compatibility

All four properties are optional. With no configuration the emitted output is byte-for-byte identical to today's hardcoded `1970-01-01` fallback. No schema changes (sentinel only affects values, not Avro/Protobuf logical types). Other Debezium connectors are unaffected.

## Test plan

- [x] Unit tests for `BinlogConnectorConfig` cascading lookup and validators (`MySqlConnectorConfigTest` +9 cases)
- [x] Unit tests for `MySqlDefaultValueConverter` zero-date sentinel (`MySqlZeroDateFallbackTest`, 7 cases covering DDL parsing path)
- [x] Unit tests for `ZeroDateFallback` parameter object (`ZeroDateFallbackTest`, 7 cases)
- [x] Unit tests for `RowDeserializers.deserializeTimestamp(V2)` zero-detection (`RowDeserializersTest`, 6 cases incl. V2 nanos guard)
- [x] Integration test extension in `BinlogDefaultValueAllZeroTimeIT` for the streaming row payload (covers both MySQL and MariaDB via the abstract base)
- [x] End-to-end Docker validation: MySQL source → Kafka Avro + Schema Registry → JDBC sink to MySQL, across three scenarios (default / uniform `9999-12-31` / per-type `.datetime=1001-01-01`), both snapshot and streaming paths

All units green: `debezium-connector-common` 386, `debezium-connector-binlog` 30, `debezium-connector-mysql` 325, `debezium-connector-mariadb` 289, `debezium-connector-postgres / oracle / sqlserver` compile (no behavior change).

## Out of scope

A `zero.date.handling.mode=null` mode that promotes the schema to `optional` and emits Kafka `null` instead of a sentinel — a schema-breaking change that warrants a separate proposal. The cascading getter introduced here is shaped to support it.

### DIFF
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
@@ -99,9 +99,6 @@ public class ZeroDateFallbackConverter
 
     private static final DateTimeFormatter DATETIME_FMT = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
     private static final String NULL_TOKEN = "NULL";
-    private static final String SCHEMA_DATE = "io.debezium.time.Date";
-    private static final String SCHEMA_TIMESTAMP = "io.debezium.time.Timestamp";
-    private static final String SCHEMA_ZONED_TIMESTAMP = "io.debezium.time.ZonedTimestamp";
     private static final int SCHEMA_VERSION = 1;
 
     private boolean applyToDate;
@@ -188,7 +185,7 @@ public class ZeroDateFallbackConverter
         final Integer fallback = effective == null ? null : (int) effective.toEpochDay();
         final boolean hasDefault = field.hasDefaultValue();
 
-        SchemaBuilder schema = SchemaBuilder.int32().name(SCHEMA_DATE).version(SCHEMA_VERSION);
+        SchemaBuilder schema = SchemaBuilder.int32().name(io.debezium.time.Date.SCHEMA_NAME).version(SCHEMA_VERSION);
         if (fallback == null) {
             schema = schema.optional();
         }
@@ -223,7 +220,7 @@ public class ZeroDateFallbackConverter
         final Long fallback = effective == null ? null : effective.toInstant(ZoneOffset.UTC).toEpochMilli();
         final boolean hasDefault = field.hasDefaultValue();
 
-        SchemaBuilder schema = SchemaBuilder.int64().name(SCHEMA_TIMESTAMP).version(SCHEMA_VERSION);
+        SchemaBuilder schema = SchemaBuilder.int64().name(io.debezium.time.Timestamp.SCHEMA_NAME).version(SCHEMA_VERSION);
         if (fallback == null) {
             schema = schema.optional();
         }
@@ -258,7 +255,7 @@ public class ZeroDateFallbackConverter
         final String fallback = effective == null ? null : effective.format(DateTimeFormatter.ISO_OFFSET_DATE_TIME);
         final boolean hasDefault = field.hasDefaultValue();
 
-        SchemaBuilder schema = SchemaBuilder.string().name(SCHEMA_ZONED_TIMESTAMP).version(SCHEMA_VERSION);
+        SchemaBuilder schema = SchemaBuilder.string().name(io.debezium.time.ZonedTimestamp.SCHEMA_NAME).version(SCHEMA_VERSION);
         if (fallback == null) {
             schema = schema.optional();
         }
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
@@ -1,0 +1,354 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.binlog.converters;
+
+import java.sql.Date;
+import java.sql.Timestamp;
+import java.sql.Types;
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+import java.time.OffsetDateTime;
+import java.time.ZoneOffset;
+import java.time.ZonedDateTime;
+import java.time.format.DateTimeFormatter;
+import java.util.HashSet;
+import java.util.Properties;
+import java.util.Set;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.function.Predicate;
+
+import org.apache.kafka.connect.data.SchemaBuilder;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.debezium.config.Field;
+import io.debezium.function.Predicates;
+import io.debezium.metadata.ConfigDescriptor;
+import io.debezium.spi.converter.CustomConverter;
+import io.debezium.spi.converter.RelationalColumn;
+import io.debezium.util.Strings;
+
+/**
+ * Replaces MySQL/MariaDB zero-date sentinels ({@code 0000-00-00} / {@code 0000-00-00 00:00:00})
+ * with either {@code null} or a user-chosen literal, per JDBC type and optionally per column.
+ *
+ * <h2>Supported types</h2>
+ * <ul>
+ *   <li>{@code DATE} &rarr; {@code io.debezium.time.Date} (INT32 epoch days)</li>
+ *   <li>{@code DATETIME} &rarr; {@code io.debezium.time.Timestamp} (INT64 epoch millis)</li>
+ *   <li>{@code TIMESTAMP} &rarr; {@code io.debezium.time.ZonedTimestamp} (STRING ISO-8601)</li>
+ * </ul>
+ *
+ * <h2>Two policies, selectable per type</h2>
+ * <ul>
+ *   <li><b>NULL policy</b> &mdash; the row and the column schema's {@code defaultValue} slot both
+ *       become {@code null}. The schema is promoted to nullable. Active when the fallback option
+ *       is {@code NULL} (case-insensitive) or unset (default).</li>
+ *   <li><b>User-value policy</b> &mdash; the row and the column schema's {@code defaultValue} are
+ *       set to the configured literal. Schema stays NOT NULL. Any DDL {@code DEFAULT} on the
+ *       column is overridden by this value.</li>
+ * </ul>
+ *
+ * <h2>Resolution order</h2>
+ * Column-level override &gt; type-level fallback &gt; NULL.
+ *
+ * <h2>Configuration</h2>
+ * <ul>
+ *   <li>{@code <prefix>.target.types} &mdash; comma-separated subset of
+ *       {@code DATE,DATETIME,TIMESTAMP}. Default applies to all three.</li>
+ *   <li>{@code <prefix>.fallback.date} / {@code .fallback.datetime} / {@code .fallback.timestamp}
+ *       &mdash; per-type replacement. {@code NULL} (default) or a type-appropriate literal.</li>
+ *   <li>{@code <prefix>.column.<db>.<table>.<column>.fallback} &mdash; column-specific override of
+ *       the type-level value. Look-up key is built from
+ *       {@code field.dataCollection() + "." + field.name()}.</li>
+ *   <li>{@code <prefix>.selector} &mdash; comma-separated regex applied to
+ *       {@code <db>.<table>.<column>}. Default {@code .*} (every matching-type column).</li>
+ * </ul>
+ *
+ * <h2>How it interacts with DDL defaults</h2>
+ * The implementation relies on {@code TableSchemaBuilder} invoking the registered conversion
+ * lambda exactly once at schema-build time (for resolving the column's {@code defaultValue}
+ * slot). A per-column {@link AtomicBoolean} guards that first call:
+ * <ul>
+ *   <li>Under the NULL policy, the lambda returns {@code null} on the first call, which leaves
+ *       the schema's {@code defaultValue} slot empty so AvroConverter does not back-fill null rows
+ *       with the DDL default.</li>
+ *   <li>Under the user-value policy, the lambda returns the configured literal on the first call,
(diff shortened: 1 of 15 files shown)


Naros (commented):
Hi @minleejae, I am curious if there was a reason why introducing a CustomConverter for environments that wanted this level of granularity couldn't be used instead?

minleejae:
Hi @Naros, thank you for the review.

The reason a `CustomConverter` was not used here is that, by the time a converter is invoked, the zero-date pattern has already been erased and cannot be recovered. On the streaming path, `RowDeserializers` decodes the raw binlog bytes and collapses zero-dates inline — `null` for nullable columns, and the historical `1970-01-01` epoch fallback (via `ZeroDateFallback.defaultEpoch()` in `JdbcValueConverters`) for non-nullable columns. On the snapshot path, the JDBC driver's `zeroDateTimeBehavior` rewrites the value before Debezium observes it. 

As a result, a converter cannot distinguish a genuine `NULL` from a nulled zero-date, nor a legitimate `1970-01-01` row from a row that was epoch-fallbacked, which makes the SPI an unsuitable place to apply per-type policy.

Implementing this in the source connector lets the decision be made where the raw pattern is still observable, and it allows a single `ZeroDateFallback` policy to be shared between the snapshot and streaming paths. Given that zero-date can arise in any `DATE` / `DATETIME` / `TIMESTAMP` column, a connector-level configuration also felt like the appropriate scope. From a user's perspective as well, I believe a natively supported option in Debezium has a much lower barrier to adoption than authoring, packaging, and registering a dedicated `CustomConverter`. That said, if you see a converter-side approach I've overlooked, I'd be glad to revisit the design.

vjuranek:
Thanks for the PR @minleejae . Could you please take a look on failed tests? The failures are related to this changes.

minleejae:
Hi, @Naros , @vjuranek 

First, thanks for the careful review and discussion so far—it genuinely helped me think more carefully about the design boundary between core and SPI here.

Once this PR lands, I'd like to discuss a follow-up I've prototyped as a `CustomConverter` SPI. I initially assumed these capabilities weren't expressible through the SPI, 
but after working through a prototype it turned out both were achievable — and that prototype also made the case for landing these in core clearer to me. The two gaps I'd like to address:

1. **Column-level granularity** — adding `zero.date.fallback.value.column.<db>.<table>.<column>` to allow precise overrides for specific legacy columns.
2. **`null` as a fallback for NOT NULL columns** — currently, `ZeroDateFallback` only accepts `LocalDate`, making `null` inexpressible for NOT NULL fields. Internally, this would involve conditionally promoting the Connect schema to optional and clearing its `defaultValue` when this specific fallback is chosen.

--- 
While I have implemented this as a private SPI internally, I believe it belongs in Debezium core because:
- Zero-date isn't team-specific; it's a hurdle for anyone migrating legacy MySQL data. 
- Standardizing this in core reduces the operational burden for users who would otherwise have to maintain a custom `CustomConverter` implementation for a common edge case.

Happy to open follow-up PRs if this direction sounds reasonable, or take guidance if SPI is the preferred long-term path for these edge cases.

vjuranek:
I tend to agree to keep this in Debezium core as it's legitimate use case to customize zero fallback date. However, not sure how common this requirement is.
I did initial review and the code LGTM. If the teams agree to keep it in Debezium core, I'll do eventually more details review (documentation, tests, nitpicks in the code etc.)

minleejae:
Hi, @Naros @vjuranek 
Thanks for the feedback.

For our use case, our data engineers concluded that transforming MySQL zerodate into null is the best approach, even for NOT NULL columns.

Although I only submitted the custom value conversion in this PR, implementing this via a custom SPI adds maintenance overhead for our team. That is why we are hoping to see this capability supported natively in Debezium core.

Please let me know if you or the team have any further thoughts on this. Thank you!

Naros:
> For our use case, our data engineers concluded that transforming MySQL zerodate into null is the best approach, even for NOT NULL columns.

I definitely agree if this is a requirement, this highlights the significant value of a `CustomConverter`, as I believe it keeps the changes required minimal.

> Standardizing this in core reduces the operational burden for users who would otherwise have to maintain a custom CustomConverter implementation for a common edge case.

I would also argue here that ZERODATE support should be something very specific to Binlog, and not Core. That's probably one of the biggest concerns I have with the current implementation is that we're doing a lot minor tweaks inside `JdbcValueConverters`, which I believe should be specific covered by a dedicated `CustomConverter` that plugs in overrides for all these use cases.  But ymmv.

@vjuranek thoughts?

vjuranek:
> I would also argue here that ZERODATE support should be something very specific to Binlog, and not Core.

Well, this is a good point and I have to agree here. @minleejae if I understand correctly, you main concern if maintenance cost if you develop and maintain your own converter. However, the custom converters like this can be part of Debezium, we already have sevelar of them, those specific to binlog (mysql, mariadb) are [here](https://github.com/debezium/debezium/tree/main/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters). Would that work for you to re-implement it as a custom converter which is part of Debezium codebase so if can be use out-of-the-box with any future Debezium release?

vjuranek (approved):
LGTM

minleejae:
@Naros @vjuranek Thanks for taking the time to weigh in — I agree with both of your suggestions, so I force-pushed a reimplementation that keeps `debezium-connector-common` untouched and ships the feature as an in-tree CustomConverter.                                                                                                                                                                                                                                  
The two remaining CI failures don't touch the zero-date code path and look unrelated to this PR.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
```diff
@@ -0,0 +1,354 @@
...
+public class ZeroDateFallbackConverter
+        implements CustomConverter<SchemaBuilder, RelationalColumn>, ConfigDescriptor {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(ZeroDateFallbackConverter.class);
+
+    private static final DateTimeFormatter DATETIME_FMT = DateTimeFormatter.ofPattern("yyyy-MM-dd HH:mm:ss");
+    private static final String NULL_TOKEN = "NULL";
+    private static final String SCHEMA_DATE = "io.debezium.time.Date";
+    private static final String SCHEMA_TIMESTAMP = "io.debezium.time.Timestamp";
+    private static final String SCHEMA_ZONED_TIMESTAMP = "io.debezium.time.ZonedTimestamp";
```
Naros: Should we not use the `SCHEMA_NAME` exposed values on the class types?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
```diff
@@ -0,0 +1,354 @@
...
+        final int jdbcType = field.jdbcType();
+        final String typeName = field.typeName() != null ? field.typeName().toUpperCase() : "";
+
+        if (jdbcType == Types.DATE && applyToDate) {
+            registerDate(field, registration);
+        }
+        else if (jdbcType == Types.TIMESTAMP && applyToDatetime && "DATETIME".equals(typeName)) {
+            registerDatetime(field, registration);
+        }
+        else if ((jdbcType == Types.TIMESTAMP_WITH_TIMEZONE || "TIMESTAMP".equals(typeName)) && applyToTimestamp) {
```
Naros: We use `"DATE"`, `"DATETIME"`, and `"TIMESTAMP"` strings in multiple places. Should we define these as private static strings in the converter?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
```diff
@@ -0,0 +1,354 @@
...
+    private static void logRegistration(String typeLabel, RelationalColumn field, String columnRaw,
+                                        boolean hasDefault, Object fallback) {
+        if (!LOGGER.isDebugEnabled()) {
+            return;
+        }
+        LOGGER.debug("Registered {} converter on '{}.{}' source={} hasDefault={} fallback={}",
+                typeLabel, field.dataCollection(), field.name(),
+                columnRaw != null ? "column-level" : "type-level",
+                hasDefault,
+                fallback == null ? "NULL" : fallback);
```
Naros: Should this use the `NULL_TOKEN`?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
```diff
@@ -0,0 +1,354 @@
...
+
+    static OffsetDateTime parseTimestampFallback(String raw) {
+        if (isNullToken(raw)) {
+            return null;
+        }
+        return OffsetDateTime.parse(raw.trim()).withOffsetSameInstant(ZoneOffset.UTC);
+    }
+
+    private static boolean isNullToken(String raw) {
+        return raw == null || raw.trim().isEmpty() || NULL_TOKEN.equalsIgnoreCase(raw.trim());
```
Naros:
```suggestion
        return Strings.isNullOrBlank(raw) || NULL_TOKEN.equalsIgnoreCase(raw.trim());
```

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
```diff
@@ -0,0 +1,354 @@
...
+    private Properties rawProps;
+
+    @Override
+    public void configure(Properties props) {
+        this.rawProps = props;
+
+        final String targets = props.getProperty(
+                ZeroDateFallbackConverterConfig.TARGET_TYPES.name(),
+                ZeroDateFallbackConverterConfig.DEFAULT_TARGET_TYPES).toUpperCase();
+        final Set<String> tokens = parseCsv(targets);
```
Naros, line 124: Could this not use the `Strings` class, e.g. `listOf` or `listOfTrimmed`, `setOf`?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
minleejae: Removed the local SCHEMA_* constants and now reference io.debezium.time{Date,Timestamp,ZonedTimestamp}.SCHEMA_NAME [758bc39](https://github.com/debezium/debezium/pull/7415/commits/758bc3951ea4c4777c3d7457a8663095b2a1185c)

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
minleejae: Extracted TYPE_DATE, TYPE_DATETIME, TYPE_TIMESTAMP constants and replaced all five call sites.[0852182](https://github.com/debezium/debezium/pull/7415/commits/0852182ccdab87fe3e565db5cdefd3beffd5e9b5)

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
minleejae: Yes done in [43931a2](https://github.com/debezium/debezium/pull/7415/commits/43931a2c07a3f525b4c35040277dc7a031110f3d)

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
minleejae: Applied your suggestion in [0be53a6](https://github.com/debezium/debezium/pull/7415/commits/0be53a6b1fa0dedbd889f57bf1601dc5bdc621ef)

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
minleejae: Done in [7fb4fb0](https://github.com/debezium/debezium/pull/7415/commits/7fb4fb0fd85ca2894be53cc818532a912b6722b8)

Naros (approved):
@vjuranek if this still LGTY, it's good for me.

vjuranek:
Merged, thanks for you effort @minleejae !

PR #7423, opened 2026-05-09
debezium/dbz#1914: idempotent logminer partial rollbacks (memory)

### LINKED ISSUE #1914: Non-idempotent partial rollbacks
## Bug report

`debezium-connector-oracle:3.6.0.Alpha1`

---

`database.connection.adapter=logminer`

---

`container-registry.oracle.com/database/express:21.3.0-xe`

---

When debezium finds and applies a partial rollback event in a transaction that doesn't end in the current mining session, it shouldn't try to apply this rollback event again to another event in the next mining session.

---

**What behavior do you see?**

Example 1
```sql
savepoint s1;
insert into t1(id, data) values (1, 'insert');
rollback to savepoint s1;
```
logminer
```tsv
scn	tran	type	rollback	rowid
1	ta		start
2	ta		insert	0			ra
3	ta		delete	1			ra
4	ta		commit
```
If the first mining session ends at scn 3, the insert will be marked as rolled back, during the next session it will be skipped, and we'll get a warning `Cannot apply undo change in transaction 'ta' with SCN '3' on table 't1' since event with row-id ra was not found`.

Example 2
```sql
insert into t1(id, data) values (1, 'insert');
savepoint s1;
update t1 set data = 'update' where id = 1;
rollback to savepoint s1;
```
```tsv
scn	tran	type	rollback	rowid
1	tb		start
2	tb		insert	0			rb
3	tb		update	0			rb
4	tb		update	1			rb
5	tb		commit
```
If the first mining session ends at scn 4, the update will be marked as rolled back, during the next session it will be skipped and the insert will be marked as rolled back.

---

Yes

---

```java
package io.debezium.connector.oracle.logminer.buffered;

public abstract class AbstractBufferedLogMinerStreamingChangeEventSourceTest extends AbstractAsyncEngineConnectorTest {
    // ...
    @Test
    public void testSavepointRollbackIdempotence() throws Exception {
        final Configuration config = getConfig().build();

        try (var source = getChangeEventSource(config)) {
            source.processEvent(getStartLogMinerEventRow(1, TRANSACTION_ID_1));
            source.processEvent(getInsertLogMinerEventRow(2, TRANSACTION_ID_1, Instant.now(), "TEST_TABLE", "AAAAAAAAAAAAAAAAAB", "'insert'"));
            source.processEvent(getUpdateLogMinerEventRow(3, TRANSACTION_ID_1, Instant.now(), "TEST_TABLE", "AAAAAAAAAAAAAAAAAB", "'update'"));
            source.processEvent(getRollbackToSavepointLogMinerEventRow(4, TRANSACTION_ID_1, Instant.now(), "TEST_TABLE", "AAAAAAAAAAAAAAAAAB", "'insert'"));
            // Simulate a new mining session
            source.processEvent(getStartLogMinerEventRow(1, TRANSACTION_ID_1));
            source.processEvent(getInsertLogMinerEventRow(2, TRANSACTION_ID_1, Instant.now(), "TEST_TABLE", "AAAAAAAAAAAAAAAAAB", "'insert'"));
            source.processEvent(getUpdateLogMinerEventRow(3, TRANSACTION_ID_1, Instant.now(), "TEST_TABLE", "AAAAAAAAAAAAAAAAAB", "'update'"));
            source.processEvent(getRollbackToSavepointLogMinerEventRow(4, TRANSACTION_ID_1, Instant.now(), "TEST_TABLE", "AAAAAAAAAAAAAAAAAB", "'insert'"));
            source.processEvent(getCommitLogMinerEventRow(5, TRANSACTION_ID_1));
            Mockito.verify(dispatcher, Mockito.times(1))
                    .dispatchDataChangeEvent(any(), any(), argThat(emitter -> emitter.getOperation() == Operation.CREATE));
        }
    }
    // ...
}
```

kinolaev:
Fixes debezium/dbz#1914

## Description
This PR makes partial rollbacks idempotent when using memory transaction cache.

## PR Checklist

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
@@ -947,7 +947,8 @@ public class BufferedLogMinerStreamingChangeEventSource extends AbstractLogMiner
      * @return true if an event was found and undone, false otherwise
      */
     private boolean rollbackTransactionEventWithRowId(Transaction transaction, LogMinerEventRow row) {
-        if (getTransactionCache().rollbackTransactionEventWithRowId(transaction, row.getRowId())) {
+        final int rollbackId = transaction.getNextEventId();
+        if (getTransactionCache().rollbackTransactionEventWithRowId(transaction, rollbackId, row.getRowId())) {
             // This metric won't necessarily be accurate when LOB is enabled, it will scale based on the
             // number of times a given transaction is re-mined.
             getMetrics().increasePartialRollbackCount();
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/LogMinerTransactionCache.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/LogMinerTransactionCache.java
@@ -151,10 +151,11 @@ public interface LogMinerTransactionCache<T extends Transaction> {
      * Marks as rolled back a specific transaction event by unique row identifier.
      *
      * @param transaction the transaction, should not be {@code null}
+     * @param rollbackId the rollback event's identifier
      * @param rowId the event's unique row identifier
      * @return {@code true} if the event was found and removed, {@code false} if it was not found
      */
-    boolean rollbackTransactionEventWithRowId(T transaction, String rowId);
+    boolean rollbackTransactionEventWithRowId(T transaction, int rollbackId, String rowId);
 
     /**
      * Checks whether a specific transaction's event with the event key is cached.
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/ehcache/EhcacheLogMinerTransactionCache.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/ehcache/EhcacheLogMinerTransactionCache.java
@@ -154,7 +154,7 @@ public class EhcacheLogMinerTransactionCache extends AbstractLogMinerTransaction
     }
 
     @Override
-    public boolean rollbackTransactionEventWithRowId(EhcacheTransaction transaction, String rowId) {
+    public boolean rollbackTransactionEventWithRowId(EhcacheTransaction transaction, int rollbackId, String rowId) {
         final long encodedRowId = RowIdCodec.encode(rowId);
         final TreeSet<Integer> eventIds = eventIdsByTransactionId.get(transaction.getTransactionId());
         for (Integer eventId : eventIds.descendingSet()) {
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/infinispan/InfinispanLogMinerTransactionCache.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/infinispan/InfinispanLogMinerTransactionCache.java
@@ -142,7 +142,7 @@ public class InfinispanLogMinerTransactionCache extends AbstractLogMinerTransact
     }
 
     @Override
-    public boolean rollbackTransactionEventWithRowId(InfinispanTransaction transaction, String rowId) {
+    public boolean rollbackTransactionEventWithRowId(InfinispanTransaction transaction, int rollbackId, String rowId) {
         final long encodedRowId = RowIdCodec.encode(rowId);
         final TreeSet<Integer> eventIds = eventIdsByTransactionId.get(transaction.getTransactionId());
         for (Integer eventId : eventIds.descendingSet()) {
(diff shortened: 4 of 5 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/LogMinerTransactionCache.java
```diff
@@ -154,7 +154,7 @@ public interface LogMinerTransactionCache<T extends Transaction> {
      * @param rowId the event's unique row identifier
```
Naros, line 155: Please update javadocs.

Naros (commented):
I still need to wait on CI, but the solution appears sound. I've left just one nitpick.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/LogMinerTransactionCache.java
kinolaev: done in 4f9d05aedc962de81219efa9c8879469e342be9d

kinolaev:
@Naros , why do you think these changes are backward incompatible? Is it because of the additional calls of `Transaction.getNextEventId` for rollback events? Yes, it can cause duplicated events when using eh/ispn cache and transactions there have rollbacks. But nothing will be lost and no actions required during upgrade. Am I missing something?

I moved the test `testSavepointRollbackIdempotence` to the branch `dbz-1914-eh-ispn` (#7424), because it fails with eh/ispn caches in this branch.

Naros:
@kinolaev we use the backward-incompatible label for dual purpose. It can indicate where a feature breaks backward compatibility or we use it to indicate that the incorporation of the feature should have some explicit "breaking changes" section in the release notes.

While this does not directly break the heap-based implementation, the entire feature of dbz#1914 as a whole does introduce a release notes breaking change notice for Infinispan/Ehcache, which is why I marked it as such. No other reason.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -947,7 +947,8 @@ else if (!getConfig().isLobEnabled()) {
      * @return true if an event was found and undone, false otherwise
      */
     private boolean rollbackTransactionEventWithRowId(Transaction transaction, LogMinerEventRow row) {
-        if (getTransactionCache().rollbackTransactionEventWithRowId(transaction, row.getRowId())) {
+        final int rollbackId = transaction.getNextEventId();
```
Naros, line 950: I am specifically concerned with this next event ID assignment.

Inside `removeEventWithRowId`, there are two blocks that call into this method, one that returns true because the transaction ID matched, and then N potential calls while we iterate, where the transaction ID is matched by prefix.

If nothing else, I would argue that this `rollbackId` should be assigned one level higher when we call `removeEventWithRowId` and then passed in here, so that we are only doing a single allocation rather than potentially N.

kinolaev: But we don't have a transaction object to call `getNextEventId` one level higher.
In practice `rollbackTransactionEventWithRowId` is always called once per row: a transaction is either matched by full id or by prefix. And in any particular moment there is only one active transaction with the given prefix (because prefix is a pointer to a physical location as I know).

Naros:
Hi @kinolaev can you rebase this on main, so we get the latest TF updates?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
Naros: True.

My biggest concern right now is the performance impact of having to mine from the start of the transaction. If we decide to change this behavior for any reason, all logic around `getNextEventId` becomes problematic, and I would genuinely prefer to avoid adding to it if at all possible.

I guess for now we can leave this as is, but I do think we need to take a hard pass over all the recent changes since DBZ-9615. I think now that we've iterated on it over a period of multiple PRs, and now that we see the larger picture, perhaps there is a more efficient, more importantly, less complicated approach for savepoints that maybe doesn't require buffering them at all or do buffer them makes the process less complicated?

For example, just buffer the `isRollbackFlag()` based events like any other; they get an `eventId`. Then, at commit time, we try to do all this complicated post-processing. We already do this for merging and a few other use cases, so why not all this for savepoint/partial rollbacks? 

This would eliminate `rollbacksByTransactionId` with no Set, Map, or rollbackId.

Naros:
Hi @kinolaev can you remind me, is this superseded by #7472 ?

kinolaev:
Yes, since in #7472 we store `RollbackToSavepoint` events in the cache like other events, we always know when a rollback has already been processed.

Naros:
Perfect, superseded by #7472

PR #7425, opened 2026-05-10
debezium/dbz#1735: fix partial rollbacks for out-of-line lobs

### LINKED ISSUE #1735: Oracle Savepoint Partial Rollback not handled when LOB column changed
## Bug report

Sourced from https://redhat.atlassian.net/browse/DBZ-9615 and https://github.com/debezium/dbz/issues/1422.

This is a follow-up where a user inserts a non-null vaule into a LOB field within a SAVEPOINT operation and then the transaction rolls back the SAVEPOINT, seen here:

Image

In this specific use case, the `DELETE` operation ROWID will not directly map to the `UPDATE` event's ROWID, and this is problematic because the connector cannot make the connection between the two operations, and successfully roll the update event back.

Delete
```sql
delete from "DEBEZIUM"."DBZ915" where ROWID = 'AAAS2MAAMAAAACUAAB';
```
Update
```sql
update "DEBEZIUM"."DBZ9615" set "DATA" = NULL where "ID" = '1' AND ROWID = 'AAAS2MAAMAAAACUAAA`;
```

kinolaev:
Fixes debezium/dbz#1735

## Description
This PR is based on the assumption that all cached events with an empty ROW_ID, followed by an event with a non-empty ROW_ID, must be merged. In other words, a transaction can have only one active complex operation at any given moment.

The approach is simple: when an incoming INTERNAL event has a non-empty ROW_ID, update the ROW_ID of the last transaction event if it is empty.

This PR adds INTERNAL events with non-empty ROW_IDs to the oracle logminer buffered query builder, but they aren't stored in the transaction cache.

## PR Checklist

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -593,6 +593,17 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             .withDefault(false)
             .withValidation(OracleConnectorConfig::validateLogMiningIncludeRedoSql);
 
+    public static final Field LOG_MINING_INCLUDE_INTERNAL_EVENTS = Field.create("log.mining.include.internal.events")
+            .withDisplayName("Specifies whether the connector supports mining INTERNAL events")
+            .withType(Type.BOOLEAN)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withDefault(false)
+            .withDescription("When set to 'false', the default, INTERNAL events will not be captured. " +
+                    "When 'lob.enabled' is 'true' and all LOB columns of an operation are stored out-of-line, " +
+                    "INTERNAL events provide the ROW_IDs necessary for accurate ROLLBACK TO SAVEPOINT handling. " +
+                    "NOTE: Enabling this may significantly increase the volume of events returned by LogMiner.");
+
     public static final Field SNAPSHOT_DATABASE_ERRORS_MAX_RETRIES = Field.create("snapshot.database.errors.max.retries")
             .withDisplayName("The maximum number of retries before snapshot database errors are not retried")
             .withType(Type.INT)
@@ -930,6 +941,7 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
     private final String logMiningInifispanGlobalConfiguration;
     private final Set<String> logMiningSchemaChangesUsernameExcludes;
     private final Boolean logMiningIncludeRedoSql;
+    private final boolean logMiningIncludeInternalEvents;
     private final boolean logMiningContinuousMining;
     private final Configuration logMiningEhCacheConfiguration;
     private final boolean logMiningUseSqlRelaxedQuoteDetection;
@@ -1010,6 +1022,7 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
         this.logMiningInifispanGlobalConfiguration = config.getString(LOG_MINING_BUFFER_INFINISPAN_CACHE_GLOBAL);
         this.logMiningSchemaChangesUsernameExcludes = Strings.setOf(config.getString(LOG_MINING_SCHEMA_CHANGES_USERNAME_EXCLUDE_LIST), String::new);
         this.logMiningIncludeRedoSql = config.getBoolean(LOG_MINING_INCLUDE_REDO_SQL);
+        this.logMiningIncludeInternalEvents = config.getBoolean(LOG_MINING_INCLUDE_INTERNAL_EVENTS);
         this.logMiningContinuousMining = config.getBoolean(LOG_MINING_CONTINUOUS_MINE);
         this.logMiningUseSqlRelaxedQuoteDetection = config.getBoolean(LOG_MINING_SQL_RELAXED_QUOTE_DETECTION);
         this.logMiningClientIdIncludes = Strings.setOfTrimmed(config.getString(LOG_MINING_CLIENTID_INCLUDE_LIST), String::new);
@@ -2016,6 +2029,13 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
      */
     public boolean isLogMiningIncludeRedoSql() {
         return logMiningIncludeRedoSql;
+    }
+
+    /**
+        * @return true if INTERNAL events are to be captured.
+        */
+    public boolean isLogMiningIncludeInternalEvents() {
+        return logMiningIncludeInternalEvents;
     }
 
     /**
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerQueryBuilder.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerQueryBuilder.java
@@ -136,12 +136,21 @@ public class BufferedLogMinerQueryBuilder extends AbstractLogMinerQueryBuilder {
         operationInClause.withValues(getOperationCodesList(isCteQuery));
         predicate.append("(").append(operationInClause.build());
 
+        // Handle INTERNAL operations that can contain ROW_IDs for INSERT/UPDATE operations with empty ROW_IDs
+        if (connectorConfig.isLobEnabled() && connectorConfig.isLogMiningIncludeInternalEvents()) {
+            predicate.append(getInternalPredicate());
+        }
+
         // Handle DDL operations
         if (connectorConfig.storeOnlyCapturedTables()) {
             predicate.append(getDdlPredicate());
         }
 
         return predicate.append(")").toString();
+    }
+
+    private static String getInternalPredicate() {
+        return " OR (OPERATION_CODE = 0 AND ROLLBACK = 0 AND ROW_ID NOT LIKE '%AAAAAAAAAAAA')";
     }
 
     private static String getDdlPredicate() {
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
@@ -366,6 +366,16 @@ public class BufferedLogMinerStreamingChangeEventSource extends AbstractLogMiner
     }
 
     @Override
+    protected void handleInternalEvent(LogMinerEventRow event) {
+        if (!event.isRollbackFlag() && event.getRowId() != null && !event.getRowId().endsWith("AAAAAAAAAAAA")) {
+            final Transaction transaction = getTransactionCache().getTransaction(event.getTransactionId());
+            if (transaction != null) {
+                getTransactionCache().updateLastEventEmptyRowId(transaction, event.getRowId());
+            }
+        }
+    }
+
+    @Override
     protected void handleStartEvent(LogMinerEventRow event) {
         final String transactionId = event.getTransactionId();
         if (!isRecentlyProcessed(transactionId)) {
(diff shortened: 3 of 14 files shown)


kinolaev:
@Naros , I've added `log.mining.include.internal.events` flag and a comprehensive integration test `AbstractBufferedLogMinerStreamingChangeEventSourceIT#shouldRollbackEventsWithLobs`. The PR is ready for review. I suggest to review and merge https://github.com/debezium/debezium/pull/7427 first as it contains a part of this PR. After merging https://github.com/debezium/debezium/pull/7427 I'll do a rebase on main to keep changes related to out-of-line LOB rollbacks isolated.

kinolaev:
Fixed `LogMinerQueryBuilderTest`

Naros:
Hi @kinolaev regarding this

> This PR is based on the assumption that all cached events with an empty ROW_ID, followed by an event with a non-empty ROW_ID, must be merged. In other words, a transaction can have only one active complex operation at any given moment.

I am not sure we can rely on this. At one point I had wanted to to simplify the `TransactionCommitConsumer` where I too had hoped that at only one given time Oracle would emit a series of LOB operations that were specific to a single column and that there would never be any interleaved, but that's not always the case.

So I really question whether we can truly rely on what you outlined above for this reason.

Naros:
If I understand correctly, there may be situations where the merge logic in `TransactionCommitConsumer` can avoid merging LOB data due to patching of `ROW_ID`. Which leads me to ask if there might be a cleaner, more solid solution that gets us where you want to be, but with a more surgical approach.

Could we consider introducing a heap-backed ROW_ID resolution map that would map event keys to row ids. When an `INTERNAL` event arrives with a real `ROW_ID`, find the most recent event with an empty `ROW_ID` in the transaction, record the event's event key to real `ROW_ID` in this resolution map. 

Then inside `rollbackTransactionEventWithRowId`, when we scan the transaction in reverse, if an event has an empty ROW_ID, we lookup the resolved ROW_ID in the resolution map and compare against the rollback event's ROW_ID. I'm thinking something like this:

```java
  // In the existing rollback scan loop:
  long effectiveRowId = entry.event.getRowId();
  if (effectiveRowId == RowIdCodec.EMPTY_ROW_ID) {
      Long resolved = resolvedRowIds.get(transaction.getEventId(entry.eventId));
      if (resolved != null) {
          effectiveRowId = resolved;
      }
  }
  if (effectiveRowId == encodedRowId && !rollbacks.contains(entry.eventId)) {
      rollbacks.add(entry.eventId);
      return true;
  }
```

In this case, `LogMinerEvent` remains immutable, `TransactionCommitConsumer` logic remains as is, and we don't risk touching any merge logic that could be brittle in some corner cases.

What's your thoughts?

kinolaev:
> I too had hoped that at only one given time Oracle would emit a series of LOB operations that were specific to a single column and that there would never be any interleaved, but that's not always the case.

Could you please share how to reproduce the case when a row with non-empty ROW_ID isn't related to a previous row with empty ROW_ID in the same transaction? I'd like to try find a way to make correct correlations.

> `TransactionCommitConsumer` logic remains as is

All changes to `TransactionCommitConsumer` are actually related to the issue https://github.com/debezium/dbz/issues/1917. I've extracted this changes to a separate PR but also kept them here because `AbstractBufferedLogMinerStreamingChangeEventSourceIT#shouldRollbackEventsWithLobs` fails otherwise. After https://github.com/debezium/debezium/pull/7427 is merged and this branch is rebased, there are will be no changes to `TransactionCommitConsumer`.

> `LogMinerEvent` remains immutable

Of course, I can add a heap-backed map. But then we also need to pass `effectiveRowId` to `TransactionCommitConsumer` for the reason described in https://github.com/debezium/dbz/issues/1917. You don't like the idea of updating a cached item or a `LogMinerEvent` object? I mean, do you want to have non-empty ROW_ID in a merged `LogMinerEvent`? In https://github.com/debezium/debezium/pull/7427 I check ROW_ID of merged event and stop merging if it isn't empty, but of course I can add a property to `RowState` instead of mutating `LogMinerEvent`.

kinolaev:
I've simplified the implementation: we don't actually need a map of last events with empty ROW_IDs per transaction, instead when we receive INTERNAL with non-empty ROW_ID we can update ROW_ID of the last transaction event in the cache if the ROW_ID is empty. I've also updated the integration test, it is now focused on handling INTERNALs instead of listing all possible LOB modification event combinations.

I've also added a map<eventId, rowId> to keep events in the cache immutable. I prefer to update cache items, because 1) it is easier implementation, 2) the behavior is behind the flag, so no cache mutations by default, 3) even with the flag enabled I don't expect many additional writes to the cache.

@Naros , let me know your preference:
- mutable `LogMinerEvent.rowId` and mutable cache (50af9383d43caae8dd47a8a5f637a11097d6f103)
- mutable `LogMinerEvent.rowId` and immutable cache ([both commits](https://github.com/debezium/debezium/pull/7425/changes))
- immutable `LogMinerEvent.rowId` (not a problem to implement)

kinolaev:
Rebased to get the latest TF updates

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/ehcache/EhcacheLogMinerTransactionCache.java
```diff
@@ -151,6 +165,17 @@ public void removeTransactionEvents(EhcacheTransaction transaction) {
...
+    }
+
+    @Override
+    public void updateLastEventEmptyRowId(EhcacheTransaction transaction, String rowId) {
+        final int numberOfEvents = transaction.getNumberOfEvents();
+        if (numberOfEvents > 0) {
+            rowIdsByEventIdByTransactionId
+                    .computeIfAbsent(transaction.getTransactionId(), id -> new HashMap<>())
+                    .computeIfAbsent(numberOfEvents - 1, id -> RowIdCodec.encode(rowId));
+        }
```
Naros, line 178: In the memory/heap implementation, we check whether the `ROW_ID` for a given event index is the `EMPTY_ROW_ID` before we set the value on the `LogMinerEvent`; however, here and in the Infinispan implementations, we blindly set it in the map regardless, as long as the key doesn't already exist. Given the nature of how these two implementations work, that would always be true unless this method has been called for that event key. This creates a slight difference in behavior, no?

To maintain the same behavior and minimize cache reads for Ehcache/Infinispan, should we check during `addTransactionEvent` whether the event associated with the eventKey has an `EMPTY_ROW_ID` and, if not, pre-register it in the map so we don't mistakenly register it here?

If we align this, then the config option about including internal events should guard whether we populate the `rowIdsByEventIdByTransactionId` during the `addTransactionEvent` to keep the heap low when the feature is not enabled.

debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/logminer/buffered/AbstractBufferedLogMinerStreamingChangeEventSourceTest.java
```diff
@@ -703,6 +721,16 @@ private LogMinerStreamingChangeEventSourceMetrics createMetrics(OracleDatabaseSc
...
 
+    private LogMinerEventRow getInternalLogMinerEventRow(long scn, String transactionId, Instant changeTime, String rowId) {
+        LogMinerEventRow row = Mockito.mock(LogMinerEventRow.class);
+        Mockito.when(row.getEventType()).thenReturn(EventType.INTERNAL);
+        Mockito.when(row.getTransactionId()).thenReturn(transactionId);
+        Mockito.when(row.getScn()).thenReturn(Scn.valueOf(scn));
+        Mockito.when(row.getChangeTime()).thenReturn(changeTime);
+        Mockito.when(row.getRowId()).thenReturn(rowId);
+        return row;
+    }
```
Naros, line 732: This runtime code path explicitly checks `isRollbackFlag`, so should that be mocked here?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -2018,6 +2031,13 @@ public boolean isLogMiningIncludeRedoSql() {
         return logMiningIncludeRedoSql;
     }
 
+    /**
+        * @return true if INTERNAL events are to be captured.
+        */
```
Naros, line 2036: Javadoc indentation needs adjustment.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/events/LogMinerEvent.java
```diff
@@ -54,6 +54,10 @@ public Long getRowId() {
         return rowId;
     }
 
+    public void setRowId(long rowId) {
+        this.rowId = rowId;
+    }
```
Naros, line 59: Making this mutable could create issues with hash-based collections that rely on hashCode/equals. When the `rowId` changes, those values change. Should we consider dropping `rowId` from the `equals` and `hashCode` computations?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -593,6 +593,17 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             .withDefault(false)
             .withValidation(OracleConnectorConfig::validateLogMiningIncludeRedoSql);
 
+    public static final Field LOG_MINING_INCLUDE_INTERNAL_EVENTS = Field.create("log.mining.include.internal.events")
```
Naros, line 596: This should be added to the `oracle.adoc` connector properties section. In addition, this field should be added to the `CONFIG_DEFINITION` static variable's declaration.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -365,6 +365,16 @@ protected boolean isEventSkipped(LogMinerEventRow event) {
         return false;
     }
 
+    @Override
+    protected void handleInternalEvent(LogMinerEventRow event) {
+        if (!event.isRollbackFlag() && event.getRowId() != null && !event.getRowId().endsWith("AAAAAAAAAAAA")) {
```
Naros, line 370: Could we add the 12 `A` string as a public static value in `RowIdCodec` and document it as follows, just so others who read this understand Oracle's internals:

> A ROW_ID is composed of 18 characters, which adhere to the following rules:
> Format: OOOOOOFFFBBBBBBRRR
> OOOOOO - 6 characters - represents the data object number (never zero)
> FFF - 3 characters - represents the relative file number (never zero)
> BBBBBB - 6 characters - represents the block number (rows are never in the header, which is 0 block)
> RRR - 3 characters - row number in the block
>
> So a string that ends with 12 `A` characters represents file=0, block=0, row=0, which is effectively impossible for a row location inside any Oracle database file.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerQueryBuilder.java
```diff
@@ -144,6 +149,10 @@ private String getOperationCodePredicate(boolean isCteQuery) {
         return predicate.append(")").toString();
     }
 
+    private static String getInternalPredicate() {
+        return " OR (OPERATION_CODE = 0 AND ROLLBACK = 0 AND ROW_ID NOT LIKE '%AAAAAAAAAAAA')";
```
Naros, line 153: Let's also refer to the same 12-character `A` string declaration with the documentation I mentioned above.

Naros (commented):
Hi @kinolaev, this is looking really close. I've left a few inline comments.

With regard to your comment above about cache immutability, I think it's a matter of my point below about whether we want `rowId` to participate in the event's equality and hash algorithms. I generally prefer that we try to be consistent, but I can also understand the trade-offs, too. 

So our option is either (a) `rowId` is removed from `equals`/`hashCode` or (b) it remains and we force `LogMinerEvent` to be universally immutable. Do you have a preference?

Naros:
> Could you please share how to reproduce the case when a row with non-empty ROW_ID isn't related to a previous row with empty ROW_ID in the same transaction? I'd like to try find a way to make correct correlations.

I'm not entirely sure how to reproduce it yet, but an Oracle colleague of mine suggested not to make too many assumptions about the order of inter-operations, particularly the materialized `INTERNAL` operations since the public LogMiner API does not provide sufficient data for Debezium to make too many decisions based on.

Naros:
Superseded by #7472 ?

kinolaev:
Yes, because in #7472 we remove all rolled-back events before passing them to `TransactionCommitConsumer`.

Naros:
Superseded by #7472

PR #7434, opened 2026-05-13
debezium/dbz#1928 Add YDB storage module for offsets and schema history

vyrexd:
Fixes debezium/dbz#1928

## Description
  Adds a new `debezium-storage-ydb` module providing YDB-backed implementations
  of the offset backing store and the database schema history store.

  The module includes:
  - `YdbOffsetBackingStore` — Kafka Connect `OffsetBackingStore` persisting
    connector offsets in a YDB table.
  - `YdbSchemaHistory` — `SchemaHistory` implementation persisting DDL history
    in a YDB table.
  - Shared YDB transport/configuration/initialiser code (`YdbTransport`,
    `YdbCommonConfig`, `YdbInitialiser`).
  - Unit tests for config validation and integration tests for both stores,
    with a `docker-compose-ydb.yml` for a local YDB instance.

  The new module is wired into `debezium-storage/pom.xml`.

## PR Checklist


debezium-storage/debezium-storage-ydb/src/main/java/io/debezium/storage/ydb/history/YdbSchemaHistory.java
```diff
@@ -0,0 +1,206 @@
...
+import tech.ydb.table.values.PrimitiveValue;
+
+/**
+ * {@link SchemaHistory} backed by the {@code dbz_schema_history} table. The table and its
+ * parent directories are created on start via {@link YdbInitialiser#createSchemaHistoryTable} (idempotent).
+ * Records are appended with a monotonically growing {@code seq_no} (per connector); recovery
+ * is a single {@code ORDER BY seq_no ASC} scan.
+ *
+ * <p>By design only one source instance writes per connector_name (one PG slot reader),
+ * so the in-memory {@link AtomicLong} sequence is safe.
```
vjuranek: This assumption is not correct, SQL server can have multiple tasks per one connector.

debezium-storage/debezium-storage-ydb/src/test/java/io/debezium/storage/ydb/YdbTestEnvironment.java
```diff
@@ -0,0 +1,98 @@
...
+import org.testcontainers.containers.ComposeContainer;
+
+import tech.ydb.auth.NopAuthProvider;
+import tech.ydb.common.transaction.TxMode;
+import tech.ydb.core.grpc.GrpcTransport;
+import tech.ydb.query.QueryClient;
+import tech.ydb.query.tools.SessionRetryContext;
+
+/**
+ * Boots a local YDB instance via docker-compose for integration tests. The compose file
```
vjuranek: As we aim to manage all out test containers via testcontainers, would it be possible to use testcontainers for it?

vjuranek (commented):
Thanks for the PR @LevinMK23 ! Overall LGTM, just once concert about using Docker compose directly for testing.

vjuranek:
And additional high-level question. YDB provide [JDBC driver](https://github.com/ydb-platform/ydb-jdbc-driver), so why is this store needed? Cannot YDB users use JDBC store to store the offsets/history into YDB?

debezium-storage/debezium-storage-ydb/src/test/java/io/debezium/storage/ydb/YdbTestEnvironment.java
vyrexd: Yes, the required dependencies exist — I'll rewrite the tests on testcontainers.

debezium-storage/debezium-storage-ydb/src/main/java/io/debezium/storage/ydb/history/YdbSchemaHistory.java
vyrexd: For the multi-task case we can partition the keys by database name.

vjuranek: yes, this is how it's typically handled

vyrexd:
> And additional high-level question. YDB provide [JDBC driver](https://github.com/ydb-platform/ydb-jdbc-driver), so why is this store needed? Cannot YDB users use JDBC store to store the offsets/history into YDB?

The main idea was to use transactions across topic and table within a single database. Although in reality this is a prerequisite for another task — the YDB sink. I've 
  now tested the sink with a JDBC store that writes offsets to YDB, and everything works.

vjuranek:
> The main idea was to use transactions across topic and table within a single database.

Just to make sure I understand correctly, so basically you aim for exactly once delivery, when offset is written iff (if and only if) all the records in given batch are written to the target DB as well and vice versa (if offset is not written, TX with batch records is rollback as well). Is this correct?

mfvitale:
I will also add that given the recent introduction of [DS JDBC Sink](https://github.com/debezium/dbz/issues/1842) is this and the [Add YDB sink](https://github.com/debezium/dbz/issues/1927) still relevant?

vyrexd:
> I will also add that given the recent introduction of [DS JDBC Sink](https://github.com/debezium/dbz/issues/1842) is this and the [Add YDB sink](https://github.com/debezium/dbz/issues/1927) still relevant?

Yes, the commit with the ydb sink is almost ready, I'm writing tests.

vyrexd:
> > The main idea was to use transactions across topic and table within a single database.
> 
> Just to make sure I understand correctly, so basically you aim for exactly once delivery, when offset is written iff (if and only if) all the records in given batch are written to the target DB as well and vice versa (if offset is not written, TX with batch records is rollback as well). Is this correct?

Yes, that was the idea and the advantage of YDB in such scenarios; on top of that, our event delivery is built on YDB, and we'd like to build migrations on the same bus
   as well.

debezium-storage/debezium-storage-ydb/src/main/java/io/debezium/storage/ydb/history/YdbSchemaHistoryConfig.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class YdbSchemaHistoryConfig extends YdbCommonConfig {
+
+    public static final Field CONNECTOR_NAME = Field.create("schema.history.internal.ydb.connector.name")
+            .withDescription("Connector identifier used as PK prefix in dbz_schema_history")
+            .required();
+
+    public static final Field TABLE_NAME = Field.create("schema.history.internal.ydb.table.name")
+            .withDescription("YDB table that stores schema history records")
+            .withDefault("debezium/dbz_schema_history");
```
Naros, line 20: These don't follow the convention of the other storage modules, which should start with `ydb.*`.  The `schema.history.internal.` prefix is automatically stripped in the common config class. Ideally we should follow the same conventions here.

debezium-storage/debezium-storage-ydb/src/main/java/io/debezium/storage/ydb/offset/YdbOffsetBackingStoreConfig.java
```diff
@@ -0,0 +1,38 @@
...
+
+public class YdbOffsetBackingStoreConfig extends YdbCommonConfig {
+
+    public static final Field CONNECTOR_NAME = Field.create("offset.storage.ydb.connector.name")
+            .withDescription("Connector identifier used as the PK prefix in dbz_offsets")
+            .required();
+
+    public static final Field TABLE_NAME = Field.create("offset.storage.ydb.table.name")
+            .withDescription("YDB table that stores Connect offsets")
+            .withDefault("debezium/dbz_offsets");
```
Naros, line 20: Much like the schema history configuration, we should follow a similar convention here.  Please see how the JDBC or Redis storage modules handle a prefix subset in the common config class.

debezium-storage/debezium-storage-ydb/src/main/java/io/debezium/storage/ydb/history/YdbSchemaHistory.java
```diff
@@ -0,0 +1,232 @@
...
+    }
+
+    private static String extractDbName(HistoryRecord record) {
+        String db = record.document().getString(HistoryRecord.Fields.DATABASE_NAME);
+        return db == null ? "" : db;
+    }
+
+    @Override
+    public boolean storageExists() {
+        return true;
```
Naros, line 210: This seems unusual to always assume it exists, no?

debezium-storage/debezium-storage-ydb/src/main/java/io/debezium/storage/ydb/offset/YdbOffsetBackingStore.java
```diff
@@ -0,0 +1,208 @@
...
+    private YdbOffsetBackingStoreConfig config;
+    private ExecutorService executor;
+    private GrpcTransport transport;
+    private QueryClient queryClient;
+    private SessionRetryContext retryCtx;
+
+    private String selectAllSql;
+    private String upsertHeaderSql;
+
+    protected ConcurrentHashMap<String, String> data = new ConcurrentHashMap<>();
```
Naros, line 67:
```suggestion
    private final ConcurrentHashMap<String, String> data = new ConcurrentHashMap<>();
```

debezium-storage/debezium-storage-ydb/src/main/java/io/debezium/storage/ydb/history/YdbSchemaHistory.java
```diff
@@ -0,0 +1,232 @@
...
+        }
+        this.transport = YdbTransport.open(config);
+        this.queryClient = QueryClient.newClient(transport).build();
+        this.retryCtx = SessionRetryContext.create(queryClient).build();
+        this.upsertSql = String.format(UPSERT_RECORD_SQL, config.getTableName());
+        this.selectAllSql = String.format(SELECT_ALL_SQL, config.getTableName());
+        this.selectMaxSeqPerDbSql = String.format(SELECT_MAX_SEQ_PER_DB_SQL, config.getTableName());
+        this.selectExistsSql = String.format(SELECT_EXISTS_SQL, config.getTableName());
+
+        YdbInitialiser.createSchemaHistoryTable(transport, queryClient, config.getDatabase(), config.getTableName());
```
Naros, line 119: Could we follow the same convention as the other modules, like JDBC, where start checks whether the storage exists and, if it doesn't, calls `initializeStorage`. This just separates concerns cleanly in case in the future we need to explicitly initialize vs start.

debezium-storage/debezium-storage-ydb/pom.xml
```diff
@@ -0,0 +1,134 @@
...
+                <artifactId>grpc-netty-shaded</artifactId>
+                <version>${version.ydb.grpc}</version>
+            </dependency>
+        </dependencies>
+    </dependencyManagement>
+
+    <properties>
+        <version.ydb.sdk>2.4.3</version.ydb.sdk>
+        <version.ydb.grpc>1.68.3</version.ydb.grpc>
+    </properties>
```
Naros, line 70: Let's move these above the dependency management section and then use `${version.ydb.sdk}` to define the BOM version, so it's defined only once.

Naros (changes requested):
Hi @LevinMK23 a few inline questions/comments.

Naros:
Hi @vyrexd could you take a look at the remaining items as you have time?

PR #7444, opened 2026-05-15
debezium/dbz#1944 Avoid NPE in ApproximateStructSizeCalculator for null source partition or offset

### LINKED ISSUE #1944: [Spanner] NullPointerException in ApproximateStructSizeCalculator when finished-partition cleanup dispatches a null-offset heartbeat (max.queue.size.in.bytes > 0)
## Bug report

Debezium Spanner connector (incubating), `io.debezium.connector.spanner.SpannerConnector`, **v3.5.0.Final**. The same code path also exists in `v2.6.1.Final` (confirmed identical for the relevant files; see *Affected versions* below).

---

The relevant subset (full Avro / SASL / reroute config elided for brevity):

```json
{
  "connector.class": "io.debezium.connector.spanner.SpannerConnector",
  "tasks.max": "1",
  "gcp.spanner.project.id": "<my-project>",
  "gcp.spanner.instance.id": "<my-instance>",
  "gcp.spanner.database.id": "<my-database>",
  "gcp.spanner.change.stream": "<my-change-stream>",
  "connector.spanner.finished.partition.deletion.delay": "3600",

  // ← this is what flips the latent bug from silent to fatal:
  "max.queue.size.in.bytes": "268435456"
}
```

`max.queue.size.in.bytes` was set to `256 MiB` after a wide-table backfill OOM, to back-pressure heap pressure earlier than the default 8192-record count limit. Once the byte limit is on, the latent bug below fires every time the Spanner connector cleans up a finished partition.

---

- Source: Google Cloud Spanner (change streams), regional instance.
- Connect runtime: Strimzi Kafka Operator v0.51.0 / Apache Kafka 4.2.0 on Kubernetes.
- Kafka backend: Confluent Cloud (managed), SASL_SSL / PLAIN.

---

When the Spanner connector clears a finished partition by dispatching a heartbeat `SourceRecord` with `sourceOffset = null` from `RemoveFinishedPartitionOperation.removeFinishedPartitions`, `ChangeEventQueue.doEnqueue` should still be able to enqueue it without aborting the task. Either:

- `ApproximateStructSizeCalculator.getApproximateRecordSize` should handle a null `sourceOffset()` gracefully (count it as 0 bytes), or
- The Spanner connector should not emit a `SourceRecord` whose `sourceOffset()` is null.

The connector should not enter a permanent task-restart loop in either case.

---

**What behavior do you see?**

With `max.queue.size.in.bytes > 0`, the source task throws a NullPointerException the moment the Spanner partition-cleanup heartbeat is enqueued:

```text
java.lang.NullPointerException: Cannot invoke "java.util.Map.size()"
  because the return value of "org.apache.kafka.connect.source.SourceRecord.sourceOffset()" is null
    at io.debezium.util.ApproximateStructSizeCalculator.getApproximateRecordSize(ApproximateStructSizeCalculator.java:30)
    at io.debezium.pipeline.DataChangeEvent.objectSize(DataChangeEvent.java:31)
    at io.debezium.connector.base.ChangeEventQueue.doEnqueue(ChangeEventQueue.java:254)
    at io.debezium.connector.spanner.processor.heartbeat.SpannerHeartbeat.emit(SpannerHeartbeat.java:69)
    at io.debezium.pipeline.EventDispatcher.alwaysDispatchHeartbeatEvent(EventDispatcher.java:448)
    at io.debezium.connector.spanner.task.operation.RemoveFinishedPartitionOperation
        .lambda$removeFinishedPartitions$0(RemoveFinishedPartitionOperation.java:81)
    ...
```

`SpannerErrorHandler.setProducerThrowable` wraps it as a retriable exception, the task restarts, and the next `RemoveFinishedPartitionOperation` tick re-fires the NPE. The connector-level state remains `RUNNING` so external dashboards miss it; per-pod logs show roughly 5–10 NPE-driven task restarts per minute and offsets stop advancing.

### Root cause analysis

The Spanner connector intentionally builds a heartbeat `SourceRecord` with `sourceOffset = null` when removing a finished partition. See `RemoveFinishedPartitionOperation` (v3.5.0.Final and v2.6.1.Final are byte-identical here):

```java
// debezium-connector-spanner / src/main/java/io/debezium/connector/spanner/task/operation/RemoveFinishedPartitionOperation.java
LOGGER.info("Task {}, Dispatching null offset for partition {} because it is removed", ...);
PartitionOffset partitionOffset = new PartitionOffset();   // empty ctor → offset=null
SpannerOffsetContext ctx = new SpannerOffsetContext(partitionOffset, new TransactionContext());
spannerEventDispatcher.alwaysDispatchHeartbeatEvent(partition, ctx);
```

`PartitionOffset.getOffset()` returns `null` (not an empty map) for the empty-ctor case:

```java
// debezium-connector-spanner / .../context/offset/PartitionOffset.java
public Map<String, String> getOffset() {
    if (this.offset == null) {
        return null;            // ← intentional
    }
    return Map.of(...);
}
```

`SpannerHeartbeat.emit` propagates that null straight into a new `SourceRecord`:

```java
// debezium-connector-spanner / .../processor/heartbeat/SpannerHeartbeat.java
this.queue.enqueue(new DataChangeEvent(
    heartbeatRecord((Map<String, String>) partition, offset.getOffset())));   // offset.getOffset() == null

private SourceRecord heartbeatRecord(Map<String, String> sourcePartition, Map<String, ?> sourceOffset) {
    return new SourceRecord(sourcePartition, sourceOffset, topicName, 0, ...);
}
```

Then in Debezium core's size calculator the null is dereferenced unconditionally:

```java
// debezium-core (3.x: debezium-connector-common) / src/main/java/io/debezium/util/ApproximateStructSizeCalculator.java
public static long getApproximateRecordSize(SourceRecord changeEvent) {
    long value = changeEvent.sourcePartition().size() * 100L
               + changeEvent.sourceOffset().size() * 100L        // ← NPE when sourceOffset == null
               + changeEvent.headers().size() * 100L;
    ...
}
```

…and the call site is gated by the byte-limit feature switch:

```java
// debezium-core (3.x: debezium-connector-common) / .../connector/base/ChangeEventQueue.java doEnqueue
queue.add(record);
if (maxQueueSizeInBytes > 0) {            // ← byte-level queue tracking on?
    long messageSize = record.objectSize();   // calls the calculator above
    sizeInBytesQueue.add(messageSize);
    currentQueueSizeInBytes += messageSize;
}
```

So the bug stays latent as long as `max.queue.size.in.bytes = 0` (Debezium default) but lights up the moment byte-level queue accounting is enabled.

### Why only Spanner triggers this

`SourceRecord(... sourceOffset = null ...)` is permitted by the Kafka Connect API. Other Debezium connectors (MySQL, Postgres, MongoDB, Vitess, …) never construct one in practice — their offsets (binlog position, LSN, GTID, resume token) are always populated. The Spanner connector is unique in deliberately using `sourceOffset = null` as a signaling convention for the partition-lifecycle "this partition is being removed" event. Future connectors that adopt a similar partition-lifecycle pattern will hit the same NPE.

### Proposed fix (core; one-line, defensive)

Make `ApproximateStructSizeCalculator.getApproximateRecordSize` null-safe so the Kafka Connect API contract (which permits null offsets) is honoured at the framework layer rather than relying on every connector to never emit one:

```diff
 public static long getApproximateRecordSize(SourceRecord changeEvent) {
-    long value = changeEvent.sourcePartition().size() * 100L
-               + changeEvent.sourceOffset().size() * 100L
-               + changeEvent.headers().size() * 100L;
+    final Map<String, ?> partition = changeEvent.sourcePartition();
+    final Map<String, ?> offset    = changeEvent.sourceOffset();
+    long value = (partition == null ? 0 : partition.size()) * 100L
+               + (offset    == null ? 0 : offset.size())    * 100L
+               + changeEvent.headers().size() * 100L;
     value += 8; // timestamp
     return value + getStructSize((Struct) changeEvent.key()) + getStructSize((Struct) changeEvent.value())
             + changeEvent.topic().getBytes().length;
 }
```

Happy to send a PR with this change plus a unit test covering both `sourcePartition == null` and `sourceOffset == null` cases. The Spanner connector side can stay as-is (the null-offset dispatch is by design).

### Affected versions

The relevant files are byte-identical between Debezium 2.6.1.Final and 3.5.0.Final:

- `debezium-connector-spanner/.../RemoveFinishedPartitionOperation.java` — diffs only in unrelated code
- `debezium-connector-spanner/.../PartitionOffset.java` — identical
- `debezium-connector-spanner/.../SpannerHeartbeat.java` — identical
- `debezium-core` (2.6.1) / `debezium-connector-common` (3.x) `ApproximateStructSizeCalculator.java` — identical
- `ChangeEventQueue.doEnqueue` byte-size gate — identical

So this is **not** a 2.x ↔ 3.x regression; it's a latent bug across the whole 2.x/3.x line. Any 2.x or 3.x Spanner connector configured with `max.queue.size.in.bytes > 0` will hit it the first time a partition reaches the finished-deletion-delay.

---

Yes — the code on `main` and `v3.5.1.Final` is identical to `v3.5.0.Final` for these files. The NPE reproduces on `v3.5.0.Final` today, and we have verified the proposed null-safe patch eliminates it end-to-end on the same image with `max.queue.size.in.bytes = 256 MiB`.

---

Representative segment after the connector reaches `RUNNING` and starts dispatching cleanup heartbeats (NPE count: ~42 in 10 minutes over a ~100 minute observation window):

```text
NullPointerException: Cannot invoke "java.util.Map.size()" because the return value of "SourceRecord.sourceOffset()" is null
    at io.debezium.util.ApproximateStructSizeCalculator.getApproximateRecordSize(ApproximateStructSizeCalculator.java:30)
    at io.debezium.pipeline.DataChangeEvent.objectSize(DataChangeEvent.java:31)
    at io.debezium.connector.base.ChangeEventQueue.doEnqueue(ChangeEventQueue.java:254)
    at io.debezium.connector.spanner.processor.heartbeat.SpannerHeartbeat.emit(SpannerHeartbeat.java:69)
    at io.debezium.pipeline.EventDispatcher.alwaysDispatchHeartbeatEvent(EventDispatcher.java:448)
    at io.debezium.connector.spanner.task.operation.RemoveFinishedPartitionOperation.lambda$removeFinishedPartitions$0(RemoveFinishedPartitionOperation.java:81)
    ...

# SpannerErrorHandler converts to RetriableException, task restarts:
WARN  AbstractWorkerSourceTask: WorkerSourceTask{id=...} failed to poll records from SourceTask. Will retry operation.
org.apache.kafka.connect.errors.RetriableException: An exception occurred in the change event producer. This connector will be restarted.
    at io.debezium.connector.spanner.SpannerErrorHandler.setProducerThrowable(SpannerErrorHandler.java:58)
    ...
Caused by: java.lang.NullPointerException

# Loops at ~5-10 restarts per minute, offsets stop advancing
```

Happy to attach a full DEBUG-level log on request.

---

The tutorial doesn't ship a Spanner setup, but the bug only needs three things, all confirmable from the source:

1. Run any Debezium Spanner connector (2.x or 3.x) with `max.queue.size.in.bytes > 0` in its config.
2. Wait until at least one change-stream partition transitions to FINISHED and the `connector.spanner.finished.partition.deletion.delay` window elapses, so `RemoveFinishedPartitionOperation.removeFinishedPartitions` dispatches the null-offset heartbeat.
3. The first such heartbeat enqueue throws the NPE shown above.

A unit test against `ApproximateStructSizeCalculator.getApproximateRecordSize(new SourceRecord(Map.of(), null, "topic", null, null))` reproduces the underlying defect deterministically without any source DB.

---

### Workaround

Until a null-safe core lands, two viable workarounds, both confirmed:

- Set `max.queue.size.in.bytes = 0` on the Spanner connector (loses byte-level queue back-pressure).
- Build Debezium core with the one-line null-safe fix above (the NPE is gone end-to-end with `max.queue.size.in.bytes = 256 MiB`).

I will open a PR with the fix and a unit test as a follow-up; happy to incorporate any feedback first.

laughingman7743:
Fixes debezium/dbz#1944

## Description

`SourceRecord#sourcePartition()` and `SourceRecord#sourceOffset()` may legitimately return `null`. For example, the Spanner connector's `RemoveFinishedPartitionOperation` enqueues finished-partition heartbeat records with a `null` source offset.

When `max.queue.size.in.bytes > 0`, `ChangeEventQueue` calls `ApproximateStructSizeCalculator.getApproximateRecordSize(record)`, which previously dereferenced both `sourcePartition()` and `sourceOffset()` unconditionally and threw `NullPointerException`. The exception propagated out of the producer thread, causing the connector task to restart repeatedly. Reproduced on 2.6.1 and 3.5.0.Final.

This change treats a `null` source partition or offset as size `0` to match the semantics of an empty map. Only the two affected map lookups are guarded; all other behavior is unchanged.

Added `testGetApproximateRecordSizeWithNullSourcePartitionAndOffset` in `ApproximateStructSizeCalculatorTest`, covering null partition, null offset, and both null. Each case asserts the computed size matches the empty-map baseline.

### AI tool usage disclosure
I diagnosed the NPE and authored the fix and tests myself. I used Claude Code assistively to navigate the codebase and review the patch. All changes were reviewed, built, and tested locally before submission. I take full responsibility for the contents of this PR.

## PR Checklist

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/util/ApproximateStructSizeCalculator.java
+++ b/debezium-connector-common/src/main/java/io/debezium/util/ApproximateStructSizeCalculator.java
@@ -29,6 +29,9 @@ public class ApproximateStructSizeCalculator {
         // assuming 100 bytes per entry of partition / offset / header
         final Map<String, ?> sourcePartition = changeEvent.sourcePartition();
         final Map<String, ?> sourceOffset = changeEvent.sourceOffset();
+        // The Spanner connector's RemoveFinishedPartitionOperation emits finished-partition
+        // heartbeat records with a null source offset, violating Kafka's SourceRecord
+        // contract (non-null maps, possibly empty). Guard against NPE here.
         long value = (sourcePartition == null ? 0 : sourcePartition.size()) * 100L
                 + (sourceOffset == null ? 0 : sourceOffset.size()) * 100L
                 + changeEvent.headers().size() * 100L;
--- a/debezium-connector-common/src/main/java/io/debezium/util/ApproximateStructSizeCalculator.java
+++ b/debezium-connector-common/src/main/java/io/debezium/util/ApproximateStructSizeCalculator.java
@@ -27,7 +27,11 @@ public class ApproximateStructSizeCalculator {
 
     public static long getApproximateRecordSize(SourceRecord changeEvent) {
         // assuming 100 bytes per entry of partition / offset / header
-        long value = changeEvent.sourcePartition().size() * 100L + changeEvent.sourceOffset().size() * 100L + changeEvent.headers().size() * 100L;
+        final Map<String, ?> sourcePartition = changeEvent.sourcePartition();
+        final Map<String, ?> sourceOffset = changeEvent.sourceOffset();
+        long value = (sourcePartition == null ? 0 : sourcePartition.size()) * 100L
+                + (sourceOffset == null ? 0 : sourceOffset.size()) * 100L
+                + changeEvent.headers().size() * 100L;
         value += 8; // timestamp
 
         // key and value, ignoring schemas, assuming they are constant, shared on the heap
--- a/debezium-connector-common/src/test/java/io/debezium/util/ApproximateStructSizeCalculatorTest.java
+++ b/debezium-connector-common/src/test/java/io/debezium/util/ApproximateStructSizeCalculatorTest.java
@@ -39,4 +39,32 @@ public class ApproximateStructSizeCalculatorTest {
         actual = ApproximateStructSizeCalculator.getApproximateRecordSize(sourceRecord);
         assertEquals(actual, 115);
     }
+
+    @Test
+    public void testGetApproximateRecordSizeWithNullSourcePartitionAndOffset() {
+        Schema valueSchema = SchemaBuilder.struct().build();
+
+        // baseline with empty source partition and offset
+        SourceRecord baseline = new SourceRecord(Collections.emptyMap(), Collections.emptyMap(), "dummy",
+                valueSchema, new Struct(valueSchema));
+        long expected = ApproximateStructSizeCalculator.getApproximateRecordSize(baseline);
+
+        // test null source partition
+        SourceRecord withNullPartition = new SourceRecord(null, Collections.emptyMap(), "dummy",
+                valueSchema, new Struct(valueSchema));
+        long actual = ApproximateStructSizeCalculator.getApproximateRecordSize(withNullPartition);
+        assertEquals(actual, expected);
+
+        // test null source offset
+        SourceRecord withNullOffset = new SourceRecord(Collections.emptyMap(), null, "dummy",
+                valueSchema, new Struct(valueSchema));
+        actual = ApproximateStructSizeCalculator.getApproximateRecordSize(withNullOffset);
+        assertEquals(actual, expected);
+
+        // test both null
+        SourceRecord withNulls = new SourceRecord(null, null, "dummy",
+                valueSchema, new Struct(valueSchema));
+        actual = ApproximateStructSizeCalculator.getApproximateRecordSize(withNulls);
+        assertEquals(actual, expected);
+    }
 }


debezium-connector-common/src/main/java/io/debezium/util/ApproximateStructSizeCalculator.java
```diff
@@ -27,7 +27,11 @@ public class ApproximateStructSizeCalculator {
 
     public static long getApproximateRecordSize(SourceRecord changeEvent) {
         // assuming 100 bytes per entry of partition / offset / header
-        long value = changeEvent.sourcePartition().size() * 100L + changeEvent.sourceOffset().size() * 100L + changeEvent.headers().size() * 100L;
+        final Map<String, ?> sourcePartition = changeEvent.sourcePartition();
+        final Map<String, ?> sourceOffset = changeEvent.sourceOffset();
```
Naros, line 31: Hi @laughingman7743, thanks for the change. I believe it might be useful here to document in the code exactly why we need to check for these. The reason is that Kafka expects at least the partition value to be non-null map (the map can be empty), so if Spanner never supplies a partition, it's non-conforming and would break if such an event were serialized through the offset store, etc.

laughingman7743: Thanks @Naros — added an inline comment in 01e0e58 explaining the Spanner-specific case (RemoveFinishedPartitionOperation emitting a null source offset) and a brief note on the Kafka SourceRecord contract. Placed it right above the null-guarded expression. Happy to adjust the wording if anything reads off.

Naros (approved):
LGTM, I'd like to get @jpechane opinion as well.

jpechane (approved):
@laughingman7743 LGTM, thanks! The PR can be merged when CI run is finished.

Naros:
Thanks, applied @laughingman7743

PR #7448, opened 2026-05-15
debezium/dbz#1693 Deprecate `redo_log_catalog` mining strategies

### LINKED ISSUE #1693: Deprecate "redo_log_catalog" Oracle mining strategy
**Maintenance burden**: Supporting multiple mining modes requires every change to be fully tested. This isn't always the case, given the complexities that redo log catalog implies with the volume of archive logs created over the course of the test suite.

**Behavioral differences**: The mode relies on Oracle's redo log for catalog/dictionary information rather than the online catalog. This can lead to subtle differences and nuances that require performing specific tasks, such as a data dictionary build step, at carefully selected points in the mining process. This adds a level of complexity that differs from the `online_catalog` mode, making introducing code changes unnecessarily complex and often completely different because of how the modes operate.

**Performance**: Using the redo log as the catalog source forces Oracle to do additional work to reconstruct the dictionary state from the redo log stream. This can increase the session startup time, memory consumption, and increase the frequency and size of archive logs.

**Testing**: Given that we're already only testing this mode in limited capacity and can't run the entire test suite with the mode turned on, this creates the potential for behavior gaps, undiagnosed regressions, etc.

So, deprecating this in Debezium 3.x and removing this mining mode, most likely in Debezium 4.0, reduces the connector's overall behavior footprint, removes a constant pain point with multi-connector deployment, and eliminates performance, behavior, and testing concerns mentioned above.

Naros:
Fixes debezium/dbz#1693

## Description
Deprecate the `redo_log_catalog` log mining strategy. This mining strategy is scheduled for removal in 3.7.

## PR Checklist

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -162,7 +162,10 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             .withImportance(Importance.HIGH)
             .withGroup(Field.createGroupEntry(Field.Group.CONNECTION_ADVANCED, 8))
             .withValidation(OracleConnectorConfig::validateLogMiningStrategy)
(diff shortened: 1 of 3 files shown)


vjuranek (approved):
LGTM

PR #7452, opened 2026-05-18
debezium/dbz#1164 Add unavailable value placeholder for SQL Server

### LINKED ISSUE #1164: SQL Server varbinary(max), varchar(max), and nvarchar(max) should use unavailable.value.placeholder [DBZ-6884]
Migrated from [DBZ-6884](https://issues.redhat.com/browse/DBZ-6884)

When an UPDATE is applied on a table that has a {{VARBINARY(MAX)}}, {{VARCHAR(MAX)}}, or {{NVARCHAR(MAX)}} data type, the column will be assigned NULL in the capture table.  

We should treat these column mappings identical to PostgreSQL TOAST columns or Oracle's BLOB/CLOB columns where if the value isn't being set, we should perhaps expose these using the {{unavailable.value.placeholder}} concept.

Open question is whether we can determine if the column being NULL in the capture table is actually distinguishable from the user explicitly setting the column as NULL vs it being NULL due to the fact the column wasn't changed.

https://debezium.zulipchat.com/#narrow/stream/348252-community-sqlserver/topic/error.20varchar.28max.29/near/390307416

Naros:
Fixes debezium/dbz#1164

## Description
Adds `unavailable.value.placeholder` support for SQL Server's `varchar(max)`, `nvarchar(max)`, and `varbinary(max)` columns that are treated as unchanged during UPDATE operations, matching the existing behavior used by PostgreSQL TOAST and Oracle LOB column types.

## PR Checklist

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerChangeTablePointer.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerChangeTablePointer.java
@@ -43,6 +43,7 @@ public class SqlServerChangeTablePointer extends ChangeTableResultSet<SqlServerC
     private static final int COL_COMMIT_LSN = 1;
     private static final int COL_ROW_LSN = 2;
     private static final int COL_OPERATION = 3;
+    private static final int COL_UPDATE_MASK = 4;
     private static final int COL_DATA = 5;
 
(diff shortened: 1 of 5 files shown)


vjuranek (approved):
LGTM, let's wait for CI.

vjuranek:
Merged, thanks @Naros !

PR #7459, opened 2026-05-19
debezium/dbz#1879 Add SMT EnforceRecordSize

### LINKED ISSUE #1879: Auto truncate columns to adhere to a max message size

We spend a decent bit of toil resizing and updating column truncation settings to ensure our kafka max message size is not breached. We have some tooling around auto detecting and validating this, but it would be ideal if Debezium had a config for a max message size or max length that it could use to calculate truncation thresholds and dynamically apply across all columns (eg as schema evolutions occur). This way it's fully handled by Debezium. Although not all users run kafka, there is often some restriction for output message size in bytes on whichever platform is used.

---

I am drafting a PR in the debezium vitess connector. If we believe it is useful enough for core I am happy to move it there.

twthorn:
Fixes debezium/dbz#1879

## Description

https://github.com/debezium/dbz/issues/1879

Add an SMT for enforcing record size with the current strategy of proportional column truncation. Use min field size to avoid truncating smaller columns that may lose meaning with truncation. Use shared helper method to estimate source record size.

We may support other strategies in the future, one such is payload offloading https://github.com/debezium/dbz/issues/1949

## PR Checklist

### DIFF
--- a/debezium-connect-plugins/src/main/java/io/debezium/transforms/EnforceRecordSize.java
+++ b/debezium-connect-plugins/src/main/java/io/debezium/transforms/EnforceRecordSize.java
@@ -7,7 +7,6 @@ package io.debezium.transforms;
 
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.List;
 import java.util.Map;
 
@@ -16,13 +15,14 @@ import org.apache.kafka.common.config.ConfigDef;
 import org.apache.kafka.common.config.ConfigException;
 import org.apache.kafka.connect.components.Versioned;
 import org.apache.kafka.connect.connector.ConnectRecord;
-import org.apache.kafka.connect.data.Field;
 import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.source.SourceRecord;
 import org.apache.kafka.connect.transforms.Transformation;
 
 import io.debezium.Module;
+import io.debezium.config.Field;
+import io.debezium.metadata.ConfigDescriptor;
 import io.debezium.util.ApproximateStructSizeCalculator;
 
 /**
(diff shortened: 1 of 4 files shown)


debezium-connect-plugins/src/main/java/io/debezium/transforms/EnforceRecordSize.java
```diff
@@ -0,0 +1,248 @@
...
+            truncateField(dataStruct, field, newSizeBytes);
+        }
+    }
+
+    private static void truncateField(Struct dataStruct, TruncatableField field, int newSizeBytes) {
+        if (field.value instanceof String) {
+            String original = (String) field.value;
+            if (original.length() > newSizeBytes) {
+                dataStruct.put(field.fieldName, original.substring(0, newSizeBytes));
+            }
```
Naros, line 189: Here, we are comparing the length of a `String` (in characters) to a byte size limit. For multi-byte UTF-8 characters, this creates a semantic mismatch. Should we care about this?

debezium-connect-plugins/src/main/java/io/debezium/transforms/EnforceRecordSize.java
```diff
@@ -0,0 +1,248 @@
...
+        }
+        else if (field.value instanceof ByteBuffer) {
+            ByteBuffer original = (ByteBuffer) field.value;
+            if (original.limit() > newSizeBytes) {
+                dataStruct.put(field.fieldName, ByteBuffer.wrap(Arrays.copyOfRange(original.array(), 0, newSizeBytes)));
```
Naros: If the `ByteBuffer` is a direct `ByteBuffer`, the call to `array()` will throw an unsupported exception.  Should we use a similar technique to `ApproximateStructSizeCalculator#toArray`?

debezium-connect-plugins/src/main/java/io/debezium/transforms/EnforceRecordSize.java
```diff
@@ -0,0 +1,248 @@
...
+        long currentSize = (long) Math.ceil(rawEstimate * compressionRatio);
+        if (currentSize <= maxBytes) {
+            return record;
+        }
+
+        Struct value = (Struct) record.value();
+        long excess = currentSize - maxBytes;
+
+        truncateStructFields(value, "before", excess);
+        truncateStructFields(value, "after", excess);
```
Naros: This results in an over-truncation for update events.

For updates, should we consider halving the `excess` so that if the excess is 900, we apply 450 to each of the two blocks rather than 900 to each, resulting in 1800 bytes truncated?

If we are concerned about disproportionality between `before` and `after` (for example, where an LOB column is populated in `after` but not in the `before`), perhaps a recalculation makes sense in the case of updates?

debezium-connect-plugins/src/main/java/io/debezium/transforms/EnforceRecordSize.java
```diff
@@ -0,0 +1,248 @@
...
+ * Supported strategies:
+ * <ul>
+ *   <li>Proportional column truncation: truncates string/bytes columns proportionally
+ *       (larger columns are truncated more). Columns at or below the configured
+ *       minimum field size are excluded from truncation.</li>
+ * </ul>
+ *
+ * @author Thomas Thornton
+ */
+public class EnforceRecordSize<R extends ConnectRecord<R>> implements Transformation<R>, Versioned {
```
mfvitale: Could you please implement the `io.debezium.metadata.ConfigDescriptor` interface? Also register it in io.debezium.transforms.metadata.TransformsMetadataProvider#getConnectorMetadata? Thanks

debezium-connect-plugins/src/main/java/io/debezium/transforms/EnforceRecordSize.java
twthorn: I did this so we have constant time, even though the size may differ.

A future version could add a config for getting exact size with getBytes.

debezium-connect-plugins/src/main/java/io/debezium/transforms/EnforceRecordSize.java
twthorn: Nice catch, updated to use a similar approach

debezium-connect-plugins/src/main/java/io/debezium/transforms/EnforceRecordSize.java
twthorn: Good point, for this first strategy we apply proportionally to them based on their size

debezium-connect-plugins/src/main/java/io/debezium/transforms/EnforceRecordSize.java
twthorn: Yes, updated, thanks

twthorn:
The tests appear flakey and unrelated. Requesting a re-review.

Naros (approved):

jpechane:
@twthorn Applied, thanks! Could you please send a follow-up docs PR?

twthorn:
@jpechane here is the docs PR https://github.com/debezium/debezium/pull/7478

PR #7461, opened 2026-05-20
DBZ-9755: Detect and fail on stale MySQL binlog connections

### LINKED ISSUE #1474: Restart the connector or Throw an exception if the TCP connection remains ESTABLISHED, but the MySQL session is DEAD, leaving the connection in a stale state indefinitely
Migrated from [DBZ-9755)](https://issues.redhat.com/browse/DBZ-9755)

We suspect this cause of the issue. The TCP connection remains {*}ESTABLISHED{*}, but the MySQL session is {*}DEAD{*}, leaving the connection in a stale state indefinitely.
In this scenario, the connector still shows as {*}RUNNING{*}, which is the problem. 
*Possible causes:* * MySQL process termination
 * Network layer issues
 * MySQL server-side timeouts:
 ** Connection limits reached
 ** Idle connection timeout (MySQL closes inactive sessions)

Debezium Connector → TCP Layer → MySQL Application Layer
1. TCP Handshake (Layer 4)
   ├── SYN packet sent
   ├── SYN-ACK received
   └── ACK sent → TCP ESTABLISHED
2. MySQL Authentication (Layer 7)
   ├── MySQL handshake packet
   ├── Username/password exchange
   └── Authentication OK → MySQL Session Active
3. Binlog Streaming (Layer 7)
   ├── SHOW MASTER STATUS
   ├── COM_BINLOG_DUMP command
   └── Binlog events flowing → CDC Working

{*}One of the Problem Scenario{*}:Debezium Connector
       ↓
TCP Connection: ESTABLISHED   !https://a.slack-edge.com/production-standard-emoji-assets/14.0/google-medium/2705@2x.png!   ← Shows "Connected"
       ↓
MySQL Session: DEAD   !https://a.slack-edge.com/production-standard-emoji-assets/14.0/google-medium/274c@2x.png!          ← Actually disconnected
       ↓
Result: Connector thinks it's working but no data flows 
 * TCP (Layer 4) only knows about network connectivity
 * MySQL (Layer 7) handles actual database sessions

 
h3. Replicating the Issue

To reproduce the issue, I am running a MySQL database in a local Docker container and have completed the local setup to run a MySQL Debezium source Kafka connector (version {*}3.3.2.Final{*}).

I tested the end-to-end flow successfully. When I make changes to a database table, the correct events are pushed to the Kafka topic. I have also configured a heartbeat with a {*}30-second interval{*}, and the heartbeat topic is receiving heartbeat events as expected. Also i configured the heartbeat.action.query to insert current timestamp.

When I pause the MySQL Docker container, the connector does not fail. Instead, the keepalive thread continuously retries the database connection, which can be observed in the Kafka Connect cluster logs. However, the connector task itself does not log any errors or exceptions. Also heartbeat events is stopped.

When I unpause the MySQL Docker container, the keepalive thread restores the connection successfully. This reconnection works correctly in my {*}local PC environment{*}.

However, when the database and the connector are running on {*}different servers{*}, the connection is not restored successfully even when the db is active and able to make new connection. In this scenario, when we restart the connector, it re-establishes the connection and events start flowing again.

The expectation is that, at the connector level, there should be a mechanism to re-establish the connection using the keepalive thread, automatically restart the connector, retry the connection, or at least fail the connector explicitly.

abothli:
## Problem

When a TCP connection remains ESTABLISHED but the MySQL session is dead (e.g., due to AWS NAT Gateway idle timeout of 350 seconds silently dropping the connection mapping), the binlog client's read thread blocks indefinitely on `socket.read()`. The task remains in RUNNING state with no error logs, but no data flows — a "silent disconnect."

This was observed in production with 17/19 connectors to the same RDS instance simultaneously entering this state. The only indicator was `Connected=0` in JMX metrics, while task status remained RUNNING.

## Root Cause

The main streaming loop in `BinlogStreamingChangeEventSource` only checks `context.isRunning()` without any connection health validation. When the TCP connection enters a half-dead state (ESTABLISHED but no data flowing), neither the keepalive thread nor the IO thread raises an exception, so the loop continues indefinitely.

## Fix

Added stale connection detection to the main streaming loop. When `keepAlive` is enabled and no binlog events (including MySQL heartbeats) have been received for `3 × keepAliveInterval`, the task is failed with a descriptive error. This allows Kafka Connect's restart policy to re-establish a fresh connection.

The detection uses the existing `MilliSecondsSinceLastEvent` metric from `BinaryLogClientStatistics`.

## Prerequisites

- `connect.keep.alive=true` (default)
- `heartbeat.interval.ms` configured (recommended)

Without heartbeat, idle databases with no writes could trigger false positives.

## Testing

Reproduced by pausing a MySQL Docker container on a remote server. Without the fix, the connector stays RUNNING indefinitely. With the fix, the task fails after `3 × keepAliveInterval` and Kafka Connect restarts it.

Fixes: https://github.com/debezium/dbz/issues/1474

### DIFF
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
@@ -108,6 +108,7 @@ public abstract class BinlogStreamingChangeEventSource<P extends BinlogPartition
     private static final String KEEPALIVE_THREAD_NAME = "blc-keepalive";
     private static final String SET_STATEMENT_REGEX = "SET STATEMENT .* FOR";
     private static final Pattern TRUNCATE_STATEMENT_PATTERN = Pattern.compile("(SET STATEMENT .*)?TRUNCATE TABLE .*", Pattern.CASE_INSENSITIVE | Pattern.DOTALL);
+    private static final int STALE_CONNECTION_CHECK_MULTIPLIER = 3;
 
     private final BinaryLogClient client;
     private final BinlogStreamingChangeEventSourceMetrics<?, P> metrics;
@@ -341,6 +342,45 @@ public abstract class BinlogStreamingChangeEventSource<P extends BinlogPartition
             while (context.isRunning()) {
                 Thread.sleep(100);
                 waitWhenStreamingPaused(context);
+
+                // DBZ-9755: Detect stale/dead connections that don't cause the task to fail.
+                if (client.isKeepAlive()) {
+                    long keepAliveInterval = connectorConfig.getConfig().getLong(BinlogConnectorConfig.KEEP_ALIVE_INTERVAL_MS);
+                    long staleThreshold = keepAliveInterval * STALE_CONNECTION_CHECK_MULTIPLIER;
+
+                    // Case 1: Client reports disconnected but task is still running.
+                    // This happens when the keepalive thread detects a dead connection but
+                    // fails to reconnect, leaving the task in a zombie RUNNING state.
+                    if (!client.isConnected()) {
+                        long millisSinceLastEvent = metrics.getMilliSecondsSinceLastEvent();
+                        if (millisSinceLastEvent > staleThreshold && millisSinceLastEvent > 0) {
+                            LOGGER.warn("BinaryLogClient is disconnected and no events received for {} ms. "
+                                    + "Triggering task failure to force reconnection (DBZ-9755).",
+                                    millisSinceLastEvent);
+                            errorHandler.setProducerThrowable(new DebeziumException(
+                                    String.format("Binlog client disconnected: no events for %d ms. "
+                                            + "See https://github.com/debezium/dbz/issues/1474",
+                                            millisSinceLastEvent)));
+                            break;
+                        }
+                    }
+                    // Case 2: Client reports connected but connection is actually stale.
+                    // TCP remains ESTABLISHED but MySQL session is dead (half-open connection).
+                    else {
+                        long millisSinceLastEvent = metrics.getMilliSecondsSinceLastEvent();
+                        if (millisSinceLastEvent > staleThreshold && millisSinceLastEvent > 0) {
+                            LOGGER.warn("No binlog events received for {} ms (threshold: {} ms). "
+                                    + "Connection may be stale (DBZ-9755). Triggering task failure to force reconnection.",
+                                    millisSinceLastEvent, staleThreshold);
+                            errorHandler.setProducerThrowable(new DebeziumException(
+                                    String.format("Stale binlog connection detected: no events received for %d ms "
+                                            + "(keepAliveInterval=%d ms, threshold=%d ms). "
+                                            + "See https://github.com/debezium/dbz/issues/1474",
+                                            millisSinceLastEvent, keepAliveInterval, staleThreshold)));
+                            break;
+                        }
+                    }
+                }
             }
         }
         finally {


vjuranek (commented):
I'm wondering if we should fix if on the BinLogClient and throw proper exception from there.

abothli:
Thanks for the review @vjuranek!

I agree that fixing this at the BinLogClient level would be more robust — ideally the client should detect the stale connection and throw an IOException, which would then trigger `onCommunicationFailure` naturally.

However, the BinLogClient is a separate library (mysql-binlog-connector-java by shyiko/osheroff), and changes there would require a dependency update. This PR provides an immediate mitigation at the Debezium layer that works with the current BinLogClient behavior.

If you'd prefer the fix to be in BinLogClient, I'm happy to:
1. Open an issue/PR on the mysql-binlog-connector-java repo to add SO_TIMEOUT support
2. Keep this PR as a fallback until that upstream fix is available

What approach would you prefer?

abothli:
> I'm wondering if we should fix if on the BinLogClient and throw proper exception from there.

Thanks for the review @vjuranek!

I agree that fixing this at the BinLogClient level would be more robust — ideally the client should detect the stale connection and throw an IOException, which would then trigger `onCommunicationFailure` naturally.

However, the BinLogClient is a separate library (mysql-binlog-connector-java), and changes there would require an upstream PR + dependency update. This PR provides an immediate mitigation at the Debezium layer that works with the current BinLogClient behavior.

I'm happy to also open an issue on the mysql-binlog-connector-java repo to add SO_TIMEOUT support as the proper long-term fix. In the meantime, this detection serves as a defensive fallback — even if the BinLogClient is fixed later, having an application-level check doesn't hurt.

What do you think?

Naros:
> I'm wondering if we should fix if on the BinLogClient and throw proper exception from there.

I concur here. My concern with the approach in the PR is that we just blindly assume that just because there has been no activity in the last 3 minutes, things have gone awry. Aren't there potentially corner cases, like severely low activity or a bulk operation, that might take some time before the events are serialized to the binlog and available for the client to stream? I think we should aim to avoid false positives as much as possible.

abothli:
> have

@Naros Good point. However, this detection only triggers when `heartbeat.interval.ms` is configured (which is the recommended setup). MySQL's binlog heartbeat events are sent by the server independently of transaction activity — even during large bulk operations or DDL, the server continues to send heartbeat events at the configured interval.

So the scenario is:
- With heartbeat configured (e.g., 15s): the timer resets every 15s regardless of transaction activity. Only a truly dead connection would cause 180s of silence.
- Without heartbeat: if the source database has zero writes for 3+ minutes, this could theoretically false-positive. But in practice, databases with no activity for 3 minutes are rare in production, and even if triggered, the restart is harmless (no data loss).

Would it help to add a condition that only enables this check when `heartbeat.interval.ms` is configured? That would eliminate the false-positive concern entirely.

abothli:
Thanks for the review @vjuranek @Naros!

Production Validation

I've deployed and validated this fix in production (107 connectors, Aurora MySQL 8.0.42, Debezium 2.7.3). The fix correctly detected a stale connection and failed the task after 181 seconds:

[2026-05-23 18:33:05,698] WARN BinaryLogClient is disconnected and no events received 
for 181000 ms. Triggering task failure to force reconnection (DBZ-9755).

io.debezium.DebeziumException: Binlog client disconnected: no events for 181000 ms. 
See https://github.com/debezium/dbz/issues/1474
Before this fix, the task would remain in RUNNING state indefinitely with no data flowing and no error logs. Now it fails explicitly, allowing Kafka Connect's restart mechanism (or external tooling) to recover it.

Root Cause Finding

During investigation, I discovered that in our environment the disconnections were triggered by MySQL's net_write_timeout=60s — when the binlog consumer momentarily falls behind (e.g., Kafka Producer backpressure, GC pause), MySQL aborts the connection after 60s of write failure. The Debezium side never receives the RST/FIN packet (TCP half-dead state), so the task stays in a zombie RUNNING state. Databases with net_write_timeout=28800 never experienced this issue.

Regarding False Positives (@Naros)

This detection relies on heartbeat.interval.ms being configured (which is the recommended production setup). MySQL's binlog heartbeat events are sent by the server independently of transaction activity — even during large bulk operations, DDL, or periods of zero writes, the server continues sending heartbeat events at the configured interval.

With heartbeat.interval.ms=15000 (our config), the MilliSecondsSinceLastEvent timer resets every 15 seconds under normal operation. The threshold is 3 × keepAliveInterval = 180s. So only a truly dead/disconnected connection would trigger this — not low activity or bulk operations.

Without heartbeat configured, a database with zero writes for 3+ minutes could theoretically false-positive. But this is an edge case, and even if triggered, the restart is harmless (no data loss, at-least-once semantics preserved).

Updated Implementation

I've updated the PR to cover both disconnection scenarios observed in production:

Case 1: client.isConnected() == false — The keepalive thread detected the dead connection but reconnection keeps failing. Task remains RUNNING in a zombie state.
Case 2: client.isConnected() == true — TCP connection appears alive but MySQL session is dead (half-open connection). No events arrive.
Regarding BinLogClient-level fix

I agree that fixing this at the BinLogClient level (SO_TIMEOUT on the socket) would be more robust as a long-term solution. I'm happy to also open a PR on the mysql-binlog-connector-java repo for that. In the meantime, this application-level detection serves as a defensive fallback — even if the BinLogClient is fixed later, having this check doesn't hurt and provides an additional safety net.

vjuranek:
Thanks for your effort @abothli. Personally, I'd prefer the fix in mysql-binlog-connector-java (we maintain our own fork [here](https://github.com/debezium/mysql-binlog-connector-java/) so there shouldn't be any delays in releasing new version and using it in Debezium). Current implementation replies on bilog client to create and maintain the connection and if possible I'd like to keep this separation of responsibilities - if the connection is closed by mysql, binlog client should detect it and recreate, not the connector itself.

Naros:
> During investigation, I discovered that in our environment the disconnections were triggered by MySQL's net_write_timeout=60s — when the binlog consumer momentarily falls behind (e.g., Kafka Producer backpressure, GC pause), MySQL aborts the connection after 60s of write failure. The Debezium side never receives the RST/FIN packet (TCP half-dead state), so the task stays in a zombie RUNNING state. Databases with net_write_timeout=28800 never experienced this issue.

Hi @abothli, this is actually a very nice find and probably deserves a separate GitHub issue to document this in either the FAQ or [here](https://debezium.io/documentation/reference/3.6/connectors/mysql.html#mysql-when-things-go-wrong) in the MySQL docs.

That said, I also concur with @vjuranek. Since we maintain our own fork of the client, making a change there or in the connector will happen with the same level of speed.

Naros:
Superseded by https://github.com/debezium/debezium/pull/7897, attribution retained.

PR #7462, opened 2026-05-21
DBZ-1683: Fix PostgreSQL connector slow startup with many custom types

Vinu2111:
Fixes https://github.com/debezium/dbz/issues/2041

## Problem

The PostgreSQL connector could take 30+ minutes to start on databases with large numbers of custom types (e.g. databases migrated from Oracle using AWS Schema Conversion Tool, which can generate 30,000+ custom types).

## Root Cause

`prime()` uses a two-pass approach: base types are registered immediately, and types with dependencies (array types, domains) are deferred to a second pass. However, when a deferred type A's dependency is another deferred type B that hasn't been built yet, `builder.build()` calls `typeRegistry.get(oid)` which falls through to `resolveUnknownType(int)` and fires an individual `SQL_OID_LOOKUP` query per unresolved type.

With ~30,000 custom types and 20ms network latency, this produces thousands of sequential round trips at connector startup. The underlying trigger is PostgreSQL's query planner choosing a MERGE JOIN for `SQL_TYPES`, which returns results where array types arrive before their base types.

## Fix

**1. Add ORDER BY to SQL_TYPES** — sorts base types first so most dependencies are already registered when their dependents are processed, eliminating `SQL_OID_LOOKUP` fallbacks for the common case regardless of query plan.

2. Iterative resolution in prime() — replaces the single delayed pass with an iterative loop using a progress flag. Each pass builds types whose dependencies are already registered; the loop exits when no further progress is made or all types are resolved. The dependency graph is a DAG (PostgreSQL enforces acyclicity for both typelem and typbasetype), so this always converges.

**3. Batch fallback in resolveRemainingInBatch()** — any types still unresolved after iterative passes have their missing dependency OIDs fetched in a single `WHERE oid = ANY(?)` query instead of N individual round trips.

## Additional Changes

- Adds `getElementTypeOid()` and `getParentTypeOid()` accessors to `PostgresType.Builder` to support dependency readiness checks in `canResolveNow()` without triggering `builder.build()` prematurely.
- Adds `TypeRegistryTest` — the first dedicated unit test coverage for `TypeRegistry` — verifying the `ORDER BY` clause and Builder getter behaviour.

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
@@ -14,6 +14,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
@@ -70,15 +71,19 @@ public class TypeRegistry {
     private static final String SQL_ENUM_VALUES = "SELECT t.enumtypid as id, array_agg(t.enumlabel ORDER BY t.enumsortorder) as values "
             + "FROM pg_catalog.pg_enum t GROUP BY id";
 
-    private static final String SQL_TYPES = "SELECT t.oid AS oid, t.typname AS name, n.nspname AS schema_name, t.typelem AS element, t.typbasetype AS parentoid, t.typtypmod as modifiers, t.typcategory as category, e.values as enum_values "
+    public static final String SQL_TYPES_BASE = "SELECT t.oid AS oid, t.typname AS name, t.typbasetype AS parentoid, t.typtypmod AS modifiers, t.typelem AS elementoid, n.nspname AS schema_name, t.typtype AS type, t.typcategory AS category, e.values AS enum_values "
             + "FROM pg_catalog.pg_type t "
             + "JOIN pg_catalog.pg_namespace n ON (t.typnamespace = n.oid) "
             + "LEFT JOIN (" + SQL_ENUM_VALUES + ") e ON (t.oid = e.id) "
             + "WHERE n.nspname != 'pg_toast'";
 
-    private static final String SQL_NAME_LOOKUP = SQL_TYPES + " AND t.typname = ?";
+    // SQL_TYPES_BASE with ORDER BY applied so dependent types are
+    // resolved after their dependencies during prime()
+    static final String SQL_TYPES = SQL_TYPES_BASE + " ORDER BY (t.typelem != 0 OR t.typbasetype != 0), t.oid";
 
-    private static final String SQL_OID_LOOKUP = SQL_TYPES + " AND t.oid = ?";
+    private static final String SQL_NAME_LOOKUP = SQL_TYPES_BASE + " AND t.typname = ?";
+
+    private static final String SQL_OID_LOOKUP = SQL_TYPES_BASE + " AND t.oid = ?";
 
     private static final Map<String, String> LONG_TYPE_NAMES = Collections.unmodifiableMap(getLongTypeNames());
 
@@ -476,15 +481,69 @@ public class TypeRegistry {
                 }
 
                 // For types with base or element type mappings, they need to be delayed.
-                // Otherwise their base/element types has not yet be registered,
-                // which triggers additional SQL_OID_LOOKUP queries to PostgreSQL.
                 delayResolvedBuilders.add(builderWithSchema);
             }
 
-            // Resolve delayed builders
-            for (TypeBuilderWithSchema builderWithSchema : delayResolvedBuilders) {
-                addType(builderWithSchema.builder().build(), builderWithSchema.schemaName());
+            // Resolve delayed builders iteratively until no progress is made or all are resolved
+            boolean progress = true;
+            while (!delayResolvedBuilders.isEmpty() && progress) {
+                progress = false;
+                final List<TypeBuilderWithSchema> resolvedInThisPass = new ArrayList<>();
+                for (TypeBuilderWithSchema builderWithSchema : delayResolvedBuilders) {
+                    if (canResolveNow(builderWithSchema.builder())) {
+                        addType(builderWithSchema.builder().build(), builderWithSchema.schemaName());
+                        resolvedInThisPass.add(builderWithSchema);
+                        progress = true;
+                    }
+                }
+                delayResolvedBuilders.removeAll(resolvedInThisPass);
             }
+
+            if (!delayResolvedBuilders.isEmpty()) {
+                resolveRemainingInBatch(delayResolvedBuilders);
+            }
+        }
+    }
+
+    private boolean canResolveNow(PostgresType.Builder builder) {
+        if (builder.hasElementType() && oidToType.get(builder.getElementTypeOid()) == null) {
+            return false;
+        }
+        if (builder.hasParentType() && oidToType.get(builder.getParentTypeOid()) == null) {
+            return false;
+        }
+        return true;
+    }
+
+    private void resolveRemainingInBatch(List<TypeBuilderWithSchema> remaining) throws SQLException {
+        final Set<Integer> missingOids = new LinkedHashSet<>();
+        for (TypeBuilderWithSchema builderWithSchema : remaining) {
+            PostgresType.Builder builder = builderWithSchema.builder();
+            if (builder.hasElementType() && oidToType.get(builder.getElementTypeOid()) == null) {
+                missingOids.add(builder.getElementTypeOid());
+            }
+            if (builder.hasParentType() && oidToType.get(builder.getParentTypeOid()) == null) {
+                missingOids.add(builder.getParentTypeOid());
+            }
+        }
+
+        if (!missingOids.isEmpty()) {
+            String batchSql = SQL_TYPES_BASE + " AND t.oid = ANY(?)";
+            try (PreparedStatement statement = connection.connection().prepareStatement(batchSql)) {
+                Integer[] oidsArray = missingOids.toArray(new Integer[0]);
+                statement.setArray(1, connection.connection().createArrayOf("integer", oidsArray));
+                try (ResultSet rs = statement.executeQuery()) {
+                    while (rs.next()) {
+                        TypeBuilderWithSchema builderWithSchema = createTypeBuilderFromResultSet(rs);
+                        addType(builderWithSchema.builder().build(), builderWithSchema.schemaName());
+                    }
+                }
+            }
+        }
+
+        // Now build the originally delayed types with their dependencies in place.
+        for (TypeBuilderWithSchema builderWithSchema : remaining) {
+            addType(builderWithSchema.builder().build(), builderWithSchema.schemaName());
         }
     }
 
@@ -514,7 +573,7 @@ public class TypeRegistry {
             builder = builder.enumValues(Arrays.asList(enumValues));
         }
         else if (CATEGORY_ARRAY.equals(category)) {
-            builder = builder.elementType((int) rs.getLong("element"));
+            builder = builder.elementType((int) rs.getLong("elementoid"));
         }
         return new TypeBuilderWithSchema(builder.parentType(parentTypeOid), schemaName);
     }
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TypeRegistryIT.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TypeRegistryIT.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.postgresql;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.sql.SQLException;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.kafka.connect.data.Struct;
+import org.apache.kafka.connect.source.SourceRecord;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import io.debezium.connector.postgresql.PostgresConnectorConfig.SnapshotMode;
+import io.debezium.doc.FixFor;
+
+/**
+ * Integration test verifying that the TypeRegistry initialises correctly when
+ * dependent custom types are present, without triggering per-OID SQL_OID_LOOKUP
+ * round trips for each unresolved dependency (fix for DBZ-1683).
+ */
+public class TypeRegistryIT extends AbstractRecordsProducerTest {
+
+    @BeforeEach
+    void before() throws SQLException {
+        TestHelper.dropAllSchemas();
+        TestHelper.execute("CREATE SCHEMA typeregistry;");
+        // Chain: numeric82 <- numericex (domain of domain)
+        TestHelper.execute("CREATE DOMAIN typeregistry.numeric82 AS numeric(8,2);");
+        TestHelper.execute("CREATE DOMAIN typeregistry.numericex AS typeregistry.numeric82;");
+        // Array of a custom domain type
+        TestHelper.execute("CREATE DOMAIN typeregistry.posint AS integer CHECK (VALUE > 0);");
+        TestHelper.execute("CREATE TABLE typeregistry.t1 ("
+                + "id serial primary key, "
+                + "amount typeregistry.numericex, "
+                + "quantity typeregistry.posint"
+                + ");");
+    }
+
+    @Test
+    @FixFor("DBZ-2041")
+    public void shouldStartConnectorWithDependentDomainTypes() throws Exception {
+        // The connector must start successfully without timeouts caused by
+        // per-OID SQL_OID_LOOKUP queries for each unresolved dependent type.
+        start(PostgresConnector.class, TestHelper.defaultConfig()
+                .with(PostgresConnectorConfig.SNAPSHOT_MODE, SnapshotMode.NO_DATA)
+                .with(PostgresConnectorConfig.SCHEMA_INCLUDE_LIST, "typeregistry")
+                .build());
+        assertConnectorIsRunning();
+
+        TestHelper.execute(
+                "INSERT INTO typeregistry.t1 (amount, quantity) VALUES (123.45, 10);");
+
+        final TestConsumer consumer = testConsumer(1, "typeregistry");
+        consumer.await(TestHelper.waitTimeForRecords() * 30, TimeUnit.SECONDS);
+
+        SourceRecord record = consumer.remove();
+        Struct after = ((Struct) record.value()).getStruct("after");
+        assertThat(after.get("id")).isEqualTo(1);
+        assertThat(after.get("quantity")).isEqualTo(10);
+        assertThat(consumer.isEmpty()).isTrue();
+    }
+
+    @Test
+    @FixFor("DBZ-2041")
+    public void shouldResolveChainedDomainTypesWithoutOidLookupFallback() throws Exception {
+        // Verifies that a domain built on top of another domain resolves correctly
+        // through the iterative prime() loop without falling back to SQL_OID_LOOKUP.
+        start(PostgresConnector.class, TestHelper.defaultConfig()
+                .with(PostgresConnectorConfig.SNAPSHOT_MODE, SnapshotMode.NO_DATA)
+                .with(PostgresConnectorConfig.SCHEMA_INCLUDE_LIST, "typeregistry")
+                .with(PostgresConnectorConfig.INCLUDE_UNKNOWN_DATATYPES, true)
+                .build());
+        assertConnectorIsRunning();
+
+        TestHelper.execute(
+                "INSERT INTO typeregistry.t1 (amount, quantity) VALUES (99.99, 5);");
+
+        final TestConsumer consumer = testConsumer(1, "typeregistry");
+        consumer.await(TestHelper.waitTimeForRecords() * 30, TimeUnit.SECONDS);
+
+        SourceRecord record = consumer.remove();
+        Struct after = ((Struct) record.value()).getStruct("after");
+        assertThat(after.get("quantity")).isEqualTo(5);
+        assertThat(consumer.isEmpty()).isTrue();
+    }
+}
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresType.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresType.java
@@ -292,6 +292,14 @@ public class PostgresType {
             return this.elementTypeOid != 0;
         }
 
+        public int getElementTypeOid() {
+            return this.elementTypeOid;
+        }
+
+        public int getParentTypeOid() {
+            return this.parentTypeOid;
+        }
+
         public Builder enumValues(List<String> enumValues) {
             this.enumValues = enumValues;
             return this;
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TypeRegistryTest.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TypeRegistryTest.java
@@ -1,0 +1,76 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.postgresql;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import org.junit.jupiter.api.Test;
+
+/**
+ * Unit tests for {@link TypeRegistry} logic that does not require a live database connection.
+ *
+ * Covers the fixes introduced for dbz#1683: SQL_TYPES ordering and iterative
+ * delayed-type resolution to eliminate N+1 SQL_OID_LOOKUP round trips at startup.
+ */
+public class TypeRegistryTest {
+
+    // --- SQL_TYPES ORDER BY tests ---
+
+    @Test
+    void sqlTypesQueryOrdersBaseTypesBeforeDependentTypes() {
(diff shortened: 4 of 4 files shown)


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
```diff
@@ -476,16 +480,70 @@ private void prime() throws SQLException {
...
+            return false;
+        }
+        if (builder.hasParentType() && oidToType.get(builder.getParentTypeOid()) == null) {
+            return false;
+        }
+        return true;
+    }
+
+    private void resolveRemainingInBatch(List<TypeBuilderWithSchema> remaining) throws SQLException {
+        final List<Integer> missingOids = new ArrayList<>();
```
vjuranek: Would be better to use `Set` here to avoid duplicates?

vjuranek (commented):
Overall LGTM, I have one minor comment. Also adding an integration test which would create dependent custom types and check IOD lookup would be nice.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
```diff
@@ -476,16 +480,70 @@ private void prime() throws SQLException {
...
-                // which triggers additional SQL_OID_LOOKUP queries to PostgreSQL.
                 delayResolvedBuilders.add(builderWithSchema);
             }
 
-            // Resolve delayed builders
-            for (TypeBuilderWithSchema builderWithSchema : delayResolvedBuilders) {
-                addType(builderWithSchema.builder().build(), builderWithSchema.schemaName());
+            // Resolve delayed builders iteratively until no progress is made or all are resolved
+            boolean progress = true;
+            while (!delayResolvedBuilders.isEmpty() && progress) {
```
vjuranek, line 489: once more thing, you probably forgot to use `maxPasses` here (actually not declated anywhere in the code, but it's mentioned in the commit message)?

Vinu2111:
Thanks for the review!

Fixed the List → LinkedHashSet in resolveRemainingInBatch() to avoid duplicate OIDs updated in the latest commit.

Also updated the commit message to correctly describe the progress flag approach rather than the maxPasses guard that was mentioned earlier but not present in the code.

Regarding the integration test happy to add one. Could you point me to an existing integration test that creates custom types (e.g. DomainTypesIT or similar) so I can follow the same pattern for the new test?

vjuranek:
Hi @Vinu2111 , you can take a look e.g. on `RecordsStreamProducerIT` which has tests with domain types.

Vinu2111:
Added TypeRegistryIT with two tests, one verifying the connector starts successfully with chained domain types (domain of domain), and one verifying correct streaming of values through the dependent type chain. Followed the same pattern as DomainTypesIT. Updated in the latest commit.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TypeRegistryIT.java
```diff
@@ -0,0 +1,91 @@
...
+        TestHelper.execute("CREATE DOMAIN typeregistry.posint AS integer CHECK (VALUE > 0);");
+        TestHelper.execute("CREATE TABLE typeregistry.t1 ("
+                + "id serial primary key, "
+                + "amount typeregistry.numericex, "
+                + "quantity typeregistry.posint"
+                + ");");
+    }
+
+    @Test
+    @FixFor("DBZ-1683")
```
vjuranek: This is wrong issue ID. Could you please recreate the issue under https://github.com/debezium/dbz/issues and use `dbz#xxxx` ?

vjuranek (commented):
One more thing regarding issue ID. Could you please create new issue for this taks? Thanks!

Naros:
> Added TypeRegistryIT with two tests, one verifying the connector starts successfully with chained domain types (domain of domain), and one verifying correct streaming of values through the dependent type chain. Followed the same pattern as DomainTypesIT. Updated in the latest commit.

Please sort the imports in the new test class, so checkstyle passes @Vinu2111.  Thanks.

Vinu2111:
Hi @vjuranek and @Naros,

Both issues have been addressed in the latest commit (65f49ff):

1. Sorted imports in TypeRegistryIT.java - checkstyle should now pass.

2. Created new issue debezium/dbz#2041 and updated both 
   @FixFor annotations from DBZ-1683 to DBZ-2041.

Please let me know if anything else is needed. Thanks!

vjuranek:
@Vinu2111 could you please take a look on failing tests?

Vinu2111:
Hi @vjuranek, found and fixed the root cause of the CI failures.

The ORDER BY clause was appended directly to SQL_TYPES which broke SQL_NAME_LOOKUP and SQL_OID_LOOKUP since AND cannot follow ORDER BY in PostgreSQL syntax, causing a SQLException on all database-dependent integration tests.

Fixed by:
- Reverting SQL_TYPES to end with the WHERE clause only
- Appending ORDER BY only in the prime() execution path via SQL_TYPES_FOR_TEST

Unit tests passing locally. Updated in latest commit 29c67ea. 
Please let me know if anything else is needed. Thanks!

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
```diff
@@ -463,7 +467,7 @@ public static String normalizeTypeName(String typeName) {
     private void prime() throws SQLException {
         LOGGER.trace("Priming type registry with database types");
         try (Statement statement = connection.connection().createStatement();
-                ResultSet rs = statement.executeQuery(SQL_TYPES)) {
+                ResultSet rs = statement.executeQuery(SQL_TYPES_FOR_TEST)) {
```
atorik: According to the comment, SQL_TYPES_FOR_TEST is supposed to be "exposed for testing only".
However, if we also use it here in prime(), wouldn't it also be run in normal execution?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
```diff
@@ -476,16 +480,70 @@ private void prime() throws SQLException {
...
+                    if (canResolveNow(builderWithSchema.builder())) {
+                        addType(builderWithSchema.builder().build(), builderWithSchema.schemaName());
+                        resolvedInThisPass.add(builderWithSchema);
+                        progress = true;
+                    }
+                }
+                delayResolvedBuilders.removeAll(resolvedInThisPass);
+            }
+
+            if (!delayResolvedBuilders.isEmpty()) {
```
atorik, line 502: Is there a concrete case where this condition can be true?

That is, can the previous while loop end with `progress == false` and `delayedResolvedBuilders` still being non-empty?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
vjuranek: @atorik IMHO the comment is related to the visibility of `SQL_TYPES_FOR_TEST` constant - it's public so you can use it in the tests, otherwise it would be private.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
vjuranek: Thia happens whenever `delayResolvedBuilders` contains types for which `canResolveNow()` returns `false`, do you agree @Vinu2111 ?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
atorik: @vjuranek 
`SQL_TYPES_FOR_TEST` is intended only for regression test queries, right?

If that's the case, shouldn't we avoid replacing `SQL_TYPES` with `SQL_TYPES_FOR_TEST` inside prime()? The query in prime() is executed during normal initialization.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
```diff
@@ -70,12 +71,15 @@ public class TypeRegistry {
     private static final String SQL_ENUM_VALUES = "SELECT t.enumtypid as id, array_agg(t.enumlabel ORDER BY t.enumsortorder) as values "
             + "FROM pg_catalog.pg_enum t GROUP BY id";
 
-    private static final String SQL_TYPES = "SELECT t.oid AS oid, t.typname AS name, n.nspname AS schema_name, t.typelem AS element, t.typbasetype AS parentoid, t.typtypmod as modifiers, t.typcategory as category, e.values as enum_values "
```
atorik, line 73: This commit removes column 'element', but it still seems to be in use.
When I load a connector config, I encountered the following error:

```
ERROR [test-connector|task-0] WorkerSourceTask{id=test-connector-0} Task threw an uncaught and unrecoverable exception. Task is being killed and will not recover until manually restarted (org.apache.kafka.connect.runtime.WorkerTask:251)
io.debezium.DebeziumException: Failed to create TypeRegistry
        at io.debezium.connector.postgresql.connection.PostgresConnection.createTypeRegistry(PostgresConnection.java:131) ~[debezium-connector-postgres-3.6.0-SNAPSHOT.jar:3.6.0-SNAPSHOT]
        at io.debezium.connector.postgresql.PostgresConnectorTask.start(PostgresConnectorTask.java:116) ~[debezium-connector-postgres-3.6.0-SNAPSHOT.jar:3.6.0-SNAPSHOT]
        at io.debezium.connector.common.BaseSourceTask.start(BaseSourceTask.java:281) ~[debezium-connector-common-3.6.0-SNAPSHOT.jar:3.6.0-SNAPSHOT]
        at org.apache.kafka.connect.runtime.AbstractWorkerSourceTask.initializeAndStart(AbstractWorkerSourceTask.java:288) ~[connect-runtime-4.1.1.jar:?]
        at org.apache.kafka.connect.runtime.WorkerTask.doStart(WorkerTask.java:191) ~[connect-runtime-4.1.1.jar:?]
        at org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:242) [connect-runtime-4.1.1.jar:?]
        at org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:298) [connect-runtime-4.1.1.jar:?]
        at org.apache.kafka.connect.runtime.AbstractWorkerSourceTask.run(AbstractWorkerSourceTask.java:83) [connect-runtime-4.1.1.jar:?]
        at org.apache.kafka.connect.runtime.isolation.Plugins.lambda$withClassLoader$1(Plugins.java:254) [connect-runtime-4.1.1.jar:?]
        at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:545) [?:?]
        at java.base/java.util.concurrent.FutureTask.run(FutureTask.java:328) [?:?]
        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1095) [?:?]
        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:619) [?:?]
        at java.base/java.lang.Thread.run(Thread.java:1447) [?:?]
Caused by: io.debezium.DebeziumException: Couldn't initialize type registry
        at io.debezium.connector.postgresql.TypeRegistry.<init>(TypeRegistry.java:141) ~[debezium-connector-postgres-3.6.0-SNAPSHOT.jar:3.6.0-SNAPSHOT]
        at io.debezium.connector.postgresql.connection.PostgresConnection.createTypeRegistry(PostgresConnection.java:128) ~[debezium-connector-postgres-3.6.0-SNAPSHOT.jar:3.6.0-SNAPSHOT]
        ... 13 more
Caused by: org.postgresql.util.PSQLException: The column name element was not found in this ResultSet.
        at org.postgresql.jdbc.PgResultSet.findColumn(PgResultSet.java:3211) ~[postgresql-42.7.11.jar:42.7.11]
        at org.postgresql.jdbc.PgResultSet.getLong(PgResultSet.java:3077) ~[postgresql-42.7.11.jar:42.7.11]
        at io.debezium.connector.postgresql.TypeRegistry.createTypeBuilderFromResultSet(TypeRegistry.java:575) ~[debezium-connector-postgres-3.6.0-SNAPSHOT.jar:3.6.0-SNAPSHOT]
        at io.debezium.connector.postgresql.TypeRegistry.prime(TypeRegistry.java:473) ~[debezium-connector-postgres-3.6.0-SNAPSHOT.jar:3.6.0-SNAPSHOT]
        at io.debezium.connector.postgresql.TypeRegistry.<init>(TypeRegistry.java:138) ~[debezium-connector-postgres-3.6.0-SNAPSHOT.jar:3.6.0-SNAPSHOT]
        at io.debezium.connector.postgresql.connection.PostgresConnection.createTypeRegistry(PostgresConnection.java:128) ~[debezium-connector-postgres-3.6.0-SNAPSHOT.jar:3.6.0-SNAPSHOT]
        ... 13 more
```

Vinu2111:
Hi @atorik, thanks again for catching this and verifying it against a live connector that was a real production-breaking bug.

Root cause confirmed: SQL_TYPES aliases t.typelem as "elementoid", but createTypeBuilderFromResultSet() was still reading the old column name "element" at line 575. Fixed by updating the ResultSet read to use "elementoid", matching the query alias. Confirmed no other occurrences of the old column name exist in the file.

Unit tests passing. Pushed in latest commit 14e57c2.

@vjuranek confirmed re: canResolveNow() - yes, the loop can exit with progress == false while delayResolvedBuilders is still 
non-empty, e.g. when a type's dependency is itself unresolved or missing. That's exactly the scenario resolveRemainingInBatch() exists to handle as the fallback.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
vjuranek: Yes, I agree that the name is confusing. @Vinu2111 could you please rename `SQL_TYPES_FOR_TEST` to `SQL_TYPES`? Thanks!

Vinu2111:
Hi @vjuranek and @atorik,

Renamed the constants for clarity:- The original SQL_TYPES (no ORDER BY) is now SQL_TYPES_BASE, used  by SQL_NAME_LOOKUP, SQL_OID_LOOKUP, and the batch fallback query these must stay free of ORDER BY since they append AND conditions.- SQL_TYPES_FOR_TEST is now SQL_TYPES, used in prime() for the production type-loading query. Updated the comment above it to reflect that it's used in normal execution, not just tests.- Updated TypeRegistryTest.java references accordingly.

Unit tests passing. Pushed in latest commit 784ce82.

Vinu2111:
Hi @vjuranek, latest commit (784ce82) is pushed with the constant rename. Could you please approve the workflow run when you get a chance, so the PostgreSQL test suite can verify everything? Thanks!

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
atorik: @Vinu2111 
(You seem to write the reply in the conversation, so I quote it here:)
> confirmed re: canResolveNow() - yes, the loop can exit with progress == false while delayResolvedBuilders is still
> non-empty, e.g. when a type's dependency is itself unresolved or missing. 


Do such cases actually exist in practice?
I’d like to know what kind of type definitions could cause this situation to occur.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
```diff
@@ -476,16 +481,70 @@ private void prime() throws SQLException {
...
+
+        if (!missingOids.isEmpty()) {
+            String batchSql = SQL_TYPES_BASE + " AND t.oid = ANY(?)";
+            try (PreparedStatement statement = connection.connection().prepareStatement(batchSql)) {
+                Integer[] oidsArray = missingOids.toArray(new Integer[0]);
+                statement.setArray(1, connection.connection().createArrayOf("integer", oidsArray));
+                try (ResultSet rs = statement.executeQuery()) {
+                    while (rs.next()) {
+                        TypeBuilderWithSchema builderWithSchema = createTypeBuilderFromResultSet(rs);
+                        addType(builderWithSchema.builder().build(), builderWithSchema.schemaName());
```
atorik, line 538: The commit log says:

> Add a batch fallback using WHERE oid = ANY(?) with a LinkedHashSet to
collect unique unresolved OIDs and fetch them in a single round trip
instead of N individual SQL_OID_LOOKUP queries

However, isn't it still possible for Builder.build() to invoke typeRegistry.get(oid), which would then fall back to individual SQL_OID_LOOKUP queries even when the batch fallback path is used?

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TypeRegistryIT.java
```diff
@@ -0,0 +1,91 @@
...
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import io.debezium.connector.postgresql.PostgresConnectorConfig.SnapshotMode;
+import io.debezium.doc.FixFor;
+
+/**
+ * Integration test verifying that the TypeRegistry initialises correctly when
+ * dependent custom types are present, without triggering per-OID SQL_OID_LOOKUP
+ * round trips for each unresolved dependency (fix for DBZ-1683).
```
atorik, line 24: Does this test have a test that verifies that no individual SQL_OID_LOOKUP round trips occur? I couldn't find one.

atorik, line 24: I ran this test on the current main branch without this PR, and it passed.

Given that this PR aims to fix DBZ-1683, shouldn't an appropriate test fail on main and pass only with this PR applied?

atorik:
@Vinu2111 Thank you for working on this!
 I left a few comments/questions on the PR, so I would appreciate it if you could take a look.

Regarding the current approach..

> Fix
> 1. Add ORDER BY to SQL_TYPES
> 2. Iterative resolution in prime()
> 3. Batch fallback in resolveRemainingInBatch()

As far as I understand, this approach may repeatedly try to register the same type, skip it if its dependent type has not been registered yet, and then retry it later. That works, but I am not sure it is the best approach from an efficiency and simplicity perspective.

I think the reason this is necessary is that, when Debezium registers a new type, its `typbasetype` / `typelem` must already be registered in the `TypeRegistry`.

However, both `pg_type.typelem` and `pg_type.typbasetype` reference `pg_type.oid`, according to the PostgreSQL documentation:
https://www.postgresql.org/docs/current/catalog-pg-type.html

So if we load all rows from `pg_type` first, those references should always be resolvable from the loaded set. Based on that, I wonder if we can simplify the registration logic.

For example, the flow could be:

1. Read all rows from `pg_type` into `PostgresType.Builder` instances first.
2. Build each `PostgresType` without resolving `typbasetype` / `typelem` yet.
3. Register all of those unresolved `PostgresType` instances in `TypeRegistry`, so `oidToType` contains every loaded type.
4. In a second pass, resolve `typbasetype` and `typelem` directly from `oidToType` and attach the references to each `PostgresType`.
5. If a referenced OID is not found at this point, log it as an error, because all referenced types should already have been loaded from `pg_type`.

With this approach, we would not need to depend on `ORDER BY`, repeated iterative retries, or a batch fallback. The initial loading phase would also avoid calling `resolveUnknownType()`, because type references would be resolved only after all types are already registered.

WDYT?

Vinu2111:
Hi @atorik, I went and traced this through the actual code rather than just asserting it works, and you're right on both counts.

On the test (point 4): I tried reproducing your finding by running TypeRegistryIT against main directly. Confirmed the structural reason it passes either way the test only asserts the connector starts and reads correctly, never asserts on query count. With Testcontainers on localhost, the N+1 fallback queries complete in milliseconds either way, so the test can't distinguish "fixed" from "broken."

On the batch fallback (your build() question): traced this fully. PostgresType.Builder.build() calls typeRegistry.get(oid) directly for parentType/elementType. resolveRemainingInBatch() only fetches ONE level of missing dependencies via the batch query but when build() runs on those freshly-batched types (or on the original remaining types), if THEIR dependency is still uncached (e.g. a chained/nested miss), get() falls through to resolveUnknownType() and fires an individual SQL_OID_LOOKUP anyway. So you're correct: the batch fallback reduces but doesn't eliminate individual lookups for multi-level missing chains. That's also the concrete case answering your canResolveNow() question a type whose dependency is itself unresolved AND has its own unresolved dependency.

Given this, I think the right fix is to either (a) make the batch resolution recursive/transitive so it keeps batching until nothing's missing, or (b) add a query-count assertion to the test so this can't silently regress again. Want to think through which approach you'd prefer before I implement it, don't want to add complexity you'd rather avoid given your earlier point about simplicity.

atorik:
@Vinu2111  Hmm, I understood (a) as a discussion about how to fix the implementation, and (b) as a discussion about how to improve the test coverage. It seems those are separate concerns, so I'm not sure it necessarily has to be an either/or decision.

Regarding (a), how is that approach different from the earlier "iterative resolution in prime()" idea?
Also, what are your thoughts on [the idea](https://github.com/debezium/debezium/pull/7462#issuecomment-4766433810) I raised here?

Regarding (b), how would we determine the expected query count in a way that remains stable over time?
For example, one concern I have is that the number of built-in PostgreSQL types may change between PostgreSQL versions. Debezium's regression tests also make use of several extensions, and I'm not sure we can assume those extensions will never introduce additional types in the future.

One other thing: it looks like all of your replies have been posted in the main Conversation thread. Personally, I find it easier to follow code-specific discussions when they're attached directly to the relevant lines in the "[Files changed](https://github.com/debezium/debezium/pull/7462/files)" view.

Would you mind replying there for comments related to specific code changes?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
```diff
@@ -70,15 +71,19 @@ public class TypeRegistry {
     private static final String SQL_ENUM_VALUES = "SELECT t.enumtypid as id, array_agg(t.enumlabel ORDER BY t.enumsortorder) as values "
             + "FROM pg_catalog.pg_enum t GROUP BY id";
 
-    private static final String SQL_TYPES = "SELECT t.oid AS oid, t.typname AS name, n.nspname AS schema_name, t.typelem AS element, t.typbasetype AS parentoid, t.typtypmod as modifiers, t.typcategory as category, e.values as enum_values "
+    public static final String SQL_TYPES_BASE = "SELECT t.oid AS oid, t.typname AS name, t.typbasetype AS parentoid, t.typtypmod AS modifiers, t.typelem AS elementoid, n.nspname AS schema_name, t.typtype AS type, t.typcategory AS category, e.values AS enum_values "
```
Naros, line 74: Is there a reason why `SQL_TYPES_BASE` was made unnecessarily public?

Additionally, `t.typetype as type` was added to the query, but unless I've missed it, I don't see where this column is used, so it could be dropped.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
```diff
@@ -476,16 +481,70 @@ private void prime() throws SQLException {
...
+        if (builder.hasElementType() && oidToType.get(builder.getElementTypeOid()) == null) {
+            return false;
+        }
+        if (builder.hasParentType() && oidToType.get(builder.getParentTypeOid()) == null) {
+            return false;
+        }
+        return true;
+    }
+
+    private void resolveRemainingInBatch(List<TypeBuilderWithSchema> remaining) throws SQLException {
```
Naros, line 518: Is it possible to sort the `remaining` in topological order so that in the case the `remaining` has both `A` and `B` where `A` depends on `B` and `A` is processed first, we can avoid the one-off OID queries because we would guarantee that we process `B` before `A`?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
```diff
@@ -476,16 +481,70 @@ private void prime() throws SQLException {
...
+                progress = false;
+                final List<TypeBuilderWithSchema> resolvedInThisPass = new ArrayList<>();
+                for (TypeBuilderWithSchema builderWithSchema : delayResolvedBuilders) {
+                    if (canResolveNow(builderWithSchema.builder())) {
+                        addType(builderWithSchema.builder().build(), builderWithSchema.schemaName());
+                        resolvedInThisPass.add(builderWithSchema);
+                        progress = true;
+                    }
+                }
+                delayResolvedBuilders.removeAll(resolvedInThisPass);
```
Naros, line 499: As this PR is aimed at performance, I want to caution that `removeAll` here is `O(n*m)` since both are `ArrayList`. With many entries, this can really be slow.

I would prefer using something like `Iterator#remove` during the pass or a `Set`-based approach where the cost would be amortized as `O(n)` at least.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TypeRegistryIT.java
```diff
@@ -0,0 +1,91 @@
...
+        TestHelper.dropAllSchemas();
+        TestHelper.execute("CREATE SCHEMA typeregistry;");
+        // Chain: numeric82 <- numericex (domain of domain)
+        TestHelper.execute("CREATE DOMAIN typeregistry.numeric82 AS numeric(8,2);");
+        TestHelper.execute("CREATE DOMAIN typeregistry.numericex AS typeregistry.numeric82;");
+        // Array of a custom domain type
+        TestHelper.execute("CREATE DOMAIN typeregistry.posint AS integer CHECK (VALUE > 0);");
+        TestHelper.execute("CREATE TABLE typeregistry.t1 ("
+                + "id serial primary key, "
+                + "amount typeregistry.numericex, "
```
Naros, line 39: Should these tests also assert on `amount` rather than just a combination of `id` and `quantity`?

Naros (commented):
Hi @Vinu2111 a few inline comments.

PR #7471, opened 2026-05-24
DBZ-1976: set autoCommit=false on connection init; use execute() in getEstimatedTableSize

ptqa:
Fixes [debezium/dbz#1976](https://github.com/debezium/dbz/issues/1976)

## Description
getEstimatedTableSize() calls executeWithoutCommitting("USE db"). That method guards against auto-commit being enabled and throws DebeziumException (a RuntimeException) if it is. But getEstimatedTableSize() only catches SQLException, so DebeziumException escapes uncaught and kills the snapshot.

Auto-commit is enabled on the connection when the connection drops and is re-established between steps 6 and 7 — new connections default to `auto-commit=true`. With `locking.mode=none` on RDS, no table locks are held so the connection can go fully idle during the long schema history write in step 6, making reconnection likely in production.

executeWithoutCommitting is wrong here: USE db has no transactional semantics. It doesn't need to run inside a transaction. executeWithoutCommitting is intended for statements that must run within an existing transaction (e.g. LOCK TABLES). execute() is the correct method — it works regardless of auto-commit state.

## PR Checklist

### DIFF
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogConnectorConnection.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogConnectorConnection.java
@@ -55,9 +55,13 @@ public abstract class BinlogConnectorConnection extends JdbcConnection {
     private final BinlogFieldReader fieldReader;
 
     public BinlogConnectorConnection(ConnectionConfiguration configuration, BinlogFieldReader fieldReader) {
-        super(configuration.config(), configuration.factory(), QUOTED_CHARACTER, QUOTED_CHARACTER);
+        super(configuration.config(), configuration.factory(), initialOperations(), QUOTED_CHARACTER, QUOTED_CHARACTER);
         this.connectionConfig = configuration;
         this.fieldReader = fieldReader;
+    }
+
+    private static Operations initialOperations() {
+        return statement -> statement.getConnection().setAutoCommit(false);
     }
 
     @Override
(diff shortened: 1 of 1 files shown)


Naros:
Hi kindly refer to the issue in your commit messages as `debezium/dbz#1976` instead of `DBZ-1976`. Thanks.

Naros:
> Auto-commit is enabled on the connection when the connection drops and is re-established between steps 6 and 7 — new connections default to auto-commit=true. With locking.mode=none on RDS, no table locks are held so the connection can go fully idle during the long schema history write in step 6, making reconnection likely in production.

I do wonder, @ptqa, if the ideal solution is to address either the `auto-commit` problem or the exception handling. I believe some of the connectors set their auto-commit as an init operation, and Oracle does it as a ctor argument, so on reconnect, the state is retained. 

That seems like a more apprpropriate solution here most lkely, wdyt?

ptqa:
> I do wonder, @ptqa, if the ideal solution is to address either the `auto-commit` problem or the exception handling. I believe some of the connectors set their auto-commit as an init operation, and Oracle does it as a ctor argument, so on reconnect, the state is retained.
> 
> That seems like a more apprpropriate solution here most lkely, wdyt?

@Naros ok, I think it's still good to replace executeWithoutCommitting() with execute(). So I've added the same pattern used in Oracle connector.

Seems to work on my test setup.

vjuranek (approved):
LGTM

vjuranek:
Merged, thanks @ptqa !

PR #7482, opened 2026-05-28
debezium/dbz#1987 Upgrade Kafka to 4.3.0

jpechane:
Fixes https://github.com/debezium/dbz/issues/1987


jpechane:
@Naros Test failures are unrelated

PR #7483, opened 2026-05-28
debezium/dbz#1988 Quote identifiers in rowCountForTableChunked to fix SQL syntax error on hyphenated database names

### LINKED ISSUE #1988: rowCountForTableChunked generates unquoted SQL identifiers, causing SQLSyntaxErrorException on hyphenated database names
## Bug report

`Debezium MySQL Connector 3.5.1.Final`

---

 ```json
  {
    "snapshot.mode": "when_needed",
    "snapshot.max.threads": "2",
    "internal.legacy.snapshot.max.threads": "false"
  }
```
  The key combination is snapshot.max.threads > 1 with internal.legacy.snapshot.max.threads=false
  (the default). Any connector config that triggers the chunked snapshot path against a database
  whose name contains a hyphen will reproduce this.

---

`MySQL 8.0, AWS RDS (also reproducible with any MySQL deployment). The database name contains
  a hyphen, e.g. my-database.`

---

The connector should successfully complete a snapshot of all tables in the database, regardless
  of whether the database name contains hyphens or other special characters.

---

**What behavior do you see?**

The connector crashes during snapshot with a SQLSyntaxErrorException. MySQL interprets the
  unquoted hyphenated database name as arithmetic subtraction:
```
      at io.debezium.pipeline.ErrorHandler.setProducerThrowable(ErrorHandler.java:67)
      ...
  Caused by: io.debezium.DebeziumException: java.sql.SQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near
  '-database.my_table' at line 1
      at io.debezium.pipeline.source.AbstractSnapshotChangeEventSource.execute(AbstractSnapshotChangeEventSource.java:111)
      at io.debezium.pipeline.ChangeEventSourceCoordinator.doSnapshot(ChangeEventSourceCoordinator.java:311)
      ...
  Caused by: java.sql.SQLSyntaxErrorException: You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-database.my_table' at line 1
      at io.debezium.relational.RelationalSnapshotChangeEventSource.rowCountForTableChunked(RelationalSnapshotChangeEventSource.java:1103)
      at io.debezium.relational.RelationalSnapshotChangeEventSource.lambda$createChunkedDataEvents$7(RelationalSnapshotChangeEventSource.java:572)
      at io.debezium.relational.RelationalSnapshotChangeEventSource.prepareTables(RelationalSnapshotChangeEventSource.java:1364)
      at io.debezium.relational.RelationalSnapshotChangeEventSource.createChunkedDataEvents(RelationalSnapshotChangeEventSource.java:570)
      at io.debezium.relational.RelationalSnapshotChangeEventSource.doExecute(RelationalSnapshotChangeEventSource.java:197)
```
  Root cause: rowCountForTableChunked builds its COUNT query using
  jdbcConnection.getQualifiedTableName(tableId), which returns an unquoted
  catalog.table string. BinlogConnectorConnection.getQualifiedTableName is
  implemented as tableId.catalog() + "." + tableId.table() — no quoting. When
  the catalog (database) name contains a hyphen, MySQL sees my - database.table
  and raises a syntax error.

  Every other method in BinlogSnapshotChangeEventSource that constructs SQL table
  references correctly uses quotedTableIdString() (which returns `catalog`.`table`
  with backtick quoting). rowCountForTableChunked is the only outlier.

  Workaround: set internal.legacy.snapshot.max.threads=true to bypass the chunked
  code path entirely.`

---

Confirmed on 3.5.1.Final. Not yet verified on a newer version — the bug is in
  RelationalSnapshotChangeEventSource.rowCountForTableChunked which was introduced
  with the chunked snapshot feature and the fix has not been observed in any changelog.

---

The relevant stack trace is included above.

---

1. Start MySQL and create a database with a hyphen in its name:
 ```
CREATE DATABASE `my-test-db`;
CREATE TABLE `my-test-db`.`items` (id INT PRIMARY KEY, name VARCHAR(50));
INSERT INTO `my-test-db`.`items` VALUES (1, 'a'), (2, 'b'), (3, 'c');
```
  2. Configure the Debezium MySQL connector with:
  ```
{
    "database.include.list": "my-test-db",
    "snapshot.max.threads": "2",
    "internal.legacy.snapshot.max.threads": "false"
}
```
  3. Start the connector. It will crash immediately during snapshot with the
  SQLSyntaxErrorException shown above.

  Proposed fix: In RelationalSnapshotChangeEventSource.rowCountForTableChunked
  (line 1103), replace jdbcConnection.getQualifiedTableName(tableId) with
  jdbcConnection.quotedTableIdString(tableId). A PR with this fix and a
  regression test is available at:
  https://github.com/ryandsouzaappfolio/debezium/tree/fix/DBZ-hyphenated-db-name-chunked-snapshot

### PR DESCRIPTION
**Pull request by ryandsouzaappfolio**
Fixes debezium/dbz#1988

## Description
When snapshot.max.threads > 1 and internal.legacy.snapshot.max.threads=false (the default chunked snapshot path), rowCountForTableChunked generates a COUNT query using getQualifiedTableName(), which returns an unquoted catalog.table string. MySQL database names that contain hyphens (e.g. "my-database") cause MySQL to interpret the hyphen as arithmetic subtraction, producing a SQLSyntaxErrorException and stopping the connector.

Fix: use quotedTableIdString() instead, which the binlog connector overrides to return `catalog`.`table` with backtick quoting.

## PR Checklist

### DIFF
--- a/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogChunkedSnapshotIT.java
+++ b/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogChunkedSnapshotIT.java
@@ -12,13 +12,17 @@ import java.util.List;
 import org.apache.kafka.connect.source.SourceConnector;
 import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
 
+import io.debezium.config.CommonConnectorConfig;
 import io.debezium.config.Configuration;
 import io.debezium.connector.binlog.util.BinlogTestConnection;
 import io.debezium.connector.binlog.util.TestHelper;
 import io.debezium.connector.binlog.util.UniqueDatabase;
+import io.debezium.doc.FixFor;
 import io.debezium.jdbc.JdbcConnection;
 import io.debezium.pipeline.AbstractChunkedSnapshotTest;
+import io.debezium.relational.RelationalDatabaseConnectorConfig;
 
 /**
  * Abstract binlog chunked table snapshot integration tests.
@@ -124,5 +128,42 @@ public abstract class BinlogChunkedSnapshotIT<T extends SourceConnector>
     @Override
     protected String getFullyQualifiedTableName(String tableName) {
         return DATABASE.qualifiedTableName(tableName);
+    }
+
+    /**
+     * Regression test for DBZ-1988: rowCountForTableChunked uses getQualifiedTableName (unquoted)
+     * instead of quotedTableIdString, causing a SQL syntax error when the database name contains
+     * hyphens (e.g. "my-db" becomes the expression `my - db` in MySQL).
+     */
+    @FixFor("debezium/dbz#1988")
+    @Test
+    public void shouldSnapshotTableInDatabaseWithHyphenatedName() throws Exception {
+        final String hyphenDb = "debezium-hyphen-test";
+        final int rowCount = 10;
+
+        try {
+            connection.execute(
+                    "CREATE DATABASE IF NOT EXISTS `" + hyphenDb + "`",
+                    "CREATE TABLE `" + hyphenDb + "`.`items` (id INT PRIMARY KEY, name VARCHAR(50))",
+                    "INSERT INTO `" + hyphenDb + "`.`items` VALUES " +
+                            "(1,'a'),(2,'b'),(3,'c'),(4,'d'),(5,'e'),(6,'f'),(7,'g'),(8,'h'),(9,'i'),(10,'j')");
+
+            final Configuration config = DATABASE.defaultConfig()
+                    .with(BinlogConnectorConfig.DATABASE_INCLUDE_LIST, hyphenDb)
+                    .with(RelationalDatabaseConnectorConfig.TABLE_INCLUDE_LIST, hyphenDb + ".items")
+                    .with(CommonConnectorConfig.SNAPSHOT_MAX_THREADS, 2)
+                    .build();
+
+            start(getConnectorClass(), config);
+            assertConnectorIsRunning();
+            waitForSnapshotToBeCompleted();
+
+            final var records = consumeRecordsByTopic(rowCount);
+            assertThat(records.allRecordsInOrder()).hasSize(rowCount);
+        }
+        finally {
+            connection.execute("DROP DATABASE IF EXISTS `" + hyphenDb + "`");
+            stopConnector();
+        }
     }
 }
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogChunkedSnapshotIT.java [hunk 1]
```diff
@@ -125,4 +129,41 @@ protected String getTableTopicName(String tableName) {
     protected String getFullyQualifiedTableName(String tableName) {
         return DATABASE.qualifiedTableName(tableName);
     }
+
+    /**
+     * Regression test for DBZ-1988: rowCountForTableChunked uses getQualifiedTableName (unquoted)
+     * instead of quotedTableIdString, causing a SQL syntax error when the database name contains
+     * hyphens (e.g. "my-db" becomes the expression `my - db` in MySQL).
+     */
+    @FixFor("DBZ-1988")
```
Naros:
```suggestion
    @FixFor("debezium/dbz#1988")
```

Naros (commented):
Hi @ryandsouzaappfolio one small inline point, and it LGTM.

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogChunkedSnapshotIT.java
ryandsouzaappfolio: Thanks @Naros! I've updated the commit with the requested change.

Naros (approved):

Naros:
Hi @ryandsouzaappfolio there are some test failures, if you could take a look.

ryandsouzaappfolio:
> Hi @ryandsouzaappfolio there are some test failures, if you could take a look.

Hi @Naros, I've updated the tests. Could you re-review?

vjuranek (approved):

debezium-connector-common/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
```diff
@@ -1106,7 +1106,7 @@ protected OptionalLong rowCountForTable(TableId tableId) {
     protected Long rowCountForTableChunked(TableId tableId) throws SQLException {
         // todo: snapshot select overrides?
         return jdbcConnection.queryAndMap(
-                "SELECT COUNT(1) FROM %s".formatted(jdbcConnection.getQualifiedTableName(tableId)),
+                "SELECT COUNT(1) FROM %s".formatted(jdbcConnection.quotedTableIdString(tableId)),
```
Naros, line 1109: The `tableId` has the database name encoded, so for Oracle this creates:
```
"DB"."SCHEMA"."TABLE"
```
This is not going to work for databases like Oracle, which is why the old code used the qualified name method so that it returned:
```
SCHEMA.TABLE
```
Can you add an override method in `OracleSnapshotChangeEventSource` where the supplied `TableId` is copied into a new `TableId` with a null database, see inside `buildSelectWithRowLimits` for an example.

Naros (changes requested):

Naros:
It appears this was fixed and merged as part of debezium/dbz#2173 - closing. Thanks for the contribution @ryandsouzaappfolio.

PR #7490, opened 2026-06-01
debezium/dbz#2002 add cache in ReselectColumnPostProcessor

gaurav7261:
Fixes debezium/dbz#2002
## Description

Adds an optional, **pluggable per-column cache** to the `ReselectColumnsPostProcessor`
(`debezium-connector-common`).

Today, every event carrying an unavailable-value placeholder (e.g. unchanged PostgreSQL TOAST
columns, Oracle LOBs) or a `null` to refresh triggers a `SELECT` against the source database. For
frequently-updated rows with large, rarely-changing out-of-line columns, this produces many repeated,
identical re-selection queries for the same `(table, key, column)` tuple, adding avoidable load and
latency (notably on managed databases such as Aurora reader endpoints).

This change lets reselected values be reused across events:

- Cache exposed behind a **strategy interface** (`ReselectColumnCache`) so alternative backends (e.g.
  an embedded key/value store such as RocksDB, or a user-custom implementation) can be plugged in
  without changing the post-processor. A default in-memory implementation (`MemoryReselectColumnCache`)
  is provided and is selected via `reselect.cache.type`.
- **Hierarchical `table -> row -> column` layout.** The post-processor resolves a row-scoped handle
  once per event (`forRow(tableId, keyValues)`), so the table and row-key identity are serialized a
  single time and reused for every column of that row, instead of being re-serialized per column. This
  keeps the cost proportional to one row + N columns for a row with N LOB/TOAST columns.
- On a qualifying event, only the **cache misses** are queried (in a single `SELECT`); hits are served
  from the cache.
- **Invalidate-on-modify**: any event that carries a real (non-placeholder) value for a column evicts
  that column's cache entry, so a cached value is never served after the column has changed.
  Correctness therefore does **not** depend on the TTL.
- The default in-memory cache is bounded by size (LRU on the number of cached rows) via the existing
  `io.debezium.util.BoundedConcurrentHashMap` — **no new dependency** — and by a per-value TTL that
  only bounds memory retention for cold rows.
- Cached values are the **raw** JDBC values; the existing `getConvertedValue(...)` conversion is
  applied uniformly to both cache hits and freshly queried values.

New configuration (all optional, **cache disabled by default** — no behavioural change for existing
users):

| Property | Default | Description |
|---|---|---|
| `reselect.cache.enabled` | `false` | Enable the re-selection cache. |
| `reselect.cache.type` | `MemoryReselectColumnCache` | `ReselectColumnCache` implementation class. |
| `reselect.cache.max.size` | `10000` | Max cached rows (LRU). Default in-memory cache only. |
| `reselect.cache.ttl.ms` | `600000` | Entry time-to-live (ms). Default in-memory cache only. |

Includes unit tests for the post-processor (cache hit/miss, partial-hit single-column query,
invalidate-on-modify, TTL expiry, key isolation, read-event skip) and for the in-memory cache
(get/put/invalidate, TTL expiry, binary-key isolation), plus documentation for the new properties in
`reselect-columns.adoc`.

## PR Checklist

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
+++ b/debezium-connector-common/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
@@ -10,7 +10,7 @@ import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.HashMap;
-import java.util.HashSet;
+import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -257,7 +257,7 @@ public class ReselectColumnsPostProcessor implements PostProcessor, BeanRegistry
             }
         }
 
-        final List<String> requiredColumnSelections = getRequiredColumnSelections(tableId, after);
+        final Set<String> requiredColumnSelections = getRequiredColumnSelections(tableId, after);
 
         // Resolve the row-scoped cache view once, keyed by the event's message key struct, so the row
         // identity is resolved a single time and reused for every column of this row. Using the key struct
@@ -364,14 +364,13 @@ public class ReselectColumnsPostProcessor implements PostProcessor, BeanRegistry
      * @param requiredColumnSelections columns being re-selected this event, which are skipped here since
      *                                 they will be cached after re-selection below.
      */
-    private void cacheModifiedColumns(ReselectColumnCache.RowCache rowCache, TableId tableId, Struct after, List<String> requiredColumnSelections) {
+    private void cacheModifiedColumns(ReselectColumnCache.RowCache rowCache, TableId tableId, Struct after, Set<String> requiredColumnSelections) {
         if (rowCache == null) {
             return;
         }
-        final Set<String> requiredColumnSelectionsSet = new HashSet<>(requiredColumnSelections);
         for (org.apache.kafka.connect.data.Field field : after.schema().fields()) {
             final String columnName = field.name();
-            if (requiredColumnSelectionsSet.contains(columnName)) {
+            if (requiredColumnSelections.contains(columnName)) {
                 continue;
             }
             final String fullyQualifiedName = jdbcConnection.getQualifiedTableName(tableId) + ":" + columnName;
@@ -418,8 +417,8 @@ public class ReselectColumnsPostProcessor implements PostProcessor, BeanRegistry
         return key.get(field);
     }
 
-    private List<String> getRequiredColumnSelections(TableId tableId, Struct after) {
-        final List<String> columnSelections = new ArrayList<>();
+    private Set<String> getRequiredColumnSelections(TableId tableId, Struct after) {
+        final Set<String> columnSelections = new LinkedHashSet<>();
         for (org.apache.kafka.connect.data.Field field : after.schema().fields()) {
             final Object value = after.get(field);
             if (reselectUnavailableValues && isUnavailableValueHolder(field, value)) {
--- a/debezium-connector-common/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
+++ b/debezium-connector-common/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
@@ -9,8 +9,12 @@ import java.nio.ByteBuffer;
 import java.sql.SQLException;
 import java.util.ArrayList;
 import java.util.Collection;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
 import java.util.function.Predicate;
 import java.util.regex.Pattern;
 
@@ -29,6 +33,7 @@ import io.debezium.common.annotation.Incubating;
 import io.debezium.config.Configuration;
 import io.debezium.config.EnumeratedValue;
 import io.debezium.config.Field;
+import io.debezium.config.Instantiator;
 import io.debezium.connector.AbstractSourceInfo;
 import io.debezium.data.Envelope;
 import io.debezium.data.Json;
@@ -36,6 +41,8 @@ import io.debezium.data.SpecialValueDecimal;
 import io.debezium.data.VariableScaleDecimal;
 import io.debezium.function.Predicates;
 import io.debezium.jdbc.JdbcConnection;
+import io.debezium.processors.reselect.cache.MemoryReselectColumnCache;
+import io.debezium.processors.reselect.cache.ReselectColumnCache;
 import io.debezium.processors.spi.PostProcessor;
 import io.debezium.relational.Column;
 import io.debezium.relational.CustomConverterRegistry;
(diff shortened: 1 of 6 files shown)


Naros (commented):
As discussed offline, the principle of a cache makes sense, but this should be designed to be extendible, so that other implementations like RocksDB or even a user-custom implementation could be plugged in.

debezium-connector-common/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
```diff
@@ -68,10 +72,33 @@ public class ReselectColumnsPostProcessor implements PostProcessor, BeanRegistry
...
+    // key values or column names, avoiding delimiter-collision ambiguity between key components.
+    private static final char CACHE_KEY_SEPARATOR = '\u0001';
+
     private Predicate<String> selector;
     private boolean reselectUnavailableValues;
     private boolean reselectNullValues;
     private boolean reselectUseEventKeyFields;
+    private boolean reselectCacheEnabled;
+    private long reselectCacheTtlMs;
+    private BoundedConcurrentHashMap<String, CachedColumn> reselectCache;
```
Naros: Given the cache key design, if I have multiple LOB columns in a single table, we incur the cost of serializing the primary key values for each LOB because of how the map is designed.  

I would have ideally expected we pay for the `TableId` serialization once, the row's key `Struct` once, and then each column/value tuple once. Could a hierarchical map help here or use different maps for different purposes that orchestrate like a hierarchical map?

debezium-connector-common/src/main/java/io/debezium/processors/reselect/cache/MemoryReselectColumnCache.java
```diff
@@ -0,0 +1,140 @@
...
+     * rendered by content (not identity) so binary primary keys produce stable, collision-free keys.
+     */
+    private static String rowKey(TableId tableId, List<Object> keyValues) {
+        final StringBuilder sb = new StringBuilder(tableId.toString());
+        for (Object value : keyValues) {
+            sb.append(KEY_SEPARATOR);
+            sb.append(value instanceof byte[] ? Arrays.toString((byte[]) value) : String.valueOf(value));
+        }
+        return sb.toString();
+    }
```
Naros: I'm still not convinced this is the best approach for keys.  

For example, let's say that you have a table with PK `{ID1 INT, ID2 INT}`. And at some point, you decide that you want to modify the primary key to be `{ID2 INT, ID1 INT}` by reversing the two columns. 

Before the DDL change, you had an entry `{20,30}` for row A, but row B `{30,20}` hadn't been modified, so it's currently not registered. After the DDL change, you modify row B, but you now get a hit for row A from earlier and return the wrong cached values.

This is why I really suggest we should use the supplied `messageKey`, which is the first argument to the `PostProcessor#apply` call, rather than some synthetic construction. 

Given that Debezium never emits events with a primitive event key, this is generally a `Struct` object. The nice thing about using this as the key is that factors like default value changes, DDL changes, schema-to-schema parameters, or how a column is serialized control whether there is or is not a hit.

When we look at persistent data stores, e.g. RocksDB for example, having a reliable key value that can be treated as a cache miss when these secondary attributes change becomes important.

debezium-connector-common/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
```diff
@@ -225,35 +255,110 @@ public void apply(Object messageKey, Struct value) {
...
+        final List<String> requiredColumnSelections = getRequiredColumnSelections(tableId, after);
+
+        // Resolve the row-scoped cache view once so the row identity (table + key) is serialized a single
+        // time and reused for every column of this row, rather than re-derived per column.
+        final ReselectColumnCache.RowCache rowCache = isCacheEnabled() ? reselectCache.forRow(tableId, keyValues) : null;
+
+        // Invalidate-on-modify: any column that arrived with a real (non-placeholder) value must not be
+        // served from a stale cache entry, so evict it. This keeps the cache correct across row
+        // modifications, so a long TTL never serves a value that has since changed.
+        invalidateModifiedColumns(rowCache, after, requiredColumnSelections);
```
Naros: But if it does arrive in the `after` with a value, would it make sense to cache the updated value so that if the next event arrives for the same row but did not modify the LOB column, we can again skip the lookup for the column and instead use the cache?

debezium-connector-common/src/main/java/io/debezium/processors/reselect/cache/MemoryReselectColumnCache.java
```diff
@@ -0,0 +1,140 @@
...
+            }
+            final CachedValue cached = columns.get(column);
+            if (cached == null || cached.isExpired(System.currentTimeMillis(), ttlMs)) {
+                return Optional.empty();
+            }
+            return Optional.ofNullable(cached.value);
+        }
+
+        @Override
+        public void put(String column, Object value) {
```
Naros, line 98: If the LOB value is actually `null`, we likely would want to cache that value. But then doing that makes it indistinguishable from `Optional.empty()` in the `get`. 

I would expect if we're going to truly cache all LOB column values, `null` would be kept for future events. So we probably need a marker object that we can use to identify if the cached value is `null` or wasn't cached in the `get` call, no?

Naros (commented):
Hi @gaurav7261 I have a few inline comments, if you could take a look and share some feedback.

gaurav7261:
@Naros please review again

debezium-connector-common/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
```diff
@@ -225,35 +255,125 @@ public void apply(Object messageKey, Struct value) {
...
+        // identity is resolved a single time and reused for every column of this row. Using the key struct
+        // means schema/DDL/default-value changes naturally produce a cache miss rather than a false hit.
+        final ReselectColumnCache.RowCache rowCache = isCacheEnabled() ? reselectCache.forRow(key) : null;
+
+        // Cache-on-modify: any column that arrived with a real (non-placeholder) value reflects the row's
+        // current state, so refresh the cache with it. A later event that does not modify that column
+        // (e.g. an unchanged TOAST/LOB re-emitted as a placeholder) can then be served from the cache
+        // instead of re-querying. This also keeps the cache correct across row modifications without
+        // depending on the TTL.
+        cacheModifiedColumns(rowCache, tableId, after, requiredColumnSelections);
```
Naros: I think it would read more clearly here if the cache calls were wrapped by `if (isCacheEnabled())`. It yields the same net result as the code does now, but without having to navigate into `cacheModifiedColumns` to realize that when `isCacheEnabled()` is false, that method is a no-op.

debezium-connector-common/src/main/java/io/debezium/processors/reselect/cache/MemoryReselectColumnCache.java
```diff
@@ -0,0 +1,128 @@
...
+
+    public static final String MAX_SIZE = "reselect.cache.max.size";
+    public static final String TTL_MS = "reselect.cache.ttl.ms";
+
+    private static final int DEFAULT_MAX_SIZE = 10_000;
+    private static final long DEFAULT_TTL_MS = 600_000L;
+
+    private long ttlMs;
+    // message key struct -> (column -> value). The outer map is bounded/LRU on the number of rows.
+    private BoundedConcurrentHashMap<Struct, Map<String, CachedValue>> rows;
```
Naros, line 52: In general, this shouldn't be an issue, but there are corner cases where users have tables that contain 500-1000 columns. Since the reselect post-process can reselect on `null` values, I wonder if there should be guards here to ensure the inner map is bounded, too?  WDYT?

Naros (commented):
Hi @gaurav7261 LGTM, I've left just two inline comments.

debezium-connector-common/src/main/java/io/debezium/processors/reselect/cache/MemoryReselectColumnCache.java
gaurav7261: yes make sense, doing changes @Naros

gaurav7261:
@Naros please review changes

debezium-connector-common/src/main/java/io/debezium/processors/reselect/ReselectColumnsPostProcessor.java
```diff
@@ -225,35 +255,127 @@ public void apply(Object messageKey, Struct value) {
...
+     */
+    private void cacheModifiedColumns(ReselectColumnCache.RowCache rowCache, TableId tableId, Struct after, List<String> requiredColumnSelections) {
+        if (rowCache == null) {
             return;
         }
+        for (org.apache.kafka.connect.data.Field field : after.schema().fields()) {
+            final String columnName = field.name();
+            if (requiredColumnSelections.contains(columnName)) {
+                continue;
+            }
```
Naros, line 375: We'll need to keep an eye on performance here. The `contains` method performs an O(n) lookup on the `List`, whereas using a `Set` would make the cost amortized O(1). For small tables, this is likely negligible, but for large, wide tables, it could add up and lead to higher-than-expected CPU usage.

Do you want to tackle that now or wait @gaurav7261 ?

Naros (approved):

gaurav7261:
@Naros can we merge it

jpechane:
@gaurav7261 We are on 3.7 version now, could you please do lat rebase? I think the PR is good to go.
There is one question for Chris related to performance, did you see it?

gaurav7261:
done @jpechane , please review cc @Naros

Naros (approved):
I made one small change and moved the `Set` construction earlier and replaced the `List` with a `LinkedHashSet`. This preserves the insertion order of `List` but gives the `O(1)` lookup semantics while reducing 1 allocation per event.

@gaurav7261 do you agree with that change?

gaurav7261:
make sense @Naros ,we can go ahead now

Naros:
Applied, thanks @gaurav7261 :tada:

PR #7494, opened 2026-06-01
debezium/dbz#1081 Update quarkus extension documentation with event capturing advanced filtering

vsantonastaso:
Related to https://github.com/debezium/debezium-quarkus/pull/55

### DIFF
--- a/documentation/modules/ROOT/pages/integrations/quarkus-debezium-engine-extension.adoc
+++ b/documentation/modules/ROOT/pages/integrations/quarkus-debezium-engine-extension.adoc
@@ -159,6 +159,8 @@ The `CapturingEvent<T>` contains information related to the kind of database ope
 
 === Filtering Change Data Capture Events
 
+==== By destination
+
 It's possible to filter events by `destination`:
 
 ```java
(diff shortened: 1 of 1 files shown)


documentation/modules/ROOT/pages/integrations/quarkus-debezium-engine-extension.adoc
```diff
@@ -171,6 +173,78 @@ It's possible to filter events by `destination`:
...
+    }
+```
+
+[NOTE]
+====
+The `filter` attribute is mutually exclusive with `destination` and `engine`.
+Multiple `@Capturing` methods with the same destination and engine are not allowed.
+Both validation are performed at build time.
+
+Similarly, multiple filter-based methods whose filters accept the same event will result in a runtime error.
```
Naros: I was talking with Giovanni about this, and he explained the reasoning to me. 

I do think we might benefit from framing this more along the lines that the extension does not provide multiplexing capabilities where multiple change consumers can consume from the same destination. If an application requires multiple consumers for the same destination, its the responsibility of the hosting application to handle the multiplexing concerns accordingly.

WDYT?

vsantonastaso: Yes, I totally agree. It should be the responsibility of the host to handle multiple capture methods for the same event. 
We could avoid mentioning it in the documentation and open a dedicated task to extend this kind of compatibility. WDYT? @Naros @kmos

kmos (approved):

documentation/modules/ROOT/pages/integrations/quarkus-debezium-engine-extension.adoc
kmos: We need to have a deeper discussion on this topic. The direction we’ve taken so far may not be the right one. The actual use of the extensions will likely be driven by Debezium Server, and we can adjust or refine our direction based on community needs and feedback.

documentation/modules/ROOT/pages/integrations/quarkus-debezium-engine-extension.adoc
```diff
@@ -171,6 +173,89 @@ It's possible to filter events by `destination`:
...
+```
+
+[NOTE]
+====
+The `filter` attribute is mutually exclusive with `destination` and `engine`, validated at build time.
+====
+
+==== By field changes
+
+A built-in `CapturingFieldsFilterStrategy` is provided for filtering events based on whether specific fields are impacted.
```
Naros: I think we should pull up the bullet section under "The field filter behavior depends on the event type:" here and group this together. The term _impacted_ isn't clear just by the initial description, and brining the event type matrix up gives more clarity.

> A built-in `CapturingFieldsFilterStrategy` can filter events based on the state of specific fields, where that state depends on the event type. The behavior is outlined below for each event type: 

Then we can move on to the examples.

documentation/modules/ROOT/pages/integrations/quarkus-debezium-engine-extension.adoc
```diff
@@ -171,6 +173,89 @@ It's possible to filter events by `destination`:
...
+
+For batch events, each record in the batch is individually evaluated by `shouldCapture` and only matching records are delivered to the filter handler.
+
+==== Fallback chain
+
+When multiple `@Capturing` methods are defined, events are dispatched through a fallback chain:
+
+1. *Filter registry*: methods with `filter` are checked first. The filter's `shouldCapture` decides.
+2. *Object invoker*: methods with a registered deserializer for the destination (single event only).
+3. *Default registry*: methods with `destination` (exact match, then wildcard `*` fallback).
```
Naros: For an end user that is using the extensions, I think its sufficient to provide the text after the `:`. I don't believe knowing if its a Filter Registry, Invoker, or Default Registry add any value or context.

documentation/modules/ROOT/pages/integrations/quarkus-debezium-engine-extension.adoc
```diff
@@ -171,6 +173,89 @@ It's possible to filter events by `destination`:
...
+
+First match wins. A `@Capturing()` with no destination acts as a wildcard and catches any event not matched by the previous levels. If no wildcard is defined, unmatched events are skipped.
```
Naros:
```suggestion
First match wins. 
A `@Capturing()` with no destination acts as a wildcard and catches any event not matched by the previous levels. 
If no wildcard is defined, unmatched events are skipped.
```

kmos (approved):

Naros (approved):

Naros:
Applied, thanks @vsantonastaso.

PR #7497, opened 2026-06-01
dbz#2009 - fix bug where errors.max.retries is not being used

DanGrabski:
Fixes debezium/dbz#2009

## Description
Fixes an issue where errors.max.retries was not being used. 

When a source connector enters RESTARTING after a retriable error, BaseSourceTask tries to restart the connector using start(). If start() then throws a RetriableException, like if the source DB is still unavailable, the exception propagates unhandled, and errors.max.retries has no effect on the number of restart attempts.

## PR Checklist

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/connector/common/BaseSourceTask.java
+++ b/debezium-connector-common/src/main/java/io/debezium/connector/common/BaseSourceTask.java
@@ -198,6 +198,8 @@ public abstract class BaseSourceTask<P extends Partition, O extends OffsetContex
     private final Map<Map<String, ?>, Map<String, ?>> lastOffsets = new HashMap<>();
 
     private Duration retriableRestartWait;
+    private int maxRestartRetries;
+    private int restartRetries;
 
     private final ElapsedTimeStrategy pollOutputDelay;
 
@@ -264,6 +266,8 @@ public abstract class BaseSourceTask<P extends Partition, O extends OffsetContex
                     DebeziumTaskState.INITIAL);
 
             retriableRestartWait = config.getDuration(CommonConnectorConfig.RETRIABLE_RESTART_WAIT, ChronoUnit.MILLIS);
+            maxRestartRetries = config.getInteger(CommonConnectorConfig.MAX_RETRIES_ON_ERROR);
+            restartRetries = 0;
             // need to reset the delay or you only get one delayed restart
             restartDelay = null;
             if (!config.validateAndRecord(getAllConfigurationFields(), LOGGER::error)) {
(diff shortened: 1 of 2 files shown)


Naros (commented):
While this may fix some corner cases, I'm not sure this will fix all cases.

If I set up my Kafka Connect environment with infinite retries, the problem here is that after the maximum number of restart retries is exhausted, the connector throws a hard exception that isn't retried. Obviously, a user could set Debezium's max retry to unlimited, but should users be expected to define this twice? 🤔 

While we could consider moving `start(Configuration)` to be called only during the `poll` method, hard restarts of a connector task will always go through validation, which performs a database ping. If the database is gone, validation will fail, and the connector will not start, regardless of what we do in the connector.

aindriu-aiven:
> While this may fix some corner cases, I'm not sure this will fix all cases.
> 
> If I set up my Kafka Connect environment with infinite retries, the problem here is that after the maximum number of restart retries is exhausted, the connector throws a hard exception that isn't retried. Obviously, a user could set Debezium's max retry to unlimited, but should users be expected to define this twice? 🤔
> 
> While we could consider moving `start(Configuration)` to be called only during the `poll` method, hard restarts of a connector task will always go through validation, which performs a database ping. If the database is gone, validation will fail, and the connector will not start, regardless of what we do in the connector.

Hey @Naros, 
I saw this particular issue in 3.5.2.Final as well, and I would like to contribute to help resolve it. 
Do you have a specific implementation that you think would be the best way forward. 

I would be happy to look at adding to the existing PR or creating an alternative that would allow us to utilise the retry configuration.

Thanks!

Naros:
Hi @DanGrabski could we get a rebase on `main`?

aindriu-aiven:
Hey @DanGrabski do you have time to rebase this so we can hopefully get it merged in? 🙏

DanGrabski:
> Hey @DanGrabski do you have time to rebase this so we can hopefully get it merged in? 🙏

done! Anything else I need to do on my end to get it pushed in?

aindriu-aiven:
> Hi @DanGrabski could we get a rebase on `main`?

Hey @Naros would you be able to take a look at this?

As far as I can tell existing default behaviour (unlimited retries) does not change unless the user explicitly changes `MAX_RETRIES_ON_ERROR`

debezium-connector-common/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -482,13 +486,25 @@ else if (currentState == DebeziumTaskState.RESTARTING) {
...
+                        preStart(config);
+                        this.coordinator = start(config);
+                        LOGGER.info("Successfully restarted task");
+                        restartDelay = null;
+                        restartRetries = 0;
+                        setTaskState(DebeziumTaskState.RUNNING);
+                        result = true;
+                    }
+                    catch (RetriableException e) {
+                        restartRetries++;
```
AnatolyPopov, line 500: I think we need to call `stop(true)` here, similar to the catch block in `poll()` method, to avoid a resource leak. The `RetriableException` was previously handled there, but now we're catching it here and re-throwing a      
  different exception type once retries are exhausted, so that cleanup never runs.

debezium-connector-common/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -482,13 +486,25 @@ else if (currentState == DebeziumTaskState.RESTARTING) {
...
+                        if (maxRestartRetries != ErrorHandler.RETRIES_UNLIMITED && restartRetries >= maxRestartRetries) {
```
AnatolyPopov, line 501: `restartRetries >= maxRestartRetries` (checked after incrementing) tolerates one fewer retry than `ErrorHandler.hasMoreRetries()` does for the same `errors.max.retries` value (checked before incrementing). E.g.               
  `errors.max.retries=2` allows 2 retries on the streaming-error path but only 1 here.                                                                                                                                             
                                                                                                                                                                                                                                   
  Should we do it `restartRetries > maxRestartRetries` instead, to match?

AnatolyPopov (commented):
Thanks for the PR @DanGrabski!
Left a couple of comments.

PR #7498, opened 2026-06-01
Debezium connector Maven archetype - relational non-historized debezium/dbz#1931

siddhantcvdi:
Fixes debezium/dbz#1931<the Debezium Issue Number>
## Description

- Adds `debezium-connector-archetype`, a new multi-module Maven archetype under the root POM, that lets a developer bootstrap a working Debezium source connector skeleton with a single `mvn archetype:generate` command.
- The generated project contains 17 pre-wired Java classes covering all five layers of the Debezium pipeline: Kafka Connect entry points (`SourceConnector`, `ConnectorTask`, `ConnectorConfig`), offset and source info (`Partition`, `OffsetContext`, `OffsetLoader`, `SourceInfo`, `SourceInfoStructMaker`), the change event pipeline (`DataCollectionId`, `DatabaseSchema`, `ChangeEventSourceFactory`, `ChangeRecordEmitter`, `SnapshotChangeEventSource`, `StreamingChangeEventSource`), and cross-cutting concerns (`ErrorHandler`, `EventMetadataProvider`). The skeleton compiles out of the box; change-emitting paths are left as `TODO`s for the implementer to fill in.
- Includes a Maven Archetype Integration Test (`archetype-it` module) that exercises `mvn archetype:generate` end-to-end against the local repository and verifies the generated project builds successfully.

## PR Checklist


debezium-connector-archetype/pom.xml
```diff
@@ -0,0 +1,21 @@
...
+
+    <parent>
+        <groupId>io.debezium</groupId>
+        <artifactId>debezium-parent</artifactId>
+        <version>3.6.0-SNAPSHOT</version>
+        <relativePath>../debezium-parent/pom.xml</relativePath>
+    </parent>
+
+    <artifactId>debezium-connector-archetype</artifactId>
+    <name>Debezium Connector Archetypes</name>
```
kmos: call the artifact/name Debezium Archetypes

kmos:
@siddhantcvdi the new module is causing a fail in the CI. Please take a look to the GH actions and introduce also an action for the archetype mvn module

debezium-connector-archetype/debezium-source-connector-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -0,0 +1,36 @@
...
+            </includes>
+        </fileSet>
+        <!-- Resource files: service loader descriptor, build.version, etc. -->
+        <fileSet filtered="true">
+            <directory>src/main/resources</directory>
+            <includes>
+                <include>**/*</include>
+            </includes>
+        </fileSet>
+ok    </fileSets>
```
vsantonastaso: Is this `ok` a typo?

debezium-connector-archetype/debezium-source-connector-archetype/src/main/resources/archetype-resources/src/main/java/__connectorName__SnapshotChangeEventSource.java
```diff
@@ -0,0 +1,86 @@
...
+        this.config = config;
+        this.dataCollectionId = dataCollectionId;
+        this.dispatcher = dispatcher;
+        this.progressListener = progressListener;
+    }
+
+    @Override
+    public SnapshottingTask getSnapshottingTask(${connectorName}Partition partition,
+                                                ${connectorName}OffsetContext offsetContext) {
+        boolean shouldSnapshot = offsetContext.getPosition() == 0;
```
vsantonastaso: Shouldn't we evaluate the connector configuration as well? I think we can use the SnapshotMode to control the snapshot and eventually disable it

debezium-connector-archetype/debezium-source-connector-archetype/src/main/resources/archetype-resources/src/main/java/__connectorName__OffsetContext.java
```diff
@@ -0,0 +1,61 @@
...
+    }
+
+    @Override
+    public void event(DataCollectionId dataCollectionId, Instant instant) {
+        // Update sourceInfo with the current position before each event is enqueued.
+    }
+
+    @Override
+    public TransactionContext getTransactionContext() {
+        return new TransactionContext();
```
vsantonastaso: Is it correct to create a new instance on every call to `getTransactionContext()`.
If I'm not wrong `TransactionContext` tracks transaction state across events so creating a fresh one per call means transaction metadata is lost between events. WDTY?

debezium-connector-archetype/debezium-source-connector-archetype/src/main/resources/archetype-resources/src/main/java/__connectorName__ChangeEventSourceFactory.java
```diff
@@ -0,0 +1,46 @@
...
+class ${connectorName}ChangeEventSourceFactory
+        implements ChangeEventSourceFactory<${connectorName}Partition, ${connectorName}OffsetContext> {
+
+    private final ${connectorName}ConnectorConfig config;
+    private final ${connectorName}DataCollectionId dataCollectionId;
+    private final EventDispatcher<${connectorName}Partition, ${connectorName}DataCollectionId> dispatcher;
+
+    ${connectorName}ChangeEventSourceFactory(${connectorName}ConnectorConfig config,
+                                              ${connectorName}DataCollectionId dataCollectionId,
+                                              EventDispatcher<${connectorName}Partition, ${connectorName}DataCollectionId> dispatcher) {
```
Naros: We should also pass the ErrorHandler here, since you typically need it inside your streaming change event source class to propagate streaming errors to the runtime poll loop.

This also means there should be a `${connectorName}ErrorHandler` class created that extends the base `ErrorHandler` class.

debezium-connector-archetype/debezium-source-connector-archetype/src/main/resources/archetype-resources/src/main/java/__connectorName__ChangeRecordEmitter.java
```diff
@@ -0,0 +1,72 @@
...
+import io.debezium.schema.DataCollectionSchema;
+
+/**
+ * Emits a pre-parsed row as a Debezium change record.
+ *
+ * <p>Holds the key and value {@link Struct}s built by the snapshot or streaming source
+ * and hands them to the framework's {@link io.debezium.pipeline.EventDispatcher} via the
+ * {@link ChangeRecordEmitter.Receiver} callback.
+ */
+class ${connectorName}ChangeRecordEmitter implements ChangeRecordEmitter<${connectorName}Partition> {
```
Naros: This raises a question for me as to whether or not we should model this off the relational classes or not. 

There is a lot of heavy lifting that the `RelationalChangeRecordEmitter` provides OOTB here, and a user creating a new connector ideally should either use it or `AbstractChangeRecordEmitter`. 

In either case, I'm not sure I agree with passing the `Struct` and saying that the snapshot/streaming sources generate the `Struct` objects. This is generally the responsibility of the emitter, which is supplied with the `DataCollectionSchema` during `emitChangeRecords`, and that is used to take the raw values and create the `Struct` types.

See `RelationalChangeRecordEmitter`.

This highlights a discussion I had with @kmos that we should probably focus on creating dual-archetype setups: 1 for relational and 1 for non-relational. My concern, which I expressed, is that users who use the archetype for relational database connector creation will have to do a bit of rework for it to be useful, and I don't believe that's overly helpful.

WDYT?

Naros (commented):
I've left an inline comment about having dual archetypes.

debezium-connector-archetype/debezium-source-connector-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
siddhantcvdi: Yes it's a typo🥲

debezium-connector-archetype/debezium-source-connector-archetype/src/main/resources/archetype-resources/src/main/java/__connectorName__OffsetContext.java
siddhantcvdi: You're right..Fixed by storing TransactionContext as a field initialized once in the constructor, so the same instance is returned on every call and transaction state is preserved across events.

debezium-connector-archetype/debezium-source-connector-archetype/src/main/resources/archetype-resources/src/main/java/__connectorName__SnapshotChangeEventSource.java
siddhantcvdi: Hmm..The logic ignored the SnapshotMode config entirely. Fixed to delegate to `config.getSnapshotMode().shouldSnapshotData(offsetExists, false)`, so users can control snapshot behavior via the `snapshot.mode` config option.

debezium-connector-archetype/debezium-source-connector-archetype/src/main/resources/archetype-resources/src/main/java/__connectorName__ChangeEventSourceFactory.java
siddhantcvdi: Added ErrorHandler as a constructor parameter on both ChangeEventSourceFactory and StreamingChangeEventSource, and updated the task to pass it through. The ${connectorName}ErrorHandler class extending the base ErrorHandler was already present.

debezium-connector-archetype/debezium-source-connector-archetype/src/main/resources/archetype-resources/src/main/java/__connectorName__ChangeRecordEmitter.java
kmos: @Naros, in your opinion, which archetype should we prioritize if we aim to develop two?

siddhantcvdi: @Naros, I have addressed the Struct construction issue. The emitter now extends AbstractChangeRecordEmitter and takes raw source data, building Structs itself using the DataCollectionSchema supplied at emit time.             
                                                                                                                                                                                                                                   Should I extend the current PR to add a relational archetype as a second module, or is that a separate concern?

Naros: @kmos, that's a good question :thinking: 

If we look at the most recent connector additions to Debezium's portfolio, those were relational. Even today, the dominant use case for CDC is moving relational data into analytics, reporting, caches, search indexes, and for microservice communication layers with outbox/saga patterns.

However, if you look at the number of databases, non-relational databases outpace relational databases at 1.5 to 1. So if you go off that specific metric alone, one might think non-relational or more generic makes more sense.

My personal opinion, I still think relational makes the most sense. @jpechane wdyt?

jpechane: For me I'd say relational first by far. There is a stronger knowledge base around it and then there is common snapshot logic.
Generic connector can follow as the second one.
Also to keep the matter open -historized or non-historized? I'd argue there should be there flvour is order of priority

* relational non-historized
* generic
* relational historized

debezium-connector-archetype/debezium-source-connector-relational-archetype/src/main/resources/archetype-resources/pom.xml
```diff
@@ -0,0 +1,95 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
```
mfvitale, line 5: Should we add the `debezium-parent`?

debezium-connector-archetype/debezium-source-connector-archetype/src/main/resources/archetype-resources/src/main/java/__connectorName__ChangeRecordEmitter.java
kmos: ok I think that we can proceed in this way: @siddhantcvdi , Could you please update this pull request so that it targets a relational non-historized archetype? Please adjust the code, the pull request description, and the related issue accordingly. I will open two new issues for the generic and relational historized archetypes.

siddhantcvdi: > ok I think that we can proceed in this way: @siddhantcvdi , Could you please update this pull request so that it targets a relational non-historized archetype? Please adjust the code, the pull request description, and the related issue accordingly. I will open two new issues for the generic and relational historized archetypes.

Sure @kmos, I'll make those changes✌🏽

I'd also be happy to take on the new archetypes once the issues are created, if that works for you.

debezium-connector-archetype/debezium-source-connector-relational-archetype/src/main/resources/META-INF/maven/archetype-metadata.xml
```diff
@@ -0,0 +1,36 @@
...
+                      name="debezium-source-connector-archetype">
+
+    <requiredProperties>
+        <!-- PascalCase connector name used as class-name prefix (e.g. "MyDb") -->
+        <requiredProperty key="connectorName">
+            <defaultValue>MyConnector</defaultValue>
+        </requiredProperty>
+        <!-- Debezium version to target -->
+        <requiredProperty key="debeziumVersion">
+            <defaultValue>3.6.0.Final</defaultValue>
```
Naros, line 15: Should this use `${project.version}` or do we intend to apply a script to update this per final release so that the default value does not go stale over time?

debezium-connector-archetype/debezium-source-connector-relational-archetype/src/main/resources/archetype-resources/src/main/java/__connectorName__EventMetadataProvider.java
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * <p>Implement {@link #getEventTimestamp} and {@link #getEventSourcePosition} with
+ * values meaningful to your data source.
+ */
+public class ${connectorName}EventMetadataProvider implements EventMetadataProvider {
+
+    @Override
+    public Instant getEventTimestamp(DataCollectionId source, OffsetContext offset,
+                                     Object key, Struct value) {
+        return Instant.now();
```
Naros, line 29: There are a host of other `TODO` markers throughout the code, and I wonder if it makes sense to add one here to indicate that `Instant.now()` is a placeholder so it isn't overlooked?

debezium-connector-archetype/debezium-source-connector-relational-archetype/src/main/resources/archetype-resources/src/main/java/__connectorName__SourceInfo.java
```diff
@@ -0,0 +1,37 @@
...
+    private final CommonConnectorConfig config;
+
+    public ${connectorName}SourceInfo(CommonConnectorConfig config) {
+        super(config);
+        this.config = config;
+    }
+
+    @Override
+    protected Instant timestamp() {
+        return Instant.now();
```
Naros, line 30: There are a host of other `TODO` markers throughout the code, and I wonder if it makes sense to add one here to indicate that `Instant.now()` is a placeholder so it isn't overlooked?

debezium-connector-archetype/pom.xml
```diff
@@ -0,0 +1,21 @@
+<?xml version="1.0"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>io.debezium</groupId>
+        <artifactId>debezium-parent</artifactId>
+        <version>3.6.0-SNAPSHOT</version>
```
jpechane, line 9: Please change here and in the other place to 3.7.0-SNAPSHOT.

debezium-connector-archetype/debezium-source-connector-relational-archetype/pom.xml
```diff
@@ -0,0 +1,59 @@
...
+            <testResource>
+                <directory>src/test/resources</directory>
+                <filtering>true</filtering>
+            </testResource>
+        </testResources>
+        <extensions>
+            <extension>
+                <groupId>org.apache.maven.archetype</groupId>
+                <artifactId>archetype-packaging</artifactId>
+                <version>3.2.1</version>
```
jpechane, line 29: No versions should be hardcoded. Ideally they should go into archetype parent pom as properties and if possible dependency/plugin management.

PR #7509, opened 2026-06-05
debezium/dbz#2032 Assign history topic partition during recovery instead of subscribing

100maygupta:
Fixes debezium/dbz#2032

### Problem
- `KafkaSchemaHistory` recovery uses a consumer-group `subscribe()` to read the single-partition history topic; `group.id` is the schema-history name.
- During a multi-worker Connect rebalance, multiple recovery instances of the same connector join the same group to read the single partition. Kafka assigns the partition to only one consumer; the others get an empty assignment (`partitions=[]`).
- The unassigned consumers poll zero records, exhaust `recovery.attempts`, and the task fails with `The database schema history couldn't be recovered.`

### Fix
- Replace `subscribe()` with `assign()` on the single partition followed by `seekToBeginning()`.
- Recovery is a full single-partition read that needs no group coordination, so pinning the consumer to the partition removes the group-assignment contention entirely - overlapping recovery instances can each read independently.
- Fix lives in the shared `debezium-storage-kafka` module, so it applies to all connectors that do schema history recovery from Kafka.

### Test Plan
New test in `AbstractKafkaSchemaHistoryTest`
- `shouldRecoverWhenAnotherConsumerHoldsTheGroupPartition` - populates the history topic, has a separate consumer join the recovery group (`group.id == history name`) and claim the partition, then runs recovery and asserts the schema is fully recovered. Fails on `subscribe()`, passes with `assign()`.

## PR Checklist

### DIFF
--- a/debezium-connector-binlog/src/test/java/io/debezium/relational/history/AbstractKafkaSchemaHistoryTest.java
+++ b/debezium-connector-binlog/src/test/java/io/debezium/relational/history/AbstractKafkaSchemaHistoryTest.java
@@ -11,15 +11,19 @@ import static org.junit.jupiter.api.Assertions.assertFalse;
 import static org.junit.jupiter.api.Assertions.assertThrows;
 import static org.junit.jupiter.api.Assertions.assertTrue;
 
+import java.time.Duration;
+import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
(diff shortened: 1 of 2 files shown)


vjuranek (approved):
LGTM

jpechane:
@100maygupta Applied, thanks!

PR #7502, opened 2026-06-02
debezium/dbz#1849 Rework connection factory for read-only connections

Naros:
Fixes debezium/dbz#1849

## Description
There are multiple use cases where a read-only connection should be possible, including against the primary connection when the user does not have write access, or to a physical standby where the database is not in write mode.

Notable Changes:
* `internal.log.mining.read.only` promoted to `log.mining.read.only`
* `log.mining.readonly.hostname` changed to `database.standby.hostname`
* New `database.standby.url` (allowing TNS connection-string configurations)
* New `database.standby.port` (allowing standby connections to different port configurations)
* Dedicated connection factories depending on configuration (Standalone vs PhysicalStandby)
* Dedicated OracleJdbcConfiguration class to handle Oracle-specific standby configurations

## PR Checklist

- [x] I have read the [contribution guidelines](https://github.com/debezium/debezium/blob/main/CONTRIBUTING.md) and the [governance document](https://github.com/debezium/governance/blob/main/GOVERNANCE.md) on PR expectations.
- [x] Minimal changes to code not directly related to your change (e.g. no unnecessary formatting changes or refactoring to existing code)
- [x] One feature/change per PR unless tightly coupled
- [x] Do a rebase on upstream `main`

PR #7510, opened 2026-06-05
debezium/dbz#1868 Oracle XStream Downstream Mining Support

### LINKED ISSUE #1868: Support downstream mining across all Oracle adapters

Currently, there is a standby/downstream mining solution built for LogMiner, but no such mechanism exists for XStream, because XStream is currently designed to connect and operate against the primary configured host, with no way to obtain snapshot or DDL metadata from the primary.

This forces XStream into a primary-server deployment mode only. To accomplish this shift in topology support, several changes are needed.

### Pass `OracleConnectionFactory` and not `OracleConnection`

We have built the `OracleConnectionFactory` as the go-to place to obtain the main, new, and read-only connections. While we can pass these connections separately to various classes, in practice, we should aim to pass the factory around and let it hold those references. This provides a number of benefits:

1. A single class can obtain the writable and read-only connection as needed with one argument.
2. New standalone connections can be opened without having to pass around JDBC configuration objects.

This change should be applied across all adapters (OLR/XStream/LM).

### Modify XStream

In XStream, we'd need to modify the code so that all Outbound Server communication occurs over the downstream mining connection, where we capture changes, while all other operations, such as DDL queries and snapshots, are directed to the primary connection, where the database allows writes.

## What benefits does this feature provided?

The main benefit for users is that they can now move all mining operations to a standalone environment where changes are merely replayed for analytics, business intelligence, and CDC purposes. This avoids the CDC mining load on production, and allows for DBAs to tune the downstream environment without concern to user workloads.

Naros:
Fixes debezium/dbz#1868

## Description
Support for downstream mining of real-time changes, compatible with XStream. LogMiner will be introduced in a future pass after we address all required adapter code changes.

## PR Checklist

This builds on top of the work from #7502 for physical standbys.

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -250,9 +250,11 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             .withEnum(CaptureMode.class, CaptureMode.PRIMARY)
             .withWidth(Width.MEDIUM)
             .withImportance(Importance.HIGH)
+            .withValidation(OracleConnectorConfig::validateCaptureMode)
             .withDescription("Specifies the capture mode used to capture streaming changes from Oracle" +
                     "'primary' (the default) captures changes from the primary, specified by database.* configurations, " +
-                    "'physical_standby' captures changes from a read-only physical standby, specified by secondary.* configurations.");
+                    "'physical_standby' captures changes from a read-only physical standby, specified by secondary.* configurations, " +
+                    "'downstream' captures changes from a downstream real-time mining database, specified by secondary.* configurations.");
 
     public static final Field SECONDARY_DATABASE = Field.create(OracleJdbcConfiguration.SECONDARY_DATABASE.name())
             .withDisplayName("The secondary Oracle instance database name")
@@ -2369,6 +2371,15 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
                     LOGGER.warn("The property '{}' is set, but is ignored due to using read-only mode.", RAC_NODES.name());
                 }
             }
+        }
+        return 0;
+    }
+
+    public static int validateCaptureMode(Configuration config, Field field, ValidationOutput problems) {
+        final CaptureMode captureMode = CaptureMode.parse(config.getString(CAPTURE_MODE));
+        if (CaptureMode.DOWNSTREAM.equals(captureMode) && isLogMiner(config)) {
+            problems.accept(CAPTURE_MODE, config.getString(CAPTURE_MODE), "Downstream mining is currently only supported by XStream");
+            return 1;
         }
         return 0;
     }
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jdbc/CaptureMode.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jdbc/CaptureMode.java
@@ -21,7 +21,12 @@ public enum CaptureMode implements EnumeratedValue {
     /**
      * Captures changes from a physical standby instance.
      */
-    PHYSICAL_STANDBY("physical_standby");
+    PHYSICAL_STANDBY("physical_standby"),
+
+    /**
+     * Downstream mining instance.
+     */
+    DOWNSTREAM("downstream");
 
     private final String value;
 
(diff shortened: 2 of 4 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jdbc/OracleJdbcConfiguration.java
```diff
@@ -0,0 +1,189 @@
...
+        return new OracleJdbcConfigurationBuilder(Configuration.create());
+    }
+
+    static Builder copy(Configuration config) {
+        return new OracleJdbcConfigurationBuilder(Configuration.copy(config));
+    }
+
+    static OracleJdbcConfiguration adaptWithSubset(Configuration config) {
+        return OracleJdbcConfiguration.adapt(config.subset(ConfigurationNames.DATABASE_CONFIG_PREFIX, true)
+                .merge(config.subset(DRIVER_CONFIG_PREFIX, true))
```
jpechane, line 83: How does this work with draive passthrough properties? Are they shared between primary and seconday connectotions or primary and seconday has its own set?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/AbstractLogMinerStreamingChangeEventSource.java
```diff
@@ -903,34 +908,37 @@ protected Scn calculateUpperBounds(Scn lowerBoundsScn, Scn currentScn) throws SQ
...
-            minOpenRedoThreadLastScn = minOpenRedoThreadLastScn.subtract(
-                    Scn.valueOf(connectorConfig.getLogMiningRedoThreadScnAdjustment()));
-
-            if (minOpenRedoThreadLastScn.compareTo(result) < 0) {
-                // There are situations where on first start-up that the startScn may be higher
-                // than the last flushed redo thread SCN, in which case we should delay by one
-                // iteration until the startScn is before the minOpenRedoThreadLastScn
-                if (minOpenRedoThreadLastScn.compareTo(lowerBoundsScn) < 0) {
-                    return Scn.NULL;
+        if (!(connectionFactory instanceof DualOracleConnectionFactory)) {
```
jpechane: Show we base condition on class compaison?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -225,6 +227,16 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             .withDefault(ARCHIVE_LOG_ONLY_POLL_TIME.toMillis())
             .withDescription("The interval in milliseconds to wait between polls checking to see if the SCN is in the archive logs.");
 
+    public static final Field LOG_MINING_READ_ONLY = Field.create("log.mining.read.only")
```
jpechane, line 230: Do we need separate config option?

jpechane (commented):
@Naros Nice job, thanks! I left couple of comments for clarification, when addressed we are good to go.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/jdbc/OracleJdbcConfiguration.java
Naros: I honestly hadn't considered that, so right now they're shared.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
Naros: For this yes, when you are using the `primary` mode, the primary can be writable or even read-only where the user itself has not been given any write permissions, but the database is effectively open and writable. So this option restores the original intent of having Debezium not create a flush table and not writing writing to the database.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/AbstractLogMinerStreamingChangeEventSource.java
Naros: This was an oversight before adding the config option, so we can use it instead.

vjuranek (approved):
LGTM

Naros:
@jpechane is this good to merge from your PoV?

jpechane:
@Naros Applied, thanks!

PR #7521, opened 2026-06-10
debezium/dbz#2053 Add pull request safety check steps

### LINKED ISSUE #2053: Guard against malicious and unwanted files in PRs
Introduce `step-security/file-existence-action` to handle checking for specific file patterns and failing the PR, along with using another action, `dcondrey/unicode-safety-check`, to check for unicode exploits.

Naros:
Fixes debezium/dbz#2053

## Description
Adds several security checks for unwanted files/directories and Unicode exploits.

## PR Checklist

## Open Tasks
- [x] Move the checks to `dbz` to be shared across all repositories

### DIFF
--- a/.github/workflows/debezium-workflow-pr.yml
+++ b/.github/workflows/debezium-workflow-pr.yml
@@ -26,11 +26,44 @@ jobs:
     steps:
       - uses: debezium/dbz/.github/actions/check-dco@main
 
+  # Checks that the PR does not contain forbidden files or directories
+  check-forbidden-files:
+    name: "Check for forbidden files"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Action
+        uses: actions/checkout@v6
+
+      - name: Check for forbidden directories
+        id: forbidden
+        uses: step-security/file-existence-action@v3
+        with:
+          files: ".claude, .vscode"
+
+      - name: Reject forbidden files
+        if: steps.forbidden.outputs.files_exists == 'true'
+        run: |
+          echo "::error::PR contains .claude/ or .vscode/ directories which are not permitted."
+          exit 1
+
+  # Checks for malicious or adversarial Unicode patterns in changed files
+  check-unicode-safety:
+    name: "Check for malicious Unicode"
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Action
+        uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
+
+      - name: Unicode safety check
+        uses: dcondrey/unicode-safety-check@v2
+
   # This job is responsible for inspecting the changes in the repository and setting specific output variables
   # that subsequent jobs can use to determine whether a particular job should be skipped or included.
   file_changes:
     name: "Detect repository changes"
-    needs: [ check-signed-commits ]
+    needs: [ check-signed-commits, check-forbidden-files, check-unicode-safety ]
     runs-on: ubuntu-latest
     outputs:
       ai-changed: ${{ steps.changed-files-ai.outputs.any_changed }}
--- a/.github/workflows/debezium-workflow-pr.yml
+++ b/.github/workflows/debezium-workflow-pr.yml
@@ -32,6 +32,8 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: debezium/dbz/.github/actions/check-hidden-files@main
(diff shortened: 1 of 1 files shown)


.github/workflows/debezium-workflow-pr.yml
```diff
@@ -26,11 +26,44 @@ jobs:
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout Action
+        uses: actions/checkout@v6
+
+      - name: Check for forbidden directories
+        id: forbidden
+        uses: step-security/file-existence-action@v3
+        with:
+          files: ".claude, .vscode"
```
vjuranek: Add also `.bob`? Not sure if it can be abused for any kind of attacks, but better to be careful (and we have it in `.gitignore`, so it can be deduced at least someone uses it)

kmos: I think using a whitelist instead of a blacklist is an approach worth evaluating: there are a few hidden directories that we should allow to be committed.

AlvarVG: Well. We can also add .idea, .eclipse and so on. So i guess what @kmos is proposing makes more sense.

kmos: exactly, all these hidden dirs are source of executables

Naros: I think we should add `.bob`, `eclipse`, and others as suggested, but I would also say that we likely should make these subdirectories, e.g. `**/.claude`. Right now, it only matches the repository root? wdyt?

> I think using a whitelist instead of a blacklist is an approach worth evaluating: there are a few hidden directories that we should allow to be committed.

Wouldn't a whitelist be more overall maintenance?

Naros: Also whatever we land on, this likely should be moved to the `dbz` repository to be shared across all repos, and then change the repo's main CI jobs to be workflow dispatches post security check. Just adding that here so I don't forget.

kmos: > Also whatever we land on, this likely should be moved to the `dbz` repository to be shared across all repos, and then change the repo's main CI jobs to be workflow dispatches post security check. Just adding that here so I don't forget.

Yeah, totally agree

kmos: > I think we should add `.bob`, `eclipse`, and others as suggested, but I would also say that we likely should make these subdirectories, e.g. `**/.claude`. Right now, it only matches the repository root? wdyt?
> 
> 
> 
> > I think using a whitelist instead of a blacklist is an approach worth evaluating: there are a few hidden directories that we should allow to be committed.

For sure we need to take under control also the subdirectories.
> 
> 
> Wouldn't a whitelist be more overall maintenance?

What do you mean ?

Naros: > > Wouldn't a whitelist be more overall maintenance?
> 
> What do you mean ?

To your original point, I took it that you were saying we should specify what is allowed rather than trying to define a list of what isn't. But in this case, that would mean we'd need to be explicit about every directory in the project (every submodule, package, etc). That becomes cumbersome and maintenance-heavy, rather than defining a set of obvious offenders.

Maybe I misunderstood?

jpechane: @Naros I think that @kmos meant whitelist for hidden directories and probably files starting with dot.
There should be very limited number of them so I don't see it as burden.

kmos: > > > Wouldn't a whitelist be more overall maintenance?
> > 
> > 
> > What do you mean ?
> 
> To your original point, I took it that you were saying we should specify what is allowed rather than trying to define a list of what isn't. But in this case, that would mean we'd need to be explicit about every directory in the project (every submodule, package, etc). That becomes cumbersome and maintenance-heavy, rather than defining a set of obvious offenders.
> 
> Maybe I misunderstood?

yeah sorry as @jpechane mentioned, I mean a whitelist for hidden directories :D

Naros: Ok, all these changes are applied in https://github.com/debezium/dbz/pull/2482 and this PR now focuses on incorporating those composite actions in the PR workflow.

kmos (approved):
🚀

PR #7522, opened 2026-06-10
debezium/dbz#2055 Revert Fix read-only snapshotOnlyWithRestart

vjuranek:
Fixes debezium/dbz#2055

This reverts commit c4870265f43c54ca03ec952bbf2b3a668cd38d74.

Claude suggests that the test hangs due to heartbeats (as a result of reusing thread-unsafe JDBC connection). This explanation seems to be plausible. Moreover, we started to observe hanged test short after merging c4870265 which added heartbeats into the test.

Summary of the root cause from Claude:

    ReadOnlyIncrementalSnapshotIT hangs due to thread-unsafe shared JDBC connection.

    The signal processor thread and streaming thread both use the same
    mainConnection instance returned by `DefaultMainConnectionProvidingConnectionFactory`.
    When a snapshot signal arrives, the signal processor calls `readChunk()`
    (executing `SELECT` queries and `pg_current_snapshot()`) while the streaming
    thread concurrently calls `connection.commit()` on the same connection.
    This race corrupts the JDBC connection state, causing subsequent snapshot
    operations to hang. The hang is sustained because heartbeat records
    (when enabled) keep resetting the test's empty-poll counter, preventing
    the test from ever timing out.

Aforemention timeout should eventually happen in
`AbstractSnapshotTest#consumeMixedWithIncrementalSnapshot` which uses `for (;;)` without any timeout or any other guardrails (and which is another place for possible improvements).

## PR Checklist

### DIFF
--- a/debezium-embedded/src/test/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotTest.java
+++ b/debezium-embedded/src/test/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotTest.java
@@ -42,7 +42,6 @@ import io.debezium.config.CommonConnectorConfig;
 import io.debezium.config.Configuration;
 import io.debezium.data.Envelope;
 import io.debezium.doc.FixFor;
-import io.debezium.heartbeat.Heartbeat;
 import io.debezium.jdbc.JdbcConnection;
 import io.debezium.junit.EqualityCheck;
 import io.debezium.junit.SkipWhenConnectorUnderTest;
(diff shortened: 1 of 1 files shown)


vjuranek:
Now `snapshotOnlyWithRestart` fails again :-) Looks like we need to PG connection issue.

vjuranek:
AI was wrong, the root cause seems to be completely different (still to be confirmed, but definitely it has nothing to do with corrupted SQL connection). Removing heartbeats doesn't help anything (besides not running tests indefinitely).

PR #7527, opened 2026-06-12
debezium/dbz#2045 Map varbit(1) as boolean

jpechane:
Fixes debezium/dbz#2045


vjuranek (approved):
LGTM, test failures seem to be unrelated.

PR #7529, opened 2026-06-15
debezium/dbz#2075 Fail-fast on NanoTimestamp overflow instead of silent wrap

minleejae:
Fixes debezium/dbz#2075

## Description

When `time.precision.mode=nanoseconds` is selected, `io.debezium.time.NanoTimestamp.toEpochNanos` silently wraps to a negative `long` for any value outside the INT64-representable range (approximately **1677-09-21 to 2262-04-11**). End-of-time sentinels like `9999-12-31 23:59:59` (extremely common in production) are silently corrupted into year **1816**, and `2286-11-20` becomes year **1702**, with no error or warning surfaced anywhere in the pipeline. Full reproduction matrix and root-cause analysis are in debezium/dbz#2075.

This PR makes the converter **fail-fast** instead of silently corrupting:

- `Math.multiplyExact` is used in `toEpochNanos(LocalDate)` for the `epochDay * NANOSECONDS_PER_DAY` step.
- `Math.addExact` is used in `toEpochNanos(LocalDateTime)` for the `nanosOfDay + nanoInDay` step.
- Either overflow is converted to an `IllegalArgumentException` whose message names the offending value and points the user to `time.precision.mode=microseconds` or `isostring` as supported alternatives.

`IllegalArgumentException` thrown from a value converter already routes through Debezium's existing `event.processing.failure.handling.mode` framework (`fail` / `warn` / `skip`), so operators retain full control over the failure mode without any new option surface.

For comparison with prior art:
- `java.time.Instant.toEpochMilli()` / `Instant.toEpochNano()` throw `ArithmeticException` on overflow.
- pandas raises `OutOfBoundsDatetime` at exactly the same boundary (1677~2262).
- ClickHouse documents the limit explicitly: *"If the maximum precision of 9 digits (nanoseconds) is used, the maximum supported value is `2262-04-11 23:47:16` in UTC."*

### Tests

New `NanoTimestampTest` (JUnit 5 + AssertJ, mirroring the style of the existing `NanoDurationTest`) covering:

- **In-range values**: epoch (`1970-01-01T00:00:00Z` → `0`), mid-range (`2001-09-09T01:46:40Z` → `1_000_000_000_000_000_000L`), and the exact upper boundary `2262-04-11T23:47:16.854775807Z` → `Long.MAX_VALUE`.
- **Out-of-range values**: `9999-12-31` (the sentinel that motivated this fix), `1000-01-01` (well below the 1677 lower bound), one-nanosecond-past the upper boundary, and `2286-11-20` (the additional value documented in debezium/dbz#2075).

Each out-of-range test asserts both the `IllegalArgumentException` type and that the cause chain preserves the original `ArithmeticException`.

Local verification:
```
./mvnw -pl debezium-connector-common test
[INFO] Tests run: 394, Failures: 0, Errors: 0, Skipped: 0
[INFO] BUILD SUCCESS
```
394 tests pass — the 5 new tests plus the entire existing suite (including `JdbcValueConvertersTemporalPrecisionTest`, which exercises the `nanoseconds` mode end-to-end) green with no regression.

## PR Checklist

- [x] Minimal changes to code not directly related to your change (single class touched + one new test class)
- [x] Do a rebase on upstream `main` (branched from current `main` head)
- [x] Commit signed off (DCO)


minleejae:
The CI failures all stem from a single ConditionTimeout in `ApicurioRegistryTestIT.shouldNotErrorWithBadHeader`, which appears unrelated to this change (this PR touches only `NanoTimestamp` and                                                                                            documentation; the test exercises a deliberately-bad SSL truststore                                                                                                                                                           and does not use `time.precision.mode` or any nanosecond conversion).

debezium-connector-common/src/test/java/io/debezium/time/NanoTimestampTest.java
```diff
@@ -0,0 +1,72 @@
...
+
+import java.time.LocalDate;
+import java.time.LocalDateTime;
+
+import org.junit.jupiter.api.Test;
+
+public class NanoTimestampTest {
+
+    @Test
+    public void shouldConvertWithinRangeValues() {
```
Naros, line 22:
```suggestion
    @Test
    @FixFor("debezium/dbz#2075")
    public void shouldConvertWithinRangeValues() {
```
Lets add the `FixFor` annotation for this and other methods here.

Naros (commented):
Hi @minleejae LGTM, I've left just one nit-pick wrt tests.

vjuranek (approved):

vjuranek:
@minleejae please rebase on the latest `main` so it has the fix for Apicurio tests (https://github.com/debezium/debezium/pull/7533) and we can re-schedule the tests. Thanks!

Naros (approved):

jpechane:
@minleejae Applied, thanks!

PR #7531, opened 2026-06-15
debezium/dbz#1554 Fix data loss on restart when multiple events share the same LSN

### LINKED ISSUE #1554: Postgres Source with jdbc copy will lost data.
## Bug report

3.4

---

default config

---

any version

When insert with jdbc copy( using mutl-insert will merger multiple message into one lsn is always true: 

Image

For example:
```sql
psql -h 127.0.0.1 -p5438 -U postgres -d postgres  -c "\copy public.lsn_test(id,val) FROM '//mnt/disk1/wudi/test.csv' WITH (FORMAT csv, HEADER false)
```

Image

We can see the lsn of multiple message is same. While isLastEventForLsn of PgOutputReplicationMessage is always true. If the job is stopped at the first one, then restarted. Data will be lost.

---

**What behavior do you see?**

Postgres Source loses data.

---

`<Your answer>`

---

`<Your answer>`

---

psql -h 127.0.0.1 -p5438 -U postgres -d postgres  -c "\copy public.lsn_test(id,val) FROM '//mnt/disk1/wudi/test.csv' WITH (FORMAT csv, HEADER false)

ganeshannt:
Fixes debezium/dbz#1554

## Description

When using PostgreSQL COPY or batched multi-row inserts, multiple DML events can share
the same LSN within a transaction. Previously, PgOutputReplicationMessage.isLastEventForLsn()
always returned true, causing the connector to mark the LSN as fully processed after the
first event. On restart, remaining events at that LSN would be skipped, resulting in data loss.

This fix introduces a per-LSN event counter in the connector offsets that tracks how many
events at the current LSN have been successfully processed. On restart, WalPositionLocator
uses this counter to resume precisely after the last processed event within a same-LSN group.

Changes:
- PgOutputReplicationMessage.isLastEventForLsn() now returns false for DML events
- Added lsn_events_processed field to PostgresOffsetContext (backward compatible, defaults to 0)
- Updated WalPositionLocator to use LSN + event count for resume point calculation
- PostgresStreamingChangeEventSource increments counter per DML, resets on COMMIT
- Added unit tests for same-LSN multi-event restart scenarios

## PR Checklist

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresOffsetContext.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresOffsetContext.java
@@ -37,6 +37,7 @@ public class PostgresOffsetContext extends CommonOffsetContext<SourceInfo> {
 
     public static final String LAST_COMPLETELY_PROCESSED_LSN_KEY = "lsn_proc";
     public static final String LAST_COMMIT_LSN_KEY = "lsn_commit";
+    public static final String LSN_EVENTS_PROCESSED_KEY = "lsn_events_processed";
 
     private final Schema sourceInfoSchema;
     private boolean lastSnapshotRecord;
@@ -45,6 +46,8 @@ public class PostgresOffsetContext extends CommonOffsetContext<SourceInfo> {
     private Lsn streamingStoppingLsn = null;
     private final TransactionContext transactionContext;
     private final IncrementalSnapshotContext<TableId> incrementalSnapshotContext;
+    private long lsnEventsProcessed;
+    private Lsn previousLsn;
 
     private PostgresOffsetContext(PostgresConnectorConfig connectorConfig, Lsn lsn, Lsn lastCompletelyProcessedLsn, Lsn lastCommitLsn, Long txId, Operation messageType,
                                   Instant time,
@@ -99,6 +102,9 @@ public class PostgresOffsetContext extends CommonOffsetContext<SourceInfo> {
         }
         if (sourceInfo.messageType() != null) {
             result.put(SourceInfo.MSG_TYPE_KEY, sourceInfo.messageType().toString());
+        }
+        if (lsnEventsProcessed > 0) {
+            result.put(LSN_EVENTS_PROCESSED_KEY, lsnEventsProcessed);
         }
         return sourceInfo.isSnapshot() ? result : incrementalSnapshotContext.store(transactionContext.store(result));
     }
@@ -176,6 +182,33 @@ public class PostgresOffsetContext extends CommonOffsetContext<SourceInfo> {
         return sourceInfo.messageType();
     }
 
+    public long lsnEventsProcessed() {
+        return lsnEventsProcessed;
+    }
+
+    /**
+     * Increments the per-LSN event counter. If the LSN has changed since the last call,
+     * the counter resets to 1. This enables precise restart recovery when multiple events
+     * share the same LSN (e.g. COPY or batched inserts).
+     */
+    public void incrementLsnEventsProcessed(Lsn currentLsn) {
+        if (previousLsn != null && previousLsn.equals(currentLsn)) {
+            lsnEventsProcessed++;
+        }
+        else {
+            lsnEventsProcessed = 1;
+            previousLsn = currentLsn;
+        }
+    }
+
+    /**
+     * Resets the per-LSN event counter, typically called on transaction commit
+     * when the LSN is fully processed.
+     */
+    public void resetLsnEventsProcessed() {
+        lsnEventsProcessed = 0;
+    }
+
     /**
      * Returns the LSN that the streaming phase should stream events up to or null if
      * a stopping point is not set. If set during the streaming phase, any event with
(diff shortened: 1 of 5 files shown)


debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/junit/WALPositionLocatorTest.java
```diff
@@ -344,6 +344,82 @@ void testToStringAfterProcessingMessages() {
                 .contains("startStreamingLsn=" + lsn);
     }
 
+    // =============================================================================
+    // Same-LSN Event Counter Tests (dbz#1554)
+    // =============================================================================
+
+    @Test
+    void whenMultipleEventsShareSameLsnShouldResumeAfterProcessedCount() {
```
Naros, line 350:
```suggestion
    @Test
    @FixFor("debezium/dbz#1554")
    void whenMultipleEventsShareSameLsnShouldResumeAfterProcessedCount() {
```

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/junit/WALPositionLocatorTest.java
```diff
@@ -344,6 +344,82 @@ void testToStringAfterProcessingMessages() {
...
+        assertThat(result).isEmpty();
+
+        // Next LSN arrives - should resume from here
+        result = locator.resumeFromLsn(nextLsn, insertMessage);
+        assertThat(result).isPresent();
+        assertThat(result.get()).isEqualTo(nextLsn);
+    }
+
+    @Test
+    void whenLsnEventsProcessedIsZeroShouldReplayAllEventsAtThatLsn() {
```
Naros, line 424:
```suggestion
    @Test
    @FixFor("debezium/dbz#1554")
    void whenLsnEventsProcessedIsZeroShouldReplayAllEventsAtThatLsn() {
```

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/junit/WALPositionLocatorTest.java
```diff
@@ -344,6 +344,82 @@ void testToStringAfterProcessingMessages() {
                 .contains("startStreamingLsn=" + lsn);
     }
 
+    // =============================================================================
+    // Same-LSN Event Counter Tests (dbz#1554)
+    // =============================================================================
```
Naros: These comments are superfluous, can you remove them.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresOffsetContext.java
```diff
@@ -45,6 +46,8 @@ public class PostgresOffsetContext extends CommonOffsetContext<SourceInfo> {
     private Lsn streamingStoppingLsn = null;
     private final TransactionContext transactionContext;
     private final IncrementalSnapshotContext<TableId> incrementalSnapshotContext;
+    private long lsnEventsProcessed;
```
vjuranek, line 49: If I'm not mistaken, you don't set `lsnEventsProcessed` anywhere. How does it survive Debezium restart?

ganeshannt: Good catch for asking! It's actually persisted in getOffset() where result.put(LSN_EVENTS_PROCESSED_KEY, lsnEventsProcessed) writes it to the offset map, and restored in Loader.load() via getOrDefault(LSN_EVENTS_PROCESSED_KEY, 0L). 
default 0 keeps it backward compatible with old offsets.

vjuranek: Thanks @ganeshannt , I completely missed that.

vjuranek:
@ganeshannt please rebase also this PR on the latest main to avoid Apicurio failures. Thanks!

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresOffsetContext.java
ganeshannt: no worries 👍🏼

ganeshannt:
@vjuranek , Rebased on latest main

jpechane:
@ganeshannt Test failures seems to be related.

ganeshannt:
> @ganeshannt Test failures seems to be related.

@jpechane Updated the approach to handle this properly.

The previous approach broke `lastCompletelyProcessedLsn` tracking by changing `isLastEventForLsn()` to return false for DML events. Also, `skipMessage()` filtered by LSN only, so unprocessed events sharing the same LSN got filtered during replay.

Updated logic reverts `isLastEventForLsn()` back to true, and instead teaches `WalPositionLocator` to set the resume point at the stored LSN with a skip count. `skipMessage()` now counts down processed events before passing, so same-LSN events that weren't processed don't get lost. Tests cover both search and replay phases now. Let me know if you have any other thoughts.

ganeshannt:
@jpechane, Pushed a fix for the CI failures. issue was the skip threshold so changed from `> 0` to `> 1` in `WalPositionLocator.resumeFromLsn()`.
In pgoutput, BEGIN and the first DML can share the same LSN. With `lsnEventsProcessed = 1` (the normal single-event case), the search phase counted both as 2 events at the stored LSN, exceeded the threshold, and incorrectly activated the skip path causing one record to be lost on restart.
Fix only activates same-LSN skip counting when `lsnEventsProcessed > 1`, meaning there were genuinely multiple DML events at that LSN (COPY/batch scenario). For counter 0 or 1, the existing "resume from next different LSN" logic handles restart correctly. Added a test for the single-event case.

correct me if i missing any scenario here

ganeshannt:
@jpechane @Naros @vjuranek  - CI is green now. Ready for review when you get a chance

vjuranek (approved):
LGTM, but as it's a change with potentially big impact, let's wait for more reviews.

Naros (approved):
LGTM, just one question @ganeshannt if its possible to have an integration test in addition to the WALPositionLocator logic in isolation, or is that a bit difficult to do?

ganeshannt:
> LGTM, just one question @ganeshannt if its possible to have an integration test in addition to the WALPositionLocator logic in isolation, or is that a bit difficult to do?

@Naros, Yeah, it’s good to add integration tests too. I added edge case unit tests and an integration test as requested. Unit tests cover the boundary at lsnEventsProcessed=2, all events already processed at stored LSN, and correct skip counter decrement. Integration test inserts 20 rows in a single transaction, stops mid-batch, restarts, and verifies no data loss. Pgoutput-only.

ganeshannt:
@Naros @vjuranek - I’ve added the integration test fixes. can you trigger the CI pipeline

ganeshannt:
@vjuranek @Naros ,  All PostgreSQL, JDBC, Apicurio, and Core Component jobs pass. The only failing job is "Debezium Platform - conductor", which fails resolving debezium-schema-generator:3.6.0-SNAPSHOT while building debezium-connector-db2. This is unrelated to the changes in this PR. 
Could a maintainer retry that job?

vjuranek:
@ganeshannt Debezium Platform failure is fine, it's obviously unrelated. IIRC @Naros is fine with the change, so IMHO this is ready to be merge. However, as we are going to release 3.6.0.Final tomorrow and as I wrote, this change has potentially big impact, I'd suggest merge it right after 3.6.0.Final (later tomorrow or on Wed) and release in 3.7.0.Alpha1.

vjuranek:
Merged, thanks @ganeshannt !

PR #7533, opened 2026-06-16
debezium/dbz#2084 Use Apicurio 3 property names

jpechane:
Signed-off-by: Jiri Pechanec <jiri.pechanec@centrum.cz>

Fixes debezium/dbz#2084


vjuranek (approved):
Out of curiosity, are these options documented somewhere (as I failed to fined them in Apicurio documentation) or you have to read the [source code](https://github.com/Apicurio/apicurio-registry/blob/main/schema-resolver/src/main/java/io/apicurio/registry/resolver/config/SchemaResolverConfig.java) to learn about them?

jpechane:
@vjuranek I've found it in the source code, not sure about docs.

PR #7536, opened 2026-06-17
debezium/dbz#2046 Allow Oracle buffered transaction creation to be deferred until first DML

nathan-smit-1:
Fixes debezium/dbz#2046

## Description
The goal of this change is to reintroduce an approximation of the deferred transaction start that was first worked on here:  https://github.com/debezium/debezium/pull/5252

The PR creates a new map that stores required metadata.  A transaction stays in this deferred state until new DML is seen at which point it is promoted to a "real" transaction.  This change allows for a sliding window similar to Debezium behaviour in versions prior to 3.2.4.  It tries to approximate the previous behaviour of the internal config log.mining.buffer.memory.legacy.transaction.start but without losing metadata like user_name and clientid.

In this implementation, the lower bound of the session window does not get pinned by in-flight transactions.  The offset scn though is impacted by both deferred and true transactions.  This is to prevent metadata being lost on connector restart.  There is a configuration that allows clean up of the deferred map.  By default, it allows transactions to be stored there for 24 hours which seems like a reasonable amount of time to store in this map (in practice you'd likely want this window to be much smaller as this means that there has been no DML on the transaction in 24 hours).

In log.mining.buffer.memory.legacy.transaction.start it explicitly excluded the non-memory types.  Given that in this new mode we capture the transaction starts, I can't think of any reason not to allow infinispan and ehcache to make use of it and so the naming reflects that.

## PR Checklist

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -716,16 +716,29 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
     public static final Field SIGNAL_DATA_COLLECTION = CommonConnectorConfig.SIGNAL_DATA_COLLECTION
             .withValidation(OracleConnectorConfig::validateSignalDataCollection);
 
-    public static final Field LOG_MINING_BUFFER_MEMORY_LEGACY_TRANSACTION_START = Field.createInternal("log.mining.buffer.memory.legacy.transaction.start")
-            .withDisplayName("Use legacy transaction start behavior")
+    public static final Field LOG_MINING_BUFFER_DEFERRED_TRANSACTION_START = Field.create("log.mining.buffer.deferred.transaction.start")
+            .withDisplayName("Use deferred transaction start behavior")
             .withType(Type.BOOLEAN)
             .withWidth(Width.SHORT)
             .withImportance(Importance.LOW)
             .withDefault(false)
-            .withValidation(OracleConnectorConfig::validateIncludeTransactionStartEvents)
-            .withDescription("Controls whether transaction start events are buffered when using the heap/memory buffer type. " +
-                    "true: transaction start events are not buffered; " +
-                    "false: (the default) transaction start events are buffered");
+            .withValidation(OracleConnectorConfig::validateDeferredTransactionStart)
+            .withDescription("Controls whether transaction start events are deferred when using buffered LogMiner. " +
+                    "When enabled, transaction start events are stored in a lightweight metadata map. " +
+                    "Transactions are only promoted to the transaction cache when a DML event is observed. " +
+                    "The mining window is not pinned by transactions, allowing a block-by-block sliding window.")
+            .withDeprecatedAliases("log.mining.buffer.memory.legacy.transaction.start");
+
+    public static final Field LOG_MINING_BUFFER_DEFERRED_TRANSACTION_RETENTION_MS = Field.create("log.mining.buffer.deferred.transaction.retention.ms")
+            .withDisplayName("Deferred transaction retention")
+            .withType(Type.LONG)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.MEDIUM)
+            .withDefault(TimeUnit.HOURS.toMillis(24))
+            .withValidation(Field::isNonNegativeLong)
+            .withDescription("Duration in milliseconds to retain deferred transaction metadata for transactions that never " +
+                    "emit a DML event. By default, deferred transaction metadata is retained for 24 hours. " +
+                    "This is independent of log.mining.transaction.retention.ms which governs cached transactions with events.");
 
     public static final Field LEGACY_DECIMAL_HANDLING_STRATEGY = Field.create("legacy.decimal.handling.strategy")
             .withDisplayName("Use legacy decimal handling strategy")
@@ -863,7 +876,8 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
                     LOG_MINING_CLIENTID_INCLUDE_LIST,
                     LOG_MINING_CLIENTID_EXCLUDE_LIST,
                     LOG_MINING_RESUME_POSITION_INTERVAL_MS,
-                    LOG_MINING_BUFFER_MEMORY_LEGACY_TRANSACTION_START,
+                    LOG_MINING_BUFFER_DEFERRED_TRANSACTION_START,
+                    LOG_MINING_BUFFER_DEFERRED_TRANSACTION_RETENTION_MS,
                     LOG_MINING_PATH_DICTIONARY,
                     LOG_MINING_READONLY_HOSTNAME,
                     LEGACY_DECIMAL_HANDLING_STRATEGY,
@@ -951,6 +965,8 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
     private final boolean logMiningBufferTrackUsername;
     private final boolean logMiningBufferTrackCommitTimestamp;
     private final boolean logMiningBufferTrackStartTimestamp;
+    private final boolean logMiningDeferredTransactionStart;
+    private final Duration logMiningDeferredTransactionRetention;
 
     private final String openLogReplicatorSource;
     private final String openLogReplicatorHostname;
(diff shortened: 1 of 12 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -577,44 +623,52 @@ protected void handleRollbackEvent(LogMinerEventRow event) {
...
             getMetrics().setBufferedEventCount(getTransactionCache().getTransactionEvents());
         }
+        else if (removeDeferredTransaction(transactionId)) {
+            LOGGER.debug("Transaction {} was in deferred state, removed on rollback.", transactionId);
+        }
         else {
             LOGGER.debug("Transaction {} not found in cache, no events to rollback.", transactionId);
 
-            if (transactionId.endsWith(NO_SEQUENCE_TRX_ID_SUFFIX)) {
+            if (transactionId != null && transactionId.endsWith(NO_SEQUENCE_TRX_ID_SUFFIX)) {
```
Naros:
```suggestion
            if (!Strings.isNullOrEmpty(transactionId) && transactionId.endsWith(NO_SEQUENCE_TRX_ID_SUFFIX)) {
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -369,6 +394,15 @@ protected boolean isEventSkipped(LogMinerEventRow event) {
     protected void handleStartEvent(LogMinerEventRow event) {
         final String transactionId = event.getTransactionId();
         if (!isRecentlyProcessed(transactionId)) {
+            if (getConfig().isDeferredLogMinerTransactionStartBehaviorEnabled() && transactionId != null) {
```
Naros:
```suggestion
            if (getConfig().isDeferredLogMinerTransactionStartBehaviorEnabled() && !Strings.isNullOrEmpty(transactionId)) {
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -390,7 +424,14 @@ protected void handleCommitEvent(LogMinerEventRow row) throws InterruptedExcepti
         }
 
         final Transaction transaction = getTransactionCache().getAndRemoveTransaction(transactionId);
+        boolean removedDeferredTransaction = false;
         if (transaction == null) {
+            // Check if the transaction was deferred and never promoted
+            if (transactionId != null && deferredTransactions.remove(transactionId) != null) {
```
Naros:
```suggestion
            if (!Strings.isNullOrEmpty(transactionId) && deferredTransactions.remove(transactionId) != null) {
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -629,6 +683,30 @@ else if (matchingTransactions.size() == 1) {
         getBatchMetrics().rollbackObserved();
     }
 
+    private boolean removeDeferredTransaction(String transactionId) {
+        return transactionId != null && deferredTransactions.remove(transactionId) != null;
```
Naros:
```suggestion
        return !Strings.isNullOrEmpty(transactionId) && deferredTransactions.remove(transactionId) != null;
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/logminer/buffered/BufferedLogMinerStreamingChangeEventSource.java
```diff
@@ -1054,11 +1165,24 @@ protected void enqueueEvent(LogMinerEventRow event, LogMinerEvent dispatchedEven
 
         Transaction transaction = getTransactionCache().getTransaction(transactionId);
         if (transaction == null) {
-            LOGGER.trace("Transaction {} is not in cache, creating.", transactionId);
-            transaction = transactionFactory.createTransaction(event);
-            getTransactionCache().addTransaction(transaction);
+            // Check if this transaction is in the deferred state and promote it
+            if (transactionId != null) {
```
Naros:
```suggestion
            if (!Strings.isNullOrEmpty(transactionId)) {
```

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -727,6 +727,29 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
                     "true: transaction start events are not buffered; " +
                     "false: (the default) transaction start events are buffered");
 
+    public static final Field LOG_MINING_BUFFER_DEFERRED_TRANSACTION_START = Field.create("log.mining.buffer.deferred.transaction.start")
```
Naros, line 719: Should we remove `LOG_MINING_BUFFER_MEMORY_LEGACY_TRANSACTION_START` and instead add to this option a `deprecatedAlias` with the value:
```
internal.log.mining.buffer.memory.legacy.transaction.start
```
WDYT?

Naros (commented):
Hi @nathan-smit-1, I'm starting CI below, but wanted to give some feedback.

nathan-smit-1:
@Naros  will go through your feedback.  Am I right in saying the test failures seem unrelated?

nathan-smit-1:
@Naros Just regarding the deprecation:  I had been thinking about something along these lines (actually just removing the old configuration), but was just worried that someone might be relying on that functionality (specifically the fact that the query removes the START markers).  Maybe it's better though to focus on a single way to do deferred transactions to avoid cognitive load of tracking all these mechanisms

nathan-smit-1:
@Naros   I've pushed changes now and assuming deprecation.  Just pulled out the usages.  Let me know if that makes sense and is line with what you were thinking?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
```diff
@@ -716,16 +716,29 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
...
-                    "true: transaction start events are not buffered; " +
-                    "false: (the default) transaction start events are buffered");
+            .withValidation(OracleConnectorConfig::validateDeferredTransactionStart)
+            .withDescription("Controls whether transaction start events are deferred when using buffered LogMiner. " +
+                    "When enabled, transaction start events are stored in a lightweight metadata map. " +
+                    "Transactions are only promoted to the transaction cache when a DML event is observed. " +
+                    "The mining window is not pinned by transactions, allowing a block-by-block sliding window.")
+            .withDeprecatedAliases("log.mining.buffer.memory.legacy.transaction.start");
+
+    public static final Field LOG_MINING_DEFERRED_TRANSACTION_RETENTION_MS = Field.create("log.mining.deferred.transaction.retention.ms")
```
Naros: We use `log.mining.deferred` here, but the other option uses `log.mining.buffer.deferred`. Ideally, I would expect these two properties to carry the same namespace. Would you agree?

Naros (commented):
Hi @nathan-smit-1 LGTM, just one final nit-pick.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
nathan-smit-1: Yes you're right this was a mistake.  Initially I named it similar to the legacy config but then adjusted when I realised there was no need to focus on memory only.  Didn't rename this particular one properly.  Will correct now.

nathan-smit-1:
@Naros should be corrected now.  Both configs are in the buffer namespace which I think is correct.

Naros:
Thanks @nathan-smit-1, there were some recent changes I pushed to the Oracle connector and it's CI in the last 24 hours. So I can kick-off CI, I'm going to rebase on the latest main for you. I hope that's okay.

nathan-smit-1:
@Naros failures seem to be on my test class I will check

nathan-smit-1:
@Naros it's because of the test changes by dbz#2049 in AbstractBufferedLogMinerStreamingChangeEventSourceTest.  I'm actually not certain now about the amount of duplication between these classes.  What are your thoughts?

For now the test is updated and I ran tests locally which are now passing and I've pushed again.  Sorry about that, just a timing thing I think.  I also rebased

Naros (approved):

PR #7538, opened 2026-06-17
debezium/dbz#2099 Preserve PostgreSQL BIT and VARBIT values in JDBC sink

minleejae:
Fixes https://github.com/debezium/dbz/issues/2099

The PostgreSQL JDBC sink converted Debezium `Bits` payloads with `BigInteger(byte[])`. Debezium `Bits` uses `BitSet` byte-array semantics, while `BigInteger` interprets bytes as signed big-endian values. This could silently reverse multi-byte `BIT`/`VARBIT` values or fail for high-bit bytes such as `0xff`.

This change binds PostgreSQL bit values by decoding the payload with `BitSet.valueOf(byte[])` and emitting the bit string from the schema width. It also adds unit coverage for byte order, high-bit bytes, and bounded leading zeroes, plus PostgreSQL source-to-sink e2e coverage for `bit(8/16/24)` and `bit varying(8/16/24)`.

Testing:

* Ran `BitTypeTest` for PostgreSQL BIT/VARBIT binding, covering byte order, high-bit bytes, bounded leading zeroes, and unbounded VARBIT payloads.
* Ran PostgreSQL source-to-PostgreSQL sink e2e tests for `bit(8/16/24)` and `bit varying(8/16/24)` across the JDBC sink pipeline source option combinations.
* Built a local JDBC connector plugin and verified a Docker Compose PostgreSQL-to-PostgreSQL pipeline. The sink preserved `0000000100000010`, `000000110000001000000001`, `10000000`, and `11111111` for `BIT` and `VARBIT` columns.

### DIFF
--- a/debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/dialect/postgres/BitTypeTest.java
+++ b/debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/dialect/postgres/BitTypeTest.java
@@ -7,6 +7,7 @@ package io.debezium.connector.jdbc.dialect.postgres;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.nio.ByteBuffer;
 import java.util.List;
 
 import org.apache.kafka.connect.data.Schema;
@@ -41,6 +42,14 @@ class BitTypeTest {
 
     @Test
     @FixFor("debezium/dbz#2099")
+    @DisplayName("Should bind ByteBuffer bit values from Avro converter")
+    void testBindByteBufferBitValues() {
+        assertBindValue(Bits.schema(16), ByteBuffer.wrap(new byte[]{ 0x02, 0x01 }), "0000000100000010");
+        assertBindValue(Bits.schema(24), ByteBuffer.wrap(new byte[]{ 0x01, 0x02, 0x03 }), "000000110000001000000001");
+    }
+
+    @Test
+    @FixFor("debezium/dbz#2099")
     @DisplayName("Should bind values with the high bit set as unsigned bit strings")
     void testBindHighBitValues() {
         assertBindValue(Bits.schema(8), new byte[]{ (byte) 0x80 }, "10000000");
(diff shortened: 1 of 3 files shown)


debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/dialect/postgres/BitTypeTest.java
```diff
@@ -0,0 +1,68 @@
...
+import io.debezium.data.Bits;
+import io.debezium.sink.valuebinding.ValueBindDescriptor;
+
+/**
+ * Unit tests for the PostgreSQL {@link BitType} handler.
+ */
+@Tag("UnitTests")
+class BitTypeTest {
+
+    @Test
```
Naros, line 28: Lets be sure to include `@FixFor("debezium/dbz#2099` on the test methods

Naros (commented):
Overall LGTM, I've left one comment below.

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/dialect/postgres/BitTypeTest.java
minleejae: Thanks, applied in 7a62e13.

minleejae:
Hi @Naros 
Added an additional fix for the Confluent AvroConverter path.
During the broader PG -> PG type audit in debezium/dbz#2100, I found that BIT/VARBIT values still failed when using Schema Registry + AvroConverter because Avro `bytes` values are delivered to the sink as `ByteBuffer`, while `BitType` assumed `byte[]`.

This update changes PostgreSQL `BitType` binding to use the existing `ByteArrayUtils.getByteArrayFromValue()` helper, so both `byte[]` and `ByteBuffer` inputs are handled consistently.

jpechane (approved):

jpechane:
@minleejae Applied, thanks!

PR #7540, opened 2026-06-17
debezium/dbz#2100 Preserve PostgreSQL TIME and TIMETZ boundary values

minleejae:
## Description

Fixes debezium/dbz#2100.

PostgreSQL `time` and `time with time zone` can represent the boundary value `24:00:00`. This differs from the Java time objects used in the connector paths: `LocalTime` supports `00:00:00` through `23:59:59.999999999`, and `OffsetTime` uses `LocalTime` internally. Because of that range mismatch, PostgreSQL values that are valid at the database level can fail during conversion or be normalized to `00:00:00Z` when routed through Java time objects.

This change preserves PostgreSQL `TIME` and `TIMETZ` boundary values in the affected PostgreSQL source and JDBC sink paths.

Specifically, this PR:

* handles JDBC sink `TIME` values that represent `24:00:00` without converting them through `LocalTime`;
* preserves scalar PostgreSQL `TIMETZ 24:00:00+00` as `24:00:00Z` instead of normalizing it to `00:00:00Z`;
* preserves `TIMETZ[]` elements by reading array elements from the raw `PgArray` result set representation, avoiding `PgArray#getArray()` conversion to `java.sql.Time`, which loses the distinction between `24:00:00+00` and `00:00:00+00`.

The scope is intentionally limited to preserving PostgreSQL boundary values in the existing Debezium temporal mappings.

## Testing

Verified with focused unit and integration coverage for the affected paths:

* JDBC sink `TIME` boundary coverage for PostgreSQL target columns;
* PostgreSQL snapshot and streaming coverage for scalar `TIMETZ 24:00:00+00`;
* PostgreSQL snapshot and streaming coverage for `TIMETZ[]` boundary values;
* existing PostgreSQL time array regression coverage;
* JDBC connector module compile check.

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresTimeBoundary.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresTimeBoundary.java
@@ -1,0 +1,44 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.jdbc.dialect.postgres;
+
+import java.util.concurrent.TimeUnit;
+import java.util.regex.Pattern;
+
+/**
+ * Boundary helpers for PostgreSQL {@code TIME} values.
+ */
+final class PostgresTimeBoundary {
+
+    static final String BOUNDARY_TIME = "24:00:00";
+    static final String BOUNDARY_TIME_WITH_TIMEZONE = "24:00:00Z";
+    static final String TIME_QUERY_BINDING = "cast(? as time)";
+    static final String TIME_WITH_TIMEZONE_QUERY_BINDING = "cast(? as time with time zone)";
+
+    private static final long DAY_MILLIS = TimeUnit.DAYS.toMillis(1);
+    private static final long DAY_MICROS = TimeUnit.DAYS.toMicros(1);
+    private static final long DAY_NANOS = TimeUnit.DAYS.toNanos(1);
+    private static final Pattern BOUNDARY_TIME_WITH_TIMEZONE_PATTERN = Pattern.compile("^24:00:00(?:\\.0{1,6})?(?:Z|[+-]00(?::?00)?)$");
+
+    static boolean isBoundaryMilliseconds(Number value) {
+        return value.longValue() == DAY_MILLIS;
+    }
+
+    static boolean isBoundaryMicroseconds(Number value) {
+        return value.longValue() == DAY_MICROS;
+    }
+
+    static boolean isBoundaryNanoseconds(Number value) {
+        return value.longValue() == DAY_NANOS;
+    }
+
+    static boolean isBoundaryTimeWithTimezone(Object value) {
+        return value instanceof String && BOUNDARY_TIME_WITH_TIMEZONE_PATTERN.matcher((String) value).matches();
+    }
+
+    private PostgresTimeBoundary() {
+    }
+}
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresTimeBoundary.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresTimeBoundary.java
@@ -23,20 +23,20 @@ final class PostgresTimeBoundary {
     private static final long DAY_NANOS = TimeUnit.DAYS.toNanos(1);
     private static final Pattern BOUNDARY_TIME_WITH_TIMEZONE_PATTERN = Pattern.compile("^24:00:00(?:\\.0{1,6})?(?:Z|[+-]00(?::?00)?)$");
 
-    static boolean isBoundaryMilliseconds(Number value) {
-        return value.longValue() == DAY_MILLIS;
+    static boolean isBoundaryMilliseconds(Object value) {
+        return value instanceof Number number && number.longValue() == DAY_MILLIS;
     }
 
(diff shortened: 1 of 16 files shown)


jpechane:
@minleejae Could you please rebase the PR?

minleejae:
hi, @jpechane I rebased : https://github.com/debezium/debezium/pull/7540/changes/4c77d72577f7ca030c4cde097221b42bb00e731e

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresTimeBoundary.java
```diff
@@ -0,0 +1,44 @@
...
+        return value.longValue() == DAY_MICROS;
+    }
+
+    static boolean isBoundaryNanoseconds(Number value) {
+        return value.longValue() == DAY_NANOS;
+    }
+
+    static boolean isBoundaryTimeWithTimezone(Object value) {
+        return value instanceof String && BOUNDARY_TIME_WITH_TIMEZONE_PATTERN.matcher((String) value).matches();
+    }
```
Naros, line 40: For this use case, we pass the `Object` and handle the cast/type-check internally, but for the others in this class, we delegate that responsibility to the caller. Should we always accept `Object` at all call sites and handle the cast internally?

This would cleanly allow for dropping the `isBoundary` helpers in each of the Type classes, too.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java
```diff
@@ -1010,6 +1011,10 @@ else if (data instanceof OffsetDateTime) {
     protected Object convertTimeWithZone(Column column, Field fieldDefn, Object data) {
         // during snapshotting; already receiving OffsetTime @ UTC during streaming
         if (data instanceof String) {
+            if (PostgresTimeBoundary.isTimeWithTimeZoneBoundaryAtUtc((String) data)) {
```
Naros:
```suggestion
if (data instanceof String value) {
            if (PostgresTimeBoundary.isTimeWithTimeZoneBoundaryAtUtc(value)) {
```

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java
```diff
@@ -1138,10 +1143,16 @@ protected Object convertArray(Column column, Field fieldDefn, PostgresType eleme
             }
             else if (data instanceof PgArray) {
```
Naros:
```suggestion
            else if (data instanceof PgArray array) {
```
And then in the code that follows use `array` so we only cast the value once.

Naros (commented):
Hi @minleejae great work, I've left a few nit-picks.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresTimeBoundary.java
minleejae: Thanks, updated the boundary helpers to accept Object and perform the type checks internally.  [377d8af](https://github.com/debezium/debezium/pull/7540/commits/377d8afe6b63c194e3a0ec8cbadab92c28b88bf6)

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java
minleejae: Updated in https://github.com/debezium/debezium/pull/7540/changes/93264ab1ce80ae734ba97b04672755608e72cc7f

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java
minleejae: The PgArray branch now uses a pattern variable and reuses it for both array conversion paths. [91551e0](https://github.com/debezium/debezium/pull/7540/commits/91551e01dce0e7c60d40869d222bce8a42f48626)

Naros (approved):
Hi @minleejae overall LGTM, I just have one final question. What about a scenario where the value is emitted as `24:00:00+05:30`, isn't this case bypassed with the current code?

minleejae:
Hi @Naros

I haven’t added another commit yet. My current thinking is to keep this PR scoped to the original issue: making the JDBC sink handle PostgreSQL `24:00:00` boundary values correctly when such a value reaches the sink.

  While looking into the reviewer’s `24:00:00+05:30` case, I checked PostgreSQL’s `time with time zone` behavior more closely. PostgreSQL records the offset as part of a `time with time zone` value and outputs it as stored; unlike `timestamp with time zone`, it is not adjusted to the active session time zone:
  https://www.postgresql.org/docs/current/datatype-datetime.html

  So from a PostgreSQL source-to-PostgreSQL sink fidelity perspective, preserving the TIMETZ value and offset seems to be the correct target, e.g.:

  `source PG: 24:00:00+05:30` -> `sink PG: 24:00:00+05:30`

The current Debezium PostgreSQL `TIMETZ` handling is based on converting `ZonedTime` values to UTC, which means the original offset is not preserved before the JDBC sink sees the value. That looks like a broader source-side semantic issue rather than something this sink-focused PR should change.

I think the right place to address full TIMETZ fidelity is the structured temporal work in #7554, where we can model the raw time components and offset explicitly. That work should also account for PostgreSQL’s `24:00:00` boundary, since Java `OffsetTime`/`LocalTime` cannot represent hour 24 directly. What do you think?

Naros:
Applied thanks @minleejae :tada:

PR #7542, opened 2026-06-18
debezium/dbz#2100 Preserve PostgreSQL MONEY snapshot precision

minleejae:
## Problem

Fixes debezium/dbz#2100.

PostgreSQL `MONEY` values read during snapshots were converted through `PGmoney.val`, which is a `double`. For large values this can lose precision before Debezium emits the snapshot record.

This was observed in a PostgreSQL source to PostgreSQL JDBC sink pipeline where the source value `-92233720368547758.08` was written to the sink as `-92233720368547760.00`. The sink made the precision loss visible, but the lossy conversion happened earlier in the PostgreSQL source snapshot path.

## Solution

Parse PostgreSQL `MONEY` string values directly into `BigDecimal` instead of converting through `double`.

This change:

* adds a shared PostgreSQL MONEY parser for snapshot and replication fallback paths
* updates the snapshot `PgOid.MONEY` path to use the shared parser
* updates the pgproto int64 MONEY path to use `BigDecimal.valueOf(value, 2)`
* adds a snapshot regression test for the large negative boundary value

## Testing

```bash
JAVA_HOME=/opt/homebrew/opt/openjdk@25 PATH=/opt/homebrew/opt/openjdk@25/bin:$PATH ./mvnw -U -pl debezium-connector-postgres -am -Dtest=NoUnitTests -Dsurefire.failIfNoSpecifiedTests=false -Dit.test=PostgresMoneyIT -Dfailsafe.failIfNoSpecifiedTests=false verify

JAVA_HOME=/opt/homebrew/opt/openjdk@25 PATH=/opt/homebrew/opt/openjdk@25/bin:$PATH ./mvnw -U -pl debezium-connector-postgres -am clean verify -Dtest=NoUnitTests -Dsurefire.failIfNoSpecifiedTests=false -Dit.test=RecordsSnapshotProducerIT#shouldGenerateSnapshotForPositiveMoney+shouldGenerateSnapshotForNegativeMoney+shouldGenerateSnapshotForNullMoney+shouldGenerateSnapshotForLargeNegativeMoneyWithoutPrecisionLoss -Dfailsafe.failIfNoSpecifiedTests=false
```

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/AbstractColumnValue.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/AbstractColumnValue.java
@@ -23,7 +23,6 @@ import org.postgresql.geometric.PGpoint;
 import org.postgresql.geometric.PGpolygon;
 import org.postgresql.jdbc.PgArray;
 import org.postgresql.util.PGInterval;
-import org.postgresql.util.PGtokenizer;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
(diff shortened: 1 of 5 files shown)


jpechane:
@minleejae Applied, thanks!

PR #7543, opened 2026-06-18
debezium/dbz#2102 Preserve MySQL BIGINT UNSIGNED in JDBC sink

minleejae:
## Problem

Fixes debezium/dbz#2102.

For MySQL source to MySQL JDBC sink pipelines, `BIGINT UNSIGNED` could lose its target type and value semantics when source type metadata was propagated.

| Source mode / sink setup | Kafka Connect representation | Behavior before this change | Impact |
|---|---|---|---|
| `bigint.unsigned.handling.mode=long` with auto-DDL | `INT64`, for example `-1L` for `18446744073709551615` | JDBC sink created a signed `bigint` | The MySQL unsigned type was not preserved |
| `bigint.unsigned.handling.mode=long` with a pre-created `bigint unsigned` sink column | `INT64`, for example `-1L` for `18446744073709551615` | JDBC sink bound the negative signed value | Overflow-range unsigned values could be written incorrectly |
| `bigint.unsigned.handling.mode=precise` with auto-DDL | Decimal scale 0, for example `BigDecimal("18446744073709551615")` | JDBC sink resolved the Debezium Decimal schema before propagated source metadata and created `decimal(20,0)` | The value was preserved, but the MySQL unsigned type was not |

## Solution

This change:

* adds a MySQL JDBC sink type handler for propagated `BIGINT UNSIGNED` and `INT8 UNSIGNED` source column metadata
* handles that source metadata in the MySQL dialect only, so the general schema logical-name precedence remains unchanged for other dialects and types
* binds signed long payloads with `Long.toUnsignedString(...)` so overflow-range values are written as the intended unsigned decimal value
* adds unit and MySQL integration coverage for `long` and `precise` unsigned bigint records

The MySQL-specific handling is intentionally narrow. Other logical schemas, such as Kafka Connect `Decimal`, continue to resolve through the normal logical schema path unless the target dialect explicitly handles the propagated source type.

## Testing

```bash
./mvnw -pl debezium-connector-jdbc process-sources

./mvnw -pl debezium-connector-jdbc -DskipITs -Dtest=io.debezium.connector.jdbc.dialect.mysql.BigIntUnsignedTypeTest test

./mvnw -pl debezium-connector-jdbc -Dtest=NoUnitTests -Dsurefire.failIfNoSpecifiedTests=false -Dit.test=io.debezium.connector.jdbc.integration.mysql.JdbcSinkColumnTypeMappingIT#testShouldCreateAndWriteUnsignedBigIntColumnType verify

./mvnw -pl debezium-connector-jdbc -Dtest=NoUnitTests -Dsurefire.failIfNoSpecifiedTests=false -Dit.test=io.debezium.connector.jdbc.e2e.JdbcSinkPipelineToPostgresIT#testMoneyDataType -Dfailsafe.failIfNoSpecifiedTests=false -Dtest.tags=e2e-postgresql -Dsource.connectors=sqlserver verify
```

The last command verifies that source column type propagation no longer causes SQL Server `money` to be auto-created as PostgreSQL `money`; it remains `NUMERIC(19,4)` through the Decimal logical schema path.

Also verified with Docker Compose using `.experiment/precision-mode-avro/docker-compose.yml`, a local `3.6.0-SNAPSHOT` JDBC sink plugin, JSON converter and Avro/Schema Registry, and both `long` and `precise` source modes. The sink auto-created `bigint unsigned` and stored `18446744073709551615` in all four combinations.

## Non-blocking follow-up

This PR only fixes the JDBC sink behavior and intentionally leaves the MySQL source connector default `bigint.unsigned.handling.mode=long` unchanged.

Separately from this PR, would it be worth discussing whether `precise` should become the default in the future? `long` preserves the existing source schema, but values in `9223372036854775808` through `18446744073709551615` are represented as negative INT64 values in Kafka Connect. `precise` avoids that at the source, although it would change the emitted schema and Java value type, so I am raising it only as a follow-up topic.

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/BigIntUnsignedType.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/BigIntUnsignedType.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.jdbc.dialect.mysql;
+
+import java.math.BigDecimal;
+import java.math.BigInteger;
+import java.util.List;
+
+import org.apache.kafka.connect.data.Schema;
+
+import io.debezium.connector.jdbc.type.AbstractType;
+import io.debezium.connector.jdbc.type.JdbcType;
+import io.debezium.sink.valuebinding.ValueBindDescriptor;
+
+/**
+ * An implementation of {@link JdbcType} that provides support for {@code BIGINT UNSIGNED} data types.
+ */
+class BigIntUnsignedType extends AbstractType {
+
+    public static final BigIntUnsignedType INSTANCE = new BigIntUnsignedType();
+    private static final String[] REGISTRATION_KEYS = new String[]{ "BIGINT UNSIGNED", "BIGINT UNSIGNED ZEROFILL", "INT8 UNSIGNED", "INT8 UNSIGNED ZEROFILL" };
+
+    @Override
+    public String[] getRegistrationKeys() {
+        return REGISTRATION_KEYS;
+    }
+
+    public boolean matchesSourceColumnType(Schema schema) {
+        if (schema.parameters() == null) {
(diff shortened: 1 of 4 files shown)


jpechane:
@minleejae Applied, thanks!

PR #7545, opened 2026-06-18
debezium/dbz#2103 Forbid using shaded classes

jpechane:
Fixes debezium/dbz#2103


mfvitale (approved):

Naros (approved):
Thanks for picking this one up @jpechane 🚀

PR #7546, opened 2026-06-18
debezium/dbz#2102 Preserve MySQL BIT values in JDBC sink

minleejae:
## Problem

Fixes debezium/dbz#2102.

In a MySQL source to MySQL JDBC sink pipeline, MySQL `BIT` values were not converted back into MySQL `BIT` values before being written by the sink.

The MySQL source connector emits multi-bit `BIT(n)` columns as Kafka Connect `io.debezium.data.Bits` values. That payload is not a SQL-ready MySQL bit literal such as `b'101010101'`. It is Debezium's `Bits` byte representation. With AvroConverter, the same logical value arrives at the sink as a `ByteBuffer`.

Before this change, the MySQL JDBC sink had a `BitType` handler for type names and default values, but it did not override runtime value binding. The generic binding path therefore passed the raw Java value to the MySQL driver.

| MySQL source value | Expected sink value | Numeric readback | Value bound before this change | Resulting problem |
|---|---|---|---|---|
| `BIT(8)` `b'11111111'` | `BIT(8)` `b'11111111'` | `255` | `byte[] { (byte) 0xff }`, or Avro `HeapByteBuffer[pos=0 lim=1 cap=1]` | The sink did not explicitly bind a value representing the original `BIT(8)` pattern |
| `BIT(9)` `b'101010101'` | `BIT(9)` `b'101010101'` | `341` | `byte[] { 0x55, 0x01 }`, or Avro `HeapByteBuffer[pos=0 lim=2 cap=2]` | The raw Debezium little-endian `Bits` bytes were passed through instead of being decoded back to the original `BIT(9)` pattern; this can be rejected by MySQL as an oversized binary value for `BIT(9)` |
| `BIT(16)` `b'0000000100000010'` | `BIT(16)` `b'0000000100000010'` | `258` | `byte[] { 0x02, 0x01 }` | The raw byte order is not the MySQL bit pattern itself; it must be decoded from Debezium `Bits` first |
| `BIT(64)` all bits set | `BIT(64)` with all bits set | `18446744073709551615` | eight raw `0xff` bytes | The high-bit payload needed unsigned conversion before binding |

This affects both sink table setups:

| Sink table setup | Why it is affected |
|---|---|
| Auto-created sink table | The sink may create `BIT(n)`, but the row value still used the generic raw-value binding path |
| Pre-created sink table | The column type is already correct, but inserts/upserts still depend on the same broken runtime binding path |

## Root Cause

`io.debezium.data.Bits` uses `BitSet` byte-array semantics. That representation must be decoded using the schema bit width before it can be written to a MySQL `BIT` column.

The MySQL JDBC sink `BitType` previously inherited `AbstractType.bind(...)`, whose behavior is effectively:

```java
return List.of(new ValueBindDescriptor(index, value));
```

So for a source value like `BIT(9) b'101010101'`, the sink attempted to bind the raw payload `byte[] { 0x55, 0x01 }` instead of binding a value representing the original MySQL bit pattern. When read numerically, that bit pattern is `341`. With Avro and Schema Registry, it attempted to bind the corresponding `ByteBuffer` object, which the generic binding path cannot write as a MySQL `BIT` value.

## Solution

This change updates the MySQL JDBC sink `BitType` to:

* normalize `byte[]` and Avro `ByteBuffer` payloads through `ByteArrayUtils.getByteArrayFromValue(...)`
* decode Debezium `Bits` payloads with `BitSet.valueOf(...)`
* use the schema `length` parameter so bounded values such as `BIT(9)` preserve their width and byte order
* bind the decoded bit pattern as a positive `BigInteger`; for example, `byte[] { 0x55, 0x01 }` is decoded back to `BIT(9) b'101010101'`, which MySQL reads numerically as `341`
* format default values as MySQL bit literals, for example `b'101010101'`, instead of byte-array hex literals

## Testing

```bash
./mvnw -pl debezium-connector-jdbc process-sources

./mvnw -pl debezium-connector-jdbc -DskipITs -Dtest=io.debezium.connector.jdbc.dialect.mysql.BitTypeTest test

./mvnw -pl debezium-connector-jdbc -Dtest=NoUnitTests -Dsurefire.failIfNoSpecifiedTests=false -Dit.test=io.debezium.connector.jdbc.integration.mysql.JdbcSinkColumnTypeMappingIT#testShouldCreateAndWriteMultiByteBitColumnType verify
```

The new unit coverage checks:

* `BIT(1)` zero and one values
* multi-byte `BIT(9)` and `BIT(16)` byte order
* Avro `ByteBuffer` values
* high-bit `BIT(8)` values such as `0x80` and `0xff`
* maximum-width `BIT(64)` values, including all bits set
* default value formatting as MySQL bit literals

The MySQL integration test verifies JDBC sink auto-DDL and writes for `BIT(1)`, `BIT(8)`, `BIT(9)`, and `BIT(64)`.

Also verified with Docker Compose using `.experiment/precision-mode-avro/docker-compose.yml`, a local `3.6.0-SNAPSHOT` JDBC sink plugin, AvroConverter, and Schema Registry. Both auto-created and pre-created MySQL sink tables preserved these MySQL source values. The table below shows numeric readback from MySQL:

```text
id  b1  b8   b9   b64
1   1   255  341  18446744073709551615
2   0   128  511  0
```

### DIFF
--- a/debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/dialect/mysql/BitTypeTest.java
+++ b/debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/dialect/mysql/BitTypeTest.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.jdbc.dialect.mysql;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.math.BigInteger;
+import java.nio.ByteBuffer;
+import java.util.List;
+
+import org.apache.kafka.connect.data.Schema;
+import org.junit.jupiter.api.DisplayName;
+import org.junit.jupiter.api.Tag;
+import org.junit.jupiter.api.Test;
+
+import io.debezium.data.Bits;
+import io.debezium.doc.FixFor;
+import io.debezium.sink.valuebinding.ValueBindDescriptor;
+
+/**
+ * Unit tests for the MySQL {@link BitType} handler.
+ */
+@Tag("UnitTests")
+class BitTypeTest {
+
+    private static final BigInteger UNSIGNED_64BIT_MAX_VALUE = new BigInteger("18446744073709551615");
+
+    @Test
+    @FixFor("debezium/dbz#2102")
+    @DisplayName("Should bind one-bit boundary values")
+    void shouldBindOneBitBoundaryValues() {
+        assertBindValue(Bits.schema(1), new byte[]{ 0x00 }, BigInteger.ZERO);
+        assertBindValue(Bits.schema(1), new byte[]{ 0x01 }, BigInteger.ONE);
+    }
+
+    @Test
+    @FixFor("debezium/dbz#2102")
+    @DisplayName("Should bind multi-byte bit values using Debezium Bits byte order")
+    void shouldBindMultiByteBitValues() {
+        assertBindValue(Bits.schema(9), new byte[]{ 0x55, 0x01 }, new BigInteger("341"));
+        assertBindValue(Bits.schema(16), new byte[]{ 0x02, 0x01 }, new BigInteger("258"));
+    }
+
+    @Test
+    @FixFor("debezium/dbz#2102")
+    @DisplayName("Should bind ByteBuffer bit values from Avro converter")
+    void shouldBindByteBufferBitValues() {
+        assertBindValue(Bits.schema(9), ByteBuffer.wrap(new byte[]{ 0x55, 0x01 }), new BigInteger("341"));
+    }
+
+    @Test
+    @FixFor("debezium/dbz#2102")
+    @DisplayName("Should bind values with the high bit set as unsigned values")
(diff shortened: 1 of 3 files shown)


debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/dialect/mysql/BitTypeTest.java
```diff
@@ -0,0 +1,91 @@
...
+/**
+ * Unit tests for the MySQL {@link BitType} handler.
+ */
+@Tag("UnitTests")
+class BitTypeTest {
+
+    private static final BigInteger UNSIGNED_64BIT_MAX_VALUE = new BigInteger("18446744073709551615");
+
+    @Test
+    @FixFor("DBZ-2102")
```
jpechane:
```suggestion
    @FixFor("debezium/dbz#2102")
```

Same for other instances

jpechane (commented):
@minleejae LGTM, thanks! Could you please rebase as I merged the other PR?

minleejae:
Hi @jpechane,
Rebased on latest main and updated the `@FixFor` annotations to use `debezium/dbz#2102`. Thanks!

jpechane:
@minleejae Applied, thanks!

PR #7549, opened 2026-06-18
DBZ-2113 Reduce MySQL poll conversion overhead

minleejae:
## Description

This PR reduces allocation and CPU overhead in `MySqlConnectorTask#doPoll()`.

The method is on the repeated Kafka Connect source task poll path. It does not read the binlog or produce records to Kafka directly. At this point, Debezium has already created `DataChangeEvent` objects and `doPoll()` only unwraps their underlying `SourceRecord` values.

The previous implementation used a sequential stream and `Collectors.toList()` for this simple one-to-one conversion. This PR replaces that path with a pre-sized `ArrayList` and a direct loop.

This preserves behavior:

- record order is unchanged
- the same `SourceRecord` instances are returned
- connector configuration is unchanged
- emitted records and schemas are unchanged
- the returned list remains mutable, like the current `Collectors.toList()` result

Closes debezium/dbz#2113

## Benchmark

I added a JMH benchmark for the MySQL poll batch conversion path.

Environment used for the reported local benchmark:

- JDK: Amazon Corretto 17.0.11
- JMH: 1.37
- Profiler: `-prof gc`
- Mode: average time
- Benchmark type: sustained repeated conversion, 1,000 poll conversions per invocation

The benchmark numbers below were produced by running the JMH benchmark directly with the GC profiler after compiling the benchmark module:

```bash
./mvnw -pl debezium-microbenchmark -am -DskipTests clean compile -U

java -cp "<benchmark classpath>" org.openjdk.jmh.Main \
  'io.debezium.performance.connector.mysql.MySqlConnectorTaskPollPerf.*' \
  -prof gc
```

mysql-poll-benchmark-comparison-detailed-logscale

Detailed benchmark setup, chart, and single-poll results are in debezium/dbz#2113.

### Sustained repeated conversion

| batch size | loop | stream | time reduction | loop allocation | stream allocation | allocation reduction |
|---:|---:|---:|---:|---:|---:|---:|
| 2048 | 5.465 us/op | 7.926 us/op | 31.1% | 8,232 B/op | 33,776 B/op | 75.6% |
| 16384 | 28.244 us/op | 48.891 us/op | 42.2% | 65,576 B/op | 253,576 B/op | 74.1% |
| 65536 | 114.714 us/op | 192.932 us/op | 40.5% | 262,185 B/op | 854,369 B/op | 69.3% |

For the default `max.batch.size=2048`, allocation decreased from about `33.8 KB/op` to about `8.2 KB/op` in the sustained benchmark.

This does not claim an end-to-end connector throughput improvement by itself. The goal is to remove unnecessary allocation and CPU overhead from a hot poll conversion path.

## Verification

```bash
./mvnw -pl debezium-connector-mysql -am -DskipITs -Dtest=MySqlConnectorConfigTest -Dsurefire.failIfNoSpecifiedTests=false test
./mvnw -pl debezium-microbenchmark -am -DskipTests clean compile -U
```

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorTask.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlConnectorTask.java
@@ -6,11 +6,11 @@
 package io.debezium.connector.mysql;
 
 import java.sql.SQLException;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
-import java.util.stream.Collectors;
 
 import org.apache.kafka.connect.source.SourceRecord;
 import org.slf4j.Logger;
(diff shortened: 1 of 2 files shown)


jpechane:
@minleejae Good catch, applied. Thanks!

PR #7550, opened 2026-06-18
debezium/dbz#2105 Harden emitWindowOpen() so that it attempts a confi…

rainerschamm:
…gurable number of JDBC reconnections before giving up and abandoning an incremental snapshot

Fixes debezium/dbz#2105

## PR Checklist


jpechane:
@rainerschamm Could you please associate the commit with the email you use on GitHub? Thanks

jpechane:
@vjuranek There are multiple place in our codebase doing retyring logic. I wonder if `io.debezium.embedded.async.RetryingCallable<V>` can be generalized or the retrying logic in it into a class located into `io.debezium.util` package so it can be reused in all such places. It might even use a builder patter to provide a nice way how to configure it.

rainerschamm:
> @rainerschamm Could you please associate the commit with the email you use on GitHub? Thanks

Done! Both in my .gitconfig for future commits and in the commit of the PR.

rainerschamm:
> @vjuranek There are multiple place in our codebase doing retyring logic. I wonder if `io.debezium.embedded.async.RetryingCallable<V>` can be generalized or the retrying logic in it into a class located into `io.debezium.util` package so it can be reused in all such places. It might even use a builder patter to provide a nice way how to configure it.

Yes, let me look at that! 

Also later in the day I found this example which uses io.debezium.util.DelayStrategy. 

At a  minimum let me refactor the sleep to use  DelayStrategy and rename the "int count" to "int attempt" etc.



```java
    public LogFilesResult getLogs(Scn offsetScn) throws SQLException, LogFileNotFoundException {
        LOGGER.debug("Collecting logs based on the read SCN position {}.", offsetScn);
        final DelayStrategy retryStrategy = DelayStrategy.exponential(initialDelay, maxRetryDelay);
        for (int attempt = 0; attempt <= maxAttempts; ++attempt) {
            // Fetch current Redo Thread State
            final RedoThreadState currentRedoThreadState = connection.getRedoThreadState();
            for (RedoThread redoThread : currentRedoThreadState.getThreads()) {
                LOGGER.debug("Thread {}: {}", redoThread.getThreadId(), redoThread);
            }

            // Fetch logs
            final List<LogFile> files = getLogsForOffsetScn(offsetScn);
            if (!isLogFileListConsistent(offsetScn, files, currentRedoThreadState)) {
                LOGGER.info("No logs available yet (attempt {})...", attempt + 1);
                retryStrategy.sleepWhen(true);
                continue;
            }

            return new LogFilesResult(files, currentRedoThreadState);
        }
        throw new LogFileNotFoundException(offsetScn);
    }
```

rainerschamm:
FYI, I am working on a utility with builder pattern loosely based on: 

`io.debezium.embedded.async.RetryingCallable<V>`

The new utility class has this signature: 

```java
public class RetryingRunnable<E extends Exception> {}
```

And it uses a new functional interface:

```java
@FunctionalInterface
public interface ThrowingRunnable<E extends Exception> {

    void run() throws E, InterruptedException;

}
```

The usage looks like this in our case:

```java
        RetryingRunnable.<SQLException> builder()
                .retries(connectorConfig.getSignalEmitFailureMaxRetries())
                .delayStrategy(DelayStrategy.constant(connectorConfig.getSignalEmitFailureBackoff()))
                .doRun(() -> {
                    String signalWindowStatement = "INSERT INTO " + signalTableName + " VALUES (?, ?, ?)";
                    signalMetadata = new SignalMetadata(Instant.now(), null);
                    jdbcConnection.prepareUpdate(signalWindowStatement, x -> {
                        LOGGER.trace("Emitting open window for chunk = '{}' to signal table '{}'", context.currentChunkId(), signalTableName);
                        x.setString(1, context.currentChunkId() + "-open");
                        x.setString(2, OpenIncrementalSnapshotWindow.NAME);
                        x.setString(3, signalMetadata.metadataString());
                    });
                    jdbcConnection.commit();
                })
                .doAutoHeal(() -> {
                    LOGGER.error("Error emitting open window to signal table '{}'. Closing connection and re-connecting", signalTableName);
                    try {
                        jdbcConnection.close();
                    }
                    catch (Exception ignore) {
                        // Ignore
                    }
                    LOGGER.trace("Reconnecting JDBC connection");
                    jdbcConnection.connect();
                })
                .build()
                .runWrapped(cause -> new SQLException("Interrupted", cause));
```

The runWrapped() is only required here if we keep the signature of the method the same. Else if would just be .run();

I am still working on the utility class tests, so not ready to commit yet....

debezium-util/src/main/java/io/debezium/util/RetryingRunnable.java
```diff
@@ -0,0 +1,138 @@
...
+import io.debezium.function.ThrowingRunnable;
+
+/**
+ * Allows to re-try a runnable action if exception is thrown during the execution.
+ * The action is re-tried {@code retries} number of times.
+ * The delay between retries is defined by {@link DelayStrategy}, which needs to be provided by the implementing class.
+ * Optionally, an auto-heal action can be provided, which is executed before each retry.
+ * Inspired by: io.debezium.embedded.async.RetryingCallable.
+ */
+public class RetryingRunnable<E extends Exception> {
```
mfvitale, line 27: I could see this moved to `debezium-util` module. WDYT?

rainerschamm: Oh yes definitely, much better location!

debezium-util/src/main/java/io/debezium/util/RetryingRunnable.java
```diff
@@ -0,0 +1,138 @@
...
+    private final ThrowingRunnable<E> doRun;
+    private final ThrowingRunnable<E> doAutoHeal;
+    private final DelayStrategy delayStrategy;
+
+    // package-private / private: construction goes through the builder
+    private RetryingRunnable(Builder<E> b) {
+        this.retries = b.retries;
+        this.doRun = b.doRun;
+        this.doAutoHeal = b.doAutoHeal;
+        this.delayStrategy = b.delayStrategy;
```
mfvitale, line 42: Wondering if we could also add the concept of retryable and not retryable exceptions. 
So for example passing a list of exception when on error start the retry and for exceptions not in this list just exit after one failure?

@jpechane @Naros WDYT?

jpechane: @mfvitale Good proposal!

debezium-util/src/main/java/io/debezium/util/RetryingRunnable.java
```diff
@@ -0,0 +1,138 @@
...
+    public void runWrapped(Function<Throwable, E> exceptionWrapper) throws E {
+        try {
+            run();
+        }
+        catch (InterruptedException ex) {
+            throw exceptionWrapper.apply(ex);
+        }
+    }
+
+    public void run() throws E, InterruptedException {
```
vjuranek, line 59: Would it be possible to adjust `RetryingCallable`, mvoe it info debezium-utils and use across the code as Jiri has suggested? If possible, I'd like to avoid duplicating the retrying logic as any change then needs to be done on several places and sooner or later start to differ and introduce inconsistent behavior in various parts of the application.

rainerschamm: Maybe we can try, but the main reason I did not refactor RetryingCallback is that its based on the Callable interface, which is primary used for asynchronous workloads, and becomes a bit awkward for synchronous void method style scenarios. Also, we lack control over the Exception being thrown.


```java
@FunctionalInterface
public interface Callable<V> {
    V call() throws Exception;
}
```

I ended up introducing this interface:

```java
@FunctionalInterface
public interface ThrowingRunnable<E extends Exception> {
    void run() throws E, InterruptedException;
}
```

This interface makes it a bit easier to introduce retries in synchronous use cases without having to catch (Exception ex) {} at some point.

I had to add this run variant though, to wrap InterruptedException whenever the signature of the calling method does not support it:

```java
    public void runWrapped(Function<Throwable, E> exceptionWrapper) throws E {
        try {
            run();
        }
        catch (InterruptedException ex) {
            throw exceptionWrapper.apply(ex);
        }
    }
```

I am open to looking into an approach that works without a new interface though.

rainerschamm:
Hi all, I'll move the utility classes into "debezium-util" a bit later today. 

Then I will try and add the idea of an optional list of "Retriable" exceptions. If the list is null or empty we retry on all exceptions, else we only retry for the supplied exceptions.

rainerschamm:
Please review this version of RetryingRunnable utility.

debezium-util/src/main/java/io/debezium/util/RetryingRunnable.java
```diff
@@ -0,0 +1,205 @@
...
+     * Returns {@code true} if the given exception should be retried. When no retriable exception types were
+     * configured (empty list), every exception is retriable. Otherwise the exception is retriable only if it is
+     * an instance of one of the configured types (or a descendant thereof).
+     */
+    private boolean isRetriable(Exception ex) {
+        if (retriableExceptions.isEmpty()) {
+            return true;
+        }
+        for (Class<? extends Exception> retriable : retriableExceptions) {
+            if (retriable.isInstance(ex)) {
```
mfvitale: Should we navigate the stacktrace until we reach a `null` cause so that if a retryable exception is wrapped into a non retryable one we can caught it?

debezium-util/src/main/java/io/debezium/util/RetryingRunnable.java
vjuranek: IMHO the class implementing retrying mechanism don't have to extend `Callable` interface. I can be ordinary class like your `RetryingRunnable`, which would be used later on e.g. for implementing `Callable` interface.
`PollRecords` would look like

```
private static class PollRecords implements Callable<Void> {
    private final RetryingCallable<Void, Exception> retrying;

    this.retrying = RetryingRunnable.<Void, Exception> builder(this::doPoll). ... .build();

    public Void call() throws Exception {
        return retrying.call();
    }
}
```

My point here is not to reuse `RetryingCallable` itself, but avoid to introduce another place where we implement retrying mechanism. We already have several such places and I'd like to avoid introducing one more.

debezium-util/src/main/java/io/debezium/util/RetryingRunnable.java
Naros: I would think so, because then internally in our code we can just toss up a `DebeziumException` and let the stack trace walk control the behavior, particularly for things like `SQLRetriableException`(s).

Naros (commented):
Hi @rainerschamm I realize this PR is addressing `emitWindowOpen`, but isn't the same exact issue present in `emitWindowClose` as well?

debezium-util/src/main/java/io/debezium/util/RetryingRunnable.java
rainerschamm: Ok cool, I will have a look!

rainerschamm:
> Hi @rainerschamm I realize this PR is addressing `emitWindowOpen`, but isn't the same exact issue present in `emitWindowClose` as well?

Yea I would also think so, but we have never observed it. Let me add it to emitWindowClose() too.

rainerschamm:
Let me look at changing the defaults of the new configuration setting for those 3 failed integration tests. So that the old tests still behave the same (no retries)

rainerschamm:
> Let me look at changing the defaults of the new configuration setting for those 3 failed integration tests. So that the old tests still behave the same (no retries)

Or else, maybe better, we change the default of the new configuration setting to no retries.

rainerschamm:
Please verify again:)

rainerschamm:
Sorry, I saw I had a mistake in the common config validation for retries, it was set to isPositiveInteger, I changed it to:

```java
    public static final Field SIGNAL_EMIT_FAILURE_MAX_RETRIES = Field.create("signal.emit.failure.max.retries")
            .withDisplayName("Signal emit window event failure maximum retries")
            .withType(Type.INT)
            .withGroup(Field.createGroupEntry(Field.Group.ADVANCED, 24))
            .withWidth(Width.SHORT)
            .withImportance(Importance.MEDIUM)
            .withDefault(0)
            .withValidation(Field::isNonNegativeInteger)
            .withDescription(
                    "The maximum number of retries after encountering an error emitting a window event (-1 = no limit, 0 = disabled, > 0 = num of retries). Defaults to 0.");
```

Does the rest look ok? In particular I am not sure if the group setting is correct.

Also I'll setup a small build server so that I can run the integration tests in future myself, my laptop is not really suitable.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/SignalBasedIncrementalSnapshotChangeEventSource.java
```diff
@@ -137,7 +147,25 @@ protected void emitWindowClose(P partition, OffsetContext offsetContext) throws
...
+                .run();
+    }
+
+    private void autoHealJdbcConnection() throws SQLException {
+        LOGGER.info("Closing JDBC connection and re-connecting");
+        try {
+            jdbcConnection.close();
+        }
+        catch (Exception ignore) {
+            // Ignore
```
mfvitale: Could you at least log something?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/SignalBasedIncrementalSnapshotChangeEventSource.java
```diff
@@ -115,15 +117,23 @@ protected void emitWindowOpen(P partition, OffsetContext offsetContext) throws S
...
+                    String signalWindowStatement = "INSERT INTO " + signalTableName + " VALUES (?, ?, ?)";
+                    signalMetadata = new SignalMetadata(Instant.now(), null);
+                    jdbcConnection.prepareUpdate(signalWindowStatement, x -> {
+                        LOGGER.trace("Emitting open window for chunk = '{}' to signal table '{}'", context.currentChunkId(), signalTableName);
+                        x.setString(1, context.currentChunkId() + "-open");
+                        x.setString(2, OpenIncrementalSnapshotWindow.NAME);
+                        x.setString(3, signalMetadata.metadataString());
+                    });
+                    jdbcConnection.commit();
+                })
```
mfvitale: Maybe you can put this into a method

mfvitale (approved):

mfvitale:
Hi @rainerschamm Overall it LGTM. I left some minor comments.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/SignalBasedIncrementalSnapshotChangeEventSource.java
rainerschamm: Yes, will do!

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/SignalBasedIncrementalSnapshotChangeEventSource.java
rainerschamm: Cool yes!

rainerschamm:
I managed to run some of the integrations test on my new build server and its looking better now!
Will try and run all...

jpechane:
@rainerschamm Applied, very useful, thanks!

rainerschamm:
Awesome thanks:)

PR #7554, opened 2026-06-20
debezium/dbz#2119 Introduce structured temporal types beyond INT64 epoch limits

minleejae:
Fixes debezium/dbz#2119.

## Description

This PR introduces `time.precision.mode=structured` as discussed in debezium/dbz#2119.

The goal is to preserve high-precision temporal values that cannot be safely represented by flat INT64 epoch-based schemas such as `MicroTimestamp` or `NanoTimestamp`. The new mode emits structured Debezium semantic schemas for temporal values instead of collapsing range and precision into one integer.

## Changes

* Added structured temporal semantic schemas for date, time, timestamp, zoned timestamp, zoned time, and duration values.
* Added source conversion support for MySQL, MariaDB, PostgreSQL, Oracle, and SQL Server.
* Preserved temporal edge cases that exceed flat INT64 timestamp limits, including far-future PostgreSQL timestamps and high-precision Oracle / SQL Server values.
* Added JDBC sink support for decoding structured temporal schemas and binding them to MySQL and PostgreSQL target columns.
* Added documentation for the new `structured` precision mode and connector-specific behavior.

## Notes

This PR intentionally focuses on structured temporal representation and sink support for MySQL/PostgreSQL JDBC targets. It does not attempt to change existing default temporal precision modes.

### CDC fidelity for invalid temporal values

In structured mode, the source connector preserves raw temporal components even when they cannot be represented as Java temporal values. For example, MySQL and MariaDB can contain temporal values such as `0000-00-00`, `0000-00-00 00:00:00`, or calendar-invalid values allowed by the source database configuration.

This behavior is intentional for CDC fidelity. The source connector should describe what existed in the source database rather than silently converting such values to `null`, epoch values, or another normalized timestamp. If a target system cannot store those values, sink connectors or SMTs should decide explicitly whether to reject, transform, or map them for the target database.

For PostgreSQL logical decoding, `pgoutput` can preserve the tested structured temporal values. The `decoderbufs` path remains a separate limitation because it depends on the external decoder payload representation.

## Testing

Passed locally:

* `./mvnw -pl debezium-connector-common,debezium-connector-binlog,debezium-connector-mysql,debezium-connector-mariadb,debezium-connector-postgres,debezium-connector-oracle,debezium-connector-sqlserver,debezium-connector-jdbc -am -DskipTests process-sources test-compile`
* `./mvnw -pl debezium-connector-common,debezium-connector-binlog,debezium-connector-mysql,debezium-connector-mariadb,debezium-connector-oracle,debezium-connector-sqlserver,debezium-connector-jdbc -am -Dtest=StructuredTemporalTest,JdbcValueConvertersTemporalPrecisionTest,RowDeserializersTest,MySqlValueConvertersTest,ValueConvertersTest,OracleValueConvertersTest,SqlServerValueConvertersTest,StructuredTemporalTypeTest -Dsurefire.failIfNoSpecifiedTests=false test`
* MySQL snapshot structured temporal edge IT
* MariaDB snapshot structured temporal edge IT
* PostgreSQL structured temporal snapshot/streaming IT with `pgoutput`
* JDBC sink structured temporal MySQL/PostgreSQL integration tests
* `git diff --check`

Not completed locally:

* SQL Server and Oracle container ITs did not reach test execution because the local Docker daemon started leaving new containers in `Created` state. This was reproduced even with a small probe container, so it appears to be a local Docker start-path issue rather than a test failure.


debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/oracle/StructuredDateType.java
```diff
@@ -0,0 +1,14 @@
...
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.jdbc.dialect.oracle;
+
+/**
+ * Oracle implementation of {@link io.debezium.time.StructuredDate} values.
+ */
+public class StructuredDateType extends io.debezium.connector.jdbc.type.debezium.StructuredDateType {
```
Naros: If a specific dialect does not introduce any unique change in behavior from the base type, please just register the common type in the dialect and avoid creating wrapper classes that add no value. This keeps the code aligned with what we've done with other types.

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
```diff
@@ -140,9 +144,19 @@ public SchemaBuilder schemaBuilder(Column column) {
...
             case OracleTypes.TIMESTAMPLTZ:
+                if (temporalPrecisionMode == TemporalPrecisionMode.STRUCTURED) {
+                    return StructuredZonedTimestamp.builder();
+                }
                 return ZonedTimestamp.builder();
-            case OracleTypes.INTERVALYM:
             case OracleTypes.INTERVALDS:
+                if (temporalPrecisionMode == TemporalPrecisionMode.STRUCTURED) {
+                    return StructuredDuration.builder();
+                }
```
Naros, line 154: There is no `StructuredDurationType` handler on the Oracle or SQL Server JDBC sink side. This creates a gap in the solution. The Oracle JDBC sink dialect should support this. The same is for both interval types.

So Oracle-to-Oracle or Oracle-to-SQL Server would fail.

debezium-connector-common/src/main/java/io/debezium/jdbc/JdbcValueConverters.java
```diff
@@ -364,6 +379,19 @@ protected Object convertTimestampWithZone(Column column, Field fieldDefn, Object
...
+        final int precision = getTimePrecision(column);
+        return convertValue(column, fieldDefn, data,
+                StructuredZonedTimestamp.from(fieldDefn.schema(), OffsetDateTime.of(LocalDate.ofEpochDay(0), LocalTime.MIDNIGHT, defaultOffset), null, precision),
+                (r) -> {
+                    try {
+                        r.deliver(StructuredZonedTimestamp.toStructuredZonedTimestamp(fieldDefn.schema(), data, defaultOffset, adjuster, precision));
+                    }
+                    catch (IllegalArgumentException e) {
+                    }
+                });
```
Naros, line 395: I realize there are several catches in the converters that do nothing, but should we at least log a warning here rather than silently swallowing the exception?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/type/AbstractTimeType.java
```diff
@@ -31,15 +30,14 @@ public String getTypeName(Schema schema, boolean isKey) {
...
         }
         return dialect.getJdbcTypeName(Types.TIME, Size.precision(dialect.getDefaultTimePrecision()));
     }
 
     protected int getTimePrecision(Schema schema) {
-        final String length = getSourceColumnSize(schema).orElse("0");
-        final Optional<String> scale = getSourceColumnPrecision(schema);
-        return scale.map(Integer::parseInt).orElseGet(() -> Integer.parseInt(length));
+        final String length = getSourceColumnSize(schema).orElse("-1");
+        return getSourceColumnPrecision(schema).map(Integer::parseInt).orElseGet(() -> Integer.parseInt(length));
```
Naros: Does this change potentially lead to different SQL generated at targets when schema evolution is enabled?  The same applies inside `AbstractTimestampType`, too.

Naros (commented):
Hi @minleejae besides the few points below, a few other things.

1. If this is being offered as a common connector framework solution for emitting structured temporal values, ideally there would be adjacent PRs in the sibling connector repositories, along with updates to all connector docs.
2. One thing I noticed that was missing on the JDBC sink is no support for dialects like Db2 and Db2 iSeries. We should add those to make sure all dialects are feature complete.

minleejae:
Hi @Naros Thanks, agreed. 

1. I’ll keep this PR focused on the common types and the core connectors in this repository, then follow up with separate PRs for the sibling connector repositories (for example, debezium-connector-db2, debezium-connector-ibmi, debezium-connector-cassandra, debezium-connector-vitess, debezium-connector-spanner, and debezium-connector-informix), including their tests and documentation updates. 

2. For the JDBC sink dialect coverage you mentioned, I’ll address the missing Db2 and Db2 iSeries support in this PR.  [97a0879](https://github.com/debezium/debezium/pull/7554/commits/97a08798330f849748f492819580ec7176eead09)

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/oracle/StructuredDateType.java
minleejae: I removed the Oracle and SQL Server structured temporal wrapper classes that did not add dialect-specific behavior, and removed their explicit dialect registrations.  [36a9df0](https://github.com/debezium/debezium/pull/7554/commits/36a9df0b179cc7aaee0c0d2212fb5f9a8dd7ae35)

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
minleejae: I added StructuredDurationType handlers for both Oracle and SQL Server JDBC sink dialects and registered them in the dialects. [9affa28](https://github.com/debezium/debezium/pull/7554/commits/9affa287e4a82f4e82867a3a4f3280769f847450)

debezium-connector-common/src/main/java/io/debezium/jdbc/JdbcValueConverters.java
minleejae: I replaced the empty catch blocks with warn logging. [6c09fbb](https://github.com/debezium/debezium/pull/7554/commits/6c09fbb0be4130bb42dfe8a074b1ed674915b231)

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/type/AbstractTimeType.java
minleejae: Yes, the previous change could have changed the generated target DDL with schema evolution enabled.
I updated the base time/timestamp types to keep the exisiting behavior for non-structured temporal types,
and limited explicit source precision 0 handling to structured types only. [db1afea](https://github.com/debezium/debezium/pull/7554/commits/db1afeaeb53eb3ed3495af3c6c21294e451e3852)

Naros:
Hi @minleejae, could I get a rebase?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogValueConverters.java
```diff
@@ -833,6 +846,75 @@ protected Object convertDurationToMicroseconds(Column column, Field fieldDefn, O
...
+                            0,
+                            Math.toIntExact(abs.toHours() * sign),
+                            abs.toMinutesPart() * sign,
+                            abs.toSecondsPart() * sign,
+                            abs.toNanosPart() * sign,
+                            precision));
+                }
+            }
+            catch (IllegalArgumentException | ArithmeticException e) {
+            }
```
Naros, line 875: We shouldn't silently swallow an exception.

documentation/modules/ROOT/pages/connectors/db2.adoc
```diff
@@ -1897,6 +1898,44 @@ Represents timestamp values in UTC format, according to the ISO 8601 standard, f
...
+|`STRUCT`
+|`io.debezium.time.StructuredTimestamp` +
+ +
+Contains `year`, `month`, `day`, `hour`, `minute`, `second`, `nanos`, and optional `precision` fields.
+
+|`DATETIMEOFFSET`
+|`STRUCT`
+|`io.debezium.time.StructuredZonedTimestamp` +
+ +
+Contains calendar and clock component fields, `offsetSeconds`, optional `zoneId`, and optional `precision`.
```
Naros: I don't believe these are valid data types for Db2. I believe the equivalent in Db2 is `TIMESTAMP`.

Naros (commented):
A few inline comments.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogValueConverters.java
minleejae: Now logs a warning instead of swallowing the exception. [2cd8e3e](https://github.com/debezium/debezium/pull/7554/commits/2cd8e3eb9a6279f8c26cc6aab546029e72634692)

documentation/modules/ROOT/pages/connectors/db2.adoc
minleejae: You're right — DATETIME/DATETIMEOFFSET aren't Db2 types. Corrected them across the Db2 temporal tables and prose: DATETIME → `TIMESTAMP`, and the zoned entry → `TIMESTAMP WITH TIME ZONE` aecab62

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/oracle/StructuredZonedTimeType.java
```diff
@@ -0,0 +1,53 @@
...
+    @Override
+    public List<ValueBindDescriptor> bind(int index, Schema schema, Object value) {
+        if (value == null) {
+            return List.of(new ValueBindDescriptor(index, null));
+        }
+        final Struct struct = requireStruct(value);
+        final OffsetTime offsetTime = OffsetTime.of(
+                StructuredTemporalSupport.toLocalTime(struct),
+                ZoneOffset.ofTotalSeconds(struct.getInt32(StructuredTemporal.OFFSET_SECONDS_FIELD)));
+        final ZonedDateTime zonedDateTime = offsetTime.atDate(LocalDate.now()).toZonedDateTime();
```
Naros: `LocalDate.now()` introduces host-clock non-determinism.

When the JDBC sink processes a structured zoned time value, it anchors the time-only components to today's date. This means:
- The same Kafka record produces different database values depending on *when* it's processed
- Re-delivery after a consumer rebalance could write a different date
- Records processed around midnight may get inconsistent dates within a single batch

Since the date component is semantically meaningless for time-only values (it's just needed to construct a `ZonedDateTime` for JDBC binding), consider using a fixed epoch date instead:

```suggestion
        final ZonedDateTime zonedDateTime = offsetTime.atDate(LocalDate.EPOCH).toZonedDateTime();
```

The same issue appears in:
- `debezium-connector-jdbc/.../sqlserver/StructuredTimeType.java:35`
- `debezium-connector-jdbc/.../sqlserver/StructuredZonedTimeType.java:56`

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/sqlserver/StructuredTimeType.java
```diff
@@ -0,0 +1,38 @@
...
+    public String getQueryBinding(ColumnDescriptor column, Schema schema, Object value) {
+        return "cast(? as time(7))";
+    }
+
+    @Override
+    public List<ValueBindDescriptor> bind(int index, Schema schema, Object value) {
+        if (value == null) {
+            return List.of(new ValueBindDescriptor(index, null));
+        }
+        final LocalDateTime localDateTime = StructuredTemporalSupport.toLocalTime(requireStruct(value)).atDate(LocalDate.now());
```
Naros: Same `LocalDate.now()` non-determinism issue as noted on `oracle/StructuredZonedTimeType.java:50`.

```suggestion
        final LocalDateTime localDateTime = StructuredTemporalSupport.toLocalTime(requireStruct(value)).atDate(LocalDate.EPOCH);
```

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/sqlserver/StructuredZonedTimeType.java
```diff
@@ -0,0 +1,59 @@
...
+    @Override
+    public List<ValueBindDescriptor> bind(int index, Schema schema, Object value) {
+        if (value == null) {
+            return List.of(new ValueBindDescriptor(index, null));
+        }
+        final Struct struct = requireStruct(value);
+        final OffsetTime offsetTime = OffsetTime.of(
+                StructuredTemporalSupport.toLocalTime(struct),
+                ZoneOffset.ofTotalSeconds(struct.getInt32(StructuredTemporal.OFFSET_SECONDS_FIELD)));
+        final OffsetDateTime offsetDateTime = offsetTime.atDate(LocalDate.now());
```
Naros: Same `LocalDate.now()` non-determinism issue as noted on `oracle/StructuredZonedTimeType.java:50`.

```suggestion
        final OffsetDateTime offsetDateTime = offsetTime.atDate(LocalDate.EPOCH);
```

Naros (commented):
Hi @minleejae I found one reminaing item below, if you could look.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/sqlserver/StructuredZonedTimeType.java
minleejae: Applied in https://github.com/debezium/debezium/pull/7554/changes/af945a249c31f00c5cfaf695701b5049258556b0

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/sqlserver/StructuredTimeType.java
minleejae: Applied in https://github.com/debezium/debezium/commit/af945a249c31f00c5cfaf695701b5049258556b0

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/oracle/StructuredZonedTimeType.java
minleejae: Good point, thanks. I changed the affected Oracle and SQL Server structured time bindings to use `LocalDate.EPOCH` and added test assertions for the epoch date. 
af945a2

Naros:
Hi @minleejae the test failures with SQL Server seem related, could you take a look?

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
```diff
@@ -156,6 +170,11 @@ public SchemaBuilder schemaBuilder(Column column) {
         }
     }
 
+    @Override
+    protected int getTimePrecision(Column column) {
+        return column.scale().orElse(column.length());
```
Naros, line 177: Could you add a small comment here that explains we do this for INTERVALDS. For TIMESTAMP, we already handle this inside `OracleConnection#overrideColumn`. This is just so I don't forget when I look at this in the future and wonder why 😉

Naros (approved):
So overall this LGTM. I've left one tiny item below, and I have one question: do you intend to extend this PR beyond the core connectors for the other community source connectors: Vitess, Cassandra, CockroachDB, etc?

Lastly, this PR is blocked by the following:

debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleValueConverters.java
minleejae: Applied in a5babbe.

minleejae:
Hi, @Naros Yes, I plan to follow up with connector-specific PRs for Vitess, Cassandra, CockroachDB, and the other community connectors.

Naros:
Rebased after applying #7411 to make sure CI is still green.

PR #7560, opened 2026-06-23
[ci] Error in Oracle 23 expression

jpechane:


Naros (approved):
Thanks for catching this, @jpechane.

PR #7565, opened 2026-06-25
debezium/dbz#2102 Handle MySQL ENUM and SET literal escaping

minleejae:
Fixes debezium/dbz#2102

## Description
This change improves MySQL and MariaDB ENUM/SET allowed value handling by keeping parser output as parsed option values and moving schema-parameter escaping to the common enumerated-values helper.

It also updates MySQL JDBC sink type rendering tests to cover comma, quote, backslash, trailing backslash, and empty ENUM labels.

## Testing
- `git diff --check`
- `JAVA_HOME=/opt/homebrew/opt/openjdk/libexec/openjdk.jdk/Contents/Home ./mvnw -pl debezium-connector-common test -Dtest=io.debezium.data.EnumeratedValuesTest`

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/MySqlPtAntlrDdlParser.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/MySqlPtAntlrDdlParser.java
@@ -419,7 +419,7 @@ public class MySqlPtAntlrDdlParser extends AntlrDdlParser<MySqlLexer, MySqlParse
     }
 
     /**
-     * Extracts the enumeration values properly parsed and escaped.
+     * Extracts the enumeration values properly parsed and unescaped.
      *
      * @param enumValues the raw enumeration values from the parsed column definition
      * @return the list of options allowed for the {@code ENUM} or {@code SET}; never null.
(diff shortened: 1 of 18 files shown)


jpechane:
@minleejae I know this is fix for source connector issue but could you please introduce a test on the sink side too to make sure it works end to end?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/MySqlPtAntlrDdlParser.java
```diff
@@ -432,10 +432,9 @@ public static List<String> extractEnumAndSetOptions(List<String> enumValues) {
     }
 
     public static String escapeOption(String option) {
```
Naros: I wonder if this should be named `unescapeOption`? The same applies to the other classes.

Naros (commented):
Overall this LGTM, I've left just one nit-pick below. Feel free to follow-up on this separately if you'd prefer.

Naros (approved):

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/MySqlPtAntlrDdlParser.java
minleejae: Applied in [36b582a](https://github.com/debezium/debezium/pull/7565/commits/36b582ae92293fa790a36361cb73462b84b4774d)

Naros (approved):

Naros:

PR #7570, opened 2026-06-26
debezium/dbz#1639 Add CockroachDB dialect to the JDBC sink

### LINKED ISSUE #1639: JDBC Sink: Add CockroachDBDatabaseDialectProvider to support Hibernate CockroachDialect
# 🚀 Feature Request

## Problem Description

The Debezium JDBC sink currently requires the following configuration when writing to CockroachDB:

```
hibernate.dialect=org.hibernate.dialect.PostgreSQLDialect
```

Hibernate 6.x provides a first-class dialect for CockroachDB:

```
org.hibernate.dialect.CockroachDialect
```

However, when `CockroachDialect` is used, the JDBC sink's `DatabaseDialectResolver` does not recognize it as PostgreSQL-compatible.

Because `CockroachDialect` extends `Dialect` directly (not `PostgreSQLDialect`), the resolver falls back to:

```
GeneralDatabaseDialect
```

`GeneralDatabaseDialect`:

* Does not support upsert
* Does not generate `INSERT ... ON CONFLICT ... DO UPDATE`

CockroachDB fully supports PostgreSQL-style upsert syntax, so this fallback results in missing upsert capability unless users manually override the dialect to `PostgreSQLDialect`.

---

## Proposed Change

Add a `CockroachDBDatabaseDialectProvider` to the JDBC sink that:

* Detects `org.hibernate.dialect.CockroachDialect`
* Maps it to a Cockroach-specific dialect implementation
* Or internally reuses `PostgresDatabaseDialect`
* Enables proper generation of `INSERT ... ON CONFLICT ... DO UPDATE`

This would allow users to use:

```
hibernate.dialect=org.hibernate.dialect.CockroachDialect
```

without losing upsert support.

---

## Expected Behavior

When using the Hibernate `CockroachDialect`, the Debezium JDBC sink should:

* Automatically resolve to a dialect that supports upsert
* Generate PostgreSQL-compatible `ON CONFLICT DO UPDATE` statements
* Avoid falling back to `GeneralDatabaseDialect`
* Remove the need for manual dialect overrides

---

## Alternatives Considered

Users can currently work around the issue by forcing:

```
hibernate.dialect=org.hibernate.dialect.PostgreSQLDialect
```

However, this:

* Is unintuitive
* Hides CockroachDB-specific optimizations
* Breaks principle-of-least-surprise behavior
* Prevents future CockroachDB-specific enhancements in the sink

---

## Additional Context

The JDBC sink requires hibernate.dialect=org.hibernate.dialect.PostgreSQLDialect. While CockroachDB has a first-class [CockroachDialect](https://docs.hibernate.org/orm/6.3/javadocs/org/hibernate/dialect/CockroachDialect.html) in Hibernate 6.x ([CockroachDB + Hibernate docs](https://www.cockroachlabs.com/docs/stable/build-a-java-app-with-cockroachdb-hibernate)), the Debezium JDBC sink's DatabaseDialectResolver does not yet have a CockroachDB-specific DatabaseDialectProvider. Since CockroachDialect extends Dialect directly (not PostgreSQLDialect), the resolver falls back to GeneralDatabaseDialect which lacks upsert support. Using PostgreSQLDialect correctly maps to the Debezium PostgresDatabaseDialect, which generates the INSERT ... ON CONFLICT ... DO UPDATE syntax that CockroachDB supports. Adding a CockroachDBDatabaseDialectProvider to the Debezium JDBC sink is a natural follow-up enhancement.

viragtripathi:
Closes debezium/dbz#1639.

Draft: opening for CI and maintainer input on the approach (extend `PostgresDatabaseDialect` vs a dedicated dialect, how much CockroachDB-specific type coverage to add, and whether to gate serialization retries behind a config flag).

## Problem

The JDBC sink resolves the database dialect from the Hibernate dialect. CockroachDB maps to Hibernate's `CockroachDialect`, which is not a `PostgreSQLDialect`, so no `DatabaseDialectProvider` matches and the sink falls back to `GeneralDatabaseDialect`. That dialect does not implement `getUpsertStatement`, so upsert and delete against CockroachDB fail with:

```
java.lang.UnsupportedOperationException: Upsert configurations are not supported for this dialect
  at io.debezium.connector.jdbc.dialect.GeneralDatabaseDialect.getUpsertStatement
```

The current workaround is to force `hibernate.dialect=org.hibernate.dialect.PostgreSQLDialect`.

## Change

- Add `CockroachDBDatabaseDialect` and its provider (matching Hibernate's `CockroachDialect`), extending `PostgresDatabaseDialect` since CockroachDB speaks the PostgreSQL wire protocol and supports `INSERT ... ON CONFLICT`.
- Make `PostgresDatabaseDialect`'s constructor `protected` so the CockroachDB dialect can extend it.
- Register the provider in `META-INF/services`.
- Treat CockroachDB serialization and deadlock conflicts (SQLSTATE `40001` and `40P01`, which Hibernate maps to `LockAcquisitionException`) as retriable so the configured flush retries (`flush.max.retries` / `flush.retry.delay.ms`) handle them.

## Verification

- Unit tests: the provider resolves `CockroachDialect` (and not a plain PostgreSQL dialect), is registered via `ServiceLoader`, and a `40001` `TransactionSerializationException` is covered by the `LockAcquisitionException` retriable class.
- Behavior validated separately end to end: a CockroachDB to CockroachDB pipeline with `insert.mode=upsert` and `delete.enabled=true` replicates correctly when the PostgreSQL dialect is selected, which this change makes automatic for CockroachDB targets.

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/cockroachdb/CockroachDBDatabaseDialect.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/cockroachdb/CockroachDBDatabaseDialect.java
@@ -1,0 +1,73 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.jdbc.dialect.cockroachdb;
+
+import java.util.Set;
+
+import org.hibernate.SessionFactory;
+import org.hibernate.dialect.CockroachDialect;
+import org.hibernate.dialect.Dialect;
+import org.hibernate.exception.LockAcquisitionException;
+import org.hibernate.exception.TransactionSerializationException;
+
+import io.debezium.connector.jdbc.JdbcSinkConnectorConfig;
+import io.debezium.connector.jdbc.dialect.DatabaseDialect;
+import io.debezium.connector.jdbc.dialect.DatabaseDialectProvider;
+import io.debezium.connector.jdbc.dialect.postgres.PostgresDatabaseDialect;
+
+/**
+ * A {@link DatabaseDialect} implementation for CockroachDB.
+ *
+ * <p>CockroachDB speaks the PostgreSQL wire protocol and supports {@code INSERT ... ON CONFLICT},
+ * so it reuses {@link PostgresDatabaseDialect}. The distinct dialect exists because Hibernate maps
+ * CockroachDB to {@link CockroachDialect}, which is not a {@code PostgreSQLDialect}; without a
+ * matching provider the resolver falls back to the general dialect, which does not support upsert.</p>
+ *
+ * @author Virag Tripathi
+ */
+public class CockroachDBDatabaseDialect extends PostgresDatabaseDialect {
+
+    public static class CockroachDBDatabaseDialectProvider implements DatabaseDialectProvider {
+        @Override
+        public boolean supports(Dialect dialect) {
+            return dialect instanceof CockroachDialect;
+        }
+
+        @Override
+        public Class<?> name() {
+            return CockroachDBDatabaseDialect.class;
+        }
+
+        @Override
+        public DatabaseDialect instantiate(JdbcSinkConnectorConfig config, SessionFactory sessionFactory) {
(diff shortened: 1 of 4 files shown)


Naros (approved):

jpechane:
@viragtripathi Applied, thanks!

PR #7574, opened 2026-06-28
debezium/dbz#248 Resolve CREATE TABLE LIKE schema when reference table not in cache

### LINKED ISSUE #248: Support CREATE TABLE ... LIKE syntax for blacklisted source table [DBZ-1496]
Migrated from [DBZ-1496](https://issues.redhat.com/browse/DBZ-1496)

See https://groups.google.com/forum/#!topic/debezium/ZnrOv3gK2BQ. I believe this was supported in the legacy parser, and we should make it work for Antlr, too. Essentially, {{CreateTableParserListener()}} will have to override {{enterLikePredicate()}} and fetch the metadata from the referenced table.

leosanqing:
Fixes debezium/dbz#248
## Description
Resolve CREATE TABLE LIKE schema when reference table not in cache

## Implementation / Solution
Due to the limitation of the `store.only.captured.databases.ddl` parameter, a "table not found" exception occurs if the referenced table falls outside the regex rules. This happens because the regex for the template table does not always match the regex for the fact table.

* Used string replacement to fully complete the DDL of the created table. 
* The newly created tables are now stored in memory for independent management.

When the template table is updated, newly created tables will still fetch and apply the most up-to-date DDL.

## PR Checklist

### DIFF
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
@@ -28,6 +28,7 @@ import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.function.Predicate;
+import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
 import javax.net.ssl.KeyManager;
@@ -114,6 +115,17 @@ public abstract class BinlogStreamingChangeEventSource<P extends BinlogPartition
     private static final String DML_DELETE_PREFIX = "DELETE ";
     private static final String DML_REPLACE_PREFIX = "REPLACE ";
 
+    /**
+     * Regex to detect Create TABLE ... LIKE statements and extract both the new table and the
+     * reference (template) table.
+     */
+    private static final Pattern CREATE_TABLE_LIKE_PATTERN = Pattern.compile(
+            "(?i)CREATE\\s+TABLE\\s+"
+                    + "(?:IF\\s+NOT\\s+EXISTS\\s+)?"
+                    + "(?:(`[^`]+`|[^.`\\s]+)\\.)?(`[^`]+`|[^.`\\s]+)"
+                    + "\\s+LIKE\\s+"
+                    + "(?:(`[^`]+`|[^.`\\s;]+)\\.)?(`[^`]+`|[^.\\s;]+)");
+
     private final BinaryLogClient client;
     private final BinlogStreamingChangeEventSourceMetrics<?, P> metrics;
     // todo: can we go back to accessing schema via task context? issue is all the generics :/
@@ -126,6 +138,7 @@ public abstract class BinlogStreamingChangeEventSource<P extends BinlogPartition
     private final ErrorHandler errorHandler;
     private final EventProcessingFailureHandlingMode eventDeserializationFailureHandlingMode;
     private final EventProcessingFailureHandlingMode inconsistentSchemaHandlingMode;
+    private final boolean resolveLikeTableSchema;
     private final SnapshotterService snapshotterService;
     private final Predicate<String> gtidDmlSourceFilter;
     private final boolean isGtidModeEnabled;
@@ -164,6 +177,7 @@ public abstract class BinlogStreamingChangeEventSource<P extends BinlogPartition
         this.metrics = metrics;
         this.eventDeserializationFailureHandlingMode = connectorConfig.getEventProcessingFailureHandlingMode();
         this.inconsistentSchemaHandlingMode = connectorConfig.getInconsistentSchemaFailureHandlingMode();
+        this.resolveLikeTableSchema = connectorConfig.isResolveLikeTableSchema();
         this.snapshotterService = snapshotterService;
         this.client = client;
         configureBinaryLogClient(client, connectorConfig, binaryLogClientThreads, connection);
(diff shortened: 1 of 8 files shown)


leosanqing:
I want to add some msg to explain why I implement by this way.

I Use Flink CDC（which based on dbz 1.9.8 version）in my company. Firstly I use Second way, and occur some problems in PRO env as below. 

1. **(Chosen) Resolve-and-retry** — when a `CREATE TABLE ... LIKE` references a table whose schema is not in the cache, fetch the reference table's structure via `SHOW CREATE TABLE`, build a self-contained `CREATE TABLE` DDL for the new table, and re-parse it.
2. **(Alternative) Cache the template table in memory** — keep the excluded reference (template) table's schema in memory so the `LIKE` statement can be resolved against it directly.

Both approaches have trade-offs. On balance I chose approach (1), for the following reasons:

### 1. The in-memory cache conflicts with `store.only.captured.databases.ddl`
If that option is enabled, the template table's DDL is never written to the schema history (the template table is typically excluded from capture). The cached schema would then live only in memory — so at runtime, after a restart/recovery, the reference table can no longer be found and schema-history recovery fails. The resolve-and-retry approach instead emits a synthetic `CREATE TABLE` for the *new, captured* table, which is persisted to schema history normally and recovers correctly.

### 2. The in-memory cache goes stale when the template table changes
A cached template schema is not updated when the template table's DDL evolves. Tables created later would be materialized from the stale template, leaving them inconsistent with the latest DDL schema on the server. Resolving via `SHOW CREATE TABLE` always reflects the t
able's current structure.

### 3. DDL completion does not depend on the template table, so maintenance stays uniform
Because we don't keep a separate cached representation, tables created via `CREATE TABLE ... LIKE` are maintained exactly like any other table. Future/new features (e.g. schema evolution) require no special-casing for `LIKE`-created tables.

### 4. The per-query cost is acceptable
`CREATE TABLE ... LIKE` is not a high-frequency operation, so the extra `SHOW CREATE TABLE` round-trip on the streaming ice.

jpechane:
@leosanqing Are you sure you point to the correct GitHub issue?

leosanqing:
> @leosanqing Are you sure you point to the correct GitHub issue?

hi It's this ISSUE, https://github.com/debezium/dbz/issues/248
Its original number is 1496, should I change to 248？ dbz 248 is correct?

leosanqing:
@jpechane Hi, could you  re-run  failed jobs again? PG and Sql server run  failed, looks like infra flakiness.

jpechane:
@leosanqing Yes, please the new issue number - following the GHI numbering

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
```diff
@@ -114,6 +115,17 @@ public abstract class BinlogStreamingChangeEventSource<P extends BinlogPartition
     private static final String DML_DELETE_PREFIX = "DELETE ";
     private static final String DML_REPLACE_PREFIX = "REPLACE ";
 
+    /**
+     * Regex to detect Create TABLE ... LIKE statements and extract both the new table and the
+     * reference (template) table.
+     */
+    private static final Pattern CREATE_TABLE_LIKE_PATTERN = Pattern.compile(
```
jpechane: PLease use multiline String.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
```diff
@@ -820,6 +843,87 @@ private String removeSetStatement(String sql) {
...
+                newTableId, refTableId);
+
+        final String refCreateDdl = fetchShowCreateTable(refTableId);
+        if (refCreateDdl == null) {
+            LOGGER.warn("Failed to resolve reference table schema for {}; " +
+                    "the new table {} will not be tracked until its schema is resolved", refTableId, newTableId);
+            return List.of();
+        }
+
+        final String syntheticDdl = refCreateDdl.replaceFirst(
```
jpechane, line 877: Please use `String.formatted`  methods.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
```diff
@@ -820,6 +843,87 @@ private String removeSetStatement(String sql) {
...
+        return new TableId[]{ newTableId, refTableId };
+    }
+
+    private static String unquote(String identifier, String fallback) {
+        return identifier == null ? fallback : identifier.replace("`", "");
+    }
+
+    private String fetchShowCreateTable(TableId tableId) {
+        String quotedRef = (tableId.catalog() != null)
+                ? "`" + tableId.catalog() + "`.`" + tableId.table() + "`"
```
jpechane: Use `String.formatted`

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogCreateTableLikeSchemaResolutionIT.java
```diff
@@ -0,0 +1,144 @@
...
+
+        Struct after = value.getStruct("after");
+        assertThat(after.getString("name")).isEqualTo("from_like_table");
+        assertThat(after.getInt32("status")).isEqualTo(2);
+
+        stopConnector();
+    }
+
+    @Test
+    @FixFor("DBZ-1496")
```
jpechane:
```suggestion
    @FixFor("debezium/dbz#248")
```

Also for the other instances in annotations or comments

jpechane (commented):
@leosanqing LGTM, thanks. I left few minor comments and then we are good to go.

jpechane (approved):
@leosanqing LGTM, thanks! Can be merged when CI passes.

leosanqing:
@jpechane Hello, all CI checks passed successfully earlier, but due to other PRs being merged, there were some merge conflicts. I have just resolved them and updated the branch. Could you please help trigger the CI again? Thank you.

Naros:
Thanks @leosanqing, applied! :tada:

PR #7575, opened 2026-06-29
debezium/dbz#1479 Reset RemainingTableCount metric when data snapshot is skipped

### LINKED ISSUE #1479: MySQL with snapshot mode no_data produces snapshot remaining tables JMX metric
## Bug report

`MySQL 3.3.1`

---

`{
    "database.dbname": "**REDACTED**",
    "database.hostname": "**REDACTED**",
    "database.include.list": "**REDACTED**",
    "database.include.schema.changes": "true",
    "database.password": "**REDACTED**",
    "database.port": "3306",
    "database.server.id": "010101",
    "database.user": "**REDACTED**",
    "decimal.handling.mode": "string",
    "errors.log.enable": "true",
    "errors.log.include.messages": "true",
    "field.name.adjustment.mode": "avro",
    "key.converter.compatibility": "NONE",
    "name": "debezium-mysql-agribusiness-roger",
    "publication.autocreate.mode": "filtered",
    "publication.name": "**REDACTED**",
    "read.only": "true",
    "schema.history.internal.kafka.bootstrap.servers": "kafka-cluster-kafka-bootstrap:9092",
    "schema.history.internal.kafka.topic": "**REDACTED**",
    "schema.name.adjustment.mode": "avro",
    "snapshot.lock.timeout.ms": "900000",
    "snapshot.locking.mode": "minimal",
    "snapshot.mode": "no_data",
    "tasks.max": "1",
    "tombstones.on.delete": "false",
    "topic.prefix": "**REDACTED**",
    "value.converter.compatibility": "NONE"
}`

---

`AWS RDS MySQL 8.0.42`

---

`Connectors with snapshot.mode=no_data should not generate snapshot metrics`

---

**What behavior do you see?**

`debezium.mysql:name=null,type=connector-metrics,attribute=RemainingTableCount shows remaining snapshot tables with the total number of included tables forever for a connector with snapshot.mode=no_data`

---

`3.3.1 is fairly recent, but I haven't tested 3.4 yet`

---

`This bug doesn't show up in logs, only in the JMX metrics`

---

`Create a MySQL connection in streaming only mode (snapshot.mode=no_data) and look into the aforementioned JMX metric`

minleejae:
Fixes debezium/dbz#1479

## Description

When a connector runs with `snapshot.mode=no_data` (or `schema_only`), the `RemainingTableCount` JMX metric stays at the full captured-table count forever, even though no table is ever data-snapshotted.

**Root cause:** during the snapshot, `RelationalSnapshotChangeEventSource.doExecute()` always calls `SnapshotProgressListener.monitoredDataCollectionsDetermined(...)`, which populates both the captured-table set (backing `TotalTableCount`/`CapturedTables`) and the remaining-table set (backing `RemainingTableCount`). The remaining-table set is only ever drained by `dataCollectionSnapshotCompleted(...)`, which is invoked exclusively from `createDataEvents(...)` — i.e. only while data is being snapshotted. With `no_data`, the "Snapshot step 7 - Snapshotting data" stage is skipped, so the remaining-table set is never drained and the metric is stuck at the captured-table count.

**Fix:** when the data snapshot step is skipped, mark every captured table as completed with zero rows. This drains `RemainingTableCount` to `0` while leaving `TotalTableCount`/`CapturedTables` intact. This mirrors the existing handling for tables without a snapshot select query (which already calls `dataCollectionSnapshotCompleted(..., 0)`).

The fix lives in the shared `RelationalSnapshotChangeEventSource`, so it applies uniformly to all relational connectors (MySQL, MariaDB, PostgreSQL, Oracle, SQL Server, Db2). Modes that skip the snapshot entirely (e.g. `never`) are unaffected, since they never enter `doExecute()`.

### Tests
- Added an assertion to the shared `AbstractMetricsTest.assertSnapshotNotExecutedMetrics()` that `RemainingTableCount == 0` for streaming-only (`no_data`) connectors. Each relational connector's metrics IT maps `noSnapshot` to `NO_DATA`, so this exercises the path on all of them. Verified RED (`expected: 0 but was: 1`) without the fix and GREEN with it on `MySqlMetricsIT#testStreamingOnlyMetrics`.
- Added `SnapshotMeterTest`, a unit test covering the `TotalTableCount`/`RemainingTableCount` bookkeeping (previously untested).

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
+++ b/debezium-connector-common/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
@@ -199,6 +199,12 @@ public abstract class RelationalSnapshotChangeEventSource<P extends Partition, O
             else {
                 LOGGER.info("Snapshot step 7 - Skipping snapshotting of data");
                 releaseDataSnapshotLocks(ctx);
+                // debezium/dbz#1479: no data is snapshotted (e.g. snapshot.mode=no_data/schema_only), so mark every
+                // captured table as completed with zero rows; otherwise the RemainingTableCount metric would
+                // stay at the full captured-table count forever, since only the data snapshot drains it.
+                for (TableId tableId : ctx.capturedTables) {
+                    snapshotProgressListener.dataCollectionSnapshotCompleted(ctx.partition, tableId, 0);
+                }
                 ctx.offset.preSnapshotCompletion();
                 ctx.offset.postSnapshotCompletion();
             }
--- a/debezium-connector-common/src/test/java/io/debezium/pipeline/meters/SnapshotMeterTest.java
+++ b/debezium-connector-common/src/test/java/io/debezium/pipeline/meters/SnapshotMeterTest.java
@@ -1,0 +1,61 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.pipeline.meters;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.util.List;
+
+import org.junit.jupiter.api.Test;
+
+import io.debezium.doc.FixFor;
+import io.debezium.relational.TableId;
+import io.debezium.util.Clock;
+
+/**
+ * Unit tests for {@link SnapshotMeter}, in particular the {@code TotalTableCount}/{@code RemainingTableCount}
+ * bookkeeping that backs the snapshot JMX metrics.
+ */
+public class SnapshotMeterTest {
+
+    private final SnapshotMeter meter = new SnapshotMeter(Clock.system());
+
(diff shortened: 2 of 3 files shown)


debezium-connector-common/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
```diff
@@ -199,6 +199,12 @@ public SnapshotResult<O> doExecute(ChangeEventSourceContext context, O previousO
             else {
                 LOGGER.info("Snapshot step 7 - Skipping snapshotting of data");
                 releaseDataSnapshotLocks(ctx);
+                // DBZ-1479: no data is snapshotted (e.g. snapshot.mode=no_data/schema_only), so mark every
```
jpechane: Please use `debezium/dbz#1479` here and othr places, `DBZ-` is the obsolee Jira id

jpechane (approved):
@minleejae LGTM, thanks. Can be merged when CI passes.

Naros (approved):

Naros:
Test failures are unrelated.  Thanks @minleejae, applied!

PR #7576, opened 2026-06-29
debezium/dbz#1174 Add JMX metrics to the JDBC sink connector

### LINKED ISSUE #1174: Add JMX metrics on JDBC connector [DBZ-7261]
Migrated from [DBZ-7261](https://issues.redhat.com/browse/DBZ-7261)

h1. Enhancement

Like asked in the following conversation : https://debezium.zulipchat.com/#narrow/stream/383534-community-jdbc/topic/JMX.20metrics

Would it be possible to add JMX metrics on the JDBC connector ? this will provided a uniform view with other existing Debezium connector

minleejae:
Fixes debezium/dbz#1174

## Description

The JDBC sink connector currently exposes no Debezium JMX metrics. Kafka Connect already provides built-in sink-task metrics (e.g. `sink-record-send-rate`, `put-batch-avg-time-ms`), but those stop at the connector's `put()` boundary and report nothing about the **target database** — they cannot tell you what was actually applied, nor whether the target connection is up. This PR adds Debezium-domain metrics for the JDBC sink so it offers a uniform view alongside the other Debezium connectors.

### Metrics

Registered while the task runs, under:

```
debezium.jdbc:type=connector-metrics,context=sink,server=<connector>,task=<id>
```

| Attribute | Description |
|---|---|
| `TotalNumberOfUpsertEventsSeen` | upsert (insert/update) operations applied to the target |
| `TotalNumberOfDeleteEventsSeen` | delete operations applied to the target |
| `TotalNumberOfTruncateEventsSeen` | truncate operations applied to the target |
| `NumberOfFilteredEvents` | records skipped without being applied (unresolvable table, deletes/truncates disabled, schema-change events) |
| `Connected` | whether the target database connection is established |

**Why upsert (not insert/update split):** the sink models a change as delete / truncate / otherwise-upsert. It cannot reliably distinguish an insert from an update — flattened records (the common `ExtractNewRecordState` setup) carry no `op` field, and the sink applies both `c` and `u` as an upsert. Counting upserts reflects what the connector actually does and stays consistent regardless of flattening.

**Why not throughput/batch counts:** those duplicate Kafka Connect's built-in sink-task metrics, so they are intentionally omitted. The metrics here are limited to what Connect does not provide (target-side operation breakdown and connection health).

### Implementation

- New `io.debezium.connector.jdbc.metrics` package: `JdbcSinkConnectorMetricsMXBean` + `JdbcSinkConnectorMetrics` (thread-safe `AtomicLong`/`AtomicBoolean` counters).
- Registration reuses `io.debezium.pipeline.JmxUtils`, so the MBean lives in the standard `debezium.<type>` domain and benefits from the existing re-registration retry handling on connector restarts.
- Wired into `JdbcSinkConnectorTask` (register/`connected(true)` on start, unregister/`connected(false)` on stop) and `JdbcChangeEventSink.execute()` (counts at the record-classification points).
- Connector name and task id are taken from the standard connector properties (`name`, `task.id`).

### Tests
- `JdbcSinkConnectorMetricsTest` — unit tests for the counters and the JMX register/unregister round-trip (no database).
- `AbstractJdbcSinkMetricsTest` + `mysql/JdbcSinkMetricsIT` — integration test that drives the real sink task against a Testcontainers MySQL, consumes create/update/delete records, asserts the metric values via `MBeanServer.getAttribute(...)`, and verifies the MBean is unregistered on stop.

Documentation: added a Monitoring section to `connectors/jdbc.adoc`.

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
@@ -33,6 +33,7 @@ import io.debezium.openlineage.DebeziumOpenLineageEmitter;
 import io.debezium.openlineage.dataset.DatasetMetadata;
 import io.debezium.sink.DebeziumSinkRecord;
 import io.debezium.sink.spi.ChangeEventSink;
+import io.debezium.sink.spi.SinkProgressListener;
 import io.debezium.util.Stopwatch;
 
 /**
@@ -52,14 +53,16 @@ public class JdbcChangeEventSink implements ChangeEventSink {
 
     private final RecordWriter recordWriter;
     private final ConnectorContext connectorContext;
+    private final SinkProgressListener progressListener;
 
     public JdbcChangeEventSink(JdbcSinkConnectorConfig config, StatelessSession session, DatabaseDialect dialect, RecordWriter recordWriter,
-                               ConnectorContext connectorContext) {
+                               ConnectorContext connectorContext, SinkProgressListener progressListener) {
         this.config = config;
         this.dialect = dialect;
         this.session = session;
         this.recordWriter = recordWriter;
         this.connectorContext = connectorContext;
+        this.progressListener = progressListener;
 
         final DatabaseVersion version = this.dialect.getVersion();
         LOGGER.info("Database version {}.{}.{}", version.getMajor(), version.getMinor(), version.getMicro());
@@ -75,6 +78,7 @@ public class JdbcChangeEventSink implements ChangeEventSink {
 
             if (record.isSchemaChange()) {
                 SCHEMA_CHANGE_LOGGER.warn("Ignored schema change event for topic '{}'. " + FOUND_SCHEMA_CHANGE_RECORD_MSG, record.topicName());
+                progressListener.filtered();
                 continue;
             }
 
@@ -82,12 +86,14 @@ public class JdbcChangeEventSink implements ChangeEventSink {
             if (null == collectionId) {
                 LOGGER.warn("Ignored to write record from topic '{}' partition '{}' offset '{}'. No resolvable table name", record.topicName(), record.partition(),
                         record.offset());
+                progressListener.filtered();
                 continue;
             }
 
             if (record.isTruncate()) {
                 if (!config.isTruncateEnabled()) {
                     LOGGER.debug("Truncates are not enabled, skipping truncate for topic '{}'", record.topicName());
+                    progressListener.filtered();
                     continue;
                 }
 
@@ -98,6 +104,7 @@ public class JdbcChangeEventSink implements ChangeEventSink {
                 try {
                     final TableDescriptor table = recordWriter.checkAndApplyTableChangesIfNeeded(collectionId, record);
                     recordWriter.writeTruncate(table.getId());
+                    progressListener.truncated();
                     continue;
                 }
                 catch (SQLException | JDBCException e) {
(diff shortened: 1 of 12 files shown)


jpechane:
Triggering the rebuild to see if the NPE is permanent or not.

jpechane:
Please take a look at the source connectors. The connector code is not hooked directly into metrics code but uses indirect interfaces like `SnapshotProgressListener` and `StreamingProgressListener`. It would be good to follow the same pattern on the sink side.

minleejae:
@jpechane   I updated the JDBC sink to report progress through a listener instead of updating the JMX metrics directly. [de32e6b](https://github.com/debezium/debezium/pull/7576/commits/de32e6bbdf748719e4a9eba5d9bcbce200cd2477)

Naros (commented):
Should we consider also tracking the number of schema changes the sink applies when `schema.evolution` is enabled @minleejae ?

minleejae:
Good idea — added in c2950757d5.

Naros (approved):

Naros:
Thanks, applied @minleejae :tada:

mfvitale:
@Naros Unfortunately this broke Debezium server jdbc sink. 
Wondering if now that jdbc sink is in the DS we should build it when a change occurs in the `debezium-connector-jdbc`

@minleejae Could you do the CI changes so it will trigger also the DS build when there is a change on the jdbc connector? 

Fixed in https://github.com/debezium/debezium-server/pull/291

minleejae:
Hi @mfvitale 
Sorry for the breakage, and thank you for the quick fix. I opened #7622, which adds the existing jdbc-changed filter (covering debezium-sink/** and debezium-connector-jdbc/**) to the build_server job condition, so the Debezium Server build now runs on PRs touching the JDBC sink connector.

rk3rn3r:
@jpechane @Naros I have a some generic questions on this one:

1) Those metrics are applied on the event sink side before buffers. This does not guarantee that they get processed downstream and they might not be accurate, right? The initial comment says `they cannot tell you what was actually applied`, the point where they were added still don't guarantee accuracy of those numbers, do they? Furthermore progress of the processed messages will show sink progress, no?
The new metrics only report what their name says, how many events of a type were seen. Not how many were processed. @minleejae 
It seems there is also a increment of 1 applied when an entire buffer was flushed, could this be?

2) Instead of adding it statically like this, did we ever consider using AOP for adding behavior like metrics processing?

3) I also miss a NOP-Listener. Did we consider adding one and store this as ori

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcChangeEventSink.java
```diff
@@ -139,13 +148,22 @@ public void execute(Collection<SinkRecord> records) {
                 }
 
                 flushBufferRecordsWithRetries(collectionId, getRecordsToFlush(upsertBufferByTable, collectionId, record));
+                recordWriteOperation();
```
rk3rn3r, line 151: Am I right, that counts a single event as processsed in all those cases:

1) there was nothing (not enough records in the buffer / batch.size not reached) sent

2) multiple records from the buffer were sent downstream (batch.size number of events)

3) a single record from the buffer was sent (correct case)
@minleejae

minleejae: Hi @rk3rn3r Thanks for the review.

Yes, +1 in all three cases — the unit being counted is the event, not the flush. `getRecordsToFlush()` always adds the current record to the buffer in the same iteration, so each event is counted once as it enters the write path:

1. Nothing sent: the record was just buffered — it is still flushed before `execute()` returns (the trailing `flushBuffers()`), so this is early counting, not a miscount.
2. `batch.size` records sent: the other records were already counted in their own iterations — no double counting in the totals.
3. Single record sent: counted once.

---
Two caveats, consistent with the `...EventsSeen` naming: counting is at-least-once across task restarts (redelivered records are counted again, as with other connector metrics), and with the reduction buffer the counts reflect events seen rather than deduplicated rows written. If strict "applied" semantics are preferred, the increments could move to after each successful flush commit.

rk3rn3r:
I am addressing my concerns in the mentioned PR.

PR #7580, opened 2026-06-30
debezium/dbz#2095 Fix recovery procedure for PostgreSQL with failover slot

atorik:
Fixes debezium/dbz#2095

## Description

Commit 32e94290a0 updates the recovery procedure for PostgreSQL 17 clusters that use failover replication slots.

Commit b0f4cc232b1 fixes the procedure linked from that update. The previous text could be read as instructing users to set `synchronized_standby_slots` to the replication slot name used by Debezium. However, `synchronized_standby_slots` should reference the physical replication slot used for standby synchronization, so this commit updates the wording and example accordingly.

## PR Checklist

I think the latter commit is "tightly coupled" with the former one, but if you think it should be handled separately, feel free to exclude it from this PR.


jpechane:
@atorik Applied, thanks

PR #7586, opened 2026-07-01
debezium/dbz#2163 Replace throwing helper with exception factory to remove dead code

minleejae:
## Summary

`AbstractType.throwUnexpectedValue(Object)` in the JDBC sink was declared `void` but always threw a `ConnectException`. Because the compiler cannot know the method never returns, every value-returning caller had to append an unreachable `return null;`.

This replaces the throwing helper with an exception factory:

```java
protected ConnectException unexpectedValue(Object value) {
    return new ConnectException(...);
}
```

Callers now use `throw unexpectedValue(value);`. The `throw` statement makes control-flow termination explicit to the compiler, so the four unreachable `return null;` statements are removed. Behavior is unchanged — the same `ConnectException` is thrown.

## Changes (4 files)
- `type/AbstractType.java`: `void throwUnexpectedValue(...)` → `ConnectException unexpectedValue(...)` (returns the exception instead of throwing; drops the now-unnecessary `throws` clause). `requireStruct` uses `throw unexpectedValue(value)`.
- `type/connect/ConnectDecimalType.java`, `dialect/mysql/BigIntUnsignedType.java`, `dialect/mysql/BitType.java`: `throwUnexpectedValue(value); return null;` → `throw unexpectedValue(value);`.

In `BitType.toBitString` the removed `return null;` was true dead code (unreachable statement inside an `if` block already followed by a normal return).

## Notes
- The helper is `protected` in `AbstractType`; it is referenced only at these four call sites and is not overridden anywhere, so the rename is safe.

## Validation
- `mvn -pl debezium-connector-jdbc -am compile` passes
- Unit tests pass: `BitTypeTest`, `BigIntUnsignedTypeTest` (mysql + postgres) — 19 tests, 0 failures
- `checkstyle:check` and `impsort` pass; `git diff --check` clean

Closes debezium/dbz#2163

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/BigIntUnsignedType.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/BigIntUnsignedType.java
@@ -81,7 +81,6 @@ class BigIntUnsignedType extends AbstractType {
             return new BigDecimal(string);
         }
 
-        throwUnexpectedValue(value);
-        return null;
+        throw unexpectedValue(value);
     }
 }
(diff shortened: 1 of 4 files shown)


jpechane:
@minleejae Applied, thanks

PR #7532, opened 2026-06-15
debezium/dbz#2020 Fix incremental snapshot on tables with generated columns

ganeshannt:
Fixes debezium/dbz#2020

## Description

Incremental snapshot fails with `Column 'X' not found in result set` when a PostgreSQL table has generated columns and uses pgoutput plugin without `column.exclude.list` configured.

**Root cause:** `AbstractChunkQueryBuilder.buildProjection()` defaulted to `SELECT *` when no column filter was configured, but `refreshTableSchema()` already removed generated columns from the in-memory table model (since pgoutput doesn't deliver them). This caused a mismatch when mapping result set columns.

**Changes:**
* Always build explicit column list from `table.columns()` instead of defaulting to `*`
* Column names are properly quoted via `jdbcConnection.quoteIdentifier()`
* Column filter logic preserved when configured

**Testing:**
+ Added regression test for pgoutput with generated column without `column.exclude.list`
+ Verified existing `ignoreGeneratedColumn` test (with column.exclude.list) still passes
+ Verified compatibility with Oracle PhysicalRowIdentifier pattern

## PR Checklist

- [x] I have read the [contribution guidelines](https://github.com/debezium/debezium/blob/main/CONTRIBUTING.md) and the [governance document](https://github.com/debezium/governance/blob/main/GOVERNANCE.md) on PR expectations.
- [x] Minimal changes to code not directly related to your change (e.g. no unnecessary formatting changes or refactoring to existing code)
- [x] One feature/change per PR unless tightly coupled
- [x] Do a rebase on upstream `main`

PR #7588, opened 2026-07-01
debezium/dbz#2020 Fix incremental snapshot on tables with generated columns

ganeshannt:
Fixes debezium/dbz#2020

## Summary

Incremental snapshot fails with `Column 'X' not found in result set` when a PostgreSQL table has `GENERATED` columns and no `column.exclude.list` is configured.

**Root cause:** The Postgres JDBC driver does not populate `IS_GENERATEDCOLUMN`, so `Column.isGenerated()` stays false. For pgoutput, `refreshFromIncrementalSnapshot()` then prunes generated columns from the in-memory `Table`, while the chunk query still used `SELECT *`, so the JDBC result set contained columns the pruned table no longer knew about.

**Fix:**
- Populate `Column.isGenerated()` on Postgres at schema discovery via `pg_attribute.attgenerated` (STORED on PG12+, VIRTUAL on PG18+)
- Keep a Postgres-side map of generated column names (`PostgresSchema.tableIdToGeneratedColumns`) so projection still knows about them after the in-memory `Table` is pruned
- `PostgresChunkQueryBuilder` uses that map through `hasAdditionalGeneratedColumns` / `isAdditionalGeneratedColumn` hooks
- `AbstractChunkQueryBuilder.buildProjection()` keeps the two-branch behavior agreed on Zulip: `SELECT *` when nothing is filtered or generated (avoids widening the schemaChanges race from DBZ-6000 / DBZ-7532); explicit column list only when column filters or generated columns apply
- Treat MySQL/MariaDB `AUTO_INCREMENT` carefully: those connectors set both `autoIncremented` and `generated`, so projection only omits generated columns that are not auto-incremented
- Add `Column.isInvisible()` / `ColumnEditor.invisible()` to the relational model + schema-history SerDe (optional, default false; no consumer yet, Oracle follow-up)
- Postgres IT regression + unit tests for both projection paths and the AUTO_INCREMENT guard

Leaving `refreshFromIncrementalSnapshot()` pruning as-is for this PR. Broader cleanup (stop pruning / filter at schema-build) is a follow-up.

Supersedes #7532 (will close once this lands).

Related Zulip: https://debezium.zulipchat.com/#narrow/channel/302533-dev/topic/DBZ-2020.3A.20generated.20columns.20in.20incremental.20snapshot

## PR Checklist

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
+++ b/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
@@ -91,15 +91,18 @@ public abstract class AbstractChunkQueryBuilder<T extends DataCollectionId>
     }
 
     protected String buildProjection(Table table) {
-        String projection = "*";
-        if (connectorConfig.isColumnsFiltered()) {
-            TableId tableId = table.id();
-            projection = table.columns().stream()
-                    .filter(column -> columnFilter.matches(tableId.catalog(), tableId.schema(), tableId.table(), column.name()))
-                    .map(column -> jdbcConnection.quoteIdentifier(column.name()))
-                    .collect(Collectors.joining(", "));
-        }
-        return projection;
+        // DBZ-2020: always build an explicit column list from the in-memory Table, same pattern as
+        // the initial snapshot (getPreparedColumnNames). Generated columns that were removed from the
+        // Table by refreshFromIncrementalSnapshot are already absent; any still marked isGenerated()
+        // (e.g. connectors that mark but do not remove) are filtered here.
+        final boolean columnsFiltered = connectorConfig.isColumnsFiltered();
+        final TableId tableId = table.id();
+        return table.columns().stream()
+                .filter(column -> !column.isGenerated())
+                .filter(column -> !columnsFiltered
+                        || columnFilter.matches(tableId.catalog(), tableId.schema(), tableId.table(), column.name()))
+                .map(column -> jdbcConnection.quoteIdentifier(column.name()))
+                .collect(Collectors.joining(", "));
     }
 
     protected void addLowerBound(IncrementalSnapshotContext<T> context, Table table, Object[] boundaryKey, StringBuilder sql) {
--- a/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
+++ b/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
@@ -91,18 +91,40 @@ public abstract class AbstractChunkQueryBuilder<T extends DataCollectionId>
     }
 
     protected String buildProjection(Table table) {
-        // DBZ-2020: always build an explicit column list from the in-memory Table, same pattern as
-        // the initial snapshot (getPreparedColumnNames). Generated columns that were removed from the
-        // Table by refreshFromIncrementalSnapshot are already absent; any still marked isGenerated()
-        // (e.g. connectors that mark but do not remove) are filtered here.
+        // DBZ-2020: only fall back to an explicit column list when we actually need to drop
+        // columns from the projection (generated columns, or column include/exclude filters).
+        // In the common case where neither applies, keep emitting SELECT * so we do not widen
+        // the schemaChanges race window observed in DBZ-6000 / DBZ-7532.
         final boolean columnsFiltered = connectorConfig.isColumnsFiltered();
+        final boolean hasGeneratedColumns = table.columns().stream().anyMatch(Column::isGenerated)
+                || hasAdditionalGeneratedColumns(table);
+        if (!columnsFiltered && !hasGeneratedColumns) {
+            return "*";
+        }
         final TableId tableId = table.id();
         return table.columns().stream()
                 .filter(column -> !column.isGenerated())
+                .filter(column -> !isAdditionalGeneratedColumn(table, column.name()))
                 .filter(column -> !columnsFiltered
                         || columnFilter.matches(tableId.catalog(), tableId.schema(), tableId.table(), column.name()))
                 .map(column -> jdbcConnection.quoteIdentifier(column.name()))
                 .collect(Collectors.joining(", "));
+    }
+
+    /**
+     * Hook for connectors that track generated columns outside the in-memory {@link Table}
+     * (for example after {@code refreshFromIncrementalSnapshot} has pruned them). Default is none.
+     */
+    protected boolean hasAdditionalGeneratedColumns(Table table) {
+        return false;
+    }
+
+    /**
+     * Hook for connectors that track generated columns outside the in-memory {@link Table}.
+     * Default is never generated.
+     */
+    protected boolean isAdditionalGeneratedColumn(Table table, String columnName) {
+        return false;
     }
 
     protected void addLowerBound(IncrementalSnapshotContext<T> context, Table table, Object[] boundaryKey, StringBuilder sql) {
--- a/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
+++ b/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
@@ -96,19 +96,29 @@ public abstract class AbstractChunkQueryBuilder<T extends DataCollectionId>
         // In the common case where neither applies, keep emitting SELECT * so we do not widen
         // the schemaChanges race window observed in DBZ-6000 / DBZ-7532.
         final boolean columnsFiltered = connectorConfig.isColumnsFiltered();
-        final boolean hasGeneratedColumns = table.columns().stream().anyMatch(Column::isGenerated)
+        // MySQL/MariaDB mark AUTO_INCREMENT (and SERIAL / ON UPDATE) as both autoIncremented and
+        // generated. Only treat "true" generated columns as ones we must omit from the projection.
+        final boolean hasGeneratedColumns = table.columns().stream().anyMatch(this::isGeneratedNonAutoIncrementColumn)
                 || hasAdditionalGeneratedColumns(table);
         if (!columnsFiltered && !hasGeneratedColumns) {
             return "*";
         }
         final TableId tableId = table.id();
         return table.columns().stream()
-                .filter(column -> !column.isGenerated())
+                .filter(column -> !isGeneratedNonAutoIncrementColumn(column))
                 .filter(column -> !isAdditionalGeneratedColumn(table, column.name()))
                 .filter(column -> !columnsFiltered
                         || columnFilter.matches(tableId.catalog(), tableId.schema(), tableId.table(), column.name()))
                 .map(column -> jdbcConnection.quoteIdentifier(column.name()))
                 .collect(Collectors.joining(", "));
+    }
+
+    /**
+     * True when the column is generated and not auto-incremented. MySQL/MariaDB set both flags for
+     * {@code AUTO_INCREMENT}, so those columns must remain in the chunk projection.
+     */
+    private boolean isGeneratedNonAutoIncrementColumn(Column column) {
+        return column.isGenerated() && !column.isAutoIncremented();
     }
 
     /**
--- a/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
+++ b/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
@@ -93,15 +93,18 @@ public abstract class AbstractChunkQueryBuilder<T extends DataCollectionId>
     }
 
     protected String buildProjection(Table table) {
-        String projection = "*";
-        if (connectorConfig.isColumnsFiltered()) {
-            TableId tableId = table.id();
-            projection = table.columns().stream()
-                    .filter(column -> columnFilter.matches(tableId.catalog(), tableId.schema(), tableId.table(), column.name()))
-                    .map(column -> jdbcConnection.quoteIdentifier(column.name()))
-                    .collect(Collectors.joining(", "));
-        }
-        return projection;
+        // DBZ-2020: always build an explicit column list from the in-memory Table, same pattern as
+        // the initial snapshot (getPreparedColumnNames). Generated columns that were removed from the
+        // Table by refreshFromIncrementalSnapshot are already absent; any still marked isGenerated()
+        // (e.g. connectors that mark but do not remove) are filtered here.
+        final boolean columnsFiltered = connectorConfig.isColumnsFiltered();
+        final TableId tableId = table.id();
+        return table.columns().stream()
+                .filter(column -> !column.isGenerated())
+                .filter(column -> !columnsFiltered
+                        || columnFilter.matches(tableId.catalog(), tableId.schema(), tableId.table(), column.name()))
+                .map(column -> jdbcConnection.quoteIdentifier(column.name()))
+                .collect(Collectors.joining(", "));
     }
 
     public void addLowerBound(List<Column> pkColumns, Object[] boundaryKey, StringBuilder condition, boolean inclusiveFinal) {
--- a/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
+++ b/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
@@ -93,34 +93,19 @@ public abstract class AbstractChunkQueryBuilder<T extends DataCollectionId>
     }
 
     protected String buildProjection(Table table) {
-        // DBZ-2020: only fall back to an explicit column list when we actually need to drop
-        // columns from the projection (generated columns, or column include/exclude filters).
-        // In the common case where neither applies, keep emitting SELECT * so we do not widen
-        // the schemaChanges race window observed in DBZ-6000 / DBZ-7532.
+        // DBZ-2020: keep SELECT * unless column filters apply or a connector opts in via the hooks
+        // below (Postgres pgoutput), so we avoid widening the schemaChanges race window (DBZ-6000).
         final boolean columnsFiltered = connectorConfig.isColumnsFiltered();
-        // MySQL/MariaDB mark AUTO_INCREMENT (and SERIAL / ON UPDATE) as both autoIncremented and
-        // generated. Only treat "true" generated columns as ones we must omit from the projection.
-        final boolean hasGeneratedColumns = table.columns().stream().anyMatch(this::isGeneratedNonAutoIncrementColumn)
-                || hasAdditionalGeneratedColumns(table);
-        if (!columnsFiltered && !hasGeneratedColumns) {
+        if (!columnsFiltered && !hasAdditionalGeneratedColumns(table)) {
             return "*";
         }
         final TableId tableId = table.id();
         return table.columns().stream()
-                .filter(column -> !isGeneratedNonAutoIncrementColumn(column))
                 .filter(column -> !isAdditionalGeneratedColumn(table, column.name()))
                 .filter(column -> !columnsFiltered
                         || columnFilter.matches(tableId.catalog(), tableId.schema(), tableId.table(), column.name()))
                 .map(column -> jdbcConnection.quoteIdentifier(column.name()))
                 .collect(Collectors.joining(", "));
-    }
-
-    /**
-     * True when the column is generated and not auto-incremented. MySQL/MariaDB set both flags for
-     * {@code AUTO_INCREMENT}, so those columns must remain in the chunk projection.
-     */
-    private boolean isGeneratedNonAutoIncrementColumn(Column column) {
-        return column.isGenerated() && !column.isAutoIncremented();
     }
 
     /**
--- a/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
+++ b/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
@@ -93,18 +93,40 @@ public abstract class AbstractChunkQueryBuilder<T extends DataCollectionId>
     }
 
     protected String buildProjection(Table table) {
-        // DBZ-2020: always build an explicit column list from the in-memory Table, same pattern as
-        // the initial snapshot (getPreparedColumnNames). Generated columns that were removed from the
-        // Table by refreshFromIncrementalSnapshot are already absent; any still marked isGenerated()
-        // (e.g. connectors that mark but do not remove) are filtered here.
+        // DBZ-2020: only fall back to an explicit column list when we actually need to drop
+        // columns from the projection (generated columns, or column include/exclude filters).
+        // In the common case where neither applies, keep emitting SELECT * so we do not widen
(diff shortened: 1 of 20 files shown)


debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
```diff
@@ -83,15 +83,18 @@ public String buildChunkQuery(IncrementalSnapshotContext<T> context, Table table
...
-        }
-        return projection;
+        // DBZ-2020: always build an explicit column list from the in-memory Table, same pattern as
+        // the initial snapshot (getPreparedColumnNames). Generated columns that were removed from the
+        // Table by refreshFromIncrementalSnapshot are already absent; any still marked isGenerated()
+        // (e.g. connectors that mark but do not remove) are filtered here.
+        final boolean columnsFiltered = connectorConfig.isColumnsFiltered();
+        final TableId tableId = table.id();
+        return table.columns().stream()
+                .filter(column -> !column.isGenerated())
```
Naros: Won't this cause MySQL `AUTO_INCREMENT` columns to be excluded, when they shouldn't? Perhaps guard this also as `!isAutoIncrement()`?

Additionally, I believe this leaves some dead code in the subclasses that still expect this method to return `*` in specific use cases that needs to be addressed.

Naros (changes requested):

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
ganeshannt: Thanks Chris. You’re right about MySQL/MariaDB. The DDL listeners mark AUTO_INCREMENT / SERIAL (and ON UPDATE) as both autoIncremented and generated, so a plain !isGenerated() filter would drop those columns once we expand the projection. I've adjusted the check to generated && !autoIncremented, so AUTO_INCREMENT stays selected and only actual generated columns are skipped. Covered that with unit tests as well.

On the SELECT * side, I agree always expanding was the wrong trade-off given the schemaChanges race from DBZ-6000 / DBZ-7532. Projection is back to the two-branch approach: keep * when there are no column filters and no true generated columns, and expand only when we need to omit something. For Postgres specifically, generated column names live in a side map on PostgresSchema because refreshFromIncrementalSnapshot still prunes them from the in-memory Table, and PostgresChunkQueryBuilder uses that map. Leaving the broader "stop pruning / filter at schema-build" cleanup as a follow-up.

debezium-connector-common/src/main/java/io/debezium/relational/history/ConnectTableChangeSerializer.java
```diff
@@ -121,6 +122,7 @@ private Struct toStruct(Column column) {
         struct.put(OPTIONAL_KEY, column.isOptional());
         struct.put(AUTO_INCREMENTED_KEY, column.isAutoIncremented());
         struct.put(GENERATED_KEY, column.isGenerated());
+        struct.put(INVISIBLE_KEY, column.isInvisible());
```
Naros, line 125: Could we add a test in `HistoryRecordTest` that validates the serialization of `INVISIBLE_KEY`?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresChunkQueryBuilder.java
```diff
@@ -0,0 +1,47 @@
...
+ * PostgreSQL chunk query builder that consults {@link PostgresSchema}'s generated-column side map.
+ *
+ * <p>For pgoutput, {@link PostgresSchema#refreshFromIncrementalSnapshot} prunes generated columns from
+ * the in-memory {@link Table}. Without the side map, {@code buildProjection} would see no generated
+ * columns and fall back to {@code SELECT *}, which returns those columns from the database and fails
+ * with "Column 'X' not found in result set" (DBZ-2020).</p>
+ *
+ * @param <T> data collection id type
+ */
+public class PostgresChunkQueryBuilder<T extends DataCollectionId> extends RowValueConstructorChunkQueryBuilder<T> {
```
Naros, line 24: Should we have a unit test with coverage of this with the schema side-map interaction? We have unit tests for things like `DefaultChunkQueryBuilder`, perhaps we should here?

Naros (commented):
Hi @ganeshannt this is a well structured and analyzed PR. I've added a few findings below, if you could take a look as you have time. If you could also rebase on `main`, that would be great.

debezium-connector-common/src/main/java/io/debezium/relational/history/ConnectTableChangeSerializer.java
ganeshannt: Added. HistoryRecordTest covers SerDe for invisible/generated, including older history without the field.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresChunkQueryBuilder.java
ganeshannt: added PostgresChunkQueryBuilderTest for the side-map cases after prune.

vjuranek (approved):
LGTM, but I appreciate another pair of eyes on this.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
```diff
@@ -93,15 +93,50 @@ else if (upperBound.isPresent()) {
...
+        // DBZ-2020: only fall back to an explicit column list when we actually need to drop
+        // columns from the projection (generated columns, or column include/exclude filters).
+        // In the common case where neither applies, keep emitting SELECT * so we do not widen
+        // the schemaChanges race window observed in DBZ-6000 / DBZ-7532.
+        final boolean columnsFiltered = connectorConfig.isColumnsFiltered();
+        // MySQL/MariaDB mark AUTO_INCREMENT (and SERIAL / ON UPDATE) as both autoIncremented and
+        // generated. Only treat "true" generated columns as ones we must omit from the projection.
+        final boolean hasGeneratedColumns = table.columns().stream().anyMatch(this::isGeneratedNonAutoIncrementColumn)
+                || hasAdditionalGeneratedColumns(table);
+        if (!columnsFiltered && !hasGeneratedColumns) {
```
Naros: This change not only addresses PostgreSQL but also affects every connector path, and because we may not have sufficient test coverage for auto-generated columns across all connectors, side effects may be hard to spot.

For example, Oracle snapshots include `generated=true`, but this filters them out, and they drop from the chunks. Similarly, PostgreSQL only prunes when the plugin is `pgoutput`, not `decoderbufs`. The net result, I believe, is that the incremental snapshot yields the column as `null`.

I wonder if we should have a bit more cross-database coverage here to be safe? WDYT?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSchema.java
```diff
@@ -89,6 +94,11 @@ private static TableSchemaBuilder getTableSchemaBuilder(PostgresConnectorConfig
     protected PostgresSchema refresh(PostgresConnection connection, boolean printReplicaIdentityInfo) throws SQLException {
         // read all the information from the DB
         connection.readSchema(tables(), null, null, getTableFilter(), null, true);
+        // The Postgres JDBC driver does not expose the IS_GENERATEDCOLUMN metadata slot that
+        // JdbcConnection.readSchema() relies on, so Column.isGenerated() is left false after the
+        // initial read. Populate it from pg_attribute.attgenerated so incremental snapshot and any
+        // other consumer that relies on isGenerated() can behave correctly.
+        tableIds().forEach(tableId -> refreshGeneratedColumnsMap(connection, tableId));
```
Naros: This issues one `pg_attribute` query per captured table on every full refresh, unconditionally for pgoutput, whether or not any table has a generated column. Compare `refreshToastableColumnsMap` just below, which does the same per-table pattern but is gated behind `skipRefreshSchemaOnMissingToastableData` precisely because it is expensive.

For a deployment capturing a few thousand tables that is a few thousand extra round trips at connector start, and line 147 adds one more on every DDL-triggered refresh during streaming.

Could we collapse this into a single query over the captured tables? Joining `pg_attribute` against the set of relations once and grouping by `TableId` would give the same map in one round trip, and the per-table refresh path could then read from it rather than re-querying.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSchema.java
```diff
@@ -238,6 +259,103 @@ private void refreshToastableColumnsMap(PostgresConnection connection, TableId t
...
+     * {@link Table} and the chunk projection instead of being pruned as generated.</p>
+     */
+    private List<String> trackedGeneratedColumnNames(Table table, PostgresConnection connection, TableId tableId) {
+        final List<String> allGeneratedColumnNames = readGeneratedColumnNames(connection, tableId);
+        final List<String> primaryKeyColumnNames = table.primaryKeyColumnNames();
+        if (primaryKeyColumnNames.isEmpty()) {
+            return allGeneratedColumnNames;
+        }
+        return allGeneratedColumnNames.stream()
+                .filter(name -> !primaryKeyColumnNames.contains(name))
```
Naros, line 287: This filter drops primary key columns from the list, and the same filtered list is then passed to `applyGeneratedColumnFlags` (lines 149 and 277). The result is that a generated column that is also the primary key keeps `isGenerated() == false`.

That undercuts the stated goal of the PR, which is to populate `Column.isGenerated()` correctly on Postgres, for exactly the case the PR went out of its way to handle. It also flows outward: the schema change topic will report `generated: false` for a column that is generated.

Suggest separating the two concerns. Set the flags from the full `readGeneratedColumnNames` result, and use the PK-filtered list only for `tableIdToGeneratedColumns` (the projection side map) and for the pruning decision. As a bonus that makes the `primaryKeyColumnNames.contains(x.name())` guard at line 152 do real work; right now the flag it checks was never set, so that guard cannot fire.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSchema.java
```diff
@@ -238,6 +259,103 @@ private void refreshToastableColumnsMap(PostgresConnection connection, TableId t
...
+            }
+            if (existing.isGenerated()) {
+                continue;
+            }
+            editor.updateColumn(existing.edit().generated(true).create());
+        }
+        return editor.create();
+    }
+
+    private List<String> readGeneratedColumnNames(PostgresConnection connection, TableId tableId) {
```
Naros, line 315: This is a near-verbatim copy of `refreshToastableColumnsMap` above: same joins, same bind parameters, same autocommit-and-commit dance, same `ConnectException` wrapping. Only the column predicate differs (`attgenerated <> ''` versus `attstorage in ('x','e','m')`).

Could we extract a shared private helper, something like `readColumnNames(PostgresConnection connection, TableId tableId, String predicate, String errorContext)`, and have both call it? That keeps the transaction handling in one place, which matters here because getting the commit behaviour subtly different between the two would be easy to miss in review.

Minor while you are in here: line 338 can use `Strings.isNullOrEmpty(tableId.schema())` from `io.debezium.util.Strings` instead of the null-and-length check, and line 298's `Collectors.toList()` can be `Stream.toList()`.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresChunkQueryBuilder.java
```diff
@@ -0,0 +1,47 @@
...
+
+    @Override
+    protected boolean hasAdditionalGeneratedColumns(Table table) {
+        return !schema.getGeneratedColumnsForTableId(table.id()).isEmpty();
+    }
+
+    @Override
+    protected boolean isAdditionalGeneratedColumn(Table table, String columnName) {
+        // After pruning, generated columns are absent from table.columns(), so this filter is a
+        // no-op for the stream over table.columns(). Kept for symmetry with the detection hook.
```
Naros: This comment says the filter is a no-op for the stream over `table.columns()`, but `PostgresChunkQueryBuilderTest.shouldOmitSideMapGeneratedColumnStillPresentOnTable` deliberately asserts the opposite. Is my understanding correct that the comment likely needs updating?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresChunkQueryBuilder.java
```diff
@@ -0,0 +1,47 @@
...
+        return schema.getGeneratedColumnsForTableId(table.id()).stream()
+                .anyMatch(name -> name.equalsIgnoreCase(columnName));
```
Naros: `equalsIgnoreCase` here does not match how the same names are compared elsewhere: `trackedGeneratedColumnNames` uses `primaryKeyColumnNames.contains(name)` and `applyGeneratedColumnFlags` uses `table.columnWithName(...)`, both exact. PostgreSQL identifiers are case sensitive once quoted, so a table with both `"Foo"` and `"foo"` where only one is generated would lose both from the projection.

```suggestion
                .anyMatch(columnName::equals);
```

Also worth noting that this is called once per column and re-fetches plus re-streams the list each time. Hoisting the names into a `Set<String>` (either a field refreshed alongside the side map, or a local in `buildProjection`) would make it a single lookup per column.

documentation/modules/ROOT/pages/connectors/postgresql.adoc
```diff
@@ -274,6 +274,13 @@ The {prodname} connector for PostgreSQL does not support schema changes while an
 If a schema change is performed _before_ the incremental snapshot start but _after_ sending the signal then passthrough config option `database.autosave` is set to `conservative` to correctly process the schema change.
 ====
 
+[NOTE]
+====
+When a captured table has `GENERATED` columns, the connector detects them at schema discovery time from `pg_attribute.attgenerated` and omits them from the chunk `SELECT` used during an incremental snapshot.
+Values for those columns are not present in the resulting snapshot events; recompute them on the consumer side if needed.
+Tables without any generated columns continue to use `SELECT *` and are unaffected.
+====
```
Naros, line 285: This note reads as unconditional, but the behaviour is pgoutput only. With `decoderbufs` the generated column stays in `SELECT *` and its value is emitted, which this PR's own `incrementalSnapshotKeepsGeneratedColumnValueForDecoderbufs` IT asserts. "detects them at schema discovery time" is likewise pgoutput only, since `tracksGeneratedColumns()` gates it. A user on decoderbufs reading this would expect missing values that they will in fact receive.

```suggestion
[NOTE]
====
When you use the `pgoutput` logical decoding plug-in and a captured table has `GENERATED` columns, the connector detects them at schema discovery time from `pg_attribute.attgenerated` and omits them from the chunk `SELECT` that it uses during an incremental snapshot.
Values for those columns are not present in the resulting snapshot events; recompute them on the consumer side if needed.
A `GENERATED` column that is also part of the primary key is retained, along with its computed value.

Other decoding plug-ins, such as `decoderbufs`, retain `GENERATED` columns and emit their computed values.
Tables without any generated columns continue to use `SELECT *` and are unaffected.
====
```

Naros (commented):
Hi @ganeshannt I gave this another pass and found a few more things.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
ganeshannt: @Naros I reduced the blast radius in a later commit. Updated `buildProjection()` no longer checks `column.isGenerated()` for every connector — pruning now only happens through the `isAdditionalGeneratedColumn()` hook, which defaults to `false` and is overridden solely in `PostgresChunkQueryBuilder`, scoped to pgoutput. So Oracle (and all other connectors) keep SELECT * and generated columns aren't dropped or nulled there. Let me know if you'd still like broader cross-database test coverage added.

documentation/modules/ROOT/pages/connectors/postgresql.adoc
ganeshannt: I updated the note to match the behaviour

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSchema.java
ganeshannt: Agreed. I refactored the duplicated metadata lookup path into shared helpers (`readColumnNames(...)` and `runColumnMetadataQuery(...)`) so generated/toastable lookups share the same query, bind, commit, and exception handling flow. I also switched schema normalization to `Strings.isNullOrEmpty(...)` and used `.toList()` in the stream path.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSchema.java
ganeshannt: Great point. I changed full refresh to batch generated-column metadata lookup in a single query across captured `(schema, table)` pairs, so we no longer do one `pg_attribute` round trip per captured table at startup.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSchema.java
ganeshannt: I split the concerns as suggested: `applyGeneratedColumnFlags(...)` now uses the full generated-column list, and only the projection side-map uses the PK-filtered list. This keeps generated PK columns correctly marked as generated while still retaining them in chunk projection/pruning.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresChunkQueryBuilder.java
ganeshannt: my bad. that comment was stale. I updated it to match behavior and tests: side-map entries can still apply when a column is present in `table.columns()`, so projection filtering should always consult the side map.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSchema.java
```diff
@@ -89,6 +98,9 @@ private static TableSchemaBuilder getTableSchemaBuilder(PostgresConnectorConfig
     protected PostgresSchema refresh(PostgresConnection connection, boolean printReplicaIdentityInfo) throws SQLException {
         // read all the information from the DB
         connection.readSchema(tables(), null, null, getTableFilter(), null, true);
+        // The Postgres JDBC driver does not reliably populate IS_GENERATEDCOLUMN metadata.
+        // Populate generated-column metadata from pg_attribute.attgenerated.
+        refreshGeneratedColumnsMap(connection);
```
Naros, line 103: This runs inside the initial snapshot's transaction. `PostgresSnapshotChangeEventSource#connectionCreated` pins the exported snapshot with `SET TRANSACTION ISOLATION LEVEL REPEATABLE READ; SET TRANSACTION SNAPSHOT ...` on the main connection (autocommit is off), and `readTableStructure` then calls `schema.refresh(jdbcConnection, false)` before any data is read. This new call goes through `runColumnMetadataQuery`, which commits whenever autocommit is off (line 396), so the snapshot transaction is ended before `createDataEvents` and the table reads happen under a fresh transaction without the exported snapshot.

That matches what CI shows: all seven PostgreSQL failures are `expected: "c" but was: "r"`, i.e. rows inserted after slot creation are picked up by the snapshot as reads.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSchema.java
```diff
@@ -205,37 +229,180 @@ private void refreshToastableColumnsMap(PostgresConnection connection, TableId t
...
+        final Map<TableId, List<String>> allGeneratedColumnsByTable = readGeneratedColumnNames(connection, tableIds);
+        tableIds.forEach(tableId -> refreshGeneratedColumnsMap(tableId, allGeneratedColumnsByTable.getOrDefault(tableId, List.of())));
+    }
+
+    /**
+     * Populates {@link Column#isGenerated()} from {@code pg_attribute.attgenerated} (the Postgres JDBC
+     * driver does not surface {@code IS_GENERATEDCOLUMN}) and records the names in
+     * {@link #tableIdToGeneratedColumns}. Runs for pgoutput only; see {@link #tracksGeneratedColumns()}.
+     */
+    private void refreshGeneratedColumnsMap(PostgresConnection connection, TableId tableId) {
```
Naros, line 254: I believe this is now dead-code and can be removed.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSchema.java
```diff
@@ -205,37 +229,180 @@ private void refreshToastableColumnsMap(PostgresConnection connection, TableId t
...
+        if (tableIds.isEmpty()) {
+            return Collections.emptyMap();
+        }
+
+        // See applyGeneratedColumnFlags for why attgenerated <> '' captures both STORED (PG12+) and
+        // VIRTUAL (PG18+) generated columns.
+        final Set<TableId> capturedLookupKeys = tableIds.stream()
+                .map(tableId -> new TableId(null, schemaName(tableId), tableId.table()))
+                .collect(Collectors.toSet());
+        final String tablePredicates = String.join(" OR ", Collections.nCopies(capturedLookupKeys.size(), "(ns.nspname = ? and tbl.relname = ?)"));
```
Naros, line 330: This emits one `(ns.nspname = ? and tbl.relname = ?)` pair per captured table, so the statement carries two bind parameters per table. pgjdbc caps a single statement at 32767 parameters, so a deployment capturing more than roughly 16k tables would fail at connector start. 

Since the result rows are already filtered client-side through `capturedTablesByName`, a single `ns.nspname = ANY(?)` over the distinct captured schemas (array parameter) would give the same map with one bind variable. 

Chunking the predicate list is the other option if you would rather keep the exact-match shape.

If we leave the 2 bind parameters in, we should document this limit.

Naros (commented):
Hi @ganeshannt so I've left a few more findings. The one about the commit during initial snapshot is the most critical, as it impacts snapshot consistency.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSchema.java
ganeshannt: removed

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSchema.java
ganeshannt: fixed. batched generated-column refresh no longer commits inside the pinned exported-snapshot transaction. `runColumnMetadataQuery` now takes a `commitAfterQuery` flag; the full-refresh path passes `false` so the snapshot owner keeps the transaction open through `createDataEvents`, while the per-table/toastable paths (outside the pinned snapshot) keep committing to avoid idle-in-transaction. This clears the `expected: "c" but was: "r"` failures.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSchema.java
ganeshannt: switched to a single `ns.nspname = ANY(?)` array parameter over the distinct captured schemas, so the statement uses one bind variable regardless of table count; rows are still narrowed to the captured tables client-side via `capturedTablesByName`. No 16k-table limit to document now.

PR #7592, opened 2026-07-02
debezium/dbz#2173 Fix chunked snapshot failure on tables whose names require quoting

mly-zju:
# What

Debezium's chunked/parallel snapshot algorithm (used when `snapshot.max.threads > 1` and `legacy.snapshot.max.threads` is not enabled) computes a per-table row count in `RelationalSnapshotChangeEventSource.rowCountForTableChunked`:

```java
"SELECT COUNT(1) FROM %s".formatted(jdbcConnection.getQualifiedTableName(tableId))
```

`JdbcConnection.getQualifiedTableName()` returns `schema + "." + table` **without quoting**. For any schema/table name that requires quoting (special characters, dots, `#`, `/`, leading digit, uppercase, reserved words) this generates invalid SQL and crashes the snapshot. For example a Postgres table `table_with_pk.1#2/3` yields:

```
SELECT COUNT(1) FROM public.table_with_pk.1#2/3
```

which PostgreSQL rejects with `ERROR: syntax error at or near ".1"`.

This changes the call site to `quotedTableIdString(tableId)`, which is the same method each connector already uses to qualify tables in its snapshot `SELECT`, so the row-count query is quoted consistently per dialect:

- **Postgres / SQL Server** – double-quoted (`"schema"."table"` / `"db"."schema"."table"`); SQL Server's snapshot select literally uses `jdbcConnection.quotedTableIdString(...)`.
- **MySQL / MariaDB** – backtick-quoted (`` `db`.`table` ``) via the binlog `quotedTableIdString` override.
- **Oracle** – `TableId` carries a CDB/PDB catalog that cannot appear in a qualified name, so `OracleSnapshotChangeEventSource` overrides `rowCountForTableChunked` to strip the catalog before quoting, mirroring what its `getSnapshotSelect` already does. Without this it would emit an invalid `"catalog"."schema"."table"`.

`getQualifiedTableName()` itself is intentionally left unchanged, because it is also used as an identifier / map key (not SQL) by `ReselectColumnsPostProcessor`.

# Why

Fixes https://github.com/debezium/dbz/issues/2173

This is a regression from the new chunked snapshot; the non-chunked path already quoted identifiers correctly, so these tables snapshotted fine before. Only `rowCountForTableChunked` used the unquoted form; the other chunked-path queries (chunk boundaries, chunk data selects) already quote via `quotedTableIdString` / `quoteIdentifier`.

# Test

## Integration test

Added `PostgresChunkedSnapshotIT#shouldSnapshotChunkedTableWhoseNameRequiresQuoting`, which creates a primary-keyed table named `public."table_with_pk.1#2/3"`, runs a chunked snapshot with `snapshot.max.threads=2`, and asserts all rows are captured and the chunked worker pool was used. Without the fix this fails during snapshot with `ERROR: syntax error at or near ".1"`.

```bash
mvn test -pl debezium-connector-postgres -Dit.test=PostgresChunkedSnapshotIT#shouldSnapshotChunkedTableWhoseNameRequiresQuoting -Dcheckstyle.skip=true -Dformat.skip=true -Denforcer.skip=true -Drevapi.skip=true
```

The existing `OracleChunkedSnapshotIT` covers the Oracle override (the shared change alone would have regressed it by emitting a 3-part catalog-qualified name).

## Test in real env

Reproduced the original crash in our system (a PostgreSQL data migration with `snapshot.max.threads=2` capturing tables with special-character names). With this fix in place the snapshot completes successfully.

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
+++ b/debezium-connector-common/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
@@ -1100,8 +1100,10 @@ public abstract class RelationalSnapshotChangeEventSource<P extends Partition, O
 
     protected Long rowCountForTableChunked(TableId tableId) throws SQLException {
         // todo: snapshot select overrides?
+        // Use the quoted, fully-qualified identifier (as the rest of the snapshot SQL does) so that
+        // schema/table names requiring quoting - special characters, reserved words, etc. - remain valid.
         return jdbcConnection.queryAndMap(
-                "SELECT COUNT(1) FROM %s".formatted(jdbcConnection.getQualifiedTableName(tableId)),
+                "SELECT COUNT(1) FROM %s".formatted(jdbcConnection.quotedTableIdString(tableId)),
                 rs -> rs.next() ? rs.getLong(1) : 0L);
     }
 
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
@@ -242,6 +242,14 @@ public class OracleSnapshotChangeEventSource extends RelationalSnapshotChangeEve
     }
 
     @Override
+    protected Long rowCountForTableChunked(TableId tableId) throws SQLException {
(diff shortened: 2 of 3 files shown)


jpechane:
@mly-zju Debezium 3.6 is now stable. I don't think we need to backport the fix to 3.5 so I propose to close the PR.

mly-zju:
> @mly-zju Debezium 3.6 is now stable. I don't think we need to backport the fix to 3.5 so I propose to close the PR.

@jpechane sure, closing this PR now

PR #7593, opened 2026-07-02
debezium/dbz#978 Add opt-in binlog-metadata-based schema mode for MySQL and MariaDB

### LINKED ISSUE #978: MySQL 8.x and binlog_row_metadata = FULL [DBZ-7794]
Migrated from [DBZ-7794](https://issues.redhat.com/browse/DBZ-7794)

Hello,

I have read that starting from MySQL 8 the binlog contains more information about the table being modified in the Table_map event, priors to any insert/update event. In full mode
set global binlog_row_metadata = FULL, all metadata are present, see for example:
 
{code:java}
insert into test (id, c1, c3) values (8, 0, 'bye'); {code}
from the binlog:
 
{code:java}
#240419  9:35:26 server id 1  end_log_pos 1509 CRC32 0x1dab5f9b     Table_map: `mydb`.`test` mapped to number 90
# has_generated_invisible_primary_key=0
# Columns(`id` INT NOT NULL,
#         `c1` BIT(1),
#         `c3` TEXT CHARSET utf8mb4 COLLATE utf8mb4_0900_ai_ci)
# Primary Key(id)
# at 1509
#240419  9:35:26 server id 1  end_log_pos 1555 CRC32 0x2571c7db     Write_rows: table id 90 flags: STMT_END_FBINLOG '
vh4iZhMBAAAATQAAAOUFAAAAAFoAAAAAAAEABG15ZGIABHRlc3QAAwMQ/AMBAAIGAQEAAgP8/wAE
CQJpZAJjMQJjMwgBAAwB4Jtfqx0=
vh4iZh4BAAAALgAAABMGAAAAAFoAAAAAAAEAAgAD/wAIAAAAAAMAYnll28dxJQ==
'/*!*/;
###   @3='bye'
# at 1555
#240419  9:35:26 server id 1  end_log_pos 1586 CRC32 0x7188858d     Xid = 281
COMMIT/*!*/; {code}
 
Thank you.
Best regards,
Emmanuel

minleejae:
Fixes https://github.com/debezium/dbz/issues/978
This is an opt-in mode that lets the MySQL and MariaDB connectors reconstruct the streaming table schema directly from the metadata carried by binlog `TABLE_MAP` events, instead of relying on a persisted database schema history topic.

## Motivation
When a server runs with `binlog_row_metadata=FULL` (MySQL 8.0+, MariaDB 10.5+), every `TABLE_MAP` event carries the full column metadata (names, types, character sets, `ENUM`/`SET` values, primary key) — the same self-describing information the PostgreSQL connector gets from pgoutput relation messages. In that case the connector can rebuild schema from the log itself and no longer needs a schema history topic, removing the bootstrap/recovery/operational overhead of managing that topic.

## What this adds
A new opt-in property, `binlog.metadata.based.schema` (default `false`). When set to `true`:
- The streaming table schema is reconstructed from the FULL `TABLE_MAP` metadata via a new `BinlogMetadataTableBuilder`.
- The connector behaves as a non-historized connector (like PostgreSQL): no schema history topic is created or required (an in-memory history satisfies the historized base lifecycle).
- It fails fast at startup if `binlog_row_metadata` is not `FULL`.
- For snapshot modes that do not snapshot data (for example `no_data`), the current binlog position is recorded before streaming, so a fresh connector starts from the current point rather than the earliest binlog.
- It correctly resumes after a restart in the middle of a multi-row `INSERT`/`UPDATE`: the schema is rebuilt from the re-read `TABLE_MAP` event before the resumed rows are decoded.

The change lives in the shared `debezium-connector-binlog` module, so it applies to both the MySQL and MariaDB connectors with no connector-specific production code. The mode is disabled by default.

## Testing
- Unit test for the type mapping (`BinlogMetadataTableBuilderTest`), seeded with real-server ground-truth `TABLE_MAP` values.
- Integration tests (`BinlogMetadataBasedSchemaIT`, run for both MySQL and MariaDB): schema reconstruction without a history topic; initial snapshot → streaming handoff; **mid-multi-row-INSERT restart resume**; fail-fast on non-FULL; `no_data` starting from the current position.
- Manually verified end-to-end against MySQL 8.0 and MariaDB 11.4 with Kafka and Confluent Schema Registry (Avro): all-types reconstruction, DDL evolution via `TABLE_MAP`, GTID stop/resume, and confirmed that no schema history topic is created.

### DIFF
--- a/.github/workflows/connector-mysql-workflow.yml
+++ b/.github/workflows/connector-mysql-workflow.yml
@@ -17,9 +17,9 @@ jobs:
       max-parallel: ${{ inputs.max-parallel }}
       fail-fast: false
       matrix:
-        profile: [ "mysql-ci", "mysql-ci-gtids", "mysql-ci-percona", "mysql-ci-ssl" ]
+        profile: [ "mysql-ci", "mysql-ci-gtids", "mysql-ci-percona", "mysql-ci-ssl", "mysql-ci-binlog-metadata" ]
         version-mysql-server: [ "8.0", "8.4", "9.7" ]
-        # Always exclude mysql-ci-gtids and mysql-ci-ssl profiles for all versions except most recent
+        # Always exclude mysql-ci-gtids, mysql-ci-ssl and mysql-ci-binlog-metadata profiles for all versions except most recent
         exclude:
           - profile: "mysql-ci-gtids"
             version-mysql-server: "8.0"
@@ -28,6 +28,10 @@ jobs:
           - profile: "mysql-ci-ssl"
             version-mysql-server: "8.0"
           - profile: "mysql-ci-ssl"
+            version-mysql-server: "8.4"
+          - profile: "mysql-ci-binlog-metadata"
+            version-mysql-server: "8.0"
+          - profile: "mysql-ci-binlog-metadata"
             version-mysql-server: "8.4"
     name: MySQL ${{ matrix.version-mysql-server }} - ${{ matrix.profile }}
     runs-on: ubuntu-latest
--- a/.github/workflows/connector-mysql-workflow.yml
+++ b/.github/workflows/connector-mysql-workflow.yml
@@ -17,9 +17,9 @@ jobs:
       max-parallel: ${{ inputs.max-parallel }}
       fail-fast: false
       matrix:
-        profile: [ "mysql-ci", "mysql-ci-gtids", "mysql-ci-percona", "mysql-ci-ssl", "mysql-ci-binlog-metadata" ]
+        profile: [ "mysql-ci", "mysql-ci-gtids", "mysql-ci-percona", "mysql-ci-ssl", "mysql-ci,mysql-feature-binlog-metadata" ]
         version-mysql-server: [ "8.0", "8.4", "9.7" ]
-        # Always exclude mysql-ci-gtids, mysql-ci-ssl and mysql-ci-binlog-metadata profiles for all versions except most recent
+        # Always exclude the mysql-ci-gtids, mysql-ci-ssl and binlog-metadata builds for all versions except most recent
         exclude:
           - profile: "mysql-ci-gtids"
             version-mysql-server: "8.0"
@@ -29,9 +29,9 @@ jobs:
             version-mysql-server: "8.0"
           - profile: "mysql-ci-ssl"
             version-mysql-server: "8.4"
-          - profile: "mysql-ci-binlog-metadata"
+          - profile: "mysql-ci,mysql-feature-binlog-metadata"
             version-mysql-server: "8.0"
-          - profile: "mysql-ci-binlog-metadata"
+          - profile: "mysql-ci,mysql-feature-binlog-metadata"
             version-mysql-server: "8.4"
     name: MySQL ${{ matrix.version-mysql-server }} - ${{ matrix.profile }}
     runs-on: ubuntu-latest
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogDatabaseSchema.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogDatabaseSchema.java
@@ -7,9 +7,12 @@ package io.debezium.connector.binlog;
 
 import java.time.Instant;
 import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.BitSet;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
 import java.util.Set;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.regex.Pattern;
@@ -18,8 +21,10 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import com.github.shyiko.mysql.binlog.event.TableMapEventData;
+import com.github.shyiko.mysql.binlog.event.TableMapEventMetadata;
 
 import io.debezium.config.CommonConnectorConfig;
+import io.debezium.connector.binlog.charset.BinlogCharsetRegistry;
 import io.debezium.connector.binlog.metadata.BinlogMetadataTableBuilder;
 import io.debezium.connector.common.CdcSourceTaskContext;
 import io.debezium.relational.CustomConverterRegistry;
@@ -65,8 +70,8 @@ public abstract class BinlogDatabaseSchema<P extends BinlogPartition, O extends 
     private final Map<Long, TableId> tableIdsByTableNumber = new ConcurrentHashMap<>();
     private final Map<Long, TableId> excludeTableIdsByTableNumber = new ConcurrentHashMap<>();
     private final BinlogConnectorConfig connectorConfig;
-    private final BinlogMetadataTableBuilder metadataTableBuilder = new BinlogMetadataTableBuilder();
-    private final Set<TableId> tablesRegisteredFromBinlogMetadata = ConcurrentHashMap.newKeySet();
+    private final BinlogMetadataTableBuilder metadataTableBuilder;
+    private final Map<TableId, TableMapMetadataSnapshot> tablesRegisteredFromBinlogMetadata = new ConcurrentHashMap<>();
 
     /**
      * Creates a binlog-connector based relational schema based on the supplied configuration. The DDL
@@ -105,19 +110,13 @@ public abstract class BinlogDatabaseSchema<P extends BinlogPartition, O extends 
         this.ddlParser = createDdlParser(connectorConfig, valueConverter);
         this.connectorConfig = connectorConfig;
         this.filters = connectorConfig.getTableFilters();
+        this.metadataTableBuilder = new BinlogMetadataTableBuilder(
+                () -> connectorConfig.getServiceRegistry().tryGetService(BinlogCharsetRegistry.class));
     }
 
     @Override
     protected DdlParser getDdlParser() {
         return ddlParser;
-    }
-
-    @Override
-    public boolean isHistorized() {
-        // In binlog-metadata-based schema mode the schema is rebuilt from the log itself (TABLE_MAP events),
-        // so the connector behaves like a non-historized connector (e.g. PostgreSQL): no schema snapshot,
-        // no history recovery and no persisted schema history topic.
-        return !connectorConfig.isBinlogMetadataBasedSchema();
     }
 
     @Override
@@ -234,16 +233,93 @@ public abstract class BinlogDatabaseSchema<P extends BinlogPartition, O extends 
         // A TABLE_MAP event precedes the row events for its table within every transaction, so it is seen
         // very frequently. Rebuilding the table and its Kafka Connect schema on every event would be
         // wasteful (the PostgreSQL connector rebuilds unconditionally, but its pgoutput relation messages
-        // are rare, whereas TABLE_MAP is per-transaction). Instead, each table is populated once and only
-        // rebuilt after a DDL statement for it has been parsed from the binlog, which marks it dirty by
-        // removing it from this registry (see the invalidation in parseDdl).
-        if (tablesRegisteredFromBinlogMetadata.contains(tableId) && schemaFor(tableId) != null) {
+        // are rare, whereas TABLE_MAP is per-transaction). Instead, each table is populated once and
+        // rebuilt when either:
+        // - a DDL statement for it has been parsed from the binlog, which marks it dirty by removing it
+        // from this registry (see the invalidation in parseDdl); or
+        // - the TABLE_MAP metadata no longer structurally matches what was registered, which catches
+        // schema drift that never appears as a binlog DDL statement (for example an ALTER executed
+        // with sql_log_bin=OFF); the comparison is a cheap field-by-field check without allocations.
+        final TableMapMetadataSnapshot registered = tablesRegisteredFromBinlogMetadata.get(tableId);
+        if (registered != null && registered.matches(metadata) && schemaFor(tableId) != null) {
             return true;
         }
         final Table table = metadataTableBuilder.build(tableId, metadata);
         refresh(table);
-        tablesRegisteredFromBinlogMetadata.add(tableId);
+        tablesRegisteredFromBinlogMetadata.put(tableId, TableMapMetadataSnapshot.of(metadata));
         return true;
+    }
+
+    /**
+     * Snapshot of the {@code TABLE_MAP} fields the {@link BinlogMetadataTableBuilder} consumes, used to
+     * detect whether a newly received event still structurally matches the registered schema.
+     */
+    private static final class TableMapMetadataSnapshot {
+
+        private final byte[] columnTypes;
+        private final int[] columnMetadata;
+        private final BitSet columnNullability;
+        private final List<String> columnNames;
+        private final BitSet signedness;
+        private final List<Integer> columnCharsets;
+        private final Integer defaultCharsetCollation;
+        private final Map<Integer, Integer> defaultCharsetOverrides;
+        private final List<String[]> enumStrValues;
+        private final List<String[]> setStrValues;
+        private final List<Integer> simplePrimaryKeys;
+
+        private TableMapMetadataSnapshot(TableMapEventData data, TableMapEventMetadata meta) {
+            this.columnTypes = data.getColumnTypes() != null ? data.getColumnTypes().clone() : null;
+            this.columnMetadata = data.getColumnMetadata() != null ? data.getColumnMetadata().clone() : null;
+            this.columnNullability = data.getColumnNullability() != null ? (BitSet) data.getColumnNullability().clone() : null;
+            this.columnNames = meta.getColumnNames();
+            this.signedness = meta.getSignedness() != null ? (BitSet) meta.getSignedness().clone() : null;
+            this.columnCharsets = meta.getColumnCharsets();
+            final TableMapEventMetadata.DefaultCharset defaultCharset = meta.getDefaultCharset();
+            this.defaultCharsetCollation = defaultCharset != null ? defaultCharset.getDefaultCharsetCollation() : null;
+            this.defaultCharsetOverrides = defaultCharset != null ? defaultCharset.getCharsetCollations() : null;
+            this.enumStrValues = meta.getEnumStrValues();
+            this.setStrValues = meta.getSetStrValues();
+            this.simplePrimaryKeys = meta.getSimplePrimaryKeys();
+        }
+
+        static TableMapMetadataSnapshot of(TableMapEventData data) {
+            return new TableMapMetadataSnapshot(data, data.getEventMetadata());
+        }
+
+        boolean matches(TableMapEventData data) {
+            final TableMapEventMetadata meta = data.getEventMetadata();
+            if (meta == null) {
+                return false;
+            }
+            final TableMapEventMetadata.DefaultCharset defaultCharset = meta.getDefaultCharset();
+            return Arrays.equals(columnTypes, data.getColumnTypes())
+                    && Arrays.equals(columnMetadata, data.getColumnMetadata())
+                    && Objects.equals(columnNullability, data.getColumnNullability())
+                    && Objects.equals(columnNames, meta.getColumnNames())
+                    && Objects.equals(signedness, meta.getSignedness())
+                    && Objects.equals(columnCharsets, meta.getColumnCharsets())
+                    && Objects.equals(defaultCharsetCollation, defaultCharset != null ? defaultCharset.getDefaultCharsetCollation() : null)
+                    && Objects.equals(defaultCharsetOverrides, defaultCharset != null ? defaultCharset.getCharsetCollations() : null)
+                    && stringArrayListsEqual(enumStrValues, meta.getEnumStrValues())
+                    && stringArrayListsEqual(setStrValues, meta.getSetStrValues())
+                    && Objects.equals(simplePrimaryKeys, meta.getSimplePrimaryKeys());
+        }
+
+        private static boolean stringArrayListsEqual(List<String[]> left, List<String[]> right) {
+            if (left == null || right == null) {
+                return left == right;
+            }
+            if (left.size() != right.size()) {
+                return false;
+            }
+            for (int i = 0; i < left.size(); i++) {
+                if (!Arrays.equals(left.get(i), right.get(i))) {
+                    return false;
+                }
+            }
+            return true;
+        }
     }
 
     /**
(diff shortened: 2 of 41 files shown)


debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogDatabaseSchema.java
```diff
@@ -201,6 +216,77 @@ public boolean assignTableNumber(long tableNumber, TableId id) {
...
+            return false;
+        }
+        // A TABLE_MAP event precedes the row events for its table within every transaction, so it is seen
+        // very frequently. Rebuilding the table and its Kafka Connect schema on every event would be
+        // wasteful (the PostgreSQL connector rebuilds unconditionally, but its pgoutput relation messages
+        // are rare, whereas TABLE_MAP is per-transaction). Skip the rebuild when the metadata is unchanged
+        // from what was last registered for this table. The signature is a deterministic serialization of
+        // every field the builder reads (not a hash), so an unnoticed change cannot leave a stale schema;
+        // in the worst case a signature bug only causes an unnecessary rebuild.
+        final String signature = metadataSignature(metadata);
```
jpechane: @minleejae Is this necessary? I suppose that even in this mode we are still received DDLs. So if we see DDL for a certain table we can mark it as dirty or remove from the list of tables so it is populated again?

jpechane:
Also I wonder if it is possible to modify the testsuite in a way that it will either use schema history or not.
That way we can (at least once) make sure there is no compatibility issue.

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogDefaultValueIT.java
```diff
@@ -337,6 +338,7 @@ public void stringTest() throws InterruptedException {
 
     @Test
     @SkipWhenKafkaVersion(check = EqualityCheck.EQUAL, value = KafkaVersion.KAFKA_1XX, description = "Not compatible with Kafka 1.x")
+    @SkipWhenSchemaHistoryDisabled(reason = "Column default values are not carried by the binlog TABLE_MAP metadata")
```
jpechane: @minleejae These are initeresting points? Does it mean that historized and non-historized mode are not equivavent?
In that case we need to clearly document pros and cons using both modes.

minleejae: You're right that they're not fully equivalent. I've restructured the documentation into a dedicated comparison section under the schema history topic — what is identical, the advantages, the limitations
  (each with its downstream impact and workaround), and guidance on choosing a mode. https://github.com/debezium/debezium/pull/7593/changes/3f658db8eb5c16cf28d2253d62c954b69f2e139a

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogDatabaseSchema.java
minleejae: Parsed DDL removes the table from the registry, so the next TABLE_MAP event rebuilds it, as you suggest.
The comparison itself remained as the skip-rebuild check and to catch changes that never reach the binlog as DDL. 30c5862322

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/history/BinlogMetadataSchemaHistory.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * A no-op {@link SchemaHistory} used by the binlog-metadata-based schema mode (debezium/dbz#978). In
+ * this mode the streaming table schema is reconstructed from the FULL metadata carried by binlog
+ * {@code TABLE_MAP} events, so no persistent schema history storage is needed: records are discarded and
+ * recovery is a no-op, while the snapshot phase runs exactly as with a persistent history. The storage
+ * always reports itself as present so that connector startup never attempts to create or recover it.
+ */
+@ThreadSafe
+public final class BinlogMetadataSchemaHistory extends AbstractSchemaHistory {
```
Naros, line 23: I wonder if, as a follow-up, we should consider moving this up to the common connector framework and shifting the PG connector to use a `DummySchemaHistory` implementation that effectively uses this no-op approach @jpechane ? With the archetype work we're doing, and how to document creating connectors, anywhere we can think about homogenizing the connectors with a strategy pattern would likely be seen as a win.

debezium-connector-mysql/pom.xml
```diff
@@ -1104,6 +1104,87 @@
             </build>
         </profile>
 
+        <!-- Runs the same testsuite as mysql-ci, but with the binlog-metadata-based schema mode enabled,
+             so the suite is executed without a schema history (debezium/dbz#978) -->
+        <profile>
+            <id>mysql-ci-binlog-metadata</id>
```
Naros: I'm concerned that this is effectively `mysql-ci` with a property toggle, but we take a high cost in XML duplication here to support it. Could we reimagine this profile as a feature toggle rather than a full CI profile?

For example, the main pom properties section defines this
```
<properties>
  <use.binlog.metadata.based.schema>false</use.binlog.metadata.based.schema>
</properties>
```

Then we modify all the `mysql-ci*` profiles to include this property; by default, it is `false`. 

Then we introduce a `mysql-feature-binlog-metadata` profile or similar that toggles this this property to be `true`, so the expectation is to run the Binlog Metadata CI, we'd pass `mysql-ci` plus `mysql-feature-binlog-metadata` together.

This composition just minimizes technical debt imo, and we should consider this for MariaDB too.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogDatabaseSchema.java
```diff
@@ -201,6 +213,165 @@ public boolean assignTableNumber(long tableNumber, TableId id) {
...
+                return;
+            }
+            final TableId tableId = getTableId(event);
+            if (tableId == null) {
+                tablesRegisteredFromBinlogMetadata.clear();
+                return;
+            }
+            tablesRegisteredFromBinlogMetadata.remove(tableId);
+            if (event instanceof DdlParserListener.TableAlteredEvent) {
+                final TableId previousTableId = ((DdlParserListener.TableAlteredEvent) event).previousTableId();
```
Naros: Could we use `instanceof` pattern matching here?

debezium-connector-mariadb/pom.xml
```diff
@@ -956,6 +956,85 @@
             </build>
         </profile>
 
+        <!-- Runs the same testsuite as mariadb-ci, but with the binlog-metadata-based schema mode enabled,
+             so the suite is executed without a schema history (debezium/dbz#978) -->
+        <profile>
+            <id>mariadb-ci-binlog-metadata</id>
```
Naros: Same point here I made about the profiles in MySQL.

Naros (commented):
Hi @minleejae, this is coming along nicely. I've left a few comments below.

debezium-connector-mysql/pom.xml
minleejae: Restructured as suggested in [b2374ce](https://github.com/debezium/debezium/pull/7593/commits/b2374ceff4bea2d1d9b6848b366565173f67e905): `use.binlog.metadata.based.schema` now defaults to false in the pom properties, and the duplicated profile is replaced by a slim `mysql-feature-binlog-metadata` profile that only flips the property to true, so the binlog-metadata CI now runs -P mysql-ci,mysql-feature-binlog-metadata.

debezium-connector-mariadb/pom.xml
minleejae: Addressed in [b2374ce](https://github.com/debezium/debezium/pull/7593/commits/b2374ceff4bea2d1d9b6848b366565173f67e905)

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogDatabaseSchema.java
minleejae: Addressed in [69489d4](https://github.com/debezium/debezium/pull/7593/commits/69489d48349227b4bab25d70f5f262fb3decfc54)

Naros (approved):
LGTM, I'm going to leave this waiting for one more pass from @jpechane. I'd like his input in case I missed anything.

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogMetadataBasedSchemaIT.java
```diff
@@ -0,0 +1,570 @@
...
+                    rs -> rs.next() ? rs.getString(2) : null);
+        }
+    }
+
+    private void setBinlogRowMetadata(String value) throws SQLException {
+        try (BinlogTestConnection db = getTestDatabaseConnection("mysql")) {
+            db.execute("SET GLOBAL binlog_row_metadata = '" + value + "'");
+        }
+    }
+}
```
rk3rn3r, line 570: Is there any test to check for a schema evolution?

documentation/modules/ROOT/partials/modules/all-connectors/shared-mariadb-mysql.adoc
```diff
@@ -125,6 +133,39 @@ If consumers do not need the records the connector generates for helper tables,
...
 * xref:{context}-topic-names[Default names for topics] that receive {prodname} event records.
 
+[id="{context}-comparison-schema-history-binlog-metadata"]
+==== Comparison between the schema history and the binlog metadata mode
+
+The two modes produce identical change events for the same row changes.
+Every change that the connector reads from the binlog carries the complete row together with the schema information that is needed to decode it, so change event values and keys, snapshots, and restart behavior do not differ between the modes.
+The modes are not fully equivalent, however: they differ in where the table schema comes from, and in the schema attributes that are available from that source.
+
+.Advantages of the binlog metadata mode
```
rk3rn3r, line 155: This mixes advantages of the binlog metadata mode with the disadvantages of the schema history mode. Furthermore it would be good to have separate advantages and disadvantage sections for both modes. Can you do this?

documentation/modules/ROOT/partials/modules/all-connectors/shared-mariadb-mysql.adoc
```diff
@@ -125,6 +133,39 @@ If consumers do not need the records the connector generates for helper tables,
...
+* The connector restarts without a history recovery phase.
+* Schema changes that are never written to the binlog (for example, an `ALTER TABLE` executed with `sql_log_bin=OFF`) heal themselves as soon as the connector reads the next change for the table, whereas the schema history mode keeps using the stale schema.
+* The connector offset can be seeded or rewound to any position that the binlog still retains, even across past DDL statements, because the binlog carries the schema that was in effect at every position.
+With a schema history, pointing a new connector at an earlier binlog position stops the connector when a schema change lies between that position and the current table structure, because the schema history topic does not describe the schema at that position and decoding the older rows with the current schema fails.
+The xref:{context}-connector-snapshot-mode-options[`recovery` snapshot mode] cannot repair this, because it rebuilds the history from the current schema; the binlog metadata mode does not have this restriction.
+
+.Limitations of the binlog metadata mode
```
rk3rn3r, line 161: Can we have a paragraph for schema history disadvantages (see above)?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/listener/AlterTableParserListener.java
```diff
@@ -51,7 +51,17 @@ public AlterTableParserListener(MySqlAntlrDdlParser parser, List<ParseTreeListen
...
+                // The table is captured but not yet part of the in-memory model, e.g. when the schema
+                // is reconstructed from binlog TABLE_MAP metadata and no row event has been seen for
+                // the table since the restart. Signal the event without applying the statement, so
+                // that the schema change is still emitted; the schema itself is restored by the next
+                // TABLE_MAP event.
+                parser.signalAlterTable(tableId, null, ctx);
+            }
+            else {
+                LOG.debug("Ignoring ALTER TABLE statement for non-captured table {}", tableId);
+            }
```
rk3rn3r: Whats the difference between this implementation and the legacy one? The code looks identical.

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/antlr/listener/AlterTableParserListener.java
```diff
@@ -48,7 +48,17 @@ public AlterTableParserListener(MariaDbAntlrDdlParser parser, List<ParseTreeList
...
+                // The table is captured but not yet part of the in-memory model, e.g. when the schema
+                // is reconstructed from binlog TABLE_MAP metadata and no row event has been seen for
+                // the table since the restart. Signal the event without applying the statement, so
+                // that the schema change is still emitted; the schema itself is restored by the next
+                // TABLE_MAP event.
+                parser.signalAlterTable(tableId, null, ctx);
+            }
+            else {
+                LOG.debug("Ignoring ALTER TABLE statement for non-captured table {}", tableId);
+            }
```
rk3rn3r: same here? Could this logic be unified and shared when it is identical?

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/junit/SkipWhenSchemaHistoryDisabled.java
```diff
@@ -0,0 +1,31 @@
...
+ * Marker annotation for tests that inherently depend on the schema history (its storage, recovery, or the
+ * schema change events emitted during a schema snapshot). Such tests are skipped when the testsuite runs
+ * with the binlog-metadata-based schema mode enabled (see
+ * {@link io.debezium.connector.binlog.util.UniqueDatabase#BINLOG_METADATA_BASED_SCHEMA_PROPERTY}), where no
+ * schema history is used.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target({ ElementType.METHOD, ElementType.TYPE })
+@ExtendWith(SkipTestWhenSchemaHistoryDisabledExtension.class)
+public @interface SkipWhenSchemaHistoryDisabled {
```
rk3rn3r, line 25: Instead of running all tests (even many unrelated tests) that are not annotated to be skipped with parallel CI profiles, could we instead have 2 test suites, one that covers the schema history related tests and one that covers the binlog metadata tests and run those in one CI run like: run metadata test suite related tests, run schema history specific tests, run all other tests, instead of running all tests twice and sometimes skip the one or the other binlog metadata / schema history specific tests?

.github/workflows/connector-mysql-workflow.yml
```diff
@@ -17,9 +17,9 @@ jobs:
       max-parallel: ${{ inputs.max-parallel }}
       fail-fast: false
       matrix:
-        profile: [ "mysql-ci", "mysql-ci-gtids", "mysql-ci-percona", "mysql-ci-ssl" ]
+        profile: [ "mysql-ci", "mysql-ci-gtids", "mysql-ci-percona", "mysql-ci-ssl", "mysql-ci,mysql-feature-binlog-metadata" ]
```
rk3rn3r, line 20: @jpechane @Naros Looking at how the tests are executed with another parallel matix run utilizing GH runners (which are already a problem), which also run unrelated tests cases again twice, I suggested a change below (on the annotation):

Instead of running all tests (even many unrelated tests) that are not annotated to be skipped with parallel CI profiles, could we instead have 2 test suites, one that covers the schema history related tests and one that covers the binlog metadata tests and run those in one CI run like: run metadata test suite related tests, run schema history specific tests, run all other tests, instead of running all tests twice and sometimes skip the one or the other binlog metadata / schema history specific tests?

wdyt?

rk3rn3r:
While the code looks good, I have some problems with adding another 2 new test runs to the matrix, blocking GH runners. Especially considering that the schema history code cannot be removed until we drop support for older database versions that don't emit binlog metadata.

.github/workflows/connector-mysql-workflow.yml
Naros: @rk3rn3r,  Jiri got a runner cap increase, so this shouldn't generally be an issue with 2 new runs. I don't think we should be concerned with this atm.

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogMetadataBasedSchemaIT.java
minleejae: Yes. `shouldRefreshSchemaWhenTableIsAltered()` covers DML -> `ALTER TABLE` -> DML and verifies that the added column appears in the schema rebuilt from the next `TABLE_MAP` event ([59390f9](https://github.com/debezium/debezium/pull/7593/commits/59390f9d4eae1dc6f7373ccc16d91753b399815e)). `shouldStreamFromRewoundOffsetAcrossDdlChange()` also verifies the point-in-time schemas on both sides of a DDL change during replay ([49ebb71](https://github.com/debezium/debezium/pull/7593/commits/49ebb7134d61f845890457dd88b2bba3bc23cf0c)).

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/junit/SkipWhenSchemaHistoryDisabled.java
minleejae: The full second run is intentional because it exercises the existing connector tests unchanged with the metadata-based mode enabled. A dedicated metadata suite would only cover scenarios that we explicitly classified and could miss broader compatibility regressions. [b2374ce](https://github.com/debezium/debezium/pull/7593/commits/b2374ceff4bea2d1d9b6848b366565173f67e905) consolidates the duplicated connector profiles into a feature toggle while retaining that broad compatibility run. Given the runner capacity increase mentioned by Chris, I would prefer to keep this coverage for now.

.github/workflows/connector-mysql-workflow.yml
minleejae: Thanks. Since Chris confirmed the runner capacity increase, I kept the additional matrix entry so that the existing suite runs with both schema history and the metadata-based mode. The duplicated connector profiles were consolidated into a shared feature toggle in [b2374ce](https://github.com/debezium/debezium/pull/7593/commits/b2374ceff4bea2d1d9b6848b366565173f67e905).

minleejae:
@rk3rn3r, the additional run is intentional for broader compatibility coverage. Since Chris confirmed the runner capacity increase, I would prefer to keep it for now.

documentation/modules/ROOT/partials/modules/all-connectors/shared-mariadb-mysql.adoc
minleejae: Agreed. I split the comparison into separate advantages and limitations for schema history and binlog metadata: [3c48c70](https://github.com/debezium/debezium/commit/3c48c704279fa35a4c4271c9bd18e056b80dbe60).

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/listener/AlterTableParserListener.java
minleejae: There is no behavioral difference here. I moved the shared handling into `AntlrDdlParser`, and both MySQL parsers now reuse it: [657fa19](https://github.com/debezium/debezium/commit/657fa199587b9af9a60e9423bcc531898daacbf7).

documentation/modules/ROOT/partials/modules/all-connectors/shared-mariadb-mysql.adoc
minleejae: Added the schema history limitations, including history recovery, rewound-offset coverage, and unlogged DDL: [3c48c70](https://github.com/debezium/debezium/commit/3c48c704279fa35a4c4271c9bd18e056b80dbe60).

debezium-connector-mariadb/src/main/java/io/debezium/connector/mariadb/antlr/listener/AlterTableParserListener.java
minleejae: Agreed. MariaDB now reuses the same `AntlrDdlParser` helper as the MySQL parsers: [657fa19](https://github.com/debezium/debezium/commit/657fa199587b9af9a60e9423bcc531898daacbf7).

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/metadata/BinlogMetadataTableBuilder.java
```diff
@@ -0,0 +1,498 @@
...
+                    pkNames.add(names.get(idx));
+                }
+                table.setPrimaryKeyNames(pkNames);
+            }
+        }
+
+        return table.create();
+    }
+
+    private void intType(ColumnEditor column, String baseName, int jdbcType, boolean unsigned) {
```
Naros:
```suggestion
    private void numericType(ColumnEditor column, String baseName, int jdbcType, boolean unsigned) {
```
Since this is used for a combination of whole-number integers and floating-point single/double precision, using `numericType` seems more appropriate.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/metadata/BinlogMetadataTableBuilder.java
```diff
@@ -0,0 +1,498 @@
...
+                case TYPE_LONG -> intType(column, "INT", Types.INTEGER, unsigned);
+                case TYPE_LONGLONG -> intType(column, "BIGINT", Types.BIGINT, unsigned);
+                case TYPE_FLOAT -> intType(column, "FLOAT", Types.FLOAT, unsigned);
+                case TYPE_DOUBLE -> intType(column, "DOUBLE", Types.DOUBLE, unsigned);
+                case TYPE_YEAR -> column.type("YEAR").jdbcType(Types.INTEGER);
+                case TYPE_DECIMAL, TYPE_NEWDECIMAL -> {
+                    column.type(unsigned ? "DECIMAL UNSIGNED" : "DECIMAL").jdbcType(Types.DECIMAL);
+                    column.length(columnMeta[i] & 0xFF);
+                    column.scale((columnMeta[i] >> 8) & 0xFF);
+                }
```
Naros, line 175: Could we add a comment here, similar to BIT about the encoding?

Naros (commented):
Hi @minleejae this looks quite good, I've left a few comments.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/metadata/BinlogMetadataTableBuilder.java
minleejae: Renamed intType to numericType since it handles both integer and floating-point numeric types.  [d667efd](https://github.com/debezium/debezium/pull/7593/commits/d667efdfb9f186510adbe2022f249388be2652bd)

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/metadata/BinlogMetadataTableBuilder.java
minleejae: Added a comment documenting that DECIMAL metadata is encoded as (scale << 8) | precision. Thanks! [d667efd](https://github.com/debezium/debezium/pull/7593/commits/d667efdfb9f186510adbe2022f249388be2652bd)

minleejae:
Hi @Naros @jpechane, I fixed the metadata-mode CI failure in `shouldHandleNegativeTinyIntOneAsTrue()` by setting `boolean.length.checker=false` together with its exact selector. TABLE_MAP metadata does not carry the `TINYINT(1)` display width, so the selector—not the unavailable width—identifies the Boolean column in this mode. I also cross-checked this with #7748: both paths represent `BOOL`/`BOOLEAN` as `TINYINT`/`INT8` by default, while Boolean conversion remains an explicit opt-in through `TinyIntOneToBooleanConverter`.

PR #7599, opened 2026-07-03
debezium/dbz#2179 JDBC sink: add none-validated schema evolution mode

minleejae:
Closes debezium/dbz#2179.

## Summary

This adds a `none-validated` schema evolution mode for the JDBC sink connector.

When `schema.evolution=none-validated`, the connector validates statically resolvable target tables during task startup and fails fast if a required table does not exist. After startup, this mode behaves like `schema.evolution=none`: the connector still does not create or alter target tables while processing records.

## Details

The startup validation is intentionally limited to collection names that can be resolved before consuming records:

- literal `collection.name.format` values
- default `${topic}`-based names when `topics` is statically configured

The connector rejects startup validation for dynamic cases such as `topics.regex`, source-field placeholders, or custom collection naming strategies, because those target tables cannot be determined safely at task startup.

## Testing

```bash
./mvnw -q -pl debezium-connector-jdbc -am -DskipTests -DskipITs -DskipLongRunningTests test-compile
./mvnw -q -pl debezium-connector-jdbc -am -DskipITs -DskipIT -DskipLongRunningTests -DskipTests=false -Dsurefire.failIfNoSpecifiedTests=false -Dtest=StartupCollectionValidatorTest,JdbcSinkConnectorConfigTest test
./mvnw -q -pl debezium-connector-jdbc -DskipTests -DskipITs -DskipLongRunningTests checkstyle:check
git diff --check
./mvnw -q -pl debezium-connector-jdbc -DskipITs=false -DskipLongRunningTests -Dtest.tags=it-mysql -Dfailsafe.failIfNoSpecifiedTests=false '-Dit.test=io.debezium.connector.jdbc.integration.mysql.JdbcSinkSchemaEvolutionIT#testStartShould*NoneValidated*' failsafe:integration-test failsafe:verify
```

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/CollectionsExistsValidator.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/CollectionsExistsValidator.java
@@ -148,6 +148,8 @@ final class CollectionsExistsValidator {
     @SuppressWarnings("deprecation")
     private DefaultCollectionNamingStrategy getDefaultCollectionNamingStrategy() {
         CollectionNamingStrategy strategy = config.getCollectionNamingStrategy();
+        // The connector configuration wraps even the default naming strategy in this proxy.
+        // Unwrap it so that validation can recognize and reuse the underlying default strategy.
         if (strategy instanceof TemporaryBackwardCompatibleCollectionNamingStrategyProxy proxy) {
             strategy = proxy.getOriginalStrategy();
         }
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/CollectionsExistsValidator.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/CollectionsExistsValidator.java
@@ -54,7 +54,7 @@ final class CollectionsExistsValidator {
             return;
         }
         if (collectionNames.isEmpty()) {
-            throw new DebeziumException("'" + JdbcSinkConnectorConfig.SCHEMA_EVOLUTION + "=" + SchemaEvolutionMode.NONE_VALIDATED.getValue()
+            throw new DebeziumException("'" + JdbcSinkConnectorConfig.SCHEMA_EVOLUTION + "=" + SchemaEvolutionMode.VALIDATE_ONLY.getValue()
                     + "' requires '" + SinkTask.TOPICS_CONFIG + "' or '" + SinkTask.TOPICS_REGEX_CONFIG + "' when '"
                     + JdbcSinkConnectorConfig.COLLECTION_NAME_FORMAT + "' contains '${topic}'.");
         }
@@ -84,7 +84,7 @@ final class CollectionsExistsValidator {
                                     + "Create the target table before the connector processes records or use '%s=%s'.",
                             collectionId.toFullIdentiferString(),
                             JdbcSinkConnectorConfig.SCHEMA_EVOLUTION,
-                            SchemaEvolutionMode.NONE_VALIDATED.getValue(),
+                            SchemaEvolutionMode.VALIDATE_ONLY.getValue(),
                             JdbcSinkConnectorConfig.SCHEMA_EVOLUTION,
                             SchemaEvolutionMode.BASIC.getValue()));
                 }
@@ -108,18 +108,18 @@ final class CollectionsExistsValidator {
     Set<String> resolveCollectionNames(Collection<String> topics) {
         final String collectionNameFormat = config.getCollectionNameFormat();
         if (Strings.isNullOrEmpty(collectionNameFormat)) {
-            throw new DebeziumException("'" + JdbcSinkConnectorConfig.SCHEMA_EVOLUTION + "=" + SchemaEvolutionMode.NONE_VALIDATED.getValue()
+            throw new DebeziumException("'" + JdbcSinkConnectorConfig.SCHEMA_EVOLUTION + "=" + SchemaEvolutionMode.VALIDATE_ONLY.getValue()
                     + "' requires '" + JdbcSinkConnectorConfig.COLLECTION_NAME_FORMAT + "' to be configured.");
         }
 
         final DefaultCollectionNamingStrategy namingStrategy = getDefaultCollectionNamingStrategy();
         if (namingStrategy == null) {
-            throw new DebeziumException("'" + JdbcSinkConnectorConfig.SCHEMA_EVOLUTION + "=" + SchemaEvolutionMode.NONE_VALIDATED.getValue()
+            throw new DebeziumException("'" + JdbcSinkConnectorConfig.SCHEMA_EVOLUTION + "=" + SchemaEvolutionMode.VALIDATE_ONLY.getValue()
                     + "' supports table existence validation only with the default collection naming strategy.");
         }
 
         if (collectionNameFormat.contains(SOURCE_PLACEHOLDER_PREFIX)) {
-            throw new DebeziumException("'" + JdbcSinkConnectorConfig.SCHEMA_EVOLUTION + "=" + SchemaEvolutionMode.NONE_VALIDATED.getValue()
+            throw new DebeziumException("'" + JdbcSinkConnectorConfig.SCHEMA_EVOLUTION + "=" + SchemaEvolutionMode.VALIDATE_ONLY.getValue()
                     + "' cannot validate target tables when '" + JdbcSinkConnectorConfig.COLLECTION_NAME_FORMAT
                     + "' contains source field placeholders.");
         }
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/CollectionsExistsValidator.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/CollectionsExistsValidator.java
@@ -7,7 +7,6 @@ package io.debezium.connector.jdbc;
 
 import java.util.LinkedHashSet;
 import java.util.List;
-import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -33,16 +32,19 @@ final class CollectionsExistsValidator {
     private final JdbcSinkConnectorConfig config;
     private final SessionFactory sessionFactory;
     private final DatabaseDialect dialect;
-    private final Map<String, String> props;
+    private final Set<String> staticallyConfiguredTopics;
+    private final boolean topicsRegexConfigured;
 
     CollectionsExistsValidator(JdbcSinkConnectorConfig config,
                                SessionFactory sessionFactory,
                                DatabaseDialect dialect,
-                               Map<String, String> props) {
+                               String topics,
+                               boolean topicsRegexConfigured) {
         this.config = config;
         this.sessionFactory = sessionFactory;
         this.dialect = dialect;
-        this.props = props;
+        this.staticallyConfiguredTopics = parseTopics(topics);
+        this.topicsRegexConfigured = topicsRegexConfigured;
     }
 
     void validate() {
@@ -103,19 +105,18 @@ final class CollectionsExistsValidator {
             return Set.of(collectionNameFormat);
         }
 
-        if (!Strings.isNullOrEmpty(props.get(SinkTask.TOPICS_REGEX_CONFIG))) {
+        if (topicsRegexConfigured) {
             throw new DebeziumException("'" + JdbcSinkConnectorConfig.SCHEMA_EVOLUTION + "=" + SchemaEvolutionMode.NONE_VALIDATED.getValue()
                     + "' cannot validate target tables at startup when '" + SinkTask.TOPICS_REGEX_CONFIG + "' is used.");
         }
 
-        final Set<String> topics = parseTopics(props.get(SinkTask.TOPICS_CONFIG));
-        if (topics.isEmpty()) {
+        if (staticallyConfiguredTopics.isEmpty()) {
             throw new DebeziumException("'" + JdbcSinkConnectorConfig.SCHEMA_EVOLUTION + "=" + SchemaEvolutionMode.NONE_VALIDATED.getValue()
                     + "' requires statically configured '" + SinkTask.TOPICS_CONFIG + "' when '" + JdbcSinkConnectorConfig.COLLECTION_NAME_FORMAT
                     + "' contains '${topic}'.");
         }
 
-        return topics
+        return staticallyConfiguredTopics
                 .stream()
                 .map(topic -> collectionNameFormat.replace(TOPIC_PLACEHOLDER, topic.replace(".", "_")))
                 .collect(Collectors.toCollection(LinkedHashSet::new));
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/CollectionsExistsValidator.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/CollectionsExistsValidator.java
@@ -90,7 +90,8 @@ final class CollectionsExistsValidator {
                     + "' requires '" + JdbcSinkConnectorConfig.COLLECTION_NAME_FORMAT + "' to be configured.");
         }
 
-        if (!usesDefaultCollectionNamingStrategy()) {
+        final DefaultCollectionNamingStrategy namingStrategy = getDefaultCollectionNamingStrategy();
+        if (namingStrategy == null) {
             throw new DebeziumException("'" + JdbcSinkConnectorConfig.SCHEMA_EVOLUTION + "=" + SchemaEvolutionMode.NONE_VALIDATED.getValue()
                     + "' supports startup table validation only with the default collection naming strategy.");
         }
@@ -118,7 +119,7 @@ final class CollectionsExistsValidator {
 
         return staticallyConfiguredTopics
                 .stream()
-                .map(topic -> collectionNameFormat.replace(TOPIC_PLACEHOLDER, topic.replace(".", "_")))
+                .map(topic -> namingStrategy.resolveCollectionName(topic, collectionNameFormat))
                 .collect(Collectors.toCollection(LinkedHashSet::new));
     }
 
@@ -134,11 +135,11 @@ final class CollectionsExistsValidator {
     }
 
     @SuppressWarnings("deprecation")
-    private boolean usesDefaultCollectionNamingStrategy() {
+    private DefaultCollectionNamingStrategy getDefaultCollectionNamingStrategy() {
         CollectionNamingStrategy strategy = config.getCollectionNamingStrategy();
         if (strategy instanceof TemporaryBackwardCompatibleCollectionNamingStrategyProxy proxy) {
             strategy = proxy.getOriginalStrategy();
         }
-        return strategy instanceof DefaultCollectionNamingStrategy;
+        return strategy instanceof DefaultCollectionNamingStrategy defaultStrategy ? defaultStrategy : null;
     }
 }
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/CollectionsExistsValidator.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/CollectionsExistsValidator.java
@@ -5,6 +5,7 @@
  */
 package io.debezium.connector.jdbc;
 
+import java.util.Collection;
 import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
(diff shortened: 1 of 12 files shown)


debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -390,6 +395,10 @@ public String getValue() {
             return mode;
         }
 
+        public boolean isSchemaEvolutionEnabled() {
+            return this == BASIC;
```
mfvitale: Can we use the opposite logic so that further schema evolution mode are correctly checked?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/StartupCollectionValidator.java
```diff
@@ -0,0 +1,131 @@
...
+        if (config.getSchemaEvolutionMode() != SchemaEvolutionMode.NONE_VALIDATED) {
+            return;
+        }
+
+        final Set<CollectionId> collectionIds = resolveCollectionIds(config, dialect, props);
+        try (StatelessSession validationSession = sessionFactory.openStatelessSession()) {
+            for (CollectionId collectionId : collectionIds) {
+                final boolean exists = validationSession.doReturningWork(connection -> dialect.tableExists(connection, collectionId));
+                if (!exists) {
+                    throw new ConnectException(String.format(
```
mfvitale: Please throw a `DebeziumException`

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/StartupCollectionValidator.java
```diff
@@ -0,0 +1,131 @@
...
+        }
+    }
+
+    static Set<CollectionId> resolveCollectionIds(JdbcSinkConnectorConfig config, DatabaseDialect dialect, Map<String, String> props) {
+        return resolveCollectionNames(config, props)
+                .stream()
+                .map(collectionName -> {
+                    final CollectionId collectionId = dialect.getCollectionId(collectionName);
+                    if (collectionId == null) {
+                        throw new ConnectException("Unable to resolve target table name '" + collectionName + "' for startup validation.");
```
mfvitale: Same as above

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/StartupCollectionValidator.java
```diff
@@ -0,0 +1,131 @@
...
+                    }
+                    return collectionId;
+                })
+                .collect(Collectors.toCollection(LinkedHashSet::new));
+    }
+
+    static Set<String> resolveCollectionNames(JdbcSinkConnectorConfig config, Map<String, String> props) {
+        final String collectionNameFormat = config.getCollectionNameFormat();
+        if (Strings.isNullOrEmpty(collectionNameFormat)) {
+            throw new ConnectException("'" + JdbcSinkConnectorConfig.SCHEMA_EVOLUTION + "=" + SchemaEvolutionMode.NONE_VALIDATED.getValue()
```
mfvitale: Same as above

jpechane:
Please add sign-off

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/StartupCollectionValidator.java
```diff
@@ -0,0 +1,131 @@
...
+
+import io.debezium.connector.jdbc.JdbcSinkConnectorConfig.SchemaEvolutionMode;
+import io.debezium.connector.jdbc.dialect.DatabaseDialect;
+import io.debezium.connector.jdbc.naming.TemporaryBackwardCompatibleCollectionNamingStrategyProxy;
+import io.debezium.metadata.CollectionId;
+import io.debezium.sink.naming.CollectionNamingStrategy;
+import io.debezium.sink.naming.DefaultCollectionNamingStrategy;
+import io.debezium.util.Strings;
+
+final class StartupCollectionValidator {
```
jpechane: @minleejae I don't see any reason why this class should be just a static/singleton.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -390,6 +395,10 @@ public String getValue() {
             return mode;
         }
 
+        public boolean isSchemaEvolutionEnabled() {
```
jpechane, line 408: There should be two new methods

`boolean validateOnStartup()` and `boolean schemaEvolution()`

The enum onstructor should have two booleans setting flags for these two methods and the methods will be used in the code later instead of enamu comparison.

jpechane (changes requested):

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/StartupCollectionValidator.java
minleejae: Addressed in [199af19](https://github.com/debezium/debezium/pull/7599/commits/199af195929d0aa53abf58b63aa7e03cc585d371).

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/StartupCollectionValidator.java
minleejae: Addressed in [bd76809](https://github.com/debezium/debezium/pull/7599/commits/bd76809c760248cf8012b17b51bf47534ab10f94)

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/StartupCollectionValidator.java
minleejae: Addressed in [7325d11](https://github.com/debezium/debezium/pull/7599/commits/7325d118a744619ed7996f2ff8c3dd7ae2e753b1).

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
minleejae: Good point, thanks. SchemaEvolutionMode now has validatorOnStartup() and schemaEvolution(), backed by explicit boolean flags in the enum constructor. 5b48482

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/StartupCollectionValidator.java
minleejae: I changed `StartupCollectionValidator` from a static utility-style class to an instance-based validator. fd5f5ae.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -255,6 +255,9 @@ You can use the following schema evolution modes with the {prodname} JDBC sink c
 |`none`
 |The connector does not perform any DDL schema evolution.
 
+|`none-validated`
+|The connector does not perform any DDL schema evolution, but validates statically resolvable destination tables during startup.
```
Naros: I believe we need to improve the documentation here.

When I, and I suspect others, see `validated`, the first thing that I expect is that the connector makes sure that on startup, the table exists and has the right structure for my topic events. But we know that the latter is not possible without ingestion, and the former has limitations, as seen in the code.

The documentation should be more precise about what is being validated, what scenarios can and cannot be validated, etc. Otherwise, I'm afraid this leaves the docs quite vague on what this truly does or gives the wrong impression.

Naros (changes requested):

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
minleejae: I replcaed the enum comparision logic with explicit capability flas on SchemaEvolutionMode.
The call sites now use schemaEvolution() and validateOnStartup(), so future modes can define ther behavior explicitly. 5b48482

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: I clarifiy that `none-validated` only checks whether statically resolvable destination tables exist during startup.
The docs now also state that it does not validate table structure, and list the cases that can and cannot be validated at startup. 
78e1260

Naros (approved):

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -333,25 +355,34 @@ public enum SchemaEvolutionMode implements EnumeratedValue {
         /**
          * No schema evolution occurs, assumed that the destination table's structure matches the event.
          */
-        NONE("none"),
+        NONE("none", false, false),
+
+        /**
+         * No schema evolution occurs, but statically resolvable destination tables are validated at startup.
+         */
+        NONE_VALIDATED("none-validated", true, false),
```
rk3rn3r: @Naros is the wording okay? `none validated` sounds like it is not validated but actually is because it means no schema evolution but validated.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/DefaultRecordWriter.java
```diff
@@ -185,7 +183,7 @@ private TableDescriptor alterTableIfNeeded(CollectionId collectionId, JdbcSinkRe
             }
         }
 
-        if (NONE.equals(config.getSchemaEvolutionMode())) {
+        if (!config.getSchemaEvolutionMode().schemaEvolution()) {
```
rk3rn3r: I am wondering why this isn't `NONE || NONE_VALIDATED` ? Ah wait, it is replaced with the value of `schemaEvolution()` never mind.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -333,25 +355,34 @@ public enum SchemaEvolutionMode implements EnumeratedValue {
...
-        BASIC("basic");
+        BASIC("basic", false, true);
 
         // /**
         // * When an event is received, the table will be created if it does not exist, and any new fields
         // * found in the event will be added to the existing table's schema. Any columns from the table
         // * schema not found in the event will be dropped.
         // */
-        // ADVANCED("advanced");
+        // ADVANCED("advanced", false, true);
```
rk3rn3r, line 379: What about this enum?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/StartupCollectionValidator.java
rk3rn3r: Is it mandatory to only call / use it on `Startup`? If not, probably it could be simply renamed to `CollectionValidator` or `CollectionsExistsValidator` in order to describe a little more WHAT it is doing instead of WHEN it is supposed to be used.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/StartupCollectionValidator.java
```diff
@@ -0,0 +1,143 @@
...
+                            JdbcSinkConnectorConfig.SCHEMA_EVOLUTION,
+                            SchemaEvolutionMode.NONE_VALIDATED.getValue(),
+                            JdbcSinkConnectorConfig.SCHEMA_EVOLUTION,
+                            SchemaEvolutionMode.BASIC.getValue()));
+                }
+            }
+        }
+    }
+
+    Set<CollectionId> resolveCollectionIds() {
```
rk3rn3r: I have the feeling we are having the same or very similar code somewhere else. Can you have a look if this is the case and probably reuse/share the code?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/StartupCollectionValidator.java
```diff
@@ -0,0 +1,143 @@
...
+            throw new DebeziumException("'" + JdbcSinkConnectorConfig.SCHEMA_EVOLUTION + "=" + SchemaEvolutionMode.NONE_VALIDATED.getValue()
+                    + "' cannot validate target tables at startup when '" + JdbcSinkConnectorConfig.COLLECTION_NAME_FORMAT
+                    + "' contains source field placeholders.");
+        }
+
+        if (!collectionNameFormat.contains(TOPIC_PLACEHOLDER)) {
+            return Set.of(collectionNameFormat);
+        }
+
+        if (!Strings.isNullOrEmpty(props.get(SinkTask.TOPICS_REGEX_CONFIG))) {
```
rk3rn3r: Shouldn't this be moved to top so that we return / fail early? The previous checks are rather complex and then it might fail here.
Furthermore this constraint makes it probably very unusable.
Can we get around this? Deriving matching topic names and pass them to the validator?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/CollectionsExistsValidator.java
```diff
@@ -0,0 +1,143 @@
...
+        return parsedTopics.stream()
+                .filter(topic -> !topic.isEmpty())
+                .collect(Collectors.toCollection(LinkedHashSet::new));
+    }
+
+    @SuppressWarnings("deprecation")
+    private boolean usesDefaultCollectionNamingStrategy() {
+        CollectionNamingStrategy strategy = config.getCollectionNamingStrategy();
+        if (strategy instanceof TemporaryBackwardCompatibleCollectionNamingStrategyProxy proxy) {
+            strategy = proxy.getOriginalStrategy();
```
rk3rn3r, line 154: Whats the reason for handling the Proxy explicitly?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
```diff
@@ -368,6 +399,14 @@ public String getValue() {
             return mode;
         }
 
+        public boolean validateOnStartup() {
+            return validateOnStartup;
+        }
+
+        public boolean schemaEvolution() {
```
rk3rn3r: When this is not returning a SchemaEvolution strategy (which might be a cleaner solution) then this should be renamed to reflect it is a boolean information, like `schemaEvolutionEnabled()` or similar. Or move the logic into a class instance and return the appropiate strategy.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorTask.java
```diff
@@ -126,8 +126,10 @@ public void start(Map<String, String> props) {
             config.validate();
 
             sessionFactory = config.getHibernateConfiguration().buildSessionFactory();
-            StatelessSession session = sessionFactory.openStatelessSession();
             DatabaseDialect dialect = DatabaseDialectResolver.resolve(config, sessionFactory);
+            new StartupCollectionValidator(config, sessionFactory, dialect, props).validate();
```
rk3rn3r: Instead of passing all props, could you just pass `TOPICS_REGEX_CONFIG` and `TOPICS_CONFIG` that are used inside the validator? And probably we can fetch the matching topics directly and pass them to the validator? In oder to avoid that it won't work with `TOPICS_REGEX_CONFIG` ?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
rk3rn3r: I am sorry, that I added my comment below that I don't think `schemaEvolution()`  is a good method name when it is not returning a SchemaEvolution strategy, but a boolean information.

rk3rn3r:
I have added some nit-pick comments, not all might require changes but a short conversation maybe.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
Naros: I'm not really sure there is any better wording for this aside from just calling it `validated`, but the concern is that you lose the intent of no evolution. We even discussed my concerns about "validated" because it could be interpreted in several ways, and we landed on being clear in the docs about what the mode does. I think that's really the best we can do for this tbh, but I'm open to suggestions.

rk3rn3r: `only-validate` or `validate-only` maybe?

Naros:
Currently waiting for:
* #7411

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/StartupCollectionValidator.java
minleejae: Agreed. I renamed it to `CollectionsExistsValidator`, since validation can also happen during partition assignment. Addressed in [99a964a](https://github.com/debezium/debezium/pull/7599/commits/99a964a6538c933ba5a511a09c135c29124492a9).

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
minleejae: Good point. I renamed `schemaEvolution()` to `isSchemaEvolutionEnabled()` to make its boolean semantics clearer. Addressed in [b8499e9](https://github.com/debezium/debezium/pull/7599/commits/b8499e9d8eca907d9217b1545c6e392d670d95d8).

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorTask.java
minleejae: Agreed. The validator now receives only the topic configuration it needs instead of the full properties map. Addressed in [8f3f66b](https://github.com/debezium/debezium/pull/7599/commits/8f3f66bb142ebc389e84739b9b5f7b48c2afd7b0).

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/StartupCollectionValidator.java
minleejae: Good suggestion. I added a topic-based collection name resolver and reused the default naming strategy during validation. Addressed in [fdea229](https://github.com/debezium/debezium/pull/7599/commits/fdea229618bb7d0fb9e1f8350c4c518936d681c7).

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/StartupCollectionValidator.java
minleejae: Agreed. For `topics.regex`, table validation now runs in `SinkTask#open()` after partition assignment and before any records are processed. I also added test coverage and updated the documentation. Addressed in [625bea4](https://github.com/debezium/debezium/pull/7599/commits/625bea4d15b9cc75f3c22b7e724202be0dd722cf).

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
minleejae: @rk3rn3r 
I renamed it to isSchemaEvolutionEnabled() to make the boolean semantics explicit. Addressed in b8499e9                                                                .

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
minleejae: `ADVANCED` is currently commented out, but I updated its example constructor arguments to (false, true) for consistency: it would not validate tables at startup, but it would enable schema evolution.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/CollectionsExistsValidator.java
minleejae: The configuration wraps even the default naming strategy in the backward-compatibility proxy. We unwrap it so the validator can recognize and reuse the underlying default strategy. I added a comment to clarify this in [ad38ca3](https://github.com/debezium/debezium/pull/7599/commits/ad38ca3fa12181fdaa8943a32c10f898ed67c8bc).

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
Naros: I'll leave this up to @minleejae, but I'd be fine with `validate-only` if he wants it changed.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
Naros: I could be mistaken, but I think ADVANCED has always been commented out. The goal was to eventually introduce it if we could design a good way to handle real schema change ingestion from Debezium sources. 

OpenCDC is laying some great groundwork for this.

rk3rn3r: @Naros Do you have a link to OpenCDC for this?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/CollectionsExistsValidator.java
rk3rn3r: Gotcha. Why does it require the unwrapped strategy? This can potentially return `TableNamingStrategy` instance based on the connector config. We are going to remove the proxy in 3.7 now, I assume.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
Naros: The spec is still in flux and subject to change. But my personal goal is to ensure it can really push toward handling more advanced replication at the DDL level, rather than just DML with basic add-new-field logic.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/JdbcSinkConnectorConfig.java
minleejae: @rk3rn3r @Naros Thanks for letting me decide. 
I renamed the mode to `validate-only` and updated the enum, tests, error messages, and documentation. Addressed in [bf8d952](https://github.com/debezium/debezium/pull/7599/commits/bf8d952f2c8bb9cd29bf0afc8a0e6a6f4dc406c0)

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/CollectionsExistsValidator.java
minleejae: The unwrapping is needed because the connector currently wraps even the default CollectionNamingStrategy. The validator needs the underlying DefaultCollectionNamingStrategy to resolve a collection name from a topic without a record. A custom legacy TableNamingStrategy does not pass the subsequent type check and is rejected.                    

I agree that removing the proxy in 3.7 would simplify this. Since that also removes compatibility for the deprecated TableNamingStrategy API and related configuration alias, I think the full cleanup should be tracked separately unless you prefer it to be included in this PR.

Naros:
@rk3rn3r could you follow-up on minleejae's comments?

rk3rn3r:
@minleejae Great work! Thx for all those adjustments! LGTM @Naros

rk3rn3r (approved):

Naros:
Applied, thanks @rk3rn3r and @minleejae, great collab! :tada:

PR #7600, opened 2026-07-03
debezium/dbz#2135 Map PostgreSQL box, circle, line, lseg, path, polygon to first-class schemas

sundongkim-dev:
Fixes debezium/dbz#2135

## Description

The PostgreSQL connector mapped only `point` to a Connect schema. The other built-in geometric types — `box`, `circle`, `line`, `lseg`, `path`, `polygon` — had no converter, so they were emitted as opaque `bytea` (or dropped when `include.unknown.datatypes` was off). Their values were unusable downstream and PostgreSQL → PostgreSQL replication was broken because `bytea` does not re-insert into a geometric column.

This maps all six to first-class schemas and adds the PostgreSQL sink bindings needed to round-trip them, following the direction agreed on the issue.

## Source mapping (debezium-connector-postgres)

- `box`, `lseg`, `path`, `polygon` → `io.debezium.data.geometry.Geometry` (WKB). These four have a faithful OGC WKB form (a `box` is a coordinate-exact 5-point ring), so they reuse the existing cross-vendor `Geometry` type. An `extensions` map carries the originating native type, plus the open/closed flag for `path`, so the value can be rebuilt losslessly.
- `circle` → `io.debezium.data.geometry.Circle` `{ center: Point, radius }`.
- `line` → `io.debezium.data.geometry.Line` `{ a, b, c }`.

`circle` and `line` get dedicated types because WKB has no curve or infinite-line primitive; carrying the exact values lets each sink pick the best (possibly lossy) target representation instead of baking a polygon approximation into the wire at the source.

## connector-common

- `Geometry` gains an optional `MAP<STRING,STRING>` `extensions` field (schema version 1 → 2). The field is optional, so existing consumers that read only `wkb`/`srid` see no change from its addition.
- New `Circle` and `Line` semantic types, modelled on `Point`'s shape but as standalone structs (not `extends Geometry`, which would add always-null `wkb`/`srid` fields). `Geography` rejects a non-empty `extensions` map, since its v1 schema has no such field.
- New `io.debezium.spatial.WkbWriter` / `WkbReader` encode and decode `LineString` and `Polygon` coordinates, reusing the existing spatial tooling. `WkbReader` rejects a multi-ring (polygon-with-holes) geometry rather than flattening its rings, since a native PostgreSQL `polygon` is a single ring.

## JDBC sink (PostgreSQL)

- `GeometryType` reconstructs the native PostgreSQL text from the WKB and binds via `cast(? as box|lseg|path|polygon)` when the propagated source-column type is one of the native geometric types, so no PostGIS is required; otherwise it keeps the existing `ST_GeomFromWKB` binding. The native-vs-PostGIS decision is made from the schema rather than the value, because the batch SQL string is built once from the first record — a value-based decision would make the placeholder count unstable across a mixed or null-leading batch.
- New `CircleType` / `LineType` bind to native `circle` / `line`.
- The UNNEST batch path binds each column as a SQL array via `createArrayOf`, which cannot take a struct element, so a batch carrying any struct-valued column (the geometric types, including the existing `Point`) falls back to the row-wise path that runs the per-value reconstruction.
- Other dialects' `circle`/`line` sink types are left for a follow-up, matching how `Point` is currently bound only for PostgreSQL and the MySQL family.

## connect-plugins

`GeometryFormatTransformer` previously threw when asked to add an SRID to a geometry that has none. The native geometric types are emitted without an SRID, so it now leaves such a value unchanged instead of failing the record.

## Scope / limitation

- Native (non-PostGIS) reconstruction of `box`/`lseg`/`path`/`polygon` on the sink relies on the source type reaching the schema via `column.propagate.source.type`; without it, these fall back to the PostGIS `ST_GeomFromWKB` binding.
- `circle` / `line` are bound natively on the PostgreSQL sink only.
- For a NOT NULL geometric column with no default and a null value, the fallback is shaped so the sink can rebuild a value the native type accepts: a full ring for `box`, a two-point segment for `lseg`, and `{0,1,0}` for `line` (rather than the all-zero triple, which PostgreSQL rejects).

## Tests

- Unit: WKB round-trip, the new `Circle`/`Line`/`Geometry` schemas, and the PostgreSQL sink bindings.
- Integration: a PG → PG native round-trip with no PostGIS, plus the existing PostgreSQL domain-alias snapshot/streaming assertions updated to the structured schemas.

Documentation updated for both the PostgreSQL source and the JDBC sink type mappings.

## PR Checklist

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java
@@ -1301,27 +1301,32 @@ public class PostgresValueConverter extends JdbcValueConverters {
 
     /**
      * Builds the non-null fallback a NOT NULL geometric column falls back to when a null value is
-     * received with no column default (see {@link #convertValue}). The WKB must be shaped so the sink
-     * can reconstruct a value the target native type actually accepts: a {@code box} is rebuilt from
-     * ring vertices 0 and 2 (so it needs a full ring), an {@code lseg} is a two-point segment, while a
-     * {@code path}/{@code polygon} tolerate a single point. All coordinates are the origin.
+     * received with no column default (see {@link #convertValue}). When {@code column.propagate.source.type}
+     * is unset the sink binds these through PostGIS {@code ST_GeomFromWKB}, which enforces the OGC
+     * minimum-vertex rules (a ring needs at least four points, a line string at least two) and rejects
+     * degenerate geometry. So each fallback is a small but genuinely valid shape rather than a repeated
+     * origin point: a {@code box}/{@code polygon} is a unit square/triangle ring (the box is rebuilt from
+     * ring vertices 0 and 2, which the square places on opposite corners), and an {@code lseg}/{@code path}
+     * is a two distinct-point segment.
      */
-    private static Struct geometryFallback(Schema schema, String type) {
+    // package-private for direct unit testing of the OGC minimum-vertex invariant
+    static Struct geometryFallback(Schema schema, String type) {
         final byte[] wkb;
         switch (type) {
             case "box":
                 wkb = WkbWriter.buildPolygon(List.of(List.of(
-                        new double[]{ 0, 0 }, new double[]{ 0, 0 }, new double[]{ 0, 0 },
-                        new double[]{ 0, 0 }, new double[]{ 0, 0 })));
+                        new double[]{ 0, 0 }, new double[]{ 0, 1 }, new double[]{ 1, 1 },
+                        new double[]{ 1, 0 }, new double[]{ 0, 0 })));
                 break;
             case "lseg":
-                wkb = WkbWriter.buildLineString(List.of(new double[]{ 0, 0 }, new double[]{ 0, 0 }));
+                wkb = WkbWriter.buildLineString(List.of(new double[]{ 0, 0 }, new double[]{ 1, 0 }));
                 break;
             case "polygon":
-                wkb = WkbWriter.buildPolygon(List.of(List.of(new double[]{ 0, 0 })));
+                wkb = WkbWriter.buildPolygon(List.of(List.of(
+                        new double[]{ 0, 0 }, new double[]{ 1, 0 }, new double[]{ 0, 1 }, new double[]{ 0, 0 })));
                 break;
             default: // path
-                wkb = WkbWriter.buildLineString(List.of(new double[]{ 0, 0 }));
+                wkb = WkbWriter.buildLineString(List.of(new double[]{ 0, 0 }, new double[]{ 1, 0 }));
                 break;
         }
         return Geometry.createValue(schema, wkb, null, Map.of(Geometry.EXTENSION_TYPE_KEY, type));
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java
@@ -31,6 +31,7 @@ import java.util.Arrays;
 import java.util.BitSet;
 import java.util.Collections;
 import java.util.Date;
+import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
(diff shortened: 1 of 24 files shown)


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java
```diff
@@ -1128,6 +1160,186 @@ else if (data instanceof PgProto.Point) {
...
+        });
+    }
+
+    /**
+     * Converts a PostgreSQL {@code polygon} to a {@link Geometry} value encoded as a single closed
+     * polygon ring.
+     */
+    protected Object convertPolygon(Column column, Field fieldDefn, Object data) {
+        final Schema schema = fieldDefn.schema();
+        return convertValue(column, fieldDefn, data, geometryFallback(schema, "polygon"), (r) -> {
```
Naros, line 1236: Correct me if I am not mistaken, but doesn't a polygon require 4 points (3 distinct + closing)?  Wouldn't this lead to a PG sink rejecting this?

Naros (commented):
Hi @sundongkim-dev overall I think the PR looks good, just one question.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresValueConverter.java
sundongkim-dev: You're right that a WKB polygon ring has to be closed (first == last) and needs at least 4 points. But that's a `ST_GeomFromWKB` constraint, and for `box`/`lseg`/`path`/`polygon` the value never goes through PostGIS, so a live value won't be rejected.

On the source side the ring is always closed — `convertPolygon` calls `closeRing(...)` before encoding. (A PG `polygon` isn't actually bound by the OGC 4-point minimum — I checked, and PG 14/17/18 all accept even a 1- or 2-vertex `polygon`; only an empty `()` is rejected — but since real values go through the native cast below, not PostGIS, that minimum never comes into play here anyway.)

On the sink side there's no PostGIS in the path for these four. `getQueryBinding` returns `cast(? as polygon)` rather than `ST_GeomFromWKB`, and `reconstructNativeText` rebuilds the canonical PG text from the WKB, dropping the closing vertex that WKB requires but the native `polygon` text doesn't:

```java
final List<double[]> ring = WkbReader.readPolygonRing(wkb);
final List<double[]> vertices = isRingClosed(ring) ? ring.subList(0, ring.size() - 1) : ring;
return "(" + joinPoints(vertices) + ")";
```

Native `polygon` is just a vertex list — it doesn't enforce closure or the OGC 4-point minimum — so the original polygon round-trips unchanged and the cast accepts it. WKB is only the transport here. (A real PostGIS `geometry` source still uses `ST_GeomFromWKB`, but that WKB is PostGIS-produced and already valid.)

Where your comment does bite is the NOT NULL fallback. `geometryFallback` currently emits a single-point ring for polygon:

```java
case "polygon":
    wkb = WkbWriter.buildPolygon(List.of(List.of(new double[]{ 0, 0 })));
```

`WkbWriter` doesn't validate, so this open 1-point ring slips through. On the native `cast(? as polygon)` path it would actually still be accepted (I confirmed `cast('((0.0,0.0))' as polygon)` round-trips on PG 14/17/18), but it breaks on the PostGIS fallback — i.e. when `column.propagate.source.type` is unset and the value falls back to `ST_GeomFromWKB`, which rejects it with *"Polygon must have at least four points in each ring"*. The `box` fallback right above it is already a closed 5-point ring and passes that same path, so polygon is the odd one out, and the comment claiming path/polygon "tolerate a single point" is wrong. I'll fix the fallback and the comment.

One thing I'd check with you there: how strict should the synthesized fallback ring be? A degenerate closed ring of repeated `(0,0)` is enough for both the native path and PostGIS (it matches the box fallback), but if you'd rather every ring we build be strictly OGC-valid — 3 distinct vertices plus closing, e.g. `(0,0),(1,0),(0,1),(0,0)` — I'll hold both the normal and fallback rings to that same rule. Which do you prefer?

Naros: My apologies if I wasn't sufficiently clear, but you articulated my concern precisely, so thank you.

I genuinely think we should have fallbacks that cover the common denominator for maximum compatibility. In this case, with `column.propagate.source.type` unset, we should ensure we adhere to the `ST_GeomFromWKB` requirement with 3 distinct vertices plus a closing.

The implementation of how we arrive at that, can be one of a few things:

1. We modify the fallback logic on the source
2. A transform/custom converter can be plugged in to handle these use cases selectively
3. JDBC sink is made adaptive

I genuinely think (1) makes the most sense here with (2) as the preferred fallback. 

@jpechane any opinion/

sundongkim-dev: Thanks — went with your call here: I've held every synthesized ring/line to the strict OGC-valid rule on the source side (option 1), so the common-denominator `ST_GeomFromWKB` path works even with `column.propagate.source.type` unset.

`geometryFallback` now emits:
- `box` → unit square ring `(0,0),(0,1),(1,1),(1,0),(0,0)` (the sink still rebuilds the box from ring vertices 0 and 2, which the square places on opposite corners)
- `polygon` → triangle `(0,0),(1,0),(0,1),(0,0)` — 3 distinct vertices + closing
- `lseg` / `path` → two distinct-point segment `(0,0),(1,0)`

Verified on real PostGIS: all four now pass `ST_GeomFromWKB` (and the native `cast(? as …)` path), while the old single-point polygon is rejected with exactly *"Polygon must have at least four points in each ring"*. Also fixed the stale comment that claimed path/polygon tolerate a single point, and added `GeometryFallbackTest` asserting the OGC minimum-vertex invariant. Pushed in `3f270cc`.

Naros (approved):
Thanks @sundongkim-dev, the code LGTM. This just waits on:
* #7411

Naros:
Rebased after applying #7411, lets make sure CI is still clean.

Naros:

PR #7602, opened 2026-07-03
debezium/dbz#2180 Add CockroachDB integration tests for the JDBC sink dialect

viragtripathi:
Closes https://github.com/debezium/dbz/issues/2180
Closes https://github.com/debezium/dbz/issues/2182

Follow-up to #7570, as requested in https://github.com/debezium/dbz/issues/1639#issuecomment-4874195760.

Adds CockroachDB to the JDBC sink integration test matrix: the testcontainers-cockroachdb dependency, a COCKROACHDB SinkType, a CockroachDbSinkDatabaseContextProvider running docker.io/cockroachdb/cockroach:v25.4.12, and an it-cockroachdb entry in the CI workflow matrix.

Integration tests under integration/cockroachdb:
- Insert modes and delete enabled reuse the shared abstract suites; upsert exercises the dialect's INSERT ... ON CONFLICT statement.
- Column type mapping covers uuid and jsonb coercion, bytes, and text, int and bool arrays.
- JdbcSinkTransactionConflictIT verifies the transaction-conflict retry contract: a conflicting flush recovers through the retriable conflict exceptions the dialect registers once the contending transaction ends, and flush.max.retries=0 disables retries and fails fast with the conflict exception as cause (per https://github.com/debezium/dbz/issues/1639#issuecomment-4876725311, reusing the existing flush.max.retries and flush.retry.delay.ms options).

The new tests exposed a dialect bug, fixed in the first commit (dbz#2182): string array writes failed with PSQLException: Unable to find server array type for provided name string, because the Hibernate CockroachDialect names character types 'string', which is absent from pg_type and cannot be resolved by PgJDBC in Connection#createArrayOf. CockroachDBDatabaseDialect now returns 'text' for the character JDBC types.

Verification: all 62 CockroachDB integration tests pass against the container; the it-postgresql suite result is unchanged from main; unit tests, formatting, and checkstyle pass.


Naros (approved):
LGTM, thanks @viragtripathi 🎉

PR #7604, opened 2026-07-04
debezium/dbz#2100 Bind numeric array element type without precision modifier

sundongkim-dev:
Fixes debezium/dbz#2100 (PostgreSQL → PostgreSQL JDBC sink type-preservation audit), case 8.

## What

A `numeric(N,S)[]` source column fails to sink into a native PostgreSQL array column, even with the target table pre-created and `schema.evolution=none`:

```
org.postgresql.util.PSQLException: Unable to find server array type for provided name decimal(10,2)
```

`ArrayType.bind` passes the element type name straight to `Connection#createArrayOf`. For a decimal element that name is `decimal(10,2)` (it carries the precision/scale, which is correct for the DDL), but `createArrayOf` looks the argument up as a server type and only accepts the bare base type name — `numeric`/`decimal`, no brackets, no modifier.

## Fix

Strip the array brackets and the length/precision modifier from the element type name before binding, so `createArrayOf` receives the base type. The DDL type name (`getTypeName`) is untouched, so the target column is still created as `numeric(N,S)[]` and precision is preserved end to end.

The UNNEST batch path (`UnnestRecordWriter.getSqlTypeName`) already applies the same stripping; this brings the row-wise binding path in line. Keeping the change local to `ArrayType` avoids a `core → dialect.postgres` dependency.

## Tests

- Unit: `ArrayTypeTest` covers stripping the precision/scale modifier, stripping array brackets, and leaving a bare base type unchanged.
- Integration: `JdbcSinkColumnTypeMappingIT#testShouldWorkWithNumericArrayWithPrecisionAndScale` round-trips `numeric(10,2)[]` = `{1.25, 2.50, -9999999.99}` into a pre-created native `numeric(10,2)[]` column with `schema.evolution=none`, across both the standard and UNNEST insert paths. It fails on `main` and passes with this change.

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/ArrayType.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/ArrayType.java
@@ -44,6 +44,26 @@ public class ArrayType extends AbstractType {
         if (value == null) {
             return List.of(new ValueBindDescriptor(index, null));
         }
-        return List.of(new ValueBindDescriptor(index, value, java.sql.Types.ARRAY, getElementTypeName(this.getDialect(), schema, false)));
+        final String elementTypeName = baseElementTypeName(getElementTypeName(this.getDialect(), schema, false));
+        return List.of(new ValueBindDescriptor(index, value, java.sql.Types.ARRAY, elementTypeName));
+    }
+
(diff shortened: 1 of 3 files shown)


Naros (approved):
LGTM, just waiting on CI.

Naros:
This waits for:

Naros:

PR #7605, opened 2026-07-05
DBZ-2187 Fix DATETIME(4-6) precision truncation in ZeroDateFallbackConverter

skdus531:
Fixes debezium/dbz#2187

## Description

`ZeroDateFallbackConverter.registerDatetime()` previously hardcoded the `io.debezium.time.Timestamp` (milliseconds) schema for every `DATETIME` column, regardless of the column's fractional-second precision. Because the converter registers itself as a `CustomConverter` and fully replaces the schema/value conversion for the column, this silently truncated `DATETIME(4-6)` values to milliseconds — even though the normal Debezium converter path (`JdbcValueConverters` / `TemporalPrecisionMode`) maps `DATETIME(4-6)` to `io.debezium.time.MicroTimestamp` (microseconds).

For example, `2024-06-15 12:34:56.123456` in a `DATETIME(6)` column was emitted as `1718454896123` (`Timestamp`, millis) instead of `1718454896123456` (`MicroTimestamp`, micros).

Changes:

* Branch on `RelationalColumn.length()` to select `MicroTimestamp` (INT64 micros) for precision `>= 4`, and keep `Timestamp` (INT64 millis) for precision `0-3` or when length is absent.
* Use `io.debezium.time.Conversions.toEpochMicros` for the fallback value and row-level conversion on the micros path.
* Add unit tests for `DATETIME(3)`, `DATETIME(4)`, `DATETIME(5)`, `DATETIME(6)` and the no-precision case; extend `mockColumn()` to stub `RelationalColumn.length()`.

MySQL/MariaDB cap `DATETIME` precision at 6, so no nanosecond mapping is required.

**Scope note:** this change only aligns the converter with the precision-based *adaptive* mapping. It intentionally does not make the converter honor `time.precision.mode` (`connect`, `isostring`,`microseconds`, `nanoseconds`) — the converter already hardcodes semantic types for `DATE` and `TIMESTAMP` and never consulted that property. Wiring `time.precision.mode` through the converter is a larger, separate change and is left as a follow-up, keeping this PR minimal.

## PR Checklist

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -841,3 +841,4 @@ Roshni R
 刘镕通
 Sundong Kim
 Timur Rakhmatullin
+skdus531
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/ZeroDateFallbackConverter.java
@@ -37,9 +37,15 @@ import io.debezium.util.Strings;
  * <h2>Supported types</h2>
  * <ul>
  *   <li>{@code DATE} &rarr; {@code io.debezium.time.Date} (INT32 epoch days)</li>
- *   <li>{@code DATETIME} &rarr; {@code io.debezium.time.Timestamp} (INT64 epoch millis)</li>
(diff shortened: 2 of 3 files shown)


Naros (approved):
LGTM, just waiting on CI.

Naros:
Thanks @skdus531, applied to main and 3.6. :tada:

PR #7608, opened 2026-07-06
debezium/dbz#2188 Mask RocketMQ ACL secret key in RocketMqConfig#toString()

TimurRakhmatullin86:
Fixes debezium/dbz#2188

This masks the RocketMQ ACL `secretKey` in `RocketMqConfig#toString()`, which was rendered in clear text and could leak into logs, exception messages, or debugger output. It uses the project's standard `"********"` placeholder (as in `Configuration#withMaskedPasswords`); a `null` secret is still shown as `null`, and the non-sensitive fields (including the `accessKey` identifier) remain visible. Adds `RocketMqConfigTest` covering the masked and null cases.

### DIFF
--- a/debezium-storage/debezium-storage-rocketmq/src/main/java/io/debezium/storage/rocketmq/RocketMqConfig.java
+++ b/debezium-storage/debezium-storage-rocketmq/src/main/java/io/debezium/storage/rocketmq/RocketMqConfig.java
@@ -81,7 +81,7 @@ public class RocketMqConfig {
                 ", groupId='" + groupId + '\'' +
                 ", aclEnable=" + aclEnable +
                 ", accessKey='" + accessKey + '\'' +
-                ", secretKey='" + secretKey + '\'' +
+                ", secretKey='" + (secretKey == null ? null : "********") + '\'' +
                 '}';
     }
 
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -839,3 +839,4 @@ Aleksandr Pavlyuk
 Roshni R
 刘镕通
 Sundong Kim
+Timur Rakhmatullin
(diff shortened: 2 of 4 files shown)


mfvitale (approved):

debezium-storage/debezium-storage-rocketmq/src/main/java/io/debezium/storage/rocketmq/RocketMqConfig.java
```diff
@@ -81,7 +81,7 @@ public String toString() {
                 ", groupId='" + groupId + '\'' +
                 ", aclEnable=" + aclEnable +
                 ", accessKey='" + accessKey + '\'' +
-                ", secretKey='" + secretKey + '\'' +
+                ", secretKey='" + (secretKey == null ? null : "********") + '\'' +
```
Naros, line 84: Should we use `Strings.mask` instead?

TimurRakhmatullin86: Thanks @Naros — I looked at `Strings.mask(original, mask, sensitives...)` when writing this. It works by value substitution: it reduces over the sensitives doing `masked.replace(sensitive, "***")` on the whole string (side note: the `mask` parameter is currently unused — the replacement is hardcoded to `"***"`). Two edge cases made me keep the explicit per-field masking:

- the varargs are filtered for `null` only, not for empty strings — so if `secretKey` is ever `""` (e.g. ACL disabled), the reduction becomes `fullString.replace("", "***")`, which per `String.replace` empty-target semantics inserts `***` between every character and mangles the entire line;
- it replaces the secret's value anywhere it occurs, so if the secret happens to be a substring of another field's value, the output gets rewritten unpredictably.

Field-level masking sidesteps both and keeps `toString()` deterministic. That said, if you'd prefer `Strings.mask` for consistency, I can apply it over the built string with a null/empty guard on `secretKey` — happy to push that right away. And if the empty-string behavior of `Strings.mask` itself is worth hardening, I'd be glad to open a small follow-up.

Naros:
Rebased to fix the Aliases.txt/COPYRIGHT.txt conflicts.

Naros (approved):

Naros:
Thanks @TimurRakhmatullin86 applied :tada:

PR #7611, opened 2026-07-06
debezium/dbz#1176 Fix database name filtering to treat names as literals

Chaitanya-030:
Support database names with special regex characters via escaping

## Problem

Database and schema names containing special regex metacharacters were misinterpreted as regex patterns, preventing users from capturing databases with legitimate naming conventions:

- `Test$user` → `$` treated as regex "end-of-line" anchor instead of literal character
- `db[prod]` → `[prod]` interpreted as character class instead of literal text
- `db.config` → `.` matches any character instead of literal dot
- `app{v1}` → `{}` treated as regex quantifier

## Solution

Implemented character escaping approach for database and schema filtering:

**Key Features:**
- ✅ Users escape special characters in config: `Test\$user` matches literal `Test$user`
- ✅ Unescaped patterns work as regex: `db[0-9]` still matches `db0`, `db1`, etc.
- ✅ Supports all regex metacharacters: `$ . [ ] ( ) ? * + | ^ \ { } -`
- ✅ Smart transformation: only escapes values that were explicitly escaped
- ✅ Uses Java's `Pattern.quote()` for safe, standardized literal matching
- ✅ 100% backward compatible with existing regex patterns

## Changes

Modified `debezium-connector-common/src/main/java/io/debezium/relational/Selectors.java`:
- Added `unescapeAndQuoteIfNeeded()` transformation function
- Updated 6 builder methods to use new escaping logic

Modified `debezium-connector-common/src/test/java/io/debezium/relational/SelectorsTest.java`:
- Added 4 comprehensive consolidated test methods
- 19 total tests (15 original + 4 new consolidated)

## Testing

✅ All 411 connector-common tests passing
✅ New test cases verify escaped special characters match literals correctly
✅ Unescaped regex patterns continue to work as expected
✅ Mixed configurations with both escaped literals and patterns work correctly

## User Impact

Users can now:
- Use database names with `$` characters (common in cloud: `prod$us-east$db`)
- Use database names with brackets, dots, parentheses, etc.
- Continue using unescaped regex patterns
- Mix literals and patterns in the same configuration

Example Config:
`database.include.list=user$db,prod_.*,sys[internal],app[0-9]+`

This captures:
- Literal `user$db`
- Pattern matching `prod_main`, `prod_backup`, etc.
- Literal `sys[internal]`
- Pattern matching `app0`, `app1`, `app123`, etc.

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/relational/Selectors.java
+++ b/debezium-connector-common/src/main/java/io/debezium/relational/Selectors.java
@@ -6,6 +6,7 @@
 package io.debezium.relational;
 
 import java.util.function.Predicate;
+import java.util.regex.Pattern;
 
 import io.debezium.annotation.Immutable;
 import io.debezium.function.Predicates;
@@ -41,6 +42,14 @@ public class Selectors {
         return value == null || value.trim().isEmpty();
     }
 
+    private static String unescapeAndQuoteIfNeeded(String value) {
+        String unescaped = value.replaceAll("\\\\([$\\.\\[\\]()\\?*+|\\^\\\\{}\\-])", "$1");
+        if (!unescaped.equals(value)) {
+            return Pattern.quote(unescaped);
+        }
+        return unescaped;
+    }
+
     /**
      * Implementations convert given {@link TableId}s to strings, so regular expressions can be applied to them for the
      * purpose of table filtering.
@@ -69,7 +78,7 @@ public class Selectors {
                 dbInclusions = null;
             }
             else {
-                dbInclusions = Predicates.includes(databaseNames);
+                dbInclusions = Predicates.includesLiteralsOrPatterns(databaseNames, value -> false, Selectors::unescapeAndQuoteIfNeeded);
             }
             return this;
         }
@@ -87,7 +96,7 @@ public class Selectors {
                 dbExclusions = null;
             }
             else {
-                dbExclusions = Predicates.excludes(databaseNames);
+                dbExclusions = Predicates.excludesLiteralsOrPatterns(databaseNames, value -> false, Selectors::unescapeAndQuoteIfNeeded);
             }
             return this;
         }
@@ -137,7 +146,7 @@ public class Selectors {
                 dbInclusions = null;
             }
             else {
-                dbInclusions = Predicates.includes(databaseNames);
+                dbInclusions = Predicates.includesLiteralsOrPatterns(databaseNames, value -> false, Selectors::unescapeAndQuoteIfNeeded);
             }
             return this;
         }
@@ -155,7 +164,7 @@ public class Selectors {
                 dbExclusions = null;
             }
             else {
-                dbExclusions = Predicates.excludes(databaseNames);
+                dbExclusions = Predicates.excludesLiteralsOrPatterns(databaseNames, value -> false, Selectors::unescapeAndQuoteIfNeeded);
             }
             return this;
         }
@@ -172,7 +181,7 @@ public class Selectors {
                 schemaInclusions = null;
             }
             else {
-                schemaInclusions = Predicates.includes(schemaNames);
+                schemaInclusions = Predicates.includesLiteralsOrPatterns(schemaNames, value -> false, Selectors::unescapeAndQuoteIfNeeded);
             }
             return this;
         }
@@ -190,7 +199,7 @@ public class Selectors {
                 schemaExclusions = null;
             }
             else {
-                schemaExclusions = Predicates.excludes(schemaNames);
+                schemaExclusions = Predicates.excludesLiteralsOrPatterns(schemaNames, value -> false, Selectors::unescapeAndQuoteIfNeeded);
             }
             return this;
         }
--- a/debezium-connector-common/src/test/java/io/debezium/relational/SelectorsTest.java
+++ b/debezium-connector-common/src/test/java/io/debezium/relational/SelectorsTest.java
@@ -221,6 +221,149 @@ public class SelectorsTest {
         assertNotAllowed(filter, "db2", "sc4", "B");
     }
 
+    @Test
+    public void shouldCreateFilterWithEscapedDatabaseNamesContainingSpecialCharacters() {
+        // Test escaped literals: users must escape special regex chars to match literals
+        filter = Selectors.tableSelector()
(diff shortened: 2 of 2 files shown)


debezium-connector-common/src/main/java/io/debezium/relational/Selectors.java
```diff
@@ -69,7 +69,7 @@ public DatabaseSelectionPredicateBuilder includeDatabases(String databaseNames)
                 dbInclusions = null;
             }
             else {
-                dbInclusions = Predicates.includes(databaseNames);
+                dbInclusions = Predicates.includesLiterals(databaseNames, (s) -> s, true);
```
Naros: I'm afraid the use of `includeLiterals` breaks compatibility. Any include/exclude list property is designed to support regular expressions, and this breaks that capability. Have you checked if you could use `includesLiteralsOrPatterns` instead for these use cases?

Naros (changes requested):

debezium-connector-common/src/main/java/io/debezium/relational/Selectors.java
Chaitanya-030: Ohh, let me check once.

Chaitanya-030: yes, you're correct. schema include list and database include list can have regex based patterns.

I was seeing specific to sqlserver which takes database.names , hence missed.

changed the code.

debezium-connector-common/src/main/java/io/debezium/relational/Selectors.java
```diff
@@ -41,6 +43,18 @@ private static boolean isEmpty(String value) {
         return value == null || value.trim().isEmpty();
     }
 
+    private static boolean isRegexPattern(String value) {
+        try {
+            Pattern.compile(value);
+            // Only treat as pattern if it has definite regex operator characters
+            // Literal database names typically don't have these operators
+            return value.matches(".*[*+?|{}\\\\].*"); // Regex operators and escape character
```
Naros: This would break any existing configuration where a user may have set the include list to `db[0-9]` to match `db0`, `db1`, ... `db9` because it wouldn't be recognized as a regular expression.

Naros (changes requested):

debezium-connector-common/src/main/java/io/debezium/relational/Selectors.java
Chaitanya-030: yes you're correct, I can modify the method though to include all the regex patterns notation but I think this can be a bit risky overall.

Naros: I concur. What I might recommend is take a step back and look at it from what we would expect, and that is if you have a special regex character like `$` in a name, you escape it so its treated as a literal from the maching point of view. 

Then I would suggest isolate where the connector fails and see if you can apply logic that unescapes the regex so that `Test\$Db` becomes `Test$Db` in those use cases.

This way you get the best of both worlds, regex-based include/exclude selectors plus fixes the bug you're trying to address in the first place.

Chaitanya-030: okay so whenever the user will escape something in the config, we will treat it as a literal after unescaping else will continue with our regex flow.

Naros: The unescape use case is likely on a case-by-case basis. In some scenarios, like the Oracle connector, the regex patterns are passed to a database query. In other connectors, they handle this differently, so you need to look at the specific cases where it fails.

Naros (changes requested):
Hi @sundongkim-dev so looking back over the GH issue report itself, I wonder if the issue here isn't just configuration and not an issue with the code. If you try using `main` without this PR, the user reported using:
```
database.dbname=Test$user
table.include.list=dbo.user
database.include.list=Test$user
```
The problem is this is an invalid configuration, and the right config would have been:
```
database.dbname=Test$user
table.include.list=dbo.user
database.include.list=Test\$user
```
This way the `Selectors` woudl have treated the `$` as a literal rather than a regex character. This is what would be expected for any special character in database, schema, or table include/exclude list configurations.

Then in situations where we use the `database.dbname` or the more relevant and recent `database.names` property, this would not require being escaped because it's used as a literal, so the value supplied should just work.

Could you check if that works as expected, as I genuinely think no changes are required.

sundongkim-dev:
Hi @Naros — I think you may have meant to ping @Chaitanya-030 (the PR author) here rather than me 🙂. That said, I was curious so I went ahead and verified your hypothesis, and I can confirm you're right: **escaping already works on `main` with no code changes.**

I exercised the actual filtering path on `main` (`Predicates.includes` → `Strings.setOfRegex` → `RegExSplitter` → `Pattern.compile`) directly, reproducing the scenario from debezium/dbz#1176:

| Scenario | Config | DB name | Matches? |
|---|---|---|---|
| Reporter's original (broken) config | `Test$user` | `Test$user` | ❌ (`$` is a regex end-anchor — reproduces the bug) |
| **Escaped (your suggestion)** | `Test\$user` | `Test$user` | ✅ |
| No over-matching | `Test\$user` | `Testuser` | ❌ |
| Backward-compat regex | `db[0-9]` | `db0`, `db7` | ✅ |
| Other specials | `db\.config`, `sys\[internal\]`, `app\{v1\}` | literal | ✅ |
| Mixed list | `user\$db,prod_.*` | `user$db` + `prod_main` | ✅ both |

`RegExSplitter` treats `\` as an escape sequence, so the backslash survives the comma-split, and `Pattern.compile` then interprets `\$` as a literal `$`. So the reporter's real problem was an invalid config — `database.include.list=Test\$user` is what they needed, and `database.dbname` / `database.names` are used as literals so they need no escaping at all.

So it looks like no code change is required here — the existing escaping behavior covers all the cases in the issue. The only real gap seems to be discoverability/docs (users not knowing to escape special characters). Hope this helps!

Chaitanya-030:
ohh, yes @Naros agreed 🙂. No code changes will be required in this case. Will be more careful next time and will check if the issue is genuine before starting with the changes.

Naros:
As we are in agreement, we will close the PR and mark the issue as invalid. Thanks to you both :tada:

PR #7616, opened 2026-07-07
debezium/dbz#2100 Bind macaddr8 to its native PostgreSQL type

sundongkim-dev:
Fixes debezium/dbz#2100, case 10.

## Description

`MacAddressType` only registered the `MACADDR` source column type and always emitted `cast(? as macaddr)` with a hard-coded `macaddr` type name. As a result a PostgreSQL `macaddr8` column was mapped to `text` on the sink (type downgrade), and inserting into a pre-created native `macaddr8` column failed with `column "..." is of type macaddr8 but expression is of type character varying`.

This registers `MACADDR8` as well and derives both the query-binding cast and the DDL type name from the source column type (`getSourceColumnType(schema)`), following the existing `RangeType` pattern. Both `macaddr` and `macaddr8` now resolve to their native PostgreSQL type. The 6-byte vs 8-byte distinction needs no special handling because the source connector already reports the exact type via `__debezium.source.column.type`.

Tests:
- `MacAddressTypeTest` (unit) — registration keys, and `macaddr` / `macaddr8` each resolving to the correct cast and type name.
- `JdbcSinkColumnTypeMappingIT#testShouldCoerceStringTypeToMacaddr8ColumnType` (IT) — round-trips a value into a native `macaddr8` column (auto-created via schema evolution) across all insert modes. Without the fix it reproduces the column being created as `text` instead of `macaddr8`.

## PR Checklist

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/MacAddressType.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/MacAddressType.java
@@ -12,7 +12,7 @@ import io.debezium.connector.jdbc.type.JdbcType;
 import io.debezium.sink.column.ColumnDescriptor;
 
 /**
- * An implementation of {@link JdbcType} for {@code MACADDR} column types.
+ * An implementation of {@link JdbcType} for {@code MACADDR} and {@code MACADDR8} column types.
  *
  * @author Chris Cranford
  */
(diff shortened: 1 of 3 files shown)


debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/MacAddressType.java
```diff
@@ -22,17 +22,17 @@ class MacAddressType extends AbstractType {
...
     @Override
     public String[] getRegistrationKeys() {
-        return new String[]{ "MACADDR" };
+        return new String[]{ "MACADDR", "MACADDR8" };
     }
 
     @Override
     public String getQueryBinding(ColumnDescriptor column, Schema schema, Object value) {
-        return "cast(? as macaddr)";
+        return "cast(? as " + getSourceColumnType(schema).orElseThrow() + ")";
```
Naros: This and the other use case are both behavior regressions. Now if type propagaion is not enabled, these will always throw an exception, where before it would always map to `macaddr`.

I would recommend the following instead:

```suggestion
        return "cast(? as " + getSourceColumnType(schema).orElse("macaddr") + ")";
```
I would apply the same behavior inside `getTypeName()` below.

Naros (changes requested):

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/MacAddressType.java
sundongkim-dev: Thanks, you're right — fixed in 88901e9. Both `getQueryBinding()` and `getTypeName()` now use `.orElse("macaddr")`, so with type propagation disabled they fall back to `macaddr` exactly as before instead of throwing.

Naros (approved):

Naros:

PR #7617, opened 2026-07-08
debezium/dbz#2185 Add CockroachDB target to the JDBC sink e2e pipeline tests

viragtripathi:
Closes https://github.com/debezium/dbz/issues/2185
Closes https://github.com/debezium/dbz/issues/2197

Follow-up to #7602, adding CockroachDB as a target in the JDBC sink end-to-end pipeline test matrix.

First commit (dbz#2197): the CockroachDB dialect inherits PostgreSQL type registrations that CockroachDB rejects, verified against CockroachDB v25.4.12 and v26.3.0-alpha.1 (both behave identically): xml, hstore, the range types, macaddr, cidr, money, point, halfvec, and sparsevec do not exist, and the spatial types were emitted with a PostGIS extension schema qualifier. The dialect now registers CockroachDB-compatible overrides: XML, ranges, macaddr, and cidr are stored as text; Json logical values, including propagated HSTORE columns, and MAP schema types are stored as jsonb; money as decimal(19,2); geometry, geography, and point use the built-in spatial types with unqualified names and ST_GeomFromWKB; FloatVector uses the native vector type; SparseDoubleVector reports the standard unsupported-schema-type error. Types verified as natively supported remain inherited: ltree, citext, inet, oid, tsvector, vector, bit, varbit, enum, interval, uuid.

Second commit (dbz#2185): JdbcSinkPipelineToCockroachDbIT tagged e2e-cockroachdb, modeled on the PostgreSQL target, with CockroachDB-specific expected types, an infinity timestamp override, COCKROACHDB arms in the shared expectation branches, vector coverage in the it-cockroachdb column type mapping suite, and the CI matrix entry extended to it-cockroachdb,e2e-cockroachdb.

Verification: the it-cockroachdb integration tests pass (66 tests) with formatting and checkstyle enabled. With the PostgreSQL source the e2e suite passes, including every data type affected by the dialect changes. With the MySQL source 93 tests pass with no failures; the remaining invocations and the SQL Server source could not be completed in the local environment due to container startup timeouts under emulation and rely on the CI matrix in this PR.

### DIFF
--- a/debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/cockroachdb/JdbcSinkColumnTypeMappingIT.java
+++ b/debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/cockroachdb/JdbcSinkColumnTypeMappingIT.java
@@ -47,14 +47,14 @@ public class JdbcSinkColumnTypeMappingIT extends AbstractJdbcSinkTest {
 
     @ParameterizedTest
     @ArgumentsSource(SinkRecordFactoryArgumentsProvider.class)
-    @FixFor("DBZ-1639")
+    @FixFor("debezium/dbz#1639")
     public void testShouldCoerceStringTypeToUuidColumnType(SinkRecordFactory factory) throws Exception {
         shouldCoerceStringTypeToColumnType(factory, "uuid", "9bc6a215-84b5-4865-a058-9156427c887a", "f54c2926-076a-4db0-846f-14cad99a8307");
     }
 
     @ParameterizedTest
     @ArgumentsSource(SinkRecordFactoryArgumentsProvider.class)
-    @FixFor("DBZ-1639")
+    @FixFor("debezium/dbz#1639")
     public void testShouldCoerceStringTypeToJsonbColumnType(SinkRecordFactory factory) throws Exception {
         shouldCoerceStringTypeToColumnType(factory, "jsonb", "{\"id\": 12345}", "{\"id\": 67890}");
     }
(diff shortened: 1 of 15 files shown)


debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/cockroachdb/JdbcSinkColumnTypeMappingIT.java
```diff
@@ -209,6 +212,53 @@ public void testShouldWorkWithIntArray(SinkRecordFactory factory) throws Excepti
         });
     }
 
+    @ParameterizedTest
+    @ArgumentsSource(SinkRecordFactoryArgumentsProvider.class)
+    @FixFor("DBZ-2197")
```
Naros:
```suggestion
    @FixFor("debezium/dbz#2197")
```
or simply
```suggestion
    @FixFor("dbz#2197")
```
The old `DBZ-XXXX` format is for Jira issue keys, which we no longer use. Could you adjust all references in this test class. Otherwise, then I think the PR is good to go.

Naros (changes requested):

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/cockroachdb/JdbcSinkColumnTypeMappingIT.java
viragtripathi: Done, updated all the references in this class to the debezium/dbz#NNNN format, and also fixed the DBZ-2180 references in JdbcSinkTransactionConflictIT for consistency across the CockroachDB test classes.

Naros (approved):

Naros:
Blocked by
* #7411

Naros:
Rebased to resolve conflicts after applying #7411

Naros:

PR #7621, opened 2026-07-08
debezium/dbz#2202 Avoid allocating a list to count non-toasted columns

sundongkim-dev:
Fixes https://github.com/debezium/dbz/issues/2202

## Description

`PostgresChangeRecordEmitter#columnValues()` runs on the streaming hot path — once per INSERT/DELETE and twice per UPDATE (old + new tuple), for every captured row. To size the `values` array it materialized a filtered list via `stream().filter().collect(Collectors.toList())` and then read only its `size()`; the list itself was never iterated (the following loop walks the original `columns`). The `Stream`, filter lambda, and intermediate `ArrayList` were allocated solely to obtain a count and then discarded.

## Changes

- Replace the stream/collector with a plain counting loop.
- Simplify the sizing expression to `Math.max(schemaColumns.size(), columnsWithoutToasted)`; the ternary `a < b ? b : a` is exactly `Math.max(b, a)`.
- Remove the now-unused `io.debezium.function.Predicates` import (`Collectors` is still used elsewhere in the file).
- Add a JMH benchmark (`PostgresColumnValuesCountPerf`) in `debezium-microbenchmark` comparing both implementations.

Behavior is preserved: identical array size, same predicate, no change to emitted records, schemas, order, or connector configuration. `isToastedColumn()` is a side-effect-free getter in all implementations.

## Benchmark results

Single call, `gc.alloc.rate.norm` (bytes allocated per operation), `toastedPercent = 0`:

| columnCount | stream (current) | loop (proposed) | time reduction |
|---:|---:|---:|---:|
| 4   | 296 B/op   | ≈ 0 | 0.041 → 0.002 us/op |
| 16  | 544 B/op   | ≈ 0 | 0.104 → 0.005 us/op |
| 64  | 1,160 B/op | ≈ 0 | 0.304 → 0.018 us/op |
| 256 | 4,792 B/op | ≈ 0 | 1.029 → 0.066 us/op |

The loop allocates essentially nothing; the stream allocation grows linearly with the column count.

## Testing

- `./mvnw -pl debezium-microbenchmark -am -DskipTests compile` — builds.
- `RecordsStreamProducerIT` toasted-column and REPLICA IDENTITY FULL streaming tests pass against PostgreSQL (pgoutput): `shouldNotPropagateUnchangedToastedData`, `shouldHandleToastedArrayColumn`, `shouldHandleToastedArrayColumnForReplicaIdentityFullTable`, `shouldHandleToastedByteArrayColumn`, `shouldHandleToastedIntegerArrayColumn`, `shouldReceiveChangesForReplicaIdentityFullTableWithToastedValueTableFromStreaming` — 6 run, 0 failures.

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresChangeRecordEmitter.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresChangeRecordEmitter.java
@@ -26,7 +26,6 @@ import org.slf4j.LoggerFactory;
 import io.debezium.connector.postgresql.connection.PostgresConnection;
 import io.debezium.connector.postgresql.connection.ReplicationMessage;
 import io.debezium.data.Envelope.Operation;
-import io.debezium.function.Predicates;
 import io.debezium.pipeline.spi.ChangeRecordEmitter;
 import io.debezium.pipeline.spi.OffsetContext;
 import io.debezium.pipeline.spi.Partition;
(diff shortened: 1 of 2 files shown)


Naros (approved):
LGTM, great find @sundongkim-dev. Once CI passes, we can merge.

Naros:

PR #7626, opened 2026-07-09
debezium/dbz#2199 Bind PostgreSQL tsvector verbatim instead of stripping quotes

sundongkim-dev:
Fixes debezium/dbz#2199

## Description

`TsvectorType` overrode `bind()` only to run `replaceAll("'", "")` on the value. PostgreSQL tsvector text wraps each lexeme in single quotes and escapes an internal apostrophe by doubling it, so stripping every quote merges the wrapping quotes into the lexeme content: a space then reads as a lexeme boundary and a doubled apostrophe collapses. This silently corrupts any lexeme containing a space or apostrophe — e.g. `'new york'` became `'new' 'york'`, and `'it''s'` became `'its'`.

The source connector always emits the canonical quoted tsvector text, so no unquoting is needed on the sink side. This removes the `bind()` override so the value binds verbatim through `cast(? as tsvector)`, matching how `LtreeType` handles its native cast. This is a pure removal of an unnecessary override, not new logic.

`TsvectorType` is registered only for the PostgreSQL dialect, so non-PostgreSQL sinks (which map tsvector to text) never ran the quote-stripping and are unaffected.

## Tests

- Added an end-to-end regression test (`testTsvectorDataTypePreservesQuotedLexemes`) covering a space-containing lexeme (`new york`) and an escaped apostrophe (`it's`), run across the sink matrix. Verified it fails on `main` with the reported corruption and passes with this change.
- The existing `testTsvectorDataTypeWithStaticValue` is unchanged: its value has no space or apostrophe, so it round-trips identically.

## PR Checklist

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/TsvectorType.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/TsvectorType.java
@@ -5,13 +5,10 @@
  */
 package io.debezium.connector.jdbc.dialect.postgres;
 
-import java.util.List;
-
 import org.apache.kafka.connect.data.Schema;
 
 import io.debezium.connector.jdbc.type.AbstractType;
 import io.debezium.sink.column.ColumnDescriptor;
-import io.debezium.sink.valuebinding.ValueBindDescriptor;
 
 /**
  * An implementation of {@link AbstractType} for the PostgreSQL {@code tsvector} data type.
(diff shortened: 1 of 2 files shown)


sundongkim-dev:
@Naros a small housekeeping note: my commits are authored as `Paul_Kim` (my DCO sign-off name), while I'm already registered in `COPYRIGHT.txt`/`Aliases.txt` as `Sundong Kim` (handle `sundongkim-dev`). The contributor-check bot matches the commit author name against those files, so it keeps flagging me as a new contributor even though I'm already listed.

Could you add an alias mapping for `Paul_Kim` when convenient? Something like `Paul_Kim,Sundong Kim` in `Aliases.txt` should let the check resolve it. Not urgent — just to quiet the false "new contributor" prompt on my PRs. Thanks!

Naros:
Added in https://github.com/debezium/debezium/commit/3841fa0856fd92eeb3aaf3bf17a270a6fdf7da7b

Naros (approved):

Naros:
Blocked by
* #7411

Naros:

PR #7641, opened 2026-07-10
debezium/dbz#2217 Map MySQL and MariaDB REAL columns to DOUBLE

minleejae:
Fixes debezium/dbz#2217

## Description

MySQL and MariaDB treat `REAL` as a synonym for `DOUBLE` unless the `REAL_AS_FLOAT` sql_mode is enabled ([MySQL manual](https://dev.mysql.com/doc/refman/8.4/en/numeric-type-syntax.html)); the server rewrites the column type at DDL time (`SHOW CREATE TABLE` reports `double`), stores 8-byte doubles, and binlog row events carry 8-byte doubles.

The ANTLR DDL parsers resolved the `REAL` keyword to `java.sql.Types.REAL` (single precision in JDBC terms). As a result, the same column produced different schemas depending on the capture path:

| Capture path | Emitted schema | Value `3.141592653589793` |
|---|---|---|
| Snapshot (parses `SHOW CREATE TABLE`, already normalized to `double`) | `FLOAT64` | preserved |
| Streaming (parses original binlog DDL with `REAL`) | `FLOAT32` | **silently truncated to `3.1415927`** |

This PR resolves `REAL` to `Types.DOUBLE` in `MySqlAntlrDdlParser`, `MySqlPtAntlrDdlParser` and `MariaDbAntlrDdlParser`, so the streaming path matches the snapshot path and the server behavior, and no precision is lost. Default values for `REAL` columns are now emitted as `Double` accordingly (two existing test assertions updated). The shared data type mapping documentation row for `REAL` is updated to `FLOAT64` with a `REAL_AS_FLOAT` caveat.

This applies the conclusion of DBZ-6226 ("Regardless of when the table is created, REAL data types should be treated as a DOUBLE") to the default mapping; that issue shipped only the opt-in `JdbcSinkDataTypesConverter` workaround.

Note for existing deployments: tables captured before this fix keep their recorded schema-history model (`Types.REAL`) until the table model is rebuilt (re-snapshot or a subsequent DDL statement for the table); newly parsed DDL is emitted correctly.

Verified end-to-end against MySQL 8.0.46 (`sql_mode` without `REAL_AS_FLOAT`): with the patched plugin, a `CREATE TABLE ... (v REAL)` issued during streaming now emits `"type":"double"` and `3.141592653589793` intact; on the unpatched connector the same statement emitted `"type":"float"` and `3.1415927`. New regression test `shouldTreatRealDataTypeAsDouble` in the shared `BinlogAntlrDdlParserTest` fails before the fix (`Types.REAL` vs expected `Types.DOUBLE`) and passes after; full unit suites of the binlog, mysql and mariadb modules pass.

## PR Checklist

### DIFF
--- a/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlAntlrDdlParserTest.java
+++ b/debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlAntlrDdlParserTest.java
@@ -8,6 +8,7 @@ package io.debezium.connector.mysql;
 
 import static org.assertj.core.api.Assertions.assertThat;
 
+import java.sql.Types;
 import java.util.List;
 
 import org.junit.jupiter.api.Disabled;
@@ -18,6 +19,7 @@ import io.debezium.config.CommonConnectorConfig.EventConvertingFailureHandlingMo
 import io.debezium.connector.binlog.BinlogAntlrDdlParserTest;
 import io.debezium.connector.binlog.BinlogConnectorConfig;
 import io.debezium.connector.mysql.antlr.MySqlAntlrDdlParser;
+import io.debezium.connector.mysql.antlr.MySqlPtAntlrDdlParser;
 import io.debezium.connector.mysql.charset.MySqlCharsetRegistry;
 import io.debezium.connector.mysql.jdbc.MySqlDefaultValueConverter;
 import io.debezium.connector.mysql.jdbc.MySqlValueConverters;
@@ -28,6 +30,7 @@ import io.debezium.jdbc.TemporalPrecisionMode;
 import io.debezium.relational.Column;
 import io.debezium.relational.RelationalDatabaseConnectorConfig;
 import io.debezium.relational.Table;
+import io.debezium.relational.Tables;
 import io.debezium.relational.Tables.TableFilter;
 import io.debezium.relational.ddl.DdlChanges;
 import io.debezium.relational.ddl.SimpleDdlParserListener;
@@ -56,6 +59,11 @@ public class MySqlAntlrDdlParserTest
     @Override
     protected MySqlAntlrDdlParser getParser(SimpleDdlParserListener listener, boolean includeViews, boolean includeComments) {
         return new MySqlDdlParserWithSimpleTestListener(listener, includeViews, includeComments);
+    }
+
+    @Override
+    protected MySqlAntlrDdlParser getParser(SimpleDdlParserListener listener, BinlogConnectorConfig.RealHandlingMode realHandlingMode) {
+        return new MySqlDdlParserWithSimpleTestListener(listener, realHandlingMode);
     }
 
     @Override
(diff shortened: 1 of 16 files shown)


Naros:
Hi @minleejae, I wonder if this should be guarded by a configuration option.

Depending on one's use of schema registry and whether you're using Avro/Protbufs, there are corner cases where the promotion from `FLOAT32` / `float` to `FLOAT64` / `double` is problematic and is incompatible.

We discussed this in DBZ-7693, which is why we added `JdbcSinkDataTypesConverter` for this use case, giving the user the ability to control this as they please rather than enforcing an event schema change that could be viewed as incompatible depending on pipeline setup.

If we want to fold `JdbcSinkDataTypesConverter` behavior into MySQL, I believe this should be conditionalized.

minleejae:
Hi @Naros,Thanks for the suggestion. 
I added `real.handling.mode` with `double` as the default and `float` as an explicit option for `REAL_AS_FLOAT` compatibility. Explicit FLOAT and DOUBLE mappings remain unchanged. Addressed in [1a80b35](https://github.com/debezium/debezium/pull/7641/commits/1a80b3502ecae6e33367e33ed3999e6c6be50ac4).

Naros:
Hi @minleejae could you rebase against the latest `main`?

minleejae:
Hi @Naros Rebased against the latest main.

 The two remaining failures look like flaky incremental-snapshot tests unrelated to this change:
- MariaDB 11.4 (gtids) — ReadOnlyIncrementalSnapshotIT.schemaChanges fails with [Too many no data record results, 49 < 1011]: the consume loop hit 6 consecutive empty polls (limit 5) while waiting for records, i.e. a timing starvation, not a wrong value.
- MongoDB - Apicurio (8.0) — IncrementalSnapshotIT.pauseDuringSnapshot is missing a single entry (257=256) around the pause/resume window.

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlAntlrDdlParserTest.java
```diff
@@ -100,6 +108,24 @@ public void shouldParseMySql56InitializationStatements() {
     public void shouldParseMySql57InitializationStatements() {
     }
 
+    @Test
+    @FixFor("debezium/dbz#2217")
+    public void shouldApplyRealHandlingModeWithLegacyDdlParser() {
```
Naros: One nitpick: @minleejae, should this belong in `MySqlPtAntlrDdlParserTest` instead, so we keep the legacy PT and Oracle-specific grammar tests independent?

minleejae: Thanks! Applied in https://github.com/debezium/debezium/pull/7641/changes/66321549c218ad642341647eb36108aee8da9cd8

PR #7642, opened 2026-07-10
[docs] Fix data type syntax notation in the MySQL/MariaDB type mapping tables

minleejae:
## Description

Fixes the type syntax notation in the shared MySQL/MariaDB data type mapping tables (`shared-mariadb-mysql.adoc`, rendered into both connector pages).

**Broken bracket notation** (present since 2019): the `CHAR`, `VARCHAR`, `BINARY` and `VARBINARY` rows carried a dangling closing bracket without its opening counterpart, e.g. `CHAR(M)]`. Corrected per the [MySQL string type syntax](https://dev.mysql.com/doc/refman/8.4/en/string-type-syntax.html): `CHAR[(M)]` and `BINARY[(M)]` take an optional length, while `VARCHAR(M)` and `VARBINARY(M)` require it (no brackets).

**Notation aligned with the database manuals**:

| Before | After | Reason |
|---|---|---|
| `TINYINT` | `TINYINT[(M)]` | [Numeric type syntax](https://dev.mysql.com/doc/refman/8.4/en/numeric-type-syntax.html); consistent with the SMALLINT/MEDIUMINT/BIGINT rows |
| `INT, INTEGER[(M)]` | `INT[(M)], INTEGER[(M)]` | both synonyms take the optional display width |
| `BLOB`, `TEXT` | `BLOB[(M)]`, `TEXT[(M)]` | both accept an optional length that selects the smallest sufficient variant |
| `YEAR[(2\|4)]` | `YEAR[(4)]` | `YEAR(2)` was removed in MySQL 5.7.5 and is not accepted by any MySQL version Debezium supports; the [MariaDB manual](https://mariadb.com/kb/en/year-data-type/) also documents the syntax as `YEAR[(4)]` |
| `TIME[(M)]`, `DATETIME[(M)]`, `TIMESTAMP[(M)]` | `TIME[(fsp)]`, `DATETIME[(fsp)]`, `TIMESTAMP[(fsp)]` | the manuals use `fsp` (fractional seconds precision, range 0-6) for the temporal precision argument ([date and time type syntax](https://dev.mysql.com/doc/refman/8.4/en/date-and-time-type-syntax.html)); `M` elsewhere in the table means display width/length |

The two accompanying "allows `M` to be in the range of `0-6`" sentences are updated to spell out "fractional seconds precision `fsp`".

Not touched: `FLOAT[(P)]` (already accurate), the `REAL` row (updated separately in #7641), the `DATETIME, DATETIME(0)..(3)` adaptive-mode rows (they deliberately enumerate specific precisions), and TINY/MEDIUM/LONG BLOB/TEXT variants (they take no length argument).

## PR Checklist

### DIFF
--- a/documentation/modules/ROOT/partials/modules/all-connectors/shared-mariadb-mysql.adoc
+++ b/documentation/modules/ROOT/partials/modules/all-connectors/shared-mariadb-mysql.adoc
@@ -1842,21 +1842,21 @@ include::{snippetsdir}/frag-{context}-props-snippets.adoc[leveloffset=+1,tags=ba
 |`FLOAT64`
 a|_n/a_
 
-|`CHAR(M)]`
+|`CHAR[(M)]`
 |`STRING`
 a|_n/a_
 
-|`VARCHAR(M)]`
+|`VARCHAR(M)`
 |`STRING`
 a|_n/a_
 
-|`BINARY(M)]`
+|`BINARY[(M)]`
 |`BYTES` or `STRING`
 a|_n/a_
 
 Either the raw bytes (the default), a base64-encoded String, or a base64-url-safe-encoded String, or a hex-encoded String, based on the xref:{context}-property-binary-handling-mode[`binary.handling.mode`] connector configuration property setting.
 
-|`VARBINARY(M)]`
+|`VARBINARY(M)`
 |`BYTES` or `STRING`
 a|_n/a_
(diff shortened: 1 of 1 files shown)


Naros (approved):

Naros:
Thanks @minleejae, applied. :tada:

PR #7646, opened 2026-07-11
debezium/dbz#75 Recover schema history when the binlog base name changes

### LINKED ISSUE #75: Cannot compare binlog filenames with different base names [DBZ-8981]
Migrated from [DBZ-8981](https://issues.redhat.com/browse/DBZ-8981)

h1. Bug report

My source connector randomly fails like this:
{code:java}
{
  "name": "source-db",
  "connector": {
    "state": "RUNNING",
    "worker_id": "connect-0.connect.confluent.svc.cluster.local:8083"
  },
  "tasks": [
    {
      "id": 0,
      "state": "FAILED",
      "worker_id": "connect-0.connect.confluent.svc.cluster.local:8083",
      "trace": "java.lang.IllegalArgumentException: Cannot compare binlog filenames with different base names\n\tat io.debezium.connector.binlog.history.BinlogHistoryRecordComparator$BinlogFileName.compareTo(BinlogHistoryRecordComparator.java:232)\n\tat io.debezium.connector.binlog.history.BinlogHistoryRecordComparator.isPositionAtOrBefore(BinlogHistoryRecordComparator.java:112)\n\tat io.debezium.relational.history.HistoryRecordComparator.isAtOrBefore(HistoryRecordComparator.java:33)\n\tat io.debezium.relational.history.AbstractSchemaHistory.lambda$recover$3(AbstractSchemaHistory.java:103)\n\tat io.debezium.storage.kafka.history.KafkaSchemaHistory.recoverRecords(KafkaSchemaHistory.java:338)\n\tat io.debezium.relational.history.AbstractSchemaHistory.recover(AbstractSchemaHistory.java:100)\n\tat io.debezium.relational.history.SchemaHistory.recover(SchemaHistory.java:192)\n\tat io.debezium.relational.HistorizedRelationalDatabaseSchema.recover(HistorizedRelationalDatabaseSchema.java:72)\n\tat io.debezium.schema.HistorizedDatabaseSchema.recover(HistorizedDatabaseSchema.java:40)\n\tat io.debezium.connector.common.BaseSourceTask.validateAndLoadSchemaHistory(BaseSourceTask.java:151)\n\tat io.debezium.connector.mysql.MySqlConnectorTask.start(MySqlConnectorTask.java:134)\n\tat io.debezium.connector.common.BaseSourceTask.start(BaseSourceTask.java:253)\n\tat org.apache.kafka.connect.runtime.AbstractWorkerSourceTask.initializeAndStart(AbstractWorkerSourceTask.java:282)\n\tat org.apache.kafka.connect.runtime.WorkerTask.doStart(WorkerTask.java:188)\n\tat org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:248)\n\tat org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:304)\n\tat org.apache.kafka.connect.runtime.AbstractWorkerSourceTask.run(AbstractWorkerSourceTask.java:82)\n\tat org.apache.kafka.connect.runtime.isolation.Plugins.lambda$withClassLoader$7(Plugins.java:341)\n\tat java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)\n\tat java.base/java.util.concurrent.FutureTask.run(Unknown Source)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)\n\tat java.base/java.lang.Thread.run(Unknown Source)\n"
    }
  ],
  "type": "source"
} {code}

io.debezium.connector.mysql.MySqlConnector (3.0.8)

 

 
On-premise, MySQL 8.0.41

No failing connector
h2. What behavior do you see?

Connector fails

Yes
{code:java}
{   "name": "source-translated",   "connector": {     "state": "RUNNING",     "worker_id": "connect-0.connect.confluent.svc.cluster.local:8083"   },   "tasks": [     {       "id": 0,       "state": "FAILED",       "worker_id": "connect-0.connect.confluent.svc.cluster.local:8083",       "trace": "java.lang.IllegalArgumentException: Cannot compare binlog filenames with different base names\n\tat io.debezium.connector.binlog.history.BinlogHistoryRecordComparator$BinlogFileName.compareTo(BinlogHistoryRecordComparator.java:232)\n\tat io.debezium.connector.binlog.history.BinlogHistoryRecordComparator.isPositionAtOrBefore(BinlogHistoryRecordComparator.java:112)\n\tat io.debezium.relational.history.HistoryRecordComparator.isAtOrBefore(HistoryRecordComparator.java:33)\n\tat io.debezium.relational.history.AbstractSchemaHistory.lambda$recover$3(AbstractSchemaHistory.java:103)\n\tat io.debezium.storage.kafka.history.KafkaSchemaHistory.recoverRecords(KafkaSchemaHistory.java:338)\n\tat io.debezium.relational.history.AbstractSchemaHistory.recover(AbstractSchemaHistory.java:100)\n\tat io.debezium.relational.history.SchemaHistory.recover(SchemaHistory.java:192)\n\tat io.debezium.relational.HistorizedRelationalDatabaseSchema.recover(HistorizedRelationalDatabaseSchema.java:72)\n\tat io.debezium.schema.HistorizedDatabaseSchema.recover(HistorizedDatabaseSchema.java:40)\n\tat io.debezium.connector.common.BaseSourceTask.validateAndLoadSchemaHistory(BaseSourceTask.java:151)\n\tat io.debezium.connector.mysql.MySqlConnectorTask.start(MySqlConnectorTask.java:134)\n\tat io.debezium.connector.common.BaseSourceTask.start(BaseSourceTask.java:253)\n\tat org.apache.kafka.connect.runtime.AbstractWorkerSourceTask.initializeAndStart(AbstractWorkerSourceTask.java:282)\n\tat org.apache.kafka.connect.runtime.WorkerTask.doStart(WorkerTask.java:188)\n\tat org.apache.kafka.connect.runtime.WorkerTask.doRun(WorkerTask.java:248)\n\tat org.apache.kafka.connect.runtime.WorkerTask.run(WorkerTask.java:304)\n\tat org.apache.kafka.connect.runtime.AbstractWorkerSourceTask.run(AbstractWorkerSourceTask.java:82)\n\tat org.apache.kafka.connect.runtime.isolation.Plugins.lambda$withClassLoader$7(Plugins.java:341)\n\tat java.base/java.util.concurrent.Executors$RunnableAdapter.call(Unknown Source)\n\tat java.base/java.util.concurrent.FutureTask.run(Unknown Source)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(Unknown Source)\n\tat java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(Unknown Source)\n\tat java.base/java.lang.Thread.run(Unknown Source)\n"     }   ],   "type": "source" }
 {code}

No idea.

sundongkim-dev:
Fixes debezium/dbz#75

## Description

A MySQL/MariaDB connector fails on startup during schema history recovery with:

```
java.lang.IllegalArgumentException: Cannot compare binlog filenames with different base names
	at io.debezium.connector.binlog.history.BinlogHistoryRecordComparator$BinlogFileName.compareTo(BinlogHistoryRecordComparator.java:232)
	at io.debezium.connector.binlog.history.BinlogHistoryRecordComparator.isPositionAtOrBefore(...)
	...
	at io.debezium.relational.history.AbstractSchemaHistory.recover(...)
```

This happens whenever a recorded binlog position and the desired (restart) position have different **base names** — e.g. after an RDS restore, a failover, or a `log_bin_basename` change (`mysql-bin.000010` vs `binlog.000002`). `BinlogFileName.compareTo` threw because the numeric extensions of two different base names are not directly comparable.

`isPositionAtOrBefore` already handles an analogous "coordinates are unrelated" case: when the two positions come from different servers it does not compare binlog coordinates. This change treats a differing base name the same way — instead of throwing, the recorded position is considered at-or-before the desired one so its DDL is applied during recovery.

Because connector offsets advance monotonically, a differing base name means the recorded position predates the base-name change (an older coordinate space), so applying it rebuilds the complete schema. This is the safe direction during recovery: skipping the DDL (the alternative) would leave the in-memory schema incomplete and break parsing of later events.

The `compareTo` contract is left unchanged, so genuinely unparseable filenames (no extension / non-numeric extension) still fail fast as before.

## Tests

- `BinlogSourceInfoTest#shouldNotFailWhenComparingPositionsWithDifferentBaseNames` (runs for MySQL and MariaDB): asserts that comparing positions with different, parseable base names no longer throws and is treated as at-or-before in both directions. This replaces the previous `shouldNotComparePositionsWithDifferentFilenameFormats`, which asserted the now-removed throwing behavior.
- The existing tests that assert throwing for unparseable filenames (no extension, non-numeric extension) are retained unchanged.

## PR Checklist

### DIFF
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/history/BinlogHistoryRecordComparator.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/history/BinlogHistoryRecordComparator.java
@@ -5,7 +5,12 @@
  */
 package io.debezium.connector.binlog.history;
 
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.function.Predicate;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import io.debezium.annotation.VisibleForTesting;
 import io.debezium.connector.binlog.BinlogOffsetContext;
@@ -22,8 +27,15 @@ import io.debezium.relational.history.HistoryRecordComparator;
  */
 public abstract class BinlogHistoryRecordComparator extends HistoryRecordComparator {
 
+    private static final Logger LOGGER = LoggerFactory.getLogger(BinlogHistoryRecordComparator.class);
+
     private final Predicate<String> gtidSourceFilter;
     private final GtidSetFactory gtidSetFactory;
+
+    // isPositionAtOrBefore() is invoked once per recorded history entry during recovery, so a base-name
+    // change would otherwise log for every differing record. Track the transitions already reported to
+    // keep the warning to one line per distinct base-name change.
+    private final Set<String> warnedBaseNameChanges = ConcurrentHashMap.newKeySet();
 
     public BinlogHistoryRecordComparator(Predicate<String> gtidSourceFilter, GtidSetFactory gtidSetFactory) {
         this.gtidSourceFilter = gtidSourceFilter;
@@ -115,6 +127,13 @@ public abstract class BinlogHistoryRecordComparator extends HistoryRecordCompara
             // than failing schema history recovery, treat the recorded position as at-or-before the desired
             // one so its DDL is applied and the in-memory schema is rebuilt completely. Skipping the DDL is
             // the unsafe direction: it would leave the schema incomplete and break parsing of later events.
+            final String change = recordedFileName.baseName + " -> " + desiredFileName.baseName;
+            if (warnedBaseNameChanges.add(change)) {
+                LOGGER.warn("Binlog base name changed during schema history recovery ({}); the recorded DDL is "
+                        + "applied because the numeric extensions are no longer comparable. This is expected after a "
+                        + "restore or log_bin_basename change, but if it results from switching back and forth between "
+                        + "primary and failover the recovered schema history may be incomplete.", change);
+            }
             return true;
         }
         final int fileNameCheck = recordedFileName.compareTo(desiredFileName);
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/history/BinlogHistoryRecordComparator.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/history/BinlogHistoryRecordComparator.java
@@ -109,6 +109,14 @@ public abstract class BinlogHistoryRecordComparator extends HistoryRecordCompara
         // Compare binlog file names
         final BinlogFileName recordedFileName = getBinlogFileName(recorded);
         final BinlogFileName desiredFileName = getBinlogFileName(desired);
+        if (!recordedFileName.baseName.equals(desiredFileName.baseName)) {
+            // The binlog base name changed (e.g. after a restore, failover, or a log_bin_basename change),
+            // so the numeric extensions belong to unrelated coordinate spaces and cannot be compared. Rather
+            // than failing schema history recovery, treat the recorded position as at-or-before the desired
+            // one so its DDL is applied and the in-memory schema is rebuilt completely. Skipping the DDL is
+            // the unsafe direction: it would leave the schema incomplete and break parsing of later events.
+            return true;
+        }
         final int fileNameCheck = recordedFileName.compareTo(desiredFileName);
         if (fileNameCheck != 0) {
             return fileNameCheck < 0;
--- a/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogSourceInfoTest.java
+++ b/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogSourceInfoTest.java
@@ -583,11 +583,16 @@ public abstract class BinlogSourceInfoTest<S extends BinlogSourceInfo, O extends
     }
 
     @Test
-    void shouldNotComparePositionsWithDifferentFilenameFormats() {
-        assertThrows(IllegalArgumentException.class, () -> {
-            Document history = positionWithoutGtids("mysql-bin.000001", 1, 0, 0);
-            assertThatDocument(history).isAtOrBefore(positionWithoutGtids("mysql-binlog-filename.000001", 1, 0, 0));
-        });
+    @FixFor("debezium/dbz#75")
+    void shouldNotFailWhenComparingPositionsWithDifferentBaseNames() {
+        // A changed binlog base name (e.g. after a restore, failover, or a log_bin_basename change) is a
+        // valid situation, not an error. The numeric extensions then belong to unrelated coordinate spaces,
+        // so instead of throwing the recorded position is treated as at-or-before the desired one and its
+        // DDL is applied during schema history recovery.
+        assertThatDocument(positionWithoutGtids("mysql-bin.000010", 100, 0, 0))
+                .isAtOrBefore(positionWithoutGtids("binlog.000002", 200, 0, 0));
+        assertThatDocument(positionWithoutGtids("binlog.000002", 200, 0, 0))
+                .isAtOrBefore(positionWithoutGtids("mysql-bin.000010", 100, 0, 0));
     }
 
     @Test


Naros (commented):
@vjuranek, are there any other scenarios here that you can think of that could be handled incorrectly in this use case?  I understand the premise here, but I'm curious if this could be problematic when switching from primary to failover and then back to primary.

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/history/BinlogHistoryRecordComparator.java
```diff
@@ -109,6 +109,16 @@ else if (recordedGtid != null) {
...
         final BinlogFileName desiredFileName = getBinlogFileName(desired);
+        if (!recordedFileName.hasSameBaseNameAs(desiredFileName)) {
+            // The binlog base name changed (e.g. after a restore, failover, or a log_bin_basename change),
+            // so the numeric extensions belong to unrelated coordinate spaces and cannot be compared. Rather
+            // than failing schema history recovery, treat the recorded position as at-or-before the desired
+            // one so its DDL is applied. Because offsets advance monotonically, a differing base name means
+            // the recorded position predates the change, so applying it rebuilds the complete schema. This
+            // is the safe direction during recovery: skipping the DDL would leave the in-memory schema
+            // incomplete and break parsing of later events.
+            return true;
```
vjuranek, line 137: Why return always `true` and not compare timestamps as we do few lines above in case of different servers?

sundongkim-dev: Good question — I looked hard at reusing the `getTimestamp(recorded) <= getTimestamp(desired)` comparison from the different-servers branch above, and the short version is that **implementing it here would be behaviorally identical to `return true` today**, because that timestamp comparison is itself inert during schema-history recovery.

**Why the timestamp comparison can't discriminate here.** During recovery the comparator only sees `record.position()` — the serialized offset map. `getTimestamp()` reads `BinlogSourceInfo.TIMESTAMP_KEY`, which resolves to `AbstractSourceInfo.TIMESTAMP_KEY = "ts_ms"`, but the offset map only ever carries the timestamp under `BinlogOffsetContext.TIMESTAMP_KEY = "ts_sec"` (`offsetUsingPosition`). The record *does* store a real millisecond timestamp, but under `ts_ms` at the record's doc root (`HistoryRecord`), which is structurally outside the `position()` document this comparator inspects. So `getTimestamp` falls back to its `0` default for both sides, and the different-servers branch above is really `0 <= 0 == true` regardless of the actual timestamps. Copying that pattern into the base-name branch would just be an obfuscated `return true`. (This `ts_ms`/`ts_sec` mismatch looks like a separate latent bug from the DBZ-7693 refactor; I've kept it out of this PR's scope but can file it separately.)

**On the choice of `true`.** What we'd really want is "apply DDL that predates the restart point, skip DDL that's after it." But once the base name differs the two extensions are unrelated coordinate spaces, and — per the above — there's no readable timestamp in the position document to order them by, so that discrimination isn't actually available here. Between the two reachable options, applying is the safer default during recovery: skipping DDL leaves the in-memory schema incomplete and breaks parsing of later events, which is the concrete dbz#75 failure mode. I don't want to overstate it as "always harmless" — a pathological case (same `server_id`, no GTID, a base-name change mid-stream, and post-change records flushed to history before the data offset commits) could let a later `DROP`/`ALTER` be replayed — but that corner isn't reachable through the timestamp path either, and it's outside what dbz#75 reports (RDS restore / `log_bin_basename` change, where the differing-base records genuinely predate the restart).

I also dropped the earlier "offsets advance monotonically, so the recorded position predates the change" line from the comment and commit message, since that doesn't hold under base-name reuse cycles and isn't what the fix relies on.

Naros: Before this PR, if the failover uses a different stem, recovery becomes impossible and its a hard error. 

But what happens with this PR when you switch from primary to failover, then back to primary. Doesn't this eventually lead to an incomplete schema history on recovery? To me this is a silent error rather than a hard, loud fault, which makes far more sense.

sundongkim-dev:
Thanks @Naros — let me answer the failover question and the silent-vs-loud point together, since they're related.

**Where the new branch actually sits.** The base-name check is only reached after two earlier sections of `isPositionAtOrBefore` have already returned:

- If either position carries a GTID, the comparison is resolved in the GTID section above — the base-name branch is never reached. Since GTID-based failover is the modern default, primary → failover → primary is handled entirely by `GtidSet` containment.
- If both lack GTIDs but `server_id` differs, the existing different-servers branch returns first. A genuine failover to a distinct server takes this path.

So the branch only fires when both positions lack GTIDs, share the same `server_id`, and differ in base name — a same-server base-name change (`log_bin_basename` change, `RESET MASTER`, a restore), which is exactly the dbz#75 case.

**Why this can't produce an incomplete schema history.** `recover()` uses the comparator purely as an include/skip gate while streaming history records in stored order — it never re-sorts. Pre-PR, a differing base name didn't skip the record; it threw `IllegalArgumentException` and aborted recovery entirely, so the connector failed to start. This PR converts that throwing path into "include" and changes nothing else — no record that was applied before can become skipped, so relative to `main` the set of applied DDL can only grow. Whatever primary → failover → primary sequence we construct, the outcome is over-inclusion at worst, never a gap.

**The honest limit is the opposite direction.** If the connector resumes at an offset *older* than some differing-base records (offsets rewound or restored), their DDL is applied early and the in-memory schema runs ahead of the stream position. That's usually recoverable — a complete schema replays fine — but since `DROP`/`ALTER` are subtractive I don't claim "always harmless", only that including is the least-bad option actually available: the position document carries no readable timestamp to discriminate with (the `ts_ms`/`ts_sec` mismatch discussed above), and skipping produces exactly the incomplete schema that breaks parsing of later events — the failure dbz#75 reports.

Two caveats for completeness: a shared or defaulted `server_id` (0) can route a cross-node case into this branch — same over-inclusion outcome, not a crash; and base-name toggling A→B→A isn't fully solved, because same-base records with colliding extensions are still compared numerically — a pre-existing coordinate-reuse ambiguity that only GTIDs disambiguate, untouched here.

**On silent vs loud.** Agreed — the mismatch shouldn't pass without a trace. But what the hard error was loudly reporting was an unrecoverable state whose only escape hatches are operational (rebuild the history topic or re-snapshot), even in the common case where the differing-stem records genuinely predate the restart point and recovery is perfectly possible. My proposal: keep recovery working, but WARN on the first base-name mismatch seen during recovery (once, not per record — recovery can stream thousands of old-stem records), naming both filenames so a misconfigured cluster stays visible in the logs. Happy to push that here if it addresses your concern.

I can also add a primary → failover → primary test (same-server base-name toggle, with and without GTIDs) if useful. @vjuranek does this match your read?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/history/BinlogHistoryRecordComparator.java
vjuranek: @Naros I agree this is a concern. Would logging a clear warning here be a way forward (not failing, but at the same time informing the user about possible issue)?

sundongkim-dev: Sounds good — a warning is the right middle ground: recovery keeps working for the legitimate base-name change while the assumption becomes visible, which addresses @Naros's silent-vs-loud point.

One note: `isPositionAtOrBefore` runs once per recovered record, so a bare `warn()` would fire for every differing record. I'll log it only once per base-name transition (`recordedBase` → `desiredBase`), as informational, with a note that history may be incomplete after a failover cycle. Pushing shortly.

sundongkim-dev:
@Naros @vjuranek gentle ping on this one — I posted a reply on Jul 15 walking through where the base-name branch sits relative to the earlier GTID/position checks, and the silent-vs-loud handling. Does that address the concern, or is there anything else you'd like adjusted here?

Naros (approved):
Hi @sundongkim-dev, I believe your explanations address all my concerns. I had genuinely thought about whether we should consider some sort of toggle for this to hard-fail or gracefully recover as the PR is done, but AFAIK I can't seem to find any use case where the hard-fail isn't already guarded by other hard-fail checks that come prior.

Let's wait for @vjuranek on Monday to see if he has any other input.

vjuranek (approved):
LGTM

vjuranek:
> Let's wait for @vjuranek on Monday to see if he has any other input.

I was waiting for your input, if you have any objection :-) , but forgot to approve the PR.

vjuranek:

vjuranek:
And of course sorry for the delay on our side @sundongkim-dev !

PR #7648, opened 2026-07-11
debezium/dbz#2100 Guard pgvector sink DDL against unresolved dimension

sundongkim-dev:
Fixes debezium/dbz#2100, case 12.

## Description

pgvector types (`vector` / `halfvec` / `sparsevec`) are provided by an extension and get **dynamically assigned OIDs**, so they are not in the PostgreSQL JDBC driver's hard-coded type table. The driver therefore cannot resolve their dimension, and the propagated `__debezium.source.column.length` arrives as `Integer.MAX_VALUE` (`2147483647`) on both the snapshot and streaming paths. The JDBC sink rendered that verbatim, so automatic schema evolution attempted:

```sql
CREATE TABLE vecaudit_sink (id integer NOT NULL, vec vector(2147483647) NULL,
  half halfvec(2147483647) NULL, sparse sparsevec(2147483647) NULL, PRIMARY KEY(id))
```

which PostgreSQL rejects with `ERROR: dimensions for type vector cannot exceed 16000`.

This guards `getTypeName` in the three PostgreSQL vector handlers: when the propagated size is `Integer.MAX_VALUE`, emit an unqualified `vector` / `halfvec` / `sparsevec` instead of a qualified type. This mirrors the existing `BitType` guard, which emits `bit varying` for the same `Integer.MAX_VALUE` sentinel. An unqualified pgvector column accepts a vector of any dimension, so values still round-trip; only the dimension constraint (which the driver does not expose to us) is not replicated.

Recovering the real dimension on the source side is tracked separately as a longer-term follow-up, as discussed on the issue.

Tests:
- `DoubleVectorTypeTest` / `FloatVectorTypeTest` / `SparseDoubleVectorTypeTest` (unit) — each asserts a qualified type for a propagated dimension, the unqualified fallback when no size is present, and the unqualified fallback when the propagated size is `Integer.MAX_VALUE` (the regression pinned to this issue).
- Verified end-to-end (PG `vector(3)`/`halfvec(3)`/`sparsevec(25)` → JDBC sink): the sink now auto-creates unqualified `vector`/`halfvec`/`sparsevec` columns and round-trips the values, where `main` fails with the error above.

## PR Checklist

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/DoubleVectorType.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/DoubleVectorType.java
@@ -28,7 +28,12 @@ public class DoubleVectorType extends AbstractDoubleVectorType {
 
     @Override
     public String getTypeName(Schema schema, boolean isKey) {
-        final Optional<String> size = getSourceColumnSize(schema);
+        // pgvector types get dynamically assigned OIDs, so the PostgreSQL driver cannot resolve their
+        // dimension and the propagated column length arrives as Integer.MAX_VALUE. Emit an unqualified
+        // vector in that case rather than an invalid vector(2147483647). This mirrors BitType, which
+        // falls back to "bit varying" for the same Integer.MAX_VALUE sentinel.
+        final Optional<String> size = getSourceColumnSize(schema)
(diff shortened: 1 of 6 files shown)


Naros (approved):
LGTM, waits on

Naros:
Applied thanks @sundongkim-dev :tada:

PR #7650, opened 2026-07-11
debezium/dbz#798 Skip ad-hoc snapshot signals with non-string data collections

### LINKED ISSUE #798: Null pointer exception when starting ad-hoc snapshot with wrong content [DBZ-6352]
Migrated from [DBZ-6352](https://issues.redhat.com/browse/DBZ-6352)

Debezium PostgreSQL 2.0.0.Final, but the code has not changed

{code:json}
{
    "incremental.snapshot.chunk.size": "5120",
    "slot.name": "debezium_cdc",
    "tasks.max": "1",
    "publication.name": "debezium_cdc",
    "value.converter.enhanced.avro.schema.support": "true",
    "value.converter.basic.auth.credentials.source": "USER_INFO",
    "database.sslmode": "require",
    "topic.prefix": "registry",
    "signal.data.collection": "public.debezium_signals",
    "errors.log.enable": "true",
    "key.converter": "io.confluent.connect.avro.AvroConverter",
    "database.dbname": "snykregistry",
    "database.user": "username",
    "database.server.name": "registry",
    "key.converter.enhanced.avro.schema.support": "true",
    "database.port": "5432",
    "plugin.name": "pgoutput",
    "key.converter.basic.auth.user.info": "userinfo",
    "value.converter.schema.registry.url": "schema-registry-url",
    "column.exclude.list": "public.a_table.a_column",
    "value.converter.basic.auth.user.info": "user-info",
    "database.hostname": "db-hostname",
    "database.password": "password",
    "name": "debezium-cdc",
    "errors.tolerance": "all",
    "skipped.operations": "t",
    "table.include.list": "a list of tables",
    "key.converter.schema.registry.url": "url",
    "key.converter.basic.auth.credentials.source": "USER_INFO",
    "snapshot.mode": "never"
}
{code}

h2. What is the captured database version and mode of depoyment?

GCP PostgreSQL 14 (really doesn't matter)

An invalid `data` column in the signals table will trigger an error in the logs, but WAL log processing will continue.

When inserting an invalid `data` column in the signals table, Debezium will fail and completely stop processing messages. Restarting the connector won't help.

The data column had a nested array (this was an error on our side). For example:
```{"data-collections":[["tablea","tableb"]]}```

Can't try easily, but code is the same https://github.com/debezium/debezium/blob/db4c2fd91844f3401bc34d62a8033892d5b52987/debezium-core/src/main/java/io/debezium/pipeline/signal/ExecuteSnapshot.java#L71-L73

{noformat}
	at io.debezium.pipeline.ErrorHandler.setProducerThrowable(ErrorHandler.java:53)
	at io.debezium.connector.postgresql.PostgresStreamingChangeEventSource.execute(PostgresStreamingChangeEventSource.java:180)
	at io.debezium.connector.postgresql.PostgresStreamingChangeEventSource.execute(PostgresStreamingChangeEventSource.java:41)
	at io.debezium.pipeline.ChangeEventSourceCoordinator.streamEvents(ChangeEventSourceCoordinator.java:174)
	at io.debezium.pipeline.ChangeEventSourceCoordinator.executeChangeEventSources(ChangeEventSourceCoordinator.java:141)
	at io.debezium.pipeline.ChangeEventSourceCoordinator.lambda$start$0(ChangeEventSourceCoordinator.java:109)
	at java.base/java.util.concurrent.Executors$RunnableAdapter.call(Executors.java:515)
	at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)
	at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)
	at java.base/java.lang.Thread.run(Thread.java:829)
Caused by: org.apache.kafka.connect.errors.ConnectException: Error while processing event at offset {transaction_id=null, lsn_proc=40020183155160, lsn_commit=40020183155160, lsn=40020183155160, txId=714621851, ts_usec=1681843474910582}
	at io.debezium.pipeline.EventDispatcher.dispatchDataChangeEvent(EventDispatcher.java:261)
	at io.debezium.connector.postgresql.PostgresStreamingChangeEventSource.lambda$processMessages$0(PostgresStreamingChangeEventSource.java:275)
	at io.debezium.connector.postgresql.connection.pgoutput.PgOutputMessageDecoder.decodeInsert(PgOutputMessageDecoder.java:413)
	at io.debezium.connector.postgresql.connection.pgoutput.PgOutputMessageDecoder.processNotEmptyMessage(PgOutputMessageDecoder.java:191)
	at io.debezium.connector.postgresql.connection.AbstractMessageDecoder.processMessage(AbstractMessageDecoder.java:41)
	at io.debezium.connector.postgresql.connection.PostgresReplicationConnection$1.deserializeMessages(PostgresReplicationConnection.java:504)
	at io.debezium.connector.postgresql.connection.PostgresReplicationConnection$1.readPending(PostgresReplicationConnection.java:496)
	at io.debezium.connector.postgresql.PostgresStreamingChangeEventSource.processMessages(PostgresStreamingChangeEventSource.java:215)
	at io.debezium.connector.postgresql.PostgresStreamingChangeEventSource.execute(PostgresStreamingChangeEventSource.java:177)
	... 9 more
Caused by: java.lang.NullPointerException
	at io.debezium.pipeline.signal.ExecuteSnapshot.lambda$getDataCollections$0(ExecuteSnapshot.java:70)
	at java.base/java.util.stream.ReferencePipeline$3$1.accept(ReferencePipeline.java:195)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1655)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:484)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:474)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:913)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.collect(ReferencePipeline.java:578)
	at io.debezium.pipeline.signal.ExecuteSnapshot.getDataCollections(ExecuteSnapshot.java:71)
	at io.debezium.pipeline.signal.ExecuteSnapshot.arrived(ExecuteSnapshot.java:44)
	at io.debezium.pipeline.signal.Signal.process(Signal.java:138)
	at io.debezium.pipeline.signal.Signal.process(Signal.java:176)
	at io.debezium.pipeline.EventDispatcher$2.changeRecord(EventDispatcher.java:235)
	at io.debezium.relational.RelationalChangeRecordEmitter.emitCreateRecord(RelationalChangeRecordEmitter.java:79)
	at io.debezium.relational.RelationalChangeRecordEmitter.emitChangeRecords(RelationalChangeRecordEmitter.java:47)
	at io.debezium.connector.postgresql.PostgresChangeRecordEmitter.emitChangeRecords(PostgresChangeRecordEmitter.java:92)
	at io.debezium.pipeline.EventDispatcher.dispatchDataChangeEvent(EventDispatcher.java:224)
{noformat}

Create connector that supports signals, insert message like 

{noformat}
id='09ac1a68-df2e-41e3-8757-28c266b2632a', type='execute-snapshot', data='{"data-collections":[["tablea"]]}'
{noformat}

Any help to get the connector running again would be appreciated

### PR DESCRIPTION
**Pull request by sundongkim-dev**
Fixes debezium/dbz#798

## Description

An `execute-snapshot` (or `stop-snapshot`) signal whose `data-collections` array contains a non-string element — for example a nested array like `{"data-collections":[["tablea"]]}` instead of a flat `["tablea"]` — caused a `NullPointerException`.

`ExecuteSnapshot.getDataCollections` (and the identical code in `StopSnapshot.getDataCollections`) does:

```java
dataCollectionsArray.streamValues()
        .map(v -> v.asString().trim())
```

`Value.asString()` returns `null` for a non-string value, so `.trim()` throws NPE. Because the signal is processed inside the change-event producer, the exception propagates through `ErrorHandler` and stops the connector. On restart the same signal row is re-read from the log and fails again, so the connector is permanently unable to resume — the only workaround today is to remove `signal.data.collection` from the config and restart.

This matches the reporter's expectation that an invalid `data` value should be logged and skipped while log processing continues.

## Fix

When any `data-collections` element is not a string, log a warning and return `null`. This reuses the existing invalid-signal path (missing/empty `data-collections` already returns `null`), so the malformed signal is skipped:
- `ExecuteSnapshot.arrived` already treats a `null` result as "invalid signal" and returns `false` (no snapshot started).
- `StopSnapshot`: note that a `null`/empty `data-collections` means "stop the entire in-progress snapshot" per its contract. Routing a malformed signal through this path therefore escalates to stop-all rather than stopping specific collections. This is intentional and strictly safer than the previous behavior: over-stopping a snapshot is fully recoverable (just re-trigger it), whereas the prior NPE wedged the connector permanently. A warning is logged so the operator can see the signal was malformed.

## Tests

- New `SnapshotSignalDataCollectionsTest` covering both `ExecuteSnapshot` and `StopSnapshot`:
  - nested-array `data-collections` → returns `null` (previously NPE) and logs the warning
  - a non-string scalar element → returns `null`
  - valid string arrays → parsed as before
  - missing / empty `data-collections` → `null` (unchanged)
- Verified as a regression test: reverting the guard reproduces the exact `NullPointerException: Cannot invoke "String.trim()" because ... Value.asString() is null` on the malformed cases while the valid cases keep passing.

## PR Checklist

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/pipeline/signal/SignalProcessor.java
+++ b/debezium-connector-common/src/main/java/io/debezium/pipeline/signal/SignalProcessor.java
@@ -267,6 +267,10 @@ public class SignalProcessor<P extends Partition, O extends OffsetContext> {
 
         // Extract database name from first data collection
         for (Array.Entry entry : dataCollectionsArray) {
+            if (!entry.getValue().isString()) {
+                LOGGER.debug("Skipping non-string data-collections element while resolving the signal partition");
+                continue;
+            }
             String dataCollection = entry.getValue().asString().trim();
 
             String[] parts = dataCollection.split(POINT_REGEX);
--- a/debezium-connector-common/src/main/java/io/debezium/pipeline/signal/actions/AbstractSnapshotSignal.java
+++ b/debezium-connector-common/src/main/java/io/debezium/pipeline/signal/actions/AbstractSnapshotSignal.java
@@ -5,6 +5,7 @@
  */
 package io.debezium.pipeline.signal.actions;
 
+import java.util.List;
 import java.util.Optional;
 
 import org.slf4j.Logger;
@@ -31,6 +32,16 @@ public abstract class AbstractSnapshotSignal<P extends Partition> implements Sig
     public enum SnapshotType {
         INCREMENTAL,
         BLOCKING
+    }
+
+    /**
+     * Result of parsing the {@code data-collections} field of a snapshot signal.
+     *
+     * @param valid whether the signal can be acted upon; {@code false} when the field is present but malformed
+     *            (for example a non-string element), so the signal must be skipped
+     * @param collections the parsed collections, or {@code null} when the field is missing or empty
+     */
+    public record DataCollections(boolean valid, List<String> collections) {
     }
 
     public static SnapshotType getSnapshotType(Document data) {
--- a/debezium-connector-common/src/main/java/io/debezium/pipeline/signal/actions/snapshotting/ExecuteSnapshot.java
+++ b/debezium-connector-common/src/main/java/io/debezium/pipeline/signal/actions/snapshotting/ExecuteSnapshot.java
@@ -52,10 +52,11 @@ public class ExecuteSnapshot<P extends Partition> extends AbstractSnapshotSignal
     @Override
     public boolean arrived(SignalPayload<P> signalPayload) throws InterruptedException {
 
-        final List<String> dataCollections = getDataCollections(signalPayload.data);
-        if (dataCollections == null) {
+        final DataCollections dataCollectionsResult = getDataCollections(signalPayload.data);
+        if (!dataCollectionsResult.valid()) {
             return false;
         }
+        final List<String> dataCollections = dataCollectionsResult.collections();
         SnapshotType type = getSnapshotType(signalPayload.data);
 
         List<AdditionalCondition> additionalConditions = getAdditionalConditions(signalPayload.data, type);
@@ -100,18 +101,24 @@ public class ExecuteSnapshot<P extends Partition> extends AbstractSnapshotSignal
                 .build();
     }
 
-    public static List<String> getDataCollections(Document data) {
+    public static DataCollections getDataCollections(Document data) {
 
         final Array dataCollectionsArray = data.getArray(FIELD_DATA_COLLECTIONS);
         if (dataCollectionsArray == null || dataCollectionsArray.isEmpty()) {
             LOGGER.warn(
                     "Execute snapshot signal '{}' has arrived but the requested field '{}' is missing from data or is empty",
                     data, FIELD_DATA_COLLECTIONS);
-            return null;
+            return new DataCollections(false, null);
         }
-        return dataCollectionsArray.streamValues()
+        if (dataCollectionsArray.streamValues().anyMatch(v -> !v.isString())) {
+            LOGGER.warn(
+                    "Execute snapshot signal '{}' has arrived but the requested field '{}' contains a non-string data collection",
+                    data, FIELD_DATA_COLLECTIONS);
+            return new DataCollections(false, null);
+        }
+        return new DataCollections(true, dataCollectionsArray.streamValues()
                 .map(v -> v.asString().trim())
-                .collect(Collectors.toList());
+                .collect(Collectors.toList()));
     }
 
     public static Optional<String> getSurrogateKey(Document data) {
--- a/debezium-connector-common/src/main/java/io/debezium/pipeline/signal/actions/snapshotting/StopSnapshot.java
+++ b/debezium-connector-common/src/main/java/io/debezium/pipeline/signal/actions/snapshotting/StopSnapshot.java
@@ -42,7 +42,11 @@ public class StopSnapshot<P extends Partition> extends AbstractSnapshotSignal<P>
 
     @Override
     public boolean arrived(SignalPayload<P> signalPayload) throws InterruptedException {
-        final List<String> dataCollections = getDataCollections(signalPayload.data);
+        final DataCollections dataCollectionsResult = getDataCollections(signalPayload.data);
+        if (!dataCollectionsResult.valid()) {
+            return false;
+        }
+        final List<String> dataCollections = dataCollectionsResult.collections();
         final SnapshotType type = getSnapshotType(signalPayload.data);
 
         LOGGER.info("Requested stop of snapshot '{}' for data collections '{}'", type, dataCollections);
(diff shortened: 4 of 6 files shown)


Naros (commented):
Hi @sundongkim-dev, I don't necessarily agree with the premise that if there is a malformed stop signal that we treat it as stop all rather than logged error.

I genuinely think we should either consider some validate method call and/or use a record result type that contains the collection plus a valid flag so that inside the `arrived` method, we can check whether the collection data is valid before proceeding. When invalid we can log a warn/error and return `false`.

WDYT?

sundongkim-dev:
Thanks @Naros, that makes sense — I've reworked it along the lines you suggested.

`getDataCollections` now returns a `DataCollections(boolean valid, List<String> collections)` record, and `arrived()` checks `valid()` before acting. When the field is malformed (a non-string element) it logs a warning and returns `false` without touching the snapshot. Crucially this keeps the malformed case distinct from a missing/empty `data-collections`, which for `stop-snapshot` remains a valid request to stop the whole in-progress snapshot — so a malformed stop signal is now skipped rather than escalating to stop-all. There's a new `SnapshotSignalArrivedTest` that verifies (via Mockito) `requestStopSnapshot` is *never* invoked for a malformed stop signal, and *is* invoked for the missing/empty case.

While digging in I also found that the connector isn't actually stopped by this today — `SignalProcessor#processSignal` has a catch-all that logs and continues, so the real symptom is a stack trace re-logged on every poll rather than a hard stop. I've corrected the commit message to reflect that. I also noticed the same `asString().trim()` pattern in `SignalProcessor#extractPartitionFromData` (partition resolution for external-channel signals on multi-partition tasks), which would hit the same NPE before `arrived()` is even reached, so I guarded that too.

Let me know if you'd prefer the partition-resolution fix split into its own PR.

Naros:
Could you take a look at the Apicurio test failures with OpenLineage @sundongkim-dev ?
I wonder if these are related.

sundongkim-dev:
Hi @Naros — I looked into it, and the OpenLineage/Apicurio failures aren't related to this PR:

- The **same failures reproduce on `main`** — today's push build (`Build Debezium (Push)` run 29393514172) fails MongoDB 7.0/8.0 with the identical `OpenLineageIT` assertions (`assertEventContainsExpectedData`) plus `SinkConnector*IT` failing to pull `quay.io/debezium/connect:3.7` (`Tests run: 317, Failures: 3, Errors: 2`). So these are pre-existing on trunk, not introduced here.
- They're also **flaky across reruns of this PR**: `Apicurio / PostgreSQL (18)` failed on the previous run and passed on the latest one with no code change, while `PostgreSQL 18` flipped the other way on an unrelated `replication slot "debezium" is active` teardown race.
- Every test that actually exercises the changed signal/snapshot code passes in all runs (`SignalsIT`, `ReadOnlyIncrementalSnapshotIT`, MongoDB `IncrementalSnapshotIT`, `PostgresConnectorIT`, and the new unit tests).

The OpenLineage assertion is on connector-config content in the lineage START event, which this change doesn't touch. Happy to help track the OpenLineage flake separately if useful.

Naros (approved):

Naros:

PR #7651, opened 2026-07-11
debezium/dbz#683 Use unqualified type name for user-defined types in snapshot

### LINKED ISSUE #683: Inconsistent column type naming between streaming and incremental snapshot [DBZ-5571]
Migrated from [DBZ-5571](https://issues.redhat.com/browse/DBZ-5571)

h1. Bug report

PostgresConnector, v1.9.2.Final

 
{code:java}
{
    "database.hostname": "{{database_hostname}}",
    "database.port": "5432",
    "database.dbname": "db1",
    "database.user": "postgres",
    "database.password": "{{database_password}}",
    "database.server.name": "db1",
    "slot.name": "cdc_debezium_test",
    "plugin.name": "pgoutput",
    "publication.name": "cdc_dbz",
    "table.include.list": "db1.table1,db1.table2,db1.debezium_signals",
    "signal.data.collection": "db1.debezium_signals",
    "heartbeat.interval.ms": "3000",
    "heartbeat.action.query": "INSERT INTO db1.debezium_heartbeat (id, ts) VALUES ('cdc', now()) ON CONFLICT (id) DO UPDATE SET ts = now();",
    "snapshot.mode": "never",
    "key.converter": "org.apache.kafka.connect.json.JsonConverter",
    "key.converter.schemas.enable": "false",
    "value.converter.schema.registry.url": "http://host.docker.internal:8081",
    "topic.creation.default.replication.factor": 1,
    "topic.creation.default.partitions": 1,
    "topic.creation.default.cleanup.policy": "compact",
    "transforms": "route",
    "transforms.route.type": "org.apache.kafka.connect.transforms.RegexRouter",
    "transforms.route.regex": ".*",
    "transforms.route.replacement": "cdc_dbz.$0.v1",
    "converters": "financialAsset",
    "financialAsset.column.type.name": "financial_asset",
    "financialAsset.schema.name": "brex.debezium.extensions.data.FinancialAsset",
    "financialAsset.type": "brex.debezium.extensions.converters.FinancialAssetConverter",
    "include.unknown.datatypes": true
}{code}
 
h2. What is the captured database version and mode of depoyment?

Database version: PostgreSQL 13

Mode of deployment: Amazon EKS

We expect to have consistent column type naming between streaming change events and incremental snapshot change events.

When we test with user user-defined data types (UDT), we are seeing inconsistent column type names between streaming events and incremental snapshot events.

For streaming events (insert, update and delete), the column type name is {_}<type_name>{_}. But for incremental snapshotting, we are seeing fully qualified name like {_}"<schema_name>"."<type_name>"{_}.

 

Our investigation:

For incremental snapshot, when the connector reads the table schema via jdbc driver, it returns either fully qualified name or the name without schema name for user defined types. This depends on the search path. If the db user search path includes the schema name of the UDT, then it returns the short type name ([code|https://github.com/pgjdbc/pgjdbc/blob/9effea487dca529dc5cb0e71600a8c6509819c56/pgjdbc/src/main/java/org/postgresql/jdbc/TypeInfoCache.java#L518]).

For streaming change events, seems the Debezium pgoutput message decoder drops the schema name of the UDT fully qualified name from the db query results and only uses the short type name as column typeName ([code|https://github.com/debezium/debezium/blob/ce9ddb119eaebef4a78882146f20a3d3ebd7c30f/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java#L203]).

We did not test with the latest Debezium version, but we suspect to see the same issue.

The custom data type is "financial_asset" in "expenses_v2" schema and we got the following DEBUG logs during incremental snapshotting:
{code:java}
cdc-kafka-connect-connect-5f7f88b75-x4b6r cdc-kafka-connect-connect {"@timestamp":"2022-08-31T00:26:54.490Z","source_host":"cdc-kafka-connect-connect-5f7f88b75-x4b6r","file":"TableSchemaBuilder.java","method":"addField","level":"DEBUG","line_number":"418","thread_name":"debezium-postgresconnector-expenses_v2-change-event-source-coordinator","@version":1,"logger_name":"io.debezium.relational.TableSchemaBuilder","message":"- field 'amount' (BYTES) from column amount \"expenses_v2\".\"financial_asset\"(2147483647, 0) DEFAULT VALUE NULL","class":"io.debezium.relational.TableSchemaBuilder","mdc":{"dbz.connectorContext":"streaming","dbz.connectorType":"Postgres","connector.context":"[expenses-v2-debezium-connector-v2|task-0] ","dbz.taskId":"0","dbz.connectorName":"expenses_v2"}}
cdc-kafka-connect-connect-5f7f88b75-x4b6r cdc-kafka-connect-connect {"@timestamp":"2022-08-31T00:26:54.491Z","source_host":"cdc-kafka-connect-connect-5f7f88b75-x4b6r","file":"TableSchemaBuilder.java","method":"addField","level":"DEBUG","line_number":"418","thread_name":"debezium-postgresconnector-expenses_v2-change-event-source-coordinator","@version":1,"logger_name":"io.debezium.relational.TableSchemaBuilder","message":"- field 'original_amount' (BYTES) from column original_amount \"expenses_v2\".\"financial_asset\"(2147483647, 0) DEFAULT VALUE NULL","class":"io.debezium.relational.TableSchemaBuilder","mdc":{"dbz.connectorContext":"streaming","dbz.connectorType":"Postgres","connector.context":"[expenses-v2-debezium-connector-v2|task-0] ","dbz.taskId":"0","dbz.connectorName":"expenses_v2"}}
cdc-kafka-connect-connect-5f7f88b75-x4b6r cdc-kafka-connect-connect {"@timestamp":"2022-08-31T00:26:54.491Z","source_host":"cdc-kafka-connect-connect-5f7f88b75-x4b6r","file":"TableSchemaBuilder.java","method":"addField","level":"DEBUG","line_number":"418","thread_name":"debezium-postgresconnector-expenses_v2-change-event-source-coordinator","@version":1,"logger_name":"io.debezium.relational.TableSchemaBuilder","message":"- field 'billing_amount' (BYTES) from column billing_amount \"expenses_v2\".\"financial_asset\"(2147483647, 0) DEFAULT VALUE NULL","class":"io.debezium.relational.TableSchemaBuilder","mdc":{"dbz.connectorContext":"streaming","dbz.connectorType":"Postgres","connector.context":"[expenses-v2-debezium-connector-v2|task-0] ","dbz.taskId":"0","dbz.connectorName":"expenses_v2"}}
cdc-kafka-connect-connect-5f7f88b75-x4b6r cdc-kafka-connect-connect {"@timestamp":"2022-08-31T00:26:54.491Z","source_host":"cdc-kafka-connect-connect-5f7f88b75-x4b6r","file":"TableSchemaBuilder.java","method":"addField","level":"DEBUG","line_number":"418","thread_name":"debezium-postgresconnector-expenses_v2-change-event-source-coordinator","@version":1,"logger_name":"io.debezium.relational.TableSchemaBuilder","message":"- field 'budget_amount' (BYTES) from column budget_amount \"expenses_v2\".\"financial_asset\"(2147483647, 0) DEFAULT VALUE NULL","class":"io.debezium.relational.TableSchemaBuilder","mdc":{"dbz.connectorContext":"streaming","dbz.connectorType":"Postgres","connector.context":"[expenses-v2-debezium-connector-v2|task-0] ","dbz.taskId":"0","dbz.connectorName":"expenses_v2"}}{code}

We use a test database "db1" to reproduce the issue.

database setup:
{code:java}
-- connect with Postgres admin user
CREATE SCHEMA db1;
CREATE TABLE IF NOT EXISTS db1.debezium_heartbeat (id VARCHAR(255) NOT NULL PRIMARY KEY, ts TIMESTAMP NOT NULL);
CREATE TABLE IF NOT EXISTS db1.debezium_signals (id VARCHAR(42) PRIMARY KEY, type VARCHAR(32) NOT NULL, data VARCHAR(2048) NULL);

CREATE PUBLICATION cdc_dbz FOR ALL TABLES;
CREATE TYPE db1.financial_asset AS (quantity NUMERIC, instrument_code TEXT);
CREATE TABLE db1.table1 (id VARCHAR PRIMARY KEY, amount db1.financial_asset);{code}
create connector (we implemented our custom converter _"brex.debezium.extensions.converters.FinancialAssetConverter"_ and added it to the connect image):
{code:java}
# use Postgres admin user for the connector
curl --location --request POST 'localhost:8083/connectors' \
--header 'Accept: application/json' \
--header 'Content-Type: application/json' \
--data-raw '{
  "name":"cdc_connector_fa1",
  "config":{
    "database.hostname": "host.docker.internal",
    "database.port": "5432",
    "database.dbname": "db1",
    "database.user": "postgres",
    "database.password": <database_password>,
    "database.server.name": "db1",
    "slot.name": "cdc_debezium_test",
    "plugin.name": "pgoutput",
    "publication.name": "cdc_dbz",
    "table.include.list": "db1.table1,db1.table2,db1.debezium_signals",
    "signal.data.collection": "db1.debezium_signals",
    "heartbeat.interval.ms": "3000",
    "heartbeat.action.query": "INSERT INTO db1.debezium_heartbeat (id, ts) VALUES ('\''cdc'\'', now()) ON CONFLICT (id) DO UPDATE SET ts = now();",
    "snapshot.mode": "never",
    "key.converter": "org.apache.kafka.connect.json.JsonConverter",
    "key.converter.schemas.enable": "false",
    "value.converter.schema.registry.url": "http://host.docker.internal:8081",
    "topic.creation.default.replication.factor": 1,
    "topic.creation.default.partitions": 1,
    "topic.creation.default.cleanup.policy": "compact",
    "transforms": "route",
    "transforms.route.type": "org.apache.kafka.connect.transforms.RegexRouter",
    "transforms.route.regex": ".*",
    "transforms.route.replacement": "cdc_dbz.$0.v1",
    "converters": "financialAsset",
    "financialAsset.column.type.name": "financial_asset",
    "financialAsset.schema.name": "brex.debezium.extensions.data.FinancialAsset",
    "financialAsset.type": "brex.debezium.extensions.converters.FinancialAssetConverter",
    "include.unknown.datatypes": true
  }
}' {code}
create streaming events and incremental snapshotting:
{code:java}
-- connect with Postgres admin user
-- create insert events
INSERT INTO db1.table1 (id, amount) SELECT '01', '(1.5,usd)';
-- start incremental snapshot
INSERT INTO db1.debezium_signals (id, type, data) VALUES ('snapshot-test01', 'execute-snapshot', '{"data-collections": ["db1.table1"]}');{code}
Please let us know if you need any other information :)

<Your answer>

<Your answer>

sundongkim-dev:
Fixes debezium/dbz#683

## Description

For a PostgreSQL user-defined type (composite, enum, or domain), the column type name reported in change events differs between the snapshot and streaming paths:

- **Snapshot** (`PostgresConnection.doReadTableColumn`) takes the type name from the JDBC driver metadata (`ResultSet.getString(6)` / `TYPE_NAME`). The driver reports a **schema-qualified** name such as `"schema"."type"` when the type's schema is not on the `search_path`, and the unqualified name otherwise.
- **Streaming** (`PgOutputMessageDecoder`) always uses the **unqualified** name via `PostgresType#getName()`.

A custom converter that matches on `column.type.name` (via `converterFor`) therefore matches streaming change events but silently fails to match snapshot (initial or incremental) events. This was reported and confirmed as a bug by the maintainers, who asked for a test + PR to make the two paths consistent.

## Fix

In `doReadTableColumn`, after the type is resolved from the `TypeRegistry`, normalize the column type name to `PostgresType#getName()` — the same source the streaming path uses:

```java
PostgresType nativeType = getTypeRegistry().get(tableId.schema(), typeName);
...
column.type(nativeType.getName());
```

Notes:
- This uses the type's own name (`getName()`), **not** its root type (`getRootType().getName()`). The `nativeType`/`jdbcType` are still derived from the root type (unchanged), matching streaming, but the type *name* of a domain must stay its own name (e.g. `positive_int`) rather than collapsing to the base type (`int4`).
- Built-in types (e.g. `int4`, `numeric`, `text`, `jsonb`, arrays) are always reported unqualified by the driver, so normalizing them is a no-op — no special-casing is needed.
- The single-argument `ColumnEditor.type(String)` sets both the type name and the type expression, so both are normalized to the unqualified form.

## Tests

- New `ConnectionIT.shouldReadUnqualifiedUserDefinedTypeNameRegardlessOfSearchPath` reads the schema for a table with a composite, an enum, a domain, and a built-in column under two `search_path` settings (schema off-path → driver qualifies; schema on-path → driver already unqualified). In both cases the snapshot type names are asserted to be the unqualified names, matching streaming.
- Verified as a regression test: reverting the fix makes the assertion fail with `expected "financial_asset" but was ""dbz5571"."financial_asset""`.

## PR Checklist

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
@@ -726,13 +726,13 @@ public class PostgresConnection extends JdbcConnection {
             column.nativeType(nativeType.getRootType().getOid());
             column.jdbcType(nativeType.getRootType().getJdbcId());
 
-            // The JDBC driver reports a user-defined type with a schema-qualified name (for example
-            // "schema"."type") when the type's schema is not on the search_path, whereas the streaming
-            // path always uses the unqualified type name via PostgresType#getName(). Strip the schema
-            // qualifier so the column type name is consistent between snapshot and streaming
-            // (see debezium/dbz#683). Only the qualified form is normalized; type names the driver
-            // already reports unqualified (including built-in aliases such as serial) are left untouched.
-            if (typeName.contains(".")) {
+            // The JDBC driver reports a user-defined type schema-qualified (e.g. "schema"."type") when
+            // its schema is not on the search_path, whereas streaming always uses the unqualified name
+            // via PostgresType#getName(). Normalize to the unqualified form so snapshot and streaming
+            // agree (debezium/dbz#683), but skip it when the type did not resolve so an unrecognized
+            // name (e.g. from a PostgreSQL-compatible source) keeps the driver's spelling rather than
+            // the UNKNOWN placeholder.
+            if (typeName.contains(".") && nativeType != PostgresType.UNKNOWN) {
                 column.type(nativeType.getName());
             }
 
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
@@ -726,6 +726,16 @@ public class PostgresConnection extends JdbcConnection {
             column.nativeType(nativeType.getRootType().getOid());
             column.jdbcType(nativeType.getRootType().getJdbcId());
 
+            // The JDBC driver reports a user-defined type with a schema-qualified name (for example
+            // "schema"."type") when the type's schema is not on the search_path, whereas the streaming
+            // path always uses the unqualified type name via PostgresType#getName(). Strip the schema
+            // qualifier so the column type name is consistent between snapshot and streaming
+            // (see debezium/dbz#683). Only the qualified form is normalized; type names the driver
+            // already reports unqualified (including built-in aliases such as serial) are left untouched.
+            if (typeName.contains(".")) {
+                column.type(nativeType.getName());
+            }
+
             // For domain types, the postgres driver is unable to traverse a nested unbounded
             // hierarchy of types and report the right length/scale of a given type. We use
             // the TypeRegistry to accomplish this since it is capable of traversing the type
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/ConnectionIT.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/ConnectionIT.java
@@ -5,6 +5,7 @@
  */
 package io.debezium.connector.postgresql;
 
+import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.assertThatThrownBy;
 
 import java.sql.SQLException;
@@ -14,7 +15,11 @@ import org.postgresql.util.PSQLException;
 
 import io.debezium.config.Configuration;
 import io.debezium.connector.postgresql.connection.PostgresConnection;
+import io.debezium.doc.FixFor;
 import io.debezium.jdbc.JdbcConfiguration;
+import io.debezium.relational.Table;
+import io.debezium.relational.TableId;
+import io.debezium.relational.Tables;
 import io.debezium.util.Testing;
 
 public class ConnectionIT implements Testing {
@@ -49,6 +54,51 @@ public class ConnectionIT implements Testing {
             assertThatThrownBy(() -> conn.execute("SELECT pg_sleep(10)"))
                     .isInstanceOf(PSQLException.class)
                     .hasMessage("ERROR: canceling statement due to user request");
+        }
+    }
+
+    @Test
+    @FixFor("debezium/dbz#683")
+    void shouldReadUnqualifiedUserDefinedTypeNameRegardlessOfSearchPath() throws SQLException {
+        try (PostgresConnection conn = TestHelper.create()) {
+            conn.connect();
+            conn.execute(
+                    "DROP SCHEMA IF EXISTS dbz5571 CASCADE",
+                    "CREATE SCHEMA dbz5571",
+                    "CREATE TYPE dbz5571.financial_asset AS (quantity numeric, instrument_code text)",
+                    "CREATE TYPE dbz5571.mood AS ENUM ('happy', 'sad')",
+                    "CREATE DOMAIN dbz5571.positive_int AS integer CHECK (VALUE > 0)",
+                    "CREATE TABLE dbz5571.t (id varchar PRIMARY KEY, asset dbz5571.financial_asset, "
+                            + "mood dbz5571.mood, amount dbz5571.positive_int, plain integer)");
+        }
+
+        // The JDBC driver reports a user-defined type with a schema-qualified name (e.g. "dbz5571"."mood")
+        // when the type's schema is not on the search_path, but with the unqualified name otherwise. The
+        // streaming path always uses the unqualified name, so the snapshot path must do the same
+        // (debezium/dbz#683).
+        try {
+            assertUnqualifiedTypeNames("public");
+            assertUnqualifiedTypeNames("dbz5571, public");
+        }
+        finally {
+            try (PostgresConnection conn = TestHelper.create()) {
+                conn.execute("DROP SCHEMA IF EXISTS dbz5571 CASCADE");
+            }
+        }
+    }
+
+    private void assertUnqualifiedTypeNames(String searchPath) throws SQLException {
+        try (PostgresConnection conn = TestHelper.createWithTypeRegistry()) {
+            conn.execute("SET search_path TO " + searchPath);
+            Tables tables = new Tables();
+            conn.readSchema(tables, null, "dbz5571", null, null, false);
+            Table table = tables.forTable(new TableId(null, "dbz5571", "t"));
+
+            assertThat(table.columnWithName("asset").typeName()).isEqualTo("financial_asset");
+            assertThat(table.columnWithName("mood").typeName()).isEqualTo("mood");
+            assertThat(table.columnWithName("amount").typeName()).isEqualTo("positive_int");
+            // built-in types are unaffected
+            assertThat(table.columnWithName("plain").typeName()).isEqualTo("int4");
         }
     }
 }
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/IncrementalSnapshotIT.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/IncrementalSnapshotIT.java
@@ -17,6 +17,7 @@ import java.util.Map;
 import java.util.Set;
 import java.util.stream.Collectors;
 
+import org.apache.kafka.connect.data.Schema;
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.source.SourceRecord;
 import org.junit.jupiter.api.AfterAll;
@@ -40,6 +41,7 @@ import io.debezium.kafka.KafkaClusterUtils;
 import io.debezium.pipeline.signal.channels.KafkaSignalChannel;
 import io.debezium.pipeline.source.snapshot.incremental.AbstractIncrementalSnapshotTest;
 import io.debezium.relational.RelationalDatabaseConnectorConfig;
+import io.debezium.relational.mapping.PropagateSourceMetadataToSchemaParameter;
 import io.strimzi.test.container.StrimziKafkaCluster;
 
 public class IncrementalSnapshotIT extends AbstractIncrementalSnapshotTest<PostgresConnector> {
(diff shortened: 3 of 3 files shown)


Naros:
Hi @sundongkim-dev, the OpenLineage tests seem unusual. Could you double-check that this isn't somehow related to your changes? In the meantime I'll refire the failed tests again.

Naros (approved):
Code LGTM, just waiting on CI.

sundongkim-dev:
@Naros good catch on the OpenLineage tests — I dug in and they're **directly caused by this change**, and it surfaced a bigger scope question I wanted to run by you before pushing.

**What's happening.** The one-line change (`column.type(nativeType.getName())` in `doReadTableColumn`) normalizes the snapshot type name for *every* column, not just UDTs. For a real user-defined type that's exactly the dbz#683 fix — pgjdbc reports it schema-qualified (`"schema"."type"`) when the type's schema isn't on the `search_path`, and this brings it back to the unqualified name that streaming already uses.

But it also runs built-in `serial`/`smallserial`/`bigserial` through `TypeRegistry.get(...)`, which aliases them to `int4`/`int2`/`int8`. So a `serial` column's snapshot type name flips `serial → int4` — that's what breaks `OpenLineageIT` (snapshot facet now reports `pk;int4` instead of `pk;serial`). Worth noting: **streaming already reports `int4` for a `serial` column under both decoders** (OID-derived name), so on that axis the change actually makes snapshot *consistent* with streaming.

**The catch.** That same `serial → int4` normalization also flows into `__debezium.source.column.type` (`PropagateSourceMetadataToSchemaParameter`), which the JDBC sink's `SerialType` relies on to re-create `SERIAL`/`BIGSERIAL` columns. So with `column.propagate.source.type` enabled on a Postgres→Postgres pipeline, `AbstractJdbcSinkPipelineIT.testSerialDataType` / `testSmallSerialDataType` / `testBigSerialDataType` would get plain integer columns instead of `SERIAL`. (Code-level inference — I can't run the container e2e matrix locally.)

So there's a scope decision here. One option that keeps this PR focused:

- **Narrow this PR to UDTs only** — guard the normalization to the schema-qualified case, leaving `serial`/`smallserial`/`bigserial` untouched. That fixes exactly what dbz#683 reports (UDT names), needs no OpenLineage/JDBC-sink test changes, and keeps the JDBC sink's SERIAL round-trip intact.
- **Then handle `serial` separately** — the snapshot-vs-streaming inconsistency for `serial` (snapshot `serial`, streaming `int4`) is really its own question, since making them agree also drops the sink's ability to re-create `SERIAL`/`BIGSERIAL` columns. That feels like it deserves its own issue/PR where the sink impact can be weighed on its own, rather than riding along here.

Alternatively, if you'd rather normalize everything in one go here, I'll update the OpenLineage + JDBC-sink serial expectations to match.

Is there context I'm missing on how the `serial` type name is expected to behave across snapshot/streaming and the JDBC sink — or a direction you'd prefer?

Naros:
Hi @sundongkim-dev I'd prefer we go with option 1 and we keep the PR work narrow, focusing only on UDT. I suspect that for the `serial` discussion, that's going to be a much larger can of worms to crack, with decisions to be made about backward compatibility, etc. Do you concur?

sundongkim-dev:
Agreed, @Naros — let's go with option 1. I've narrowed it to only strip the schema qualifier the driver adds for user-defined types (the `"schema"."type"` form), leaving `serial`/`smallserial`/`bigserial` and other driver-reported names untouched — so there's no OpenLineage or JDBC-sink SERIAL impact. I'll open a separate issue for the `serial` snapshot-vs-streaming inconsistency, since as you say it brings in backward-compat trade-offs around the sink's SERIAL round-trip. Narrowed version force-pushed.

sundongkim-dev:
Filed as [debezium/dbz#2232](https://github.com/debezium/dbz/issues/2232) for the `serial` snapshot-vs-streaming discussion.

Naros:
Hi @sundongkim-dev, as I was going back over this PR, I had one question. What is the risk here with the `contains(".")` guard for PostgreSQL-compatible databases? I'm thinking could this be risky with Aurora PostgreSQL or any other PG compatible database where the source isn't the native open-source PG database.

sundongkim-dev:
Good catch — `contains(".")` alone is a looser gate than needed. The substituted value is never a guess (`nativeType.getName()`, the type already resolved for this column, which is what streaming emits), so it wouldn't corrupt a PG-compatible name — but the gate shouldn't fire when the type didn't resolve, which is exactly the non-native-PG case. Tightened it to `typeName.contains(".") && nativeType != PostgresType.UNKNOWN`, so an unresolved type keeps the driver's spelling instead of the `"unknown"` placeholder (`UNKNOWN` is the registry's single fallback instance, so the reference check is exact). That no-op path isn't reachable in an IT against real PostgreSQL, so it's covered by inspection. Pushed in 7e04483.

Naros (commented):
Hi @sundongkim-dev after a quick re-read of the issue, how difficult would it be to add an incremental snapshot integration test, as outlined in the issue? 

Ideally I'd like to make sure that there isn't also a corner case bug in the incremental snapshot flow too, since that was the context in which the use case was reported.

sundongkim-dev:
Added `incrementalSnapshotReportsUnqualifiedUserDefinedTypeNames`. Without the fix it fails with the type coming through as `"s1"."mood"`, and passes with it, so there's no separate bug in the incremental snapshot path. Composite types get dropped from the CDC output (JDBC 2002), so this covers the enum/domain UDTs and composite stays covered by `ConnectionIT`.

Naros:
Applied @sundongkim-dev, thanks!

PR #7492, opened 2026-06-01
debezium/dbz#2004 Fix duplicate transactional logical messages after connector restart

atorik:
Fixes debezium/dbz#2004

## Description

Looking at the logs when the issue occurs, I see the following message:

 ```
 INFO Last processed event was MESSAGE operation at LSN 'LSN{1/E41D9A90}', will restart from first LSN 'LSN{1/E41D9A90}' (io.debezium.connector.postgresql.connection.WalPositionLocator:113)
 ```

According to the comments in the code, this path is intended only for non-transactional MESSAGE operations:

```java
109:        // For non-transactional MESSAGE operations, lastCommitStoredLsn equals lastEventStoredLsn
110:        // (both set to the MESSAGE's LSN). Since the MESSAGE was fully processed and committed,
111:        // we can safely resume from the first LSN received after restart.
112:        if (lastProcessedMessageType == Operation.MESSAGE && lastCommitStoredLsn.equals(lastEventStoredLsn)) {
113:            LOGGER.info("Last processed event was MESSAGE operation at LSN '{}', will restart from first LSN '{}'",
```

However, in this case the message was emitted using `pg_logical_emit_message(true, ...)`, i.e. a transactional message, so I would not expect this condition to match.

I am not yet sure whether the following is the best fix, but my current understanding is that transactional logical messages should be handled similarly to regular DML operations such as INSERT and UPDATE.

For INSERT/UPDATE events, `isLastEventForLsn()` always returns true:

https://github.com/debezium/debezium/blob/6034037b802cee1fb04656e16883d2c5d7d97477/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/ReplicationMessage.java#L218-L220

In contrast, the implementation for `pg_logical_emit_message()` currently returns `!isTransactional`.

When I change it to always return `true`, the duplicate-message issue no longer occurs.
In that case, WalPositionLocator follows the same startup path as regular DML events and emits:

```
INFO LSN after last stored change LSN 'LSN{1/E41D9DD0}' received (io.debezium.connector.postgresql.connection.WalPositionLocator:96)
```

What do you think?

## PR Checklist

- [x] I have read the [contribution guidelines](https://github.com/debezium/debezium/blob/main/CONTRIBUTING.md) and the [governance document](https://github.com/debezium/governance/blob/main/GOVERNANCE.md) on PR expectations.
- [x] Minimal changes to code not directly related to your change (e.g. no unnecessary formatting changes or refactoring to existing code)
[...]

PR #7653, opened 2026-07-12
debezium/dbz#2004 Fix duplicate transactional logical messages after connector restart

### LINKED ISSUE #2004: Transactional pg_logical_emit_message() events are emitted twice after connector restart

main(HEAD)

---

```
% cat docker-min-no-table.json
{
  "name": "test-connector",
  "config": {
    "database.port": "5432",
    "database.user": "postgres",
    "topic.prefix": "example",
    "slot.name": "emit_log_message",
    "plugin.name": "pgoutput",
    "publication.autocreate.mode": "no_tables",
    "publication.name": "notable",
    "snapshot.mode": "no_data"
  }
}
```

---

on-premise

---

When a message is emitted using `pg_logical_emit_message()` with `transactional = true` before the connector is stopped, the message should be delivered exactly once after the connector is restarted.

According to the Debezium documentation, the PostgreSQL source connector supports Exactly Once Delivery:

https://debezium.io/documentation/reference/stable/configuration/eos.html

---

**What behavior do you see?**

After restarting the connector, the same transactional logical message is delivered twice.

For example, a message emitted via:

```
=# BEGIN;
=# SELECT * FROM pg_logical_emit_message(true, 'test1', 'aaa');
=# COMMIT;
```

is published twice to the Kafka topic after connector restart.

---

Yes.

---

Yes. I can provide them if needed.

---

Start the PostgreSQL connector using the configuration above:

```
$ curl -i -X POST -H "Accept:application/json" -H  "Content-Type:application/json" http://localhost:8083/connectors/ -d @docker-min-no-table.json
```

Emit a transactional logical message:

```
=# BEGIN; SELECT * FROM pg_logical_emit_message(true, 'test1', 'aaa'); COMMIT;
```

Stop the connector.

Restart the connector.

Consume from the message topic:

```
$ docker-compose -f docker-compose-postgres.yaml exec kafka /kafka/bin/kafka-console-consumer.sh \
    --bootstrap-server kafka:9092 \
    --from-beginning \
    --property print.key=true \
    --topic example.message

```

Observe that two identical records with prefix test1 are present in the topic, although only one message was emitted:

```
{"schema":{"type":"struct","fields":[{"type":"string","optional":true,"field":"prefix"}],"optional":false,"name":"io.debezium.connector.postgresql.MessageKey","version":1},"payload":{"prefix":"test1"}}      {"schema":{"type":"struct","fields":[{"type":"string","optional":false,"field":"op"},{"type":"int64","optional":true,"field":"ts_ms"},{"type":"struct","fields":[{"type":"string","optional":false,"field":"version"},{"type":"string","optional":false,"field":"connector"},{"type":"string","optional":false,"field":"name"},{"type":"int64","optional":false,"field":"ts_ms"},{"type":"string","optional":true,"name":"io.debezium.data.Enum","version":1,"parameters":{"allowed":"true,first,first_in_data_collection,last_in_data_collection,last,false,incremental"},"default":"false","field":"snapshot"},{"type":"string","optional":false,"field":"db"},{"type":"string","optional":true,"field":"sequence"},{"type":"int64","optional":true,"field":"ts_us"},{"type":"int64","optional":true,"field":"ts_ns"},{"type":"string","optional":false,"field":"schema"},{"type":"string","optional":false,"field":"table"},{"type":"int64","optional":true,"field":"txId"},{"type":"int64","optional":true,"field":"lsn"},{"type":"int64","optional":true,"field":"xmin"},{"type":"string","optional":true,"field":"origin"},{"type":"int64","optional":true,"field":"origin_lsn"}],"optional":false,"name":"io.debezium.connector.postgresql.Source","version":1,"field":"source"},{"type":"struct","fields":[{"type":"string","optional":false,"field":"id"},{"type":"int64","optional":false,"field":"total_order"},{"type":"int64","optional":false,"field":"data_collection_order"}],"optional":true,"name":"event.block","version":1,"field":"transaction"},{"type":"struct","fields":[{"type":"string","optional":true,"field":"prefix"},{"type":"bytes","optional":true,"field":"content"}],"optional":false,"name":"io.debezium.connector.postgresql.Message","version":1,"field":"message"}],"optional":false,"name":"io.debezium.connector.postgresql.MessageValue","version":1},"payload":{"op":"m","ts_ms":1780131651282,"source":{"version":"3.5.1.Final","connector":"postgresql","name":"example","ts_ms":1780131650920,"snapshot":"false","db":"postgres","sequence":"[\"5549461504\",\"5549461624\"]","ts_us":1780131650920271,"ts_ns":1780131650920271000,"schema":"","table":"","txId":33926,"lsn":5549461624,"xmin":null,"origin":null,"origin_lsn":null},"transaction":null,"message":{"prefix":"test1","content":"YWFh"}}}
{"schema":{"type":"struct","fields":[{"type":"string","optional":true,"field":"prefix"}],"optional":false,"name":"io.debezium.connector.postgresql.MessageKey","version":1},"payload":{"prefix":"test1"}}      {"schema":{"type":"struct","fields":[{"type":"string","optional":false,"field":"op"},{"type":"int64","optional":true,"field":"ts_ms"},{"type":"struct","fields":[{"type":"string","optional":false,"field":"version"},{"type":"string","optional":false,"field":"connector"},{"type":"string","optional":false,"field":"name"},{"type":"int64","optional":false,"field":"ts_ms"},{"type":"string","optional":true,"name":"io.debezium.data.Enum","version":1,"parameters":{"allowed":"true,first,first_in_data_collection,last_in_data_collection,last,false,incremental"},"default":"false","field":"snapshot"},{"type":"string","optional":false,"field":"db"},{"type":"string","optional":true,"field":"sequence"},{"type":"int64","optional":true,"field":"ts_us"},{"type":"int64","optional":true,"field":"ts_ns"},{"type":"string","optional":false,"field":"schema"},{"type":"string","optional":false,"field":"table"},{"type":"int64","optional":true,"field":"txId"},{"type":"int64","optional":true,"field":"lsn"},{"type":"int64","optional":true,"field":"xmin"},{"type":"string","optional":true,"field":"origin"},{"type":"int64","optional":true,"field":"origin_lsn"}],"optional":false,"name":"io.debezium.connector.postgresql.Source","version":1,"field":"source"},{"type":"struct","fields":[{"type":"string","optional":false,"field":"id"},{"type":"int64","optional":false,"field":"total_order"},{"type":"int64","optional":false,"field":"data_collection_order"}],"optional":true,"name":"event.block","version":1,"field":"transaction"},{"type":"struct","fields":[{"type":"string","optional":true,"field":"prefix"},{"type":"bytes","optional":true,"field":"content"}],"optional":false,"name":"io.debezium.connector.postgresql.Message","version":1,"field":"message"}],"optional":false,"name":"io.debezium.connector.postgresql.MessageValue","version":1},"payload":{"op":"m","ts_ms":1780131741234,"source":{"version":"3.5.1.Final","connector":"postgresql","name":"example","ts_ms":1780131650920,"snapshot":"false","db":"postgres","sequence":"[\"5549461504\",\"5549461624\"]","ts_us":1780131650920271,"ts_ns":1780131650920271000,"schema":"","table":"","txId":33926,"lsn":5549461624,"xmin":null,"origin":null,"origin_lsn":null},"transaction":null,"message":{"prefix":"test1","content":"YWFh"}}}
```

I plan to submit a PR.

atorik:
Fixes debezium/dbz#2004 and debezium/dbz#2058 <the Debezium Issue Number>

## Description
In the previous PR (#7492), we reverted the change because of an issue [#2058](https://github.com/debezium/dbz/issues/2058) that could cause data loss due to PostgreSQL reporting a different kind of LSN for decoded logical messages than for other operations.

As discussed in https://github.com/debezium/dbz/issues/2058, if PostgreSQL changes the LSN reported for logical messages only in a future major release, Debezium will still need to handle the current special behavior for earlier PostgreSQL major versions.

I was not entirely sure whether such handling would be feasible, so I tried implementing it to confirm. This is done in commit 0ab1dc4d2ab848d769d7efd4ab1e830df7067147.

If PostgreSQL decides to fix this behavior in a future major or minor release, we need to add a version check so that this special handling is applied only to the affected PostgreSQL versions. I'm going to do it after PostgreSQL-side discussion reaches a conclusion.

The other commits reintroduce the changes that were reverted previously.

WDYT?

## PR Checklist

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/LogicalDecodingMessage.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/LogicalDecodingMessage.java
@@ -35,7 +35,7 @@ public class LogicalDecodingMessage implements ReplicationMessage {
 
     @Override
     public boolean isLastEventForLsn() {
-        return !isTransactional;
+        return true;
     }
 
     @Override
@@ -74,5 +74,9 @@ public class LogicalDecodingMessage implements ReplicationMessage {
 
     public byte[] getContent() {
         return content;
+    }
+
+    public boolean isTransactional() {
+        return isTransactional;
     }
 }
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
@@ -49,6 +49,7 @@ public class WalPositionLocator {
     private Set<Lsn> lsnSeen = new HashSet<>(1_000);
     private long currentLsnEventCount = 0;
     private long startStreamingEventsToSkip = 0;
+    private boolean skipProcessedLogicalMessage = false;
 
     public WalPositionLocator(Lsn lastCommitStoredLsn, Lsn lastEventStoredLsn, Operation lastProcessedMessageType) {
         this(lastCommitStoredLsn, lastEventStoredLsn, lastProcessedMessageType, 0);
@@ -120,6 +121,19 @@ public class WalPositionLocator {
                     LOGGER.info("Will restart from LSN '{}' corresponding to the event following the BEGIN event", txStartLsn);
                     startStreamingLsn = txStartLsn;
                     return Optional.of(startStreamingLsn);
+                }
+
+                // PostgreSQL currently reports the end LSN of the MESSAGE rather than its start
+                // LSN. Since this LSN can also identify the next decoded operation, resume from
+                // that LSN and skip the already processed MESSAGE operation.
+                //
+                // This PostgreSQL behavior is under discussion and may change in a future version:
+                // https://www.postgresql.org/message-id/flat/d99c688994ab3a998afe26e61fe4f69f%40oss.nttdata.com
+                if (lastProcessedMessageType == Operation.MESSAGE) {
+                    startStreamingLsn = currentLsn;
+                    skipProcessedLogicalMessage = true;
+                    LOGGER.info("Last processed event was MESSAGE operation; will restart from LSN '{}' and skip that MESSAGE operation", currentLsn);
+                    return Optional.of(currentLsn);
                 }
                 return Optional.empty();
             }
@@ -228,6 +242,17 @@ public class WalPositionLocator {
         }
         LOGGER.debug("Message with LSN '{}' filtered", lsn);
         return true;
+    }
+
+    public boolean skipProcessedLogicalMessage(Lsn lsn) {
+        if (skipProcessedLogicalMessage && startStreamingLsn != null && startStreamingLsn.equals(lsn)) {
+            skipProcessedLogicalMessage = false;
+            passMessages = true;
+            lsnSeen = new HashSet<>();
+            LOGGER.info("Processed MESSAGE operation with LSN '{}' skipped, switching off the filtering", lsn);
+            return true;
+        }
+        return false;
     }
 
     /**
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
@@ -49,6 +49,7 @@ public class WalPositionLocator {
     private Set<Lsn> lsnSeen = new HashSet<>(1_000);
     private long currentLsnEventCount = 0;
     private long startStreamingEventsToSkip = 0;
+    private boolean skipProcessedLogicalMessage = false;
 
     public WalPositionLocator(Lsn lastCommitStoredLsn, Lsn lastEventStoredLsn, Operation lastProcessedMessageType) {
         this(lastCommitStoredLsn, lastEventStoredLsn, lastProcessedMessageType, 0);
@@ -120,6 +121,19 @@ public class WalPositionLocator {
                     LOGGER.info("Will restart from LSN '{}' corresponding to the event following the BEGIN event", txStartLsn);
                     startStreamingLsn = txStartLsn;
                     return Optional.of(startStreamingLsn);
+                }
+
+                // PostgreSQL currently reports the end LSN of the MESSAGE rather than its start
+                // LSN. Since this LSN can also identify the next decoded operation, resume from
+                // that LSN and skip the already processed MESSAGE operation.
+                //
+                // This PostgreSQL behavior is under discussion and may change in a future version:
+                // https://www.postgresql.org/message-id/flat/d99c688994ab3a998afe26e61fe4f69f%40oss.nttdata.com
+                if (lastProcessedMessageType == Operation.MESSAGE) {
+                    startStreamingLsn = currentLsn;
+                    skipProcessedLogicalMessage = true;
+                    LOGGER.info("Last processed event was MESSAGE operation; will restart from LSN '{}' and skip that MESSAGE operation", currentLsn);
+                    return Optional.of(currentLsn);
                 }
                 return Optional.empty();
             }
@@ -241,6 +255,17 @@ public class WalPositionLocator {
         }
         LOGGER.debug("Message with LSN '{}' filtered", lsn);
         return true;
+    }
+
+    public boolean skipProcessedLogicalMessage(Lsn lsn) {
+        if (skipProcessedLogicalMessage && startStreamingLsn != null && startStreamingLsn.equals(lsn)) {
+            skipProcessedLogicalMessage = false;
+            passMessages = true;
+            lsnSeen = new HashSet<>();
+            LOGGER.info("Processed MESSAGE operation with LSN '{}' skipped, switching off the filtering", lsn);
+            return true;
+        }
+        return false;
     }
 
     /**
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/LogicalDecodingMessageMonitor.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/LogicalDecodingMessageMonitor.java
@@ -108,7 +108,7 @@ public class LogicalDecodingMessageMonitor {
         sender.accept(new SourceRecord(partition.getSourcePartition(), offsetContext.getOffset(), topicName,
                 keySchema, key, value.schema(), value));
 
-        if (message.isLastEventForLsn()) {
+        if (!message.isTransactional()) {
             offsetContext.getTransactionContext().endTransaction();
         }
     }
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresStreamingChangeEventSource.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresStreamingChangeEventSource.java
@@ -358,12 +358,13 @@ public class PostgresStreamingChangeEventSource implements StreamingChangeEventS
             maybeWarnAboutGrowingWalBacklog(true);
         }
         else if (message.getOperation() == Operation.MESSAGE) {
-            offsetContext.updateWalPosition(lsn, lastCompletelyProcessedLsn, message.getCommitTime(), toLong(message.getTransactionId()),
+            final LogicalDecodingMessage logicalMessage = (LogicalDecodingMessage) message;
+            offsetContext.updateWalPosition(lsn, lastCompletelyProcessedLsn, logicalMessage.getCommitTime(), toLong(logicalMessage.getTransactionId()),
                     getSlotXmin(),
-                    message.getOperation());
+                    logicalMessage.getOperation());
 
             // non-transactional message that will not be followed by a COMMIT message
-            if (message.isLastEventForLsn()) {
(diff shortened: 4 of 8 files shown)


atorik:
The test reintroduced in this PR is failing with the following error:

```
2026-07-12T08:27:25.4722939Z [ERROR]   PostgresConnectorIT.shouldNotReemitConsumedLogicalMessageAfterRestart:4200
2026-07-12T08:27:25.4723788Z expected: 1
2026-07-12T08:27:25.4724151Z  but was: 0
```

However, line 4200 simply verifies that a single logical message was received before the connector was restarted. Therefore, I do not think this failure is directly related to the issue addressed by this change.
I have also confirmed that the test passes in my local environment.

Naros:
Hi @atorik could the failure be because `decoderbufs` and `pgoutput` handle this differently and the test method was not marked to be skipped for `decoderbufs`, which is what failed?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/LogicalDecodingMessage.java
```diff
@@ -35,7 +35,7 @@ public LogicalDecodingMessage(Operation op, Instant commitTimestamp, Long transa
 
     @Override
     public boolean isLastEventForLsn() {
-        return !isTransactional;
+        return true;
```
Naros, line 38: Inside `processReplicationMessages`, for transactional-based logic decoding messages, this will update the last processed LSN even when the commit hasn't yet been observed. Are we sure this doesn't introduce any potential side-effects?

atorik:
> could the failure be because decoderbufs and pgoutput handle this differently and the test method was not marked to be skipped for decoderbufs, which is what failed?

Thanks! That seems to be the reason.
Updated the branch.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/LogicalDecodingMessage.java
atorik: @Naros  You may have already seen this, but as noted in the previous PR https://github.com/debezium/debezium/pull/7492, this change aligns the handling with regular DML operations::


> my current understanding is that transactional logical messages should be handled similarly to regular DML operations such as INSERT and UPDATE.

> For INSERT/UPDATE events, isLastEventForLsn() always returns true:

This point was also discussed in the previous PR, and we did not identify any specific problem caused by this change.  I agree that it's worth considering this carefully.

Naros: Well, I raise this because the original PR changes were reverted.

atorik: @Naros The changes were reverted because they exposed another issue caused by PostgreSQL's behavior, tracked as [#2058](https://github.com/debezium/dbz/issues/2058):

```
It appears that the root cause is PostgreSQL's handling of the LSN returned for decoded pg_logical_emit_message() operations.
For normal WAL-decoded operations such as INSERT, PostgreSQL reports the LSN corresponding to the beginning of the operation. However, for pg_logical_emit_message(), PostgreSQL reports the LSN corresponding to the end of the operation.
```

This PR restores the reverted changes to address [#2004](https://github.com/debezium/dbz/issues/2004)., while also adding commit 0ab1dc4 to address [#2058](https://github.com/debezium/dbz/issues/2058).

Naros: Thanks for the explanation @atorik.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
```diff
@@ -121,6 +122,19 @@ public Optional<Lsn> resumeFromLsn(Lsn currentLsn, ReplicationMessage message) {
                     startStreamingLsn = txStartLsn;
                     return Optional.of(startStreamingLsn);
                 }
+
+                // PostgreSQL currently reports the end LSN of the MESSAGE rather than its start
+                // LSN. Since this LSN can also identify the next decoded operation, resume from
+                // that LSN and skip the already processed MESSAGE operation.
+                //
```
vjuranek, line 129: Just for the clarity, could you please add a note here, that this is pgoutput specific?

vjuranek (approved):
Just a nit-pick comment about adding more clarity in the comment, but the fix itself seems good to me.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/WalPositionLocator.java
atorik: Thanks for reviewing!

Hmm, since Debezium can decode logical messages only with pgoutput, as described in [the documentation](https://debezium.io/documentation/reference/stable/connectors/postgresql.html#postgresql-message-events), I am not entirely sure why we need to add a comment specifically here.

Also, there is already handling for logical message starting around line 164 in the same file, but it does not mention that the handling is specific to pgoutput either:

```
        // For non-transactional MESSAGE operations, lastCommitStoredLsn equals lastEventStoredLsn
        // (both set to the MESSAGE's LSN). Since the MESSAGE was fully processed and committed,
        // we can safely resume from the first LSN received after restart.
        if (lastProcessedMessageType == Operation.MESSAGE && lastCommitStoredLsn.equals(lastEventStoredLsn)) {
```
If we add such a note here, I think we should probably add one there as well. Personally, however, I do not feel strongly that this needs to be stated within the implementation itself.

WDYT?

..By the way, strictly speaking, the underlying issue is not specific to pgoutput; it affects all output plugins. Debezium supports two output plugins, decoderbufs and pgoutput, and decoderbufs does not support decoding logical messages, as noted above. Therefore, in practice, the issue arises only with pgoutput in Debezium.

vjuranek: Yeah fair point. My intention was that this is used also by decoderbufs so I'd like to mark it as pgoutput specific in the comment, but you are right decoderbufs doesn't support it and we are not mentioning pgoutput specific part in existing code, so this is fine.

vjuranek:
LGTM, @Naros @jpechane any opinions on this? (to avoid the situation I overlooked something and will have to revert it again :-)

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/LogicalDecodingMessage.java
Naros: @atorik, would you do me a favor? 

Just for completeness and peace of mind, could you add an integration test that creates a transaction with a real DML event, a transactional logical decoding message, and then another DML event before the transaction commits? This test should verify that the transaction metadata does not reset and that it is emitted once rather than 3 times.

@vjuranek this is the only thing I see that stands out to me.

atorik: @Naros Added a [commit](https://github.com/debezium/debezium/pull/7653/changes/135cc28eca77162f3770606a167cd8982e0071db) for this.
Does this match what you had in mind?

atorik: Hi @Naros , just a gentle ping on this.
When you have a chance, could you take a look at the additional test I added?
Please let me know if this matches what you had in mind, or if any further changes are needed.

Naros (commented):
Hi @atorik I would genuinely prefer if we had two tests added in `TransactionMetadataIT` where there is a transaction sequence as follows:
```
BEGIN
INSERT
logical message
INSERT
COMMIT
```
For this sequence we should be testing both local message as transactionl and as non-transactional and checking that we get the expected transaction metadata BEGIN and COMMIT/END notifications.

My guess here is when transaction metadata is enabled and logical message is transactional, we will get a spurious BEGIN because `isLastEventForLsn` always returns `true` and we should validate that the expected event count is correct for the transaction.

If that passes and we get the expected values ,then I'm good with the chaneg.

atorik:
> My guess here is when transaction metadata is enabled and logical message is transactional, we will get a spurious BEGIN because isLastEventForLsn always returns true and we should validate that the expected event count is correct for the transaction.

@Naros Thanks! Your assumption was correct.

After adding tests for both transactional and non-transactional logical messages, the transactional case exposed a spurious `BEGIN`.

The cause was that the condition for calling `endTransaction()` still used `isLastEventForLsn()`. Now that `isLastEventForLsn()` correctly always returns `true` for logical messages, this also cleared the transaction context for transactional messages.

I updated the PR to use `!message.isTransactional()` for that condition, so the transaction context is ended there only for non-transactional messages.

Naros (approved):
This LGTM, I'd like @vjuranek to give it one final pass as well.

vjuranek:
LGTM, going to merge it

vjuranek:
Merged, thanks @atorik !

PR #7654, opened 2026-07-13
debezium/dbz#2221 Skip corrupted records during Redis schema history recovery instead of silently truncating it

TimurRakhmatullin86:
Fixes debezium/dbz#2221

`RedisSchemaHistory.recoverRecords()` aborts on the first schema history entry that fails to deserialize:

```java
catch (IOException e) {
    LOGGER.error("Failed to convert record to string: {}", item, e);
    return;
}
```

A single corrupted entry in the Redis stream (e.g. truncated JSON) makes the method return early, silently discarding every record that comes after it. The connector then proceeds with a partially recovered schema history — no error is raised at recovery time, and the damage only shows up later as "table schema unknown" failures or wrong parsing for tables whose DDL lived in the dropped tail.

This makes the Redis store strictly worse than its siblings on the same input:

- `KafkaSchemaHistory` catches `IOException` per record, logs it via `Loggings.logErrorAndTraceRecord` and continues with the remaining records;
- `RocketMqSchemaHistory` skips invalid records and throws on read errors;
- `RedisSchemaHistory` neither fails nor recovers fully — it silently truncates.

Changes:

- Skip the corrupted record and continue recovering the remaining ones, aligning with the Kafka implementation. The record content is logged via `Loggings.logErrorAndTraceRecord` (record body at TRACE), consistent with `storeRecord()` in the same class.
- Fix the retry description for the `XRANGE` call in `recoverRecords()`: it said "Writing to database schema history stream" although the operation is a read, so retry warnings for a failed read were reported as a failed write.
- Add a unit test (no Redis instance needed): a fake `RedisClient` returns valid + corrupted + valid entries and the test asserts both valid records are recovered. It fails on the previous code with `expected: <2> but was: <1>`. To inject the fake client, `RedisSchemaHistory` gets a `@VisibleForTesting` package-private `setRedisClient()`, mirroring the existing `RedisOffsetBackingStore.setRedisClient()`.

If maintainers would rather fail loudly (`SchemaHistoryException`) than skip, happy to change it — either is an improvement over silent truncation; skipping matches the Kafka behaviour.

### DIFF
--- a/debezium-storage/debezium-storage-redis/src/test/java/io/debezium/storage/redis/history/RedisSchemaHistoryTest.java
+++ b/debezium-storage/debezium-storage-redis/src/test/java/io/debezium/storage/redis/history/RedisSchemaHistoryTest.java
@@ -1,0 +1,114 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.storage.redis.history;
+
+import static org.junit.jupiter.api.Assertions.assertEquals;
+
+import java.util.AbstractMap.SimpleEntry;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
+import org.junit.jupiter.api.Test;
+
+import io.debezium.config.Configuration;
+import io.debezium.doc.FixFor;
+import io.debezium.relational.history.HistoryRecord;
+import io.debezium.relational.history.SchemaHistory;
+import io.debezium.relational.history.SchemaHistoryListener;
+import io.debezium.storage.redis.RedisClient;
+
+public class RedisSchemaHistoryTest {
+
+    private static final String VALID_RECORD_1 = "{\"source\":{\"server\":\"test\"},\"position\":{\"file\":\"binlog.000001\",\"pos\":100},\"ddl\":\"CREATE TABLE first (id INT)\"}";
+    // truncated JSON, e.g. after a partial write to the stream
+    private static final String CORRUPTED_RECORD = "{\"source\":{\"server\":\"test\"},\"position\":";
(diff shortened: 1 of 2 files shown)


debezium-storage/debezium-storage-redis/src/test/java/io/debezium/storage/redis/history/RedisSchemaHistoryTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+public class RedisSchemaHistoryTest {
+
+    private static final String VALID_RECORD_1 = "{\"source\":{\"server\":\"test\"},\"position\":{\"file\":\"binlog.000001\",\"pos\":100},\"ddl\":\"CREATE TABLE first (id INT)\"}";
+    // truncated JSON, e.g. after a partial write to the stream
+    private static final String CORRUPTED_RECORD = "{\"source\":{\"server\":\"test\"},\"position\":";
+    private static final String VALID_RECORD_2 = "{\"source\":{\"server\":\"test\"},\"position\":{\"file\":\"binlog.000001\",\"pos\":200},\"ddl\":\"CREATE TABLE second (id INT)\"}";
+
+    @Test
+    public void shouldSkipCorruptedRecordAndRecoverSubsequentOnes() {
```
Naros, line 33:
```suggestion
    @FixFor("debezium/dbz#2221")
    public void shouldSkipCorruptedRecordAndRecoverSubsequentOnes() {
```

Naros (commented):
This LGTM, I've left one nit-pick and its good to go.

debezium-storage/debezium-storage-redis/src/test/java/io/debezium/storage/redis/history/RedisSchemaHistoryTest.java
TimurRakhmatullin86: Added the `@FixFor` annotation, thanks!

Naros:

Naros (approved):

Naros:

PR #7655, opened 2026-07-13
debezium/dbz#1340 Pin era parsing to English locale in Postgres DateTimeFormat

### LINKED ISSUE #1340: Debezium Connector Postgres  DateTimeParseException: Text '0002-11-30 00:00:00 BC' and '0002-11-30 BC' could not be parsed [DBZ-8762]
Migrated from [DBZ-8762](https://issues.redhat.com/browse/DBZ-8762)

h1. Bug report

3.0.8 Final 

!image-2025-03-07-11-34-31-709.png!

PostgreSQL 17.4 on x86_64-pc-linux-gnu, compiled by gcc (GCC) 8.5.0 20210514 (Red Hat 8.5.0-22.0.1), 64-bit

Can parse BC dates such as '0002-11-30 00:00:00 BC' or '0002-11-30 BC'.
h2. What behavior do you see?

The translation is:
"Cannot parse BC dates such as '0002-11-30 00:00:00 BC' or '0002-11-30 BC'."

Yes, I'm using the latest version 3.0.8 Final and the issue still persists.

!image-2025-03-07-11-26-51-674.png!

<Your answer>

<Your answer>

<Your answer>

sundongkim-dev:
Fixes debezium/dbz#1340

## Problem

PostgreSQL always emits the era literally as English `BC`/`AD` on the wire. The ERA-aware formatters in `ISODateTimeFormat` were built without pinning a locale, so on a JVM whose default locale is non-English (e.g. `ko_KR`, where the short era text is `기원전`) the `BC` token cannot be parsed. As reported, timestamp/date values before Christ such as `0002-11-30 00:00:00 BC` fail with:

```
DateTimeParseException: Text '0002-11-30 00:00:00 BC' could not be parsed, unparsed text found at index 19
```

## Fix

Build the five ERA-aware formatters (`TS_FORMAT`, `TS_TZ_FORMAT`, `TS_TZ_WITH_SECONDS_FORMAT`, `SYSTEM_TS_FORMAT`, `DATE_FORMAT_OPT_ERA`) with `toFormatter(Locale.ENGLISH)` so the era text is always resolved in English, matching what PostgreSQL sends. Formatters that do not use `ChronoField.ERA` are left untouched.

## Tests

`ISODateTimeFormatTest` previously derived the expected era text from `Locale.getDefault()`, so on English CI machines the test passed even with the bug present. This PR:

- hardcodes the expected era text to `"BC"` (what PostgreSQL actually sends), and
- forces a non-English default locale (`Locale.KOREAN`) in `@BeforeEach`, restoring it in `@AfterEach`, so the regression is actually exercised.

Verified: against the pre-fix source the BC assertions fail at the era offset (index 10/19/20/26); with the fix all cases pass.

## Question for maintainers

The formatters are `static final` fields, so the test only exercises the regression when `ISODateTimeFormat` is first loaded by this test (currently the only unit test that touches it). If a future unit test loads the class under an English default locale first, the formatters would already be initialized and the regression would no longer be caught. I noted this in a code comment. Would you prefer I pin the locale via a surefire fork (e.g. `-Duser.language=ko`) instead, to make it robust regardless of class-load order?

### DIFF
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/connection/ISODateTimeFormatTest.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/connection/ISODateTimeFormatTest.java
@@ -13,16 +13,39 @@ import java.time.LocalTime;
 import java.time.OffsetDateTime;
 import java.time.OffsetTime;
 import java.time.ZoneOffset;
-import java.time.chrono.IsoEra;
-import java.time.format.TextStyle;
 import java.util.Locale;
 
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
 
+import io.debezium.doc.FixFor;
+
 public class ISODateTimeFormatTest {
-    private static final String BCE_DISPLAY_NAME = IsoEra.BCE.getDisplayName(TextStyle.SHORT, Locale.getDefault());
+    // PostgreSQL always emits the era in English ("BC"/"AD") regardless of the JVM locale.
+    private static final String BCE_DISPLAY_NAME = "BC";
+
+    private Locale defaultLocale;
+
+    @BeforeEach
+    void setUp() {
+        // Force a non-English locale to guard against locale-dependent era parsing (dbz#1340).
+        // NOTE: the formatters under test are static final fields, so this only exercises the
+        // regression when ISODateTimeFormat is first loaded here (currently the only unit test
+        // that touches it). If a future unit test loads the class under an English default
+        // locale first, the formatters would be initialized before this runs and the regression
+        // would no longer be caught -- in that case pin the locale via a surefire fork instead.
+        defaultLocale = Locale.getDefault();
+        Locale.setDefault(Locale.KOREAN);
+    }
+
+    @AfterEach
+    void tearDown() {
+        Locale.setDefault(defaultLocale);
+    }
 
     @Test
+    @FixFor("debezium/dbz#1340")
     void testTimestampToInstant() {
         ZoneOffset offset = ZoneOffset.UTC;
         assertThat(DateTimeFormat.get().timestampToInstant("2016-11-04 13:51:30"))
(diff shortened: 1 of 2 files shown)


debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/connection/ISODateTimeFormatTest.java
```diff
@@ -13,16 +13,39 @@
...
+        Locale.setDefault(Locale.KOREAN);
+    }
+
+    @AfterEach
+    void tearDown() {
+        Locale.setDefault(defaultLocale);
+    }
 
     @Test
+    @FixFor("debezium/dbz#1340")
```
Naros: Given that these test methods did not change, there's no need to add a `@FixFor` annotation here and in the others that follow the same pattern.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/connection/ISODateTimeFormatTest.java
```diff
@@ -13,16 +13,39 @@
...
+    private Locale defaultLocale;
+
+    @BeforeEach
+    void setUp() {
+        // Force a non-English locale to guard against locale-dependent era parsing (dbz#1340).
+        // NOTE: the formatters under test are static final fields, so this only exercises the
+        // regression when ISODateTimeFormat is first loaded here (currently the only unit test
+        // that touches it). If a future unit test loads the class under an English default
+        // locale first, the formatters would be initialized before this runs and the regression
+        // would no longer be caught -- in that case pin the locale via a surefire fork instead.
```
Naros: Let's avoid long blocks of comments; this is really unhelpful. Can you summarize this in 1 or 2 lines?

Naros (changes requested):
This is close, I've left a few inline comments.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/connection/ISODateTimeFormatTest.java
sundongkim-dev: Good point — removed the `@FixFor` annotations from these unchanged tests and dropped the now-unused import.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/connection/ISODateTimeFormatTest.java
sundongkim-dev: Done, trimmed it down to a single line.

Naros (approved):

Naros:
LGTM @sundongkim-dev, just waits for CI.

Naros:
Applied @sundongkim-dev, thanks! :tada:

PR #7659, opened 2026-07-14
debezium/dbz#870 Declare schema history field as STRING to avoid eager class loading

### LINKED ISSUE #870: HistorizedRelationalDatabaseConnectorConfig.SCHEMA_HISTORY requires Kafka storage on the classpath [DBZ-6804]
Migrated from [DBZ-6804](https://issues.redhat.com/browse/DBZ-6804)

h1. Bug report

MySQL 2.3.2-Final

{{#name=engine}}
{{connector.class=io.debezium.connector.mysql.MySqlConnector}}
{{#offset.storage=org.apache.kafka.connect.storage.FileOffsetBackingStore}}
{{#offset.storage.file.filename=offsets7.dat}}
{{#offset.flush.interval.ms=60000}}
{{database.hostname=dre-cluster.fyre.ibm.com}}
{{database.port=3306}}
{{database.user=dbz}}
{{database.password=<PW>}}
{{database.server.id=123456}}
{{table.include.list=mqsi.Lending}}
{{schema.history.internal=io.debezium.storage.file.history.FileSchemaHistory}}
{{schema.history.internal.file.filename=mysql-schema-history.dat}}
{{#database.server.id=85802}}
{{#database.server.name=my-app-connector7}}
{{#database.history=io.debezium.relational.history.FileDatabaseHistory}}
{{#database.history.file.filename=deleteme-dbhistory7.dat}}
{{#debezium.source.offset.storage.file.filename=offset.dat}}

and

{{      Configuration results = Configuration.load(readFile(filename))}}
{{        .withSystemProperties(Function.identity()).edit()}}
{{        .with(EmbeddedEngine.ENGINE_NAME, "DebeziumWriter")}}
{{        .with(EmbeddedEngine.OFFSET_STORAGE, FileOffsetBackingStore.class)}}
{{        .with("name", "DebeziumWriter")}}
{{        .with("topic.prefix", "dbserver1")}}

As you can see from the comments, I've tried a few things...!

MySQL 8.1.0 in own kubernetes cluster

Successful connection

{{[pool-1-thread-1] ERROR io.debezium.embedded.EmbeddedEngine - null}}
{{Exception in thread "pool-1-thread-1" org.apache.kafka.common.config.ConfigException: Invalid value io.debezium.storage.kafka.history.KafkaSchemaHistory for configuration schema.history.internal: Class io.debezium.storage.kafka.history.KafkaSchemaHistory could not be found.}}
{{        at org.apache.kafka.common.config.ConfigDef.parseType(ConfigDef.java:744)}}
{{        at org.apache.kafka.common.config.ConfigDef$ConfigKey.<init>(ConfigDef.java:1172)}}
{{        at org.apache.kafka.common.config.ConfigDef.define(ConfigDef.java:152)}}
{{        at io.debezium.config.Field.group(Field.java:453)}}
{{        at io.debezium.config.ConfigDefinition.addToConfigDef(ConfigDefinition.java:104)}}
{{        at io.debezium.config.ConfigDefinition.configDef(ConfigDefinition.java:71)}}
{{        at io.debezium.connector.mysql.MySqlConnectorConfig.configDef(MySqlConnectorConfig.java:922)}}
{{        at io.debezium.connector.mysql.MySqlConnector.config(MySqlConnector.java:76)}}
{{        at io.debezium.embedded.EmbeddedEngine.run(EmbeddedEngine.java:717)}}
{{        at io.debezium.embedded.ConvertingEngineBuilder$2.run(ConvertingEngineBuilder.java:229)}}
{{        at java.base/java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1128)}}
{{        at java.base/java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:628)}}
{{        at java.base/java.lang.Thread.run(Thread.java:829)}}

 

 
I *think* this is the latest!

Yes

Dunno, I can provide the source code for the Java connector though.  I used your cache-invalidation example for partial inspiration.  This works perfectly for PostgreSQL, but I can't get it to work for MySQL

sundongkim-dev:
Fixes debezium/dbz#870

## Description

The `schema.history.internal` field defaults to the Kafka schema history implementation (`io.debezium.storage.kafka.history.KafkaSchemaHistory`) and was declared with `Type.CLASS`. Kafka's `ConfigDef` eagerly resolves a `CLASS` default via `Class.forName` when the configuration definition is built (`ConfigDef.define` → `parseType`). As a result, any runtime that does not ship the optional `debezium-storage-kafka` module on the classpath — for example the embedded engine, or a deployment using a different schema history store — fails to start with a `ConfigException`, even when a different schema history is configured.

This matches the root cause and the fix direction outlined by @jpechane in the original discussion (DBZ-6804).

**Change:** declare the field as `Type.STRING` while keeping the default value. Class resolution is deferred to when the history is actually instantiated in `getSchemaHistory()` (which resolves the class from the string value via `Configuration.getInstance` regardless of the field type). Behaviour is unchanged for Kafka-based deployments, and an invalid custom class name still fails clearly at instantiation time.

A regression test (`HistorizedRelationalDatabaseConnectorConfigTest`) builds the config definition in a module whose test classpath does not contain the Kafka storage implementation, reproducing the embedded scenario; it fails on `Type.CLASS` and passes on `Type.STRING`. The `OracleConnectorTest`/`SqlServerConnectorTest` config-definition assertions are adjusted because the field default is now surfaced as a `String` rather than a `Class`.

## PR Checklist

### DIFF
--- a/debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SqlServerConnectorTest.java
+++ b/debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SqlServerConnectorTest.java
@@ -69,8 +69,11 @@ public class SqlServerConnectorTest {
             assertThat(key.importance).isEqualTo(expected.importance());
             assertThat(key.documentation).isEqualTo(expected.description());
             assertThat(key.type).isEqualTo(expected.type());
-            if (expected.equals(SqlServerConnectorConfig.SCHEMA_HISTORY) || expected.equals(CommonConnectorConfig.TOPIC_NAMING_STRATEGY)) {
+            if (expected.equals(CommonConnectorConfig.TOPIC_NAMING_STRATEGY)) {
                 assertThat(((Class<?>) key.defaultValue).getName()).isEqualTo((String) expected.defaultValue());
+            }
+            else if (expected.equals(SqlServerConnectorConfig.SCHEMA_HISTORY)) {
+                assertThat(key.defaultValue).isEqualTo(expected.defaultValue());
             }
             else if (expected.type() == ConfigDef.Type.LIST && key.defaultValue != null) {
                 assertThat(key.defaultValue).isEqualTo(Arrays.asList(expected.defaultValue()));
--- a/debezium-connector-common/src/main/java/io/debezium/relational/HistorizedRelationalDatabaseConnectorConfig.java
+++ b/debezium-connector-common/src/main/java/io/debezium/relational/HistorizedRelationalDatabaseConnectorConfig.java
@@ -54,7 +54,9 @@ public abstract class HistorizedRelationalDatabaseConnectorConfig extends Relati
      */
     public static final Field SCHEMA_HISTORY = Field.create("schema.history.internal")
             .withDisplayName("Database schema history class")
-            .withType(Type.CLASS)
+            // STRING rather than CLASS so the default class is not eagerly loaded when the config definition is
+            // built; the default lives in an optional module that may be absent (e.g. the embedded engine).
+            .withType(Type.STRING)
             .withWidth(Width.LONG)
             .withImportance(Importance.LOW)
             .withInvisibleRecommender()
--- a/debezium-connector-common/src/test/java/io/debezium/relational/HistorizedRelationalDatabaseConnectorConfigTest.java
+++ b/debezium-connector-common/src/test/java/io/debezium/relational/HistorizedRelationalDatabaseConnectorConfigTest.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.relational;
+
+import static org.assertj.core.api.Assertions.assertThat;
+import static org.assertj.core.api.Assertions.assertThatCode;
+
+import org.apache.kafka.common.config.ConfigDef;
+import org.junit.jupiter.api.Test;
+
+import io.debezium.config.Field;
+
+public class HistorizedRelationalDatabaseConnectorConfigTest {
+
+    /**
+     * The {@code schema.history.internal} field defaults to the Kafka schema history implementation, which lives in
+     * an optional module. When the field was declared as {@link ConfigDef.Type#CLASS}, building the configuration
+     * definition eagerly loaded that default class, so any environment that does not ship the Kafka storage module on
+     * the classpath (for example the embedded engine) failed with a {@link org.apache.kafka.common.config.ConfigException}.
+     */
+    @Test
+    @io.debezium.doc.FixFor("debezium/dbz#870")
+    public void schemaHistoryFieldShouldNotEagerlyLoadDefaultClass() {
+        // The Kafka schema history class is not on this module's test classpath, mirroring an embedded-engine setup.
+        final ConfigDef configDef = new ConfigDef();
+
+        assertThatCode(() -> Field.group(configDef, "history", HistorizedRelationalDatabaseConnectorConfig.SCHEMA_HISTORY))
+                .doesNotThrowAnyException();
+
+        // The default is still advertised so environments that do ship the Kafka storage module keep working.
+        assertThat(configDef.configKeys().get(HistorizedRelationalDatabaseConnectorConfig.SCHEMA_HISTORY.name()).defaultValue)
+                .isEqualTo(HistorizedRelationalDatabaseConnectorConfig.SCHEMA_HISTORY.defaultValue());
+    }
+}
--- a/debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleConnectorTest.java
+++ b/debezium-connector-oracle/src/test/java/io/debezium/connector/oracle/OracleConnectorTest.java
@@ -75,8 +75,11 @@ public class OracleConnectorTest {
             assertThat(key.importance).isEqualTo(expected.importance());
             assertThat(key.documentation).isEqualTo(expected.description());
             assertThat(key.type).isEqualTo(expected.type());
-            if (expected.equals(OracleConnectorConfig.SCHEMA_HISTORY) || expected.equals(CommonConnectorConfig.TOPIC_NAMING_STRATEGY)) {
+            if (expected.equals(CommonConnectorConfig.TOPIC_NAMING_STRATEGY)) {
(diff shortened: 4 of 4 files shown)


debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SqlServerConnectorTest.java
```diff
@@ -69,7 +69,7 @@ protected static void assertConfigDefIsValid(Connector connector, io.debezium.co
             assertThat(key.importance).isEqualTo(expected.importance());
             assertThat(key.documentation).isEqualTo(expected.description());
             assertThat(key.type).isEqualTo(expected.type());
-            if (expected.equals(SqlServerConnectorConfig.SCHEMA_HISTORY) || expected.equals(CommonConnectorConfig.TOPIC_NAMING_STRATEGY)) {
+            if (expected.equals(CommonConnectorConfig.TOPIC_NAMING_STRATEGY)) {
```
Naros, line 72: This change removes `SCHEMA_HISTORY` as an assertion use case, but it was retained in the Oracle equivalent, with a slight change to the assertion. Could we mirror the same logic here as we use in Oracle for consistency?

Naros (changes requested):
Hi @sundongkim-dev overall the code LGTM, just one test nit-pick.

sundongkim-dev:
Good catch — done. I've mirrored the Oracle approach in `SqlServerConnectorTest`: `SCHEMA_HISTORY` now has its own `else if` branch asserting `key.defaultValue` equals the expected `String` default, rather than being dropped from the check.

`shouldReturnConfigurationDefinition` passes, and I confirmed the assertion is live (temporarily corrupting the expected value fails it with `expected "...BOGUS" but was "io.debezium.storage.kafka.history.KafkaSchemaHistory"`).

vjuranek (commented):
Informix failures seem to be related.

sundongkim-dev:
@vjuranek you're right — the Informix/Db2/Ingres failures are directly caused by this change. Each of those `*ConnectorTest.assertConfigDefIsValid` special-cases `SCHEMA_HISTORY` by field and casts `key.defaultValue` to `Class`, which is now a `String` → `ClassCastException`. MySQL/MariaDB stay green because `BinlogConnectorConfigTest` keys off `type() == Type.CLASS` instead, so `SCHEMA_HISTORY` just falls through to the plain-value branch.

I've applied the equivalent fix to Oracle and SqlServer here, but Informix, Db2, and Ingres live in their own repos, so they need the same one-line adjustment there (ideally the type-driven guard, so it won't break again). Happy to open companion PRs against those three — how would you like to sequence it relative to this one?

vjuranek:
@sundongkim-dev please open the PRs, in most of the cases the build should pick the corresponding branch in remote repo

sundongkim-dev:
Opened the three companion PRs, each on the matching `dbz-870-schema-history-classtype` branch so their Cross Maven CI builds core from this PR:

- Informix — debezium/debezium-connector-informix#181
- Db2 — debezium/debezium-connector-db2#235
- Ingres — debezium/debezium-connector-ingres#20

Each applies the same one-line adjustment as the Oracle/SQL Server tests here: comparing the `SCHEMA_HISTORY` default value directly instead of casting it to `Class`.

Naros (commented):
I'm not convinced that the original report itself is technically a bug; instead, it's a malformed deployment.

Debezium's connectors default OOTB as both Kafka and File -aware, which is exactly why, if you look at MySQL, MariaDB, Oracle, SQL Server, Db2, Informix, and technically ANY historized connector, they have a hard dependency directly on the Kafka and the File storage modules (and if they don't, they should).

So the original reporter who opted not to include the Kafka storage on their classpath chose to deploy the connector in a way it was never intended to be deployed, at least prior to the creation of the storage modules, and still not based on the compile dependency chain used today.

The explicit hard dependency guarantees that when you install the connector plugin in a Connect cluster, all required dependencies are part of the bundle, so it's a single artifact. Similarly, the file storage module plays the same role for Debezium Server, Platform, and embedded use cases. You get the most common OOTB bits in one bundle.

So while moving the property type from CLASS to STRING addresses the eager load of the `KafkaSchemaHistory` class, the eager load of that should have succeeded anyway IMO. 

The only way someone could have used the MySQL connector with the embedded engine, without the Kafka storage module on the classpath, was to explicitly exclude that transitive dependency.

@vjuranek @jpechane WDYT?

vjuranek:
Personally, I can understand the intention here (I work on removing Kafka dependencies from the engine after all :-). What I don't fully understand is why Debezium search for Kafka schema history class when the user configures `schema.history.internal` to something else.

Naros:
Hi @vjuranek, I believe the issue with the reporter's failure is because the `debezium-storage-kafka` dependency wasn't on the classpath, so when the `ConfigDef#parseType` parsed the default value, which is defined as a `CLASS` type, which refers to the `KafkaSchemaHistory` class, this block throws a ClassNotFoundException:

From `ConfigDef#parseType`:
```java
                case CLASS:
                    if (value instanceof Class)
                        return value;
                    else if (value instanceof String) {
                        return Utils.loadClass(trimmed, Object.class); // <-- this throws the CNFE
                    } else
                        throw new ConfigException(name, value, "Expected a Class instance or class name.");
```

What @sundongkim-dev is doing is changing this property to `STRING` type, which means that the `ConfigDef` won't attempt to load the class during the `ConfigDef#parseType` on the default value.

So I get the intent to at least remove Kafka dependencies on the embedded engine, but those dependencies would still be enforced at the connector module level regardless, right?

vjuranek:
> Hi @vjuranek, I believe the issue with the reporter's failure is because the `debezium-storage-kafka` dependency wasn't on the classpath, so when the `ConfigDef#parseType` parsed the default value, which is defined as a `CLASS` type, which refers to the `KafkaSchemaHistory` class, this block throws a ClassNotFoundException:

Oh, I see, thanks for explanation @Naros !

> So I get the intent to at least remove Kafka dependencies on the embedded engine, but those dependencies would still be enforced at the connector module level regardless, right?

I don't think so, if the user provides another implementation of offset store and schema history, `debezium-storage-kafka` shouldn't be needed AFAICT, so this PR makes sense to me (of course under assumption I'm not wrong here, that `debezium-storage-kafka` is not needed).

Naros:
Oracle test failures fixed in #7722

Naros:
The Oracle, Apicurio, JDBC failures are unrelated, and fixed in recent changes on `main`.

Naros (approved):

vjuranek (approved):

vjuranek:

PR #7660, opened 2026-07-14
debezium/dbz#2227 Persist purged GTID set as offset floor on no-snapshot start

akanimesh7:
Fixes debezium/dbz#2227.

## Problem
A binlog connector started in a stream-only mode (no schema or data snapshot) against a MySQL server with a non-empty `gtid_purged` starts and streams fine, but permanently fails on the first task restart with `... is no longer available on the server`.

On such a first attach, `BinlogStreamingChangeEventSource` sets the binlog client's GTID set to the purged set (so the server skips those GTIDs) but seeds the in-process accumulator empty, so committed offsets start at `:P+1` — without the `:1-` floor. On restart, `isBinlogPositionAvailable` computes `server − offset`, sees the purged `:1-P` range as needed-and-purged, and fails the task, even though those GTIDs were intentionally skipped at first attach.

Every snapshot-running mode avoids this because it seeds the offset via `setCompletedGtidSet(Executed_Gtid_Set)`, which carries the `:1-` floor.

## Fix
Seed the offset (`setCompletedGtidSet`) and the accumulator (`initializeGtidSet`) with the purged GTID set, mirroring the stored-offset branch just above and the snapshot path. Committed offsets then carry the floor and restarts validate correctly. Also corrects an inverted log message in `isBinlogPositionAvailable` (it claimed the server "does not contain" the connector's set inside the branch where it does).

## Testing
Reproduced end-to-end against MySQL 8.0 with GTIDs enabled and a non-empty `gtid_purged`, using `snapshot.mode=configuration_based` stream-only:
- Before: the committed offset is `:P+1-…` and the task fails on the first restart.
- After: the committed offset carries the `:1-` floor and the task survives restart and continues streaming.

Note: `ZZZBinlogGtidSetIT#shouldProcessPurgedGtidSet` (`@FixFor("DBZ-1184")`) asserts the previous behavior (offset must not start at `:1`) and will need updating; a dedicated restart regression test will follow.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
@@ -285,9 +285,14 @@ public abstract class BinlogStreamingChangeEventSource<P extends BinlogPartition
                 else {
                     LOGGER.info("No GTID stored in the offset, but there is non-empty purged GTID set. Registering binlog reader with purged GTID set: '{}'",
                             purgedServerGtidSet.toString());
-                    client.setGtidSet(purgedServerGtidSet.toString());
-                    // We don't have stored any GTID in the offset, so start from empty GTID set.
-                    initializeGtidSet("");
+                    final String purgedServerGtidSetStr = purgedServerGtidSet.toString();
+                    client.setGtidSet(purgedServerGtidSetStr);
+                    // The purged GTID set is the effective starting floor for a no-snapshot start: the connector
+                    // intentionally skips everything at or below it. Record it in the offset and seed the accumulator
+                    // with it (mirroring the stored-offset branch above and the snapshot path's setCompletedGtidSet),
+                    // so the committed offset carries this floor and subsequent restarts validate it correctly.
+                    effectiveOffsetContext.setCompletedGtidSet(purgedServerGtidSetStr);
+                    initializeGtidSet(purgedServerGtidSetStr);
                 }
             }
         }
--- a/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/zzz/ZZZBinlogGtidSetIT.java
+++ b/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/zzz/ZZZBinlogGtidSetIT.java
@@ -114,15 +114,16 @@ public abstract class ZZZBinlogGtidSetIT<C extends SourceConnector> extends Abst
         // Check that all records are valid, can be serialized and deserialized ...
         records.forEach(this::validate);
 
-        // Check that records have GTID that does not contain purged interval
+        // Records carry the purged GTID set as the offset's starting floor.
         records.recordsForTopic(ro_database.topicForTable("customers")).forEach(record -> {
             final String gtids = (String) record.sourceOffset().get("gtids");
 
-            // the format is <uuid:<start-tx>-<end-tx>; we don't expect any offsets with start tx = 1 due to the flush
+            // format is <uuid>:<start-tx>-<end-tx>; the offset now starts at tx 1 because the connector records
+            // the purged set as its floor (those GTIDs were intentionally skipped, not lost).
             final Pattern p = Pattern.compile(".*:(.*)-.*");
             final Matcher m = p.matcher(gtids);
             m.matches();
-            assertThat(m.group(1)).isNotEqualTo("1");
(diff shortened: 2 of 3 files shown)


akanimesh7:
@Naros @jpechane 
Can you review this one ?

Naros:
Hi @akanimesh7 do you plan to add the restart regression test to this PR?

akanimesh7:
Yes — added in `8916fea`. `ZZZBinlogGtidSetIT#shouldResumeStreamingAfterRestartWithPurgedGtidSet` starts a no-snapshot connector against a server with a non-empty purged GTID set, stops it, inserts rows, and asserts it resumes streaming after restart instead of failing with "… is no longer available on the server". Verified locally against MySQL 8.0 with GTIDs enabled.

akanimesh7:
@Naros Can you check once more

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
```diff
@@ -285,9 +285,14 @@ public void execute(ChangeEventSourceContext context, P partition, O offsetConte
...
-                    client.setGtidSet(purgedServerGtidSet.toString());
-                    // We don't have stored any GTID in the offset, so start from empty GTID set.
-                    initializeGtidSet("");
+                    final String purgedServerGtidSetStr = purgedServerGtidSet.toString();
+                    client.setGtidSet(purgedServerGtidSetStr);
+                    // The purged GTID set is the effective starting floor for a no-snapshot start: the connector
+                    // intentionally skips everything at or below it. Record it in the offset and seed the accumulator
+                    // with it (mirroring the stored-offset branch above and the snapshot path's setCompletedGtidSet),
+                    // so the committed offset carries this floor and subsequent restarts validate it correctly.
+                    effectiveOffsetContext.setCompletedGtidSet(purgedServerGtidSetStr);
```
Naros, line 294: Should we consider filtering the purged GTID set when a GTID source filter is enabled? If we do, then could we have a test that covers this use case?

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/zzz/ZZZBinlogGtidSetIT.java
```diff
@@ -114,20 +114,65 @@ public void shouldProcessPurgedGtidSet() throws SQLException, InterruptedExcepti
...
             final String gtids = (String) record.sourceOffset().get("gtids");
 
-            // the format is <uuid:<start-tx>-<end-tx>; we don't expect any offsets with start tx = 1 due to the flush
+            // format is <uuid>:<start-tx>-<end-tx>; the offset now starts at tx 1 because the connector records
+            // the purged set as its floor (those GTIDs were intentionally skipped, not lost).
             final Pattern p = Pattern.compile(".*:(.*)-.*");
             final Matcher m = p.matcher(gtids);
             m.matches();
-            assertThat(m.group(1)).isNotEqualTo("1");
+            assertThat(m.group(1)).isEqualTo("1");
```
Naros, line 126: Since we are changing the logic of this test method here, could you modify the `@FixFor` annotation on this test method as:
```
@FixFor({ "DBZ-1184", "debezium/dbz#2227" })
```

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/BinlogStreamingChangeEventSource.java
akanimesh7: `filterGtidSet` (the stored-offset path) already does this — it filters the offset and available set, but unions the purged set unfiltered. For a null offset that reduces to just `purgedServerGtidSet`, so this branch stays consistent with it.

Filtering it would actually break things: `client.setGtidSet()` tells MySQL what we already have, so any purged GTID we don't claim, the server tries to resend and fails with 1236.

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/zzz/ZZZBinlogGtidSetIT.java
akanimesh7: Done in `43c653e`.

akanimesh7:
@Naros Could you check again ?

Naros (approved):

Naros:
Applied, thanks @akanimesh7 :tada:

PR #1477, opened 2020-05-06
DBZ-2045 Refresh postgres schema when a new enum value is noticed

bradengroom:

PR #7668, opened 2026-07-16
debezium/dbz#304 Refresh cached enum metadata when a value is added at streaming time

### LINKED ISSUE #304: Postgres schema does not update when altering an enum type [DBZ-2045]
Migrated from [DBZ-2045](https://issues.redhat.com/browse/DBZ-2045)

Here is a failing test demonstrating the bug

{code:java}
    @Test
    public void shouldUpdateSchemaWhenEnumOptionsChange() throws Exception {
        TestHelper.execute("CREATE TYPE test_type AS ENUM ('V1');");
        TestHelper.execute("CREATE TABLE enum_table (pk SERIAL, value test_type NOT NULL, PRIMARY KEY (pk));");
        startConnector(config -> config
                .with(PostgresConnectorConfig.INCLUDE_UNKNOWN_DATATYPES, true)
                .with(PostgresConnectorConfig.SNAPSHOT_MODE, SnapshotMode.NEVER)
                .with("column.propagate.source.type", "public.enum_table.value")
                .with(PostgresConnectorConfig.TABLE_WHITELIST, "public.enum_table"), false);

        waitForStreamingToStart();

        // Alter the enum type to include a new option
        try (Connection connection = TestHelper.create().connection()) {
            Statement statement = connection.createStatement();
            statement.execute("ALTER TYPE test_type ADD VALUE 'V2'");
        }
        catch (Exception e) {
            throw new RuntimeException(e);
        }

        consumer = testConsumer(1);
        executeAndWait("INSERT INTO enum_table (value) VALUES ('V2');");

        SourceRecord rec = assertRecordInserted("public.enum_table", PK_FIELD, 1);
        assertSourceInfo(rec, "postgres", "public", "enum_table");

        List<SchemaAndValueField> expected = Arrays.asList(
                new SchemaAndValueField(PK_FIELD, Schema.INT32_SCHEMA, 1),
                new SchemaAndValueField("value", Enum.builder("V1,V2")
                        .parameter(TestHelper.TYPE_NAME_PARAMETER_KEY, "TEST_TYPE")
                        .parameter(TestHelper.TYPE_LENGTH_PARAMETER_KEY, String.valueOf(Integer.MAX_VALUE))
                        .parameter(TestHelper.TYPE_SCALE_PARAMETER_KEY, "0")
                        .build(), "V2"));

        assertRecordSchemaAndValues(expected, rec, Envelope.FieldName.AFTER);
        assertThat(consumer.isEmpty()).isTrue();
    }
{code}

sundongkim-dev:
Fixes debezium/dbz#304

## Problem

When an `ALTER TYPE <enum> ADD VALUE '<x>'` is executed while the connector is streaming, the enum column's type OID does not change. As a result the change is invisible to `PostgresChangeRecordEmitter.schemaChanged()` (which only reacts to column count / OID changes) and, for `pgoutput`, no new `RELATION` message is emitted either. The cached `PostgresType` therefore keeps its stale list of allowed values, and the emitted record schema advertises an incomplete enum:

```
allowed = "V1"        // expected "V1,V2"
```

The value itself is still streamed correctly as a string, but Avro / schema-registry consumers reject records carrying a value that is absent from the schema's `allowed` parameter.

This affects **both** logical-decoding plugins:
- `decoderbufs`: `schemaChanged()` never returns `true` for the added value.
- `pgoutput`: `shouldSchemaBeSynchronized()` is `false` (it relies on `RELATION` messages), so the schema is never re-synchronized at all — and `ALTER TYPE ... ADD VALUE` produces no `RELATION` message.

## Fix

- `TypeRegistry.refresh(int oid)` re-reads a type from the database and evicts both the oid-keyed and name-keyed cache entries (the latter is required because `addType()` keeps the first mapping for a given name, and the `pgoutput` decoder resolves column types by name on every row — without evicting it the lookup would stay stale and re-trigger a refresh on every subsequent row).
- `PostgresChangeRecordEmitter` detects, for either plugin, an enum column whose incoming value is missing from the cached type and refreshes that type before the table schema is rebuilt. This runs regardless of `shouldSchemaBeSynchronized()` so `pgoutput` is covered as well.

The behaviour is best-effort, as discussed on the original PR debezium/debezium#1477: the schema is updated when the first row carrying the new value arrives, and removed enum values are not detected.

## Testing

- Added `RecordsStreamProducerIT#shouldRefreshSchemaWhenEnumValueAddedViaAlterType`, which fails before the change (`allowed="V1"`) and passes after (`allowed="V1,V2"`).
- Verified on a live PostgreSQL 17 database with **both** `decoderbufs` and `pgoutput`; the existing enum and streaming tests pass on both.

## Notes / scope

Limited to scalar enum columns (the reported case). Enum arrays (`enum[]`) and domain types built on top of an enum are out of scope and could be addressed as follow-ups.

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresChangeRecordEmitter.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresChangeRecordEmitter.java
@@ -136,20 +136,60 @@ public class PostgresChangeRecordEmitter extends RelationalChangeRecordEmitter<P
     }
 
     private DataCollectionSchema synchronizeTableSchema(DataCollectionSchema tableSchema) {
-        if (getOperation() == Operation.DELETE || !message.shouldSchemaBeSynchronized()) {
+        if (getOperation() == Operation.DELETE) {
             return tableSchema;
         }
         final TableId tableId = (TableId) tableSchema.id();
-        final Table table = schema.tableFor(tableId);
         final List<ReplicationMessage.Column> columns = message.getNewTupleList();
+        // dbz#304 ALTER TYPE ... ADD VALUE keeps the column's type OID, so the added value is invisible
+        // both to schemaChanged() and to the pgoutput RELATION messages. Refresh the stale enum metadata
+        // regardless of the decoder's shouldSchemaBeSynchronized() so the rebuilt schema exposes it.
+        final boolean enumRefreshed = refreshStaleEnumTypes(columns);
+        if (!message.shouldSchemaBeSynchronized()) {
+            if (enumRefreshed) {
+                refreshTableFromDatabase(tableId);
+                return schema.schemaFor(tableId);
+            }
+            return tableSchema;
+        }
+        final Table table = schema.tableFor(tableId);
         // check if we need to refresh our local schema due to DB schema changes for this table
-        if (schemaChanged(columns, table)) {
+        if (enumRefreshed || schemaChanged(columns, table)) {
             // Refresh the schema so we get information about primary keys
             refreshTableFromDatabase(tableId);
             // Update the schema with metadata coming from decoder message
             schema.refresh(tableFromFromMessage(columns, schema.tableFor(tableId)));
         }
         return schema.schemaFor(tableId);
+    }
+
+    /**
+     * Detects enum columns whose incoming value is missing from the cached enum type metadata
+     * (the result of an {@code ALTER TYPE ... ADD VALUE}) and refreshes those types from the
+     * database so the rebuilt schema exposes the up-to-date list of allowed values.
+     *
+     * @param columns the columns of the current replication message
+     * @return {@code true} if at least one enum type was refreshed
+     */
+    private boolean refreshStaleEnumTypes(List<ReplicationMessage.Column> columns) {
+        if (columns == null) {
+            return false;
+        }
+        boolean refreshed = false;
+        for (ReplicationMessage.Column column : columns) {
+            final PostgresType type = column.getType();
+            if (!type.isEnumType()) {
+                continue;
+            }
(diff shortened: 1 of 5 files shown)


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresChangeRecordEmitter.java
```diff
@@ -152,6 +163,35 @@ private DataCollectionSchema synchronizeTableSchema(DataCollectionSchema tableSc
...
+        }
+        boolean refreshed = false;
+        for (ReplicationMessage.Column column : columns) {
+            final PostgresType type = column.getType();
+            if (!type.isEnumType()) {
+                continue;
+            }
+            final Object value = column.getValue(() -> connection.connection().unwrap(BaseConnection.class),
+                    connectorConfig.includeUnknownDatatypes());
+            if (value instanceof String && !type.getEnumValues().contains(value)) {
```
Naros, line 186: For wide tables and, most importantly, tables with multiple enum types, this can become a hot path using `List#contains`, which is `O(n)`. 

AFAIK PostgreSQL enforces uniqueness for enum values, so technically `PostgresType` could use a `LinkedHashSet` rather than `List` for the enum values, and we retain the same behavior as we did previously, while gaining `O(1)` on the `contains` call here.

@sundongkim-dev WDYT?

sundongkim-dev:
Good call, thanks @Naros. Switched to a `LinkedHashSet` — PostgreSQL enforces label uniqueness, so we get `O(1)` on `contains()` while insertion order (`enumsortorder`) is preserved, leaving the `allowed` parameter unchanged. Pushed as a follow-up commit.

Naros:
The StarRocks test failure is fixed in https://github.com/debezium/debezium/pull/7684

Naros (approved):

Naros:

PR #7669, opened 2026-07-16
debezium/dbz#2235 Map MySQL and MariaDB signed TINYINT columns to INT8

minleejae:
Fixes debezium/dbz#2235

## Description

MySQL/MariaDB signed `TINYINT` is an 8-bit type (-128 to 127), but the ANTLR DDL parsers folded `TINYINT`/`INT1` into `java.sql.Types.SMALLINT`, so it was emitted as `INT16` — the only integer mapping not using the smallest Kafka Connect type that holds the source range.

| Column type | Range | Before | After |
|---|---|---|---|
| `TINYINT` | -128 to 127 | `INT16` | **`INT8`** |
| `TINYINT UNSIGNED` | 0 to 255 | `INT16` | `INT16` |
| `MEDIUMINT UNSIGNED` | 0 to 16777215 | `INT32` | `INT32` |

Changes:
- Register `TINYINT`/`INT1` as `Types.TINYINT` in `MySqlAntlrDdlParser`, `MySqlPtAntlrDdlParser` and `MariaDbAntlrDdlParser`, so signed TINYINT takes the existing core `Types.TINYINT → INT8` path.
- Add explicit `TINYINT UNSIGNED → INT16` and `MEDIUMINT UNSIGNED → INT32` rules in `BinlogValueConverters`, keeping the widen-by-one-type behavior for unsigned columns now that the jdbcType no longer decides these cases.
- `convertTinyInt` now produces `Byte` (it previously delegated to `convertSmallInt` and produced `Short`, contradicting its own `INT8` schema), and the decimal-string default value handling (DBZ-5134) is extended to the tinyint converter.

`TINYINT(1)`-as-boolean handling (`TinyIntOneToBooleanConverter`, `JdbcSinkDataTypesConverter`) is unaffected. Tables captured before this fix keep their recorded schema-history model until a re-snapshot or a subsequent DDL statement for the table.

Verified with the unit suites of the common/binlog/mysql/mariadb modules and the docker-based ITs (tinyint, unsigned integer, default value, charset, jdbc sink) against MySQL 8.2 and MariaDB 11.4.3; `MySqlTinyIntIT` gains -128/-1/127 boundary coverage.


documentation/modules/ROOT/pages/transformations/gtid-watermark-filter.adoc
```diff
@@ -0,0 +1,99 @@
+// Category: debezium-using
```
vjuranek: This was probably unintentionally added?

vjuranek (commented):
Contains unrelated doc changes, otherwise LGTM

documentation/modules/ROOT/pages/transformations/gtid-watermark-filter.adoc
minleejae: Thanks! This was added unintentionally, and I’ve removed it. [bdf1e49](https://github.com/debezium/debezium/pull/7669/commits/bdf1e4913597ea6e314026c06f32cb373c5fe93e)

vjuranek (approved):
LGTM

Naros (approved):
I added the backward-incompatible label mainly to highlight this in breaking changes for those who use schema registry or downstream consumers to be aware that while upgrading may not create issues, a schema change introduces type narrowing, which can.

Naros:

PR #7628, opened 2026-07-09
debezium/dbz#815 Add StarRocks JDBC sink support

minleejae:
### Related issues

- Fixes https://github.com/debezium/dbz/issues/815 — Introduce dialect support for StarRocks database
- Fixes https://github.com/debezium/dbz/issues/979 — Support setting target schema for JDBC sink with StarRocks target

### Summary

Adds initial StarRocks support to the Debezium JDBC sink connector, following the structure of the SingleStore support in #7566.

- **Dialect & automatic detection**: StarRocks speaks the MySQL wire protocol but is not DDL-compatible with MySQL, so the Debezium dialect extends `GeneralDatabaseDialect` while a minimal Hibernate `StarRocksDialect` (mirroring Hibernate's `TiDBDialect` precedent) pins the diverging column types. The bundled `com.starrocks:starrocks-connector-j` driver reports the `StarRocks` product name, which a registered Hibernate `DialectResolver` resolves automatically — no `hibernate.dialect` configuration is needed.
- **Upserts & table creation**: StarRocks executes `INSERT` statements against PRIMARY KEY tables as upserts (it has no `ON DUPLICATE KEY UPDATE`), so `insert.mode=upsert` generates plain inserts. Auto-created tables emit `PRIMARY KEY (...) DISTRIBUTED BY HASH (...)` clauses from the record key fields, with `NULL/NOT NULL` placed before `DEFAULT` as StarRocks requires.
- **Type mappings** (verified against StarRocks 4.1.1): time logical types are stored as `HH:mm:ss.ffffff` strings (StarRocks has no TIME type; MySQL's ±838h range keeps its notation), `DATETIME` retains microsecond precision, `VARCHAR` character lengths are scaled ×4 for StarRocks' byte semantics (wide `CHAR` promotes to `VARCHAR`), `BIGINT UNSIGNED` maps to `LARGEINT` preserving the full unsigned range, `DECIMAL` precision is capped at 38, JSON/MAP map to `JSON`, and `BOOLEAN` maps to native `boolean`.
- **Metadata handling**: JDBC metadata lookups are qualified with the current database (the driver otherwise iterates every database and StarRocks errors on non-matching ones), and `BYTES` binding converts `ByteBuffer` values the driver cannot bind.
[...]

PR #7677, opened 2026-07-17
debezium/dbz#815 Document StarRocks update limitations

minleejae:
This follow-up documents that auto-created StarRocks tables without key fields are Duplicate Key tables and do not support UPDATE. It also updates the StarRocks integration test to verify the expected behavior.

Tested with both the legacy and shared sink frameworks.

### DIFF
--- a/debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/starrocks/JdbcSinkInsertModeIT.java
+++ b/debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/starrocks/JdbcSinkInsertModeIT.java
@@ -5,12 +5,25 @@
  */
 package io.debezium.connector.jdbc.integration.starrocks;
 
+import static org.junit.jupiter.api.Assertions.fail;
+
+import java.util.Map;
+
 import org.junit.jupiter.api.Tag;
 import org.junit.jupiter.api.extension.ExtendWith;
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.ArgumentsSource;
 
+import io.debezium.connector.jdbc.JdbcKafkaSinkRecord;
+import io.debezium.connector.jdbc.JdbcSinkConnectorConfig;
+import io.debezium.connector.jdbc.JdbcSinkConnectorConfig.InsertMode;
+import io.debezium.connector.jdbc.JdbcSinkConnectorConfig.SchemaEvolutionMode;
 import io.debezium.connector.jdbc.integration.AbstractJdbcSinkInsertModeTest;
 import io.debezium.connector.jdbc.junit.jupiter.Sink;
+import io.debezium.connector.jdbc.junit.jupiter.SinkRecordFactoryArgumentsProvider;
 import io.debezium.connector.jdbc.junit.jupiter.StarRocksSinkDatabaseContextProvider;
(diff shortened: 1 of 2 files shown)


minleejae:
Hi @Naros 
Follow-up to #7628: after it was merged, I noticed that the StarRocks CI job had failed. The failing test used `insert.mode=update` with `primary.key.mode=none`, which causes the connector to auto-create a Duplicate Key table. StarRocks does not support UPDATE statements on Duplicate Key tables, so I updated the test to verify the expected      
behavior and documented this limitation in this PR

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/starrocks/JdbcSinkInsertModeIT.java
```diff
@@ -24,4 +37,35 @@ public class JdbcSinkInsertModeIT extends AbstractJdbcSinkInsertModeTest {
     public JdbcSinkInsertModeIT(Sink sink) {
         super(sink);
     }
+
+    @Override
+    @ParameterizedTest
+    @ArgumentsSource(SinkRecordFactoryArgumentsProvider.class)
```
vjuranek, line 43: I think @Naros already addressed this in https://github.com/debezium/debezium/pull/7684

vjuranek (commented):
Test was fixed in another PR, doc changes LGTM.

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/starrocks/JdbcSinkInsertModeIT.java
Naros: I did, so I think @minleejae we can drop this test method entirely.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -85,6 +85,9 @@ StarRocks targets support idempotent writes through the same xref:jdbc-property-
 StarRocks executes `INSERT` statements against PRIMARY KEY tables as upserts, so the connector generates plain `INSERT` statements; `INSERT ... ON DUPLICATE KEY UPDATE ...` is not supported by StarRocks.
 Upsert and delete operations require the destination table to be a StarRocks PRIMARY KEY table.
 Tables that the connector creates automatically from events with key fields are always PRIMARY KEY tables.
+For events without key fields, the connector omits the `PRIMARY KEY` clause, and StarRocks creates its default DUPLICATE KEY table.
+DUPLICATE KEY tables retain rows with duplicate values and support append-only writes, but do not support modifying existing rows.
+As a result, xref:jdbc-property-insert-mode[`insert.mode=update`] cannot be used with xref:jdbc-property-primary-key-mode[`primary.key.mode=none`] when the connector creates the destination table.
```
Naros, line 90: Doesn't it make sense that if `insert.mode` is set to `update` and there is no primary key mode set on the sink, that should be considered an invalid configuration for all dialects? I don't quite understand how else you would be able to update a row without a PK.

minleejae:
Thanks @vjuranek and @Naros for the feedback and clarification. Since #7684 already handles the StarRocks test behavior, the additional test here is no longer needed.

The remaining documentation change raises the broader question of whether `insert.mode=update` without a primary key should be rejected for all dialects.

With that in mind, I’ll close this PR.

PR #7678, opened 2026-07-17
debezium/dbz#2220 Propagate the real pgvector dimension from the PostgreSQL source

sundongkim-dev:
Fixes debezium/dbz#2220

## Description
pgvector types (`vector`, `halfvec`, `sparsevec`) get their OIDs assigned dynamically by the extension, so they are absent from pgjdbc's hard-coded type table and `TypeInfo.getPrecision()` collapses their length to `Integer.MAX_VALUE`. With `column.propagate.source.type` enabled the propagated `__debezium.source.column.length` was therefore `2147483647` instead of the real dimension (the JDBC sink symptom `vector(2147483647)` was previously guarded sink-side in #7648). The real dimension is carried verbatim in `atttypmod` (no offset encoding), so it is recoverable.

This recovers it on every capture path that lost it:
- **pgoutput streaming** (`ColumnMetaData`): take the dimension straight from the attribute modifier for vector types instead of delegating to `getPrecision()`, and emit the type name as `vector(3)` (single modifier) rather than `vector(3,0)`.
- **snapshot** (`PostgresConnection`): the driver reports `COLUMN_SIZE` as `Integer.MAX_VALUE`, so read the dimension from `pg_attribute.atttypmod` via a catalog lookup.
- **dimensionless `vector`** (no modifier): all paths — including decoderbufs, which was already correct for a declared dimension — now report no length rather than `Integer.MAX_VALUE`.

Verified with a unit test (`ColumnMetaDataTest`), an integration test (`VectorDatabaseIT#shouldPropagateVectorDimensions`, run under both pgoutput and decoderbufs, covering snapshot + streaming, asserting the propagated dimension for `vector(3)`/`halfvec(3)`/`sparsevec(3000)` and that a bare `vector` propagates no length), and a full source→sink end-to-end run where the JDBC sink auto-creates `vector(3)`/`halfvec(3)`/`sparsevec(3000)` columns.

### Notes
- The snapshot catalog lookup runs while iterating `DatabaseMetaData.getColumns()`; pgjdbc materializes that ResultSet client-side, so the nested query on the same connection is safe (and it runs inside the snapshot transaction for a consistent read).
- `vector`/`halfvec`/`sparsevec` are classified by type name, matching how `TypeRegistry` already resolves their OIDs.
- Out of scope (still report no dimension): arrays of vectors (`vector(3)[]`) and domains over vector.

## PR Checklist
- [x] Minimal changes to code not directly related to your change

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresType.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresType.java
@@ -55,6 +55,16 @@ public class PostgresType {
     }
 
     /**
+     * @return true if this is a pgvector type ({@code vector}, {@code halfvec} or {@code sparsevec}),
+     *         whose dimension is carried verbatim in {@code atttypmod}
+     */
+    public boolean isVector() {
+        return TypeRegistry.TYPE_NAME_VECTOR.equals(name)
+                || TypeRegistry.TYPE_NAME_HALF_VECTOR.equals(name)
+                || TypeRegistry.TYPE_NAME_SPARSE_VECTOR.equals(name);
+    }
+
+    /**
      * The type system allows for the creation of user defined types (UDTs) which can be based
      * on any existing type.  When a type does not extend another type, it is considered to be
      * a base or root type in the type hierarchy.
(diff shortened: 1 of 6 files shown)


Naros (approved):

Naros:

PR #7679, opened 2026-07-17
debezium/dbz#2100 Bind PostgreSQL point to its native type without PostGIS

sundongkim-dev:
Fixes debezium/dbz#2100, case 9.

## Description

Case 9 of the `dbz#2100` sink type-preservation audit: a built-in PostgreSQL `point` column cannot be
replicated to a sink that does not have the PostGIS extension installed. The JDBC sink's `PointType`
always emits `cast(<schema>.ST_GeomFromWKB(?, ?) as point)`, so the sink fails with:

```
ERROR: function public.st_geomfromwkb(bytea, integer) does not exist
```

A built-in `point` needs no PostGIS at all — its coordinates are already carried on the
`io.debezium.data.geometry.Point` struct (`x`/`y` FLOAT64 fields). This aligns `PointType` with the
sibling `CircleType`/`LineType` added in `dbz#2135`: it now extends `AbstractType`, binds
`cast(? as point)`, and reconstructs the native text `(x,y)` directly from the struct. Because the
`Point` logical type is used only for the built-in `point` (genuine PostGIS geometry uses the
`Geometry` logical type), no source-type gating is required.

Verified end to end (source PostgreSQL `point` → pgoutput → Kafka → JDBC sink) against a plain,
non-PostGIS sink: the column is auto-created as native `point` and values round-trip exactly, including
updates. Reverting the fix reproduces the `st_geomfromwkb does not exist` failure on the same sink.

## PR Checklist

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PointType.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PointType.java
@@ -5,29 +5,26 @@
  */
 package io.debezium.connector.jdbc.dialect.postgres;
 
-import org.apache.kafka.connect.data.Schema;
+import java.util.List;
 
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.Struct;
+
+import io.debezium.connector.jdbc.type.AbstractType;
 import io.debezium.connector.jdbc.type.JdbcType;
 import io.debezium.data.geometry.Point;
 import io.debezium.sink.column.ColumnDescriptor;
+import io.debezium.sink.valuebinding.ValueBindDescriptor;
 
 /**
- * An implementation of {@link JdbcType} for {@code io.debezium.data.geometry.Point} types.
+ * An implementation of {@link JdbcType} for {@code io.debezium.data.geometry.Point} values, binding
+ * them to the native PostgreSQL {@code point} type without requiring PostGIS.
  *
  * @author Chris Cranford
  */
-class PointType extends GeometryType {
+class PointType extends AbstractType {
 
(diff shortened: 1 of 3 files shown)


Naros (commented):
Hi @sundongkim-dev overall the code LGTM. I do have one question wrt UNNEST batches.  Have you checked whether or not the latest changes in this PR for `Point`, as well as the geometry changes recently introduced for `Leg`, `Circle`, etc are properly bound when using the UNNEST dialect feature?

sundongkim-dev:
Good question, @Naros — I checked and it's handled correctly, and I've added a test to lock it down.

`point`, like the `box`/`circle`/`line`/`lseg`/`path`/`polygon` types added in dbz#2135, has a `STRUCT` Connect schema (`Point.builder()` → `SchemaBuilder.struct()`). So it's caught by the existing `hasUnnestUnsupportedField` guard in `getBatchInsertStatement()`, which returns `Optional.empty()` and drops the whole batch to the row-wise path. That path is where `PointType`'s `getQueryBinding()` (`cast(? as point)`) and `bind()` (the `(x,y)` literal) apply — the UNNEST `createArrayOf` path never sees the raw `Struct`.

The one gap was test coverage: `testPointRoundTrip` only feeds a single record, which short-circuits UNNEST (`records.size() == 1`), so it never actually exercised the batch fallback for `point`. I added `testPointBatchRoundTrip` (mirroring `testBoxBatchRoundTrip`) — a >1 batch under both Standard and UNNEST modes. Full `JdbcSinkGeometricTypesIT` is green (36/36), verified against a real PostgreSQL 17 sink.

Naros:
The StarRocks test failure is unrelated; it's fixed on `main`.

Naros (approved):

Naros:

PR #7680, opened 2026-07-17
debezium/dbz#2119 Preserve structured temporal fidelity across JDBC pipelines

minleejae:
Follow-up to #7554.

## Summary

- Represent structured fractional seconds as picoseconds and add precision and duration-kind schema metadata.
- Validate target precision, duration semantics, zoned timestamp metadata, and temporal ranges before JDBC writes, with explicit `truncate`, `round`, `fail`, and `saturate` policies.
- Model target support for date and timestamp infinity and zero dates, so PostgreSQL and MySQL special values are preserved only by compatible targets.
- Add explicit CockroachDB and SingleStore temporal bounds, including CockroachDB finite timestamp handling and SingleStore `DATE`, `DATETIME`, and `TIMESTAMP` limits.
- Preserve StarRocks `StructuredTime` and `StructuredZonedTime` values through p12 and retain the original offset by using string-backed target columns.
- Preserve Db2 and Db2i timestamp precision through p12 using text binding and a server-side cast.

## Testing

- Focused source converter and structured temporal tests, including the CockroachDB, SingleStore, and StarRocks dialect tests.
- Full JDBC unit and architecture tests (378 tests).
- Formatting, checkstyle, compilation, and verification for all changed modules.
- Added integration coverage for CockroachDB, SingleStore, StarRocks, and Db2. CockroachDB passed locally; StarRocks container startup timed out, while the SingleStore and Db2 images have no arm64 manifest, so those tests are left to CI.

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/relational/TableSchemaBuilder.java
+++ b/debezium-connector-common/src/main/java/io/debezium/relational/TableSchemaBuilder.java
@@ -36,6 +36,7 @@ import io.debezium.schema.FieldNameSelector.FieldNamer;
 import io.debezium.schema.SchemaFactory;
 import io.debezium.schema.SchemaNameAdjuster;
 import io.debezium.spi.topic.TopicNamingStrategy;
+import io.debezium.time.StructuredTemporal;
 import io.debezium.util.Loggings;
 
 /**
@@ -411,10 +412,13 @@ public class TableSchemaBuilder {
     protected void addField(SchemaBuilder builder, Table table, Column column, ColumnMapper mapper) {
         final Object defaultValue = parseDefaultValue(table.id(), column);
 
-        final SchemaBuilder fieldBuilder = customConverterRegistry.registerConverterFor(table.id(), column, defaultValue)
+        SchemaBuilder fieldBuilder = customConverterRegistry.registerConverterFor(table.id(), column, defaultValue)
                 .orElse(valueConverterProvider.schemaBuilder(column));
 
         if (fieldBuilder != null) {
+            if (mapper != null || column.comment() != null) {
+                fieldBuilder = StructuredTemporal.qualifySchemaBuilderWithSourceColumn(fieldBuilder, column.name());
+            }
             if (mapper != null) {
                 // Let the mapper add properties to the schema ...
                 mapper.alterFieldSchema(column, fieldBuilder);
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/StructuredTemporalLiteral.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/StructuredTemporalLiteral.java
@@ -7,45 +7,67 @@ package io.debezium.connector.jdbc.dialect.mysql;
 
 import java.math.BigDecimal;
 import java.math.RoundingMode;
-import java.time.LocalDateTime;
 
 import org.apache.kafka.connect.data.Struct;
 import org.apache.kafka.connect.errors.ConnectException;
 
 import io.debezium.connector.jdbc.JdbcSinkConnectorConfig.TemporalPrecisionLossHandlingMode;
+import io.debezium.connector.jdbc.JdbcSinkConnectorConfig.TemporalRangeLossHandlingMode;
 import io.debezium.connector.jdbc.type.debezium.StructuredTemporalPreflightValidator;
 import io.debezium.connector.jdbc.type.debezium.StructuredTemporalSupport;
+import io.debezium.connector.jdbc.type.debezium.TemporalRange;
+import io.debezium.connector.jdbc.type.debezium.TemporalRange.Boundary;
 import io.debezium.time.StructuredTemporal;
 
 final class StructuredTemporalLiteral {
 
     static String date(Struct value) {
-        requireFinite(value);
+        return date(value, TemporalRange.unbounded(), TemporalRangeLossHandlingMode.FAIL, "target temporal type");
+    }
+
+    static String date(Struct value, TemporalRange range, TemporalRangeLossHandlingMode handlingMode,
+                       String targetDescription) {
+        if (StructuredTemporal.isFinite(value) && hasZeroDate(value)) {
+            return formatDate(
+                    value.getInt32(StructuredTemporal.YEAR_FIELD),
+                    value.getInt8(StructuredTemporal.MONTH_FIELD),
+                    value.getInt8(StructuredTemporal.DAY_FIELD));
+        }
+        final Boundary adjusted = StructuredTemporalSupport.adjustDate(value, range, handlingMode, targetDescription);
+        return formatDate(adjusted.year(), adjusted.month(), adjusted.day());
+    }
+
+    private static String formatDate(int year, int month, int day) {
         return String.format("%04d-%02d-%02d",
-                value.getInt32(StructuredTemporal.YEAR_FIELD),
-                value.getInt8(StructuredTemporal.MONTH_FIELD),
-                value.getInt8(StructuredTemporal.DAY_FIELD));
+                year, month, day);
     }
 
     static String timestamp(Struct value, int precision, TemporalPrecisionLossHandlingMode handlingMode) {
-        requireFinite(value);
-        final var fraction = StructuredTemporalPreflightValidator.reduceFraction(
-                StructuredTemporalSupport.getPicoseconds(value), precision, handlingMode);
-        if (fraction.carrySeconds() != 0) {
-            final LocalDateTime dateTime = StructuredTemporalSupport.toLocalDateTime(value)
-                    .withNano(fraction.nanoseconds())
-                    .plusSeconds(fraction.carrySeconds());
-            return formatTimestamp(dateTime.getYear(), dateTime.getMonthValue(), dateTime.getDayOfMonth(),
-                    dateTime.getHour(), dateTime.getMinute(), dateTime.getSecond(), fraction.picoseconds(), precision);
+        return timestamp(value, precision, handlingMode, TemporalRange.unbounded(),
+                TemporalRangeLossHandlingMode.FAIL, "target temporal type");
+    }
+
+    static String timestamp(Struct value, int precision, TemporalPrecisionLossHandlingMode precisionHandlingMode,
+                            TemporalRange range, TemporalRangeLossHandlingMode rangeHandlingMode,
+                            String targetDescription) {
+        if (StructuredTemporal.isFinite(value) && hasZeroDate(value)) {
+            final var fraction = StructuredTemporalPreflightValidator.reduceFraction(
+                    StructuredTemporalSupport.getPicoseconds(value), precision, precisionHandlingMode);
+            if (fraction.carrySeconds() == 0) {
+                return formatTimestamp(
+                        value.getInt32(StructuredTemporal.YEAR_FIELD),
+                        value.getInt8(StructuredTemporal.MONTH_FIELD),
+                        value.getInt8(StructuredTemporal.DAY_FIELD),
+                        value.getInt8(StructuredTemporal.HOUR_FIELD),
+                        value.getInt8(StructuredTemporal.MINUTE_FIELD),
+                        value.getInt8(StructuredTemporal.SECOND_FIELD),
+                        fraction.picoseconds(), precision);
+            }
         }
-        return formatTimestamp(
-                value.getInt32(StructuredTemporal.YEAR_FIELD),
-                value.getInt8(StructuredTemporal.MONTH_FIELD),
-                value.getInt8(StructuredTemporal.DAY_FIELD),
-                value.getInt8(StructuredTemporal.HOUR_FIELD),
-                value.getInt8(StructuredTemporal.MINUTE_FIELD),
-                value.getInt8(StructuredTemporal.SECOND_FIELD),
-                fraction.picoseconds(), precision);
+        final Boundary adjusted = StructuredTemporalSupport.adjustTimestamp(
+                value, precision, precisionHandlingMode, range, rangeHandlingMode, targetDescription);
+        return formatTimestamp(adjusted.year(), adjusted.month(), adjusted.day(), adjusted.hour(), adjusted.minute(),
+                adjusted.second(), adjusted.picoseconds(), precision);
     }
 
     static String duration(Struct value, int precision, TemporalPrecisionLossHandlingMode handlingMode) {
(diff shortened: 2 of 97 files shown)


debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/StructuredDateType.java
```diff
@@ -39,4 +52,29 @@ public List<ValueBindDescriptor> bind(int index, Schema schema, Object value) {
...
+            }
+            if (StructuredTemporal.isNegativeInfinity(struct)) {
+                return List.of(new ValueBindDescriptor(index, "-infinity", Types.VARCHAR));
+            }
+        }
+        return super.bind(index, column, schema, value);
+    }
+
+    private boolean supportsInfinity() {
+        return getDialect() == null || getDialect().getTargetTemporalCapabilities().dateInfinitySupported();
```
Naros: Can dialect be null? And if so, wouldn't it be safer if infinity isn't supported?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/StructuredTimestampType.java
```diff
@@ -39,4 +52,29 @@ public List<ValueBindDescriptor> bind(int index, Schema schema, Object value) {
...
+            }
+            if (StructuredTemporal.isNegativeInfinity(struct)) {
+                return List.of(new ValueBindDescriptor(index, "-infinity", Types.VARCHAR));
+            }
+        }
+        return super.bind(index, column, schema, value);
+    }
+
+    private boolean supportsInfinity() {
+        return getDialect() == null || getDialect().getTargetTemporalCapabilities().timestampInfinitySupported();
```
Naros: Same concern with dialect.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/StructuredZonedTimestampType.java
```diff
@@ -39,4 +52,29 @@ public List<ValueBindDescriptor> bind(int index, Schema schema, Object value) {
...
+            }
+            if (StructuredTemporal.isNegativeInfinity(struct)) {
+                return List.of(new ValueBindDescriptor(index, "-infinity", Types.VARCHAR));
+            }
+        }
+        return super.bind(index, column, schema, value);
+    }
+
+    private boolean supportsInfinity() {
+        return getDialect() == null || getDialect().getTargetTemporalCapabilities().timestampInfinitySupported();
```
Naros: Same concern with dialect.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/StructuredTemporalLiteral.java
```diff
@@ -5,46 +5,134 @@
...
+                .add(BigDecimal.valueOf(intValue(value, StructuredTemporal.MINUTES_FIELD)).multiply(BigDecimal.valueOf(60)))
+                .add(BigDecimal.valueOf(longValue(value, StructuredTemporal.SECONDS_FIELD)))
+                .add(BigDecimal.valueOf(longValue(value, StructuredTemporal.PICOSECONDS_FIELD), 12));
+        final RoundingMode roundingMode = handlingMode == TemporalPrecisionLossHandlingMode.ROUND ? RoundingMode.HALF_UP : RoundingMode.DOWN;
+        totalSeconds = totalSeconds.setScale(precision, roundingMode);
+
+        final BigDecimal maxSeconds = BigDecimal.valueOf(838L * 3_600 + 59L * 60 + 59L)
+                .add(BigDecimal.ONE.subtract(BigDecimal.ONE.movePointLeft(precision)));
+        if (totalSeconds.abs().compareTo(maxSeconds) > 0) {
+            throw new ConnectException("Structured duration is outside the MySQL TIME range");
```
Naros: Could we include the value in this exception?

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresTemporalPrecisionHandlingIT.java
```diff
@@ -610,7 +616,7 @@ void shouldConvertSnapshotTemporalsToStructuredValues() throws Exception {
         assertEquals(4, interval.getInt32(StructuredTemporal.HOURS_FIELD));
         assertEquals(5, interval.getInt32(StructuredTemporal.MINUTES_FIELD));
         assertEquals(6L, interval.getInt64(StructuredTemporal.SECONDS_FIELD));
-        assertEquals(789_000_000, interval.getInt32(StructuredTemporal.NANOS_FIELD));
+        assertEquals(789_000_000_000L, interval.getInt64(StructuredTemporal.PICOSECONDS_FIELD));
```
Naros, line 625: Can we add the same assertions for `PRECISION_PARAMETER_KEY` and `DURATION_KIND_PARAMETER_KEY` here in the snapshot-equivalent test, as in the streaming test above?

Naros (changes requested):
Hi @minleejae this PR is coming along.

In addition to my findings below, I believe the tests lack coverage for a few corner cases, which would include:
* No E2E test tests `temporal.precision.loss.handling.mode`. While there are mock tests, it would be good to at least have one E2E for a given dialect.
* There is exactly only one IT for `saturate` for CockroachDB. But none trigger the _fail_ path for range loss.
* The Db2, SingleStore, and StarRocks run with the default of _fail_, but never send data that triggers the range rejection, so unhappy path isn't triggered.

I'd also recommend a rebase given the recent activity in the JDBC sink as of late, and check the test failures, they do appear to be related to this work.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/StructuredDateType.java
minleejae: Good point. A null dialect now means infinity is unsupported, with regression coverage in dda2c6b

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/StructuredTimestampType.java
minleejae: Updated consistently for timestamps in dda2c6b

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/StructuredZonedTimestampType.java
minleejae: Updated consistently for zoned timestamps in dda2c6b

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/StructuredTemporalLiteral.java
minleejae: The exception now includes the formatted duration value, with coverage in 6fe6efc

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresTemporalPrecisionHandlingIT.java
minleejae: Added the precision and duration-kind assertions to the snapshot test in 8665cbf

debezium-connector-common/src/main/java/io/debezium/relational/TableSchemaBuilder.java
```diff
@@ -411,10 +412,13 @@ private ValueConverter wrapInMappingConverterIfNeeded(ColumnMappers mappers, Tab
     protected void addField(SchemaBuilder builder, Table table, Column column, ColumnMapper mapper) {
         final Object defaultValue = parseDefaultValue(table.id(), column);
 
-        final SchemaBuilder fieldBuilder = customConverterRegistry.registerConverterFor(table.id(), column, defaultValue)
+        SchemaBuilder fieldBuilder = customConverterRegistry.registerConverterFor(table.id(), column, defaultValue)
                 .orElse(valueConverterProvider.schemaBuilder(column));
 
         if (fieldBuilder != null) {
+            if (mapper != null || column.comment() != null) {
+                fieldBuilder = StructuredTemporal.qualifySchemaBuilderWithSourceColumn(fieldBuilder, column.name());
```
Naros, line 420: This means that adding a column comment changes the Avro schema name for structured temporal types, potentially causing schema registry compatibility concerns. Is this absolutely required?

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/StructuredZonedTimestampType.java
```diff
@@ -14,6 +19,15 @@ public class StructuredZonedTimestampType extends StructuredTimestampType {
...
+        return io.debezium.time.StructuredZonedTimestamp.schemaNames();
+    }
+
+    @Override
+    public void validate(ColumnDescriptor column, Schema schema, Object value) {
+        super.validate(column, schema, value);
+        if (value != null) {
+            StructuredTemporalPreflightValidator.validateZonedTimestamp(
+                    requireStruct(value), getDialect().getTargetTemporalCapabilities());
+        }
```
Naros: Should this be reversed so that the exception failure is on the most semantically meaningful check first, a user would see a precision/range error first.

Naros (commented):
Hi @minleejae I've left a few final points after another review pass.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/StructuredZonedTimestampType.java
minleejae: Agreed — MySQL was the only dialect running this check last, so I moved it ahead of `super.validate(...)` to match the shared base type and Db2.

debezium-connector-common/src/main/java/io/debezium/relational/TableSchemaBuilder.java
minleejae: Hi @Naros 
Thanks — fair concern, and you are right about the cause: the qualifier is applied *conditionally*, which is why adding a comment can change the name. Two ways to stop that:

- **A — qualify unconditionally.** The name becomes a pure function of (semantic type, precision, column), so toggling a comment or `column.propagate.source.type` never changes it. Cost: a table with a single temporal column also gets a column-specific record name.
- **B — qualify only on a real collision**, i.e. when a table has two or more structured temporal columns sharing a base name. Cost: adding a second same-precision column renames the first.

I lean towards A because the rule is stateless and easy to document, but B keeps the common single-column case clean. Neither feels clearly right to me though, so if you have a better approach in mind I would rather go with that. How would you like me to proceed?

Dropping the qualifier is the one option I would avoid. What I measured:

**Why it is needed.** With `time.precision.mode=structured` temporal columns become Connect structs, so they map to *named* Avro records, and all columns of the same type and precision share one name. `column.propagate.source.type` always adds `__debezium.source.column.name` and `include.schema.comments` adds a doc, so the records diverge while keeping that shared name. Through the real `TableSchemaBuilder` path:

| Table | Result |
|---|---|
| `TIMESTAMP(6)` x2 + propagate | fails |
| `TIMESTAMP(6)` x2, comment on one column | fails |
| `TIMESTAMP(6)` x2, no propagate / no comments | ok — identical records collapse to a name reference |
| `TIMESTAMP(3)` + `TIMESTAMP(6)` + propagate | ok — the precision suffix already separates them |
| `TIMESTAMP(6)` x1 + propagate | ok |

**How it surfaces**, on stock `debezium/connect` images:

| Stack | Behaviour |
|---|---|
| 3.1 or 3.6, Confluent converter | the second record silently becomes a name reference, so the second column inherits the first one's `__debezium.source.column.name`; the source stays RUNNING and the **JDBC sink** dies on a duplicate column |
| 3.6, Apicurio 3.2.5 converter + registry | Apicurio 3.x registers each named record as its own artifact, so the second one is a conflicting write to the same artifact ID; the **source task** fails on `409` before anything reaches Kafka |

For a source table with two such columns, `amt_a` and `amt_b`:

```
CREATE TABLE sink_two_numeric (id integer NOT NULL,
    amt_a double precision NULL, amt_a double precision NULL, PRIMARY KEY(id))
ERROR: column "amt_a" specified more than once
```

Values are bound correctly and the database rejects the statement, so nothing is written wrongly — the pipeline just stops, far from the setting that caused it. The second row is what Apicurio users hit, and 3.2.5 is the version `main` pins today (`version.apicurio`). Single-column controls keep running in both.

**Why only structured temporals.** This is not a new class of problem: every struct-valued semantic type has one fixed schema name and therefore one Avro record name. Primitive-backed ones are safe because Avro only tracks named types — `io.debezium.data.Enum` carries per-column `allowed` parameters and never collides, being a string. Seven struct-valued types are already released:

<details>
<summary>the seven, and the doc each one presets</summary>

| Type | Emitted by | Preset doc |
|---|---|---|
| `io.debezium.data.geometry.Geometry` | PostgreSQL, MySQL/MariaDB | `"Geometry"` |
| `io.debezium.data.geometry.Geography` | PostgreSQL | `"Geography"` |
| `io.debezium.data.geometry.Point` | PostgreSQL, MySQL/MariaDB | `"Geometry (POINT)"` |
| `io.debezium.data.geometry.Line` | PostgreSQL | `"Geometry (LINE)"` |
| `io.debezium.data.geometry.Circle` | PostgreSQL | `"Geometry (CIRCLE)"` |
| `io.debezium.data.VariableScaleDecimal` | PostgreSQL, Oracle | `"Variable scaled decimal"` |
| `io.debezium.data.SparseDoubleVector` | PostgreSQL | `"Sparse double vector"` |

</details>

Against stock 3.1 and 3.6 images — released code, nothing from this PR — those seven hit two defects I could not find an existing issue for:

1. **The same record-name collision**, reproduced with two unconstrained `NUMERIC` columns on PostgreSQL and two `GEOMETRY` columns on MySQL, with `column.propagate.source.type` enabled. Outcomes exactly as in the table above.
2. **`include.schema.comments=true` breaks them outright**, because `TableSchemaBuilder.addField` calls `fieldBuilder.doc(column.comment())` over the preset doc: `SchemaBuilderException: Invalid SchemaBuilder call: doc has already been set.` This one is converter-independent and fails before Avro is involved. Only MySQL, MariaDB and Oracle read column comments. Structured temporals preset no doc, so they are unaffected.

Unlike structured temporals, those seven names are already in consumers' registries, so how far this should reach is your call too — happy to follow whatever scope you think is right here.

Naros: I do tend to agree with you regarding A. 

My only hesitation is the hex-encoding. I looked across the supported databases, and I couldn't find anyone where the encoding itself would present a subject name limit (Apicurio itself has a 512 upper size limit) and database column names won't push us over that limit; however, the other concern is readability. 

I don't think subject names are looked at much in the grand scheme, and the hex encoding prevents any sort of name collisions with Avro requirements, but we do have adjusters for this reason and I wonder if using those might be overall better, so we get:
```
io.debezium.data.StructuredTimestampPrecisionSourceColumn_create_timestamp
OR
io.debezium.data.StructuredTimestampPrecisionSourceColumn.create_timestamp
```
@jpechane do you have an opinion one way or the other?

Naros (approved):
Overall code LGTM, just one outstanding ping to Jiri and if he's okay with it, this would be good for merging from my PoV.

PR #7681, opened 2026-07-17
debezium/dbz#2236 Add signal id and data collection attribution to incremental snapshot watermark metadata

cuong-dang:
Fixes debezium/dbz#2236

## Description

Incremental snapshot watermark rows (`snapshot-window-open` / `snapshot-window-close`) currently carry only a per-chunk UUID and the open/close timestamps, so external tooling consuming the signal table's CDC events cannot attribute a window to the `execute-snapshot` signal that requested it or to the data collection being chunked — particularly when one signal lists several collections, or a second signal arrives while a snapshot is running.

This PR threads that attribution through to the watermark metadata:

* `DataCollection` now carries the `correlationId` of the signal that requested it, assigned in `addDataCollectionNamesToSnapshot` and round-tripped through the offsets. Tracking it per collection (rather than only on the context) keeps attribution correct when collections queued by different signals are pending in the same context.
* The watermark `SignalMetadata` JSON gains two fields: `correlationId` (the requesting signal's id) and `dataCollectionId` (the fully-qualified identifier of the collection the chunk belongs to). Both are captured at window-open and carried to the matching close, since the collection queue may already have advanced by the time the close is emitted.
* `SignalMetadata` serialization uses Jackson (correct escaping for arbitrary signal ids and identifiers) with a custom printer that keeps the emitted format byte-identical to previous versions — pinned by test.

Backward compatibility:

* The `data` column of watermark rows is write-only metadata; nothing on the connector side parses it (`OpenIncrementalSnapshotWindow`/`CloseIncrementalSnapshotWindow` key off the `id` column only), and the legacy field layout is unchanged.
* Offsets written by older versions deserialize with a null per-collection correlation id; the new offset key is ignored on downgrade.
* The pre-existing context-level `correlationId` (read by the notification service) is kept as-is; its overwrite-on-latest-signal behavior and its relationship to the per-collection value are now documented on the field.
* The `INSERT_DELETE` watermarking strategy is unaffected (its close deletes the open row and writes no metadata; open rows carry attribution under both strategies).

Unit tests cover the metadata format and escaping, the offset round-trip (including legacy offsets without the new key), and open/close attribution across collections queued by different signals.

## PR Checklist

### DIFF
--- a/debezium-connector-common/src/test/java/io/debezium/pipeline/source/snapshot/incremental/SignalBasedIncrementalSnapshotContextTest.java
+++ b/debezium-connector-common/src/test/java/io/debezium/pipeline/source/snapshot/incremental/SignalBasedIncrementalSnapshotContextTest.java
@@ -50,6 +50,29 @@ public class SignalBasedIncrementalSnapshotContextTest {
     }
 
     @Test
+    public void shouldLoadOffsetsWrittenWithPerCollectionCorrelationId() {
+        final Map<String, Object> offsets = Map.of(
+                "incremental_snapshot_collections",
+                "[{\"incremental_snapshot_collections_id\":\"public.a\","
+                        + "\"incremental_snapshot_collections_additional_condition\":null,"
+                        + "\"incremental_snapshot_collections_surrogate_key\":null,"
+                        + "\"incremental_snapshot_collections_correlation_id\":\"signal-1\"},"
+                        + "{\"incremental_snapshot_collections_id\":\"public.b\","
+                        + "\"incremental_snapshot_collections_additional_condition\":null,"
+                        + "\"incremental_snapshot_collections_surrogate_key\":null,"
+                        + "\"incremental_snapshot_collections_correlation_id\":\"signal-2\"}]");
+
+        final IncrementalSnapshotContext<TableId> restored = SignalBasedIncrementalSnapshotContext.load(offsets);
+
+        assertThat(restored.getDataCollections())
+                .extracting(DataCollection::getCorrelationId)
+                .containsExactly(Optional.of("signal-1"), Optional.of("signal-2"));
+        assertThat(restored.getDataCollections())
+                .extracting(dataCollection -> dataCollection.getId().identifier())
+                .containsExactly("public.a", "public.b");
+    }
+
+    @Test
     public void shouldLoadOffsetsWrittenWithoutPerCollectionCorrelationId() {
         final Map<String, Object> offsets = Map.of(
                 "incremental_snapshot_collections",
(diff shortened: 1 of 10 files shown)


debezium-connector-common/src/test/java/io/debezium/pipeline/source/snapshot/incremental/SignalBasedIncrementalSnapshotContextTest.java
```diff
@@ -0,0 +1,69 @@
...
+
+        final IncrementalSnapshotContext<TableId> restored = SignalBasedIncrementalSnapshotContext.load(offsets);
+
+        assertThat(restored.getDataCollections())
+                .extracting(DataCollection::getCorrelationId)
+                .containsExactly(Optional.empty());
+        assertThat(restored.getDataCollections())
+                .extracting(dataCollection -> dataCollection.getId().identifier())
+                .containsExactly("public.a");
+    }
```
mfvitale, line 91: Could you also add a test for the restore of the per collection correlation id?

mfvitale:
@cuong-dang Thanks for the PR! The changes LGTM. I left a minor comment.

debezium-connector-common/src/test/java/io/debezium/pipeline/source/snapshot/incremental/SignalBasedIncrementalSnapshotContextTest.java
cuong-dang: Hi @mfvitale, I added the test `shouldLoadOffsetsWrittenWithPerCollectionCorrelationId`. I'm assuming you mean loading from a literal offset JSON. Thanks!

mfvitale:
@cuong-dang Seems there is some conflict, could you please rebase with the main?

cuong-dang:
@mfvitale Done!

mfvitale:
If CI passes it can be merged.

cuong-dang:
@mfvitale CI completed with 2 failures, both of which seem unrelated to this change:
- Platform Conductor failed resolving io.debezium:debezium-schema-generator:3.7.0-SNAPSHOT: looks like a CI Maven cache restore issue.
- MariaDB 11.4.3 - mariadb-ci: appears flaky today: the other MariaDB legs (gtids, ssl) pass here, and different MariaDB legs failed on the main push build and on other PRs from the same window.

All other 55 jobs passed, including the incremental snapshot suites this PR touches. Could you re-run the two failed jobs when you get a chance? Happy to make any other changes if needed.

Naros (approved):
Hi @cuong-dang I don't think we need to rerun CI, as I concur those are unrelated.

mfvitale:
@cuong-dang Applied, thanks!

PR #7685, opened 2026-07-18
debezium/dbz#2238 Avoid NullPointerException resolving unsupported array column types

sundongkim-dev:
Fixes debezium/dbz#2238

## Description

When a PostgreSQL array column (`int[]`, `text[]`, …) is streamed to a JDBC sink whose dialect does not register an `ARRAY` type (e.g. MySQL), `GeneralDatabaseDialect#getSchemaType` fails with an opaque `NullPointerException` instead of the clear error it was designed to raise.

A Connect `ARRAY` schema has no logical name, so `schema.name()` is `null`. After the type-registry lookups miss (`ARRAY` is registered only by `PostgresDatabaseDialect`), the method reaches `switch (schema.name())`, which the compiler implements via `schema.name().hashCode()` — throwing an NPE before the `default` branch that raises `ConnectException("Failed to resolve column type ...")`.

This change guards the switch against a `null` schema name so unsupported, unnamed schemas fall through to the intended clear `ConnectException`. Behavior for named schemas (including the vector types) is unchanged.

Added a regression test (`JdbcSinkColumnTypeMappingIT#testShouldFailWithClearErrorForUnsupportedArrayColumnType`, MySQL sink) that feeds an array-typed field and asserts the clear error message; it fails with the NPE before the fix and passes after.

> Note: this PR only converts the crash into a clear error. Whether the JDBC sink should additionally map arrays to a portable type (e.g. `JSON`) on array-less dialects is an open design question — happy to follow up separately based on maintainer preference.

## PR Checklist

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/GeneralDatabaseDialect.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/GeneralDatabaseDialect.java
@@ -487,19 +487,21 @@ public class GeneralDatabaseDialect implements DatabaseDialect {
             return type;
         }
 
-        switch (schema.name()) {
-            case SparseDoubleVector.LOGICAL_NAME, FloatVector.LOGICAL_NAME, DoubleVector.LOGICAL_NAME ->
-                throw new ConnectException(
-                        String.format(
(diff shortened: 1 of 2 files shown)


Naros:
Hi @sundongkim-dev maybe you missed the issue update, https://github.com/debezium/debezium/pull/7683 :-)

PR #7688, opened 2026-07-19
debezium/dbz#2243 Add Conditions to Chunk Boundary Calculation Querie…

hagar-yasser1:
…s + Optimize single Chunk queries by reusing DB dependent condition layout logic built for incremental snapshots

Fixes debezium/dbz#2243

## PR Checklist


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
```diff
@@ -940,4 +940,5 @@ public List<Column> getTableColumnsForDecoder(TableId tableId, Tables.ColumnName
     public interface PostgresValueConverterBuilder {
         PostgresValueConverter build(TypeRegistry registry);
     }
+
```
Naros: This change to `PostgresConnection` is superfluous and can be dropped.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
```diff
@@ -128,76 +129,77 @@ protected void addLowerBound(IncrementalSnapshotContext<T> context, Table table,
         // (k1 > ?) where ? is NULL: (k1 IS NOT NULL): by definition, NULL is the first value and every value is always higher
         // (k1 = ?) where ? is non-NULL: (k1 = ?): no translation needed for non-NULL values
         // (k1 = ?) where ? is NULL: (k1 IS NULL): need to use IS NULL instead of equality comparison
-        final List<Column> pkColumns = getQueryColumns(context, table);
+
         final Optional<Boolean> nullsSortLast = jdbcConnection.nullsSortLast();
+        final String operator = inclusiveFinal ? " >= ?" : " > ?";
```
Naros: Doesn't this create invalid SQL when `inclusiveFinal` is `true`, because it uses `>=` for every OR term, not just the last OR term, making this semantically wrong for composite keys?

The `CascadingOrBoundaryConditions.buildLowerBound` used the `isLastTerm` logic and properly applied this. This impacts `ChunkBoundaryCalculator` and `SnapshotChunkQueryBuilder`, but the RowValue implementation is not because it uses `ROW() >= ROW()` which handles this tuple logic internally.

Single-column PK tables are unaffected.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
```diff
@@ -128,76 +129,77 @@ protected void addLowerBound(IncrementalSnapshotContext<T> context, Table table,
...
-            sql.append(')');
+            condition.append(')');
         }
     }
 
-    protected void addUpperBound(IncrementalSnapshotContext<T> context, Table table, Object[] boundaryKey, StringBuilder sql) {
+    public void addUpperBound(List<Column> pkColumns, Object[] boundaryKey, StringBuilder sql, boolean inclusiveFinal) {
         sql.append("NOT ");
-        addLowerBound(context, table, boundaryKey, sql);
+        addLowerBound(pkColumns, boundaryKey, sql, !inclusiveFinal);
```
Naros, line 225: When this is `false`, this renders:
```sql
NOT (k1 >= ?) = k1 < ?
```
This misses where `k1 = boundary_k1` but the remaining key columns are less than the boundary values. The `CascadingOrBoundaryConditions.buildUpperBound` built a direct cascading-OR for `</<=` and was correct.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/chunked/ChunkBoundaryCalculator.java
```diff
@@ -151,4 +166,22 @@ private Object[] queryColumnValues(List<Column> columns, String sql) throws SQLE
         });
     }
 
+    private Object[] queryColumnValues(List<Column> columns, String sql, Object[] lowerBoundBoundary) throws SQLException {
+        final PreparedStatement statement = jdbcConnection.connection().prepareStatement(sql);
+        if (lowerBoundBoundary != null) {
+            connectionChunkQueryBuilder.bindBoundaryParams(statement, columns, lowerBoundBoundary, 1, jdbcConnection);
+        }
+        try (statement; ResultSet rs = statement.executeQuery()) {
```
Naros: This creates a potential memory leak: if `bindBoundaryParams` throws an exception, the created `statement` is not cleaned up in a try-with-resources block. Please move the try-with-resources for the `statement` up to its initializer for correctness.

debezium-connector-common/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
```diff
@@ -967,7 +967,7 @@ protected void doCreateDataEventsForChunk(ChangeEventSourceContext sourceContext
         final Table table = chunk.getTable();
         final TableChunkProgress progress = progressMap.get(tableId);
 
-        final Stopwatch exportTimer = Stopwatch.accumulating();
+        final Stopwatch exportTimer = Stopwatch.accumulating().start();
```
Naros, line 970: This appears to be unrelated to the chunk boundary condition logic. While this appears to be a bug, it should ideally be in its own commit, with an appropriate description.

debezium-connector-common/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -1707,16 +1707,22 @@ else if (additionalCondition.isPresent()) {
...
+            sql.append(" WHERE ")
+                    .append(condition);
+        }
+
+        return sql.append(" ORDER BY ")
                 .append(orderBy)
                 .append(" OFFSET ").append(size)
                 .append(" ROWS FETCH NEXT 1 ROWS ONLY")
                 .toString();
+
```
Naros:
```suggestion
```

Naros (changes requested):
Hi @hagar-yasser1, the way this PR is designed conflates two concerns into a single commit, which goes against our contribution guidelines: adding the condition for initial chunked snapshots and refactoring the code for incremental snapshots. 

Could you please separate these two concerns into two independent commits, one that addresses the fix specifically for the initial chunked snapshots and one for the refactor. It makes things much easier to review and backport if we decide to do so.

hagar-yasser1:
Thanks @Naros for the helpful review!

I have broken the PR into 3 main commits and a refactor commit. 

- Commit#1: Adds conditions initial chunked snapshots. Doesn't change anything in the flow for incremental snapshots (only adds new methods that are used for initial chunked snapshots).
- Commit#2: Unifies lower/upperBound and binding logic across initial chunked snapshots and incremental snapshots. This is to make initial chunked snapshots benefit from future optimizations and fixes in incremental snapshots and vice versa.
- Commit#3: Fixes an bug related to the Stopwatch object used to log the performance of the initial chunked snapshots.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractChunkQueryBuilder.java
hagar-yasser1: Now given the main addLowerBound logic should be correct, I think the addUpperBound logic is corrected as well right?

debezium-connector-common/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
hagar-yasser1: ✅

Naros:
Hi @hagar-yasser1 it appears the test failures are related, could you take a look?

hagar-yasser1:
Hello @Naros, I fixed the bugs causing the test failures and added a related [PR](https://github.com/debezium/debezium-connector-informix/pull/182) to the informix connector.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/chunked/ChunkBoundaryCalculator.java
```diff
@@ -167,27 +167,20 @@ private Object[] queryColumnValues(List<Column> columns, String sql) throws SQLE
...
 
     private Object[] queryColumnValues(List<Column> columns, String sql, Object[] lowerBoundBoundary) throws SQLException {
-
-        try (PreparedStatement statement = prepareBoundStatement(columns, sql, lowerBoundBoundary);
-                ResultSet rs = statement.executeQuery()) {
-            if (rs.next()) {
-                final Object[] values = new Object[columns.size()];
-                for (int i = 0; i < columns.size(); i++) {
-                    values[i] = rs.getObject(i + 1);
+        try (PreparedStatement statement = jdbcConnection.connection().prepareStatement(sql)) {
```
Naros, line 170: Hi @hagar-yasser1, I made a small change here; explanation below:

When you write a `try-with-resources`, it desugars as:
```java
var statement = prepareBoundStatement(...);
try {
}
catch (Exception e) {
  if (statement != null && (statement instanceof AutoCloseable || statement instanceof Closeable)) {
    statement.close();
  }
}
```

Now, because `bindBoundaryParams` can throw an exception, after the TWR is desugared, that exception is thrown outside the try block, so the statement would never be closed. The solution is to explicitly wrap the assignment in a `try-with-resources` block so that, if the binding fails, the resource is properly closed.

Just in case you face this in the future, because it's not always what one would expect.

Naros (approved):
Code LGTM, just waits for CI.

debezium-connector-common/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
hagar-yasser1: I think the same resource leakage can happen here for example right? @Naros

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/chunked/ChunkBoundaryCalculator.java
hagar-yasser1: Thanks @Naros for the insight! 👍 
Based on this, I think this leaky pattern can be found in other places in the code as well right? like [example#1](https://github.com/debezium/debezium/blob/a28aefedb61a41668e7fe65ee2f6f5c30804e3d7/debezium-connector-common/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java#L999) and [example#2](https://github.com/debezium/debezium/blob/a28aefedb61a41668e7fe65ee2f6f5c30804e3d7/debezium-connector-common/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java#L906)

hagar-yasser1:
Hi @Naros, I have added an extra commit to isolate the logic for generating query params for chunkQueryBuilders to make it unit testable. This aims to make validating new logic and discovering bugs in this area faster and easier.

vjuranek (commented):
@hagar-yasser1 could you please take a look on Informix test failures, see to be related. Thanks!

hagar-yasser1:
Hello @vjuranek, I have added a [PR](https://github.com/debezium/debezium-connector-informix/pull/182) to informix connector but I am not sure why the tests here don't pick up the related branch at informix.

Naros:
Hi @hagar-yasser1, the CI cross-Maven builds are unidirectional, so Informix builds on your PR here as long as the two repositories share the same branch name. It will fail here but pass in Informix; that's normal.

debezium-connector-common/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
Naros: Great catch; that's correct @hagar-yasser1 👍

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/chunked/ChunkBoundaryCalculator.java
Naros: Yep, and I guess with this being a larger systemic issue, maybe fixing the first one if it's part of code changes for this PR, and the latter, maybe we do in a follow-up PR. WDYT?

hagar-yasser1: No problem, I have just added a commit to fix the first occurrence.

vjuranek (approved):
LGTM

Naros (approved):

Naros:
Applied, and thanks @hagar-yasser1 - great work! :tada:

PR #7689, opened 2026-07-19
debezium/dbz#2244 Bound retries when a JDBC storage operation keeps failing

### LINKED ISSUE #2244: JDBC storage: RetriableConnection retries a failing operation forever without honoring max.retries
## What

In the JDBC storage module, `io.debezium.storage.jdbc.RetriableConnection#executeWithRetry(...)`
retries a **failing operation** forever when the connection itself is healthy. The
attempt counter is incremented and compared against `max.retries` only on the
*reconnection* path; the *operation-failure* path is not bounded.

`RetriableConnection.java` (operation branch):

```java
try {
    if (func != null) {
        return func.accept(conn);
    }
    if (consumer != null) {
        consumer.accept(conn);
        return null;
    }
}
catch (SQLException e) {
    LOGGER.warn("Attempt {} to call '{}' failed.", attempt, name, e);
    if (rollback) {
        LOGGER.warn("'{}': doing rollback.", name);
        try {
            conn.rollback();
        }
        catch (SQLException ex) {
            // ignore
        }
    }
    close();          // conn = null
}
// loops back: `attempt` is never incremented, max.retries is never checked, no delay
```

After `close()` sets `conn = null`, the next loop iteration finds `isOpen() == false`,
successfully **reconnects** (the database is up), re-runs the identical operation, throws
the same `SQLException`, closes, reconnects again — indefinitely. This is a tight busy-loop:
the `wait.retry.delay` backoff lives only in the reconnect-*failure* branch, and here
reconnection *succeeds* every time.

## Impact

A persistent, operation-level `SQLException` (e.g. a value-too-long / constraint / type
error while writing a schema-history record or saving an offset, or a syntax error from a
misconfigured custom `table.ddl` / insert statement) makes the call **never return and
never throw**. The affected thread hangs forever and spins a CPU core; the connector
neither makes progress nor fails, and the underlying `SQLException` is masked.

Reachable from `JdbcSchemaHistory.storeRecord` / `recoverRecords`, and
`JdbcOffsetBackingStore.save` (and the `initialize*` / `exists` helpers).

## Expected

The operation-failure path should honor `max.retries` and `wait.retry.delay` just like the
reconnection path: retry up to `max.retries` times with the configured delay, then rethrow
the `SQLException` so the connector fails loudly instead of hanging.

## Fix direction

In the operation-failure `catch`, after `close()`, increment `attempt`, apply the
`wait.retry.delay` backoff, and rethrow once `attempt >= max.retries` — mirroring the
existing reconnection branch.

I have a fix + unit test ready and will open a PR referencing this issue.

TimurRakhmatullin86:
Closes https://github.com/debezium/dbz/issues/2244

### Problem

`io.debezium.storage.jdbc.RetriableConnection#executeWithRetry(...)` bounds its retries only
on the **reconnection** path — that branch increments `attempt` and rethrows once
`attempt >= maxRetryCount`. The **operation-failure** path does not:

```java
catch (SQLException e) {
    LOGGER.warn("Attempt {} to call '{}' failed.", attempt, name, e);
    if (rollback) { ... conn.rollback(); ... }
    close();                 // conn = null
}
// loops: attempt never incremented, maxRetryCount never checked, no delay
```

When the operation itself keeps throwing a `SQLException` on a healthy database (e.g. a
value that does not fit the target column, a constraint/type error, or a syntax error from a
misconfigured custom `table.ddl` / insert), `close()` sets `conn = null`, the next iteration
reconnects successfully, re-runs the identical operation, throws again, and repeats
**forever** — a tight busy-loop with no back-off that masks the error and hangs the affected
thread. Reachable from `JdbcSchemaHistory.storeRecord` / `recoverRecords` and
`JdbcOffsetBackingStore.save`.

### Change

Honor `max.retries` and `wait.retry.delay` on the operation-failure path too, mirroring the
existing reconnection branch: after closing the connection, rethrow the `SQLException` once
the attempts are exhausted, otherwise increment the counter, wait `wait.retry.delay`, and
retry. A persistent failure now surfaces as a thrown `SQLException` instead of an infinite
hang, while genuinely transient failures still succeed on a later attempt as before.

### Tests

- New `RetriableConnectionTest.shouldStopRetryingFailingOperationAfterMaxRetries` uses a real
  SQLite database and an operation that always throws `SQLException`; it asserts the call
  rethrows after exactly `maxRetryCount` invocations (with `@Timeout(SEPARATE_THREAD)` so a
  regression fails fast instead of hanging). It fails against the previous behaviour (infinite
  loop) and passes with the fix.
- `mvn -pl debezium-storage/debezium-storage-jdbc -am install`: `RetriableConnectionTest`
  passes and the reactor builds successfully.

### Author checklist
- [x] Change targets a single, self-contained fix
- [x] New unit test covering the fix
- [x] Commit is DCO signed-off

### DIFF
--- a/debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RetriableConnection.java
+++ b/debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RetriableConnection.java
@@ -113,6 +113,7 @@ public class RetriableConnection implements AutoCloseable {
     private synchronized <T> T executeWithRetry(ConnectionFunction<T> func, ConnectionConsumer consumer, String name, boolean rollback)
             throws SQLException {
         int attempt = 1;
+        DelayStrategy delayStrategy = DelayStrategy.constant(waitRetryDelay);
         while (true) {
             if (!isOpen()) {
                 LOGGER.debug("Trying to reconnect (attempt {}).", attempt);
@@ -127,7 +128,6 @@ public class RetriableConnection implements AutoCloseable {
                     }
                     attempt++;
                     LOGGER.debug("Waiting for reconnect for {} ms.", waitRetryDelay);
-                    DelayStrategy delayStrategy = DelayStrategy.constant(waitRetryDelay);
                     delayStrategy.sleepWhen(true);
                     continue;
                 }
@@ -163,7 +163,6 @@ public class RetriableConnection implements AutoCloseable {
                 }
                 attempt++;
                 LOGGER.debug("Waiting for retry for {} ms.", waitRetryDelay);
-                DelayStrategy delayStrategy = DelayStrategy.constant(waitRetryDelay);
                 delayStrategy.sleepWhen(true);
             }
         }
--- a/debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RetriableConnection.java
+++ b/debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RetriableConnection.java
@@ -153,6 +153,18 @@ public class RetriableConnection implements AutoCloseable {
                     }
                 }
                 close();
+                // The connection may be healthy while the operation itself keeps failing (e.g. a
+                // constraint violation or a value that does not fit the target column). Without
+                // bounding the retries here the loop would reconnect and re-run the same failing
+                // operation forever, so honor maxRetryCount and the retry delay as the reconnect
+                // branch above does.
+                if (attempt >= maxRetryCount) {
+                    throw e;
+                }
+                attempt++;
+                LOGGER.debug("Waiting for retry for {} ms.", waitRetryDelay);
+                DelayStrategy delayStrategy = DelayStrategy.constant(waitRetryDelay);
+                delayStrategy.sleepWhen(true);
             }
         }
     }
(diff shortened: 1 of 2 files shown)


debezium-storage/debezium-storage-jdbc/src/main/java/io/debezium/storage/jdbc/RetriableConnection.java
```diff
@@ -153,6 +153,18 @@ private synchronized <T> T executeWithRetry(ConnectionFunction<T> func, Connecti
...
+                // constraint violation or a value that does not fit the target column). Without
+                // bounding the retries here the loop would reconnect and re-run the same failing
+                // operation forever, so honor maxRetryCount and the retry delay as the reconnect
+                // branch above does.
+                if (attempt >= maxRetryCount) {
+                    throw e;
+                }
+                attempt++;
+                LOGGER.debug("Waiting for retry for {} ms.", waitRetryDelay);
+                DelayStrategy delayStrategy = DelayStrategy.constant(waitRetryDelay);
```
Naros: In a separate commit, can you move the `DelayStrategy` constructor up before the `while` (both this one and the one on line 130) and reuse the instance? There's no need for new allocations.

Naros (commented):
Hi @TimurRakhmatullin86, besides the commit message issue mentioned above, there is one small cleanup below, otherwise LGTM.

TimurRakhmatullin86:
Thanks @Naros! Addressed both: reworded the commit messages with the `debezium/dbz#2244` prefix, and moved the `DelayStrategy` allocation out of the retry loop into a single reused instance (both the reconnection and operation-failure paths) as a separate commit.

Naros (approved):
Code LGTM, lets wait on CI.

Naros:
Failures are related to a stale PR, local checks, and all tests passing.

Naros:

PR #7693, opened 2026-07-20
debezium/dbz#2248 Close MySQL connection after initialization failure

minleejae:
Fixes debezium/dbz#2248

## Description

Closes the JDBC connection when MySqlConnection initialization fails after a physical connection has been established. The original initialization exception is preserved, with any cleanup failure added as suppressed.

Adds regression coverage for the MySQL 8.4 permission failure path.

## Testing

- `./mvnw -pl debezium-connector-mysql test -DskipITs`
- `./mvnw clean install -pl debezium-connector-mysql -am -DskipITs`

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/jdbc/JdbcConnection.java
+++ b/debezium-connector-common/src/main/java/io/debezium/jdbc/JdbcConnection.java
@@ -40,6 +40,7 @@ import java.util.Set;
 import java.util.concurrent.TimeoutException;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
+import java.util.function.Supplier;
 import java.util.regex.Matcher;
 import java.util.stream.Collectors;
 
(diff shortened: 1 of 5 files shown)


Naros (approved):
Hi @minleejae so I agree with the fix, but looking at the broader connector portfolio, I believe the same pattern exists in Db2, IBMi, Ingres, and PostgreSQL.

So perhaps it would be reasonable as a follow-up to create some sort of helper method on JdbcConnection, e.g. `initializeAndClose(Runnable)` and effectively does what you have implemented uniformly?  WDYT?

Would that be worthwhile in a follow-up or would you prefer here?

minleejae:
Thanks @Naros. I added `initializeAndCloseOnFailure(Supplier<T>)` to `JdbcConnection` and applied it to MySQL and PostgreSQL in this PR.

  I used `Supplier<T>` because the constructors assign the initialization result to final fields. Db2, IBMi, and Ingres directly extend `JdbcConnection`, so they can adopt the same helper in follow-up PRs in their respective repositories. [26e1a01](https://github.com/debezium/debezium/pull/7693/commits/26e1a014eb747cbd66fc7dccaba2d54914a01a01)

vjuranek (approved):
LGTM

Naros:

PR #7682, opened 2026-07-18
debezium/dbz#2237 Fix handling `""` in single-quoted strings

Naros:
Fixes debezium/dbz#2237
Fixes debezium/dbz#2224

## Description
Fixed normalization with the new Oracle-based DDL parser where single-quoted strings contain `""` char sequences.

## PR Checklist

- [x] I have read the [contribution guidelines](https://github.com/debezium/debezium/blob/main/CONTRIBUTING.md) and the [governance document](https://github.com/debezium/governance/blob/main/GOVERNANCE.md) on PR expectations.
- [x] Minimal changes to code not directly related to your change (e.g. no unnecessary formatting changes or refactoring to existing code)
- [x] One feature/change per PR unless tightly coupled
- [x] Do a rebase on upstream `main`

PR #7701, opened 2026-07-21
debezium/dbz#2254 Apply reserved-keyword quoting only outside comments and string literals

jun-su-lee:
Fixes debezium/dbz#2254

## Description
`DdlNormalizer.addBackticksToReservedKeywords()` used to run three context-blind
`replaceAll` calls over the entire normalized statement, injecting backticks into
single-quoted string literals (column `DEFAULT` values, `COMMENT`s) and SQL comments
whose content happened to match the reserved-keyword heuristics, e.g.:

CREATE TABLE t (c VARCHAR(20) DEFAULT 'FROM RANK')
-> CREATE TABLE t (c VARCHAR(20) DEFAULT 'FROM RANK')

The corrupted statement is still valid SQL, so the connector silently records
literal values that differ from the actual database schema (schema history topic,
schema change events).

This change folds the reserved-keyword quoting into the existing
`QUOTED_LITERAL_PATTERN` matcher loop, applying it only to the plain SQL segments
between comments, string literals and quoted identifiers, and drops the global
second pass. Same class of defect as dbz#2237 (#7682), which fixed the
quote-conversion pass but did not cover this one.

## PR Checklist

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/DdlNormalizer.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/DdlNormalizer.java
@@ -61,21 +61,21 @@ public class DdlNormalizer {
 
         Matcher matcher = QUOTED_LITERAL_PATTERN.matcher(ddlContent);
         StringBuilder normalized = new StringBuilder(ddlContent.length());
+        int lastEnd = 0;
 
         while (matcher.find()) {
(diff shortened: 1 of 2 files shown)


Naros (approved):

Naros:
Applied to main (https://github.com/debezium/debezium/commit/e9f35adc2bbb22cc3abd5ce26d05550b2e62b38a) and 3.6.  Thanks @jun-su-lee :tada:

PR #7703, opened 2026-07-21
debezium/dbz#2255 Fix BoundedConcurrentHashMap eviction on put-heavy workloads

MrIvv:
Fixes https://github.com/debezium/dbz/issues/2255

`BoundedConcurrentHashMap` combines two defects into quadratic per-record cost on put-heavy workloads:

1. **`Segment.put()` on an existing key ignores the `onEntryHit` return value** — the check-and-evict contract is honored at the other three call sites (`get()` at 1292, both `replace()` overloads at 1347/1372) but not in the put hit path (1405), so a put-only caller (e.g. a dedup cache probed with unconditional `put`, like the OpenLineage SMT) grows the access queue without bound and the batch drain never triggers.
2. **`LRU`/`LIRS` threshold checks call `ConcurrentLinkedQueue.size()`** — an O(n) traversal per the Javadoc — on every hit.

Production evidence, reproduction steps and thread dump are in the linked issue: throughput decayed from ~90k to ~12.5k records/min over 18 hours with one conversion thread at ~96% CPU inside `ConcurrentLinkedQueue.size()`.

The fix:
- honors the `onEntryHit` return value in the put hit path, mirroring the exact form already used by both `replace()` overloads (`attemptEviction(true)` under the held segment lock; the evicted set flows into the existing out-of-lock `notifyEvictionListener`)
- tracks the access queue size with an `AtomicInteger` in both `LRU` and `LIRS` (increment on add, decrement on remove, reset on `execute()`/`clear()`). The counter may drift by a unit in rare races, which only delays the batch threshold slightly — the thresholds are approximate by design and the original racy `size()` snapshot had the same property.

No observable semantic change: same eviction policy, same thresholds, same listener notifications. Verified in a production-like rollout with the OpenLineage SMT active: the degradation signature is gone and the rate is stable over time.

### DIFF
--- a/COPYRIGHT.txt
+++ b/COPYRIGHT.txt
@@ -325,6 +325,7 @@ Ivan Kovbas
 Ivan Lorenz
 Ivan Luzyanin
 Ivan San Jose
+Ivan Senyk
 Ivan Trusov
 Ivan Vucina
 Jackey Zhang
--- a/debezium-microbenchmark/src/main/java/io/debezium/performance/core/BoundedConcurrentHashMapPerf.java
+++ b/debezium-microbenchmark/src/main/java/io/debezium/performance/core/BoundedConcurrentHashMapPerf.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.performance.core;
+
+import java.util.concurrent.TimeUnit;
+
+import org.openjdk.jmh.annotations.Benchmark;
+import org.openjdk.jmh.annotations.BenchmarkMode;
+import org.openjdk.jmh.annotations.Fork;
+import org.openjdk.jmh.annotations.Level;
+import org.openjdk.jmh.annotations.Measurement;
+import org.openjdk.jmh.annotations.Mode;
+import org.openjdk.jmh.annotations.OutputTimeUnit;
+import org.openjdk.jmh.annotations.Param;
+import org.openjdk.jmh.annotations.Scope;
+import org.openjdk.jmh.annotations.Setup;
+import org.openjdk.jmh.annotations.State;
+import org.openjdk.jmh.annotations.Warmup;
+import org.openjdk.jmh.infra.Blackhole;
+
+import io.debezium.util.BoundedConcurrentHashMap;
+import io.debezium.util.BoundedConcurrentHashMap.Eviction;
+
+/**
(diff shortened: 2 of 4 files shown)


mfvitale:
Hi @MrIvv thanks for the PR. 

Doyou think could be easy to add a [microbenchmark](https://github.com/debezium/debezium/tree/main/debezium-microbenchmark/src/main/java/io/debezium/performance/core) test to prove the improvement? Maybe you can report the `before/after` results.

MrIvv:
Hi @mfvitale 
I will prepare a microbenchmark for testing as soon as possible and add the report

MrIvv:
Hi @mfvitale, good idea — done. I added `BoundedConcurrentHashMapPerf` to `debezium-microbenchmark` (`io.debezium.performance.core`).

The benchmark times a batch of `putCount` puts over keys that are already present in the map (the put-hit path used by write-through dedup caches such as the OpenLineage SMT; the key set fits well within capacity, so every operation is a hit). Single-shot mode with a fresh map per iteration, `putCount` ∈ {100k, 400k} × eviction ∈ {LRU, LIRS}.

Results on JDK 21 (Temurin), 16-core x86_64 Linux, 5 iterations after 2 warmups — **before** = current `main`, **after** = this PR:

| Eviction | putCount | Before (ms/op) | After (ms/op) |
|---|---|---|---|
| LRU | 100,000 | 7,337 ± 3,605 | 11.2 ± 5.6 |
| LRU | 400,000 | 347,995 ± 420,927 | 36.3 ± 20.9 |
| LIRS | 100,000 | 5,717 ± 802 | 11.4 ± 8.4 |
| LIRS | 400,000 | 320,916 ± 344,737 | 32.5 ± 7.0 |

Two things worth reading out of the numbers besides the absolute difference:

1. **Scaling**: going 100k → 400k (4× the work) costs ~47-56× before (superlinear: the access queue is never drained on the put path, so each hit pays an O(queue) `size()` walk, plus growing GC pressure from the unbounded queue) and ~3× after (linear, as expected).
2. The wide error bars on the before side are part of the defect signature — per-invocation cost keeps growing with accumulated queue length and GC activity; the effect size (2-4 orders of magnitude) dwarfs them.

For the record, the full before suite takes ~85 minutes to run; after, ~3 seconds.

I also rebased the branch on current `main` to resolve the `Aliases.txt` conflict, and squashed the benchmark into the fix commit.

mfvitale:
@MrIvv Thanks a lot for the report. LGTM

@vjuranek I'll appreciate your eye on this too.

Naros (approved):

vjuranek (approved):
LGTM

vjuranek:
Merged, thanks @MrIvv !

PR #7707, opened 2026-07-22
debezium/dbz#2260 Treat nonzero TINYINT(1) values as true in TinyIntOneToBooleanConverter

minleejae:
Fixes debezium/dbz#2260

## Description

`TinyIntOneToBooleanConverter` converted numeric values with `intValue() > 0`, so negative values stored in a `TINYINT(1)` column converted to `false`. MySQL treats any nonzero value as true (`SELECT -1 IS TRUE` returns `1`), as does Debezium's default BOOLEAN conversion (`JdbcValueConverters#convertBoolean`), so the same stored value produced opposite results depending on the conversion path.

- Number branch: convert with `!= 0` instead of `> 0`.
- String branch: `return Integer.parseInt((String) x);` returned the parsed `Integer` itself under the registered BOOL schema — apply the same `!= 0` comparison.

Verified with a new unit test (`TinyIntOneToBooleanConverterTest`) and a new integration test (`BinlogTinyIntIT#shouldHandleNegativeTinyIntOneAsTrue`, inserting `-1` during streaming); MySQL 8.2 and MariaDB 11.4.3 `TinyIntIT` suites pass.

### DIFF
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/TinyIntOneToBooleanConverter.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/converters/TinyIntOneToBooleanConverter.java
@@ -79,11 +79,12 @@ public class TinyIntOneToBooleanConverter implements CustomConverter<SchemaBuild
                 return x;
             }
             else if (x instanceof Number) {
-                return ((Number) x).intValue() > 0;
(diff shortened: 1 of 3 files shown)


Naros (approved):

Naros:
Applied @minleejae, great job!

PR #7710, opened 2026-07-22
DBZ-1876 Reduce TypeRegistry and SqlTypeMapper WARN log volume

### LINKED ISSUE #1876: PostgreSQL connector: WARN spam from TypeRegistry in multi-schema deployments
## Bug report

`io.debezium.connector.postgresql.PostgresConnector`, version `3.4.3.Final` (also reproduced on `3.5.0.Final`)

---

```json
{
  "plugin.name": "pgoutput",
  "publication.name": "cdc",
  "table.include.list": ".+\\.(table_a|table_b|table_c|table_d|...)$",
  "tasks.max": "1",
  "snapshot.mode": "no_data",
  "heartbeat.interval.ms": "5000"
}
```

Note: no `schema.include.list` — we capture from all tenant schemas by design.

---

Amazon Aurora PostgreSQL, managed (AWS RDS), MSK Connect.

---

The "Type is already mapped" condition should not log at WARN level when the duplicate is simply a same-named type in a different schema. This is expected in any multi-tenant or multi-schema PostgreSQL deployment and is not indicative of a problem.

---

**What behavior do you see?**

~17,000 WARN-level log lines per connector startup:

```
WARN  Type [oid:278431, name:accounts] is already mapped   [io.debezium.connector.postgresql.TypeRegistry]
WARN  Type [oid:278432, name:users] is already mapped       [io.debezium.connector.postgresql.TypeRegistry]
WARN  Type [oid:278433, name:enrollments] is already mapped [io.debezium.connector.postgresql.TypeRegistry]
... (repeated for every type × every tenant schema)
```

On MSK Connect, there is no per-class log level configuration, so we cannot suppress these via logging config. The volume drowns out genuine warnings from the connector.

---

Yes — the `addType()` method is unchanged in `3.5.0.Final`. Not verified against any Alpha/Beta/CR builds, but the code path has not changed on `main`.

---

The relevant logs are the WARN lines themselves — samples are shown above.

---

**How to reproduce the issue using our [[tutorial](https://github.com/debezium/debezium-examples/tree/main/tutorial)](https://github.com/debezium/debezium-examples/tree/main/tutorial) deployment?**

1. Create a PostgreSQL database with multiple schemas sharing the same table definitions:

```sql
CREATE SCHEMA tenant_1;
CREATE SCHEMA tenant_2;
CREATE SCHEMA tenant_3;
-- repeat for N schemas

CREATE TABLE tenant_1.table_a (id bigint primary key, name text);
CREATE TABLE tenant_2.table_a (id bigint primary key, name text);
CREATE TABLE tenant_3.table_a (id bigint primary key, name text);
-- repeat for each schema
```

2. Configure the connector with `table.include.list` matching all schemas (no `schema.include.list`):

```json
{
  "table.include.list": ".+\\.table_a$"
}
```

3. Start the connector. Observe WARN lines for every `table_a` composite type after the first schema.

The number of warnings scales as `(num_schemas - 1) × num_types_per_schema`.

---

### Cause

In [`[TypeRegistry.addType()](https://github.com/debezium/debezium/blob/v3.5.0.Final/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java#L143-L147)`](https://github.com/debezium/debezium/blob/v3.5.0.Final/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java#L143-L147), the `nameToType` map is keyed solely by type name:

```java
if (!nameToType.containsKey(type.getName())) {
    nameToType.put(type.getName(), type);
}
else {
    LOGGER.warn("Type [oid:{}, name:{}] is already mapped", type.getOid(), type.getName());
}
```

The underlying `SQL_TYPES` query joins `pg_namespace` but the namespace is never propagated into the map key. When schema `tenant_1` and schema `tenant_2` both define a composite type `accounts`, the second one triggers a WARN.

In our environment (Canvas LMS — a multi-tenant PostgreSQL application), a single database has hundreds of schemas sharing identical table/type definitions. This produces **~17,000 WARN lines per connector startup**, repeated on every restart or rebalance.

### Why existing fixes don't help

- **PR #7150 (schema-filtered type loading)**: Filters `pg_type` to `schema.include.list` schemas. Our connector captures tables from _all_ schemas (`table.include.list: ".+\\.(table_names...)$"`) — we have no `schema.include.list` because every tenant schema is needed. Even if we enumerated all schemas, the duplicates across them would still trigger warnings.
- **DBZ-5182 (ascending OID order)**: Addressed type _collisions_ from extension schemas overriding built-in types, but did not change the log level or namespace handling.
- **#1683 / #1800 (performance fixes)**: Reduced startup time from OID lookups and double initialization, but the WARN-level logging remains.

### Relationship to [DBZ-8519](https://github.com/debezium/dbz/issues/1323)

[DBZ-8519](https://github.com/debezium/dbz/issues/1323) reports the same symptom but from a different angle — the reporter has many schemas and expected `schema.include.list` to suppress the warnings. Our case is distinct: we intentionally capture all schemas, so schema filtering is not a viable workaround. The underlying code issue (name-only keying, WARN log level) is the same.

### Suggested fix

Downgrade the log level from `LOGGER.warn` to `LOGGER.debug` in `TypeRegistry.addType()`. The "already mapped" condition is a normal consequence of PostgreSQL's type system when multiple schemas define same-named types. It is not a correctness issue — the first-registered type wins, and `oidToType` still maps every OID correctly. The `nameToType` map is used for well-known type name lookups (e.g., `hstore`, `geometry`) that don't carry schema context, so the first-wins behavior is the right semantics here.

MagicGeny:
Fixes debezium/dbz#1876

### Problem

In PostgreSQL deployments with many schemas (multi-tenant, schema-per-tenant, or databases with heavy use of composite types from `CREATE TABLE`), the PostgreSQL connector emits a high volume of WARN-level log lines from `io.debezium.connector.postgresql.TypeRegistry` and its nested `SqlTypeMapper` at startup and during streaming. Reported volumes reach ~17,000 WARN lines per connector startup on managed platforms such as MSK Connect, where per-class log-level overrides are not available, drowning out genuine warnings.

Four log sites contribute to the volume:

1. `TypeRegistry.addType()` — emits a WARN for every schema-qualified type name that gets re-mapped (same `schema.typeName` observed again with a different resolved type, e.g., during rapid DROP/CREATE churn). This is typically not actionable at WARN level: `oidToType` still maps every OID correctly, and name-based lookups continue to follow the existing “first one wins” alias semantics for well-known unqualified names (`hstore`, `geometry`, etc.).
2. `TypeRegistry.get(int oid)` — emits a WARN every time the streaming decoder hands the registry an OID that is not in the pre-loaded cache (e.g., types created after `prime()`, extension types, types resolved after a DROP/RECREATE cycle).
3. `TypeRegistry.get(String name)` — same pattern for name-based lookups that miss the cache.
4. `SqlTypeMapper.getSqlType()` — emits a WARN when the custom-statement resolution path throws and falls back to `TypeInfo#getSQLType()`, which can be hit repeatedly in multi-schema deployments depending on catalog state and driver behavior.

In observed multi-schema/tenant scenarios, these log events are expected and typically not actionable at WARN level.

### Solution

The change is a logging-only fix; no behavior is altered.

- `addType()` collision: per-type message demoted to `LOGGER.debug`. A single `LOGGER.warn` summary is emitted the first time a collision is detected for a given `TypeRegistry` instance, gated by a private `typeCollisionWarned` flag. This preserves the signal for operators while suppressing the per-type flood, matching the direction indicated by the maintainer in the issue thread.
- `get(int oid)` miss: `LOGGER.warn` → `LOGGER.debug`.
- `get(String name)` miss: `LOGGER.warn` → `LOGGER.debug`.
- `SqlTypeMapper.getSqlType()` exception catch-block: `LOGGER.warn` → `LOGGER.debug` (exception path only; behavior unchanged).

### Files changed

- `debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java`
- `debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TypeRegistryWarnIT.java`

### Testing

`TypeRegistryWarnIT` was added as a lightweight integration test that constructs a `TypeRegistry` directly against a PostgreSQL container (no full connector startup) and uses Debezium's `LogInterceptor` to assert the absence of WARN-level messages. Three tests cover the streaming-side miss paths:

- `shouldNotEmitWarnOnUnknownOidLookup` — single `get(int)` miss.
- `shouldNotEmitWarnForRepeatedUnknownOidLookups` — repeated `get(int)` misses, asserts zero WARNs after N lookups.
- `shouldNotEmitWarnOnUnknownTypeNameLookup` — `get(String)` miss.

The fourth test from the initial draft, `shouldNotEmitWarnOnDuplicateTypeNameMapping`, was removed before submission. Cross-schema duplicate names no longer collide because `nameToType` is keyed by schema-qualified name (`schema_a.mytype` vs `schema_b.mytype`). The remaining realistic trigger for the `addType()` re-mapping branch is DROP/CREATE churn that causes the same schema-qualified name to be observed again with a different OID. Simulating that deterministically in a hermetic test tends to be brittle (timing-dependent and/or reliant on catalog behavior), so coverage of the one-time summary WARN is left to code review and manual verification, while the high-volume streaming miss paths are verified by `TypeRegistryWarnIT`.

The `SqlTypeMapper.getSqlType()` exception-path demotion is intentionally not covered by a dedicated test. The branch only fires when the custom-statement path throws or the underlying `TypeInfo#getSQLType()` call fails, neither of which can be provoked reliably in a real-container integration test without either mocking the JDBC layer (inconsistent with the test style used here, which exercises real PostgreSQL) or corrupting a system catalog (destructive and non-portable). The change is a pure log-level demotion with no behavioral effect, so the cost/benefit of a brittle mock-based test does not justify its inclusion.

### Backward compatibility

Operators who relied on the previous per-type WARN lines for diagnostics can recover the full signal by setting the `io.debezium.connector.postgresql.TypeRegistry` logger to `DEBUG`. The one-time summary WARN preserves the high-level indication that schema-qualified type name re-mapping was encountered.

### Related

- dbz#1876
- dbz#9455 (schema-filtered type loading, complementary optimization)
- dbz#5182 (ascending OID order)

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
@@ -135,6 +135,12 @@ public class TypeRegistry {
     private int ltreeArrayOid = Integer.MIN_VALUE;
     private int tsVectorOid = Integer.MIN_VALUE;
 
+    /**
+     * Tracks whether the one-time summary WARN about schema-qualified type name re-mapping has already
+     * been emitted for this registry instance. See dbz#1876.
+     */
+    private boolean typeCollisionWarned = false;
+
     public TypeRegistry(PostgresConnection connection) {
         this(connection, Collections.emptySet());
     }
@@ -180,7 +186,13 @@ public class TypeRegistry {
             }
             PostgresType currentType = nameToType.get(qualifiedName);
             if (!currentType.equals(type)) {
-                LOGGER.warn("Type [oid:{}, name:{}] is already mapped", type.getOid(), qualifiedName);
+                LOGGER.debug("Type [oid:{}, name:{}] is already mapped", type.getOid(), qualifiedName);
+                if (!typeCollisionWarned) {
+                    typeCollisionWarned = true;
+                    LOGGER.warn(
+                            "Type name re-mapped: Type [oid:{}, name:{}] is already mapped (existing oid:{}); further occurrences for this registry instance will be logged at DEBUG level.",
+                            type.getOid(), qualifiedName, currentType.getOid());
+                }
             }
         }
 
@@ -248,7 +260,7 @@ public class TypeRegistry {
         if (r == null) {
             r = resolveUnknownType(oid);
             if (r == null) {
-                LOGGER.warn("Unknown OID {} requested", oid);
+                LOGGER.debug("Unknown OID {} requested", oid);
                 r = PostgresType.UNKNOWN;
             }
         }
@@ -356,7 +368,7 @@ public class TypeRegistry {
                 : cleanParts[0];
         r = resolveUnknownType(cleanName);
         if (r == null) {
-            LOGGER.warn("Unknown type named {} requested", name);
+            LOGGER.debug("Unknown type named {} requested", name);
             r = PostgresType.UNKNOWN;
         }
         return r;
@@ -711,7 +723,7 @@ public class TypeRegistry {
                     return getTypeInfo(connection).getSQLType(typeName);
                 }
                 catch (Exception e) {
-                    LOGGER.warn("Failed to obtain SQL type information for type {} via custom statement, falling back to TypeInfo#getSQLType()", typeName, e);
+                    LOGGER.debug("Failed to obtain SQL type information for type {} via custom statement, falling back to TypeInfo#getSQLType()", typeName, e);
                     return getTypeInfo(connection).getSQLType(typeName);
                 }
             }
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TypeRegistryWarnIT.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TypeRegistryWarnIT.java
@@ -1,0 +1,60 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.postgresql;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import java.sql.SQLException;
+
+import org.junit.jupiter.api.Test;
+
+import io.debezium.connector.postgresql.connection.PostgresConnection;
+import io.debezium.junit.logging.LogInterceptor;
+import io.debezium.util.Testing;
+
+public class TypeRegistryWarnIT implements Testing {
+
+    @Test
+    void shouldNotEmitWarnOnUnknownOidLookup() throws SQLException {
+        try (PostgresConnection connection = TestHelper.create()) {
+            final var logInterceptor = LogInterceptor.forClass(TypeRegistry.class);
+            final var registry = new TypeRegistry(connection);
+            logInterceptor.clear();
+
+            registry.get(999999);
+
+            assertThat(logInterceptor.containsWarnMessage("Unknown OID")).isFalse();
+        }
+    }
+
+    @Test
+    void shouldNotEmitWarnForRepeatedUnknownOidLookups() throws SQLException {
+        try (PostgresConnection connection = TestHelper.create()) {
+            final var logInterceptor = LogInterceptor.forClass(TypeRegistry.class);
+            final var registry = new TypeRegistry(connection);
+            logInterceptor.clear();
+
+            registry.get(999999);
+            registry.get(888888);
+            registry.get(777777);
+
+            assertThat(logInterceptor.countOccurrences("Unknown OID")).isEqualTo(0);
+        }
+    }
+
+    @Test
+    void shouldNotEmitWarnOnUnknownTypeNameLookup() throws SQLException {
+        try (PostgresConnection connection = TestHelper.create()) {
+            final var logInterceptor = LogInterceptor.forClass(TypeRegistry.class);
+            final var registry = new TypeRegistry(connection);
+            logInterceptor.clear();
+
+            registry.get("nonexistent_type_name");
+
+            assertThat(logInterceptor.containsWarnMessage("Unknown type named")).isFalse();
+        }
+    }
+}


debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/TypeRegistryWarnIT.java
```diff
@@ -0,0 +1,60 @@
...
+import org.junit.jupiter.api.Test;
+
+import io.debezium.connector.postgresql.connection.PostgresConnection;
+import io.debezium.junit.logging.LogInterceptor;
+import io.debezium.util.Testing;
+
+public class TypeRegistryWarnIT implements Testing {
+
+    @Test
+    void shouldNotEmitWarnOnUnknownOidLookup() throws SQLException {
```
Naros, line 21:
```suggestion
    @FixFor("debezium/dbz#1876")
    void shouldNotEmitWarnOnUnknownOidLookup() throws SQLException {
```
Lets please annotate this and the other newly added test methods.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/TypeRegistry.java
```diff
@@ -135,6 +135,12 @@ private static Map<String, String> getLongTypeNames() {
     private int ltreeArrayOid = Integer.MIN_VALUE;
     private int tsVectorOid = Integer.MIN_VALUE;
 
+    /**
+     * Tracks whether the one-time summary WARN about schema-qualified type name re-mapping has already
+     * been emitted for this registry instance. See dbz#1876.
+     */
+    private boolean typeCollisionWarned = false;
```
Naros, line 142:
```suggestion
    private boolean typeCollisionWarned = false;
```
We don't need a verbose comment here, the intent is clear.

Naros (commented):
Hi @MagicGeny overall LGTM, just a few nit-picks.

vjuranek:
Test failures are related, could you please take a look @MagicGeny ? Thanks!

PR #7717, opened 2026-07-24
debezium/dbz#2280 Do not apply column include/exclude list to the sig…

b-goyal:
Fixes debezium/dbz#2280

## Description
The column filter derived from `column.include.list` was applied uniformly to every table, including the signal data collection. When the signal table's columns were not listed in `column.include.list`, its value schema was built with zero fields, so a source-channel signal was constructed with an empty `Struct` and silently discarded ("The signal event should have 3 fields but has 0"). As a result, neither incremental nor blocking snapshots triggered via the source channel would run.

`RelationalDatabaseSchema` now wraps the configured column filter so that the required signal columns (`id`, `type`, `data`) of the signal data collection are always included when `column.include.list` is used. Matching is case-insensitive and limited to those three columns, so a signal table that also has extra columns still yields exactly three fields. Exclude-list mode is left untouched, since it already lets the signal columns through by default. This mirrors `RelationalTableFilters`, which always includes the signal data collection regardless of `table.include.list` / `table.exclude.list`.

Also clarifies the `CommonConnectorConfig` warning so it points at column filtering as the likely cause when a signal event does not have exactly three fields.

## Tests
- Unit test for the column-filter wrapping in `TableSchemaBuilderTest`.
- Incremental and blocking snapshot integration tests added to the shared base classes (`AbstractIncrementalSnapshotTest`, `AbstractBlockingSnapshotTest`) so every relational connector is covered. Connectors whose signal data-collection id carries a database/catalog prefix (Oracle, SQL Server) override the test's `column.include.list` value.
- Verified locally against SQL Server (incremental + blocking); the remaining connectors run in CI.

## PR Checklist

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/relational/RelationalDatabaseSchema.java
+++ b/debezium-connector-common/src/main/java/io/debezium/relational/RelationalDatabaseSchema.java
@@ -25,6 +25,7 @@ import io.debezium.relational.Tables.TableFilter;
 import io.debezium.relational.mapping.ColumnMappers;
 import io.debezium.schema.DatabaseSchema;
 import io.debezium.spi.topic.TopicNamingStrategy;
+import io.debezium.util.Strings;
 
 /**
  * A {@link DatabaseSchema} of a relational database such as Postgres. Provides information about the physical structure
@@ -34,6 +35,9 @@ import io.debezium.spi.topic.TopicNamingStrategy;
  */
 public abstract class RelationalDatabaseSchema implements DatabaseSchema<TableId> {
     private final static Logger LOG = LoggerFactory.getLogger(RelationalDatabaseSchema.class);
+
+    // The columns a signal data collection must expose (id, type, data) for source-channel signals to be parsed.
+    private static final Set<String> SIGNAL_REQUIRED_COLUMNS = Set.of("id", "type", "data");
 
     private final RelationalDatabaseConnectorConfig config;
     private final TopicNamingStrategy<TableId> topicNamingStrategy;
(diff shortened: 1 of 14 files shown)


debezium-connector-common/src/main/java/io/debezium/relational/RelationalDatabaseSchema.java
```diff
@@ -64,6 +65,25 @@ protected RelationalDatabaseSchema(RelationalDatabaseConnectorConfig config, Top
         this.taskContext = taskContext;
     }
 
+    // The columns a signal data collection must expose (id, type, data) for source-channel signals to be parsed.
+    private static final Set<String> SIGNAL_REQUIRED_COLUMNS = Set.of("id", "type", "data");
```
Naros: Can we keep this with the other static members?

debezium-connector-common/src/main/java/io/debezium/relational/RelationalDatabaseSchema.java
```diff
@@ -64,6 +65,25 @@ protected RelationalDatabaseSchema(RelationalDatabaseConnectorConfig config, Top
...
+
+    /**
+     * Wraps the configured column filter so that the required columns of the signal data collection
+     * ({@code id}, {@code type}, {@code data}) are always included when {@code column.include.list} is used.
+     */
+    static ColumnNameFilter overrideColumnFilter(RelationalDatabaseConnectorConfig config, ColumnNameFilter columnFilter) {
+        final boolean includeListMode = !Strings.isNullOrBlank(
+                config.getConfig().getString(RelationalDatabaseConnectorConfig.COLUMN_INCLUDE_LIST));
```
Naros: In `RelationalDatabaseConnectorConfig` we use `Strings.isNullOrEmpty`, so lets use that here so that the same expectation/outcome is reached between the two spots.

debezium-embedded/src/test/java/io/debezium/pipeline/source/snapshot/incremental/AbstractSnapshotTest.java
```diff
@@ -278,6 +278,16 @@ protected String pkFieldName() {
...
 
+    /**
+     * A {@code column.include.list} value covering the data table's own columns (pk and value) but not the signal
+     * data collection's columns, so the signal table is column filtered. Built from {@link #pkFieldName()} /
+     * {@link #valueFieldName()} so it is correct for every connector, including identifier case. Used by the
+     * incremental and blocking snapshot signal-filtering tests (debezium/dbz#2280).
+     */
+    protected String signalExcludingColumnIncludeList() {
+        return ".*\\." + pkFieldName() + ",.*\\." + valueFieldName();
+    }
```
Naros, line 283: While not blocking, does this introduce a potential issue with such a broad regex to match any table/schema/database?

Naros (commented):
Overall this LGTM, just a few inline nit-picks @b-goyal.

debezium-embedded/src/test/java/io/debezium/pipeline/source/snapshot/incremental/AbstractSnapshotTest.java
b-goyal: Yeah, I should have avoided regex. Updated now.

b-goyal:
Thanks @Naros — addressed the review:

1. Moved `SIGNAL_REQUIRED_COLUMNS` up with the other static members.
2. Switched to `Strings.isNullOrEmpty` to match `RelationalDatabaseConnectorConfig`.
3. Replaced the broad `.*` test regex with a table-scoped `column.include.list` (built from `tableDataCollectionId()`) so it no longer matches unrelated tables; Oracle and SQL Server override it since their data-collection id carries a database/catalog prefix.
4. Improved the warning message with more details, as per the Zulip chat feedback.
5. Trimmed/removed extra comments/java docs.

b-goyal:
The failing checks look unrelated/flaky. Could we re-run the failed jobs please?

Naros (approved):
LGTM, just waiting on CI.

Naros:
Rebased on the latest main to clear update test failures.

Naros (commented):
Hi @b-goyal after reviewing CI, it seems the Informix failures are related. Can you check and see if a companion Informix PR is needed in it's repo?

b-goyal:
@Naros created informix PR here https://github.com/debezium/debezium-connector-informix/pull/197

Naros:
@b-goyal I believe there is a small docs update needed about the behavior change, if you could take a look.

PR #7724, opened 2026-07-27
debezium/dbz#2305 Skip archive-log destination validation for snapshot-only Oracle runs

wagnercsantos:
Fixes debezium/dbz#2305: `OracleConnectorTask#start` validated the archive-log destination unconditionally, right after the redo-log validation. The archive destination is only consumed by the LogMiner streaming path, so on snapshot-only runs (`snapshot.mode=initial_only`) the check is unnecessary — and fatal on managed deployments (e.g. Oracle Autonomous Database) where `V$ARCHIVE_DEST` is hidden.

Gate the archive-dest validation on the same `redoLogRequired(...)` condition already used for the adjacent redo-log validation. Streaming runs are unaffected (the validation still runs); snapshot-only runs no longer require streaming-only configuration.

Validated on two Oracle deployments: `initial_only` now completes on Autonomous Database, and full LogMiner streaming still works (and still validates the archive destination) on a self-managed 21c instance.

This fix surfaced while contributing the Databricks Zerobus sink for Debezium Server (debezium/dbz#2279), validating Oracle as a source.

cc @Naros

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
@@ -138,7 +138,12 @@ public class OracleConnectorTask extends BaseSourceTask<OraclePartition, OracleO
 
         validateRedoLogConfiguration(connectorConfig, snapshotterService);
 
-        connectorConfig.getArchiveDestinationNameResolver().validate(jdbcConnection);
+        // The archive log destination is only consumed by the LogMiner streaming path; validating it
+        // is pointless (and, on managed/hidden-V$ deployments such as Oracle Autonomous Database,
+        // fatal) when this run will not stream. Gate it on the same redo-log requirement used above.
+        if (redoLogRequired(connectorConfig, snapshotterService)) {
+            connectorConfig.getArchiveDestinationNameResolver().validate(jdbcConnection);
+        }
 
         OracleOffsetContext previousOffset = previousOffsets.getTheOnlyOffset();
 
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
@@ -138,12 +138,7 @@ public class OracleConnectorTask extends BaseSourceTask<OraclePartition, OracleO
 
         validateRedoLogConfiguration(connectorConfig, snapshotterService);
 
-        // The archive log destination is only consumed by the LogMiner streaming path; validating it
-        // is pointless (and, on managed/hidden-V$ deployments such as Oracle Autonomous Database,
-        // fatal) when this run will not stream. Gate it on the same redo-log requirement used above.
-        if (redoLogRequired(connectorConfig, snapshotterService)) {
-            connectorConfig.getArchiveDestinationNameResolver().validate(jdbcConnection);
-        }
+        connectorConfig.getArchiveDestinationNameResolver().validate(jdbcConnection);
 
         OracleOffsetContext previousOffset = previousOffsets.getTheOnlyOffset();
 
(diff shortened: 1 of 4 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorTask.java
```diff
@@ -138,7 +138,12 @@ public ChangeEventSourceCoordinator<OraclePartition, OracleOffsetContext> start(
 
         validateRedoLogConfiguration(connectorConfig, snapshotterService);
 
-        connectorConfig.getArchiveDestinationNameResolver().validate(jdbcConnection);
+        // The archive log destination is only consumed by the LogMiner streaming path; validating it
+        // is pointless (and, on managed/hidden-V$ deployments such as Oracle Autonomous Database,
+        // fatal) when this run will not stream. Gate it on the same redo-log requirement used above.
+        if (redoLogRequired(connectorConfig, snapshotterService)) {
+            connectorConfig.getArchiveDestinationNameResolver().validate(jdbcConnection);
+        }
```
Naros: While this solves one use case, if `OracleConnection::validateLogPosition` ever fires, you'll face the same problem because that will also refire `validate`.

The main issue is that Debezium does not support Oracle Autonomous at all. I understand you're using it purely for snapshot reasons, and we should likely support that in some capacity, but that's only due to happenstance, not officially intended.

What I would recommend, for now, would be to introduce a method on `OracleConnection` that identifies if we're connected to an Autonomous database (we'll need this eventually), e.g.
```sql
select sys_context('USERENV','CLOUD_SERVICE') from dual
```
If the result of this query is `OLTP`, `DWCS`, or `JDCS`, then OA is `true`, otherwise `false`

So with something like `OracleConnection#isAutonomous()`, then inside the resolver, just guard the `validate` method and return early. Place a comment there that this only satisfies the validation code path, but that if any part of the code expects a destination name for queries, the queries will fail on Autonomous.

WDYT?

I'll think about how to redesign the resolver in a future pass, because its needed to do this complex logic once rather than everytime we build a SQL query, but maybe making it dynamic is the better solution, as it would facilitate online DBA changes without a connector restart 🤔

wagnercsantos: Reworked per your suggestion 👍

Added `OracleConnection#isAutonomous()` (via `SYS_CONTEXT('USERENV','CLOUD_SERVICE')`, true for `OLTP`/`DWCS`/`JDCS`) and moved the guard into `ArchiveDestinationNameResolver#validate`, which returns early on Autonomous with a comment noting it only satisfies the validation path — streaming queries that need a physical destination name would still fail there, which Debezium doesn't officially support. Reverted the `OracleConnectorTask` change.

Because the guard now lives in the resolver, it also covers the `validateLogPosition` path you flagged, since that re-invokes `validate()` too. Added a unit test (`shouldSkipValidationOnAutonomousDatabase`).

I couldn't re-test against a live Autonomous instance in this pass (the one I used was torn down), so I've relied on the unit test + CI.

Naros (approved):

wagnercsantos:
The two failing `testing-farm` checks (`oracle-19-xs` and `oracle-19-se-xs`) are an
unrelated flake, not caused by this change:

- The single failure is `IncrementalSnapshotCaseSensitiveIT.stopCurrentIncrementalSnapshotWithoutCollectionsAndTakeNewNewIncrementalSnapshotAfterRestart`
  — an `org.awaitility.core.ConditionTimeoutException` (incremental snapshot didn't
  complete within the 5-minute await), not an assertion failure. Run totals:
  `Tests run: 485, Failures: 0, Errors: 1`.
- It's in the incremental-snapshot area, unrelated to archive-log/Autonomous handling.
  This PR's own test, `ArchiveDestinationResolverTest`, passed (5/5), including
  `shouldSkipValidationOnAutonomousDatabase` (logs "skipping archive destination validation").
- It's flaky, not deterministic: the same test in the sibling class `IncrementalSnapshotIT`
  passed in the same run (~54s), and 9 of the 11 Oracle variants (incl. `oracle-26-xs`)
  are green — only the slower Oracle 19 XE runners timed out.

A re-run of those two jobs should clear it.

Naros:
Applied, thanks @wagnercsantos

PR #7728, opened 2026-07-28
debezium/dbz#2012: Redesign direct mode around  __$command_id

### LINKED ISSUE #2012: Sqlserver connector: Data loss with direct mode on update with  fetch size > 0
## Bug report

SQL Server connector. Reproduced on the **latest `main` (3.6.0-SNAPSHOT)**, built and run locally. The behavior also affects earlier releases (the relevant code is unchanged).

---

Key settings that trigger it:
```json
{
  "connector.class": "io.debezium.connector.sqlserver.SqlServerConnector",
  "data.query.mode": "direct",
  "streaming.fetch.size": "5",
  "snapshot.mode": "no_data",
  "database.names": "bgoyal_test",
  "table.include.list": "dbo.users"
}
```
`streaming.fetch.size` is set to `5` here to reproduce minimally; in production it is the default `10000` (any non-zero value reproduces it once the transaction exceeds the fetch boundary).

---

Microsoft SQL Server 2019 (RTM-CU32) 15.0.4430.1, self-managed (VM/EC2). CDC enabled via `sys.sp_cdc_enable_table`.

---

Every change captured by SQL Server CDC for a committed transaction should be emitted to Kafka exactly once, independent of `data.query.mode` and `streaming.fetch.size`. An update to a unique-index column should produce all of its delete/insert events.

---

**What behavior do you see?**

Data loss of update events when using `data.query.mode=direct` with `streaming.fetch.size` > 0. The loss happens only when the update is to a column that is part of the table's **unique index** — SQL Server records this as a "deferred update", i.e. a **delete followed by an insert with the same `__$seqval`**. If the transaction is a large bulk update that does not fully return in one fetch (because of `streaming.fetch.size`), the attempt to fetch the rest of the transaction returns only a subset of the rows, so some events are silently dropped.

Example. Table:
```sql
CREATE TABLE dbo.users (
    id         INT          NOT NULL PRIMARY KEY CLUSTERED,
    email      VARCHAR(100) NOT NULL,
    name       VARCHAR(100) NOT NULL,
    updated_at DATETIME     NOT NULL DEFAULT getutcdate()
);
CREATE UNIQUE NONCLUSTERED INDEX UX_users_email ON dbo.users(email);
EXEC sys.sp_cdc_enable_table @source_schema='dbo', @source_name='users',
     @role_name=NULL, @supports_net_changes=0;
```
Update to the unique-index column:
```sql
UPDATE dbo.users SET email = name + '_v0@example.com' WHERE id BETWEEN 1 AND 5;
```
On the CDC table this produces 10 rows (5 deletes + 5 inserts). Ordered as the **direct-mode** query orders them (`ORDER BY __$start_lsn, __$command_id, __$seqval, __$operation`):
```
commit_lsn            |seqval                |__$command_id|__$operation|name |email
----------------------+----------------------+-------------+------------+-----+--------------------
0x00000a24000018a80015|0x00000a24000018a80003|            1|           1|alice|alice_v1@example.com
0x00000a24000018a80015|0x00000a24000018a80004|            2|           1|bob  |bob_v1@example.com
0x00000a24000018a80015|0x00000a24000018a80005|            3|           1|carol|carol_v1@example.com
0x00000a24000018a80015|0x00000a24000018a80006|            4|           1|dave |dave_v1@example.com
0x00000a24000018a80015|0x00000a24000018a80007|            5|           1|eve  |eve_v1@example.com
0x00000a24000018a80015|0x00000a24000018a80003|            6|           2|alice|alice_v0@example.com
0x00000a24000018a80015|0x00000a24000018a80004|            7|           2|bob  |bob_v0@example.com
0x00000a24000018a80015|0x00000a24000018a80005|            8|           2|carol|carol_v0@example.com
0x00000a24000018a80015|0x00000a24000018a80006|            9|           2|dave |dave_v0@example.com
0x00000a24000018a80015|0x00000a24000018a80007|           10|           2|eve  |eve_v0@example.com
```
Ordered as the **function-mode** query orders them (`ORDER BY __$start_lsn, __$seqval, __$operation` — no `__$command_id`), the delete/insert of each row stay adjacent and `__$seqval` is monotonic:
```
commit_lsn            |seqval                |__$operation|name |email
----------------------+----------------------+------------+-----+--------------------
0x00000a24000018a80015|0x00000a24000018a80003|           1|alice|alice_v1@example.com
0x00000a24000018a80015|0x00000a24000018a80003|           2|alice|alice_v0@example.com
0x00000a24000018a80015|0x00000a24000018a80004|           1|bob  |bob_v1@example.com
0x00000a24000018a80015|0x00000a24000018a80004|           2|bob  |bob_v0@example.com
0x00000a24000018a80015|0x00000a24000018a80005|           1|carol|carol_v1@example.com
0x00000a24000018a80015|0x00000a24000018a80005|           2|carol|carol_v0@example.com
0x00000a24000018a80015|0x00000a24000018a80006|           1|dave |dave_v1@example.com
0x00000a24000018a80015|0x00000a24000018a80006|           2|dave |dave_v0@example.com
0x00000a24000018a80015|0x00000a24000018a80007|           1|eve  |eve_v1@example.com
0x00000a24000018a80015|0x00000a24000018a80007|           2|eve  |eve_v0@example.com
```
With `streaming.fetch.size=5`, the Kafka topic receives **5 deletes and only 1 create** (the `eve` / `command_id 10` row). The 4 inserts for `alice/bob/carol/dave` are missing:
```
op=d id=1 change_lsn=00000a24:000018a8:0003
op=d id=2 change_lsn=00000a24:000018a8:0004
op=d id=3 change_lsn=00000a24:000018a8:0005
op=d id=4 change_lsn=00000a24:000018a8:0006
op=d id=5 change_lsn=00000a24:000018a8:0007
op=c id=5 change_lsn=00000a24:000018a8:0007 event_serial_no=2     <-- only surviving insert
```
All inserts with `seqval < 00000a24:000018a8:0007` are dropped. Only `command_id 10` survives because its `seqval` (`...0007`) is `>=` the last-seen `seqval` and its operation is `> 1`. The issue reproduces whenever the number of (deferred) updates exceeds the `streaming.fetch.size` boundary.

**Likely cause:** the first fetch returns the 5 deletes (`command_id 1–5`, ordered by `command_id`), so the resume position becomes the *highest delete* `seqval` (`...0007`). The next fetch resumes with `getNextResultSet()` using a `WHERE __$seqval > ?` predicate that does **not** account for `__$command_id`, so the inserts whose `seqval` is below that high-water mark are skipped. In other words, in direct mode the query `ORDER BY` uses `__$command_id` but the pagination resume key uses only `__$seqval`, and the two disagree.

---

Yes — reproduced on the latest `main` (3.6.0-SNAPSHOT) built locally. It does **not** occur with `data.query.mode=function` (whose `ORDER BY` has no `__$command_id`, so `__$seqval` stays monotonic and delete/insert pairs remain adjacent).

---

```
TRACE  Getting top 5 changes for table Capture instance "dbo_orders" [...changeTableId=...cdc.dbo_orders_CT...] in range [00000a23:0001b318:0010-00000000:00000000:0000-0, 00000a23:0001b838:0015]   (io.debezium...SqlServerConnection:421)
TRACE  Processing change ChangeTableResultSet{... currentChangePosition=00000a23:0001b838:0015(00000a23:0001b838:0003,1)}   (...SqlServerStreamingChangeEventSource:314)
TRACE  Processing change ChangeTableResultSet{... currentChangePosition=00000a23:0001b838:0015(00000a23:0001b838:0008,1)}
TRACE  Processing change ChangeTableResultSet{... currentChangePosition=00000a23:0001b838:0015(00000a23:0001b838:000b,1)}
TRACE  Processing change ChangeTableResultSet{... currentChangePosition=00000a23:0001b838:0015(00000a23:0001b838:000e,1)}
TRACE  Processing change ChangeTableResultSet{... currentChangePosition=00000a23:0001b838:0015(00000a23:0001b838:0011,1)}
TRACE  Getting top 5 changes for table Capture instance "dbo_orders" [...] in range [00000a23:0001b838:0015-00000a23:0001b838:0011-1, 00000a23:0001b838:0015]   (io.debezium...SqlServerConnection:421)
INFO   11 records sent during previous ... last recorded offset ... is {event_serial_no=1, commit_lsn=00000a23:0001b838:0015, change_lsn=00000a23:0001b838:0011}   (io.debezium...BaseSourceTask:387)
TRACE  Processing change ChangeTableResultSet{... currentChangePosition=00000a23:0001b838:0015(00000a23:0001b838:0011,2)}   <-- only the op2 insert at the high-water seqval survives
```

---

**How to reproduce the issue using our tutorial deployment?**

1. Create the `dbo.users` table and unique nonclustered index, enable CDC (DDL above).
2. Insert 5 rows; start the SQL Server connector with `data.query.mode=direct`, `streaming.fetch.size=5`, `snapshot.mode=no_data`, capturing `dbo.users`.
3. In a single transaction, update the unique-index column on all 5 rows:
   `UPDATE dbo.users SET email = name + '_v0@example.com' WHERE id BETWEEN 1 AND 5;`
4. Consume the topic: expected 5 deletes + 5 creates; actual 5 deletes + 1 create (4 rows lost). Increase the row count / `fetch.size` to confirm it scales with the fetch boundary.

The potential cause seems to be the `getNextResultSet()` call not taking `__$command_id` into account in direct mode.

b-goyal:
Fixes debezium/dbz#2012

In `data.query.mode=direct` the connector reads the change table sorted by `__$command_id` but did not use it on the WHERE clause, which loses data on a deferred update. This PR introduces `__$command_id` into the query, into TxLogPosition, and into the connector offsets so paging and restarts resume on the same key the rows are ordered by.                       
`function` mode is unchanged. The command id is not added to the emitted record, so the record schema and its registered version stay the same and consumers are unaffected by the upgrade.       

### Testing
- Unit test: Added where applicable
- Integration test: Added test which cover mode switch, different streaming.fetch.size and deferred updates case. Existing ITs should be sufficient to cover other common flows.
- Manual test: ITs performed manually + upgrade/downgrade from the change with both function/direct modes + sanity test 
- Performance test : not done.

### DIFF
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerChangeTablePointer.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerChangeTablePointer.java
@@ -44,7 +44,9 @@ public class SqlServerChangeTablePointer extends ChangeTableResultSet<SqlServerC
     private static final int COL_ROW_LSN = 2;
     private static final int COL_OPERATION = 3;
     private static final int COL_UPDATE_MASK = 4;
-    private static final int COL_DATA = 5;
+    private static final int COL_COMMAND_ID = 5;
+    private static final int COL_DATA_FUNCTION_MODE = 5;
+    private static final int COL_DATA_DIRECT_MODE = 6;
 
     private ResultSetMapper<Object[]> resultSetMapper;
     private final int columnDataOffset;
@@ -52,14 +54,20 @@ public class SqlServerChangeTablePointer extends ChangeTableResultSet<SqlServerC
     private final Lsn fromLsn;
     private final Lsn toLsn;
     private final int maxRowsPerResultSet;
+    private final boolean directMode;
+    private final TxLogPosition resumeFromPosition;
 
-    public SqlServerChangeTablePointer(SqlServerChangeTable changeTable, SqlServerConnection connection, Lsn fromLsn, Lsn toLsn, int maxRowsPerResultSet) {
-        super(changeTable, COL_DATA, maxRowsPerResultSet);
+    public SqlServerChangeTablePointer(SqlServerChangeTable changeTable, SqlServerConnection connection, Lsn fromLsn, Lsn toLsn,
+                                       TxLogPosition resumeFromPosition, int maxRowsPerResultSet,
+                                       SqlServerConnectorConfig.DataQueryMode dataQueryMode) {
+        super(changeTable, dataQueryMode == SqlServerConnectorConfig.DataQueryMode.DIRECT ? COL_DATA_DIRECT_MODE : COL_DATA_FUNCTION_MODE, maxRowsPerResultSet);
+        this.directMode = dataQueryMode == SqlServerConnectorConfig.DataQueryMode.DIRECT;
         // Store references to these because we can't get them from our superclass
-        this.columnDataOffset = COL_DATA;
+        this.columnDataOffset = this.directMode ? COL_DATA_DIRECT_MODE : COL_DATA_FUNCTION_MODE;
         this.connection = connection;
         this.fromLsn = fromLsn;
         this.toLsn = toLsn;
+        this.resumeFromPosition = resumeFromPosition;
         this.maxRowsPerResultSet = maxRowsPerResultSet;
     }
 
@@ -78,11 +86,12 @@ public class SqlServerChangeTablePointer extends ChangeTableResultSet<SqlServerC
 
     @Override
     protected TxLogPosition getNextChangePosition(ResultSet resultSet) throws SQLException {
-        return isCompleted() ? TxLogPosition.NULL
+        return isCompleted() ? (directMode ? TxLogPosition.NULL : TxLogPosition.NULL_LEGACY)
                 : TxLogPosition.valueOf(
                         Lsn.valueOf(resultSet.getBytes(COL_COMMIT_LSN)),
                         Lsn.valueOf(resultSet.getBytes(COL_ROW_LSN)),
-                        resultSet.getInt(COL_OPERATION));
+                        resultSet.getInt(COL_OPERATION),
+                        directMode ? resultSet.getInt(COL_COMMAND_ID) : null);
     }
 
     /**
@@ -97,13 +106,30 @@ public class SqlServerChangeTablePointer extends ChangeTableResultSet<SqlServerC
 
     @Override
     protected ResultSet getNextResultSet(TxLogPosition lastPositionSeen) throws SQLException {
-        if (lastPositionSeen == null || lastPositionSeen.equals(TxLogPosition.NULL)) {
-            return connection.getChangesForTable(getChangeTable(), fromLsn, toLsn, maxRowsPerResultSet);
+        if (!directMode) {
+            if (lastPositionSeen == null || lastPositionSeen.equals(TxLogPosition.NULL_LEGACY)) {
+                return connection.getChangesForTable(getChangeTable(), fromLsn, toLsn, maxRowsPerResultSet);
+            }
+            else {
+                return connection.getChangesForTable(getChangeTable(), lastPositionSeen.getCommitLsn(), lastPositionSeen.getInTxLsn(), lastPositionSeen.getOperation(),
+                        toLsn, maxRowsPerResultSet);
+            }
         }
-        else {
-            return connection.getChangesForTable(getChangeTable(), lastPositionSeen.getCommitLsn(), lastPositionSeen.getInTxLsn(), lastPositionSeen.getOperation(),
-                    toLsn, maxRowsPerResultSet);
+
+        // next page in a running connector
+        if (lastPositionSeen != null && !lastPositionSeen.equals(TxLogPosition.NULL)) {
+            return connection.getChangesForTable(getChangeTable(), lastPositionSeen.getCommitLsn(), lastPositionSeen.getInTxLsn(),
+                    lastPositionSeen.getOperation(), lastPositionSeen.getCommandId(), toLsn, maxRowsPerResultSet);
         }
+
+        // restarted connector which has not seen any position yet
+        if (resumeFromPosition != null && fromLsn.equals(resumeFromPosition.getCommitLsn())) {
+            return connection.getChangesForTable(getChangeTable(), fromLsn, resumeFromPosition.getInTxLsn(), 0,
+                    resumeFromPosition.getCommandId(), toLsn, maxRowsPerResultSet);
+        }
+
+        // running connector with new range.
+        return connection.getChangesForTable(getChangeTable(), fromLsn, Lsn.ZERO, 0, -1, toLsn, maxRowsPerResultSet);
     }
 
     @Override
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerChangeTablePointer.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerChangeTablePointer.java
@@ -44,7 +44,9 @@ public class SqlServerChangeTablePointer extends ChangeTableResultSet<SqlServerC
     private static final int COL_ROW_LSN = 2;
     private static final int COL_OPERATION = 3;
     private static final int COL_UPDATE_MASK = 4;
-    private static final int COL_DATA = 5;
+    private static final int COL_COMMAND_ID = 5;
+    private static final int COL_DATA_FUNCTION_MODE = 5;
+    private static final int COL_DATA_DIRECT_MODE = 6;
 
     private ResultSetMapper<Object[]> resultSetMapper;
     private final int columnDataOffset;
@@ -52,14 +54,20 @@ public class SqlServerChangeTablePointer extends ChangeTableResultSet<SqlServerC
     private final Lsn fromLsn;
     private final Lsn toLsn;
     private final int maxRowsPerResultSet;
+    private final boolean directMode;
+    private final TxLogPosition resumeFromPosition;
 
-    public SqlServerChangeTablePointer(SqlServerChangeTable changeTable, SqlServerConnection connection, Lsn fromLsn, Lsn toLsn, int maxRowsPerResultSet) {
-        super(changeTable, COL_DATA, maxRowsPerResultSet);
+    public SqlServerChangeTablePointer(SqlServerChangeTable changeTable, SqlServerConnection connection, Lsn fromLsn, Lsn toLsn,
+                                       TxLogPosition resumeFromPosition, int maxRowsPerResultSet,
+                                       SqlServerConnectorConfig.DataQueryMode dataQueryMode) {
+        super(changeTable, dataQueryMode == SqlServerConnectorConfig.DataQueryMode.DIRECT ? COL_DATA_DIRECT_MODE : COL_DATA_FUNCTION_MODE, maxRowsPerResultSet);
+        this.directMode = dataQueryMode == SqlServerConnectorConfig.DataQueryMode.DIRECT;
         // Store references to these because we can't get them from our superclass
-        this.columnDataOffset = COL_DATA;
+        this.columnDataOffset = this.directMode ? COL_DATA_DIRECT_MODE : COL_DATA_FUNCTION_MODE;
         this.connection = connection;
         this.fromLsn = fromLsn;
         this.toLsn = toLsn;
+        this.resumeFromPosition = resumeFromPosition;
         this.maxRowsPerResultSet = maxRowsPerResultSet;
     }
 
@@ -82,7 +90,8 @@ public class SqlServerChangeTablePointer extends ChangeTableResultSet<SqlServerC
                 : TxLogPosition.valueOf(
                         Lsn.valueOf(resultSet.getBytes(COL_COMMIT_LSN)),
                         Lsn.valueOf(resultSet.getBytes(COL_ROW_LSN)),
-                        resultSet.getInt(COL_OPERATION));
+                        resultSet.getInt(COL_OPERATION),
+                        directMode ? resultSet.getInt(COL_COMMAND_ID) : null);
     }
 
     /**
@@ -97,13 +106,30 @@ public class SqlServerChangeTablePointer extends ChangeTableResultSet<SqlServerC
 
     @Override
     protected ResultSet getNextResultSet(TxLogPosition lastPositionSeen) throws SQLException {
-        if (lastPositionSeen == null || lastPositionSeen.equals(TxLogPosition.NULL)) {
-            return connection.getChangesForTable(getChangeTable(), fromLsn, toLsn, maxRowsPerResultSet);
+        if (!directMode) {
+            if (lastPositionSeen == null || lastPositionSeen.equals(TxLogPosition.NULL)) {
+                return connection.getChangesForTable(getChangeTable(), fromLsn, toLsn, maxRowsPerResultSet);
+            }
+            else {
+                return connection.getChangesForTable(getChangeTable(), lastPositionSeen.getCommitLsn(), lastPositionSeen.getInTxLsn(), lastPositionSeen.getOperation(),
+                        toLsn, maxRowsPerResultSet);
+            }
         }
-        else {
-            return connection.getChangesForTable(getChangeTable(), lastPositionSeen.getCommitLsn(), lastPositionSeen.getInTxLsn(), lastPositionSeen.getOperation(),
-                    toLsn, maxRowsPerResultSet);
+
+        // next page in a running connector
+        if (lastPositionSeen != null && !lastPositionSeen.equals(TxLogPosition.NULL)) {
+            return connection.getChangesForTable(getChangeTable(), lastPositionSeen.getCommitLsn(), lastPositionSeen.getInTxLsn(),
+                    lastPositionSeen.getOperation(), lastPositionSeen.getCommandId(), toLsn, maxRowsPerResultSet);
         }
+
+        // restarted connector which has not seen any position yet.
+        if (resumeFromPosition != null && resumeFromPosition.getCommandId() != null) {
+            return connection.getChangesForTable(getChangeTable(), fromLsn, resumeFromPosition.getInTxLsn(), 0,
+                    resumeFromPosition.getCommandId(), toLsn, maxRowsPerResultSet);
+        }
+
+        // No command id to resume on, so repeat the transaction from its start.
+        return connection.getChangesForTable(getChangeTable(), fromLsn, toLsn, maxRowsPerResultSet);
     }
 
     @Override
--- a/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
+++ b/debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
@@ -85,10 +85,10 @@ public class SqlServerConnection extends JdbcConnection {
     private static final String LSN_TIMESTAMP_SELECT_STATEMENT_JOIN = "TODATETIMEOFFSET(ltm.tran_begin_time, DATEPART(TZOFFSET, SYSDATETIMEOFFSET()))";
     private static final String GET_ALL_CHANGES_FOR_TABLE_SELECT = "SELECT [__$start_lsn], [__$seqval], [__$operation], [__$update_mask], #, "
             + LSN_TIMESTAMP_SELECT_STATEMENT;
-    private static final String GET_ALL_CHANGES_FOR_TABLE_SELECT_DIRECT = "SELECT cdc_data.[__$start_lsn], cdc_data.[__$seqval], cdc_data.[__$operation], cdc_data.[__$update_mask], #, "
+    private static final String GET_ALL_CHANGES_FOR_TABLE_SELECT_DIRECT = "SELECT cdc_data.[__$start_lsn], cdc_data.[__$seqval], cdc_data.[__$operation], cdc_data.[__$update_mask], cdc_data.[__$command_id], #, "
             + LSN_TIMESTAMP_SELECT_STATEMENT_JOIN;
     private static final String GET_ALL_CHANGES_FOR_TABLE_FROM_FUNCTION = "FROM #db.cdc.#function(?, ?, N'all update old')";
-    private static final String GET_ALL_CHANGES_FOR_TABLE_FROM_DIRECT = "FROM #db.cdc.#table AS cdc_data LEFT JOIN #db.cdc.lsn_time_mapping ltm ON ltm.start_lsn = cdc_data.[__$start_lsn]";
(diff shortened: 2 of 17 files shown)


debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/TxLogPosition.java
```diff
@@ -84,29 +101,71 @@ else if (!commitLsn.equals(other.commitLsn)) {
...
+     * The command id is only compared when both positions have one. In function mode neither side ever
+     * does, so the comparison falls through to the sequence value and behaves exactly as it did before
+     * command ids were recorded. It also falls through when only one side knows its command id, which
+     * happens while resuming from an offset written before they were recorded; the sequence value is the
+     * only thing the two positions have in common there.
+     * <p>
+     * {@code operation} is deliberately not part of the comparison: the before and after images of a
+     * normal update share a commit LSN, a command id and a sequence value, and are expected to compare
+     * equal here, as they always have.
+     */
```
Naros: Hi @b-goyal, this is a great example of where JavaDoc/comment fatigue is real. 

It's clear from the code, as it's self-documenting, how the comparison ordering is done. The specific points you make about the special use cases can honestly be represented with 1 or 2 line comments in the method itself, not 8+ lines of Javadoc.

We're not opposed to having AI help you with code, but it's notorious for excessive Javadoc and even inline documentation, so always try to evaluate whether the comment length is appropriate for the change in question.

b-goyal: Feedback noted @Naros. I am still testing these changes and will refine the PR further before sharing for review. I will make sure to address this feedback in the final polished version.

b-goyal:
@Naros the PR is now ready for review. Please take a look when you have a chance.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
```diff
@@ -85,17 +85,18 @@ public class SqlServerConnection extends JdbcConnection {
...
     private static final String GET_ALL_CHANGES_FOR_TABLE_SELECT = "SELECT [__$start_lsn], [__$seqval], [__$operation], [__$update_mask], #, "
             + LSN_TIMESTAMP_SELECT_STATEMENT;
-    private static final String GET_ALL_CHANGES_FOR_TABLE_SELECT_DIRECT = "SELECT cdc_data.[__$start_lsn], cdc_data.[__$seqval], cdc_data.[__$operation], cdc_data.[__$update_mask], #, "
+    private static final String GET_ALL_CHANGES_FOR_TABLE_SELECT_DIRECT = "SELECT cdc_data.[__$start_lsn], cdc_data.[__$seqval], cdc_data.[__$operation], cdc_data.[__$update_mask], cdc_data.[__$command_id], #, "
             + LSN_TIMESTAMP_SELECT_STATEMENT_JOIN;
     private static final String GET_ALL_CHANGES_FOR_TABLE_FROM_FUNCTION = "FROM #db.cdc.#function(?, ?, N'all update old')";
-    private static final String GET_ALL_CHANGES_FOR_TABLE_FROM_DIRECT = "FROM #db.cdc.#table AS cdc_data LEFT JOIN #db.cdc.lsn_time_mapping ltm ON ltm.start_lsn = cdc_data.[__$start_lsn]";
+    private static final String GET_ALL_CHANGES_FOR_TABLE_FROM_DIRECT = "FROM #db.cdc.#table AS cdc_data WITH (NOLOCK) LEFT JOIN #db.cdc.lsn_time_mapping ltm ON ltm.start_lsn = cdc_data.[__$start_lsn]";
```
Naros, line 91: Can you explain why introducing `WITH (NOLOCK)` is important here?

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SqlServerConnectorIT.java
```diff
@@ -635,7 +635,9 @@ void updatePrimaryKey() throws Exception {
         assertRecord(insertValueB.getStruct("after"), expectedInsertRowB);
         assertRecord(insertkeyB, expectedInsertKeyB);
         assertNull(insertValueB.get("before"));
-        assertThat(insertValueB.getStruct("source").getInt64("event_serial_no")).isEqualTo(2L);
+        // The delete and the insert of this primary key update share a commit LSN and a sequence value but
+        // carry different command ids hence event_serial_no stays 1L.
+        assertThat(insertValueB.getStruct("source").getInt64("event_serial_no")).isEqualTo(1L);
```
Naros: I believe this shifts the documented behavior in `sqlserver.adoc`, in Table 10: Descriptions of update event value fields. It might be good to review that and update accordingly.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
```diff
@@ -195,8 +196,8 @@ public SqlServerConnection(SqlServerConnectorConfig config, SqlServerValueConver
         this.optionRecompile = optionRecompile;
     }
 
-    private String buildGetAllChangesForTableQuery(SqlServerConnectorConfig.DataQueryMode dataQueryMode,
-                                                   Set<Envelope.Operation> skippedOperations) {
+    static String buildGetAllChangesForTableQuery(SqlServerConnectorConfig.DataQueryMode dataQueryMode,
```
Naros: Curious if there's a reason this needed to be made static and visibility widened? Seems like none of the changes themselves require this change.

Naros (commented):
Hi @b-goyal I've left a few inline questions.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
b-goyal: Added it as function mode uses it. I had a discussion with MS support on the need of it in the function mode and based on their response it felt like an addition worth it. Pasting their response here - 
_
Why NOLOCK is used inside CDC functions:
- NOLOCK is used by design to ensure that CDC reads do not block the CDC capture (log reader) process or impact production workloads.
- If NOLOCK were not used, CDC reads would take shared locks, and inserts from the capture job would require exclusive locks which could lead to blocking, slow down CDC processing, and in extreme cases delay transaction log truncation.
- Since CDC change tables contain only committed, insert-only data, using NOLOCK is safe and ensures smooth, non-blocking change capture.

Even though NOLOCK can normally cause dirty reads, this does not happen with CDC. This is because CDC tables only store committed data, and once a row is written to a CDC table, it is insert-only and never updated. Since there is no uncommitted data in CDC tables, there is no risk of reading data that can later be rolled back.
_

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
b-goyal: Seems like a miss. I added some UTs to verify a few changes to this method. Removed the UTs but missed reverting the visibility changes. Reverted now.

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SqlServerConnectorIT.java
b-goyal: Attempted to update, PTAL. Wanted to clarify the special case about PK updates. If we perform PK updates in bulk, the seqvals are different for the delete and insert events. Hence, the events do not make event_serial_no=2 even in the absence of command_id. This example is from some old testing -
```
CREATE TABLE dbo.test3 (id INT NOT NULL PRIMARY KEY, name VARCHAR(50));
  EXEC sys.sp_cdc_enable_table @source_schema='dbo', @source_name='test3',
       @role_name=NULL, @supports_net_changes=0;
  INSERT INTO dbo.test3(id,name) VALUES (1,'alice'),(2,'bob'),(3,'carol'),(4,'dave'),(5,'eve');
  -- update the PRIMARY KEY value in one transaction
  UPDATE dbo.test3 SET id = id + 1000 WHERE id BETWEEN 1 AND 5;


commit_lsn            |seqval                |__$command_id|__$operation|update_mask|name |id  |
----------------------+----------------------+-------------+------------+-----------+-----+----+
0x00000a2d000150b80010|0x00000a2d000150b80003|            1|           1|0x03       |alice|   1|
0x00000a2d000150b80010|0x00000a2d000150b80006|            2|           1|0x03       |bob  |   2|
0x00000a2d000150b80010|0x00000a2d000150b80007|            3|           1|0x03       |carol|   3|
0x00000a2d000150b80010|0x00000a2d000150b80008|            4|           1|0x03       |dave |   4|
0x00000a2d000150b80010|0x00000a2d000150b80009|            5|           1|0x03       |eve  |   5|
0x00000a2d000150b80010|0x00000a2d000150b8000a|            6|           2|0x03       |alice|1001|
0x00000a2d000150b80010|0x00000a2d000150b8000b|            7|           2|0x03       |bob  |1002|
0x00000a2d000150b80010|0x00000a2d000150b8000c|            8|           2|0x03       |carol|1003|
0x00000a2d000150b80010|0x00000a2d000150b8000d|            9|           2|0x03       |dave |1004|
0x00000a2d000150b80010|0x00000a2d000150b8000e|           10|           2|0x03       |eve  |1005|
```

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
vjuranek: Not directly related to this PR, but given this comment, it would be probably good to add `WITH (NOLOCK)` to other queries where we query CDC tables (like getting max LSN etc). Could you add it also there @b-goyal (in separate commit or dedicated pull request)? TIA!

vjuranek: Possibly in another PR as I created [new issue for it](https://github.com/debezium/dbz/issues/2334) so we have full explanation somewhere easily reachable.

b-goyal:
@Naros can we re-run the tests. The failing SQL Server test is not related to the changes and is flaky. Ran it a few times locally- passed 11/13 times.

vjuranek (approved):
LGTM

b-goyal:
@Naros @vjuranek - any additional feedback or is it good to merge?

vjuranek:
All good from my side, waiting for @Naros if he has any more input.

Naros:
Hi @b-goyal could you give this one quick rebase against `main`?

b-goyal:
@Naros updated.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerChangeTablePointer.java
```diff
@@ -97,13 +106,30 @@ protected boolean isNewTransaction() throws SQLException {
...
+        // next page in a running connector
+        if (lastPositionSeen != null && !lastPositionSeen.equals(TxLogPosition.NULL)) {
+            return connection.getChangesForTable(getChangeTable(), lastPositionSeen.getCommitLsn(), lastPositionSeen.getInTxLsn(),
+                    lastPositionSeen.getOperation(), lastPositionSeen.getCommandId(), toLsn, maxRowsPerResultSet);
         }
+
+        // restarted connector which has not seen any position yet.
+        if (resumeFromPosition != null && resumeFromPosition.getCommandId() != null) {
+            return connection.getChangesForTable(getChangeTable(), fromLsn, resumeFromPosition.getInTxLsn(), 0,
+                    resumeFromPosition.getCommandId(), toLsn, maxRowsPerResultSet);
```
Naros: Thanks for the thorough work on this fix, the keyset design and the test coverage are excellent. I do think this branch has a subtle reachability problem, though.

This branch is not restart-only. `SqlServerStreamingChangeEventSource` passes `offsetContext.getChangePosition()` as `resumeFromPosition` on **every** iteration, and every new pointer's first `getNextResultSet` call is made with `lastPositionSeen == null` (see `ChangeTableResultSet#next`, `currentChangePosition` starts null). 

So once any direct-mode event has been emitted, the first query of **every subsequent window** takes this branch and binds the keyset with this window's `fromLsn` paired with the *previous* position's `commandId`/`seqval`.

In steady state, `fromLsn = incrementLsn(previous window's toLsn)`, which belongs to a different transaction than the one `resumeFromPosition` points at. If a real transaction's `__$start_lsn` ever equals that computed `fromLsn`, its rows with `command_id <= resumeFromPosition.getCommandId()` match none of the OR branches and are silently dropped, yielding a partially emitted transaction, i.e., the same class of data loss this PR fixes.

The window for this is narrow (an exact LSN collision), but the guard is cheap: only apply the resume keyset when the resume position actually belongs to the transaction at the window's lower bound, e.g.

```java
if (resumeFromPosition != null && resumeFromPosition.getCommandId() != null
        && fromLsn.equals(resumeFromPosition.getCommitLsn())) {
```

and fall through to the plain `fromLsn`/`toLsn` query otherwise; the existing in-memory skip on `lastProcessedPositionOnStart` already handles any overlap there, exactly as it does today in function mode.

WDYT?

Naros (commented):
Aside from my finding below, I am still concerned with downstream consumer ordering guarantees.

The new design emits both the DELETE followed by the INSERT with exactly the same `event_serial_no`, `commit_lsn`, and `change_lsn`, and a downstream consumer can no longer accurately identify ordering without now introducing yet another value. This effectively makes `event_serial_no`, which exists to disambiguate this use case useless.

One option is to introduce `command_id` into the source information block and document this behavior change, but I'm not sure that's necessarily the right path forward.

@jpechane @vjuranek thoughts?

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerChangeTablePointer.java
b-goyal: Yeah, makes sense. I misunderstood and thought an active connector will always go to branch-1. I will update this branch with the suggestion.

b-goyal: Updated PR

b-goyal:
Re-ran the failing IT multiple times locally - no failures observed. Failure looks like pre-existing flake and not related to PR changes.

documentation/modules/ROOT/pages/connectors/sqlserver.adoc
```diff
@@ -1378,12 +1378,13 @@ The source metadata includes:
 * If the event was part of a snapshot
 * Name of the table that contains the new row
 * Server log offsets
-+
-The `event_serial_no` field differentiates events that have the same commit and change LSN.
-Typical situations for when this field has a value other than `1`:
-+
+
```
Naros: I believe we need to keep the `+` line continuations, otherwise the text here and below no longer left-aligns with the prior text under the `source` list item, does it not? The same would apply at line 1383, too.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerStreamingChangeEventSource.java
```diff
@@ -148,6 +148,21 @@ public void execute(ChangeEventSourceContext context, SqlServerPartition partiti
...
 
+    private TxLogPosition resumePosition(TxLogPosition position) {
+        final boolean directMode = connectorConfig.getDataQueryMode() == SqlServerConnectorConfig.DataQueryMode.DIRECT;
+        final boolean commandIdExists = position.getCommandId() != null;
+
+        // direct and function mode orders change events differently hence a mode change contains a risk of data loss.
+        // To prevent data loss, we re-read the last transaction in full if the mode changes.
+        if ((directMode && commandIdExists) || (!directMode && !commandIdExists)) {
+            return position;
+        }
```
Naros, line 162: In `directMode`, the commandIdExists is absent in two conditions:
* The first streaming pass, after the initial snapshot.
* After any idle iteration pass.

This causes the connector to report the INFO message below, indicating a change in mode, and re-stream the last transaction. At the event stream level, I don't see any changes, but this log entry may confuse users when the query mode truly hasn't changed.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/TxLogPosition.java
```diff
@@ -41,16 +46,21 @@ public int getOperation() {
...
     }
 
     @Override
     public int hashCode() {
         final int prime = 31;
         int result = 1;
         result = prime * result + ((commitLsn == null) ? 0 : commitLsn.hashCode());
+        result = prime * result + ((commandId == null) ? 0 : commandId.hashCode());
         result = prime * result + ((inTxLsn == null) ? 0 : inTxLsn.hashCode());
         result = prime * result + operation;
```
Naros: Now this predates this PR, but one observation since the code changed is that `hashCode` has the operation but `equals` does not. I do wonder while we're in here, should we clean that up, or at the very least raise a follow-up.

debezium-connector-sqlserver/src/test/java/io/debezium/connector/sqlserver/SqlServerConnectorIT.java
Naros: So, if I understand correctly, one side effect is that `event_serial_no` now differs depending on the query mode used. For example, any two changes that share `{commitLsn, seqVal}` and have different command IDs are one position apart in function mode, but two in direct mode.

Given that we do support the ability to change the query mode at any time, I believe this potentially creates an unusual situation for consumers, where we probably should re-evaluate how this can be done consistently for both modes, right?

b-goyal:
@Naros 
About the difference in behaviour, I tried to compile the difference across both modes in terms of `event_serial_no` for the cases I am familiar. Attaching a screenshot of the result. There are two cases where there would be difference in behaviour-
1. Bulk deferred update : This was the case where data loss was experienced and why we decided to work on reordering. I dont see a way to make it consistent with function mode, fix OOM and still handle data loss. Please note the PR only fixes the data loss for this case. The event_serial_no difference exists without the PR changes too today.
2. Single PK update: This got introduced due to this PR change. We can still handle it in the code in `setChangePosition` of `SqlServerOffsetContext` class. 

Are there other cases where you are concerned about the `event_serial_no` difference?

Screenshot 2026-08-31 at 1 27 59 PM

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerStreamingChangeEventSource.java
b-goyal: I have reworked on the PR. Now the command_id is set to `-1` when direct mode is used and `null` for function mode. The `command_id` is now set with snapshot offsets and idle intervals too. Please see if you think any case is still missed.

b-goyal:
@Naros , I have reworked on the PR and modified to use `-1` as the base state for `direct` mode. Also changed the event_serial_no piece to keep PK update behaviour unchanged across both modes. Please re-review when you have a chance and let me know your thoughts.

Naros (approved):
Thanks @b-goyal, LGTM.

Naros:
Thanks @b-goyal, applied! :tada:

PR #7736, opened 2026-07-30
debezium/dbz#2275 Close connection on non-transient connection errors…

iPave:
When the database closes the connection on its end, the JDBC driver throws a SQLNonTransientException during an incremental snapshot. This exception indicates that retry of the same operation would fail unless the cause of the SQLException is corrected. In such cases, we should close the connection and clear its prepared statement cache so that the connection is reinitialized on the next retry.

Fixes debezium/dbz#2275

## PR Checklist

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
+++ b/debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
@@ -16,6 +16,7 @@ import java.sql.PreparedStatement;
 import java.sql.ResultSet;
 import java.sql.ResultSetMetaData;
 import java.sql.SQLException;
+import java.sql.SQLNonTransientConnectionException;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.LinkedHashMap;
@@ -288,6 +289,9 @@ public abstract class AbstractIncrementalSnapshotChangeEventSource<P extends Par
                         context.maximumKey(maximumKey);
                     }
                     catch (SQLException e) {
+                        if (e instanceof SQLNonTransientConnectionException) {
+                            closeJdbcConnection();
+                        }
                         LOGGER.error("Failed to read maximum key for table {}", currentTableId, e);
                         notificationService.incrementalSnapshotNotificationService().notifyTableScanCompleted(context, partition, offsetContext, totalRowsScanned,
                                 SQL_EXCEPTION);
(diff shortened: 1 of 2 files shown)


vjuranek (approved):
LGTM, please sign your commit @iPave

iPave:
@vjuranek I've signed, sorry for the previous one.

iPave:
@vjuranek  I've changed code a bit, since we still need to warnAndSkip because it was unsucess readChunk, and I've missed this part. Correct me if I'm wrong.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -346,6 +347,15 @@ protected void readChunk(P partition, OffsetContext offsetContext) throws Interr
             LOGGER.trace("Window close emitted");
         }
         catch (SQLException e) {
+            if (e instanceof SQLNonTransientConnectionException) {
+                try {
+                    LOGGER.info("Connection is being closed after Not recoverable SQL exception");
+                    jdbcConnection.close();
```
Naros: Do the inner catch blocks for the max-key query and `createDataEventsForTable` also suffer from the same problem? I don't think they check for this use case. They simply skip to the next table and continue the loop, allowing each remaining table to fail with a closed connection.

Should we close this gap?

Naros (commented):
LGTM, I just have one question about a gap I potentially see.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
vjuranek: Good catch

iPave: @Naros @vjuranek so should I also check for the same exception for max-key and createDataEventsForTable?

vjuranek: yes please @iPave

iPave: @vjuranek done

Naros (approved):

Naros:
Applied, thanks @iPave :tada:

PR #7738, opened 2026-07-30
DBZ-2325 Preserve pending incremental snapshot across boot-time snapshot

MrIvv:
Fixes https://github.com/debezium/dbz/issues/2325

When a boot-time snapshot runs on a connector restart (`snapshot.mode=always` or `configuration_based` with data/schema enabled), `determineSnapshotOffset` builds a fresh offset context whose incremental snapshot context is empty, discarding the pending incremental snapshot restored from the offsets. The skipped-snapshot path is unaffected because it hands `previousOffset` to streaming unchanged.

This change extends the existing carry-over (the two LSNs gated by `shouldStreamEventsStartingFromSnapshot`) to the incremental snapshot context: `initialContext` gains an overload accepting the restored context, and `determineSnapshotOffset` passes it in both branches. First-boot behavior is unchanged (`previousOffset == null` keeps creating a fresh context as today); no SPI or offset format changes.

Field-tested on a production-like fleet: connector killed (SIGKILL) at 30% of an incremental snapshot with persistent offsets, restarted with `configuration_based`; the pending table resumed from the exact stored chunk position and completed. Without the fix the queue is silently lost and the snapshot never resumes.

The MySQL/MariaDB family has the same structural pattern (`MySqlOffsetContext.initial`); happy to follow up with an equivalent change there if this approach is accepted.

## PR Checklist

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresOffsetContext.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresOffsetContext.java
@@ -287,6 +287,11 @@ public class PostgresOffsetContext extends CommonOffsetContext<SourceInfo> {
 
     public static PostgresOffsetContext initialContext(PostgresConnectorConfig connectorConfig, PostgresConnection jdbcConnection, Clock clock, Lsn lastCommitLsn,
                                                        Lsn lastCompletelyProcessedLsn) {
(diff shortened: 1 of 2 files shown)


Naros (changes requested):
I saw your scope in the issue, but for behavior changes like this, ideally this should be applied across all connectors to keep the documented behavior uniform and consistent. 

Please add all other connectors, and include integration tests.

PR #7740, opened 2026-07-30
debezium/dbz#2330 Pin the exported snapshot on pooled connections for parallel PostgreSQL snapshots

TimurRakhmatullin86:
Fixes debezium/dbz#2330

### What

When `snapshot.max.threads > 1`, the PostgreSQL connector's initial snapshot uses a pool of extra connections. The exported snapshot from the replication slot (`SET TRANSACTION SNAPSHOT '<name>'`) is applied only on the main connection in `connectionCreated`; the pooled connections created in `createConnectionPool` copy only the isolation level (`initializePooledConnection`), not the exported snapshot. So each pooled connection reads under its own MVCC snapshot, and a parallel snapshot is **not consistent across connections** — a row committed concurrently after the slot's snapshot was taken can appear in the output of a table read by a pooled connection while being absent from tables read by the main one.

### Fix

Override `connectionPoolConnectionCreated` in `PostgresSnapshotChangeEventSource` to import the exported snapshot on each pooled connection, guarded by exactly the conditions under which the main connection sets it (`slotCreatedInfo != null && !onDemand && shouldStreamEventsStartingFromSnapshot() && startingSlotInfo == null`). It `rollback()`s first so `SET TRANSACTION SNAPSHOT` runs as the first statement of a fresh transaction, and reuses the existing `snapshotTransactionIsolationLevelStatement(...)`. On-demand/blocking snapshots, pre-existing slots, and catch-up streaming are skipped, matching the main connection.

Raised during the DDD-59 review (debezium/debezium-design-documents#60) — @Naros suggested fixing this independently of the DDD and backporting to 3.6.

### Testing

I have not added a test yet because there isn't an existing one to re-enable (the `@Disabled` tests in `RecordsSnapshotParallelProducerIT` assert per-record snapshot ordering, which is nondeterministic under parallel threads for unrelated reasons and wouldn't be fixed by this). I'd like to add a new IT (e.g. `PostgresParallelSnapshotConsistencyIT`) that runs an initial-snapshot-with-streaming with `snapshot.max.threads > 1` and more tables than threads, commits a change from a separate session between slot establishment and the pooled reads, and asserts no table's snapshot reflects that concurrent change. Before writing it — does that approach look right to you, or would you prefer a lighter `LogInterceptor` test asserting the `SET TRANSACTION SNAPSHOT` is issued on the pooled connections?

Backport candidate for 3.6.

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSnapshotChangeEventSource.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSnapshotChangeEventSource.java
@@ -21,6 +21,7 @@ import io.debezium.connector.postgresql.connection.Lsn;
 import io.debezium.connector.postgresql.connection.PostgresConnection;
 import io.debezium.connector.postgresql.spi.SlotCreationResult;
 import io.debezium.connector.postgresql.spi.SlotState;
+import io.debezium.jdbc.JdbcConnection;
 import io.debezium.jdbc.MainConnectionProvidingConnectionFactory;
 import io.debezium.pipeline.EventDispatcher;
 import io.debezium.pipeline.notification.NotificationService;
@@ -106,6 +107,26 @@ public class PostgresSnapshotChangeEventSource extends RelationalSnapshotChangeE
         if (snapshotterService.getSnapshotter().shouldStreamEventsStartingFromSnapshot() && startingSlotInfo == null) {
             setSnapshotTransactionIsolationLevel(snapshotContext.onDemand);
         }
+    }
+
+    @Override
+    protected void connectionPoolConnectionCreated(RelationalSnapshotContext<PostgresPartition, PostgresOffsetContext> snapshotContext,
+                                                   JdbcConnection connection)
+            throws SQLException {
+        // The exported-snapshot pin that connectionCreated() sets on the main connection is not shared by the
+        // extra connections created when snapshot.max.threads > 1. Without importing the same exported snapshot
+        // here, each of those connections reads under its own MVCC snapshot, so a parallel snapshot is not
+        // consistent across connections. Import the exported snapshot on every pooled connection, under the same
+        // conditions the main connection uses to set it.
+        if (slotCreatedInfo == null
+                || snapshotContext.onDemand
+                || !snapshotterService.getSnapshotter().shouldStreamEventsStartingFromSnapshot()
+                || startingSlotInfo != null) {
+            return;
+        }
+        // SET TRANSACTION SNAPSHOT must be the first statement of a fresh transaction.
+        connection.rollback();
+        connection.executeWithoutCommitting(snapshotTransactionIsolationLevelStatement(slotCreatedInfo, snapshotContext.onDemand));
     }
 
     @Override
--- a/debezium-connector-common/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
+++ b/debezium-connector-common/src/main/java/io/debezium/relational/RelationalSnapshotChangeEventSource.java
@@ -838,7 +838,7 @@ public abstract class RelationalSnapshotChangeEventSource<P extends Partition, O
                                                               SnapshotReceiver<P> snapshotReceiver, Table table, boolean firstTable, boolean lastTable, int tableOrder,
                                                               int tableCount, String selectStatement, OptionalLong rowCount, Set<TableId> rowCountTablesKeySet,
                                                               Queue<JdbcConnection> connectionPool, Queue<O> offsets) {
-        return createPooledResourceCallable(connectionPool, offsets,
+        return createPooledResourceCallable(snapshotContext, connectionPool, offsets,
                 (connection, offset) -> {
                     LoggingContext.PreviousContext previousLoggingContext = LoggingContext.forConnector(
                             connectorConfig.getContextName(), connectorConfig.getLogicalName(), null, "snapshot", snapshotContext.partition);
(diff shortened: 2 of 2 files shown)


Naros:
Hi @TimurRakhmatullin86, the test with the concurrent commit with concurrent reads directly validates the fix. While the `LogInterceptor` is lighter, it only tests the mechanism but not the invariant result. So the former makes more sense, if it's easy enough to create a reliable integration test.

TimurRakhmatullin86:
Added `PostgresParallelSnapshotConsistencyIT` — the concurrent-commit route you preferred, made deterministic rather than timing-based.

A helper transaction holds `ACCESS EXCLUSIVE` on all captured tables and stages the phantom rows, then commits exactly when the snapshot reaches step 7 (`Snapshot step 7 - Snapshotting data`). With the default `snapshot.locking.mode=none` the snapshot takes no locks itself, so every data SELECT blocks on those locks and can't complete until the commit — the commit therefore lands strictly after the exported snapshot was taken and the pool was built. Without the fix, an unpinned pooled connection's `REPEATABLE READ` snapshot is only taken when its SELECT is unblocked (post-commit), so the phantom leaks into the snapshot; with the fix every pooled connection is pinned to the exported snapshot, so the row is invisible to the snapshot and arrives exactly once via streaming.

`snapshot.max.threads=3` over 4 tables (legacy table-per-thread) guarantees non-main connections are exercised. The step-7 log wait keys the window off an enforced ordering rather than a sleep, so it shouldn't be flaky in CI. Happy to tighten anything.

TimurRakhmatullin86:
Update on the test: my first attempt hit a slot-creation deadlock I should have foreseen. Staging the phantom rows behind an `ACCESS EXCLUSIVE` lock in an open transaction *before* the connector starts blocks `CREATE_REPLICATION_SLOT ... EXPORT_SNAPSHOT`, which waits for concurrent transactions to finish before it can export a consistent snapshot — so the connector never gets past slot creation (`assertConnectorIsRunning` times out).

I've dropped that test commit for now — the fix itself is unchanged — and I'm reworking the window to take the lock only *after* the slot's exported snapshot exists (i.e. after the connector is running), so there's no interference with slot creation. I'll re-push once it's reliably green in CI.

If you have a preferred harness pattern for "commit a row between slot creation and the parallel data reads", I'm glad to follow it rather than reinvent one.

Naros:
Hi @TimurRakhmatullin86, if you're unable to find a solid way to verify it in the tests, but you can reproduce it with manual steps, that might be sufficient if those are documented in the associated issue in case we need to revisit it.

TimurRakhmatullin86:
Thanks @Naros. That works for me — I couldn't get a reliably non-flaky IT together (my first attempt deadlocked slot creation by holding the lock before `EXPORT_SNAPSHOT`, and the reworked lock-after-step-1 version I couldn't validate locally because of an unrelated `support/checkstyle` compile issue in my environment). I've documented the root cause and a deterministic manual reproduction procedure in debezium/dbz#2330 for future reference. The fix itself stands on the code path (pooled connections never receive the exported-snapshot pin).

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSnapshotChangeEventSource.java
```diff
@@ -108,6 +109,26 @@ protected void connectionCreated(RelationalSnapshotContext<PostgresPartition, Po
         }
     }
 
+    @Override
+    protected void connectionPoolConnectionCreated(RelationalSnapshotContext<PostgresPartition, PostgresOffsetContext> snapshotContext,
```
Naros, line 113: What about `initializePooledConnection`, should that get similar treatment?

Naros (commented):
I've left one comment below.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresSnapshotChangeEventSource.java
TimurRakhmatullin86: Good catch — that's the more important half, actually. `initializePooledConnection` is called in two places: at pool creation (`createConnectionPool`), immediately followed by `connectionPoolConnectionCreated`, and again on the reconnect path in `createPooledResourceCallable`, where only `initializePooledConnection` ran. So a pooled connection replaced mid-snapshot got the isolation level copied but not the exported-snapshot re-pin — which is exactly what the existing warning there ("Snapshot consistency for subsequent tables may be affected") was conceding.

I've threaded the snapshot context through `createPooledResourceCallable` and call `connectionPoolConnectionCreated(snapshotContext, connection)` right after the reconnect, so the reconnect path applies the same pin as pool creation. Connectors that don't override the hook are unaffected; Oracle re-points at its PDB and PostgreSQL re-imports the exported snapshot. Pushed as a follow-up commit on this PR (base-class change, compiles against connector-common).

One thing I deliberately left out, so it's a conscious choice rather than an oversight: the reconnect path still doesn't re-run `getSnapshotConnectionFirstSelect` (the `LIMIT 1` first-read some connectors use to pin their own view). For PostgreSQL that's the empty default, so the hook is the whole story. For a first-select connector like MySQL, re-running it on reconnect would only re-pin a *new* REPEATABLE READ view, which restores consistency solely while the global read lock is still held — so it's connector-dependent whether that's meaningful. Happy to add it for full symmetry if you'd prefer, but it felt like it belonged in the retry/consistency-mode discussion rather than this fix. Same for tightening the warning text now that a re-pin is attempted — I can soften it to only fire for connectors that can't restore a shared view.

Naros: Hi @TimurRakhmatullin86, I definitely think that given this issue was targeted at PostgreSQL, leaving the MySQL repeatable read on the retry/consistency discussion makes sense to me.

Naros (approved):

Naros:
Thanks, applied @TimurRakhmatullin86 :tada:

PR #7742, opened 2026-07-31
debezium/dbz#2333 Unwrap SpecialValueDecimal when binding query column values

MrIvv:
Closes https://github.com/debezium/dbz/issues/2333

An incremental snapshot of a table whose `NUMERIC` key holds one of PostgreSQL's special values (`NaN`, `Infinity`, `-Infinity`) silently loses data. `keyFromRow()` unwraps `ValueWrapper` boundary values via `getWrappedValue()`, which returns `null` for the special values since they have no `BigDecimal` form; `bindBoundaryParams()` then silently skips the null, the chunk query matches nothing, and the snapshot completes early without errors. Since PostgreSQL orders `NaN` above every other numeric, a single `NaN` row immediately becomes the snapshot's `maximumKey` — which is bound into every chunk query — so one such row is enough to lose everything past the first chunk.

This PR:
- handles the special values where the chunk boundaries are extracted (`keyFromRow()`): ordinary values keep their exact `BigDecimal` form, `NaN`/`±Infinity` fall back to their double form (`SpecialValueDecimal.toDouble()`), which drivers bind as `float8` and PostgreSQL compares against `numeric` with the same ordering semantics;
- reverts the previous bind-time unwrap in `JdbcConnection.setQueryColumnValue()`: the ordinary-value case it addressed is already covered on main by `keyFromRow()`'s `ValueWrapper` unwrap (the symptom I originally reported came from custom chunk-reading code in #7362, and will be fixed there);
- adds `IncrementalSnapshotIT.insertsNumericPkWithSpecialValues`, sized so that with the test chunk size of 10 the `Infinity` row lands exactly on a chunk boundary and `NaN` is the maximum key bound into every chunk query.

## PR Checklist

### DIFF
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/IncrementalSnapshotIT.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/IncrementalSnapshotIT.java
@@ -431,7 +431,7 @@ public class IncrementalSnapshotIT extends AbstractIncrementalSnapshotTest<Postg
         final Map<Integer, Integer> dbChanges = consumeMixedWithIncrementalSnapshot(
                 expectedRecordCount,
                 x -> true,
-                k -> specialAwareKey(Double.parseDouble(k.getString("pk"))),
+                k -> specialAwareKey(k.getString("pk")),
                 record -> ((Struct) record.value()).getStruct("after").getInt32(valueFieldName()),
                 "test_server.s1.anumeric",
                 null);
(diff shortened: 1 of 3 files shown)


Naros (changes requested):
Hi @MrIvv, this change should be accompanied with a test case. There doesn't need to be one for each connector, but at least one test in one of the connector suites to guard against this behavior breaking in the future.

There test should also cover expectations when `SpecialValueDecimal` has `NaN` or `Infinity` values encoded, and how the query builder should handle this.

MrIvv:
Hi @Naros, working on the test you asked for turned out to be more useful than expected. 
It showed that the original problem I described is actually already covered on main, and that what I had originally hit came from the custom chunk-reading code in my
multi-threading PR (#7362). The real remaining issue on main is the NaN/Infinity case you mentioned: I could reproduce it on our side, and it silently drops a large part of the table during an incremental snapshot.
So I reworked the PR: the original change is reverted, the fix now handles the special values where the chunk boundaries are extracted, and the new IT covers both NaN and Infinity as real boundaries. 
The issue description is updated to match.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/IncrementalSnapshotIT.java
```diff
@@ -401,6 +401,61 @@ record -> ((Struct) record.value()).getStruct("after").getInt32(valueFieldName()
...
+        sendAdHocSnapshotSignal("s1.anumeric");
+
+        // 21 rows with chunk size 10: the ascending key order is -Infinity, 1..18, Infinity, NaN, so the
+        // second chunk ends exactly on Infinity (bound as the next chunk's lower bound) and NaN is the
+        // maximum key, bound into every chunk query.
+        final int expectedRecordCount = 21;
+        final Map<Integer, Integer> dbChanges = consumeMixedWithIncrementalSnapshot(
+                expectedRecordCount,
+                x -> true,
+                k -> specialAwareKey(Double.parseDouble(k.getString("pk"))),
```
Naros: I believe the test failures are related and its because when using string-based decimal handling mode, the values are `NEGATIVE_INFINITY`, `POSITIVE_INFINITY`, and `NAN` and `Double.parseDouble` does not understand those, resulting in `NumberFormatException`.

I believe `specialAwareKey` needs to handle these specific cases, too.

Naros (changes requested):
This looks close @MrIvv, just one issue with the test.

Naros:

Naros (approved):

Naros:
Applied, thanks @MrIvv :tada:

PR #7744, opened 2026-08-01
debezium/dbz#806 Support the Lettuce client in debezium-storage-redis

### LINKED ISSUE #806: [debezium-storage-redis] Support lettuce [DBZ-6390]
Migrated from [DBZ-6390](https://issues.redhat.com/browse/DBZ-6390)

debezium-storage-redis is currently implemented using jedis.

Could you add support for lettuce?
https://github.com/lettuce-io/lettuce-core

 

I am using Spring Boot + debezium embedded engine.
For Spring Boot Application, lettuce is often used in the redis client.
(This is partly due to the fact that both spring and lettuce are developed by pivotal...haha)

Using both jedis and lettuce is a bit wasteful from a dependency point of view, so it would be great if it is supported.

doljae:
Fixes debezium/dbz#806

## Description

The Redis storage module is currently Jedis-only. This adds a config-driven choice of driver through a new `redis.client.library` property (`jedis` | `lettuce`), so that users who already run Lettuce — Spring Boot applications, typically — do not have to bring Jedis along as well.

Both the offset store and the schema history obtain their client from the single `RedisConnection.getRedisClient(...)` factory, so the selection is made in one place and reaches the Debezium Server Redis sink without further changes (`RedisStreamChangeConsumerConfig` extends `RedisCommonConfig` and the consumer goes through the same factory).

**Default behaviour is unchanged.** Without the new property, everything resolves to Jedis exactly as before.

### Lettuce is a `provided` dependency

`io.lettuce:lettuce-core` is declared with `<scope>provided</scope>`. At compile scope it pulls in nine Netty artifacts plus `reactor-core`, which every Jedis user would then carry — the opposite of what this feature is for. Measured on the distribution archive, that chain is about 6.9 MB of the 11 MB the archive would otherwise hold.

`provided` rather than `optional` so that the chain is kept out of the distribution archive as well, not only out of downstream Maven consumers' transitive dependencies. Users who set `redis.client.library=lettuce` add `lettuce-core` (with its Netty and Reactor dependencies) to the runtime classpath themselves.

To make that safe, all Lettuce types are confined to `LettuceClient`, including the connection setup. `RedisConnection` probes for the driver first and reports a configuration error naming the artifact, rather than failing later with `NoClassDefFoundError`. `LettuceOptionalDependencyTest` asserts the isolation by scanning the compiled constant pool of the default-path classes, so a stray import cannot creep back in unnoticed.

I would appreciate a maintainer opinion on this trade-off — see the inline comment on `pom.xml`. If you would rather have the driver bundled, it is a one-line change and nothing else in the PR depends on it.

### Known limitations of the Lettuce driver

| | Jedis | Lettuce |
|---|---|---|
| Single instance | yes | yes |
| Cluster mode | yes | **no** — fails fast with a message pointing at `redis.client.library=jedis` |
| `redis.ssl.truststore.type` / `redis.ssl.keystore.type` | applied | **ignored** — Lettuce derives the store type from the file |

Both are documented alongside the property.

### Implementation notes

**Single connection.** `RedisClient` mixes `String`-based commands with a single `byte[]`-based `hset`, while Lettuce fixes the codec per connection. The obvious shape — a UTF-8 connection plus a second byte-array connection dedicated to `hset` — turns out to break `wait.enabled`: Redis evaluates `WAIT` against the write offset of the *calling* connection, so a `WAIT` issued on a different connection than the preceding `HSET` judges the offset as it stood before that write and reports success without ever covering it. `LettuceClient` therefore holds a single UTF-8 string connection and decodes the `hset` arguments. That adds no new assumption, because `hgetAll` already returns `Map<String, String>` for every `RedisClient` implementation.

**Pipelined `xadd` budget.** `LettuceFutures.awaitAll` spends a single timeout across the whole batch, whereas the Jedis socket timeout applies to each response read, so the batch budget is scaled by the batch size to keep the two drivers comparable. `awaitAll` neither cancels the futures nor throws on timeout, so its return value is checked and turned into a timeout exception instead of blocking on `get()`.

**Failure semantics.** Care was taken to keep them identical to `JedisClient`, since the retry and reconnect loops in `RedisOffsetBackingStore` and `RedisSchemaHistory` are driven by whether a failure is reported as `RedisClientConnectionException`. Server-side command errors (`RedisCommandExecutionException`) are let through unwrapped, as `JedisClient` does with `JedisDataException`. The inline comments call out the specific places.

### Testing

- `RedisOffsetBackingStoreConfigTest` — default, explicit `lettuce`, and rejection of an unknown value.
- `LettuceOptionalDependencyTest` — no Lettuce reference on the default code path.
- `LettuceClientTest` — batch timeout scaling (including overflow and empty batch), cluster mode rejected before any connection attempt, and the missing-driver error path via a class loader that hides `io.lettuce`.
- `LettuceClientIT` — `xadd` / pipelined `xadd` / `xrange` / `xlen` / `hset` / `hgetAll` / `info` / `clientList` round-trip against a Testcontainers Redis, plus the `hset` return value (1 for a new field, 0 for an update) and a non-ASCII `hset` round-trip.
- `LettuceWaitReplicasIT` — master plus a replica that is frozen with `docker pause` so it can no longer send `REPLCONF ACK`; asserts that `hset`, `xadd` and pipelined `xadd` block for the wait timeout, that exactly one connection is open, and that a wait timeout above the socket timeout surfaces as `RedisClientConnectionException`. Parameterised over both `RedisClientLibrary` values, so it doubles as the behavioural parity check between Jedis and Lettuce.
- `RedisOffsetBackingStoreIT` — now parameterised over `(cluster, library)`, so the existing connection / load-retry / save-retry scenarios also run against Lettuce.
- Manually verified with `lettuce-core`, Netty and Reactor stripped from the classpath that the Jedis path is unaffected and that requesting Lettuce yields the actionable error rather than `NoClassDefFoundError`.

## PR Checklist

### DIFF
--- a/debezium-storage/debezium-storage-redis/pom.xml
+++ b/debezium-storage/debezium-storage-redis/pom.xml
@@ -62,6 +62,18 @@
             <groupId>redis.clients</groupId>
             <artifactId>jedis</artifactId>
         </dependency>
+        <!--
+            Provided rather than compile so that the default (Jedis) deployments neither pull in Lettuce
+            transitively nor carry it, Netty and Reactor in the distribution archive: that chain is about
+            6.9 MB of the 11 MB the archive would otherwise hold, and Jedis users never load a class of it.
+            Users who set redis.client.library=lettuce add lettuce-core to the runtime classpath themselves;
+            RedisConnection fails with an actionable message when it is missing.
+        -->
+        <dependency>
+            <groupId>io.lettuce</groupId>
+            <artifactId>lettuce-core</artifactId>
+            <scope>provided</scope>
+        </dependency>
         <dependency>
             <groupId>io.smallrye.reactive</groupId>
             <artifactId>mutiny</artifactId>
--- a/debezium-storage/debezium-storage-redis/pom.xml
+++ b/debezium-storage/debezium-storage-redis/pom.xml
@@ -63,6 +63,10 @@
             <artifactId>jedis</artifactId>
         </dependency>
         <dependency>
+            <groupId>io.lettuce</groupId>
+            <artifactId>lettuce-core</artifactId>
+        </dependency>
+        <dependency>
             <groupId>io.smallrye.reactive</groupId>
             <artifactId>mutiny</artifactId>
         </dependency>
--- a/debezium-storage/debezium-storage-redis/pom.xml
+++ b/debezium-storage/debezium-storage-redis/pom.xml
@@ -62,9 +62,15 @@
             <groupId>redis.clients</groupId>
             <artifactId>jedis</artifactId>
         </dependency>
+        <!--
+            Optional so that the default (Jedis) deployments do not pull in Lettuce and its Netty/Reactor
+            transitive dependencies. Users who set redis.client.library=lettuce must add lettuce-core to
+            the runtime classpath themselves.
+        -->
         <dependency>
             <groupId>io.lettuce</groupId>
             <artifactId>lettuce-core</artifactId>
+            <optional>true</optional>
         </dependency>
         <dependency>
             <groupId>io.smallrye.reactive</groupId>
--- a/debezium-storage/debezium-storage-redis/src/main/java/io/debezium/storage/redis/LettuceClient.java
+++ b/debezium-storage/debezium-storage-redis/src/main/java/io/debezium/storage/redis/LettuceClient.java
@@ -1,0 +1,339 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.storage.redis;
+
+import java.io.File;
+import java.nio.charset.StandardCharsets;
+import java.time.Duration;
+import java.util.AbstractMap.SimpleEntry;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.TimeUnit;
+import java.util.function.Supplier;
+import java.util.stream.Collectors;
+
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.debezium.DebeziumException;
+import io.debezium.util.Strings;
+import io.lettuce.core.ClientOptions;
+import io.lettuce.core.LettuceFutures;
+import io.lettuce.core.Range;
+import io.lettuce.core.RedisCommandExecutionException;
+import io.lettuce.core.RedisCommandTimeoutException;
+import io.lettuce.core.RedisException;
+import io.lettuce.core.RedisFuture;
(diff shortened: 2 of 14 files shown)


debezium-storage/debezium-storage-redis/pom.xml
```diff
@@ -62,6 +62,16 @@
...
         </dependency>
+        <!--
+            Optional so that the default (Jedis) deployments do not pull in Lettuce and its Netty/Reactor
+            transitive dependencies. Users who set redis.client.library=lettuce must add lettuce-core to
+            the runtime classpath themselves.
+        -->
+        <dependency>
+            <groupId>io.lettuce</groupId>
+            <artifactId>lettuce-core</artifactId>
+            <optional>true</optional>
```
doljae: **This is the decision I would most like a second opinion on.**

At compile scope, `lettuce-core` adds ten artifacts to every deployment of this module:

```
io.lettuce:lettuce-core
 +- io.netty:netty-common / handler / resolver / buffer / codec /
    transport / transport-native-unix-common / resolver-dns / codec-dns
 \- io.projectreactor:reactor-core
```

Measured on the module's distribution archive, that chain is about 6.9 MB of the 11 MB the archive would otherwise hold. Since the point of the feature is to let Lettuce users avoid Jedis, having everyone carry both drivers seemed like the wrong default — and pushing Netty into the Kafka Connect plugin path felt like something not to do quietly.

It is `provided` rather than `optional` so that the chain stays out of the distribution archive too, not only out of downstream Maven consumers' transitive dependencies.

The cost is that users who opt into Lettuce must add the artifact themselves; `RedisConnection` fails at startup with a message naming it when it is missing. If you would rather bundle it — either at compile scope here, or `provided` here and bundled in the Server distribution — it is a one-line change and no other part of this PR depends on it. Happy to go whichever way you prefer.

debezium-storage/debezium-storage-redis/src/main/java/io/debezium/storage/redis/LettuceClient.java
```diff
@@ -0,0 +1,306 @@
...
+            return null;
+        });
+    }
+
+    private <R> R tryErrors(Supplier<R> supplier) {
+        try {
+            return supplier.get();
+        }
+        catch (RedisCommandExecutionException e) {
+            // Errors returned by the server (WRONGTYPE, NOAUTH, OOM, ...) are not connection failures and
```
doljae, line 328: One consequence of this split lives in another file, so noting it here: `RedisClientConnectionException` is what drives the reconnect loop in `RedisSchemaHistory`. If server-side command errors were mapped to it as well, a permanent `WRONGTYPE` or `NOAUTH` would burn through `maxRetryCount` and then fail with `"Failed to connect to Redis after N attempts"`, pointing at the network instead of the actual problem.

doljae (commented):
Adding a few notes on the decisions that seemed worth a maintainer's attention, so they do not have to be reverse-engineered from the diff. The first one is a genuine question rather than an explanation.

PR #7746, opened 2026-08-01
debezium/dbz#2335: Add support for TimescaleDB 2.29+ transform

MV-GH:
I first tried solving this by using the public chunk schema that has stronger compatibility guarantees.

```sql
SELECT c.chunk_schema, c.chunk_name, c.hypertable_schema, c.hypertable_name FROM timescaledb_information.chunks c
```

But that didnt workout because that view doesn't contain the compressed chunks.

So I Added a version switch between query methods.

I used docker.io image because there was no quay.io image see https://quay.io/repository/debezium/timescale-timescaledb?tab=tags

Fixes debezium/dbz#2335

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/transforms/timescaledb/QueryInformationSchemaMetadata.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/transforms/timescaledb/QueryInformationSchemaMetadata.java
@@ -37,10 +37,7 @@ public class QueryInformationSchemaMetadata extends AbstractTimescaleDbMetadata 
             "SELECT ht.schema_name, ht.table_name, agg.user_view_schema, agg.user_view_name FROM %s.continuous_agg agg"
                     + " LEFT JOIN %s.hypertable ht ON agg.mat_hypertable_id = ht.id",
             CATALOG_SCHEMA, CATALOG_SCHEMA);
-    private static final String QUERY_CHUNK_TO_HYPERTABLE = String.format(
-            "SELECT c.schema_name, c.table_name, ht.schema_name, ht.table_name FROM %s.chunk c "
-                    + "LEFT JOIN %s.hypertable ht ON c.hypertable_id = ht.id",
-            CATALOG_SCHEMA, CATALOG_SCHEMA);
+    private static final String QUERY_CHUNK_TO_HYPERTABLE = "SELECT c.chunk_schema, c.chunk_name, c.hypertable_schema, c.hypertable_name FROM timescaledb_information.chunks c";
 
     private static final Logger LOGGER = LoggerFactory.getLogger(QueryInformationSchemaMetadata.class);
 


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/transforms/timescaledb/QueryInformationSchemaMetadata.java
```diff
@@ -80,7 +112,7 @@ public void close() throws IOException {
     private void loadTimescaleMetadata() {
         try {
             chunkToHypertable.clear();
-            connection.query(QUERY_CHUNK_TO_HYPERTABLE, rs -> {
+            connection.query(getChunkToHypertableQuery(), rs -> {
```
abbudao: On the first call, `getChunkToHypertableQuery()` runs the `pg_extension` probe, which can throw. `resolveChunkToHypertableQuery()` wraps that as a `DebeziumException`, so it skips the catch (SQLException) below but by then `chunkToHypertable.clear()` has already run, so a failed probe leaves the cache empty with nothing loaded to replace it. Previously the query was a compile-time constant so nothing between the clear and the query could throw. Resolving the query before mutating gets that back:

````suggestion
    private void loadTimescaleMetadata() {
        // Resolved before any mutation below: this issues the version probe on
        // first call and can throw, and clearing first would empty an already
        // populated cache without loading anything in its place.
        final String chunkToHypertableSql = getChunkToHypertableQuery();
        try {
            chunkToHypertable.clear();
            connection.query(chunkToHypertableSql, rs -> {
````

Comment (MV-GH) [reply debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/transforms/timescaledb/QueryInformationSchemaMetadata.java]: But on the first call `chunkToHypertable` is always empty? 

chunkToHypertable can't be populated without having the version first, it needs the query to populate that map?

Comment (abbudao) [reply debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/transforms/timescaledb/QueryInformationSchemaMetadata.java]: Sorry for the dumb comment, I completely misread the code.  You are right and the suggestion doesn't make any sense. Apologies for the unnecessary noise.

File: debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/transforms/timescaledb/QueryInformationSchemaMetadata.java [hunk 2]
```diff
@@ -56,6 +64,30 @@ public QueryInformationSchemaMetadata(Configuration config) {
                 "Debezium TimescaleDB metadata");
     }
 
+    private synchronized String getChunkToHypertableQuery() {
```
jpechane: I'd recommend eager initialization in constructor or itroduce a new init method so there is no need for synchronized method for regular execution.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/transforms/timescaledb/QueryInformationSchemaMetadata.java
```diff
@@ -56,6 +64,30 @@ public QueryInformationSchemaMetadata(Configuration config) {
...
+        if (chunkToHypertableQuery == null) {
+            chunkToHypertableQuery = resolveChunkToHypertableQuery();
+        }
+        return chunkToHypertableQuery;
+    }
+
+    private String resolveChunkToHypertableQuery() {
```
jpechane, line 68: The method should be called only once and an instance field should be set and use.

jpechane (changes requested):
@MV-GH Good start, thanks! I just belive we should do an eager initializaion to identify the statetemnt, set it in a field and use it.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/transforms/timescaledb/QueryInformationSchemaMetadata.java
MV-GH: I was thinking about that approach but I didnt go with that because I don't know enough about this codebase to do a behavioural change like that. Doing it in the constructor would mandate an already working connection towards the database. As it would query the TimescaleDB version during the construction. That's why I went with the lazy init.

But I can make the change if you prefer this.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/transforms/timescaledb/QueryInformationSchemaMetadata.java
MV-GH: Isn't that not the case already?

debezium-testing/debezium-testing-testcontainers/src/main/java/io/debezium/testing/testcontainers/ImageNames.java
```diff
@@ -12,6 +12,7 @@ public final class ImageNames {
     private static final String POSTGRES_IMAGE = "quay.io/debezium/postgres:15";
 
     private static final String TIMESCALE_DB_IMAGE = "quay.io/debezium/timescale-timescaledb:latest-pg15";
+    private static final String TIMESCALE_DB_PG18_IMAGE = "docker.io/timescale/timescaledb:latest-pg18";
```
jpechane:
```suggestion
    private static final String TIMESCALE_DB_PG18_IMAGE = "mirror.gcr.io/timescale/timescaledb:latest-pg18";
```

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/transforms/timescaledb/QueryInformationSchemaMetadata.java
jpechane: Yes, I'd prefer that. Probably the init method would be the best.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/transforms/timescaledb/QueryInformationSchemaMetadata.java
jpechane: You are right, the gurad is on the caller side.

MV-GH:
I have applied the requested changes

KGaneshDatta:
Thanks for the changes @MV-GH . I see you had already made the changes to make it backward compatible.

KGaneshDatta:
@MV-GH Requesting to please rebase the PR to resolve the merge conflicts.

MV-GH:
@KGaneshDatta I have rebased this PR

PR #7747, opened 2026-08-03
debezium/dbz#2338 Compare Durations with > 0 in Temporals.max/min

### LINKED ISSUE #2338: Temporals.max/min compare against == 1, giving wrong results for sub-second Durations with equal whole seconds
## Bug report

**Component:** `debezium-connector-common` — `io.debezium.time.Temporals`

`Temporals.max(d1, d2)` returns the longer duration and `Temporals.min(d1, d2)` the shorter, per their Javadoc.

**What behavior do you see?**
Both methods decide the order with `d1.compareTo(d2) == 1`. `Comparable.compareTo` is only contracted to return a negative / zero / positive int — not a normalized `-1/0/1`. `Duration.compareTo` returns the raw nanosecond difference when the two durations have the **same whole-second count**, so `== 1` is false and the methods return the wrong operand:

```
Temporals.max(500ms, 200ms) -> 200ms   (should be 500ms)
Temporals.min(500ms, 200ms) -> 500ms   (should be 200ms)
```

**Concrete impact:** `ChangeEventQueue.poll()` uses `Temporals.min(pollInterval, 5000ms)` to enforce the documented `poll.interval.ms` cap of 5000 ms (the config validator only warns, it doesn't clamp). For `poll.interval.ms` in 5001–5999 (same whole seconds as 5000), the cap leaks and the connector waits longer than the documented ceiling.

**Fix:** compare with `> 0` instead of `== 1`. Verified with a DB-free unit test (fails before, passes after).

_(Verified empirically on Corretto 17.)_

TimurRakhmatullin86:
### What

`Temporals.max`/`min` decide the order with `d1.compareTo(d2) == 1`. `Comparable.compareTo` is only contracted to return a negative/zero/positive int, not a normalized `-1/0/1`, and `Duration.compareTo` returns the raw nanosecond difference when the two durations share the same whole-second count. So `== 1` is false in that case and both methods return the wrong operand:

```
Temporals.max(500ms, 200ms) -> 200ms   (should be 500ms)
Temporals.min(500ms, 200ms) -> 500ms   (should be 200ms)
```

The production caller is `ChangeEventQueue.poll()`, which uses `Temporals.min(pollInterval, 5000ms)` to enforce the documented `poll.interval.ms` cap (the config validator only warns; it doesn't clamp). For `poll.interval.ms` values in 5001–5999 ms — same whole seconds as 5000 — the cap silently leaks and the connector waits longer than its documented ceiling.

### Fix

Compare with `> 0` per the `Comparable` contract.

### Test

Added `maxAndMinHandleSubSecondDurationsWithSameWholeSeconds` and `minCapsValueJustAboveTheLimit` to `TemporalsTest`. Both fail on the current code (max returns `0.2S`, the cap returns `5.001S`) and pass with the fix. Pure logic, no DB/container. The existing tests didn't catch it because they only compare durations whose whole seconds differ — the one case where `compareTo` happens to return `1`.

Closes debezium/dbz#2338

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/time/Temporals.java
+++ b/debezium-connector-common/src/main/java/io/debezium/time/Temporals.java
@@ -20,7 +20,7 @@ public class Temporals {
      * returned.
      */
     public static Duration max(Duration d1, Duration d2) {
-        return d1.compareTo(d2) == 1 ? d1 : d2;
+        return d1.compareTo(d2) > 0 ? d1 : d2;
     }
 
     /**
@@ -29,6 +29,6 @@ public class Temporals {
      * returned.
      */
     public static Duration min(Duration d1, Duration d2) {
-        return d1.compareTo(d2) == 1 ? d2 : d1;
+        return d1.compareTo(d2) > 0 ? d2 : d1;
     }
 }
--- a/debezium-connector-common/src/test/java/io/debezium/time/TemporalsTest.java
+++ b/debezium-connector-common/src/test/java/io/debezium/time/TemporalsTest.java
@@ -39,4 +39,22 @@ public class TemporalsTest {
         assertThat(Temporals.max(oneMilli, oneMillionNanos)).isEqualTo(oneMilli);
         assertThat(Temporals.max(oneMilli, oneMillionNanos)).isEqualTo(oneMillionNanos);
     }
+
+    @Test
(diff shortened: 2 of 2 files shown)


vjuranek (approved):
LGTM

Naros (approved):

Naros:
The test failures are all unrelated.

Naros:

PR #7748, opened 2026-08-03
debezium/dbz#2189 Normalize MySQL and MariaDB boolean aliases to TINYINT

minleejae:
Fixes debezium/dbz#2189

## Description

This change normalizes MySQL and MariaDB `BOOL` and `BOOLEAN` aliases to `TINYINT(1)` during DDL parsing.

As a result, columns created before startup and columns added during streaming are consistently emitted as Kafka Connect `INT8`. When source type propagation is enabled, the propagated type is `TINYINT` with a length of `1`.

Parser tests, streaming integration tests, and the MySQL/MariaDB documentation were updated accordingly.

## PR Checklist


Naros (commented):
So I'm curious what impact this change will have on backward compatibility with schema registry, shifting from `BOOL` to `INT8` in Kafka. @jpechane WDYT?

minleejae:
Hi @Naros,

Good question. I verified this using `AvroData` and Confluent's `BACKWARD` compatibility checker. A direct change of an existing field from Connect `BOOLEAN` to `INT8` is not backward-compatible. The Avro converter maps these types to `boolean` and `int`, respectively, and Avro does not define a `boolean` to `int` promotion. Schema Registry would therefore reject the new schema in `BACKWARD` mode.

For the nullable `ALTER TABLE ... ADD COLUMN ... BOOLEAN` scenario covered by this PR, however, the column is a new optional field, so there is no previous field type being replaced. I also verified that this schema change is backward-compatible.

With the default schema history recovery path, a normal connector upgrade does not rewrite previously recorded `BOOLEAN` columns. Schema history stores the parsed `TableChanges`, including the JDBC type, and recovery restores those logical changes. Therefore, a `BOOLEAN` column added during streaming by an older connector remains `BOOLEAN` after restart. This also means that the fix is not retroactive for already affected columns.

An incompatible transition can still occur if an existing column is redefined, DDL-preferred history recovery is enabled, or the schema is rebuilt and published to the same subject.

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlAntlrDdlParserTest.java
```diff
@@ -163,6 +166,26 @@ public void shouldParseMultipleAddColumnsWithRepeatedInstantAlgorithm() {
         assertThat(eventRefId.typeName()).isEqualTo("BIGINT");
     }
 
+    @Test
+    @FixFor("debezium/dbz#2189")
+    public void shouldMapBooleanAliasesToTinyIntWithLegacyParser() {
+        final MySqlPtAntlrDdlParser legacyParser = new MySqlPtAntlrDdlParser();
```
Naros: This appears to be another misplaced PT grammar test; move to `MySqlPtAntlrDdlParserTest`.

Naros (commented):
Hi @minleejae in `JdbcSinkDataTypeConverter`, there's an explicit cast 
```
return toTinyInt((Boolean) field.defaultValue());
```
With these changes, wouldn't that eventually throw a CCE after a DDL?

I wonder if we need to add a guard there for `Number`?

minleejae:
> in JdbcSinkDataTypeConverter, there's an explicit cast
return toTinyInt((Boolean) field.defaultValue());
With these changes, wouldn't that eventually throw a CCE after a DDL?
I wonder if we need to add a guard there for Number?

Hi @Naros Thanks for catching this - fixed by routing the default through the same conversion as regular values, with unit tests. 566feade6

debezium-connector-mysql/src/test/java/io/debezium/connector/mysql/MySqlAntlrDdlParserTest.java
minleejae: The shared `BinlogAntlrDdlParserTest#shouldMapBooleanAliasesToTinyInt` already covers this against the PT parser — it runs as `MySqlPtAntlrDdlParserTest.shouldMapBooleanAliasesToTinyInt` via inheritance. So I removed the redundant test instead of moving it: [5ff5a76](https://github.com/debezium/debezium/pull/7748/commits/5ff5a760f71b3d042b4d50c61693df3196e3e444)

Naros (approved):
This LGTM, just I would like another maintainer to give this one additional pass, particularly with the normalization.

vjuranek (approved):
LGTM

minleejae:
Hi @jpechane, I'd appreciate your feedback on this. Thanks!

PR #7750, opened 2026-08-03
debezium/dbz#2345 Consolidate retry utilities around RetryingSupplier

MrIvv:
Closes https://github.com/debezium/dbz/issues/2345

This consolidates the duplicated retry loops around a single implementation, as requested in the review of #7362 (which will drop its `RetryExecutor` class and build on `RetryingSupplier` at its next iteration).

Commit 1 (`debezium-util`, no behavior change):

- New `RetryingSupplier<V, E extends Exception>`: value-returning twin of `RetryingRunnable` with the same builder contract, generic in the checked exception type so callers keep their `throws` contracts. It hosts the single retry loop with the semantics established by dbz#2329: the `DelayStrategy` is applied after every failed attempt when no auto-heal is configured, a successful auto-heal retries immediately, and a failed auto-heal applies the delay. Log messages carry an operation name and report success after retries.
- `RetryingRunnable` becomes a thin adapter over `RetryingSupplier<Void>`. Public API and behavior are unchanged; `RetryingRunnableTest` passes as is.
- New `DelayStrategy.exponentialWithJitter(initialDelay, maxDelay, backOffMultiplier, jitterFactor)`: each sleep is randomized within +/- jitterFactor of the current delay so that concurrent retry loops hitting the same resource do not wake up in lockstep. Existing factories are untouched.
- New `RetryingSupplierTest` mirroring `RetryingRunnableTest`, plus dedicated cases asserting when the delay strategy is invoked across the no-heal / heal-succeeds / heal-fails paths.

Commit 2 (`debezium-embedded`, one deliberate behavior change):

- `RetryingCallable` becomes a thin adapter over `RetryingSupplier`, removing the second copy of the loop.
- Behavior change, intentional: `RetriableException` is now also matched through the cause chain (same principle as `io.debezium.util.ErrorHandler`), so a retriable exception wrapped by a connector triggers a retry instead of failing the task. Bounded by the same `errors.max.retries` configuration. If you prefer to keep the engine untouched here, this commit can be dropped and proposed separately; commit 1 is self-contained.

## PR Checklist

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/async/RetryingCallable.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/async/RetryingCallable.java
@@ -29,9 +29,15 @@ public abstract class RetryingCallable<V> implements Callable<V> {
     private static final Logger LOGGER = LoggerFactory.getLogger(RetryingCallable.class);
 
     private final int retries;
+    private final String customRetriableMessagePattern;
 
     public RetryingCallable(final int retries) {
+        this(retries, null);
+    }
+
+    public RetryingCallable(final int retries, final String customRetriableMessagePattern) {
         this.retries = retries;
+        this.customRetriableMessagePattern = customRetriableMessagePattern;
     }
 
     public abstract V doCall() throws Exception;
@@ -43,6 +49,7 @@ public abstract class RetryingCallable<V> implements Callable<V> {
                 .retries(retries)
                 .doGet(this::doCall)
                 .retriableExceptions(RetriableException.class)
+                .customRetriableMessagePattern(customRetriableMessagePattern)
                 .delayStrategy(delayStrategy())
                 .name("Callable")
                 .logger(LOGGER)
--- a/debezium-embedded/src/main/java/io/debezium/embedded/async/RetryingCallable.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/async/RetryingCallable.java
@@ -50,6 +50,11 @@ public abstract class RetryingCallable<V> implements Callable<V> {
                 .doGet(this::doCall)
                 .retriableExceptions(RetriableException.class)
                 .customRetriableMessagePattern(customRetriableMessagePattern)
+                // A terminal failure can wrap a retriable one on purpose (ErrorHandler wraps a
+                // retriable whose connector-side retries are exhausted in a ConnectException, which
+                // ChangeEventQueue raises from poll() to stop the task): examining the thrown
+                // exception alone preserves that contract.
+                .walkCauseChain(false)
                 .delayStrategy(delayStrategy())
                 .name("Callable")
                 .logger(LOGGER)
--- a/debezium-embedded/src/main/java/io/debezium/embedded/async/RetryingCallable.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/async/RetryingCallable.java
@@ -8,6 +8,8 @@ package io.debezium.embedded.async;
 import java.util.concurrent.Callable;
 
 import org.apache.kafka.connect.errors.RetriableException;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import io.debezium.util.DelayStrategy;
 import io.debezium.util.RetryingSupplier;
@@ -23,6 +25,8 @@ import io.debezium.util.RetryingSupplier;
  * @author vjuranek
  */
 public abstract class RetryingCallable<V> implements Callable<V> {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(RetryingCallable.class);
 
     private final int retries;
 
@@ -41,6 +45,7 @@ public abstract class RetryingCallable<V> implements Callable<V> {
                 .retriableExceptions(RetriableException.class)
                 .delayStrategy(delayStrategy())
                 .name("Callable")
+                .logger(LOGGER)
                 .build()
                 .get();
     }
--- a/debezium-embedded/src/main/java/io/debezium/embedded/async/RetryingCallable.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/async/RetryingCallable.java
@@ -8,21 +8,21 @@ package io.debezium.embedded.async;
 import java.util.concurrent.Callable;
 
 import org.apache.kafka.connect.errors.RetriableException;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 import io.debezium.util.DelayStrategy;
+import io.debezium.util.RetryingSupplier;
 
 /**
  * Extension to {@link Callable}, which allows to re-try the action if exception is thrown during the execution.
  * The action is re-tried {@code retries} number of times.
  * The delay between retries is defined by {@link DelayStrategy}, which needs to be provided by the implementing class.
+ * The action is re-tried when a {@link RetriableException} is thrown, either directly or as any exception in the
+ * cause chain of the thrown exception. The retry loop is implemented by {@link RetryingSupplier}, to which this
+ * class delegates.
  *
  * @author vjuranek
  */
 public abstract class RetryingCallable<V> implements Callable<V> {
-
-    private static final Logger LOGGER = LoggerFactory.getLogger(RetryingCallable.class);
 
     private final int retries;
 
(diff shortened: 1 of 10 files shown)


debezium-util/src/main/java/io/debezium/util/DelayStrategy.java
```diff
@@ -183,4 +184,60 @@ private boolean maxDelayedReached() {
...
+     * not wake up in lockstep. The exponential progression is computed on the un-jittered delay and stays at the
+     * maximum once reached. As soon as the criteria is not met, the delay resets to the initial value.
+     *
+     * @param initialDelay the initial delay; must be positive
+     * @param maxDelay the maximum delay; must be no less than the initial delay
+     * @param backOffMultiplier the factor by which the delay increases each pass; must be at least 1
+     * @param jitterFactor the fraction of the current delay used as randomization range; must be in [0, 1)
+     * @return the strategy
+     */
+    static DelayStrategy exponentialWithJitter(Duration initialDelay, Duration maxDelay, double backOffMultiplier, double jitterFactor) {
```
Naros, line 200: Could you add a test for this use case, please.

debezium-util/src/main/java/io/debezium/util/DelayStrategy.java
```diff
@@ -183,4 +184,60 @@ private boolean maxDelayedReached() {
...
+        final long initialDelayInMilliseconds = initialDelay.toMillis();
+        final long maxDelayInMilliseconds = maxDelay.toMillis();
+        if (backOffMultiplier < 1.0) {
+            throw new IllegalArgumentException("Backoff multiplier must be at least 1");
+        }
```
Naros: This uses `< 1.0`, which is inconsistent with the other exponential flavors that use `<= 1.0`.

debezium-util/src/main/java/io/debezium/util/RetryingSupplier.java
```diff
@@ -0,0 +1,258 @@
...
+        // -1 means infinite retries; int range is not infinite, but in this case probably a sufficient approximation.
+        // We start from `retries` as the last call attempt is done out of the retry loop and this last call either
+        // succeeds or throws an exception which is propagated further. I.e. the actual number of calls is `retries+1`,
+        // meaning one ordinary call and #`retries` is it fails.
+        int attempts = retries;
+        while (attempts != 0) {
+            try {
+                final V result = doGet.get();
+                if (attempts != retries) {
+                    logger.info("{} succeeded after {} retry attempt(s)", name, retries - attempts);
```
Naros: Let's move this to debug; having this at `INFO` could be verbose.

debezium-util/src/main/java/io/debezium/function/ThrowingSupplier.java
```diff
@@ -0,0 +1,13 @@
...
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.function;
+
+@FunctionalInterface
+public interface ThrowingSupplier<V, E extends Exception> {
+
+    V get() throws E, InterruptedException;
+
+}
```
Naros, line 19: Interfaces should always have Javadocs.

debezium-util/src/main/java/io/debezium/util/RetryingSupplier.java
```diff
@@ -0,0 +1,258 @@
...
+                                retriesExplained,
+                                exAutoHeal);
+                        executeDelayStrategy();
+                    }
+                }
+            }
+        }
+        final V result = doGet.get();
+        if (retries > 0) {
+            logger.info("{} succeeded after {} retry attempt(s)", name, retries);
```
Naros: This should be moved to `debug`.

Naros (commented):
Hi @MrIvv, the PR looks good. I've left a few inline comments.

debezium-util/src/main/java/io/debezium/util/DelayStrategy.java
MrIvv: Done

debezium-util/src/main/java/io/debezium/util/DelayStrategy.java
MrIvv: Done

debezium-util/src/main/java/io/debezium/util/RetryingSupplier.java
MrIvv: Done

debezium-util/src/main/java/io/debezium/function/ThrowingSupplier.java
MrIvv: ok, done

debezium-util/src/main/java/io/debezium/util/RetryingSupplier.java
MrIvv: done

MrIvv:
Hi @Naros , I followed your suggestions and applied the changes. 
I checked the failed pipeline jobs, and they don't seem related to this PR. 
I hope it looks good now

MrIvv (commented):
ready for review

debezium-embedded/src/main/java/io/debezium/embedded/async/RetryingCallable.java
```diff
@@ -35,35 +39,14 @@ public RetryingCallable(final int retries) {
     public abstract DelayStrategy delayStrategy();
 
     public V call() throws Exception {
```
Naros, line 47: This is a slight behavior change, as you noted in the description.

This could bite us inside `AsyncEmbeddedEngine.PollRecords`. If a poll throws an exception that wraps a `RetriableException`, today that causes the task to end with no retries; however, this change forces it through the retry mechanism because `RetryingSupplier` walks and evaluates the entire throws chain.

So should we add a builder option to toggle off the throws-chain walk? 
@jpechane @vjuranek any thoughts?

vjuranek: I don't think that we wrap `RetriableException` anywhere, so theoretically there shouldn't be any change in the behavior. I'm slightly included to keep it as it is (unless I'm wrong and there are cases where we catch wrapped `RetriableException`).

debezium-embedded/src/main/java/io/debezium/embedded/async/RetryingCallable.java
```diff
@@ -35,35 +39,14 @@ public RetryingCallable(final int retries) {
...
-                if (Thread.currentThread().isInterrupted()) {
-                    throw new InterruptedException("Callable was interrupted while sleeping in DelayStrategy");
-                }
-            }
-        }
-        return doCall();
+        return RetryingSupplier.<V, Exception> builder()
+                .retries(retries)
+                .doGet(this::doCall)
+                .retriableExceptions(RetriableException.class)
```
vjuranek, line 51: While the original implementation is missing it (as I probably forgot on it, at least I don't recall any specific reason), we are missing internal option `custom.retriable.exception` here, so I wonder if we should honor it here. This could be potentially change in the behavior, but not considering it, I'd consider as a bug, so IMHO it would be good to add all exceptions from `custom.retriable.exception` as well.

MrIvv: i add the missing `custom.retriable.exception` in the last commit

debezium-embedded/src/main/java/io/debezium/embedded/async/RetryingCallable.java
MrIvv: @vjuranek I found a concrete case in io.debezium.pipeline.ErrorHandler#setProducerThrowable: when a retriable failure has exhausted the connector-side retries, it is deliberately wrapped in a ConnectException ("...will be stopped"), and ChangeEventQueue documents that raising it from poll() must stop the connector and mark it FAILED. With the cause-chain walk, PollRecords would reclassify that terminal failure as retriable, and with the default errors.max.retries=-1 it would retry forever.

@Naros So rather than removing the walk, I would go with the builder toggle you suggested. The walk stays the documented default (which is what the #7362 worker relies on, and what the snapshot retry proposed in [DDD-59 / PR #60](https://github.com/debezium/debezium-design-documents/pull/60) would need to unwrap its ConnectException), and only RetryingCallable opts out to preserve the current engine semantics exactly:
```
// RetryingSupplier.Builder, default true (documented)
public Builder<V, E> walkCauseChain(boolean walkCauseChain) { ... }

// RetryingCallable#call()
.retriableExceptions(RetriableException.class)
.walkCauseChain(false)
```
Should I proceed with adding this toggle, or would you prefer I drop the cause-chain walk entirely?

debezium-embedded/src/main/java/io/debezium/embedded/async/RetryingCallable.java
```diff
@@ -25,45 +29,36 @@ public abstract class RetryingCallable<V> implements Callable<V> {
...
+                .customRetriableMessagePattern(customRetriableMessagePattern)
+                // A terminal failure can wrap a retriable one on purpose (ErrorHandler wraps a
+                // retriable whose connector-side retries are exhausted in a ConnectException, which
+                // ChangeEventQueue raises from poll() to stop the task): examining the thrown
+                // exception alone preserves that contract.
+                .walkCauseChain(false)
```
Naros, line 57: I think it would be good to add a small note in the class JavaDoc that, while this class takes the custom retriable message pattern, it expects that pattern in the outermost thrown exception and doesn't check nested exceptions. If different behavior is needed, `RetryingSupplier` should be used instead.

But I wonder whether specifying the custom retriable pattern here is entirely redundant, since any error propagates up to the runtime and the ErrorHandler takes over, honoring the same property. So could this lead to more retries than a user might expect?

debezium-embedded/src/main/java/io/debezium/embedded/async/RetryingCallable.java
```diff
@@ -12,11 +12,15 @@
...
 
 import io.debezium.util.DelayStrategy;
+import io.debezium.util.RetryingSupplier;
 
 /**
  * Extension to {@link Callable}, which allows to re-try the action if exception is thrown during the execution.
  * The action is re-tried {@code retries} number of times.
  * The delay between retries is defined by {@link DelayStrategy}, which needs to be provided by the implementing class.
+ * The action is re-tried when a {@link RetriableException} is thrown, either directly or as any exception in the
+ * cause chain of the thrown exception. The retry loop is implemented by {@link RetryingSupplier}, to which this
```
Naros, line 22: This JavaDoc is a bit misleading. It only retries the outermost exception because this class's implementation disables the walk chain to match prior behavior.

Naros (commented):
Hi @MrIvv @vjuranek I have just one question on the new introduction of the retriable exception handling, specifically in `RetryingCallable`.

MrIvv:
Hi @Naros 
I've pushed two small JavaDoc commits.
Regarding the redundancy question, I went through the paths that can raise exceptions from task.poll():
Producer-thread failures: These always go through ErrorHandler.setProducerThrowable, which already honors custom.retriable.exception (walking the cause chain) and normalizes the outcome. 
A failure classified as retriable (by exception type or by the pattern) with retries left surfaces as a RetriableException, so the pattern is inert for it here. An exhausted
or non-retriable failure surfaces as the ConnectException ("...will be stopped") wrapper. Since RetryingCallable classifies only the outermost exception, that wrapper is never reclassified, which is exactly why the cause-chain walk is disabled here. Therefore, the user only gets the retries ErrorHandler already granted.
Synchronous failures: ErrorHandler never sees failures thrown synchronously by BaseSourceTask.poll() itself (e.g., the restart attempt inside startIfNeededAndPossible() or performCommit()). For those, the property was previously silently ignored and the task died even when the user had declared the failure retriable. This covers the gap @vjuranek pointed out.
Net effect: No additional retries on failures ErrorHandler already handles, and we now correctly apply retries for sync-path failures (bounded by the same errors.max.retries).

I'm open to making any further adjustments you deem appropriate

PR #7754, opened 2026-08-04
debezium/dbz#2350 Cache PostgreSQL snapshot column type resolution per column

### LINKED ISSUE #2350: PostgreSQL snapshot resolves each column's type once per row instead of once per column (getColumnValue)

Faster snapshots for **wide PostgreSQL tables**.

During a snapshot, `PostgresConnection.getColumnValue(...)` resolves the column's type on every invocation:

```java
final ResultSetMetaData metaData = rs.getMetaData();
final String columnTypeName = metaData.getColumnTypeName(columnIndex);
final PostgresType type = getTypeRegistry().get(columnTypeName);
```

`getColumnValue` is called **once per column per row** (via `JdbcConnection.rowToArray`), so for a table with N columns and R rows the snapshot performs N×R `ResultSetMetaData#getColumnTypeName` + `TypeRegistry#get` lookups — even though a column's type is constant for the lifetime of the `ResultSet`.

This is on the path for every snapshot flavor: `rowToArray` is called by `RelationalSnapshotChangeEventSource` (initial and blocking snapshots) and `AbstractIncrementalSnapshotChangeEventSource` (incremental, read-only-incremental, and ad-hoc snapshots). Streaming (logical decoding) does not use `getColumnValue`. It is also orthogonal to chunked snapshots (3.5+), which parallelize across chunks but do not reduce the per-row type-resolution cost and fall back to a single chunk for keyless tables.

The snapshot producer is single-threaded per table, and in our pipeline (built on the Debezium engine, with a downstream consumer that comfortably keeps up) it is the throughput bottleneck: end-to-end throughput for a table is governed entirely by how fast the single snapshot thread produces rows. For a wide (82-column) table that thread is pinned at ~100% CPU and the migration plateaued at **~25k rows/s**.

Instrumenting the per-row cost of the three snapshot-loop stages — `rowToArray` (column read/convert, includes `getColumnValue`), `rs.next()`, and `dispatch` (Struct build + enqueue):

| Stage | Before | After |
|---|---|---|
| `rowToArray` | ~16.8 µs/row | **~10.9 µs/row** (≈ −35%) |
| `rs.next()` | ~7.6 µs/row | ~7.6 µs/row (unchanged) |
| `dispatch` | ~16.5 µs/row | ~16.5 µs/row (unchanged) |

The three stages sum to ~100% of the thread's wall-clock time (it is never idle), so removing ~6 µs/row from `rowToArray` translates directly into more rows/s. With the caching in place the same table's snapshot climbs to **over 30k rows/s**; the gain scales with column count.

---

Cache the resolved `PostgresType` per column for the lifetime of the `ResultSet`, so the `getColumnTypeName` + `TypeRegistry#get` lookups run **once per column** instead of once per column per row. The resolution is byte-for-byte identical to the current inline path; only its frequency changes. A snapshot `ResultSet` is read by a single thread, so a per-`ResultSet` cache (keyed by `ResultSet` identity) is sufficient and thread-safe. The `ResultSet` is referenced only weakly, as a cache-identity token, so the cache does not retain a closed `ResultSet` on a long-lived connection.

PR: https://github.com/debezium/debezium/pull/7754

mly-zju:
# What

`PostgresConnection.getColumnValue(...)` resolves the column's type on every call:

```java
final ResultSetMetaData metaData = rs.getMetaData();
final String columnTypeName = metaData.getColumnTypeName(columnIndex);
final PostgresType type = getTypeRegistry().get(columnTypeName);
```

`getColumnValue` is invoked **once per column per row** (via `JdbcConnection.rowToArray`), so for a table with N columns and R rows the snapshot performs N×R `ResultSetMetaData#getColumnTypeName` + `TypeRegistry#get` lookups — even though a column's type is constant for the lifetime of the `ResultSet`.

This change resolves and caches the `PostgresType` **per column**, keyed by `ResultSet` identity in a `ThreadLocal` (a snapshot `ResultSet` is read by a single thread). The type lookups now run once per column instead of once per column per row. The resolution itself is unchanged (same `getColumnTypeName` + `TypeRegistry#get`), so behavior is identical for every type — arrays, domain/user-defined types, `NUMERIC`, temporal types, etc.; only the frequency of the lookup changes. Any `SQLException` from the lookup still propagates to the existing `catch` in `getColumnValue`, preserving the "not a known type" fallback to `super.getColumnValue`.

The `ResultSet` is held only through a `WeakReference`, purely as a cache-identity token: a `PostgresConnection` outlives any single snapshot `ResultSet`, so a strong reference would pin the (already closed) `ResultSet` and everything it retains until the next snapshot replaced it. The weak reference lets it be collected as soon as the snapshot releases it; `get()` then returns `null`, which simply triggers a harmless rebuild on the next query.

# Scope

`getColumnValue` is reached only via `JdbcConnection.rowToArray`, whose only callers are `RelationalSnapshotChangeEventSource` (initial and blocking snapshots) and `AbstractIncrementalSnapshotChangeEventSource` (incremental, read-only-incremental, and ad-hoc snapshots). Streaming (logical decoding) never calls `getColumnValue`, so it is unaffected. Because the cache is keyed by `ResultSet` identity, it is invalidated on every new `ResultSet` — so a connection reused across many chunk queries (incremental snapshot) re-resolves per chunk and never serves a stale type.

This is orthogonal to chunked snapshots (3.5+): chunking parallelizes work across chunks but does not reduce the per-row type-resolution cost this change targets, and it falls back to a single chunk for keyless tables / when `message.key.columns` is not set. The optimization is therefore complementary and also benefits the single-threaded path. (We originally reproduced the bottleneck on Debezium 3.4, which predates chunked snapshots.)

# Why

The snapshot producer is **single-threaded per table**, and in our setup it is the throughput bottleneck. We run an ingestion pipeline on top of the Debezium engine; the downstream consumer comfortably keeps up, so end-to-end throughput for a given table is governed entirely by how fast the single snapshot thread produces rows. For a wide (82-column) table, that thread is pinned at ~100% CPU and the migration plateaued at **~25k rows/s**.

To locate the cost we instrumented the per-row time of the three stages of the snapshot loop — `rowToArray` (column read/convert, which includes `getColumnValue`), `rs.next()`, and `dispatch` (Struct build + enqueue):

| Stage | Before | After |
|---|---|---|
| `rowToArray` | ~16.8 µs/row | **~10.9 µs/row** (≈ −35%) |
| `rs.next()` | ~7.6 µs/row | ~7.6 µs/row (unchanged) |
| `dispatch` | ~16.5 µs/row | ~16.5 µs/row (unchanged) |

The three stages sum to ~100% of the thread's wall-clock time (it is never idle), so removing ~6 µs/row from `rowToArray` translates directly into more rows/s on the single producer thread — `rs.next()` and `dispatch` are unchanged, confirming the win is isolated to the column-type lookup. With the caching in place the same table's snapshot climbs to **over 30k rows/s**. The gain scales with column count.

Fixes https://github.com/debezium/dbz/issues/2350

# Test

Added:

- `PostgresConnectionColumnTypeCacheTest` — a unit test that pins the cache contract: a column is resolved exactly once regardless of row count; a **new `ResultSet` forces re-resolution** (the guarantee incremental/blocking/ad-hoc snapshots rely on when a connection is reused across chunk queries); and the resolved type and returned value are identical to the un-cached path.
- `ConnectionIT#shouldReadEveryRowWhenColumnTypeIsCachedPerColumn` — creates a multi-column table (integer, `numeric`, text, boolean, integer[]) with several rows, reads every column of every row through `getColumnValue`, and asserts all values decode correctly, exercising the per-column cache being populated on the first row and reused across the array / numeric / default type paths.

`BlockingSnapshotIT` passes against a `pgoutput` database. Existing snapshot and type ITs (which drive every PostgreSQL type through `getColumnValue`) continue to provide behavioral coverage.


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
```diff
@@ -864,6 +874,28 @@ public Object getColumnValue(ResultSet rs, int columnIndex, Column column, Table
...
+     * Resolves the {@link PostgresType} for a column, caching it per column so the
+     * {@link ResultSetMetaData#getColumnTypeName(int)} + {@link TypeRegistry#get(String)} lookups run once per column
+     * for the lifetime of a {@link ResultSet}, instead of once per column per row. The resolution is identical to the
+     * inline lookup; only its frequency changes. Any {@link SQLException} propagates to the caller so the
+     * "not a known type" fallback in {@link #getColumnValue} is preserved.
+     */
+    private PostgresType resolveColumnType(ResultSet rs, int columnIndex) throws SQLException {
+        final ColumnTypeCache cache = columnTypeCache.get();
+        if (cache.resultSet != rs) {
+            cache.resultSet = rs;
```
Naros: The issue description highlights that snapshots are a single-table per thread, but that's no longer the case when `snapshot.max.threads` is greater than `1` and you're using the new default chunked snapshots used in Debezium 3.5+. 

Furthermore, this method is used by all snapshot flavors: incremental, blocking, and initial.

My concern is this line where we create a hard reference to the `ResultSet` and there is no place where the type cache is cleared post-snapshot. So at the very least, the last result set read by either of these three snapshot flavors will be held until the thread terminates, which would be when the connector is stopped.

Rather than relying on the `ResultSet` for cache identity, could `TableId` serve the same purpose and we avoid a reference to a short-lived object type?

Naros (commented):
Hi @mly-zju the performance numbers look good, but I have some doubts. Could you take a look at my concern below?

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
mly-zju: Thanks @Naros  — good catches on all three.

**Snapshot threading / chunked snapshots:** you're right, the "single-table per thread" wording doesn't hold for 3.5+ with the default chunked snapshot. For context on where the framing came from: we reproduced this on Debezium 3.4, which predates chunked snapshots — so there a table snapshots single-threaded even when it has a primary key, and that's how we hit the per-row ResultSetMetaData#getColumnTypeName + TypeRegistry#get cost on a keyed table. I'll reword the issue to reflect 3.5+ semantics. Two reasons the change still helps on 3.5+:

* Chunked snapshots still call getColumnValue once per column per row within each chunk, so dropping the per-row lookup speeds up each parallel chunk — not only a single-threaded scan.
* Chunking needs a key (PK or message.key.columns); a table without one collapses to a single chunk, so this remains the only lever in that case.

So it's complementary to chunking rather than superseded by it — it helps 3.5+ users per-chunk and on keyless tables.

**All snapshot flavors:** understood — since getColumnValue is shared by initial/blocking/incremental, I'll make sure the reworked version is correct across all three (including an incremental snapshot that spans a schema change).

**ResultSet reference / cache lifetime:** you're absolutely right, and this is the important one — the ThreadLocal keeps a hard reference to the last ResultSet with no post-snapshot cleanup, so on a pooled snapshot thread it's retained until the connector stops. I'd rather not just key by TableId blindly, since the cache is indexed by column position and I want to be sure that's stable across flavors and the DDL-during-incremental case. Let me rework it so it doesn't retain the ResultSet at all, validate it across the snapshot flavors, and update the PR — I'll ping you when it's ready for another look. Thanks for the thorough review.

Naros: Thanks for the update @mly-zju, that sounds like a plan. Looking forward to the changes. :rocket:

mly-zju:
Thanks for the detailed review — addressed both points, and dug into the "other scenarios" question.

**On the leak (ThreadLocal holding the `ResultSet`)**

Good catch — a strong reference here pins the (already-closed) snapshot `ResultSet` and everything it retains until the next snapshot replaces it, on a connection that outlives any single snapshot. Fixed by holding the `ResultSet` only through a `WeakReference`, used purely as a cache-identity token:

```java
private static final class ColumnTypeCache {
    private WeakReference<ResultSet> resultSetRef;
    private PostgresType[] typesByColumnIndex;
}
...
if (cache.resultSetRef == null || cache.resultSetRef.get() != rs) {
    cache.resultSetRef = new WeakReference<>(rs);
    cache.typesByColumnIndex = new PostgresType[rs.getMetaData().getColumnCount() + 1];
}
```

While the snapshot is reading, the source holds the `ResultSet` strongly, so `get()` returns it and the cache stays valid across all rows. Once the snapshot releases it, the GC is free to reclaim it; `get()` then returns `null`, which just triggers a harmless rebuild on the next query. No retention beyond the snapshot.

**On keying by `TableId` instead**

I considered it but stayed with `ResultSet` identity: `TableId` isn't safe across an incremental snapshot that spans a schema change (same table, changed column layout between chunks), whereas each chunk is a distinct `ResultSet`, so identity-keying is inherently precise and self-invalidating. It also keeps the resolution byte-for-byte identical to the original inline path — the only change is *frequency* (once per column per `ResultSet` instead of once per column per row).

**On "does this work for scenarios other than snapshot?"**

I traced the call graph. `getColumnValue` is reached through exactly one path — `JdbcConnection.rowToArray()` — and `rowToArray` has only two callers: `RelationalSnapshotChangeEventSource` (initial + blocking snapshot) and `AbstractIncrementalSnapshotChangeEventSource` (incremental, read-only-incremental, ad-hoc). Streaming (logical decoding) never calls `getColumnValue`, so it's unaffected. The one case that could break a naive cache is incremental snapshot, which runs many chunk queries on the same connection — but since the cache invalidates on every new `ResultSet`, each chunk re-resolves from scratch. Correct by construction.

To lock that down I added a unit test (`PostgresConnectionColumnTypeCacheTest`) asserting (1) one resolution per column regardless of row count, (2) re-resolution on a new `ResultSet`, and (3) identical resolved type and value. I also ran `BlockingSnapshotIT` green against a `pgoutput` database as an end-to-end check of the non-initial snapshot path; the incremental / read-only-incremental IT matrix runs in CI.

For context on the chunked-snapshot discussion: we originally reproduced this on **Debezium 3.4**, which predates chunked snapshots, on a PK table. Chunking helps wall-clock time when it can parallelize, but (a) it falls back to a single chunk for keyless tables / no `message.key.columns`, and (b) it doesn't reduce the *per-row* type-resolution cost this PR targets — so the single-threaded improvement remains meaningful alongside chunking.

I've also updated the PR and issue descriptions to reflect the above (scope across snapshot flavors, the WeakReference, and the chunked-snapshot relationship).

Naros:
> I considered it but stayed with `ResultSet` identity: `TableId` isn't safe across an incremental snapshot that spans a schema change (same table, changed column layout between chunks), whereas each chunk is a distinct `ResultSet`, so identity-keying is inherently precise and self-invalidating.

Hi @mly-zju, why is `TableId#compareTo` not safe? The `TableId` is just a fully qualified identifier of the table, and it doesn't have anything directly tied to the column layouts in the relational `Table` model class.

> Fixed by holding the ResultSet only through a WeakReference, used purely as a cache-identity token

The `WeakReference` addresses the GC concerns with `ResultSet`, but see my above question.

Additionally, I'm still not convinced that a `ThreadLocal` is needed. Every thread we spawn for snapshots uses its own JdbcConnection instance, and given that these classes are inherently not thread-safe, nor is `ResultSet`, I don't really see any value add.

Without the `ThreadLocal` and using a plain instance field, that removes:
* Per-thread allocation overhead
* Avoids any conceptual expectation that the class might be thread-safe, when it isn't.

@jpechane @vjuranek any opinion?

mly-zju:
Thanks @Naros — both good points.

**On `ThreadLocal`: you're right, dropped it** (pushed as a follow-up commit). Each snapshot task polls its own `JdbcConnection` from the pool (`RelationalSnapshotChangeEventSource`, `connectionPool.poll()` per table callable, returned when done), so a given connection is only ever touched by one thread at a time. `ResultSet`/`JdbcConnection` aren't thread-safe anyway, so the `ThreadLocal` only added per-thread allocation and a misleading hint of thread-safety. It's now a plain instance field:

```java
private WeakReference<ResultSet> cachedResultSetRef;
private PostgresType[] cachedColumnTypes;
```

**On `TableId`: you're right, and let me retract "not safe" — that was wrong.** `TableId#compareTo` is a correct, cheap FQN comparison; nothing wrong with it. My point was only about cache-key *granularity*, and I stated it badly.

The cache maps `columnIndex → PostgresType`; the only question is when to rebuild it. Keying by `ResultSet` identity rebuilds per `ResultSet`, so the array is always scoped to exactly the columns whose metadata produced it. Keying by `TableId` (names only, no column info) would instead share one array across every `ResultSet` of the same table — which is fine as long as `columnIndex → type` never shifts for a table mid-snapshot. That assumption holds in practice; I just didn't see it buying anything, since `ResultSet` identity already removes the hot path (N×R → N metadata reads) and the per-chunk rebuild it implies is negligible (~N reads per chunk, not per row).

So it's not a safety issue either way — `ResultSet` identity is simply the stricter key for the same cost, which is why I leaned to it. Happy to switch to `TableId` if you'd prefer it; performance is identical.

Naros:
Hi @mly-zju, let me talk with a colleague and get back to you. I definitely now understand more clearly the reason why you want to use the `ResultSet` in the first place, it just wasn't immediately obvious to me. In the meantime, lets wait on CI.

Naros:
Hi @mly-zju, would it be possible to avoid the member-level cache completely where we need to reference the `ResultSet`, and instead use an override of `rowToArray` for `PostgresConnection` that creates a column metadata lookup context object before the `ResultSet` iteration? That could be passed to a new custom `getColumnValue` method that's used for `rowToArray` and the current method contents could be refactored so we can share the bulk of the switch lookups.

rk3rn3r:
One thing that was coming to my mind without getting deeper into the details of this topic, instead of using the "formal" way of fetching the metadata separately I wonder if there is any property in the jdbc config that fetches the metadata along with the payload and avoids IO for the metadata fetch or of the way how rows are fetched can be changed to fetch required metadata along the payload in some manual way?

mly-zju:
Thanks both, @Naros @rk3rn3r.

Quick note for @rk3rn3r first: the type metadata isn't a per-row fetch — column OIDs arrive once with the query's `RowDescription`, and pgjdbc's `getColumnTypeName` resolves OID→name from its cached `TypeInfoCache` (a DB round-trip only on a miss). So the per-row cost is CPU (OID→name + the `TypeRegistry` lookup), not IO.

On dropping the member cache: I'd actually tried resolving straight from the `Column` model early on — `getColumnValue` receives the `Column`, and `Column#nativeType()` is the OID resolved at schema discovery (`PostgresConnection` already does `typeRegistry.get(column.nativeType())` elsewhere):

```java
final PostgresType type = getTypeRegistry().get(column.nativeType());
```

It measured a touch faster than the cache and needs no member field, no context object, nothing for a refactor to preserve. I'd shelved it over a worry that it wasn't equivalent for domain types — but digging in, that doesn't hold. `getColumnValue` only branches on `isArrayType()` and the built-in base-type OIDs (never a domain OID), and Postgres reports the *base* type OID for scalar domains on the wire, so `getColumnTypeName` already resolves a domain to its base today (confirmed on PG 17: domains over numeric/text/money report `numeric`/`text`/`money`). `nativeType()` is that same base OID, so both paths take the same branch.

I verified rather than asserted: `RecordsSnapshotProducerIT` (the full type dataset via snapshot) stays green, and I added a `DomainTypesIT` case that snapshots plain/domain column pairs over numeric, money and timestamp and asserts each domain column emits identically to its base — passing both before and after the change. Pushed as a commit.

If you'd rather keep the exact `getColumnTypeName` resolution or prefer the per-`ResultSet` context route, happy to roll back — but this looked like the simplest thing that satisfies both of your points.

rk3rn3r:
When the process is only CPU bound and not IO (beside RAM / cache lookup - and which is only moved up by one layer), I lean towards saying that all speed improvements here are marginal.

mly-zju:
Thanks @rk3rn3r — you're right that for a typical, narrow table this is in the noise; there's no IO here, it's CPU plus a map lookup. But a few reasons I'd still lean toward keeping it:

1. **The bottleneck is measured in a real system, not theoretical.** With this fix the single-threaded snapshot went from ~25k to over 30k rows/s on a live migration — real and reproducible, so I wouldn't call it marginal. And "CPU-bound, not IO" is actually *why* it matters here: the snapshot producer is single-threaded and sits pinned at ~100% CPU, so there's no IO wait to hide the cost behind — CPU time is the throughput ceiling, and shaving it converts directly to rows/s.

2. **The hotspot is a genuine inefficiency, and the fix is easy.** We re-resolve each column's type from `ResultSetMetaData` on every row even though it's constant for the lifetime of the `ResultSet`. And it's a bit more than "moved up one layer" — the change drops the per-row `getColumnTypeName` OID→name derivation (and its `String` handling) entirely, resolving from the OID already on the column model: one `oidToType` lookup instead of a name conversion plus a name lookup. The fix is small and actually *removes* code — no cache, no per-row metadata access. A real inefficiency with a cheap, simplifying fix seems worth taking.

3. **It scales with table width.** That was an 82-column table; the cost is per column per row, so it only gets worse — a 100- or 1000-column table pays proportionally more. Wide tables are exactly where this bites, and where the ~25k→30k gain came from.

Happy to keep discussing, but on balance this looks like a clear, low-cost win.

rk3rn3r:
Got your point and it could be be valid then. But I hope no one will ever have a 1000 column table. 😅 
Out of curiosity, do you maybe have a profiler run for both scenarios? That could be really interesting to see differences.

Naros:
> Got your point and it could be be valid then. But I hope no one will ever have a 1000 column table. 

It's actually more common than you might think for denormalized table representations @rk3rn3r. We have users with these types of tables in other connector deployments (the largest I've seen was just under 800 columns), and with the increasing popularity of PostgreSQL and enterprises migrating toward it, who knows ;-)

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/PostgresConnection.java
```diff
@@ -814,12 +813,14 @@ public TypeRegistry getTypeRegistry() {
...
     public Object getColumnValue(ResultSet rs, int columnIndex, Column column, Table table) throws SQLException {
         try {
-            final ResultSetMetaData metaData = rs.getMetaData();
-            final String columnTypeName = metaData.getColumnTypeName(columnIndex);
-            final PostgresType type = getTypeRegistry().get(columnTypeName);
+            // Resolve the column's type from the relational model's OID (captured once at schema discovery) rather
+            // than re-deriving it from ResultSetMetaData#getColumnTypeName on every column of every row. Postgres
+            // collapses scalar domains to their base type on the wire, so this matches the ResultSetMetaData
+            // resolution for everything the logic below distinguishes -- array-ness and the built-in base-type OIDs.
+            final PostgresType type = getTypeRegistry().get(column.nativeType());
```
Naros, line 820: @mly-zju I definitely think this is the cleanest solution yet.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/ConnectionIT.java
```diff
@@ -87,6 +91,54 @@ void shouldReadUnqualifiedUserDefinedTypeNameRegardlessOfSearchPath() throws SQL
...
+            Table table = tables.forTable(new TableId(null, "dbz2350", "t"));
+
+            List<Object> ids = new ArrayList<>();
+            List<Object> labels = new ArrayList<>();
+            conn.query("SELECT id, amount, label, flag, tags FROM dbz2350.t ORDER BY id", rs -> {
+                ResultSetMetaData metaData = rs.getMetaData();
+                while (rs.next()) {
+                    // Exercise getColumnValue for every column of every row. The per-column type cache is
+                    // populated on the first row and reused for the remaining rows, and must still decode each
+                    // value; a stale/incorrect cache would surface as a null or wrong value here.
```
Naros: This comment appears stale after the cache change.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/DomainTypesIT.java
```diff
@@ -80,6 +80,42 @@ public void shouldExportDomainTypeInArrayAsUnknown() throws Exception {
         assertThat(after.get("tokens")).isEqualTo(Arrays.asList(ByteBuffer.wrap("bar".getBytes()), ByteBuffer.wrap("baz".getBytes())));
     }
 
+    @Test
+    public void shouldSnapshotDomainColumnsIdenticallyToTheirBaseType() throws Exception {
```
Naros, line 85: Let's add the `@FixFor` annotation here, too.

Naros (commented):
Hi @mly-zju just a few tiny fix-ups and I think it's G2G 🚀

Naros (approved):
LGTM @mly-zju, great job!  Just waits for CI now 🚀

Naros:
Applied, thanks @mly-zju :tada:

PR #7757, opened 2026-08-04
debezium/dbz#1344 Cast bound values to the destination enum type on PostgreSQL

### LINKED ISSUE #1344: JDBC Sink connector: allow sink to an enum column [DBZ-8900]
Migrated from [DBZ-8900](https://issues.redhat.com/browse/DBZ-8900)

 
h1. Bug report

I am trying to sink from a source table that has an ENUM column to a destination table that has the same column, same enum type. I have configured my source PG connector with {{column.propagate.source.type}} and I can see the avro schema in Apicurio has:
{noformat}
        {
          "name": "role",
          "type": {
            "type": "string",
            "connect.version": 1,
            "connect.parameters": {
              "allowed": "system,assistant,user",
              "__debezium.source.column.type": "ROLE",
              "__debezium.source.column.length": "2147483647",
              "__debezium.source.column.scale": "0",
              "__debezium.source.column.name": "role"
            },
            "connect.name": "io.debezium.data.Enum"
          }
        },{noformat}
So that seems to be working. But when I try to sink with the connector, the PG query fails with 
{noformat}
Caused by: org.postgresql.util.PSQLException: ERROR: column "role" is of type role but expression is of type character varying{noformat}
The relevant part of the upsert query is
{noformat}
ON CONFLICT ("id","version","chat_id") DO UPDATE SET "role"=EXCLUDED."role","content"=EXCLUDED."content"{noformat}
 I think it's missing a SQL cast for my enum, like adding ::role right after the value.

3.0.8.Final

<Your answer>

PG 16 - docker

Sink-ing to an enum column should work
h2. What behavior do you see?

<Your answer>

<Your answer>

<Your answer>

<Your answer>

<Your answer>

<Your answer>

sundongkim-dev:
Fixes debezium/dbz#1344

## Description

Sinking an `io.debezium.data.Enum` field into a PostgreSQL enum column fails:

```
ERROR: column "role" is of type role but expression is of type character varying
```

PostgreSQL does not implicitly cast a bound `character varying` value to an enum column. A literal works because it is initially of unknown type, but the sink binds values as parameters, so the cast has to be explicit.

`GeneralDatabaseDialect#getSchemaType` resolves such a field to the PostgreSQL `EnumType`, but that type has no `getQueryBinding` override, so `AbstractType` falls back to `PostgresDatabaseDialect#getQueryBindingWithValueCast`, which only casts `uuid`, `json` and `jsonb`. The neighbouring `UuidType` and `InetType` already override `getQueryBinding`; enum was simply missing.

`EnumType#getQueryBinding` now casts to the destination column's type name. No catalog query or cache is needed: the name is already on `ColumnDescriptor#typeName`, taken from `DatabaseMetaData#getColumns`.

This follows the direction given in the review of #6990, which took the dialect-based approach:

> I don't think placing this logic in the Dialect is the correct location for this. [...] in general, any casting should be handled by the type subsystem in `io.debezium.connector.jdbc.type`. [...] could this be solved by using a transform to make sure the field is encoded as a special logical type, i.e. `io.debezium.data.Enum`, and then letting the `EnumType` handle the casting? And is the enum cache even necessary in this case?

`getQueryBindingWithValueCast` is therefore left untouched. Credit to @bpaquet for the original report and analysis.

**Quoting.** The driver reports the type name qualified and quoted (`"schema"."type"`) only when the type is not on the search path; otherwise it reports `pg_type.typname` verbatim. A verbatim name would be folded to lower case by the parser, so `cast(? as Role)` fails with `type "role" does not exist`. The name is quoted here unless the driver already quoted it. Quoting is a no-op for built-in names (`cast(? as "text")` is valid), which matters when the destination column is not an enum at all.

**UNNEST batch path.** `getBatchInsertStatement` derives its array cast from the field schema (`?::<type>[]`), which for an enum only resolves to the `character varying` fallback, so that path fails with `expression is of type text` regardless of the binding above. Since it never looks at the destination column, records carrying an enum field are routed to the row-wise path via the existing `hasUnnestUnsupportedField` escape hatch, which already serves the geometric `STRUCT` types. `getBatchUpsertStatement` delegates to `getBatchInsertStatement` and is covered by the same check.

**Scope.** This covers fields whose schema is named `io.debezium.data.Enum`, i.e. what a source emits with `column.propagate.source.type` enabled. debezium/dbz#1490 reports the same symptom for a plain unnamed `STRING` field, which never reaches an enum-aware type; that one is not addressed here and remains open. An enum array column is also out of scope: an array field resolves to `ArrayType`, which takes only the element type's `getTypeName` for `createArrayOf` and never reaches the binding above.

**Tests.** `JdbcSinkColumnTypeMappingIT` gains a round trip into a hand-created enum column across both insert modes, and one over the UNNEST batch path. The first uses a mixed-case enum type name so the quoting is actually exercised. Reverting each part of the change fails them with the reported errors respectively: `is of type role_x but expression is of type character varying`, `... but expression is of type text`, and `type "role_x" does not exist`.

Verified against `it-postgresql` (340 tests) and `e2e-postgresql`, including `JdbcSinkPipelineToPostgresIT#testEnumDataType`, which covers the case where schema evolution created the destination column as `text`.

The `io.debezium.data.Enum` row of the dialect type mapping table in the documentation describes the column type used when the connector creates the table. `getTypeName` is unchanged, so that row still holds and the documentation is not updated.

## PR Checklist

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/EnumType.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/EnumType.java
@@ -33,9 +33,7 @@ class EnumType extends AbstractType {
 
     @Override
     public String getQueryBinding(ColumnDescriptor column, Schema schema, Object value) {
-        // PostgreSQL does not implicitly cast a bound character varying value to an enum column, so the
-        // cast must be explicit.
-        return "cast(? as " + quoteTypeName(column.getTypeName()) + ")";
+        return "cast(? as %s)".formatted(quoteTypeName(column.getTypeName()));
     }
 
     /**
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/EnumType.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/EnumType.java
@@ -13,6 +13,7 @@ import io.debezium.connector.jdbc.type.AbstractType;
 import io.debezium.connector.jdbc.type.JdbcType;
 import io.debezium.connector.jdbc.type.connect.ConnectStringType;
 import io.debezium.data.Enum;
+import io.debezium.sink.column.ColumnDescriptor;
 
 /**
  * An implementation of {@link JdbcType} for {@link Enum} column types.
@@ -28,6 +29,25 @@ class EnumType extends AbstractType {
     @Override
     public String[] getRegistrationKeys() {
         return new String[]{ Enum.LOGICAL_NAME };
+    }
+
+    @Override
+    public String getQueryBinding(ColumnDescriptor column, Schema schema, Object value) {
+        // PostgreSQL does not implicitly cast a bound character varying value to an enum column, so the
+        // cast must be explicit.
+        return "cast(? as " + quoteTypeName(column.getTypeName()) + ")";
+    }
+
+    /**
+     * Quotes the column's type name for use in a cast. The driver reports the name qualified and quoted
+     * only for a type that is not on the search path; otherwise it reports {@code pg_type.typname}
+     * verbatim, which PostgreSQL would fold to lower case unless it is quoted here.
+     */
+    private static String quoteTypeName(String typeName) {
+        if (typeName.startsWith("\"")) {
+            return typeName;
+        }
+        return "\"" + typeName.replace("\"", "\"\"") + "\"";
     }
 
     @Override
--- a/debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/postgres/JdbcSinkColumnTypeMappingIT.java
+++ b/debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/postgres/JdbcSinkColumnTypeMappingIT.java
@@ -31,7 +31,9 @@ import io.debezium.connector.jdbc.junit.jupiter.PostgresInsertModeArgumentsProvi
 import io.debezium.connector.jdbc.junit.jupiter.PostgresInsertModeArgumentsProvider.PostgresInsertMode;
 import io.debezium.connector.jdbc.junit.jupiter.PostgresSinkDatabaseContextProvider;
 import io.debezium.connector.jdbc.junit.jupiter.Sink;
+import io.debezium.connector.jdbc.junit.jupiter.SinkRecordFactoryArgumentsProvider;
 import io.debezium.connector.jdbc.util.SinkRecordFactory;
+import io.debezium.data.Enum;
 import io.debezium.data.Json;
 import io.debezium.data.Uuid;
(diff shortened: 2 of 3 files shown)


debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/EnumType.java
```diff
@@ -30,6 +31,25 @@ public String[] getRegistrationKeys() {
         return new String[]{ Enum.LOGICAL_NAME };
     }
 
+    @Override
+    public String getQueryBinding(ColumnDescriptor column, Schema schema, Object value) {
+        // PostgreSQL does not implicitly cast a bound character varying value to an enum column, so the
+        // cast must be explicit.
+        return "cast(? as " + quoteTypeName(column.getTypeName()) + ")";
```
Naros:
```suggestion
        return "cast(? as %s)".formatted(quoteTypeName(column.getTypeName()));
```

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/EnumType.java
```diff
@@ -30,6 +31,25 @@ public String[] getRegistrationKeys() {
         return new String[]{ Enum.LOGICAL_NAME };
     }
 
+    @Override
+    public String getQueryBinding(ColumnDescriptor column, Schema schema, Object value) {
+        // PostgreSQL does not implicitly cast a bound character varying value to an enum column, so the
+        // cast must be explicit.
```
Naros:
```suggestion
```
Comment's unnecessary.

debezium-connector-jdbc/src/test/java/io/debezium/connector/jdbc/integration/postgres/JdbcSinkColumnTypeMappingIT.java
```diff
@@ -830,4 +832,86 @@ private void assertNativeArrayRoundTrip(SinkRecordFactory factory, PostgresInser
...
+
+        getSink().assertColumn(destinationTable, "data", enumType);
+        getSink().assertRows(destinationTable, rs -> {
+            assertThat(rs.getString(2)).isEqualTo("assistant");
+            return null;
+        });
+    }
+
+    @ParameterizedTest
+    @ArgumentsSource(SinkRecordFactoryArgumentsProvider.class)
```
Naros: Should this use `PostgresInsertModeArgumentsProvider`?

sundongkim-dev: Done. With UNNEST enabled the two-record batch reaches the batch path; with it disabled the same records still go row-wise, so both are worth covering.

Naros (approved):

Naros:

PR #7759, opened 2026-08-04
debezium/dbz#2355 Document UNNEST batch inserts as applicable to CockroachDB

viragtripathi:
The JDBC sink option `dialect.postgres.unnest.insert.enabled` is documented as PostgreSQL only, but the CockroachDB dialect inherits the UNNEST batch write path and the generated statements run on CockroachDB (validated against CockroachDB v25.4.13 with the 3.7.0.Alpha1 sink, details in the issue).

Updates the property description to include CockroachDB, adds a link to the CockroachDB multi-row statement guidance alongside the existing PostgreSQL reference, and documents that batches which can repeat a primary key require `use.reduction.buffer=true` (https://github.com/debezium/dbz/issues/2356).

One known limitation on CockroachDB targets, found during validation and filed separately: tables with binary columns currently fail in the UNNEST path because the dialect DDL type alias `bytes` is not accepted by `createArrayOf` (https://github.com/debezium/dbz/issues/2357).

Closes https://github.com/debezium/dbz/issues/2355

PR #7756, opened 2026-08-04
debezium/dbz#2235 Align JDBC TINYINT expectations with INT8 mapping

minleejae:
Related to debezium/dbz#2235
Follow-up to #7669

## Description

Updates the JDBC end-to-end test expectations after signed MySQL `TINYINT`
columns were changed to emit INT8 schemas.

MySQL `TINYINT` columns now expect the target INT8 type regardless of column
type propagation. SQL Server `TINYINT` retains its existing INT16 expectation
because it represents an unsigned 8-bit value.

## Testing

* `./mvnw process-sources -pl debezium-connector-jdbc -am`
* `./mvnw install -pl debezium-connector-jdbc -am -Dquick`
* `./mvnw -pl debezium-connector-jdbc -DskipTests -DskipITs verify`

The targeted JDBC MySQL end-to-end tests were not run locally because Docker
Desktop had no remaining storage space. They are covered by the JDBC MySQL CI
jobs.

## PR Checklist

- [x] I have read the [contribution guidelines](https://github.com/debezium/debezium/blob/main/CONTRIBUTING.md) and the [governance document](https://github.com/debezium/governance/blob/main/GOVERNANCE.md) on PR expectations.
- [x] Minimal changes to code not directly related to your change (e.g. no unnecessary formatting changes or refactoring to existing code)
- [x] One feature/change per PR unless tightly coupled
- [x] Do a rebase on upstream `main`

PR #7763, opened 2026-08-05
debezium/dbz#2357 Fall back to row-wise writes for UNNEST batches with binary fields

viragtripathi:
With dialect.postgres.unnest.insert.enabled=true, a multi-record batch containing a Connect BYTES field fails and kills the task. On CockroachDB targets the dialect DDL alias 'bytes' is rejected as an array element type name; renaming it only moves the failure into the driver, because the PostgreSQL JDBC driver cannot encode bytea array elements in createArrayOf (UnsupportedOperationException: byte[] nested inside Object[]). Binary fields therefore cannot go through the UNNEST bind path on any target. Single-record flushes already take the per-row path, which makes the failure intermittent under light traffic.

Extends the existing non-bindable field check, which already routes STRUCT fields to the row-wise path, to also cover BYTES fields.

Verification: new integration tests against PostgreSQL and CockroachDB replay a multi-record batch with a BYTES column; both fail without the change (each with its respective error) and pass with it. The full unit suite and the it-postgresql and it-cockroachdb integration suites pass.

Closes https://github.com/debezium/dbz/issues/2357

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresDatabaseDialect.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresDatabaseDialect.java
@@ -182,10 +182,14 @@ public class PostgresDatabaseDialect extends GeneralDatabaseDialect {
      * values to {@link Connection#createArrayOf}, which only accepts scalar elements; a {@code STRUCT}
      * field (the geometric types point/box/lseg/path/polygon and circle/line) needs the per-value
      * {@code JdbcType#bind()} the row-wise path applies, so such records are handled row-wise instead.
+     * {@code BYTES} fields are also excluded: the PostgreSQL driver cannot encode {@code bytea} array
+     * elements ({@code createArrayOf} rejects {@code byte[]} nested inside {@code Object[]}), so such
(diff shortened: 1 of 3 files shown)


viragtripathi:
The failing JDBC - it-mysql,e2e-mysql job fails on JdbcSinkPipelineToMySqlIT TINYINT column expectations (testTinyIntDataType, testTinyIntWithSizeDataType), which is the pre-existing issue tracked by https://github.com/debezium/dbz/issues/2235 with the fix pending in #7756 (same AbstractJdbcSinkPipelineIT expectations). This change only affects the PostgreSQL dialect's UNNEST batch path behind dialect.postgres.unnest.insert.enabled; the it-postgresql and it-cockroachdb jobs that exercise it pass.

viragtripathi:
Hi @Naros , since the [dbz#2235](https://github.com/debezium/dbz/issues/2235) TINYINT/INT8 change merged, the JDBC it-mysql/e2e-mysql job fails its old TINYINT expectations on every JDBC PR, including two of mine (#7759 docs, #7763 UNNEST bytes fallback). minleejae's #7756 aligns the expectations and looks clean. Could you take a look when you get a chance? Happy to rebase mine after it lands.

Naros (approved):

Naros:

PR #7764, opened 2026-08-05
DBZ-2353 Expose transaction commit LSN on PostgreSQL change and transaction-metadata events

aloksingh24:
What

Adds the transaction commit LSN (pgoutput Begin.final_lsn) to PostgreSQL connector output:

source.commit_lsn on data change events
commit_lsn on transaction-metadata BEGIN/END values and on the transaction block of data events
Why

source.lsn is the per-row WAL position and is not monotonic across overlapping/long-running transactions, so it can't be used as an ordering or idempotency watermark. The commit LSN is identical for every event in a transaction and strictly increasing across the emitted stream (transactions are decoded in commit order), which makes it a correct single-scalar watermark.

How

Begin.final_lsn is decoded in PgOutputMessageDecoder, carried on TransactionMessage, stashed on PostgresOffsetContext at BEGIN and cleared at COMMIT, then stamped onto SourceInfo (source.commit_lsn) for data events and onto the transaction structs via PostgresTransactionMonitor. Schema field is OPTIONAL_INT64 (null for snapshots); transaction value/block schema versions bumped 1 → 2.

Compatibility

Additive optional field; existing consumers unaffected. New field is null for snapshot records.

Testing

Validated end-to-end in unified-adpt-service (Intuit) running the equivalent 3.0.4 backport in preprod: 632 transaction END events showed commitLsnMatch=true, no null commit_lsn, commit_lsn strictly increasing in commit order, endIdLsn − commit_lsn a constant small positive commit-record size (48/72 bytes). (Upstream unit/IT to be added — see checklist.)

Files (10, +70/-6)

core (debezium-connector-common): SchemaFactory, TransactionStructMaker, TransactionMonitor postgres: SourceInfo, PostgresSourceInfoStructMaker, PostgresOffsetContext, PostgresStreamingChangeEventSource, connection/TransactionMessage, connection/pgoutput/PgOutputMessageDecoder, pipeline/txmetadata/PostgresTransactionMonitor


vjuranek:
Thanks for ht PR @aloksingh24 , could you please take a look on `TransactionMetadataIT.transactionMetadataContainsCommitLsn`? This newly added test fails. Thanks.

aloksingh24:
@vjuranek Updated PR , please review

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/pipeline/txmetadata/PostgresTransactionStructMaker.java
```diff
@@ -0,0 +1,67 @@
...
+
+    /**
+     * Returns a copy of {@code base} with the optional {@code commit_lsn} field appended. The
+     * schema name is preserved and the version incremented by one, since adding a field is a
+     * schema change.
+     */
+    private static Schema withCommitLsn(Schema base) {
+        final SchemaBuilder builder = SchemaBuilder.struct().name(base.name());
+        if (base.version() != null) {
+            builder.version(base.version() + 1);
```
vjuranek: This doesn't seem to me correct - if we bump the version in the future, we bump the version here as well.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
```diff
@@ -1570,6 +1573,17 @@ protected SourceInfoStructMaker<? extends AbstractSourceInfo> getSourceInfoStruc
         return getSourceInfoStructMaker(SOURCE_INFO_STRUCT_MAKER, Module.name(), Module.version(), this);
     }
 
+    @Override
+    public TransactionMetadataFactory getTransactionMetadataFactory() {
+        // Default Postgres to a factory that adds the optional commit_lsn field to its transaction
+        // metadata, while still honouring an explicitly configured non-default factory.
+        final String configured = getConfig().getString(TRANSACTION_METADATA_FACTORY);
+        if (configured == null || DefaultTransactionMetadataFactory.class.getName().equals(configured)) {
```
vjuranek: So if the user explicitly configures `DefaultTransactionMetadataFactory`, Debezium would use `PostgresTransactionMetadataFactory`? I don't think this is correct.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/pipeline/txmetadata/PostgresTransactionStructMaker.java
aloksingh24: I've decoupled it — the Postgres transaction schemas now use an explicit POSTGRES_TRANSACTION_SCHEMA_VERSION constant instead of deriving from the core schema's version, so a future bump of the core transaction schema won't implicitly change this one.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/PostgresConnectorConfig.java
aloksingh24: You're right, that would override an explicitly configured DefaultTransactionMetadataFactory. I've reworked it to override the field's default instead: PostgreSQL now defaults transaction.metadata.factory to PostgresTransactionMetadataFactory, but any explicitly configured factory (including the default one) is read from the same property and honored.

aloksingh24:
@vjuranek Can you review the PR

vjuranek (approved):
The code LGTM, could you please rebase @aloksingh24 and also update Postgres documentation?

gunnarmorling:
This seems related to https://github.com/debezium/dbz/issues/1555? I think we already have the fields in the schema, it's just that we populate them with incorrect values?

aloksingh24:
@vjuranek Rebased onto latest main and updated the Postgres docs. All checks are green and there are no conflicts

aloksingh24:
@gunnarmorling Yes, same underlying mechanism as debezium/dbz#1555 — both use the BEGIN message's final_lsn (commit LSN). The difference is intentional: this PR adds a new optional commit_lsn field rather than changing the existing transaction.id/source.sequence values, since correcting those is a breaking change that @jpechane flagged in #1555 as needing more thought. Note source.lsn is the per-row event LSN, so there wasn't an existing commit-LSN field — this fills that gap without breaking anyone. Related to #1555, but scoped to the non-breaking addition

Previous discussion on the community: https://debezium.zulipchat.com/#narrow/channel/348249-community-postgresql/topic/Expose.20pgoutput.20Begin.2Efinal_lsn.20.28transaction.20commit.20LSN.29

vjuranek:
Thanks @aloksingh24 . LGTM, let's wait for CI and it's ready to go.

aloksingh24:
@vjuranek Are we good to merge? please let me know if anything is pending from my side

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/pipeline/txmetadata/PostgresTransactionStructMaker.java
```diff
@@ -0,0 +1,70 @@
...
+    }
+
+    /**
+     * Returns a copy of {@code base} with the optional {@code commit_lsn} field appended. The
+     * schema name is preserved; the version is an explicit Postgres-owned constant so it does not
+     * track the core transaction schema version.
+     */
+    private static Schema withCommitLsn(Schema base) {
+        final SchemaBuilder builder = SchemaBuilder.struct()
+                .name(base.name())
```
Naros: The 
```suggestion
        final SchemaBuilder builder = SchemaBuilder.struct()
        		.optional()
                .name(base.name())
```
The original code drops the `Struct`'s default optionality attribute.

I honestly wonder however, if the logic in this method belongs inside `PostgresSchemaFactory` that overrides `transactionBlockSchema`?

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/pipeline/txmetadata/PostgresTransactionStructMakerTest.java
```diff
@@ -0,0 +1,59 @@
...
+ * Verifies that {@link PostgresTransactionStructMaker} adds the optional {@code commit_lsn} field to
+ * the Postgres transaction metadata schemas, and that the shared/core schema is left untouched.
+ */
+public class PostgresTransactionStructMakerTest {
+
+    private final DefaultTransactionStructMaker defaultMaker = new DefaultTransactionStructMaker(Configuration.empty());
+    private final PostgresTransactionStructMaker postgresMaker = new PostgresTransactionStructMaker(Configuration.empty());
+
+    @Test
+    void transactionValueSchemaAddsOptionalCommitLsn() {
```
Naros, line 30: These tests should also validate that the output schema includes the expected parent field types, not just that the commit LSN field was added, to ensure complete coverage.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/pipeline/txmetadata/PostgresTransactionStructMakerTest.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Verifies that {@link PostgresTransactionStructMaker} adds the optional {@code commit_lsn} field to
+ * the Postgres transaction metadata schemas, and that the shared/core schema is left untouched.
+ */
+public class PostgresTransactionStructMakerTest {
+
+    private final DefaultTransactionStructMaker defaultMaker = new DefaultTransactionStructMaker(Configuration.empty());
+    private final PostgresTransactionStructMaker postgresMaker = new PostgresTransactionStructMaker(Configuration.empty());
+
+    @Test
```
Naros, line 28: All new tests should have `@FixFor("debezium/dbz#2353")` in them.

jenkins-jobs/scripts/config/Aliases.txt
```diff
@@ -397,6 +397,8 @@ prathit06,Malik Prathit
 LucjanoC,Luciano Capo
 Chaitanya-030,Chaitanya Sheth
 pugarte7,Pablo Ugarte
+aloksingh24,Alok Singh
+MarBruhn,Marcel Bruhn
```
Naros, line 401:
```suggestion
aloksingh24,Alok Singh
```
The second name already exists.

Naros (changes requested):
Hi @aloksingh24 a few inline issues, if you could take a look.

aloksingh24:
@Naros Updated as per comment , please review

gunnarmorling:
> The difference is intentional: this PR adds a new optional commit_lsn field rather than changing the existing transaction.id/source.sequence values, since correcting those is a breaking change that @jpechane flagged in https://github.com/debezium/debezium/pull/1555 as needing more thought.

It may be breaking, but so is any bug fix potentially. The existing behavior doesn't really make sense, and we should get it fixed rather than leaving the broken behavior and adding something new at the side.

> Note source.lsn is the per-row event LSN, so there wasn't an existing commit-LSN field — this fills that gap without breaking anyone.

The commit LSN ought to be part of the per event sequence field.

Overall, I'm -1 on the proposed change as-is. There's a mechanism for this already. We should fix it instead of inventing something new.

aloksingh24:
@gunnarmorling 
To add context on "buggy": both are currently documented as intentional — source.sequence is defined as "first value is the last committed LSN, second is the current LSN," and @jpechane  described that as correct for ordering in #1555; transaction.id is documented as txID:LSN where the LSN is "the LSN of the given operation." I don't disagree that a stable, commit-ordered value is more useful — I just want to flag that changing these is a documented, breaking semantic change, which is what steered me toward an additive field no existing consumer relies on. 

Since @Naros approved the additive approach here while @jpechane  was cautious about the transaction.id format change in #1555, could we get a maintainer decision on whether to make the breaking change? I'm happy to implement either — the BEGIN final_lsn plumbing in this PR supports fixing source.sequence/transaction.id just as well as adding the field.

Naros:
For `source.sequence`, my question for the group is: how literally should we take "last committed LSN"?

Debezium doesn't receive events until it observes the COMMIT, so I'm inclined to say the first element in the JSON array should be the event's enclosing transaction commit LSN, which aligns with @gunnarmorling's suggestion. If I look at other connectors like Oracle, where you need to pair the event and commit sequence numbers, logically making this change seems like the most reasonable approach.

Is it a breaking change? That depends on how the value is interpreted. Given how the documentation is worded, and because the LSN value for the first element would move forward rather than backward, the only compatibility breakage I see is if someone takes the first element and compares it with another field. And honestly, since we don't get into that level of detail, and the original intent was a commit-order watermark, I think we have an easy out using @gunnarmorling's suggestion.

Looking closer at the `transaction.id`:

* [DBZ-1052](https://issues.redhat.com/browse/DBZ-1052) introduced the field as simply the `txId`, a simple numeric string.
* [DBZ-5329](https://issues.redhat.com/browse/DBZ-5329) introduced the LSN component, simply saying "LSN of given operation"

In DBZ-5329, we originally wanted to use commitLsn, but at the time this information wasn't available in the BEGIN event, so we had no choice but to use eventLsn as an alternative. The goal of this change was to address the issue that transaction identifiers wrap, and without an additional qualifier, uniqueness was an issue.

So from my PoV, I can easily see Gunnar's PoV for both. @jpechane do you arrive at a similar conclusion?

aloksingh24:
@Naros @gunnarmorling @jpechane Still waiting on decision how we can move ahead to fix the issue .

aloksingh24:
@Naros @gunnarmorling @jpechane Gentle reminder , can we conclude on this . This is blocker for one of our prod feature to go live

aloksingh24:
@Naros @gunnarmorling @jpechane Gentle reminder

atorik:
I agree with @gunnarmorling ’s point:
> The existing behavior doesn't really make sense, and we should get it fixed rather than leaving the broken behavior and adding something new at the side.

However, I do have a concern about backward compatibility.

Looking at the [commit that introduced source.sequence](https://github.com/debezium/debezium/commit/06b0475f172155517908c256f579897783c60a95), it was explicitly intended for consumer-side deduplication.
For example, suppose a consumer has already processed a record with sequence [100, 999]. IIUC, if the same record is replayed after upgrading and now has sequence [101, 999], the consumer could treat it as a new record and process it again.

So I think moving the first element forward could itself break deduplication across an upgrade, even without comparing it with another field.

The impact depends on how consumers use `source.sequence`, but I think we should at least highlight this scenario in the release notes if this concern is valid.

aloksingh24:
@Naros Can you please help here ? can we conclude on the approach , based on that I can update the PR

PR #7765, opened 2026-08-05
debezium/dbz#2361 Deploy snapshots in GitHub Actions

jpechane:
Fixes https://github.com/debezium/dbz/issues/2361
https://github.com/debezium/debezium/actions/runs/31002901943/job/92295808613

### DIFF
--- a/.github/workflows/deploy-snapshots.yml
+++ b/.github/workflows/deploy-snapshots.yml
@@ -1,0 +1,424 @@
+name: Deploy Debezium Snapshots
+
+on:
+  schedule:
+    - cron: '0 */12 * * *'
+  workflow_dispatch:
+    inputs:
+      debezium-repositories:
+        description: Multiline list of id#repo[#subDir]#branch entries. Include the main Debezium repository as a normal entry, for example core#debezium/debezium#main.
+        required: false
+        default: |
+          core#debezium/debezium#main
+          db2#debezium/debezium-connector-db2#main
+          vitess#debezium/debezium-connector-vitess#main
+          cassandra#debezium/debezium-connector-cassandra#main
+          spanner#debezium/debezium-connector-spanner#main
+          informix#debezium/debezium-connector-informix#main
+          ibmi#debezium/debezium-connector-ibmi#main
+          cockroachdb#debezium/debezium-connector-cockroachdb#main
+          quarkus#debezium/debezium-quarkus#main
+          server#debezium/debezium-server#main
+          operator#debezium/debezium-operator#main
+          platform#debezium/debezium-platform#debezium-platform-conductor#main
+          jbang-catalog#debezium/jbang-catalog#main
+        type: string
+      descriptors-repository:
+        description: Repository where Debezium descriptors are published
+        required: false
+        default: debezium/debezium-descriptors-registry
+        type: string
+      descriptors-branch:
+        description: Branch where descriptors are published
+        required: false
+        default: snapshot
+        type: string
+
+permissions:
+  contents: read
+
+concurrency:
+  group: deploy-snapshots-${{ github.event.inputs.descriptors-branch || 'snapshot' }}
+  cancel-in-progress: false
+
+jobs:
+  deploy_snapshots:
+    name: Build and deploy snapshots
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    env:
+      MAVEN_OPTS: -Xmx4096m -Xms512m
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy-snapshots.yml
```diff
@@ -0,0 +1,424 @@
...
+              output.write(f"core_subdir={core_repository['subdir']}\n")
+          PY
+
+      - name: Set up Java
+        uses: ./workflow-repo/.github/actions/setup-java
+
+      - name: Restore Maven cache
+        uses: ./workflow-repo/.github/actions/maven-cache
+        with:
+          key: maven-deploy-snapshots-${{ hashFiles('workflow-repo/**/pom.xml') }}
```
Naros, line 186: If I am not mistaken, this creates another ~2GB cache entry in the main repo's cache budget, right?

I wonder if we even need this, and if so, would it make sense to move this workflow to the `debezium/dbz` repository or another where cache budgets aren't as critical?

Naros (commented):
Hi @jpechane overall LGTM, I just have one comment on caches.

jpechane:
@mfvitale @Naros Tested and passing - https://github.com/debezium/debezium/actions/runs/31070148354/job/92516210459

.github/workflows/deploy-snapshots.yml
jpechane: As discuess in the chat. This is only temporary. With introduction of central push build cache it will be used instead of per-workflow cache.

.github/workflows/deploy-snapshots.yml
```diff
@@ -0,0 +1,424 @@
+name: Deploy Debezium Snapshots
+
+on:
+  schedule:
+    - cron: '0 */12 * * *'
```
mfvitale, line 5: Should we also permit to run it manually as for the image workflow?

.github/workflows/deploy-snapshots.yml
```diff
@@ -0,0 +1,424 @@
...
+        cassandra#debezium/debezium-connector-cassandra#main
+        spanner#debezium/debezium-connector-spanner#main
+        informix#debezium/debezium-connector-informix#main
+        ibmi#debezium/debezium-connector-ibmi#main
+        cockroachdb#debezium/debezium-connector-cockroachdb#main
+        quarkus#debezium/debezium-quarkus#main
+        server#debezium/debezium-server#main
+        operator#debezium/debezium-operator#main
+        platform#debezium/debezium-platform#debezium-platform-conductor#main
+        jbang-catalog#debezium/jbang-catalog#main
```
mfvitale, line 66: Why these are not get from inputs?

.github/workflows/deploy-snapshots.yml
jpechane: @mfvitale It is the `workflow_dispatch` below. Isn't it?

.github/workflows/deploy-snapshots.yml
```diff
@@ -0,0 +1,424 @@
...
+            -Dfile="${{ steps.resolve-xstreams.outputs.xstreams_dir }}/xstreams.jar"
+
+      - name: Build and deploy Debezium repositories
+        shell: bash
+        env:
+          REPOSITORIES_JSON: ${{ steps.parse-repositories.outputs.repositories_json }}
+          CORE_ID: ${{ steps.repository-metadata.outputs.core_id }}
+        run: |
+          set -euo pipefail
+          mkdir -p "${DESCRIPTORS_OUTPUT_DIR}"
```
mfvitale, line 269: Is it possible to put longer script like this externally?

mfvitale:
@jpechane Great! I left some comments but overall it LGTM

.github/workflows/deploy-snapshots.yml
mfvitale: Right, I missed it. Sorry.

.github/workflows/deploy-snapshots.yml
jpechane: @mfvitale They are - few lines below - if triggered by workflow_dispatch

mfvitale: Hum, then could this env be empty initially? It creates confusion seeing this "duplication". I suppose the scheduled run uses the default then?

jpechane: Scheudled run can't use default as that is defined at workflow_dispatch trigger level

PR #7771, opened 2026-08-06
debezium/dbz#1589 Minimize LogMiner stalls due to log inconsistency

Naros:
Fixes debezium/dbz#1589

## Description
The Oracle LogMiner implementation will periodically report that the logs are inconsistent for a redo thread when we issue a query and an online redo log is scheduled for archival, but it hasn't been archived yet. 

This PR aims to reduce the impact of inconsistency checks by attempting to cap the curated list of logs at the log sequence gap for `online_catalog` and `hybrid` mining strategies, allowing the connector to safely mine the changes without waiting and causing further backlog during high-traffic windows. Once the connector is mining real-time online redo logs, the inconsistency check will create hard stalls waiting for archival to complete.

For other mining strategies, this PR has no impact, and the existing inconsistency check behavior remains unchanged. This is because data dictionary entries are appended to the write position of the redo logs when a dictionary build operation is executed. LogMiner needs that information on the next mining pass; therefore, all logs up to where the dictionary was most recently written are necessary.

## PR Checklist

- [x] I have read the [contribution guidelines](https://github.com/debezium/debezium/blob/main/CONTRIBUTING.md) and the [governance document](https://github.com/debezium/governance/blob/main/GOVERNANCE.md) on PR expectations.
- [x] Minimal changes to code not directly related to your change (e.g. no unnecessary formatting changes or refactoring to existing code)
- [x] One feature/change per PR unless tightly coupled
- [x] Do a rebase on upstream `main`

PR #7772, opened 2026-08-06
debezium/dbz#2326 Do not allow mining upper boundary to go backward

### LINKED ISSUE #2326: Oracle log-count mining: unbounded online-mode sessions under sustained load; cap-growth heuristic inflates on slow passes
Companion report to dbz#2315, observed on the same environment with DEBUG logging on
`LogFileCollector` and `CappedLogFileSessionSelector`.

## Bug report

Oracle connector (LogMiner adapter), Debezium Server 3.6.0.Final.

---

```properties
debezium.source.database.connection.adapter=logminer
debezium.source.log.mining.strategy=hybrid
debezium.source.log.mining.query.filter.mode=in
debezium.source.log.mining.log.count.min=1
debezium.source.log.mining.archive.destination.name=LOG_ARCHIVE_DEST_1
debezium.source.snapshot.mode=no_data
```

---

Oracle 19c, 2-node RAC on Exadata, on-premises. Online redo: 6 groups × 17 GB per
thread; `ARCHIVE_LAG_TARGET=1200`. Peak redo generation ~90 GB/h (morning business
window), ~1.1 TB/day total.

---

When mining falls behind the redo head under sustained load, I would expect the mining
session workload to stay bounded (comparable to the capped/archive mode governed by
`log.mining.log.count.min`), so that events and heartbeats keep flowing incrementally
and end-to-end latency degrades gracefully.

I would also expect the "capped log set unchanged → grow the per-thread log count"
heuristic to trigger only when the lower watermark is genuinely pinned by a
long-running transaction, not merely because mining passes are slower than log
rotation.

---

**What behavior do you see?**

During a peak-load window (06:15–08:45 UTC), three interacting behaviors made latency
grow from ~2 minutes to ~50 minutes:

1. **Unbounded online-mode sessions.** While the selector decides "all threads are
   reading online redo", no upper boundary cap is applied, so each session spans
   read-position → head. Once mining rate dropped slightly below generation rate, each
   session covered more redo than the previous one and session durations compounded:

   | Session start (UTC) | Gap since previous session |
   |---|---|
   | 07:10 → 07:14 | 4.5 min |
   | 07:14 → 07:21 | 6 min |
   | 07:21 → 07:29 | 8 min |
   | 07:29 → 07:40 | 11 min |
   | 07:40 → 07:57 | 17 min |
   | 07:57 → 08:22 | 25 min |
   | 08:22 → 08:45 | 23 min |

   A later single session lasted ~50 minutes (heartbeat gap 08:22–09:12 UTC). Observed
   end-to-end latency tracked session duration (measured 25 min, then ~50 min). Since
   events and heartbeats stall while a session runs, latency ≈ session length.

2. **Cap-growth heuristic misfires when mining is merely slow.** The
   "Capped log set unchanged, growing log count per redo thread to N" path incremented
   the count from 1 up to 23 during the window, even though no long-running transaction
   was pinning the watermark — the set was unchanged simply because passes progressed
   slowly through the same large archive. The inflated count then keeps the CURRENT
   logs inside the capped set, which keeps selecting the uncapped online path — i.e.,
   the heuristic reinforces behavior (1) exactly when a bound would help most.

3. **Regressive upper boundary after count reset.** After "Capped log set changed,
   resetting log count per redo thread to 1", the next capped selection produced an
   upper boundary *below* boundaries already mined in earlier online passes — the same
   value three times over an hour:

   ```
   07:21:06 Using capped logs, reading up to 13809995057971.
   07:29:16 All threads are reading online redo, ... reading up to 13809999373425.
   07:40:00 Using capped logs, reading up to 13809995057971.   <-- regressed below 07:29 boundary
   07:57:37 All threads are reading online redo, ... reading up to 13810012077399.
   08:22:04 Using capped logs, reading up to 13809995057971.   <-- again
   ```

   These look like no-op or near-no-op sessions, adding session start/stop overhead
   under load.

The system recovered on its own once peak generation subsided; the behavior reproduces
on every sustained peak.

---

3.6.0.Final is the latest Final release at the time of reporting.

---

Yes — DEBUG captures of `LogFileCollector`/`CappedLogFileSessionSelector` spanning the
whole window (excerpts above). Full sanitized captures available on request.

---

Requires generation rate exceeding mining rate for a sustained period: large redo logs,
steady high-volume DML (in our case a reporting batch generating ~20M row operations in
the window), and DEBUG on the two classes above. The compounding session durations and
the count growth to double digits then reproduce deterministically.

---

Related: dbz#2315 (ACTIVE online log not mined at log switch — same environment). Happy
to work on a PR for either issue with maintainer guidance, and to test candidate fixes
against this RAC environment.

Naros:
Fixes debezium/dbz#2326

## Description
When using the new capped log mining strategy, users may notice that log switches can temporarily cause the upper boundary to go backward. In some instances this isn't an issue, but particularly in long-running transactions, this is not desired because the next read could re-read what's already been previously read, and the window will need to expand once more.

To address this, the upper boundary is now sticky and cannot go backward.

## PR Checklist

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/oracle.adoc
+++ b/documentation/modules/ROOT/pages/connectors/oracle.adoc
@@ -4842,7 +4842,7 @@ This configuration property has no effect when using xref:#oracle-property-log-m
 |[[oracle-property-log-mining-log-count-growth-max]]<<oracle-property-log-mining-log-count-growth-max, `+log.mining.log.count.growth.max+`>>
 |`4`
 |The maximum number of transaction logs per redo thread that the mining window grows to automatically when a long-running transaction prevents the window start from advancing.
-Worst-case work in a single mining step is approximately this count multiplied by the online redo log size.
+The growth budget contributes at most approximately this count multiplied by the online redo log size to a single mining step.
 
 A xref:#oracle-property-log-mining-log-count-min[`log.mining.log.count.min`] value above this maximum takes precedence.
 The mining window can exceed this count when the connector re-reads previously mined logs, such as after a restart or while a transaction remains in progress; the value bounds automatic growth, not the window itself.
--- a/documentation/modules/ROOT/pages/connectors/oracle.adoc
+++ b/documentation/modules/ROOT/pages/connectors/oracle.adoc
@@ -4837,17 +4837,21 @@ The schema and table configuration include/exclude lists can safely specify regu
 |`2`
 |The minimum number of transaction logs to read per redo thread in each mining step.
 
-This configuration property has no effect when using xref:#oracle-property-log-mining-strategy[`log.mining.strategy`] is set to `redo_log_catalog`.
+This property does not apply if `log.mining.strategy` is set to `redo_log_catalog`.
 
 |[[oracle-property-log-mining-log-count-growth-max]]<<oracle-property-log-mining-log-count-growth-max, `+log.mining.log.count.growth.max+`>>
 |`4`
-|The maximum number of transaction logs per redo thread that the mining window grows to automatically when a long-running transaction prevents the window start from advancing.
-The growth budget contributes at most approximately this count multiplied by the online redo log size to a single mining step.
+|Specifies the maximum number of redo logs per redo thread that the connector can add to the mining window when a long-running transaction prevents the window from advancing.
 
-A xref:#oracle-property-log-mining-log-count-min[`log.mining.log.count.min`] value above this maximum takes precedence.
-The mining window can exceed this count when the connector re-reads previously mined logs, such as after a restart or while a transaction remains in progress; the value bounds automatic growth, not the window itself.
+During a mining cycle, the connector can expand the window to a size that is approximately equal to the configured number of logs multiplied by the size of each online redo log.
 
-This configuration property has no effect when using xref:#oracle-property-log-mining-strategy[`log.mining.strategy`] is set to `redo_log_catalog`.
+If you set a value that is less than the value of the xref:oracle-property-log-mining-log-count-min[`log.mining.log.count.min`] property, the connector applies the larger value from that property to this property.
+That is, the connector effectively raises this maximum to satisfy that minimum.
+
+The specified limit applies to automatic mining window growth, not to the absolute count of the logs per thread that can populate the mining window.
+When the connector re-reads previously mined logs, as can occur after a restart, or while a transaction remains in progress, the count can exceed this value.
+
+This property does not apply if `log.mining.strategy` is set to `redo_log_catalog`.
 
 |[[oracle-property-log-mining-buffer-type]]<<oracle-property-log-mining-buffer-type, `+log.mining.buffer.type+`>>
 |`memory`
--- a/documentation/modules/ROOT/pages/connectors/oracle.adoc
+++ b/documentation/modules/ROOT/pages/connectors/oracle.adoc
@@ -4839,6 +4839,16 @@ The schema and table configuration include/exclude lists can safely specify regu
 
 This configuration property has no effect when using xref:#oracle-property-log-mining-strategy[`log.mining.strategy`] is set to `redo_log_catalog`.
 
+|[[oracle-property-log-mining-log-count-growth-max]]<<oracle-property-log-mining-log-count-growth-max, `+log.mining.log.count.growth.max+`>>
+|`4`
+|The maximum number of transaction logs per redo thread that the mining window grows to automatically when a long-running transaction prevents the window start from advancing.
+Worst-case work in a single mining step is approximately this count multiplied by the online redo log size.
+
+A xref:#oracle-property-log-mining-log-count-min[`log.mining.log.count.min`] value above this maximum takes precedence.
+The mining window can exceed this count when the connector re-reads previously mined logs, such as after a restart or while a transaction remains in progress; the value bounds automatic growth, not the window itself.
+
+This configuration property has no effect when using xref:#oracle-property-log-mining-strategy[`log.mining.strategy`] is set to `redo_log_catalog`.
+
 |[[oracle-property-log-mining-buffer-type]]<<oracle-property-log-mining-buffer-type, `+log.mining.buffer.type+`>>
 |`memory`
 |The buffer type controls how the connector manages buffering transaction data.
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleConnectorConfig.java
@@ -259,6 +259,18 @@ public class OracleConnectorConfig extends HistorizedRelationalDatabaseConnector
             .withDescription("Specifies the minimum number of logs to mine per redo thread. " +
                     "Setting this to 0 disables the cap, and all available logs are mined in a single pass.");
 
+    public static final Field LOG_MINING_LOG_COUNT_GROWTH_MAX = Field.create("log.mining.log.count.growth.max")
+            .withDisplayName("Maximum automatic growth of the log count per redo thread")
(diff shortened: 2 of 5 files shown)


denizozogularas:
Thanks for the quick turnaround @Naros. This directly addresses two of the behaviors from dbz#2326 as we observed them: the sticky boundary makes the regressed no-op sessions we logged (same effective upper boundary selected three times over an hour) impossible, and reading through to the online logs should speed up the final stretch of backlog drains. The count-growth heuristic can still inflate, but with the sticky boundary its reset no longer produces a harmful window, so from our side the practical impact is covered.

Once this merges we'll run the nightly build on our staging environment (2-node RAC, 17 GB redo logs, forced switches — the setup that reproduced all of this daily) and report results here. If a snapshot build is available earlier, happy to test that too.

Naros:
Hi @denizozogularas, Maven snapshot builds are published every ~12 hours, so once a change is merged, those changes should be available by the next working day if you want to use a snapshot build. Otherwise, you can always check out the `main` branch and build from source, too. 

In fact, if you could build this PR from source now and give that test feedback, that would be great to make sure that the change doesn't introduce any regressions or unforeseen behavior.

vjuranek (approved):
LGTM

Naros:
## Changes

* First commit: I backported the monotonic guard from PR #7771 to this PR. It should have been included earlier, but it was somehow overlooked when I reworked the commits.
* Second commit: targets the comparison logic and only focuses on the offset `scn` and is no longer concerned with drift applied to `commit_scn`, which more aptly aligns with the intended logged message.
* Third commit: changed `OffsetActivityMonitor` to be time-based. Right now it uses a hard-coded 10-minute timer. It will likely need to be configurable, but I'm just waiting on the clarifications on the associated issue.

Naros:
I dropped the `OffsetActivityMonitor` change as that is now part of PR #7805

Naros:
Hi @roldanbob, I could use a quick pass on the `oracle.adoc` changes, at your convenience.

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -4839,6 +4839,16 @@ The schema and table configuration include/exclude lists can safely specify regu
 
 This configuration property has no effect when using xref:#oracle-property-log-mining-strategy[`log.mining.strategy`] is set to `redo_log_catalog`.
 
+|[[oracle-property-log-mining-log-count-growth-max]]<<oracle-property-log-mining-log-count-growth-max, `+log.mining.log.count.growth.max+`>>
+|`4`
+|The maximum number of transaction logs per redo thread that the mining window grows to automatically when a long-running transaction prevents the window start from advancing.
```
```suggestion
|The maximum number of transaction logs per redo thread that the mining window can accommodate when a long-running transaction prevents the window start from advancing.
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -4839,6 +4839,16 @@ The schema and table configuration include/exclude lists can safely specify regu
...
+The growth budget contributes at most approximately this count multiplied by the online redo log size to a single mining step.
+
+A xref:#oracle-property-log-mining-log-count-min[`log.mining.log.count.min`] value above this maximum takes precedence.
+The mining window can exceed this count when the connector re-reads previously mined logs, such as after a restart or while a transaction remains in progress; the value bounds automatic growth, not the window itself.
```
```suggestion
|Specifies the maximum number of redo logs per redo thread that the connector can add to the mining window when a long-running transaction prevents the window from advancing.

During a mining cycle, the connector can expand the window to a size that is approximately equal to the configured number of logs multiplied by the size of each online redo log.

If you set a value that is less than the value of the xref:oracle-property-log-mining-log-count-min[`log.mining.log.count.min`] property, the connector applies the larger value from that property to this property.
That is, the connector effectively raises this maximum to satisfy that minimum.

The specified limit applies to automatic mining window growth, not to the absolute count of the logs per thread that can populate the mining window.
When the connector re-reads previously mined logs, as can occur after a restart, or while a transaction remains in progress, the count can exceed this value.
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -4839,6 +4839,16 @@ The schema and table configuration include/exclude lists can safely specify regu
...
+
+This configuration property has no effect when using xref:#oracle-property-log-mining-strategy[`log.mining.strategy`] is set to `redo_log_catalog`.
```
```suggestion
This property does not apply if log.mining.strategy is set to redo_log_catalog.
```

roldanbob (changes requested):
@Naros It took me several attempts, so that was less than quick, but I think that I understand this now. Let me know if you agree! ;)

documentation/modules/ROOT/pages/connectors/oracle.adoc
denizozogularas: Small catch: this suggestion inverts the meaning. The capped selector (and therefore this property) is active for the `hybrid` and `online_catalog` mining strategies and does **not** apply under `redo_log_catalog` — see `resolveLogFileSessionSelector` and the strategy guard in `validateLogMiningLogCountGrowthMax`. The original direction was right, just awkwardly phrased; maybe: "This property does not apply if `log.mining.strategy` is set to `redo_log_catalog`."

Naros: @roldanbob, I believe the last sentence in the property before this one shared the same verbiage, so I would guess we'd want to align both of them to the same verbiage?

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -4839,6 +4839,16 @@ The schema and table configuration include/exclude lists can safely specify regu
 
 This configuration property has no effect when using xref:#oracle-property-log-mining-strategy[`log.mining.strategy`] is set to `redo_log_catalog`.
```
```suggestion
This property does not apply if log.mining.strategy is set to redo_log_catalog.
```

documentation/modules/ROOT/pages/connectors/oracle.adoc
roldanbob: Oh, dang. Yes, of course. What I was trying to do was to express the _positive_ case, i.e., tell readers that the property applies only when a specific setting is in place, but I neglected to change that value, because I meant to look up the alternative setting. But, of course there's more than one alternative, so that approach doesn't work. :|
So I'll slink off and withdraw my comment and go with your suggestion, @denizozogularas!
And per the comment from @Naros, I agree, about implementing the same update to L4840.

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -4837,7 +4837,21 @@ The schema and table configuration include/exclude lists can safely specify regu
 |`2`
 |The minimum number of transaction logs to read per redo thread in each mining step.
 
-This configuration property has no effect when using xref:#oracle-property-log-mining-strategy[`log.mining.strategy`] is set to `redo_log_catalog`.
+This property does not apply if `log.mining.strategy` is set to `redo_log_catalog`.
```
roldanbob, line 4840:
```suggestion
```
Remove redundant sentence that's repeated on L4854. I think that placing that caveat at the end of the description makes the most sense.

roldanbob (changes requested):
@Naros One minor oversight to fix. Outside of that, I approve the content.

documentation/modules/ROOT/pages/connectors/oracle.adoc
Naros: Both properties, `log.mining.log.count.min` and the newly added `log.mining.log.count.growth.max` are only applicable for non-redo log catalog mining strategies, so I believe L4840 should remain. WDYT?

roldanbob: Oh, yes, of course. I mistakenly read the L4840 as part of the description of `log.mining.log.count.growth.max`. Thanks for the correction.

PR #7775, opened 2026-08-06
debezium/dbz#2364 Make OpenLineage test transport thread safe

vjuranek:
Signed-off-by: Vojtech Juranek <vjuranek@redhat.com>

Fixes debezium/dbz#2364


vjuranek:
As described in the issue, this is very likely a concurrency issue and unfortunately I wasn't able to reproduce it locally, so I'm not completely sure this will fix the issue. But in any case, making OpenLineage test transport thread safe shouldn't make it worse.

mfvitale (approved):

PR #7783, opened 2026-08-07
debezium/dbz#2376 Align streamed JSON values with the snapshot format

suhwan-cheon:
Fixes debezium/dbz#2376

## Description

Snapshots deliver `JSON` values as the server returns them over JDBC. Streaming parses the binary JSON from the binlog with the binlog client's `JsonStringFormatter` (`JsonBinary.parseAsString(byte[])`, called from `BinlogValueConverters#convertJson`), which formats them differently:

| Column value | Snapshot (JDBC) | Streaming (binlog) |
|---|---|---|
| `{"a": 2}` | `{"a": 2}` | `{"a":2}` |
| `CAST('2015-01-15 23:24:25' AS DATETIME)` | `"2015-01-15 23:24:25.000000"` | `"2015-01-15 23:24:25"` |
| `CAST('23:24:25.12' AS TIME(3))` | `"23:24:25.120000"` | `"23:24:25.12"` |
| `CAST(x'cafe' AS JSON)` | `"base64:type15:yv4="` | `"yv4="` |

See the issue for details.

## Changes

- **`DatabaseJsonStringFormatter`** (new): a `JsonStringFormatter` subclass that reproduces the server's output: a space after each key colon and entry comma, six fractional digits on temporal scalars, and the `base64:typeNN:` prefix on opaque scalars. Escaping and all other value conversion are inherited.
- **`json.string.formatting.mode`**: `legacy` (default, current behavior) or `database`. The option is defined on `BinlogConnectorConfig`, where `convertJson` also lives, so both connectors expose it.
- **`BinlogValueConverters#convertJson`** selects the formatter based on the mode. It keeps the text pass-through that `parseAsString` applies to values that are not binary JSON, so MariaDB, which stores `JSON` as `LONGTEXT` and replicates it as text, behaves the same in both modes.
- Documentation for the new property and a note on the `JSON` type mapping.

## Tests

- `BinlogValueConvertersTest`, which runs for both MySQL and MariaDB: hand-assembled binary JSON fixtures (object, array, nested document, datetime/time/opaque scalars, empty documents, escaped scalar) assert the output of both modes, and check that escaping and the MariaDB text pass-through are unaffected.
- `BinlogJsonIT` (MySQL): streams the full `json_test.sql` fixture with `json.string.formatting.mode=database` and asserts that every streamed value is byte-identical to what the server returns over JDBC for the same row. Run against MySQL 8.0 in a container.

## Backward compatibility

The default is `legacy`, so nothing changes for existing users.

## AI usage

I used Claude Code as an assistant on this change: to explore how the binlog client parses binary JSON, to draft the formatter and the tests, and to run the tests locally. I reviewed every line, checked the expected server output against MySQL 8.0 myself, and I take responsibility for the change.

## PR Checklist

### DIFF
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogValueConverters.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogValueConverters.java
@@ -45,6 +45,7 @@ import io.debezium.DebeziumException;
 import io.debezium.annotation.Immutable;
 import io.debezium.config.CommonConnectorConfig.BinaryHandlingMode;
 import io.debezium.config.CommonConnectorConfig.EventConvertingFailureHandlingMode;
+import io.debezium.connector.binlog.BinlogConnectorConfig.JsonStringFormattingMode;
 import io.debezium.connector.binlog.BinlogGeometry;
 import io.debezium.connector.binlog.BinlogUnsignedIntegerConverter;
 import io.debezium.connector.binlog.charset.BinlogCharsetRegistry;
@@ -117,6 +118,7 @@ public abstract class BinlogValueConverters extends JdbcValueConverters {
 
     private final EventConvertingFailureHandlingMode eventConvertingFailureHandlingMode;
     private final BinlogCharsetRegistry charsetRegistry;
+    private final JsonStringFormattingMode jsonStringFormattingMode;
     private final byte[] unavailableValuePlaceholderBinary;
     private final String unavailableValuePlaceholderString;
 
@@ -130,6 +132,8 @@ public abstract class BinlogValueConverters extends JdbcValueConverters {
      * @param binaryHandlingMode how binary columns should be treated
      * @param adjuster a temporal adjuster to make a database specific time before conversion
      * @param eventConvertingFailureHandlingMode how to handle conversion failures
+     * @param jsonStringFormattingMode how {@code JSON} values read from the binlog are serialized; may be null,
+     *            which is treated as {@link JsonStringFormattingMode#LEGACY}
      * @param serviceRegistry the service registry instance, should not be {@code null}
      * @param unavailableValuePlaceholder the placeholder bytes for unavailable column values; may be null
      */
@@ -139,10 +143,12 @@ public abstract class BinlogValueConverters extends JdbcValueConverters {
                                  BinaryHandlingMode binaryHandlingMode,
                                  TemporalAdjuster adjuster,
                                  EventConvertingFailureHandlingMode eventConvertingFailureHandlingMode,
+                                 JsonStringFormattingMode jsonStringFormattingMode,
                                  ServiceRegistry serviceRegistry,
                                  byte[] unavailableValuePlaceholder) {
         super(decimalMode, temporalPrecisionMode, ZoneOffset.UTC, adjuster, bigIntUnsignedMode, binaryHandlingMode);
         this.eventConvertingFailureHandlingMode = eventConvertingFailureHandlingMode;
+        this.jsonStringFormattingMode = jsonStringFormattingMode;
         this.charsetRegistry = serviceRegistry.getService(BinlogCharsetRegistry.class);
         this.unavailableValuePlaceholderBinary = unavailableValuePlaceholder;
         this.unavailableValuePlaceholderString = unavailableValuePlaceholder != null ? new String(unavailableValuePlaceholder) : null;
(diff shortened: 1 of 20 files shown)


debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogValueConverters.java
```diff
@@ -476,6 +482,25 @@ else if (data instanceof String) {
...
+
+    /**
+     * MariaDB stores {@code JSON} columns as {@code LONGTEXT} and replicates them as text, which
+     * {@link JsonBinary#parseAsString(byte[])} passes through unchanged; the same check is applied here so
+     * that text values do not reach the binary parser.
+     */
+    private static boolean isTextJson(byte[] data) {
+        // Binary JSON type markers are all <= 0x0f
+        return data[0] > 0x0f;
+    }
```
suhwan-cheon: For MariaDB

MySQL loads JSON columns into the binlog as binary JSON. The first byte is the type marker, and **all defined values ​​are 0x0f or less** (0x00 small object, 0x02 small array, 0x0c string, 0x0f opaque, etc.).

**MariaDB is different**: there is no JSON type, and since it is an alias for LONGTEXT, JSON columns are replicated as plain UTF-8 text. The first byte of {"a": 2} is { = 0x7B, which is greater than 0x0f.

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogJsonIT.java
```diff
@@ -114,6 +122,73 @@ public void shouldConsumeAllEventsFromDatabaseUsingStreaming() throws SQLExcepti
...
+                assertThat(i).isNotNull();
+                String json = after.getString("json");
+                String expectedJdbc = jdbcValues.get(i);
+                if (expectedJdbc != null && SCIENTIFIC_NOTATION.matcher(expectedJdbc).matches()) {
+                    // The binlog client renders doubles of large or small magnitude differently from the
+                    // server whatever the formatting mode; the only such value here is one the server
+                    // prints in scientific notation
+                    skipped.add(i);
+                    return;
+                }
```
suhwan-cheon: If Double value exceed 10^-15 ~ 10^14, serialized differently by the binlog client regardless of the mode: it expands them through new BigDecimal(double).toPlainString(), so 18446744073709551616 is emitted where the server returns 1.8446744073709552e19.

I think, this changes the numeric literal rather than the formatting
so it is out of scope, and i will rais it separately (if this issue accept & PR merged)

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/jdbc/DatabaseJsonStringFormatter.java
```diff
@@ -0,0 +1,52 @@
...
+        super.nextEntry();
+        appendString(" ");
+    }
+
+    @Override
+    public void valueDatetime(int year, int month, int day, int hour, int min, int sec, int microSeconds) {
+        // The base class omits a zero fractional part and trims trailing zeros; the server does neither
+        value(String.format(Locale.ROOT, "%s%04d-%02d-%02d %02d:%02d:%02d.%06d",
+                year < 0 ? "-" : "", Math.abs(year), month, day, hour, min, sec, microSeconds));
+    }
```
suhwan-cheon, line 39: String.format uses the default locale, which selects the digit characters.
But i found, if user locale is ar-EG(Arabic - Egypt), this renders 2015-01-15 as ٢٠١٥-٠١-١٥. (corrupt the emitted value.)
So set Locale.ROOT to locale-independent

Naros:
Hi @suhwan-cheon could we get a rebase?

suhwan-cheon:
> Hi @suhwan-cheon could we get a rebase?

@Naros 
Hi. Rebased on main!

Naros (commented):
Hi @suhwan-cheon so if I understand correctly, this JSON opt-in feature is designed to target MySQL, and yet its predominately defined in the `BinlogConnectorConfig` and inherited by both MySQL and MariaDB.

If there is no benefit to MariaDB here, I think it would be better to handle this as an explicit MySQL feature in MySQL's config class, value converter, and test suite, without involving the binlog module or MariaDB.

Would you concur?

If so, then I would recommend we rework the PR in that direction and for the changes in the docs, we qualify those to be conditionalized to MySQL, e.g.
```
ifdef::MYSQL[]
// The stuff here
endif::MYSQL[]
```

This prevents those sections from being rendered when the partial is rendered for the MariaDB connector.

suhwan-cheon:
@Naros 
Agree. MariaDB's `JSON` is a `LONGTEXT` alias replicated as text, so the option has no effect there.
so I Moved everything to the MySQL Connector and wrapped the docs in `ifdef::MYSQL[] ~ endif`

Naros (approved):
Great work @suhwan-cheon. This PR can be merged after CI finishes.

suhwan-cheon:
@Naros  can you approve the workflow run?

PR #7784, opened 2026-08-07
debezium/dbz#2381 Allow nonreserved keywords as unquoted identifiers in the MySQL DDL parser

### LINKED ISSUE #2381: MySQL DDL parser rejects unquoted identifier `url` on MySQL 8.0.32+ (URL_SYMBOL missing from identifier rules)
### Current behavior

The MySQL DDL parser (Oracle grammar, default since 3.6) rejects statements that use `url` as an unquoted table name when `serverVersion >= 80032`:

```sql
CREATE TABLE url (id INT);
-- mismatched input 'url' expecting {'.', ACCOUNT_SYMBOL, ACTION_SYMBOL, ADDDATE_SYMBOL, ...}

ALTER TABLE url ADD COLUMN c INT;
-- same failure
```

MySQL itself accepts these statements — `URL` is a nonreserved keyword, so it is valid as an unquoted identifier (verified against MySQL 8.0.46: `PREPARE` succeeds for both statements above).

### Root cause

`URL_SYMBOL` was introduced for the MySQL 8.0.32 `LOAD DATA ... URL` syntax and is gated in the lexer (`MySqlLexer.g4`):

```
URL_SYMBOL
    : U R L    {this.serverVersion >= 80032}?
```

but it was never added to `identifierKeywordsUnambiguous` in `MySqlParser.g4`. Its only parser usage is the `{this.isServerVersionGe80200()}? (URL_SYMBOL | S3_SYMBOL)` alternative. As a result, on any source at 8.0.32 or later, `url` lexes as a keyword token that no identifier rule accepts:

- unquoted table name `url` → parse error
- column named `url` and qualified `db.url` still parse (different rule paths)
- `S3_SYMBOL`, introduced by the same 8.0.32 gate, *is* usable as an identifier — only `URL` is missing

This is the same class of gap as the window-function keywords (`LAG`, `RANK`, ...) that were added to `identifierKeywordsUnambiguous` during the grammar migration. The upstream Oracle grammar in antlr/grammars-v4 has the identical omission — reported as [antlr/grammars-v4#4962](https://github.com/antlr/grammars-v4/issues/4962) with fix [antlr/grammars-v4#4963](https://github.com/antlr/grammars-v4/pull/4963).

### Impact

Tables literally named `url` exist in production databases (we hit this against a real 8.0.46 source: any DDL for such a table fails to parse, so schema history cannot be updated for it). The legacy grammar parses the same statements fine, so this is a regression relative to the pre-3.6 parser.

### Proposed fix

Add `URL_SYMBOL` to `identifierKeywordsUnambiguous` in `MySqlParser.g4`, mirroring MySQL's treatment of `URL` as a nonreserved keyword. PR follows.

minleejae:
Fixes debezium/dbz#2381

## Description

Nine nonreserved MySQL keywords cannot be used as unquoted identifiers, because their lexer tokens are missing from `identifierKeywordsUnambiguous`. Each token appears in the parser grammar only in the clause that introduced it, so a table, index or constraint with one of these names fails to parse:

```sql
CREATE TABLE url (id INT);           -- mismatched input 'url' expecting {'.', ACCOUNT_SYMBOL, ...}
CREATE TABLE vector (id INT);        -- no viable alternative at input 'vector'
CREATE TABLE auto (auto INT);        -- no viable alternative at input 'auto'
ALTER TABLE t1 ADD INDEX qualify (id);
```

Per review feedback, the fix is implemented in `DdlNormalizer` instead of `MySqlParser.g4`, keeping the grammar files aligned with upstream antlr/grammars-v4: the nine keywords (`URL`, `AUTO`, `MANUAL`, `OFFLINE`, `ONLINE`, `PARALLEL`, `QUALIFY`, `TABLESAMPLE`, `VECTOR`) are added to the reserved-keyword backticking that already handles the window function keywords (`RANK`, `LAG`, ...). The keyword list is defined once, split by provenance into `WINDOW_FUNCTION_KEYWORDS` (reserved words since MySQL 8.0, valid identifiers on pre-8.0 sources) and `VERSION_GATED_KEYWORDS` (the nine keywords above — removable as a group once the upstream fix is synced), combined into `KEYWORDS_REQUIRING_BACKTICKS` and shared by all context patterns. The identifier contexts were extended to cover the failing DDL shapes:

- table names after `CREATE/ALTER/DROP/RENAME/TRUNCATE TABLE` (including `TEMPORARY` and `IF [NOT] EXISTS` variants), `FROM`, `JOIN`, `INTO`, `TO`, `ON` (`CREATE INDEX/TRIGGER ... ON t`), `LIKE`, `REFERENCES`
- `ADD/CHANGE/MODIFY/ALTER [COLUMN]`, `DROP COLUMN`, `RENAME COLUMN` targets and `AFTER <col>` placement
- column definitions (`<keyword> <type>`), with the recognized type list extended (`TINYINT` family, `CHARACTER`/`NCHAR`/`NVARCHAR`, `SERIAL`, `DEC`/`FIXED`, `LONG`, spatial types, `VECTOR`, ...)
- index and constraint names (`[ADD] [UNIQUE|FULLTEXT|SPATIAL] INDEX/KEY`, `CREATE/DROP INDEX`, `[ADD] CONSTRAINT`, `DROP/ALTER CHECK`)
- other schema object names: `DATABASE`/`SCHEMA`/`VIEW`/`TRIGGER`/`PROCEDURE`/`FUNCTION`/`EVENT` (with `IF [NOT] EXISTS`), `USE`, and `PARTITION` names
- stored-routine labels

Every rewritten shape was validated against a real MySQL 8.4 server (statements executed with keyword identifiers) before being added to the patterns.

Once the upstream fix (antlr/grammars-v4#4963) lands and the grammar is synced, these entries can be removed together with the window-function ones.

Scope note: like the existing window-keyword handling, the normalizer only rewrites identifier positions in DDL clauses. It does not rewrite keywords inside expressions or index column lists (e.g. `CHECK (online > 0)`), which also keeps it from ever touching legitimate keyword usage — the `QUALIFY` query clause, the `VECTOR(n)` data type, `AUTO_INCREMENT`, or string literals such as `DEFAULT 'http://url:8080'`.

debezium/dbz#2381 was filed for `URL`; the other eight share the identical root cause and are handled here rather than split across separate PRs.

### How these were found and verified

The keyword tokens defined in `MySqlLexer.g4` were substituted into identifier positions (table name, column name, index name, constraint name, partition name, `ALTER TABLE` target, `RENAME ... TO` target), and every candidate statement was replayed against real MySQL servers before being compared against the parser.

Their status was then confirmed against the server's own keyword catalogue rather than inferred:

```sql
SELECT WORD, RESERVED FROM INFORMATION_SCHEMA.KEYWORDS WHERE WORD IN (...);
```

| keyword | lexer gate | MySQL 8.0.46 | MySQL 8.4.11 | MySQL 9.7.2 |
|---|---|---|---|---|
| `url` | `>= 80032` | keyword, nonreserved | keyword, nonreserved | keyword, nonreserved |
| `s3`, `bulk` *(already allowed)* | `>= 80032` / `>= 80200` | nonreserved | nonreserved | nonreserved |
| `auto`, `manual`, `parallel` | `>= 80200` | not a keyword | keyword, nonreserved | keyword, nonreserved |
| `vector` | none | not a keyword | not a keyword | keyword, nonreserved |
| `online`, `offline` | none | not a keyword | not a keyword | not a keyword |
| `qualify`, `tablesample` | `>= 80200` | not a keyword | **reserved** | **reserved** |

Nonreserved keywords are usable as unquoted identifiers by definition, so seven of the nine are valid identifiers on every MySQL release line the connector supports (8.0.x, 8.4.x, 9.0, 9.1).

Two entries stand out:

* `ONLINE` and `OFFLINE` are not MySQL keywords in any supported version — they appear to be left over from the MySQL 5.6 `ALTER TABLE ... ONLINE/OFFLINE` syntax. The grammar defines tokens for them with no version gate, so it rejects two perfectly ordinary words as identifiers on every server.
* `VECTOR` only became a keyword in MySQL 9.0, with the `VECTOR` data type, and it is nonreserved there too. Because its token also has no version gate, `CREATE TABLE vector (...)` fails even against 8.0 and 8.4 servers where the word has no special meaning at all.

### Why `QUALIFY` and `TABLESAMPLE` are included too

They are the only two of the nine that MySQL actually reserves, from 8.2 onwards, so on a modern server they are not valid identifiers — but on MySQL 8.0.x they are not keywords at all, and that release line is supported by the connector and currently breaks on them.

Being permissive here is safe: the DDL parser consumes statements from the binlog, which the source server has already executed. A statement that MySQL 8.4 rejects can never reach the parser from an 8.4 source, so accepting it cannot produce an incorrect table schema. The opposite error is not free — rejecting valid DDL stops the connector.

Version-gating the two entries would be the precise alternative, but it would have no effect today: `MySqlLexerBase`/`MySqlParserBase` set `serverVersion` to `80200` in their constructors and `MySqlAntlrDdlParser` never overrides it with the actual server version, so every source is parsed as if it were MySQL 8.2 regardless of what it really is. That hardcoding is worth addressing on its own; this PR does not attempt it.

Backticking `QUALIFY` only in identifier positions does not disturb the `QUALIFY` query clause — `SELECT id, ROW_NUMBER() OVER (ORDER BY id) rn FROM t1 QUALIFY rn = 1` passes through the normalizer unchanged, and a test case covering it is included.

## Testing

* `DdlNormalizerTest` — each of the nine keywords in every rewritten context, plus negative cases: `VECTOR(3)` as a data type, `AUTO_INCREMENT`, keyword text inside string literals and comments (`DEFAULT 'http://url:8080'`), and the `QUALIFY` query clause are left untouched
* `MySqlAntlrDdlParserTest#shouldParseNonReservedKeywordsAsUnquotedIdentifiers` — parses the statements above end-to-end through `MySqlAntlrDdlParser` (which applies the normalizer) and asserts no parsing errors and correct table/column registration, including `RENAME COLUMN`
* `debezium-ddl-parser` is untouched relative to `main`; the antlr4test corpus passes unchanged
* The keyword/server matrix above was verified against real MySQL 8.0.46, 8.4.11 and 9.7.2 via `INFORMATION_SCHEMA.KEYWORDS` and `PREPARE`

### DIFF
--- a/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/DdlNormalizer.java
+++ b/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/DdlNormalizer.java
@@ -18,7 +18,8 @@ import java.util.regex.Pattern;
  * text represents identifiers and must not be converted. In that case, callers
  * should pass {@code ansiQuotesMode=true} to skip the conversion.
  *
- * Also adds backticks around reserved keywords when used as identifiers.
+ * Also adds backticks around keywords that the grammar cannot parse as unquoted
+ * identifiers (see {@code KEYWORDS_REQUIRING_BACKTICKS}) when they are used as identifiers.
  *
  * @author Debezium Authors
  */
@@ -38,33 +39,128 @@ public class DdlNormalizer {
      * - Group 4: Backtick identifiers   {@code `identifier`}
      *
      */
-    private static final Pattern QUOTED_LITERAL_PATTERN = Pattern.compile(
-            "(--[^\\n]*|#[^\\n]*|/\\*(?!!).*?\\*/)|(\'(?:\'\'|[^\'\\\\]|\\\\.)*\')|\"((?:[^\"\\\\]|\\\\.|\"\")*)\"|(`(?:``|[^`])*`)",
+    private static final Pattern QUOTED_LITERAL_PATTERN = Pattern.compile("""
+            (--[^\\n]*|#[^\\n]*|/\\*(?!!).*?\\*/)\
+            |('(?:''|[^'\\\\]|\\\\.)*')\
+            |"((?:[^"\\\\]|\\\\.|"")*)"\
+            |(`(?:``|[^`])*`)""",
             Pattern.DOTALL);
 
     /**
-     * Reserved keywords that need backticks when used as identifiers.
+     * Window function keywords (plus JSON_TABLE): reserved words since MySQL 8.0, but valid
+     * unquoted identifiers on MySQL 5.7 and older, so DDL using them as identifiers can still
+     * reach the parser (older sources, schema history). Missing from
+     * {@code identifierKeywordsUnambiguous} in the migrated grammar.
      */
-    private static final Pattern RESERVED_KEYWORD_PATTERN = Pattern.compile(
-            "\\b(CUME_DIST|DENSE_RANK|FIRST_VALUE|JSON_TABLE|LAG|LAST_VALUE|LEAD|NTH_VALUE|NTILE|PERCENT_RANK|RANK|ROW_NUMBER)\\b",
+    private static final String WINDOW_FUNCTION_KEYWORDS = "CUME_DIST|DENSE_RANK|FIRST_VALUE|JSON_TABLE|LAG|LAST_VALUE|LEAD|NTH_VALUE|NTILE|PERCENT_RANK|RANK|ROW_NUMBER";
+
+    /**
+     * Version-gated keywords introduced by MySQL 8.0.32+ (URL) and 8.2.0+/9.x lexer gates:
+     * mostly nonreserved keywords (URL, AUTO, MANUAL, PARALLEL, VECTOR) or not keywords at all
+     * (ONLINE, OFFLINE) on real servers, so MySQL accepts them as unquoted identifiers — only
+     * QUALIFY and TABLESAMPLE are reserved, from 8.2 on. Their tokens are missing from
+     * {@code identifierKeywordsUnambiguous}, so the grammar rejects them in identifier
+     * positions (debezium/dbz#2381). The upstream Oracle grammar in antlr/grammars-v4 has the
+     * same omission; this group can be removed once the upstream fix (antlr/grammars-v4#4963)
+     * is synced into the Debezium grammar.
+     */
+    private static final String VERSION_GATED_KEYWORDS = "AUTO|MANUAL|OFFLINE|ONLINE|PARALLEL|QUALIFY|TABLESAMPLE|URL|VECTOR";
+
+    /**
+     * All keywords the grammar cannot parse as unquoted identifiers; they are backtick-quoted
+     * when they appear in identifier positions.
+     */
+    private static final String KEYWORDS_REQUIRING_BACKTICKS = WINDOW_FUNCTION_KEYWORDS + "|" + VERSION_GATED_KEYWORDS;
+
+    private static final Pattern KEYWORDS_REQUIRING_BACKTICKS_PATTERN = Pattern.compile(
+            "\\b(" + KEYWORDS_REQUIRING_BACKTICKS + ")\\b",
+            Pattern.CASE_INSENSITIVE);
+
+    /**
+     * Table name positions: after CREATE/ALTER/DROP/RENAME/TRUNCATE TABLE (including
+     * TEMPORARY and IF [NOT] EXISTS variants), FROM, JOIN, INTO, REFERENCES,
+     * ON ({@code CREATE INDEX ... ON t}, {@code CREATE TRIGGER ... ON t}),
+     * LIKE ({@code CREATE TABLE t2 LIKE t}),
+     * and after TO ({@code RENAME TABLE ... TO t}, {@code RENAME COLUMN ... TO c}).
+     */
+    private static final String TABLE_NAME_ANCHORS = """
+            CREATE\\s+(?:TEMPORARY\\s+)?TABLE(?:\\s+IF\\s+NOT\\s+EXISTS)?\
+            |ALTER\\s+TABLE|DROP\\s+(?:TEMPORARY\\s+)?TABLE(?:\\s+IF\\s+EXISTS)?\
+            |RENAME\\s+TABLE|TRUNCATE(?:\\s+TABLE)?\
+            |REFERENCES|FROM|JOIN|INTO|TO|ON|LIKE""";
+
+    /**
+     * Column clause positions: ADD/CHANGE/MODIFY/ALTER/DROP [COLUMN] (the COLUMN word is
+     * optional in all of them, e.g. {@code ALTER TABLE t DROP url}), RENAME COLUMN,
+     * and AFTER ({@code ADD COLUMN c INT AFTER c2}).
+     */
+    private static final String COLUMN_CLAUSE_ANCHORS = "ADD(?:\\s+COLUMN)?|CHANGE(?:\\s+COLUMN)?|MODIFY(?:\\s+COLUMN)?|ALTER(?:\\s+COLUMN)?|DROP(?:\\s+COLUMN)?|RENAME\\s+COLUMN|AFTER";
+
+    /**
+     * Column definitions: keyword directly followed by a data type.
+     */
+    private static final Pattern COLUMN_DEFINITION_CONTEXT_PATTERN = Pattern.compile(
+            "\\b(" + KEYWORDS_REQUIRING_BACKTICKS + ")\\s+(" + """
+                    (?:VAR)?CHAR|CHARACTER|NCHAR|NVARCHAR|NATIONAL|(?:TINY|SMALL|MEDIUM|BIG)?INT|INTEGER|INT[12348]|MIDDLEINT\
+                    |DECIMAL|DEC|FIXED|NUMERIC|FLOAT|FLOAT[48]|DOUBLE|REAL|SERIAL\
+                    |JSON|(?:TINY|MEDIUM|LONG)?TEXT|(?:TINY|MEDIUM|LONG)?BLOB|LONG|DATE|DATETIME|TIMESTAMP|TIME|YEAR\
+                    |ENUM|SET|BINARY|VARBINARY|BIT|BOOL|BOOLEAN|VECTOR\
+                    |GEOMETRY|GEOMETRYCOLLECTION|POINT|LINESTRING|POLYGON|MULTIPOINT|MULTILINESTRING|MULTIPOLYGON""" + ")\\b",
+            Pattern.CASE_INSENSITIVE);
+
+    /**
+     * Index and constraint name positions: [ADD] [UNIQUE|FULLTEXT|SPATIAL] INDEX/KEY,
+     * CREATE [UNIQUE|FULLTEXT|SPATIAL] INDEX, DROP INDEX, [ADD] CONSTRAINT,
+     * DROP/ALTER CHECK.
+     */
+    private static final String INDEX_OR_CONSTRAINT_ANCHORS = """
+            (?:ADD\\s+)?(?:UNIQUE\\s+|FULLTEXT\\s+|SPATIAL\\s+)?(?:INDEX|KEY)\
+            |CREATE\\s+(?:UNIQUE\\s+|FULLTEXT\\s+|SPATIAL\\s+)?INDEX|DROP\\s+INDEX\
+            |(?:ADD\\s+)?CONSTRAINT|(?:DROP|ALTER)\\s+CHECK""";
+
+    /**
+     * Non-table schema object name positions: DATABASE/SCHEMA/VIEW/TRIGGER/PROCEDURE/
+     * FUNCTION/EVENT (including IF [NOT] EXISTS variants), USE, and PARTITION names.
+     * The anchor word always directly precedes the name, so leading verbs
+     * ({@code CREATE OR REPLACE VIEW}, {@code REORGANIZE PARTITION}) need no handling.
+     */
+    private static final String OBJECT_NAME_ANCHORS = "(?:DATABASE|SCHEMA|VIEW|TRIGGER|PROCEDURE|FUNCTION|EVENT)(?:\\s+IF(?:\\s+NOT)?\\s+EXISTS)?|USE|PARTITION";
+
+    /**
+     * All identifier-position anchors above, combined into a single alternation: they share
+     * the same {@code anchor <keyword>} shape and the same replacement, so one scan covers
+     * table names, column clauses, index/constraint names and schema object names.
+     */
+    private static final Pattern IDENTIFIER_CONTEXT_PATTERN = Pattern.compile(
+            "\\b(" + TABLE_NAME_ANCHORS + "|" + COLUMN_CLAUSE_ANCHORS + "|" + INDEX_OR_CONSTRAINT_ANCHORS + "|" + OBJECT_NAME_ANCHORS + ")\\s+("
+                    + KEYWORDS_REQUIRING_BACKTICKS + ")\\b",
+            Pattern.CASE_INSENSITIVE);
+
+    /**
+     * Labels in stored routines: {@code <keyword>:}.
+     */
+    private static final Pattern LABEL_CONTEXT_PATTERN = Pattern.compile(
+            "\\b(" + KEYWORDS_REQUIRING_BACKTICKS + ")\\s*:",
             Pattern.CASE_INSENSITIVE);
 
     /**
      * Normalizes MySQL DDL by converting double-quoted strings to single-quoted strings
-     * while preserving backtick-quoted identifiers, and adding backticks around reserved
-     * keywords when used as identifiers. Assumes the server is NOT in ANSI_QUOTES mode.
+     * while preserving backtick-quoted identifiers, and adding backticks around
+     * grammar-restricted keywords used as identifiers. Assumes the server is NOT in
+     * ANSI_QUOTES mode.
      *
      * @param ddlContent The DDL statement to normalize
      * @return Normalized DDL with double-quoted strings converted to single quotes
-     *         and reserved keywords backtick-quoted, or the original input if null or empty
+     *         and grammar-restricted keywords backtick-quoted, or the original input
+     *         if null or empty
      */
     public static String normalize(String ddlContent) {
         return normalize(ddlContent, false);
     }
 
     /**
-     * Normalizes MySQL DDL by adding backticks around reserved keywords when used as
-     * identifiers. When {@code ansiQuotesMode} is false, also converts double-quoted
+     * Normalizes MySQL DDL by adding backticks around grammar-restricted keywords used
+     * as identifiers. When {@code ansiQuotesMode} is false, also converts double-quoted
      * strings to single-quoted strings (for servers using the default sql_mode where
      * double quotes delimit string literals). When {@code ansiQuotesMode} is true,
      * double-quoted text is left as-is because it represents identifiers.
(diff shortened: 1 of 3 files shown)


jpechane:
@vsantonastaso Please review and merge if OK.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/DdlNormalizer.java
```diff
@@ -125,28 +213,25 @@ private static void convertToSingleQuoted(Matcher matcher, StringBuilder result)
...
-
-        // Pattern for labels in stored procedures: <keyword>:
-        result = result.replaceAll(
-                "(?i)\\b(CUME_DIST|DENSE_RANK|FIRST_VALUE|JSON_TABLE|LAG|LAST_VALUE|LEAD|NTH_VALUE|NTILE|PERCENT_RANK|RANK|ROW_NUMBER)\\s*:",
-                "`$1`:");
+        if (!KEYWORDS_REQUIRING_BACKTICKS_PATTERN.matcher(ddl).find()) {
+            return ddl;
+        }
+
+        String result = TABLE_NAME_CONTEXT_PATTERN.matcher(ddl).replaceAll("$1 `$2`");
```
vsantonastaso: Maybe we could optimize it since that the patterns TABLE_NAME_CONTEXT_PATTERN, COLUMN_CLAUSE_CONTEXT_PATTERN, INDEX_OR_CONSTRAINT_CONTEXT_PATTERN, OBJECT_NAME_CONTEXT_PATTERN use the same replacement. We could combine their anchors into a single reducing matcher creations, full scans, allocations from 4 to 1.

Try with 
` Pattern.compile(
          "\\b("
          // table name anchors
          + "CREATE\\s+(?:TEMPORARY\\s+)?TABLE(?:\\s+IF\\s+NOT\\s+EXISTS)?|ALTER\\s+TABLE|DROP\\s+(?:TEMPORARY\\s+)?TABLE(?:\\s+IF\\s+EXISTS)?"
          + "|RENAME\\s+TABLE|TRUNCATE(?:\\s+TABLE)?|REFERENCES|FROM|JOIN|INTO|TO|ON|LIKE"
          // column clause anchors
          + "|ADD(?:\\s+COLUMN)?|CHANGE(?:\\s+COLUMN)?|MODIFY(?:\\s+COLUMN)?|ALTER(?:\\s+COLUMN)?|DROP\\s+COLUMN|RENAME\\s+COLUMN|AFTER"
          // index/constraint anchors
          + "|(?:ADD\\s+)?(?:UNIQUE\\s+|FULLTEXT\\s+|SPATIAL\\s+)?(?:INDEX|KEY)|CREATE\\s+(?:UNIQUE\\s+|FULLTEXT\\s+|SPATIAL\\s+)?INDEX|DROP\\s+INDEX|(?:ADD\\s+)?CONSTRAINT|(?:DROP|ALTER)\\s+CHECK"
          // object name anchors
          + "|(?:DATABASE|SCHEMA|VIEW|TRIGGER|PROCEDURE|FUNCTION|EVENT)(?:\\s+IF(?:\\s+NOT)?\\s+EXISTS)?|USE|PARTITION"
          + ")\\s+(" + KEYWORDS_REQUIRING_BACKTICKS + ")\\b",
          Pattern.CASE_INSENSITIVE);`

It should contain all mentioned patterns. WDYT?

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/DdlNormalizer.java
```diff
@@ -43,28 +44,115 @@ public class DdlNormalizer {
...
+            "\\b(CREATE\\s+(?:TEMPORARY\\s+)?TABLE(?:\\s+IF\\s+NOT\\s+EXISTS)?|ALTER\\s+TABLE|DROP\\s+(?:TEMPORARY\\s+)?TABLE(?:\\s+IF\\s+EXISTS)?"
+                    + "|RENAME\\s+TABLE|TRUNCATE(?:\\s+TABLE)?|REFERENCES|FROM|JOIN|INTO|TO|ON|LIKE)\\s+("
+                    + KEYWORDS_REQUIRING_BACKTICKS + ")\\b",
+            Pattern.CASE_INSENSITIVE);
+
+    /**
+     * Column clause positions: ADD/CHANGE/MODIFY/ALTER [COLUMN], DROP COLUMN, RENAME COLUMN,
+     * and AFTER ({@code ADD COLUMN c INT AFTER c2}).
+     */
+    private static final Pattern COLUMN_CLAUSE_CONTEXT_PATTERN = Pattern.compile(
```
vsantonastaso: Here you are using `DROP\s+COLUMN` so COLUMN is required.
MySQL accepts both `ALTER TABLE t DROP url` and `ALTER TABLE t DROP COLUMN url`.
Also the grammar confirms _COLUMN_SYMBOL?_ is optional in the alterListItem _DROP_ branch (MySqlParser.g4:270).
This means ALTER TABLE t1 DROP url won't backtick url and will fail to parse when url is one of the nine keywords.
Could you check this please?

vsantonastaso (commented):
@minleejae thanks for your efforts. I left two comments. Could you please take a look at them?
Thanks

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/DdlNormalizer.java
minleejae: Benchmarked it with JMH to double-check (both variants built from the same anchor constants, outputs verified identical on a corpus of keyword DDL) — you're right, the single-pass version is measurably faster on DDL that reaches the rewrite path:

| workload | 4-pass (ns/op) | combined (ns/op) | delta |
|---|---|---|---|
| `CREATE TABLE url (id INT PRIMARY KEY)` | 4,035 | 2,948 | -27% |
| ~4.3KB CREATE TABLE + 3 keyword identifiers | 216,272 | 191,783 | -11% |
| ~4.3KB CREATE TABLE + `VECTOR(768)` column (guard passes, nothing to rewrite) | 206,343 | 183,075 | -11% |

Applied in d2c873dbb9 — the four anchor groups are kept as documented constants and combined into one compiled pattern.

<details>
<summary>JMH benchmark source (JDK 22, JMH 1.37, avgt, 2 forks × 3 warmup + 5 measurement × 500 ms)</summary>

```java
package bench;

import java.util.List;
import java.util.concurrent.TimeUnit;
import java.util.regex.Pattern;

import org.openjdk.jmh.annotations.Benchmark;
import org.openjdk.jmh.annotations.BenchmarkMode;
import org.openjdk.jmh.annotations.Fork;
import org.openjdk.jmh.annotations.Measurement;
import org.openjdk.jmh.annotations.Mode;
import org.openjdk.jmh.annotations.OutputTimeUnit;
import org.openjdk.jmh.annotations.Param;
import org.openjdk.jmh.annotations.Scope;
import org.openjdk.jmh.annotations.Setup;
import org.openjdk.jmh.annotations.State;
import org.openjdk.jmh.annotations.Warmup;

/**
 * Compares the current DdlNormalizer keyword-backticking implementation (four sequential
 * replaceAll passes sharing the same replacement) against the reviewer-proposed single
 * combined-alternation pattern. Pattern bodies are copied verbatim from the PR branch;
 * both variants are built from the same anchor constants so the only variable is
 * 4-pass-sequential vs 1-pass-combined. The COLUMN_DEFINITION and LABEL passes have
 * different replacements and stay separate in both variants, exactly as proposed.
 */
@BenchmarkMode(Mode.AverageTime)
@OutputTimeUnit(TimeUnit.NANOSECONDS)
@State(Scope.Benchmark)
@Fork(2)
@Warmup(iterations = 3, time = 500, timeUnit = TimeUnit.MILLISECONDS)
@Measurement(iterations = 5, time = 500, timeUnit = TimeUnit.MILLISECONDS)
public class DdlNormalizerBench {

    // ---- verbatim from DdlNormalizer on the PR branch ----

    private static final String WINDOW_FUNCTION_KEYWORDS = "CUME_DIST|DENSE_RANK|FIRST_VALUE|JSON_TABLE|LAG|LAST_VALUE|LEAD|NTH_VALUE|NTILE|PERCENT_RANK|RANK|ROW_NUMBER";
    private static final String VERSION_GATED_KEYWORDS = "AUTO|MANUAL|OFFLINE|ONLINE|PARALLEL|QUALIFY|TABLESAMPLE|URL|VECTOR";
    private static final String KEYWORDS_REQUIRING_BACKTICKS = WINDOW_FUNCTION_KEYWORDS + "|" + VERSION_GATED_KEYWORDS;

    private static final Pattern KEYWORDS_REQUIRING_BACKTICKS_PATTERN = Pattern.compile(
            "\\b(" + KEYWORDS_REQUIRING_BACKTICKS + ")\\b",
            Pattern.CASE_INSENSITIVE);

    // Anchor groups extracted as constants; the four sequential patterns and the combined
    // pattern are built from these same strings.
    private static final String TABLE_NAME_ANCHORS = "CREATE\\s+(?:TEMPORARY\\s+)?TABLE(?:\\s+IF\\s+NOT\\s+EXISTS)?|ALTER\\s+TABLE|DROP\\s+(?:TEMPORARY\\s+)?TABLE(?:\\s+IF\\s+EXISTS)?"
            + "|RENAME\\s+TABLE|TRUNCATE(?:\\s+TABLE)?|REFERENCES|FROM|JOIN|INTO|TO|ON|LIKE";
    private static final String COLUMN_CLAUSE_ANCHORS = "ADD(?:\\s+COLUMN)?|CHANGE(?:\\s+COLUMN)?|MODIFY(?:\\s+COLUMN)?|ALTER(?:\\s+COLUMN)?|DROP\\s+COLUMN|RENAME\\s+COLUMN|AFTER";
    private static final String INDEX_OR_CONSTRAINT_ANCHORS = "(?:ADD\\s+)?(?:UNIQUE\\s+|FULLTEXT\\s+|SPATIAL\\s+)?(?:INDEX|KEY)|CREATE\\s+(?:UNIQUE\\s+|FULLTEXT\\s+|SPATIAL\\s+)?INDEX|DROP\\s+INDEX|(?:ADD\\s+)?CONSTRAINT|(?:DROP|ALTER)\\s+CHECK";
    private static final String OBJECT_NAME_ANCHORS = "(?:DATABASE|SCHEMA|VIEW|TRIGGER|PROCEDURE|FUNCTION|EVENT)(?:\\s+IF(?:\\s+NOT)?\\s+EXISTS)?|USE|PARTITION";

    private static final Pattern TABLE_NAME_CONTEXT_PATTERN = Pattern.compile(
            "\\b(" + TABLE_NAME_ANCHORS + ")\\s+(" + KEYWORDS_REQUIRING_BACKTICKS + ")\\b",
            Pattern.CASE_INSENSITIVE);
    private static final Pattern COLUMN_CLAUSE_CONTEXT_PATTERN = Pattern.compile(
            "\\b(" + COLUMN_CLAUSE_ANCHORS + ")\\s+(" + KEYWORDS_REQUIRING_BACKTICKS + ")\\b",
            Pattern.CASE_INSENSITIVE);
    private static final Pattern INDEX_OR_CONSTRAINT_CONTEXT_PATTERN = Pattern.compile(
            "\\b(" + INDEX_OR_CONSTRAINT_ANCHORS + ")\\s+(" + KEYWORDS_REQUIRING_BACKTICKS + ")\\b",
            Pattern.CASE_INSENSITIVE);
    private static final Pattern OBJECT_NAME_CONTEXT_PATTERN = Pattern.compile(
            "\\b(" + OBJECT_NAME_ANCHORS + ")\\s+(" + KEYWORDS_REQUIRING_BACKTICKS + ")\\b",
            Pattern.CASE_INSENSITIVE);

    // Reviewer-proposed combined pattern: same four anchor groups in one alternation.
    private static final Pattern COMBINED_CONTEXT_PATTERN = Pattern.compile(
            "\\b(" + TABLE_NAME_ANCHORS + "|" + COLUMN_CLAUSE_ANCHORS + "|" + INDEX_OR_CONSTRAINT_ANCHORS + "|" + OBJECT_NAME_ANCHORS
                    + ")\\s+(" + KEYWORDS_REQUIRING_BACKTICKS + ")\\b",
            Pattern.CASE_INSENSITIVE);

    private static final Pattern COLUMN_DEFINITION_CONTEXT_PATTERN = Pattern.compile(
            "\\b(" + KEYWORDS_REQUIRING_BACKTICKS + ")\\s+("
                    + "(?:VAR)?CHAR|CHARACTER|NCHAR|NVARCHAR|NATIONAL|(?:TINY|SMALL|MEDIUM|BIG)?INT|INTEGER|INT[12348]|MIDDLEINT"
                    + "|DECIMAL|DEC|FIXED|NUMERIC|FLOAT|FLOAT[48]|DOUBLE|REAL|SERIAL"
                    + "|JSON|(?:TINY|MEDIUM|LONG)?TEXT|(?:TINY|MEDIUM|LONG)?BLOB|LONG|DATE|DATETIME|TIMESTAMP|TIME|YEAR"
                    + "|ENUM|SET|BINARY|VARBINARY|BIT|BOOL|BOOLEAN|VECTOR"
                    + "|GEOMETRY|GEOMETRYCOLLECTION|POINT|LINESTRING|POLYGON|MULTIPOINT|MULTILINESTRING|MULTIPOLYGON)\\b",
            Pattern.CASE_INSENSITIVE);

    private static final Pattern LABEL_CONTEXT_PATTERN = Pattern.compile(
            "\\b(" + KEYWORDS_REQUIRING_BACKTICKS + ")\\s*:",
            Pattern.CASE_INSENSITIVE);

    // ---- the two implementations under test ----

    static String sequentialImpl(String ddl) {
        if (!KEYWORDS_REQUIRING_BACKTICKS_PATTERN.matcher(ddl).find()) {
            return ddl;
        }
        String result = TABLE_NAME_CONTEXT_PATTERN.matcher(ddl).replaceAll("$1 `$2`");
        result = COLUMN_CLAUSE_CONTEXT_PATTERN.matcher(result).replaceAll("$1 `$2`");
        result = COLUMN_DEFINITION_CONTEXT_PATTERN.matcher(result).replaceAll("`$1` $2");
        result = INDEX_OR_CONSTRAINT_CONTEXT_PATTERN.matcher(result).replaceAll("$1 `$2`");
        result = OBJECT_NAME_CONTEXT_PATTERN.matcher(result).replaceAll("$1 `$2`");
        result = LABEL_CONTEXT_PATTERN.matcher(result).replaceAll("`$1`:");
        return result;
    }

    static String combinedImpl(String ddl) {
        if (!KEYWORDS_REQUIRING_BACKTICKS_PATTERN.matcher(ddl).find()) {
            return ddl;
        }
        String result = COMBINED_CONTEXT_PATTERN.matcher(ddl).replaceAll("$1 `$2`");
        result = COLUMN_DEFINITION_CONTEXT_PATTERN.matcher(result).replaceAll("`$1` $2");
        result = LABEL_CONTEXT_PATTERN.matcher(result).replaceAll("`$1`:");
        return result;
    }

    // ---- workloads ----

    @Param({ "small_noKw", "large_noKw", "small_kw", "large_kw", "large_vectorType" })
    public String ddlKind;

    String ddl;

    @Setup
    public void setup() {
        verifyEquivalence();
        ddl = switch (ddlKind) {
            case "small_noKw" -> "ALTER TABLE customers ADD COLUMN email VARCHAR(255) NOT NULL AFTER last_name";
            case "large_noKw" -> largeTable(false, false);
            case "small_kw" -> "CREATE TABLE url (id INT PRIMARY KEY)";
            case "large_kw" -> largeTable(true, false);
            case "large_vectorType" -> largeTable(false, true);
            default -> throw new IllegalArgumentException(ddlKind);
        };
    }

    /** Realistic-sized CREATE TABLE (~45 columns, a few KB), optionally containing
     * keyword identifiers and/or a VECTOR(n) typed column (guard passes, nothing to rewrite). */
    private static String largeTable(boolean keywordIdentifiers, boolean vectorType) {
        StringBuilder sb = new StringBuilder(4096);
        sb.append("CREATE TABLE order_line_item_history (\n");
        sb.append("  id BIGINT UNSIGNED NOT NULL AUTO_INCREMENT,\n");
        for (int i = 0; i < 40; i++) {
            sb.append("  col_").append(i).append(i % 3 == 0 ? " VARCHAR(255) DEFAULT NULL,\n" : i % 3 == 1 ? " DECIMAL(18,4) NOT NULL DEFAULT 0,\n" : " DATETIME(6) DEFAULT CURRENT_TIMESTAMP(6),\n");
        }
        if (keywordIdentifiers) {
            sb.append("  url VARCHAR(2048) NOT NULL,\n");
            sb.append("  auto INT DEFAULT 0,\n");
            sb.append("  INDEX vector (col_1),\n");
        }
        if (vectorType) {
            sb.append("  embedding VECTOR(768) NOT NULL,\n");
        }
        sb.append("  created_at TIMESTAMP NOT NULL DEFAULT CURRENT_TIMESTAMP,\n");
        sb.append("  PRIMARY KEY (id),\n");
        sb.append("  KEY idx_created (created_at)\n");
        sb.append(") ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_0900_ai_ci");
        return sb.toString();
    }

    /** Both variants must produce byte-identical output on a corpus of tricky statements. */
    private static void verifyEquivalence() {
        List<String> corpus = List.of(
                "CREATE TABLE url (id INT)",
                "CREATE TABLE vector (id INT)",
                "CREATE TABLE auto (auto INT)",
                "ALTER TABLE t1 ADD INDEX qualify (id)",
                "ALTER TABLE t1 ADD COLUMN auto INT AFTER url",
                "ALTER TABLE t1 ADD url VARCHAR(10)",
                "ALTER TABLE t1 CHANGE COLUMN url url2 VARCHAR(10)",
                "ALTER TABLE t1 MODIFY vector INT",
                "ALTER TABLE t1 DROP COLUMN url",
                "ALTER TABLE t1 RENAME COLUMN url TO auto",
                "ALTER TABLE online RENAME COLUMN url TO auto",
                "ALTER TABLE t1 ADD CONSTRAINT auto CHECK (id > 0)",
                "ALTER TABLE t1 ADD UNIQUE INDEX manual (id)",
                "CREATE INDEX auto ON url (col)",
                "CREATE UNIQUE INDEX offline ON t1 (col)",
                "DROP INDEX vector ON t1",
                "RENAME TABLE auto TO url",
                "TRUNCATE TABLE parallel",
                "TRUNCATE tablesample",
                "DROP TABLE IF EXISTS url",
                "CREATE TABLE t2 LIKE url",
                "USE url",
                "ALTER TABLE t1 DROP PARTITION auto",
                "CREATE TRIGGER trg BEFORE INSERT ON url FOR EACH ROW SET @x = 1",
                "CREATE VIEW auto AS SELECT 1",
                "CREATE DATABASE IF NOT EXISTS vector",
                "CREATE TABLE t1 (id INT, rank INT, lag VARCHAR(10))",
                "CREATE TABLE t3 (id INT, embedding VECTOR(768))",
                "CREATE TABLE t4 (id INT) AUTO_INCREMENT=100",
                "ALTER TABLE t1 ALTER COLUMN url SET DEFAULT 'x'",
                largeTable(true, true),
                largeTable(false, false));
        for (String stmt : corpus) {
            String a = sequentialImpl(stmt);
            String b = combinedImpl(stmt);
            if (!a.equals(b)) {
                throw new IllegalStateException("Implementations diverge!\n  input:      " + stmt
                        + "\n  sequential: " + a + "\n  combined:   " + b);
            }
        }
    }

    @Benchmark
    public String sequential() {
        return sequentialImpl(ddl);
    }

    @Benchmark
    public String combined() {
        return combinedImpl(ddl);
    }
}
```

</details>

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/DdlNormalizer.java
```diff
@@ -43,28 +44,113 @@ public class DdlNormalizer {
...
+    /**
+     * Column definitions: keyword directly followed by a data type.
+     */
+    private static final Pattern COLUMN_DEFINITION_CONTEXT_PATTERN = Pattern.compile(
+            "\\b(" + KEYWORDS_REQUIRING_BACKTICKS + ")\\s+("
+                    + "(?:VAR)?CHAR|CHARACTER|NCHAR|NVARCHAR|NATIONAL|(?:TINY|SMALL|MEDIUM|BIG)?INT|INTEGER|INT[12348]|MIDDLEINT"
+                    + "|DECIMAL|DEC|FIXED|NUMERIC|FLOAT|FLOAT[48]|DOUBLE|REAL|SERIAL"
+                    + "|JSON|(?:TINY|MEDIUM|LONG)?TEXT|(?:TINY|MEDIUM|LONG)?BLOB|LONG|DATE|DATETIME|TIMESTAMP|TIME|YEAR"
+                    + "|ENUM|SET|BINARY|VARBINARY|BIT|BOOL|BOOLEAN|VECTOR"
+                    + "|GEOMETRY|GEOMETRYCOLLECTION|POINT|LINESTRING|POLYGON|MULTIPOINT|MULTILINESTRING|MULTIPOLYGON)\\b",
```
mfvitale: If possible, you could use the [text blocks](https://docs.oracle.com/en/java/javase/15/text-blocks/index.html) to improve the readability. Also in other places.

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/DdlNormalizer.java
minleejae: Good catch — you're right, the bare `DROP` form wasn't covered by any of the context patterns, so `ALTER TABLE t1 DROP url` reached the parser unquoted and failed.

Fixed in ac2725a67f

debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/antlr/DdlNormalizer.java
minleejae: Applied in ce18252334 — thanks!

vsantonastaso (approved):
LGTM thanks @minleejae. Please squash your commits before merging

vsantonastaso:
Thanks @minleejae, merged

PR #7787, opened 2026-08-07
debezium/dbz#2349 Implement Neo4j CUD SMT relational-to-graph

### LINKED ISSUE #2274: Neo4j SMT relational-to-graph
DDD related https://github.com/debezium/debezium-design-documents/issues/55

vsantonastaso:
Fixes https://github.com/debezium/dbz/issues/2274
Based on https://github.com/debezium/debezium-design-documents/issues/55
  ## Description

  Implement a new `Neo4jCudConverter` Single Message Transform that converts Debezium CDC envelope events into Neo4j's CUD format
Add SMT documentation with configuration details, examples and split step section

  ### Tests

  - unit tests for each compontents + a\ PostgreSQL integration test** (`Neo4jCudConverterIT`) that covers  INSERT, UPDATE, DELETE, FK relationships, join tables, and a full use case (e-commerce) pipeline (customers → orders → order\_items)

  ## PR Checklist

### DIFF
--- a/debezium-connect-plugins/src/main/java/io/debezium/transforms/Neo4jCudConverter.java
+++ b/debezium-connect-plugins/src/main/java/io/debezium/transforms/Neo4jCudConverter.java
@@ -1,0 +1,167 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.transforms;
+
+import java.util.Map;
+
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.connect.components.Versioned;
+import org.apache.kafka.connect.connector.ConnectRecord;
+import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.Struct;
+import org.apache.kafka.connect.errors.DataException;
+import org.apache.kafka.connect.transforms.Transformation;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.debezium.Module;
+import io.debezium.config.Configuration;
+import io.debezium.config.Field;
+import io.debezium.data.Envelope;
+import io.debezium.metadata.ConfigDescriptor;
+import io.debezium.transforms.neo4j.CudEvent;
+import io.debezium.transforms.neo4j.CudEventFactory;
+import io.debezium.transforms.neo4j.CudEventSerializer;
+import io.debezium.transforms.neo4j.Neo4jCudConverterConfig;
+import io.debezium.transforms.neo4j.Neo4jCudConverterConfig.OutputMode;
+
+/**
+ * A Kafka Connect SMT that converts Debezium change events into Neo4j CUD format,
+ * enabling CDC pipelines from relational databases to Neo4j via the Neo4j Kafka sink connector.
+ *
+ * @param <R> the subtype of {@link ConnectRecord} on which the transformation will operate
+ * @see <a href="https://neo4j.com/docs/kafka/current/sink/cud-file-format/">Neo4j CUD format</a>
+ */
+public class Neo4jCudConverter<R extends ConnectRecord<R>> implements Transformation<R>, Versioned, ConfigDescriptor {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(Neo4jCudConverter.class);
+
+    private SmtManager<R> smtManager;
+    private Neo4jCudConverterConfig converterConfig;
+    private CudEventFactory eventFactory;
+    private CudEventSerializer serializer;
+
+    @Override
+    public void configure(Map<String, ?> props) {
+        final var config = Configuration.from(props);
+        this.smtManager = new SmtManager<>(config);
+        this.smtManager.validate(config, Neo4jCudConverterConfig.ALL_FIELDS);
+        this.converterConfig = Neo4jCudConverterConfig.from(config, props);
+        this.eventFactory = new CudEventFactory(converterConfig);
+        this.serializer = new CudEventSerializer();
+    }
+
+    @Override
+    public R apply(R record) {
+        if (record.value() == null) {
+            return handleTombstone(record);
+        }
+        if (isNotValidMessage(record)) {
+            return record;
+        }
+
+        return handleMessage(record);
+    }
+
+    @Override
+    public ConfigDef config() {
+        final var config = new ConfigDef();
+        // Per-table mapping keys (table.<name>.*) are dynamic and parsed from the raw properties, so
+        // only the global output settings are statically declared here.
+        Field.group(config, null,
+                Neo4jCudConverterConfig.OUTPUT_MODE,
+                Neo4jCudConverterConfig.TOMBSTONES_ENABLED);
+        return config;
+    }
+
+    @Override
+    public void close() {
+    }
+
+    @Override
+    public String version() {
+        return Module.version();
+    }
+
+    @Override
+    public Field.Set getConfigFields() {
+        return Neo4jCudConverterConfig.ALL_FIELDS;
+    }
+
+    private R handleMessage(R record) {
+        final var message = (Struct) record.value();
+
+        final var table = getTableFromSource(message);
+        final var mapping = converterConfig.mappingFor(table);
+        if (mapping == null) {
+            LOGGER.debug("No Neo4j mapping configured for table '{}'; passing record through unchanged", table);
+            return record;
+        }
+
+        final var op = resolveOperation(message);
+        if (op == null) {
+            return record;
+        }
+
+        final var data = resolveData(message, op);
+        final var cudOp = op == Envelope.Operation.DELETE ? CudEvent.Operation.DELETE : CudEvent.Operation.MERGE;
+        final var events = eventFactory.buildEvents(data, cudOp, mapping);
+
+        final var serializedContent = converterConfig.outputMode() == OutputMode.ARRAY
+                ? serializer.serializeArray(events)
+                : serializer.serializeSingle(events.get(0));
+
+        return record.newRecord(
+                record.topic(),
+                record.kafkaPartition(),
+                record.keySchema(),
+                record.key(),
+                Schema.STRING_SCHEMA,
+                serializedContent,
+                record.timestamp(),
+                record.headers());
+    }
+
+    private boolean isNotValidMessage(R record) {
+        return !smtManager.isValidEnvelope(record);
+    }
+
+    private R handleTombstone(R record) {
+        if (converterConfig.tombstonesEnabled()) {
+            LOGGER.trace("Passing through tombstone record");
+            return record;
+        }
+        return null;
+    }
+
+    private Envelope.Operation resolveOperation(Struct envelope) {
+        final var op = Envelope.Operation.forCode(envelope.getString(Envelope.FieldName.OPERATION));
+
+        if (op == Envelope.Operation.TRUNCATE || op == Envelope.Operation.MESSAGE) {
+            LOGGER.debug("Skipping non-DML operation: {}", envelope.getString(Envelope.FieldName.OPERATION));
+            return null;
+        }
+        return op;
+    }
+
+    private Struct resolveData(Struct envelope, Envelope.Operation op) {
+        if (op == Envelope.Operation.DELETE) {
+            return envelope.getStruct(Envelope.FieldName.BEFORE);
+        }
+        return envelope.getStruct(Envelope.FieldName.AFTER);
+    }
+
+    private String getTableFromSource(Struct envelope) {
+        try {
+            final var source = envelope.getStruct(Envelope.FieldName.SOURCE);
+            return source == null ? null : source.getString("table");
+        }
+        catch (DataException e) {
+            return null;
+        }
+    }
+
+}
--- a/debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/CudEventFactory.java
+++ b/debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/CudEventFactory.java
@@ -1,0 +1,266 @@
+/*
+ * Copyright Debezium Authors.
+ *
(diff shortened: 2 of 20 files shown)


debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/Neo4jCudConfigParser.java
```diff
@@ -0,0 +1,128 @@
...
+        final var relationships = parseRelationships(rawProps);
+
+        return new Neo4jCudConverterConfig(
+                nodeMode, nodeLabels, nodeIdProperties,
+                propertiesInclude, propertiesExclude, deleteDetach,
+                outputMode, tombstonesEnabled, relationships);
+    }
+
+    static List<String> parseCommaSeparated(String value) {
+        if (value == null || value.isBlank()) {
```
Naros:
```suggestion
        if (Strings.isNullOrBlank(value)) {
```
There may be other spots in the PR where `Strings` utility class should be preferred, if you could take a look.

debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/Neo4jCudConfigParser.java
```diff
@@ -0,0 +1,128 @@
...
+            return Collections.emptyList();
+        }
+        return Arrays.stream(value.split(","))
```
Naros: Perhaps here `Strings.listOfTrimmed`?

debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/Neo4jCudConfigParser.java
```diff
@@ -0,0 +1,128 @@
...
+        try {
+            return CudEvent.Operation.valueOf(value.toUpperCase());
+        }
+        catch (IllegalArgumentException e) {
+            throw new ConfigException(key, value, "Must be 'match' or 'merge'");
+        }
+    }
+
+    private static String requireNonBlank(String value, String key) {
+        if (value == null || value.isBlank()) {
```
Naros:
```suggestion
        if (Strings.isNullOrBlank(value)) {
```

debezium-connect-plugins/src/main/java/io/debezium/transforms/Neo4jCudConverter.java
```diff
@@ -0,0 +1,145 @@
...
+import io.debezium.config.Field;
+import io.debezium.data.Envelope;
+import io.debezium.metadata.ConfigDescriptor;
+import io.debezium.transforms.neo4j.CudEvent;
+import io.debezium.transforms.neo4j.CudEventFactory;
+import io.debezium.transforms.neo4j.CudEventSerializer;
+import io.debezium.transforms.neo4j.Neo4jCudConverterConfig;
+import io.debezium.transforms.neo4j.Neo4jCudConverterConfig.OutputMode;
+
+public class Neo4jCudConverter<R extends ConnectRecord<R>> implements Transformation<R>, Versioned, ConfigDescriptor {
```
Naros, line 38: Could we add some Javadoc?

Naros (commented):
This looks really great @vsantonastaso, I've left a few stylistic points.

vsantonastaso:
Thanks @Naros for the review. Feedback addressed

debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/Neo4jCudConverterConfig.java
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    public List<RelationshipMapping> relationships() {
+        return relationships;
+    }
+
+    public Set<String> fkColumns() {
+        return relationships.stream()
+                .map(RelationshipMapping::fkColumn)
+                .collect(Collectors.toSet());
```
Naros: Should we precreate the `Set` rather than re-creating it on each converter call?

vsantonastaso: Sure, good catch 👍

mfvitale (approved):

debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/CudEventSerializer.java
```diff
@@ -0,0 +1,147 @@
...
+            return s.longValue();
+        }
+        if (value instanceof Integer i) {
+            return i.longValue();
+        }
+        if (value instanceof Float f) {
+            return f.doubleValue();
+        }
+        if (value instanceof BigDecimal bd) {
+            return bd.doubleValue();
```
mfvitale: Could we loose precision also here?

vsantonastaso: good catch. I'm going to fix it

vsantonastaso: fixed

debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/CudEventFactory.java
```diff
@@ -0,0 +1,206 @@
...
+
+        final var excludedColumns = columnsExcludedFromProperties();
+        final var properties = extractProperties(data, excludedColumns);
+        return new CudNodeEvent(cudOp, config.nodeLabels(), ids, properties, null);
+    }
+
+    private CudRelationshipEvent buildRelationshipForNodeMode(Struct data, Operation cudOp,
+                                                              RelationshipMapping mapping, Object fkValue) {
+        final var sourceIds = extractIdProperties(data);
+        final var targetIds = Map.<String, Object> of(mapping.targetId(), fkValue);
```
Naros: Is there any concern here with `mapping.targetId()` being null, causing an NPE? We guard against this in Node mode, and I'm not sure if on relationship mode it makes sense.

debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/Neo4jCudConfigParser.java
```diff
@@ -0,0 +1,118 @@
...
+
+    private static Set<String> toSet(List<String> list) {
+        if (list.isEmpty()) {
+            return Collections.emptySet();
+        }
+        return Set.copyOf(list);
+    }
+
+    private static List<RelationshipMapping> parseRelationships(Map<String, ?> rawProps) {
+        final var grouped = groupByFkColumn(rawProps);
```
Naros: Is there any concern here about the `Map` implementation given by Kafka Connect? I'm concerned that the order of the rawProps, based on the ordering implied in `groupByFkColumn`, if we could map what is considered _from_ and _to_ differently across perhaps Connect versions, etc since there is no order guarantee in the configuration properties.

Naros (commented):
Overall LGTM @vsantonastaso great work. I just have two points below, and they may be nothing, but I just wanted to confirm.

debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/CudEventFactory.java
vsantonastaso: The mapping object is built [here](https://github.com/debezium/debezium/blob/bb86a29f5fc7812de3f73de0f2ef1c2569eab497/debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/Neo4jCudConfigParser.java#L80) validating all configurations needed. So the targetId cannot be null at this point. I'll enforce validation methods and highlight it in doc

debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/Neo4jCudConfigParser.java
vsantonastaso: Really good catch, this is non-deterministic.
To fix it,  for `node.mode=relationship`, the endpoint roles are now fixed by the existing `direction`property, declaring which node is outgoing and which is incoming. 
In this way the endpoints are selected by direction rather than position, so orientation becomes deterministic and independent of config order. 
Added validation, tests and docs about it.
Thanks @Naros

vsantonastaso:
> Overall LGTM @vsantonastaso great work. I just have two points below, and they may be nothing, but I just wanted to confirm.

Addressed 👍

debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/CudEventFactory.java
```diff
@@ -0,0 +1,253 @@
...
+
+    /**
+     * Converts a Debezium date type value into an ISO-8601 string, or returns the value
+     * unchanged for any other type.
+     *
+     * @param value the raw column value (epoch integer for the handled temporal types)
+     * @param schema the value's schema, used to detect the logical type; may be {@code null}
+     * @return the ISO-8601 string, or {@code value} unchanged when it is not a handled temporal type
+     */
+    protected static Object normalizeTemporal(Object value, Schema schema) {
```
Naros, line 189: Temporal coverage here is partial: `io.debezium.time.NanoTimestamp` and the TIME family (`Time`, `MicroTime`, `NanoTime`, `ZonedTime`) fall through as raw epoch numbers, and the Connect logical types produced under `time.precision.mode=connect` arrive as `java.util.Date` and get Jackson's default serialization. The doc's type-mapping table also omits the TIME types entirely.

Could we either extend the switch to cover these, or explicitly document them as unhandled?

documentation/modules/ROOT/pages/transformations/neo4j-cud-converter.adoc
```diff
@@ -0,0 +1,477 @@
...
+|`io.debezium.time.ZonedTimestamp`
+|`String`
+|ISO-8601 date-time with its original zone offset, for example `2024-01-01T12:30:45Z`. Passed through unchanged. Compatible with Neo4j `datetime()`.
+
+|`io.debezium.data.Json`
+|`String`
+|Stored as a JSON string.
+
+|`org.apache.kafka.connect.data.Decimal`, `io.debezium.data.VariableScaleDecimal`
+|`String`, `Double`, or `String`
```
Naros: `String` is listed twice here:

```suggestion
|`String` or `Double`
```

documentation/modules/ROOT/pages/transformations/neo4j-cud-converter.adoc
```diff
@@ -0,0 +1,477 @@
+// Category: debezium-using
```
Naros: Anything that intends to be community-only shouldn't carry the `// Key: Value` comments unless it intends to be productized, per Bob's request. We should likely remove these for now.

Naros (commented):
Hi @vsantonastaso, thanks for the quick turnaround. I think this is close. I don't see anything from my side after this. Not sure if anyone else does.

debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/CudEventFactory.java
vsantonastaso: Great catch @Naros , thanks.
I extended `normalizeTemporal` to cover the whole family, including the TIME types (Time/MicroTime/NanoTime → ISO time-of-day at UTC) and the `time.precision.mode=connect `logical types (org.apache.kafka.connect.data.Date/Time/Timestamp), which are handled via .toInstant() at UTC. 
Regarding **UTC**, the Debezium types based on the epoch carry no zone of their own, so I have to choose a zone to render them and I've chosen UTC because it should match the 'time.precision.mode=isostring' convention in Debezium, and it keeps the output deterministic. WDYT?

debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/CudEventFactory.java
```diff
@@ -0,0 +1,273 @@
...
+     * The schema is required to recognize these types, which is why this normalization happens
+     * here rather than in the serializer.
+     */
+    private static Object columnValue(Struct data, Field field) {
+        return normalizeTemporal(data.get(field), field.schema());
+    }
+
+    private static Object columnValue(Struct data, String fieldName) {
+        final var field = data.schema().field(fieldName);
+        return normalizeTemporal(data.get(fieldName), field != null ? field.schema() : null);
```
Naros: A typo in `node.id.properties` could lead to a `DataException` here because we lookup the field name and the API throws this exception if the field doesn't exist, so we never get a chance to evaluate `field != null` on the second line.

debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/CudEventFactory.java
```diff
@@ -0,0 +1,273 @@
...
+     *
+     * @param value the raw column value (an epoch number for the Debezium types, a {@link java.util.Date}
+     *            for the Connect logical types)
+     * @param schema the value's schema, used to detect the logical type; may be {@code null}
+     * @return the ISO-8601 string, or {@code value} unchanged when it is not a handled temporal type
+     */
+    protected static Object normalizeTemporal(Object value, Schema schema) {
+        if (value == null || schema == null || schema.name() == null) {
+            return value;
+        }
```
Naros, line 192: This method is used in several spots, e.g. Line 128, where the `fkValue` is resolved through this call. If the value is ever null for any reason, this method returns null, we don't validate that, and attempt to write a null value into `Map.of`, which fails.

Naros (approved):
Hi @vsantonastaso I see just two things, that I think you could address as follow-ups with their own tests if you'd prefer. Otherwise LGTM; feel free to merge yourself.

debezium-connect-plugins/src/main/java/io/debezium/transforms/neo4j/CudEventFactory.java
vsantonastaso: Added a validation 👍

PR #7797, opened 2026-08-11
debezium/dbz#2398 Bind PostgreSQL numeric arrays without a type modifier to their element values

### LINKED ISSUE #2398: JDBC sink cannot write a PostgreSQL numeric[] column that has no type modifier
## Description

Follow-up to #2100 (case 4). A PostgreSQL `numeric` column **without a type modifier** carries no precision or scale, so the source emits every element of a `numeric[]` column as a `io.debezium.data.VariableScaleDecimal` struct instead of a `Decimal`. The JDBC sink cannot write such a column to a PostgreSQL target: the task fails on the first record.

`numeric(10,2)[]` (with a modifier) already round-trips since debezium/debezium#7604; only the unbounded form is affected. The default `decimal.handling.mode=precise` is required to hit this — `double` and `string` never produce the struct.

## What behavior do you see?

With a pre-created `numeric[]` column:

```
INSERT INTO public.t (id,data) VALUES (('1'::int2),('{"Struct{scale=2,value=[B@2e51ede1}","Struct{scale=2,value=[B@12051d38}"}'))
ERROR: invalid input syntax for type double precision: "Struct{scale=2,value=[B@2e51ede1}"
```

With `schema.evolution=basic` the sink creates the column as `double precision[]`, so even a write that succeeds rounds the exact values away.

## Root cause (verified on `main`)

`ArrayType#getElementTypeName` (PostgreSQL dialect) resolves the element type through the element schema's handler, and `VariableScaleDecimalType#getTypeName` deliberately answers `double precision`, since a single value carries no precision information. `ArrayType#bind` then passes the element values straight to `Connection#createArrayOf`, which cannot encode a `Struct`, so pgjdbc renders it through `toString()` — the `Struct{scale=2,value=[B@...}` text above.

The scalar path does not have this problem: `VariableScaleDecimalType#bind` already unwraps the struct to a `BigDecimal`. Only the array path is missing the equivalent.

## Scope

The fix belongs in the PostgreSQL `ArrayType`: resolve the element type name to `numeric` and unwrap `VariableScaleDecimal` elements before binding.

The scalar `VariableScaleDecimalType#getTypeName` mapping is intentionally left untouched — changing that mapping is what #1163 and #854 cover, and other dialects (e.g. Oracle `NUMBER`) depend on the current behaviour.

## Notes

A PR is ready and will follow shortly.

sundongkim-dev:
Fixes debezium/dbz#2398

## Description

A PostgreSQL `numeric` column without a type modifier carries no precision or scale, so the source emits the elements of a `numeric[]` column as `io.debezium.data.VariableScaleDecimal` structs. Writing such a column to a PostgreSQL sink fails on the first record:

```
INSERT INTO public.t (id,data) VALUES (('1'::int2),('{"Struct{scale=2,value=[B@2e51ede1}",...}'))
ERROR: invalid input syntax for type double precision: "Struct{scale=2,value=[B@2e51ede1}"
```

`ArrayType#getElementTypeName` resolves the element type through the element schema's handler, and `VariableScaleDecimalType#getTypeName` answers `double precision` — correct for a scalar, where a single value carries no precision information, but wrong for an array: with `schema.evolution=basic` the sink then creates a `double precision[]` column and rounds the exact values away. `ArrayType#bind` also passes the elements to `Connection#createArrayOf` untouched, which cannot encode a `Struct`, so the driver renders it through `toString()`.

The scalar path already unwraps the struct in `VariableScaleDecimalType#bind`; only the array path was missing the equivalent. `ArrayType` now resolves such an element to `numeric` and unwraps the elements to their `BigDecimal` value before binding.

`numeric(10,2)[]` is unaffected: its element schema is a `Decimal` carrying the precision and scale, and keeps resolving as before (debezium/dbz#2100).

**Scope.** The scalar `VariableScaleDecimalType#getTypeName` mapping is deliberately left untouched. Changing it is what debezium/dbz#1163 and debezium/dbz#854 are about, and other dialects depend on the current behaviour — the e2e pipeline tests assert `double precision` for scalar `VariableScaleDecimal` columns, including for Oracle `NUMBER`. Special values (`NaN`, `Infinity`) behave exactly as they do for a scalar, which cannot be represented as a `BigDecimal` either.

The UNNEST batch path is out of scope here: it fails for every array column, not only this one, which is debezium/dbz#2399.

The dialect type mapping table in the documentation lists `io.debezium.data.VariableScaleDecimal` as `Types.DOUBLE`, which is the scalar mapping and is unchanged here, so the table still holds and the documentation is not updated.

**Tests.** `ArrayTypeTest` covers the element unwrapping directly (including a null element and the pass-through for other element types). `JdbcSinkColumnTypeMappingIT` gains a round trip into a pre-created `numeric[]` column and one asserting that `schema.evolution=basic` creates the column as `numeric[]` rather than `double precision[]`. Reverting either half of the change fails them with the reported errors: dropping the unwrapping gives `invalid input syntax for type numeric: "Struct{scale=2,value=[B@...}"`, and dropping the element type name silently loses the scale (`2.50` comes back as `2.5`).

Verified with `it-postgresql` (`JdbcSinkColumnTypeMappingIT`, 120 tests across PostgreSQL, MySQL and SQL Server) and the module's 337 unit tests.

## PR Checklist

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/ArrayType.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/ArrayType.java
@@ -5,15 +5,18 @@
  */
 package io.debezium.connector.jdbc.dialect.postgres;
 
+import java.util.Collection;
 import java.util.List;
 import java.util.Locale;
 import java.util.Set;
 
 import org.apache.kafka.connect.data.Schema;
+import org.apache.kafka.connect.data.Struct;
 
 import io.debezium.connector.jdbc.dialect.DatabaseDialect;
 import io.debezium.connector.jdbc.type.AbstractType;
 import io.debezium.connector.jdbc.type.JdbcType;
+import io.debezium.data.VariableScaleDecimal;
 import io.debezium.sink.valuebinding.ValueBindDescriptor;
 
 /**
@@ -54,8 +57,17 @@ public class ArrayType extends AbstractType {
         if (nativeElementType != null) {
             return nativeElementType;
         }
+        if (isVariableScaleDecimalElement(schema)) {
+            // The scalar handler maps VariableScaleDecimal to double precision, which would round the
+            // exact values of a numeric[] away.
+            return "numeric";
+        }
         JdbcType elementJdbcType = dialect.getSchemaType(schema.valueSchema());
         return elementJdbcType.getTypeName(schema.valueSchema(), isKey);
+    }
+
+    private static boolean isVariableScaleDecimalElement(Schema schema) {
+        return VariableScaleDecimal.LOGICAL_NAME.equals(schema.valueSchema().name());
     }
 
     /**
@@ -77,7 +89,20 @@ public class ArrayType extends AbstractType {
             return List.of(new ValueBindDescriptor(index, null));
         }
         final String elementTypeName = baseElementTypeName(getElementTypeName(this.getDialect(), schema, false));
-        return List.of(new ValueBindDescriptor(index, value, java.sql.Types.ARRAY, elementTypeName));
+        return List.of(new ValueBindDescriptor(index, unwrapElements(schema, value), java.sql.Types.ARRAY, elementTypeName));
+    }
+
+    /**
+     * Unwraps {@link VariableScaleDecimal} elements to their {@code BigDecimal} value, which is what
+     * {@link java.sql.Connection#createArrayOf} can encode. Every other element type is passed through.
+     */
+    static Object unwrapElements(Schema schema, Object value) {
+        if (!isVariableScaleDecimalElement(schema)) {
+            return value;
+        }
+        return ((Collection<?>) value).stream()
+                .map(element -> element == null ? null : VariableScaleDecimal.toLogical((Struct) element).getDecimalValue().orElseThrow())
+                .toList();
     }
 
     /**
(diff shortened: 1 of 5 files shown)


vjuranek (approved):
LGTM

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/ArrayType.java
```diff
@@ -54,10 +57,19 @@ private String getElementTypeName(DatabaseDialect dialect, Schema schema, boolea
...
+            // exact values of a numeric[] away.
+            return "numeric";
+        }
         JdbcType elementJdbcType = dialect.getSchemaType(schema.valueSchema());
         return elementJdbcType.getTypeName(schema.valueSchema(), isKey);
     }
 
+    private static boolean isVariableScaleDecimalElement(Schema schema) {
+        return VariableScaleDecimal.LOGICAL_NAME.equals(schema.valueSchema().name());
+    }
```
Naros: Could we put this in `SchemaUtils`? There could be reasons for this elsewhere in the future.

Naros (approved):
Hi @sundongkim-dev, the change LGTM, I have just one organizational nit-pick.

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/ArrayType.java
sundongkim-dev: Moved it to `SchemaUtils#isVariableScaleDecimal(Schema)`, taking the schema to test directly so callers outside arrays can use it too. Covered in `SchemaUtilsTest`.

Naros:

PR #7798, opened 2026-08-11
debezium/dbz#2399 Write records with an array column row-wise instead of an UNNEST batch

sundongkim-dev:
Fixes debezium/dbz#2399

## Description

With `dialect.postgres.unnest.insert.enabled=true`, a batch holding more than one record fails for any array column:

```
ERROR: column "col_array" is of type text[] but expression is of type text
```

`getBatchInsertStatement` derives each column's cast from the field schema, and for an array field that schema type is `ArrayType`, whose `getTypeName` already ends in `[]`. The placeholder therefore becomes `?::text[][]`, and `UNNEST` over it flattens one level too many and yields `text` scalars the column rejects. The values cannot be reshaped either: this path binds one array per column holding the value of every row, so an array column would need an array of arrays.

Records carrying an array field now take the row-wise path through the existing `hasUnnestUnsupportedField` escape hatch, which already serves the geometric `STRUCT` types (debezium/dbz#2135), `bytea` fields (debezium/dbz#2357) and enum fields (debezium/dbz#1344).

A single-record batch was never affected, because `getBatchInsertStatement` already returns empty for one record and the write takes the row-wise path — which is why the existing array tests did not catch this.

**Tests.** `JdbcSinkUnnestArrayIT` writes two records with a `text[]` column in one batch with UNNEST enabled, mirroring `JdbcSinkUnnestBytesIT`. Reverting the change fails it with the error above. Verified alongside `JdbcSinkUnnestBehaviorIT`, `JdbcSinkUnnestBytesIT`, `JdbcSinkColumnTypeMappingIT` and `JdbcSinkInsertModeIT` (242 tests across PostgreSQL, MySQL and SQL Server) and the module's 335 unit tests.

## PR Checklist

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresDatabaseDialect.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/postgres/PostgresDatabaseDialect.java
@@ -188,11 +188,15 @@ public class PostgresDatabaseDialect extends GeneralDatabaseDialect {
      * records take the row-wise path as well. An enum field is handled row-wise for a related reason:
      * the array cast above is derived from the field schema, which for an enum only resolves to the
      * character varying fallback, whereas the cast has to name the destination column's enum type.
+     * An {@code ARRAY} field cannot be batched at all: this path binds one array per column holding the
+     * value of every row, so an array column would need an array of arrays, and the cast derived above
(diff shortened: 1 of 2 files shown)


vjuranek (approved):
LGTM

vjuranek:
CI failures are unrelated.

Naros (approved):

Naros:
Thanks, applied @sundongkim-dev :tada:

PR #7801, opened 2026-08-11
debezium/dbz#2403 Document secondary unique constraint limitations

### LINKED ISSUE #2403: Document secondary unique constraint limitations in the JDBC sink connector
## Task

Document the data integrity limitations of using secondary unique constraints with the Debezium JDBC sink connector.

## Background

The JDBC sink documentation describes `insert.mode=upsert` in terms of the configured primary key: an existing primary key is updated, and a missing primary key is inserted. It also documents that `use.reduction.buffer=true` consolidates events independently by primary key.

Secondary unique constraints introduce dependencies between rows with different primary keys. Kafka preserves ordering only within a partition, and the reduction buffer reasons only about the configured primary key. As a result, preserving secondary unique constraints requires additional ordering guarantees that are not currently documented.

There are two distinct failure scenarios.

## 1. Events for different primary keys are applied out of source order

Assume that the sink initially contains:

```text
id=1, email=x@a.com, name=A
```

The source releases the unique value and then assigns it to a different primary key:

```sql
-- 1. Release the unique value from the existing row.
DELETE FROM t WHERE id = 1;

-- 2. Reuse the value for a different primary key.
INSERT INTO t (id, email, name)
VALUES (2, 'x@a.com', 'B');
```

The source event order is therefore:

```text
1. DELETE id=1
2. INSERT id=2, email=x@a.com
```

If `id=1` and `id=2` are assigned to different Kafka partitions, the sink can observe the events in the opposite order:

```text
2. INSERT id=2, email=x@a.com
1. DELETE id=1
```

At this point, the sink still contains `id=1, email=x@a.com`, so writing the new row conflicts with the secondary unique constraint.

### 1.1 Dialects that target the configured primary key

PostgreSQL uses `ON CONFLICT (<key columns>)`, while Db2, Oracle, and SQL Server use a `MERGE` condition built from the configured key columns. The incoming `id=2` does not match `id=1`, so the connector attempts an insert and the secondary unique constraint rejects it.

The failure is visible, but the sink task can fail or repeatedly retry the same batch.

### 1.2 MySQL-compatible `ON DUPLICATE KEY UPDATE`

MySQL and SingleStore use `INSERT ... ON DUPLICATE KEY UPDATE`. MySQL applies this clause when either a primary key or any unique index conflicts; the statement cannot select the primary key as its conflict target.

For example, the sink can execute an upsert equivalent to:

```sql
INSERT INTO t (id, email, name)
VALUES (2, 'x@a.com', 'B') AS new
ON DUPLICATE KEY UPDATE
  email = new.email,
  name = new.name;
```

Although `id=2` does not exist, `email=x@a.com` matches the secondary unique index of the existing `id=1` row. The insert therefore updates `id=1` instead of creating `id=2`. Because the connector does not update primary-key fields, the intermediate sink state becomes:

```text
id=1, email=x@a.com, name=B
```

The delayed delete is then applied:

```sql
DELETE FROM t WHERE id = 1;
```

The final states differ:

```text
source: id=2, email=x@a.com, name=B
sink:   empty
```

Both sink writes succeed, so the inconsistency might not be detected through connector error monitoring. This behavior was reproduced with MySQL 8.0 by applying the insert/upsert before the delete; the final sink row count was zero.

`ON DUPLICATE KEY UPDATE` can also fail visibly if the resulting update violates another unique constraint.

## 2. Reduction removes an intermediate state required by a unique constraint

This issue can occur even if the records reach the connector in source order.

Assume the initial state:

```text
id=1, email=a@a.com
id=2, email=b@a.com
```

The source swaps the two unique values by using an intermediate value:

```sql
-- 1. Release a@a.com.
UPDATE t SET email='tmp@a.com' WHERE id=1;

-- 2. Assign a@a.com to id=2.
UPDATE t SET email='a@a.com' WHERE id=2;

-- 3. Assign b@a.com to id=1.
UPDATE t SET email='b@a.com' WHERE id=1;
```

Every source transition satisfies the unique constraint. With `use.reduction.buffer=true`, however, the first and third events have the same primary key. Last-write-wins reduction can discard `id=1, email=tmp@a.com` and retain only:

```text
id=1, email=b@a.com
id=2, email=a@a.com
```

Applying either final update first conflicts with the value still held by the other row. The final state is valid, but it cannot be reached without the discarded intermediate transition.

This limitation is database-independent because reduction happens before the dialect writes to the destination database.

## Required documentation

The JDBC sink documentation should explain that:

1. Secondary unique constraints create dependencies between records with different primary keys.
2. Out-of-order processing can cause a unique-constraint failure for dialects that target the configured primary key.
3. For MySQL-compatible dialects that use `ON DUPLICATE KEY UPDATE`, a secondary unique-key conflict can update a row with a different primary key and cause silent data inconsistency.
4. `use.reduction.buffer=true` can remove intermediate row states required to satisfy secondary unique constraints.
5. If secondary unique constraints must be preserved, all dependent events must be routed through the same Kafka partition and applied in source order, with `use.reduction.buffer=false`.
6. Setting `tasks.max=1` alone does not restore ordering between events that are already in different Kafka partitions.

The general warning fits in the **Idempotent writes** section, with a related note in the `use.reduction.buffer` property description. The MySQL-compatible behavior should be called out separately because it can produce silent inconsistency rather than a visible constraint violation.

## References

- [Debezium JDBC sink connector documentation](https://debezium.io/documentation/reference/stable/connectors/jdbc.html#jdbc-idempotent-writes)
- [MySQL `INSERT ... ON DUPLICATE KEY UPDATE` documentation](https://dev.mysql.com/doc/refman/8.4/en/insert-on-duplicate.html)

minleejae:
Fixes debezium/dbz#2403

## Description

Documents the JDBC sink limitations when destination tables contain secondary unique constraints.

The update explains that:

- dependent records from different Kafka partitions can be applied out of source order;
- MySQL and SingleStore upserts can match a secondary unique index and silently update a row with a different primary key; and
- primary-key reduction can discard intermediate states required to satisfy a secondary unique constraint.

It also documents the safe processing conditions: route dependent events through the same partition, preserve source order, and disable the reduction buffer.

## PR Checklist

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -312,9 +312,11 @@ endif::community[]
 |`INSERT ... ON CONFLICT ... DO UPDATE SET ...`
 |Configured primary key only
 
+ifdef::community[]
 |SingleStore
 |`INSERT ... ON DUPLICATE KEY UPDATE ...`
 |Any unique index
+endif::community[]
 
 |SQL Server
 |`MERGE ...`
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -281,49 +281,75 @@ If the primary key value already exists, the operation updates values in the row
 If the specified primary key value doesn't exist, an `insert` adds a new row.
 
 Each database dialect handles idempotent writes differently, because there is no SQL standard for _upsert_ operations.
-The following table shows the `upsert` DML syntax for the database dialects that {prodname} supports:
+The following table shows the `upsert` DML syntax for the database dialects that {prodname} supports, and the constraint that each dialect evaluates to decide between the insert and the update action:
 
 |===
-|Dialect |Upsert Syntax
+|Dialect |Upsert Syntax |Constraint that triggers the update action
 
 |CockroachDB
 |`INSERT ... ON CONFLICT ... DO UPDATE SET ...`
+|Configured primary key only
 
 |Db2
 |`MERGE ...`
+|Configured primary key only
+
+|MariaDB
+|`INSERT ... ON DUPLICATE KEY UPDATE ...`
+|Any unique index
 
 |MySQL
 |`INSERT ... ON DUPLICATE KEY UPDATE ...`
+|Any unique index
 
 |Oracle
 |`MERGE ...`
+|Configured primary key only
 
 |PostgreSQL
 |`INSERT ... ON CONFLICT ... DO UPDATE SET ...`
+|Configured primary key only
 
 |SingleStore
 |`INSERT ... ON DUPLICATE KEY UPDATE ...`
+|Any unique index
 
 |SQL Server
 |`MERGE ...`
+|Configured primary key only
 
 |StarRocks
 |`INSERT ...` (PRIMARY KEY tables execute inserts as upserts)
+|Table primary key
 
 |===
 
 [WARNING]
 ====
 Secondary unique constraints can create ordering dependencies between records that have different primary keys.
-Kafka guarantees record order only within a partition.
-If dependent records are routed to different partitions, the sink connector might apply them out of source order, causing a unique constraint violation.
+Kafka guarantees record order only within a partition, and the connector batches and reduces events based only on the configured primary key.
+For these reasons, avoid defining secondary unique constraints or secondary unique indexes on destination tables.
+If a uniqueness rule must be enforced, enforce it in the source database, and omit the constraint from the destination table.
 
-MySQL and SingleStore use `INSERT ... ON DUPLICATE KEY UPDATE`, which reacts to a conflict with any unique index, not only the configured primary key.
-A secondary unique-key conflict can therefore update an existing row whose primary key differs from the incoming record, potentially causing data inconsistency without reporting an error.
+If dependent records are routed to different partitions, the sink connector might apply them out of source order.
+How the resulting conflict surfaces depends on the dialect, as shown in the preceding table:
 
-If a destination table contains secondary unique constraints, route all dependent events through the same Kafka partition, preserve source event order, and set xref:jdbc-property-use-reduction-buffer[`use.reduction.buffer=false`].
-For a table with dependencies between arbitrary rows, using a single partition for its Kafka topic provides this ordering.
+* Dialects that use `INSERT ... ON DUPLICATE KEY UPDATE` react to a conflict with any unique index, not only the configured primary key.
+A secondary unique-key conflict can therefore update an existing row whose primary key differs from the incoming record, causing silent data inconsistency.
+* Dialects that use `INSERT ... ON CONFLICT` or `MERGE` detect conflicts only on the configured primary key.
+A conflict with a secondary unique constraint causes the write to fail with a constraint violation error, rather than silently modifying a different row.
+
+If a destination table must retain a secondary unique constraint, all of the following conditions are required:
+
+* Route all dependent events through the same Kafka partition, and preserve the source event order.
+For a table with dependencies between arbitrary rows, use a single partition for its Kafka topic.
 Setting `tasks.max=1` alone does not restore ordering between records that are already in different partitions.
+* Keep xref:jdbc-property-use-reduction-buffer[`use.reduction.buffer`] at its default value of `false`.
+Reduction by primary key can remove intermediate row states that are required to satisfy secondary unique constraints.
+* Set xref:jdbc-property-batch-size[`batch.size`] to `1` if a batch might contain multiple events for a primary key that does not yet exist in the destination table.
+Some dialects and driver configurations process a batch in a single pass, rather than validating each row against the effects of the preceding rows in the same batch.
+For example, `MERGE`-based dialects can attempt two `INSERT` operations for the same primary key instead of an `INSERT` followed by an `UPDATE`, and PostgreSQL combines the batch into a single multi-row statement when xref:jdbc-property-dialect-postgres-unnest-insert-enabled[`dialect.postgres.unnest.insert.enabled`] is enabled.
+In both cases, the batch fails.
 ====
 
 // Type: reference
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -362,7 +362,7 @@ A conflict involving a secondary unique constraint can therefore update an exist
 A conflict involving a secondary unique constraint causes the write operation to fail with a constraint violation error rather than updating a different row.
 |===
 
-For a destination table to retain a secondary unique constraint, the data pipeline must meet all of the following requirements:
+To prevent data inconsistencies when a destination table includes a secondary unique constraint, the data pipeline must meet all of the following requirements:
 
 Dependent events routed through the same partition::
 Route all dependent events through the same Kafka partition, and preserve the source event order.
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -281,7 +281,7 @@ If the primary key value already exists, the operation updates values in the row
 If the specified primary key value doesn't exist, an `insert` adds a new row.
 
 Each database dialect handles idempotent writes differently, because there is no SQL standard for _upsert_ operations.
-The following table shows the `upsert` DML syntax for the database dialects that {prodname} supports, and the constraint that each dialect evaluates to decide between the insert and the update action:
+The following table shows the upsert DML syntax for each database dialect that {prodname} supports and which constraints trigger an update action when processing an `upsert` operation.
 
 |===
 |Dialect |Upsert Syntax |Constraint that triggers the update action
@@ -326,31 +326,55 @@ The following table shows the `upsert` DML syntax for the database dialects that
 
 [WARNING]
 ====
+Secondary unique constraints can create ordering dependencies.
+For more information about their use, see xref:sink-connector-secondary-unique-constraints[secondary unique constraints].
+====
+
+// Type: concept
+// ModuleID: debezium-jdbc-connector-secondary-unique-constraints
+[id="sink-connector-secondary-unique-constraints"]
+=== Secondary unique constraints in sink connectors
+
 Secondary unique constraints can create ordering dependencies between records that have different primary keys.
 Kafka guarantees record order only within a partition, and the connector batches and reduces events based only on the configured primary key.
-For these reasons, avoid defining secondary unique constraints or secondary unique indexes on destination tables.
-If a uniqueness rule must be enforced, enforce it in the source database, and omit the constraint from the destination table.
+For these reasons, you should avoid defining secondary unique constraints or secondary unique indexes on destination tables.
+If a uniqueness rule must be enforced, enforce the rule in the source database, and omit the constraint from the destination table.
+
+.Understanding the risk
 
 If dependent records are routed to different partitions, the sink connector might apply them out of source order.
-How the resulting conflict surfaces depends on the dialect, as shown in the preceding table:
+The following table shows how different database dialects respond to a resulting conflict:
 
-* Dialects that use `INSERT ... ON DUPLICATE KEY UPDATE` react to a conflict with any unique index, not only the configured primary key.
+.Dialect-specific behavior when secondary unique constraint conflicts occur
+|===
+|Dialect |Behavior
+
+|Dialects that use `INSERT ... ON DUPLICATE KEY UPDATE`
+|React to a conflict with any unique index, not only the configured primary key.
 A secondary unique-key conflict can therefore update an existing row whose primary key differs from the incoming record, causing silent data inconsistency.
-* Dialects that use `INSERT ... ON CONFLICT` or `MERGE` detect conflicts only on the configured primary key.
+
+|Dialects that use `INSERT ... ON CONFLICT` or `MERGE`
+|Detect conflicts only on the configured primary key.
 A conflict with a secondary unique constraint causes the write to fail with a constraint violation error, rather than silently modifying a different row.
+|===
 
-If a destination table must retain a secondary unique constraint, all of the following conditions are required:
+For a destination table to retain a secondary unique constraint, the data pipeline must meet all of the following requirements:
 
-* Route all dependent events through the same Kafka partition, and preserve the source event order.
+Dependent events routed through the same partition::
+Route all dependent events through the same Kafka partition, and preserve the source event order.
 For a table with dependencies between arbitrary rows, use a single partition for its Kafka topic.
-Setting `tasks.max=1` alone does not restore ordering between records that are already in different partitions.
-* Keep xref:jdbc-property-use-reduction-buffer[`use.reduction.buffer`] at its default value of `false`.
+Setting `tasks.max=1` is not sufficient to restore ordering between records that are already in different partitions.
+
+Reduction buffering disabled::
+The default value (`false`) is retained for the xref:jdbc-property-use-reduction-buffer[`use.reduction.buffer`] property.
 Reduction by primary key can remove intermediate row states that are required to satisfy secondary unique constraints.
-* Set xref:jdbc-property-batch-size[`batch.size`] to `1` if a batch might contain multiple events for a primary key that does not yet exist in the destination table.
+
+Batch size set to `1`::
+The xref:jdbc-property-batch-size[`batch.size`] property is set to `1` if a batch might contain multiple events for a primary key that does not yet exist in the destination table.
++
 Some dialects and driver configurations process a batch in a single pass, rather than validating each row against the effects of the preceding rows in the same batch.
 For example, `MERGE`-based dialects can attempt two `INSERT` operations for the same primary key instead of an `INSERT` followed by an `UPDATE`, and PostgreSQL combines the batch into a single multi-row statement when xref:jdbc-property-dialect-postgres-unnest-insert-enabled[`dialect.postgres.unnest.insert.enabled`] is enabled.
-In both cases, the batch fails.
-====
+In each of these cases, the batch fails.
 
 // Type: reference
 // Title: Schema evolution modes for the {prodname} JDBC connector
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -338,7 +338,7 @@ For more information about their use, see xref:sink-connector-secondary-unique-c
 Secondary unique constraints can create ordering dependencies between records that have different primary keys.
 Kafka guarantees record order only within a partition, and the connector batches and reduces events based only on the configured primary key.
 For these reasons, you should avoid defining secondary unique constraints or secondary unique indexes on destination tables.
-If a uniqueness rule must be enforced, enforce the rule in the source database, and omit the constraint from the destination table.
+If a uniqueness rule is required, enforce the rule in the source database, and omit the constraint from the destination table.
 
 .Understanding the risk
 
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -346,8 +346,6 @@ Kafka guarantees record order only within a partition, and the connector batches
 For these reasons, you should avoid defining secondary unique constraints or secondary unique indexes on destination tables.
 If a uniqueness rule is required, enforce the rule in the source database, and omit the constraint from the destination table.
 
-.Understanding the risk
-
 If dependent records are routed to different partitions, the sink connector might apply them out of source order.
 The following table shows how different database dialects respond to a resulting conflict:
 
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -286,9 +286,11 @@ The following table shows, for each database dialect that {prodname} supports, t
 |===
 |Dialect |Upsert Syntax |Constraint that triggers the update action
 
+ifdef::community[]
 |CockroachDB
 |`INSERT ... ON CONFLICT ... DO UPDATE SET ...`
 |Configured primary key only
+endif::community[]
 
 |Db2
 |`MERGE ...`
--- a/documentation/modules/ROOT/pages/connectors/jdbc.adoc
+++ b/documentation/modules/ROOT/pages/connectors/jdbc.adoc
@@ -312,6 +312,20 @@ The following table shows the `upsert` DML syntax for the database dialects that
 
 |===
 
+[WARNING]
+====
+Secondary unique constraints can create ordering dependencies between records that have different primary keys.
+Kafka guarantees record order only within a partition.
+If dependent records are routed to different partitions, the sink connector might apply them out of source order, causing a unique constraint violation.
+
+MySQL and SingleStore use `INSERT ... ON DUPLICATE KEY UPDATE`, which reacts to a conflict with any unique index, not only the configured primary key.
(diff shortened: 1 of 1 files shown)


Naros (commented):
Overall, I think at the very least documenting this now makes sense. 

But I am concerned with the wording, particularly for the corner case where you have this secondary unique constraint, you setup the transport layer as you described, using a single topic with a single partition with 1 task consumer. In this scenario, you can still face problems because you set `use.reduction.buffer` as `false`.

The way this happens is the source performs two operations for the same PK, and this PK does not yet exist at the target. If the `batch.size` isn't explicitly set to `1`, the `MERGE` operation fails due to conflicts if those two source operations arrive in the same batch at the sink connector.

This is because not every dialect's `MERGE` operation performs the INSERT-v-UPDATE validation per row after each previous row operation. Instead it does a single pass to determine the data set eligibility and fires those operations. So in the case I described, `MERGE` attempts double `INSERT` rather than `INSERT` followed by an `UPDATE`.

So perhaps in this context we may need to explicitly call out that the `batch.size` must also be set to `1`, if you expect a batch to contain multiple events for the same PK that doesn't already exist at the target.

minleejae:
Hi, @Naros Thanks — I've updated the PR. The warning now recommends avoiding secondary unique constraints on destination tables entirely and, where they must be kept, requires same-partition routing, use.reduction.buffer=false, and batch.size=1 when a batch might contain multiple events for the same new primary key.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,77 @@ If the primary key value already exists, the operation updates values in the row
 If the specified primary key value doesn't exist, an `insert` adds a new row.
 
 Each database dialect handles idempotent writes differently, because there is no SQL standard for _upsert_ operations.
-The following table shows the `upsert` DML syntax for the database dialects that {prodname} supports:
+The following table shows the `upsert` DML syntax for the database dialects that {prodname} supports, and the constraint that each dialect evaluates to decide between the insert and the update action:
```
roldanbob:
```suggestion
The following table shows the upsert DML syntax for each database dialect that {prodname} supports and which constraints trigger an update action when processing an `upsert` operation.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,77 @@ If the primary key value already exists, the operation updates values in the row
...
+For a table with dependencies between arbitrary rows, use a single partition for its Kafka topic.
+Setting `tasks.max=1` alone does not restore ordering between records that are already in different partitions.
+* Keep xref:jdbc-property-use-reduction-buffer[`use.reduction.buffer`] at its default value of `false`.
+Reduction by primary key can remove intermediate row states that are required to satisfy secondary unique constraints.
+* Set xref:jdbc-property-batch-size[`batch.size`] to `1` if a batch might contain multiple events for a primary key that does not yet exist in the destination table.
+Some dialects and driver configurations process a batch in a single pass, rather than validating each row against the effects of the preceding rows in the same batch.
+For example, `MERGE`-based dialects can attempt two `INSERT` operations for the same primary key instead of an `INSERT` followed by an `UPDATE`, and PostgreSQL combines the batch into a single multi-row statement when xref:jdbc-property-dialect-postgres-unnest-insert-enabled[`dialect.postgres.unnest.insert.enabled`] is enabled.
+In both cases, the batch fails.
+====
```
roldanbob:
```suggestion
[WARNING]
====
Secondary unique constraints can create ordering dependencies.
For more information about their use, see xref:secondary-unique-constraints[secondary unique constraints].
====

// Type: concept
// ModuleID: debezium-jdbc-connector-connector-secondary-unique-constraints
[id="sink-connector-secondary-unique-constraints"]
=== Secondary unique constraints in sink connectors

Secondary unique constraints can create ordering dependencies between records that have different primary keys.
Kafka guarantees record order only within a partition, and the connector batches and reduces events based only on the configured primary key.
For these reasons, you should avoid defining secondary unique constraints or secondary unique indexes on destination tables.
If a uniqueness rule is required enforced, enforce the rule in the source database, and omit the constraint from the destination table.

.Understanding the risk

If dependent records are routed to different partitions, the sink connector might apply them out of source order.
The following table shows how different database dialects respond to a resulting conflict:

.Dialect-specific behavior when secondary unique constraint conflicts occur
|===
|Dialect |Behavior

|Dialects that use `INSERT ... ON DUPLICATE KEY UPDATE`
|React to a conflict with any unique index, not only the configured primary key.
A secondary unique-key conflict can therefore update an existing row whose primary key differs from the incoming record, causing silent data inconsistency.

|Dialects that use `INSERT ... ON CONFLICT` or `MERGE`
|Detect conflicts only on the configured primary key.
A conflict with a secondary unique constraint causes the write to fail with a constraint violation error, rather than silently modifying a different row.
|===

For a destination table to retain a secondary unique constraint, the data pipeline must meet all of the following requirements:

Dependent events routed through the same partition::
Route all dependent events through the same Kafka partition, and preserve the source event order.
For a table with dependencies between arbitrary rows, use a single partition for its Kafka topic.
Setting `tasks.max=1` is not sufficient to restore ordering between records that are already in different partitions.

Reduction buffering disabled::
The default value (`false`) is retained for the xref:jdbc-property-use-reduction-buffer[`use.reduction.buffer`] property.
Reduction by primary key can remove intermediate row states that are required to satisfy  secondary unique constraints.

Batch size set to `1`::
The xref:jdbc-property-batch-size[`batch.size`] property is set to `1` if a batch might contain multiple events for a primary key that does not yet exist in the destination table.
+
Some dialects and driver configurations process a batch in a single pass, rather than validating each row against the effects of the preceding rows in the same batch.
For example, `MERGE`-based dialects can attempt two `INSERT` operations for the same primary key instead of an `INSERT` followed by an `UPDATE`, and PostgreSQL combines the batch into a single multi-row statement when xref:jdbc-property-dialect-postgres-unnest-insert-enabled[`dialect.postgres.unnest.insert.enabled`] is enabled.
In each of these cases, the batch fails.

```
Placing so much information into an admonition and mixing problem description, database-specific behavior, and procedural requirements makes it difficult to parse. Let's convert the bulk of the content into a separate concept topic to make it more consumable.

roldanbob (changes requested):
@minleejae Hi. I made some minor language edits and suggested restructuring the admonition to make the information easier to parse. Thanks!

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Applied in e34b4d0. Thanks!

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Applied in e34b4d0 — thanks, splitting this into its own concept topic reads much better than the long admonition.

I made three small adjustments while applying the suggestion. Please let me know if you would prefer any of them handled differently:

1. **xref target.** The suggestion used `xref:secondary-unique-constraints[...]`, but the section id is `sink-connector-secondary-unique-constraints`, so the xref would not have resolved. I pointed the xref at the section id. I kept your id value as-is, but I am happy to rename it to `jdbc-secondary-unique-constraints` instead, to match the `[[jdbc-*]]` anchors used by the other sections in this file — just let me know which you prefer.
2. **ModuleID.** Changed `debezium-jdbc-connector-connector-secondary-unique-constraints` to `debezium-jdbc-connector-secondary-unique-constraints`, since the repeated `connector-` looked unintentional.
3. **Wording.** "If a uniqueness rule is required enforced" -> "If a uniqueness rule must be enforced".

I left out a `// Title:` line, since the other `// Type: concept` module in this file omits it as well. Happy to add one if you would rather have it.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,101 @@ If the primary key value already exists, the operation updates values in the row
...
+
+// Type: concept
+// ModuleID: debezium-jdbc-connector-secondary-unique-constraints
+[id="sink-connector-secondary-unique-constraints"]
+=== Secondary unique constraints in sink connectors
+
+Secondary unique constraints can create ordering dependencies between records that have different primary keys.
+Kafka guarantees record order only within a partition, and the connector batches and reduces events based only on the configured primary key.
+For these reasons, you should avoid defining secondary unique constraints or secondary unique indexes on destination tables.
+If a uniqueness rule must be enforced, enforce the rule in the source database, and omit the constraint from the destination table.
```
roldanbob:
```suggestion
If a uniqueness rule is required, enforce the rule in the source database, and omit the constraint from the destination table.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
roldanbob: Thanks for the corrections! I always seem to have problems with making structural changes when working in the small comment window!

The  wording for #3 requires a slight adjustment, which I've added in a new comment. Otherwise, I agree with your on all of these.

roldanbob (changes requested):
@minleejae I made a small adjustment to the wording on L341. If you could apply that then I think that this is good to go. Thanks!

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Thanks! Applied in bbe7369.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,101 @@ If the primary key value already exists, the operation updates values in the row
...
-The following table shows the `upsert` DML syntax for the database dialects that {prodname} supports:
+The following table shows the upsert DML syntax for each database dialect that {prodname} supports and which constraints trigger an update action when processing an `upsert` operation.
 
 |===
-|Dialect |Upsert Syntax
+|Dialect |Upsert Syntax |Constraint that triggers the update action
 
 |CockroachDB
 |`INSERT ... ON CONFLICT ... DO UPDATE SET ...`
+|Configured primary key only
```
roldanbob, line 292:
```suggestion
ifdef::community[]
|CockroachDB
|`INSERT ... ON CONFLICT ... DO UPDATE SET ...`
|Configured primary key only
endif::community[]
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,101 @@ If the primary key value already exists, the operation updates values in the row
 If the specified primary key value doesn't exist, an `insert` adds a new row.
 
 Each database dialect handles idempotent writes differently, because there is no SQL standard for _upsert_ operations.
-The following table shows the `upsert` DML syntax for the database dialects that {prodname} supports:
+The following table shows the upsert DML syntax for each database dialect that {prodname} supports and which constraints trigger an update action when processing an `upsert` operation.
```
roldanbob:
```suggestion
The following table shows, for each database dialect that {prodname} supports, the upsert DML syntax and the constraints that trigger an update action during an upsert operation.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,101 @@ If the primary key value already exists, the operation updates values in the row
...
 |`MERGE ...`
+|Configured primary key only
 
 |PostgreSQL
 |`INSERT ... ON CONFLICT ... DO UPDATE SET ...`
+|Configured primary key only
 
 |SingleStore
 |`INSERT ... ON DUPLICATE KEY UPDATE ...`
+|Any unique index
```
roldanbob, line 318:
```suggestion
ifdef::community[]
|SingleStore
|`INSERT ... ON DUPLICATE KEY UPDATE ...`
|Any unique index
endif::community[]
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,101 @@ If the primary key value already exists, the operation updates values in the row
...
 
 |SQL Server
 |`MERGE ...`
+|Configured primary key only
 
 |StarRocks
 |`INSERT ...` (PRIMARY KEY tables execute inserts as upserts)
+|Table primary key
```
roldanbob, line 328:
```suggestion
ifdef::community[]
|StarRocks
|`INSERT ...` (PRIMARY KEY tables execute inserts as upserts)
|Table primary key
endif::community[]
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,101 @@ If the primary key value already exists, the operation updates values in the row
...
+// ModuleID: debezium-jdbc-connector-secondary-unique-constraints
+[id="sink-connector-secondary-unique-constraints"]
+=== Secondary unique constraints in sink connectors
+
+Secondary unique constraints can create ordering dependencies between records that have different primary keys.
+Kafka guarantees record order only within a partition, and the connector batches and reduces events based only on the configured primary key.
+For these reasons, you should avoid defining secondary unique constraints or secondary unique indexes on destination tables.
+If a uniqueness rule is required, enforce the rule in the source database, and omit the constraint from the destination table.
 
+.Understanding the risk
```
roldanbob:
```suggestion
```
Remove.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,101 @@ If the primary key value already exists, the operation updates values in the row
...
+
+If dependent records are routed to different partitions, the sink connector might apply them out of source order.
```
roldanbob:
```suggestion
If dependent records are routed to different partitions, the sink connector might apply records out of source order, which can lead to inconsistent results.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,101 @@ If the primary key value already exists, the operation updates values in the row
...
+The following table shows how different database dialects respond to a resulting conflict:
```
roldanbob:
```suggestion
The following table shows how different database dialects handle conflicts involving secondary unique constraints:
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,101 @@ If the primary key value already exists, the operation updates values in the row
...
+
+.Dialect-specific behavior when secondary unique constraint conflicts occur
+|===
+|Dialect |Behavior
+
+|Dialects that use `INSERT ... ON DUPLICATE KEY UPDATE`
+|React to a conflict with any unique index, not only the configured primary key.
+A secondary unique-key conflict can therefore update an existing row whose primary key differs from the incoming record, causing silent data inconsistency.
```
roldanbob:
```suggestion
|Consider conflicts on any unique index, not only the configured primary key.
A conflict involving a secondary unique constraint can therefore update an existing row whose primary key differs from the incoming record, causing silent data inconsistency.
```
Update the descriptions to use parallel language so that it's easier to compare them.

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,101 @@ If the primary key value already exists, the operation updates values in the row
...
+
+|Dialects that use `INSERT ... ON CONFLICT` or `MERGE`
+|Detect conflicts only on the configured primary key.
+A conflict with a secondary unique constraint causes the write to fail with a constraint violation error, rather than silently modifying a different row.
```
roldanbob:
```suggestion
|Consider conflicts only on the configured primary key.
A conflict involving a secondary unique constraint causes the write operation to fail with a constraint violation error rather than updating a different row.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,101 @@ If the primary key value already exists, the operation updates values in the row
...
 |===
 
+For a destination table to retain a secondary unique constraint, the data pipeline must meet all of the following requirements:
```
roldanbob:
```suggestion
To prevent data inconsistencies when a destination table includes a secondary unique constraint, the data pipeline must meet all of the following requirements:
```
I was not sure how to interpret the phrase "retain a secondary unique constraint". Does this alternative capture the intended meaning?

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,101 @@ If the primary key value already exists, the operation updates values in the row
...
+
+Dependent events routed through the same partition::
+Route all dependent events through the same Kafka partition, and preserve the source event order.
+For a table with dependencies between arbitrary rows, use a single partition for its Kafka topic.
+Setting `tasks.max=1` is not sufficient to restore ordering between records that are already in different partitions.
+
+Reduction buffering disabled::
+The default value (`false`) is retained for the xref:jdbc-property-use-reduction-buffer[`use.reduction.buffer`] property.
```
roldanbob:
```suggestion
Retain the default value (`false`) of the xref:jdbc-property-use-reduction-buffer[`use.reduction.buffer`] property.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,101 @@ If the primary key value already exists, the operation updates values in the row
...
+Reduction by primary key can remove intermediate row states that are required to satisfy secondary unique constraints.
+
+Batch size set to `1`::
+The xref:jdbc-property-batch-size[`batch.size`] property is set to `1` if a batch might contain multiple events for a primary key that does not yet exist in the destination table.
```
roldanbob:
```suggestion
Set the value of the xref:jdbc-property-batch-size[`batch.size`] property to `1` if any batches might contain multiple events for a primary key that is not defined in the destination table.
```

documentation/modules/ROOT/pages/connectors/jdbc.adoc
```diff
@@ -281,37 +281,101 @@ If the primary key value already exists, the operation updates values in the row
...
++
+Some dialects and driver configurations process a batch in a single pass, rather than validating each row against the effects of the preceding rows in the same batch.
+For example, `MERGE`-based dialects can attempt two `INSERT` operations for the same primary key instead of an `INSERT` followed by an `UPDATE`, and PostgreSQL combines the batch into a single multi-row statement when xref:jdbc-property-dialect-postgres-unnest-insert-enabled[`dialect.postgres.unnest.insert.enabled`] is enabled.
```
roldanbob:
```suggestion
For example, `MERGE`-based dialects can attempt two `INSERT` operations for the same primary key instead of an `INSERT` followed by an `UPDATE`.
Similarly, when xref:jdbc-property-dialect-postgres-unnest-insert-enabled[`dialect.postgres.unnest.insert.enabled`] is set to `true`, PostgreSQL combines a batch into a single multi-row statement .
```

roldanbob (changes requested):
@minleejae cc: @Naros  
Several comments for you to consider. Let me know if you have any questions. Thanks!

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Applied in [`7b8d0a8`](https://github.com/debezium/debezium/pull/7801/commits/7b8d0a81919c49174f581515b4c213561cfcf7ab).

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Applied in [`65d2e16`](https://github.com/debezium/debezium/pull/7801/commits/65d2e167ed3ea1ed124d19167f35121d6c59eab0).

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Applied in [`577ae89`](https://github.com/debezium/debezium/pull/7801/commits/577ae896f2b1ff81c5fa95f4951f2a455e4e49ae).

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Applied in [`18ea36e`](https://github.com/debezium/debezium/pull/7801/commits/18ea36e8e3ecd80f2bb1cccabdb9a19c482d2aa3).

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Applied in [`3a7c980`](https://github.com/debezium/debezium/pull/7801/commits/3a7c980c6106562f311b6fbb2e0097d3a440c3cb).

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Applied in [`6b4b6f1`](https://github.com/debezium/debezium/pull/7801/commits/6b4b6f1f3b07cca4553f695d6ce2f8d3ab75eaed).

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Applied in [`bc2df95`](https://github.com/debezium/debezium/pull/7801/commits/bc2df95e73c94e7c6bcd467e8a2df2c0dd95768f).

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Applied in [`8f0a698`](https://github.com/debezium/debezium/pull/7801/commits/8f0a698163f069c48482949dfa827c25684d50af).

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Applied in [`76b1b15`](https://github.com/debezium/debezium/pull/7801/commits/76b1b1596bffb9d368f75203cff9ed4f93c647ab).

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Yes, that captures the intended meaning. Applied in [`384198f`](https://github.com/debezium/debezium/pull/7801/commits/384198f6e94bf63efe1d7c08173070c53e49dada). Thanks!

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Applied in [`c65a21a`](https://github.com/debezium/debezium/pull/7801/commits/c65a21a920b8c345a78a3cc2e69886103b7a3f72).

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Applied in [`49ae21e`](https://github.com/debezium/debezium/pull/7801/commits/49ae21efd5fc1cc0f0fa2468ce42183bd618ce04).

documentation/modules/ROOT/pages/connectors/jdbc.adoc
minleejae: Applied in [`f8c5749`](https://github.com/debezium/debezium/pull/7801/commits/f8c574950cecc4283ec81492472c360ae5b28d15) (removed the stray space before the period).

roldanbob (approved):
Thanks, @minleejae!  LGTM!

Naros (approved):
Thanks @minleejae and @roldanbob. Applied 🎉

Naros:
@roldanbob backported these changes to 3.6, sans the changes specifically related to the 3 new dialects, as those are not present in the 3.6 JDBC sink. :rocket:

PR #7804, opened 2026-08-11
debezium/dbz#2352 Map propagated INT16 TINYINT to smallint in the MySQL JDBC sink

### LINKED ISSUE #2352: JDBC sink maps propagated SQL Server TINYINT to signed MySQL TINYINT
### Problem

SQL Server `TINYINT` has a range of `0..255`, so the SQL Server connector correctly emits it as Kafka Connect `INT16`. When `column.propagate.source.type` is enabled, the field also contains:

```text
__debezium.source.column.type=TINYINT
```

The JDBC sink prioritizes this propagated type over the Connect schema. For a MySQL target, `TinyIntType` therefore creates a signed `TINYINT` column with a range of `-128..127`. Values from `128` through `255` then fail to write.

### Reproducer

1. Produce a field equivalent to a SQL Server `TINYINT` with propagation enabled:

   ```java
   SchemaBuilder.int16()
           .parameter("__debezium.source.column.type", "TINYINT")
           .build();
   ```

2. Send the value `(short) 255` to a MySQL JDBC sink with `schema.evolution=basic`.

The sink creates `data TINYINT` and the write fails with:

```text
Out of range value for column 'data' at row 1
```

### Suggested improvement

Keep SQL Server `TINYINT` mapped to Connect `INT16`. In the MySQL JDBC dialect, a propagated `TINYINT` paired with an `INT16` schema should fall back to `SMALLINT` instead of narrowing it to signed `TINYINT`.

Add coverage for both mappings:

* `INT8` plus propagated `TINYINT` maps to MySQL `TINYINT`.
* `INT16` plus propagated `TINYINT` maps to MySQL `SMALLINT` and preserves the value `255`.

This is separate from #2349, which updates JDBC test expectations for signed MySQL `TINYINT` after #2235.

minleejae:
Fixes debezium/dbz#2352

## Description

A propagated source `TINYINT` emitted as an INT16 schema — SQL Server's unsigned 0-255 `TINYINT` — was mapped to a signed MySQL `tinyint`, so values 128-255 failed with "Out of range value". `TinyIntType` now widens an INT16 schema to `smallint`; a signed MySQL `TINYINT` (INT8) still maps to `tinyint`. The JDBC e2e pipeline expectation is updated to match.

## Testing

* `./mvnw process-sources -pl debezium-connector-jdbc -Dformat.formatter.goal=validate -Dformat.imports.goal=check`
* `./mvnw test -pl debezium-connector-jdbc -Dtest=TinyIntTypeTest`
* JDBC MySQL end-to-end and column-type-mapping ITs are covered by the JDBC CI jobs.

## PR Checklist

### DIFF
--- a/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/TinyIntType.java
+++ b/debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/TinyIntType.java
@@ -26,16 +26,21 @@ public class TinyIntType extends AbstractType {
 
     @Override
     public String getTypeName(Schema schema, boolean isKey) {
-        // A propagated source TINYINT that is emitted as an INT16 schema (e.g. SQL Server's
-        // unsigned 0-255 TINYINT) does not fit a signed MySQL tinyint (-128 to 127); widen it to
-        // smallint. This is checked before the display width, because a propagated width does not
-        // change the range of values that the column has to hold. A signed MySQL TINYINT is
-        // emitted as INT8 and is unaffected.
-        if (schema.type() == Schema.Type.INT16) {
+        final int columnSize = Integer.parseInt(getSourceColumnSize(schema).orElse("0"));
+        // A propagated source TINYINT emitted as an INT16 schema cannot fit a signed MySQL tinyint
+        // (-128 to 127) and must be widened to smallint. The tricky part is that both a MySQL
+        // BOOLEAN and SQL Server's unsigned 0-255 TINYINT reach this branch as INT16 while also
+        // carrying a propagated display width, so the presence of a width cannot tell them apart -
+        // only its value can, and ordering the width check either before or after the INT16 check
+        // breaks one of the two cases. A width of exactly 1 is MySQL's tinyint(1) BOOLEAN
+        // convention and must be preserved (the MySQL driver reports tinyint(1) as BIT), whereas
+        // any other width (e.g. SQL Server's length 3) denotes a real 0-255 value that has to be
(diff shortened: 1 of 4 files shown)


debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/TinyIntType.java
```diff
@@ -26,6 +26,12 @@ public String[] getRegistrationKeys() {
 
     @Override
     public String getTypeName(Schema schema, boolean isKey) {
+        // A propagated source TINYINT that is emitted as an INT16 schema (e.g. SQL Server's
+        // unsigned 0-255 TINYINT) does not fit a signed MySQL tinyint (-128 to 127); widen it
+        // to smallint. A signed MySQL TINYINT is emitted as INT8 and remains a tinyint.
+        if (schema.type() == Schema.Type.INT16) {
+            return "smallint";
+        }
```
Naros: Hi @minleejae I believe the failures are related to the PR, if you could take a look.

I believe you may need to place this after the `columnSize > 0` check.

Naros (changes requested):

debezium-connector-jdbc/src/main/java/io/debezium/connector/jdbc/dialect/mysql/TinyIntType.java
minleejae: Thanks! Applied in https://github.com/debezium/debezium/pull/7804/changes/8ecc176164348a5fded7740c6ca4648629608b17

Naros (approved):

Naros:

PR #7805, opened 2026-08-11
debezium/dbz#2406 Use `OffsetActivityMonitor` across connectors

Naros:
Fixes debezium/dbz#2406

## Description
The Oracle connector for LogMiner has had a longstanding feature called `OffsetActivityMonitor` utility that tracks when the offsets have not progressed over a window, logging a warning when a specific threshold is reached. 

This PR's goal is to centralize this logic in `debezium-connector-common` with a unified service, service provider, and contract that every connector can choose to implement.

The design is based on `internal.offset.activity.monitor.interval.ms`, which defaults to 10 minutes. The service is pulsed on each stream loop iteration, and once the elapsed window has expired, the monitor implementation is invoked if supplied. If the connector does not supply a monitoring implementation, the default pulse is a no-op.

For Oracle specifically, this changes the implementation from being iteration-based to the new centralized time-based approach, so that it provides better granular control by users.

## Questions

1. Should the configuration property be internal or should it be documented?
2. Does anyone see any refactoring options to minimize duplication that I may have missed?

## Non-core implementations

* https://github.com/debezium/debezium-connector-cockroachdb/pull/76
* https://github.com/debezium/debezium-connector-db2/pull/239
* https://github.com/debezium/debezium-connector-ibmi/pull/109
* https://github.com/debezium/debezium-connector-ingres/pull/22
* https://github.com/debezium/debezium-connector-informix/pull/189
* https://github.com/debezium/debezium-connector-milvus/pull/9
* https://github.com/debezium/debezium-connector-spanner/pull/168
* https://github.com/debezium/debezium-connector-sqlite/pull/9
* https://github.com/debezium/debezium-connector-tidb/pull/4
* https://github.com/debezium/debezium-connector-vitess/pull/294
* https://github.com/debezium/debezium-connector-yashandb/pull/19

## Connectors not changed - currently don't use the ServiceRegistry

* Cassandra
* YugabyteDB

## PR Checklist


debezium-connector-common/src/main/java/io/debezium/pipeline/monitor/DefaultOffsetActivityMonitorService.java
```diff
@@ -0,0 +1,77 @@
...
+    private OffsetActivityMonitor<Partition, OffsetContext> monitor;
+
+    public DefaultOffsetActivityMonitorService(Duration checkInterval) {
+        this(checkInterval, Clock.SYSTEM);
+    }
+
+    /**
+     * Creates a disabled service instance whose pulses are no-ops.
+     */
+    public static DefaultOffsetActivityMonitorService disabled() {
```
mfvitale: Maybe we can have a static field called `NO_OP`?

debezium-connector-common/src/main/java/io/debezium/pipeline/monitor/OffsetActivityMonitor.java
```diff
@@ -0,0 +1,27 @@
...
+ */
+@Incubating
+public interface OffsetActivityMonitor<P extends Partition, O extends OffsetContext> {
+    /**
+     * Checks for stale offsets.
+     *
+     * @param partition the partition currently being streamed, should not be {@code null}
+     * @param offsetContext the partition's offsets, should not be {@code null}
+     */
+    void checkForStaleOffsets(P partition, O offsetContext);
```
mfvitale: @Naros What if we let this return a result object, with a `String` message for now, and then let in the service `pulse` method do the action? I could see not only a log but maybe a notification sent too?

Naros: Sure, that makes sense. I guess something like this should be sufficient.
```java
public record StaleOffsetResult(boolean stale, String message) {
}
```
Then the activity monitor would fully serialize the `String` using `formatted`. WDYT @mfvitale ?

mfvitale: Or maybe something like this?

```java
public sealed interface StaleOffsetResult
        permits StaleOffsetResult.Stale, StaleOffsetResult.Fresh {

    record Stale(String message) implements StaleOffsetResult {
        public Stale {
            Objects.requireNonNull(message, "message must not be null");
        }
    }

    record Fresh() implements StaleOffsetResult {}

    static StaleOffsetResult stale(String message) {
        return new Stale(message);
    }

    static StaleOffsetResult fresh() {
        return new Fresh();
    }
}
```

kmos: I am not sure that we can use `sealed class` in connectors (java 21)

kmos: but I like the idea

Naros: If I am not mistaken, they were added in 17 thankfully, so I believe we go :partying_face:

PR #7808, opened 2026-08-12
debezium/dbz#2041 Register the delayed PostgreSQL types in dependency order

sundongkim-dev:
Fixes debezium/dbz#2041

## Description

`TypeRegistry#prime()` collects the types that have a base type (`typbasetype`) or an element type (`typelem`) and registers them after the types that have neither, in the order in which they were read from the database. That order is whatever order the catalog is scanned in, not a dependency order, so whenever a delayed type's dependency is registered later, `PostgresType.Builder#build()` resolves it through `TypeRegistry#get(oid)`, which misses the cache and issues an individual `SQL_OID_LOOKUP` query.

On a database with many custom types that turns startup into thousands of sequential round trips, which is what the issue reports as a 30+ minute startup.

The delayed types are now registered dependency-first. The dependencies are walked with an explicit stack, so that long chains of types stay off the JVM stack, and a builder is removed from the pending map when it is first visited, so that a cyclic catalog cannot loop. A dependency that is not part of the result set at all — for instance when priming is filtered to a set of schemas — is still resolved with a single lookup, as before.

Reproducer and effect, on PostgreSQL 15 with 3000 domains plus 3000 domains derived from them (12611 rows in `pg_type`). A plain `ALTER DOMAIN ... SET NOT NULL` on the base domains is enough to have the catalog return the dependent types first:

| | `SQL_OID_LOOKUP` queries per start | priming |
|---|---|---|
| before | 6000 | 392-564 ms |
| after | 0 | 23-134 ms |

The queries were counted on the server with `log_statement = all`. The timings are from a connection to localhost; with a 20 ms round trip those 6000 queries alone are two minutes of startup for this database, and the issue reports databases an order of magnitude larger.

Two side effects worth naming:

- Before this change a dependency that was resolved out of order was built twice — once by `resolveUnknownType`, once when the outer loop reached its own builder — leaving two instances for one OID. Now it is built once.
- The unqualified type name aliases registered in `addType` keep the first type of a given name, so with same-named types in different schemas the alias can now point to a different one. That tie was already decided by the physical scan order, i.e. it was never stable; this test relies on exactly that instability to reproduce the issue.

This picks up the analysis and the review discussion from debezium/debezium#7462 (closed for inactivity), including @Naros' suggestion to resolve the remaining types in a dependency order rather than to batch the lookups.

Tests: `ConnectionIT#shouldPrimeDependentTypesWithoutIndividualLookups` creates a domain and a domain derived from it, inverts the order in which the catalog returns them, and asserts that priming does not look up the base domain by OID. It fails on `main` and passes with the change. `ConnectionIT`, `DomainTypesIT`, `PostgresEnumIT`, `PostgresSchemaIT`, `RecordsSnapshotProducerIT` and `RecordsStreamProducerIT` pass (193 tests).

## PR Checklist

### DIFF
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/ConnectionIT.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/ConnectionIT.java
@@ -20,11 +20,14 @@ import io.debezium.config.Configuration;
 import io.debezium.connector.postgresql.connection.PostgresConnection;
 import io.debezium.doc.FixFor;
 import io.debezium.jdbc.JdbcConfiguration;
+import io.debezium.junit.logging.LogInterceptor;
 import io.debezium.relational.Column;
 import io.debezium.relational.Table;
 import io.debezium.relational.TableId;
 import io.debezium.relational.Tables;
 import io.debezium.util.Testing;
+
+import ch.qos.logback.classic.Level;
 
 public class ConnectionIT implements Testing {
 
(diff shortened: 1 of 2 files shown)


debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/ConnectionIT.java
```diff
@@ -139,6 +142,65 @@ else if ("label".equals(column.name())) {
...
+                    .as("the types are no longer read with the dependent one first, so debezium/dbz#2041 is not reproduced")
+                    .containsSubsequence("dependent_domain", "base_domain");
+
+            final LogInterceptor logInterceptor = new LogInterceptor(TypeRegistry.class);
+            logInterceptor.setLoggerLevel(TypeRegistry.class, Level.TRACE);
+
+            TestHelper.getTypeRegistry();
+
+            // Priming has to register the base domain first, rather than look it up on its own
+            assertThat(logInterceptor.containsMessage("Type OID '" + baseTypeOid + "' not cached")).isFalse();
```
Naros, line 177: If `TypeRegistry#prime` is never called (someone changes how the code works in the connector), this assertion by itself would never indicate any issue, so it should be paired with something we do expect, so that we know the omission of it have merit. Consider adding before it:
```java
assertThat(logInterceptor.containsMessage("Priming type registry with database types"))
        .as("the interceptor is attached and TRACE is enabled")
        .isTrue();
```

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/ConnectionIT.java
```diff
@@ -139,6 +142,65 @@ else if ("label".equals(column.name())) {
...
+                    "DROP SCHEMA IF EXISTS dbz2041 CASCADE",
+                    "CREATE SCHEMA dbz2041",
+                    "CREATE DOMAIN dbz2041.base_domain AS varchar(50)",
+                    "CREATE DOMAIN dbz2041.dependent_domain AS dbz2041.base_domain",
+                    // Rewrites the pg_type row of the base domain, so that the types are read back with the
+                    // dependent one first, as it happens on databases where the catalog has been updated.
+                    "ALTER DOMAIN dbz2041.base_domain SET NOT NULL");
+        }
+
+        try (PostgresConnection conn = TestHelper.create()) {
```
Naros: Should we combine these two try blocks together? Do they need to be different?

Naros (commented):
Hi @sundongkim-dev I've left two minor comments about the test.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/ConnectionIT.java
sundongkim-dev: Added, thanks.

debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/ConnectionIT.java
sundongkim-dev: No reason to keep them apart — combined. That also brings the DDL under the existing `finally`.

Naros (approved):

Naros:
Applied @sundongkim-dev :tada:

PR #7811, opened 2026-08-12
debezium/dbz#2391 Document stale cursor recovery caveats

antonio0:
## Summary

- document that `snapshot.mode=when_needed` can re-emit existing rows after an expired cursor
- warn that deletes occurring while the connector is offline cannot be recovered by the new snapshot
- add the caveats to both the snapshot-mode table and the GC-TTL troubleshooting guidance

Companion documentation for debezium/debezium-connector-cockroachdb#74 and debezium/dbz#2391.

## Testing

- `git diff --check`
- documentation-only change; the repository's documentation workflow validates the published content


Naros (approved):

Naros:
Applied, thanks @antonio0 :tada:

PR #7812, opened 2026-08-12
debezium/dbz#2392 Document startup-frozen table discovery

antonio0:
## Summary

- document that CockroachDB table discovery occurs when the connector task starts
- explain that tables created later are not added automatically to the running changefeed
- document the restart-and-reissue workflow for incremental snapshot signals
- clarify that the behavior applies to both Kafka and sinkless delivery modes

Companion documentation for debezium/debezium-connector-cockroachdb#75 and debezium/dbz#2392.

## Testing

- `git diff --check`
- documentation-only change; the repository's documentation workflow validates the published content


Naros (approved):

Naros:
Thanks @antonio0 applied! :tada:

PR #7815, opened 2026-08-13
debezium/dbz#2418 Resolve the unchanged toast placeholder of an array from its element schema

sundongkim-dev:
Fixes debezium/dbz#2418

## Description

An unchanged TOASTed array column is emitted with the plain string `unavailable.value.placeholder`, which does not fit the field's `ARRAY` schema:

```
org.apache.kafka.connect.errors.DataException: Invalid Java object for schema with type ARRAY: class java.lang.String for field: "numeric_array"
```

With the default `event.converting.failure.handling.mode=warn` this only reaches the log and the column is dropped from the event, so a sink cannot tell it apart from a column that was never captured.

`UnchangedToastedReplicationMessageColumn#setUnchangedToastValue` picks the placeholder by matching the column's type name against a fixed list, and only `text[]`, `varchar[]`, `bpchar[]`, `json[]`, `jsonb[]`, `bytea[]`, `int4[]`, `date[]`, `int8[]` and `uuid[]` are on it. Everything else falls back to `UNCHANGED_TOAST_VALUE`, which `UnchangedToastedPlaceholder` maps to a `String`. `convertArray` delivers nothing for a marker value, so `handleUnknownData` returns that string for an `ARRAY` field.

That list has grown one element type at a time (DBZ-4941, DBZ-5944, DBZ-5935, DBZ-5936, DBZ-6122, DBZ-6720, DBZ-8416) and cannot be completed by enumeration — enum arrays, domain arrays and arrays of user-defined types have no type name that can be known up front.

The placeholder is now resolved from the array's element schema, which `convertArray` already has. `UnchangedToastedPlaceholder#getArrayValue` builds it for `STRING` (including `Uuid`), `BYTES` (including `Decimal`), `INT16`/`INT32`/`INT64`, `FLOAT32`/`FLOAT64` and `VariableScaleDecimal`, so the type name list is no longer needed and is removed along with the per-type marker constants.

The existing representations are kept unchanged: text-like elements carry the placeholder as a single element, numeric elements carry one element per placeholder byte, exactly as the `integer[]` and `bigint[]` placeholders have always been built.

**Scope.** Element types whose schema is a `STRUCT` keep the current behaviour, except `VariableScaleDecimal` where the numeric representation is obvious. A placeholder for geometry and the other `Struct` types is the separate design question in debezium/dbz#1601. `boolean[]` also keeps the current behaviour: every placeholder byte is non-zero, so its placeholder would be an all-`true` list that no placeholder choice can make distinguishable from a real value (see the review discussion); it stays dropped until debezium/dbz#1601 settles a representation. `hstore` is not an array and is untouched.

This is the part of debezium/dbz#680 that is still open — the `text[]` case in that issue's description was fixed by DBZ-4941, while the reporter's follow-up comment ("this issue happens for any array type") was not.

**Re-selection.** A dropped field left the value null, which `reselect.null.values` re-selected; a placeholder the post-processor does not recognize is re-selected by neither branch. `isUnavailableArrayValueHolder` knew only `INT32` and `INT64`, so it now covers the element types the connector emits.

Not covered: `uuid[]`, whose placeholder the post-processor never matched, before this change as well; and the columns that are still dropped (`boolean[]`, temporal arrays under `time.precision.mode=connect`, whose Connect logical element schemas expect `java.util.Date`), which keep arriving as null and are repaired by `reselect.null.values` as on `main`.

**Docs.** The TOAST section of `postgresql.adoc` now describes the placeholder representation per array element type, the `boolean[]` and `time.precision.mode=connect` omissions, and the `uuid[]` re-selection gap.

**Tests.** `RecordsStreamProducerIT#shouldHandleToastedArrayColumnsOfAnyElementType` covers `numeric[]`, `numeric(10,2)[]`, `float8[]`, `real[]`, `smallint[]`, `inet[]` and `timestamp[]` in one table, which exercises every branch of `getArrayValue` together with the existing `text[]`, `bytea[]`, `int4[]`, `date[]`, `int8[]` and `uuid[]` tests; `bool[]` is asserted to stay on the existing path (dropped from the event). Without the change all placeholder-carrying columns fail with the `DataException` above.

`PostgresReselectColumnsProcessorIT#testToastColumnArrayReselectedForEveryElementType` covers the re-selection of `numeric[]`, `numeric(10,2)[]`, `float8[]`, `real[]`, `integer[]` and `smallint[]` via the placeholder branch and `bool[]` via the null branch; reverting only the post-processor makes it fail.

Verified with the full `RecordsStreamProducerIT` on pgoutput (131 tests) and the 21 toasted-column tests on both pgoutput and decoderbufs, the full `PostgresReselectColumnsProcessorIT`, plus the unit tests of both modules.

## PR Checklist

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/UnchangedToastedPlaceholder.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/UnchangedToastedPlaceholder.java
@@ -5,13 +5,20 @@
  */
 package io.debezium.connector.postgresql;
 
+import java.math.BigDecimal;
 import java.util.ArrayList;
-import java.util.Arrays;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.UUID;
+import java.util.function.IntFunction;
+
+import org.apache.kafka.connect.data.Decimal;
+import org.apache.kafka.connect.data.Schema;
+
+import io.debezium.data.Uuid;
+import io.debezium.data.VariableScaleDecimal;
 
 /**
  * Helper that returns placeholder values for unchanged toasted columns.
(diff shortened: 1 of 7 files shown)


debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/UnchangedToastedPlaceholder.java
```diff
@@ -40,27 +47,63 @@ public UnchangedToastedPlaceholder(PostgresConnectorConfig connectorConfig) {
...
+     * Returns the placeholder for an array column, expressed in the array's element type.
+     * <p>
+     * Text-like elements carry the placeholder as a single element, every other element type carries one
+     * element per placeholder byte, which is how the {@code integer[]} and {@code bigint[]} placeholders
+     * have always been built. An empty result means the element type has no placeholder representation
+     * yet, in which case the caller keeps the existing behaviour.
+     *
+     * @param elementSchema schema of the array's elements; never null
+     */
+    public Optional<List<Object>> getArrayValue(Schema elementSchema) {
```
Naros, line 72: One area where this changes the landscape is with the ReselectColumnsPostProcessor. 

Previously, the RCPP would never see these columns because the conversion failed and the field was dropped. Now, this changes the failure from something visible to something entirely quiet.

I believe PostgresReselectColumnsPostProcessor already handles `text[]` reselection in its tests, so these adjustments deserve their own tests and the RCPP should be adjusted to support them.

Naros (changes requested):

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/UnchangedToastedPlaceholder.java
sundongkim-dev: Good catch — worse than quiet: the dropped field used to be re-selected via `reselect.null.values`, so the placeholder replaced a working repair with none. `isUnavailableArrayValueHolder` now covers the element types the connector emits; the new test fails if only the post-processor is reverted.

`uuid[]` stays uncovered — it never was, and matching it needs a second convention there. Happy to follow up with a shared helper if you'd rather the mapping live in one place.

Naros: Hi @sundongkim-dev I do think having a shared helper makes sense to minimize this moving forward. If you have the cycles to come up with a desgn, we'd welcome an issue to discuss strategy/approach.

sundongkim-dev: Opened debezium/dbz#2495 to discuss the shared-helper design. This PR keeps the mirrored mapping so it stays self-contained; the consolidation lands there.

sundongkim-dev:
@Naros ready for another look. Rebased on current main and re-ran CI.

The changes since your review are the element-type cases in `ReselectColumnsPostProcessor#isUnavailableArrayValueHolder` and the new `testToastColumnArrayReselectedForEveryElementType`, so a toasted array of an element type the placeholder list never covered is still re-selected for users on `reselect.null.values`. Your changes-requested review is still open on the PR.

The two remaining red jobs each fail on a single unrelated test (`ConnectionIT` scan-order guard on PG 18, `ReadOnlyIncrementalSnapshotIT` on Apicurio); the toast and reselect suites are green on both PG jobs.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/UnchangedToastedPlaceholder.java
```diff
@@ -40,27 +47,67 @@ public UnchangedToastedPlaceholder(PostgresConnectorConfig connectorConfig) {
...
+            case INT32:
+                return Optional.of(placeholderBytesAs(b -> b));
+            case INT64:
+                return Optional.of(placeholderBytesAs(b -> (long) b));
+            case FLOAT32:
+                return Optional.of(placeholderBytesAs(b -> (float) b));
+            case FLOAT64:
+                return Optional.of(placeholderBytesAs(b -> (double) b));
+            case BOOLEAN:
+                return Optional.of(placeholderBytesAs(b -> b != 0));
```
Naros, line 92: A question on the `boolean[]` representation: every byte of the placeholder is non-zero, so `b -> b != 0` yields a list of all `true` values. With the default `__debezium_unavailable_value` that is 28 `true` elements, which is a fairly plausible real value for a `boolean[]` column.

For the reselect post-processor a collision only costs a redundant re-select, but a downstream consumer has no way to tell a genuine 28-element all-true array apart from an unchanged TOAST value. The numeric element types have the same theoretical collision, but it is far less likely there.

Do you think this is acceptable as is, or would it be safer to leave `boolean[]` on the existing path (returning `Optional.empty()`) until debezium/dbz#2495 or debezium/dbz#1601 settles on a representation that cannot be mistaken for data?

Naros:
Looks like a comment of mine was dropped:
> One question on the user-visible side: columns that previously disappeared from the event (with a WARN in the log) will now arrive as placeholder arrays, one element per placeholder byte for the numeric, boolean and decimal element types. That is the right outcome, but it is a behavior change that sink authors and consumers will notice.
> 
> Could we add a short note to the TOAST section of postgresql.adoc describing which array element types get which placeholder representation, along with the gaps you already called out in the description (uuid[] not being re-selected, and temporal arrays under time.precision.mode=connect still being dropped)?
> 
> I left one inline question about the boolean[] representation below.

debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/UnchangedToastedPlaceholder.java
sundongkim-dev: Agreed — and for `boolean[]` the collision is not even placeholder-dependent: every non-zero byte maps to `true`, so any configured placeholder yields the same all-`true` list. The numeric collisions can be steered away from by the placeholder choice; this one cannot.

Reverted `boolean[]` to the existing path: the column is dropped as on `main` and still repaired through `reselect.null.values`, which the tests now pin. A representation that cannot be mistaken for data stays with debezium/dbz#1601.

sundongkim-dev:
Added the note to the TOAST section of `postgresql.adoc`: the placeholder representation per array element type, the `boolean[]` and `time.precision.mode=connect` omissions, and the `uuid[]` re-selection gap. `boolean[]` is reverted to the existing path per the inline thread, and the PR description is updated to match.

PR #7826, opened 2026-08-15
debezium/dbz#2435 Introduce a RocksDB cache for ReselectColumnsPostProcessor

### LINKED ISSUE #2435: Introduce RocksDB cache for `ReselectColumnsPostProcessor`

The `ReselectColumnsPostProcessor` provides a mechanism to requery and replace unavailable value placeholders in events produced by TOASTed or LOB columns that the database doesn't capture unless the column actually changes. 

By default, it always reselects values from the data source; however, an existing JVM heap-based backend stores column values in a bounded fashion to minimize queries to the database, as added in #2002.

This improvement aims to introduce a standard contract that any storage solution can implement to cache reselected column values in persistent storage, shipping a RocksDB implementation by default.

Naros:
Fixes debezium/dbz#2435

## Description
* Standardizes the ReselectColumnCache contract
* Implements a ReselectValueSerde for value (de)serialization
* Adds a RockDB reselect columns cache implementation

## PR Checklist

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/processors/reselect/cache/AbstractSerializingReselectColumnCache.java
+++ b/debezium-connector-common/src/main/java/io/debezium/processors/reselect/cache/AbstractSerializingReselectColumnCache.java
@@ -1,0 +1,215 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.processors.reselect.cache;
+
+import java.nio.charset.StandardCharsets;
+import java.util.Arrays;
+import java.util.Optional;
+import java.util.Set;
+import java.util.concurrent.ConcurrentHashMap;
+
+import org.apache.kafka.common.config.ConfigDef;
+import org.apache.kafka.connect.data.Struct;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import io.debezium.common.annotation.Incubating;
+import io.debezium.config.Configuration;
+import io.debezium.config.Field;
+
+/**
+ * Base class for {@link ReselectColumnCache} implementations backed by a byte-oriented store (embedded
+ * key/value store, remote cache, object store). Subclasses implement only the byte-level storage
+ * primitives; this class owns the mapping between the SPI's live Kafka Connect objects and stable bytes.
+ * <p>
+ * <b>Row key encoding.</b> The SPI's row identity is the event's message key {@link Struct}, whose
+ * {@code equals} contract makes the key <em>schema</em> part of the identity: a DDL change or primary-key
+ * reordering yields a different key and a natural cache miss rather than a false hit against a stale
+ * entry. To preserve those semantics in a byte-keyed store, the row key is encoded via
+ * {@link ReselectValueSerde#serializeRowKey(Struct)} as a fingerprint of the key schema followed by the
+ * key's field values written positionally (see that method for details). The per-column storage key
+ * appends a {@code 0x00} separator and the UTF-8 column name, so all columns of a row share a common byte
+ * prefix and sort adjacently in ordered stores.
+ * <p>
+ * <b>Value encoding.</b> Values are serialized via {@link ReselectValueSerde}: a versioned envelope
+ * carrying the write timestamp (for read-side TTL enforcement) and a type-tagged, value-only payload.
+ * Values whose type the serde does not support are skipped with a one-time warning; the column is simply
+ * re-queried on a later miss. Undecodable stored bytes (e.g. after a format change) are deleted and
+ * treated as a miss.
+ * <p>
+ * <b>TTL.</b> {@code reselect.cache.ttl.ms} defaults to {@code 0} (no expiration) here: persistence is
+ * the point of a durable backend, and correctness never depends on the TTL because the post-processor
+ * refreshes entries on modification. This deliberately differs from {@link MemoryReselectColumnCache},
+ * whose TTL defaults to ten minutes to bound heap retention. When a positive TTL is configured, expiry is
+ * enforced strictly at read time against the envelope's write timestamp (expired entries are deleted and
+ * reported as misses); subclasses may additionally reclaim expired data in storage.
+ * <p>
+ * Storage primitives are invoked concurrently (streaming and snapshot threads) and must be thread-safe.
+ *
+ * @author Chris Cranford
+ */
+@Incubating
+public abstract class AbstractSerializingReselectColumnCache implements ReselectColumnCache {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractSerializingReselectColumnCache.class);
+
+    public static final Field TTL_MS = Field.create("reselect.cache.ttl.ms")
+            .withDisplayName("Reselect cache TTL")
+            .withType(ConfigDef.Type.LONG)
+            .withDefault(0L)
+            .withWidth(ConfigDef.Width.SHORT)
+            .withImportance(ConfigDef.Importance.LOW)
+            .withDescription("Time-to-live in milliseconds for cached values; 0 disables expiration so "
+                    + "entries persist until overwritten or invalidated.");
+
+    private static final byte KEY_COLUMN_SEPARATOR = 0x00;
+
+    private final Set<String> unsupportedValueTypes = ConcurrentHashMap.newKeySet();
+    private final Set<String> unsupportedKeyTypes = ConcurrentHashMap.newKeySet();
+
+    private ReselectValueSerde serde;
+    private long ttlMs;
+
+    @Override
+    public void configure(Configuration config) {
+        this.ttlMs = config.getLong(TTL_MS);
+        this.serde = new ReselectValueSerde();
+        configureStorage(config);
+    }
+
+    @Override
+    public RowCache forRow(Struct messageKey) {
+        final byte[] rowKey;
+        try {
+            rowKey = serde.serializeRowKey(messageKey);
+        }
+        catch (Exception e) {
+            // An unencodable key disables caching for this row only; processing must not fail because of it.
+            final String keyType = messageKey.schema().name();
+            if (unsupportedKeyTypes.add(String.valueOf(keyType))) {
+                LOGGER.warn("Unable to encode reselect cache row key for key schema '{}'; caching is skipped for such rows.", keyType, e);
+            }
+            return NoOpRowCache.INSTANCE;
+        }
+        return new SerializingRowCache(rowKey);
+    }
+
+    /**
+     * Configure the underlying storage. Invoked once from {@link #configure(Configuration)} after the
+     * common cache options have been read.
+     *
+     * @param config the connector configuration, including any storage-specific properties
+     */
+    protected abstract void configureStorage(Configuration config);
+
+    /**
+     * Return the stored bytes for the given storage key, or {@code null} if absent. Must be thread-safe.
+     */
+    protected abstract byte[] getFromStorage(byte[] key);
+
+    /**
+     * Store the given bytes under the given storage key. Must be thread-safe.
+     */
+    protected abstract void putToStorage(byte[] key, byte[] value);
+
+    /**
+     * Remove the entry for the given storage key, if any. Must be thread-safe.
+     */
+    protected abstract void removeFromStorage(byte[] key);
+
+    /**
+     * The configured time-to-live in milliseconds; {@code 0} when expiration is disabled.
+     */
+    protected long getTtlMs() {
+        return ttlMs;
+    }
+
+    /**
+     * The current wall-clock time; overridable by tests to exercise TTL behavior deterministically.
+     */
+    protected long currentTimeMillis() {
+        return System.currentTimeMillis();
+    }
+
+    private final class SerializingRowCache implements RowCache {
+
+        private final byte[] rowKey;
+
+        private SerializingRowCache(byte[] rowKey) {
+            this.rowKey = rowKey;
+        }
+
+        @Override
+        public Optional<Hit> get(String column) {
+            final byte[] storageKey = storageKey(column);
+            final byte[] stored = getFromStorage(storageKey);
+            if (stored == null) {
+                return Optional.empty();
+            }
+            final ReselectValueSerde.DeserializedValue deserialized;
+            try {
+                deserialized = serde.deserialize(stored);
+            }
+            catch (Exception e) {
+                LOGGER.warn("Discarding undecodable reselect cache entry for column '{}'.", column, e);
+                removeFromStorage(storageKey);
+                return Optional.empty();
+            }
+            if (ttlMs > 0 && currentTimeMillis() - deserialized.timestampMs() >= ttlMs) {
+                removeFromStorage(storageKey);
+                return Optional.empty();
+            }
+            return Optional.of(new Hit(deserialized.value()));
+        }
+
+        @Override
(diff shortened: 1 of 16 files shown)


mfvitale (approved):

debezium-connector-common/src/main/java/io/debezium/processors/reselect/cache/ReselectValueSerde.java
```diff
@@ -0,0 +1,517 @@
...
+ * <p>
+ * Row keys use a separate, compare-only encoding produced by {@link #serializeRowKey(Struct)}: keys are
+ * pure identity and are never deserialized, so nested structs are written positionally with no schema
+ * detail, and a fingerprint of the key schema pins the shape (see that method for details).
+ *
+ * @author Chris Cranford
+ */
+@Incubating
+@ThreadSafe
+public final class ReselectValueSerde {
```
jpechane: @Naros I can envision `Struct` serialization/deserialization to be superuseful in other (future) memory caching/optimizations. Would it make sense to move it to a common library - maybe into util?

debezium-connector-common/src/main/java/io/debezium/serde/ConnectValueSerde.java
```diff
@@ -0,0 +1,517 @@
...
+     *
+     * @param value the value to serialize; may be null
+     * @param timestampMs the wall-clock time of the write, stored for read-side TTL enforcement
+     * @return the serialized bytes
+     * @throws DebeziumException if the value's type is not supported by the encoding
+     */
+    public byte[] serialize(Object value, long timestampMs) {
+        try (ByteArrayOutputStream baos = new ByteArrayOutputStream();
+                DataOutputStream dos = new DataOutputStream(baos)) {
+            dos.writeByte(FORMAT_VERSION);
```
jpechane, line 163: @Naros Thinking about that I'd add some magical sequence at the beginning to denote this is Debezium expected format and version add only after it. Just in case.

debezium-connector-common/src/main/java/io/debezium/processors/reselect/cache/ReselectValueSerde.java
```diff
@@ -0,0 +1,517 @@
...
+            return baos.toByteArray();
+        }
+        catch (IOException e) {
+            throw new DebeziumException("Failed to serialize reselect cache row key", e);
+        }
+    }
+
+    private void writeTaggedValue(DataOutputStream dos, Object value) throws IOException {
+        if (value == null) {
+            dos.writeByte(TAG_NULL);
```
jpechane: Do we really want to take this approach? We are losing datattype here. Would not it be better to keep the tag type? How about using the tag but dedicate one bit as nul/not-null flag?

debezium-connector-common/src/main/java/io/debezium/processors/reselect/cache/AbstractSerializingReselectColumnCache.java
```diff
@@ -0,0 +1,217 @@
...
+                serialized = serde.serialize(value, schema, currentTimeMillis());
+            }
+            catch (Exception e) {
+                final String valueType = value != null ? value.getClass().getName() : "null";
+                if (unsupportedValueTypes.add(valueType)) {
+                    LOGGER.warn("Values of type '{}' cannot be serialized and will not be cached; they will be re-queried on demand.", valueType, e);
+                }
+                return;
+            }
+            putToStorage(storageKey(column), serialized);
```
Naros: Right now, `putToStorage`, along with `getFromStorage` and `removeFromStorage`, are not wrapped in any sort of try-catch block, and any transient disk or IO operation inside these fails the task, which in turn bypasses the `reselect.error.handling.mode` configuration.

Either these storage methods should be treated as a simple cache miss, falling back to requery from the source, or they should be governed by the reselect error handling mode.

debezium-connector-common/src/main/java/io/debezium/processors/reselect/cache/AbstractSerializingReselectColumnCache.java
```diff
@@ -0,0 +1,217 @@
...
+ * Storage primitives are invoked concurrently (streaming and snapshot threads) and must be thread-safe.
+ *
+ * @author Chris Cranford
+ */
+@Incubating
+public abstract class AbstractSerializingReselectColumnCache implements ReselectColumnCache {
+
+    private static final Logger LOGGER = LoggerFactory.getLogger(AbstractSerializingReselectColumnCache.class);
+
+    public static final Field TTL_MS = Field.create("reselect.cache.ttl.ms")
```
Naros, line 66: Switching from `RocksDB` to `TtlDB` based on a TTL value will introduce undecodable bytes because they handle serialization very differently. @jpechane would a docs warning about this be sufficient?

mfvitale:
@Naros LGMT.

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -68,6 +69,137 @@ After you define a custom key, set the xref:reselect-columns-post-processor-prop
 Setting this option enables the post processor to use the specified event key fields as selection criteria in lieu of a primary key column.
 Be sure to test the configuration to ensure that the reselection query provides the expected results.
 
+// Type: concept
+// ModuleID: use-of-value-caching-with-the-debezium-reselect-columns-post-processor
+// Title: Use of value caching with the {prodname} `ReselectColumnsPostProcessor`
+[id="value-caching"]
+== Value caching
```
roldanbob, line 76: @naros Should this be conditionalized as community content? I assume that is certainly the case for the RocksDB cache section.
For any portions that we're not exposing downstream, we should remove the modularization comments.

Naros: Done @roldanbob

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -68,6 +69,139 @@ After you define a custom key, set the xref:reselect-columns-post-processor-prop
...
+
+By default, the post processor queries the database for every event that requires reselection.
+For workloads in which the same rows are reselected repeatedly (for example, TOAST or LOB columns that the database re-emits as placeholders on every unrelated update of a row), you can enable an optional cache of reselected values to reduce the number of database round-trips.
+
+Enable caching by setting `reselect.cache.enabled` to `true`.
+Values are cached per row and per column, keyed by the event's message key, so events that touch different placeholder subsets of the same row reuse each other's results.
```
roldanbob:
```suggestion
Reselected values are cached per row and per column.
The cache key is the event's message key, which uniquely identifies the row. 
As a result, if two events affect different columns of a row, they share the same cache entry.
In this way, events that affect different columns in a row can reuse each other's results.
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -68,6 +69,139 @@ After you define a custom key, set the xref:reselect-columns-post-processor-prop
...
+Before the post processor queries the database, it checks the cache for each column that requires reselection, and only queries the columns that are not present in the cache.
```
roldanbob:
```suggestion
Reselected values are cached per row and per column, using the event's message key as the cache key.
Before querying the database, the post processor checks the cache for each column that requires reselection, and only queries columns that are not already cached.
Eligible columns whose values are present in the event record are also written to the cache, even if no reselection is needed, so a later event that omits the same column can be served from the cache without querying the database again.
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -68,6 +69,139 @@ After you define a custom key, set the xref:reselect-columns-post-processor-prop
...
 Be sure to test the configuration to ensure that the reselection query provides the expected results.
 
+// Type: concept
+// ModuleID: use-of-value-caching-with-the-debezium-reselect-columns-post-processor
+// Title: Use of value caching with the {prodname} `ReselectColumnsPostProcessor`
+[id="value-caching"]
+== Value caching
+
+By default, the post processor queries the database for every event that requires reselection.
+For workloads in which the same rows are reselected repeatedly (for example, TOAST or LOB columns that the database re-emits as placeholders on every unrelated update of a row), you can enable an optional cache of reselected values to reduce the number of database round-trips.
```
roldanbob:
```suggestion
For workloads in which the same rows are reselected repeatedly (for example, TOAST or LOB columns that {prodname} emits as placeholders on every unrelated row update), you can enable an optional cache of reselected values to reduce the number of database queries.
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -68,6 +69,139 @@ After you define a custom key, set the xref:reselect-columns-post-processor-prop
...
+
+Enable caching by setting `reselect.cache.enabled` to `true`.
+Values are cached per row and per column, keyed by the event's message key, so events that touch different placeholder subsets of the same row reuse each other's results.
+Before the post processor queries the database, it checks the cache for each column that requires reselection, and only queries the columns that are not present in the cache.
+
+The cache is refreshed on modification: any event that carries a real (non-placeholder) value for an eligible column updates that column's cache entry with the current value.
+As a result, a later event that does not modify the column can be served from the cache, and a cached value is never served after the column has changed.
+Correctness therefore never depends on a cache entry expiring.
+
+Because the row identity is the message key, including its schema, a change to the key's schema (for example, a DDL change or a primary-key reordering) yields a different cache key and a natural cache miss rather than a stale hit.
```
roldanbob:
```suggestion
Because the row identity is the message key, including its schema, a change to the key's schema (for example, a DDL change or a primary-key reordering) yields a different cache key, so the post processor queries the database for the new key rather than returning a stale cached value.
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -68,6 +69,139 @@ After you define a custom key, set the xref:reselect-columns-post-processor-prop
...
+By default, the post processor queries the database for every event that requires reselection.
+For workloads in which the same rows are reselected repeatedly (for example, TOAST or LOB columns that the database re-emits as placeholders on every unrelated update of a row), you can enable an optional cache of reselected values to reduce the number of database round-trips.
+
+Enable caching by setting `reselect.cache.enabled` to `true`.
+Values are cached per row and per column, keyed by the event's message key, so events that touch different placeholder subsets of the same row reuse each other's results.
+Before the post processor queries the database, it checks the cache for each column that requires reselection, and only queries the columns that are not present in the cache.
+
+The cache is refreshed on modification: any event that carries a real (non-placeholder) value for an eligible column updates that column's cache entry with the current value.
+As a result, a later event that does not modify the column can be served from the cache, and a cached value is never served after the column has changed.
+Correctness therefore never depends on a cache entry expiring.
```
roldanbob:
```suggestion
```
The preceding sentences already communicate that the cache stays current, so this comes across as a bit pedagogical.

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -68,6 +69,139 @@ After you define a custom key, set the xref:reselect-columns-post-processor-prop
...
+
+When the limit is reached, the least-recently-used column entries for that row are evicted.
+This bound guards against wide tables (for example, tables with 500-1000 columns) growing the per-row inner map without limit.
+
+|===
+
+ifdef::community[]
+// Type: concept
+// ModuleID: debezium-reselect-columns-post-processor-rocksdb-cache
+// Title: {prodname} `ReselectColumnsPostProcessor` RocksDB cache
```
roldanbob:
```suggestion
```
Remove modularization comments. The Nebel tooling sometimes picks these up, even when content is designated for community use and that can result in downstream errors.

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -68,6 +69,139 @@ After you define a custom key, set the xref:reselect-columns-post-processor-prop
...
+[id="value-caching-rocksdb"]
+=== RocksDB cache
+
+The `io.debezium.storage.rocksdb.reselect.RocksDbReselectColumnCache` implementation persists cached values to disk in an embedded RocksDB database.
+Cached values survive connector restarts, so columns that would otherwise need to be reselected again after every restart can continue to be served from the cache.
```
roldanbob:
```suggestion
Cached values persist after you a connector restarts, so columns that would otherwise need to be reselected continue to be served from the cache after a restart.
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -68,6 +69,139 @@ After you define a custom key, set the xref:reselect-columns-post-processor-prop
...
+Entries remain correct across restarts because the post processor refreshes the cache whenever a column arrives with a real value.
+
+To use the RocksDB cache, the `debezium-storage-rocksdb` artifact must be on the connector plugin path, and `reselect.cache.type` must be set to the implementation's fully-qualified class name.
+
+Storage is permanent by default: no time-to-live is applied, and the database directory is retained when the connector stops.
+If you configure a positive `reselect.cache.ttl.ms`, expired entries are treated as cache misses and removed when they are read, and RocksDB additionally reclaims expired entries from disk during compaction.
```
roldanbob:
```suggestion
If you specify a positive `reselect.cache.ttl.ms` value, expired entries are treated as cache misses and removed when they are read.
RocksDB reclaims expired entries from disk during compaction.
```

documentation/modules/ROOT/pages/post-processors/reselect-columns.adoc
```diff
@@ -163,41 +297,17 @@ When set to `FAIL`, the connector will stop processing changes if the row no lon
...
-
-|[[reselect-columns-post-processor-property-reselect-cache-max-columns-per-row]]<<reselect-columns-post-processor-property-reselect-cache-max-columns-per-row, `+post.processors.reselect.cache.max.columns.per.row+`>>
-|`200`
-|The maximum number of columns cached per row in the reselection cache.
-
-When the limit is reached, the least-recently-used column entries for that row are evicted.
-This bound guards against wide tables (for example, tables with 500–1000 columns) growing the per-row inner map without limit.
-Applies only to the default in-memory cache, when `reselect.cache.enabled` is `true`.
-
+{prodname} provides the xref:value-caching-in-memory[in-memory cache] (the default) and the xref:value-caching-rocksdb[RocksDB cache] (`io.debezium.storage.rocksdb.reselect.RocksDbReselectColumnCache`); each supports additional `reselect.cache.*` properties described in its section.
```
roldanbob:
```suggestion
ifdef::product[]

{prodname} provides the xref:value-caching-in-memory[in-memory cache].
You can set additional `reselect.cache.*` properties, as described in xref:value-caching-in-memory[In-memory cache].
endif::product[]
ifdef::community[]
{prodname} provides the xref:value-caching-in-memory[in-memory cache] (the default) and the xref:value-caching-rocksdb[RocksDB cache] (`io.debezium.storage.rocksdb.reselect.RocksDbReselectColumnCache`); each caches supports additional `reselect.cache.*` properties, as described in xref:value-caching-rocksdb[RocksDB cache].
endif::community[]
```

roldanbob (changes requested):
Suggested some changes. Let me know if you have questions. Thanks!

roldanbob (approved):
LGTM.

PR #7827, opened 2026-08-15
docs(oracle): add microseconds and nanoseconds time.precision.mode documentation

### LINKED ISSUE #1440: time.precision.mode not fully documented [DBZ-9706]
Migrated from [DBZ-9706](https://issues.redhat.com/browse/DBZ-9706)

In Debezium 3.1.0 was introduced nano seconds precision for temporal types. To keep backward compatibility, several options for {{time.precision.mode}} were added ({{isostring}}, {{microseconds}} and {{nanoseconds}}). However, it seems the documentation wasn't updated and these options are missing from the documentation.

bisbeb:
## Summary

Closes debezium/dbz#1440

The `microseconds` and `nanoseconds` values of the `time.precision.mode` configuration property were introduced in Debezium 3.1.0 (`TemporalPrecisionMode` enum) but were missing from the documentation of several connectors.

This PR adds the missing sections to the Oracle, MySQL, MariaDB, Vitess, Db2, and Informix connector docs, following the pattern established in the PostgreSQL connector documentation.

### Changes

**Oracle** (`connectors/oracle.adoc`):
- Added `time.precision.mode=microseconds` section with type mapping table
- Added `time.precision.mode=nanoseconds` section with type mapping table and range warning

**MySQL / MariaDB** (`shared-mariadb-mysql.adoc`):
- Replaced the existing IMPORTANT note with proper sections including full type mapping tables
- Added precision note: `DATETIME`/`TIME` limited to microsecond precision (fsp 0-6)
- Included `NanoTimestamp` range warning (`DATETIME` supports up to `9999-12-31`, exceeding the ~2262 limit)

**Vitess** (`connectors/vitess.adoc`):
- Added `time.precision.mode=microseconds`, `nanoseconds`, and `structured` sections
- Fixed `isostring` section: corrected table title (was `=connect`), replaced `n/a` semantic types with proper `io.debezium.time.IsoDate/IsoTime/IsoTimestamp`

**Db2** (`connectors/db2.adoc`):
- Added `time.precision.mode=microseconds` and `nanoseconds` sections
- Notable: `TIME(7)` supports tenth-of-microsecond precision, so `nanoseconds` provides genuine additional precision over `microseconds` for `TIME(7)` columns

**Informix** (`connectors/informix.adoc`):
- Added `time.precision.mode=microseconds` and `nanoseconds` sections
- Informix has only `DATE` and `DATETIME` — no `TIME` type; `INTERVAL` is not supported by the Change Stream client (confirmed by commented-out tests in `AbstractInformixDatatypesTest`)

### Type mappings summary

| Connector | microseconds | nanoseconds |
|---|---|---|
| Oracle | `DATE`/`TIMESTAMP` → `MicroTimestamp`, `INTERVAL` → `MicroDuration` | same but `NanoTimestamp`; INTERVAL stays `MicroDuration` |
| MySQL/MariaDB | `DATE` → `Date`, `TIME` → `MicroTime`, `DATETIME` → `MicroTimestamp`, `TIMESTAMP` → `ZonedTimestamp` | same but `NanoTime`/`NanoTimestamp` |
| Vitess | identical to MySQL | identical to MySQL |
| Db2 | `DATE` → `Date`, `TIME([P])` → `MicroTime`, `TIMESTAMP` → `MicroTimestamp` | same but `NanoTime`/`NanoTimestamp` |
| Informix | `DATE` → `Date`, `DATETIME` → `MicroTimestamp` | same but `NanoTimestamp` |

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/documentation/modules/ROOT/pages/connectors/db2.adoc
+++ b/documentation/modules/ROOT/pages/connectors/db2.adoc
@@ -1736,6 +1736,8 @@ The following sections describe these mappings:
 * xref:db2-time-precision-mode-adaptive[`time.precision.mode=adaptive`]
 * xref:db2-time-precision-mode-connect[`time.precision.mode=connect`]
 * xref:db2-time-precision-mode-isostring[`time.precision.mode=isostring`]
+* xref:db2-time-precision-mode-microseconds[`time.precision.mode=microseconds`]
+* xref:db2-time-precision-mode-nanoseconds[`time.precision.mode=nanoseconds`]
 * xref:db2-time-precision-mode-structured[`time.precision.mode=structured`]
 
 [[db2-time-precision-mode-adaptive]]
@@ -1840,6 +1842,78 @@ Represents time values in UTC format, according to the ISO 8601 standard, for ex
 |`io.debezium.time.IsoTimestamp`
 
 Represents timestamp values in UTC format, according to the ISO 8601 standard, for example, `2019-07-09T02:28:57.123456Z`.
+
+|===
+
+[[db2-time-precision-mode-microseconds]]
+.`time.precision.mode=microseconds`
+Set the `time.precision.mode` property to `microseconds` to configure the connector to map all temporal values to microsecond-precision integers.
+All `DATE` values are represented as the number of days since the epoch, and all `TIME` and `TIMESTAMP` values are represented as the number of microseconds since the epoch or since midnight.
+
+.Mappings when `time.precision.mode` is `microseconds`
+[cols="25%a,20%a,55%a",options="header"]
+|===
+|Db2 data type |Literal type (schema type) |Semantic type (schema name) and Notes
+
+|`DATE`
+|`INT32`
+|`io.debezium.time.Date`
+
+Represents the number of days since the epoch.
+
+|`TIME([P])`
+|`INT64`
+|`io.debezium.time.MicroTime`
+
+Represents the number of microseconds past midnight, and does not include timezone information.
+Db2 allows `P` to be in the range 0-7.
+
+|`TIMESTAMP`
+|`INT64`
+|`io.debezium.time.MicroTimestamp`
+
+Represents the number of microseconds since the epoch, and does not include timezone information.
+
+|===
+
+[[db2-time-precision-mode-nanoseconds]]
+.`time.precision.mode=nanoseconds`
+Set the `time.precision.mode` property to `nanoseconds` to configure the connector to map all temporal values to nanosecond-precision integers.
+All `DATE` values are represented as the number of days since the epoch, and all `TIME` and `TIMESTAMP` values are represented as the number of nanoseconds since the epoch or since midnight.
+
+NOTE: Db2 `TIME` columns support up to `TIME(7)` (tenth of a microsecond precision), so `nanoseconds` mode provides finer precision than `microseconds` for `TIME(7)` columns.
+
+[IMPORTANT]
+====
+Nanoseconds since the Unix epoch are stored in a signed `INT64`, restricting the representable range to approximately *`1677-09-21T00:12:43Z` to `2262-04-11T23:47:16Z`*.
+Values outside this range cannot be represented and cause the connector to throw an `IllegalArgumentException` from the value converter.
+The exception is routed through the standard `event.processing.failure.handling.mode` configuration, so operators can choose `fail` (the default), `warn`, or `skip`.
+Use `time.precision.mode=microseconds` (or `isostring`) to avoid the limitation.
+====
+
+.Mappings when `time.precision.mode` is `nanoseconds`
+[cols="25%a,20%a,55%a",options="header"]
+|===
+|Db2 data type |Literal type (schema type) |Semantic type (schema name) and Notes
+
+|`DATE`
+|`INT32`
+|`io.debezium.time.Date`
+
+Represents the number of days since the epoch.
+
+|`TIME([P])`
+|`INT64`
+|`io.debezium.time.NanoTime`
+
+Represents the number of nanoseconds past midnight, and does not include timezone information.
+Db2 allows `P` to be in the range 0-7.
+
+|`TIMESTAMP`
+|`INT64`
+|`io.debezium.time.NanoTimestamp`
+
+Represents the number of nanoseconds since the epoch, and does not include timezone information.
 
 |===
 
--- a/documentation/modules/ROOT/pages/connectors/db2.adoc
+++ b/documentation/modules/ROOT/pages/connectors/db2.adoc
@@ -1881,13 +1881,13 @@ Represents the number of microseconds since the epoch, and does not include time
 Set the `time.precision.mode` property to `nanoseconds` to configure the connector to map all temporal values to nanosecond-precision integers.
 All `DATE` values are represented as the number of days since the epoch, and all `TIME` and `TIMESTAMP` values are represented as the number of nanoseconds since the epoch or since midnight.
 
-NOTE: Db2 `TIME` columns support up to `TIME(7)` (tenth of a microsecond precision), so `nanoseconds` mode provides finer precision than `microseconds` for `TIME(7)` columns.
+NOTE: Db2 `TIMESTAMP` supports up to 12 fractional digits (`TIMESTAMP(0-12)`), so `nanoseconds` mode provides additional precision over `microseconds` for `TIMESTAMP(7)` and above. Sub-nanosecond digits beyond position 9 are truncated.
 
 [IMPORTANT]
 ====
 Nanoseconds since the Unix epoch are stored in a signed `INT64`, restricting the representable range to approximately *`1677-09-21T00:12:43Z` to `2262-04-11T23:47:16Z`*.
 Values outside this range cannot be represented and cause the connector to throw an `IllegalArgumentException` from the value converter.
-The exception is routed through the standard `event.processing.failure.handling.mode` configuration, so operators can choose `fail` (the default), `warn`, or `skip`.
+The exception is handled according to the `event.converting.failure.handling.mode` configuration: with `warn` (the default) or `skip`, the affected column value is emitted as `null` and the failure is logged; with `fail`, the connector stops.
 Use `time.precision.mode=microseconds` (or `isostring`) to avoid the limitation.
 ====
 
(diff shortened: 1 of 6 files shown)


Naros (commented):
Hi @bisbeb, I don't think this is documented for the other connectors either. Would you be willing to review and supply the changes for all connectors in this PR? What you've done for Oracle here looks good.

bisbeb:
Hi @Naros, yes I do.

documentation/modules/ROOT/pages/connectors/vitess.adoc
```diff
@@ -1229,24 +1229,134 @@ MySQL converts `TIMESTAMP` values to UTC when they are written.
 |===
 
 time.precision.mode=isostring::
-The Vitess connector uses string representations for all temporal types.
+The Vitess connector represents date, time, and datetime values as ISO-8601 formatted strings at the UTC time zone.
+Because the values are represented as strings, this mode preserves the fractional-second precision that the `connect` mode can lose.
 +
```
Naros: I'm afraid this section (the intro here plus the three `Iso*` rows below) documents behavior the Vitess connector doesn't have. `VitessValueConverter.schemaBuilder` explicitly overrides ISOSTRING:

```java
if (isTemporal(typeName) && temporalPrecisionMode.equals(TemporalPrecisionMode.ISOSTRING)) {
    return SchemaBuilder.string();
}
```

That's a plain `STRING` schema with **no semantic schema name** — not `IsoTimstamp`. Furthermore, the converter passes the raw source string through unchanged. So the values aren't ISO-8601 (`2019-07-09 02:28:57`, space-separated, no `T`, no `Z`), and they aren't normalized to UTC.

The original text ("uses string representations for all temporal types") and the `n/a` semantic-type cells were correct for this connector; the real fix in this hunk is the table title (`=connect` to `=isostring`), which is a good catch. Could we keep that fix but revert the semantic-type and format claims, perhaps adding a note that Vitess emits the raw source string rather than a normalized ISO-8601/UTC value?

documentation/modules/ROOT/pages/connectors/vitess.adoc
```diff
@@ -1229,24 +1229,134 @@ MySQL converts `TIMESTAMP` values to UTC when they are written.
...
+
+|`DATE`
+|`STRUCT`
+a|`io.debezium.time.StructuredDate` +
+Contains `year`, `month`, and `day` fields.
+
+|`TIME[(M)]`
+|`STRUCT`
+a|`io.debezium.time.StructuredDuration` +
+Contains signed `years`, `months`, `days`, `hours`, `minutes`, `seconds`, `nanos`, and optional `precision` fields.
```
Naros: The `StructuredDuration` mapping for `TIME` comes from an override in `BinlogValueConverters` that only the MySQL/MariaDB connectors have. The Vitess converter has no such override, so it inherits the core mapping from `TemporalPrecisionMode.STRUCTURED`, which is `io.debezium.time.StructuredTime`:

```suggestion
a|`io.debezium.time.StructuredTime` +
```

(The field description below may need adjusting to match `StructuredTime`'s fields as well.)

documentation/modules/ROOT/pages/connectors/oracle.adoc
```diff
@@ -2537,6 +2537,123 @@ Contains calendar and clock component fields, `offsetSeconds`, optional `zoneId`
...
+
+When the `time.precision.mode` configuration property is set to `nanoseconds`, the connector maps temporal values to nanosecond-precision integers.
+Oracle `DATE` and `TIMESTAMP` values are represented as the number of nanoseconds since the UNIX epoch.
+
+[IMPORTANT]
+====
+Because nanoseconds since the Unix epoch are stored in a signed `INT64`, the representable range of `io.debezium.time.NanoTimestamp` is approximately *`1677-09-21T00:12:43Z` to `2262-04-11T23:47:16Z`*.
+Values outside this range — for example a `9999-12-31` end-of-time sentinel stored in a `TIMESTAMP(7 - 9)` column — cannot be represented and cause the connector to throw an `IllegalArgumentException` from the value converter.
+The exception is routed through the standard `event.processing.failure.handling.mode` configuration, so operators can choose `fail` (the default), `warn`, or `skip` per their requirements.
+
```
Naros, line 2621: Value-conversion exceptions are not handled by `event.processing.failure.handling.mode`, they're caught in `TableSchemaBuilder.createValueGenerator` and governed by **`event.converting.failure.handling.mode`**, whose default is **`warn`**, not `fail`. The behavior also differs from what's described: with `warn` (default) or `skip` the column value is emitted as `null` and the failure is logged; with `fail` the connector stops with a `DebeziumException`.

Suggested replacement for the sentence:

> The exception is handled according to the `event.converting.failure.handling.mode` configuration: with `warn` (the default) or `skip`, the affected column value is emitted as `null` and the failure is logged; with `fail`, the connector stops.

The same sentence appears in D2, Informix, and Vitess, and is retained from the pre-existing note in `shared-mariadb-mysql.adoc`. The error pre-dates this PR there, but since the PR touches all these blocks, it would be great to fix them together.

documentation/modules/ROOT/pages/connectors/sqlserver.adoc
```diff
@@ -1898,6 +1898,106 @@ Contains calendar and clock component fields, `offsetSeconds`, optional `zoneId`
...
+
+|===
+
+When the `time.precision.mode` configuration property is set to `nanoseconds`, the connector maps temporal values to nanosecond-precision integers.
+All `TIME` values are represented as the number of nanoseconds past midnight, and all date-time values are represented as the number of nanoseconds past the epoch.
+
+[IMPORTANT]
+====
+Because nanoseconds since the Unix epoch are stored in a signed `INT64`, the representable range of `io.debezium.time.NanoTimestamp` is approximately *`1677-09-21T00:12:43Z` to `2262-04-11T23:47:16Z`*.
+If a column may legitimately contain values outside the representable range, declare it with `DATETIME2(0 - 6)` so that Debezium emits `Timestamp` (millisecond) or `MicroTimestamp` (microsecond) instead, or set `time.precision.mode=isostring` to capture the value as an ISO 8601 string.
```
Naros: This advice contradicts the table just below: in `nanoseconds` mode, `DATETIME2(0)`–`(7)` all map to `NanoTimestamp` regardless of the declared column precision; the precision-based `Timestamp`/`MicroTimestamp` selection only happens in `adaptive` mode. The escape hatch here is a different `time.precision.mode`, not a different column precision:

```suggestion
If a column may legitimately contain values outside the representable range, set `time.precision.mode=microseconds` or `time.precision.mode=isostring` to capture the value with lower precision or as an ISO 8601 string.
```

This block is also the only one of the six that omits the failure-handling sentence mentioned in another comment; for consistency, add the corrected sentence here too.

documentation/modules/ROOT/pages/connectors/db2.adoc
```diff
@@ -1900,6 +1902,78 @@ Represents timestamp values in UTC format, according to the ISO 8601 standard, f
...
+Represents the number of microseconds since the epoch, and does not include timezone information.
+
+|===
+
+[[db2-time-precision-mode-nanoseconds]]
+.`time.precision.mode=nanoseconds`
+Set the `time.precision.mode` property to `nanoseconds` to configure the connector to map all temporal values to nanosecond-precision integers.
+All `DATE` values are represented as the number of days since the epoch, and all `TIME` and `TIMESTAMP` values are represented as the number of nanoseconds since the epoch or since midnight.
+
+NOTE: Db2 `TIME` columns support up to `TIME(7)` (tenth of a microsecond precision), so `nanoseconds` mode provides finer precision than `microseconds` for `TIME(7)` columns.
```
Naros: Question on this one: Db2 (LUW) `TIME` doesn't support fractional seconds at all.

"`TIME(7)` (tenth of a microsecond precision)" is a SQL Server fact. The existing adaptive table in this doc has carried `TIME(0)`–`TIME(7)` rows since the original DBZ-695 documentation.

The genuinely useful Db2-specific statement would be about `TIMESTAMP`: Db2 supports `TIMESTAMP(0-12)`, so `nanoseconds` mode provides real additional precision for `TIMESTAMP(7)` and above (with sub-nanosecond digits truncated). I'd suggest replacing this NOTE with that, or dropping it, and we can deal with the pre-existing TIME(P) rows separately.

documentation/modules/ROOT/pages/connectors/vitess.adoc
```diff
@@ -1229,24 +1229,134 @@ MySQL converts `TIMESTAMP` values to UTC when they are written.
...
+time.precision.mode=nanoseconds::
+The Vitess connector maps all temporal values to nanosecond-precision integers.
+All `DATE` values are represented as the number of days since the epoch, and all `TIME` and `DATETIME` values are represented as the number of nanoseconds since the epoch or since midnight.
++
+NOTE: Vitess `DATETIME` and `TIME` columns are limited to microsecond precision (`DATETIME(0-6)`, `TIME(0-6)`), so `nanoseconds` does not provide additional precision beyond what `microseconds` already offers.
++
+[IMPORTANT]
+====
+Nanoseconds since the Unix epoch are stored in a signed `INT64`, restricting the representable range to approximately *`1677-09-21T00:12:43Z` to `2262-04-11T23:47:16Z`*.
+Values outside this range — for example a `9999-12-31` end-of-time sentinel — cannot be represented and cause the connector to throw an `IllegalArgumentException` from the value converter.
```
Naros: This is at least one instance (there are others) where I see em dashes added, which is a very common construct with AI, and one that we should strive to avoid and prefer natural punctuation (commas, semicolons, colons, and periods) to separate concerns.

Naros (changes requested):
Hi @bisbeb great work. I've left a few inline things I observed and once fixed, I think this can be merged.

documentation/modules/ROOT/pages/connectors/vitess.adoc
bisbeb: @Naros 
You're right, I'm working with AI assistance (OpenClaw and Claude). I also updated the memory for the agent. But I'm also interested in providing good quality and appreciate your feedback. This is the first opensource project I'm trying to contribute to.

Naros: No problem at all, AI gets all of us with these em dashes at some point ;-)

Naros (approved):

Naros:
Applied thanks @bisbeb :tada:

PR #7828, opened 2026-08-16
DBZ-2440 Fix PostgreSQL text defaults with escaped quotes

### LINKED ISSUE #2440: PostgreSQL connector incorrectly parses text defaults containing escaped single quotes
### Describe the bug

The PostgreSQL connector incorrectly parses column default expressions that contain escaped single quotes. PostgreSQL represents a single quote inside a string literal by doubling it, for example `'can''t'::text`.

`PostgresDefaultValueConverter` currently extracts only `can` from that expression instead of `can't`, so the Kafka Connect schema default can be wrong even though the emitted row value is correct.

### Expected behavior

A PostgreSQL default expression such as:

```sql
'can''t'::text
```

should be parsed as:

```text
can't
```

### Reproducer

A focused converter test demonstrates the issue:

```java
final Column textColumn = Column.editor().type("text").jdbcType(Types.VARCHAR).defaultValueExpression("'can''t'::text").create();
final Optional<Object> convertedValue = postgresDefaultValueConverter.parseDefaultValue(
        textColumn,
        textColumn.defaultValueExpression().orElse(null));

Assertions.assertEquals(Optional.of("can't"), convertedValue);
```

Before the fix this fails with:

```text
expected: <Optional[can't]> but was: <Optional[can]>
```

### Environment

Observed on current Debezium main while testing the PostgreSQL connector default value converter.

mathewjustin:
Fixes debezium/dbz#2440

This updates the PostgreSQL default value parser to handle PostgreSQL string literals that escape single quotes by doubling them, for example `'can''t'::text`. Previously the parser stopped at the first escaped quote and produced `can` instead of `can't`.

The change keeps the parser scoped to literal extraction and adds converter coverage for the escaped quote case.

Validation:

```
./mvnw -pl debezium-connector-postgres -Dtest=PostgresDefaultValueConverterIT#shouldParseTextDefaultValueWithEscapedQuote test
```


jpechane (approved):
@mathewjustin LGTM, thanks. Can be merged when CI passes.

mathewjustin:
Hi @jpechane

Thanks for taking a look and approviing. Well, I was checking what is happening to this pr. 

I checked the non-passing check and it seems to be:

Debezium Platform - conductor / Build Debezium Platform - Conductor. It looks like a flaky connection issue to postgres not something to do with code change itself.

jpechane:
@mathewjustin Applied, thanks!

PR #7839, opened 2026-08-17
debezium/dbz#478 Support capturing changes from multiple PDBs

### LINKED ISSUE #478: Support capturing changes from multiple PDBs with one connector [DBZ-3666]
Migrated from [DBZ-3666](https://issues.redhat.com/browse/DBZ-3666)

In an effort to continue to standardize the Oracle connector like others, we should look to introduce database include/exclude lists while deprecating the use of {{database.pdb.name}}.  

This brings about several benefits, most notably that it allows a single connector to support streaming changes across multiple pluggable databases on Oracle without needing to deploy different connectors.  By using a single connector rather than multiple, you reduce the LogMiner overhead on the system while also keeping log generation in check when wanting to track DDL changes.

I propose that these database include/exclude lists are only applicable when the database is deployed in container database mode.  When CDB is not active, these settings make no sense given that you have 1 single database then, so effectively database include/exclude lists would translate to Oracle pluggable database / container database include/exclude lists supporting things like CDB$ROOT, ORCLPDB1, etc.

While this may be somewhat straight forward with LogMiner, its important to remember that for Xstream, some of this configuration is specified when creating the XOUT server in PL/SQL and so this may have less impact overall on that adapter than LogMiner.

Naros:
Fixes debezium/dbz#478

## Description
The Oracle connector has inherently supported capturing changes from a single-tenant database or a single PDB within a multi-tenant database.

This change introduces `database.pdb.names`, a new property to specify a comma-separated list of PDB names that Debezium will use to capture changes for. When specifying `database.pdb.name`, the legacy property, the connector operates as `database.pdb.names` with a single value, with no change in current behavior.

## PR Checklist

### DIFF
--- a/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
+++ b/debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
@@ -12,7 +12,10 @@ import java.time.Duration;
 import java.time.Instant;
 import java.time.ZoneId;
 import java.util.Collection;
+import java.util.HashSet;
+import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Map;
 import java.util.Optional;
 import java.util.OptionalLong;
 import java.util.Queue;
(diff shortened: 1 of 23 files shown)


debezium-connector-oracle/src/main/java/io/debezium/connector/oracle/OracleSnapshotChangeEventSource.java
```diff
@@ -52,78 +60,169 @@ public class OracleSnapshotChangeEventSource extends RelationalSnapshotChangeEve
...
+            // A transaction cannot span containers, so each pluggable database's tables are locked on a
+            // dedicated connection whose transaction is held until the schema snapshot completes.
+            final Map<String, List<TableId>> tablesByCatalog = snapshotContext.capturedTables.stream()
+                    .collect(Collectors.groupingBy(TableId::catalog, LinkedHashMap::new, Collectors.toList()));
+            for (Map.Entry<String, List<TableId>> catalogTables : tablesByCatalog.entrySet()) {
+                final OracleConnection connection = connectionFactory.newConnection();
+                pdbLockConnections.put(catalogTables.getKey(), connection);
+                connection.setSessionToPdb(catalogTables.getKey());
+                lockTables(sourceContext, connection, catalogTables.getValue());
+            }
```
Naros, line 152: Per our Zulip discussion, there's concern that configuring many PDBs for capture in `database.pdb.names` creates a barrage of connections. 

We should first log an INFO message to the user, explaining that the connector is opening N connections (1 per PDB) where we resolve N from the number of captured PDBs. 

To minimize the quick barrage of connections, we should also consider two loops: one that opens connections with a throttle, perhaps configurable via an internal property, and a second that locks the tables after all connections are open. This prevents the throttle from impacting the lock duration.

Naros: Added documentation note as pointed out by Giovanni & Vojta.

PR #7026, opened 2026-01-20
debezium/dbz#1541 use correct timestamp key in BinlogHistoryRecordComparator

zwang28:
Fix https://github.com/debezium/dbz/issues/1541

PR #7842, opened 2026-08-18
debezium/dbz#1541 Do not treat a missing server id as a different server during history recovery

### LINKED ISSUE #1541: bug: AbstractHistoryRecordComparator incorrectly evaluate to true every time.
## Bug report

MySql connector. Debezium v3.2.4.

---

When recovering schema history from my customized `AbstractFileBasedSchemaHistory`, HistoryRecord are not correctly filtered out by position. i.e. [isPositionAtOrBefore](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/strategy/AbstractHistoryRecordComparator.java#L48) incorrectly evaluate to true every time. 

---

**Root Cause**

The root cause is the inconsistent usage of field names [ts_sec](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/MySqlOffsetContext.java#L102) and [ts_ms](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/strategy/AbstractHistoryRecordComparator.java#L104). It causes [this comparison](https://github.com/debezium/debezium/blob/4159397bd40dd5d0c8efe2b97ba61ba8eb220b58/debezium-connector-mysql/src/main/java/io/debezium/connector/mysql/strategy/AbstractHistoryRecordComparator.java#L106) to incorrectly evaluate to true every time, because ts_ms field is never set.

sundongkim-dev:
Fixes debezium/dbz#1541

## Description

Schema history recovery compares every recorded position against the restart position. That restart position is not the stored offset as written, it is regenerated from the loaded offset context (`SchemaHistory#recover(Offsets, …)` calls `getOffset()` on it), and neither `MySqlOffsetContext.Loader` nor `MariaDbOffsetContext.Loader` restores the server id or the source time. So the restart position carries neither key, while every position recorded during streaming carries a server id.

`getServerId` defaults a missing key to `0`, so those pairs took the "different servers" branch. There the comparator read `ts_ms`, although an offset only ever stores `ts_sec` (`BinlogOffsetContext#offsetUsingPosition`), so the comparison was permanently `0 <= 0`. Every streaming-recorded position was accepted regardless of the restart point, and recovery replayed DDL from after it, leaving the in-memory schema ahead of where streaming resumes.

The fix stops reading a missing server id as a different server. The absence is deliberate — there is no way to tell which primary of a topology wrote the change — so those pairs are now compared by their binlog coordinates, which is what actually filters them.

The timestamp key is corrected as well. On its own it changes nothing today, because the restart position carries no timestamp either. That is also why the earlier #7026, which swapped only the key, turned every non-GTID MySQL IT red: recorded positions have `ts_sec > 0` and the restart position reads `0`, so all streaming-recorded DDL was dropped instead. The key is still worth correcting, and it keeps the branch honest for positions that do reach it — restoring a server id on load without also restoring `ts_sec` would otherwise reproduce that failure exactly.

One consequence worth naming: when two genuinely different servers are compared and the recorded side has no server id, the comparison now falls through to coordinates from unrelated coordinate spaces. Differing base names are already handled (they warn and apply the DDL). Identical base names compare as if they were one server. This is inherent to keeping the server id absent during snapshots, and the alternative is what failed CI.

The stored offset format is unchanged, and `ts_ms` was never written to an offset, so there is nothing to migrate.

The file-based schema history rework asked for in the first comment on the issue is not part of this PR — that code is being touched by #7063.

Credit to @zwang28, who reported the issue and identified the key mismatch in #7026.

## Tests

Two tests in `BinlogSourceInfoTest`, inherited by both the MySQL and MariaDB `SourceInfoTest`. The existing `positionWith` helper sets neither `server_id` nor `ts_sec`, so no test reached this branch before.

|  | `shouldCompareTimestampsOfPositionsFromDifferentServers` | `shouldNotTreatPositionWithoutServerIdAsDifferentServer` |
| --- | --- | --- |
| `main` | fails | fails |
| key swap only | passes | fails |
| this change | passes | passes |

The second test is the one that pins the fix; the first pins the contract of the branch itself, including that equal server ids keep comparing by coordinates.

Verified against the non-GTID `mysql-ci` profile: 433 tests, 0 failures, 0 errors. Reverting only the server id guard reproduces what closed #7026 (`MySqlConnectorIT.shouldConsumeAllEventsFromDatabaseUsingSnapshot`, no records for the topic). The GTID profiles and MariaDB are unaffected by construction — every GTID branch returns before this code, and MariaDB sets a GTID set even during snapshots — which is also why they stayed green on #7026.

## PR Checklist

### DIFF
--- a/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/history/BinlogHistoryRecordComparator.java
+++ b/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/history/BinlogHistoryRecordComparator.java
@@ -110,8 +110,9 @@ public abstract class BinlogHistoryRecordComparator extends HistoryRecordCompara
             return false;
         }
 
-        // Both positions are missing GTIDs, compare servers
-        if (getServerId(recorded) != getServerId(desired)) {
+        // Both positions are missing GTIDs, compare servers. A missing server id is not a different server:
+        // snapshot offsets never carry one, as there is no way to tell which primary of a topology wrote the change.
+        if (hasServerId(recorded) && hasServerId(desired) && getServerId(recorded) != getServerId(desired)) {
             // These are from different servers.
             // Their binlog coordinates are not related, so the only thing that is possible is to compare
             // timestamps, and assume that the server timestamps can be compared.
@@ -175,6 +176,16 @@ public abstract class BinlogHistoryRecordComparator extends HistoryRecordCompara
     }
 
     /**
+     * Get whether the position carries a server unique identifier.
+     *
+     * @param document the document to inspect, should not be null
+     * @return true if the document has a server identifier, false otherwise
+     */
+    protected boolean hasServerId(Document document) {
+        return document.has(BinlogSourceInfo.SERVER_ID_KEY);
+    }
+
+    /**
      * Get the server unique identifier.
      *
      * @param document the document to inspect, should not be null
@@ -198,10 +209,10 @@ public abstract class BinlogHistoryRecordComparator extends HistoryRecordCompara
      * Get the timestamp.
      *
      * @param document the document to inspect, should not be null
-     * @return the timestamp value
+     * @return the timestamp value, in seconds
      */
     protected long getTimestamp(Document document) {
-        return document.getLong(BinlogSourceInfo.TIMESTAMP_KEY, 0);
+        return document.getLong(BinlogOffsetContext.TIMESTAMP_KEY, 0);
     }
 
     /**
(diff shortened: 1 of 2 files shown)


Naros (approved):
LGTM, just waits on CI.

vjuranek (approved):
LGTM

vjuranek:
Test failures are unrelated.

vjuranek:

PR #7844, opened 2026-08-18
debezium/dbz#2442 AsyncEmbeddedEngine SourceTask.stop() is never called when the ChangeConsumer is blocked at shutdown, leaving the postgresql replication slot alive

### LINKED ISSUE #2442: AsyncEmbeddedEngine SourceTask.stop() is never called when the ChangeConsumer is blocked at shutdown, leaving the postgresql replication slot alive
## Bug report

`v3.7.0.Alpha2`

---

   connector.class = io.debezium.connector.postgresql.PostgresConnector
   slot.name = abandoned_walsender_slot
   record.processing.shutdown.timeout.ms = 1000
   bootstrap.servers = localhost:9092
   slot.max.retries = 3
   record.processing.order = ORDERED
   topic.prefix = test_server
   database.sslmode = disable
   slot.retry.delay.ms = 1000
   offset.storage.file.filename = 
   record.processing.threads = 
   errors.retry.delay.initial.ms = 300
   value.converter = org.apache.kafka.connect.json.JsonConverter
   database.topic.prefix = dbserver1
   status.update.interval.ms = 100
   key.converter = org.apache.kafka.connect.json.JsonConverter
   database.user = postgres
   database.dbname = postgres
   slot.drop.on.stop = false
   offset.storage = org.apache.kafka.connect.storage.MemoryOffsetBackingStore
   offset.flush.timeout.ms = 5000
   errors.retry.delay.max.ms = 10000
   database.port = 5432
   plugin.name = pgoutput
   offset.flush.interval.ms = 60000
   internal.task.management.timeout.ms = 5000
   record.processing.with.serial.consumer = false
   errors.max.retries = -1
   database.hostname = localhost
   database.password = ********
   name = abandoned-walsender-connector
   table.include.list = zombie.test

---

`Postgres 15.X`

---

SourceTask.stop should stop all the tasks before the engine report STOPPED. An engine that reports STOPPED must not leave a live task behind.

---

**What behavior do you see?**

AsyncEmbeddedEngine.stopSourceTasks() submits the task's stop to taskService which uses the same pool that runs record polling so a consumer which is still blocked at shutdown leaves no thread for the stop to run on. After task.management.timeout.ms (40 seconds default) the engine logs "Stopping of the tasks was interrupted, shutting down immediately", the finally block's taskService.shutdownNow() now discards the queued callable and SourceTask.stop() never runs. The engine reports STOPPED anyway. The abandoned task keeps its walsender, and the slot stays active and whenerver you start the next connector it will fail with replication slot is active for PID N

---

`Yes`

---

`Yes`

---

**Steps to Reproduce**

1. Start the debezium engine with default config options for timeout
2. Use some mechanism to block the consumer (processRecords) to simulate a situation where consumer takes time to process the records for more than 40 sec (default task management timeout).
3. Call engine.close() from another thread. After the timeout the engine logs "Stopping of the tasks was interrupted, shutting down immediately" and taskService.shutdownNow() discards the queued stop callable.
4. Release the lock and let the engine shutdown.
5. Read the pg_replication_slots the slot will be still active under the abandoned tasks pid.
6. Start a second engine for the same replication slot it fails with ERROR: replication slot "walsender_slot" is active for PID.

Sample repro at https://github.com/chiragkava/debezium/blob/stop-starvation-repro/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/AbandonedWalsenderOnBlockedStopIT.java

 **Potential Fix**
Keep track of the tasks that are to be stopped in AsyncEmbeddedEngine's stopSourceTasks method and in finally block check if tasks were discarded due to timeout and close was never called then in finally block call the task.close for graceful stop

chiragkava:
Fixes debezium/dbz#2442

## Description

If the ChangeConsumer is still blocked when engine.close() is called, the stop callable never gets a chance to run . It goes to the same taskService pool that runs record polling, and the blocked consumer is holding that thread. After `task.management.timeout.ms` the engine gives up and shutdownNow() just throws the queued callable away, so                                SourceTask.stop() is never called. The engine still reports STOPPED though. Now this abandoned task keeps its walsender and the slot stays active. So whenever the next replacement connector tries to bring it up it will fail witj replication slot is active for PID N.                            
                                                                                                                    
The fix is to stop these leftover tasks directly on the shutdown thread after shutdownNow() in finally block.
                                                                                                                    
Added a test to AsyncEmbeddedEngineTest that blocks the consumer, closes the engine, and                        
checks the task actually got stopped before close() returned. It fails without the fix.  

## PR Checklist

### DIFF
--- a/debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEmbeddedEngine.java
+++ b/debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEmbeddedEngine.java
@@ -77,6 +77,7 @@ import io.debezium.spi.storage.OffsetStoreProvider;
 import io.debezium.storage.kafka.offset.KafkaConnectOffsetUtil;
 import io.debezium.util.DelayStrategy;
 import io.debezium.util.KafkaConnectUtil;
+import io.debezium.util.Threads;
 
 /**
  * Implementation of {@link DebeziumEngine} which allows to run multiple tasks in parallel and also
@@ -701,19 +702,32 @@ public final class AsyncEmbeddedEngine<R> implements DebeziumEngine<R>, AsyncEng
      * @param tasks {@link List<EngineSourceTask>} of source tasks which should be stopped.
      */
     private void stopSourceTasks(final List<EngineSourceTask> tasks) {
-        final Set<EngineSourceTask> pendingStopTasks = ConcurrentHashMap.newKeySet();
-        pendingStopTasks.addAll(tasks);
+        final ExecutorService stopTaskService = Threads.newFixedThreadPool(AsyncEmbeddedEngine.class,
+                config.getString(EmbeddedEngineConfig.ENGINE_NAME), "task-stop", Math.max(tasks.size(), 1));
+        final Set<ConnectorTaskId> tasksNotStopped = ConcurrentHashMap.newKeySet();
+        tasks.forEach(task -> tasksNotStopped.add(task.context().connectorTaskId()));
         try {
             LOGGER.debug("Stopping source connector tasks.");
-            final ExecutorCompletionService<Void> taskCompletionService = new ExecutorCompletionService(taskService);
+            final ExecutorCompletionService<Void> taskCompletionService = new ExecutorCompletionService(stopTaskService);
+            final ClassLoader originalTccl = Thread.currentThread().getContextClassLoader();
             for (EngineSourceTask task : tasks) {
                 final long commitTimeout = Configuration.from(task.context().config()).getLong(EmbeddedEngineConfig.OFFSET_COMMIT_TIMEOUT_MS);
                 taskCompletionService.submit(() -> {
-                    LOGGER.debug("Committing task's offset.");
-                    commitOffsets(task.context().offsetStorageWriter(), task.context().clock(), commitTimeout, task.connectTask());
-                    LOGGER.debug("Stopping Connect task.");
-                    if (pendingStopTasks.remove(task)) {
-                        task.connectTask().stop();
+                    LoggingContext.clear();
+                    try (LoggingContext loggingContext = LoggingContext.forTask(task.context().connectorTaskId())) {
+                        Thread.currentThread().setContextClassLoader(this.classLoader);
+                        try {
+                            LOGGER.debug("Committing task's offset.");
+                            commitOffsets(task.context().offsetStorageWriter(), task.context().clock(), commitTimeout, task.connectTask());
+                        }
+                        finally {
+                            LOGGER.debug("Stopping Connect task.");
+                            task.connectTask().stop();
+                            tasksNotStopped.remove(task.context().connectorTaskId());
+                        }
+                    }
+                    finally {
+                        Thread.currentThread().setContextClassLoader(originalTccl);
                     }
                     return null;
                 });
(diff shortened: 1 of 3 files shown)


debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEmbeddedEngine.java
```diff
@@ -745,6 +751,25 @@ private void stopSourceTasks(final List<EngineSourceTask> tasks) {
...
+    }
+
+    private void stopAbandonedTasks(final Set<EngineSourceTask> pendingStopTasks) {
+        for (EngineSourceTask task : List.copyOf(pendingStopTasks)) {
+            if (pendingStopTasks.remove(task)) {
+                LOGGER.warn("Stop of task {} was discarded together with the task service, stopping the task directly to avoid leaving it running.",
+                        task.context().connectorTaskId());
+                final long startTime = System.nanoTime();
+                try {
+                    task.connectTask().stop();
```
vjuranek: We can evetualy hand here forever.

vjuranek (commented):
Thanks for the PR @chiragkava ! My main concern with this approach is that we can eventually hang in the stop tasks forever. Won't be sufficient fix to run the shutdown tasks in their own executor pool and from user perspective adjust shutdown timeout?

chiragkava:
@vjuranek agreed that dedicated executor is the right approach. i have made the change so now the stop callables run on their own pool . But doing so introduced a few new concerns (since we moved away from the single thread executor )

1. Concurrent offset flushes. The stop thread's final commitOffsets() can now overlap the record thread's commit on the same shared OffsetStorageWriter worst case, one thread cancels the other's in flight flush.  added a synchronized in commitOffset method not so sure whether this is a clean way of doing it.
2. Skipped stop on a failed commit. If the final offset commit throws or is interrupted, stop() must still run so kept the stop inside finally block.
3. Offset store torn down under a late stop. After shutdownNow(), the shutdown waits (bounded by the same timeout) for interrupted callables to finish while the offset store is still available. Worst case close() takes 2 x task.management.timeout.ms, and only when both the final flush and the stop hang.

let me know your views for above points.

debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEmbeddedEngine.java
```diff
@@ -743,8 +763,30 @@ private void stopSourceTasks(final List<EngineSourceTask> tasks) {
...
+
+    /**
+     * Shuts down the executor which runs the stop callables. Interrupts right away the callables already got the
+     * whole {@code task.management.timeout.ms}, so whatever still runs is overdue. Then waits, bounded by the same
+     * timeout, for the interrupted callables to finish stopping their tasks while the offset store is still available.
+     */
+    private void shutdownStopTaskServiceForcefully(final ExecutorService stopTaskService) {
+        stopTaskService.shutdownNow();
+        try {
+            if (!stopTaskService.awaitTermination(config.getLong(AsyncEngineConfig.TASK_MANAGEMENT_TIMEOUT_MS), TimeUnit.MILLISECONDS)) {
```
vjuranek, line 783: This seems to be redundant, `shutdownNow()` should terminate the tasks immediately.

vjuranek (commented):
Well, obvisouly I haven't think about it very carefully when suggesting dedicated executors for shutdown - it complicates the design and makes the whole flow complicated and error-prone. I'll try to re-think it, if there is any better solution.

debezium-embedded/src/main/java/io/debezium/embedded/async/AsyncEmbeddedEngine.java
chiragkava: wont the `shutdownNow()` only send interrupt and the running callable will still keep on running until it reacts to the interrupt ?

PR #7845, opened 2026-08-18
debezium/dbz#2431 Fix read.only incremental snapshot failing on hot standby due to pg_current_xact_id()

### LINKED ISSUE #2431: read.only=true fails on PostgreSQL hot standby due to pg_current_xact_id() call
## Bug description

When `read.only=true` is configured, `PostgresReadOnlyIncrementalSnapshotChangeEventSource`
calls `pg_current_xact_id()` at the start of each incremental snapshot
(`preIncrementalSnapshotStart` → `forceNewTransactionId()`). This forces a real
transaction ID to be assigned to the connector's session.

On a PostgreSQL hot standby replica, this fails because standbys cannot assign
transaction IDs to user sessions — they are read-only at the database level.

**Note:** this failure only manifests when the connector **restarts mid-snapshot**. On restart, an in-progress incremental snapshot is resumed, which triggers `preIncrementalSnapshotStart` and the `pg_current_xact_id()` call again.

## Steps to reproduce

1. Configure the PostgreSQL connector with `read.only=true` against a hot standby (replica) server
2. Start an incremental snapshot
3. Restart the connector before the snapshot completes
4. Observe failure on restart when `pg_current_xact_id()` is called to resume the snapshot

## Expected behaviour

`read.only=true` should be usable against a hot standby replica. The feature
description says it is intended for cases where the connector "cannot write to
a signaling table" — a hot standby is the primary such use case.

## Actual behaviour

The connector fails to resume the snapshot on restart because `pg_current_xact_id()` requires the ability to
assign a transaction ID, which is not available on a standby.

## Root cause

`PostgresReadOnlyIncrementalSnapshotChangeEventSource.java` line 73–74, 97, 233–240:

```java
private static final String FORCE_NEW_TRANSACTION = "SELECT * FROM pg_current_xact_id();";
...
protected void preIncrementalSnapshotStart() {
    super.preIncrementalSnapshotStart();
    forceNewTransactionId();  // fails on hot standby
}
```

The call is needed so the connector's own transaction appears in subsequent
`pg_current_snapshot()` watermarks. An alternative approach may be needed for
true standby support (e.g. using virtual transaction IDs or `pg_current_xact_id_if_assigned()`).

## Environment

- Debezium version: 3.5.2 (confirmed unchanged through current main)
- PostgreSQL: 18.3

rick020:
## Summary

Fixes [2431](https://github.com/debezium/dbz/issues/2431)

When `read.only=true` is configured and the connector restarts mid-snapshot, `preIncrementalSnapshotStart` calls `forceNewTransactionId()` which executes `pg_current_xact_id()`. On a PostgreSQL hot standby this fails because standbys cannot assign transaction IDs to user sessions.

## Changes

- Replace the bare `pg_current_xact_id()` call in `FORCE_NEW_TRANSACTION` with a `CASE pg_is_in_recovery()` guard that returns `NULL` on a hot standby — the same pattern already used in `PostgresConnection.currentTransactionId()`
- Handle the `NULL` result gracefully in `forceNewTransactionId()` with a trace log instead of attempting to log the transaction ID

On a primary the behaviour is completely unchanged.

## Test plan

- `forceTransactionQueryIsRecoveryAware` — asserts the SQL constant contains both `pg_is_in_recovery()` and `pg_current_xact_id()`
- `forceNewTransactionIdIsNoOpOnHotStandby` — mocks a `NULL` result (standby) and verifies no exception is thrown and the standby skip message is logged
- `forceNewTransactionIdLogsTransactionOnPrimary` — mocks a real transaction ID result and verifies the primary path is unchanged

### DIFF
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresReadOnlyIncrementalSnapshotChangeEventSourceTest.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresReadOnlyIncrementalSnapshotChangeEventSourceTest.java
@@ -71,7 +71,7 @@ public class PostgresReadOnlyIncrementalSnapshotChangeEventSourceTest {
     }
 
     @Test
-    @FixFor("DBZ-2431")
+    @FixFor("debezium/dbz#2431")
     void forceTransactionQueryIsRecoveryAware() throws Exception {
         Field field = PostgresReadOnlyIncrementalSnapshotChangeEventSource.class
                 .getDeclaredField("FORCE_NEW_TRANSACTION");
@@ -83,7 +83,7 @@ public class PostgresReadOnlyIncrementalSnapshotChangeEventSourceTest {
     }
 
     @Test
-    @FixFor("DBZ-2431")
+    @FixFor("debezium/dbz#2431")
     void forceNewTransactionIdIsNoOpOnHotStandby() throws Exception {
         // Simulate a hot standby: pg_is_in_recovery() = true, so the CASE returns NULL
         stubQueryResult(null);
@@ -95,7 +95,7 @@ public class PostgresReadOnlyIncrementalSnapshotChangeEventSourceTest {
     }
 
     @Test
-    @FixFor("DBZ-2431")
+    @FixFor("debezium/dbz#2431")
     void forceNewTransactionIdLogsTransactionOnPrimary() throws Exception {
         // Simulate a primary: pg_is_in_recovery() = false, so a real transaction ID is returned
         stubQueryResult("12345");
--- a/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresReadOnlyIncrementalSnapshotChangeEventSourceTest.java
+++ b/debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresReadOnlyIncrementalSnapshotChangeEventSourceTest.java
@@ -26,6 +26,8 @@ import io.debezium.connector.postgresql.connection.PostgresConnection;
 import io.debezium.doc.FixFor;
 import io.debezium.jdbc.JdbcConnection;
 import io.debezium.junit.logging.LogInterceptor;
+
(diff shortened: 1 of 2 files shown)


debezium-connector-postgres/src/test/java/io/debezium/connector/postgresql/PostgresReadOnlyIncrementalSnapshotChangeEventSourceTest.java
```diff
@@ -0,0 +1,126 @@
...
+        stubQueryResult(null);
+
+        LogInterceptor logInterceptor = new LogInterceptor(PostgresReadOnlyIncrementalSnapshotChangeEventSource.class);
+
+        assertThatNoException().isThrownBy(this::invokeForceNewTransactionId);
+        assertThat(logInterceptor.containsMessage("Skipping transaction ID assignment on hot standby")).isTrue();
+    }
+
+    @Test
+    @FixFor("DBZ-2431")
```
Naros:
```suggestion
    @FixFor("debezium/dbz#2431")
```
This and others should follow this convention for GH issues.

Naros (approved):
LGTM @rick020, just one small nit-pick with annotations.

mfvitale (approved):
LGTM. Let's see the CI

rick020:
@mfvitale I updated my test file. Could you re-trigger the CI?

rick020:
@Naros Could you trigger the CI for me again please?

vjuranek:
@rick020 re-triggered, there are some formatting issues, you can use `mvn process-sources` to fix it.

rick020:
> @rick020 re-triggered, there are some formatting issues, you can use `mvn process-sources` to fix it.

@vjuranek  Done

Naros:
Applied, thanks @rick020 :tada:

rick020:
@Naros Thanks! Do need to add myself to https://github.com/debezium/debezium/blob/main/jenkins-jobs/scripts/config/Aliases.txt https://github.com/debezium/debezium/blob/main/COPYRIGHT.txt

PR #7849, opened 2026-08-19
debezium/dbz#2459 Supply OVERRIDING SYSTEM VALUE for PostgreSQL identity targets

### LINKED ISSUE #2459: JDBC sink omits OVERRIDING SYSTEM VALUE, so a PostgreSQL target whose primary key is GENERATED ALWAYS AS IDENTITY fails the task
## Bug report

JDBC sink connector, 3.7.0-SNAPSHOT (`main`, `e441730799`). Reproduced against `quay.io/debezium/connect:nightly`.

---

PostgreSQL source → Kafka → JDBC sink → PostgreSQL, where the sink table already exists and its primary key is `GENERATED ALWAYS AS IDENTITY`.

Sink:

```json
{
  "connector.class": "io.debezium.connector.jdbc.JdbcSinkConnector",
  "topics": "ident.s.ident",
  "connection.url": "jdbc:postgresql://sink-pg:5432/sinkdb",
  "insert.mode": "upsert",
  "primary.key.mode": "record_key",
  "primary.key.fields": "id",
  "schema.evolution": "basic",
  "delete.enabled": "true",
  "collection.name.format": "ident_always"
}
```

Both the source and the pre-created sink table:

```sql
CREATE TABLE s.ident (
  id   integer GENERATED ALWAYS AS IDENTITY PRIMARY KEY,
  name text NOT NULL,
  amt  numeric(10,2)
);
```

---

PostgreSQL 17 (`quay.io/debezium/postgres:17`) for both source and sink, local Docker Compose.

---

The sink writes the primary key value it received from the source into the target table, as it does for a non-identity or `GENERATED BY DEFAULT AS IDENTITY` primary key.

---

**What behavior do you see?**

The sink task fails immediately and permanently:

```
ERROR: cannot insert a non-DEFAULT value into column "id"
  Detail: Column "id" is an identity column defined as GENERATED ALWAYS.
  Hint: Use OVERRIDING SYSTEM VALUE to override.
```

The generated statement omits the SQL standard override clause:

```sql
INSERT INTO public.ident_always (id,name,amt) VALUES (('1'::int4),('snap-a'),('1.11'::double precision))
ON CONFLICT (id) DO UPDATE SET name=EXCLUDED.name,amt=EXCLUDED.amt
```

`OVERRIDING SYSTEM VALUE` is missing after the column list, so PostgreSQL rejects the explicitly supplied identity value. Zero rows are written and the task is unrecoverable.

The PostgreSQL dialect builds `INSERT INTO t (cols) ...` in three places, and none of them emits the clause (`grep -r OVERRIDING` over the repository returns no hits):

- `PostgresDatabaseDialect#getUpsertStatement` — the path above
- `PostgresDatabaseDialect#getBatchInsertStatement` — the UNNEST batch path (`getBatchUpsertStatement` composes it)
- `GeneralDatabaseDialect#getInsertStatement` — inherited, so `insert.mode=insert` fails the same way

Scope of the problem:

- `GENERATED BY DEFAULT AS IDENTITY` works, because that form permits an explicitly supplied value. Only `GENERATED ALWAYS` is affected.
- A sink table that the connector creates itself does not get an identity column, which is why this only surfaces when the target table is pre-created — a homogeneous PostgreSQL-to-PostgreSQL migration, where the target DDL mirrors the source.

There is a precedent for the same class of problem in another dialect: `SqlServerDatabaseDialect#getInsertStatement` wraps the statement in `SET IDENTITY_INSERT ... ON/OFF`, gated by `dialect.sqlserver.identity.insert` (DBZ-6801).

A related defect makes a fix incomplete on its own. `TableDescriptor` decides whether a table has an identity column by looking only at the first column:

```java
this.autoGeneratedIdentityColumn = columns.stream().findFirst().filter(ColumnDescriptor::isAutoIncrement).isPresent();
```

So `hasAutoGeneratedIdentityColumn()` returns `false` whenever the identity column is not the first one, which silently disables the SQL Server path above as well. `anyMatch` is what the method name promises.

---

Yes. Reproduced on current `main` (`e441730799`) and on the nightly container image.

---

```
org.apache.kafka.connect.errors.ConnectException: Exiting WorkerSinkTask due to unrecoverable exception.
Caused by: org.apache.kafka.connect.errors.ConnectException: JDBC sink connector failure
Caused by: org.apache.kafka.connect.errors.ConnectException: Failed to flush records for table 'ident_always'
Caused by: org.hibernate.exception.SQLGrammarException: Error executing work [Batch entry 0
  INSERT INTO public.ident_always (id,name,amt) VALUES (('1'::int4),('snap-a'),('1.11'::double precision))
  ON CONFLICT (id) DO UPDATE SET name=EXCLUDED.name,amt=EXCLUDED.amt
  was aborted: ERROR: cannot insert a non-DEFAULT value into column "id"
    Detail: Column "id" is an identity column defined as GENERATED ALWAYS.
    Hint: Use OVERRIDING SYSTEM VALUE to override.
Caused by: org.postgresql.util.PSQLException: ERROR: cannot insert a non-DEFAULT value into column "id"
```

---

Starting from the tutorial's PostgreSQL setup, with a JDBC sink writing back into a second PostgreSQL:

1. On the source, create a table whose primary key is `GENERATED ALWAYS AS IDENTITY` and insert a row.
2. On the sink database, pre-create the destination table with the same DDL, so the connector does not create it.
3. Register the PostgreSQL source and the JDBC sink shown above.
4. The sink task fails on the first flush with the error above; the destination table stays empty.

Replacing `GENERATED ALWAYS` with `GENERATED BY DEFAULT` in step 2 makes the same pipeline succeed.

### A note on the fix

`OVERRIDING SYSTEM VALUE` is accepted on any insert target in PostgreSQL — I verified on 17 that it is inert on a plain `integer` column, a `serial` column and a `GENERATED BY DEFAULT` identity column — so emitting it whenever the target table has an identity column is enough, and `ALWAYS` and `BY DEFAULT` need not be told apart.

One question worth settling before this is fixed: should the clause be behind a configuration option, mirroring `dialect.sqlserver.identity.insert`? The two cases differ. `SET IDENTITY_INSERT` is session scoped, needs `ALTER` on the table and allows only one table to be ON per session, so an opt-in is warranted there. `OVERRIDING SYSTEM VALUE` is statement scoped, needs no privilege and does nothing unless a column is `GENERATED ALWAYS`. A flag would mostly give users the ability to leave the task broken, so I have not added one — but I am happy to follow the SQL Server precedent instead if you prefer the consistency.

Worth flagging separately: `CockroachDBDatabaseDialect` extends `PostgresDatabaseDialect`, and CockroachDB 25.4 raises the same `GENERATED ALWAYS` restriction but does not parse `OVERRIDING SYSTEM VALUE` at all (SQLSTATE 42601), so it has to opt out of the clause explicitly.

I have a fix and tests ready and will open a pull request shortly.

sundongkim-dev:
Fixes debezium/dbz#2459

## Description

A column declared `GENERATED ALWAYS AS IDENTITY` rejects an explicitly supplied value, and a sink replicating the source's primary key has no choice but to supply one. The PostgreSQL dialect never emitted the SQL standard override clause, so writing into a pre-created target table failed the task outright:

```
ERROR: cannot insert a non-DEFAULT value into column "id"
  Detail: Column "id" is an identity column defined as GENERATED ALWAYS.
  Hint: Use OVERRIDING SYSTEM VALUE to override.
```

This surfaces in a homogeneous PostgreSQL-to-PostgreSQL migration, where the target DDL mirrors the source. A table the connector creates itself does not get an identity column, which is why the default path never hit this.

`OVERRIDING SYSTEM VALUE` is now emitted after the column list on all three PostgreSQL insert paths — the inherited `getInsertStatement`, `getUpsertStatement`, and the UNNEST `getBatchInsertStatement`, which `getBatchUpsertStatement` composes and so inherits the clause for free. Fixing only `getUpsertStatement` would have left `insert.mode=insert` and the UNNEST batch broken.

The clause is added through a `protected getOverrideClause(TableDescriptor)` hook on `GeneralDatabaseDialect`, defaulting to an empty string. `getUpsertStatement` builds its own statement and PostgreSQL does not override `getInsertStatement`, so the clause has to be injected mid-statement; the SQL Server approach of wrapping `super`'s output does not apply. Since `SqlStatementBuilder.append` is a plain `StringBuilder.append`, the empty default leaves every other dialect's SQL byte-identical.

Two details worth calling out:

**`TableDescriptor` (first commit).** `hasAutoGeneratedIdentityColumn()` inspected only the first column, so it returned `false` whenever the identity column was not first. It is kept as a separate commit because it changes behaviour beyond PostgreSQL: its only other caller is `SqlServerDatabaseDialect#wrapWithIdentityInsert`, which will now wrap statements for tables it previously skipped. That is the intended reading of the method name, and it is bounded by the opt-in `dialect.sqlserver.identity.insert`, but it is a behaviour change and I would rather it be easy to revert on its own. Without it, four of the twelve new tests fail, so it cannot simply be split into a separate PR.

**CockroachDB.** `CockroachDBDatabaseDialect` extends `PostgresDatabaseDialect`. CockroachDB 25.4 enforces the same `GENERATED ALWAYS` restriction but does not parse `OVERRIDING SYSTEM VALUE` at all (SQLSTATE 42601), so it overrides the hook to emit nothing. This keeps its existing, clearer error rather than turning it into a syntax error. Being an unsupported target either way, there is no success path to assert, so this is a guard rather than something the test suite covers.

`GENERATED BY DEFAULT AS IDENTITY` was never affected, since that form permits an explicitly supplied value. No configuration option gates the clause: it is statement scoped, needs no privilege, and is inert unless a column is `GENERATED ALWAYS` — verified on PostgreSQL 17 against a plain `integer` column, a `serial` column and a `BY DEFAULT` identity column. Gating it would mostly offer users the ability to leave the task failing. I am happy to follow the `dialect.sqlserver.identity.insert` precedent instead if you would rather have the consistency.

## Verification

`JdbcSinkIdentityColumnIT` covers upsert and insert modes against a pre-created identity target, plus a case where the identity column is not the first column, across both record factories and with the UNNEST batch path on and off — twelve cases. Two records per case, so the batch path is actually taken rather than falling back for a single record.

Reverting the dialect change fails all twelve with the error above; reverting only the `TableDescriptor` change fails exactly the four not-first-column cases, so both parts are load-bearing.

Regression runs on this branch: `it-postgresql` 348 passing, the same tag with `-Denable.sces=true` 366 passing with nothing skipped, `it-cockroachdb` 76 passing, `it-sqlserver` 122 passing. I could not run `e2e-sqlserver` locally and am relying on CI for it. End to end, a PostgreSQL 17 → Kafka → JDBC sink → PostgreSQL 17 pipeline now replicates snapshot, insert, update and delete into a `GENERATED ALWAYS AS IDENTITY` target, which failed on the first flush before.

## PR Checklist


Naros (approved):

Naros (changes requested):
Actually @sundongkim-dev could you do me a favor, as this may be useful to backport to 3.6 but in 3.6 CockroachDB does not exist.  Could you split the code changes into two buckets (sets of commits):

* Changes that target general JDBC sink + PostgreSQL
* Changes that target specifically the CockroachDB dialect (applied last in the commit history)

This way we can backport the former, and leave the latter on 3.7.

sundongkim-dev:
Split as asked, @Naros. The first two commits are the general JDBC sink and PostgreSQL change; the CockroachDB opt-out is last on its own.

jpechane (approved):

Naros (approved):

Naros:

PR #7850, opened 2026-08-19
debezium/dbz#2452 Recover incremental snapshot chunks on schema mismatch

### LINKED ISSUE #2452: Incremental snapshot fails fatally when the table schema changes mid-snapshot: recover the chunk instead of stopping the connector
### What Debezium connector do you use and what version?

PostgreSQL connector, current main. Structurally, every relational connector built on `AbstractIncrementalSnapshotChangeEventSource` is affected.

### What behavior do you expect / see?

A DDL statement landing on a table while an incremental snapshot is reading its chunks makes the connector fail fatally. On restart the pending snapshot resumes on the same table, so depending on timing this ranges from one restart per DDL to a permanent crash loop. The mismatch can surface in either direction:

- **column added**: the cached table schema and the result set disagree → `IllegalArgumentException` from `ColumnUtils.toArray` (the message itself hints at this family: *"This might be caused by DBZ-4350"*);
- **column dropped**: with an explicit column projection the chunk query references a column that no longer exists → SQLSTATE `42703` (Postgres) / `42S22` (MySQL family).

This is reachable on current main today: with a column filter configured, `buildProjection` already produces an explicit column list, and the `schemaChanges` suite combined with a `column.exclude.list` fails — as observed during the review of #6000 (DBZ-8430). The same race is what is blocking https://github.com/debezium/debezium/pull/7532 (#2020, generated columns) on the `schemaChanges` suites. Quoting the open question from the #6000 review:

> *"There is a window between the schema change validation step and the execution of the query where I could imagine this error happening, and we should likely guard against this race condition. The question is whether or not we can efficiently guard against this"*

### Proposal

A bounded recovery guard instead of the fatal: on a schema mismatch (either direction), refresh the table schema through the connection, discard the buffered window rows, revert the chunk position, re-read with a fresh projection — with a small bounded attempt counter so a genuinely broken state still fails fast. A DDL landing mid-snapshot then costs one chunk re-read instead of the connector.

We run this guard in production on our fork (built for the parallel-snapshot work in #7362, where the read-only Postgres source's connection-aware schema refresh pattern already pointed the way). A live example — `DROP COLUMN` issued while chunk reads were active on a 484k-row table:

```
12:59:33  incremental snapshot signalled
13:00:05  ALTER TABLE ... DROP COLUMN (snapshot in flight)
13:00:29  WARN Cached schema for table '...' is stale against the database (recovery 1/3):
          refreshing schema, reverting chunk and re-reading
13:04:14  snapshot completed (483,989 rows = exact source count)
```

One recovered chunk, no connector failure, exact row count. A subtle detail worth guarding as well: an undefined-column error raised while reading the maximum key can currently be swallowed by the generic catch in `readMaximumKey` and silently skip the table as "empty".

### Relationship with PR #7532

We hope this also helps unblock https://github.com/debezium/debezium/pull/7532: the explicit projection there is correct, and with this guard in place the `schemaChanges` suites should pass with it, since the race window becomes recoverable.

MrIvv:
Closes https://github.com/debezium/dbz/issues/2452
Related: https://github.com/debezium/debezium/pull/7588

### What

A DDL landing while an incremental snapshot chunk is in flight currently makes the connector fail, skip the table silently, or hang, depending on where the schema mismatch surfaces. This PR turns all of these into a bounded chunk deferral: refresh the table schema, discard the buffered window rows, revert the chunk position, and re-read the chunk in the next window.

Four mismatch surfaces are covered in `AbstractIncrementalSnapshotChangeEventSource`:

1. **Result set ahead of the cache** (column added): `IllegalArgumentException` from `ColumnUtils.toArray`, which previously bubbled to the generic handler and skipped the table.
2. **Cache ahead of the database** (column dropped, explicit projection): undefined-column SQLException (`42703` PostgreSQL/Db2, `42S22` MySQL/MariaDB, `S0022` SQL Server). In the maximum-key query this was previously caught generically and the table skipped.
3. **The schema-verification query itself**: `verifySchemaUnchanged`/`readSchema` builds its LIMIT-0 probe with the same cached projection, so a drop landing between verification cycles fails it too. The `DebeziumException` wrapper reached the generic handler, which skipped the table. Worse, because the exception bypasses `emitWindowClose`, the window transaction stayed uncommitted: the connection kept the table's metadata lock and blocked every subsequent DDL on that table indefinitely. The deferral path closes the window instead, which commits the transaction and releases the lock. The other failure paths, which skip the window close, now roll the transaction back explicitly for the same reason.
4. **Buffered rows against a rotated emission schema**: rows buffered before a DDL fail to emit once the streamed DDL rotates the `TableSchema` (`Data row is smaller than a column index`). The failure was swallowed by the signal processor ("Action snapshot-window-close failed ... may not have been processed") and the snapshot hung silently, forever. On deferral the chunk position returns to the start of the window, so the whole window is re-read: at-least-once semantics, since the rows already delivered from the failed window are emitted again, but the row whose emission failed is never lost.

The deferral counter is progress-aware: consecutive deferrals only count against the same observed schema, so a DDL storm keeps making progress while a genuinely broken state (the refreshed schema never changes) still fails after the bound (3).

### Why now

This race is what blocks the explicit chunk projection of #7532 (dbz#2020, generated columns) on the `schemaChanges` suites, and it is reachable on main today: configuring `column.include.list`/`column.exclude.list` switches the chunk query to an explicit projection, which hides concurrent DDL from the result-set schema comparison (`checkSchemaChanges` sees two identical stale shapes) until the window-close emission blows up.

### Testing

New `schemaChangesWithColumnFilter` in `AbstractIncrementalSnapshotWithSchemaChangesSupportTest`: the existing `schemaChanges` DDL-storm body, started with a non-matching `column.exclude.list` so the chunk query uses an explicit projection. Without the guard this hangs on MySQL (the snapshot stalls after the first ADD COLUMN with no failure surfaced).

The test runs on the suite extenders (binlog family, SQL Server). The PostgreSQL connector does not extend this suite (no incremental-snapshot schema-changes support), so the PostgreSQL side of the guard (undefined-column recovery) is exercised in combination with the projection and tests of #7532; the full PostgreSQL `IncrementalSnapshotIT` was run against PostgreSQL 14, 17 and 18/pgoutput as a shared-path regression: green.

## PR Checklist


debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -746,6 +816,55 @@ private boolean checkSchemaChanges(ResultSet rs) throws SQLException {
...
+    }
+
+    private static boolean isUndefinedColumn(SQLException e) {
+        for (Throwable t = e; t != null; t = t.getCause()) {
+            if (t instanceof SQLException sql && UNDEFINED_COLUMN_SQL_STATES.contains(sql.getSQLState())) {
+                return true;
+            }
+        }
+        return false;
+    }
```
Naros, line 922: I don't like mixing different dialect SQLState values like this; it could easily become brittle when two different dialects share the same error codes but mean different things. 

Either each connector extends the class with its own method override, or this class accepts a FunctionalInterface that an implementation provides to drive this dialect-wise.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -173,8 +182,21 @@ protected String getSignalTableName(String dataCollectionId) {
...
         LOGGER.debug("Sending {} events from window buffer", window.size());
         offsetContext.incrementalSnapshotEvents();
-        for (Object[] row : window.values()) {
-            sendEvent(partition, dispatcher, offsetContext, row);
+        try {
+            for (Object[] row : window.values()) {
+                sendEvent(partition, dispatcher, offsetContext, row);
+            }
+        }
+        catch (ConnectException e) {
```
Naros: Are we sure that `ConnectException` is the best choice here? This assumes Kafka Connect, and the connector could run in other runtimes that wouldn't necessarily throw this exception.

debezium-embedded/src/test/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotWithSchemaChangesSupportTest.java
```diff
@@ -50,6 +52,23 @@ public void schemaChanges() throws Exception {
 
         populateTable();
         startConnector();
+        executeSchemaChangesTest();
+    }
+
+    @Test
+    @FixFor("debezium/dbz#2452")
+    public void schemaChangesWithColumnFilter() throws Exception {
```
Naros, line 61: This test should identify that the deferral mechanism was actually used. One common way to do this is with a `LogInterceptor` that looks for a specific debug log entry on certain code paths. Could we update the test to look for something like this?

Naros (changes requested):

debezium-embedded/src/test/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotWithSchemaChangesSupportTest.java
MrIvv: Hope is better now

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -172,9 +178,40 @@ protected String getSignalTableName(String dataCollectionId) {
...
+            window.clear();
+            warnAndSkip(partition, offsetContext,
+                    SQL_EXCEPTION,
+                    "Error while emitting the incremental snapshot window of table '%s', skipping and continuing streaming"
+                            .formatted(context.currentDataCollectionId().getId()),
+                    e);
+            return;
         }
         offsetContext.postSnapshotCompletion();
         window.clear();
```
Naros: Should `postSnapshotCompletion` be called here regardless of early return? If I am not mistaken, this is potentially a latent bug where the `snapshot` attribute continues to carry `incremental` in the streamed events `source` information block.

debezium-connector-sqlserver/src/main/java/io/debezium/connector/sqlserver/SqlServerConnection.java
```diff
@@ -163,6 +163,15 @@ public class SqlServerConnection extends JdbcConnection {
      * @param valueConverters     {@link SqlServerValueConverters} instance
      * @param skippedOperations   a set of {@link Envelope.Operation} to skip in streaming
      */
+    private static final int ERROR_INVALID_COLUMN_NAME = 207;
```
Naros: Hi, this has cut the javadoc off from the method. Please move this `private static` final with the other private static final members in the class.

debezium-connector-common/src/main/java/io/debezium/jdbc/JdbcConnection.java
```diff
@@ -412,6 +413,16 @@ public JdbcConnection commit() throws SQLException {
...
 
+    /**
+     * Whether the given exception reports a reference to a column that does not exist in the
+     * database. Used by the incremental snapshot to classify a chunk query failure as a
+     * stale-schema condition. Defaults to the SQL standard SQLSTATE; dialects that report a
+     * vendor-specific code override this.
+     */
+    public boolean isUndefinedColumnError(SQLException exception) {
+        return SQLSTATE_UNDEFINED_COLUMN.equals(exception.getSQLState());
+    }
```
Naros: While doing it this way benefits dialects like Db2, which get this for free, my concern is that 42703 also refers in Db2 to cases where undefined parameters or variables are invalid, so the catch-all could be overly broad.

I still wonder if something like this on `IncrementalSnapshotChangeEventSource` is more appropriate, and then each connector's change event source factory could pass its specific value in the ctor for `AbstractIncrementalSnapshotChangeEventSource`, and the current ctor could explicitly pass `NONE` as the fallback for existing connectors not implementing the behavior.
```
@FunctionalInterface
interface UndefinedColumnClassifier {
    UndefinedColumnClassifier NONE = exception -> false;

    boolean isUndefinedColumn(SQLException exception);
}
```
This means connectors would need to opt in to the behavior, but it also forces connectors to be explicit about what they mean by an undefined column for incremental snapshots. 

I'm curious what my colleagues think here.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -172,9 +178,40 @@ protected String getSignalTableName(String dataCollectionId) {
...
         LOGGER.debug("Sending {} events from window buffer", window.size());
+        if (!window.isEmpty() && windowSchema != null && databaseSchema.schemaFor(currentTable.id()) != windowSchema) {
+            // The emission schema rotated after the window was buffered (every schema refresh
+            // replaces the TableSchema instance, and the pipeline is single-threaded, so the
+            // identity comparison detects the rotation deterministically): the buffered rows
+            // no longer match it. The chunk position returns to the window start and the whole
+            // window is re-read (at-least-once); sendEvent advances lastEventKeySent before
+            // dispatching, so reverting to it would skip a row.
+            deferChunkOnStaleSchema(new DebeziumException(
+                    "The schema of table '%s' was refreshed after the window was buffered".formatted(currentTable.id())));
```
Naros: I believe this may require a try-catch block paired with a warnAndSkip. Without it, a failure here would reach the signal processor, which would log and drop it without opening another window, stalling the snapshot and skipping the table instead. Is that right?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -172,9 +178,40 @@ protected String getSignalTableName(String dataCollectionId) {
...
+            }
+        }
+        catch (InterruptedException e) {
+            throw e;
+        }
+        catch (Exception e) {
+            // Same policy as a failed chunk read: emitting this window is not going to succeed
+            // on a retry either, so skip the table visibly instead of letting the failure reach
+            // the signal processor, which swallows it and stalls the snapshot silently.
+            window.clear();
```
Naros, line 235:
```suggestion
            offsetContext.postSnapshotCompletion();
            window.clear();
```
Without the `postSnapshotCompletion`, any further streamed events, I believe, show up with their `snapshot` source information block attribute showing `incremental`, which would be incorrect.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -746,6 +837,55 @@ private boolean checkSchemaChanges(ResultSet rs) throws SQLException {
...
+            throw new DebeziumException("Schema refresh failed while deferring a chunk of table '%s'".formatted(currentTable.id()), e);
+        }
+    }
+
+    private void resetStaleSchemaDeferrals() {
+        staleSchemaDeferrals = 0;
+        lastStaleTable = null;
+    }
+
+    private boolean isUndefinedColumn(SQLException e) {
```
Naros:
```suggestion
    private boolean isUndefinedColumn(Exception e) {
```
Then, at the call site, you can pass the Exception without any specific handling. This already traverses the exception hierarchy, checking for SQLException as it traverses upward.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -746,6 +837,55 @@ private boolean checkSchemaChanges(ResultSet rs) throws SQLException {
...
+     * {@code ColumnUtils.toArray}), the chunk or schema-verification query can reference a column
+     * that no longer exists (undefined-column SQLSTATE), or rows buffered against the old schema
+     * can fail to emit once the schema rotates before the window closes. None of these is fatal: refreshing
+     * the schema and closing the window lets the connector re-verify it (and streaming deliver
+     * the pending DDL events) before the chunk is re-read. Consecutive deferrals are counted
+     * against the same observed schema, so a DDL storm keeps making progress while a genuinely
+     * broken state (the refreshed schema never changes) still fails after the bound.
+     */
+    private void deferChunkOnStaleSchema(Exception cause) {
+        if (lastStaleTable != null && currentTable != null && !currentTable.equals(lastStaleTable)) {
```
Naros, line 883: I think we should call `window.clear()` here first. 

My reasoning is partial-buffer safety. In `createDataEventsForTable` a SQLException doesn't have to surface through an `executeQuery` call, it can actually surface mid `ResultSet#next` too, after rows are already in the window. Without the clear, the mid-iteration deferral leaves a partial chunk in the buffer, and when the window-close arrives, that partial chunk is emitted. 

The other point is that the bound-exceeded path clears before throwing, so that the `warnAndSkip` handlers never inherit any sort of stale buffer.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -746,6 +837,55 @@ private boolean checkSchemaChanges(ResultSet rs) throws SQLException {
...
+        if (staleSchemaDeferrals > MAX_STALE_SCHEMA_DEFERRALS) {
+            resetStaleSchemaDeferrals();
+            throw new DebeziumException(
+                    "Cached schema for table '%s' is still stale against the database after %d chunk deferrals"
+                            .formatted(currentTable.id(), MAX_STALE_SCHEMA_DEFERRALS),
+                    cause);
+        }
+        LOGGER.warn("Cached schema for table '{}' is stale against the database (deferral {}/{}): refreshing schema and re-reading the chunk in the next window",
+                currentTable.id(), staleSchemaDeferrals, MAX_STALE_SCHEMA_DEFERRALS, cause);
+        context.setSchemaVerificationPassed(false);
```
Naros, line 901:
```suggestion
        commitWindowTransaction();
        context.setSchemaVerificationPassed(false);
```
Some drivers, once an error surfaces, strictly follow the JDBC specification that says the connection can no longer be used because it treats the current transaction as aborted, and the only way forward is to commit/rollback the transaction to return the connection to a clean state. Since this path is called when the deferral has seen a JDBC exception, this is good to do for those strict drivers, as the following step is to refresh the schema.

Naros (commented):
Hi @MrIvv, I've left a few comments.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
MrIvv: yes, my missing

debezium-connector-common/src/main/java/io/debezium/jdbc/JdbcConnection.java
MrIvv: Hi @Naros, I added this part to the PR, but I would appreciate getting your team's opinion, since you know the product and its future evolution better

MrIvv:
I've rebased the code and it's ready for review whenever you have a chance, @Naros
Thanks a lot

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -173,9 +194,52 @@ protected String getSignalTableName(String dataCollectionId) {
 
     protected void sendWindowEvents(P partition, OffsetContext offsetContext) throws InterruptedException {
         LOGGER.debug("Sending {} events from window buffer", window.size());
+        if (!window.isEmpty() && windowSchema != null && databaseSchema.schemaFor(currentTable.id()) != windowSchema) {
```
Naros: Question: on PostgreSQL with pgoutput, every RELATION message rebuilds the `TableSchema` unconditionally (`PgOutputMessageDecoder#handleRelationMessage` calls `PostgresSchema#applySchemaChangesForTable`, which calls `refresh`), even when the table definition hasn't changed. If one arrives for the current table between the chunk read and the window close, this identity check defers the window without any DDL.

That likely costs only an extra re-read, but would comparing the `Table` definitions rather than the `TableSchema` instance be more precise?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -173,9 +194,52 @@ protected String getSignalTableName(String dataCollectionId) {
...
-            sendEvent(partition, dispatcher, offsetContext, row);
+        try {
+            for (Object[] row : window.values()) {
+                sendEvent(partition, dispatcher, offsetContext, row);
+            }
+        }
+        catch (InterruptedException e) {
+            throw e;
+        }
+        catch (Exception e) {
```
Naros: Thanks for adding this, it covers the signal-processor stall nicely. One concern: `sendWindowEvents` is also called directly from the streaming thread by the read-only sources (`BinlogReadOnlyIncrementalSnapshotChangeEventSource#readUntilGtidChange` and `PostgresReadOnlyIncrementalSnapshotChangeEventSource#readUntilNewTransactionChange`). On those paths an emission failure used to propagate and fail the task visibly. With this catch-all, any emission failure (a converter, a custom `ChangeRecordEmitter`, etc.) now skips the table with only a WARN.

Could we narrow this to the schema-rotation failure, or limit the skip to the signal-based path so the read-only sources keep their existing fail-fast behavior?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -295,12 +370,21 @@ protected void readChunk(P partition, OffsetContext offsetContext) throws Interr
...
+                            deferChunkOnStaleSchema(e);
+                            break;
+                        }
                         LOGGER.error("Failed to read maximum key for table {}", currentTableId, e);
                         notificationService.incrementalSnapshotNotificationService().notifyTableScanCompleted(context, partition, offsetContext, totalRowsScanned,
                                 SQL_EXCEPTION);
                         nextDataCollection(partition, offsetContext);
                         continue;
                     }
+                    catch (IllegalArgumentException e) {
```
Naros: The comment points at `ColumnUtils.toArray`, but this catch covers the whole `queryAndMap` lambda, including `jdbcConnection.rowToArray` and `keyFromRow`. An unrelated `IllegalArgumentException` would be treated as a stale schema, deferred three times, and then the table skipped with a "still stale" message that sends the user in the wrong direction. Could the catch be scoped to the `ColumnUtils.toArray` call only?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -727,9 +835,19 @@ private boolean createDataEventsForTable(P partition) throws SQLException {
...
         }
         catch (SQLException e) {
+            if (isUndefinedColumn(e)) {
+                deferChunkOnStaleSchema(e);
+                return false;
+            }
             LOGGER.error("Snapshotting of table {} failed. Skipping it", currentTable.id(), e);
             throw e;
         }
+        catch (IllegalArgumentException e) {
```
Naros: Same concern as the maximum-key query: this try block also covers `rowToArray`, `tableSchema.keyFromColumnData` and `chunkQueryBuilder.resolveChunkEndPosition`, so any `IllegalArgumentException` from value conversion ends up on the stale-schema path. Scoping it to `ColumnUtils.toArray` would keep the classification accurate.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -727,9 +835,19 @@ private boolean createDataEventsForTable(P partition) throws SQLException {
...
+            // ColumnUtils.toArray: the result set carries a column the cached schema does not know yet
+            deferChunkOnStaleSchema(e);
+            return false;
+        }
+        resetStaleSchemaDeferrals();
```
Naros: Because the counter resets after every successful chunk read, and an emission-path deferral is always followed by a re-read, deferrals raised from `sendWindowEvents` never accumulate past 1/3. That makes the bound-exceeded `catch (DebeziumException)` in `sendWindowEvents` effectively unreachable; the CI logs show `deferral 1/3` repeating for the same table.

Is that intended? If the bound should also apply to emission deferrals, the reset might need to happen only once a window has actually been emitted.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
```diff
@@ -748,6 +866,63 @@ private boolean checkSchemaChanges(ResultSet rs) throws SQLException {
...
+        lastStaleTable = currentTable;
+        staleSchemaDeferrals++;
+        if (staleSchemaDeferrals > MAX_STALE_SCHEMA_DEFERRALS) {
+            resetStaleSchemaDeferrals();
+            throw new DebeziumException(
+                    "Cached schema for table '%s' is still stale against the database after %d chunk deferrals"
+                            .formatted(currentTable.id(), MAX_STALE_SCHEMA_DEFERRALS),
+                    cause);
+        }
+        LOGGER.warn("Cached schema for table '{}' is stale against the database (deferral {}/{}): refreshing schema and re-reading the chunk in the next window",
```
Naros: Nit: outside PostgreSQL, `refreshTableSchema` is a no-op (the base implementation returns the table unchanged), so for binlog and SQL Server nothing is refreshed here; the deferral waits for streaming to deliver the DDL. Maybe reword to something like "re-reading the chunk in the next window" so the log doesn't suggest a refresh took place?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/IncrementalSnapshotChangeEventSource.java
```diff
@@ -23,6 +24,21 @@
  */
 public interface IncrementalSnapshotChangeEventSource<P extends Partition, T extends DataCollectionId> {
 
+    /**
+     * Recognizes the dialect-specific SQL error a chunk query fails with when it references a
+     * column that no longer exists in the database, which the incremental snapshot treats as a
+     * stale cached schema and recovers from by deferring the chunk. Connectors opt in by passing
+     * their classifier to the change event source; with {@link #NONE} the recovery never
```
Naros: This reads as though connectors that keep `NONE` see no behavior change, but that only holds for the undefined-column SQLSTATE path. The `IllegalArgumentException` deferral, the emission schema identity check in `sendWindowEvents`, and the emission catch-all apply to every connector, regardless of the classifier or `incremental.snapshot.allow.schema.changes`. For example, Oracle still uses the old constructor and picks up all three.

Could we either gate those paths as well, or adjust the Javadoc (and the PR description) to reflect the actual scope?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogConnectorConnection.java
```diff
@@ -55,6 +55,12 @@ public abstract class BinlogConnectorConnection extends JdbcConnection {
     private final ConnectionConfiguration connectionConfig;
     private final BinlogFieldReader fieldReader;
 
+    private static final String SQLSTATE_UNDEFINED_COLUMN = "42S22";
```
Naros: Small one, mirroring the SQL Server change in cf6cd775: could this constant move up with the other `private static final` members?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogConnectorConnection.java
```diff
@@ -55,6 +55,12 @@ public abstract class BinlogConnectorConnection extends JdbcConnection {
...
+
+    public boolean isUndefinedColumnError(SQLException exception) {
```
Naros: Nice to see `SqlServerConnectionIT#shouldClassifyUndefinedColumnError` for the SQL Server classifier. Could we add the equivalent for this one, and for `PostgresConnection#isUndefinedColumnError`? The PostgreSQL recovery path isn't exercised by any test in this PR, so a classifier-level test would at least guard the SQLSTATE mapping.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: Question: on PostgreSQL with pgoutput, every RELATION message rebuilds the `TableSchema` unconditionally (`PgOutputMessageDecoder#handleRelationMessage` calls `PostgresSchema#applySchemaChangesForTable`, which calls `refresh`), even when the table definition hasn't changed. If one arrives for the current table between the chunk read and the window close, this identity check defers the window without any DDL.

That likely costs only an extra re-read, but would comparing the `Table` definitions rather than the `TableSchema` instance be more precise?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: Thanks for adding this, it covers the signal-processor stall. One concern about its scope: `sendWindowEvents` is also called directly on the streaming thread by the read-only sources (`BinlogReadOnlyIncrementalSnapshotChangeEventSource#readUntilGtidChange` and `PostgresReadOnlyIncrementalSnapshotChangeEventSource#readUntilNewTransactionChange`). On those paths an emission failure used to propagate and fail the task (`BinlogStreamingChangeEventSource#handleEvent` hands it to the error handler). With this catch-all, any emission failure (a converter, a custom `ChangeRecordEmitter`, etc.) now skips the table with a WARN instead.

Since the identity check above already handles the schema-rotation case before emission, what reaches this catch is by definition something unrelated to schema drift. Could we limit the skip to the signal-based path (or handle the swallowing in `closeWindow` / `SignalProcessor`), so the read-only sources keep their fail-fast behavior?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: The comment points at `ColumnUtils.toArray`, but this catch covers the whole `queryAndMap` lambda, including `jdbcConnection.rowToArray` and `keyFromRow`. An unrelated `IllegalArgumentException` would be treated as a stale schema, deferred three times, and then the table skipped with a "still stale" message that sends the user in the wrong direction. Could the catch be scoped to the `ColumnUtils.toArray` call only?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: Same concern as the maximum-key query: this try block also covers `rowToArray`, `tableSchema.keyFromColumnData` and `chunkQueryBuilder.resolveChunkEndPosition`, so any `IllegalArgumentException` from value conversion ends up on the stale-schema path. Scoping it to `ColumnUtils.toArray` would keep the classification accurate.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: A non-empty window at emission time always follows a successful `createDataEventsForTable`, and this reset runs at the end of every successful read. So when the identity check in `sendWindowEvents` defers, the counter is always 0 and becomes 1; emission-path deferrals can never accumulate to the bound. The CI logs show this as `deferral 1/3` repeating for the same table.

That leaves the `catch (DebeziumException)` in `sendWindowEvents` reachable only through a failed schema refresh on PostgreSQL, never through the bound. Is that intended? If the bound should also cover repeated emission deferrals, the reset would need to happen after a window is actually emitted rather than after a read.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: Nit: outside PostgreSQL, `refreshTableSchema` is a no-op (the base implementation returns the table unchanged), so for binlog and SQL Server nothing is refreshed here; the deferral waits for streaming to deliver the DDL. Maybe reword to something like "re-reading the chunk in the next window" so the log doesn't suggest a refresh took place?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/IncrementalSnapshotChangeEventSource.java
Naros: This reads as though connectors that keep `NONE` see no behavior change, but that only holds for the undefined-column SQLSTATE path. The `IllegalArgumentException` deferral, the emission schema identity check in `sendWindowEvents`, and the emission catch-all apply to every connector, regardless of the classifier or `incremental.snapshot.allow.schema.changes`. For example, Oracle still uses the old constructor and picks up all three.

Could we either gate those paths as well, or adjust the Javadoc (and the PR description) to reflect the actual scope?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogConnectorConnection.java
Naros: Small one, mirroring the SQL Server change in cf6cd775: could this constant move up with the other `private static final` members?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogConnectorConnection.java
Naros: Nice to see `SqlServerConnectionIT#shouldClassifyUndefinedColumnError` for the SQL Server classifier. Could we add the equivalent for this one, and for `PostgresConnection#isUndefinedColumnError`? The PostgreSQL recovery path isn't exercised by any test in this PR, so a classifier-level test would at least guard the SQLSTATE mapping.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: On PostgreSQL with pgoutput, a relation (`R`) message for a table triggers :

`PgOutputMessageDecoder#handleRelationMessage` 
  → `PostgresSchema#applySchemaChangesForTable` 
  → `refresh`

which always builds a new `TableSchema` instance. pgoutput sends the `R` message the first time a table appears in the session, not only after DDL. So if the first DML on the snapshotted table lands between the chunk read and the window close, this identity check defers the window and logs "stale against the database" even though nothing changed.

Is this intended?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: One concern about its scope: `sendWindowEvents` is also called directly on the streaming thread by the read-only sources:
* `BinlogReadOnlyIncrementalSnapshotChangeEventSource#readUntilGtidChange`
* `PostgresReadOnlyIncrementalSnapshotChangeEventSource#readUntilNewTransactionChange`

On those paths, an emission failure used to propagate and fail the task (e.g Binlog's `handleEvent` hands it to the error handler). With this catch-all, any emission failure (a converter, a custom `ChangeRecordEmitter`, etc.) now skips the table with a WARN instead.

Since the identity check above already handles the schema-rotation case before emission, what reaches this catch is, by definition, unrelated to schema drift. Could we limit the skip to the signal-based path (or handle the swallowing in `closeWindow` / `SignalProcessor`), so the read-only sources keep their fail-fast behavior?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: The comment points at `ColumnUtils.toArray`, but this catch covers the whole `queryAndMap` lambda, including `jdbcConnection.rowToArray` and `keyFromRow`. An unrelated `IllegalArgumentException` would be treated as a stale schema, deferred three times, and then the table skipped with a "still stale" message that sends the user in the wrong direction. Could the catch be scoped to the `ColumnUtils.toArray` call only?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: Same concern as the maximum-key query: this try block also covers `rowToArray`, `tableSchema.keyFromColumnData` and `chunkQueryBuilder.resolveChunkEndPosition`, so any `IllegalArgumentException` from value conversion ends up on the stale-schema path. 

Scoping it to `ColumnUtils.toArray` would keep the classification accurate.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: A non-empty window at emission time always follows a successful `createDataEventsForTable`, and this reset runs at the end of every successful read. So when the identity check in `sendWindowEvents` defers, the counter is always 0 and becomes 1; emission-path deferrals can never accumulate to the bound. The CI logs show this as `deferral 1/3` repeating for the same table.

That leaves the `catch (DebeziumException)` in `sendWindowEvents` reachable only through a failed schema refresh on PostgreSQL, never through the bound. Is that intended? If the bound should also cover repeated emission deferrals, the reset would need to happen after a window is actually emitted rather than after a read.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
Naros: Outside PostgreSQL, `refreshTableSchema` is a no-op (the base implementation returns the table unchanged), so for binlog and SQL Server nothing is refreshed here; the deferral waits for streaming to deliver the DDL. Maybe reword it to something like "re-reading the chunk in the next window" so the log doesn't suggest a refresh took place?

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/IncrementalSnapshotChangeEventSource.java
Naros: This reads as though connectors that keep `NONE` see no behavior change, but that only holds for the undefined-column SQLSTATE path. The `IllegalArgumentException` deferral, the emission schema identity check in `sendWindowEvents`, and the emission catch-all apply to every connector, regardless of the classifier or `incremental.snapshot.allow.schema.changes`. For example, Oracle still uses the old constructor and picks up all three.

Could we either gate those paths as well, or adjust the Javadoc to reflect the actual scope?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogConnectorConnection.java
Naros: Mirroring the SQL Server change in cf6cd775: could this constant move up with the other `private static final` members?

debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/jdbc/BinlogConnectorConnection.java
Naros: Nice to see `SqlServerConnectionIT#shouldClassifyUndefinedColumnError` for the SQL Server classifier. Could we add the equivalent for this one, and for `PostgresConnection#isUndefinedColumnError`? The PostgreSQL recovery path isn't exercised by any test in this PR, so a classifier-level test would at least guard the SQLSTATE mapping.

Naros (commented):
Thanks for the rebase @MrIvv, I've left several more comments while we wait for CI for the Oracle connector.

debezium-connector-common/src/main/java/io/debezium/pipeline/source/snapshot/incremental/AbstractIncrementalSnapshotChangeEventSource.java
MrIvv: Hello Naros, no, that isn't intended. is a big mistake from my side. I will push a fix soon

MrIvv:
Hi @Naros, the changes are ready for review. I'm always available for any further updates

PR #7851, opened 2026-08-19
debezium/dbz#2232 Name serial columns as the driver does while streaming

### LINKED ISSUE #2232: PostgreSQL: snapshot reports serial/bigserial column type name while streaming reports int4/int8
## Summary

For a PostgreSQL `serial` / `smallserial` / `bigserial` column, the reported `column.type.name` for the same column can differ between the snapshot and streaming paths, and the streaming value is decoder-dependent:

- **Snapshot** (`PostgresConnection#doReadTableColumn`) takes the type name from the JDBC driver metadata, which reports `serial` / `smallserial` / `bigserial`.
- **Streaming** typically derives the name from the column OID via `PostgresType#getName()`, i.e. the underlying `int4` / `int2` / `int8`. The value is not uniform across decoders, though — `OpenLineageIT#shouldProduceOpenLineageInputDatasetUponDDLEvent` currently expects `int4` for the pgoutput decoder but `serial` for the others on a DDL-refreshed column.

A custom converter (or any downstream consumer) matching on `column.type.name` therefore sees `serial` during snapshot but `int4` during streaming for the same column.

## Why this wasn't folded into the UDT fix

This surfaced while narrowing [debezium/debezium#7651](https://github.com/debezium/debezium/pull/7651) (fix for [#683](https://github.com/debezium/dbz/issues/683), unqualified UDT type names). That fix deliberately leaves `serial` untouched, because normalizing `serial → int4` in the snapshot path has a backward-compatibility cost:

- The type name flows into `__debezium.source.column.type` via `PropagateSourceMetadataToSchemaParameter`.
- The JDBC sink's `SerialType` relies on that value to re-create `SERIAL` / `BIGSERIAL` columns (`AbstractJdbcSinkPipelineIT.testSerialDataType` / `testSmallSerialDataType` / `testBigSerialDataType`). Collapsing `serial → int4` would make a PG→PG pipeline with `column.propagate.source.type` emit plain integer columns instead of `SERIAL`.

## Decision needed

Should snapshot and streaming agree on the type name for serial columns, and in which direction?

- Snapshot → `int4` (match streaming) is more internally consistent but drops the sink's ability to re-create `SERIAL` / `BIGSERIAL`.
- Keeping/deriving `serial` on the streaming side would preserve that ability, but the information isn't readily available from the replication stream OID (and is already decoder-inconsistent today).

Filing separately so the trade-off can be weighed on its own, independent of the UDT fix in [debezium/debezium#7651](https://github.com/debezium/debezium/pull/7651).

sundongkim-dev:
Fixes debezium/dbz#2232

## Description
A `serial` column is reported as `serial` during snapshot but as `int4` once streaming with _pgoutput_ starts, because `PgOutputMessageDecoder` names the column after the OID the relation message carries. The name reaches consumers as `__debezium.source.column.type`, and the JDBC sink keys its `SERIAL` re-creation on it, so the column silently turns into `INTEGER` at the snapshot-to-streaming boundary. _decoderbufs_ is unaffected: it synchronizes the schema from the driver, which is why `OpenLineageIT` carried a decoder-dependent expectation.

Per @jpechane's call in the issue this takes option 1, aligning streaming with the snapshot, but not by flipping `shouldSchemaBeSynchronized()` for _pgoutput_. That flag alone does not fix it:

* the refresh it enables is gated on `schemaChanged()`, which compares OIDs, length, scale and optionality — a `serial` column carries the OID of its underlying `int4`, so nothing looks changed and no refresh happens;
* when a refresh does fire, `tableFromFromMessage()` rebuilds the column with `.type(type.getName())`, deriving the name from the OID again and overwriting what the driver reported.

`handleRelationMessage()` already queries the driver for every relation message and already harvests defaults and nullability from that result — the driver's type name is in the same rows and was simply dropped. Keeping it reaches option 1's outcome with no extra round trip and no `nextval(` heuristic of our own. `TypeRegistry` already aliases `serial`, `smallserial` and `bigserial` to their integer types, so type resolution and value conversion are untouched. A column that the driver metadata does not cover, such as one excluded by `column.exclude.list`, keeps falling back to the OID-derived name, exactly as today.

Verified on both decoders locally, including a negative control: reverting the single `.type(...)` line makes the new IT fail with `"INT4" (expected: "SERIAL")`.

## PR Checklist

### DIFF
--- a/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/ColumnMetaData.java
+++ b/debezium-connector-postgres/src/main/java/io/debezium/connector/postgresql/connection/pgoutput/ColumnMetaData.java
@@ -27,6 +27,7 @@ public class ColumnMetaData {
     private final int length;
     private final int scale;
     private final String typeName;
+    private final String driverTypeName;
 
     /**
      * Create a metadata structure representing a column.
@@ -38,9 +39,10 @@ public class ColumnMetaData {
      * @param hasDefaultValue {@code true} if the column has a default value specified, {@code false} otherwise
      * @param defaultValueExpression the parsed default value literal for the column
      * @param typeModifier the attribute type modifier
+     * @param driverTypeName the type name the JDBC driver reports for the column, {@code null} when unavailable
      */
     ColumnMetaData(String columnName, PostgresType postgresType, boolean key, boolean optional, boolean hasDefaultValue, String defaultValueExpression,
-                   int typeModifier) {
+                   int typeModifier, String driverTypeName) {
         this.columnName = columnName;
         this.postgresType = postgresType;
         this.key = key;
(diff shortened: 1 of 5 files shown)


Naros (approved):

jpechane:
@sundongkim-dev Applied, thanks!

PR #7852, opened 2026-08-19
debezium/dbz#2460 Acknowledge source offsets from the previous flush snapshot

### LINKED ISSUE #2460: Source offsets acknowledged to the database can run ahead of the persisted offsets, breaking recovery after a crash
After a hard crash (SIGKILL, OOM), a Postgres connector can permanently fail to restart with:

    The connector is trying to read change stream starting at ... but this is no longer available on the
    server. Reconfigure the connector to use a snapshot mode when needed.

even though the replication slot still exists and was never recreated: the stored offset LSN sits slightly
below the slot's `restart_lsn`, so the WAL it points to has been recycled.

We hit this on three connectors sharing one PostgreSQL instance, which all crashed hard within the same
weekend: in each case the persisted offset was between a few hundred KB and ~55MB behind the slot's
`restart_lsn`, and its timestamp matched the last successful offset flush before the crash.

Root cause: the runtime side is correctly ordered (both `AsyncEmbeddedEngine` and the Kafka Connect worker
call `SourceTask#commit()` only after a successful offset flush), but `BaseSourceTask#performCommit()` then
acknowledges `lastOffsets`, the most recent in-memory offsets recorded by `poll()`. By the time the
notification arrives, those can be up to a full offset flush interval ahead of what was durably persisted.
The acknowledgment flows into `StreamingChangeEventSource#commitOffset`, whose contract is to "indicate how
far the source TX log can be discarded": the connector therefore authorizes discarding WAL it may still
need. On Postgres this maps to `setFlushedLSN` and the standby status update, whose field is defined by the
protocol as the position "flushed to disk" on the client, which an in-memory position is not.

In steady state the mismatch is invisible: on a clean restart the WAL below the acknowledged position is
still retained. The failure needs a hard crash inside the window where the slot's `restart_lsn` has already
moved past the persisted offset; with the default 60s flush interval that window reopens on every cycle.
The same applies to any offset store (file, Kafka, custom) and to every connector implementing
`commitOffset` (Postgres, Oracle, SQL Server); connectors with the default no-op are unaffected.

The direction of a fix follows from the asymmetry of the two errors: under-acknowledging costs at most one
extra flush interval of source-side retention and heals on the next cycle, while over-acknowledging can
lose data irrecoverably. Proposal: acknowledge, at each `commit()` notification, the offsets captured at
the previous notification. Everything captured then is provably covered by the flush that triggered the
current notification, so the acknowledged position can never run ahead of the store. The cost is a source
acknowledgment lagging one flush interval behind.

Validated with an invariant probe on a live incremental snapshot over a 17.5M-row table: with the fix, the
slot's `confirmed_flush_lsn` never exceeded the persisted offset across the whole run (before the fix the
violation reappeared on every flush cycle), and a SIGKILL mid-snapshot was followed by a clean restart and
an exact resume.

MrIvv:
Closes https://github.com/debezium/dbz/issues/2460

`BaseSourceTask#performCommit()` used to acknowledge `lastOffsets`, the most recent in-memory offsets from
`poll()`, which can be up to a full offset flush interval ahead of what was durably persisted when the
`commit()` notification arrives. Since the acknowledgment authorizes the source database to discard its TX
log (`StreamingChangeEventSource#commitOffset`), a hard crash inside that window leaves the persisted
offset pointing at WAL the server has already recycled, and the connector cannot resume.

This keeps two generations of offsets: each `commit()` notification acknowledges the snapshot captured at
the previous notification, which is provably covered by the flush that triggered the current one, so the
acknowledged position can never run ahead of the offset store.

Notes:

- The source acknowledgment now lags one flush interval behind; the retention cost is bounded and the lag
  is absorbed on the next cycle. The first notification acknowledges nothing.
- On a graceful shutdown the last generation is not acknowledged; the source retains up to one extra flush
  interval of log, which the restarting connector replays from its (consistent) offsets.
- The debug log statement in `performCommit` had its arguments swapped with respect to the message text;
  fixed in passing since the line was touched anyway.
- A unit test in `BaseSourceTaskTest` covers the contract: the first notification acknowledges nothing, the
  following ones acknowledge exactly the previous generation.

Validated on a live incremental snapshot (17.5M rows) with an invariant probe comparing the slot's
`confirmed_flush_lsn` with the persisted offset on every cycle: no violation across the run, clean restart
and exact resume after a SIGKILL mid-snapshot.

## PR Checklist

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/connector/common/BaseSourceTask.java
+++ b/debezium-connector-common/src/main/java/io/debezium/connector/common/BaseSourceTask.java
@@ -13,7 +13,6 @@ import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
-import java.util.Iterator;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
@@ -200,6 +199,15 @@ public abstract class BaseSourceTask<P extends Partition, O extends OffsetContex
      * (which may be a no-op depending on the connector).
      */
     private final Map<Map<String, ?>, Map<String, ?>> lastOffsets = new HashMap<>();
+
+    /**
+     * Offsets captured at the previous flush notification. Everything in here was already covered by the
+     * offset flush that triggered the current {@link #commit()}, so acknowledging these positions with the
+     * source database can never run ahead of the durably persisted offsets. Acknowledging
+     * {@link #lastOffsets} directly would confirm positions up to a flush interval ahead of the store and,
+     * after a hard crash, the persisted offset could fall below the WAL retained by the source.
+     */
+    private final Map<Map<String, ?>, Map<String, ?>> flushedOffsets = new HashMap<>();
 
     private Duration retriableRestartWait;
 
@@ -583,15 +591,13 @@ public abstract class BaseSourceTask<P extends Partition, O extends OffsetContex
         if (locked) {
             try (var rootLoggingContext = LoggingContext.initRootContext()) {
                 if (coordinator != null) {
-                    Iterator<Map<String, ?>> iterator = lastOffsets.keySet().iterator();
-                    while (iterator.hasNext()) {
-                        Map<String, ?> partition = iterator.next();
-                        Map<String, ?> lastOffset = lastOffsets.get(partition);
-
-                        LOGGER.debug("Committing offset '{}' for partition '{}'", partition, lastOffset);
-                        coordinator.commitOffset(partition, lastOffset);
-                        iterator.remove();
+                    for (Map.Entry<Map<String, ?>, Map<String, ?>> entry : flushedOffsets.entrySet()) {
+                        LOGGER.debug("Committing offset '{}' for partition '{}'", entry.getValue(), entry.getKey());
+                        coordinator.commitOffset(entry.getKey(), entry.getValue());
                     }
+                    flushedOffsets.clear();
+                    flushedOffsets.putAll(lastOffsets);
+                    lastOffsets.clear();
                 }
             }
             finally {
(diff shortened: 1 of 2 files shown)


Naros (approved):
These changes seem to LGTM, and what I found nice is that the XStream tests for some of our more problematic environments passed, so I do wonder if this was perhaps the issue with XStream incremental snapshots. I would appreciate @vjuranek giving me an extra review to be sure.

debezium-connector-common/src/main/java/io/debezium/connector/common/BaseSourceTask.java
```diff
@@ -583,15 +591,13 @@ public void performCommit() {
...
                 if (coordinator != null) {
-                    Iterator<Map<String, ?>> iterator = lastOffsets.keySet().iterator();
-                    while (iterator.hasNext()) {
-                        Map<String, ?> partition = iterator.next();
-                        Map<String, ?> lastOffset = lastOffsets.get(partition);
-
-                        LOGGER.debug("Committing offset '{}' for partition '{}'", partition, lastOffset);
-                        coordinator.commitOffset(partition, lastOffset);
-                        iterator.remove();
+                    for (Map.Entry<Map<String, ?>, Map<String, ?>> entry : flushedOffsets.entrySet()) {
```
vjuranek, line 594: Would that work for Postgres and Oracle which override `performCommit()` method?

vjuranek (commented):
The fix itself LGTM, but I'm not sure if it actually fix the issue also for Postgres.

debezium-connector-common/src/main/java/io/debezium/connector/common/BaseSourceTask.java
MrIvv: For Postgres this PR generalizes that guarantee to every connector still on the base implementation, so the two are consistent rather than conflicting. The Oracle override only diverges for the XStream adapter, with LogMiner it calls super.performCommit() and picks up the fixed base. 
What do you think @vjuranek?

vjuranek: As in the issue you explicitly mention that it's related to Postgres, I'm just wondering, how this fixes the issue as Postgres won't use this patch as it overrides this method. (As I wrote above, the change itself LGTM)

MrIvv (commented):
so is it ok for be merged?

vjuranek:
> so is it ok for be merged?

First I'd like to clarify it if actually fixes the issue and we can close https://github.com/debezium/dbz/issues/2460 or not @MrIvv

MrIvv:
@vjuranek do you have any updates about this pr?

PR #7856, opened 2026-08-20
debezium/dbz#2464 Guard against a missing SQL state in the heartbeat error handlers

### LINKED ISSUE #2464: NullPointerException in heartbeat error handlers when the driver reports no SQL state
`BinlogHeartbeatErrorHandler.onError` (MySQL/MariaDB) and the equivalent lambda in `PostgresConnectorTask` switch on `SQLException#getSQLState()` without a null check. A `switch` on a String dereferences it via `hashCode()`, so a driver reporting no SQL state makes the handler itself throw NPE. The NPE escapes `DatabaseHeartbeatImpl.forcedBeat()` before the `LOGGER.error` that would have recorded the real cause, so the underlying heartbeat failure is lost and the task fails with a stack trace pointing only at the error handler.

Seen in production on MySQL:
```
Caused by: io.debezium.DebeziumException: Error processing binlog event
Caused by: java.lang.NullPointerException: Cannot invoke "String.hashCode()" because "<local3>" is null
    at io.debezium.connector.binlog.BinlogSourceTask$BinlogHeartbeatErrorHandler.onError(BinlogSourceTask.java:85)
    at io.debezium.heartbeat.DatabaseHeartbeatImpl.forcedBeat(DatabaseHeartbeatImpl.java:58)
```

suryadanny:
Fixes https://github.com/debezium/dbz/issues/2464

`BinlogHeartbeatErrorHandler.onError` (MySQL/MariaDB) and the equivalent lambda in `PostgresConnectorTask` switch on `SQLException#getSQLState()` without a null check. A `switch` on a String dereferences it via `hashCode()`, so when a driver reports no SQL state — typical of a connection-level failure rather than an error response from the server — the handler itself throws `NullPointerException`.

That NPE escapes `DatabaseHeartbeatImpl.forcedBeat()` before the `LOGGER.error` that would have recorded the underlying `SQLException`, so the original heartbeat failure is never logged and the task fails with a stack trace pointing only at the error handler.

Observed in production on MySQL:
```
Caused by: io.debezium.DebeziumException: Error processing binlog event
Caused by: java.lang.NullPointerException: Cannot invoke "String.hashCode()" because "<local3>" is null
    at io.debezium.connector.binlog.BinlogSourceTask$BinlogHeartbeatErrorHandler.onError(BinlogSourceTask.java:85)
    at io.debezium.heartbeat.DatabaseHeartbeatImpl.forcedBeat(DatabaseHeartbeatImpl.java:58)
    at io.debezium.heartbeat.HeartbeatImpl.heartbeat(HeartbeatImpl.java:71)
    at io.debezium.pipeline.EventDispatcher.dispatchHeartbeatEvent(EventDispatcher.java:438)
```

This returns early when no SQL state is reported, leaving the caller to log the exception — the same outcome an unrecognised SQL state already gets, so a transient failure is logged rather than terminating the task.

Covered by a new `BinlogSourceTaskTest`; removing the guard makes `shouldNotFailWhenSqlStateIsNotReported` fail with the NPE above.

### DIFF
--- a/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogSourceTaskTest.java
+++ b/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogSourceTaskTest.java
@@ -1,0 +1,62 @@
+/*
+ * Copyright Debezium Authors.
+ *
+ * Licensed under the Apache Software License version 2.0, available at http://www.apache.org/licenses/LICENSE-2.0
+ */
+package io.debezium.connector.binlog;
+
+import static org.assertj.core.api.Assertions.assertThatCode;
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+import java.sql.SQLException;
+
+import org.junit.jupiter.api.Test;
+
+import io.debezium.DebeziumException;
+import io.debezium.connector.binlog.BinlogSourceTask.BinlogHeartbeatErrorHandler;
+
+/**
+ * Unit tests for {@link BinlogSourceTask.BinlogHeartbeatErrorHandler}.
+ */
+public class BinlogSourceTaskTest {
+
+    private final BinlogHeartbeatErrorHandler errorHandler = new BinlogHeartbeatErrorHandler();
+
+    @Test
+    void shouldNotFailWhenSqlStateIsNotReported() {
+        // A driver is not required to populate the SQL state; a connection-level failure
+        // commonly leaves it unset. The handler must not dereference it unconditionally.
+        final SQLException exception = new SQLException("Communications link failure");
+
+        assertThatCode(() -> errorHandler.onError(exception)).doesNotThrowAnyException();
+    }
+
+    @Test
+    void shouldThrowWhenDatabaseAccessIsDenied() {
+        final SQLException exception = new SQLException("Access denied", "42000");
+
+        assertThatThrownBy(() -> errorHandler.onError(exception))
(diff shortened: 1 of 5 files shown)


debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogSourceTaskTest.java
```diff
@@ -0,0 +1,62 @@
...
+import java.sql.SQLException;
+
+import org.junit.jupiter.api.Test;
+
+import io.debezium.DebeziumException;
+import io.debezium.connector.binlog.BinlogSourceTask.BinlogHeartbeatErrorHandler;
+
+/**
+ * Unit tests for {@link BinlogSourceTask.BinlogHeartbeatErrorHandler}.
+ */
```
vjuranek, line 21: Could you add `@FixFor("dbz#2464")`

vjuranek (approved):
LGTM, just one small thing

jpechane (approved):

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogSourceTaskTest.java
suryadanny: Added @FixFor("dbz#2464") to the tests as suggested. @vjuranek

Naros:
Applied, thanks @suryadanny :tada:

PR #7862, opened 2026-08-20
debezium/dbz#2470 Propagate standalone DDL GTID events with binlog buffering

### LINKED ISSUE #2470: MariaDB EventBuffer drops standalone DDL GTID events when binlog.buffer.size is enabled
## Bug report

MariaDB connector on current `main` at `dc6df25384100e9441a3306ac79e6b2865eeebf6`.

---

The relevant settings are:

```properties
binlog.buffer.size=10000
include.schema.changes=true
```

MariaDB GTID is enabled.

---

MariaDB 11.4 in a standalone deployment.

---

A `MARIADB_GTID` event whose flag 1 is set should bypass transaction buffering because no commit event follows, but it should still be forwarded immediately to the streaming event handler.

The following standalone DDL schema change record should contain the current GTID in `source.gtid`, consistently with the behavior when `binlog.buffer.size=0`.

---

**What behavior do you see?**

When `binlog.buffer.size` is greater than zero, `EventBuffer` does not open a transaction for a flag 1 `MARIADB_GTID` event, but it also does not call `consumeEvent()` or `handleEvent()` for that event.

As a result, `MariaDbStreamingChangeEventSource.handleGtidEvent()` is not invoked and the schema change record emitted for the following DDL has a null `source.gtid`.

The flag 1 branch that drops the event is here:

https://github.com/debezium/debezium/blob/dc6df25384100e9441a3306ac79e6b2865eeebf6/debezium-connector-binlog/src/main/java/io/debezium/connector/binlog/EventBuffer.java#L110-L119

---

The behavior is present on the current `main` branch.

---

A MariaDB integration regression test reproduces the issue. Before the proposed change, it fails because `source.gtid` is null. After immediately forwarding the flag 1 GTID event, the test passes and the existing `MariaDbBinlogReaderBufferIT` suite remains green.

---

1. Start MariaDB 11.4 with binary logging and GTID enabled.
2. Start the MariaDB connector with `binlog.buffer.size=10000` and `include.schema.changes=true`.
3. Execute a standalone DDL statement:

   ```sql
   ALTER TABLE products ADD COLUMN volume FLOAT;
   ```

4. Inspect the emitted schema change record.
5. Compare its `source.gtid` with `@@GLOBAL.gtid_binlog_pos`.

With the current implementation, `source.gtid` is null. With buffering disabled, the GTID event reaches the normal handler and `source.gtid` contains the current DDL GTID.

The expected fix is to process this GTID event immediately without opening a buffered transaction.

Restart-position handling for commit-less DDL behaves the same with buffering enabled and disabled, so DDL replay after an immediate restart is not claimed as part of this bug.

Gimini-3:
Fixes debezium/dbz#2470

## Description

When MariaDB emits a standalone GTID event for a DDL statement, the GTID event has flag 1 set because no commit event follows it. With `binlog.buffer.size` enabled, `EventBuffer` correctly avoids opening a transaction for that event, but previously returned without forwarding it to the streaming event handler. As a result, the DDL schema change record could be emitted without its GTID and the GTID accumulator would not observe that position.

This change forwards standalone MariaDB GTID events directly to the event consumer while leaving regular transaction buffering unchanged.

A MariaDB integration test executes a standalone `ALTER TABLE` with buffering enabled and verifies that the schema change record contains a GTID present in `@@GLOBAL.gtid_binlog_pos`. The test fails before the fix because `source.gtid` is null.

## Testing

- `MariaDbBinlogReaderBufferIT#shouldPropagateGtidForStandaloneDdl`
- Full `MariaDbBinlogReaderBufferIT` test class: 5 tests passed
- Binlog connector build with formatting and checkstyle validation

## PR Checklist

### DIFF
--- a/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogReaderBufferIT.java
+++ b/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogReaderBufferIT.java
@@ -69,6 +69,7 @@ public abstract class BinlogReaderBufferIT<C extends SourceConnector> extends Ab
     @SkipWhenDatabaseIs(value = SkipWhenDatabaseIs.Type.MYSQL, reason = "MariaDB-specific GTID behavior")
     @SkipWhenDatabaseIs(value = SkipWhenDatabaseIs.Type.PERCONA, reason = "MariaDB-specific GTID behavior")
     @Test
+    @FixFor("debezium/dbz#2470")
     void shouldPropagateGtidForStandaloneDdl() throws SQLException, InterruptedException {
         config = DATABASE.defaultConfig()
                 .with(BinlogConnectorConfig.HOSTNAME, System.getProperty("database.replica.hostname", "localhost"))
--- a/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogReaderBufferIT.java
+++ b/debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogReaderBufferIT.java
@@ -23,6 +23,7 @@ import org.junit.jupiter.api.Test;
 
 import io.debezium.config.CommonConnectorConfig;
 import io.debezium.config.Configuration;
+import io.debezium.connector.binlog.junit.SkipWhenDatabaseIs;
 import io.debezium.connector.binlog.util.BinlogTestConnection;
 import io.debezium.connector.binlog.util.TestHelper;
 import io.debezium.connector.binlog.util.UniqueDatabase;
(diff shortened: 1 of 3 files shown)


debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogReaderBufferIT.java
```diff
@@ -65,6 +66,49 @@ void afterEach() {
         }
     }
 
+    @SkipWhenDatabaseIs(value = SkipWhenDatabaseIs.Type.MYSQL, reason = "MariaDB-specific GTID behavior")
+    @SkipWhenDatabaseIs(value = SkipWhenDatabaseIs.Type.PERCONA, reason = "MariaDB-specific GTID behavior")
+    @Test
+    void shouldPropagateGtidForStandaloneDdl() throws SQLException, InterruptedException {
```
Naros, line 73:
```suggestion
    @FixFor("debezium/dbz#2470")
    void shouldPropagateGtidForStandaloneDdl() throws SQLException, InterruptedException {
```

Naros (changes requested):
The change LGTM, just one small issue with test annotation.

debezium-connector-binlog/src/test/java/io/debezium/connector/binlog/BinlogReaderBufferIT.java
Gimini-3: Added the requested `@FixFor("debezium/dbz#2470")` annotation in `6ed2db9a7b`. The targeted `MariaDbBinlogReaderBufferIT#shouldPropagateGtidForStandaloneDdl` integration test passes on MariaDB 11.4.

Naros (approved):
LGTM, just waits on CI.

Gimini-3:
The MariaDB jobs failed in unrelated ReadOnlyIncrementalSnapshotIT timeouts, while MariaDbBinlogReaderBufferIT passed. I don't have permission to re-run the workflow. Could you please re-run the failed jobs?

vjuranek:
I re-triggered the jobs. All failed with `AbstractSnapshotTest.consumeRecordsMixedWithIncrementalSnapshot` - this test IIRC fails randomly, but I find strange that all the jobs failed here. But TBH without digging into it, I don't see any relation of the failure with this PR.

Gimini-3:
Thanks for checking and re-running the jobs!

vjuranek:
@Gimini-3 all the MariaDB tests failed with error like this in the same tests:

```
2026-08-24T10:57:14.5869664Z [ERROR]   ReadOnlyIncrementalSnapshotIT>AbstractIncrementalSnapshotWithSchemaChangesSupportTest.renameTable:117->AbstractSnapshotTest.consumeRecordsMixedWithIncrementalSnapshot:270->AbstractSnapshotTest.consumeMixedWithIncrementalSnapshot:201->AbstractSnapshotTest.consumeMixedWithIncrementalSnapshot:211->AbstractSnapshotTest.consumeMixedWithIncrementalSnapshot:230 [Too many no data record results, 101 < 1000] 
```

So it seems the failure is related to the change. Could you please take a look on it?

Gimini-3:
Thanks for catching this. The failures were caused by this PR.

Forwarding standalone GTID events through `handleGtidEvent()` correctly advanced the GTID accumulator and offset, but it also emitted a transaction-start callback. Because a standalone DDL has no following commit, read-only incremental snapshots could start the next chunk before the DDL query updated the schema cache.

I pushed a fix that keeps the GTID/offset update while skipping the transaction-start callback only when `FL_STANDALONE` is set. I also replaced the raw `0x01` check in `EventBuffer` with the named constant.

Verified locally on MariaDB 11.4 with `binlog.buffer.size=10000`:
- `MariaDbBinlogReaderBufferIT#shouldPropagateGtidForStandaloneDdl`
- `ReadOnlyIncrementalSnapshotIT#renameTable`
- `ReadOnlyIncrementalSnapshotIT#schemaChanges`

All three pass.

Gimini-3:
When you have a chance, could you please re-trigger the full CI for the latest commit? Thanks!

Naros:
Applied, thanks @Gimini-3 :tada:

PR #7863, opened 2026-08-20
debezium/dbz#2458 Skip signal processing when the semaphore cannot be acquired

### LINKED ISSUE #2458: SignalProcessor.executeWithSemaphore ignores tryAcquire result and runs the critical section unguarded on timeout
### Bug report

`SignalProcessor.executeWithSemaphore(...)` discards the boolean result of `Semaphore.tryAcquire(timeout, unit)` and runs the guarded operation unconditionally, so on a timeout the signal-processing critical section executes **without holding the permit** — defeating the mutual exclusion the semaphore exists to provide.

`debezium-connector-common/.../pipeline/signal/SignalProcessor.java` (around the `executeWithSemaphore` method, `SEMAPHORE_WAIT_TIME = 10`):

```java
private void executeWithSemaphore(Runnable operation) {
    boolean acquired = false;
    try {
        acquired = semaphore.tryAcquire(SEMAPHORE_WAIT_TIME, TimeUnit.SECONDS);
        operation.run();                       // runs even when acquired == false
    }
        LOGGER.error("Not able to acquire semaphore after {}s", SEMAPHORE_WAIT_TIME);
        throw new DebeziumException("Not able to acquire semaphore during signaling processing", e);
    }
    finally {
        if (acquired) {
            semaphore.release();
        }
    }
}
```

`tryAcquire(timeout, unit)` returns `false` on timeout — it does not throw. The `catch (InterruptedException)` only fires if the waiting thread is itself interrupted, not when the permit is unavailable for 10s. So on timeout, execution falls through to `operation.run()` regardless. (The `finally` correctly skips `release()` since `acquired == false`, so there is no permit leak — but the exclusivity was already violated.)

### Why the guard matters

The single-permit `semaphore` serializes two entry points that run on **different threads**:

- `process()` — scheduled on `signalProcessorExecutor` (`scheduleAtFixedRate(this::process, ...)`).
- `processSourceSignal(partition)` — called inline from `EventDispatcher.dispatchDataChangeEvent()` on the streaming thread ("a synchronization point to immediately execute an eventual stop signal").

So the semaphore is exactly what keeps an external-channel signal action from running concurrently with a source-channel one.

### Failure scenario

1. A signal requests a long-running action — a blocking snapshot, or an `execute-snapshot` (INCREMENTAL) that loops `readChunk()` over a large table. `process()` holds the permit and runs it; this routinely exceeds 10s.
2. Meanwhile a signal-table row arrives; the streaming thread reaches `processSourceSignal()` → `executeWithSemaphore()`.
3. `tryAcquire(10, SECONDS)` waits 10s, returns `false`, no exception.
4. `operation.run()` executes anyway, concurrently with the still-running first action.
5. Both can mutate shared `@NotThreadSafe` state (e.g. `AbstractIncrementalSnapshotChangeEventSource.context` / its `window` `LinkedHashMap`), risking corrupted chunk bookkeeping, `ConcurrentModificationException`, or a stop-signal racing a snapshot start.

### Fix

Only enter the critical section when the permit was acquired. Two reasonable behaviours on timeout:

- **Fail-loud (matches the intent already encoded in the error message):** log and `throw new DebeziumException("Not able to acquire semaphore ...")`, i.e. make that existing message actually reachable for the timeout it was written for.
- **Fail-safe:** log a warning and skip this cycle (`return`); the periodic `process()` retries next interval and the source signal is re-read on the next dispatch.

I lean fail-safe (`return`) to avoid turning transient contention into a task failure, but I'm happy to follow the maintainers' preference. One related question: `SEMAPHORE_WAIT_TIME` is a hardcoded 10s, which makes a fast, non-flaky regression test awkward — would you be open to a small test seam (making the wait injectable/package-private) so the fix can be covered by a unit test?

I have a fix ready and can open a PR once the preferred behaviour is confirmed.

TimurRakhmatullin86:
Fixes [debezium/dbz#2458](https://github.com/debezium/dbz/issues/2458).

### Problem

`SignalProcessor.executeWithSemaphore(...)` discarded the boolean result of `Semaphore.tryAcquire(timeout, unit)` and ran the guarded operation unconditionally. `tryAcquire` returns `false` on timeout (it does not throw, and the `catch` only handles `InterruptedException`), so on a timeout the signal-processing critical section executed **without holding the permit** — defeating the mutual exclusion between the two threads that enter it:

- `process()` — the `signalProcessorExecutor` thread (external channels), and
- `processSourceSignal()` — the streaming thread, reached inline from `EventDispatcher.dispatchDataChangeEvent()`.

A long-running signal action (a blocking snapshot, or an incremental `execute-snapshot` looping `readChunk()`) can exceed the 10s wait; a second action arriving meanwhile would then run concurrently and mutate shared `@NotThreadSafe` snapshot state (e.g. the incremental-snapshot context/window), risking corrupted chunk bookkeeping or `ConcurrentModificationException`.

### Fix

Per @jpechane on the issue:

- Only run the operation when the permit was actually acquired; otherwise log a warning and skip the cycle (fail-safe — the periodic `process()` retries next interval and the source signal is re-read on the next dispatch).
- Make the wait configurable via the internal `signal.processor.semaphore.wait.ms` property (default 10s, replacing the hardcoded constant) so it can be tuned from user reports and, importantly, lowered for a fast unit test.
- Restore the interrupt flag on `InterruptedException` and correct the (previously misleading) message.

### Test

`SignalProcessorTest.shouldSkipProcessingWhenSemaphoreCannotBeAcquired` holds the single permit, sets the wait to 1 ms, and asserts the signal action does not run. Verified it fails against the previous code (the action ran unguarded) and passes with the fix; the existing `SignalProcessorTest` cases stay green.

### DIFF
--- a/debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
+++ b/debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
@@ -990,6 +990,17 @@ public abstract class CommonConnectorConfig {
             .withValidation(Field::isPositiveInteger)
             .withDescription("Interval for looking for new signals in registered channels, given in milliseconds. Defaults to 5 seconds.");
 
+    public static final Field SIGNAL_PROCESSOR_SEMAPHORE_WAIT_MS = Field.createInternal("signal.processor.semaphore.wait.ms")
+            .withDisplayName("Signal processor semaphore wait time (ms)")
+            .withGroup(Field.createGroupEntry(Field.Group.ADVANCED))
+            .withType(Type.LONG)
+            .withWidth(Width.SHORT)
+            .withImportance(Importance.LOW)
+            .withDefault(10000L)
+            .withValidation(Field::isPositiveInteger)
+            .withDescription("The maximum time in milliseconds the signal processor waits to acquire its internal lock before "
+                    + "skipping a signal processing cycle. Defaults to 10 seconds.");
+
     public static final Field SIGNAL_EMIT_FAILURE_MAX_RETRIES = Field.create("signal.emit.failure.max.retries")
             .withDisplayName("Signal emit window event failure maximum retries")
             .withType(Type.INT)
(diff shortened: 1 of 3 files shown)


debezium-connector-common/src/main/java/io/debezium/config/CommonConnectorConfig.java
```diff
@@ -990,6 +990,17 @@ public static SnapshotQueryMode parse(String value, String defaultValue) {
             .withValidation(Field::isPositiveInteger)
             .withDescription("Interval for looking for new signals in registered channels, given in milliseconds. Defaults to 5 seconds.");
 
+    public static final Field SIGNAL_PROCESSOR_SEMAPHORE_WAIT_MS = Field.createInternal("signal.processor.semaphore.wait.ms")
```
mfvitale, line 993: Wondering if this could be internal for now? @jpechane

TimurRakhmatullin86: It already is — this uses `Field.createInternal(...)` (per @jpechane's suggestion on the issue), so the property stays out of the public connector configuration documentation and is only meant for tuning/testing. Happy to change it if you had something else in mind.

mfvitale (approved):

PR #7865, opened 2026-08-20
debezium/dbz#2472 Document the consumer group offset commit option

viragtripathi:
Documents `cockroachdb.changefeed.kafka.consumer.offset.commit.enabled` added by https://github.com/debezium/debezium-connector-cockroachdb/pull/82 (https://github.com/debezium/dbz/issues/2472): an opt-in, disabled by default, that mirrors the connector's consumed positions to its Kafka consumer group after each processed batch so external lag tooling reflects progress on the intermediate topics. Restart positions always come from the Debezium offsets, so the option has no effect on delivery semantics. Adds the property row to the changefeed consumer table in cockroachdb.adoc.


Naros (approved):

viragtripathi:
Connector side merged (debezium/debezium-connector-cockroachdb#82), so this is good to go whenever.

Naros:

PR #7879, opened 2026-08-22
debezium/dbz#2483 Apply snapshot.fetch.size to chunked snapshot chunk statements

minleejae:
Fixes debezium/dbz#2483

The chunked parallel initial snapshot created chunk statements with a raw
`connection().prepareStatement(...)`, so the configured `snapshot.fetch.size` was never
applied — unlike the legacy snapshot path (`JdbcConnection#readTableStatement`) and
incremental snapshot chunks (`JdbcConnection#readTablePreparedStatement`). On large
tables this leads to client-side `OutOfMemoryError` (pgJDBC materializes the whole
result set with its default fetch size 0; MySQL's streaming cursor handling is bypassed).

Route the chunk statement through the existing `JdbcConnection#readTablePreparedStatement`,
as incremental snapshot chunks already do.

Verified with a new unit test that fails before the fix (`setFetchSize` never invoked on
the chunk statement) and passes after. The existing chunked snapshot IT suites pass
locally against real databases: `PostgresChunkedSnapshotIT` 11/11 (PostgreSQL 17),
`MySqlChunkedSnapshotIT` 10/10 — these now exercise the applied default fetch sizes,
including MySQL's `Integer.MIN_VALUE` streaming fetch size.


Naros (approved):

Naros:

PR #7880, opened 2026-08-22
debezium/dbz#1201 Mask passwords in task configuration debug logging

### LINKED ISSUE #1201: Debezium server logging doesn't mask password [DBZ-8343]
Migrated from [DBZ-8343](https://issues.redhat.com/browse/DBZ-8343)

During the start, Debezium server prints engine configuration which includes sensitive data like database password. This data are printed into the log in plain text format. We need to mask the passwords as we do in other parts of Debezium.

The problematic logging is [here|https://github.com/debezium/debezium-server/blob/v3.0.0.Final/debezium-server-core/src/main/java/io/debezium/server/DebeziumServer.java#L168]

minleejae:
Fixes debezium/dbz#1201

The code originally referenced in the issue (`DebeziumServer.java`) was removed when the
server bootstrap moved to the Quarkus extensions, but the same exposure remains in the
shared engine: `AsyncEmbeddedEngine` logs raw task configurations — including
`database.password` — at DEBUG level when creating source tasks.

Mask the values via `Configuration#withMaskedPasswords()`, as `BaseSourceTask` already
does for its startup configuration logging. Verified with a unit test that fails before
the fix (plaintext password present in the log output) and passes after.


Naros (approved):

Naros:
Thanks @minleejae applied, :tada:

